PR #39459, opened 2021-11-03
[8.x] Make view errors shareable globally

bilfeldt:
This PR makes it possible to add errors to a view

```php
view('post-index')
    ->withErrors(['field' => 'Some error'], 'default', true); // only the third parameter is new
```

where the errors are now globally available meaning that not just `post-index.blade.php` has access to the `$errors` but also parent/child views like for example the main layout `resources/views/layouts/app.blade.php`. Note that this global `$errors` have already been made available using the existing middleware `ShareErrorsFromSession`.

This is aligned with the way that the redirect method of same name (`withErrors`) makes errors globally available on the redirected request: https://github.com/laravel/framework/blob/master/src/Illuminate/Http/RedirectResponse.php#L131

## Why

This is great if you would like to render errors from the `default` error bag as part of you main layout file

## Considerations

Personally I think the **default** behaviour should be that any errors added via `viewErrors()` were made globally available no matter if called on a view or a redirect. Usually if someone would like to target some errors for a specific view it would make sense to use a **named bag** like:

```php
view('post-index')
    ->withErrors(['field' => 'Some error'], 'port-index-bag'); // here I am assuming that errors are globally accepted by default
```

and if this is strictly not wanted, then simply set the `$share = false` or use the `view(..)->with()` method instead.
 
**But** sharing these errors globally to all views could probably be considered a breaking change - maybe not if we simply **added** them to any existing errors in the global error bag.

Questions for the editors:

1. Should the default be globally accessable?
2. Should errors be **added** (pushed) to any existing error bag or simply override any bag that already exist (current behaviour)?

### DIFF
--- a/src/Illuminate/View/View.php
+++ b/src/Illuminate/View/View.php
@@ -207,10 +207,20 @@ class View implements ArrayAccess, Htmlable, ViewContract
      *
      * @param  \Illuminate\Contracts\Support\MessageProvider|array  $provider
      * @param  string  $bag
+     * @param  bool  $share
      * @return $this
      */
-    public function withErrors($provider, $bag = 'default')
+    public function withErrors($provider, $bag = 'default', $share = false)
     {
+        if ($share) { // Add the errors to the globally shared $errors
+            $this->factory->share('errors', (new ViewErrorBag)->put(
+                $bag, $this->formatErrors($provider)
+            ));
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
No plans to change the behavior of errors, etc. at the moment.

PR #39461, opened 2021-11-03
[8.x] Added varbinary schema type

pintend:
Adds

```php
$table->varbinary('column_name', 255);
```

I was dealing with a legacy project and I was using https://github.com/bennett-treptow/laravel-migration-generator to generate migrations so laravel can be in sync with the legacy applications database migrations, which is when i noticed varbinary is not supported by laravel


taylorotwell:
What about the other grammars?

PR #39462, opened 2021-11-03
[8.x] Improvement for Log::build()

Air-Petr:
This PR is connected with "On-Demand Channels" (https://laravel.com/docs/8.x/logging#on-demand-channels).
"On-Demand Channels" have been introduced by https://github.com/laravel/framework/pull/39273

## Problem

Every call of a Log::build(...) uses the same log channel. Example (can be executed in tinker):

```

// ... some code

Log::build([
  'driver' => 'single',
  'path' => storage_path('foo/custom.log'),
])->info('foo');

// Now we have a "foo" record in a "foo" directory

// ... some code

Log::build([
  'driver' => 'single',
  'path' => storage_path('bar/custom.log'),
])->info('bar');

// New "bar" record, but still in a "foo" directory

```

## Case

We have a queue running. Some jobs need certain channels (based on the names from a database). We've tried to use "On-Demand Channels". And now all information goes to one channel (the one is used by a first job appeared in a queue).

## Possible solution

We can extend a "build" method by adding second optional parameter - the name of a channel.
For me it's hard to say is it good or bad from architectural perspective. But it certainly can be useful in some situations.
I've prepared a code change and a test method. 
I've run tests by calling:
```
./vendor/bin/phpunit --filter=LogManagerTest
```

### DIFF
--- a/src/Illuminate/Log/LogManager.php
+++ b/src/Illuminate/Log/LogManager.php
@@ -65,11 +65,12 @@ class LogManager implements LoggerInterface
      * Build an on-demand log channel.
      *
      * @param  array  $config
+     * @param  string|null  $name
      * @return \Psr\Log\LoggerInterface
      */
-    public function build(array $config)
+    public function build(array $config, ?string $name = null)
     {
-        return $this->get('ondemand', $config);
+        return $this->get($name ?? 'ondemand', $config);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


driesvints:
There's breaking changes here so please send this to master.

Air-Petr:
@driesvints 
Does an optional parameter considered as breaking change? It won't hurt existing functionality.

driesvints:
Changing a method signature is a breaking change unfortunately.

dennisprudlo:
@Air-Petr classes that extend `LogManager` and overwrite the `build` method will break, because the signature does not match anymore. That's what is Breaking, unfortunately.

PR #39475, opened 2021-11-04
Resolve an issue with ->load or ->fresh on models with different connections defaulting to parent connection when $connection is not specified

CountTo25:
This is a followup to this post. Since i'm dumb, i wanted to avoid creating PR because there's most likely better way of fixing this. In the end though i decided to create one anyway.

Here's description of an issue with examples:
https://github.com/laravel/framework/discussions/39262

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1626,7 +1626,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
      */
     public function getConnectionName()
     {
-        return $this->connection;
+        return $this->connection ?? config('database.default');
     }
 
     /**


taylorotwell:
Hey there - we can't call the global configuration helper from the Eloquent model. It's also a bit scary to make this change on a patch release. 

For now I recommend calling `setConnection` on your model explicitly if you would like to tell it what connection to use.

CountTo25:
> Hey there - we can't call the global configuration helper from the Eloquent model. It's also a bit scary to make this change on a patch release.
> 
> For now I recommend calling `setConnection` on your model explicitly if you would like to tell it what connection to use.

Yeah, no problem there, as i told in the message, im pretty dumb, no surprises it getting rolled back.

While i surely can fix that myself just by specifying $connection on my default models, is there any possible out of the box solution to this in the future?

PR #39415, opened 2021-10-29
[8.x] MySQL Builds

driesvints:
This PR adds a separate build for MySQL that runs our database integration tests against MySQL 5.7 & MySQL 8.0. What I did was add separate database builds that run our database integration tests to validate that all functionality works in every engine. I've already discovered several places in our tests which were set up incorrectly or validating data incorrectly. I've added comments to each part that got changed to explain why that happened. Feel free to "Resolve conversation" those that are clear and add comments to the ones which you have questions about. 

A few follow up PR's that I'd like to make after this is to consolidate how the environment gets set up and data gets refreshed as well as to look in the unit test suite which tests might be able to move to the integration test suite.

Builds for PostgreSQL, MSSQL & MariaDB will be added after this gets merged to keep the scope of this PR small.

Issue #27778, opened 2019-03-05
SQLSRV driver does not support ID inserts


laurencei:
- Laravel Version: 5.8.0
- PHP Version: 7.2
- Database Driver & Version: MSSQL 2012 with SQLSRV driver

### Description:

As part of an enterprise project, my client needs to use MSSQL. Trying to insert a seed with a specific id gives an error:

    DB::table('test_table')->insert(['id' => 1, 'name' => 'example']);

    SQLSTATE[23000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]Cannot insert explicit value for identity column in table 'test_table' when IDENTITY_INSERT is set to OFF. (SQL: insert into [test_table] ([id], [name]) values (1, example))

This is expected behavior, as MSSQL works a bit differently to most other databases, and does not allow IDs to be inserted by default. The method to solve is to turn `IDENTITY_INSERT` to `ON` for the table, insert the record, then turn it `OFF` again:

    DB::statement('SET IDENTITY_INSERT test_table ON');
    DB::table('test_table')->insert(['id' => 1, 'name' => 'example']);
    DB::statement('SET IDENTITY_INSERT test_table OFF');

Except this doesnt work, and MSSQL still thinks the `IDENTITY_INSERT` is `OFF`. If you run the above as native SQL queries - it works, so the problem is how Laravel is handling it.

Technically, we should be able to wrap the whole thing in a transaction, but that doesnt work either:

    DB::beginTransaction();
    DB::statement('SET IDENTITY_INSERT test_table ON');
    DB::table('test_table')->insert(['id' => 1, 'name' => 'example']);
    DB::statement('SET IDENTITY_INSERT test_table OFF');
    DB::commit();

Googling around, I found this [SO thread](https://stackoverflow.com/q/34622395/1317935) which people have the same issue. There is a manual workaround, but the answer given there is not practical for me or suitable for long term use.

I only really need the `id` inserting for `seeding` and a few edge cases, - but this seems like a bug to me?

laurencei:
I'm wondering if we need to create some extra commands in the SQL driver - and do something like:

```
DB::beginTransaction();
DB::setIdentityInsertOn();
DB::table('test_table')->insert(['id' => 1, 'name' => 'example']);
DB::setIdentityInsertOff();
DB::commit();
```

Or even better:

   DB::table('test_table')->withIdentityInsert()->insert(['id' => 1, 'name' => 'example']);

[...]

PR #39477, opened 2021-11-04
[8.x] PostgreSQL, MSSQL & MariaDB builds

driesvints:
Continuation of https://github.com/laravel/framework/pull/39415 for more engines.

### DIFF
--- a/tests/Integration/Database/EloquentBelongsToManyTest.php
+++ b/tests/Integration/Database/EloquentBelongsToManyTest.php
@@ -204,6 +204,7 @@ class EloquentBelongsToManyTest extends DatabaseTestCase
         );
     }
 
+    /** @group SkipMSSQL */
     public function testCustomPivotClassUpdatesTimestamps()
     {
         Carbon::setTestNow('2017-10-10 10:10:10');
@@ -645,6 +646,7 @@ class EloquentBelongsToManyTest extends DatabaseTestCase
         $this->assertNotSame('2017-10-10 10:10:10', Tag::find(300)->updated_at);
     }
 
+    /** @group SkipMSSQL */
     public function testWherePivotOnString()
     {
         $tag = Tag::create(['name' => Str::random()]);
@@ -693,6 +695,7 @@ class EloquentBelongsToManyTest extends DatabaseTestCase
         $this->assertEquals($relationTag->getAttributes(), $tag->getAttributes());
     }
 
+    /** @group SkipMSSQL */
     public function testWherePivotInMethod()
     {
         $tag = Tag::create(['name' => Str::random()]);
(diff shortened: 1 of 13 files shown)


tests/Integration/Database/EloquentBelongsToManyTest.php
```diff
@@ -48,7 +48,8 @@ protected function setUp(): void
 
         Schema::create('posts_tags', function (Blueprint $table) {
             $table->integer('post_id');
-            $table->string('tag_id');
+            $table->integer('tag_id')->default(0);
+            $table->string('tag_name')->default('')->nullable();
```
driesvints, line 52: Using an integer for the related custom key leads to errors in PostgreSQL. Since the purpose for the related test is a custom related string key we should also make sure the pivot column has the correct type. There for we need to add one explicitly.

tests/Integration/Database/EloquentBelongsToManyTest.php
```diff
@@ -402,8 +403,8 @@ public function testFindOrNewMethod()
 
         $this->assertEquals($tag->id, $post->tags()->findOrNew($tag->id)->id);
 
-        $this->assertNull($post->tags()->findOrNew('asd')->id);
-        $this->assertInstanceOf(Tag::class, $post->tags()->findOrNew('asd'));
+        $this->assertNull($post->tags()->findOrNew(666)->id);
```
driesvints, line 406: PostgreSQL specifically fails here if you provide an incorrect type. Since the purpose is just to provide an invalid error, we can simply provide an invalid integer ID.

tests/Integration/Database/EloquentModelRefreshTest.php
```diff
@@ -59,6 +59,7 @@ public function testItSyncsOriginalOnRefresh()
     public function testAsPivot()
     {
         Schema::create('post_posts', function (Blueprint $table) {
+            $table->increments('id');
```
driesvints, line 62: PostgreSQL requires an incrementing ID for this table.

tests/Integration/Database/EloquentBelongsToManyTest.php
```diff
@@ -598,6 +600,7 @@ public function testNoTouchingHappensIfNotConfigured()
         $this->assertNotSame('2017-10-10 10:10:10', $tag->fresh()->updated_at->toDateTimeString());
     }
 
+    /** @group SkipMSSQL */
```
driesvints, line 603: We're explicitly skipping some tests on MSSQL because of the following issue: https://github.com/laravel/framework/issues/27778

At a later point we should refactor these maybe to not insert incremental identifiers.

tests/Integration/Database/EloquentWhereHasMorphTest.php
```diff
@@ -135,6 +135,10 @@ public function testWhereHasMorphWithOwnerKey()
             $table->string('slug')->nullable();
         });
 
+        Schema::table('comments', function (Blueprint $table) {
+            $table->dropIndex('comments_commentable_type_commentable_id_index');
+        });
```
driesvints, line 140: MSSQL was nagging about the index when changing the related column so let's drop it explicitly.

PR #39483, opened 2021-11-04
[8.x] Adds dynamic middleware set on route declaration

DarkGhostHunter:
## What?

Based on #39464. Allows dynamically setting middleware in route declarations, including parameters, fluently.

```php
// Before
Route::get('flights', [FlightController::class , 'show'])
    ->middleware(['can:viewAny:\App\Models\Flight', 'auth:pilot']);

// After
Route::get('flights', [FlightController::class , 'show'])
    ->can('viewAny', Flight::class)
    ->auth('pilot');
```

## How?

When passing a method that doesn't exist or is not defined as a macro when declaring a route, the router will check if the method name is set as [route middleware](https://github.com/laravel/laravel/blob/5d22b2464c9d64443ecf60d5788d753430eff7fa/app/Http/Kernel.php#L49-L66). If that's the case, it will set is as a middleware for the given route.

> If these conditions don't met, an `BadMethodCallException` will be thrown. This avoids the developer to push any typo accidentally, where `middleware()` would, pushing the error when the route is called. For this to work, an `app(Kernel::class)` was implemented to check the registered array.

It also works on dot notation middleware, like `password.confirm`, with parameters:

```php
// Before
Route::get('flights', [FlightController::class , 'show'])
    ->middleware('password.confirm:auth.confirm');

// After
Route::get('flights', [FlightController::class , 'show'])
    ->passwordConfirm('auth.confirm');
```

It will also work when:

- using the `Router` facade directly:

```php
Route::passwordConfirm()->get('flights', ...);
```

- using a route group

```php
Route::auth()->group(fn() => ... );
```

- using a resource

```php
Route::resource('flight')->authBasic();
```

## BC?

1. No, as is expected route declarations using the available methods.
2. No, there are no method collisions when cross-checking the [default route middleware](https://github.com/laravel/laravel/blob/5d22b2464c9d64443ecf60d5788d753430eff7fa/app/Http/Kernel.php#L49-L66).
3. No. If there is a macro with the same name, macro takes precedence.

## Notes

Because it uses `app(Kernel::class)->getRouteMiddleware()` behind the scenes to check if the middleware exists or a `BadMethodCallException` should be thrown, some tests must be rewritten to add a container instance.

### DIFF
--- a/src/Illuminate/Routing/ForwardsCallToMiddleware.php
+++ b/src/Illuminate/Routing/ForwardsCallToMiddleware.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace Illuminate\Routing;
+
+use Illuminate\Foundation\Http\Kernel;
+
+trait ForwardsCallToMiddleware
+{
+    /**
+     * Sets the middleware in the route if set in the HTTP Kernel.
+     *
+     * @param  string  $middleware
+     * @param  array  $parameters
+     * @return static|void
+     */
+    protected function forwardToMiddleware($middleware, $parameters)
+    {
+        return $this->middleware($this->parseMiddlewareFromMethod($middleware, $parameters));
+    }
+
+    /**
+     * Transforms method call to a middleware declaration.
+     *
+     * @param  string  $middleware
+     * @param  array  $parameters
+     * @return string
+     */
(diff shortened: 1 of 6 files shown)


taylorotwell:

PR #39497, opened 2021-11-05
[8.x] Fifo support for queue name suffix

ArronKing:
PR to add support for fifo naming conventions.

A FIFO SQS queue must always end with `.fifo`. For example, you set a queue name of `emails.fifo` and a suffix of `-staging` you end up with a queue name of `emails.fifo-staging`. After this PR you get `emails-staging.fifo`.

I appriciate FIFO queues aren't supported by the Framework. But there are packages that add support. This PR just adds support for the naming convention.

Thanks

### DIFF
--- a/src/Illuminate/Queue/SqsQueue.php
+++ b/src/Illuminate/Queue/SqsQueue.php
@@ -188,8 +188,22 @@ class SqsQueue extends Queue implements QueueContract, ClearableQueue
         $queue = $queue ?: $this->default;
 
         return filter_var($queue, FILTER_VALIDATE_URL) === false
-            ? rtrim($this->prefix, '/').'/'.Str::finish($queue, $this->suffix)
+            ? $this->suffixQueue($queue, $this->suffix)
             : $queue;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
I don't think just jamming fifo on a queue is enough. Laravel doesn't understand group ids.

ArronKing:
Yeah I agree, but I'm not trying to add FIFO support. This focused on fixing(?) the naming of fifo queues when using a suffix.

Using a suffix of `-production` and a FIFO SQS queue with the name `emails.fifo`. We should end up with the name `emails-production.fifo` not `emails.fifo-production`?

If it helps, my use case is I'm using https://github.com/shiftonelabs/laravel-sqs-fifo-queue to add Fifo support. But would still like to use some Laravel queue features, like `artisan queue:work` and `artisan queue:monitor` etc.  These don't really care if it's a standard queue or fifo queue, but in order for them to work we need the suffix to be in the correct place.

Thanks again

taylorotwell:
Thanks for explaining. Will look it over soon.

Issue #39452, opened 2021-11-02
Event's `withoutOverlapping` mutex is not released if another before/after callback fails


inxilpro:
- Laravel Version: 8.68.1
- PHP Version: 7.4.21
- Database Driver & Version: N/A

### Description:

When you call [`Event::withoutOverlapping()`](https://github.com/laravel/framework/blob/de0071debe5ea7362fd5f2c99a0bf1f24f8106e5/src/Illuminate/Console/Scheduling/Event.php#L644) it sets up an "after" callback that removes the event mutex:

```php
$this->then(function () {
  $this->mutex->forget($this);
});
```

If any other before or after callback fails before the mutex is cleared, this callback won't be triggered.

### Steps To Reproduce:

```php
$schedule->command('inspire')
  ->everyMinute()
  ->withoutOverlapping()
  ->before(function() {
    throw new \Exception('Something unexpected happened.');
  });
```

### Potential Solutions:

There are three possible solutions that I can think of:

1. Run each "before" and "after" callback in a `try`/`catch` to ensure that they all run. This would fix the problem, but also potentially break code that expects previous callbacks to have run.
2. Add special handling for the "withoutOverlapping" case, maybe by just checking the `$withoutOverlapping` value after [`Event::run()`](https://github.com/laravel/framework/blob/d37d032832717737f99fed3b454d8ca4666b0822/src/Illuminate/Console/Scheduling/Event.php#L191) has executed, and clearing the mutex there.
3. Introduce a new kind of callback… something like `Event::alwaysBefore()` and `Event::alwaysAfter()` that is executed separately from `Event::before()` and `Event::after()` and is run within a `try`/`catch` block to ensure that they always run no matter what. We could then move the call to `$mutex->forget()` into `alwaysAfter()` to ensure the mutex is cleared regardless of what exceptions were thrown.

I'm happy to PR a solution, but I wanted some feedback before I do.

### Additional Context:

This came up because we have a `Log::info()` call inside a `before()` callback, and our logging service had downtime causing an exception to be thrown. We've added `try`/`catch` statements to our application code and added an exception handler to that logger, but it was a beast to debug. I ended up having to manually find and clear the mutex value in our Redis cache, which was *not* fun!

driesvints:
Paging @taylorotwell to see what the best solution to this would be.

taylorotwell:
@inxilpro I would probably go for solution #2... just handle it totally separately from the `then` callbacks. Note that some commands run in the foreground and some run in the background.

PR #39498, opened 2021-11-05
[8.x] Ensure event mutex is always removed

inxilpro:
This is a fix for #39452

- Added a new `removeMutex` method (which already exists on `CallbackEvent`) that only clears the mutex if necessary
- For commands running foreground, we wrap everything in a `try`/`finally` to call `removeMutex`
- For commands running the the background, we wrap the before callbacks in a `try`/`catch` that re-throws any exceptions after first calling `removeMutex`. Then, we call `removeMutex` in `callAfterCallbacksWithExitCode` after the background job completes.

### DIFF
--- a/src/Illuminate/Console/Scheduling/Event.php
+++ b/src/Illuminate/Console/Scheduling/Event.php
@@ -228,7 +228,7 @@ class Event
 
             $this->callAfterCallbacks($container);
         } finally {
-            $this->deleteMutex();
+            $this->removeMutex();
         }
     }
 
@@ -245,7 +245,7 @@ class Event
 
             Process::fromShellCommandline($this->buildCommand(), base_path(), null, null, null)->run();
         } catch (Throwable $exception) {
-            $this->deleteMutex();
+            $this->removeMutex();
 
             throw $exception;
         }
@@ -291,7 +291,7 @@ class Event
         try {
             $this->callAfterCallbacks($container);
         } finally {
-            $this->deleteMutex();
+            $this->removeMutex();
         }
     }
 
@@ -936,7 +936,7 @@ class Event
      *
      * @return void
      */
-    protected function deleteMutex()
+    protected function removeMutex()
     {
         if ($this->withoutOverlapping) {
             $this->mutex->forget($this);
--- a/src/Illuminate/Console/Scheduling/Event.php
+++ b/src/Illuminate/Console/Scheduling/Event.php
@@ -228,7 +228,7 @@ class Event
 
             $this->callAfterCallbacks($container);
         } finally {
-            $this->removeMutex();
+            $this->deleteMutex();
         }
     }
 
@@ -245,7 +245,7 @@ class Event
 
             Process::fromShellCommandline($this->buildCommand(), base_path(), null, null, null)->run();
         } catch (Throwable $exception) {
-            $this->removeMutex();
+            $this->deleteMutex();
 
             throw $exception;
         }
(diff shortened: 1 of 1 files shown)


taylorotwell:
The `withoutOverlapping` command still releases the mutex. Could this not create a race condition where the `withoutOverlapping` callback releases the mutex, then a new scheduled command grabs a new mutex, but then the `removeMutex` method is called and deletes the newly obtained mutex for the other command?

inxilpro:
> Could this not create a race condition

Oh shoot. Good point. I decided to leave that in there to change as few things possible—I have a `9.x` version of the PR that I'm going to push up that cleans things up a little more but adds a few new methods to `Event` that could be theoretically backwards-incompatible. But yeah, I'll remove the `after()` callback from this, too.

src/Illuminate/Console/Scheduling/Event.php
```diff
@@ -915,4 +930,16 @@ public function preventOverlapsUsing(EventMutex $mutex)
 
         return $this;
     }
+
+    /**
+     * Delete the mutex for the event.
+     *
+     * @return void
+     */
+    protected function deleteMutex()
```
inxilpro, line 939: @taylorotwell The reason I called this `removeMutex` is because that's what it's already called in `CallbackEvent`. It probably either needs to be renamed in both places or left as `removeMutex`.

https://github.com/laravel/framework/blob/10483fbd04643135c242850c20467e3c182dce17/src/Illuminate/Console/Scheduling/CallbackEvent.php#L101-L106

taylorotwell: OK we rename it back to removeMutex

inxilpro:
@taylorotwell It kinda looks like `CallbackEvent` has the same issue:

https://github.com/laravel/framework/blob/cb46639988f817be01c81056351361e699897438/src/Illuminate/Console/Scheduling/CallbackEvent.php#L67-L73

I'm having trouble understanding why `CallbackEvent` handles mutex's so differently than normal `Event` objects. It seems to me that the `register_shutdown_function` call either should exist on both objects or should be removed. I'm working on a `9.x` version of this that includes a little refactoring, and I'd like to bring the implementations of the two closer together.  Do you have any more context about why they're so different right now?

taylorotwell:
I don't know why but I would be very cautious changing things we don't fully understand. ALWAYS leads to breaking changes.

taylorotwell:
@inxilpro, so there is some remaining issue that prevents this particular PR from being merged?

inxilpro:
I think that this can be merged as-is. If there's an issue with `register_shutdown_function` it was already there. It's a pretty edge-case race condition, so I'm not surprised it hasn't come up.

But `register_shutdown_function` callbacks **are** called if an exception is thrown. That's the issue. The mutex is [removed inside a `finally` block](https://github.com/laravel/framework/blob/cb46639988f817be01c81056351361e699897438/src/Illuminate/Console/Scheduling/CallbackEvent.php#L88) and also [during the shutdown function](https://github.com/laravel/framework/blob/cb46639988f817be01c81056351361e699897438/src/Illuminate/Console/Scheduling/CallbackEvent.php#L71) so theoretically the same race condition you identified could happen for `CallbackEvent` execution as well:

1. Callback A starts executing
2. Callback A triggers an exception, causing the `removeMutex` call in the `finally` block. At the exact same moment in time, callback B starts and is able to run because the mutex was removed.
3. The `register_shutdown_function` callback from Callback A is executed, removing the mutex for callback B.
4. Callback C starts and is able to run because the mutex was removed twice by callback A.

It's very very unlikely for that sequence to happen, but I think it's theoretically possible…

PR #39499, opened 2021-11-05
[8.x] Added a `shouldRun` method to event listeners

mateusjunges:
This PR adds a `shouldRun` method to event listeners. This method determines if the event should run or not, similar to `shouldQueue`, for queud event listeners.

The `shouldRun` method receives the same parameters as `handle`, so you can define if your event listener sould run based on the event payload:

```php
public function handle($event)
{
    // Handle the event event
}

function shouldRun($event) 
{
    // Add your logic here
}
```

This removes logic to determine if the listener should be executed from the `handle` method to the newly added `shouldRun`, cleaning up event listeners substantially.


taylorotwell:

PR #39500, opened 2021-11-05
[9.x] Split scheduled event handling into discrete parts

inxilpro:
This is a continuation of #39498. I'm submitting it against `9.x` because I've changed some method names. This would only affect applications that have custom versions of `Scheduling\Event` or `Scheduling\CallbackEvent`.

This PR splits scheduled event execution into three parts:

1. `start`: Run before callbacks (removing mutex on exception)
2. `execute`: Execute the command or run the callback
3. `finish`: Set the exit code, run after callbacks, remove mutex

By doing so, we can improve the handling of mutex's and minimize the difference between foreground/background execution and across regular and callback events:

- This removes the need for `runCommandInBackground` and `runCommandInForeground` because start/finish are considered separate actions in both cases.
- We can centralize the calls to `callBeforeCallbacks` and `callAfterCallbacks` and easily perform exception handling around them.
- We can eliminate the need for `callAfterCallbacks` vs `callAfterCallbacksWithExitCode` — the `finish` method is used in foreground and background execution in exactly the same way.
- Much of the logic in `CallbackEvent` can be removed now that `execute` is its own discrete step. All the mutex handling and callback logic stays the same and the only real difference is the execution of the event.

This **should not be merged** until [my comment on #39498 is addressed](https://github.com/laravel/framework/pull/39498#issuecomment-962181501), though.

### DIFF
--- a/src/Illuminate/Console/Scheduling/CallbackEvent.php
+++ b/src/Illuminate/Console/Scheduling/CallbackEvent.php
@@ -25,6 +25,13 @@ class CallbackEvent extends Event
     protected $parameters;
 
     /**
+     * The result of the callback execution.
+     *
+     * @var mixed
+     */
+    protected $result;
+
+    /**
      * Create a new event instance.
      *
      * @param  \Illuminate\Console\Scheduling\EventMutex  $mutex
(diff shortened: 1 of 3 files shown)


driesvints:
Closing due to inactivity. Feel free to resubmit once it's ready.

PR #39502, opened 2021-11-06
with php artisan test --filter it need @test to recognize tests and r…

rivassh:
a little attribute in php doc that needed to run --filter of php artisan test


driesvints:
The method already starts with `test_` which is the default for PHPUnit to detect a test method.

PR #38538, opened 2021-08-25
[9.x] Improves `Support\Collection` and `Database\Eloquent\Collection` type definitions

nunomaduro:
Following the discussion on https://github.com/laravel/framework/pull/38257, this pull request starts the work of introducing **serious strong type definitions** on Laravel core. What this pull request **include**:

1. GitHub action job that validates/tests the type definitions.
2. Improves type definitions of `Support\Collection` and `Database\Eloquent\Collection`.

generics_laravel_collections-3

What this pull request **does not** include:
- Usage of this new type definitions on `Database\Eloquent`. In other words, `collect(['string' => 1])->flip()->all()` or `collect()->map()` is 100% static analyzable and understood by static analysis tools - PHPStorm + PHPStan + ~~[Psalm](https://github.com/vimeo/psalm/issues/6357)~~, but not yet `User::all()`.

As the next step, going to improve the type definitions of `Database\Eloquent`. That's only possible now because of the this work we did with collections.

PR #39506, opened 2021-11-06
[9.x] Improve `Support\LazyCollection` type definitions

sebdesign:
Similarly to #38538, this PR makes the `LazyCollection` generic.

I haven't tested this with PHPStorm or Psalm, but in VSCode + PHPStan extension it works great.

Also:

- Fixed some minor issues with the `Support\Collection` types
- Added missing tests for some methods

There are still a few issues with the constructors and the `make()` method of the collections:
- The `Support\LazyCollection` needs to widen the type of items it accepts, because it accepts `callable`s that return `Generator`s, but the `Support\Collection` does not accept that. I have extended the `make` method and it works as expected.

- For some reason `new Eloquent\Collection([new User()])` returns `Eloquent\Collection<(int|string), Eloquent\Model>` instead of `Eloquent\Collection<int, User>`. If I extend the constructor it works (but I haven't included this).


driesvints:
Seems like tests are failing here. Feel free to markt his as ready when tests are passing.

sebdesign:
All good now! Maybe @nunomaduro would like to review this?

driesvints:
Ping @JosephSilber

sebdesign:
Also, @JosephSilber I think there is an issue with the internal types the `LazyCollection`: The `makeIterator` returns a `Traversable` but it is used as if it's an `Iterator`.

PR #39507, opened 2021-11-06
Add unauthorized method for Laravel Http Client

marvelefe:
While working with the Http Client and making requests to services that require some form of authentication,  I found that I have been doing:
```php
$response = Http::withToken('apikey')->post('http://example.com', [
      'data' => $data,
]);

if ($response->status() == 401 ||  $response->status() == 403) {
    # authentication error
}
```

and it became repetitive. With this pull request, everything written above can now be achieved with:

```php
$response = Http::withToken('apikey')->post('http://example.com', [
      'data' => $data,
]);

if ($response->unauthorized()) {
    # authentication error
}
```
Hopefully someone else finds it useful 😀

### DIFF
--- a/src/Illuminate/Http/Client/Response.php
+++ b/src/Illuminate/Http/Client/Response.php
@@ -191,6 +191,16 @@ class Response implements ArrayAccess
     }
 
     /**
+     * Determine if the response indicates the request is unauthorized or forbidden.
+     *
+     * @return bool
+     */
+    public function unauthorized()
+    {
+        return $this->status() == 401 || $this->status() == 403;
+    }
+
+    /**
      * Execute the given callback if there was a server or client error.
      *
      * @param  callable  $callback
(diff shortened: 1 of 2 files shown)


src/Illuminate/Http/Client/Response.php
```diff
@@ -190,6 +190,16 @@ public function serverError()
...
     }
 
+    /**
+     * Determine if the response indicates the request is unauthorized or forbidden.
+     *
+     * @return bool
+     */
+    public function unauthorized()
+    {
+        return $this->status() == 401 || $this->status() == 403;
```
driesvints, line 200: This definitely cannot cover 403 which is forbidden and a totally different thing than unauthorized.

taylorotwell:
Just check the status code you want directly.

dennisprudlo:
If you still want to use this as a method the Response class is macroable.

marvelefe:
Alright Thanks for the tip @dennisprudlo

PR #39508, opened 2021-11-06
[8.x] Add @count() to BladeEngine

faissaloux:
I frequently use @if ( count($array) ) like below, and sure a lot of devs do too.

```
@if ( count($array) )
//
@endif
```
 so I thought to simplify it by using @count
```
@count ( $array )
//
@endcount
```

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php
@@ -160,6 +160,17 @@ trait CompilesConditionals
     }
 
     /**
+     * Compile the if-count statements into valid PHP.
+     *
+     * @param  array  $expression
+     * @return string
+     */
+    protected function compileCount($expression)
+    {
+        return "<?php if(count{$expression}): ?>";
+    }
+
+    /**
      * Compile the unless statements into valid PHP.
      *
      * @param  string  $expression
@@ -197,6 +208,16 @@ trait CompilesConditionals
      * @return string
      */
     protected function compileEndif()
+    {
+        return '<?php endif; ?>';
+    }
+
+    /**
+     * Compile the end-count statements into valid PHP.
+     *
+     * @return string
+     */
+    protected function compileEndCount()
     {
         return '<?php endif; ?>';
     }
--- a/tests/View/Blade/BladeCountStatementsTest.php
+++ b/tests/View/Blade/BladeCountStatementsTest.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Illuminate\Tests\View\Blade;
+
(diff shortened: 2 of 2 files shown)


DarkGhostHunter:
Empty arrays are considered false. And you can use `blank()` and `filled()` helpers.

faissaloux:
> Empty arrays are considered false. And you can use `blank()` and `filled()` helpers.

But still have to use it like @if ( filled($array) )

dennisprudlo:
@faissaloux You don't need this helper.
> Empty arrays are considered false.

`@if ($array) ... @endif` works.

faissaloux:
> @faissaloux You don't need this helper.
> 
> > Empty arrays are considered false.
> 
> `@if ($array) ... @endif` works.

No, we want to check the array length we need the count helper

dennisprudlo:
Maybe I am missing something but you want the activate an if statement, if the array has values.
```php
count([]); // 0
count([1]); // 1
count([1, 2]); // 2
count([1, 2, 3]); // 3

(bool) []; // false
(bool) [1]; // true
(bool) [1, 2]; // true
(bool) [1, 2, 3]; // true
(bool) [0]; // true
(bool) [false]; // true
```

So `if (count($array))` and `if ($array)` will both evaluate to `false` if the array is empty and to `true` if the array is not empty. There's also `@empty(...)` to check if something is empty, but a variable not existing is also considered empty.

faissaloux:
> Maybe I am missing something but you want the activate an if statement, if the array has values.
> 
> ```
> count([]); // 0
> count([1]); // 1
> count([1, 2]); // 2
> count([1, 2, 3]); // 3
> 
> (bool) []; // false
> (bool) [1]; // true
> (bool) [1, 2]; // true
> (bool) [1, 2, 3]; // true
> (bool) [0]; // true
> (bool) [false]; // true
> ```
> 
> So `if (count($array))` and `if ($array)` will both evaluate to `false` if the array is empty and to `true` if the array is not empty. There's also `@empty(...)` to check if something is empty, but a variable not existing is also considered empty.

Yes but for collections:
```
(bool) collect([]); // true
```

So here we need the count
```
(bool) count( collect([]) ); // false
```

To achieve this in blade we use
```
@if ( count( $collection ) )
  //
@endif
```

In this PR I simplified it using @count
```
@count( $collection )
  //
@endcount
```

DarkGhostHunter:
`@if($collection->isEmpty())`

faissaloux:
> `@if($collection->isEmpty())`
> 
> This PR is unnecessary IMO.

Yes this is correct but this PR is just to simplify this line of code.

Isn't @count( $collection ) better than @if( $collection->isEmpty() ) !!

taylorotwell:
This has more downvotes than upvotes for now so going to hold off on merging this.

PR #39515, opened 2021-11-07
bugfix getTrustedHeaderNames never return headers

wuwx:
getTrustedHeaderNames allways return from switch's default
delete default from switch then return headers property

### DIFF
--- a/src/Illuminate/Http/Middleware/TrustProxies.php
+++ b/src/Illuminate/Http/Middleware/TrustProxies.php
@@ -116,9 +116,6 @@ class TrustProxies
             case 'HEADER_X_FORWARDED_PROTO':
             case Request::HEADER_X_FORWARDED_PROTO:
                 return Request::HEADER_X_FORWARDED_PROTO;
-
-            default:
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Tests don't agree?

src/Illuminate/Http/Middleware/TrustProxies.php
```diff
@@ -116,9 +116,6 @@ protected function getTrustedHeaderNames()
             case 'HEADER_X_FORWARDED_PROTO':
             case Request::HEADER_X_FORWARDED_PROTO:
                 return Request::HEADER_X_FORWARDED_PROTO;
-
-            default:
-                return Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_HOST | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PROTO | Request::HEADER_X_FORWARDED_AWS_ELB;
         }
 
         return $this->headers;
```
GrahamCampbell, line 121: instead, that line should be deleted.

GrahamCampbell (changes requested):

taylorotwell:
See comment above.

PR #39522, opened 2021-11-07
[8.x] Introduce `@js()` directive

inxilpro:
This is a rebased version of #39478 that I'm resubmitting [after a conversation](https://twitter.com/inxilpro/status/1456826254817103873) with @calebporzio. If this gets merged, either he or I can add a `version_compare` check to Livewire that would only register the existing Livewire directive for Laravel `v8.69.0` and lower (or whatever release is one below the version this ended up in).

### DIFF
--- a/src/Illuminate/View/Compilers/BladeCompiler.php
+++ b/src/Illuminate/View/Compilers/BladeCompiler.php
@@ -20,6 +20,7 @@ class BladeCompiler extends Compiler implements CompilerInterface
         Concerns\CompilesIncludes,
         Concerns\CompilesInjections,
         Concerns\CompilesJson,
+        Concerns\CompilesJs,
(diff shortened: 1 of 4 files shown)


calebporzio:
I approve this message

lupinitylabs:
> I approve this message

I have just noticed that in a project using Laravel 9.7.0 and Livewire 2.10.5 the Livewire @js directive seems to take precedence over the framework implementation.

Following this PR, which states that Livewire should not register its directive on Laravel 8.71.0+, it seems like this is not the intended behavior?

@calebporzio Can you confirm this?

lupinitylabs:
I opened a discussion at https://github.com/livewire/livewire/discussions/4875 with more details as this seems to be an issue of Livewire rather than Laravel.

PR #31917, opened 2020-03-11
[7.x] Intelligently drop unnamed prefix name routes when caching

taylorotwell:
In Laravel 6.x, when using route name groups, any routes not given a more an explicit name would be given the group name. For example:

```php
Route::name('admin.')->group(function () {
    Route::get('/dashboard');
    Route::get('/users');
});
```

Both of the routes above would be receive the route name `admin.`. This, of course, is odd behavior but was not particularly problematic until Laravel 7.x route caching was introduced.

Laravel 7.x uses Symfony's very fast route matcher, but that route matcher requires **every** route to have a unique name. This has introduced a significant barrier to upgrade to Laravel 7.x because people are now forced to remove the route name prefix from the group or literally assign a route name to every single route within the group explicitly. 

In my opinion, when caching, the framework should assign a generated name to a route **if the route's name ends in a `.` and it will create a duplicate route name error**. When we encounter routes like these, I think it is safe to make the assumption that the developer did not actually intend for them to have a name at all and likely have been unaware they were receiving a name in previous Laravel releases. In fact, this caused a problem with Laravel Nova itself that even I did not expect.

This will significantly ease the burden of upgrading to Laravel 7.x when using route name groups with no breaking changes since we are only changing behavior in a situation where the code was about to throw an exception anyways.

PR #39525, opened 2021-11-08
[8.x] Don't name unnamed routes inside a group with a prefix

tobyzerner:
Currently if you register an unnamed route within a group with a name prefix, the route's name is set to the prefix.

```php
$router->name('api.')->group(function ($router) {
    $router->get('users', 'UsersController@index');
});

$routes = $router->getRoutes()->get();
echo last($routes)->getName(); // "api."
```

While the consequences of this are somewhat obscure ([but definitely exist](https://github.com/laravel/framework/pull/31917)), it is indeed odd behavior and I would expect `getName()` to return `null` in the above example since I never explicitly set a name on the route.

This PR attempts to fix that within the constraints of the current architecture. I'm not sure if this would be considered a breaking change vs a bug fix, but I can PR this instead to master if desired.

Basically, instead of passing the group's `$action['as']` value into each new route created within it, it passes it as `$action['as_prefix']`. Then when a name is set on a route, it prepends `$action['as_prefix']`.

This also changes the behavior of calling `name` more than once on a route. Previously each time you called it, the value passed would be continually appended:

```php
$router->name('api.')->group(function ($router) {
    $router->get('users', 'UsersController@index')->name('foo')->name('bar') // "api.foobar"
});
```

Now, the value will be appended onto the initial group prefix each time:

```php
$router->name('api.')->group(function ($router) {
    $router->get('users', 'UsersController@index')->name('foo')->name('bar') // "api.bar"
});
```

### DIFF
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -855,7 +855,7 @@ class Route
      */
     public function name($name)
     {
-        $this->action['as'] = isset($this->action['as']) ? $this->action['as'].$name : $name;
+        $this->action['as'] = isset($this->action['as_prefix']) ? $this->action['as_prefix'].$name : $name;
 
         return $this;
     }
--- a/src/Illuminate/Routing/RouteGroup.php
+++ b/src/Illuminate/Routing/RouteGroup.php
@@ -92,8 +92,12 @@ class RouteGroup
      */
     protected static function formatAs($new, $old)
     {
-        if (isset($old['as'])) {
-            $new['as'] = $old['as'].($new['as'] ?? '');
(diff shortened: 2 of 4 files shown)


devcircus:
Seems like a common sense change but worried it would be a breaking change for a lot of apps. I've see a number of apps with route "groups" for individual routes and using the group name for the route. I know I've ended up with one route in a group after moving other routes out of the group and never refactoring. IMO this should go into the next version with a note in the upgrade guide.

driesvints:
This is a huge breaking change.

tobyzerner:
@taylorotwell would you accept this for master/9.x? Or not interested?

PR #39527, opened 2021-11-08
[8.x] Update placeholders for stubs

VladyslavChernyshov:
Updated deprecated placeholders for stubs.
StubPublishCommand.php also updated (paths to stubs optimized)


Jubeki:
I think you should still replace `DummyClass` and so one because it is a breaking change for all applications which use custom/published stubs and which are not updated accordingly.

PR #39529, opened 2021-11-08
[8.x] Configurable lost connection and concurrency error checks

martindilling:
Currently, the checks for a lost connection or a concurrency error is not configurable in any way (not any way I could figure out at least).

This adds a way to set a callback to be used as part of the checks, so I can just configure this callback, and expect Laravel to behave as it normally would when encountering these two cases.

I don't know how your process for backporting things is, but if possible, I would love for this to be backported to previous versions.

Open to suggestion, if you have a better idea on how to solve this issue :)

### DIFF
--- a/src/Illuminate/Database/DetectsConcurrencyErrors.php
+++ b/src/Illuminate/Database/DetectsConcurrencyErrors.php
@@ -38,7 +38,7 @@ trait DetectsConcurrencyErrors
     {
         if (
             is_callable($this->concurrencyErrorCheck)
-            && call_user_func($this->concurrencyErrorCheck, $e)
+            && ($this->concurrencyErrorCheck)($e)
         ) {
             return true;
         }
--- a/src/Illuminate/Database/DetectsConcurrencyErrors.php
+++ b/src/Illuminate/Database/DetectsConcurrencyErrors.php
@@ -18,7 +18,7 @@ trait DetectsConcurrencyErrors
     /**
      * Set a custom check to be used in the concurrency error check.
      *
-     * @param callable $check
+     * @param  callable  $check
      * @return $this
      */
     public function setConcurrencyErrorCheck(callable $check)
--- a/src/Illuminate/Database/DetectsConcurrencyErrors.php
+++ b/src/Illuminate/Database/DetectsConcurrencyErrors.php
@@ -9,6 +9,26 @@ use Throwable;
 trait DetectsConcurrencyErrors
 {
     /**
+     * A configurable check for a concurrency error exception.
+     *
+     * @var callable|null
+     */
+    protected $concurrencyErrorCheck;
+
+    /**
+     * Set a custom check to be used in the concurrency error check.
+     *
+     * @param callable $check
+     * @return $this
+     */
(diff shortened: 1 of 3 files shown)


src/Illuminate/Database/DetectsConcurrencyErrors.php
```diff
@@ -16,6 +36,13 @@ trait DetectsConcurrencyErrors
      */
     protected function causedByConcurrencyError(Throwable $e)
     {
+        if (
+            is_callable($this->concurrencyErrorCheck)
+            && call_user_func($this->concurrencyErrorCheck, $e)
```
GrahamCampbell, line 41: just call it with native syntax

src/Illuminate/Database/DetectsConcurrencyErrors.php
```diff
@@ -16,6 +36,13 @@ trait DetectsConcurrencyErrors
      */
     protected function causedByConcurrencyError(Throwable $e)
     {
+        if (
+            is_callable($this->concurrencyErrorCheck)
```
GrahamCampbell, line 40: logic seems wrong. if is callable but returns false, it's output is ignored and the original is used

GrahamCampbell (changes requested):

src/Illuminate/Database/DetectsConcurrencyErrors.php
martindilling: Not entirely sure what you mean. But I'll answer to what I think you mean.
My goal was to get a way to add more definitions of what a (in this case) concurrency error looks like. If we just returned the result from `concurrencyErrorCheck()`. It wouldn't do the existing checks. And I would have to reimplement the existing checks.
However I do agree it seems strange to only use it if it's true. But I'm not sure how the logic should work another way.

src/Illuminate/Database/DetectsConcurrencyErrors.php
martindilling: Done. I just haven't seen that syntax used as much as this when I've looked at the framework code :)

taylorotwell:

martindilling:
@taylorotwell Unless I'm missing something (can't rule that out), the problem is there is no way to "hook" into these checks. It's a hardcoded list of text strings to look for in exception messages. And if I need another exception to be handled as a lost connection, there is no way for me to do that. 
If you (or anyone else) have an idea about how to hook into this currently, I'd love to hear it. I'm trying to get a project off an annoying fork of the framework :)

Specifically I have these two extra lines:
```
    'Error while reading line from the server', // Predis
    'went away', // PhpRedis
```
But I thought it would be better with some way in, instead of PRs with text checks for everything. And then the 'went away' seems like a very broad thing to check for per default.

If you have any suggestions for a solution I'll gladly work on a PR :)

martindilling:
@taylorotwell Any chance you could look at this :) I wouldn't mind making this as a package, but I can't see any way to do that as the code is now.

As I mentioned, if you can hint at an approach you might accept, I'll gladly work on a PR :)

PR #35588, opened 2020-12-12
[9.x] Fix several lingering issues with PostgreSQL search_path handling

cbj4074:
In previous PRs I submitted, in relation to https://github.com/laravel/ideas/issues/918 , I spotted two other issues that should be fixed while we're at it. These issues shouldn't cause any further breaking changes.

1. Crucially, within the `parseSearchPath()` method, PostgreSQL's special `$user` variable wasn't being resolved to the username specified on the connection (just as is done in the `parseSchemaAndTable()` method).
2. The `getAllTables()` and `getAllViews()` methods, which are used when querying the list of tables/views to be dropped, were still referring to the `schema` configuration variable, which should be removed completely from the keys defined on the configuration by default, in favor of `search_path` (in the `laravel/laravel` repository). This PR updates that reference to `search_path` and wraps the value in the `parseSearchPath()` method.
3. I had overlooked a handful of `schema` references that should also have been updated to `search_path` or `searchPath`.

Also, while not _directly_ related to these changes, I added a couple of new tests to cover previously-uncovered code that is tangentially related to these changes.

Finally, it's worth cross-linking https://github.com/laravel/ideas/issues/2438 so that my reasoning is documented for posterity.

I swear that this is my last PR on this subject! Unless it's to add more tests for uncovered code! ;-)

PR #35463, opened 2020-12-02
[9.x] Address confusion between PostgreSQL concepts "schema" and "search_path"

cbj4074:
This PR is intended to fulfill the objectives described in https://github.com/laravel/ideas/issues/918 , as they apply to this repository (related changes may be needed in `laravel/laravel`).

In short, this PR replaces the PostgreSQL term "schema" with "search_path". While related and conceptually similar, there is a crucial distinction to be made between the two, which I describe exhaustively in the above-cited post.

Beyond the terminology corrections, we have also sought to fix a significant limitation whereby a `search_path` that contains more than one schema cannot be specified as a string, e.g., via the effective environment (i.e., an `.env` file or similar) because the implementation does not parse and quote the specified value correctly.

Finally, we have added several new tests to ensure complete consistency in how the `search_path` value is passed on to PostgreSQL, regardless of whether the value is specified as:

- a string that contains only one schema (with or without double-quotes around the schema)
- a comma-separated string that contains more than one schema (with or without double-quotes around each schema)
- an array that contains any number of schemas (with or without double-quotes around each schema)
- any of the above variations, but with variable notation, e.g., the schema name begins with a `$` sign

At this juncture, I am unaware of any other valid `search_path` syntax that could possibly need to be supported.

~~Also, per https://github.com/laravel/ideas/issues/918 , @mfn had expressed an interest in completely expunging the term "schema" from Laravel's lexicon in relation to this issue, with which I more or less agree. However, I left the reference in the `configureSearchPath()` method so as to provide backwards compatibility, which makes sense only if this can be merged into the `8.x` branch. If the changes are already breaking, then I agree that we should completely expunge "schema" and target the next major release, at which time we should also update the default configuration array in https://github.com/laravel/laravel/blob/8.x/config/database.php (by replacing `schema` with `search_path`), and update any affected documentation, if any actually exists (I have yet to see any description of PostgreSQL's `schema`/`search_path` directive, or what format it should take when specified in Laravel's database configuration).~~

(UPDATE: This is breaking, so I expunged the incorrect terminology everywhere, with no concern for preserving BC with regard to the database connection configuration key name.)

~~I'm looking for some guidance from @taylorotwell or whomever else may be qualified in this regard.~~ (UPDATE: driesvints already addressed this.)

[...]

PR #39531, opened 2021-11-08
[9.x] Fix postgresql build

driesvints:
More graceful phase out of the `schema` key for Postgres. Introduced in https://github.com/laravel/framework/pull/35463 & https://github.com/laravel/framework/pull/35588.

This PR will fallback to the old `schema` key if no `search_path` key is present in the database.php file. This way there's a smooth upgrade path for Postgres users.

### DIFF
--- a/.github/workflows/databases.yml
+++ b/.github/workflows/databases.yml
@@ -132,49 +132,49 @@ jobs:
           DB_CONNECTION: mysql
           DB_USERNAME: root
 
-  # pgsql:
-  #   runs-on: ubuntu-20.04
+  pgsql:
+    runs-on: ubuntu-20.04
 
-  #   services:
-  #     postgresql:
-  #       image: postgres:14
-  #       env:
-  #         POSTGRES_DB: forge
-  #         POSTGRES_USER: forge
-  #         POSTGRES_PASSWORD: password
-  #       ports:
-  #         - 5432:5432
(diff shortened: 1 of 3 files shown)


.github/workflows/databases.yml
```diff
@@ -132,49 +132,49 @@ jobs:
...
-  #       run: vendor/bin/phpunit tests/Integration/Database --verbose --exclude-group MySQL
-  #       env:
-  #         DB_CONNECTION: pgsql
-  #         DB_PASSWORD: password
+  pgsql:
+    runs-on: ubuntu-20.04
+
+    services:
+      postgresql:
+        image: postgres:14
```
GrahamCampbell, line 140: Would be cool to test on earlier versions too. Almost nobody is using version 14 yet. ;)

driesvints: I don't think it's necessary to test too many versions. We already have quite a bit of builds. It's not like these major postgres versions are so vastly different that they warrant separate builds.

GrahamCampbell: Maybe just test the lowest supported, instead of the highest?

driesvints: Yeah that maybe makes sense.

cbj4074 (commented):
Looks good to me!

PR #39535, opened 2021-11-08
[8.x] Cleanup tests

driesvints:
This PR contains the following changes:

- Do not explicitly set debug mode anymore. These calls were probably legacy code lingering.
- Move MySQL specific tests in a separate MySql directory so they're grouped per-engine and easier discoverable. This way we can remove the `@group` call as well.
- Modify `EloquentUpdateTest` & `EloquentDeleteTest` to run on all engines and skip a few tests on MSSQL.

### DIFF
--- a/tests/Integration/Database/EloquentDeleteTest.php
+++ b/tests/Integration/Database/EloquentDeleteTest.php
@@ -13,8 +13,6 @@ class EloquentDeleteTest extends TestCase
 {
     protected function getEnvironmentSetUp($app)
     {
-        $app['config']->set('app.debug', 'true');
-
         $app['config']->set('database.default', 'testbench');
 
         $app['config']->set('database.connections.testbench', [
--- a/tests/Integration/Database/EloquentDeleteTest.php
+++ b/tests/Integration/Database/EloquentDeleteTest.php
@@ -34,6 +34,7 @@ class EloquentDeleteTest extends DatabaseTestCase
         });
     }
 
+    /** @group SkipMSSQL */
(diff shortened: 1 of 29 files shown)


tests/Integration/Database/EloquentDeleteTest.php
```diff
@@ -59,7 +45,11 @@ public function testOnlyDeleteWhatGiven()
         Post::latest('id')->limit(1)->delete();
         $this->assertCount(9, Post::all());
 
-        Post::join('comments', 'comments.post_id', '=', 'posts.id')->where('posts.id', '>', 1)->orderBy('posts.id')->limit(1)->delete();
+        Post::join('comments', 'comments.post_id', '=', 'posts.id')
+            ->where('posts.id', '>', 8)
```
driesvints, line 49: SQLite and MySQL behave different here so we should adjust the test accordingly.

tests/Integration/Database/MySql/MySqlTestCase.php
```diff
@@ -0,0 +1,17 @@
...
+
+abstract class MySqlTestCase extends DatabaseTestCase
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        if ($this->driver !== 'mysql') {
+            $this->markTestSkipped('Test requires a MySQL connection.');
+        }
```
driesvints, line 15: Skip tests on non-MySQL connections so they don't error out when using other drivers.

GrahamCampbell (approved):
:shipit:

PR #39540, opened 2021-11-09
[8.x] TestResponse::statusMessageWithException - Caters for situations where exception may be an array

specialtactics:
Sometime fairly recently, some additional functionality was added to output exceptions when using (in our case) the assertJsonValidationErrors() function. For us at least, sometimes the exception is not an instance of throwable, but an array - ie.

```php
array(4) {
["code"]=>
int(0)
["message"]=>
string(26) "Could not create resource."
["line"]=>
int(71)
["file"]=>
string(34) "/var/www/app/Helpers/Utilities.php"
}
```

These tests then will result in an ErrorException being thrown with the message `Array to string conversion`, as php won't cast-convert an array into a string, and the tests fail with this error. Potentially there can be other situations with a variable that cannot be stringified in this way.

What I propose is to keep the current functionality as it is, but try and catch an ErrorException on this cast conversion, and if there is a problem, use the more universal var_export instead.

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -19,6 +19,7 @@ use Illuminate\Testing\Assert as PHPUnit;
 use Illuminate\Testing\Constraints\SeeInOrder;
 use Illuminate\Testing\Fluent\AssertableJson;
 use LogicException;
+use ErrorException;
 use Symfony\Component\HttpFoundation\StreamedResponse;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you resubmit this with instructions how to to reproduce it in a fresh Laravel application?

specialtactics:
@taylorotwell would take way less of my time to just fix something which was functional prior.

PR #39555, opened 2021-11-10
Changing the function authorize to return true

abhilash14001:
As most of the developers when generate form request they surely return to true. Instead of changing by themselves we make them easier by changing it to default value to return true

### DIFF
--- a/src/Illuminate/Foundation/Console/stubs/request.stub
+++ b/src/Illuminate/Foundation/Console/stubs/request.stub
@@ -13,7 +13,7 @@ class {{ class }} extends FormRequest
      */
     public function authorize()
     {
-        return false;
+        return true;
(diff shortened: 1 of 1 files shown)


Neol3108:
If I don't need authorization, (so always true) I just remove the method...

abhilash14001:
Yes

> If I don't need authorization, (so always true) I just remove the method...

Yes that could be other way, but what if some developers should change it to false, so the function is required

axlon:
For authorization I'd say `false` is a way more sensible default than `true`, because you can't accidentally forget to implement it as your `FormRequest` simply won't work

devcircus:
Agreed. that's why it's the way it is. It's been discussed before.

PR #39556, opened 2021-11-10
[8.x] Using a different composer file

dikopylov:
For now, users should only use `composer.json` file with no choice.
Even though the official Composer documentation - https://getcomposer.org/doc/03-cli.md#composer allows it.
Multiple composer files are required if a project has multiple configurations. For example, composer-dev.json specifies the dependencies for the latest (develop) versions, and composer.json specifies the dependencies for specific versions.


taylorotwell:

PR #39561, opened 2021-11-10
Fix Mailable markdown text view

clytras:
Markdown mailables are using html views when there is a text view specified.
That's wrong, any markup in `text-plain` MIME type is useless and it will result in countless issues not having proper text messages for emails (accessibility, parsers, email clients that only support text like CLI clients). Instead the specified text view should be used to render plain text.

See the detailed description on this discussion: https://github.com/laravel/framework/discussions/39520

### DIFF
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -344,8 +344,7 @@ class Mailable implements MailableContract, Renderable
      */
     protected function buildMarkdownText($markdown, $data)
     {
-        return $this->textView
-                ?? $markdown->renderText($this->markdown, $data);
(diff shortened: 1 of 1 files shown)


taylorotwell:
If you call the `text` method explicitly - it's just going to render what you give it. I'm not going to change this for existing applications. So, if you pass a `text` view explicitly that view should render to pure plain text. Note that on Markdown mailables you do not need to typically do this as the `markdown` method will add a plain/text version of the message with unrendered Markdown to the email for you.

clytras:
Thank you @taylorotwell for the clarification.
I really struggled to figure out how this is working but it wasn't very clear to me.
For DRY templates, I wanted to mix markdown and text views which apparently is not a good idea.

PR #39567, opened 2021-11-10
[8.x] Add new allows blade directive

khalyomede:
This pull request adds a new "allows" directive. It is an alternative to the "can" directive, and only provides the possibility to get the "reason" (message) that is the result of the Gate inspect message.

Users can check if an action is authorized.

```php
@can("create", App\Models\Post::class)
    You can create a post!
@elsecan
    You cannot create a post...
@endcan
```

To get the message/reason for the denial, we have to pass a variable holding the result of the Gate's message method call like this:

```php
@php
    use App\Models\Post;

    $creatingPost = Gate::inspect("create", Post::class);
@endphp

@if($creatingPost->allowed())
    You can create a post!
@else
    {{ $creatingPost->message() }}
@endif
```

(Note that you would have probably pass the `$creatingPost` from your controller)

This pull request offers a more straight forward way to do all of this using a shortened code.

```php
@allows("create", App\Models\Post::class)
    You can create a post!
@elseallows
    {{ $message }}
@endallows
```

It works exactly like the `@error` directive, exposing a local `$message` variable that holds the result of the Gate's message call. Note that like the `@error` directive, if there was already an existing `$message` variable, its value will be stored in a variable called `$__messageOriginal`.

Also note that you can pass the message of the allowing result as well:

```php
namespace App\Policies;

use App\Models\Post;
use Illuminate\Auth\Access\HandlesAuthorization;
use Illuminate\Auth\Access\Response;

final class PostPolicy
{
  use HandlesAuthorization;

  public function update(User $user, Post $post): Response
  {
    return $post->author->is($user) ?
      Response::allow("Update your post.") :
      Response::deny("Only author can update their post.");
  }
}
```

```php
@allows("update", $post)
    {{ $message }}
@elseallows
    {{ $message }}
@endallows
```

No breaking changes should occur after merging this PR.

I shamelessly copy/pasted the source code on my own L8 project to test if it works (and it did), since I cannot figure out how to test the whole integration.

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesAuthorizations.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesAuthorizations.php
@@ -132,6 +132,8 @@ trait CompilesAuthorizations
      */
     protected function compileEndallows()
     {
-        return '<?php endif; ?>';
+        return '<?php unset($message);
+if (isset($__messageOriginal)) { $message = $__messageOriginal; }
+endif; ?>';
     }
 }
--- a/src/Illuminate/View/Compilers/Concerns/CompilesAuthorizations.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesAuthorizations.php
@@ -45,7 +45,8 @@ trait CompilesAuthorizations
      */
     protected function compileAllows($expression)
     {
-        return '<?php $access = app(\Illuminate\\Contracts\\Auth\\Access\\Gate::class)->inspect'.$expression.'; ?>
+        return '<?php if (isset($access)) { $__accessOriginal = $access; }
+$access = app(\Illuminate\\Contracts\\Auth\\Access\\Gate::class)->inspect'.$expression.'; ?>
 <?php if ($access->allowed()): ?>
 <?php if (isset($message)) { $__messageOriginal = $message; } $message = $access->message(); ?>';
     }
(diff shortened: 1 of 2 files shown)


sebdesign:
If think you should unset the `$access` and `$message` variables inside `compileEndallows`, like `compileEndError` does:

```php
<?php unset($message);
if (isset($__messageOriginal)) { $message = $__messageOriginal; }
endif;
unset($access); ?>
```

khalyomede:
I also would like your opinion because I thought of even more "readable" Blade keys. What do you think of this:

```php
@allow("update", $post) 
  You can update your post! 
@deny
  {{ $message }}
@endallows
```

dennisprudlo:
It would be breaking but I kind of think this should be integrated into the `@can` directive because they do basically the same thing. The only difference is that `@allows` exposes that message. But as I said this would be problematic for all cases where people use the `@can` directive with a `$message` var inside.

khalyomede:
That's for this exact reason I've decided to make a new directive @dennisprudlo. I am eager to make `@can` benefit from this, and target 9.x branch. I guess it's up to the Laravel team to decide 🤔 

If you need my opinion -> I guess 9.x is close, so it's time to stop adding new features, and postpone anything for the next big release.

inxilpro:
I agree that I'd just add this to `@can()` in `9.x` and wait until then for its release. If `$message` is too general, you could name the new variable something like `$gateMessage` to minimize conflicts.

IgorGaming:
Yes, I agree with that too. There is no need to add another directive that actually does the same thing.

taylorotwell:

PR #39568, opened 2021-11-10
[8.x] Allow assertion of multiple JSON validation errors.

marshall-davis:
Allow multiple validation error messages to be expected when testing responses from JSON API.

The `TestResponse::assertJsonValidationErrors()` method already accepts a single message to be expected for any given request attribute. It seems intuitive that since the actual JSON response might contain an array of validation errors, if more than one rule is violated, that this method could expect multiple messages per attribute.

Currently this assumes that each attribute will expect a single message and if an array is provided it will throw an exception for an array to string conversion when creating the error message.

The change allows for multiple messages to be expected and for the assertion to fail if any of the provided messages is not in the response.

All previous tests continue to pass and previous usages of this method still work so this should not break any existing code based on this method.

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -871,7 +871,7 @@ EOF;
      */
     public function assertJsonValidationErrorFor($key, $responseKey = 'errors')
     {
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #39120, opened 2021-10-06
[8.x] New flag --requests -R to make:controller and make:model Commands

PovilasKorop:
This PR adds a new flag to `php artisan make:controller` command, which creates FormRequest class and uses it immediately in the generated controller.

## Problem

How I personally use Resourceful Controllers with Form Requests:

1. `php artisan make:controller UserController --resource`
2. `php artisan make:request StoreUserRequest`
3. `php artisan make:request UpdateUserRequest`
4. Go to the `UserController` and manually replace `Illuminate\Http\Request` with those new FormRequest classes

Why couldn't we generate it all at once?

This problem resonated with my Twitter audience, with 180+ likes on [this Tweet](https://twitter.com/PovilasKorop/status/1444530282078613506).

- - - - -

## My Proposed Solution (updated October 7)

Now, when you run `php artisan make:controller UserController --resource --model=User --requests`, it will generate two FormRequest classes: `StoreUserRequest` and `UpdateUserRequest`, and will replace them in the Controller.

A short version of `--requests` flag is `-R`, not to mix with `-r` for resources.

It is done by reusing the `make:request` command and calling it with appropriate class name.

This functionality works __only__ together with `--model` flag, otherwise it would be impossible/hard to guess the name for the FormRequest classes.

Also, it works one "layer" above in the Model: `php artisan make:model Project -mcrR` would generate `StoreProjectRequest` and `UpdateProjectRequest`

- - - - - 

## Breaking changes

It shouldn't break any existing functionality, it's just a new flag. If that flag isn't used, it all falls back to the old default values. 

- - - - - 

Any other suggestions for improvement are welcome.

PR #39578, opened 2021-11-11
[8.x] Make:model --all flag would auto-fire make:controller with --requests

PovilasKorop:
Following up on my [recently-merged PR](https://github.com/laravel/framework/pull/39120) about `--requests` flag, I received at least a few comments on my Youtube/Twitter that it would be beneficial to include Form Requests generation when launching `php artisan make:model Post --all` (or `-a` for short).


taylorotwell:

PR #39579, opened 2021-11-11
[8.x] Added declined and declined_if validation rules

Frozire:
Adds `declined` and `declined_if` validation rules.

Reverse off the `accepted` and `accepted_if` rules that require 'yes', 'on', '1', 1, true, 'true', these rules require a 'no', 'off', '0', 0, false, 'false' value.

As these functions are only new additions, they should not break anything.

Documentation PR: https://github.com/laravel/docs/pull/7425
Laravel PR: https://github.com/laravel/laravel/pull/5723

### DIFF
--- a/src/Illuminate/Validation/Concerns/ReplacesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ReplacesAttributes.php
@@ -25,6 +25,24 @@ trait ReplacesAttributes
     }
 
     /**
+     * Replace all place-holders for the declined_if rule.
+     *
+     * @param  string  $message
+     * @param  string  $attribute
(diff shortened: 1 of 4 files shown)


taylorotwell:
Can you explain the use case from your application for this?

Frozire:
> Can you explain the use case from your application for this?

We are running a financial application where the customer answers yes and no selectors, such as registered as a bad payer or having existing debt. We could solve it by reversing the question's phrasing, e.g. "not_registered_as_a_bad_payer", but to stay as close to the original intent as possible and limit developer errors or interpretation errors when comparing to business rules.

taylorotwell:

PR #39583, opened 2021-11-12
[9.x] Use match expression where possible

oliverquynh:
Laravel 9 will require PHP 8.0 as the minimum version, so we are free to use match expression. In this PR, I try to convert a few switch statements into match expressions that makes code easier to read.

Please review the code changes and see how nice the match expression is.

### DIFF
--- a/src/Illuminate/Auth/CreatesUserProviders.php
+++ b/src/Illuminate/Auth/CreatesUserProviders.php
@@ -33,16 +33,13 @@ trait CreatesUserProviders
             );
         }
 
-        switch ($driver) {
-            case 'database':
-                return $this->createDatabaseProvider($config);
-            case 'eloquent':
-                return $this->createEloquentProvider($config);
-            default:
-                throw new InvalidArgumentException(
-                    "Authentication user provider [{$driver}] is not defined."
-                );
-        }
+        return match ($driver) {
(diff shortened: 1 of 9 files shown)


src/Illuminate/Auth/CreatesUserProviders.php
```diff
@@ -33,16 +33,13 @@ public function createUserProvider($provider = null)
...
-                return $this->createEloquentProvider($config);
-            default:
-                throw new InvalidArgumentException(
-                    "Authentication user provider [{$driver}] is not defined."
-                );
-        }
+        return match ($driver) {
+            'database' => $this->createDatabaseProvider($config),
+            'eloquent' => $this->createEloquentProvider($config),
+            default => (throw new InvalidArgumentException(
```
taylorotwell, line 39: Why is the throw wrapped in parentheses?

seriquynh: There are not necessary. I removed them.

driesvints:
@seriquynh this seem to have broken the default `null` return in a couple of places: https://github.com/laravel/framework/pull/40880

seriquynh:
@driesvints I'm sorry for missing your mention because I haven't check github for a while. I see a fixed PR has been merged. Thanks for you guys!

PR #39587, opened 2021-11-12
[8.x] Make Listener Artisan command overwrites existing listener

npostman:
The `artisan make:listener` command overwrites an existing listener. Updating the `alreadyExists()` method with the method taken from the EventMakeCommand fixes this.

To reproduce:

run `artisan make:listener TestListener`

Make changes to TestListener.php and run `artisan make:listener TestListener` again.

File overwritten en changes gone.


taylorotwell:
This has been tried several times and breaks other use cases. You should be able to see them in the PR history.

npostman:
Darn. Sorry for not checking. I could have known there was a reason for that behavior.

PR #39593, opened 2021-11-12
[8.x] Add capability to ignore chars on `Str::random()`

xiCO2k:
This PR adds a capability to ignore certain chars on the `Str::random()`:

# Usage:

```php
/* It will ignore the chars "I, i, o, O, 0, l and 1" when generating a random string */
$code = Str::random(6, ['I', 'i', 'o', 'O', '0', 'l', '1']);
```

Let me know if its a good addition of if I need to make any change.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -549,9 +549,10 @@ class Str
      * Generate a more truly "random" alpha-numeric string.
      *
      * @param  int  $length
+     * @param  array<int, string>  $ignoreChars
      * @return string
      */
-    public static function random($length = 16)
+    public static function random($length = 16, $ignoreChars = [])
     {
         $string = '';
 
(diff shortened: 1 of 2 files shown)


xiCO2k:
Another Idea, is to have the capability to add just a string with all the chars to be ignored, like this:

```php
$code = Str::random(6, 'IioO0l1');
```

hicham-saddek:
The idea is good but the ignore string array is right after the length, it might be a bit confusing for developers, you should maybe consider using php 8.x in order to use variable placements like
````php
Str::random(32, ignore: "Ihgsbi7");
````
but other than that its quite a good idea.

should also consider adding a variable called `chain` which contains the chars to be generating with, so users can generate strings rven with special characters :).

xiCO2k:
Yes both great ideas, I will wait first for @taylorotwell input and then work on the refactor if needed.

taylorotwell:

PR #39597, opened 2021-11-13
[8.x] Fix: prevent re-casting of enum values

tomschlick:
I ran into an issue in testing 8.1 enums where when using observers/jobs to set an enum after a previous update it was trying to re-cast the enum value into the enum class using `EnumClass::from()` which only accepts an integer/string value. A simple check here stops that from happening.


taylorotwell:

PR #39608, opened 2021-11-14
[8.x] Enum casts accept backed values

Propaganistas:
This PR builds upon https://github.com/laravel/framework/pull/39315 to accept an enum's backed value as well.

Currently enum castable attributes first have to be instantiated (or be null) before they are accepted for casting:

```php
enum MyEnum: string
{
    case Pending = 'pending';
}
```

```php
$model = new Model;
$model->enum = MyEnum::Pending;
```

This can get verbose when handling requests:

```php
// Request input = ['enum' => 'pending']

$request->merge([
   'enum' => MyEnum::from($request->input('enum')),
]);

$model->fill(
    $request->all()
);
```

By also accepting the backed values of an enum when casting, fluent model filling will be enabled (again).

```php
$model = new Model;
$model->enum = 'pending';
```

```php
// Request input = ['enum' => 'pending']

$model->fill(
    $request->all()
);
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -926,7 +926,15 @@ trait HasAttributes
      */
     protected function setEnumCastableAttribute($key, $value)
     {
-        $this->attributes[$key] = isset($value) ? $value->value : null;
+        $enumClass = $this->getCasts()[$key];
+
+        if (! isset($value)) {
+            $this->attributes[$key] = null;
+        } else if ($value instanceof $enumClass) {
+            $this->attributes[$key] = $value->value;
+        } else {
+            $this->attributes[$key] = $enumClass::from($value)->value;
+        }
     }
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
```diff
@@ -926,7 +926,15 @@ function () {
...
     {
-        $this->attributes[$key] = isset($value) ? $value->value : null;
+        $enumClass = $this->getCasts()[$key];
+
+        if (! isset($value)) {
+            $this->attributes[$key] = null;
+        } else if ($value instanceof $enumClass) {
+            $this->attributes[$key] = $value->value;
+        } else {
+            $this->attributes[$key] = $enumClass::from($value)->value;
```
Propaganistas, line 936: This forces values to be valid a enum case.

themsaid:
This looks ok to me. We do the same with date casting where we accept a date object or a string or an integer.

PR #39613, opened 2021-11-15
[8.x] Add `allowIf` and `denyIf` helpers to `HandlesAuthorization`

lioneaglesolutions:
This PR allows a way to quickly return a policy response conditionally using similar syntax that is used elsewhere in the framework such as `$this->when( ... )`, `abort_if( ... )` and `throw_if( ... )`.

The two methods would accept either a value or a callback.

I have only allowed the message and code to be specified for the `deny` response as I was unsure as to where/when an `allow` response would actually ever be returned and used as a response.

An example of its usage is;
```php

// Current implementation
public function delete(User $user, Patient $patient): Response
{
    if ($user->hasPatient($patient)) {
        return $this->allow();
    }

    return $this->deny('You must be assigned to this patient.');
}

// Using new methods
public function delete(User $user, Patient $patient): Response
{
    return $this->allowIf($user->hasPatient($patient), 'You must be assigned to this patient.');
}
```

And to deny;
```php
public function delete(User $user, Product $product)
{
    return $this->denyIf($user->isCustomer());
}
```

### DIFF
--- a/src/Illuminate/Auth/Access/HandlesAuthorization.php
+++ b/src/Illuminate/Auth/Access/HandlesAuthorization.php
@@ -32,7 +32,7 @@ trait HandlesAuthorization
      * Conditionally create an access response or throw
      * an unauthorized exception.
      *
-     * @param  mixed $condition
+     * @param  mixed  $condition
      * @param  string|null  $message
      * @param  mixed|null  $code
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #39615, opened 2021-11-15
[9.x] Makes `with` generic

nunomaduro:
This pull request makes the `with` helper generic. The current type definition is not perfect, assuming that the return value from `with` callback will be same as the given value. Yet, it's better then what we had before. At PHPStan level 5 works perfectly, and PHPStorm, of course, is able to infer the types correctly.

Screenshot 2021-11-15 at 12 05 10

Screenshot 2021-11-15 at 12 06 17

### DIFF
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -368,9 +368,11 @@ if (! function_exists('with')) {
     /**
      * Return the given value, optionally passed through the given callback.
      *
-     * @param  mixed  $value
-     * @param  callable|null  $callback
-     * @return mixed
+     * @template TValue
+     *
+     * @param  TValue  $value
+     * @param  (callable(TValue): TValue)|null  $callback
+     * @return TValue
      */
     function with($value, callable $callback = null)
     {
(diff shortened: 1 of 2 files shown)


sebdesign:
@nunomaduro If think we should handle the case where the callback returns a different type than the value:

E.g.: https://github.com/laravel/framework/blob/master/src/Illuminate/Log/LogManager.php#L132

```php
/**
 * Return the given value, optionally passed through the given callback.
 *
 * @template TValue
 * @template TReturn
 *
 * @param  TValue  $value
 * @param  (callable(TValue): TReturn)|null  $callback
 * @return TValue|TReturn
 */
function with($value, callable $callback = null)
{
    return is_null($callback) ? $value : $callback($value);
}
```

sebdesign:
Your tests are asserting that the `$user` parameter is `int|User`, while it's `User`. The return type is `int|User` which is correct.

nunomaduro:
@sebdesign I am aware of that. Yet, PHPStan does not support default generics values - meaning that the code you proposed will make a phpstan error. So this pull request does the possible job for now.

sebdesign:
@nunomaduro thank you for your explanation! Indeed you are right we cannot use default/optional generics, I'm sorry for the noise on this PR.

PR #39619, opened 2021-11-15
[9.x] Use jsonb type for pgsql whereJsonLength

tpetry:
In PostgreSQL there are historically two different datatypes to store json data: The old `json` datatype storing everything as text and the new `jsonb` datatype with binary storage. The old textual storage is discouraged because it is in all respects inferior to binary storage, there are very few use cases where textual storage is advantageous.

Laravel uses the new jsonb datatype for all query builder operations, only `whereJsonLength` was forgotten to be adapted. Thus, any calculation of array length on jsonb columns first leads to a conversion of binary representation to textual representation. Since all other query builder operations work natively on jsonb, it is reasonable to adapt the only remaining function to fit the other query operations.

### DIFF
--- a/src/Illuminate/Database/Query/Grammars/PostgresGrammar.php
+++ b/src/Illuminate/Database/Query/Grammars/PostgresGrammar.php
@@ -138,7 +138,7 @@ class PostgresGrammar extends Grammar
     {
         $column = str_replace('->>', '->', $this->wrap($column));
 
-        return 'json_array_length(('.$column.')::json) '.$operator.' '.$value;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Does this work on both `json` and `jsonb` column types?

tpetry:
Yes it will work for both. In case a legacy json type is used it will be cast to jsonb.

taylorotwell:
I already merged this, but is there a big reason this couldn't just go into Laravel 8.x?

tpetry:
In a very very rare case this could be a breaking change. If someone built a functional index on the expression the index can‘t be used anymore because it‘s using a different function now.

PR #37128, opened 2021-04-26
[6.x] Fix nullable values for required_if

driesvints:
> `required_if`, `required_unless`, `exclude_if`, `exclude_unless` (referred to below as "these rules").

This PR fixes several things:

1. A regression introduced in https://github.com/laravel/framework/pull/36969#issuecomment-826140483: we now will take into account `nullable` values the same way as we do for`boolean` values. This will make sure that if someone adds `null` to these rules, it'll get properly handled for the targetted key's value.
2. These rules will now automatically pass when the targeted key isn't present in the data. That's the `if (! Arr::has($this->data, $parameters[0])) {` part that was added to each of them. This was already expected I believe but should now fail early.
3. I've added an `if (is_null($value)) {` to `FormatsMessages` so that validation messages can display things like `The baz field is required when foo is null.`. This might be a bigger breaking change so if you want that part reverted I'll do so. Keep in mind that that will cause the validation message to display `The baz field is required when foo is .` instead.

I'd very much like a some second pair of eyes on this. @timacdonald @jessarcher if you're willing?

PR #39625, opened 2021-11-16
fix(rollback) required_if and exclude_if validation

s4muel:
i think you do not want to return early in these cases as @driesvints changed in this PR: https://github.com/laravel/framework/pull/37128
that PR broke using this validation for `null` values (e.g. checkboxes that are not checked) so you cannot use it like `exclude_if,otherfield,null`

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1509,10 +1509,6 @@ trait ValidatesAttributes
     {
         $this->requireParameterCount(2, $parameters, 'required_if');
 
-        if (! Arr::has($this->data, $parameters[0])) {
-            return true;
-        }
-
         [$values, $other] = $this->parseDependentRuleParameters($parameters);
 
         if (in_array($other, $values, is_bool($other) || is_null($other))) {
@@ -1611,10 +1607,6 @@ trait ValidatesAttributes
     public function validateExcludeIf($attribute, $value, $parameters)
     {
         $this->requireParameterCount(2, $parameters, 'exclude_if');
-
-        if (! Arr::has($this->data, $parameters[0])) {
-            return true;
-        }
 
         [$values, $other] = $this->parseDependentRuleParameters($parameters);
 


driesvints:
Yes you do. This was a vital part of the fix. The problem is that you're not using `nullable` in the validation rules.

s4muel:
hm, the checkbox (otherfield in my example), even if nullable, is not present (null) in `$this->data`, so this always passes: https://github.com/laravel/framework/blob/8.x/src/Illuminate/Validation/Concerns/ValidatesAttributes.php#L1615

driesvints:
Please post the full validation rules and data that's being passed so we can reproduce this.

s4muel:
minimum to reproduce:
**validation rules:**
```
[
    'email' => 'exclude_if:owner,null|email',
    'owner' => 'nullable',
]
```
**form contains two inputs:**
email: input field
owner: checkbox

submit form with empty email and unchecked checkbox
dump of `$request->input()`:
```
[
  "_token" => "rjGnNZQ2OP9oqRQtn7y6yKFK8izvlh4wFXiRtSGF"
  "email" => null
  "send" => "Submit"
]
```

the email is not excluded and it fails on 'email' rule.

note: if i switch the logic around and use `exclude_unless`
```
[
    'email' => 'exclude_unless:owner,on|email',
    'owner' => 'nullable',
]
```
it works as expected

driesvints:
Screen Shot 2021-11-16 at 11 41 13

This is with your example. Like I said: you need to add `nullable` to your `email` rule:

Screen Shot 2021-11-16 at 11 42 01

s4muel:
oh, i see. kind of makes sense now. and since i want that email not to be null if the checkbox is true, i need a `sometimes` rule instead of `nullable` (`'email' => 'exclude_if:owner,null|sometimes|email'`). at least it seems to be working.

it wasnt clear though and when i updated laravel recently (from 8.38.0 to 8.68.1) my tests just broke (phew, at least i got that covered) because of that PR mentioned earlier. it previously worked like that so i got false understanding how it should work.

thank you!

driesvints:
@s4muel no worries. Glad you got that fixed 👍

PR #39627, opened 2021-11-16
Implement lazyById in descending order

moshe-autoleadstar:
Sometimes, when dealing with a lot of data, you need to lazily loop through models.
However, often, one would care more about the "newer" objects being processed first.

```
$query->lazyByIdDesc()
```


driesvints:
Changing a method signature is a breaking change. Please send this to master instead.

PR #39628, opened 2021-11-16
[8.x] Update PusherBroadcaster to allow use of the auth batching plugin

hjbdev:
I ran into an issue where a page had many calls to `/broadcasting/auth` which was delaying other ajax calls on the page. [In the readme of the pusher-js library](https://github.com/pusher/pusher-js#batching-auth-requests-aka-multi-auth) there's a link to [pusher-js-auth](https://github.com/dirkbonhomme/pusher-js-auth) which batches auth calls together.

Rather than returning a response with either 200 or 403 status codes, you instead return a format like this:

```
{
    "private-a": {
        "status": 200, // HTTP status codes, optional on success
        "data": {
            "auth": "xxxxxx:xxxxxxxxxxxxx"
        }
    },
    "private-b": {
        "status": 200,
        "data": {
            "auth": "xxxxxx:xxxxxxxxxxxxx"
        }
    },
    "private-c": {
        "status": 403
    }
}
```

I could do with some help on this though. My implementation seems to work when I test in my application, but I think there's something going on with Mockery that I don't understand. When running the tests, `PusherBroadcaster.php:68` throws an exception because `$authResponse` is null.

Also I think my implementation could be cleaner but it could require refactoring of a method in the parent class, so I wanted to get input before proceeding further.

Could also be that this is better placed in a package rather than as part of the main framework.

Thanks!

### DIFF
--- a/src/Illuminate/Broadcasting/Broadcasters/PusherBroadcaster.php
+++ b/src/Illuminate/Broadcasting/Broadcasters/PusherBroadcaster.php
@@ -41,16 +41,56 @@ class PusherBroadcaster extends Broadcaster
      */
     public function auth($request)
     {
-        $channelName = $this->normalizeChannelName($request->channel_name);
-
-        if (empty($request->channel_name) ||
-            ($this->isGuardedChannel($request->channel_name) &&
-            ! $this->retrieveUser($request, $channelName))) {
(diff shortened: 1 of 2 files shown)


driesvints:
This seems to break the test suite. Please fix the tests and mark as ready for review.

PR #39632, opened 2021-11-16
[8.x] Adds support for `pusher/pusher-php-server: ^7.0`

nunomaduro:
This pull request adds support for `pusher/pusher-php-server: ^7.0`. Note that, at this point, we are using a few [deprecated methods](https://github.com/pusher/pusher-http-php/blob/master/CHANGELOG.md), such as `$this->pusher->socket_auth(...`, etc. They will still work for now, but i guess they will be removed on `^8.0`.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -148,7 +148,7 @@
         "phpunit/phpunit": "Required to use assertions and run tests (^8.5.19|^9.5.8).",
         "predis/predis": "Required to use the predis connector (^1.1.9).",
         "psr/http-message": "Required to allow Storage::put to accept a StreamInterface (^1.0).",
-        "pusher/pusher-php-server": "Required to use the Pusher broadcast driver (^4.0|^5.0|^6.0).",
+        "pusher/pusher-php-server": "Required to use the Pusher broadcast driver (^4.0|^5.0|^6.0|^7.0).",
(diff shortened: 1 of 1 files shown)


driesvints:
@nunomaduro maybe we can already drop/remove some of them in Laravel 9?

nunomaduro:
@driesvints Agree. @taylorotwell would you be ok having make Laravel 9 "pusher-php-server: ^7.0" only? There is no reason why would keep supporting those old versions, and is a good moment to add a note to the upgrade guide.

composer.json
```diff
@@ -148,7 +148,7 @@
         "phpunit/phpunit": "Required to use assertions and run tests (^8.5.19|^9.5.8).",
         "predis/predis": "Required to use the predis connector (^1.1.9).",
         "psr/http-message": "Required to allow Storage::put to accept a StreamInterface (^1.0).",
-        "pusher/pusher-php-server": "Required to use the Pusher broadcast driver (^4.0|^5.0|^6.0).",
+        "pusher/pusher-php-server": "Required to use the Pusher broadcast driver (^4.0|^5.0|^6.0|^7.0).",
```
X-Coder264, line 151: This change should also be made in the composer.json of the broadcasting component.

PR #39634, opened 2021-11-16
[8.x] Use parents to resolve middleware priority

netpok:
This PR allows using parents of the middleware as fallback to resolve the middleware priority. This allows overriding system middlewares while it keeps the correct ordering.

Also this fixes a bug where Jetstream overrides the AuthenticateSession middleware but since the priority was lost, a password update will log out the user if it was authenticated on a non-default guard.


taylorotwell:

Thanks!

PR #34758, opened 2020-10-08
Allow `array` and `Countable` on `Str::plural()` and `Str::pluralStudly()`

squatto:
`Str::plural()` and `Str::pluralStudly()` currently require the count to be an `int`, which forces you to call `count()` on `array` and `Countable`. This PR allows you to pass the `array` or `Countable` itself to the method and it will count them for you.

### `array`

```php
$items = ['item1', 'item2', 'item3'];

// before:
echo Str::plural('item', count($items));

// after:
echo Str::plural('item', $items);
```

### `Countable`

```php
$items = collect(['item1', 'item2', 'item3']);

// before:
echo Str::plural('item', $items->count());

// after:
echo Str::plural('item', $items);
```

The tests check with an `array`, `Collection`, and `MessageBag`

PR #39641, opened 2021-11-17
[8.x] Add support for countables to the pluralizer

michaeldyrynda:
This PR adds support for countable values to the `Str::plural` method.

In doing so, you will be able to pass anything that passes PHP's `is_countable` check i.e. arrays, objects that implement `Countable` as the second parameter of the `plural` method.

This is useful when you want to pluralize a string based on a collection returned directly from an Eloquent query or otherwise filtered collection.

### DIFF
--- a/src/Illuminate/Support/Pluralizer.php
+++ b/src/Illuminate/Support/Pluralizer.php
@@ -64,7 +64,7 @@ class Pluralizer
      * Get the plural form of an English word.
      *
      * @param  string  $value
-     * @param  int|Countable  $count
+     * @param  int|array|Countable  $count
      * @return string
      */
     public static function plural($value, $count = 2)
--- a/src/Illuminate/Support/Pluralizer.php
+++ b/src/Illuminate/Support/Pluralizer.php
@@ -64,7 +64,7 @@ class Pluralizer
      * Get the plural form of an English word.
      *
      * @param  string  $value
-     * @param  int|array|Countable  $count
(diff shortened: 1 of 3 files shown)


driesvints (approved):

src/Illuminate/Support/Pluralizer.php
```diff
@@ -64,11 +64,15 @@ class Pluralizer
      * Get the plural form of an English word.
      *
      * @param  string  $value
-     * @param  int  $count
+     * @param  int|array|Countable  $count
```
nunomaduro, line 67:
```suggestion
     * @param  int|array|\Countable  $count
```

src/Illuminate/Support/Str.php
```diff
@@ -521,7 +521,7 @@ public static function parseCallback($callback, $default = null)
      * Get the plural form of an English word.
      *
      * @param  string  $value
-     * @param  int  $count
+     * @param  int|array|Countable  $count
```
nunomaduro, line 524:
```suggestion
     * @param  int|array|\Countable  $count
```

nunomaduro (changes requested):
The pull request looks good to me. Left some feedback regarding the PHP annotations.

nunomaduro (approved):

squatto:
I'm happy to see this make it's way into the framework, it's been needed for a while. I tried adding the same thing but it was declined for some reason: https://github.com/laravel/framework/pull/34758

PR #39642, opened 2021-11-17
[9.x] Use Relative View Path Name When Hashing Compiled Views

bogdankharchenko:
### Issue

We are not able to "pre-compile" views on say a local machine/build box, and push it into prod "pre-compiled". This is because the filename of the cached view is a sha1 of the absolute path of the file. Meaning, if the directory is different on which the view was initially compiled, then the view is then recompiled. 

This PR introduces a new `basePath` param into the the Compiler - which then allows us to take the absolute path of the view, and remove the base path from it - which is then used as the "name" before being hashed

Targeted into 9.x as this is a possible breaking change.

Driver behind this was during a vapor deploy, we can pre-compile the views (locally) into bootstrap/cache/views and serve them already compiled. Which is great for a read-only filesystem.

### DIFF
--- a/src/Illuminate/View/Compilers/Compiler.php
+++ b/src/Illuminate/View/Compilers/Compiler.php
@@ -38,7 +38,7 @@ abstract class Compiler
      *
      * @throws \InvalidArgumentException
      */
-    public function __construct(Filesystem $files, $cachePath, $basePath)
+    public function __construct(Filesystem $files, $cachePath, $basePath = '')
     {
         if (! $cachePath) {
             throw new InvalidArgumentException('Please provide a valid cache path.');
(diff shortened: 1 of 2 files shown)


dennisprudlo:
Why not simply call `view:cache` as part of the deployment process?

bogdankharchenko:
@dennisprudlo yes I already do that. But this will allow for the ability to pre-compile views before packaging the app up in a container or deploying to read only file systems

inxilpro:
I could definitely also see using this in a CI pipeline to pre-compile views before deploy.

taylorotwell:
Please detail the breaking changes.

taylorotwell:
I placed this behind a new `view.relative_hash` configuration option to minimize risk of breakage. Defaults to `false`.

PR #39644, opened 2021-11-17
[6.x] Fixes Query Builder to only cast integer when given other than `null`

crynobone:
The regression bug was introduced in [v6.20.39](https://github.com/laravel/framework/releases/tag/v6.20.39) and [v8.71.0](https://github.com/laravel/framework/releases/tag/v8.71.0)

Fixes laravel/nova-issues#3637


GrahamCampbell (approved):
Good catch.

PR #39645, opened 2021-11-17
[8.x] Fix \Illuminate\Support\Facades\App::storagePath() method ignores $path argument

afilippov1985:
`\Illuminate\Support\Facades\App::storagePath()` method ignores `$path` argument

This fix makes `App::storagePath()` method and `storage_path` helper return same result when `$path` is passed


driesvints:
Changing a method isgnature is a breaking change unfortunately. Please send this to master.

PR #39646, opened 2021-11-17
[8.x] Implement lazyById in descending order

moshe-autoleadstar:
Sometimes, when dealing with a lot of data, you need to lazily loop through models.
However, often, one would care more about the "newer" objects being processed first.

```
$query->lazyByIdDesc()
```


ronapelbaum2 (commented):
💪

GrahamCampbell:
Don't we need additional tests here to ensure it is actually lazy? cc @JosephSilber

PR #39647, opened 2021-11-17
[8.x] Use parents to resolve middleware priority

netpok:
Currently when middlewares are ordered it uses the class name and its interfaces for sorting but not it's parents.

This is a problem for example in Jetstream because:
- Jetstream extends the `AuthenticateSession` middleware with it's own.
- Since it doesn't have a contract it will run before any `Authenticate` middleware
- Resulting in it's always being initialized via the default guard instead of the authenticated one
- This prevents using two guards simultaniously (for example a user and an admin guard)

This PR fixes this behaviour by using parent classes for ordering so the Jetstream `AuthenticateSession` middleware will be in the correct place because it extends the Laravel `AuthenticateSession` middleware.

I also added test for the parent sorting and the already present interface sorting.

### DIFF
--- a/src/Illuminate/Routing/SortedMiddleware.php
+++ b/src/Illuminate/Routing/SortedMiddleware.php
@@ -101,6 +101,14 @@ class SortedMiddleware extends Collection
                 yield $interface;
             }
         }
+
+        $parents = @class_parents($stripped);
+
+        if ($parents !== false) {
+            foreach ($parents as $parent) {
+                yield $parent;
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


driesvints:
We're going to have to revert this: https://github.com/404labfr/laravel-impersonate/issues/145

netpok:
@driesvints I'm pretty sure the package has some bugs in it but probably it's best to revert it and I will target 9.x later if I have some time

driesvints:
@netpok as said in the other thread: it would be good if we could identify the bug itself so we know how this will impact others.

PR #39648, opened 2021-11-17
[9.x] Fix \Illuminate\Support\Facades\App::storagePath() method ignores $path argument

afilippov1985:
\Illuminate\Support\Facades\App::storagePath() method ignores $path argument

This fix makes App::storagePath() method and storage_path helper return same result when $path is passed

### DIFF
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -454,11 +454,13 @@ class Application extends Container implements ApplicationContract, CachesConfig
     /**
      * Get the path to the storage directory.
      *
+     * @param  string  $path
      * @return string
      */
-    public function storagePath()
+    public function storagePath($path = '')
     {
-        return $this->storagePath ?: $this->basePath.DIRECTORY_SEPARATOR.'storage';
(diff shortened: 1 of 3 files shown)


taylorotwell:
We're not able to add this argument on a patch release in case people are extending this method.

src/Illuminate/Foundation/Application.php
```diff
@@ -454,11 +454,13 @@ public function publicPath()
     /**
      * Get the path to the storage directory.
      *
+     * @param  string  $path
      * @return string
      */
-    public function storagePath()
+    public function storagePath($path = '')
```
nunomaduro, line 460: If we do this change, the interface implemented "the contract" by this class needs to be changed too. And it's a breaking change, that needs to be documented.

nunomaduro (changes requested):

src/Illuminate/Foundation/Application.php
GrahamCampbell: We can wait till 9.x to change the contract tbh?

nunomaduro: This is targeting already `master` @GrahamCampbell .

PR #38766, opened 2021-09-11
[8.x] Add `assertExists` testing method

RVxLab:
This PR adds the `assertExists` method for use in testing.

This method aims to make checking whether a model exists in the database easier, much like `assertDeleted` did for `assertDatabaseMissing`.

With this method, the following:

```php
public function testProductIsNotDeleted(): void
{
    $product = Product::find(1);

    // ...

    $this->assertDatabaseHas('products', [
        'id' => $product->getKey(),
    ]);
}
```

Could be rewritten to:

```php
public function testProductIsNotDeleted(): void
{
    $product = Product::find(1);

    // ...

    $this->assertExists($product);
}
```

PR #30648, opened 2019-11-21
[6.x] Add assertDeleted for database testing

jasonmccreary:
While there are numerous ways to test a record has been deleted, all are relatively lengthy to assert and don't communicate well. 

**Before**
```php
$this->assertDatabaseMissing('posts', [
    'id' => $post->id
]);
```

There is an `assertSoftDeleted` which can actually be passed a _model_. Upon doing so, it conveniently builds the appropriate expectations.

An `assertDeleted` method is symmetrical with this existing method and improves the developer experience by wrapping all the same behavior of `assertDatabaseMissing` in an intently named method.

**After**
```php
$this->assertDeleted($post);
```

PR #39661, opened 2021-11-17
[9.x] Remove `assertDeleted`

jasonmccreary:
This removes the asymmetrical, proxy assertion `assertDeleted`. Unlike other assertions, its opposite assertion does not exist (`assertNotDeleted`). It is also superseded in #38766 by `assertModelExists` and `assertModelMissing`.

Tests using this assertion may either adopt one of these more communicative methods, or fallback to the original `assertDatabaseMissing`.

Targeting Laravel 9 as this is a breaking change. Alternatively, I am glad to open a PR to add `assertNotDeleted` to provide symmetry. I believe removal is cleaner though as all paths should be covered by the other 4 assertions.

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php
@@ -69,25 +69,6 @@ trait InteractsWithDatabase
     }
 
     /**
-     * Assert the given record has been deleted.
-     *
-     * @param  \Illuminate\Database\Eloquent\Model|string  $table
-     * @param  array  $data
-     * @param  string|null  $connection
-     * @return $this
-     */
-    protected function assertDeleted($table, array $data = [], $connection = null)
-    {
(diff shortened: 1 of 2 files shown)


taylorotwell:
What are the current differences between `assertDeleted` and `assertModelMissing`?

jasonmccreary:
@taylorotwell, when passed a model, none. Then both call assertDatabaseMissing underneath ([ref](https://github.com/laravel/framework/blob/master/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php#L82), [ref](https://github.com/laravel/framework/blob/master/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php#L157-L160))

I originally added `assertDeleted` (https://github.com/laravel/framework/pull/30648). For the life of me, I'm not sure why I made it a proxy method and didn't add its counterpart. 😅 

As you can see from the example, it was intended to do what `assertModelExists` and `assertModelMissing` ultimately did.

PR #39668, opened 2021-11-18
set read or write type when select

bearzlh:
I want to know the read or write type from QueryExecuted event when i dont set ::read or ::write after connection name,because not all model use R/W Splitting


GrahamCampbell:
This breaks the tests.

PR #39672, opened 2021-11-18
[8.x] Add Support for custom Increment / Decrement in Collection range

ManojKiranA:
By default [php range](https://www.php.net/manual/en/function.range.php) function will accept 3rd argument for `increment/decrement` between elements in the sequence. 

Since [`Collection::range`](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Collections/Collection.php#L43) function uses native php function I have added support for 3rd argument.

Before:

`collect(range(2,10,2));`

After:

`collect()->range(2,10,2);`

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -34,13 +34,14 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     /**
      * Create a collection with the given range.
      *
-     * @param  int  $from
-     * @param  int  $to
+     * @param  int|float  $from
+     * @param  int|float  $to
+     * @param  int|float  $step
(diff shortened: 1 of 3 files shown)


src/Illuminate/Collections/Collection.php
```diff
@@ -34,13 +34,14 @@ public function __construct($items = [])
     /**
      * Create a collection with the given range.
      *
-     * @param  int  $from
-     * @param  int  $to
+     * @param  int|float  $from
```
GrahamCampbell, line 37: 👎 for allowing floats here. What was the use case?

GrahamCampbell (changes requested):

driesvints:
This PR contains breaking changes to the method signature.

PR #39673, opened 2021-11-18
[8.x] Fix assertSoftDeleted & assertNotSoftDeleted

### LINKED ISSUE #39667: assertSoftDeleted() and assertNotSoftDeleted() discard $data argument value when first argument is a model
- Laravel Version: 8.71.0
- PHP Version: 8.0.2

### Description:
The testing helpers `assertSoftDeleted()` and `assertNotSoftDeleted()` both accept either a table name provided as a string, or a model that implements the `SoftDeletes` trait. When a model is provided these functions automatically filter the query to the table for the model, as well as looking for the record with the provided primary key.

However, when a model is provided any values provided in the second argument, `$data`, are discarded when the function recursively calls itself. As a result, what appears to be a valid test case is actually not executed with the expected behavior:

```
// Assume a database table (example_table) has the following record:
id  | col_a | col_b | deleted_at
1   | foo     | null     | 2021-11-16 00:00:00

// Assume we have a "$model" variable that exists for this record.
// Now, in a test we add the following assertion:

$this->assertSoftDeleted($model, ['col_b' => 'bar']);
```

The example above will pass, even though the database does _not_ contain a record for that model where `col_b` is equal to `bar`. However, if you give that assertion the following code, it _does_ fail as expected:

```
$this->assertSoftDeleted('example_table', ['id' => 1, 'col_b' => 'bar']);
```

The code responsible for these functions is here:
https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php#L90-L132

The specific chunk responsible for dropping the `$data` parameter is here:
https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php#L123-L125

driesvints:
This PR fixes an issue where `$data` wasn't passed along properly as a where clause for the `assertSoftDeleted` & `assertNotSoftDeleted` methods. By passing the data properly, more granular assertions can be made.

Fixes #39667

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php
@@ -121,7 +121,12 @@ trait InteractsWithDatabase
     protected function assertNotSoftDeleted($table, array $data = [], $connection = null, $deletedAtColumn = 'deleted_at')
     {
         if ($this->isSoftDeletableModel($table)) {
-            return $this->assertNotSoftDeleted($table->getTable(), [$table->getKeyName() => $table->getKey()], $table->getConnectionName(), $table->getDeletedAtColumn());
+            return $this->assertNotSoftDeleted(
+                $table->getTable(),
+                array_merge($data, [$table->getKeyName() => $table->getKey()]),
+                $table->getConnectionName(),
+                $table->getDeletedAtColumn()
+            );
         }
 
         $this->assertThat(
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Is this an issue on 6.x too?

driesvints:
@GrahamCampbell no, the method doesn't exists there.

GrahamCampbell (approved):
🚢

PR #39677, opened 2021-11-18
[8.x] Added Prohibits validation rule to dependentRules property

dees040:
The prohibits validation rule depends on other fields. However the rule was not added to the `$dependentRules` property.


GrahamCampbell:
Could you add a unit test please?

dees040:
@GrahamCampbell of course. I've just done that.

PR #39678, opened 2021-11-18
[8.x] Allow a Closure to be passed as a ttl in Cache remember() method

gerardnll:
This PR simplifies the code needed in case the $ttl is calculated each time we have to store a value. I simply added the value() function for the $ttl parameter.

In the following code i want to know how many events the User has hosted so I'm calculating the ttl taking in account when is the next hosted event for that user ending, so the counter will be recalculated because the stored cached value will be expired by then.

```
function countEventsHostedCached($user) {
    if (Cache::has('count_events_hosted')) {
        return Cache::get('count_events_hosted', 0);
    }

    $closestEventEndsAt = $user->hostedEvents()
        ->select('ends_at')
        ->where('starts_at', '>', Date::now())
        ->orderBy('ends_at')
        ->limit(1)
        ->value('ends_at');

    $value = $user->hostedEvents()->ended()->count();

    Cache::put(
        'count_events_hosted',
        $value,
        // Expire when closest event ends
        $closestEventEndsAt ? Date::parse($closestEventEndsAt) : 60
    );

    return $value;
}
```

Turns into:

```
function countEventsHostedCached($user) {
    return Cache::remember(
        'count_events_hosted',
        function () use ($user) {
            // Expire when closest event ends
            $closestEventEndsAt = $user->hostedEvents()
                ->select('ends_at')
                ->where('starts_at', '>', Date::now())
                ->orderBy('ends_at')
                ->limit(1)
                ->value('ends_at');

            return $closestEventEndsAt ? Date::parse($closestEventEndsAt) : 60;
        },
        function () use ($user) {
            return $user->hostedEvents()->ended()->count();
        }
    );
} 
```

### DIFF
--- a/src/Illuminate/Cache/Repository.php
+++ b/src/Illuminate/Cache/Repository.php
@@ -373,7 +373,7 @@ class Repository implements ArrayAccess, CacheContract
      * Get an item from the cache, or execute the given Closure and store the result.
      *
      * @param  string  $key
-     * @param  \DateTimeInterface|\DateInterval|int|null  $ttl
+     * @param  \Closure|\DateTimeInterface|\DateInterval|int|null  $ttl
      * @param  \Closure  $callback
      * @return mixed
      */
@@ -388,7 +388,7 @@ class Repository implements ArrayAccess, CacheContract
             return $value;
         }
 
-        $this->put($key, $value = $callback(), $ttl);
+        $this->put($key, $value = $callback(), value($ttl));
 
         return $value;
     }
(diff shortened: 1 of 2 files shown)


src/Illuminate/Cache/Repository.php
```diff
@@ -373,7 +373,7 @@ public function forever($key, $value)
      * Get an item from the cache, or execute the given Closure and store the result.
      *
      * @param  string  $key
-     * @param  \DateTimeInterface|\DateInterval|int|null  $ttl
+     * @param  \Closure|\DateTimeInterface|\DateInterval|int|null  $ttl
```
GrahamCampbell, line 376: PR description says callable, but this is not a callable. this is more specific, a closure, not all callables are closures. all closures are callable.

gerardnll: Excuse my wording. I guess Closure is more appropriate.

PR #39680, opened 2021-11-18
Reattempt of #39672

ManojKiranA:
Reattempt of #39672 without changing method signature

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -34,8 +34,8 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     /**
      * Create a collection with the given range.
      *
-     * @param  int|float  $from
-     * @param  int|float  $to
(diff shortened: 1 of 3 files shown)


src/Illuminate/Collections/Collection.php
```diff
@@ -36,11 +36,12 @@ public function __construct($items = [])
      *
      * @param  int  $from
      * @param  int  $to
+     * @param  int|float  $step
      * @return static
      */
-    public static function range($from, $to)
+    public static function range($from, $to, $step = 1)
```
driesvints, line 42: The method signature is changed again on these methods.

ManojKiranA: can I target it to 9.x

ManojKiranA: can I target it to 9.x

driesvints: You can try

PR #39687, opened 2021-11-18
[8.x] Add Str::onlyNumbers()

Dahmer49:
### Description
This PR adds a new `Stringable` method to get only the numbers from a string.

### Purpose
There's not a Laravel way to get only the numbers from a string, just using `regex` or `filter_var`

### Use Case

```
// Displays: "123456"
echo Str::onlyNumbers("foobar123456");
```

```
// Displays: "1256"
echo Str::onlyNumbers("U$ 12,56");
```

When no number is find in string will return `null`

### DIFF
--- a/tests/Support/SupportStrTest.php
+++ b/tests/Support/SupportStrTest.php
@@ -536,6 +536,15 @@ class SupportStrTest extends TestCase
         $this->assertSame('Мама мыла раму', Str::ucfirst('мама мыла раму'));
     }
 
+    public function testOnlyNumbers()
+    {
+        $this->assertSame(123, Str::onlyNumbers("f1o2o3"));
+        $this->assertSame(45612, Str::onlyNumbers("U$ 456.12"));
+        $this->assertSame(9870, Str::onlyNumbers("foo987bar ,.0-."));
+        $this->assertSame(0, Str::onlyNumbers("0foo"));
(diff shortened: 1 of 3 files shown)


tests/Support/SupportStrTest.php
```diff
@@ -536,6 +536,15 @@ public function testUcfirst()
         $this->assertSame('Мама мыла раму', Str::ucfirst('мама мыла раму'));
     }
 
+    public function testOnlyNumbers()
+    {
+        $this->assertSame(123, Str::onlyNumbers("f1o2o3"));
+        $this->assertSame(45612, Str::onlyNumbers("U$ 456.12"));
```
crynobone, line 542: I would expect we would get `456.12` instead of `45612`

powellblyth: also this would produce `143` for `Your balance is £-143`
and It would beautifully mangle exponential numbers.
and doesn't consider localised decimals  https://www.languageediting.com/format-numbers-eu-vs-us/


Perhaps the function should be called `onlyDigits`?  or `Str::digits` ?

taylorotwell:

tests/Support/SupportStrTest.php
Dahmer49: @powellblyth
You're right, I will change the function name to `Str::digits`, this will let the function more readable

PR #39688, opened 2021-11-18
[8.x] Added ignorable functionality to migration files

binarygeotech:
### Description
This PR adds the ability to ignore migration files either by setting the `ignore` property to true or returning a truthy evaluation in the `ignore()` method of the migration file.

### Purpose
This will enable developers to ignore migration files that are not ready/needed to be migrated yet at the time of running the migrate command. There are some occasions that we needed to ignore some files from been migrated but had to move the files out of the migrations directory.

### DIFF
--- a/src/Illuminate/Database/Migrations/Migrator.php
+++ b/src/Illuminate/Database/Migrations/Migrator.php
@@ -106,12 +106,24 @@ class Migrator
             $files, $this->repository->getRan()
         ));
 
+        // This section filters ignorable migrations out from pending migration files.
+        $runnableMigrations = Collection::make($migrations)->reject(function($file) {
+            $migration = $this->resolve(
+                $this->getMigrationName($file)
+            );
+
+            $hasIgnoreProperty = property_exists($migration, 'ignore') && $migration->ignore;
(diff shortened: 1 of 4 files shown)


emmanuelizzy:
+1

LastDragon-ru:
> There are some occasions that we needed to ignore some files from been migrated but had to move the files out of the migrations directory.

You can redefine Migrator in service provider and implement any logic to skip migrations, please see as example https://github.com/LastDragon-ru/lara-asp/blob/febb6bb3f5cc524b0bec4694349553a1c33fc961/packages/migrator/src/Extenders/SmartMigrator.php#L29

binarygeotech:
@LastDragon-ru, I think it's best for non-advanced users to be able to ignore some migrations files without going the advanced way of extending the `Migrator` class. Moreover, ignoring migration files might be on temporary basis, providing an out-of-the-box simpler method is good approach.

taylorotwell:

LastDragon-ru:
@binarygeotech from my point of view non-advanced users should not ignore migrations because it makes app code more complicated eg you never know exists a table or not, etc.

Another possible solution - move migrations into a separate package. After that, they can be ignored in the provider based on the conditions. Probably you can check how Telescope does it.

PR #39697, opened 2021-11-19
set read or write type when select

bearzlh:
I want to know the read or write type from QueryExecuted event when i dont set ::read or ::write after connection name,because not all model use R/W Splitting


driesvints:
Seems this is breaking tests.

bearzlh:
I only change the property of readWriteType in Connection,adding one line.
I dont know where breaking the tests.
I use pull request for the first time,Thanks for pointing it out for me

PR #39698, opened 2021-11-19
[8.x] New static method getTableName to get table name without create new instance

AhmedHelalAhmed:
This PR adds a new method to the model that get the table name for the model statically without creating new instance.

## Problem
We should create instance of model to be able to use getTable method that may take unneeded memory when the model contains a lot of logic or appends.
```  (new \App\Models\User)->getTable(); ```

## My Proposed Solution
To encapsulate the logic of getting table name in new method that can be callable through class itself and then we can call it statically and reduce the memory needed for instance.

## Usage 

```  \App\Models\User::getTableName(); ```

## Breaking changes
It shouldn't break any existing functionality, it's just encapsulate existing logic making new reusable method.

Any other suggestions for improvement are welcome.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1685,13 +1685,23 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     }
 
     /**
+     * Get the table name associated with the model.
+     *
+     * @return string
+     */
+    public static function getTableName()
+    {
+        return Str::snake(Str::pluralStudly(class_basename(get_called_class())));
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


driesvints:
Hah, quite a coincidence seeing this PR and @freekmurze's tweet a little earlier: https://twitter.com/freekmurze/status/1461342728613076994

crynobone:
This would only work until you define table name directly using `protected $table = 'users';`

driesvints:
Yeah, @crynobone. This isn't a valid implementation unfortunately.

AhmedHelalAhmed:
@crynobone @driesvints  this PR Fix mentioned issue https://github.com/laravel/framework/pull/39701
Thank you

PR #39700, opened 2021-11-19
[9.x] Add getTableName() on models

chrysanthos:
Replaces https://github.com/laravel/framework/pull/39698

Targeting 9.x mainly because I think that some people may already have this helper in their models (or at least I have).

The nice thing about this is that now, you will be able to use this in your tests like so instead of referencing the table name directly. 

```
$this->assertDatabaseHas(Category::getTableName(), [
     'id'   => $id,
     'name' => 'Abc'
])
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1685,6 +1685,16 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     }
 
     /**
+     * Get the table name for the model.
+     *
+     * @return string
+     */
+    public static function getTableName()
+    {
+        return (new static)->getTable();
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


faissaloux:
In your tests you can use ```model::class``` instead of table name

src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -1684,6 +1684,16 @@ public static function unsetConnectionResolver()
...
     }
 
+    /**
+     * Get the table name for the model.
+     *
+     * @return string
+     */
+    public static function getTableName()
+    {
+        return (new static)->getTable();
```
GrahamCampbell, line 1694: Seems like changing getTable to be static instead would be the better solution in 9.x.

GrahamCampbell (changes requested):

PR #39701, opened 2021-11-19
[8.x] New static method getDefaultTableName to get table default name without create new instance

AhmedHelalAhmed:
This PR adds a new method to the model that get the default table name for the model statically without creating new instance.

## Problem
We should create instance of model to be able to use getTable method that may take unneeded memory when the model contains a lot of logic or appends.
```  (new \App\Models\User)->getTable(); ```

## My Proposed Solution
To encapsulate the logic of getting table name in new method that can be callable through class itself and then we can call it statically and reduce the memory needed for instance.

## Usage 

```  \App\Models\User::getDefaultTableName(); ```

## Breaking changes
It shouldn't break any existing functionality, it's just encapsulate existing logic making new reusable method.

## Note
I have Fixed issue mentioned in https://github.com/laravel/framework/pull/39698 that if we use setTable.

Any other suggestions for improvement are welcome.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1687,7 +1687,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     /**
      * Get the default table name associated with the model.
      *
-     * @param self|null  $class
+     * @param string|null $class
      * @return string
      */
     public static function getDefaultTableName($class = null)
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1685,13 +1685,14 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     }
 
     /**
-     * Get the table name associated with the model.
+     * Get the default table name associated with the model.
      *
+     * @param self|null $class
      * @return string
      */
-    public static function getTableName()
+    public static function getDefaultTableName($class = null)
     {
-        return Str::snake(Str::pluralStudly(class_basename(get_called_class())));
+        return Str::snake(Str::pluralStudly(class_basename($class ?? get_called_class())));
     }
 
     /**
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/Illuminate/Database/Eloquent/Model.php [hunk 1]
```diff
@@ -1684,14 +1684,25 @@ public static function unsetConnectionResolver()
         static::$resolver = null;
     }
 
+    /**
+     * Get the default table name associated with the model.
+     *
+     * @param self|null $class
```
GrahamCampbell: this phpdoc is wrong. this should say `class-string<self>|null` or just `string|null`.

GrahamCampbell (changes requested):

src/Illuminate/Database/Eloquent/Model.php
AhmedHelalAhmed: Thank you. I modified it

src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -1687,7 +1687,7 @@ public static function unsetConnectionResolver()
     /**
      * Get the default table name associated with the model.
      *
-     * @param self|null $class
+     * @param self|null  $class
```
GrahamCampbell: incorrect phpdoc. should be `class-string<self>|null` or just `string|null`

GrahamCampbell (changes requested):

src/Illuminate/Database/Eloquent/Model.php
AhmedHelalAhmed: I modified it this is old version kindly check this https://github.com/laravel/framework/pull/39701/files

AhmedHelalAhmed: image

taylorotwell:

AhmedHelalAhmed:
@taylorotwell @GrahamCampbell 
Thanks for your time

rogervila:
Could this PR be revised to be implemented for Laravel 9?

PR #39702, opened 2021-11-19
[6.x] Only allow PHP 8.0.x versions

canvural:
Looks like the PHP 8.1 compatibility PR is not merged into 6.x So it doesn't make sense to allow installing 6.x on PHP 8.1

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "^7.2.5|^8.0",
+        "php": "^7.2.5|~8.0.0",
         "ext-json": "*",
         "ext-mbstring": "*",
         "ext-openssl": "*",


driesvints:
This has no effect. Older Laravel 6 versions will still be installed on php 8. We also made the choice to not adopt this syntax for versioning.

canvural:
It has an effect. It will disallow installing 6.x on PHP 8.1. Now, you can install it but at runtime your project will fail.

GrahamCampbell:
> It has an effect. It will disallow installing 6.x on PHP 8.1. Now, you can install it but at runtime your project will fail.

No it does not. Composer will just resolve the version of Laravel 6 before you added this restriction. All this does is lock people of of security updates, so no benefit.

PR #39703, opened 2021-11-19
New hasManyThroughPivot table relation

chrisjk123:
This PR adds a new eloquent relationship to the pivot model that allows for many relations through the given pivot table

## Problem
I have tried using `hasManyThrough()` in order to get many permissions through another pivot table. However, since my pivot table has no primary ID column this becomes difficult. Below is the following ERD database structure for the given scenario:

```
users
- id
- name
- password 
etc...

team_user
- team_id
- user_id
- created_at
- updated_at

team_user_permissions
- team_id
- user_id
- permission_id
- created_at
- updated_at

permissions
- id
- name
- created_at
- updated_at
```

## My Proposed Solution
Create a new eloquent relationship where you can specify the given column names to match against the related table through another pivot table.

```
<?php

namespace App\Models;

use App\Relations\CustomRelations;

class TeamUser extends Pivot
{
    use CustomRelations;

    /**
     * Has many Permission through the pivot TeamUserPermission table
     */
    public function permissions()
    {
        return $this->hasManyThroughPivot(
            Permission::class,
            TeamUserPermission::class,
            'user_id', // team_user.user_id - local key
            'user_id', // team_user_permissions.user_id - foreign related key
            // 'permission_id', // team_user_permissions.permission_id - far foreign related key
            // 'id', // permissions.id - related key
        );
    }
}
```

## Breaking changes
This shouldn't break anything, it's a completely new suggested relationship which you can load custom relationships through a pivot table.

## Note
Any other suggestions for are definitely welcome! 👍

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
@@ -11,6 +11,7 @@ use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\HasMany;
 use Illuminate\Database\Eloquent\Relations\HasManyThrough;
+use Illuminate\Database\Eloquent\Relations\HasManyThroughPivot;
 use Illuminate\Database\Eloquent\Relations\HasOne;
 use Illuminate\Database\Eloquent\Relations\HasOneThrough;
 use Illuminate\Database\Eloquent\Relations\MorphMany;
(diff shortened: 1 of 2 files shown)


taylorotwell:

chrisjk123:
Hey @taylorotwell, 

No worries at all, completely understand.

I do believe this solves some highly complicated relations involving pivot tables only whereby the given table does not have a primary key so if you do get the opportunity to take a further look then please do. For now, I will release this as a package next week.

neilgilmour:
Hi @chrisjk123, I'm genuinely interested to learn here - why not add in the `id` field to the pivot and then use a pivot model?

Or have you seen Staudenmeir's [packages](https://github.com/staudenmeir)?

PR #39704, opened 2021-11-19
[8.x] Make view component namespace configurable

heyjoe1984:
these change will allow us to configure the view component namespace via the `view` config file

e.g inside `config/view.php`: 

```
'component_namespace' => 'Http\\View\\Components',
```

it falls back on the current namespace so there should be 0 conflicts.

### DIFF
--- a/src/Illuminate/View/Compilers/BladeCompiler.php
+++ b/src/Illuminate/View/Compilers/BladeCompiler.php
@@ -566,7 +566,7 @@ class BladeCompiler extends Compiler implements CompilerInterface
 
         if (is_null($alias)) {
             $namespace = config('view.component_namespace', 'View\\Components');
-            
+
             $alias = Str::contains($class, '\\'.$namespace.'\\')
                             ? collect(explode('\\', Str::after($class, '\\'.$namespace.'\\')))->map(function ($segment) {
                                 return Str::kebab($segment);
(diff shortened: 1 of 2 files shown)


driesvints:
I don't think it's a good idea to make this configurable. If you already have a components directory, I suggest you rename that to something else.

src/Illuminate/View/Compilers/BladeCompiler.php
```diff
@@ -565,8 +565,10 @@ public function component($class, $alias = null, $prefix = '')
         }
 
         if (is_null($alias)) {
-            $alias = Str::contains($class, '\\View\\Components\\')
-                            ? collect(explode('\\', Str::after($class, '\\View\\Components\\')))->map(function ($segment) {
+            $namespace = config('view.component_namespace', 'View\\Components');
```
GrahamCampbell, line 568: You cannot call that function here.

GrahamCampbell (changes requested):

heyjoe1984:
> I don't think it's a good idea to make this configurable. If you already have a components directory, I suggest you rename that to something else.

Why isn't it a good idea to make it configurable? We can already configure things like the model directory.

The reason I'm submitting this is so I can move the `View/Components` folder into a different folder.

src/Illuminate/View/Compilers/BladeCompiler.php
heyjoe1984: Does this compiler fire before the config values are processed?

taylorotwell:

PR #39708, opened 2021-11-20
[8.x] Add array create helper method

Ivanshamir:
In here, I have created new helper method: `Arr::create($keys, $values = null, $join = false)` .  This method is for if we want to create new array from string which contains dot(.) or without dot. Also it can join two arrays if we pass additional param `$join=true`. In here the `$keys` parameter is required and it should be string. Example:
1. If we pass only one parameter with dot(.) notation
```
    $string = "first.second.third";
    $result = Arr::create($string);

    //result
    [
      0 => "first"
      1 => "second"
      2 => "third"
    ]
```
2.  If we pass first parameter with dot(.) notation and second parameter of single string then the second parameter will be added in the most depth key of array.:
```
    $items = "first.second.third";
    $values = "john";
    $result = Arr::create($items, $values);

    //result
      [
        "first" => [
          "second" => [
            "third" => "john"
          ]
        ]
      ]
```
If we pass string with dot(.) then the second parameter will be converted to array and it will be added into the most depth key of final array.
```
    $items = "first.second.third";
    $values = "john.doe";
    $result = Arr::create($items, $values);

    //result
      [
        "first" => [
          "second" => [
            "third" => [
                 0 => "john"
                 1 => "doe"
             ]
          ]
        ]
      ]
```
3. Finally if we want to join the two arrays, then pass `$join = true` in the helper function.
```
    $items = "first.second.third";
    $values = "john.doe";
    $result = Arr::create($items, $values, $join = true);

   //result
    [
      "first" => "john"
      "second" => "doe"
      "third" => null
    ]
```
**Difference between Arr::wrap() and Arr::create():** The `Arr::wrap()` method wraps a single string into array, where `Arr::create()` can create new array from string with/without dot(.) notation. Also it can join the two arrays.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -718,9 +718,9 @@ class Arr
     /**
      * Create array from strings.
      *
-     * @param string $keys
-     * @param string|null $values
-     * @param bool $join
+     * @param  string  $keys
+     * @param  string|null  $values
+     * @param  bool  $join
      * @return array
      */
     public static function create($keys, $values = null, $join = false)
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -714,4 +714,35 @@ class Arr
 
         return is_array($value) ? $value : [$value];
     }
+
+    /**
+     * Create array from strings.
+     *
+     * @param string $keys
+     * @param string|null $values
+     * @param bool $join
+     * @return array
+     */
+    public static function create($keys, $values = null, $join = false)
+    {
+        $result = [];
+
+        $keysArray = explode('.', $keys);
+
+        if (empty($values)) {
+            return $keysArray;
+        }
+
+        if ($join) {
+            $values = explode('.', $values);
+            foreach ($keysArray as $i => $segment) {
+                $result[$segment] = $values[$i] ?? null;
+            }
+            return $result;
+        }
+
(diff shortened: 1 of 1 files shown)


dennisprudlo:
I don't think it's necessary to have a helper method to wrap three different use cases and determine which use case to use by passing the correct set of parameters.

The first use case (`Array::create($someString)`) is just a split by a `.`. But the naming of the function is confusing then:
```php
Arr::create('github.com'); // [ 'github', 'com' ]
```
But why does it split my domain name? I would expect that a `create` function simply creates an array and I receive `[ 'github.com' ]`. The whole function kind of assumes that you want to work with string that is splitted by dots but `create` is a very generic name for that. Why not use these examples?
```php
explode('.', 'github.com');
Str::of('github.com')->explode('.'); // If you are ok with a collection as a return type
```

The second use case works with the dot notation as it is often used in laravel (well for the first parameter, the second is different again). The first parameter uses the dots to specify a new level in the array but in the second parameter the value is again splitted by the dots. The third use case uses the dot-splitting method for both parameters.

I think these are very niche utilities wrapped in a single function that doesn't need to be in the core.

Ivanshamir:
Hello @dennisprudlo you are absolutely correct. May be I need to rethink the `create` helper method. But first I need to clear the purpose. I want to create simple array from one string. If the string contains dot then the array will be a multidimensional array. Kind of opposite of `Arr::dot()` helper. What do you think? If the `create` method makes more confusion, I will close this PR.

taylorotwell:

PR #39709, opened 2021-11-20
[8.x] Add `modelKeys` method to Eloquent Builder

lupinitylabs:
Why?
====
There is a `modelKeys()` method which returns an array of the primary keys of all the models in an Eloquent Collection by calling:

```php
MyModel::get()->modelKeys();
```

This is favorable over hard coding the primary key name in the code like

```php
MyModel::get()->pluck('id');
```

However, we never use the code like this, but instead prefer to pluck directly on the Eloquent Builder:

```php
MyModel::pluck('id');
```

The advantage of this approach is that the resulting query with limit the results to the id column instead of needing to fetch all columns for all rows, which improves performance, limits traffic and saves memory.

However, there is no `modelKeys()` method available on the Builder.

So, in order to achieve the same result like with `modelKeys()`, you would need to do:

```php
MyModel::pluck(MyModel::getModel()->getKeyName());
```

...which is neither nice to write nor to read.

Secondly, I rarely encountered a situation where I need just all available primary keys - instead, we are having dozens of uses for preparing a list, for example for the options in select elements, where you need a display name keyed by the id:

```php
MyModel::pluck('name', 'id');
```

It would be cool to be able to just do

```php
MyModel::modelKeys('name');
```

in those cases and get an array or collection of all name values indexed by the model's primary key.

Let's do it
====

This PR adds a `modelKeys()` method to the Eloquent Builder that behaves much alike the `modelKeys()` method in the Eloquent Collection class, with the following exceptions:

1. it returns `static` instead of `array`, because it is using the pluck method and I found it makes sense it stays in line with pluck.
2. it accepts an optional `$column` parameter that indexes the given column values by the model's primary keys.

This PR also extends the existing Eloquent Collection's `modelKeys()` method by also accepting an optional `$column` parameter to replicate the behavior of the Builder's `modelKeys()` method.

Issues up for discussion
====

We run into an ugly situation here, because the existing `modelKeys()` method, unlike most other methods in the class, returns an array instead of an Eloquent Collection. 

In order not to introduce a breaking change, I left the return type to be an array.

I would leave up for discussion whether both the Collection AND Builder implementations should both return arrays. Currently, the Builder implementation stays in line with the `pluck` method it uses and returns an `Illuminate\Support\Collection`.

Personally, I would expect a collection return in order to be able to chain further method calls, and returning an array would in most cases lead to having to `collect()` the result again...

Let me know what you think!

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -779,6 +779,17 @@ class Builder
     }
 
     /**
+     * Get an array of primary keys, or the values of a given column, indexed by the primary keys.
+     *
+     * @param $column
+     * @return \Illuminate\Support\Collection
+     */
+    public function modelKeys($column = null)
+    {
+        return $this->pluck(...array_filter([$column, $this->model->getKeyName()]));
+    }
+
+    /**
      * Paginate the given query.
      *
      * @param  int|null  $perPage
--- a/src/Illuminate/Database/Eloquent/Collection.php
+++ b/src/Illuminate/Database/Eloquent/Collection.php
@@ -309,15 +309,19 @@ class Collection extends BaseCollection implements QueueableCollection
     }
 
     /**
-     * Get the array of primary keys.
+     * Get an array of primary keys, or the values of a given column, indexed by the primary keys.
      *
-     * @return array
+     * @return array|static
      */
-    public function modelKeys()
+    public function modelKeys($column = null)
     {
-        return array_map(function ($model) {
-            return $model->getKey();
-        }, $this->items);
+        if ($column === null) {
+            return $this->map->getKey()->toArray();
+        }
+
+        return $this->mapWithKeys(function ($model) use ($column) {
+            return [$model->getKey() => $model->getAttribute($column)];
+        });
     }
 
     /**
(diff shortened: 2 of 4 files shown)


devfrey:
`modelKeys()` on the Eloquent builder seems like a nice addition, but I'm not sure why you're changing the behaviour.

I also don't see the benefit of returning a collection from `Builder::modelKeys()`. Seems inconsistent with `Collection::modelKeys()`.

lupinitylabs:
> `modelKeys()` on the Eloquent builder seems like a nice addition, but I'm not sure why you're changing the behaviour.
> 
> I also don't see the benefit of returning a collection from `Builder::modelKeys()`. Seems inconsistent with `Collection::modelKeys()`.

The behavior doesn't change if used the way it was used before. However, neither me nor anyone I know ever used `modelKeys()` on its own (the old behavior), because you almost always want to match it with another column (name, title, ...). Using `modelKeys($column)` felt natural to me.

I see the inconsistency to `Collection::modelKeys()`, that's why I mentioned it in the description.
What made me ponder returning Collections is that `Collection::modelKeys()` returning an array feels inconsistent with most other Collection methods. There are only three other Collection methods returning an array: `getDictionary`, `getQueueableIds` and `getQueueableRelations`. `pluck` and `keys`, which are probably the most closely related methods, are both returning Collections.

I do see, however, that discussing the return type is not beneficial, because we certainly want to avoid making this a breaking change. So I am open to the idea of Eloquent returning an array as well.

taylorotwell:

PR #39710, opened 2021-11-20
[8.x] Add `elseisset` directive

faissaloux:
Currently we have only ```@isset``` and ```@endisset``` directives and we need ```@elseisset``` directive (that's the goal of this PR)

```
@isset($foo)
  //
@elseisset($bar)
  //
@endisset
```

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php
@@ -223,6 +223,17 @@ trait CompilesConditionals
     }
 
     /**
+     * Compile the else-isset statements into valid PHP.
+     *
+     * @param  string  $expression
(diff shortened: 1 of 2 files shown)


taylorotwell:
Just use @if and @else

faissaloux:
I just thought while ```@isset()``` exist we need a ```@elseisset()``` so we can use it with it 👀

faissaloux:
@taylorotwell I just thought while @isset() exist we need a @elseisset() so we can use it with it eyes

faissaloux:
@taylorotwell I just thought while @Isset() exist we need a @elseisset() so we can use it with it 👀

faissaloux:
@taylorotwell I just thought while ```@isset()``` exist we need a ```@elseisset()``` so we can use them together 👀

PR #39712, opened 2021-11-20
[8.x] Add `getDeeply` collection method

michael-rubel:
### Why?

Always when we want to get the data from a collection with multidimensional arrays, it was a way to use `pull` or multidimensional collections with `->get()->get()` which is not very elegant. The pull works well, but it actually removes the item from the collection besides retrieving. Most times working in Laravel I wanted to use dot notation, but to leave the item in the collection as other parts of the application might want access that data as well. The new method `getDeeply` works in a similar way as `pull` (using dot notation), but without removing an item from the collection.

### B/C?
- No. It's a completely new method, and under the hood, it's just an alias to `Arr::get()`.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -893,6 +893,18 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
+     * Get an item from the collection using "dot" notation.
+     *
+     * @param  mixed  $key
+     * @param  mixed  $default
+     * @return mixed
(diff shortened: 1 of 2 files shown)


foremtehan:
I always wonder why we cant use dot notation in `get()` method itself

taylorotwell:

michael-rubel:
If someone's interested in this function, it's released under the [`Spatie's Collection Macros`](https://github.com/spatie/laravel-collection-macros/pull/220) as `path` function. For me, this name is a bit confusing but @freekmurze decided to name it like this.

PR #39713, opened 2021-11-20
[8.x] Forget and clear a mocked/spied instance of an object in the container

khalilst:
This PR was inspired by `Jest`'s `mockRestore()` method.
It introduces new `forgetMock()` method which uses `forgetInstance()` method internally and has a better DX.

The use-case:
Consider we have multiple test methods using a `mocked/spied` instance of an object which is prepared in the `setUp()` method.
If in that class some tests want to run tests without mock, they can call the `forgetMock()` method.


driesvints:
Can't we just also call this forgetInstance to keep it the same and generic?

khalilst:
Just hard to guess without knowing what is going on behind the scene.

PR #39714, opened 2021-11-21
[8.x] Add integer() method to InteractsWithInput trait

BrendanKoral:
This PR is inspired by the `boolean()` method on the `InteractsWithInput` trait. I often find that I have to write something like this when reading numbers out of a request:

```
$myVal = (int)$request->query('myval');
```

This PR eliminates the typecasting by adding an `integer()` method similar to the `boolean()` method.

```
$myVal = $request->integer('myval');
```

If there's already an existing way to do something like this and I've overlooked it, please let me know. I didn't see anything like this in the documentation.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -303,7 +303,7 @@ trait InteractsWithInput
      *
      * @param  null  $key
      * @param  false  $default
-     * @return  mixed
+     * @return mixed
      */
     public function integer($key = null, $default = false)
     {
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -297,6 +297,20 @@ trait InteractsWithInput
     }
 
     /**
+     * Retrieve input as an integer value.
+     *
+     * Returns an integer when value is a boolean or a number in string form i.e. "123". Otherwise, returns false.
+     *
+     * @param  null  $key
+     * @param  false  $default
+     * @return  mixed
+     */
+    public function integer($key = null, $default = false)
+    {
+        return filter_var($this->input($key, $default), FILTER_VALIDATE_INT);
+    }
+
+    /**
      * Retrieve input from the request as a collection.
      *
      * @param  array|string|null  $key
--- a/tests/Http/HttpRequestTest.php
+++ b/tests/Http/HttpRequestTest.php
@@ -503,6 +503,17 @@ class HttpRequestTest extends TestCase
         $this->assertFalse($request->boolean('some_undefined_key'));
     }
 
+    public function testIntegerMethod()
+    {
+        $request = Request::create('/', 'GET', ['with_trashed' => 'false', 'price' => '1.23', 'download' => true, 'checked' => 1, 'unchecked' => '0', 'id' => '123']);
+        $this->assertFalse($request->integer('with_trashed'));
+        $this->assertFalse($request->integer('price'));
+        $this->assertIsInt($request->integer('download'));
+        $this->assertIsInt($request->integer('checked'));
+        $this->assertIsInt($request->integer('unchecked'));
+        $this->assertIsInt($request->integer('id'));
+    }
+
     public function testCollectMethod()
     {
         $request = Request::create('/', 'GET', ['users' => [1, 2, 3]]);


seriquynh:
The $query and $request properties of the $request object are instances of Symfony\Component\HttpFoundation\ParameterBag, so you can use the getInt method of it like this:

```
$request->query->getInt('your var');
$request->request->getInt('your_var');
```

Anyway, \Illuminate\Http\Request class already has a method called boolean. It depends, these proxy methods may be helpful or I can just cast values manually, it's fine.

src/Illuminate/Http/Concerns/InteractsWithInput.php
```diff
@@ -296,6 +296,20 @@ public function boolean($key = null, $default = false)
         return filter_var($this->input($key, $default), FILTER_VALIDATE_BOOLEAN);
     }
 
+    /**
+     * Retrieve input as an integer value.
+     *
+     * Returns an integer when value is a boolean or a number in string form i.e. "123". Otherwise, returns false.
+     *
+     * @param  null  $key
```
GrahamCampbell, line 304: incorrect type

src/Illuminate/Http/Concerns/InteractsWithInput.php
```diff
@@ -296,6 +296,20 @@ public function boolean($key = null, $default = false)
...
+     * @param  false  $default
```
GrahamCampbell, line 305: same here

GrahamCampbell (changes requested):

derekmd:
## `integer()` never returns `bool`

The parameter `$default = false` makes sense for `Request@boolean()` that _always_ returns `bool` but may not for a new `integer()` method. The intention here seems to be coercing to single type yet it may return a mix.

```php
Request::create('/', 'GET', [])->integer('foo');
// false

Request::create('/', 'GET', ['foo' => ''])->integer('foo');
// false

Request::create('/', 'GET', ['foo' => 'bar'])->integer('foo');
// false
```

To **support passing the `integer()` return value into a method with type-checked parameter `?int $arg`** or to fill in a strict-type database column, this new method should probably fallback to `null` instead of `false`. 

```php
public function integer($key = null, $default = null)
{
    return filter_var($this->input($key, $default), FILTER_VALIDATE_INT, FILTER_NULL_ON_FAILURE);
}
```

## Casting vs. `filter_var()`

Now comparing the `filter_var()` implementation and `(int)` casting:

```php
use Illuminate\Support\Facades\Request;
Request::macro('integer', function ($key = null, $default = null) {
    return filter_var($this->input($key, $default), FILTER_VALIDATE_INT, FILTER_NULL_ON_FAILURE);
});

Request::create('/', 'GET', [])->integer('foo');
// null
Request::create('/', 'GET', [])->query->getInt('foo');
// 0

Request::create('/', 'GET', ['foo' => ''])->integer('foo');
// null
Request::create('/', 'GET', ['foo' => ''])->query->getInt('foo');
// 0

// I assume request validation makes the next 2 use cases irrelevant:
Request::create('/', 'GET', ['foo' => '1.234'])->integer('foo');
// null
Request::create('/', 'GET', ['foo' => '1.234'])->query->getInt('foo');
// 1

Request::create('/', 'GET', ['foo' => '1bar'])->integer('foo');
// null
Request::create('/', 'GET', ['foo' => '1bar'])->query->getInt('foo');
// 1

Request::create('/', 'GET', ['foo' => 'bar2'])->integer('foo');
// null
Request::create('/', 'GET', ['foo' => 'bar2'])->query->getInt('foo');
// 0

// integer disallows leading zeroes? see below
Request::create('/', 'GET', ['minute' => '07'])->integer('minute');
// null
Request::create('/', 'GET', ['minute' => '07'])->query->getInt('minute');
// 7

Request::create('/', 'GET', ['minute' => '7'])->integer('minute');
// 7
Request::create('/', 'GET', ['minute' => '7'])->query->getInt('minute');
// 7
```

So casting and `filter_var()` have some differing behavior that may make the `filter_var()` implementation more trouble than its worth although it does match the `integer` validation rule:

https://github.com/laravel/framework/blob/ccea1bfcbb37cf923dc1bb30cdbf2effbfb1619c/src/Illuminate/Validation/Concerns/ValidatesAttributes.php#L1188-L1191

e.g., edge case for input of minute/hour/month/day with a leading 0:

```php
filter_var('07', FILTER_VALIDATE_INT, FILTER_NULL_ON_FAILURE)
// null

filter_var('07', FILTER_VALIDATE_INT, FILTER_NULL_ON_FAILURE | FILTER_FLAG_ALLOW_OCTAL)
// 7
```

## Userland solution for handling base 10 integer input

I found the best catch-all solution was avoiding `filter_var()` by validating with a regex and then casting:

```php
class BaseTenInteger implements \Illuminate\Contracts\Validation\Rule
{
    {
        return preg_match('/^[0-9]+$/', $value);
    }

    public function message()
    {
        return trans('validation.integer');
    }
}
```

#### Required integer parameter

```php
request()->validate(['foo' => ['required', new BaseTenInteger]);
$foo = (int) request('foo');
```

#### Optional integer parameter

```php
request()->validate(['foo' => new BaseTenInteger]);
$foo = request()->filled('foo')
    ? (int) request('foo')
    : null;
```

taylorotwell:

PR #39721, opened 2021-11-22
Initializable traits in controllers

MFlor:
Inspired by the ability to initialize traits for models by adding a static method called `boot[Trait Name]`, I've introduced a similar feature for controllers. See [`bootTraits()`](https://github.com/laravel/framework/blob/36dfae9d9ef7f88e8f9489c484a0a0609592bc21/src/Illuminate/Database/Eloquent/Model.php#L260)

This will allow us to add a simple trait to many controllers, adding logic that could be placed in a constructor of a controller.

This comes from a need in our company, where we are adding a middleware to many controllers, caching an entire response from the controller. But instead of adding a constructor to all controllers, the trait could be used. This will also allow for defaults to be placed in the trait instead of adding it to all controllers.

Given the trait:
```php
trait CacheResponseTrait
{
    public function initializeCacheResponseTrait()
    {
        $ttl = 10;
        if (property_exists($this, 'ttl')) {
            $ttl = $this->ttl;
        }
        $this->middleware(CacheRequestsMiddleware::class . ':' . $ttl);
    }
}
```
We will be able to add the trait to all controllers needing to cache an entire response. And for the given controllers, that need a longer `ttl`, than the default, we'll simply be able to define the controller as such:

```php
class ApiController extends Controller
{
    use CacheResponseTrait;
    
    protected int $ttl = 3600;
```

Thank you for considering my pull request.

### DIFF
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -291,7 +291,7 @@ class Route
     {
         foreach (class_uses_recursive($this->controller) as $trait) {
             $method = 'initialize'.class_basename($trait);
-            if (method_exists($this->controller, $method) && !in_array($method, $this->initializedTraits)) {
+            if (method_exists($this->controller, $method) && ! in_array($method, $this->initializedTraits)) {
                 $this->controller->{$method}();
                 $this->initializedTraits[] = $method;
             }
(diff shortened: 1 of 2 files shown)


taylorotwell:
I suggest using a route group to assign the middleware to multiple controllers.

PR #39724, opened 2021-11-22
[8.x] Fix wrong return type in PHPDoc

marvin-wtt:
I think `#resolveChildRouteBindingQuery` actually returns a Builder instead of the Model.


sebdesign:
Actually this is not correct. Method calls from the `Relation` objects to the underlying `Builder` object return the relation and not the query. So the correct return type is `Illuminate\Database\Eloquent\Relations\Relation`.

See https://github.com/laravel/framework/blob/8.x/src/Illuminate/Support/Traits/ForwardsCalls.php#L55

GrahamCampbell:
@sebdesign can you submit a PR please?

PR #39725, opened 2021-11-23
[8.x] Fixed return type of `Grammar::getValue`

robotsniper:
this return type should be consistent with Expression class getValue method return type

### DIFF
--- a/src/Illuminate/Database/Grammar.php
+++ b/src/Illuminate/Database/Grammar.php
@@ -179,7 +179,7 @@ abstract class Grammar
      * Get the value of a raw expression.
      *
      * @param  \Illuminate\Database\Query\Expression  $expression
-     * @return string
+     * @return mixed
      */
     public function getValue($expression)
     {


sebdesign:
I would suggest that we revert this change.

The actual return type of this method depends on the return type of `Expression::getValue()`.
While this method returns `mixed`, this is something that was added in the first commit of Laravel!
See: https://github.com/laravel/framework/blame/599e7851988ce88aae745847389dcefd3c4cc6f6/src/Illuminate/Database/Query/Expression.php#L26

The `Expression` also has a `__toString()` method that casts the result of `getValue()` to `string`, which is risky, because if the value cannot be casted to `string`, that would raise an error.

Also the result of `Grammar::getValue()` is used in many methods that return `string`, like `wrapTable()`, `wrap`, `parameter`.

The most sane thing to do is to change all the references of `mixed` types to `\Stringable|scalar` in the `Expression` class in 9.x as it is a breaking change.

PR #38167, opened 2021-07-29
Add: Arr:undot() helper

BenCavens:
This PR adds a new array helper method `Arr::undot`. 

A `Arr::dot` method exists for quite a while which converts an multidimensional array to an array with dot notation.

_Out-of-the-framework-box_ there isn't a method available for the opposite behaviour. 
This PR adds functionality to transform an array with dot notation to a multi-dimensional array. Although a macro can be set, it feels right to add it to core since the dot method is already present. 

Example usage:
```php 

Illuminate\Support\Arr::undot(['foo.bar' => 'baz']);

// Result: ['foo' => ['bar' => 'baz']]


```

PR #33845, opened 2020-08-13
[7.x] Add `undot` array helper

divspace:
Adds the inverse to the `dot` array helper by expanding array keys with "dot" notation into multi-dimensional arrays.

Besides adding the inverse functionality, there are times when certain functions/methods in your code may not allow multi-dimensional arrays and you need a way to pass flattened "dot" notated arrays to them (or maybe you find the syntax easier to write). For example, if you were to [add a message](https://laravel.com/api/7.x/Illuminate/Contracts/Support/MessageBag.html#method_add) with something like:

```php
$messages = new MessageBag([
    'error.code' => 403,
    'error.message' => 'Unauthorized request',
]);
```

You could store those messages correctly, and after retrieving them, simply `undot` the messages to get a proper array:

```php
[
  'error' => [
    'code' => 403,
    'message' => 'Unauthorized request'
  ]
]
```

This does not break existing functionality since it's a new helper and tests have been provided.

*This is my first pull request **ever** to a project that isn't work related, so hopefully I did this right (even watched Jeffrey Way's video on Laracasts). I think I spent more time writing the method's description than anything else* 😬

PR #13897, opened 2016-06-06
[5.2] Add Arr::unDot() method

mustafaaloko:
Expand a dotted-key array. It acts the opposite way of Arr::dot(). 

**Note:** I am not sure about the method name itself (is it good or not). If acceptable for merge, then we can change the method name.

PR #39729, opened 2021-11-23
[8.x] Add `undot()` method to Arr helpers and Collections

amiranagram:
This PR adds support for `elevate` method in array helpers and collections.

This method basically does the opposite of `Arr::dot()`, it takes an associative array and converts it to a multi-dimensional one based on keys with dot notation.

For example if we take this array and pass it through `Arr::dot()`:

```php
$original = [
    'user' => [
        'name' => 'foo',
        'occupation' => 'bar',
    ]
];

$dotted = Arr::dot($original);
```

We would get a _flattened_ associative array:

```php
$dotted = [
    'user.name' => 'foo',
    'user.occupation' => 'bar',
];
```

Now if we _elevate_ the _dotted_ array we should get the original.

```php
$elevated = Arr::elevate($dotted);
```

`$elevated ≡ $original`

Same goes for collections and lazy collections. Not sure if "elevate" is the best name for it, so I'm open to suggestions.

Cheers.

P.S. Will take a look on how to test if a collection is actually lazy and add an additional test. Some help is welcome, of course.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -122,6 +122,23 @@ class Arr
     }
 
     /**
+     * Elevate an associative array with dots to a multi-dimensional array.
+     *
+     * @param iterable $array
+     * @return array
+     */
+    public static function elevate($array)
+    {
+        $results = [];
+
+        foreach ($array as $key => $value) {
+            static::set($results, $key, $value);
+        }
+
+        return $results;
+    }
+
+    /**
      * Get all of the given array except for a specified array of keys.
      *
      * @param  array  $array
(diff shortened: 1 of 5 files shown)


johanrosenson:
This has been suggested before as `undot`, which i personally think is a great name, pretty self explanatory.

https://github.com/laravel/framework/pull/38167
https://github.com/laravel/framework/pull/33845
https://github.com/laravel/framework/pull/13897
https://github.com/laravel/framework/pull/4735

sadly `undot` have not been merged so far.

amiranagram:
@johanrosenson Welp, wasn't aware that this was suggested before.

Agree, `undot` sounds way better. _Elevate_ is an antonym of _flatten_ as far as I know and _flatten_ is something else entirely.

taylorotwell:
Since this keeps coming back up as a PR I will go ahead and include it.

PR #39738, opened 2021-11-23
[8.x] Use `$this` in order to check methods in traits

glorand:
we will determine if the model base class (!!traits included!!) contains this given key


taylorotwell:
Breaking change.

PR #39739, opened 2021-11-23
[8.x] Avoid overwrite on `php artisan make:listener` command.

xiCO2k:
Hey, This PR fixes an issue that I running `php artisan make:listener`.

So if I run it as it is right now it overwrites the file, instead of returning `Listener already exists!`.

## Before:
image

## After:
image

Thanks,
Francisco.


taylorotwell:
Comes up every few months. This is a breaking change. Please review other PRs.

xiCO2k:
Should I push it to the 9.x?

Thanks.

PR #39741, opened 2021-11-23
[8.x] Avoid overwrite last ip or user_agent with empty ones on DatabaseSessionHandler

PaolaRuby:
https://github.com/tlaverdure/laravel-echo-server/pull/546
In some cases request does'nt has `ipAddress` or `userAgent`, with this change it avoid overwrite the last reported data

### DIFF
--- a/src/Illuminate/Session/DatabaseSessionHandler.php
+++ b/src/Illuminate/Session/DatabaseSessionHandler.php
@@ -174,6 +174,15 @@ class DatabaseSessionHandler implements ExistenceAwareInterface, SessionHandlerI
      */
     protected function performUpdate($sessionId, $payload)
     {
+        if ($this->container->bound('request')) {
+            if (! $payload['ip_address']) {
+                unset($payload['ip_address']);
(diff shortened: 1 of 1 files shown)


taylorotwell:
What if it's the very first request and there is no ip_address information in the session payload? Should we initialize it to empty string then? Otherwise this could be a breaking change if people are always expecting it to be there.

PaolaRuby:
https://github.com/laravel/framework/blob/277c2fbd0cebd2cb194807654d870f4040e288c0/src/Illuminate/Session/Console/stubs/database.stub#L19-L20
They are `nullable`, but could be a breaking change

PaolaRuby:
@taylorotwell fixed

taylorotwell:
Not sure I want to resolve the request from the container here. It looks like once that PR is merged on the echo-server then the problem will be solved.

PaolaRuby:
@taylorotwell **echo-server** is already abandoned, that PR will not be merged, and it was just an example

PR #39752, opened 2021-11-24
[8.x] Add Event::fakeExcept() to fake all but certain events

devfrey:
I sometimes find myself working on feature tests that trigger multiple events. I try to fake all events to avoid unexpected side effects, but sometimes certain events may be required to simulate expected behaviour. I would use this to fake all but certain events:

```php
Event::fake([
    fn (string $eventClass) => $eventClass !== ServiceSaving::class,
]);
```

This PR adds a new `fakeExcept()` method to the `Event` facade to allow this:

```php
Event::fakeExcept(ServiceSaving::class);
Event::fakeExcept([ServiceSaving::class, 'some-other-event']);
```

### DIFF
--- a/src/Illuminate/Support/Facades/Event.php
+++ b/src/Illuminate/Support/Facades/Event.php
@@ -46,6 +46,21 @@ class Event extends Facade
     }
 
     /**
+     * Replace the bound instance with a fake, but allow the given events.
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #39755, opened 2021-11-24
[8.x] Add App::isDebug() method to detect debug mode

joelmellon:
Like App::isLocal(), isProduction(), runningInConsole(), isLocale(), etc. it makes sense to have a method on the Container to detect debug mode.

I struggled with the method name, but isDebug() sort of matches most of the methods with similar functionality. I also like:

* debugMode()
* debugEnabled()
* debugModeEnabled()
* inDebugMode()


williamdes:
Thank you so much !
I needed this right now

PR #39763, opened 2021-11-25
[6.x] append cookie

Zsfzy:


driesvints:
Most likely, this isn't needed. Also: please provide an explanation next time why you sent this in.

Zsfzy:
> Most likely, this isn't needed. Also: please provide an explanation next time why you sent this in.

i hava a auth token in cookies.
but json function cannot attach cookies like other functions.
and In 7.x 8.x versions this function attach cookies

My English is not very good, I hope you can understand what I say

PR #39768, opened 2021-11-25
add Redis::bulk to simplify bulk operation

nasservb:
With this PR we could able to work with bulk operations simply.
For example:

```php
 Redis::bulk('like-count')
            ->add(auth()->id())
            ->count(100)
            ->then(function ($likes) use ($post) {
                $post->likes()->sync($likes);
            });
```

This code store post likes in one bulk query(every 100 like).


taylorotwell:
This PR doesn't explain how this works internally, performance improvements, etc.

PR #39770, opened 2021-11-25
[9.x] Adds the ability to delay retries on database transactions

lukeraymonddowning:
Howdy guys!

When performing a database transaction with attempts to avoid deadlocks, it is often desirable to wait a certain number of milliseconds before retrying.

Inspired by this tweet https://twitter.com/mattkingshott/status/1463572180546068490?s=20, I thought it would be nice in Laravel 9 to support this on the method itself. 

Here's the before and after:

```php
// Before
retry(3, fn() => DB::transaction(User::update(['first_name' => 'Taylor'])), 250);

// After
DB::transaction(User::update(['first_name' => 'Taylor']), 3, 250);
```

Obviously, this needs to target 9.x because it affects the signature of the `Connection` interface.

On a side note, has the `value` function been altered? It no longer seems to pass properties, but I could be missing something.

Thanks for all the continued hard work.

Kind Regards,
Luke


lukeraymonddowning:
Yeah, better docs around this would be a good idea @driesvints. I guess if it is going to cause issues for a lot of 3rd party libs, perhaps that would be better 👍 Just wouldn't want the solution to go un-missed by those not on social media etc if that makes sense?

mfn:
The benefit of the built-in way is that it works also when using illuminate/database directory, where the global helpers are not available (because they're part of `laravel/framework`).

But besides the signature (thus: breaking) change, I feel like the current `transaction` method is very concise and should not be changed _that_ particular way (though, unfortunately, I don't have a better suggestion at hand 😅)

driesvints:

taylorotwell:
I think instead of introducing the breaking change I'll just let people use the existing `retry` helper approach.

PR #39773, opened 2021-11-25
Fixed ErrorException - Division by zero in LengthAwarePaginator

hafizshahid:
$this->lastPage = max((int) ceil($total / $perPage), 1);

This line was causing the error and the applications were breaking due to the zero value of $perPage.

It was giving Error Exception - Division by zero

We have replaced it with the following code

$this->lastPage = ((int) $perPage != 0) ? max((int) ceil($total / $perPage), 1) : 1;


taylorotwell:
I suggest not passing perPage of 0.

PR #39774, opened 2021-11-25
[9.x] Cast values on route binding

sl0wik:
Using [route model binding](https://laravel.com/docs/8.x/routing#route-model-binding), primary key is passed as a string causing exception on databases with strict typing.

### Example

`GET /users/100` call with route:
```
Route::get('/users/{user}', function (User $user) {
  // ...
});
```

Currently this request would execute `where('id', '100')` query.
Improved behavior would be `where('id', 100)` because the primary key should be an integer by default.

### Benefits

- Extra security layer
- This feature is required to utilize strongly-typed databases (like Google Cloud Spanner).

### Risk

- This feature might break routes using weak-typing (for example `/users/taylor` would not work if model key type is set to `int`)


sl0wik:
If anyone knows alternatives, please share.

chikaldirick:
If casting to an integer happens for each route binding param, I would use at least `is_numeric` to cast only those strings like `100` that look like actual numbers, not strings like `taylor` that would be casted to `0`.
But even taking this approach we're not safe with cases where, for example, user name contains only digits (and therefore would be casted to an integer) but internally is stored as a string.

antonkomarev:
> If anyone knows alternatives, please share.

Don't use implicit binding 😉

X-Coder264:
IMO this should be merged, it can be considered as a bug fix too. The cast should be respected, the query should not bind a string if the (primary) key is set to be an integer in the model. The suggestion to NOT use implicit binding also doesn't make sense IMO. Implicit binding is one of the most used and loved Laravel features, why would we need to use more code with explicit binding if we don't need to and there's no justified reason for it?

src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -1877,7 +1877,9 @@ public function getRouteKeyName()
      */
     public function resolveRouteBinding($value, $field = null)
     {
-        return $this->where($field ?? $this->getRouteKeyName(), $value)->first();
+        $field = $field ?? $this->getRouteKeyName();
```
arthurtavaresdev, line 1880:
```suggestion
        $field ??= $this->getRouteKeyName();
```

sl0wik:
I got access to a preview of [Google Cloud Spanner PostgreSQL](https://cloud.google.com/blog/topics/developers-practitioners/postgresql-interface-adds-familiarity-and-portability-cloud-spanner) and working on making it compatible with Eloquent. "Don't use implicit binding" to all users won't work. The only alternative to route parameter casting I can see is casting directly on all queries, which would require significant changes.

There are downsides to this solution, but throwing route string directly to where without space for any sanitization is also questionable.

taylorotwell:
The other route binding methods that add where conditions in this file required no updates?

sl0wik:
@taylorotwell I'm updating `resolveChildRouteBindingQuery`. I will also make https://github.com/laravel/framework/blob/23912e779356715a1719eb3720bd7b9327c5e463/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L576 public unless there is a reason to keep it protected.

src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
```diff
@@ -573,7 +573,7 @@ public function mergeCasts($casts)
      * @param  mixed  $value
      * @return mixed
      */
-    protected function castAttribute($key, $value)
+    public function castAttribute($key, $value)
```
X-Coder264, line 576: Is there a need for this? Why does this method need to be `public`? This seems to me like an unnecessary breaking change.

sl0wik:
If model key type is not part of `getCasts()`, we could cast only variables in `$casts` and keep the primary key as is. With the current `getCasts()` approach, this is the best way I found to handle cast on route binding. 
 https://github.com/laravel/framework/blob/23912e779356715a1719eb3720bd7b9327c5e463/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1264

src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
sl0wik: I will restore it as protected. Perhaps it can be extended in the future with something public.

Using cast looks like a good way to control types in strongly-typed databases.

Currently, there is no way to access cast method if for example user would like to do something like `where('user_id', $user->castAttribute('id', $user->id))`.

driesvints:

taylorotwell:
Is this actually broken right now on Laravel 8 and Postgres? The database connection class uses `PDO::BIND_INT` if the value is an integer.

sl0wik:
I will run extra test on L8 + Postgres.

sl0wik:
L8 Postgres unit tests are green with those updates. I also did a basic test on the live app and behave as expected (default primary key "1" => 1).

taylorotwell:
Nobody answered my question that I can see. You're telling me that *all* implicit route model binding is currently broken on Laravel 8 without this PR?

taylorotwell:
Just tested a simple implicit binding route on Postgres and Laravel 8 and it's working fine:

```php
Route::get('/users/{user}', function (User $user) {
    return $user;
});
```

What should I try that is broken?

sl0wik:
It will work fine in L8 + Postgres. It`s needed for [Google Cloud Spanner Postgres interface](https://cloud.google.com/blog/topics/developers-practitioners/postgresql-interface-adds-familiarity-and-portability-cloud-spanner) implementation which is not fully compatible with Postgres but with a couple of small changes can be compatible with Laravel. Spanner is very strict about types, and to implement this interface value must pass through `casts` before hitting where. It can be done by extending Eloquent Model under a separate package (example  https://github.com/colopl/laravel-spanner/blob/master/src/Colopl/Spanner/Eloquent/Model.php), but there seems to be no reason to not have it implemented in Eloquent by default.

Its all about `/users/1` to call `->where('id', 1)` instead of `->where('id', '1')` (in default setup).

taylorotwell:
If you need this I suggest customizing your base model for now. That is why we have customization hooks on that model.

PR #39778, opened 2021-11-25
[8.x] Adds on-demand authorization check

DarkGhostHunter:
## What

Allows the developer to throw an `AuthotizationException` using a condition, bypassing before/after callbacks or registering a gate or policy that's only used once.

Before:
```php
if (auth()->user()->comments()->count() > 3) {
    throw new AuthorizationException();
}
```

After:
```php
Gate::authorizeUnless(auth()->user()->comments()->count() > 3);
```

It also supports callbacks, which allows to use the authenticated user as parameter, and a message with a code.

```php
Gate::authorizeIf(function ($user) {
    return $user->comments()->count() < 2;
}, "You have allocated all of your comments", 406);
```

And supports using the `Response` itself as a callback result. When doing so, it can override the false/truthy result, giving more control on the callback.

```php
Gate::authorizeUnless(function ($user) {
    if ($user->notPremium()) {
        return Response::deny('You have to activate your account before commenting');
    }

    return $user->comments()->count() > 3;
}, "You have allocated all of your comments", 406);
```

> Includes `authorizeIf()` and `authorizeUnless()`  to do the opposite.

# BC?

None, only additive.

### DIFF
--- a/src/Illuminate/Auth/Access/Gate.php
+++ b/src/Illuminate/Auth/Access/Gate.php
@@ -391,7 +391,7 @@ class Gate implements GateContract
      */
     public function authorizeIf($condition, $arguments = [])
     {
-        $response = value($condition, ...array_merge([$this->resolveUser()], $arguments));
+        $response = value($condition, ...array_merge([$this->resolveUser()], Arr::wrap($arguments)));
 
         if (! $response instanceof Response) {
             $response = new Response($response);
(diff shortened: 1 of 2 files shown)


DarkGhostHunter:
If you're asking why I made so much commits, it's because I used VSCode Web on a tablet.

taylorotwell:
Doesn't read right to me? If sounds like the user *will* be authorized if the condition is truthy, but they will actually be denied.

DarkGhostHunter:
Fucked up the example. Now is understandable.

DarkGhostHunter:
Gonna scrap this and add this functionality to `authorize()` itself, because you can do something similar with `throw_if()` and `throw_unless()`.

PR #39783, opened 2021-11-26
[8.x] Fix sqlserver paging issue

joelharkes:
Sql server does not support paging without a `order by` column.

But the current implementation leads to issues when using `GUID` column as ID's and generating those id's in laravel. (sql orders id's differently than alphabetically, so the order of the models inserted is differently than the order of ID's).

### Current setup
Old setup when paging and sorting:

page1:
```sql
SELECT TOP 50 [users].* FROM [users] WHERE [users].[deleted_at] IS NULL ORDER BY [id] ASC
```

page2:
```sql
SELECT * FROM (SELECT [users].*, row_number() OVER (ORDER BY [id] ASC) AS row_num FROM [users] WHERE [users].[deleted_at] IS NULL) AS temp_table WHERE row_num between 51 and 100 ORDER BY row_num
```
### Proposed solution

page1:
```sql
SELECT TOP 50 [users].* FROM [users] WHERE [users].[deleted_at] IS NULL ORDER BY [id] ASC
```
page2:
```sql
SELECT [users].* FROM [users] WHERE [users].[deleted_at] IS NULL ORDER BY [id] ASC OFFSET 50 ROWS FETCH NEXT 50 ROWS ONLY
```

### Edge cases

- `OFFSET 50 ROWS FETCH NEXT 10` is a feature added to sql server in 2012. But you are already using features that are not supported by older versions.
- this only works when you provide an order column as this is mandatory in sql server.

## improvements

- instead of needing to query the whole table, now it will stop after Offset+Limit. (eg page 2 on 50 results = after 100 results instead of the whole table). 
- when ordering you will no longer get the issue of rows being ordered differently on the first and subsequent pages.

### DIFF
--- a/CHANGELOG-7.x.md
+++ b/CHANGELOG-7.x.md
@@ -1,6 +1,22 @@
 # Release Notes for 7.x
 
-## [Unreleased](https://github.com/laravel/framework/compare/v7.30.4...7.x)
+## [Unreleased](https://github.com/laravel/framework/compare/v7.30.5...7.x)
+
+
+## [v7.30.5 (2021-11-17)](https://github.com/laravel/framework/compare/v7.30.4...v7.30.5)
+
+### Added
+- Added new line to `DetectsLostConnections` ([#36373](https://github.com/laravel/framework/pull/36373))
+
+### Fixed
+- Fixed `Illuminate\View\ViewException::report()` ([#36110](https://github.com/laravel/framework/pull/36110))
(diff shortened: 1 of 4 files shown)


GrahamCampbell:
Laravel 7 is EOL and not accepting bug fixes. Please upgrade to Laravel 8.

driesvints:
Seems like some branching went wrong

PR #39784, opened 2021-11-26
add Redis::bulk to simplify bulk operation

nasservb:
**What's the Problem and How to solve it?**
when you need to save data to the DB, and this data is obtained over time such as page visit count or the number of like, it is not a good idea to insert data into the database for each request. 

We can use Redis to cache this data, and when the number of items reaches a significant amount, read the data from Redis and save it to DB via a bulk insert query. 

This PR helps us to do this simply in a few lines of code for saving visits:
```php 
	Redis::bulk('page-visit')
            ->add(request()->ip())
	    ->then(function($visits)use($page){
		$page->visits()->attach($visits);
	    });
```
or for saving likes:
```php 
	Redis::bulk('like-count')
            ->add(auth()->id())
            ->then(function ($likes) use ($post) {
                $post->likes()->sync($likes);
            });
```
 
**How does this work internally?** 
- This code uses the Redis RPUSH command to add data to the cache,
- Check cache size with Redis LLEN command,
- And when the cache is full, release them with the Redis LRANGE command and delete the cache items with the Redis DEL command. 
 
**What are the benefits?** 
If you have a direct insert in each request and replace it with the Redis::bulk function, your performance will be 10 times better in the default configuration. you can configure your cache size with: 
```php 
	Redis::bulk('key')->count(cache_size)...; 
```	
for saving likes:
```php 
	Redis::bulk('like-count')
            ->add(auth()->id())
            ->count(100)
            ->then(function ($likes) use ($post) {
                $post->likes()->sync($likes);
            });
```
**How to extract cache data?**
And improve the performance again, but keep in mind that when the cache is large, saving your data will be delayed until the cache is full and remains in memory. 

When the execution of your code is going to finish, you need to release data from the cache and save it to DB.

To do this, you can use the forceRelease function as follows: 
```php 
	Redis::bulk('like-count')
            ->forceRelease()
            ->then(function ($likes) use ($post) {
                $post->likes()->sync($likes);
            });
```
You can add this code in Application::shutdown() function or any other place that runs at the end.

### DIFF
--- a/src/Illuminate/Redis/Connections/Connection.php
+++ b/src/Illuminate/Redis/Connections/Connection.php
@@ -5,9 +5,9 @@ namespace Illuminate\Redis\Connections;
 use Closure;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Redis\Events\CommandExecuted;
+use Illuminate\Redis\Limiters\BulkBuilder;
 use Illuminate\Redis\Limiters\ConcurrencyLimiterBuilder;
 use Illuminate\Redis\Limiters\DurationLimiterBuilder;
-use Illuminate\Redis\Limiters\BulkBuilder;
 use Illuminate\Support\Traits\Macroable;
 
 abstract class Connection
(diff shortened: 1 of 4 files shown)


taylorotwell:

PR #39786, opened 2021-11-26
Map Illuminate\Support\ItemNotFoundException to 404 Not Found HTTP error

riccardodallavia:
The pull request aims to map the `Illuminate\Support\ItemNotFoundException` exception, thrown by the `firstOrFail()` methods both in Collection and LazyCollection, to a 404 HTTP error in `Illuminate\Foundation\Exceptions\Handler`.

The main reason is I think the exception should throw the same error code as of the `firstOrFail()` method of the Eloquent query.
When I, for some reason, must fetch all results of an eloquent query and filter them using the `filter()` method of `Illuminate\Support\Collection`, I get a different error code when executing the `firstOrFail()` method.

Here is an example of what I'm talking about: in both cases, I'm working with models, but the exception thrown is different, which should not be correct.

```
// throws 404 not found exception
User::query()
  ->where('created_at', '>=', now())
  ->firstOrFail();
```

```
// throws 500 internal server error exception
User::query()
  ->where('created_at', '>=', now())
  ->get()
  ->filter(
      fn ($user) => $user->someMethod()
  )
  ->firstOrFail();
```

I think in both cases the exception is about an entry not found (whether it is a model, a string, or whatever), so the exception should always be a 404 Not Found error.

### DIFF
--- a/src/Illuminate/Foundation/Exceptions/Handler.php
+++ b/src/Illuminate/Foundation/Exceptions/Handler.php
@@ -21,6 +21,7 @@ use Illuminate\Routing\Router;
 use Illuminate\Session\TokenMismatchException;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Auth;
+use Illuminate\Support\ItemNotFoundException;
 use Illuminate\Support\Reflector;
 use Illuminate\Support\Traits\ReflectsClosures;
 use Illuminate\Support\ViewErrorBag;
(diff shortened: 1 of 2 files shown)


dennisprudlo:
Wouldn't that throw a 404 when I use a generic collection with strings or numbers (or some other data) in it and `firstOrFail()` fails?

```php
collect([1, 2, 3])->filter(fn ($value) => $value > 5)->firstOrFail()
```

riccardodallavia:
> Wouldn't that throw a 404 when I use a generic collection with strings or numbers (or some other data) in it and `firstOrFail()` fails?
> 
> ```
> collect([1, 2, 3])->filter(fn ($value) => $value > 5)->firstOrFail()
> ```

Yes, it would - but isn't that correct? The exception is about an item **not found** in the collection, which should indeed be a 404 error and not a 500 internal server error.

dennisprudlo:
I don't know IMO a 404 is for an actual resource that could not be found. When I request `GET /users/423` and receive a 404 I assume the User with the ID 423 does not exist (anymore). But the 404 would also appear when I have a collection of some data like email addresses that belongs to the user and for some reason the collection is empty.

I would consider this a server error because the requested resource exists.

taylorotwell:
Not sure I want this to map to 404 all the time.

PR #39789, opened 2021-11-26
[8.x] Adds on-demand gate authorization

DarkGhostHunter:
## What?

Allows the gate to _permit_ or _forbid_ a procedure by a condition, instead of using the gate to register a one-time ability or using verbose checks. Rework of #39778.

Before
```php
use Illuminate\Auth\Access\AuthorizationException;
use App\Models\Wallet;

$wallet = Wallet::find(1);

if ($wallet->overQuota()) {
    throw new AuthorizationException("This action exceeds your cuota");
}
```
After
```php
use Illuminate\Support\Facades\Gate;
use App\Models\Wallet;

$wallet = Wallet::find(1);

// Forbids an action...
Gate::forbid($wallet->overQuota(), "This action exceeds your cuota");

// or permits an action.
Gate::permit($wallet->underQuota(), "This action exceeds your cuota");
```

> This is intended is only for one-time checks. Developers are encouraged to register abilities when these can be reused.

## Why?

Because you can bypass registering an action in the Gate, without abandoning the Gate powers. For example, you can also issue callbacks to retrieve the currently authenticated user and even set a message...

```php
use Illuminate\Support\Facades\Gate;
use App\Models\Wallet;

Gate::forbid(fn ($user) => $user->wallet->overQuota(), "You exceed your cuota", 409);
```

... or hijack the callback with a `Response` instance anytime.

```php
use Illuminate\Auth\Access\Response;use Illuminate\Support\Facades\Gate;
use App\Models\Wallet;

Gate::forbid(function ($user) {
    if ($user->can('bypassQuota')) {
        // This takes precedence. 
        return Response::allow('Quota limits are not applied to you.');
    }
    
    return $user->wallet->overQuota();
});
```

## BC?

Nope, only additive.


DarkGhostHunter:
Help pls Docker decided to die.

DarkGhostHunter:
Can someone help me re-run the jobs?

driesvints:
@DarkGhostHunter done

DarkGhostHunter:
Tx

taylorotwell:
Converting to draft while you potentially address this. Please mark as ready for review when you would like me to look at this PR again.

DarkGhostHunter:
I reworked the method to always throw an exception if the callback requires the authenticated user and there is none.

I thought a `AuthenticationException` would make more sense, but since the user requirement is part of the callback itself, I think it's fine to even respect the same message and code issued (if any).

taylorotwell:
I find passing an `$allow` flag as the last parameter of `allowIf` utterly confusing. Please remove that and just duplicate some code if you have to between the two methods. It makes them both much easier to read.

DarkGhostHunter:
Can I just create a “onDemand” protected method and use allowIf and denyIf as the fronts?

Not a fan of copy pasting anything. 

Italo Baeza C.

taylorotwell:

DarkGhostHunter:
You rock dude. If I wasn’t for you, I would probably be stealing on the streets. 

Italo Baeza C.

PR #39791, opened 2021-11-27
[8.x] name convention changed for -R flag

abbasudo:
this pr will change nameing convention of -R flag #39120
R flag creates form requests starting with the **method** name followed by the **model** name like that :
`StoreUserRequest.php`
`UpdateUserRequest.php`
this will bundle and sort files by their methods . 
i have changed this naming to be **model** name followed by **method** name like that : 
`UserStoreRequest.php`
`UserUpdateRequest.php`
this way its easy to find relevant files to a model if you have a lot models and will bundle files by their models

thanks for reading and have a good day


taylorotwell:
No plans to change this.

PR #39794, opened 2021-11-27
[9.x] View component `x-for` loop

sethsandaru:
## What & Why?

Hey guys, just a small idea for rendering components.

Inspired by `v-for` of VueJS - with the `x-for`, we don't have to wrap our component inside `@foreach`

### Before

```php
@foreach ($users as $user)
    <x-user-card :user="$user" />
@endforeach
```

### After & Usage
Pass an array, first param is the **iterable**, second is the **attribute key name** for your component.

```php
<x-user-card x-for="[$users, 'user']" />
```

Thanks and have a great day!

P/s: not sure this one is better for 9.x or 8.x, let me know if I need to bring it to 8.x instead.

### DIFF
--- a/src/Illuminate/View/Compilers/ComponentTagCompiler.php
+++ b/src/Illuminate/View/Compilers/ComponentTagCompiler.php
@@ -502,6 +502,8 @@ class ComponentTagCompiler
                 $attribute = Str::after($attribute, 'bind:');
 
                 $this->boundAttributes[$attribute] = true;
+            } elseif (Str::startsWith($attribute, 'x-for')) {
+                $this->boundAttributes[$attribute] = true;
             } else {
                 $value = "'".$this->compileAttributeEchos($value)."'";
             }
--- a/src/Illuminate/View/Compilers/ComponentTagCompiler.php
+++ b/src/Illuminate/View/Compilers/ComponentTagCompiler.php
@@ -498,11 +498,10 @@ class ComponentTagCompiler
 
             $value = $this->stripQuotes($value);
 
-            if (Str::startsWith($attribute, 'bind:')) {
+            // x-for should be considered as bound attribute
+            if (Str::startsWith($attribute, 'bind:') || $attribute === 'x-for') {
                 $attribute = Str::after($attribute, 'bind:');
 
-                $this->boundAttributes[$attribute] = true;
-            } elseif (Str::startsWith($attribute, 'x-for')) {
                 $this->boundAttributes[$attribute] = true;
             } else {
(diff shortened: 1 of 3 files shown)


ahinkle:
`x-for` is used by AlpineJS and along with other JavaScript frameworks (as you mentioned). This is a breaking change as `x-for` can be used with anonymous Component Attributes. 

https://laravel.com/docs/8.x/blade#component-attributes

sethsandaru:
> `x-for` is used by AlpineJS and along with other JavaScript frameworks (as you mentioned). This is a breaking change as `x-for` can be used with anonymous Component Attributes.
> 
> https://laravel.com/docs/8.x/blade#component-attributes

It should be fine with your normal HTML tags since `x-for` here is just for View Component and won't affect others.

sethsandaru:
> `x-for` is used by AlpineJS and along with other JavaScript frameworks (as you mentioned). This is a breaking change as `x-for` can be used with anonymous Component Attributes.
> 
> https://laravel.com/docs/8.x/blade#component-attributes

It won't affect other HTML tags - `x-for` here is just for View Component.

dennisprudlo:
@sethsandaru What @ahinkle means is that when using:
```html
<x-some-component x-for="..." />
```
You may want the `x-for` attribute to simply render as an HTML attribute so a Javascript Framework can pick it up and do something with it. Imagine we have a list component that shows items using AlpineJS. The base AlpineJS container has all kinds of data like `foo` and `bar`. With the component we can define which list should be displayed:

```html
<!-- resources/views/index.blade.php -->
<h1>All Foos</h1>
<x-list x-for="item in foo" />

<h1>All Bars</h1>
<x-list x-for="item in bar" />
```

```html
<!-- resources/views/components/list.blade.php -->
<ul x-data="{ foo: [...], bar: [...], ...}">
    <template {{ $attributes }}>
        <li x-text="item"></li>
    </template
</ul>
```

In this case we need the `x-for` to be rendered into the HTML the way we specified it in the anonymous component. If we hijack the attribute name for a blade-render-thing no one can use it anymore as an anonymous component attribute. Actually you expect the value of `x-for` to be an array with two values (the first iterable, the second a string). Wouldn't the blade rendering simply break with my code above because my `x-for` attribute has the value `"item in foo"`?

sethsandaru:
@dennisprudlo ah right, thanks. the `x-` prefix just came to my mind since it matched the `<x-` prefix of View Component 🤔  

Probably it would be better to change to another property name there

sethsandaru:
Closed for now to find a better solution

PR #39810, opened 2021-11-29
[8.x] Add collection methond double() and triple()

andiLeong:
Hi All,

I proposed adding the double() and triple() method to collection , which will double/triple up each value.
Without letting users manually call map or reduce methods.
I added a ```Arr::multiples($array,$multiples)``` which you can also multiply by any values.

### User Case

1, for example , an e-commerce website, I fetch api from a provider to get their product price. Now I want to double up our product price or I want to multiply 1.2.

2, when doing some kind of promotion , for example , if you spend at least 5000 with your credit card on a holiday, your total reward point will be doubled up at the end of today. so i can make a cron job get all qualified users, double up their points, update the db.

I think this one can be useful in many calculational scenarios that are business specific.

Thanks

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -432,7 +432,7 @@ class Arr
      */
     public static function multiples($array, $multiple)
     {
-        return array_map(function($item) use ($multiple) {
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #39811, opened 2021-11-29
[8.x] Customize type column in database notifications using databaseType me…

mvaliolahi:
By this pull request, we can customize the `type` column in the database notifications.

By default, Laravel use class_name($notification) for the type column, and this can be undesirable on the client-side.

```php
class InvoiceNotification extends Notification implements ShouldQueue {

    {
        return ['database'];
    }

    public function databaseType() 
    {
        return 'DAILY';
    }
```

### DIFF
--- a/src/Illuminate/Notifications/Channels/DatabaseChannel.php
+++ b/src/Illuminate/Notifications/Channels/DatabaseChannel.php
@@ -55,24 +55,22 @@ class DatabaseChannel
     {
         return [
             'id' => $notification->id,
-            'type' => $this->chooseType($notification),
+            'type' => $this->storageType($notification),
             'data' => $this->getData($notifiable, $notification),
             'read_at' => null,
         ];
     }
 
     /**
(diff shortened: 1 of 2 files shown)


driesvints:
Are you completely sure this actually works? Have you tested this in a real Laravel application. I've been trying to do this myself a couple of times but there was always a snag I ran into. Can't exactly remember which ones though 🤔

mvaliolahi:
> 

Yes, I test it, and works for me.

taylorotwell:

mvaliolahi:
@taylorotwell Thanks for your reply, This PR is similar to Relation::enforceMorphMap and your reply does not convince me! 

Notification record in DB is like below:

{
"notifications": [
	{
		"id" : "",
		"type" : "App\\Notifications\\LikeNotification",
		"notifiable_type" : "",
		"notifiable_id" :,
		"data" : "{}",
		"read_at" : ,
		"created_at" : "",
		"updated_at" : ""
	},
]}

PR #39817, opened 2021-11-30
[8.x] Allow blade view index default

faissaloux:
This PR allows to use `index.blade.php` as a default view for views folders.

For example we have this structure:
- `users`
  - `index.blade.php`
  - `create.blade.php`
  - `edit.blade.php`

Without this PR we have to call it like `view('users.index');`.
With this PR we can call it like `view('users');`.


taylorotwell:

PR #38925, opened 2021-09-23
[9.x] Add missing giveConfig method in ContextualBindingBuilder

denniscoorn-paqt:
This PR adds the method `giveConfig` to the `ContextualBindingBuilder` contract. A similar PR, https://github.com/laravel/framework/pull/36361, was targeted at `8.x` and rejected because it would introduce a breaking change for any implementation implementing this contract. That's why this PR is targeted at `master` to make it available for the upcoming `9.x` release.

One of the reasons why I think this method should be part of the contract is that the `ContextualBindingBuilder` contract interface is declared as return type by the method `Container::when()`. When someone uses the "[Binding primitives](https://laravel.com/docs/8.x/container#binding-primitives)" example from the docs, where the existence of the `giveConfig` method is documented, they may encounter warnings from their IDE claiming that the method doesn't exists. 

In my opinion this is small fix that can prevent possible confusion and improves developer experience.

PR #36361, opened 2021-02-23
Add missing giveConfig method in ContextualBindingBuilder

ahmadmayahi:
Add missing `giveConfig` method in `ContextualBindingBuilder` interface.

PR #36241, opened 2021-02-12
[8.x] Allow users to specify configuration keys to be used for primitive binding

simensen:
Syntactic sugar around fancy calls to `->give()` related to getting primitives from configuration.

Most of the time when I'm doing primitive binding it is for the purpose of using configuration values as constructor arguments. A lot of function calls are used and feel pretty bulky compared to being able to just say, "hey, get the value from this config key."

This PR provides the ability to do just that. Need an API key? `->need('$apiKey')->giveConfig('awesome_api.key')` is all you'd need to set in a service provider now.

```
$container
    ->when(ContainerTestContextInjectFromConfigIndividualValues::class)
    ->needs('$username')
    ->giveConfig('test.username');
```

Is the same as:

```
$container
    ->when(ContainerTestContextInjectFromConfigIndividualValues::class)
    ->needs('$username')
    ->give(function () {
        return config('test.username');
    });
```

and

```
$container
    ->when(ContainerTestContextInjectFromConfigIndividualValues::class)
    ->needs('$username')
    ->give(fn() => config('test.username'));
```

----

I believe this may require additional work on the contract (?) that I'll do if this gets approved. Would just need to know which branch to target that change on. I can also do the docs on this if it looks like this will be able to be merged. :)

PR #32514, opened 2020-04-23
[7.x] Provide syntactic sugar for contextually binding tagged services

simensen:
Handling tagged services feels heavier than it needs to be.

Currently, we can use primitive binding for primitive (`array`) constructor arguments to avoid having to use a factory for the whole class. That's a plus.

However, they almost always look the same:

```php
$this->app->when(WantFactory::class)
    ->needs('$wantConfigurers')
    ->give(function ($container) {
        return iterator_to_array($container->tagged('want_configurer'));
    });
```

There are some weird oddities with `Container::tagged`, though. So, even if one figures out they can use `iterator_to_array` to convert the response into something that can be consumed of the receiving object's constructor, it might not *always* work.

This is because `Container::tagged` returns an array early if nothing has been tagged with the tag in question:

```php
if (! isset($this->tags[$tag])) {
    return [];
}
```

So, to be safe, one would need to be more specific by doing something like this:

```php
$this->app->when(WantFactory::class)
    ->needs('$wantConfigurers')
    ->give(function ($container) {
        $services = $container->tagged('want_configurer');

        return is_array($services) ? $services : iterator_to_array($services);
    });
```

It is easy to get it wrong. It is easy to end up in a situation where it works until it doesn't.

It could be easier.

----

Tagged services are baked into the Container as first-class citizens. By extending the Contextual Binding Builder with a bit of sugar, we can simplify this common task:

```php
$this->app->when(WantFactory::class)
    ->needs('$wantConfigurers') // primitive binding w/ array type
    ->giveTagged('want_configurer');
```

Building on the new variadic constructor argument support, this can also be used to satisfy typed variadic constructor arguments as well:

```php
$this->app->when(WantFactory::class)
    ->needs(WantConfigurer::class) // variadic binding w/ WantConfigurer type
    ->giveTagged('want_configurer');
```

If you are using a variadic constructor argument and are using the type as a tag (something I've started to do more often) you can do this:

```php
$this->app->when(WantFactory::class)
    ->needs(WantConfigurer::class) // variadic binding w/ WantConfigurer type
    ->giveTagged(WantConfigurer::class);
```

~~Taking it to the next level, if someone wants to `giveTagged` and haven't set a `needs`, we can use the tag as the assumed abstract. That looks like this:~~

> ~~$this->app->when(WantFactory::class)->giveTagged(WantConfigurer::class); ~~

This PR makes these last ~~four~~ three configuration options possible.

----

[...]

PR #39819, opened 2021-11-30
[9.x] Add giveConfig to ContextualBindingBuilder

simensen:
This is a follow-up to #36241. I had intended to work on the contract eventually and thought maybe it would just get added eventually by someone else.

> I believe this may require additional work on the contract (?) that I'll do if this gets approved. Would just need to know which branch to target that change on. I can also do the docs on this if it looks like this will be able to be merged. :)

This afternoon I decided to go ahead and add it. In the process, I discovered I'm not the first to send a PR for this addition! Unfortunately, both #36361 and #38925 were rejected.

I'm going to send my PR anyway to open the discussion again. In my initial PR, it was not stated that this was syntactical sugar that would never make it to the contract. Similar to `giveTagged` in #32514, I fully expected we'd be able to get this into the contract eventually.

Going to bring this discussion up in Discord, too. :)

Ultimately, nobody will ever be able to discover this method exists and probably shouldn't be relying on it being available:

image

### DIFF
--- a/src/Illuminate/Contracts/Container/ContextualBindingBuilder.php
+++ b/src/Illuminate/Contracts/Container/ContextualBindingBuilder.php
@@ -27,4 +27,13 @@ interface ContextualBindingBuilder
      * @return void
      */
     public function giveTagged($tag);
+
+    /**
+     * Specify the configuration item to bind as a primitive.
+     *
+     * @param  string  $key
+     * @param  ?string  $default
+     * @return void
+     */
+    public function giveConfig($key, $default = null);
 }


simensen:
Discord discussion here: https://discord.com/channels/297040613688475649/486650494790402061/915044075823050762

taylorotwell:
I doubt many people (if anyone) are writing their own implementation of this interface so will go ahead and merge it.

simensen:
@taylorotwell Thank you so very much!

PR #39821, opened 2021-11-30
[8.x] Fulltext Index

taylorotwell:
This PR adds the ability to create fulltext indexes in your migrations when using MySQL. In the future we may expand this to Postgres.


humaneguy:
Looks like the future is now. Checkout https://github.com/laravel/framework/pull/39875 by @tpetry

PR #39822, opened 2021-11-30
Optimize the execution time of the unique method

procket-org:
When passing a large array, the unique method is very slow. 
This modification can increase the speed by hundreds of times, 
and it is also suitable for most scenarios.


tonylevid:
This is a screenshot of the execution time comparison.

PR #39827, opened 2021-11-30
[8.x] Bump Symfony to 5.4

driesvints:
This bumps Symfony to the latest minor version which will be the last iteration of the 5.x series. Doing this will allow us to remove the versioning workaround in our builds and so we can use the latest features of Symfony 5.4 in Laravel 8.


GrahamCampbell:
This seams reasonable. Going forward, the 5.4.x series will be only one (from 5.x) to get bug and security fixes, and it is (mostly) BC with all 5.x, so I don't see why anyone would object to us doing this. :shipit:

GrahamCampbell (approved):

PR #39831, opened 2021-11-30
[9.x] Adds Route Discovery

nunomaduro:
This pull request is a **proof of concept** for what could potentially be an interesting feature for Laravel 9: **Route Discovery**.

It's known that we use the pattern "Discover by convention" in multiple places on the framework these days, discovering console command classes, container bindings, model policies, register packages, and more. And, seems that developers embrace so well these "Discover" features, that they even forget that you can still register things manually. 

Now, this pull request proposes applying the pattern "Discover by convention" to our HTTP routes. In short, it's about generating routes based on the project's folder/file structure. It's open to debate, but it could be something like this:

```php
<?php // routes/web.php

Route::discover(app_path('Http/Controllers'));
```

``` php
// Assuming the following structure within Http/Controllers:
- Api/Cars.php { public function show($car): Response }
- Photos.php { public function index(): Response }
```

```bash
$ php artisan route:list
+----------+---------------------+---------------+--------------------------------------+
| Method   | URI                 | Name          | Action                                                     |       
+----------+---------------------+---------------+--------------------------------------+
| GET|HEAD | api/cars/{car}      | api.cars.show | App\Http\Controllers\Api\Cars@show   |
| GET|HEAD | photos              | photos.index  | App\Http\Controllers\Photos@index    |
+----------+---------------------+---------------+--------------------------------------+
```

Of course, there are multiple things to investigate such as applying Middlewares, adding extra route parameters, etc. But for now, let me know what do you think about the concept. 👍🏻

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -24,6 +24,7 @@ use JsonSerializable;
 use Psr\Http\Message\ResponseInterface as PsrResponseInterface;
 use ReflectionClass;
 use Symfony\Bridge\PsrHttpMessage\Factory\HttpFoundationFactory;
+use Symfony\Component\Finder\Finder;
 use Symfony\Component\HttpFoundation\Response as SymfonyResponse;
 
 /**
@@ -1307,6 +1308,45 @@ class Router implements BindingRegistrar, RegistrarContract
         $this->container = $container;
 
         return $this;
+    }
+
+    /**
+     * Discover and routes resource controllers by the given path.
+     *
+     * @param  string  $path
+     * @return void
+     */
+    public function discover($path)
+    {
+        $files = (new Finder)->in($path)->files();
+
+        $namespace = app()->getNamespace();
+
+        return collect($files)
+            ->map(fn ($file) => $namespace.str_replace(
+                ['/', '.php'],
+                ['\\', ''],
+                Str::after($file->getRealPath(), realpath(app_path()).DIRECTORY_SEPARATOR)
+            ))->map(fn ($fileClass) => new ReflectionClass($fileClass))
+            ->reject->isAbstract()
+            ->map(function (ReflectionClass $reflection) use ($path) {
+                $methods = collect($reflection->getMethods())
+                    ->filter->isPublic()
+                    ->map->getName()
+                    ->filter(fn ($name) => in_array($name, ['index', 'show', 'store', 'update', 'destroy']));
+
+                $relativeRouteName = rtrim($reflection->getShortName(), 'Controller');
+
+                if (empty($relativeRouteName)) {
+                    return;
+                }
+
+                $routeName = strtolower(ltrim($reflection->getFileName(), $path));
+                $routeName = rtrim($routeName, '.php');
+                $this->resource(Str::kebab($routeName), $reflection->getName(), [
+                    'only' => $methods->all(),
+                ])->names(str_replace('/', '.', $routeName));
+            });
     }
 
     /**
--- a/src/Illuminate/Support/Facades/Route.php
+++ b/src/Illuminate/Support/Facades/Route.php
@@ -35,6 +35,7 @@ namespace Illuminate\Support\Facades;
  * @method static string|null currentRouteName()
  * @method static void apiResources(array $resources, array $options = [])
  * @method static void bind(string $key, string|callable $binder)
+ * @method static void discover(string $path)
  * @method static void model(string $key, string $class, \Closure|null $callback = null)
  * @method static void pattern(string $key, string $pattern)
  * @method static void resources(array $resources, array $options = [])


lukeraymonddowning:
If this goes forward, it might be nice to return a class from discover that allows for similar syntax to an individual route:

```php
Route::discover(app_path('Http/Controllers/Admin'))
  ->middleware(['web', 'admin'])
  ->name('admin');
```

The above example would discover all routes in the Admin directory, assign the `web` and `admin` middleware and prefix the route name with `admin`.

stancl:
For distinguishing the HTTP verb, the syntax could be injecting specific children of `Request`:
```php
public function store(PostRequest $request): Response
```
Basically just simple classes that extend `Request` and are bound to e.g. return the `request()` when its verb matches the expected one (`POST` in `PostRequest`), and `abort()` execution if they're resolved but don't match the `request()`'s HTTP verb.

This is, if you want to avoid using attributes. If attributes would be used here then that opens a lot more possibilities — this tweet by Brent can be a good source of inspiration: https://twitter.com/brendt_gd/status/1443463018910728195

mpociot:
I'm personally not a huge fan of this, as this would be a too-tight coupling between HTTP controller naming and routes which could lead to accidentally renaming a route, just by refactoring its class name.

JustSteveKing:
This is an awesome idea, and without going too much down the Symfony route, PHP 8 attributes could be leveraged nicely here. Using attributes for the HTTP verb and middleware would work well. 

Could it be reverse engineered to also: route:discover as an artisan command - to build a web.php file 🤔 so it could have a fallback?

beko1997:
Agree with @JustSteveKing attributes could be a nice way to go. It would make it possible to define a explicit naming  which will not change based on the controller name. With that in mind it would be very practically for not having to switch between de routes file and Controllers when working on some routes. Nice feature as long as it is opt in 👍

erikaraujo:
I believe this is a cool idea IF combined with attributes.

Maybe we'd have this only discover routes by attributes instead of automatically doing so.

For example, for the following controller nothing would be added to the routes:
```php
class MyController
{
    public function index()
    {
        //...
    }
}
```

But, for the following one, it would:
```php
class MyController
{
    #[Get('my-route-path', 'my-route-name')]
    #[Middlewares(['auth'])
    public function anyMethod()
    {
        //...
    }
}

---

We could even potentially have both, with a `Route::discover` only discovering ones with attributes, and a `Route::discoverDefaults` which generates the route based on convention.

**Comment by hlipnick** [comment]
Perhaps a 'withPath' or similar modifier to allow updating the path of a specific route from a controller?  Decoupling the controller terms (index, store, update, etc) from the user facing url path can sometime be helpful.
A (somewhat contrived) example:
```php
Route::discover(app_path('Http/Controllers'))
  ->withPath([
        'photos.index' => 'photos/all',
        'photos.store' => 'photos/new'
    ])
```

ConsoleTVs:
Althought I like the idea, I think PHP 8's attributes can make this easier.

abdumu:
I think you are reinventing the wheel here, just introduce attributes to Laravel as it’s better, and the mainstream solution to what you are trying to achieve.

Thanks 🥰

Bojan121:
It's cool as a concept, but I hate it. It's way too much relying on convention. I don't like attributes but even that would be a better idea.

shirkerohit:
Wouldn't mind this but, I love the explicit nature of routes.. so less burdon on understanding. Also, the resource are the way that already does this sort of in a way.. 😅

mrbenosborne:
Although it sounds like a good idea on the surface I think being explicit is beneficial in terms of your applications routes.

JosephSilber:
So we've come full circle, back to the CodeIgniter days. Also, early versions of Laravel actually had this as `Route::controllers()`

P.S. See @philsturgeons post from eons ago: https://phil.tech/2013/beware-the-route-to-evil/

thomasilmer:
I prefer the explicit routes, because this feels slightly too much like "magic".

lupinitylabs:
So that would mean with a boilerplate controller created with `artisan make:controller`, this would create the full set of routes even though none of the action methods contain any code? 🤔

mateusjunges:
I think I love and hate at the same time. It's very practical not to have to define routes, but I don't like to rely too much on convention.

stancl:
I think a lot of the arguments against the feature come from "would I use this in my current projects" rather than "would others use this". The PR itself adds a feature, as the upside, and more code to maintain, as the downside. The discussion should be about whether the trade-off is worth it.

To address some other concerns people have brought up, here are my thoughts.

> this would be a too-tight coupling between HTTP controller naming and routes which could lead to accidentally renaming a route, just by refactoring its class name

This is already the case with many things in Laravel. I see how it can be painful though, especially when developing an app that uses API calls from a frontend framework. On the backend you can use `url()->action()` to avoid this issue.

But at the same time, renaming controllers isn't really that much more common than renaming routes. And when **intentionally** using route autodiscovery, you'd expect this behavior. (I think the feature shouldn't be in the default laravel/laravel routes, but instead something users can reach for.)

I think a good comparison to renaming API routes is renaming things in a Livewire component:
- when you rename a component view, all views that use it will need to be changed as well
- when you rename a Livewire component, you need to fix all uses of it in Blade
- when you rename a Livewire method, you need to change `wire:click` calls from the frontend
- when you rename a Livewire property, you need to change the all references (Blade and JS) in the template

So I think the tight coupling, while a valid concern, isn't very unique to this and it's just about having the right expectations when reaching for this feature.

> It's cool as a concept, but I hate it. It's way too much relying on convention

I structure big applications in a DDD way but I don't hate `laravel/laravel` which heavily relies on convention. The thing about this feature is that you **can** use it but you **don't have to**, so there's no reason to argue against it being added on the basis of relying on convention too much. Convention works great for many things, that's why it's the convention.

> While I do love a lot of Laravel's "Discover by convention" features applying it to routes is not one I'm a fan of. Having inherited a number of codebases in a number of programming languages it is always a headache to determine the controller file and method I'm looking for since I have to figure out the convention that is used

"I have to figure out the convention that is used" is the key part and here you see basically no configurability. At most you'll have to check the routes file to see which directories contain the controllers, but that's it. Any codebase using this in a reasonable way will be **more** understandable than codebases using custom named routes.

---

I personally think this feature could be a great addition. At the same time, I wouldn't use it in the majority of my apps. But I still think it can be very valuable for specific use cases **assuming the implementation complexity is worth it**.

Some good use cases I would've had for this in the past:
- Tiny Laravel apps that only have a few routes
    - tools that transform another API's output
    - webhook handlers that execute side effects, e.g. invite a user to a GitHub repo after successful payment
- Debugging / closure routes
    - Closure routes are often used for debugging. With this feature you could create a full class with multiple methods to debug things effectively. The routes could be `/debug/*` and they'd only get registered in non-production environments. Much easier to keep classes organized (compared to closure routes) for debugging purposes.

Customization-wise, I wouldn't go further than configuring things like the HTTP verb. The main point of this is convention over configuration, and for more configuration-heavy things people can use [Spatie's package](https://github.com/spatie/laravel-route-attributes) or just resource route groups.

I suggested using children of `Request` for the HTTP verb in a comment above, because while being a bit non-standard, it'd allow the basic level of configurability that's commonly needed. I love PHP 8 attributes, but also think that introducing them in *any* way here opens a huge can of worms. Which is better handled by a dedicated package, like the one by Spatie.

JustSteveKing:
I think @JosephSilber has a point here - while adding some auto-discovery is a nice touch, we need fallback modes and any routes that can/will be discovered needs a place to sync back to otherwise we are going to fall into issues in application land where we cannot override because of discovery.


My personal thoughts would be an artisan command that looks inside `App\Http\Controllers` and uses PHP 8 Attributes to  get method, middleware and route naming to add into the web.php. This is discovering, but at the same time it could also sync - so if there are routes in the web.php that don't have attributes, we can feedback and say what cannot be found as a warning (not an error as it isn't a requirement).

```bash
php artisan routes:discover
```
I believe @freekmurze or @brendt did something at Spatie with Route Attributes which would boilerplate at least half of the discovery process, but it is key to sync this and not just lean on attribute for those who want to opt out. As Phil said in his article "web.php is documentation" and I am sure a lot of people (myself included) use these routes files to discover what they can about the application, before diving into controllers.

A great idea that would work really well for the route that PHP itself is going down, as long as we cater for both aspects.

src/Illuminate/Routing/Router.php
```diff
@@ -1309,6 +1310,45 @@ public function setContainer(Container $container)
...
+     * @return void
+     */
+    public function discover($path)
+    {
+        $files = (new Finder)->in($path)->files();
+
+        $namespace = app()->getNamespace();
+
+        return collect($files)
+            ->map(fn ($file) => $namespace.str_replace(
```
sebdesign, line 1326:
```php
fn ($file) => (string) Str::of($file->getRealPath())
    ->replaceFirst(realpath(app_path()).DIRECTORY_SEPARATOR, $namespace)
    ->replaceLast('.php', '')
    ->replace('/', '\\')
```

src/Illuminate/Routing/Router.php
```diff
@@ -1309,6 +1310,45 @@ public function setContainer(Container $container)
...
+                ['/', '.php'],
+                ['\\', ''],
+                Str::after($file->getRealPath(), realpath(app_path()).DIRECTORY_SEPARATOR)
+            ))->map(fn ($fileClass) => new ReflectionClass($fileClass))
+            ->reject->isAbstract()
+            ->map(function (ReflectionClass $reflection) use ($path) {
+                $methods = collect($reflection->getMethods())
+                    ->filter->isPublic()
+                    ->map->getName()
+                    ->filter(fn ($name) => in_array($name, ['index', 'show', 'store', 'update', 'destroy']));
```
sebdesign, line 1336:
```php
->intersect(['index', 'show', 'store', 'update', 'destroy'])
```

binaryk:
Stupid Q - do we still have the classic way available?

src/Illuminate/Routing/Router.php
```diff
@@ -1309,6 +1310,45 @@ public function setContainer(Container $container)
...
 
+    /**
+     * Discover and routes resource controllers by the given path.
+     *
+     * @param  string  $path
+     * @return void
+     */
+    public function discover($path)
+    {
+        $files = (new Finder)->in($path)->files();
```
sebdesign, line 1321: This will remove the need to check the filename in line 1340:

```php
$files = (new Finder)->in($path)->files()->notName('Controller.php');
```

src/Illuminate/Routing/Router.php
nunomaduro: This is a POC.

src/Illuminate/Routing/Router.php
nunomaduro: This is a POC.

src/Illuminate/Routing/Router.php
nunomaduro: This is a POC.

mateusjatenee:
> Stupid Q - do we still have the classic way available? 

Yes. This is a new feature that wouldn't impact the current way of doing things.

taylorotwell:
Since this appears to be quite controversial, I think this could be released as a package. Though I am sympathetic to the idea of convention based routing in general if we had more consensus on the feature implementation.

PR #39832, opened 2021-11-30
Adds target status code to HTTP client retry method

OfficialOzzy:
This PR adds the ability to pass an additional argument to the retry method called "$targetStatusCode", see below:

```php
Http::retry(5, 1000, targetStatusCode: 200)->get('http://example.com');
```

The reason for this addition is that the current logic uses *$response->successful()* which quite rightly returns *true* if the status code is between 200 and 299.

This doesn't work if you are polling an endpoint and you're expecting a specific status code within the successful region, ie. 200-299, such as a 200 OK but the endpoint returns a 202 Accepted for example until the endpoint is ready to return its response.


driesvints:
This PR's tests fail and it changes a method signature which is a breaking change. Please send this to master with passing tests. Thanks.

PR #39833, opened 2021-11-30
[9.x] Add target status code to HTTP client retry method

OfficialOzzy:
This PR adds the ability to pass an additional argument to the retry method called "$targetStatusCode", see below:

```php
Http::retry(5, 1000, targetStatusCode: 200)->get('http://example.com');
```

The reason for this addition is that the current logic uses *$response->successful()* which quite rightly returns *true* if the status code is between 200 and 299.

This doesn't work if you are polling an endpoint and you're expecting a specific status code within the successful region, ie. 200-299, such as a 200 OK but the endpoint returns a 202 Accepted for example until the endpoint is ready to return its response.

### DIFF
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -25,7 +25,7 @@ use PHPUnit\Framework\Assert as PHPUnit;
  * @method \Illuminate\Http\Client\PendingRequest contentType(string $contentType)
  * @method \Illuminate\Http\Client\PendingRequest dd()
  * @method \Illuminate\Http\Client\PendingRequest dump()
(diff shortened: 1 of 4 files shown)


DarkGhostHunter:
Sounds fine, but it should allow also for a range of codes. An API may tell me it's 204 or 202 depending on their load.

mrbenosborne:
> Sounds fine, but it should allow also for a range of codes. An API may tell me it's 204 or 202 depending on their load.

That sounds like a good plan, if the maintainers are happy with the PR I can work on allowing a range of codes as you suggested.

taylorotwell:
I would maybe suggest just using the `retry` global helper directly to give you full flexibility in behavior.

PR #39853, opened 2021-12-01
[8.x] Remove unused parameter and redundant comments in `StartSession` middleware

bernardwiesner:
Just removed a dead parameter and some redundant comments. I find the comments are not very useful and add some noise. The code is self explanatory.


DarkGhostHunter:
If I'm extending this middleware what are my options?

taylorotwell:
See above. Would be a breaking change.

PR #39854, opened 2021-12-01
[8.x] [POC] Allow passing in Enum class to migration

jimmypuckett:
Now that we can cast values to `Enum's`, it seems as though it would make sense to allow specifying the `Enum` class in the migration so that the class is the source of truth for the values

If you had the following enum class:

```php
<?php

namespace App\Enums;

enum ServerStatus: string
{
    case Online = 'online';
    case Offline = 'offline';
}
```

You could pass the class to the migration:

```php
<?php

use App\Enums\ServerStatus;

class CreateServersTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::create('servers', function (Blueprint $table) {
            $table->id();

            $table->enum('status', ServerStatus::class);

            $table->timestamps();
        });
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::dropIfExists('servers');
    }
}
```

I have not added any tests as I am not sure if anyone would like this idea, but I would be happy to add if this concept would be of value.

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -91,7 +91,7 @@ class Blueprint
         $this->table = $table;
         $this->prefix = $prefix;
 
-        if (!is_null($callback)) {
+        if (! is_null($callback)) {
             $callback($this);
         }
     }
@@ -129,10 +129,10 @@ class Blueprint
         $this->ensureCommandsAreValid($connection);
 
         foreach ($this->commands as $command) {
-            $method = 'compile' . ucfirst($command->name);
+            $method = 'compile'.ucfirst($command->name);
 
             if (method_exists($grammar, $method) || $grammar::hasMacro($method)) {
-                if (!is_null($sql = $grammar->$method($this, $command, $connection))) {
+                if (! is_null($sql = $grammar->$method($this, $command, $connection))) {
                     $statements = array_merge($statements, (array) $sql);
                 }
             }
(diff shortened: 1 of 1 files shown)


src/Illuminate/Database/Schema/Blueprint.php
```diff
@@ -1000,11 +1000,17 @@ public function boolean($column)
      * Create a new enum column on the table.
      *
      * @param  string  $column
-     * @param  array  $allowed
+     * @param  string|array  $allowed
      * @return \Illuminate\Database\Schema\ColumnDefinition
      */
-    public function enum($column, array $allowed)
+    public function enum($column, string|array $allowed)
```
taylorotwell, line 1006: Union types are PHP 8+ I believe? Laravel 8 supports PHP 7.4.

jimmypuckett: Ya, sorry.  Good catch.  I will loosen the signature like the other methods.

jimmypuckett:
Argh, sorry. I had my VS Code formatted on.  Will fix.  Sorry.

X-Coder264:
This looks like a cool feature at first, but actually I'm not sure if this is a good idea to merge. The enum class can change over the development lifetime of your application (for example adding a new case in it) but the same migration should always produce the same database schema no matter at which time it has been run.

antonkomarev:
Agree with @X-Coder264 
Database changes should not be tied with codebase

jimmypuckett:
Ya, I am very torn about this too.  I am very torn about `enum` in a database as a whole, as it gets tricky when you need to add a value.  I was not sure what is the best way to deal with adding cases to the enum class & keeping the DB in sync or even if that is the responsibility of the framework or the responsibility of the programmer.

Great feedback.  Hopefully, someone has a good suggestion.

driesvints:
What's the status on this PR? Can this be reviewed again?

jimmypuckett:
> What's the status on this PR? Can this be reviewed again?

From a coding perspective, it is working, but I would be happy to add any tests if people approve/like the POC.

driesvints:
I'll let Taylor review this again.

taylorotwell:

jimmypuckett:
Totally understand.  I think that there is a need here, but just don't know if we have a good solution.  Maybe someone will come up with a great idea.

PR #39859, opened 2021-12-01
[6.x] Update lock return type

maxsky:


driesvints:
This definitely isn't correct. The return type should always be a boolean.

maxsky:
> This definitely isn't correct. The return type should always be a boolean.

maybe return by `$callback`

```php
public function block($seconds, $callback = null)
{
    $starting = $this->currentTime();

    while (! $this->acquire()) {
        usleep(250 * 1000);

        if ($this->currentTime() - $seconds >= $starting) {
            throw new LockTimeoutException;
        }
    }

    if (is_callable($callback)) {
        try {
            return $callback();
        } finally {
            $this->release();
        }
    }

    return true;
}
```

PR #39860, opened 2021-12-01
[9.x] Use environment method to determine which environment is

oliverquynh:
`Illuminate\Contracts\Foundation\Application::environment` already has a purpose to check which environment the application is running in. We should keep it consistent by using this method instead of manually comparisons.

### DIFF
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -584,7 +584,7 @@ class Application extends Container implements ApplicationContract, CachesConfig
      */
     public function isLocal()
     {
-        return $this['env'] === 'local';
+        return $this->environment('local');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Illuminate/Foundation/Application.php
```diff
@@ -605,9 +605,7 @@ public function isProduction()
      */
     public function detectEnvironment(Closure $callback)
     {
-        $args = $_SERVER['argv'] ?? null;
```
seriquynh, line 608: We needn't this temporary variable.

seriquynh:
I can't figure out why there are a few errors of testing. They don't seem to be related to this change. For example,

```
1) Illuminate\Tests\Database\DatabaseMigrationMigrateCommandTest::testBasicMigrationsCallMigratorWithProperArguments
Mockery\Exception\InvalidCountException: Method paths(<Any Arguments>) from Mockery_96_Illuminate_Database_Migrations_Migrator should be called
 exactly 1 times but called 0 times.
```
 I will try in another time later.

PR #39864, opened 2021-12-02
[9.x] Register `db.schema` as deferred service

crynobone:
`Illuminate\Database` only has two service providers, `DatabaseServiceProvider` (eager-loaded) and `MigrationServiceProvider` (deferred). Don't think it worth it to add another service provider specifically for `Schema`.


taylorotwell:
What does this fix?

crynobone:
Laravel 9 change `Schema` behavior from `singleton` to `bind`. So, given the following:

```php
Schema::defaultStringLength(30);
Schema::morphUsingUuids();
```

The above code would resolve `db.schema` twice, however seems like this isn't the best solution for above problem. Will review this again at another time.

PR #39870, opened 2021-12-02
[8.x] Add support to convert Boolean-like strings to actual Boolean

nagi1:
This simple PR adds the convenient ability to convert bool-like string i.e. `True` to actual bool type `true`.

This can be useful when dealing with API that uses only strings to communicate where "bool" fields from http responses isn't guaranteed to be "bool". 

This simple situation when you thought that you are dealing with bool and instead you got "True"!

Before

```php
$overwrite = (bool) $request->input('overwrite', 'false');

// when "overwrite" is "false", $overwrite will be always true
if($overwrite) {
   Storage::delete($file);
}

```

After

```php
$overwrite = Str::toBool($request->input('overwrite', 'false'));

if($overwrite) {
   Storage::delete($file);
}
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -845,6 +845,31 @@ class Str
     }
 
     /**
+     * Convert boolean-like strings to actual boolean.
+     *
+     * @param  string|bool  $exp
+     * @return bool
+     */
+    public static function toBool($exp)
+    {
+        if (is_bool($exp)) {
+            return $exp;
+        }
+
+        $exp = strtolower($exp);
+
+        if ($exp === 'true') {
+            return true;
+        }
+
+        if ($exp === 'false') {
+            return false;
+        }
+
+        return false;
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


derekmd:
A `Request::boolean()` method exists for converting HTTP parameter values:

https://github.com/laravel/framework/blob/72aeac2e55cfe3e169a7e5b5c84222c284a04f2b/src/Illuminate/Http/Concerns/InteractsWithInput.php#L294-L297

```php
Request::create('', 'GET', ['foo' => 'true'])->boolean('foo')
// true

Request::create('', 'GET', ['foo' => 'on'])->boolean('foo')
// true

Request::create('', 'GET', ['foo' => '1'])->boolean('foo')
// true

Request::create('', 'GET', ['foo' => 'false'])->boolean('foo')
// false

Request::create('', 'GET', ['foo' => 'off'])->boolean('foo')
// false

Request::create('', 'GET', ['foo' => '0'])->boolean('foo')
// false
```

taylorotwell:

PR #39875, opened 2021-12-02
[8.x] Fulltext index for PostgreSQL

tpetry:
As promised on Twitter, I want to help with the PostgreSQL support for fulltext search.

But as fulltext searching is no SQL standard, every database has an entirely different implementation. The most significant difference is that PostgreSQL fulltext searching is language-dependent because the stemming algorithms are always only designed for a specific language. Also, adding a fulltext index for multiple columns is different. The vectors for multiple columns need to be combined for the index creation. But when querying the database the exact same syntax for combining the vectors needs to be used for the query or the index will not be used. As there is no query builder method to build these conditions, I restricted the usage to a single column for the moment. With an official way to query the index the restriction could be lifted.

So the following changes had to be made:

- An optional language parameter was added to `Blueprint::fulltext()`
- The PostgreSQL driver will fail if no language was provided
- The PostgreSQL driver will fail when trying to add a fulltext index with multiple columns

### DIFF
--- a/src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
@@ -188,17 +188,16 @@ class PostgresGrammar extends Grammar
      */
     public function compileFulltext(Blueprint $blueprint, Fluent $command)
     {
-        if (is_null($command->language)) {
-            throw new RuntimeException('The PostgreSQL driver requires a language for fulltext index creation.');
-        }
+        $language = $command->language ?: 'english';
+
         if (count($command->columns) > 1) {
-            throw new RuntimeException('The PostgreSQL driver does not support fulltext index creation with multiple columns. Please create the index by yourself.');
+            throw new RuntimeException('The PostgreSQL driver does not support fulltext index creation using multiple columns.');
         }
 
         return sprintf('create index %s on %s using gin (to_tsvector(%s, %s))',
             $this->wrap($command->index),
             $this->wrapTable($blueprint),
-            $this->quoteString($command->language),
+            $this->quoteString($language),
             $this->wrap($command->columns[0])
         );
     }
--- a/src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
@@ -4,6 +4,7 @@ namespace Illuminate\Database\Schema\Grammars;
 
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Fluent;
+use RuntimeException;
 
 class PostgresGrammar extends Grammar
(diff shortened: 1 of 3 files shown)


src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
```diff
@@ -176,6 +177,32 @@ public function compileIndex(Blueprint $blueprint, Fluent $command)
...
+     *
+     * @param  \Illuminate\Database\Schema\Blueprint  $blueprint
+     * @param  \Illuminate\Support\Fluent  $command
+     * @return string
+     *
+     * @throws \RuntimeException
+     */
+    public function compileFulltext(Blueprint $blueprint, Fluent $command)
+    {
+        if (null === $command->language) {
```
faissaloux, line 191:
```suggestion
        if (is_null($command->language)) {
```

tpetry: A specific reason why you suggest replacing a language construct to a function? The behavior will be the same.

faissaloux: Consistency

tpetry: Both ways are used in laravel. I can‘t say which one is the preferred one 🤷‍♂️

mfn: I see Taylor often converting them to the `is_null` variant.

I don't understand why one does this manually and there's no fixer in place doing this automagically 😅

CHANGELOG-8.x.md
```diff
@@ -2,6 +2,8 @@
 
 ## [Unreleased](https://github.com/laravel/framework/compare/v8.74.0...8.x)
 
+### Added
+- Added Fulltext Index support for PostgreSQL
```
driesvints, line 6: Please remove this again.

src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
driesvints: @mfn what fixer do you recommend?

CHANGELOG-8.x.md
tpetry: Done

src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
mfn: @driesvints none actually 😅

sorry for the noise, was thinking about https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/language_construct/is_null.rst but that's the other way around and Style CI can do this too but it's not he CS of the project

PR #39877, opened 2021-12-03
[8.x] Test Improvements

crynobone:
* Replace Capsule (using in-memory sqlite) test to extends `DatabaseTestCase`.
* Replace mocking `Container` to asserts on `events` and use `Event::fake()` 
* Purge all database connections on `tearDown()`.


driesvints (approved):

driesvints:
Nice @crynobone! Now we can re-enable the Postgres build on master after this is merged.

PR #39882, opened 2021-12-03
[8.x] Cleanup request test

oliverquynh:
There are two things I try to clean up in this PR:

1. Request::create() sets the second parameter (method) as "GET" by default. If a test doesn't care about the method, we needn't to pass it explicitly.

2. prefers method test doesn't look good to read. Using data provider make it easier to read.

For example, Within the case ['application/json', ['json'], 'json'], we can describe it like this:

> If the request has the header **accept** to be 'application/json', the prefers method that takes ['json'] as it's parameter will return 'json';

### DIFF
--- a/tests/Http/HttpRequestTest.php
+++ b/tests/Http/HttpRequestTest.php
@@ -757,29 +757,41 @@ class HttpRequestTest extends TestCase
         $this->assertEquals($payload, $data);
     }
 
-    public function testPrefers()
+    public function getPrefersCases()
     {
-        $this->assertSame('json', Request::create('/', 'GET', [], [], [], ['HTTP_ACCEPT' => 'application/json'])->prefers(['json']));
-        $this->assertSame('json', Request::create('/', 'GET', [], [], [], ['HTTP_ACCEPT' => 'application/json'])->prefers(['html', 'json']));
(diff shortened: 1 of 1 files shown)


driesvints:
Any reason why you sent this in to master and not to 8.x?

seriquynh:
- Firstly, I think 8.x is the stable version and it avoids these MRs like this.
- Secondly, I remember that Taylor often prevents these things from being merged into the current stable branch. I may be wrong.
- Finally if it's not a bug and I am not sure which branch I should target, I will send to the master branch.

driesvints:
@seriquynh these are only changes to the test suite. Nothing will be broken by these. I think you're free to send this in to 8.x

seriquynh:
@driesvints Thank you for you help! I will send this to 8x.

I'm sorry if I miss something. However, This MR is absolute not a minor or major change, it is not also a bug fix. Do you think Laravel should add one more a description to https://laravel.com/docs/8.x/contributions#which-branch  like this?

> Improvements such as fixing docblock, clean up code/tests that do not introduce any behavior changes should be send to the current release (stable) branch.

(just an example)

driesvints:
@seriquynh this falls under "minor" imo.

driesvints (approved):

seriquynh:
@driesvints I am sorry! I click accidently the request review button.

driesvints (approved):

driesvints:
@seriquynh no worries 👍

PR #39395, opened 2021-10-28
[9.x] Opened test traits initialization

DarkGhostHunter:
## What?

Test traits now can be initialized automatically using `setUp{trait}`, `created{trait}` and `destryoing{trait}`, making traits shareable across multiple Test Classes.

* `setUp{trait}` runs as soon the application is already created.
* `created{trait}` runs after all traits have been set up.
* `destroying{trait}` runs before the application is shut down.

## How?

The main `TestCase::setUpTraits()` has been changed to get an array of all traits used recursively by the Test Case. While this may take the main Test Case Concerns (like `InteractsWithContainer()`, the difference in performance is negligible, specially if OPcache is enabled on CLI.

```php
/**
 * Boot the testing helper traits.
 *
 * @return array
 */
protected function setUpTraits()
{
    // ...

    foreach (array_reverse($uses) as $trait => $int) {
        if (method_exists($trait, $method = 'setUp' . $trait)) {
            $this->{$method}();
        }

        if (method_exists($trait, $method = 'created' . $trait)) {
            $this->afterApplicationCreated([$this, $method]);
        }

        if (method_exists($trait, $method = 'destroying' . $trait)) {
            $this->beforeApplicationDestroyed([$this, $method]);
        }
    }

    return $uses;
}
```

## Why?

To allow reusing the same trait across multiple Test Classes out of the box. This traits setups everything with just including it in the Test Classes the developer may need it.

```php
<?php

namespace Tests\Feature\Concerns;

use App\Models\User;

trait WithSubscribedUser
{
    protected $user;
    
    protected $subscriptionsOpen = true; 
    
    protected function setUpWithSubscribedUser()
    {
        $this->app['config']->set('subscriptions.open', $this->subscriptionsOpen);
    }
    
    protected function createdWithSubscribedUser()
    {
        $this->user = User::factory()->subscribed()->createOne();
        
        $this->actingAs($this->user);
    }
    
    protected function destroyingWithSubscribedUser()
    {
        // Always assert the user is still subscribed whatever may happen!
        static::assertTrue($this->user->subscription->exists);

        // Free up the user from memory.
        $this->user = null;
    }
}
```

Later in their Test Classes:

```php
<?php

namespace Tests\Feature;

use App\Models\User;
use Illuminate\Foundation\Testing\RefreshDatabase;
use Tests\TestCase;

class ExampleTest extends TestCase
{
    use RefreshDatabase;
    use Concerns\WithSubscribedUser;

    protected $subscriptionsOpen = false;

[...]

PR #39883, opened 2021-12-03
[8.x] Adds first-level test traits setUp/tearDown (opt-in)

DarkGhostHunter:
## What?

Allows the developer to include and reuse traits that automatically set-up and tear-down in each test, when declared at first-level.

Given a reusable test `WithExamplePost`:

```php
namespace Tests\Feature\Concerns;

use App\Models\Post;

trait WithExamplePost
{    
    protected $post;

    protected function setUpWithExamplePost()
    {
        Post::factory()->createOne();
    }
    
    protected function tearDownWithExamplePost()
    {
        $this->post = null;
    }
}
```

The test class implementing `InitializeTraits`, will automatically call `setUpWithExamplePost` after the application is created, and `tearDownWithExamplePost` before is destroyed.

```php
<?php

namespace Tests\Feature;

use App\Models\User;
use Illuminate\Contracts\Testing\InitializeTraits;
use Tests\TestCase;

class PostTest extends TestCase implements InitializeTraits
{
    use RefreshDatabase;
    use Concerns\WithSubscribedUser;

    public function test_post_is_visible()
    {
        $this->get('post/' . $this->post->id)->assertOk();
    }
}
```

Since this is an additive opt-in feature, no test will break unless the interface is added on tests classes.

## Why?

There are three keys for this implementation:

- Reusable traits out of the box, no need to implement the feature by the developer.
- Opt-in
- Runs after the application is created, and before it's destroyed.
- Only works for **first-level traits**. Nested traits are not affected (like in PHPUnit), so initialization is _controllable_.

## BC?

Nope, opt-in additive.

## Notes

I decided to use an interface-check in the `TestCase` rather than adding a trait because:

- Avoids getting the trait into the initializers array (or even alone by itself, making no sense)
- Avoids polluting a test properties or methods if it were to check them.

## Didn't I see this before?

[Kinda](https://github.com/laravel/framework/pull/39395)

### DIFF
--- a/src/Illuminate/Contracts/Testing/InitializeTraits.php
+++ b/src/Illuminate/Contracts/Testing/InitializeTraits.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Illuminate\Contracts\Testing;
+
+interface InitializeTraits
+{
+    //
+}
--- a/src/Illuminate/Foundation/Testing/TestCase.php
+++ b/src/Illuminate/Foundation/Testing/TestCase.php
@@ -5,6 +5,7 @@ namespace Illuminate\Foundation\Testing;
 use Carbon\Carbon;
 use Carbon\CarbonImmutable;
 use Illuminate\Console\Application as Artisan;
+use Illuminate\Contracts\Testing\InitializeTraits;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Queue\Queue;
 use Illuminate\Support\Facades\Facade;
@@ -94,6 +95,10 @@ abstract class TestCase extends BaseTestCase
             $callback();
         }
 
+        if ($this instanceof InitializeTraits) {
+            $this->initializeFirstLevelTraits();
+        }
+
         Model::setEventDispatcher($this->app['events']);
 
         $this->setUpHasRun = true;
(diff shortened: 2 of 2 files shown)


inxilpro:
I think this would be a nice addition. I did want to mention that you can accomplish this right now with the following:

```php
trait MyTrait
{
  /** @before */
  protected function setUpMyTrait()
  {
    $this->afterApplicationCreated(function() {
      // Perform setup (must happen in afterApplicationCreated because
      // everything with the @before annotation runs BEFORE setUp() is called
    });
  }
  
  /** @after */
  protected function tearDownMyTrair()
  {
    // Perform tear down (no need for closure, because this runs before tearDown()
  }
}
```

But I think the confusion about how the `@before` and `@after` hooks are ordered makes it a pain—it'd be nice for it to just be handled by the framework.

DarkGhostHunter:
I also do it in some of my testing traits but feels less straightforward using annotations.

driesvints:
Ah I didn't immediately remember that you could do this using the annotations. In that case I wouldn't add this to the core framework. There's also a slight chance of breakages with this PR if people have already defined such methods.

DarkGhostHunter:
If there is a minimal collision risk, I can push this to 9.x then, and remove the interface.

inxilpro:
> Ah I didn't immediately remember that you could do this using the annotations. In that case I wouldn't add this to the core framework. There's also a slight chance of breakages with this PR if people have already defined such methods.

I still feel like the `@before` and `@after` annotation behavior is confusing due to the way they're applied. Adding framework-level support for dynamically applying traits in a predictable way would be great ([it's something I proposed a few years ago, as well](https://github.com/laravel/framework/pull/21126))

I suppose another option would be to add a new `@afterApplicationCreated` and `@beforeApplicationDestroyed` annotation that's Laravel-specific…

inxilpro:
I actually just dug around some more and I was wrong, the `@after` annotation will cause the method to be called after the test class' `tearDown()` method, which means you won't have access to any of the mocked Laravel app. You're better off adding `afterApplicationCreated` and `beforeApplicationDestroyed` closures inside a `@before` hook if you need the Laravel app to still be set up.

taylorotwell:

PR #39884, opened 2021-12-03
 [8.x] Make worker avoid timeout mechanism when queue is empty

ryanlcr55:
When the sleep option bigger than (or equal to) timeout option, worker process will be killed.
In my opinion, when queue is empty, worker should sleep well until next job is pushed.
And the timeout mechanism should only restrict the job processing time.
So I moved registerTimeoutHandler in the if statement, which its condition is having a job.

image

### DIFF
--- a/src/Illuminate/Queue/Worker.php
+++ b/src/Illuminate/Queue/Worker.php
@@ -159,14 +159,14 @@ class Worker
                 $this->manager->connection($connectionName), $queue
             );
 
-            if ($supportsAsyncSignals) {
-                $this->registerTimeoutHandler($job, $options);
-            }
-
             // If the daemon should run (not in maintenance mode, etc.), then we can run
             // fire off this job for processing. Otherwise, we will need to sleep the
             // worker so no more jobs are processed until they should be processed.
             if ($job) {
+                if ($supportsAsyncSignals) {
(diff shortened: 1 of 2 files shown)


src/Illuminate/Queue/Worker.php
```diff
@@ -243,7 +243,7 @@ protected function resetTimeoutHandler()
     /**
      * Get the appropriate timeout for the given job.
      *
-     * @param  \Illuminate\Contracts\Queue\Job|null  $job
+     * @param  \Illuminate\Contracts\Queue\Job  $job
```
driesvints, line 246: Revert these docblock changes.

ryanlcr55: @driesvints done!

ryanlcr55: @driesvints done!

GrahamCampbell:
Isn’t this a breaking change?

taylorotwell:

ryanlcr55:
@taylorotwell, 
Thanks for your reply.
But I did encounter this problem.
So I am eager to fix it.
I will be really appreciate if I could get some advices to fix this pr from you.

Thanks again for your kindly help!

PR #39885, opened 2021-12-03
[8.x] Add the possibility to translate arrays and do replacements

gilbertorussi:
Currently it is possible to translate arrays, but only if there's nothing to replace in them.

This makes possible to translate an array of results and make replacements recursively.

```
[
  "foo" => [
    "bar" => ":baz",
  ],
]
```

I already can  call `__('foo')` and it would return `["bar" => ":baz"]`,

Why can't I call `__('foo', ['baz' => 'one'])` and get back `["bar" => "one"]`?

### DIFF
--- a/src/Illuminate/Translation/Translator.php
+++ b/src/Illuminate/Translation/Translator.php
@@ -238,7 +238,7 @@ class Translator extends NamespacedItemResolver implements TranslatorContract
         if (is_string($line)) {
             return strtr($line, $shouldReplace);
         }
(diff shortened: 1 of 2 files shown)


taylorotwell:

gilbertorussi:
@taylorotwell 
Before commit [#024067f](https://github.com/laravel/framework/commit/024067fd56b227d74ce0b70d6b78ef5ad7095d6d), translating arrays with replacements used to work.

But thanks anyway, it is probably a code smell on my end.

PR #39901, opened 2021-12-06
[8.x] Added `--migration` argument to specify a migration to be used with the `migrate:rollback` command

promatik:
~I'm not sure if this ever worked before, but by reading the code I believe it never did, at least it doesn't right now.
Although, I found some [stack overflow answers](https://stackoverflow.com/questions/30287896/rollback-one-specific-migration-in-laravel) where some suggest the usage of `--path` arg to specify a migration to be rolled back.~

~This PR allows to specify one or multiple `--path` with the migrations we want to rollback.~

---
This PR adds `--migration` argument to specify one or many migrations we want to rollback.

```cmd
php artisan migrate:rollback --migration="database/migrations/first.php" --migration "database/migrations/second.php"
```

```cmd
Rolling back: first
Rolled back:  first (49.60ms)
Rolling back: second
Rolled back:  second (450.67ms)
```

This is particularly useful to rollback a single migration, in the middle of any batch.

### DIFF
--- a/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
@@ -57,7 +57,11 @@ class RollbackCommand extends BaseCommand
 
         $this->migrator->usingConnection($this->option('database'), function () {
             $this->migrator->setOutput($this->output)->rollback(
-                $this->getMigrationPaths(), $this->options()
+                $this->getMigrationPaths(), [
+                    'pretend' => $this->option('pretend'),
+                    'step' => (int) $this->option('step'),
+                    'path' => $this->option('path'),
+                ]
             );
         });
 
--- a/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
@@ -57,10 +57,7 @@ class RollbackCommand extends BaseCommand
 
         $this->migrator->usingConnection($this->option('database'), function () {
             $this->migrator->setOutput($this->output)->rollback(
-                $this->getMigrationPaths(), [
-                    'pretend' => $this->option('pretend'),
-                    'step' => (int) $this->option('step'),
-                ]
+                $this->getMigrationPaths(), $this->options()
             );
         });
 
--- a/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
@@ -60,7 +60,7 @@ class RollbackCommand extends BaseCommand
                 $this->getMigrationPaths(), [
                     'pretend' => $this->option('pretend'),
                     'step' => (int) $this->option('step'),
-                    'path' => $this->option('path'),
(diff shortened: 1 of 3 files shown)


src/Illuminate/Database/Console/Migrations/RollbackCommand.php
```diff
@@ -57,10 +57,7 @@ public function handle()
 
         $this->migrator->usingConnection($this->option('database'), function () {
             $this->migrator->setOutput($this->output)->rollback(
-                $this->getMigrationPaths(), [
-                    'pretend' => $this->option('pretend'),
-                    'step' => (int) $this->option('step'),
-                ]
+                $this->getMigrationPaths(), $this->options()
```
promatik, line 60: Since we now need the `--path` also, I think it's easier to send all the options, instead of rebuild an array with the same unmodified data.

src/Illuminate/Database/Migrations/Migrator.php
```diff
@@ -248,6 +248,12 @@ protected function getMigrationsForRollback(array $options)
             return $this->repository->getMigrations($steps);
         }
 
+        if ($path = $options['path'] ?? []) {
+            return collect($path)->map(function ($m) {
+                return (object) ['migration' => (string) \Str::of($m)->basename('.php')];
```
derekmd, line 253: `use Illuminate\Support\Str;` is already defined in this file so `Str::of($m)->basename('.php')` without `\` is OK.

This also wouldn't be able to assume root facade `Str`  is defined since upgraded apps made not have added it or userland may have decided to remove some aliases:

https://github.com/laravel/laravel/blob/399d435c4f0b41a5b6d3e14894195f9196d36bb8/config/app.php#L228

promatik:
```suggestion
                return (object) ['migration' => (string) Str::of($m)->basename('.php')];
```

promatik: Fixed, thank you @derekmd.

driesvints:
Seems like this breaks the test suite. Please fix the tests and mark this as ready.

promatik:
Hi @derekmd, hi @driesvints!

I noticed there is a problem with this part of code. `--path` argument is available on all migration commands, it is actually in the `BaseCommand`, it is used to define the working folder(s) where Laravel will search for migrations.

It would be possible to use path to define specific migrations to be rolled back, and I did it in the beginning, but many tests were failing because by changing `--path` argument behavior, we change the way those functions work for the worst, at some point I noticed I had to mess a lot the tests for this solution to work.

To overcome this, I created a new argument, `--migration`, it is also an array, so multiple migrations can be specified.

Since `--path` was not working for defining specific migrations, I believe this is not a breaking change, this is more like a new feature.
I've also added tests for it.

Sample command;
```php
php artisan migrate:rollback
  --migration="2014_10_12_100000_create_password_resets_table.php"
  --migration="2014_10_12_000000_create_users_table.php"
```


Please let me know what you think about this 🙌

taylorotwell:
I'm not sure I want to go down the rabbit hole of rolling back and executing migrations out of order.

PR #39906, opened 2021-12-06
Parent replacement fix

taylorotwell:
This adjusts how we replace the parent placeholder in Blade views. The parent placeholder hash is now salted if possible.


buismaarten:
LGTM!

PR #39907, opened 2021-12-06
add countable interface to eloquent factory sequence

acurrieclark:
I recently submitted [a discussion](https://github.com/laravel/framework/discussions/39655) regarding this, but following [Dries Vints' request on twitter today](https://twitter.com/driesvints/status/1467838887921958913), I thought I would just go ahead and submit a PR.

Essentially, I tried to call `count($sequence)` on an instance of `Illuminate\Database\Eloquent\Factories\Sequence` and was surprised to find that it wasn't, well, `\Countable`.

Given the existence of the `$count` class variable, I thought it might be a sensible addition.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/Sequence.php
+++ b/src/Illuminate/Database/Eloquent/Factories/Sequence.php
@@ -2,7 +2,9 @@
 
 namespace Illuminate\Database\Eloquent\Factories;
 
-class Sequence
+use Countable;
+
+class Sequence implements Countable
 {
     /**
      * The sequence of return values.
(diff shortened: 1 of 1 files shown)


taylorotwell:

acurrieclark:
Most welcome. And an enormous thank you to you too.

src/Illuminate/Database/Eloquent/Factories/Sequence.php
```diff
@@ -37,6 +39,16 @@ public function __construct(...$sequence)
         $this->count = count($sequence);
     }
 
+    /**
+     * Get the current count of the sequence items.
+     *
+     * @return int
+     */
+    public function count()
```
X-Coder264, line 47: @acurrieclark @taylorotwell The `int` return typehint is missing so this will throw a deprecation on PHP 8.1. Since this is completely new code there's no need to add the `ReturnTypeWillChange` attribute when the return type can be added instead.

https://wiki.php.net/rfc/internal_method_return_types

driesvints: Thanks. I fixed that.

PR #39916, opened 2021-12-06
[8.x] Allows to use a closure to group nullable columns

fernandokbs:
This PR introduces the use of a closure to group a nullable columns,
Sometimes we need to create columns and every one needs to be defined as nullable, instead of appends the method nullable() for every column, we can group it by using the clousure nullable

I think this should make the developer's life easier by avoiding to define nullable every time

```php
  Schema::create('posts', function (Blueprint $table) {
      $table->nullable(function($table) {
          $table->integer('foo'); // nullable
          $table->string('bar'); // nullable
      });
      $table->string('note'); // not nullable
  });
```

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -79,6 +79,13 @@ class Blueprint
     public $after;
 
     /**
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #39920, opened 2021-12-07
[8.x] Pusher 6.x+ does not support TLS options anymore

rennokki:
### The problem
Pusher 6.x+ swapped their support from CURLOPT to Guzzle, and removed the TLS options, and this brought a [series of issues](https://github.com/pusher/pusher-http-php/issues/313) for TLS users that use self-signed certs in servers like [Laravel WebSockets](https://github.com/beyondcode/laravel-websockets/issues/879) and [soketi](https://github.com/soketi/soketi/issues/191), locally or in private networks.

### Potential fixes 

There may be [an approach to fix this](https://github.com/beyondcode/laravel-websockets/issues/879#issuecomment-987780296) by overwriting the `BroadcastManager` class at bootstrapping, but I had to find a way to overcome this by solving it at the root so that it isn't required.

Along the way, pre-6.x Pusher support has to be deprecated, and this might lead to issues in 8.x and should be in 9.x, but at the same time, it should be seen as a fix to support the new Pusher version's client argument.

This PR initializes Pusher with a Guzzle client whose options are defined in `config/broadcasting.php`. `curl_options` are no longer used, so in case it's defined, it should be removed:

```php
'pusher' => [
    'driver' => 'pusher',
    'key' => env('PUSHER_APP_KEY', 'app-key'),
    'secret' => env('PUSHER_APP_SECRET', 'app-secret'),
    'app_id' => env('PUSHER_APP_ID', 'app-id'),
    'options' => [
        // ...
    ],
    'client_options' => [
        // You can pass request options: https://docs.guzzlephp.org/en/stable/request-options.html
        'verify' => false, // to disable TLS checks
    ],
],
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -148,7 +148,7 @@
         "phpunit/phpunit": "Required to use assertions and run tests (^8.5.19|^9.5.8).",
         "predis/predis": "Required to use the predis connector (^1.1.9).",
         "psr/http-message": "Required to allow Storage::put to accept a StreamInterface (^1.0).",
-        "pusher/pusher-php-server": "Required to use the Pusher broadcast driver (^4.0|^5.0|^6.0|^7.0).",
+        "pusher/pusher-php-server": "Required to use the Pusher broadcast driver (^6.0|^7.0).",
         "symfony/cache": "Required to PSR-6 cache bridge (^5.4).",
         "symfony/filesystem": "Required to enable support for relative symbolic links (^5.4).",
         "symfony/psr-http-message-bridge": "Required to use PSR-7 bridging features (^2.0).",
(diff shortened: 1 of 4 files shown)


francoism90:
@rennokki Many thanks!

This is really helpful and needed to allow self-signed certificates on local/non-production environments. Even on LAN, you may need to use SSL/TLS, including testing.

GrahamCampbell:
I don't think it makes sense to make this change in Laravel 8.

GrahamCampbell:
Dropping old versions of the pusher sdk is a major breaking change because of the way it cannot be enforced easily.

GrahamCampbell:
I'd honestly recommend just not using self-signed stuff. There are ways around this, even for local dev.

francoism90:
@GrahamCampbell I do understand this is not a support forum, but how should this work for things like this?

rennokki:
@GrahamCampbell At least would make sense to find a way to be able to initialize the custom Guzzle client passed to the Pusher instance for custom Guzzle options. :)

taylorotwell:
This will need to go to Laravel 9 (`master` branch).

PR #39921, opened 2021-12-07
[8.x] Add method `filterNulls()` to Arr

chu121su12:
This PR adds array helper to filter array from nulls, compared to base `array_filter` which filters on falsy values.


JurianArie:
Collections have a method called `whereNotNull` https://laravel.com/docs/8.x/collections#method-wherenotnull. Maybe it should be named the same here?

chu121su12:
I’m open to name change, however changing to `whereX` may lead to implied dynamic parameter support based on `operatorForWhere`.

Personally I think more advanced `where` uses is reserved for collection instead of arr.

PR #39922, opened 2021-12-07
[9.x] Adds `WithoutModelEvents` trait for running seeds without Model Events

nunomaduro:
This pull request proposes a creation of a public trait — `WithoutModelEvents` — that prevents **Model Events** from being executed while running seeds.

Laravel developers often use Model events for firing queueable or not queueable tasks, that will send emails, index their Laravel Scout data, logging, etc. And, while useful, I don't necessarily want these tasks to be executed by my seeders.

So, having this `WithoutModelEvents` trait, which is very similar to the testing trait `WithoutModelEvents`, it's super useful, especially when using Factories on my seeds:

```php
<?php

namespace Database\Seeders;

use Illuminate\Database\Seeder;
use Illuminate\Database\Console\Seeds\WithoutModelEvents;

class DatabaseSeeder extends Seeder
{
    use WithoutModelEvents;

    public function run()
    {
        // these models are created, but no Model Events are
        // fired thanks to the `WithoutModelEvents` trait...
        User::factory(10)->create(); 
    }
}
``` 

Regarding the implementation there are two things to keep in mind:
- First, we are **only disabling Model Events**, and not events in general.
- Second, when using the method `$this->call` to call multiple seeders, the `WithoutModelEvents` propagates down. Let's see two examples:

1. In this example, the DatabaseSeeder is called, and because it contains the `WithoutModelEvents` trait, no Models Events are ever fired when using this seeder:
```php
<?php

namespace Database\Seeders;

use Illuminate\Database\Seeder;
use Illuminate\Database\Console\Seeds\WithoutModelEvents;

class DatabaseSeeder extends Seeder
{
    use WithoutModelEvents;

    public function run()
    {
        User::factory(10)->create(); // no Model Events are fired... 🔇

        $this->call([
            UserSeeder::class, // no Model Events are fired... 🔇
            UserWithoutModelEventsSeeder::class, // no Model Events are fired... 🔇
        ]);
    }
}
``` 

2. In this example, the DatabaseSeeder is called, and it do not contain the `WithoutModelEvents` trait, so Model Events get fired by the `UserSeeder::class` but not by `UserWithoutModelEventsSeeder::class`:
```php
<?php

namespace Database\Seeders;

use Illuminate\Database\Seeder;

class DatabaseSeeder extends Seeder
{
    public function run()
    {
        User::factory(10)->create(); // Model Events are fired... 🔈✅

        $this->call([
            UserSeeder::class, // Model Events are fired... 🔈✅
            UserWithoutModelEventsSeeder::class,  // no Model Events are fired... 🔇
        ]);
    }
}
```

### DIFF
--- a/src/Illuminate/Database/Console/Seeds/WithoutModelEvents.php
+++ b/src/Illuminate/Database/Console/Seeds/WithoutModelEvents.php
@@ -7,7 +7,7 @@ use Illuminate\Database\Eloquent\Model;
 trait WithoutModelEvents
 {
     /**
-     * Prevent model event handles from being executed by the given callback.
+     * Prevent model events from being dispatched by the given callback.
      *
      * @param  callable  $callback
      * @return callable
--- a/src/Illuminate/Database/Console/Seeds/WithoutModelEvents.php
+++ b/src/Illuminate/Database/Console/Seeds/WithoutModelEvents.php
@@ -4,15 +4,15 @@ namespace Illuminate\Database\Console\Seeds;
 
 use Illuminate\Database\Eloquent\Model;
 
-trait WithoutEvents
+trait WithoutModelEvents
 {
     /**
-     * Prevent all event handles from being executed by the given callback.
+     * Prevent model event handles from being executed by the given callback.
      *
      * @param  callable  $callback
      * @return callable
      */
-    public function withoutEvents(callable $callback)
+    public function withoutModelEvents(callable $callback)
     {
         return fn () => Model::withoutEvents($callback);
     }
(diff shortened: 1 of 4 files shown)


driesvints (approved):

src/Illuminate/Database/Console/Seeds/WithoutModelEvents.php
```diff
@@ -0,0 +1,19 @@
+<?php
+
+namespace Illuminate\Database\Console\Seeds;
+
+use Illuminate\Database\Eloquent\Model;
+
+trait WithoutModelEvents
+{
+    /**
+     * Prevent model event handles from being executed by the given callback.
```
jbrooksuk, line 10: Could maybe be better written as this?

```suggestion
     * Prevent model events from being dispatched by the given callback.
```

jbrooksuk (approved):

PR #39925, opened 2021-12-07
[8.x] Add explode method to Request

haider00125:
This PR introduces a new helper method to the `Illuminate\Http\Request: explode($key, $delimiter)` which takes input using the input() method and explode value using given delimiter.

```
// Given that
// https://xyz.com?tags=Taylor,Laravel,Framework

// Before
$tags = !empty($input = $request->input('tags')) ? explode(',', $input) : [];

// After
$tags = $request->explode('tags');

// Or you can specify the delimiter
$tags = $request->explode('tags', ':');
```

Unsure how everyone feels about this and the naming of it though.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -304,7 +304,7 @@ trait InteractsWithInput
      * @param  mixed  $default
      * @return mixed
      */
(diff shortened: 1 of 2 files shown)


inxilpro:
You should just be able to do `explode(',', $request->input('tags', ''))` — set the second `$default` parameter to an empty string and then you don't need the ternary.

taylorotwell:

PR #39437, opened 2021-11-01
[8.x] Add an Enum validation rule

themsaid:
This PR adds a new Enum validation rule that will make sure the provided data has a corresponding case in the enum.

```php
validator(
    ['status' => 'pending'],
    ['status' => [new Enum(ServerStatus::class)]]
)->validate();
```

PR #39926, opened 2021-12-07
[8.x] Fail enum validation with pure enums

goodevilgenius:
The current enum validation (#39437) throws an error if a pure enum is used. Since a pure enum doesn't make sense in this context, the validation should simply fail. This is similar to the behavior for if the validation is used on PHP < 8.1.

Consider the following code:

```php
enum PureEnum
{
    case ONE;
    case TWO;
}

validator(
    ['status' => 'ONE'],
    ['status' => [new Enum(PureEnum::class)],
)->validate();
```

This code will throw an error, because the code expects a `tryFrom` static method on the enum. This method only exists on backed enums.

### DIFF
--- a/src/Illuminate/Validation/Rules/Enum.php
+++ b/src/Illuminate/Validation/Rules/Enum.php
@@ -33,7 +33,7 @@ class Enum implements Rule
      */
     public function passes($attribute, $value)
     {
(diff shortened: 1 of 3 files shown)


marcoaureliocardoso:
Enum Validation Rules only works with the BackedEnum Interface
https://laracasts.com/index.php/index.php/discuss/channels/laravel/enum-validation-rules-only-works-by-backedenum-interface

PR #39927, opened 2021-12-07
[8.x] Add type to foreignId()

AmirRezaM75:
I think most developers use ``foreignId()``  method to generate foreign key constraints. But using certain integer types could just be an overkill or under used thus better to have a way to customize it.


taylorotwell:
If you want to use another type, just call that migration method directly.

PR #39928, opened 2021-12-07
[8.x] Remove redundant description & localize template

xanderificnl:
Screen readers will announce the type of role the landmark is, in this case screen readers would say "Pagination Navigation Navigation" -- this PR fixes that. This should be of benefit to accessibility.

See also "Redundant descriptions" @ mdn: developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/form_role

Off-topic: I'd like to thank the maintainers of Laravel. You guys rock!

### DIFF
--- a/src/Illuminate/Pagination/resources/views/simple-tailwind.blade.php
+++ b/src/Illuminate/Pagination/resources/views/simple-tailwind.blade.php
@@ -1,5 +1,5 @@
 @if ($paginator->hasPages())
-    <nav role="navigation" aria-label="Pagination Navigation" class="flex justify-between">
+    <nav role="navigation" aria-label="{{ __('Pagination') }}" class="flex justify-between">
         {{-- Previous Page Link --}}
         @if ($paginator->onFirstPage())
             <span class="relative inline-flex items-center px-4 py-2 text-sm font-medium text-gray-500 bg-white border border-gray-300 cursor-default leading-5 rounded-md">
--- a/src/Illuminate/Pagination/resources/views/tailwind.blade.php
+++ b/src/Illuminate/Pagination/resources/views/tailwind.blade.php
@@ -1,5 +1,5 @@
 @if ($paginator->hasPages())
-    <nav role="navigation" aria-label="{{ __('Pagination Navigation') }}" class="flex items-center justify-between">
+    <nav role="navigation" aria-label="{{ __('Pagination') }}" class="flex items-center justify-between">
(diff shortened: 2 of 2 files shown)


taylorotwell:

aaronhuisinga:
@taylorotwell  @xanderificnl This actually introduces a bug that breaks Tailwind pagination out of the box, and it should ideally be reverted.

When attempting to use Tailwind pagination, I'm seeing the following:
```htmlspecialchars(): Argument #1 ($string) must be of type string, array given (View: /var/www/html/vendor/laravel/framework/src/Illuminate/Pagination/resources/views/tailwind.blade.php)```

The line causing the issue:
```
<nav role="navigation" aria-label="{{ __('Pagination') }}" class="flex justify-between">
```

This is due to Laravel shipping with a `resources/lang/en/pagination.php` file by default. This PR causes the array returned by the pagination language file to be passed into the `__()` method, which results in the parsing exception.

**Comment by xanderificnl** [comment]
I'm using this exact template in my app, and not running into this issue. I doubt this PR is causing your issue.

The line you quoted is from the `simple-tailwind.blade.php` and the error describes `tailwind.blade.php`, are you sure you're looking at the right view?

You can easily test via tinker (`php artisan tinker`):

```
❯ sail art tink
Psy Shell v0.10.12 (PHP 8.1.0 — cli) by Justin Hileman
>>> __('Pagination')
=> "Pagination"
```

You really shouldn't be getting an array conversion when localizing strings.

xanderificnl:
As said, I'm not running into this issue, but I did check the docs and they do agree that `__('Pagination')` isn't a great idea when there's `pagination.php`.

I'll be submitting a PR shortly to resolve this matter by adding a new translation key for pagination.

aaronhuisinga:
@xanderificnl I did indeed paste the change from the wrong file. I was experiencing the issue in `tailwind.blade.php`.

I'm really not sure why it was triggering the error for us and not for your installation. We were just running a pretty standard Laravel app using Sail for testing, and ran into the issue after updating the framework.

Thanks for taking the time to check this out and submitting an updated PR!

xanderificnl:
Hi.

I was just looking through my source and I have a file called resources/lang/en.json. Considering the documentation, that may be the reason why I haven't hit this issue. 

Currently in the train w/o access to a machine besides my phone so I can't actually verify it by trying it out.

Thanks again for bringing this to my attention!

Cheers,

- Xander

PR #34134, opened 2020-09-04
[9.x] Fix scoped implicit route binding query

iamgergo:
This PR provides a fix for scoped implicit route bindings.

Resubmit of https://github.com/laravel/framework/pull/34131

------

Let's say we have **scoped** route binding definition like the following:

```php
Route::get('api/users/{user}/posts/{post:slug}', function (User $user, Post $post) {
    return $post;
});
```

The `resolveChildRouteBinding` method never calls the `resolveRouteBinding`  method on the child model, while it should (in my opinion). This means we can't customize the binding logic for the child model using the `resolveRouteBinding` method.

I was trying different solutions, but for now, it looks the only way is passing the relationship query when resolving the model, and use that as the base query.

------

This is definitely a BC, and probably too late for Laravel 8.x, so I targeted 9.x

Updated tests to use the `resolveRouteBinding` and `resolveChildRouteBinding` methods.

PR #39929, opened 2021-12-08
[8.x] Adds possibility to customize child model route binding resolution

marvin-wtt:
This is an approach to allow the user to define a custom model binding query by overwriting the added function `resolveRouteBindingQuery`. It is needed if lookup of the model cannot be achieved by comparing the value with the field.
The actual problem is briefly explained in #39779

For an for me unknown reason, the test models actually return a model instead of a relation. I considered this even though I don't understand the intention for that.

I am open for other ideas

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1877,7 +1877,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
      */
     public function resolveRouteBinding($value, $field = null)
     {
-        return $this->where($field ?? $this->getRouteKeyName(), $value)->first();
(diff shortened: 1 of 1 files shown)


marvin-wtt:
An example for localized fields where the translations are stored as JSON:
```php
class TestModel extends Model
{
    // ...

    public function resolveRouteBindingQuery($query, $value, $field = null): Builder
    {
        return $query->where("{$field}->{$this->getLocale()}", $value);
    }
}
```

iamgergo:
I had a similar attempt here: https://github.com/laravel/framework/pull/34134, however it was a BC.

An addition like this, would be very handy in my opinion, thanks for the PR.

marvin-wtt:
@taylorotwell Thanks for merging. Should I create a separate PR for 7.x as well?

PR #39931, opened 2021-12-08
[8.x] Fix request dump and dd methods

oliverquynh:
## What do I do?
1. I fix the `dump` docblock because it doesn't only accept array as parameters but also an argument list. These two calls produce the same result.

```php
// it takes an array.
$request->dump(['name', 'age']);

// it takes an argument list.
$request->dump('name', 'age');
```

2. Because `dd` method uses ... oparator, `$keys` in this context will be always an array and the check is unnecessary. `dd` just need to forward the call to `dump` and then to end the script.

```php
class Request
{
    public function dd(...$keys)
    {
        $this->dump(...$keys);

        exit(1);
    }
}
```

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -496,9 +496,7 @@ trait InteractsWithInput
      */
     public function dd(...$keys)
     {
-        $keys = is_array($keys) ? $keys : func_get_args();
-
-        call_user_func_array([$this, 'dump'], $keys);
+        $this->dump(...$keys);
 
         exit(1);
     }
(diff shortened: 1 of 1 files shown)


src/Illuminate/Http/Concerns/InteractsWithInput.php
```diff
@@ -496,17 +496,15 @@ protected function retrieveItem($source, $key, $default)
...
 
         exit(1);
     }
 
     /**
      * Dump the items.
      *
-     * @param  array  $keys
+     * @param  array|mixed  $keys
      * @return $this
```
driesvints, line 508: Mixed can never be used at the same time with another type. Please use mixed here and fix the dd type as well.

seriquynh: I have changed it.

PR #39935, opened 2021-12-08
[8.x] Replace escaped dot with place holder in dependent rules parameters

glamax13:
Hi,
While using the validation tool i stumbled upon an issue with the "required_without" rule.
My use case was somewhat unusual but i needed to check that a field was required if a key was present in an array.
The catch is that the key had a dot in it and i couldn't escape it.

Example:
data: `["country" => ["country_code" => "FR"], "city" => ["iata.alternate_name" => "ORY"]]`
rules: `['country.country_code' => ['required_without:city.iata\.alternate_name', 'size:2'], 
'city.iata\.alternate_name' => ['required_without:country.country_code', 'size:3'],];`

Issue:
The rule wouldn't passe because the dot was interpreted as a nested array.

Solution:
I replaced the dot with the dot placeholder for dependent rules.


taylorotwell:

PR #39936, opened 2021-12-08
[8.x] Add default message to password rule

ricklambrechts:
Currently it is not ease to override all the error messages of the password rule at once.

With this change it is possible to add a defaultMessage that will override other messages.
You can use it by using:
```
Password::min(8)->withDefaultMessage('This is the default message.')
```


taylorotwell:
This is more customizable in the upcoming Laravel 9 release.

PR #39937, opened 2021-12-08
[8.x] Fix php 8.1 deprecation in ValidatesAttributes::checkDateTimeOrder

bastien-phi:
Calling `DateTime::createFromFormat($format, null)` is deprecated with PHP 8.1

Validating 
```php
Validator::make(
    ['start' => '31/12/2012', 'ends' => null],
    ['start' => 'date_format:d/m/Y|before:ends', 'ends' => 'date_format:d/m/Y|after:start']
)
```

triggers this deprecation in `ValidatesAttributes::getDateTimeWithOptionalFormat`.


bastien-phi:
Thanks !

PR #39938, opened 2021-12-08
[8.x] Fixes reporting deprecations when logger is not ready yet

nunomaduro:
This pull request addresses [#286](https://github.com/laravel/sail/issues/286), and [laracasts.com/discuss/channels/laravel/logmanager-not-found](https://laracasts.com/discuss/channels/laravel/logmanager-not-found), by not trying to log deprecations if the deprecation itself comes while composer is trying to load the `LogManager` class.

It's unclear, at this time, where the deprecation comes from, but we suspect that it's from one of the `LogManager` class dependencies at one **specific version** that is allowed by our **composer.json** file.

### DIFF
--- a/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php
+++ b/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php
@@ -86,6 +86,10 @@ class HandleExceptions
      */
     public function handleDeprecation($message, $file, $line)
     {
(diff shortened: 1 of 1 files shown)


Levivb:
Unfortunately the issue isn't really fixed.
It seems to be related to a bug in php's opcache: https://github.com/php/php-src/issues/8164

Will leave a comment there tomorrow

PR #39940, opened 2021-12-08
Introduction of factory property on eloquent models?

rschoonheim:
Hello Artisans,

I use the Eloquent model factories on a daily basis. They really improve my ability to generate fake / mock data while testing. The only downside is I have to stick with the default namespacing or implement `newFactory()` on my Eloquent model. 

```php
/**
* Create a new factory instance for the model.
*
* @return \Illuminate\Database\Eloquent\Factories\Factory
*/
protected static function newFactory()
{
  return $instanceofFactory;
}
``` 

### **Suggested change**

I would love to have a property available where I can define the namespacing of my custom factory. This PR implements such property. What do you people think?

**Example**

```php
<?php

use SomeNamespace\MyCustomUserFactory;

class User extends Eloquent
{
    use HasFactory;

    public static $factory = MyCustomUserFactory::class;
}
```

Feedback please!


taylorotwell:

PR #39944, opened 2021-12-08
[8.0] Making `throw()` callback function argument explicit.

joelharkes:
Made the callback function more explicit, by adding the parameter to the method.

This is probably breaking change for anyone overriding the `Response::throw` method, so you would probably only want to merge this in master?

Other suggestions:
* throw function let it also accept a class name of an exception eg: `->throw(MyExceptionClassWithResponseAsSingleParamterInConstructor::class)`
* toException() can return `null` => which is very weird as in this function it excepts to always return an exception. If user override the Response class and make it return null, this `throw()` method would crash because it would `throw null;` => thus i would suggest it not allow to return null, ever.

I would like to hear feedback on these ideas
I can make these changes if wanted.

### DIFF
--- a/src/Illuminate/Http/Client/Response.php
+++ b/src/Illuminate/Http/Client/Response.php
@@ -267,10 +267,8 @@ class Response implements ArrayAccess
      *
      * @throws \Illuminate\Http\Client\RequestException
      */
-    public function throw()
+    public function throw($callback = null)
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
I believe we did it this way for now in order to make it backwards compatible. However, you could send this change to `master` branch.

PR #39945, opened 2021-12-08
[8.x] Adds datetime parsing to `Request` instance

DarkGhostHunter:
## What?

Allows to conveniently retrieve an input from the Request as a `Carbon` object (or `DateTimeInterface`). It uses the `Date` facade behind the scenes.

```php
public function post(Requets $request)
{
    // ...

    $post->publish_at = $request->date('publish_at')->addHour()->startOfHour();

    $post->save();

    return back();
}
```

## Why?

Because you can quickly get a lot of Carbon or Date calls to parse the date from the request, specially from different keys.

```php
// Before
if ($date = $request->input('when')) {
    $date = Carbon::parse($datetime);
}

// After
$date = $request->date('when');
```

It supports for custom formats (otherwise it will try to parse it).

```php
$timestamp = $request->date('when', 'U');
```

And also supports timezoning.

```php
$timestamp = $request->date('when', 'U', 'America/Santiago');
```

Since the input can be null, it's easier to create defaults:

```php
$now = $request->date('when', 'U') ?? now();
```

# BC?

No, only additive.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -9,6 +9,8 @@ use SplFileInfo;
 use stdClass;
 use Symfony\Component\VarDumper\VarDumper;
 
+use function is_null;
+
 trait InteractsWithInput
 {
     /**
@@ -314,11 +316,12 @@ trait InteractsWithInput
      * @param  string  $key
      * @param  string|null  $format
      * @param  string|null  $tz
+     * @param  \Closure|string|null  $default
      * @return \DateTimeInterface|\Illuminate\Support\Carbon|null
      */
-    public function datetime($key, $format = null, $tz = null)
+    public function datetime($key, $format = null, $tz = null, $default = null)
     {
-        $date = $this->input($key);
+        $date = $this->input($key) ?? value($default, $this);
 
         if (is_null($date)) {
             return null;
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -314,19 +314,22 @@ trait InteractsWithInput
      * @param  string  $key
      * @param  string|null  $format
      * @param  string|null  $tz
-     * @param  \Closure|string|null  $default
-     * @return \DateTimeInterface|\Illuminate\Support\Carbon|null
+     * @return \Illuminate\Support\Carbon|bool|null
      */
-    public function datetime($key, $format = null, $tz = null, $default = null)
+    public function datetime($key, $format = null, $tz = null)
     {
-        $date = $this->input($key) ?? value($default, $this);
+        $date = $this->input($key);
 
         if (is_null($date)) {
             return null;
         }
 
         if ($format) {
-            return Date::createFromFormat($format, $date, $tz);
+            try {
+                return Date::createFromFormat($format, $date, $tz);
+            } catch (InvalidArgumentException $e) {
+                return false;
+            }
         }
 
         return Date::parse($date, $tz);
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -310,7 +310,7 @@ trait InteractsWithInput
 
     /**
      * Retrieve input from the request as a Carbon instance.
-     * 
+     *
      * @param  string  $key
      * @param  string|null  $format
      * @param  string|null  $tz
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -298,17 +298,6 @@ trait InteractsWithInput
     }
 
     /**
-     * Retrieve input from the request as a collection.
-     *
-     * @param  array|string|null  $key
-     * @return \Illuminate\Support\Collection
-     */
-    public function collect($key = null)
-    {
-        return collect(is_array($key) ? $this->only($key) : $this->input($key));
-    }
-
-    /**
      * Retrieve input from the request as a Carbon instance.
      *
      * @param  string  $key
@@ -327,6 +316,17 @@ trait InteractsWithInput
         }
 
         return Date::createFromFormat($format, $this->input($key), $tz);
+    }
+
+    /**
+     * Retrieve input from the request as a collection.
+     *
+     * @param  array|string|null  $key
+     * @return \Illuminate\Support\Collection
+     */
+    public function collect($key = null)
+    {
+        return collect(is_array($key) ? $this->only($key) : $this->input($key));
     }
 
     /**
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -4,6 +4,7 @@ namespace Illuminate\Http\Concerns;
 
 use Illuminate\Http\UploadedFile;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Facades\Date;
 use SplFileInfo;
 use stdClass;
 use Symfony\Component\VarDumper\VarDumper;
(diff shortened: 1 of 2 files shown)


src/Illuminate/Http/Concerns/InteractsWithInput.php
```diff
@@ -307,6 +308,30 @@ public function collect($key = null)
...
+    public function datetime($key, $format = null, $tz = null, $default = null)
+    {
+        $date = $this->input($key) ?? value($default, $this);
+
+        if (is_null($date)) {
+            return null;
+        }
+
+        if ($format) {
+            return Date::createFromFormat($format, $date, $tz);
```
devfrey, line 329: `createFromFormat()` may return `false`, which is not accounted for. Here's an example of how this is handled elsewhere in the framework:

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1141-L1147

Also note that `createFromFormat()` may throw `InvalidFormatException` (hence the `InvalidArgumentException` catch in the implementation above).

src/Illuminate/Http/Concerns/InteractsWithInput.php
```diff
@@ -307,6 +308,30 @@ public function collect($key = null)
...
     }
 
+    /**
+     * Retrieve input from the request as a Carbon instance.
+     *
+     * @param  string  $key
+     * @param  string|null  $format
+     * @param  string|null  $tz
+     * @param  \Closure|string|null  $default
+     * @return \DateTimeInterface|\Illuminate\Support\Carbon|null
```
devfrey, line 318: `DateTimeInterface` is redundant because it's implemented by `Illuminate\Support\Carbon`.

Because `$default` (or its return value if passed a closure) can be mixed, this method should return `mixed` as well. Since this would not be very helpful type-wise, I'd suggest removing the `$default` altogether. The same fallback behaviour can be achieved with a null coalesce operator after `$request->datetime('...')`.

devfrey:
This example is not very accurate:
```php
if ($datetime = $request->input('when')) {
    $datetime = Carbon::parse($datetime);
}
```

It can already be written like this:
```php
$datetime = Carbon::make($request->input('when'));
```
... which does not seem very bad.

src/Illuminate/Http/Concerns/InteractsWithInput.php
DarkGhostHunter: That's assuming the developer doesn't validates an input from being a valid datetime. I think it should return `false` to keep it consistent with how it's handled on models.

src/Illuminate/Http/Concerns/InteractsWithInput.php
DarkGhostHunter: Yep, makes sense.

DarkGhostHunter:
> This example is not very accurate:
> 
> ```
> if ($datetime = $request->input('when')) {
>     $datetime = Carbon::parse($datetime);
> }
> ```
> 
> It can already be written like this:
> 
> ```
> $datetime = Carbon::make($request->input('when'));
> ```
> 
> ... which does not seem too bad to me.

Same case for `collect()`. Also, you're accounting the input is not `null`.

devfrey:
> Same case for `collect()`.

I'm not sure what you mean by this?

> Also, you're accounting the input is not null.

That's not true. `Carbon::make(null)` returns `null`. It's different from `Carbon::parse()`.

src/Illuminate/Http/Concerns/InteractsWithInput.php
```diff
@@ -307,6 +308,33 @@ public function collect($key = null)
...
+
+        if (is_null($date)) {
+            return null;
+        }
+
+        if ($format) {
+            try {
+                return Date::createFromFormat($format, $date, $tz);
+            } catch (InvalidArgumentException $e) {
+                return false;
```
devfrey, line 331: This should be `null`. This convenience method will only work if it returns either `\Illuminate\Support\Carbon` or `null`. Anything else will require additional type checks on the consumer side to ensure types are compatible (eg. when setting a model attribute to the return value of `$request->datetime()`).

DarkGhostHunter:
Just reworked it to return `null` if the datetime is invalid.

crnkovic:
I like the idea as I needed this many times, but you should probably take care of few stuff.

First, not sure why the use of Date facade, when DateTime and/or Carbon classes are sufficient for this, IMHO. Also, Carbon throws an `InvalidFormatException`, which I think is more expressive than just catching all invalid argument exceptions. Likewise, `createFromFormat` will [return `false` on failure](https://www.php.net/manual/en/datetime.createfromformat.php#refsect1-datetime.createfromformat-returnvalues), so this should also be taken care of as the typehint of your `date()` method is Carbon or null.

DarkGhostHunter:
> First, not sure why the use of Date facade when DateTime and/or Carbon classes are sufficient for this, IMHO

The developer may be using another library. Otherwise, the `Date` facade wouldn't exists. I'm just using what is available. Even `$model->freshTimestamp()` uses the `Date` facade behind the scenes.

> Also, Carbon throws an `InvalidFormatException`, which I think is more expressive than just catching all invalid argument exceptions.

Same case as above. `InvalidFormatException` is exclusive to the Carbon library, as far as I know. Can't catch what doesn't exists, if it doesn't.

> Likewise, `createFromFormat` will [return `false` on failure](https://www.php.net/manual/en/datetime.createfromformat.php#refsect1-datetime.createfromformat-returnvalues), so this should also be taken care of as the typehint of your `date()` method is Carbon or null.

I would back to my original idea: if the format or the value to be formatted is invalid, throw an exception. It will return `null` only if the input is `null` or doesn't exists.

crnkovic:
> I would back to my original idea: if the format or the value to be formatted is invalid, throw an exception. It will return `null` only if the input is `null` or doesn't exists.

Yeah okay makes sense, if it can't be parsed by Carbon then should throw an exception just like Carbon would.

taylorotwell:
> I would back to my original idea: if the format or the value to be formatted is invalid, throw an exception. It will return null only if the input is null or doesn't exists.

@DarkGhostHunter  I agree that this should be the behavior.

DarkGhostHunter:
It's ready.

DarkGhostHunter:
It's ready for review.

DarkGhostHunter:
👍

src/Illuminate/Http/Concerns/InteractsWithInput.php
```diff
@@ -296,6 +297,27 @@ public function boolean($key = null, $default = false)
...
 
+    /**
+     * Retrieve input from the request as a Carbon instance.
+     *
+     * @param  string  $key
+     * @param  string|null  $format
+     * @param  string|null  $tz
+     * @return \Illuminate\Support\Carbon|null
+     */
+    public function date($key, $format = null, $tz = null)
```
amitmerchant1990, line 308: Why not `$timezone`?

iKlsR:
I like the datetime parsing but I wish the function handled defaults better when the input isn't null.

```php
// Old
$startDate = Carbon::parse(Request::input('from', Carbon::now()->subYears(5)))->format('Ymd');
$endDate = Carbon::parse(Request::input('to', $lastTradeDate))->format('Ymd');

// New
$startDate = Request::date('from', 'Ymd') ?? Carbon::now()->subYears(5)->format('Ymd');
$endDate = Request::date('to', 'Ymd') ?? Carbon::parse($lastTradeDate)->format('Ymd');

// Expected?
$startDate = Request::date('from', ['Ymd'], Carbon::now()->subYears(5));
$endDate = Request::date('to', ['Ymd', 'America/Santiago'], $lastTradeDate);
$otherDate = Request::date('other', ['Ymd', 'America/Santiago']);
```

iKlsR:
I like this but I wish the function handled defaults better sans coalescing when the input isn't null so one could benefit more from the parsing upfront and I think the timezone is not needed in most cases so it could be optional in an array and use the param space for the default.

```php
// Old
$startDate = Carbon::parse(Request::input('from', Carbon::now()->subYears(5)))->format('Ymd');
$endDate = Carbon::parse(Request::input('to', $lastTradeDate))->format('Ymd');

// New
$startDate = Request::date('from', 'Ymd') ?? Carbon::now()->subYears(5)->format('Ymd');
$endDate = Request::date('to', 'Ymd') ?? Carbon::parse($lastTradeDate)->format('Ymd'); 

// Expected?
$startDate = Request::date('from', ['Ymd'], Carbon::now()->subYears(5));
$endDate = Request::date('to', ['Ymd', 'America/Santiago'], $lastTradeDate);
$otherDate = Request::date('other', ['Ymd', 'America/Santiago']);
$anotherDate = Request::date('another', 'Ymd');
```

PR #39946, opened 2021-12-08
Validation Rules: remove duplicate table name resolving

devpreview:
I don't understand why need table name resolving on DatabaseRule.

Rules in _Illuminate\Validation\Rules_ converted to a validation string.
Database validation rules (e.g. _exists_, _unique_) accepted DB table name or model class, this is implemented in _ValidatesAttributes::parseTable_ method.
https://github.com/laravel/framework/blob/8.x/src/Illuminate/Validation/Concerns/ValidatesAttributes.php#L887

May need to remove table name resolving duplication?

### DIFF
--- a/src/Illuminate/Validation/Rules/DatabaseRule.php
+++ b/src/Illuminate/Validation/Rules/DatabaseRule.php
@@ -4,7 +4,6 @@ namespace Illuminate\Validation\Rules;
 
 use Closure;
 use Illuminate\Database\Eloquent\Model;
-use Illuminate\Support\Str;
 
 trait DatabaseRule
 {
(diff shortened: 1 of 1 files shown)


devpreview:
Another implementation:

```
namespace Illuminate\Validation\Rules;

use Illuminate\Support\Traits\Conditionable;

class Exists
{
    use Conditionable, DatabaseRule;

    /**
     * Create a new rule instance.
     *
     * @param  string  $table
     * @param  string  $column
     * @return void
     */
    public function __construct($table, $column = 'NULL')
    {
        parent::__construct($table, $column);
        $this->table = $table;
    }

    ...
}
```

Similar for _Illuminate\Validation\Rules\Unique_.

taylorotwell:
I don't want to remove code if we don't understand it.

PR #39950, opened 2021-12-08
[8.x] Add cache-busting for public assets

mhmdomer:
This PR allows the use of cache-busting on public assets,
Using:
```php
asset('filePath', version: true)
```
generates a URL for the asset and appends an id with this format:
`https://your-domain.com/filePath?id=1639002822`
the value of the id is a UNIX timestamp of the last time the file was modified.
So using this feature we can make sure the browser always requests the new file whenever the file is changed, even if the file was cached on the browser. also, it eliminates the annoying need for updating file names when they change to make sure clients get the newest file every time the file is changed.

This PR also doesn't break the normal usage of the `asset` helper.
So the function:
```php
asset('filePath')
```
generates `https://your-domain.com/filePath` as expected

These changes don't introduce any security vulnerability as the files are already public and the only exposed info is the last modified time of the file.

### DIFF
--- a/src/Illuminate/Routing/UrlGenerator.php
+++ b/src/Illuminate/Routing/UrlGenerator.php
@@ -255,10 +255,11 @@ class UrlGenerator implements UrlGeneratorContract
     protected function assetWithVersion($path, $url)
     {
         $fullPath = public_path($path);
-        if (!file_exists($fullPath)) {
+        if (! file_exists($fullPath)) {
             return $url;
         }
-        return $url . '?id=' . filemtime($fullPath);
+
+        return $url.'?id='.filemtime($fullPath);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
I don't think this will play nice across multi-server deployments. It is more robust to use ETAG caching (default in cloudfront) or to version the URIs at compile time (mix supports this, easily), instead of using a query string param.

src/Illuminate/Routing/UrlGenerator.php
```diff
@@ -230,7 +230,7 @@ public function secure($path, $parameters = [])
      * @param  bool|null  $secure
      * @return string
      */
-    public function asset($path, $secure = null)
+    public function asset($path, $secure = null, $version = false)
```
driesvints, line 233: This is a breaking change to the method signature.

mhmdomer:
@GrahamCampbell @mhmdomer thanks for taking the time to review this.
I think I will just add a new helper function for this feature to my collection of helpers 😅

taylorotwell:

PR #39958, opened 2021-12-09
[8.x] `withTrashed` on routes check if `SoftDeletes` is used in `Model`

Slubby:
When using `withTrashed()` on a route with two bind models and one model doesn't have `SoftDeletes` its throws a `BadMethodCallException`

For example, you have a route like this:

`
Route::put('admin/articles/(article)/comments/{comment}/restore')->withTrashed()
`

The article model doesn't have `SoftDeletes` but the comment model has. When you want to restore the comment now its throws an exception because the model article doesn't have `SoftDeletes`


Quick00:
Exactly what i need! 🌮

PR #39965, opened 2021-12-09
[9.0] Making `throw()` callback function argument explicit.

joelharkes:
Breaking changes in `Response` class.

* Cleanup implicit argument in method `toException()` and more helpfull exceptions if fails
* Make `toException` always return an exception, seems much more logical otherwise i would call method:`toExceptionIfFailed()` 

## Why

To me this feature was not clear because not all IDE's show you the option for this parameter. ie: PHPstorm:
image


taylorotwell:
Hmm, I'm not sure I wanted all the behavior changes. The callback doesn't receive the exception anymore either. If you want to make just the signature change that is fine.

Issue #35798, opened 2021-01-06
BelongsToMany#updateOrCreate is not merging arguments before create


marcoboers:
- Laravel Version: 8.20.1
- PHP Version: 7.4.8
- Database Driver & Version: MariaDB 10.5.8

### Description:
For `Illuminate\Database\Eloquent\Relations\BelongsToMany` relationships, the `updateOrCreate` method is not merging the $attributes and $values arrays. 

This is different from the behaviour of the `updateOrCreate` methods on `Builder`,   `HasOneOrMany` and `HasManyThrough` classes and from the documented behaviour "If no such flight exists, a new flight will be created which has the attributes resulting from merging the first argument array with the second argument array" [eloquent#upserts](https://laravel.com/docs/8.x/eloquent#upserts). 

**Illuminate\Database\Eloquent\Relations\BelongsToMany**
```
    public function updateOrCreate(array $attributes, array $values = [], array $joining = [], $touch = true)
    {
        if (is_null($instance = $this->where($attributes)->first())) {
            return $this->create($values, $joining, $touch);
        }

        $instance->fill($values);

        $instance->save(['touch' => false]);

        return $instance;
    }
```
**Proposed solution**
This solution is based on master (9.x) branch where the fix for [#34083](https://github.com/laravel/framework/issues/34083) is taken into account.
```
    public function updateOrCreate(array $attributes, array $values = [], array $joining = [], $touch = true)
    {
        if (is_null($instance = $this->related->where($attributes)->first())) {
            $instance = $this->create($attributes, $joining, $touch);
        }

        $instance->fill($values);

        $instance->save(['touch' => false]);

        return $instance;
    }
```
### Steps To Reproduce:
```
Schema::create('roles', function (Blueprint $table) {
    $table->increments('id');
    $table->unsignedBigInteger('created_by_id');
    $table->string('name'); // <-- required
    $table->timestamps();
});
```
```
$user = User::factory()->create();
Role::create(['name' => 'existing.role', 'description' => 'foo']);
$user->roles()->updateOrCreate(['name' => 'existing.role'], ['description' => 'bar']);

parent::assertEquals(1, Role::count()); // assertion fails (or is never reached); count equals 2
```
```
$user = User::factory()->create();

// throws: Illuminate\Database\QueryException: SQLSTATE[HY000]: General error: 1364 Field 'name' doesn't have a default value
$user->roles()->updateOrCreate(['name' => 'existing.role'], ['description' => 'bar']); 

[...]

Issue #34083, opened 2020-09-01
BelongsToMany#updateOrCreate is checking the wrong table for Model existence


Harti:
- Laravel Version: 7.25.0
- PHP Version: 7.4.7
- Database Driver & Version: MySQL 8.0.20

### Description:

For `Illuminate\Database\Eloquent\Relations\BelongsToMany` relations only, the `updateOrCreate()` method is checking _the pivot table_ instead of the _related_ table for Model existence:

##### Illuminate\Database\Eloquent\Relations\BelongsToMany
```php
    public function updateOrCreate(array $attributes, array $values = [], array $joining = [], $touch = true)
    {
        if (is_null($instance = $this->where($attributes)->first())) {
            return $this->create($values, $joining, $touch);
        }

        $instance->fill($values);

        $instance->save(['touch' => false]);

        return $instance;
    }
```

It appears this was an oversight because the `BelongsToMany#create` method however would then proceed to create an entity in `$this->related`. 

As a side note, the `HasOneOrMany#updateOrCreate` method is vastly different with its `tap($this->firstOrNew(...), ...)` call.

##### Proposed fix (change first line in method body):
```php
    public function updateOrCreate(array $attributes, array $values = [], array $joining = [], $touch = true)
    {
        if (is_null($instance = $this->related->where($attributes)->first())) {
            return $this->create($values, $joining, $touch);
        }

        $instance->fill($values);

        $instance->save(['touch' => false]);

        return $instance;
    }
```

### Steps To Reproduce:
Example Gist with migrations, test and stacktrace (friend User-User relation): https://gist.github.com/Harti/2afb5afd524f3639045b3d57756655fd

Inline Example:
```php
Role::create(['name' => 'existing.role', 'description' => 'foo']);
$user->roles()->updateOrCreate(['name' => 'existing.role'], ['description' => 'bar']);

parent::assertEquals(1, Role::count()); // assertion fails (or is never reached); count equals 2
``` 

A second/duplicate entry is inserted into the `roles` table (or a Integrity Constraint Violation is thrown, should `name` be unique).
Expected behavior: The existing Role is updated, and saved to the `$user->roles()` relation (which also only seems to happen if the instance was newly created).

Best regards,

Harti

driesvints:
Can you please post your models/pivots?

Harti:
@driesvints 
This was entirely made up due to NDA, so I created another example with as little boilerplate as possible in a fresh Laravel installation:
[...]

PR #39970, opened 2021-12-10
[8.x] Fixed BelongsToMany UpdateOrCreate Create operation

matiaslauriti:
I was working on a project and I was trying to do `$model->relation()->updateOrCreate($attributesToLookFor, $attributesToUpdateOrCreateWith)` and the issue is that the `create` flow is not adding the first attributes.

Related issue https://github.com/laravel/framework/issues/35798 and https://github.com/laravel/framework/issues/34083 (this last one is not the same issue, but is related to the same part of the code).

---

I just saw there is a fix for this only for Laravel 9.x on https://github.com/laravel/framework/pull/35827, I am working on 8.x and I need this fix as soon as possible...


taylorotwell:
I can't make this change on a patch release. Sorry.

PR #39972, opened 2021-12-10
[8.x] Added reason phrase getter to Response class

markvesterskov:
I am interfacing with a very legacy XML API in my application, which sends back error hints in the reasonPhrase field of a response - not super kosher, but it works, i guess.

This simply adds a getter to Illuminate\Http\Client\Response to get the reason phrase from the underlying PSR response.


taylorotwell:

PR #39973, opened 2021-12-10
[8.x] Prevent further lookups of user

jaulz:
In case there is still a user id in the session and the user was deleted, the framework tries to retrieve the user every time the `user` method is called. This PR sets a flag to prevent further lookups.

### DIFF
--- a/src/Illuminate/Auth/SessionGuard.php
+++ b/src/Illuminate/Auth/SessionGuard.php
@@ -147,7 +147,7 @@ class SessionGuard implements StatefulGuard, SupportsBasicAuth
             return $this->user;
         }
 
-        // Indicate the we already attempted to retrieve the user 
+        // Indicate the we already attempted to retrieve the user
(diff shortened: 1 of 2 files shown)


jaulz:
Hm, for some reason the tests fail and I face similar issues in my local environment as well but it seems that all the issues are unrelated to my changes. My new specific test ran through without issues as far as I could see.

taylorotwell:
Hmm, I'm not sure I want to change this behavior on a patch release. Fear it could have some unintended breakages for edge cases.

jaulz:
@taylorotwell thanks for your feedback. Shall I raise the PR for a different branch then? Right now it has the possibility to significantly decrease the performance of the app because the `user()` function can easily be triggered a few dozens times and each call causes a new database request. I admit that the likelihood of this event is minimal and requires the user to be deleted after the user logged in but the same argument would speak for having it as part of a patch release.

PR #38629, opened 2021-09-01
[8.x][fix bug] when search from cross database relationship ,it comes to "Property [from] does not exist on the Eloquent builder instance."

Sparkinzy:
fix bug : Property [from] does not exist on the Eloquent builder instance.

when I search from cross database relationship ,it comes to this error.

```php
OauthRecord::whereHas('user', function (Builder $query) {
    $query->where('username', 'like', 'sparkinzy%');
})
```

> Error:  Property [from] does not exist on the Eloquent builder instance.
> Error:  Property [from] does not exist on the Eloquent builder instance.
> Error:  Property [from] does not exist on the Eloquent builder instance.

Model A:
```php
class OauthRecord extends Model
{

    protected $table = 'oauth_record';
    protected $connection = 'mysql-B'

    public function user()
    {
        $instance = new User();
        $instance->setTable('db_a.users');
        $query = $instance->newQuery();
        return new BelongsTo($query, $this, 'user_id', 'id',null);
    }

}
```

Model B 
```php
class User extends Model
{

    protected $table = 'users';
    protected $connection = 'mysql-A'

}
```

config/database.php
```php
return [
    'connections' => [
           'mysql-A' => [
                'driver' => 'mysql',
                'host' => '127.0.0.1',
                'port' => '3306',
                'database'=> 'db_a',

            ],
           'mysql-B' => [
                'driver' => 'mysql',
                'host' => '127.0.0.1',
                'port' => '3306',
                'database'=> 'db_b',

            ],

    ]
];
```

Issue #35631, opened 2020-12-16
joinSub -> Property [from] does not exist on the Eloquent builder instance.


OliverZiegler:
- Laravel Version: 8.19.0
- PHP Version: 7.14.10
- Database Driver & Version: mysql 5.7.31

### Description:

when calling `joinSub` with a Model query (instance of `Eloquent\Builder`) we run in an error:
`Property [from] does not exist on the Eloquent builder instance.`

Seems like I ran into the same issue like described in #34249 and could trace it down to the following.

`Query\Builder->joinSub` calls `Query\Builder->createSub` where  `Query\Builder->parseSub` gets called.

```php
protected function parseSub($query)
    {
        if ($query instanceof self || $query instanceof EloquentBuilder || $query instanceof Relation) {
            $query = $this->prependDatabaseNameIfCrossDatabaseQuery($query);

            return [$query->toSql(), $query->getBindings()];
        } elseif (is_string($query)) {
            return [$query, []];
        } else {
            throw new InvalidArgumentException(
                'A subquery must be a query builder instance, a Closure, or a string.'
            );
        }
    }
```

Here it also checks for an `EloquentBuilder` and passes it down to `prependDatabaseNameIfCrossDatabaseQuery`.

```php
protected function prependDatabaseNameIfCrossDatabaseQuery($query)
    {
        if ($query->getConnection()->getDatabaseName() !==
            $this->getConnection()->getDatabaseName()) {
            $databaseName = $query->getConnection()->getDatabaseName();

            if (strpos($query->from, $databaseName) !== 0 && strpos($query->from, '.') === false) {
                $query->from($databaseName.'.'.$query->from);
            }
        }

        return $query;
    }
```

But if `$query` here is not an `Query\Builder` but an `Eloquent\Builder` we cannot access `$query->from`.
My IDE tells me it is available (as `Query\Builder` is `@mixin \Illuminate\Database\Query\Builder` in the PHP Doc comments of `Eloquent\Builder`, but I cannot get how PHP should get access to the mixed in properties.
The `ForwardCalls` Trait makes only functions available, but not properties and in the `__get` only `orWhere` is a magic property.

As for now I could workaround by calling `$query->toBase()` when using it in `joinSub` but the main issue is still present in the laravel core I think.

taylorotwell:
Can you call the `from` method? Please share the entire query you are trying to create?

sauladam:
[...]

PR #39981, opened 2021-12-10
[8.x] Add getSourceTable method to Builder classes

glmdev:
This PR makes the `Illuminate\Database\Query\Builder->from` property available to read from the `Illuminate\Database\Eloquent\Builder` instance.

This is done by adding a pass-through in the `__get` method of `Eloquent\Builder` that just returns the value on the base query builder.

### The Current Issue

Currently, the join/subquery methods on the Eloquent builder are documented as accepting _either_ a base query builder, or an Eloquent query builder.

However, if you attempt to do a cross-database join/subquery, the framework eventually calls `prependDatabaseNameIfCrossDatabaseQuery` on the `Query\Builder` instance. This method checks if the query is cross-database. If so, it attempts to access the `->from` property of the query.

When an Eloquent builder instance is passed along, it will error out at that point, as there is no `from` property accessible on the `Eloquent\Builder` instance.

### Rationale

The Eloquent builder currently throws an error when trying to access the `from` property, so this doesn't modify an existing accessor.

Because the PHP doc comments say you should be able to pass an Eloquent builder into the subquery/join methods, the fact that they can fail in some cases specifically with an Eloquent builder is a bug.

Adding this `from` property fixes the bug by making the call to `prependDatabaseNameIfCrossDatabaseQuery` succeed for either type of `Builder`.

This change is also fairly minor, and I've included an update to the class doc for `Eloquent\Builder` and a test for said property access.

### Related Issues

This fixes the issue described in #35631 and is an improved solution on #38629.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -21,6 +21,7 @@ use ReflectionMethod;
 
 /**
  * @property-read HigherOrderBuilderProxy $orWhere
+ * @property-read string from
  *
  * @mixin \Illuminate\Database\Query\Builder
  */
@@ -1598,6 +1599,10 @@ class Builder
     {
         if ($key === 'orWhere') {
             return new HigherOrderBuilderProxy($this, $key);
+        }
+
+        if ($key === 'from') {
+            return $this->query->from;
         }
 
         throw new Exception("Property [{$key}] does not exist on the Eloquent builder instance.");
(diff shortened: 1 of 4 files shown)


inxilpro:
Would it be better to add a method (that can be added to the Query Builder interface that's in 9.x) and then call the method? That feels a little better to me than using a condition in `__get()`…

glmdev:
I can modify the solution to use a method if you think that's best. I was just trying to find the solution with the smallest impact in the hopes that this could be fixed in 8.x.

taylorotwell:
Yeah I would maybe add a method or something to access that so we don't have to have this string checking in __get.

glmdev:
Okay, I've added a getter method to both of the Builder classes and removed the `from` pass-through.

taylorotwell:
I actually did end up solving this by adding a `propertyPassthru` array similar to the existing `passthru` array instead of having to invent a new method - sorry about that.

PR #39985, opened 2021-12-11
[8.x] Add open with IDE to all the make commands

d8vjork:
Hi everyone! 👋 

So this PR only adds a very simple feature, I always dreamed about quickly open the files I was creating with the built-in make commands.

First I thought on this as an `--open` option, but I don't know how you guys would think of it. Although if you don't have any IDE configured it won't do anything (thanks to that horrible switch, I would prefer to use match but it will break retrocompatibility with PHP 7.x versions).

Also need to add this config block on the `app.php` (I don't know a better place):

```php
/*
|--------------------------------------------------------------------------
| Application IDE
|--------------------------------------------------------------------------
|
| When your application is writing some files using make commands
| this will be used for open the files on the editor you use for
| write your application's code.
|
| Supported IDEs: "sublime", "textmate", "emacs", "macvim", "phpstorm", "idea",
|        "vscode", "vscode-insiders", "vscode-remote", "vscode-insiders-remote",
|        "atom", "nova", "netbeans"
|
*/

'ide' => env('APP_IDE'),
```

Let me know what you think

### DIFF
--- a/src/Illuminate/Console/GeneratorCommand.php
+++ b/src/Illuminate/Console/GeneratorCommand.php
@@ -168,6 +168,8 @@ abstract class GeneratorCommand extends Command
         if (in_array(CreatesMatchingTest::class, class_uses_recursive($this))) {
             $this->handleTestCreation($path);
         }
+
+        $this->openWithIde($path);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


d8vjork:
Also I don't know how this should be tested, I already tested locally on Mac and Windows (tho, [Windows seems limited on PHPStorm at least](https://youtrack.jetbrains.com/issue/IDEA-65879)), but I copied all the URLs from the awesome work done by @spatie on Ignition.

Linux needs tests with that xdg-open (I just saw it online as I'm a Mac user)

igorgaming:
But different developers use different IDEs. As a rule, there is no clear installed ide for the project. Therefore, this makes this feature useless, especially if you are not working on the project alone.
It also seems strange to me to add a used IDE to the application config, it definitely shouldn't be there.

As for the `--open` option itself, it would be cool to have it, but not in this way, imho.

d8vjork:
@igorgaming That's why is a env variable, the IDE totally depends on the user

igorgaming:
> @igorgaming That's why is a env variable, the IDE totally depends on the user

oh, yeah, i see

antonkomarev:
There is issues in Linux and maybe Windows, apps are not opening with `appname://` scheme.

d8vjork:
@antonkomarev And what are those issues?

Got it working on my Ubuntu 20.04:

image

antonkomarev:
Have you tried phpstorm? Especially when other intellij products installed?

d8vjork:
@antonkomarev Yes, look at the first comment

Thought, there's still one option for IntelliJ products (at least for PHPStorm), they've a CLI for all this kind of things, but will extend the config/env further.

taylorotwell:

d8vjork:
Ok... but this is impossible to port to a external package as the framework is not providing the output path, neither the exact command executed, this will be near impossible unless we modify `Illuminate\Console\Events\CommandFinished`

PR #39988, opened 2021-12-11
[8.x] Add `Str::substrReplace()` and `Str::of($string)->substrReplace()` methods for the PHP native `substr_replace()` function

ralphjsmit:
This pull request adds the ability to use the `substr_replace()` on the `Str` and `Stringable` classes. 

## What & why

Developers can now use directly use `Str::substrReplace(...)` and `Str::of($string)->substrReplace(...)`.

Substring replace can be used to insert a string at a certain position in another string (set the last parameter, the length, to `0`).
```php
$string = '1300';

$result = Str::substrReplace($string, ':', 2, 0); 
// '13:00'

$result = (string) Str::of($string)->substrReplace(':', 2, 0);
// '13:00'
```

It can also be used to replace the remainder of a string after a certain position:

```php
$result = (string) Str::of('Laravel Framework')->substrReplace('– The PHP Framework for Web Artisans', 8);
// 'Laravel – The PHP Framework for Web Artisans'
```
The benefit for webdevelopers is that they now have access to the `substr_replace()` method through the Str facade. The biggest benefit is (in my opinion), is that it's also a perfect candidate to be part of the fluent Stringable methods.

I'm also willing to PR a docs update for this, but I don't know if maintainers usually take care of updating the docs or if that is a task for contributors.

This is my first PR to Laravel, so please let me know if I missed anything or if there's anything I can improve. Thanks!

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -876,6 +876,20 @@ class Str
     }
 
     /**
+     * Replace text within a portion of a string
+     *
+     * @param  string|array  $string
+     * @param  string|array  $replace
+     * @param  array|int  $offset
+     * @param  array|int|null  $length
+     * @return string|array
+     */
+    public static function substrReplace($string, $replace, $offset = 0, $length = null)
+    {
+        return substr_replace($string, $replace, $offset, $length);
+    }
+
+    /**
      * Make a string's first character uppercase.
      *
      * @param  string  $string
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -886,6 +886,10 @@ class Str
      */
     public static function substrReplace($string, $replace, $offset = 0, $length = null)
     {
+        if ($length ===  null) {
+            $length = strlen($string);
+        }
+        
         return substr_replace($string, $replace, $offset, $length);
     }
 
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -886,10 +886,10 @@ class Str
      */
     public static function substrReplace($string, $replace, $offset = 0, $length = null)
     {
-        if ($length ===  null) {
+        if ($length === null) {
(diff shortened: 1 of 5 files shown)


src/Illuminate/Support/Str.php
```diff
@@ -875,6 +875,24 @@ public static function substrCount($haystack, $needle, $offset = 0, $length = nu
...
+     * @param  string|array  $replace
+     * @param  array|int  $offset
+     * @param  array|int|null  $length
+     * @return string|array
+     */
+    public static function substrReplace($string, $replace, $offset = 0, $length = null)
+    {
+        if ($length === null) {
+            $length = strlen($string);
+        }
```
derekmd, line 891: It looks like this conditional block isn't required since `substr_replace()` accepts a `$length` parameter value of `null` to replace the remaining substring:

substr_replace


```php
substr_replace(
    'Laravel Framework',
    '– The PHP Framework for Web Artisans',
    8
);
// "Laravel – The PHP Framework for Web Artisans"
```

ralphjsmit: Hey Derek, that's true, but only since PHP 8.0 is the `$length` parameter nullable. 

In PHP 7.3 and 7.4 this parameter would automatically take a value of `0` instead of `null`, causing an insert instead of a replacement of the rest of the string. So I added this conditional to keep behaviour consistent. 

The test runs pointed this out as well. 

In a newer version of Laravel (9?) we could drop the conditional. So it's perhaps an idea to add a commit to the `9.x` branch as well that removes the if-condition.

NewEXE:
It's NOT multibytes-ready.

```php
$right = 'Ларавел – PHP-фреймворк';
$res = \Illuminate\Support\Str::substrReplace('Ларавел фреймворк', '– PHP-фреймворк', 8);
dump($right, $res);

// ^ "Ларавел – PHP-фреймворк"
// ^ "Лара– PHP-фреймворк"
```

PR #39991, opened 2021-12-11
[8.x] Fixes model:prune --pretend command for models with SoftDeletes

wgmv:
This PR should fix the model:prune --pretend command when used in combination with soft deleted models.
The withTrashed method is never called because the SoftDeletes class is not found in the array. Therefore the --pretend option always returns 0 when using a condition in the model like

```php
 public function prunable()
    {
        return static::where('deleted_at', '<=', now()->subDays(30));
    }
```

the query will currently be appended with "where deleted_at is null"


derekmd:
A test case should probably be added to [tests/Database/PruneCommandTest.php ](https://github.com/laravel/framework/blob/8.x/tests/Database/PruneCommandTest.php)to ensure this works as intended going forward?

PR #39992, opened 2021-12-12
[8.X] PATCH eloquent builder method find() performance improvement.

navaneeth-raman-bhaskar:
### Included in this PR

**1.** Changes to the Eloquent Builder method `find()` . 
 Previously method would have run a database query if  `$id` parameter was `null`. 
 Now it returns early. Saves you a query 😊
 
 _**Example:**_

 Previously
 
`$user = User::find(null);`

> $user will be **null**.

Database query

> SELECT * FROM users WHERE users.id IS NULL and users.deleted_at IS NULL limit  1

After the patch 

`$user = User::find(null);`

> $user will be **null**.

Database query

> (no query at all)

**2.** Added a new `findOrCreate()` method to Eloquent Builder class.
Method is used to find a model by its primary key. If it is not found then creates a new record.

_**Example:1**_

 Lets us assume id 27 is in database and 28 is not

`$user = User::findOrcreate(27, ['name'=>'navaneeth', 'age'=>27]);`

> $user will have the record from database

Database query

> SELECT * FROM users WHERE users.id = 27 and users.deleted_at IS NULL limit  1

`$user = User::findOrcreate(28, ['name'=>'navaneeth', 'age'=>27]);`

> $user will have newly created data

Database query

> SELECT * FROM users WHERE users.id = 28 and users.deleted_at IS NULL limit  1
>  INSERT INTO users (id, name, age) VALUES (28, 'navaneeth', 27)

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -384,6 +384,10 @@ class Builder
      */
     public function find($id, $columns = ['*'])
     {
+        if (is_null($id)) {
+            return null;
+        }
+
         if (is_array($id) || $id instanceof Arrayable) {
             return $this->findMany($id, $columns);
         }
@@ -483,6 +487,24 @@ class Builder
         }
 
         return tap($this->newModelInstance(array_merge($attributes, $values)), function ($instance) {
+            $instance->save();
+        });
+    }
+
+    /**
+     * Find a model by its primary key. If not found, creates a new record.
+     *
+     * @param  mixed  $id
+     * @param  array  $values
+     * @return \Illuminate\Database\Eloquent\Model|static
+     */
+    public function findOrCreate($id, array $values = [])
+    {
+        if (! is_null($instance = $this->find($id))) {
+            return $instance;
+        }
+
+        return tap($this->newModelInstance($values), function ($instance) {
             $instance->save();
         });
     }


navaneeth-raman-bhaskar:
@taylorotwell eloquent builder method  `find() `  early return added.

src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -487,6 +491,24 @@ public function firstOrCreate(array $attributes = [], array $values = [])
...
     }
 
+    /**
+     * Find a model by its primary key. If not found, creates a new record.
+     *
+     * @param  mixed  $id
+     * @param  array  $values
+     * @return \Illuminate\Database\Eloquent\Model|static
+     */
+    public function findOrCreate($id, array $values = [])
```
mateusjunges, line 501: I don't really see a use case for this. The `firstOrCreate` method is useful because it searches models for any field you want, and if it doesn't exist, a new model will be created with the values ​​you searched for merged with the second array. Since you're not merging id into `$values`, which wouldn't even make sense since `id` is generated at the database level, why would you need to check if any models exist by their id and then create a new one with values ​​that you didn't even use it to check if the model exists in the first place?

navaneeth-raman-bhaskar: Previously
`$user = User::firstOrcreate(['id'=>28], ['name'=>'navaneeth', 'age'=>27]);`

Now
`$user = User::findOrcreate(28, ['name'=>'navaneeth', 'age'=>27]);`

2 benefits
1. No need to type key name.
2. Just in case if key is null then no need to run a DB query.

navaneeth-raman-bhaskar: @mateusjunges If you still think it is useless, what is `findOrNew()` doing in the first place 😏

taylorotwell:

navaneeth-raman-bhaskar:
@taylorotwell 

why should not we early return ? it prevents an unwanted DB query.

navaneeth-raman-bhaskar:
@GrahamCampbell @taylorotwell  Do you want to me to remove` findOrCreate()` method and only retail `find()` method changes?

PR #39997, opened 2021-12-12
[8.x] New whereIf() Base Query Builder method added

navaneeth-raman-bhaskar:
## PR includes a new base query builder` whereIf()` method
Simple where query is applied when condition is true.
If the $arguments are closures then whereIf() exactly act like when() method.

**Example:**
      assume $check = true then,

`User::whereIf($check,'name','like','%laravel%');`
> SELECT * FROM users WHERE name LIKE '%laravel%' and deleted_at IS NULL

`User::whereIf($check,'name','laravel');`
> SELECT * FROM users WHERE name = 'laravel' and deleted_at IS NULL

`User::whereIf(!$check,'name','like','%laravel%');`
> SELECT * FROM users  WHERE deleted_at IS NULL

`User::whereIf($check,'name', fn($q)=>$q->where('name','laravel');`
> SELECT * FROM users WHERE name = 'laravel' and deleted_at IS NULL

`User::whereIf(!$check,'name', fn($q)=>$q->where('name','laravel', fn($q)=>$q->where('name','lumen');`
> SELECT * FROM users WHERE name = 'lumen' and deleted_at IS NULL

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1775,11 +1775,11 @@ class Builder
      * User::whereIf(!$check,'name', fn($q)=>$q->where('name','laravel', fn($q)=>$q->where('name','lumen');
      *      # select * from users name = 'lumen' and deleted_at is null
      *
-     * @param  bool  $condition
+     * @param  mixed  $condition
      * @param  mixed  $arguments
      * @return $this
      */
-    public function whereIf(bool $condition, ...$arguments)
+    public function whereIf($condition, ...$arguments)
     {
         if ($condition) {
             return $arguments[0] instanceof \Closure ? $arguments[0]($this, $condition) : $this->where(...$arguments);
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1761,19 +1761,6 @@ class Builder
 
     /**
      * Apply the where condition if the given $condition is truthy.
-     * If the rest of the $arguments are closures then whereIf() exactly act like when() method.
-     *
-     * @example assume $check = true then
-     * User::whereIf($check,'name','like','%laravel%');
-     *      # select * from users name like '%laravel%' and deleted_at is null
-     * User::whereIf($check,'name','laravel');
-     *      # select * from users name = 'laravel' and deleted_at is null
-     * User::whereIf(!$check,'name','like','%laravel%');
-     *      # select * from users  and deleted_at is null
-     * User::whereIf($check,'name', fn($q)=>$q->where('name','laravel');
-     *      # select * from users name = 'laravel' and deleted_at is null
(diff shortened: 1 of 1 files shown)


mateusjunges:
You can do the same thing using `when`, since Builder uses the `Conditionable` trait:

```php
User::when($check, fn($query) => $query->where('name','like','%laravel%'));
```

navaneeth-raman-bhaskar:
> You can do the same thing using `when`, since Builder uses the `Conditionable` trait:
> 
> ```
> User::when($check, fn($query) => $query->where('name','like','%laravel%'));
> ```



> You can do the same thing using `when`, since Builder uses the `Conditionable` trait:
> 
> ```
> User::when($check, fn($query) => $query->where('name','like','%laravel%'));
> ```

@mateusjunges I know `whereIf() ` is just a syntactic sugar. You don't need to pass closures for simple where queries. That is why `whereIf()` .

src/Illuminate/Database/Query/Builder.php
```diff
@@ -1759,6 +1759,37 @@ public function orWhereJsonLength($column, $operator, $value = null)
         return $this->whereJsonLength($column, $operator, $value, 'or');
     }
 
+    /**
+     * Apply the where condition if the given $condition is truthy.
+     * If the rest of the $arguments are closures then whereIf() exactly act like when() method.
+     *
+     * @example assume $check = true then
```
GrahamCampbell, line 1766: Please remove this from the code, and put this in a PR to the docs repo. :)

src/Illuminate/Database/Query/Builder.php
```diff
@@ -1759,6 +1759,37 @@ public function orWhereJsonLength($column, $operator, $value = null)
...
+     *      # select * from users name = 'laravel' and deleted_at is null
+     * User::whereIf(!$check,'name','like','%laravel%');
+     *      # select * from users  and deleted_at is null
+     * User::whereIf($check,'name', fn($q)=>$q->where('name','laravel');
+     *      # select * from users name = 'laravel' and deleted_at is null
+     * User::whereIf(!$check,'name', fn($q)=>$q->where('name','laravel', fn($q)=>$q->where('name','lumen');
+     *      # select * from users name = 'lumen' and deleted_at is null
+     *
+     * @param  mixed  $condition
+     * @param  mixed  $arguments
```
GrahamCampbell, line 1779: is this phpdoc right?

GrahamCampbell (changes requested):

src/Illuminate/Database/Query/Builder.php
navaneeth-raman-bhaskar: @GrahamCampbell Yes its is. I will remove the @example from the code.

src/Illuminate/Database/Query/Builder.php
navaneeth-raman-bhaskar: Done.

taylorotwell:

PR #39999, opened 2021-12-12
Fixed function Str::slug

bald-cat:
Sometimes product models, for example, mobile phones, have a + symbol in their names.

The function now ignores the sign and leaves a space.

When there are two products that have a difference in the + symbol, two duplicates are formed.

For instance, Mi mix Alpha and Mi mix Alpha +


taylorotwell:
If you need some custom behavior for slug, create your own helper - or, make this replacement before you call Str::slug.

bald-cat:
That's exactly what I did. But I thought it might be useful.

вс, 12 дек. 2021 г., 18:11 Taylor Otwell ***@***.***>:

PR #40002, opened 2021-12-12
[8.x] Add invisible modifier for MySQL columns

lupinitylabs:
This PR adds support for the new `invisible` modifier for columns introduced in MySQL 8.0.23: https://dev.mysql.com/doc/refman/8.0/en/invisible-columns.html

Columns marked as `invisible` are not implicitly selected by `SELECT *` and therefore, the columns will also not be automatically hydrated in a Laravel model. Those columns can still be selected explicitly, however, making it useful for columns that contain data that should only be fetched if really needed.

A concrete use case might be a model that contains several big JSON structures (e.g. with translation strings in multiple languages), but you only want to select specific keys out of those structures (only translations for the current language and a fallback language, for example). If you always fetched the whole translation structure, a lot of bandwidth and memory are wasted, and you will quickly run into sort buffer limits (which is the main reason why I would love to see this supported).

With this, you can use a new `->invisible()` method on the table Blueprint:

```php
    $table->string('secret')->nullable()->invisible();
});
```

This feature is not available in PostgreSQL and I am not aware of any other DBMS supporting this. 
I am not sure if you want any modifiers in that are not widely supported, so this is only a suggestion.

However, I do not see an easy way to turn this into a package?

### DIFF
--- a/src/Illuminate/Database/Schema/ColumnDefinition.php
+++ b/src/Illuminate/Database/Schema/ColumnDefinition.php
@@ -17,6 +17,7 @@ use Illuminate\Support\Fluent;
  * @method $this from(int $startingValue) Set the starting value of an auto-incrementing field (MySQL / PostgreSQL)
  * @method $this generatedAs(string|Expression $expression = null) Create a SQL compliant identity column (PostgreSQL)
  * @method $this index(string $indexName = null) Add an index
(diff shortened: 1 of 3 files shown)


virtual-designer:
Great feature added!!!

PR #40006, opened 2021-12-12
[8.x] Calculate the backoff value from total of exceptions

### LINKED ISSUE #39976: Job backoff array accessed by number of job attempts instead of failures
- Laravel Version: 8.74.0
- PHP Version: 7.4.26

### Description:
The [documentation about job failures](https://laravel.com/docs/8.x/queues#dealing-with-failed-jobs) states about `backoff`:
> Using the --backoff option, you may specify how many seconds Laravel should wait before retrying **a job that has encountered an exception**.

And then:
> You may easily configure "exponential" backoffs by returning an array of backoff values from the backoff method

However, [when accessing this array](https://github.com/laravel/framework/blob/cf00a5621f2eacdb978d3176e6d601193d9dc57d/src/Illuminate/Queue/Worker.php#L600), the 
 `Illuminate\Queue\Worker` class uses `$job->attempts()` and not the number of failures as the index.

Since the number of attempts also increases when using the `release` function, some elements of the backoff array may never be used as the "backoff value" when unhandled exceptions are detected by the queue worker.

For example, if the backoff array is defined as: `[5, 50, 500]`, the following could happen:
```
- Attempt #1: throw new Exception();
... wait 5 seconds
- Attempt #2: $this->release();
... wait 0 seconds
- Attempt #3: throw new Exception();
... wait 500 seconds
- Attempt #4...
```

The number of failures is [counted in the Cache](https://github.com/laravel/framework/blob/cf00a5621f2eacdb978d3176e6d601193d9dc57d/src/Illuminate/Queue/Worker.php#L550) and used for the [max exceptions](https://laravel.com/docs/8.x/queues#max-exceptions) option.
Shouldn't this value also be used as the index when accessing the backoff array?

### Steps To Reproduce:

- Create a Job class with a `backoff` array with at least two elements
- In the first attempt of the job, call `$this->release()`
- In the second job attempt, throw an `Exception`
- The backoff time until the third job attempt will be the second element of the `backoff` array, and not the first.

<details>
  <summary>Example Job class (will never use the 10s backoff)</summary>

```
<?php

namespace App\Jobs;

use Exception;
use Illuminate\Bus\Queueable;

class TestJob implements ShouldQueue
{

    public $tries = 10;
    public $maxExceptions = 5;

    public function __construct()
    {
        //
    }
    
    public function backoff()
    {
        return [10, 20, 30];
    }

    public function handle()
    {
        if ($this->attempts() > 1) throw new Exception();
        $this->release();
    }
}
```
</summary>

MartinPham:
#39976 

This PR makes the Worker to prefer to calculate the backoff value from total of exceptions (when the backoff is configured as an array), and fallback to the total of attempts.

## Situation
- A Job is configured with backoff as an array `[1, 10]`

### Current behavior
- 1st attempt, the job calls `release()`.
- (After `0` seconds) 2nd attempt, the job throws an exception.
- (After `10` seconds) 3rd attempt, the job throws an exception.
- (After `10` seconds) 4th attempt, the job throws an exception.
...

### Proposed behavior
- 1st attempt, the job calls `release()`.
- (After `0` seconds) 2nd attempt, the job throws an exception.
- (After `1` seconds) 3rd attempt, the job throws an exception.
- (After `10` seconds) 4th attempt, the job throws an exception.
...

Notice that on the 2nd attempt, the backoff was `10` seconds

### DIFF
--- a/src/Illuminate/Queue/Worker.php
+++ b/src/Illuminate/Queue/Worker.php
@@ -556,7 +556,7 @@ class Worker
         $totalExceptions = $this->cache->get($jobExceptionsKey);
         if (is_null($totalExceptions)) {
             $this->cache->put($jobExceptionsKey, 0, Carbon::now()->addDay());
-            
+
             return 0;
         }
 
--- a/src/Illuminate/Queue/Worker.php
+++ b/src/Illuminate/Queue/Worker.php
@@ -529,6 +529,40 @@ class Worker
     }
 
     /**
+     * Get the stored key for job's total number of exceptions
+     *
+     * @param  \Illuminate\Contracts\Queue\Job  $job
+     * @return string
+     */
+    private function getJobTotalExceptionsKey($job)
+    {
+        return 'job-exceptions:' . $job->uuid();
+    }
+
+    /**
+     * Get the job's total number of exceptions
+     *
+     * @param  \Illuminate\Contracts\Queue\Job  $job
+     * @return int|null
+     */
+    private function getJobTotalExceptions($job)
+    {
+        if (! $this->cache || is_null($job->uuid())) {
+            return;
+        }
+
+        $jobExceptionsKey = $this->getJobTotalExceptionsKey($job);
+
+        $totalExceptions = $this->cache->get($jobExceptionsKey);
+        if (is_null($totalExceptions)) {
+            $this->cache->put($jobExceptionsKey, 0, Carbon::now()->addDay());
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think this would be a breaking change to release on the current release since people may be relying on the current behavior when manually calling `release`. We would like need to make such a change on Laravel 9 (master branch) if such a change needs to be made at all.

PR #40010, opened 2021-12-13
[8.x] Replace strpos by Str::startsWith in Application class

oliverquynh:
In in Application class, `Illuminate\Support\Str` has already been imported and used. Therefore, we can use `Str::startsWith` instead of `strpos($haystack, $needle) === 0`


taylorotwell:
I think this is more of a maximum performance thing.

PR #40014, opened 2021-12-13
[9.x] Support x-forwarded-prefix in TrustProxies.php

ppodds:
#39996 

If we want to deploy laravel project in a subfolder of the domain like the following Nginx setting. We need to use the x-forwarded-prefix header.

```
server {
  listen 80;
  listen [::]:80;

  server_name _;
  location /to/project/ {
    proxy_set_header X-Forwarded-Prefix /to/project;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
    proxy_set_header X-Forwarded-Host my.domain;
    proxy_set_header X-Forwarded-Port 443;
    # laravel backend at localhost:8080
    proxy_pass http://localhost:8080/;
}
```

Symfony has already supported the x-forwarded-prefix header. We can deploy laravel in the subfolder of domain with this feature.

### DIFF
--- a/src/Illuminate/Http/Middleware/TrustProxies.php
+++ b/src/Illuminate/Http/Middleware/TrustProxies.php
@@ -19,7 +19,7 @@ class TrustProxies
      *
      * @var int
      */
-    protected $headers = Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_HOST | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PROTO | Request::HEADER_X_FORWARDED_AWS_ELB;
+    protected $headers = Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_HOST | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PROTO | Request::HEADER_X_FORWARDED_PREFIX | Request::HEADER_X_FORWARDED_AWS_ELB;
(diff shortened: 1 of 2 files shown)


src/Illuminate/Http/Middleware/TrustProxies.php
```diff
@@ -117,8 +117,12 @@ protected function getTrustedHeaderNames()
             case Request::HEADER_X_FORWARDED_PROTO:
                 return Request::HEADER_X_FORWARDED_PROTO;
 
+            case 'HEADER_X_FORWARDED_PREFIX':
+            case Request::HEADER_X_FORWARDED_PREFIX:
+                return Request::HEADER_X_FORWARDED_PREFIX;
+
```
mateusjunges, line 123: Just a suggestion, but since you are targetting 9.x, this whole switch statement can be replaced with php's `match` function. What do you think?

ppodds: I like this suggestion. Should I close this PR and replace it with php's `match` function and open another PR for this?

mateusjunges: i can't answer that. @driesvints can help us here

driesvints: Let's focus on the change itself and we can do style changes in a different PR.

PR #40029, opened 2021-12-14
[9.x] Prevent further lookups of user if not found

jaulz:
In case there is still a user id in the session and the user was deleted, the framework tries to retrieve the user every time the user() method is called. Right now it has the possibility to significantly decrease the performance of the app because the user() function can easily be triggered a few dozens times and each call causes a new database request. This PR simply sets a flag to prevent further lookups.

This is a follow-up of https://github.com/laravel/framework/pull/39973 which suggested the change to `8.x` but it was rejected by Taylor since he did not want it to be part of a patch release.

### DIFF
--- a/src/Illuminate/Auth/SessionGuard.php
+++ b/src/Illuminate/Auth/SessionGuard.php
@@ -103,6 +103,13 @@ class SessionGuard implements StatefulGuard, SupportsBasicAuth
     protected $recallAttempted = false;
 
     /**
+     * Indicates if a user retrieval has been attempted.
+     *
+     * @var bool
+     */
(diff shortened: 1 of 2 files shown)


Jubeki:
Would it not better to logout the "deleted" user instead of setting a flag? This would also remove the lookup from the database for further requests which are made.

jaulz:
@Jubeki I also thought about it before but what if the user can simply not be found because he is locked (via a custom user provider that checks a flag in the user record). Once he is unlocked I think he should be logged in again automatically.

taylorotwell:
I'm not changing any of this.

PR #40030, opened 2021-12-14
[8.x] Fix Password validation failure to allow errors after min rule

PovilasKorop:
I've noticed this behavior after this comment on my Youtube channel:

Screenshot 2021-12-14 at 11 44 44

Indeed, if there's a `min:X` rule failure in the Password class, then it doesn't even check other errors, fails immediately with `return`. All other validation failures just call `$this->fail` but without the `return`, which makes sense, as they add more potential errors to the messages array.

With this change in PR, it would show all errors: 

Screenshot 2021-12-14 at 11 46 01

### DIFF
--- a/src/Illuminate/Validation/Rules/Password.php
+++ b/src/Illuminate/Validation/Rules/Password.php
@@ -270,34 +270,37 @@ class Password implements Rule, DataAwareRule, ValidatorAwareRule
     {
         $this->messages = [];
 
-        $validator = Validator::make($this->data, [
-            $attribute => 'string|min:'.$this->min,
-        ], $this->validator->customMessages, $this->validator->customAttributes);
+        $validator = Validator::make(
+            $this->data,
+            [$attribute => 'string|min:'.$this->min],
+            $this->validator->customMessages,
+            $this->validator->customAttributes
+        )->after(function ($validator) use ($attribute, $value) {
+            if (! is_string($value)) {
+                return;
+            }
+
+            $value = (string) $value;
+
+            if ($this->mixedCase && ! preg_match('/(\p{Ll}+.*\p{Lu})|(\p{Lu}+.*\p{Ll})/u', $value)) {
+                $validator->errors()->add($attribute, 'The :attribute must contain at least one uppercase and one lowercase letter.');
+            }
+
+            if ($this->letters && ! preg_match('/\pL/u', $value)) {
+                $validator->errors()->add($attribute, 'The :attribute must contain at least one letter.');
+            }
+
+            if ($this->symbols && ! preg_match('/\p{Z}|\p{S}|\p{P}/u', $value)) {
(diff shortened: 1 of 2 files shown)


PovilasKorop:
Ok, it seems that it's intended behavior, according to the tests that are failing.
Or maybe should I re-write the tests for the updated behavior?

@nunomaduro as the original author, maybe you can share your opinion? Thanks!

crynobone:
1. The additional rules would fail if the password value is not `string`.
2. Running `uncompromised` rules on every request when we already know it would failed would just waste 3rd party provider.

Anyway, I PR the following to handle above requirements https://github.com/PovilasKorop/framework/pull/1

PovilasKorop:
@crynobone good catch, thanks for flagging! So it's much more complicated than I thought, I guess not worth bothering then, I thought it would be a quick fix.

If no one else wants to take it further, will close this PR.

crynobone:
> If no one else wants to take it further, will close this PR.

See PovilasKorop#1

PovilasKorop:
@crynobone sorry misunderstood while reading from my phone. Merged now, thanks!

src/Illuminate/Validation/Rules/Password.php
```diff
@@ -270,34 +270,37 @@ public function passes($attribute, $value)
...
-            $this->fail('The :attribute must contain at least one symbol.');
-        }
-
-        if ($this->numbers && ! preg_match('/\pN/u', $value)) {
-            $this->fail('The :attribute must contain at least one number.');
-        }
-
-        if (! empty($this->messages)) {
-            return false;
+            $this->fail($validator->messages()->all());
```
crynobone, line 303:
```suggestion
            return $this->fail($validator->messages()->all());
```

Probably a bad merge from previous PR.

PovilasKorop: @crynobone thanks! Fixed now, tests are successful. Though, still not sure about this change, cause changing the tests is risky. We'll see if it gets merged.

humaneguy:
Nice work @PovilasKorop

src/Illuminate/Validation/Rules/Password.php
```diff
@@ -270,36 +270,39 @@ public function passes($attribute, $value)
...
-        $validator = Validator::make($this->data, [
-            $attribute => 'string|min:'.$this->min,
-        ], $this->validator->customMessages, $this->validator->customAttributes);
+        $validator = Validator::make(
+            $this->data,
+            [$attribute => 'string|min:'.$this->min],
+            $this->validator->customMessages,
+            $this->validator->customAttributes
+        )->after(function ($validator) use ($attribute, $value) {
+            if (! is_string($value)) {
```
taylorotwell, line 279: @PovilasKorop why was this added? Isn't `string` baked into the validation rules above?

PovilasKorop: @crynobone a question for you here

PR #40034, opened 2021-12-14
[9.x] Allows to remove the `server.php` file

nunomaduro:
The `serve` artisan command makes use of the `server.php` file that exists on the root of each Laravel project. This file, it's basically the same since 11 Jan 2013, and never needs to be modified at user-land.

So this pull request proposes that we remove the file from the Laravel 9 skeleton (user-land), and move it to the framework itself. 

https://github.com/laravel/laravel/pull/5747

### DIFF
--- a/src/Illuminate/Foundation/Console/ServeCommand.php
+++ b/src/Illuminate/Foundation/Console/ServeCommand.php
@@ -49,8 +49,6 @@ class ServeCommand extends Command
      */
     public function handle()
     {
-        chdir(public_path());
-
         $this->line("<info>Starting Laravel development server:</info> http://{$this->host()}:{$this->port()}");
 
         $environmentFile = $this->option('env')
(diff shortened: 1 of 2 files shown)


assertchris:
I absolutely love this change. Thank you.

haruncpi:
I like it 👍

Krisell:
Is it going to be possible to keep using a custom file? We have made some adjustments to add cache-headers to static files to improve performance of Cypress-tests. Of course, we could run the test server differently, but using the built-in server is very convenient.

BrandonSurowiec:
@Krisell Yes, you can still keep your customized `server.php` and it will read from it.

```
$server = file_exists(base_path('server.php'))
            ? base_path('server.php')
            : __DIR__.'/../resources/server.php';
```

Krisell:
> @Krisell Yes, you can still keep your customized `server.php` and it will read from it.
> 
> ```
> $server = file_exists(base_path('server.php'))
>             ? base_path('server.php')
>             : __DIR__.'/../resources/server.php';
> ```

Oh wow I can't believe I didn't see that piece when glancing at the code. Thanks!

PR #40036, opened 2021-12-14
Add string version of true and false

jhouie:
I created this pull request because the validation of boolean fields did not accept the **true** and **false** strings as valid values, which ends up causing confusion in the rule operation.


GrahamCampbell:
This is quite a breaking change.

taylorotwell:
It would be a a breaking change for us to widen the accepted values of this rule on a patch release.

jhouie:
Could it be something introduced in an upcoming release?

jhouie:
For now, I've resolved by extending the validation rule

PR #40041, opened 2021-12-14
[9.x] Added new string versions of true and false to the Boolean validation rule

jhouie:
I created this pull request because the validation of boolean fields did not accept the true and false strings as valid values, which ends up causing confusion in the rule operation.

I didn't create new tests but modified the existing ones to work correctly with the new values.

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -395,7 +395,7 @@ trait ValidatesAttributes
      */
     public function validateBoolean($attribute, $value)
     {
-        $acceptable = [true, false, 0, 1, '0', '1'];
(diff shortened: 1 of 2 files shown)


igorgaming:
It works correctly, because php does not perceive the strings `'true'` and `'false'` as boolean, but '0' and '1' does. 
Therefore, the code below will always output `'True'`

image

williamxsp:
This will also create a lot of breaking changes because if youre passing false or 0 to a database with a column tinyint, it will be store properly but 'false' and 'true' not.

taylorotwell:
I don't think I'm going to change this behavior. The accepted values are documented.

andrey-helldar:
@igorgaming, just use a boolean value along with a file upload in the same request. Not `1` or `0`, namely `true` or `false`.

PR #40047, opened 2021-12-14
Fixes slight oversight

xanderificnl:
In my previous PR, I made a small error where I used the same key for a translation string as a filename (i.e. `__("Pagination")` and `en/pagination.php`) which caused issues for someone (see: https://github.com/laravel/framework/pull/39928#issuecomment-994112464) and does not line up with best practices explained in the documentation (see "Key / File Conflicts" https://laravel.com/docs/8.x/localization#using-translation-strings-as-keys)

This PR goes hand in hand with https://github.com/laravel/laravel/pull/5748 -- and should resolve the aforementioned issue and oversight.

### DIFF
--- a/src/Illuminate/Pagination/resources/views/simple-tailwind.blade.php
+++ b/src/Illuminate/Pagination/resources/views/simple-tailwind.blade.php
@@ -1,5 +1,5 @@
 @if ($paginator->hasPages())
-    <nav role="navigation" aria-label="{{ __('pagination.label') }}" class="flex justify-between">
+    <nav aria-label="{{ __('pagination.label') }}" class="flex justify-between">
         {{-- Previous Page Link --}}
         @if ($paginator->onFirstPage())
             <span class="relative inline-flex items-center px-4 py-2 text-sm font-medium text-gray-500 bg-white border border-gray-300 cursor-default leading-5 rounded-md">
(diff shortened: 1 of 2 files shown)


derekmd:
This fix would only be appropriate for new installs. Existing apps will be updating the next 8.x minor patch release and now have a translation key `pagination.label` appearing in HTML because the `en/pagination.php` file doesn't contain this new key.

Choosing another non-conflicting human-readable phrase like `aria-label="{{ __('Page navigation') }}"` is likely a better fix. "Pagination" is a term really only known by programmers so any screenreader should interpret a component description targeted at non-technical users.

xanderificnl:
In that case, it should be `aria-label="{{ __('Page') }}"` which could potentially cause the same issue, if someone has lang/x/page.php and no lang/x.json. Setting the label to 'Page Navigation' will cause screen readers to read "Page Navigation Navigation" (because the rol of <nav> is 'navigation').

driesvints:
I feel like it's best we just revert the original PR for now.

xanderificnl:
To summarize:

- This fix won't work (as-is) for minor updates.
- We shouldn't duplicate the role (navigation) in the label.
- Using 'page' (or any single word) may introduce problems w/ language files.

So we should probably just revert it for now, and include the fix in the next major version?

taylorotwell:
If you want to change this - you can export the views and change it in your own application.

PR #40048, opened 2021-12-15
[9.x] Allow http client to call send-related method without parameter.

chu121su12:
Based on #39934.

Similar to query builder that can be passed around before actually `->get()`-ing the result,
I want to be able to pass around `PendingRequest` so it can be send later. Ex:

```php
[$method, $http] = ['post', \Http::withOptions([...])->url($url)->withData((array) $body)];
// later
$response = $http->$method(); // options, url, body, etc already defined
```

Additional benefit is that you can now do bodied-request with query string.


taylorotwell:

PR #40050, opened 2021-12-15
[9.x] Proper fix when provided path is '0'

vlakoff:
#40013 was incorrect, because when provided argument was '0',
we were appending `$path` (i.e. '0') instead of `DIRECTORY_SEPARATOR.$path` (i.e. '/0').

To summarize the behavior after this PR:

| argument       | appended string |
|----------------|-----------------|
| `'foo'`        | `'/foo'`        |
| `'0'`          | `'/0'`          |
| `''` (default) | `''`            |

In addition, as I'm using a loose comparison, it supports the following cases (they are not supported nor expected, though it doesn't harm to work better with them):

| argument | appended string |
|----------|-----------------|
| `null`   | `''`            |
| `false`  | `''`            |

### DIFF
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -460,7 +460,7 @@ class Application extends Container implements ApplicationContract, CachesConfig
     public function storagePath($path = '')
     {
         return ($this->storagePath ?: $this->basePath.DIRECTORY_SEPARATOR.'storage')
-                            .($path ? DIRECTORY_SEPARATOR.$path : $path);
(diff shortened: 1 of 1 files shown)


iamine:
i think this fix must be applied to other functions too such as: path, basePath, bootstrapPath, configPath.....

vlakoff:
As a reminder, I used `$path != ''` which is equivalent to `(string) $path !== ''`.
I choosed the former because it is shorter, but the latter might be preferable so people are not inclined to think that `$path != ''` should be replaced with with `$path !== ''`.

PR #39075, opened 2021-10-04
[9.x] Chunk collections with callback

mateusjunges:
This is a follow up on #39065

This PR adds an optional callback to the chunk method on `Illuminate\Support\Collection` and `Illuminate\Support\LazyCollection` classes.
This allows to use a callback when chunking collections:

```php
$collection = new Collection([1, 2, 3, 4, 5, 6, 7, 8, 9, 10]);

$collection->chunk(5, function(Collection $items) {
    return $items->filter(function($item) use ($items) {
        return $item > 1;
    });
})
```

The code above returns a Collection of chunked collections, rejecting the value `1` in the first collection: 

```
Illuminate\Support\Collection {#284 ▼
  #items: array:2 [▼
    0 => Illuminate\Support\Collection {#285 ▼
      #items: array:4 [▼
        1 => 2
        2 => 3
        3 => 4
        4 => 5
      ]
    }
    1 => Illuminate\Support\Collection {#286 ▼
      #items: array:5 [▼
        5 => 6
        6 => 7
        7 => 8
        8 => 9
        9 => 10
      ]
    }
  ]
}
```

PR #40053, opened 2021-12-15
Added a callback closure for chunk function.

karam-mustafa:
I was working on something that required me to do a filter every time I split an array with certain data using the chunk function, so I added this callback to the chunk function so that I could apply another collection to the data in every chunk.


mateusjunges:
This was already rejected #39075

karam-mustafa:
@mateusjunges 
I think there is a difference. My PR applies the callback for each iteration in the internal chunk loop, the `each(fn) `function applies the callback function after chunks are done.
And I think I should update the interfaces and tests as well

taylorotwell:

PR #40058, opened 2021-12-15
[8.x] Add `setApplication` method to `FilesystemManager`

AlexVanderbist:
The `setApplication` method will be used to pass in a new `Application` to the filesystem manager by Octane when relevant (cfr. https://github.com/laravel/octane/blob/1.x/src/Listeners/GiveNewApplicationInstanceToDatabaseManager.php).

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemManager.php
+++ b/src/Illuminate/Filesystem/FilesystemManager.php
@@ -312,6 +312,7 @@ class FilesystemManager implements FactoryContract
 
         return $this;
     }
+
     /**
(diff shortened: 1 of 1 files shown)


src/Illuminate/Filesystem/FilesystemManager.php
```diff
@@ -312,7 +312,7 @@ public function set($name, $disk)
 
         return $this;
     }
-
+
```
Neol3108, line 315:
```suggestion

```

PR #40059, opened 2021-12-15
[8.x] move maintenance mode logic

wimulkeman:

Overwriting the logic may be needed if for example the application is running on multiple servers.

Fixes #36474

### DIFF
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -14,6 +14,7 @@ use Illuminate\Foundation\Bootstrap\LoadEnvironmentVariables;
 use Illuminate\Foundation\Events\LocaleUpdated;
 use Illuminate\Http\Request;
 use Illuminate\Log\LogServiceProvider;
+use Illuminate\Maintenance\MaintenanceMode;
 use Illuminate\Routing\RoutingServiceProvider;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
@@ -1109,7 +1110,7 @@ class Application extends Container implements ApplicationContract, CachesConfig
      */
     public function isDownForMaintenance()
     {
-        return file_exists($this->storagePath().'/framework/down');
+        return app(MaintenanceMode::class)->isDown();
     }
 
     /**
--- a/src/Illuminate/Foundation/Console/UpCommand.php
+++ b/src/Illuminate/Foundation/Console/UpCommand.php
@@ -5,6 +5,7 @@ namespace Illuminate\Foundation\Console;
 use Exception;
 use Illuminate\Console\Command;
 use Illuminate\Foundation\Events\MaintenanceModeDisabled;
+use Illuminate\Maintenance\MaintenanceMode;
 
 class UpCommand extends Command
 {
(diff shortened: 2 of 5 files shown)


vurpa:
I think that up/down naming can be a bit confusing, because I think of it the opposite way of how it's currently implemented.

Something like enable/disable feels like more natural verbs to work with.

wimulkeman:
@vurpa The enter/leave names would be a good replacement. I used the up/down names for now to stay in sync with the commands used to start and stop the maintenance mode. If requested by the core developers I would be glad to change it.

src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php
```diff
@@ -23,15 +24,21 @@ class PreventRequestsDuringMaintenance
...
+    private $maintenanceMode;
+
     /**
      * Create a new middleware instance.
      *
      * @param  \Illuminate\Contracts\Foundation\Application  $app
      * @return void
      */
-    public function __construct(Application $app)
+    public function __construct(Application $app, MaintenanceMode $maintenanceMode)
```
driesvints, line 38: Adding a new dependency here is a breaking change unfortunately.

src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php
```diff
@@ -23,15 +24,21 @@ class PreventRequestsDuringMaintenance
      */
     protected $except = [];
 
+    /**
+     * @var MaintenanceMode
```
driesvints, line 28: Please use FQN and a short description for DocBlocks

src/Illuminate/Foundation/Console/UpCommand.php
```diff
@@ -27,16 +28,16 @@ class UpCommand extends Command
      *
      * @return int
      */
-    public function handle()
+    public function handle(MaintenanceMode $maintenanceMode)
     {
         try {
-            if (! is_file(storage_path('framework/down'))) {
+            if ($maintenanceMode->isDown() === false) {
```
driesvints, line 34: Maybe add an `isUp` method as well.

src/Illuminate/Foundation/Console/UpCommand.php
```diff
@@ -27,16 +28,16 @@ class UpCommand extends Command
      *
      * @return int
      */
-    public function handle()
+    public function handle(MaintenanceMode $maintenanceMode)
```
driesvints, line 31: This is a breaking change.

src/Illuminate/Foundation/Application.php
```diff
@@ -1109,7 +1110,7 @@ public function addAbsoluteCachePathPrefix($prefix)
      */
     public function isDownForMaintenance()
     {
-        return file_exists($this->storagePath().'/framework/down');
+        return app(MaintenanceMode::class)->isDown();
```
driesvints, line 1113:
```suggestion
        return $this->resolve(MaintenanceMode::class)->isDown();
```

driesvints:
There's quite a few breaking changes here so I suggest this is best sent to master. Please also take care of the other comments. Thanks

src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php
wimulkeman: @driesvints You're right. It would for example break implementation if people have extended the class. Should I point the update to the Laravel 9 branch instead?

PR #40062, opened 2021-12-15
[9.x] Removes core class aliases from the skeleton

nunomaduro:
This pull request allows removing core class aliases from the `config/app.php` from the skeleton in Laravel 9 — **about 46 lines removed**: **https://github.com/laravel/laravel/pull/5751**.

The motivation for this pull request is the same https://github.com/laravel/framework/pull/40034, those aliases are rarely (never?) modified at user-land. So, it's kind of safe to move them to the core, giving the skeleton a little bit more space to breathe.

Now, the approach of this pull request is simple, if this pull request gets merged, the core aliases are moved to a `Facades::defaultAliases()`, and the skeleton `app.aliases` calls that method to get the core aliases. Allowing the user to still fully disable this feature by removing the `app.aliases` configuration key.

Is worth mentioning that this pull request does not contain breaking changes. Also, if this pull request gets merged, I will remove this section in the docs too: https://laravel.com/docs/8.x/redis#the-redis-facade-alias.


taylorotwell:
One small caveat to this approach is if you want to define even one additional alias for their own classes you now need to copy this entire list into the skeleton.

You could get around that by making the `laravel/laravel` config like this in Laravel 9:

```php
'aliases' => [
    ...Application::facadeAliases(),
    // ...
],
```

src/Illuminate/Foundation/Bootstrap/RegisterFacades.php
```diff
@@ -17,12 +21,51 @@ class RegisterFacades
...
     {
-        Facade::clearResolvedInstances();
+        Facades\Facade::clearResolvedInstances();
 
-        Facade::setFacadeApplication($app);
+        Facades\Facade::setFacadeApplication($app);
 
         AliasLoader::getInstance(array_merge(
-            $app->make('config')->get('app.aliases', []),
+            $app->make('config')->get('app.aliases', [
```
igorgaming, line 29: Can't we move the core aliases above this line in `array_merge`? 
Thus, when adding our custom aliases, we do not have to copy this, as Taylor wrote above.

taylorotwell: That means there is no way to totally disable aliases.

igorgaming: Yes, but as the author of this PR noted, they are basically not modified by most people in any way.

nunomaduro: @igorgaming After discussing with Taylor internally, we have applied your suggestion.

nunomaduro:
@taylorotwell @igorgaming Updated the description of the pull request in light of our recent discussion.

tests/Foundation/Bootstrap/RegisterFacadesTest.php
```diff
@@ -0,0 +1,150 @@
...
+
+        $this->assertTrue(class_exists('MyPackageApp'));
+        $reflection = new ReflectionClass('MyPackageApp');
+
+        $this->assertSame(Facades\App::class, $reflection->getName());
+
+        $this->assertCount(39, AliasLoader::getInstance()->getAliases());
+    }
+
+    protected function getDefaultAliases()
```
mfn, line 107: Maybe this method should be part of `class RegisterFacades` so the array below does not have to be duplicated in tests?

mfn (commented):
> If the user wishes to add more aliases or override existing ones

How would I go about **replacing** the default ones?

Mine look currently like this:
```php
    'aliases' => [
        // Necessary due to use as `@mixin \Eloquent` in models due to ide-helper
        'Eloquent' => Illuminate\Database\Eloquent\Model::class,
    ],
```
And that is only out of necessity with ide-helper, actually.

I read the arguments about aliases being lazy and thus having no impact, but I can also tell from my experience, developers want to have specific control over this and _want to make sure they're disabled so they're not getting accidentally used_.

Thanks

koenhoeijmakers:
Wouldn't this make it way more inconvenient to disable all aliases?

Use case: i preferably remove all aliases so that our team properly uses the class definition so that every facade used also gives the expected intellisense

taylorotwell:
We should make the `app.aliases` configuration entry look like this:

```php
'aliases' => [
    ...Facade::aliases(),
    // ...
],
```

Then, if someone wants to totally disable the built-in aliases they can easily do so. In addition, they can disable the built-in aliases while still retaining the ability to define their own.

nunomaduro:
@taylorotwell Both pull requests ready now. Worth to mention that the result of the operation bellow won't override the default `View` alias. People would have to sum the `Facade::defaultAliases()` as last element.

```
Facade::defaultAliases() + [
    'View' => 'Nuno',
],
```

BrandonSurowiec:
Would it be weird to do:

```
'aliases' => Facade::defaultAliases([
    'View' => 'Nuno',
]),
```

Then the `Facade::defaultAliases` could merge the array on the inside. To override with the current setup, we'd `array_merge` or use the `+` operator:

```
'aliases' => array_merge(Facade::defaultAliases(), [
    'View' => 'Nuno',
]),
```

```
'aliases' => [
    'View' => 'Nuno',
] + Facade::defaultAliases(),
```

mfn:
> Would it be weird to do:

TBH, the moment a method is being called in the config file to provide the defaults (for backwards compat), started to look weird to me. Though I do in fact appreciate the suggestion because exactly of "backwards compat".

RoodFruit:
@nunomaduro Heads-up, in the latest Laravel 9.0.0-beta.3 branch, `Facade::defaultAliases()` now returns a `Collection`. Meaning the `array_merge()` in `app.php` won't work anymore.

Easy fix of course by adding `->toArray()`.
```
    'aliases' => array_merge(Facade::defaultAliases()->toArray(), [
        // ...
    ]),
```

PR #40044, opened 2021-12-14
[8.x] Add missing method to Collection and LazyCollection

tomhrtly:
Small PR to add the reverse of the `contains` method for collections.

PR #40065, opened 2021-12-15
[8.x] Update conditionals for Collection and LazyCollection contains methods

tomhrtly:
A small fix from my previous [PR](https://github.com/laravel/framework/pull/40044).

When calling the following:
```
$collection = collect([1, 2, 3, 4, 5]);

$collection->doesntContain(function ($value, $key) {
    return $value < 5;
});
```

A TypeError is returned by the method because the `operator` and `value` values are carried over to the `contains` method, therefore `func_num_args()` is equal to 3 instead of 1.

I've updated the conditionals so that it checks that the `operator` and `value` values are null instead and now the method works as expected.

Sorry bout that!

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -163,7 +163,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
      */
     public function contains($key, $operator = null, $value = null)
     {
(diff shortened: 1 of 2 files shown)


taylorotwell:
Eh, that's different behavior than func_num_args. I'll need to adjust `doesntContain`.

taylorotwell:
Fixing now.

PR #40066, opened 2021-12-15
[8.x] Add `Str::replaceBetween()` and `Str::of($string)->replaceBetween()` methods to replace part of a string between two markers

ralphjsmit:
This pull request adds the ability for developers to replace a part of string between two substrings.

## What & why

Consider the following example, where I want to replace a namespace in a string. To accomplish that, I'd need to make several calls to an `after()`, `before()` and `replace()` method. 

```php

$string = 'namespace App\Models\User;';

Str::replace(
    Str::of($string)->after('namespace ')->before(';'),
    'Support\Models\User',
    $string
);
// 'namespace Support\Models\User;'
```

This is rather verbose for a very simple action. Now you can simply do this:

```php
Str::replaceBetween('namespace ', ';', 'Support\Models\User', 'namespace App\Models\User;')
// 'Support\Models\User;'
```

The fluent method is even better:
```php
(string) Str::of('namespace App\Models\User;')->replaceBetween('namespace ', ';', 'Support\Models\User');
// 'Support\Models\User;'
```

In the case that there are multiple occurrences of the `$start` or `$end` variable, it will replace the part that matches `Str::of($string)->after($start)->before($end)`. In other words, it will replace everything directly after the first occurrence of `$start` until the first occurrence of `$end`, even if that includes a second occurrence of `$start`.

This PR would allow developers to clean up their code a lot more and it fits nicely together with the existing replace methods, like `replaceArray()`, `replaceFirst()`, `replaceLast()`, etc.

NB. There were errors with Docker when running the tests in GitHub Actions. The tests are fully working, so please trigger GitHub Actions again if you're reviewing this. Hopefully the Docker issue will then be resolved. Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 /vendor
-/vendor.nosync
 composer.phar
 composer.lock
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /vendor
+/vendor.nosync
 composer.phar
 composer.lock
 .DS_Store
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -615,7 +615,7 @@ class Str
         }
 
         return self::replace(
-            (new Stringable($subject))->after($start)->before($end),
+            self::of($subject)->after($start)->before($end),
             $replace,
             $subject
         );
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -600,6 +600,28 @@ class Str
     }
 
     /**
+     * Replace part of a string between two given substrings.
+     *
+     * @param  string  $start
+     * @param  string  $end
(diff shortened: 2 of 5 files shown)


koenhoeijmakers:
I don't see how this can give reliable results, if i wanted to replace `john` with `james` between `peter` and `daisy` in the string `peterjohndaisy peterjohndaisy` what would it do? Should it replace the first instance, or the second, or both?

ralphjsmit:
> Should it replace the first instance, or the second, or both?

It should replace the first instance in your example. 

> In the case that there are multiple occurrences of the $start or $end variable, it will replace the part that matches Str::of($string)->after($start)->before($end). In other words, it will replace everything directly after the first occurrence of $start until the first occurrence of $end, even if that includes a second occurrence of $start.

E.g. it will replace everything after the first 'peter' until the first 'daisy'.

taylorotwell:

PR #40070, opened 2021-12-16
[9.x] Move maintenance mode logic

wimulkeman:

Overwriting the logic may be needed if for example the application is run on multiple servers. In this case the developer would want to extend the storage logic for the maintenance mode to implement a solution that can change the maintenance state for all the servers at once and reverse the maintenance mode with one command.

By separating the maintenance mode logic the framework opens the door for packages that could handle the maintenance mode through different mechanisms like using the cache (as proposed in a blog https://www.raypold.com/posts/laravel-distributed-maintenance-mode/).

The need for this was shown within the discussions laravel/ideas#1944 and #36474.

This PR is a follow-up to the PR https://github.com/laravel/framework/pull/40059 with suggested changes and pointed to the master instead of the 8.x branch because it contains backward incompatible changes.


src/Illuminate/Maintenance/MaintenanceMode.php
```diff
@@ -0,0 +1,43 @@
...
+    public function getPayload(): array
+    {
+        return json_decode(file_get_contents($this->getDownFilePath()), true);
+    }
+
+    private function getDownFilePath(): string
+    {
+        return storage_path('framework/down');
+    }
+}
```
driesvints, line 43: Please move this class into the Foundation namespace. I don't think it makes sense to create an entire new component for this. Thanks

wimulkeman: Done. Thanks for reviewing the code

src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php
```diff
@@ -23,15 +24,24 @@ class PreventRequestsDuringMaintenance
      */
     protected $except = [];
 
+    /**
+     * The maintenance mode implementation. Can be used to check
+     * if the application is under maintenance.
```
driesvints, line 29:
```suggestion
     * The MaintenanceMode instance.
```

src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php
```diff
@@ -23,15 +24,24 @@ class PreventRequestsDuringMaintenance
...
+     * @var \Illuminate\Foundation\MaintenanceMode
+     */
+    protected $maintenanceMode;
+
     /**
      * Create a new middleware instance.
      *
      * @param  \Illuminate\Contracts\Foundation\Application  $app
-     * @return void
+     * @param  \Illuminate\Foundation\MaintenanceMode  $maintenanceMode
```
driesvints, line 39:
```suggestion
     * @param  \Illuminate\Foundation\MaintenanceMode  $maintenanceMode
     * @return void
```

src/Illuminate/Foundation/MaintenanceMode.php
```diff
@@ -0,0 +1,45 @@
+<?php
+
+namespace Illuminate\Foundation;
+
+use function storage_path;
+
+class MaintenanceMode
```
driesvints, line 7: Maybe this sounds better as `Maintenance`?

src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php
driesvints: I wonder if we can add a `maintenance` method to the `Application` class that returns the `Maintenance` instance? That way we don't need an extra dependency here. And you can call `$this->app->maintenance()->` for these calls.

driesvints (changes requested):

src/Illuminate/Foundation/MaintenanceMode.php
```diff
@@ -0,0 +1,45 @@
...
+
+        unlink($this->getDownFilePath());
+    }
+
+    public function getPayload(): array
+    {
+        return json_decode(file_get_contents($this->getDownFilePath()), true);
+    }
+
+    private function getDownFilePath(): string
```
driesvints, line 41: Add DocBlocks to all these method please.

driesvints (changes requested):

src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php
wimulkeman: That would indeed remove the dependency and make it more available within the application as a whole. I will add it to the Application class

src/Illuminate/Foundation/Console/DownCommand.php
```diff
@@ -42,21 +43,19 @@ class DownCommand extends Command
...
      *
+     * @param  \Illuminate\Foundation\MaintenanceMode  $maintenanceMode
      * @return int
      */
-    public function handle()
+    public function handle(MaintenanceMode $maintenanceMode)
     {
         try {
-            if (is_file(storage_path('framework/down'))) {
+            if ($maintenanceMode->isDown()) {
```
driesvints, line 52:
```suggestion
            if ($this->laravel->maintenanceMode()->isDown()) {
```

src/Illuminate/Foundation/Console/DownCommand.php
```diff
@@ -42,21 +43,19 @@ class DownCommand extends Command
...
                 $this->comment('Application is already down.');
 
                 return 0;
             }
 
-            file_put_contents(
-                storage_path('framework/down'),
-                json_encode($this->getDownFilePayload(), JSON_PRETTY_PRINT)
-            );
+            $maintenanceMode->down($this->getDownFilePayload());
```
driesvints, line 58:
```suggestion
            $this->laravel->maintenanceMode()->down($this->getDownFilePayload());
```

src/Illuminate/Foundation/Console/UpCommand.php
```diff
@@ -34,18 +35,19 @@ class UpCommand extends Command
...
      *
+     * @param  \Illuminate\Foundation\MaintenanceMode  $maintenanceMode
      * @return int
      */
-    public function handle()
+    public function handle(MaintenanceMode $maintenanceMode)
     {
         try {
-            if (! is_file(storage_path('framework/down'))) {
+            if ($maintenanceMode->isUp()) {
```
driesvints, line 44:
```suggestion
            if ($this->laravel->maintenanceMode()->isUp()) {
```

src/Illuminate/Foundation/Console/UpCommand.php
```diff
@@ -34,18 +35,19 @@ class UpCommand extends Command
...
                 $this->comment('Application is already up.');
 
                 return 0;
             }
 
-            unlink(storage_path('framework/down'));
+            $maintenanceMode->up();
```
driesvints, line 50:
```suggestion
            $this->laravel->maintenanceMode()->up();
```

driesvints (changes requested):

src/Illuminate/Foundation/Console/DownCommand.php
wimulkeman: @driesvints You were one step ahead of me :). I've replace the DI usage with the application method. That begins to beg the question, now the arguments have been removed from the Middleware and the commands, should this PR be retarget to the 8.x branch instead of the master branch?

wimulkeman:
Because of the changes made the PR no longer introduces new method arguments to existing code. Should this PR shill be pointed at master (9.x) or instead to the 8.x branch?

src/Illuminate/Foundation/MaintenanceMode.php
```diff
@@ -0,0 +1,76 @@
...
+     *
+     * @return void
+     */
+    public function up(): void
+    {
+        if ($this->isDown() === false) {
+            return;
+        }
+
+        unlink($this->getDownFilePath());
```
driesvints, line 54:
```suggestion
        if ($this->isDown() !== false) {
            unlink($this->getDownFilePath());
        }
```

src/Illuminate/Foundation/MaintenanceMode.php
```diff
@@ -0,0 +1,76 @@
...
+    {
+        return json_decode(file_get_contents($this->getDownFilePath()), true);
+    }
+
+    /**
+     * Get the path where the file is stored that signals that the application is down for maintenance.
+     *
+     * @return string
+     */
+    private function getDownFilePath(): string
```
driesvints, line 72:
```suggestion
    protected function getDownFilePath(): string
```

driesvints (changes requested):

driesvints:
> Because of the changes made the PR no longer introduces new method arguments to existing code. Should this PR shill be pointed at master (9.x) or instead to the 8.x branch?

@wimulkeman you're totally correct. Feel free to re-send this to 8.x, thanks! Good job on the PR so far 👍

One thing that can very much help this PR get merged is explain the use case a bit more thoroughly in your PR description. So state clearly why this is needed (multiple servers but also different stores). 

Thanks!

seriquynh:
This PR no longer introduces new method arguments to existing code but it adds a new method into `Application` contract. So, it's still a breaking-change PR and should be sent to 9.x.

src/Illuminate/Foundation/MaintenanceMode.php
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    /**
+     * Take the application out of maintenance.
+     *
+     * @return void
+     */
+    public function up(): void
+    {
+        if ($this->isDown() !== false) {
```
seriquynh, line 50: `isDown` method always returns a bool value. Why don't you use just `$this->isDown()`?

```php
    public function up(): void
    {
        if ($this->isDown()) {
            unlink($this->getDownFilePath());
        }
    }
```

driesvints: Ah yeah my bad, I said IsUp first but it should indeed be isDown.

wimulkeman: Good point. Changed it in the code

driesvints (approved):
I like the PR. I'd maybe even make some proxy methods on `Application` for `isUp`, `isDown`, `up` and `down` to make the code even cleaner but let's see what @taylorotwell says.

ziadoz:
This looks like really useful feature. Would it be worth adding a contract/interface for the `MaintenanceMode` class itself?

src/Illuminate/Foundation/Application.php
```diff
@@ -1105,14 +1105,24 @@ public function addAbsoluteCachePathPrefix($prefix)
...
     }
 
+    /**
+     * Get the instance of the MaintenanceMode.
+     *
+     * @return \Illuminate\Contracts\Foundation\MaintenanceMode
+     */
+    public function maintenanceMode()
+    {
+        return $this->make(MaintenanceMode::class);
```
driesvints, line 1115: Now that you have a contract, it's probably better to resolve that. Otherwise you won't be able to overwrite the concrete implementation. We'll need to bind the contract to the container.

wimulkeman: Check. I've now bind the contract to the default implementation in the src/Illuminate/Foundation/Providers/FoundationServiceProvider.php. I was not totally sure if thats the correct place to do it.

src/Illuminate/Foundation/Providers/FoundationServiceProvider.php
```diff
@@ -94,6 +97,16 @@ public function registerRequestSignatureValidation()
         });
     }
 
+    /**
+     * Register de MaintenanceMode service through a contract.
```
axlon, line 101:
```suggestion
     * Register the MaintenanceMode service through a contract.
```

taylorotwell:
Can you allow me to push to this branch so I can update the PR? Allow edits from maintainers.

wimulkeman:
> taylorotwell

I've given you permission to the repository. I couldn't find the setting on this PR page, so I gave you the needed permissions instead on the forked repo. Is that sufficient?

taylorotwell:

driesvints:
Thanks for your PR @wimulkeman 👍

PR #40072, opened 2021-12-16
[8.x] move maintenance mode logic

wimulkeman:

Overwriting the logic may be needed if for example the application is run on multiple servers. In this case the developer would want to extend the storage logic for the maintenance mode to implement a solution that can change the maintenance state for all the servers at once and reverse the maintenance mode with one command.

By separating the maintenance mode logic the framework opens the door for packages that could handle the maintenance mode through different mechanisms like using the cache (as proposed in a blog https://www.raypold.com/posts/laravel-distributed-maintenance-mode/).

The need for this was shown within the discussions https://github.com/laravel/ideas/issues/1944 and https://github.com/laravel/framework/discussions/36474.

Fixes #36474

This PR is a follow-up to the PR #40070 with suggested changes and pointed to the master instead of the 8.x branch because it contains backward incompatible changes.

### DIFF
--- a/src/Illuminate/Contracts/Foundation/Application.php
+++ b/src/Illuminate/Contracts/Foundation/Application.php
@@ -83,6 +83,13 @@ interface Application extends Container
     public function runningUnitTests();
 
     /**
+     * Get the instance of the MaintenanceMode.
+     *
+     * @return \Illuminate\Foundation\MaintenanceMode
+     */
+    public function maintenanceMode();
+
+    /**
      * Determine if the application is currently down for maintenance.
      *
      * @return bool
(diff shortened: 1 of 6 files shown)


src/Illuminate/Contracts/Foundation/Application.php
```diff
@@ -82,6 +82,13 @@ public function runningInConsole();
      */
     public function runningUnitTests();
 
+    /**
+     * Get the instance of the MaintenanceMode.
+     *
+     * @return \Illuminate\Foundation\MaintenanceMode
+     */
+    public function maintenanceMode();
```
driesvints, line 90: Forgot to mention but you'll need to remove this change since this is still a breaking change.

wimulkeman: I would suggest in that case to leave the PR to 9.x open instead? Introducing the method without updating the interface feels invalid to me.

driesvints: Yeah that's okay 👍

PR #36885, opened 2021-04-06
[8.x] Fixes array to string conversion exception when using custom validation messages for size rules

mateusjunges:
This PR fixes a Array to string conversion exception when using custom validation messages for size rules.

When using laravel validation, you can specify custom validation messages for attributes using the convention "attribute.rule" to name the lines:
```php
'custom' => [
     'attribute-name' => [
        'rule-name' => 'custom-message',
    ],
],
```
The problem is: when using size rules, such as `Between`, if you define your custom message as something like this (i used the `email` field for this test:
```php
 'custom' => [
   'email' => [
        'between' => [
            'numeric' => 'The :attribute must be between :min and :max.',
            'file' => 'The :attribute must be between :min and :max kilobytes.',
            'string' => 'Some custom message here',
            'array' => 'The :attribute must have between :min and :max items.',
        ]
    ]
],
```
it will cause an exception:
image

This occurs because the actual custom message for size rules is returned before checking if the rule being validated is a size rule or not. So, the message passed to the `Illuminate\Support\MessageBag` `transform` method is this one:

```
array:1 [▼
  0 => array:4 [▼
    "numeric" => "The email must be between 20 and 24."
    "file" => "The email must be between 20 and 24 kilobytes."
    "string" => "Some custom message here."
    "array" => "The email must have between 20 and 24 items."
  ]
]
```

This PR fixes this error by adding a check for size rules before returning the custom validation message to the MessageBag class, and, if it actually is a size rule, get the message for the correct attribute type:

```diff
    if ($customMessage !== $customKey) {
+     if (in_array($rule, $this->sizeRules)) {         
+         return $this->getCustomSizeMessage($attribute, $rule);   
+     }
    return $customMessage;
    }

+ protected function getCustomSizeMessage($attribute, $rule)
+ {
+    $lowerRule = Str::snake($rule);
+    $type = $this->getAttributeType($attribute);
+    $key = "validation.custom.{$attribute}.{$lowerRule}.{$type}";
+    return $this->translator->get($key);
+ }
```
With this fix, the validation returns the correct message, which, in this case is `Some custom message here`.

PR #40074, opened 2021-12-16
[9.x] Fix array to string conversion when using custom validation messages for size rules

mateusjunges:
## Description
This PR is a follow up on #36885. I've tried ealier this year, but i think i haven't described correctly the issues or used wrong examples and it was rejected, so i'll try it again. When using custom validation messages for size rules, if you define your custom validation messages in `resources/lang/en/validation.php, then laravel thrown an exception when accessing those messages:

image

This exception comes from the following line, while replacing the `:message`:
https://github.com/laravel/framework/blob/96d2b48e45b9d113b95ce0edf542b3ffc76c5176/src/Illuminate/Support/MessageBag.php#L268

I'm sending this to master because if some user currently uses the custom validation messages, then the response he gets when using `$validator->errors()` is the following:
```json
{
    "file":[
        {
           "file":"The file must be between 20 and 50 kilobytes.",
           "string":"Some custom message here"
        }
    ]
}
```
and this PR changes it to:
```json
{
    "file":[
        "Some custom message here"
    ]
}
```

So, this changes the way you access validation errors from `$validator->errors()->toArray()['file'][0]['string']` to `$validator->errors()->toArray()['file'][0]` or `$validator->errors()->first('file)`, which is a breaking change.

## Steps to reproduce
- Define custom validation messages in `resources/lang/en/validation.php`:
```php
'custom' => [
    'file' => [
        'between' => [
            'numeric' => 'The :attribute must be between :min and :max.',
            'file' => 'The :attribute must be between :min and :max kilobytes.',
            'string' => 'Some custom message here',
            'array' => 'The :attribute must have between :min and :max items.',
        ],
    ],
],
```

- Create a test route in `web.php` to test the validation message:
```php
Route::get('/test', function() {
   return Validator::make(['file' => 'filename.png'], ['file' => 'between:20,50'])->errors()->first('file');
});
```

- Go to `localhost:8000/test`

### DIFF
--- a/src/Illuminate/Validation/Concerns/FormatsMessages.php
+++ b/src/Illuminate/Validation/Concerns/FormatsMessages.php
@@ -31,9 +31,10 @@ trait FormatsMessages
 
         $lowerRule = Str::snake($rule);
 
-        $customKey = in_array($rule, $this->sizeRules)
-            ? "validation.custom.{$attribute}.{$lowerRule}.{$this->getAttributeType($attribute)}"
-            : "validation.custom.{$attribute}.{$lowerRule}";
+//        $customKey = in_array($rule, $this->sizeRules)
+//            ? "validation.custom.{$attribute}.{$lowerRule}.{$this->getAttributeType($attribute)}"
+//            : "validation.custom.{$attribute}.{$lowerRule}";
+            $customKey = "validation.custom.{$attribute}.{$lowerRule}";
 
         $customMessage = $this->getCustomMessageFromTranslator($customKey);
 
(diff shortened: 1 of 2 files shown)


mateusjunges:
Here https://github.com/laravel/framework/pull/36885#issuecomment-814896474 taylor said that "an email is never a file or a numeric. It's basically string". I agree and i really should have described it better. But, with the `file` attribute, you could have two or more places in your project that uses `file` sometimes as `string` (filename) and sometimes as a `file` (uploaded file). So, if you use custom validation messages, then you get the array to string conversion. That's why i'm submitting this PR again.

PR #40075, opened 2021-12-16
[8.x] Passthru Eloquent\Query::explain function to Query\Builder:explain for the ability to use database-specific explain commands

tpetry:
After a [twitter discussion](https://twitter.com/mattkingshott/status/1471117843722903553?s=20) about laravel's ability to explain database queries and the insight that many optional features are missing from the explain output for PostgreSQL, i wanted to add them to [tpetry/laravel-postgresql-enhanced](https://github.com/tpetry/laravel-postgresql-enhanced).

I then discovered that the eloquent builder _and_ the query builder are using the same trait for generating explain information. As the eloquent builder is already passing many database query specific functions to the query builder i extended it to the explain command.

By this change database connections can use database-specific explain commands when they are using a custom query builder. And there is no need to change the model to make use of it. The enhanced laravel postgresql driver will use it to include many more optional explain options.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -7,7 +7,6 @@ use Closure;
 use Exception;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Database\Concerns\BuildsQueries;
-use Illuminate\Database\Concerns\ExplainsQueries;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Query\Builder as QueryBuilder;
@@ -26,7 +25,7 @@ use ReflectionMethod;
  */
 class Builder
 {
-    use Concerns\QueriesRelationships, ExplainsQueries, ForwardsCalls;
+    use Concerns\QueriesRelationships, ForwardsCalls;
     use BuildsQueries {
         sole as baseSole;
     }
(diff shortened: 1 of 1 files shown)


tpetry:
I have my implementation for a PostgreSQL-specific query plan finished. [The difference is remarkable](https://github.com/tpetry/laravel-postgresql-enhanced#explain).

Laravel Query-Plan:
```php
DB::table('migrations')->where('batch', 1)->explain(analyze:true)->dd();

// Output:
//array:2 [
//  0 => {#882
//    +"QUERY PLAN": "Seq Scan on migrations  (cost=0.00..11.75 rows=1 width=524)"
//  }
//  1 => {#881
//    +"QUERY PLAN": "  Filter: (batch = 1)"
//  }
//]
```

My Query-Plan:
```php
DB::table('migrations')->where('batch', 1)->explain(analyze:true)->dd();

// Output:
// array:1 [
//  0 => """
//    Seq Scan on public.migrations  (cost=0.00..11.75 rows=1 width=524) (actual time=0.014..0.031 rows=1 loops=1)\n
//      Output: id, migration, batch\n
//      Filter: (migrations.batch = 1)\n
//      Buffers: shared hit=1\n
//    Settings: search_path = 'public'\n
//    Planning:\n
//      Buffers: shared hit=61\n
//    Planning Time: 0.282 ms\n
//    Execution Time: 0.100 ms
//    """
//]
```

With the current Laravel implementation, the new query plan is activated as soon as my enhanced driver is installed. But there's currently no way doing this as well for Eloquent queries without adding a trait to every single model. With my PR the eloquent `explain()` method will simply forward to the query builder `explain()` method as the eloquent method is not doing anything different anyways. By this any custom database connection driver can implement very specific functionality.

mstaack:
👍🏼

Issue #40055, opened 2021-12-15
HTTP client automatically throws exception after retrying


mattkingshott:
- Laravel Version: 8.75.0
- PHP Version: 8.1.0
- Database Driver & Version: MySQL 8.0.27

### Description:

When a retry is specified within the HTTP client, and the retry limit is reached, then the response is automatically thrown. Since there is no way to hook into the `throw` method, you'll always get a `RequestException`. Therefore, you cannot do things like `$response->failed()`

The problem lies within `Illuminate\Http\Client\PendingRequest`, on lines 682-684:

```php
if ($this->tries > 1 && ! $response->successful()) {
    $response->throw();
}
```

### Steps To Reproduce:

Simply, add a retry configuration to the HTTP client:

```php
Http::fake([
    '*' => Http::response(['error'], 403),
]);

$response = Http::retry(3, 5000)->post('https://example.com/route'); 

// The above line throws a RequestException, so the below error handling will not be reached

if ($response->failed()) {
    // error handling
}
```

driesvints:
Hey @mattkingshott, this is the expected behaviour. Please see the docs here:

![Screenshot 2021-12-16 at 14 28 39](https://user-images.githubusercontent.com/594614/146380700-f3b62baa-b45d-4b3e-9aab-d5343e5b2f0a.png)

If you want to see a different behaviour then please attempt a PR. Thanks

mattkingshott:
Oh my. Sorry I missed that. Maybe will attempt a PR for this.

PR #40076, opened 2021-12-16
[9.x] Allow HTTP client requests with retries to optionally throw RequestExceptions

mattkingshott:
This PR addresses #40055 by adding functionality to the HTTP client so that you can prevent it from throwing a `RequestException` when `retries` are configured.

## The problem

Presently, and as highlighted in the docs, the HTTP client does not throw exceptions when a request fails. As a result, you can make use of methods like `failed` on the response e.g.

```php
$response = Http::get('http://example.com');

if ($response->failed())
{
    // Handle the error
}
```

However, if you configure `retries` on the HTTP client, it will throw a `RequestException` if all attempts fail. 

```php
$response = Http::retries(3, 1000)->get('http://example.com');
```

This means:

1. You cannot configure `throw()` on the client.
2. You cannot chain methods e.g. `throw()->json()`.
3. You cannot use `throwIf`.
4. You cannot use the response's error handling methods e.g. `$response->failed()`.

In addition, you have to wrap the request within a try/catch block if you want to handle the error. 

## The solution

The PR adds a fourth parameter `throw` to the `retries` method, which allows you to control whether the HTTP client should throw an exception when the retry limit is reached. 

By default, this parameter is set to `true` to preserve backward compatibility.

```php
Http::retries(3, 1000); // throws a RequestException if all retries fail
Http::retries(3, 1000, null, true); // throws a RequestException if all retries fail
Http::retries(3, 1000, null, false); // does not throw a RequestException if all retries fail
```

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -99,6 +99,13 @@ class PendingRequest
     protected $retryDelay = 100;
 
     /**
+     * Whether to throw an exception when all retries fail.
+     *
+     * @var bool
+     */
+    protected $retryThrow = true;    
+    
+    /**
      * The callback that will determine if the request should be retried.
      *
      * @var callable|null
@@ -451,12 +458,14 @@ class PendingRequest
      * @param  int  $times
      * @param  int  $sleep
      * @param  callable|null  $when
+     * @param  bool  $throw
      * @return $this
      */
-    public function retry(int $times, int $sleep = 0, ?callable $when = null)
+    public function retry(int $times, int $sleep = 0, ?callable $when = null, bool $throw = true)
     {
         $this->tries = $times;
         $this->retryDelay = $sleep;
+        $this->retryThrow = $throw;
         $this->retryWhenCallback = $when;
 
         return $this;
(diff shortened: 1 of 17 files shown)


src/Illuminate/Http/Client/PendingRequest.php
```diff
@@ -451,12 +458,14 @@ public function timeout(int $seconds)
      * @param  int  $times
      * @param  int  $sleep
      * @param  callable|null  $when
+     * @param  bool  $throw
      * @return $this
      */
-    public function retry(int $times, int $sleep = 0, ?callable $when = null)
+    public function retry(int $times, int $sleep = 0, ?callable $when = null, bool $throw = true)
```
driesvints, line 464: We can't change the method signature on a stable release unfortunately.

mattkingshott: Should I submit this to master then?

driesvints: Yeah, if we can't solve it differently then that would be best.

mattkingshott: Okay. I’ll submit it for master.

PR #40077, opened 2021-12-16
[8.x] Consistently negate conditions in @includeUnless() Blade directive

markjaquith:
https://twitter.com/gonedark/status/1471514212363079681
CleanShot 2021-12-16 at 14 05 28@2x

Before this change, `@includeUnless($undefined ?? true, 'template')` would result in a PHP Warning and an unexpected condition result, because `$undefined ?? true` would be compiled as `! $undefined ?? true`, which due to `!` having precedence over `??` is different than `! ($undefined ?? true)` (which is how `@unless` would compile it).

```
>>> ! $undefined ?? true
<warning>PHP Warning:  Undefined variable $undefined in eval()'d code on line 1</warning>
=> true
>>> ! ($undefined ?? true)
=> false
```

This change necessitated the creation of `renderUnless()`, because `@includeUnless()` accepts multiple arguments, and parsing out the first argument to a Blade directive is fraught. So `renderUnless()` takes the parameters as-is, and negates the condition inside the method.

`@includeUnless()` also lacked unit tests, which I added.

This change makes `@includeUnless()` work more like `@unless()`, which already correctly wraps the condition in parentheses (or rather, does not remove the parentheses that are already there).

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesIncludes.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesIncludes.php
@@ -64,7 +64,7 @@ trait CompilesIncludes
     {
         $expression = $this->stripParentheses($expression);
 
-        return "<?php echo \$__env->renderWhen(! $expression, \Illuminate\Support\Arr::except(get_defined_vars(), ['__data', '__path'])); ?>";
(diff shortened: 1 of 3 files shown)


taylorotwell:

jasonmccreary:
@markjaquith, nice work!

PR #40078, opened 2021-12-16
[8.x] Adds `maskBetween()`, mask a string between two strings.

DarkGhostHunter:
## What?

Adds a way to mask a portion of a string between two given strings, without knowing its length.

```php
$text = 'The private address is Almond 259, Texas.';

Str::maskBetween($text, '*', 'address is ', '.'); 

// 'The private address is ******************.'
```

## Why?

Because without this you will need to find the start and end index manually, which is cumbersome when you don't know the mask-able text length, but you know or how is contained.

```php
$text = 'The private address is Almond 259, Texas.';

$start = strpos($text, 'address is ');

Str::maskBetween($text, '*', $start  + strlen('address is'), strlen($text) - 1;
```

## BC?

Additive. PR does not expects users adding `maskBetween` as a Macro to `Str` and `Stringable`.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -431,6 +431,29 @@ class Str
     }
 
     /**
+     * Masks a portion of a string between two given strings.
+     *
+     * @param  string  $string
+     * @param  string  $character
+     * @param  string  $start
+     * @param  string  $end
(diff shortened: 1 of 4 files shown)


DarkGhostHunter:
I hate vscode.

PR #40083, opened 2021-12-17
[8.x] Make it possible to use prefixes on Predis per Connection

tben:
This PR allows the ability to use prefixes on a per connection basis for Predis.

Basically the only way to setup a prefix is by doing the following `['redis' => ['options' => ['prefix' => 'prefix:']]]` in the  `/config/database.php` file.

So copying a similar setup in PHPRedis. Each connection can be given a unique prefix by defining them within the individual  connection config like so  `['redis' => ['connection1' => ['prefix' => 'prefix:'], 'connection2' => ['prefix' => 'prefix2:']]]`.

### DIFF
--- a/src/Illuminate/Redis/Connectors/PredisConnector.php
+++ b/src/Illuminate/Redis/Connectors/PredisConnector.php
@@ -22,6 +22,10 @@ class PredisConnector implements Connector
         $formattedOptions = array_merge(
             ['timeout' => 10.0], $options, Arr::pull($config, 'options', [])
         );
+        
+        if (! empty($config['prefix'])) {
+            $formattedOptions['prefix'] = $config['prefix'];
+        }
 
         return new PredisConnection(new Client($config, $formattedOptions));
     }
--- a/src/Illuminate/Redis/Connectors/PredisConnector.php
+++ b/src/Illuminate/Redis/Connectors/PredisConnector.php
@@ -23,7 +23,7 @@ class PredisConnector implements Connector
             ['timeout' => 10.0], $options, Arr::pull($config, 'options', [])
         );
 
-        if (! empty($config['prefix'])) {
+        if (isset($config['prefix'])) {
             $formattedOptions['prefix'] = $config['prefix'];
         }
 
@@ -42,7 +42,7 @@ class PredisConnector implements Connector
     {
         $clusterSpecificOptions = Arr::pull($config, 'options', []);
 
-        if (! empty($config['prefix'])) {
+        if (isset($config['prefix'])) {
             $clusterSpecificOptions['prefix'] = $config['prefix'];
         }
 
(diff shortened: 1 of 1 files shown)


src/Illuminate/Redis/Connectors/PredisConnector.php
```diff
@@ -42,6 +42,10 @@ public function connectToCluster(array $config, array $clusterOptions, array $op
     {
         $clusterSpecificOptions = Arr::pull($config, 'options', []);
 
+        if (! empty($config['prefix'])) {
```
GrahamCampbell, line 45: empty is not the right function here. it will not permit using 0 as a prefix.

GrahamCampbell (changes requested):

src/Illuminate/Redis/Connectors/PredisConnector.php
tben: I simply copied the code from https://github.com/laravel/framework/blob/a049279178cfea6d8e714dcbb09429b5b78603c9/src/Illuminate/Redis/Connectors/PhpRedisConnector.php#L94

Do we want to also fix the code within PHPRedis as well?

taylorotwell:
Please share what the actual formatted Redis configuration would look like using these new options. Please mark as ready for review when you have done so.

jasonvarga:
We were just looking into this and noticed it was already possible to put a prefix on a specific store. You just had to nest it under an `options` array. This works for both phpredis  and predis clients.

```php
'redis' => [
    'cache' => [
        'host' => env('REDIS_HOST', '127.0.0.1'),
        'password' => env('REDIS_PASSWORD', null),
        'port' => env('REDIS_PORT', 6379),
        'database' => env('REDIS_CACHE_DB', 1),
        
        'options' => [ // 👈 you needed this
            'prefix' => env('REDIS_PREFIX', 'prefix2:'),
        ]
    ],
],
```

So now this PR introduces an inconsistency. You can now add un-nested `prefix` keys to predis stores, but not phpredis.

tben:
It's been a while since I've look at the code but the change was centered around how predis client code expected the prefix to be presented. Atleast when I committed the code there was no way to pass this information to the predis code in a way that would allow it to function. It's possible that code changes now allow for it.

jasonvarga:
In the few lines above what was added in this PR, you can see that the `options` were already being merged in:

```php
$formattedOptions = array_merge(
    ['timeout' => 10.0], $options, Arr::pull($config, 'options', [])
);
```

jasonvarga:
The first argument passed into `connect()` would be the store's config, which could have `options` in it.

tben:
@jasonvarga Okay, I just checked on my local machine and you were correct. Yeah, that's really is my bad. :( I swear down that options didn't work for me when I wrote this fix.

PR #40084, opened 2021-12-17
[9.x] add confirmation to clear:cache command

ankurk91:
Hi,

Cache is kind of database and when cleared accidentally may create problems in production.

Imagine you have cached the API response to Redis and it gets cleared, now application has to rebuild the cache, which will slow down the application until cache is rebuild.

This is a breaking change for many, since they need to add `--force` flag to the command now.

### DIFF
--- a/src/Illuminate/Cache/Console/ClearCommand.php
+++ b/src/Illuminate/Cache/Console/ClearCommand.php
@@ -4,12 +4,15 @@ namespace Illuminate\Cache\Console;
 
 use Illuminate\Cache\CacheManager;
 use Illuminate\Console\Command;
+use Illuminate\Console\ConfirmableTrait;
 use Illuminate\Filesystem\Filesystem;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputOption;
 
 class ClearCommand extends Command
 {
+    use ConfirmableTrait;
+
     /**
      * The console command name.
      *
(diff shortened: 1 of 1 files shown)


driesvints:
> Cache is kind of database and when cleared accidentally may create problems in production.

A cache store should never be a database and clearing cache should never cause issues for an app other than re-fetching cached data and re-storing it.

mfn:
This is correct, but for example the redis cache drive uses `FLUSHDB` which wipes the entire database; should one share it with other operations, data maybe be accidentally removed.

There's a note in the manual warning about this, but it's really not so obvious because e.g. a cache prefix is supported and one may think only the prefixed keys are removed.

driesvints:
> should one share it with other operations, data maybe be accidentally removed.

You should never use the same connection for a cache and a database store.

ankurk91:
Laravel stores many other things in cache, for example the API and the Queue Jobs throttling.

When a user enter wrong password 5 times, Laravel blocks the user from login, but if we clear the cache, user is allowed to login again.
 
 Laravel may also user cache for object locks, and clearing the cache may result in unexpected behavior in a LIVE app.

There are php packages which are using cache , for example:
https://github.com/tzsk/otp

driesvints:
> When a user enter wrong password 5 times, Laravel blocks the user from login, but if we clear the cache, user is allowed to login again.

That's a really big edge case. I would have to happen around the same time that you're clearing the cache. Also, you're not clearing your cache store all the time as that would defeat the purpose of a cache store.

> Laravel may also user cache for object locks, and clearing the cache may result in unexpected behavior in a LIVE app.

That should never be an issue. Like I said before: clearing cache should never cause issues in a live app.

ankurk91:
Okay

PR #40085, opened 2021-12-17
[9.x] Add confirmation to model:prune command

ankurk91:
Hi,

Lets add confirmation to model:prune command since it is a destructive command.

Users need to add `--force` flag to command in production.

thanks

### DIFF
--- a/src/Illuminate/Database/Console/PruneCommand.php
+++ b/src/Illuminate/Database/Console/PruneCommand.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Database\Console;
 
 use Illuminate\Console\Command;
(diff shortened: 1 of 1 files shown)


JamesFreeman:
This would be a breaking change and would have to go into 9.x?

driesvints:
@JamesFreeman this is targeting master already.

driesvints:
@ankurk91 please fix the tests here first. Thanks

JamesFreeman:
> @JamesFreeman this is targeting master already.

Doh! My bad.

ankurk91:
looks like that too much work,  i can live without this feature so closing

PR #40087, opened 2021-12-17
[9.x] PSR Log v3

driesvints:
This PR bumps `psr/log` to v3. Because of the introduction of type-hints for the method arguments we'll need to explicitly add the types that we support as well to widen the method signature.


src/Illuminate/Log/Logger.php
```diff
@@ -262,7 +262,7 @@ protected function formatMessage($message)
             return var_export($message->toArray(), true);
         }
 
-        return $message;
+        return (string) $message;
```
driesvints, line 265: I added this because the `fireLogEvent` always expects the `$message` to be a string.

GrahamCampbell:
It is not necessary to add the typehints. We can still support v3 without them.

driesvints:
@GrahamCampbell can you pr?

PR #40090, opened 2021-12-17
[9.x] Add setTimeout for Queueable jobs

jmarcher:
Add new method fluent setTimeout()

Sometimes we need to set individual timeouts for our jobs (which is currently supported via public $timeout property which gets serialized) but it gets tedious and not so pretty when dispatching jobs like this:

Before:
```php
$job = (new SomeJob)
    ->onQueue('super-queue')
    ->onConnection('redis')
    ->delay(now()->addDay());
$job->timeout = 1000;
dispatch($job);
```

After:
```php
dispatch(
    (new SomeJob)
        ->onQueue('super-queue')
        ->onConnection('redis')
        ->delay(now()->addDay())
        ->setTimeout(1000)
);
```

### DIFF
--- a/src/Illuminate/Bus/Queueable.php
+++ b/src/Illuminate/Bus/Queueable.php
@@ -52,7 +52,7 @@ trait Queueable
     public $delay;
 
     /**
-     * The number of seconds the job can run
+     * The number of seconds the job can run.
      *
      * @var int|null
      */
--- a/src/Illuminate/Bus/Queueable.php
+++ b/src/Illuminate/Bus/Queueable.php
@@ -52,6 +52,13 @@ trait Queueable
     public $delay;
 
     /**
+     * The number of seconds the job can run
+     *
+     * @var int|null
+     */
+    public $timeout;
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
Is there a specific reason you chose `setTimeout` instead of just `timeout`?

jmarcher:
@taylorotwell For 2 reasons more or less:

Because we use `onQueue()`, `onConnection()`, etc. instead of `queue()`, `connection()` which one may say it is a "similar" notation

And the second one is that the abstract `Illuminate\Queue\Jobs\Job` has a `timeout()` but it is a getter of the timeout inside the job's payload.

taylorotwell:
I wonder if this should be on `PendingDispatch` as well? I think I may hold off on this because I would usually dispatch jobs like `Foo::dispatch` not using the `dispatch` helper. In addition, I'm not totally sold on the syntax and it's already a solvable problem by passing the `timeout` you want to your job's constructor and setting the property that way.

PR #40091, opened 2021-12-17
[9.x] add search method to eloquent

abbasudo:
Add new `search()` method to eloquent.
Sometimes we need to search a column for a specific text . insted of using raw query with `LIKE` , i added `search` method so you can use it like this: 
```php
Post::search('title' , 'something')->get();
```

i have added a test for search and i will make changes to the document in case you accepted it.
thanks for your time and attention.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -248,6 +248,20 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Add a basic like clause to the query.
+     *
+     * @param  \Closure|string|array|\Illuminate\Database\Query\Expression  $column
+     * @param  mixed  $value
+     * @return $this
+     */
+    public function search($column, $value)
+    {
+        $this->where($column, 'LIKE', "%{$value}%");
+
+        return $this;
+    }
+
+    /**
      * Add a basic where clause to the query, and return the first result.
      *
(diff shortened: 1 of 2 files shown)


driesvints:
This will conflict with Scout.

llabbasmkhll:
@driesvints would you suggest me another method name. dose `including()` fits?

src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -247,6 +247,20 @@ public function where($column, $operator = null, $value = null, $boolean = 'and'
...
+    /**
+     * Add a basic like clause to the query.
+     *
+     * @param  \Closure|string|array|\Illuminate\Database\Query\Expression  $column
+     * @param  mixed  $value
+     * @return $this
+     */
+    public function search($column, $value)
+    {
+        $this->where($column, 'LIKE', "%{$value}%");
```
crynobone, line 259: Not all database support `LIKE`

driesvints:
@llabbasmkhll I don't think we should merge this. You can add this syntax to your own app if you want.

llabbasmkhll:
@driesvints Ok cool . thanks for reviewing.

taylorotwell:

SocolaDaiCa:
@llabbasmkhll how to search special character https://stackoverflow.com/questions/712580/list-of-special-characters-for-sql-like-clause

PR #40092, opened 2021-12-17
Added message to status assertions

yormy:
Adding a status message to the assert response.
Especially if you want to add a status assertion in a loop you can pass additional information to your console about what issue or dataset actually triggered the assertion. This saves alot of time debugging and finding out what data item triggered the assertion

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -92,32 +92,35 @@ class TestResponse implements ArrayAccess
     /**
      * Assert that the response has a 200 status code.
      *
+     * @param string|null $message
      * @return $this
      */
-    public function assertOk()
+    public function assertOk(string $message = null)
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you show me how it saves you time?

yormy:
Sure, one of the thing I do is to run though a list of routes to see if they all get a 200 back. If one fails in the loop, I have no clue which one as it breaks on the assertion line. Echoing the route in the message and I know exactly what route fails

SjorsO:
> ...is to run though a list of routes to see if they all get a 200 back...

Laravel doesn't support making multiple HTTP calls in a single test (even though it works fine most of the time).

A better solution would be to use a phpunit dataprovider. If one of those tests fails phpunit will print the arguments the test was given.

taylorotwell:
Yes - see PHPUnit data providers as mentioned above.

PR #40098, opened 2021-12-18
[8.x] Add rule to validate MAC address

bilalalmassry:
add new validation rule "mac_address" to **ensure** that the attribute is a valid Mac Address

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1231,6 +1231,18 @@ trait ValidatesAttributes
     }
 
     /**
+     * Validate that an attribute is a valid MAC address.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @return bool
+     */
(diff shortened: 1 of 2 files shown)


bilalalmassry:
I just wondering, should i open a new PR in laravel project repo to add the validation message?

src/Illuminate/Validation/Concerns/ValidatesAttributes.php
```diff
@@ -1230,6 +1230,18 @@ public function validateIpv6($attribute, $value)
...
     }
 
+    /**
+     * Validate that an attribute is a valid MAC address.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @return bool
+     */
+    public function validateMacAddress($attribute, $value)
```
dennisprudlo, line 1240: Maybe name the validation rule just "mac" to be consistent with "email" and "ip" for example?

bilalalmassry: I don't think it will **understandable**, **clear** enough, it could be consistent with "accepted_if", "require_if"

PR #40103, opened 2021-12-18
[9.x] Posibility to calculate backoff value from total of exceptions, or from total of attempts

MartinPham:
#39976 

This PR ~~makes the Worker to prefer to calculate the backoff value from total of exceptions (when the backoff is configured as an array), and fallback to the total of attempts.~~ allows us to configure how to calculate backoff value, based on total of attempts (default), or on total of exceptions.

## Situation
- A Job is configured with backoff as an array `[1, 10]`

### Current behavior
- 1st attempt, the job calls `release()`.
- (After `0` seconds) 2nd attempt, the job throws an exception.
- (After `10` seconds) 3rd attempt, the job throws an exception.
- (After `10` seconds) 4th attempt, the job throws an exception.
...
Notice that on the 2nd attempt, the backoff was `10` seconds instead of `1`

### Proposed solution
1) Introducing `$job->backoffMode`: 
- `Illuminate\Queue\Jobs\Job::BACKOFF_FROM_ATTEMPTS` : (Default) Calculate backoff based on total of attempts
- `Illuminate\Queue\Jobs\Job::BACKOFF_FROM_EXCEPTIONS `: Calculate backoff based on total of exceptions

**Example**
```php
<?php

namespace App\Jobs;

use Exception;
use Illuminate\Bus\Queueable;
use Illuminate\Queue\Jobs\Job;

class ExampleJob implements ShouldQueue
{

    public $tries = 10;
    public $maxExceptions = 5;
    
    // Define as class variable
    // public $backoffMode = Job::BACKOFF_FROM_EXCEPTIONS;

    public function __construct()
    {
    }

    public function backoff()
    {
        return [1, 5];
    }

    // Define as class method
    public function backoffMode()
    {
        return Job::BACKOFF_FROM_EXCEPTIONS;
    }

    public function handle()
    {
        if ($this->attempts() === 1) {
            $this->release();
        } else {
            throw new Exception();
        }
    }
}

```

2) New behavior (BACKOFF_FROM_EXCEPTIONS)
- 1st attempt, the job calls `release()`.
- (After `0` seconds) 2nd attempt, the job throws an exception.
- (After `1` seconds) 3rd attempt, the job throws an exception.
- (After `10` seconds) 4th attempt, the job throws an exception.
...
Notice that on the 2nd attempt, the backoff was `1` seconds, follows the backoff array

### DIFF
--- a/src/Illuminate/Queue/Jobs/Job.php
+++ b/src/Illuminate/Queue/Jobs/Job.php
@@ -285,7 +285,6 @@ abstract class Job
         return $this->payload()['backoff'] ?? $this->payload()['delay'] ?? null;
     }
 
-
     /**
      * @const Backoff Calculation Mode: From total of "attempts" (default) o "exceptions"
      */
@@ -293,7 +292,7 @@ abstract class Job
     const BACKOFF_FROM_EXCEPTIONS = 'exceptions';
 
     /**
-     * Decide how to calculate the backoff (default: "attempts" - same behavior as before)
+     * Decide how to calculate the backoff (default: "attempts" - same behavior as before).
      *
      * @return string
      */
--- a/src/Illuminate/Queue/Jobs/Job.php
+++ b/src/Illuminate/Queue/Jobs/Job.php
@@ -12,6 +12,12 @@ abstract class Job
     use InteractsWithTime;
 
     /**
+     * @const Backoff Calculation Mode: From total of "attempts" (default) o "exceptions"
+     */
+    const BACKOFF_FROM_ATTEMPTS = 'attempts';
+    const BACKOFF_FROM_EXCEPTIONS = 'exceptions';
+
+    /**
      * The job handler instance.
      *
      * @var mixed
@@ -286,19 +292,13 @@ abstract class Job
     }
 
     /**
-     * @const Backoff Calculation Mode: From total of "attempts" (default) o "exceptions"
-     */
-    const BACKOFF_FROM_ATTEMPTS = 'attempts';
-    const BACKOFF_FROM_EXCEPTIONS = 'exceptions';
-
-    /**
      * Decide how to calculate the backoff (default: "attempts" - same behavior as before).
      *
      * @return string
      */
     public function backoffMode()
     {
-        return $this->payload()['backoffMode'] ?? self::BACKOFF_FROM_ATTEMPTS;
+        return $this->payload()['backoffMode'] ?? static::BACKOFF_FROM_ATTEMPTS;
     }
 
     /**
--- a/src/Illuminate/Queue/Jobs/Job.php
+++ b/src/Illuminate/Queue/Jobs/Job.php
@@ -285,6 +285,23 @@ abstract class Job
         return $this->payload()['backoff'] ?? $this->payload()['delay'] ?? null;
     }
 
+
+    /**
+     * @const Backoff Calculation Mode: From total of "attempts" (default) o "exceptions"
+     */
+    const BACKOFF_FROM_ATTEMPTS = 'attempts';
+    const BACKOFF_FROM_EXCEPTIONS = 'exceptions';
+
+    /**
+     * Decide how to calculate the backoff (default: "attempts" - same behavior as before)
+     *
+     * @return string
+     */
+    public function backoffMode()
+    {
+        return $this->payload()['backoffMode'] ?? self::BACKOFF_FROM_ATTEMPTS;
+    }
+
     /**
      * Get the number of seconds the job can run.
      *
--- a/src/Illuminate/Queue/Worker.php
+++ b/src/Illuminate/Queue/Worker.php
@@ -557,6 +557,7 @@ class Worker
             $this->cache->put($jobExceptionsKey, 0, Carbon::now()->addDay());
         }
     }
+
     /**
      * Get the job's total number of exceptions.
      *
--- a/src/Illuminate/Queue/Worker.php
+++ b/src/Illuminate/Queue/Worker.php
@@ -13,6 +13,7 @@ use Illuminate\Queue\Events\JobProcessing;
 use Illuminate\Queue\Events\Looping;
 use Illuminate\Queue\Events\WorkerStopping;
 use Illuminate\Support\Carbon;
+use Illuminate\Queue\Jobs\Job;
 use Throwable;
 
 class Worker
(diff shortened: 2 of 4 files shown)


MartinPham:
I create the PR on 9.x following the last comment on the same PR on 8.x


If it gets approved & merged, I will make a PR to update the docs

src/Illuminate/Queue/Worker.php
```diff
@@ -538,19 +573,23 @@ protected function markJobAsFailedIfWillExceedMaxAttempts($connectionName, $job,
...
             is_null($maxExceptions = $job->maxExceptions())) {
             return;
         }
 
-        if (! $this->cache->get('job-exceptions:'.$uuid)) {
-            $this->cache->put('job-exceptions:'.$uuid, 0, Carbon::now()->addDay());
-        }
+        $jobExceptionsKey = $this->getJobTotalExceptionsKey($job);
 
-        if ($maxExceptions <= $this->cache->increment('job-exceptions:'.$uuid)) {
```
taylorotwell, line 550: A race condition has been introduced here. Previously, the `increment` operation here was atomic for some cache systems (Redis, Database, etc.)... but now it has been changed to read, perform some PHP code, and then increment in certain conditions, meaning there could be a race condition between the read and the increment that could lead to unexpected flow.

MartinPham: @taylorotwell Thanks for your review! I've changed the way it increases and compares, to avoid the race condition. Could you please take a look at this?

humaneguy:
@MartinPham Could you provide some unit tests?

MartinPham:
@humaneguy sure! I'm gonna do now. Thanks

MartinPham:
@humaneguy 
- I've improved the behavior, so now it's compatible with previous configuration. We can config to change to behavior when we need.
- I've also added test cases against 2 configurations.
Thanks!

humaneguy:
@MartinPham Great

src/Illuminate/Queue/Worker.php
```diff
@@ -528,6 +529,55 @@ protected function markJobAsFailedIfWillExceedMaxAttempts($connectionName, $job,
...
+
+    /**
+     * Get the job's total number of exceptions.
+     *
+     * @param  \Illuminate\Contracts\Queue\Job  $job
+     * @return int|null
+     */
+    private function getJobTotalExceptions($job)
+    {
+        if (! $this->cache || is_null($job->uuid())) {
```
humaneguy, line 570: This check is already done in the ```setupJobTotalExceptionsIfMissing``` method. No need repeating it here.

tests/Queue/QueueWorkerTest.php
```diff
@@ -391,6 +430,13 @@ private function workerOptions(array $overrides = [])
 
         return $options;
     }
+
+    protected function workerCache()
+    {
+        $repository = new Repository(new ArrayStore);
+
+        return $repository;
```
humaneguy, line 438: No need to return a variable. It could just be a one-liner.

src/Illuminate/Queue/Worker.php
```diff
@@ -597,7 +647,27 @@ protected function calculateBackoff($job, WorkerOptions $options)
                         : $options->backoff
         );
 
-        return (int) ($backoff[$job->attempts() - 1] ?? last($backoff));
+        $backoffMode = $job->backoffMode();
+
+        if ($backoffMode === Job::BACKOFF_FROM_ATTEMPTS) {
```
humaneguy, line 652: Since this is the default case, I'd think you should save it for later.
First check if the backoff mode is calculated by exceptions and return its backoff calculation. Then return this default calculation mode later on.
```suggestion
        if ($backoffMode === Job:: BACKOFF_FROM_EXCEPTIONS) {
            // back off from exceptions calculation
         }
        
        // default calculation of backoff
         return (int) ($backoff[$job->attempts() - 1] ?? last($backoff));
```

src/Illuminate/Queue/Jobs/Job.php
```diff
@@ -285,6 +285,22 @@ public function backoff()
...
+    const BACKOFF_FROM_EXCEPTIONS = 'exceptions';
+
+    /**
+     * Decide how to calculate the backoff (default: "attempts" - same behavior as before).
+     *
+     * @return string
+     */
+    public function backoffMode()
+    {
+        return $this->payload()['backoffMode'] ?? self::BACKOFF_FROM_ATTEMPTS;
```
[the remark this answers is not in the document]
dammy001: The variable is a not static variable so I think using self is okay

src/Illuminate/Queue/Worker.php
```diff
@@ -528,6 +529,55 @@ protected function markJobAsFailedIfWillExceedMaxAttempts($connectionName, $job,
...
+    private function getJobTotalExceptionsKey($job)
+    {
+        return 'job-exceptions:'.$job->uuid();
+    }
+
+    /**
+     * Init the job's total number of exceptions if it's missing.
+     *
+     * @param  \Illuminate\Contracts\Queue\Job  $job
+     * @return null
```
dammy001, line 547: I think this should be `null|void`

src/Illuminate/Queue/Worker.php
MartinPham: Thanks, I was thinking about the default calculation mode also. But then I think maybe it’s better to throw an exception  when the backoffMode’s value is unsupported.
What do you think?

src/Illuminate/Queue/Jobs/Job.php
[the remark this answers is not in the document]
humaneguy: It allows for late static binding. I think it's safer that way.

src/Illuminate/Queue/Worker.php
humaneguy: I think it's just safer to use the default backoff mode.

MartinPham: ok!

MartinPham:
Thanks a lot @humaneguy @dammy001!! I've pushed the improvements following your reviews!

taylorotwell:
I think I'm just going to hold off making any changes here for the time being.

humaneguy:
Are you revisiting this sometime soon? @taylorotwell

PR #40110, opened 2021-12-20
[8.x] Apply `--except` option in combination with `--model` for `model:prune`

markwalet:
Fixed an edge case when using the new `--except` option added in https://github.com/laravel/framework/pull/39749.

When using the `--model` options. The `--except` option normally would get ignored when used in combination.

I couldn't come up with a practical use case for this. That's why I added a warning when using both parameters. I still feel that it makes more sense to support a combination of both parameters. Especially because it never states that all other parameters will be ignored when the model option is used.

Looking forward to any feedback and suggestions regarding this PR.

Thanks!


taylorotwell:
Fixed on 8.x.

PR #40112, opened 2021-12-20
Add observer mapping

punyflash:
This feature allows you to have cleaner observers mapping in `EventServiceProvider` which basically copies syntax of policies, event listeners and other similar stuff. Observers are automatically registered during `booting` hook. In general, this allows to write this:

```php
/**
 * Register any events for your application.
 *
 * @return void
 */
public function boot()
{
    User::observe([UserObserver::class, UserOrderObserver::class]);
    Blog::observe(BlogObserver::class);
    Transaction::observe(TransactionObserver::class);
    Order::observe(OrderObserver::class);
    Payout::observe(PayoutObserver::class);
}
```

like this:
```php
/**
 * The observers mapping for the application models.
 *
 * @var array
 */
protected $observers = [
    User::class             => [ UserObserver::class, UserOrderObserver::class ],
    Blog::class             => BlogObserver::class,
    Transaction::class      => TransactionObserver::class,
    Order::class            => OrderObserver::class,
    Payout::class           => PayoutObserver::class,
];
```

### DIFF
--- a/src/Illuminate/Foundation/Support/Providers/EventServiceProvider.php
+++ b/src/Illuminate/Foundation/Support/Providers/EventServiceProvider.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Foundation\Support\Providers;
 
+use Illuminate\Database\Eloquent\Model;
 use Illuminate\Foundation\Events\DiscoverEvents;
 use Illuminate\Support\Facades\Event;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Not sure I want to add this on a patch release as someone may already be defining this property and doing it manually, etc.

PR #40116, opened 2021-12-20
[8.x] Add a mergeIfMissing method to the Illuminate Http Request class

davidpeach:
When unticking a checkbox and submitting my form, no value for that field is sent to the server. I couldn't guarantee that the column actually wanted to be turned off.

So I added a little check to merge in the "off" state when that key was missing from the request.

Thought it might make a nice little addition.

Before:
```
if ($request->missing('boolean_setting')) {
    $request->merge(['boolean_setting' => 0]);
}
```
After:
```
$request->mergeIfMissing(['boolean_setting' => 0])
```

### DIFF
--- a/src/Illuminate/Http/Request.php
+++ b/src/Illuminate/Http/Request.php
@@ -344,9 +344,7 @@ class Request extends SymfonyRequest implements Arrayable, ArrayAccess
             return $this->getInputSource()->has($key);
         })->toArray();
 
-        $this->getInputSource()->add($input);
-
-        return $this;
+        return $this->merge($input);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


JustSteveKing:
Love this idea!

taylorotwell:
I'm a little concerned there is some subtle unexpected behavior here maybe.

You call `getInputSource()->has(...)`; however, the `missing` method itself just calls `has` on the request, which could have slightly different behavior. Therefore, I'm worried the definition of "missing" is not consistent between the actual `missing` method and `mergeIfMissing`. 

Is there a reason you did `getInputSource()->has(...)` instead of just calling `has`?

davidpeach:
Thanks for the feedback @taylorotwell . Much appreciated.

I see what you mean ye. Not just calling `has` was an oversight on my part.

I've now reversed the check too, so it uses the `missing` method in a `filter`, which makes more sense to me, as its inside a `mergeIfMissing` method.

Thanks again.

Issue #39959, opened 2021-12-09
Renaming timestamp columns converts them to datetime


ChrisThompsonTLDR:
- Laravel Version: 8.75.0
- PHP Version: 7.4.26
- Database Driver & Version: 8.0.26

### Description:

Renaming a timestamp column ends up converting it to a datetime.

### Steps To Reproduce:

```PHP
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

class CreateSubscriptionsTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::create('subscriptions', function (Blueprint $table) {
            $table->id();
            $table->timestamp('membership_starts_at')->nullable();
        });
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
        Schema::dropIfExists('subscriptions');
    }
}
```

```PHP
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

class ChangeSubscriptionsTable extends Migration
{
    /**
     * Run the migrations.
     *
     * @return void
     */
    public function up()
    {
        Schema::create('subscriptions', function (Blueprint $table) {
              $table->renameColumn('membership_starts_at', 'billing_starts_at');
        });
    }

    /**
     * Reverse the migrations.
     *
     * @return void
     */
    public function down()
    {
            $table->renameColumn('billing_starts_at', 'membership_starts_at');
    }
}
```

Now check the table schema and you will see that `billing_starts_at` is a `datetime` not a `timestamp`.

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

However, this issue will not be locked and everyone is still free to discuss solutions to your problem!

Thanks.

[...]

PR #35591, opened 2020-12-13
[8.x] DBAL Timestamp Type

TomHAnderson:
This PR is a response to https://github.com/laravel/framework/issues/35498
Please read that whole issue to put in context this PR.

This PR is in tandem with https://github.com/laravel/laravel/pull/5488

This PR works against all 4 supported databases.  But this repo, laravel/framework, does not have the framework in place to support unit testing all of these databases.  So, I include as an attachment the migration I tested the databases with and the notes on my testing.  

I'm open to hearing better ideas for unit testing which are supported by this repository.

[2020_12_09_123007_datetime_to_timestamp_2020120912300724.php.txt](https://github.com/laravel/framework/files/5684161/2020_12_09_123007_datetime_to_timestamp_2020120912300724.php.txt)

[timestamp.notes.txt](https://github.com/laravel/framework/files/5684160/timestamp.notes.txt)

Issue #35498, opened 2020-12-05
Timestamp is not a valid Doctrine column type


TomHAnderson:
- Laravel Version: 8.17.0
- PHP Version: 7.3.25
- Database Driver & Version: mysql / 10.2.36-MariaDB

### Preface

I created the below issue from my own research before finding this duplicate issue:

This bug report is a duplicate of:  https://github.com/laravel/framework/issues/16526

and I'd like to address the resolution of that issue here.  As a member of the Doctrine team I can assure you this is not an _issue_ in Doctrine.  It is a design decision just as not supporting `enum` is a design decision.  But Laravel uses timestamp column types throughout and timestamp is a standard in Laravel.  So just because Doctrine chooses not to support this directly is not reason enough for Laravel not to take the small steps necessary to support Laraval's own design decisions and implement the column in the DBAL library Laravel is integrating with.  I'm open to writing the PR for this.

### Description:

Attempting to change a column from a datetime to a timestamp results in the following exception:

Unknown column type "timestamp" requested. Any Doctrine type that you use has to be registered with \Doctrine\DBAL\Types\Type::addType(). You can get a list of all the known types with \Doctrine\DBAL\Types\Type::getTypesMap(). If this error occurs during database introspection then you might have forgotten to register all database types for a Doctrine Type. Use AbstractPlatform#registerDoctrineTypeMapping() or have your custom types implement Type#getMappedDatabaseTypes(). If the type name is empty you might have a problem with the cache or forgot some mapping information.

### Steps To Reproduce:

Run this migration against an empty database:

```
    public function up()
    {
        Schema::create('user_to_role', function (Blueprint $table) {
            $table->bigInteger('role_id')->unsigned();
            $table->bigInteger('user_id')->unsigned();
            $table->primary(['role_id','user_id']);
        });

        Schema::table('user_to_role', function (Blueprint $table) {
            $table->dateTime('created_at')->after('user_id');
        });

        Schema::table('user_to_role', function (Blueprint $table) {
            $table->timestamp('created_at')->after('user_id')->change();
        });
    }
```

### Discussion

Laravel uses the mysql `timestamp` field type throughout the framework.  To modify columns Laravel uses the Doctrine DBAL library.  Doctrine has a strict set of data types and uses the more generic `datetime` instead of the MySQL specific `timestamp`.

[...]

PR #40119, opened 2021-12-21
[8.x] Improve support for custom Doctrine column types

bakerkretzmar:
## Context

This PR is my and @TomHAnderson's follow-up to several other issues and PRs that address registering custom Doctrine columns in Laravel.

Using custom Doctrine column types in a Laravel app is useful because it makes renaming and changing columns in migrations really easy. While Laravel handles creating columns itself internally, it relies on Doctrine to rename columns or make changes to them. Doctrine doesn't have built-in support for all of Laravel's column types, so in a fresh Laravel app there are actually several column types that can be created fluently in migrations but require custom SQL statements to change or rename later.

The `TimestampType` added by @TomHAnderson in #35591 is a great example of this—it adds platform-specific SQL statements to support handling timestamp columns on several different database drivers, which allows developers to keep calling `->timestamp()` in migrations and always have it behave as expected regardless of which driver they're using. Without it, renaming or changing a `timestamp` column in a migration might also silently convert it to a `datetime` column, because that's the closest type Doctrine has built in (see #39959).

The `Xml` and `Bit` types used in this PR's tests are similar, platform-specific examples, which would make it possible to do things like `$table->renameColumn('old_xml_name', 'new_xml_name')` or `$table->bit('bit_column')->nullable()->change()`.

See: #35498, #39959, #35591, #39683.

## Motivation

There is existing functionality in the framework that handles most of the required setup for registering custom Doctrine types already, and this PR fills in the gaps to make this functionality more consistent.

Specifically, the Laravel documentation includes [an example](https://laravel.com/docs/8.x/migrations#prerequisites) of registering custom column types automatically by listing them in `config/database.php`, but this currently doesn't work in all contexts (multiple connections, tests, etc.):

```php
use Illuminate\Database\DBAL\TimestampType;

'dbal' => [
    'types' => [
        'timestamp' => TimestampType::class,
    ],
],
```

## Solution

This PR ensures that all custom Doctrine database column types listed in `config('database.dbal.types')` are registered consistently with the framework, so that columns using those types can be created, changed, renamed, and deleted in migrations and in tests.

## Changes

- Moves the automatic registration of custom types in config files from the framework's core `DatabaseServiceProvider` into the `DatabaseManager` class at the point where a particular database connection is actually created.
- Moves the code responsible for performing the registration of a specific custom type from the Schema `Builder` class into the database `Connection` class, because Doctrine's type "mappings" are connection-specific.

## Usage

Using the `database.dbal.types` config array as shown above, or calling `DB::connection()->registerDoctrineType()` directly, should now work as expected in all contexts, and in tests.

## Tests

We added tests for the existing `TimestampType` verifying that `timestamp` columns can be modified and/or changed to `datetime`, and vice versa, on supported database drivers.

We also added tests for completely new custom types, `xml` in Postgres and `bit` in MySQL, verifying that:

- custom types can be registered automatically simply by adding them to the `database.dbal.types` config array
- custom types can be registered on multiple database connections

Finally, we added a test to verify that custom type registration works in tests.

Note: since this feature relies on data types that are specific to certain database drivers, some of our tests are also slightly different for different databases.


src/Illuminate/Database/Schema/Builder.php
```diff
@@ -392,26 +390,10 @@ protected function createBlueprint($table, Closure $callback = null)
...
-
-        if (! Type::hasType($name)) {
-            Type::addType($name, $class);
-
-            $this->connection
-                ->getDoctrineSchemaManager()
-                ->getDatabasePlatform()
-                ->registerDoctrineTypeMapping($type, $name);
-        }
+        $this->connection->registerDoctrineType($class, $name, $type);
```
bakerkretzmar, line 396: I'll submit a separate follow-up PR to remove this method in Laravel 9 since it's now redundant and users should just call `DB::registerCustomDoctrineType` directly.

tests/Integration/Database/SchemaBuilderTest.php
```diff
@@ -65,4 +65,26 @@ public function testRegisterCustomDoctrineType()
...
+
+        $blueprint = new Blueprint('test', function (Blueprint $table) {
+            $table->tinyInteger('test_column')->change();
+        });
+
+        $blueprint->build($this->getConnection(), new SQLiteGrammar);
+
+        $this->assertArrayHasKey(TinyInteger::NAME, Type::getTypesMap());
+        $this->assertSame('tinyinteger', Schema::getColumnType('test', 'test_column'));
+    }
```
bakerkretzmar, line 89: This test is identical to the one above it, but prior to this PR it failed, because registering custom Doctrine types only worked _once_ during test runs (the very first time).

src/Illuminate/Database/DBAL/TimestampType.php
```diff
@@ -36,7 +36,7 @@ public function getSQLDeclaration(array $fieldDeclaration, AbstractPlatform $pla
                 return $this->getSQLitePlatformSQLDeclaration($fieldDeclaration);
 
             default:
-                throw new DBALException('Invalid platform: '.$name);
+                throw new RuntimeException('Invalid platform: '.$name);
```
driesvints, line 39: This is a breaking change unfortunately.

tpetry:
I am not against your pull request but wouldn't it make more sense just providing the missing doctrine types than rewriting parts of Laravel, so the user can add new doctrine types? I mean it's already possible to add custom doctrine types for migrations. I am [doing this for PostgreSQL](https://github.com/tpetry/laravel-postgresql-enhanced) where I am adding numerous missing types [[1]](https://github.com/tpetry/laravel-postgresql-enhanced/blob/0.7.0/src/PostgresqlEnhancedServiceProvider.php#L84) [[2]](https://github.com/tpetry/laravel-postgresql-enhanced/tree/0.7.0/src/Types).

bakerkretzmar:
@tpetry yeah providing more of the missing types out of the box would be great! You're right it's already possible to do this, this PR just aims to make it a bit easier.

skollro:
Using custom DBAL types via `database.dbal.types` leads to the following error with Laravel's parallel testing feature: `Doctrine\DBAL\Driver\PDO\Exception: SQLSTATE[08006] [7] FATAL:  sorry, too many clients already`. Tests work as expected when I register the custom DBAL type using `DB::registerCustomDoctrineType()` in the `AppServiceProvider`'s `register`method.

bakerkretzmar:
@skollro Interesting, do you get that error immediately, like on one of the first tests, or only after a lot of tests have run? I wonder if the code we're calling inside Doctrine is creating new connections every time and if there's a way around that.... I'll look into this asap.

skollro:
@bakerkretzmar Thanks a lot! Yes, it happens after a few tests. I also had the same thoughts and I've tracked it down to this line https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Connection.php#L1038

bakerkretzmar:
> Tests work as expected when I register the custom DBAL type using `DB::registerDoctrineType()` in the `AppServiceProvider`'s `register` method.

@skollro can you show me exactly how you're doing this? It fails both ways for me.

I _think_ the line you referenced is causing the tests to create two database connections when they only need one, but I'm not sure why that matters. Laravel is already creating connections and that doesn't cause any issues, so maybe they're getting cleaned up somehow and Doctrine's aren't?

canvural:
This change broke our tests when we upgraded to 8.78.1

We had `dbal` array in our `config/database.php` Removing that and adding a `DB::registerDoctrineType()` in `AppServiceProvider` instead fixed it. 

So this seems like a breaking change.

TomHAnderson:
@canvural Before rearranging your Laravel project did you increase your packet size?

```
max_allowed_packet=200M
```

canvural:
> @canvural Before rearranging your Laravel project did you increase your packet size?
> 
> ```
> max_allowed_packet=200M
> ```

Why would I? It's working fine with 8.77

TomHAnderson:
> Why would I? It's working fine with 8.77

I ask because the error was a MySQL error.  However, I don't want to discount your report to just be the fault of another service.  Can you provide a unit test that replicates the issue you've reported?  I'm afraid that without a test to duplicate it we're going to be at a loss.

bakerkretzmar:
@canvural are you running tests in parallel? If so, do you still see those errors when you run them 'normally' (not in parallel)?

canvural:
> @canvural are you running tests in parallel? If so, do you still see those errors when you run them 'normally' (not in parallel)?

No, I don't run parallel.

skollro:
@bakerkretzmar You're right, I've recreated the issue in a fresh Laravel project, using a PostgreSQL database. Calling `DB::registerDoctrineType()` in the `AppServiceProvider` only works for sequential test runners. Not even the test databases are created when parallel testing is used. Maybe @nunomaduro could have a look to help us here?

canvural:
Also adding `DB::registerDoctrineType` call to a service provider causes Laravel to make db connections at boot. I'm trying to avoid doing any db calls in any service provider. 

An application should be bootable even if there is no db connection or it has wrong configuration.

canvural:
I created an example repo: https://github.com/canvural/doctrine-type-bug

In the `master` branch there is just the `dbal` array in `config/database.php` And when running tests it fails. 

In `fixed` branch I removed the `dbal` array and instead added `DB::registerDoctrineType` to `AppServiceProvider` and it works.

PR #40121, opened 2021-12-21
[9.x] Added View Create Command

ahmetbarut:
This command allows to create a view file from terminal. In some cases, when projects grow, it can be annoying to navigate to the resources directory and navigate through the directories in it. With this command, it is sufficient to write only the directory name and the name of the view to be created.
> If the file exists, the command returns a warning.
## normal use
```
php artisan make:view viewName
```

If the specified directory does not exist, it creates a directory by adding the `-p` flag.

```
php artisan make:view directory.viewName -p
```

If a view file exists in the specified directory, it will issue a warning and stop the command. If the user wants to force this, they add the -f flag and the corresponding file is overwritten.

```
php artisan make:view directory.viewName -f
```

### DIFF
--- a/src/Illuminate/Foundation/Console/ViewCreateCommand.php
+++ b/src/Illuminate/Foundation/Console/ViewCreateCommand.php
@@ -57,7 +57,7 @@ class ViewCreateCommand extends Command
         if (file_exists($viewPath) && ! $force) {
 
             $this->error('View already exists!');
-            
+
             return Command::FAILURE;
         } elseif (file_exists($viewPath) && $force) {
             unlink($viewPath);
--- a/src/Illuminate/Foundation/Console/ViewCreateCommand.php
+++ b/src/Illuminate/Foundation/Console/ViewCreateCommand.php
@@ -57,6 +57,7 @@ class ViewCreateCommand extends Command
         if (file_exists($viewPath) && ! $force) {
 
             $this->error('View already exists!');
+            
             return Command::FAILURE;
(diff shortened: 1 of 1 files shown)


src/Illuminate/Foundation/Console/ViewCreateCommand.php
```diff
@@ -0,0 +1,79 @@
...
+use Illuminate\Support\Str;
+
+class ViewCreateCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'view:create {name}
```
mateusjunges, line 15: Why not `make:view`?

src/Illuminate/Foundation/Console/ViewCreateCommand.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+namespace App\Console\Commands;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Str;
+
+class ViewCreateCommand extends Command
```
mateusjunges, line 8: Laravel have a base `Illuminate\Console\GeneratorCommand` that you should extend here instead of just `Command`.

novalramdhani:
Hmm yeah, I'm interested in this feature if these PRs are merged, the reason is to simplify the performance of creating blade view files.

taylorotwell:
See comment about `GeneratorCommand`.

PR #40126, opened 2021-12-21
Remove redundant check

andershagbard:
This check should be enough to determinate if the variable is a string and empty.

```php
$value === ''
```


taylorotwell:
If it's not broken atm I will just leave as-is.

PR #40129, opened 2021-12-21
[8.x] Implement Full-Text Search for MySQL & PostgreSQL

driesvints:
This PR implements [natural language Full-Text Searches for MySQL](https://dev.mysql.com/doc/refman/8.0/en/fulltext-natural-language.html) and [PostgreSQL](https://github.com/laravel/framework/pull/40229). The columns that are searched in the `MATCH` part always need to exist as a `fulltext` index.

```php
Schema::create('articles', function (Blueprint $table) {
    $table->id('id');
    $table->string('title', 200);
    $table->text('body');
    $table->fulltext(['title', 'body']);
});
```

```php
// Search for "databases" in the title and body fulltext index...
$articles = DB::table('articles')->whereFulltext(['title', 'body'], 'database')->get();

// Search for "databases" in the title and body fulltext index with boolean mode...
$articles = DB::table('articles')->whereFulltext(['title', 'body'], 'database', ['mode' => 'boolean'])->get();

// Search for "databases" in the title and body fulltext index with an expanded query...
$articles = DB::table('articles')->whereFulltext(['title', 'body'], 'database', ['expanded' => true])->get();
```

More info:
- https://dev.mysql.com/doc/refman/8.0/en/fulltext-natural-language.html
- https://dev.mysql.com/doc/refman/8.0/en/fulltext-boolean.html
- https://dev.mysql.com/doc/refman/8.0/en/fulltext-query-expansion.html

Thanks to @tpetry for also implementing PostgreSQL support! https://github.com/laravel/framework/pull/40229

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1829,7 +1829,7 @@ class Builder
      * @param  string  $boolean
      * @return $this
      */
-    public function whereFulltext($columns, $value, array $options = [], $boolean = 'and')
+    public function whereFullText($columns, $value, array $options = [], $boolean = 'and')
     {
         $type = 'Fulltext';
 
@@ -1849,7 +1849,7 @@ class Builder
      * @param  string  $value
      * @return $this
      */
-    public function orWhereFulltext($columns, $value, array $options = [])
+    public function orWhereFullText($columns, $value, array $options = [])
     {
         return $this->whereFulltext($columns, $value, $options, 'or');
     }
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1837,6 +1837,8 @@ class Builder
 
         $this->wheres[] = compact('type', 'columns', 'value', 'options', 'boolean');
 
+        $this->addBinding($value);
+
         return $this;
     }
 
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1841,6 +1841,18 @@ class Builder
     }
 
     /**
+     * Add a "or where fulltext" clause to the query.
+     *
+     * @param  string|string[]  $columns
+     * @param  string  $value
+     * @return $this
+     */
+    public function orWhereFulltext($columns, $value, array $options = [])
+    {
+        return $this->whereFulltext($columns, $value, $options, 'or');
+    }
+
+    /**
      * Add a "group by" clause to the query.
      *
      * @param  array|string  ...$groups
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1822,18 +1822,16 @@ class Builder
     }
 
     /**
-     * Add a "match against" clause to the query.
+     * Add a "where fulltext" clause to the query.
      *
      * @param  string|string[]  $column
      * @param  string|string[]  $values
      * @param  string  $boolean
-     * @param  bool  $expanded
-     * @param  string  $mode
      * @return $this
      */
-    public function matchAgainst($columns, $values, $boolean = 'and', bool $expanded = false, string $mode = '')
+    public function whereFulltext($columns, $values, array $options = [], $boolean = 'and')
     {
-        $type = 'MatchAgainst';
+        $type = 'Fulltext';
 
         // Next, if the value is Arrayable we need to cast it to its raw array form so we
         // have the underlying array value instead of an Arrayable object which is not
@@ -1845,39 +1843,11 @@ class Builder
         $columns = (array) $columns;
         $value = collect($this->cleanBindings((array) $values))->implode(',');
 
-        $this->wheres[] = compact('type', 'columns', 'value', 'expanded', 'mode', 'boolean');
+        $this->wheres[] = compact('type', 'columns', 'value', 'options', 'boolean');
 
         $this->addBinding($value);
 
         return $this;
-    }
-
-    /**
-     * Add a "match against in boolean mode" clause to the query.
-     *
-     * @param  string|string[]  $column
-     * @param  string|string[]  $values
-     * @param  string  $boolean
-     * @param  bool  $expanded
-     * @return $this
-     */
-    public function matchAgainstBoolean($columns, $values, $boolean = 'and', bool $expanded = false)
-    {
-        return $this->matchAgainst($columns, $values, $boolean, $expanded, 'boolean');
-    }
-
-    /**
-     * Add a "match against with expanded query" clause to the query.
-     *
-     * @param  string|string[]  $column
-     * @param  string|string[]  $values
-     * @param  string  $boolean
-     * @param  string  $mode
-     * @return $this
-     */
-    public function matchAgainstExpanded($columns, $values, $boolean = 'and', string $mode = '')
-    {
-        return $this->matchAgainst($columns, $values, $boolean, true, $mode);
     }
 
     /**
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1824,7 +1824,7 @@ class Builder
     /**
      * Add a "where fulltext" clause to the query.
      *
-     * @param  string|string[]  $column
+     * @param  string|string[]  $columns
      * @param  string  $value
      * @param  string  $boolean
      * @return $this
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1822,6 +1822,65 @@ class Builder
     }
 
     /**
+     * Add a "match against" clause to the query.
+     *
+     * @param  string|string[]  $column
+     * @param  string|string[]  $values
+     * @param  string  $boolean
+     * @param  bool  $expanded
+     * @param  string  $mode
+     * @return $this
+     */
+    public function matchAgainst($columns, $values, $boolean = 'and', bool $expanded = false, string $mode = '')
+    {
+        $type = 'MatchAgainst';
+
+        // Next, if the value is Arrayable we need to cast it to its raw array form so we
+        // have the underlying array value instead of an Arrayable object which is not
+        // able to be added as a binding, etc. We will then add to the wheres array.
+        if ($values instanceof Arrayable) {
+            $values = $values->toArray();
+        }
+
+        $columns = (array) $columns;
+        $value = collect($this->cleanBindings((array) $values))->implode(',');
+
+        $this->wheres[] = compact('type', 'columns', 'value', 'expanded', 'mode', 'boolean');
+
+        $this->addBinding($value);
+
+        return $this;
+    }
+
+    /**
+     * Add a "match against in boolean mode" clause to the query.
+     *
+     * @param  string|string[]  $column
+     * @param  string|string[]  $values
+     * @param  string  $boolean
+     * @param  bool  $expanded
+     * @return $this
+     */
+    public function matchAgainstBoolean($columns, $values, $boolean = 'and', bool $expanded = false)
+    {
+        return $this->matchAgainst($columns, $values, $boolean, $expanded, 'boolean');
+    }
+
+    /**
+     * Add a "match against with expanded query" clause to the query.
+     *
+     * @param  string|string[]  $column
+     * @param  string|string[]  $values
+     * @param  string  $boolean
+     * @param  string  $mode
+     * @return $this
+     */
+    public function matchAgainstExpanded($columns, $values, $boolean = 'and', string $mode = '')
+    {
+        return $this->matchAgainst($columns, $values, $boolean, true, $mode);
+    }
+
+    /**
      * Add a "group by" clause to the query.
      *
      * @param  array|string  ...$groups
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1837,8 +1837,6 @@ class Builder
 
         $this->wheres[] = compact('type', 'columns', 'value', 'options', 'boolean');
 
-        $this->addBinding($value);
-
         return $this;
     }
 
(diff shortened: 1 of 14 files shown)


Wulfheart:
Might also be related for SQL Server: https://docs.microsoft.com/en-us/sql/relational-databases/search/full-text-search

deleugpn:
I worked on 3 database migrations (EC2 to RDS, RDS to Aurora and Aurora 5.6 to 5.7) on databases ranging from 100GB to 350GB and all 3 migrations were a nightmare because of Mysql full text indexes. When using services like AWS DMS to copy data over without downtime, large table with fulltext indexes easily crash. MySQL 5.7 is very buggy with fulltext indexes and can lead to a lot of deadlocks and semaphore waits crashing after 600 seconds. Even a table with only 2 columns (id and comment) where the comment column is fulltext indexed lead to database crashes (Aurora MySQL 5.7 compatible). I don't know how 8.0 is now but I would avoid MySQL fulltext indexes like the plague.

ostark:
@driesvints 
Nice addition 

@deleugpn
In my experience mysql fulltext search is good enough for mid size datasets, like ~1M rows.

GrahamCampbell:
RDS Aurora supports adding indexes only to read-replicas. :P

tpetry:
As promised on Twitter, I would take a look at the implementation for porting it to PostgreSQL. The implementation looks solid but is very MySQL-specific (as expected). For introducing PostgreSQL support, I would have to add entirely new functions, as the current ones are exactly designed to work with MySQL.

I had already been working on a common implementation, but that's not simple. PostgreSQL's fulltext engine is wholly different from the MySQL one:
* You need to specify a language for fulltext indexes and search because stemming on words is different for every language (a common ruleset like MySQL does will not work correctly for any non-english language)
* There is no boolean or natural language mode because you are being handed all the options on how to do matching than just two mode
  * You can build very complex matching rules by [combining text search operators](https://www.postgresql.org/docs/12/functions-textsearch.html) 
  * Some functions exist for common search behavior to build text search operators by a string
      * The `plainto_tsquery` function will transform a search string by and-ing the search terms: `plainto_tsquery('english', 'The Fat Rats door')` --> must match `'fat' & 'rat' & 'door'`
      * The `phraseto_tsquery` function will transform a search string by needing the phrases to appear in the exact order one after another: `phraseto_tsquery('english', 'The Fat Rats door')` --> must match `'fat' <-> 'rat' <-> 'door'`
      * The `websearch_to_tsquery` function is mostly like MySQL's boolean search mode, which allows positive/negative operators in the text. But contrary to MySQL the terms are combined with `and` instead of `or`:  `websearch_to_tsquery('english', 'fat +rat -door')` --> must match `'fat' & 'rat' & !'door'`

These are the most impactful differences. But still is not tackling the options PostgreSQL will provide. The article [Fine Tuning Full Text Search with PostgreSQL 12](https://rob.conery.io/2019/10/29/fine-tuning-full-text-search-with-postgresql-12/) easily explains how many options you have to really fine-tune the fulltext search behavior of PostgreSQL in contrast to MySQL.

---

My advice would be to make the following changes to be able to add a very basic PostgreSQL implementation i could work on:
* Remove the `expanded ` and `mode` parameters from `matchAgainst` as they will only make sense for MySQL
  * They could be replaced with a single `$options` array which every implementation could interpret differently: `['expanded' => boolean, 'mode' => 'boolean' | 'natural']` could work for MySQL and `['language' => 'english', 'query' => 'websearch_to_tsquery']` and countless more options could work for example for PostgreSQL. That's the implementation I were working on.
* Transforming multiple strings to a single one should be done by the database driver as PostgreSQL will need an entirely different one (imploding by space and not comma)

I am aware that supporting all PostgreSQL options for fulltext search are beyond the scope of Laravel. With my adviced changes i could add them one by one to [tpetry/laravel-postgresql-enhanced](https://github.com/tpetry/laravel-postgresql-enhanced).

tpetry:
Another question I have is the default mode for MySQL fulltext search by Laravel? Which mode will be used? It would be best aligning the default behavior for MySQL and PostgreSQL so they kind of behave similar. The results will not be exactly similar (search term combining differences in `and` and `or` logic).

I am not very experiences in MySQL's fulltext search and the documentation does not _exactly_ describe how the default mode works. Is the default mode comparable to one of the PostgreSQL string functions i described? I guess making the default search behaviour the typical google one "laravel database +postgresql" would be the one everyone is expecting?

driesvints:
Placing this in draft while I talk to @tpetry on how to improve this PR.

driesvints:
We simplified the API to `whereFulltext` so we can implement this on different engines. The extra `$options` argument can be used to use engine specific features.

src/Illuminate/Database/Query/Builder.php
```diff
@@ -1821,6 +1821,27 @@ protected function addDynamic($segment, $connector, $parameters, $index)
         $this->where(Str::snake($segment), '=', $parameters[$index], $bool);
     }
 
+    /**
+     * Add a "where fulltext" clause to the query.
+     *
+     * @param  string|string[]  $column
```
humaneguy, line 1827: ```$column``` or ```$columns```?

src/Illuminate/Database/Query/Builder.php
```diff
@@ -1821,6 +1821,27 @@ protected function addDynamic($segment, $connector, $parameters, $index)
...
+    {
+        $type = 'Fulltext';
+
+        $columns = (array) $columns;
+
+        $this->wheres[] = compact('type', 'columns', 'value', 'options', 'boolean');
+
+        $this->addBinding($value);
+
+        return $this;
```
humaneguy, line 1840: We could just return ```$this->addBinding($value)```

driesvints:
Merged @tpetry's PostgreSQL PR into this one.

src/Illuminate/Database/Query/Grammars/Grammar.php
```diff
@@ -629,6 +629,18 @@ protected function compileJsonLength($column, $operator, $value)
...
+    /**
+     * Compile a "where fulltext" clause.
+     *
+     * @param  \Illuminate\Database\Query\Builder  $query
+     * @param  array  $where
+     * @return string
+     */
+    public function whereFulltext(Builder $query, $where)
+    {
+        throw new RuntimeException('This database engine does not support fulltext search operations.');
```
humaneguy, line 641: This means for other engines where fulltext search operations have not been implemented we still get this exception🤨.

tests/Database/DatabaseQueryBuilderTest.php
```diff
@@ -858,6 +858,34 @@ public function testArrayWhereColumn()
         $this->assertEquals([], $builder->getBindings());
     }
 
+    public function testWhereFulltext()
```
humaneguy: Instead of having all of the various test cases in a singe test method, I think it'd be better if we split them to shoreter methods as this method tends to grow later in the future.

src/Illuminate/Database/Query/Grammars/Grammar.php
driesvints: You just don't use this method on other engines. It doesn't makes sense to use a method that's not supported.

tests/Database/DatabaseQueryBuilderTest.php
driesvints: This is fine.

humaneguy:
Hi @driesvints, Redist also supports Fulltext search. I'd like to work on that, too.

src/Illuminate/Database/Query/Builder.php
driesvints: Thanks

driesvints:
@humaneguy since Redis isn't a direct database engine but rather a standalone thing, I'd rather not mix up things in this PR.

humaneguy:
@driesvints Ok, great.

src/Illuminate/Database/Query/Grammars/Grammar.php
humaneguy: Yeah. But since not all of them are covered now. For instance, SQL Server supports fulltext search but it's not been added. I think, we could modify the exception message a little. Something in the lines of ```This database engine does not support fulltext search operations or there is no current implementation of it yet. Kindly revert to external packages if that's the case```. Too long, tho'

driesvints: You can send in a PR if you want mssql support 👍

src/Illuminate/Database/Query/Grammars/MySqlGrammar.php
```diff
@@ -50,6 +50,31 @@ protected function whereNotNull(Builder $query, $where)
...
+     *
+     * @param  \Illuminate\Database\Query\Builder  $query
+     * @param  array  $where
+     * @return string
+     */
+    public function whereFulltext(Builder $query, $where)
+    {
+        $columns = $this->columnize($where['columns']);
+
+        $query->addBinding($where['value']);
```
taylorotwell, line 64: We never add bindings in grammars typically. How will this binding be in the right order if where clauses are added after this operation while the developer is building the query?

src/Illuminate/Database/Query/Grammars/PostgresGrammar.php
```diff
@@ -517,4 +546,22 @@ protected function wrapJsonPathAttributes($path)
...
+     * @param  mixed  $value
+     * @return mixed
+     */
+    protected function bindParameter(Builder $query, $value)
+    {
+        if ($this->isExpression($value)) {
+            return $this->getValue($value);
+        }
+
+        $query->addBinding($value);
```
taylorotwell, line 563: Same binding order issue here. Bindings must be added while the query is being built otherwise they will never be in the correct order?

taylorotwell:
This is also missing `orWhereFulltext`.

src/Illuminate/Database/Query/Grammars/PostgresGrammar.php
humaneguy: What should then be the approach to fixing this bug?

src/Illuminate/Database/Query/Grammars/MySqlGrammar.php
tpetry: Hmm, i don't know how to solve it. The PostgreSQL version needs more user parameters than the MySQL one because a language name is required. I built the code with the idea in mind that the language parameter should be sql-injection safe. If we remove this security-barrier then the bug could be fixed. What do you think about it @taylorotwell @driesvints?

By the way it's a good example for my pull request #40115. With the grammar now having access to a sql-injection safe `quoteString` implementation, the added bindings in the grammar would not be needed anymore. The language parameter could be made safe by `quoteString` and the value binding moved to the `whereFulltext` method again.

So i see two solutions on fixing it:
1. Make the language parameter sql-injection vulnerable and move the binding of the search value to `whereFulltext`
2. Merging #40115 and moving fulltext search to `9.x` as the grammar could now use the safe `quoteString` method
3. Any other idea?

driesvints: I think it should be save to add the language param directly to the query since you're explicitly passing it as an option and not as a userland provided value. Sanitization should be done on the userside. @taylorotwell do you agree with that? If so, @tpetry could you send in those changes or do you want me to do that?

driesvints:
@taylorotwell added `orWhereFulltext`.

src/Illuminate/Database/Query/Grammars/MySqlGrammar.php
tpetry: I can work on it in a few hours.

src/Illuminate/Database/Query/Grammars/PostgresGrammar.php
driesvints: @humaneguy see the convo above.

src/Illuminate/Database/Query/Grammars/MySqlGrammar.php
driesvints: @tpetry I've already pushed the changes, thanks 👍

driesvints:
I've adjusted the bindings to be back on the query builder. Please note that this now sets `$language` directly in the query. 

A final alternative could be that we check the language option inside the query builder but then it could conflict with other grammars in the future.

jonnott:
@driesvints @taylorotwell  A common querying pattern with fulltext indexes is to use the `MATCH .. AGAINST .. AS relevance` in the SELECT clause, and then add `WHERE relevance > 0`.

e.g. as stated within https://www.cloudsavvyit.com/10172/how-to-use-full-text-searches-in-mysql/ ..

"When using MATCH ... AGAINST in a SELECT statement, you don’t need to repeat it in the WHERE clause. You could manually filter the results to include only records with a non-zero relevance score."

`SELECT content, MATCH (content) AGAINST ('database engine') AS relevance FROM articles WHERE relevance > 0 ORDER BY relevance DESC`

Is there any possibility of somehow making methods to create that kind of query a part of what's built-in to Eloquent?

I think otherwise for many uses-cases where query results need to be ordered by relevance, we'd end up still doing something like:

```
$query = $this->model
    ->selectRaw("*,MATCH (words) AGAINST (? in natural language mode) as relevance", [$search_words])
    ->having('relevance', '>', 0)
    ->orderByDesc('relevance');
```

driesvints:
@jonnott I think it's best that you attempt a PR for that.

tpetry:
@jonnott you could start experimenting on a `orderByFullText` implementation. Keep me in the loop for PostgreSQL specifics.

jonnott:
> @jonnott you could start experimenting on a `orderByFullText` implementation. Keep me in the loop for PostgreSQL specifics.

@tpetry I have a wip here https://github.com/jonnott/framework/commit/7cee6f851cf4ed4a2f83b26cae1dadea4af7dbff which could turn into a draft PR for this. I've added selectFullText() and addSelectFullText() methods. I have no clue on the PostGres side of things though..

mshamaseen:
Great addition, but what can be sent in the `$options` argument? I can't find any documentation for that

tpetry:
Look at the changes of the MySQL grammar. My enhanced PostgreSQL driver also uses the options to add some PG specific stuff.

mshamaseen:
@tpetry Thanks

So:
`array{mode:string}` for both MySql and Postgres
`array{expanded:bool}` for MySql
`array{Language:string}` for Postgres

It would be great if that could be added to the PHPDoc on the `whereFullText` method.

PR #40131, opened 2021-12-21
 [9.x] Added View Create Command

ahmetbarut:
This command allows to create a view file from terminal. In some cases, when projects grow, it can be annoying to navigate to the resources directory and navigate through the directories in it. With this command, it is sufficient to write only the directory name and the name of the view to be created.
> If the file exists, the command returns a warning.
## normal use
```
php artisan make:view viewName
```

If a view file exists in the specified directory, it will issue a warning and stop the command. If the user wants to force this, they add the -f flag and the corresponding file is overwritten.

```
php artisan make:view directory.viewName -f
```

### DIFF
--- a/src/Illuminate/Foundation/Console/ViewCreateCommand.php
+++ b/src/Illuminate/Foundation/Console/ViewCreateCommand.php
@@ -57,7 +57,7 @@ class ViewCreateCommand extends Command
         if (file_exists($viewPath) && ! $force) {
 
             $this->error('View already exists!');
-            
+
             return Command::FAILURE;
         } elseif (file_exists($viewPath) && $force) {
             unlink($viewPath);
--- a/src/Illuminate/Foundation/Console/ViewCreateCommand.php
+++ b/src/Illuminate/Foundation/Console/ViewCreateCommand.php
@@ -12,8 +12,7 @@ class ViewCreateCommand extends GeneratorCommand
      *
      * @var string
      */
-    protected $signature = 'make:view {name} 
-                {--p|path : If the specified directory does not exist, it creates a directory.}
+    protected $signature = 'make:view {name}
                 {--f|force : Overwrite existing view if any}';
 
     /**
@@ -37,12 +36,11 @@ class ViewCreateCommand extends GeneratorCommand
     public function handle()
     {
 
-        // Here, we look at whether there is a view file to be created, then the relevant  
-        // answers are returned to those who use the command, and at the end, the view file is  
+        // Here, we look at whether there is a view file to be created, then the relevant
+        // answers are returned to those who use the command, and at the end, the view file is
         // created.
 
         $force = $this->option('force');
-        $path = $this->option('path');
 
         // If it specifies directory and file as dots, it converts dots to slash(/)
         $viewName = $this->getView();
@@ -52,19 +50,15 @@ class ViewCreateCommand extends GeneratorCommand
         // If the file exists, and the user wants to overwrite it, it will be overwritten.
         if (file_exists($viewPath) && ! $force) {
             $this->error('View already exists!');
+
             return GeneratorCommand::FAILURE;
         } elseif (file_exists($viewPath) && $force) {
             unlink($viewPath);
         }
 
         // If the directory exists, and the user wants to overwrite it, it will be overwritten.
-        if (! is_dir(dirname($viewPath)) && $path) {
+        if (! is_dir(dirname($viewPath))) {
             mkdir(dirname($viewPath), 0755, true);
-        } elseif (! is_dir(dirname($viewPath)) && ! $path) {
-            $this->error("Directory does not exist!");
-            $this->info('If you want to create the directory as well add the -p flag');
-
-            return GeneratorCommand::FAILURE;
         }
 
         $this->info('Creating view:'.$viewName.'.blade.php');
--- a/src/Illuminate/Foundation/Console/ViewCreateCommand.php
+++ b/src/Illuminate/Foundation/Console/ViewCreateCommand.php
@@ -57,6 +57,7 @@ class ViewCreateCommand extends Command
         if (file_exists($viewPath) && ! $force) {
 
             $this->error('View already exists!');
+            
             return Command::FAILURE;
         } elseif (file_exists($viewPath) && $force) {
             unlink($viewPath);
--- a/src/Illuminate/Foundation/Console/ViewCreateCommand.php
+++ b/src/Illuminate/Foundation/Console/ViewCreateCommand.php
@@ -41,8 +41,8 @@ class ViewCreateCommand extends Command
     public function handle()
     {
 
-        // Here, we look at whether there is a view file to be created, then the relevant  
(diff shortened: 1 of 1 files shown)


isaac-souza:
> 
> If the specified directory does not exist, it creates a directory by adding the `-p` flag.
> 
> ```
> php artisan make:view directory.viewName -p
> ```

Just a random opinion, but I just think the folder should be created by default if it doesn't exist already, that is the default behavior for all native Laravel commands I believe.

src/Illuminate/Foundation/Console/ViewCreateCommand.php
```diff
@@ -0,0 +1,80 @@
...
+    {
+
+        // Here, we look at whether there is a view file to be created, then the relevant
+        // answers are returned to those who use the command, and at the end, the view file is
+        // created.
+
+        $force = $this->option('force');
+
+        // If it specifies directory and file as dots, it converts dots to slash(/)
+        $viewName = $this->getView();
```
mateusjunges, line 46: Personally i don't think you should extract this code to a separate method, since it's just a one line string manipulation.

src/Illuminate/Foundation/Console/ViewCreateCommand.php
```diff
@@ -0,0 +1,80 @@
...
+        if (file_exists($viewPath) && ! $force) {
+            $this->error('View already exists!');
+
+            return GeneratorCommand::FAILURE;
+        } elseif (file_exists($viewPath) && $force) {
+            unlink($viewPath);
+        }
+
+        // If the directory exists, and the user wants to overwrite it, it will be overwritten.
+        if (! is_dir(dirname($viewPath))) {
```
mateusjunges, line 60: This comment does not make sense here

src/Illuminate/Foundation/Console/ViewCreateCommand.php
```diff
@@ -0,0 +1,80 @@
...
+            mkdir(dirname($viewPath), 0755, true);
+        }
+
+        $this->info('Creating view:'.$viewName.'.blade.php');
```
mateusjunges, line 64:
```diff
- $this->info('Creating view:'.$viewName.'.blade.php');
+ $this->info('Creating view: '.$viewName.'.blade.php');
```

mateusjunges:
I agree with @isaac-souza, but looking at your code, the `-p` options seems to never be used, not even in the command signature, so i think you just forgot it in the PR description from #40121.

src/Illuminate/Foundation/Console/ViewCreateCommand.php
ahmetbarut: Thank you for your interest.

ahmetbarut:
> I agree with @isaac-souza, but looking at your code, the `-p` options seems to never be used, not even in the command signature, so i think you just forgot it in the PR description from #40121.
@mateusjunges, @isaac-souza I did what he was told. I guess you didn't review the codes? 
 [Here review](https://github.com/ahmetbarut/framework/blob/50218366634e734ee16655c7eb2dc485208e5890/src/Illuminate/Foundation/Console/ViewC)

src/Illuminate/Foundation/Console/ViewCreateCommand.php
```diff
@@ -0,0 +1,75 @@
...
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Create a view file in the resources/views directory';
+
+    /**
+     * Create a new command instance.
```
humaneguy, line 26: Do you really need this docblock?

humaneguy: It seems to be out of place here.

NickSdot:
Hey @ahmetbarut, I've down-voted this. But it's not because I don't like the idea. Please allow me to share some things I have noticed and concerns I have about the approach you have taken. 🤗 

1) **There is not only one kind of view in Laravel.** 
People use different stacks. A view for instance could be created with Blade, Vue, React or any other frontend technique/framework. Hence, reserving the command `make:view` for blade feels off to me. It might be a good idea to consider to rather chose `make:blade` for making a Blade view. So at a later point other commands like `make:vue`, `make:react` and so on could be introduced without confusion.

2) **There is not only one view path in Laravel.**
Views can live literally everywhere in the app (by loading them via `\Illuminate\Support\ServiceProvider::loadViewsFrom`). The PR doesn't cover this at all and assumes that all created views will live in a location relative to the `resource_path`.

4) **Ignores conventions/patterns introduced by other commands**
It introduces several own implementations for functionality which already exist in the parent class `GeneratorCommand`. Even though this is also the case in other commands, I think it would be a good idea to no go this path in future commands too. These things caught my attention:
a) class name should be `ViewMakeCommand` (or if you agree on what I wrote in '1', imho even better: `BladeMakeCommand`), instead of `ViewCreateCommand`. 
b) it introduces `directory.viewName` to define a directory to create the view in.  But if you, for instance, try `make:model foo/bar/MyModel`, you'll see that the model would be created under the given path anyway. I think it should work the same way in this PR. Actually I think there is no need to override `handle()` at all for this command.
c) it does not make use of: `makeDirectory()` and `alreadyExists()`, which causes a different behavior and will make it harder to maintain in the future. (also solvable by not overriding `handle()`).
d) Using unlink() feels a bit dangerous, What if in the following steps for whatever reason the creation of the new file fails? The old file would be deleted even the new one would not have been created. I believe it's a good idea to simply overwrite the old file, as in other commands too.  (also solvable by not overriding `handle()`).
e) `$type` isn't set at all.

4) **There is no way to customize the stub**
Unlike some other commands, this one doesn't make use of `resolveStubPath()` to enable developers to customize the given empty stub file by placing their own stub in `root/stubs`. Maybe you want to consider to use `resolveStubPath()`  and allow to publish the stub via `StubPublishCommand`?

5) **It might makes sense to enforce naming best practices**
For instance in `make:model` we can see that the file naming best practices are enforced. For controllers/seeders/factories studly-case and for migrations snake-case with date prefix. For blade views kebab-case could be enforced by overriding  e.g. in `getNameInput()` or `getPath()`.

humaneguy:
I quite appreciate this @NickSdot. Taking your time to put all of this time.👏🏼 I really learn a lot. It'd be good if @ahmetbarut converts it to a draft PR and review all of the things you listed.

ahmetbarut:
@NickSdot Thank you for the feedback and I'll take the time to review what you said.

@humaneguy I will definitely take the time to thank you for your feedback.

taylorotwell:

ahmetbarut:
@taylorotwell 
Hi, this command is to create a view inside the frame, not to fix a bug in the frame. Suppose a project has more than 5 folders in `resources/views` and there are too many views in them, so it can be cumbersome to go to the `resources/views` folder and search for the corresponding folder, and in this case it is easier to do it via terminal. The purpose of the command is to allow us to do this from the terminal, rather than navigating through folders. Also, when we create a `mail` class, we may need to add a view for this class again. A similar scenario occurs.

PR #40132, opened 2021-12-21
select fields for count in the paginate method

dmouse:
Hi!

Currently in the `Model::paginate()` method you can configure the columns to you want to return in the result but those columns are not used for the count operation.

image

In terms of improve the performance in the `count` my idea is add a new argument in the `paginate` method to configure separately the columns for the count and you can use some indexed fields to improve the performance.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -799,13 +799,13 @@ class Builder
      *
      * @throws \InvalidArgumentException
      */
-    public function paginate($perPage = null, $columns = ['*'], $pageName = 'page', $page = null)
+    public function paginate($perPage = null, $columns = ['*'], $pageName = 'page', $page = null, $count = ['*'])
     {
         $page = $page ?: Paginator::resolveCurrentPage($pageName);
 
         $perPage = $perPage ?: $this->model->getPerPage();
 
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -799,13 +799,13 @@ public function pluck($column, $key = null)
      *
      * @throws \InvalidArgumentException
      */
-    public function paginate($perPage = null, $columns = ['*'], $pageName = 'page', $page = null)
+    public function paginate($perPage = null, $columns = ['*'], $pageName = 'page', $page = null, $count = ['*'])
```
GrahamCampbell, line 802: missing phpdoc

GrahamCampbell (changes requested):

src/Illuminate/Database/Eloquent/Builder.php
dmouse: updated

taylorotwell:
This is apparently a performance based PR but no performance benchmarks were included.

dmouse:
@taylorotwell what kind of benchmark test do I need to include?

For example in this case, a query including all fields takes `125ms`
image

But if I count using an indexed field takes `4.43ms`
image

PR #40133, opened 2021-12-22
[8.x] Remove an useless check in Console Application class

oliverquynh:
The local variable `$input` inside the `parseCommand` method of Console Application class is always an instance of either `StringInput` or `ArrayInput`. It's never null (false), so "$input ?? null" is useless. I just remove that check.

### DIFF
--- a/src/Illuminate/Console/Application.php
+++ b/src/Illuminate/Console/Application.php
@@ -210,7 +210,7 @@ class Application extends SymfonyApplication implements ApplicationContract
             $input = new ArrayInput($parameters);
         }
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Incorrect, this is not useless. It only gets set if you enter the `else` branch of the if.

seriquynh:
@GrahamCampbell Could you show me the use-case in which the $input of this method is not set, please? I see it's always set, never null.

seriquynh:
either the block of if is executed, otherwise the block if else is executed. When both of them aren't executed.

PR #40134, opened 2021-12-22
[8.x] Replace property_exists by null coalescing operator

oliverquynh:
There are a lot of `property_exists` calls that can be replaced by using null coalescing operator. In this PR, I just help to replace where possible.


seriquynh:
I will figure out test problems and send a PR later.

PR #40135, opened 2021-12-22
[8.x] add `schedule:clear-mutex` command

bernardwiesner:
Currently there is no way to clear the mutex files created when using withoutOverlap(). The only way to clear them is using `php artisan cache:clear` however this is dangerous since it clears all other cache as well.

If your server restarts while a task is running, the mutex file does not get deleted. Then when the server comes up, it could take up to 24 hours (default value) for your task to re-run because the mutex file still exists. The only way currently to solve this is just running `cache:clear` however this is not optimal. 

In order to delete only mutex files I added a `schedule:clear-mutex` command.

### DIFF
--- a/src/Illuminate/Console/Scheduling/ScheduleClearCacheCommand.php
+++ b/src/Illuminate/Console/Scheduling/ScheduleClearCacheCommand.php
@@ -4,21 +4,21 @@ namespace Illuminate\Console\Scheduling;
 
 use Illuminate\Console\Command;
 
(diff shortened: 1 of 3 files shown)


bernardwiesner:
Adding an option like:
`schedule:clear-mutex {signature?}` would also be useful. I can add that too if there is a willingness for this feature.

bernardwiesner:
just FYI Taylor renamed the method to `schedule:clear-cache`

PR #40137, opened 2021-12-22
[8.x] Added ability to define extra default password rules

ash-jc-allen:
Hi! This PR is only fairly simple but I think could be really useful. It adds the functionality for you to define your own custom validation rules that should be run when using the `Password::defaults()` validation rule.

As far as I could see, this functionality doesn't already exist; but if it does, apologies.

To give a bit of context, I have an app that has 5 different places where users can be created and updated. At the moment, I'm using the password defaults, but also want to add a rule to make sure that every password passes a [zxcvbn](https://github.com/bjeavons/zxcvbn-php) validation check. So my form request rules look similar to this:

```php
public function rules()
{
    return  [
        // ...
       'password' => ['required', Password::defaults(), new ZxcvbnRule()],
        // ...
    ];
}
```

By default, I always want my passwords to pass the `ZxcvbnRule` validation check, so I have this in every place where I'm using `Password::defaults()`. So, in a way, I'm also treating that rule as a default one too.

So, with my proposed change, I can do something like this to define some extra default password rules:

```php
Password::defaults(function () {
    return Password::min(8)
        ->symbols()
        ->mixedCase()
        ->uncompromised()
        ->rules(new ZxcvbnRule());
});
```

This means that I can then clean up my form request rules and be confident that the `ZxcvbnRule()` will always be run wherever the password defaults are used. It could look more like this:

```php

public function rules()
{
    return  [
        // ...
       'password' => ['required', Password::defaults()],
        // ...
    ];
}
```

The new `->rules()` method that I've added supports rule objects and closures and they can be passed in individually or as an array like so:

```php
Password::defaults(function () {
    return Password::min(8)
        ->rules(new ZxcvbnRule());
});
```

```php
Password::defaults(function () {
    return Password::min(8)
        ->rules([new ZxcvbnRule(), new AnotherRule()]);
});
```

```php
Password::defaults(function () {
    return Password::min(8)
        ->rules(function ($attribute, $value, $fail) {
            if ($value !== 'foo') {
                $fail('Password is not foo');
            }
        });
});
```

As far as I can see, this appears to be working. If it's something that you might consider merging, please let me know if you need any changes making 😄

### DIFF
--- a/src/Illuminate/Validation/Rules/Password.php
+++ b/src/Illuminate/Validation/Rules/Password.php
@@ -80,7 +80,7 @@ class Password implements Rule, DataAwareRule, ValidatorAwareRule
     protected $compromisedThreshold = 0;
 
     /**
-     * Custom validation rules that should be used.
+     * Additional validation rules that should be merged into the default rules during validation.
      *
      * @var array
      */
@@ -267,10 +267,10 @@ class Password implements Rule, DataAwareRule, ValidatorAwareRule
     }
 
     /**
-     * Define custom validation rules that should be used.
+     * Specify additional validation rules that should be merged with the default rules during validation.
      *
-     * @param $rules
-     * @return Password
+     * @param  string|array  $rules
+     * @return $this
      */
     public function rules($rules)
     {
--- a/src/Illuminate/Validation/Rules/Password.php
+++ b/src/Illuminate/Validation/Rules/Password.php
@@ -80,6 +80,13 @@ class Password implements Rule, DataAwareRule, ValidatorAwareRule
     protected $compromisedThreshold = 0;
 
     /**
+     * Custom validation rules that should be used.
+     *
+     * @var array
+     */
+    protected $customRules = [];
+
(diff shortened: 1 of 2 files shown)


yusufonur:
I think very good feature, thanks @ash-jc-allen

onlime:
Thanks @ash-jc-allen for this great addition!

I only find the order where the `customRules` are applied rather confusing and in the case of `ZxcvbnRule` not really what I would expect. The custom rules are merged right at the beginning and are evaluated before any `mixedCase`, `letters`, `symbols`, `numbers`, `uncompromised` conditions.

```php
            [$attribute => array_merge(['string', 'min:'.$this->min], $this->customRules)],
```

Could this be improved by moving the custom rules to the bottom of the stack? I find this more intuitive and a custom rule usually is more complex to evaluate, so it should come after the simple built-in ones.

AshishDhamalaAD:
This is really useful. Thanks for contributing. :)

PR #40139, opened 2021-12-22
[8.x] Add the ability to select the count fields in paginate method

dmouse:
Currently in the `Model::paginate()` method you can configure the columns you want to return in the result but these columns are not used for the count operation.

image

In terms of improve the performance in the count process, my idea is add a new argument in the `paginate` method to configure separately the columns for the count and then you can use some indexed fields to improve the performance.

image

another quick test:

image

image

In my test `order_id` and `product_vendor_id` are indexed fields

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -799,13 +799,13 @@ class Builder
      *
      * @throws \InvalidArgumentException
      */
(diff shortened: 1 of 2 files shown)


taylorotwell:
I'm not sure what database you are using, but this doesn't matter on modern versions of MySQL / Postgres. Running an EXPLAIN on count(*) vs count(some_indexed_column) shows that both queries will use an index if available.

Feel free to provide more information if needed.

PR #40143, opened 2021-12-23
[8.x] Added Validation Rule for Base64

hmennen90:
For Fileuploads, e.g. Image Files, it is necessary to upload them in base64. This Rule can be applied to those Fields.


src/Illuminate/Validation/Rules/Base64.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public function passes($attribute, $value)
+    {
+        $result = base64_decode($value, true);
+
+        if($result === false) {
+            return $result;
+        }
+
+        return true;
```
dennisprudlo, line 25: Can't this be reduced to `return base64_decode($value, true) !== false;`?

hmennen90: Can it will be there now ;-)

tests/Validation/ValidationBase64Test.php
```diff
@@ -0,0 +1,28 @@
...
+
+use Illuminate\Support\Str;
+use Illuminate\Validation\Rules\Base64;
+use PHPUnit\Framework\TestCase;
+
+class ValidationBase64Test extends TestCase
+{
+    public function testBase64StringPassValidation()
+    {
+        $string = base64_encode(Str::random());
```
BrandonSurowiec, line 13: It doesn't _really_ matter, but there's no reason not to use hardcoded strings instead of `Str::random()` in your test cases.

`$string = base64_encode('Hello world');`

hmennen90: Is replaced :-)=

taylorotwell:

hmennen90:
Released as Package :-)

https://github.com/mennen-online/laravel-validation-rules

PR #40144, opened 2021-12-23
Bugfix in model:prune, order of argument check is backwards

KreshnikUkmata:
I am using Laravel 8.77.0 and when using ```php artisan model:prune``` it would throw "Undefined variable $except" at line 106.

I looked into it and saw that the file was updated in the master branch but I still think that the command is not working properly.

On line 90 the code would return if the ```$models``` argument was not empty and therefore would not reach the check in line 95 so the order of the conditionals should be flipped. That makes it necessary to define the ```$models``` variable first and then to perform the conditionals.


taylorotwell:
Update to 8.77.1.

PR #40146, opened 2021-12-23
[8.x] Fix Rule::in() to work with Enum values

haxmedia:
## Why

Validation rule `In` throws `TypeError` when we pass Enum cases from PHP 8.1:

`Rule::in(SomeEnumObject::cases())`

## Error

`TypeError: str_replace(): Argument #3 ($subject) must be of type array|string, \App\Enums\SomeEnumObject given in /vendor/laravel/framework/src/Illuminate/Validation/Rules/In.php:43`

### DIFF
--- a/src/Illuminate/Validation/Rules/In.php
+++ b/src/Illuminate/Validation/Rules/In.php
@@ -37,7 +37,7 @@ class In
     public function __toString()
     {
         $values = array_map(function ($value) {
-            if ($value instanceof \UnitEnum){
+            if ($value instanceof \UnitEnum) {
                 $value = $value->value;
             }
             return '"'.str_replace('"', '""', $value).'"';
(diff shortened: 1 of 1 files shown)


haxmedia (commented):
Style fix

src/Illuminate/Validation/Rules/In.php
```diff
@@ -37,6 +37,10 @@ public function __construct(array $values)
     public function __toString()
     {
         $values = array_map(function ($value) {
+            if ($value instanceof \UnitEnum) {
```
mateusjunges, line 40: You should use FQN here

taylorotwell:
Just use the actual `enum` validation rule: https://laravel.com/docs/8.x/validation#rule-enum

haxmedia:
@taylorotwell Enum validation rule is not 100% what I want to achieve. This validation check if value fits in full enum. In some cases, not all enum cases will be expected. For example, you can create public method inside enum, that return array of not all, but some cases.

PR #40147, opened 2021-12-23
[8.x] SimpleMessage: Add method to add multiple lines as once

andershagbard:
Not sure if we can change the phpdoc to:

```php
@param  mixed[]  $lines
```

### DIFF
--- a/src/Illuminate/Notifications/Messages/SimpleMessage.php
+++ b/src/Illuminate/Notifications/Messages/SimpleMessage.php
@@ -160,7 +160,7 @@ class SimpleMessage
     /**
      * Add lines of text to the notification.
      *
-     * @param  array  $lines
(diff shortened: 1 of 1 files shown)


src/Illuminate/Notifications/Messages/SimpleMessage.php
```diff
@@ -157,6 +157,21 @@ public function line($line)
         return $this->with($line);
     }
 
+    /**
+     * Add a line of text to the notification.
+     *
+     * @param  mixed  $lines
```
mateusjunges, line 163:
```suggestion
     * @param  array  $lines
```

mateusjunges (commented):
Mixed can be string, which will resut in a php warning:

```
PHP Warning:  Invalid argument supplied for foreach()
```

PR #40155, opened 2021-12-24
[8.x] Add serial property to loop variable for pagination

shibbirweb:
Currently, there is no default way to show serial numbers in `@foreach` or `@forelse` loop variable in pagination. As you can see in the below screenshot.

Screenshot 2021-12-25 032358

So, I add a new property for the loop variable '**serial**'. This property shows the current serial number in pagination.
```
@foreach($users as $user)

    <tr>
        <th scope="row" class="text-center">{{ $loop->serial }}</th>
        <td>{{ $user->name }}</td>
        <td>{{ $user->email }}</td>
    </tr>

@endforeach
```
So that it can show the current item serial in pagination as you can see in the below screenshot.
Screenshot 2021-12-25 032156

Note:  Using `CursorPaginator` this `$loop->serial` works like the `$loop->iteration` property because in `CursorPaginator` currently no way to get the current page number.

### DIFF
--- a/src/Illuminate/View/Concerns/ManagesLoops.php
+++ b/src/Illuminate/View/Concerns/ManagesLoops.php
@@ -3,8 +3,8 @@
 namespace Illuminate\View\Concerns;
 
 use Countable;
-use Illuminate\Support\Arr;
 use Illuminate\Pagination\AbstractPaginator;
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #40157, opened 2021-12-25
[8.x] Append data in JsonResource::collection

crazynds:
Hi, I have a project in laravel and I have organized all my modals with a controller, request, resource, migration, policy and other standard files, but I ran into a problem where I needed to add a dataset to the collection's json and to not go out of the pattern and create another Resource class I created a small code snippet in which I can add additional fields in JsonResource when I run collection with it.
I would like to share this problem and the solution and I believe it would be very useful to add to the main library.


taylorotwell:

PR #40158, opened 2021-12-25
Add Str::conact helper to concatenate strings

James4645:
`Str::concat` is a new string helper I am proposing to make concatenating strings easier.

**The Problem:**
I find myself regularly concatenating strings together, often with variables interleaved in a string, and I end up with some mess like:

`return 'Thank you, ' . $user->name . ' for your entry titled ' . $entry->title . '. It is now live.';`

You see the problem, its not very readable, Each `.` you must closely look at to determine if its a period to end a sentence or if it is a concatenation operator. Now, may be thinking what about string interpolation. It is true, you can do something like:

`return "Thank you, {$user->name} for your entry titled {$entry->title}. It is now live.";`

And while that certainly solves the readability problems, its less ideal when you are already concatenating to build a string in parts.

`Str::concat` is a proposal to add a string helper to the mix. Simply provide an array for each part of the string. And as an optional second parameter include a separator (its space by default, but you can change to anything, such as "-", "_", or "" [empty string]) or anything else you can think of.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -170,6 +170,24 @@ class Str
 
         return static::$camelCache[$value] = lcfirst(static::studly($value));
     }
+    
+    /**
+     * Concatenate an array of strings together with an optional seperator.
+     *
+     * @param  array  $strings
+     * @param  string|null  $separator
+     * @return string
+     */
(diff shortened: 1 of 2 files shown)


foremtehan:
Use `Str::append()`

taylorotwell:

PR #40159, opened 2021-12-25
[8.x] Add rails style column definitions in make migration command options

Abdul-fattah-Tayih:
Rails has a very nice [feature](https://guides.rubyonrails.org/active_record_migrations.html#creating-a-migration) which enables you to define columns in the cli when creating a migration, this cuts down time and makes development more snappy.

Especially when creating a quick update or a POC migration, this change will help make it that much quicker to be up and running.

My goal is to reach this:

```console
user@device:~$ php artisan make:migration create_news_table --c=title:string --c=body:text --c=hashed_id:string:index
```

and the result would be:

```php
Schema::create('news', function (Blueprint $table) {
    $table->id();
    $table->string('title');
    $table->text('body');
    $table->string('hashed_id')->index();
    $table->timestamps();
});
```

This change won't break previous version because it adds new functionality to `MigrationMakeCommand` and `MigrationCreator` classes and in case the column definitions were empty, then nothing will happen.

### DIFF
--- a/src/Illuminate/Database/Console/Migrations/MigrateMakeCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/MigrateMakeCommand.php
@@ -152,7 +152,6 @@ class MigrateMakeCommand extends BaseCommand
                 abort(0);
             }
 
-
             $columns[] = new MigrationLine($values[0], $values[1], array_slice($values, 2));
         }
 
--- a/src/Illuminate/Database/Console/Migrations/MigrateMakeCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/MigrateMakeCommand.php
@@ -3,8 +3,11 @@
 namespace Illuminate\Database\Console\Migrations;
 
 use Illuminate\Database\Migrations\MigrationCreator;
+use Illuminate\Database\Migrations\MigrationLine;
+use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Composer;
 use Illuminate\Support\Str;
+use InvalidArgumentException;
 
 class MigrateMakeCommand extends BaseCommand
 {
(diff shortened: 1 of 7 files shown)


src/Illuminate/Database/Console/Migrations/MigrateMakeCommand.php
```diff
@@ -132,4 +136,26 @@ protected function getMigrationPath()
...
+    {
+        $columns = [];
+        foreach ($this->input->getOption('columns') ?? [] as $column) {
+            $values = explode(':', $column);
+            if (count($values) < 2) {
+                $this->error("$column is invalid, the format is name:type");
+                abort(0);
+            }
+
+
```
dammy001, line 155: i think you should remove trailing spaces

shadoWalker89:
I'm not against your PR, but I find writing the columns using the power of auto complete more robust (hard to make a typo using auto complete) and time efficient.

src/Illuminate/Database/Console/Migrations/MigrateMakeCommand.php
Abdul-fattah-Tayih: I've removed the extra empty line

Abdul-fattah-Tayih:
> I'm not against your PR, but I find writing the columns using the power of auto complete more robust (hard to make a typo using auto complete) and time efficient.

You're right in that regard, autocomplete within your IDE/text editor would be more reliable, but there's a check in the implementation to see if the given method exists or not, so the command will tell you if your input was wrong, and personally I see this feature improving developer experience, for example you can use `make:controller Test` and add the usual crud methods manually or use `make:controller Test -r` to have these methods stubbed out for you, I believe this goes the same way.

taylorotwell:

Issue #32979, opened 2020-05-26
Mailable::mailer not used when sending


BnitoBzh:
- Laravel Version: 7.12.0
- PHP Version: 7.3.18
- Database Driver & Version: pgsql

### Description:
Mailable has a mailer property that represents the mailer that should send the message.
But it is not used to send the message.

### Steps To Reproduce:
Prerequisite : Configure 2 mailers (smtp / mailgun) with smtp as default.

Create a simple basic TestMail class that extends the Mailable class.
```php
<?php
namespace App\Mail;
use Illuminate\Mail\Mailable;
class TestMail extends Mailable
{
    public function __construct()
    {
        $this->mailer = 'mailgun';
    }

    public function build()
    {
        // The test-mail.blade.php file simple contains 'hello world'
        return $this->text('mails.test-mail');
    }
}
```
Try to send the TestMail with the Mail facade
```php
Mail::to('email@eample.com')->send(new TestMail);
```

The TestMail is sent with the smtp mailer instead of the mailgun.

Trying to configure the mailer inside the build method has no effect.
```php
    // ...
    public function build()
    {
        // The test-mail.blade.php file simple contains 'hello world'
        return $this->mailer('mailgun')->text('mails.test-mail');
    }
```

If we specify the mailer from the Facade the mailgun mailer is used.
```php
Mail::mailer('mailgun')->to('email@eample.com')->send(new TestMail);
```

koenhoeijmakers:
Seems that the given `mailer` property is only used when no `mailer` has been selected as it only uses this property when the `Factory` is passed instead of a `Mailer`.

BnitoBzh:
Ok, so how to transform this basic usage to make sure the `TestMail` mailer will be used correctly?
```php
Mail::to('email@eample.com')->send(new TestMail);
```

paul-thebaud:
Hi!
I think the `mailer` property should be used and override default mailer.
This way, it will work as same as the `connection` property on Eloquent Models, or the `queue` property on Jobs.

BnitoBzh:
The `Mail` facade always returns the default mailer `Mail::to(...)->send(...)` if no mailer is specified `Mail::mailer(...)->to(...)->send(...)`.

The `Mail` facade (which returns the mail.manager) has no method for using the mailer property of the sent mail (`TestMail`).

taylorotwell:
I'm not sure we ever document this behavior or ability.

vesper8:
[...]

PR #40161, opened 2021-12-25
Dedicated mailer support for Mailables

ReDev1L:
With this change we will be able to use dedicated mailer that was set in Mailable, instead of overriding it.
Solves this issue: https://github.com/laravel/framework/issues/32979#issuecomment-927112009
Doesnt break anything.

### DIFF
--- a/src/Illuminate/Mail/Mailer.php
+++ b/src/Illuminate/Mail/Mailer.php
@@ -299,9 +299,11 @@ class Mailer implements MailerContract, MailQueueContract
      */
     protected function sendMailable(MailableContract $mailable)
     {
+        $mailer = $mailable->mailer ?? $this->name;
+
         return $mailable instanceof ShouldQueue
-                        ? $mailable->mailer($this->name)->queue($this->queue)
-                        : $mailable->mailer($this->name)->send($this);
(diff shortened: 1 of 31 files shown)


src/Illuminate/Mail/Mailer.php
```diff
@@ -299,9 +299,10 @@ public function send($view, array $data = [], $callback = null)
      */
     protected function sendMailable(MailableContract $mailable)
     {
+        $mailer = $mailable->mailer ?? $this-name;
```
ReDev1L, line 302:
```suggestion
        $mailer = $mailable->mailer ?? $this->name;
```

src/Illuminate/Mail/Mailer.php
```diff
@@ -299,9 +299,10 @@ public function send($view, array $data = [], $callback = null)
      */
     protected function sendMailable(MailableContract $mailable)
     {
+        $mailer = $mailable->mailer ?? $this->name;
         return $mailable instanceof ShouldQueue
```
ReDev1L, line 303:
```suggestion
        $mailer = $mailable->mailer ?? $this->name;
        
        return $mailable instanceof ShouldQueue
```

PR #40162, opened 2021-12-25
Dedicated mailer support for Mailables

ReDev1L:
With this change we will be able to use dedicated mailer that was set in Mailable, instead of overriding it.
Solves this issue: https://github.com/laravel/framework/issues/32979 and doesnt break anything.


taylorotwell:
https://laravel.com/docs/8.x/mail#sending-mail-via-a-specific-mailer

You can wrap this logic behind your own method that receives a mailable, etc.

PR #40163, opened 2021-12-25
[8.x] Removed shrink-to-fit in error view

Erulezz:
This tag isn't necessary anymore. This was needed with iOS 9, until iOS 9.3 restored the normal behaviour.

This PR brings it in sync with the other error template files.


taylorotwell:
If it's not broken I don't plan to change this.

Erulezz:
This is the only file where it's in the meta tag, even the default [Laravel welcome view](https://github.com/laravel/laravel/blob/8.x/resources/views/welcome.blade.php) doesn't have it.

It's not 'broken', but it does absolutely nothing also.

PR #40164, opened 2021-12-25
[8.x] Hide or exclude pagination fields [links/meta] in Resources or Collections

hossein-zare:
I had to create collections for my api resources to override the `paginationInformation` method to hide some of the pagination array fields while that was extra work.

With this PR they can be hidden without the need for Collections.
It doesn't matter if you're in a `Resource` or `Collection`, it works both ways.

**Example:**
```php
class TransactionCollection extends ResourceCollection
{
    public $hidden = []; // hides nothing

    public $hidden = ['links', 'meta']; // hides both

    public $hidden = [
        'links' => ['first'], // excludes first
    ];

    public $hidden = [
        'meta' => [], // excludes nothing
        'links' // hides links
    ];

    ...
}

class TransactionResource extends JsonResource {
    ...
}
```


taylorotwell:

PR #40166, opened 2021-12-26
[6.x] Add varbinary type to Schema.

kudashevs:
## Motivation

According to [this]( https://laracasts.com/discuss/channels/eloquent/varbinary-equivalent-in-eloquent) and [this]( https://stackoverflow.com/questions/17795517/laravel-4-saving-ip-address-to-model) there is a convenient way to store IP addresses through inet_pton() and inet_ntop() and retain them in a database in a VARBINARY(16) column. It's efficient and suitable for both IP versions. Unfortunately, I haven’t found a way how it is possible to make a columns varbinary. So, I started to seek a solution. The first one was to use:
```code
DB::statement('ALTER TABLE `my_logs` ADD `ip_address` VARBINARY(16)');
```

But it’s not flexible, because it’s impossible to set the exact order of the column in the table. So, the next solution was almost the same, but without this problem. It is possible to create a column of any type and then modify its type.
```code
DB::statement('ALTER TABLE `my_logs` MODIFY COLUMN `ip_address` VARBINARY(16)');
```

Unfortunately, it turned out, that SQLite (which I use during testing) doesn’t support the modify command. I searched for PRs on this topic and found only #39461, which was closed.

## Proposition

This is my PR with a proposition to add the possibility to create a column of the varbinary type using Schema.

**P.S.** This solution doesn’t get any additional options (because binary() method does it neither). In case, if you need to provide any additional information to the database it is possible to do that through addColumn() method.
```code
$table->addColumn('varbinary', 'ip_address', ['length' => 16]);
```

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -1067,6 +1067,17 @@ class Blueprint
     }
 
     /**
+     * Create a new varbinary column on the table.
+     *
+     * @param  string  $column
+     * @return \Illuminate\Database\Schema\ColumnDefinition
+     */
+    public function varbinary($column)
+    {
+        return $this->addColumn('varbinary', $column);
+    }
+
+    /**
      * Create a new uuid column on the table.
      *
      * @param  string  $column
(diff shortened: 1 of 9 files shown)


driesvints:
I'm sorry but 6.x is closed for new features.

PR #40177, opened 2021-12-27
PhpDoc for Log. Message can be mixed type

adiafora:
Hello. 
I just added the types to the phpDoc for `Illuminate\Log`. The `message `parameter can be not only a string, but also an array. In addition, in many methods it is declared as `mixed `(for example, in `Illuminate\Log\Logger::FormatMessage`). 
The advantage for development is that the IDE does not highlight the code when a variable other than a string is passed.

### DIFF
--- a/src/Illuminate/Log/LogManager.php
+++ b/src/Illuminate/Log/LogManager.php
@@ -532,7 +532,7 @@ class LogManager implements LoggerInterface
     /**
      * System is unusable.
      *
-     * @param  string  $message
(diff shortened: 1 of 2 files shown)


adiafora:
How can I fix the incompatibility with the PHP 7.4 version in the phpDoc? )))

GrahamCampbell:
I don't think this is right. We could possibly widen this to what we did on 9.x though, which is larger than string, but smaller than mixed. cc @driesvints.

PR #40179, opened 2021-12-27
add package_view helper function

monaye:
Since the `view` helper function only accepts the string format for the path, to specify the path to the package's view file, it will require concatenation with special sting `::` for package and `.` syntax for the path. 

since we want to declare package slug variable during the package development to avoid string typo, most likely will end up a lot of the following code

```
view(MyPackageServiceProvider::$slug . '::' . 'viewto.my.index');
```

this new helper function help developer writer better readability and a cleaner way to specify view file from the package

```
package_view(MyPackageServiceProvider::$slug, 'viewto.my.index');
```

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -932,3 +932,18 @@ if (! function_exists('view')) {
         return $factory->make($view, $data, $mergeData);
     }
 }
+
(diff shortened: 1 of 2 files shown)


monaye:
FYI: planing to update the docs once this is accepted

taylorotwell:

PR #40180, opened 2021-12-27
[8.x] Add a default timeout in the HTTP client

themsaid:
## Update

Moved to the 9.x release: https://github.com/laravel/framework/pull/40187

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -162,7 +162,7 @@ class PendingRequest
 
         $this->options = [
             'http_errors' => false,
-            'timeout' => 10
+            'timeout' => 10,
         ];
 
         $this->beforeSendingCallbacks = collect([function (Request $request, array $options, PendingRequest $pendingRequest) {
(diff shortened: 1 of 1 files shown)


src/Illuminate/Http/Client/PendingRequest.php
```diff
@@ -162,6 +162,7 @@ public function __construct(Factory $factory = null)
 
         $this->options = [
             'http_errors' => false,
+            'timeout' => 10,
```
GrahamCampbell, line 165: 10 seconds is probably too low for a lot of people. I'd recommend going with:

```
'connect_timeout' => 10,
'timeout' => 15,
```

like in https://github.com/GrahamCampbell/Guzzle-Factory/blob/5.0/src/GuzzleFactory.php.

GrahamCampbell (changes requested):

mabdullahsari:
Please target 9.x. 

This change will break an app I work on because it has API calls that last 3 mins (I did not design it 🙃).

SjorsO:
A default timeout is a great idea. I've had queue workers get stuck because HTTP calls timed-out in a weird way (even though supervisor reported they were working fine).

leonardovee:
A default timeout is sure a good idea, but shouldn't this be treat as a breaking change?

taylorotwell:
@themsaid I agree we should probably send this to 9.x / master.

PR #40190, opened 2021-12-28
[8.x] Add ability to supply HTTP client methods with `Arrayable` instances

stevebauman:
## Problem

Currently we cannot supply `Arrayable` instances into any of the `HTTP` methods, as Guzzle will throw an exception, saying that HTTP options must be an array or string.

## Description

This PR introduces the ability to supply `Arrayable` instances into the Laravel HTTP Client for more convenient use.

This is done by simply leveraging `collect($options)->toArray()` to transform all HTTP options that are `Arrayable` to their raw array format, prior to sending them along to Guzzle.

## Usage

Using a collection for query parameters:

```php

$query = collect(['brand' => 'Samsung', 'color' => 'Black']);

Http::get('/products', $query);
```

Using a fluent instance for form data:

```php
use Illuminate\Support\Fluent;

$user = new Fluent();

$user->name = 'Steve Bauman';
$user->email = 'steve@email.com';

//...

Http::post('/users', $user);
```

The above use-cases would be more useful when their `Fluent` or `Collection` instances coming from other sources in your application, so you don't have to transform them every time you want to send a request.

## Important

I wasn't sure if I should change the `array` doc block types to `mixed`, or simply add `Arrayable` as an allowed type.

Let me know what you'd like, thanks!

---

Thanks for your time :heart:. No hard feelings on closure.

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -16,7 +16,6 @@ use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
 use Psr\Http\Message\MessageInterface;
 use Symfony\Component\VarDumper\VarDumper;
-use Illuminate\Contracts\Support\Arrayable;
 
 class PendingRequest
 {
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -16,6 +16,7 @@ use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
 use Psr\Http\Message\MessageInterface;
 use Symfony\Component\VarDumper\VarDumper;
+use Illuminate\Contracts\Support\Arrayable;
 
 class PendingRequest
 {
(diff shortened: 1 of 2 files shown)


taylorotwell:
Is changing this signature a breaking change?

stevebauman:
Dang, yes it is a breaking change. PHP will throw an incompatibility exception. I don't think there's any way around it, even if we update the type to `mixed`, it still will throw an exception with someone having the `array` type declared on their extended version.

I can move this to the 9.x branch if you'd like?

src/Illuminate/Http/Client/PendingRequest.php
```diff
@@ -570,7 +570,7 @@ public function head(string $url, $query = null)
      * @param  array  $data
      * @return \Illuminate\Http\Client\Response
      */
-    public function post(string $url, array $data = [])
+    public function post(string $url, $data = [])
```
GrahamCampbell, line 573: breaking change

GrahamCampbell (changes requested):

taylorotwell:
Yeah if we make such a change it would have to go to master I guess.

PR #40192, opened 2021-12-29
[9.x] Automatically resolve route parameters into view data

jasonmccreary:
Recently, I attempted to register the following route:

```php
Route::view('event/{event}/sign-in', 'event.sign-in');
```

While I didn't _expect_ this to work, the magic within Laravel has surprised me numerous times.

Ultimately, I had to write:

```php
Route::get('event/{event}/sign-in', function (\App\Models\Event $event) {
    return view('event.sign-in', ['event' => $event]);
});
```

However, I think it'd be pretty neat if Laravel had this magic. So that's what this PR adds - resolving route parameters and automatically injecting them into the view data.

**How it works**
Any explicit route parameter bindings are resolved like any other route. Implicit model bindings are _guessed_ using a fallback similar to how model factories are guessed.

If view data with these names is not already defined, the parameter will be automatically injected into the view data. Said another way, values passed in the `$data` argument take precedence over parameters.

---

**Note:** this is targeted for Laravel 9 due to the change in default behavior of `Route::view()`. To be compatible with Laravel 8 or to make the behavior opt-in, a chained method could be used. For example, `Route::view()->withParameterBinding();`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -119,7 +119,8 @@
     },
     "autoload-dev": {
         "files": [
-            "tests/Database/stubs/MigrationCreatorFakeMigration.php"
+            "tests/Database/stubs/MigrationCreatorFakeMigration.php",
+            "tests/Integration/Routing/Fixtures/User.php"
         ],
         "psr-4": {
             "Illuminate\\Tests\\": "tests/"
--- a/src/Illuminate/Routing/ImplicitRouteBinding.php
+++ b/src/Illuminate/Routing/ImplicitRouteBinding.php
@@ -77,7 +77,7 @@ class ImplicitRouteBinding
     /**
      * Guess the class from the parameter name.
      *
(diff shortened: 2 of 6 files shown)


taylorotwell:
Personally I think I'll hold off on this. I don't love that we have to reach out to the global `app` helper from deep in the routing layer to try to guess the class names, etc.

Can kinda already be a one-liner like this... 😅

```php
Route::get('/event/{event}/sign-in', fn (Event $event) => view('event.sign-in', compact('event')));
```

jasonmccreary:
Sure. With Laravel 9's requirement of PHP 8 you could definitely map them explicitly using a type-hinted, short closure.

I thought the implicit mapping was a nice bit of automagic that would work for most apps. However, completely understand there's probably a better implementation.

PR #40193, opened 2021-12-29
[8.x] Add shortcut for searching in Postgres Array (`[]`) type

yakatz:
Similar to `whereJsonContains()`, this makes it easier to search in columns with `[]` types. I am not sure which databases support this besides Postgres.


taylorotwell:

yakatz:
@taylorotwell How is this different that the `JsonContains` functions? Since this is a core part of the grammar, I don't see how it could easily be included in a package and it seems to be a simple addition to the framework.

PR #40198, opened 2021-12-29
[9.x] Remove unnecessary `strtolower` calls in Encrypter

Krisell:
Since lowercase cipher names are used when calling `openssl`, but the Laravel configuration file uses uppercase, the cipher name is passed through `strtolower`. This PR performs this as part of the Encryptor constructor instead of for every internal usage, and will save some unnecessary work. 

The public methods with `$cipher` as an argument still call `strtolower` to preserve BC. There is however the possibility that third-party code that extends Encryptor peeks at or modifies `$cipher` directly, and might need to be updated to use lowercase cipher names after this PR. If this is an unnecessary risk, I'm all OK with this PR being ignored and the extra work considered acceptable. Perhaps I am missing some other nuance of BC?

Of course, another solution is to require the use of lowercase in the configuration file and all usage, enabling us to skip all `strtolower` calls. However it's quite inconvenient to force everyone to update their configuration files to save just a few cycles.

### DIFF
--- a/src/Illuminate/Encryption/Encrypter.php
+++ b/src/Illuminate/Encryption/Encrypter.php
@@ -56,7 +56,7 @@ class Encrypter implements EncrypterContract, StringEncrypter
         }
 
         $this->key = $key;
-        $this->cipher = $cipher;
+        $this->cipher = strtolower($cipher);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


driesvints (approved):

taylorotwell:
I think it's fine as is.

PR #40202, opened 2021-12-29
[8.x] Extend PostgreSQL updateFrom to support nested joins

peanutsmcgee:
### Description:
This PR builds upon the recent addition of the `updateFrom` method added in commit [#re-add update](https://github.com/laravel/framework/commit/64421305c9e31dc5d909cc96a053e2bf949f7f2b) to include nested joins.

Update statements in PostgreSQL require the use of a `from` clause when joining tables, but regular `join` clauses are allowed as long as at least one `from` clause is set. This is especially useful for including `left join` statements, as `from` always acts as an `inner join`.

There is also the matter of code consistency. Using `->update()` will include all nested joins in the compiled query, but using `->updateFrom()` ignores any nested joins.

### Example of Issue:
Given the query...
```
DB::table('table1')
	->join('table2', function ($join) {
		$join->on('table1.id', '=', 'table2.fk')
			->leftJoin('table3', 'table2.id', '=', 'table3.fk');
	})
	->updateFrom(['name' => 'newName']);
```
The following statement is output (missing `table3`)
```
update "table1"
set "name" = 'newName'
from "table2"
where "table1"."id" = "table2"."fk"
```
But using `->update()` outputs the following (includes `table3`)
```
update "table1"
set "name" = 'newName'
where "ctid" in (select "table1"."ctid"
                 from "table1"
                          inner join ("table2" left join "table3" on "table2"."id" = "table3"."fk")
                                     on "table1"."id" = "table2"."fk");
```

### Proposed Solution:
After the changes made to the `compileUpdateFrom` method in `PostgresGrammar.php`, the above query outputs as follows. 
```
update "table1"
set "name" = 'newName'
from "table2"
         left join "table3" on "table2"."id" = "table3"."fk"
where "table1"."id" = "table2"."fk"
```

### DIFF
--- a/src/Illuminate/Database/Query/Grammars/PostgresGrammar.php
+++ b/src/Illuminate/Database/Query/Grammars/PostgresGrammar.php
@@ -291,6 +291,8 @@ class PostgresGrammar extends Grammar
                 $from = ' from '.implode(', ', $froms);
             }
 
+            // Nested joins must be compiled out separately and added in as normal
+            // join clauses.
             $joinFroms = collect($query->joins)->filter(function ($join) {
                 return collect($join->joins)->count();
             })->map(function ($join) {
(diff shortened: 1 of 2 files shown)


driesvints:
@peanutsmcgee can you add a test please?

peanutsmcgee:
> @peanutsmcgee can you add a test please?

Sorry i'm new to this. Do you mean like make additons to `DatabaseQueryBuilderTest.php`, or give more examples of queries I've used to test this?

peanutsmcgee:
> @peanutsmcgee can you add a test please?

I've appended an addition statement to the existing `testUpdateFromMethodWithJoinsOnPostgres` function. Not sure if that was the correct way to implement this test, or if I should have created an entirely new test function for it. Please let me know if I should change it.

peanutsmcgee:
Found an issue with the order of bindings when using `where` statements within a nested join. Added a test case to show for the issue. I'm looking for a solution now. Will close this PR if i can't find a solution.

peanutsmcgee:
@driesvints So the root of the issue seems comes from `prepareBindingsForUpdateFrom()` when bindings for the join statements are moved to match the move from a join+on statement to a from+where statement. The problem is that only the base join statements are converted to from statements, while ALL of the bindings are moved. 

Ideally I'd code it to only re-order the needed bindings, but I couldn't find a good way to differentiate the bindings of base join statements from the bindings of nested join statements. 

As an alternative, I found that reversing there where statements from `$baseWheres . $joinWheres` to `$joinWhere . $baseWheres` seeemed to resolve bindings mismatch issue. 

I'm not sure if this is a valid way to resolve this issue. I also worry there will be more issues when compiling more complicated queries (i'm still testing this). It may be for the better that nested joins are not allowed in `updateFrom` statements, just to avoid the complications that arise with rewriting the query.

driesvints:
Yeah I think it's best that we avoid any breaking changes here. Still thanks for checking in on this 👍

PR #40206, opened 2021-12-30
[8.x] Add new function called `except()` to HasAttributes trait

Solomon04:
This new `except()` function serves as an inverse to the `only()` method

### Background
In my past Laravel applications, I've extended the BaseModel and added an `except()` function to exclude specific attributes of a model. It differs from the `$hidden` attribute because `except()` method is for one-off use cases. 

```
public function showUserInformation($userID): Response {
    // Don't show sensitive information unless it belongs to the user
    return Auth::id() === $userID ? User::find($userID):   User::find($userID)->except('email', 'phone_number')
}

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1744,6 +1744,8 @@ trait HasAttributes
     }
 
     /**
+     * Exclude specific attributes from a model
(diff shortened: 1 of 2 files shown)


taylorotwell:
Could be a breaking change if people have already defined this on a base model as you have.

bkilshaw:
This seems like a great change; any chance it could be added into 10.x and listed as a breaking change?

PR #40207, opened 2021-12-30
[9.x] Use `str_contains` instead of `strpos`

selcukcukur:


dinhquochan:
👎  this PR because Laravel 8.x required `7.3` that version does not have `str_contains()` function.

selcukcukur:
> 👎 this PR because Laravel 8.x required `7.3` that version does not have `str_contains()` function.


It can be used with the `symfony/polyfill-php80` package. The `str_contains` method on php 7.x.

GrahamCampbell:
This PR is not for Laravel 8. It's for Laravel 9, which is PHP 8.0.2+ only.

dammy001:
Exactly @GrahamCampbell

PR #40208, opened 2021-12-30
[8.x] Add ability to customize json options on JsonResource response

bastien-phi:
Currently, it is possible to customize the options with `JsonResource::withResponse($request, $response)` doing something like
```php
public function withResponse($request, $response)
{
    $response->setEncodingOptions(JSON_PRETTY_PRINT);
}
```

The behaviour is that the original json is decoded then re-encoded with the given options https://github.com/laravel/framework/blob/676ba54b7d62eb76f0d1b025173b8f7134ed8b78/src/Illuminate/Http/JsonResponse.php#L119-L124
https://github.com/laravel/framework/blob/676ba54b7d62eb76f0d1b025173b8f7134ed8b78/src/Illuminate/Http/JsonResponse.php#L63-L66
https://github.com/laravel/framework/blob/676ba54b7d62eb76f0d1b025173b8f7134ed8b78/src/Illuminate/Http/JsonResponse.php#L73-L92

Unfortunately, json_encode with no option is destructive : 
`json_decode(json_encode(['float' => 3.0]), true)` gives `['float' => 3]`.

With this, `$response->setEncodingOptions(JSON_PRESERVE_ZERO_FRACTION);` returns a json like 
```json
{"data":{"float":3}}
```

That is, there is currently no way to encode JsonResources with `JSON_PRESERVE_ZERO_FRACTION`.

This PR adds the ability to customize json serialization options for responses from JsonResource at the time that the JsonResponse is created, providing the ability to preserve the zero fraction of floats.

<details>
<summary>Why do you need JSON_PRESERVE_ZERO_FRACTION btw ?</summary>

Imagine a Order model with a `quantity` property as float and you want to test some simple api call like 

```php
public function it_returns_the_order() 
{
    $order = Order::factory()->createOne();
    
    $this->getJson("api/orders/{$order->id}")
        ->assertOk()
        ->assertJsonPath('data', [
            'id' => $order->id,
            'quantity' => $order->quantity,
        ]);
}
```

Pretty easy no ? 

Well, if your factory is defined with
```php
public function definition()
{
    return [
        'quantity' => $this->faker->randomFloat(1, 0, 10), // random float with 1 digit max between 0 and 10
    ];
}
```
and faker generates a float with 1 digit, your test will pass. But if faker generates `3.0` for example, the test will fail with 

```
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
-  'quantity' => 0.0
+  'quantity' => 0
````

Quite annoying to have randomly failing tests...

</detail>

### DIFF
--- a/src/Illuminate/Http/Resources/Json/JsonResource.php
+++ b/src/Illuminate/Http/Resources/Json/JsonResource.php
@@ -165,6 +165,16 @@ class JsonResource implements ArrayAccess, JsonSerializable, Responsable, UrlRou
     }
 
     /**
+     * Get the serialization options applied to the resource response.
+     *
+     * @return int
+     */
+    public function jsonOptions()
+    {
+        return 0;
+    }
+
+    /**
      * Customize the response for a request.
      *
      * @param  \Illuminate\Http\Request  $request
--- a/src/Illuminate/Http/Resources/Json/JsonResource.php
+++ b/src/Illuminate/Http/Resources/Json/JsonResource.php
@@ -165,7 +165,7 @@ class JsonResource implements ArrayAccess, JsonSerializable, Responsable, UrlRou
     }
 
     /**
-     * Get the serialization options applied to the resource response.
+     * Get the JSON serialization options that should be applied to the resource response.
      *
      * @return int
      */
--- a/tests/Integration/Http/Fixtures/PostResourceWithJsonOptions.php
+++ b/tests/Integration/Http/Fixtures/PostResourceWithJsonOptions.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Illuminate\Tests\Integration\Http\Fixtures;
+
+use Illuminate\Http\Resources\Json\JsonResource;
+
+class PostResourceWithJsonOptions extends JsonResource
+{
+    public function toArray($request)
+    {
+        return [
+            'id' => $this->id,
+            'title' => $this->title,
+            'reading_time' => $this->reading_time,
+        ];
+    }
+
+    public function jsonOptions()
+    {
+        return JSON_PRESERVE_ZERO_FRACTION;
+    }
+}
--- a/src/Illuminate/Http/Resources/CollectsResources.php
+++ b/src/Illuminate/Http/Resources/CollectsResources.php
@@ -55,18 +55,7 @@ trait CollectsResources
     }
 
     /**
-     * Get an iterator for the resource collection.
-     *
-     * @return \ArrayIterator
-     */
-    #[\ReturnTypeWillChange]
-    public function getIterator()
-    {
-        return $this->collection->getIterator();
-    }
-
-    /**
-     * Get the serialization options applied to the resource response.
+     * Get the JSON serialization options that should be applied to the resource response.
      *
      * @return int
      */
@@ -75,5 +64,16 @@ trait CollectsResources
         $collects = $this->collects();
 
         return $collects ? (new $collects([]))->jsonOptions() : 0;
+    }
+
+    /**
+     * Get an iterator for the resource collection.
+     *
+     * @return \ArrayIterator
+     */
+    #[\ReturnTypeWillChange]
+    public function getIterator()
+    {
+        return $this->collection->getIterator();
     }
 }
--- a/src/Illuminate/Http/Resources/CollectsResources.php
+++ b/src/Illuminate/Http/Resources/CollectsResources.php
@@ -64,4 +64,16 @@ trait CollectsResources
     {
         return $this->collection->getIterator();
     }
+
+    /**
+     * Get the serialization options applied to the resource response.
+     *
+     * @return int
+     */
+    public function jsonOptions()
+    {
+        $collects = $this->collects();
+
+        return $collects ? (new $collects([]))->jsonOptions() : 0;
+    }
 }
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by taylorotwell** [comment]
What about resource collections? Does anything need to be done for them?

**Comment by bastien-phi** [comment]
No, no change needed for ResourceCollection.

But you make me remember that the same change must be done in PaginatedResourceResponse
 https://github.com/laravel/framework/blob/8.x/src/Illuminate/Http/Resources/Json/PaginatedResourceResponse.php#L15

Currently afk, I will make the change tomorrow.

**Comment by bastien-phi** [comment]
I will reopen the PR once the changes are pushed.

**Comment by bastien-phi** [comment]
Well, you were right, some changes were needed for ResourceCollections... Let me know if something looks wrong

**Comment by humaneguy** [comment]
Really good work @bastien-phi

File: src/Illuminate/Http/Resources/Json/JsonResource.php [hunk 1]
```diff
@@ -164,6 +164,16 @@ public function additional(array $data)
...
     }
 
+    /**
+     * Get the serialization options applied to the resource response.
+     *
+     * @return int
+     */
+    public function jsonOptions()
+    {
+        return 0;
```
humaneguy, line 174: Returning 0 here means what exactly?

tests/Integration/Http/Fixtures/PostResourceWithJsonOptions.php
```diff
@@ -0,0 +1,22 @@
...
+        return [
+            'id' => $this->id,
+            'title' => $this->title,
+            'reading_time' => $this->reading_time,
+        ];
+    }
+
+    public function jsonOptions()
+    {
+        return JSON_PRESERVE_ZERO_FRACTION;
```
humaneguy, line 20: This seeems to be bound to only one implementation at a time. What if I want to use it with something like `JSON_UNESCAPED_SLASHES `. How do I go about that?

src/Illuminate/Http/Resources/Json/JsonResource.php
bastien-phi: It just means no options, which is the default for json serialization

tests/Integration/Http/Fixtures/PostResourceWithJsonOptions.php
bastien-phi: The options are integers used as bit masks. If you want multiple options, go for `JSON_PRESERVE_ZERO_FRACTION|JSON_UNESCAPED_SLASHES`

bastien-phi: Take a look at https://www.php.net/manual/en/function.json-encode.php ;-)

src/Illuminate/Http/Resources/Json/JsonResource.php
humaneguy: Why not null? It's pretty self-explanatory

bastien-phi: The result of this method will be passed to `json_encode` which expects an integer. Null is not possible here

bastien-phi:
Thanks !

ninjami-matti:
This has introduced an error for me:

```php
namespace App\Http\Resources;

use App\Models\User;
use Illuminate\Http\Resources\Json\ResourceCollection;

class UserResourceCollection extends ResourceCollection
{
    public $collects = User::class;

    /**
     * Transform the resource collection into an array.
     *
     * @return array
     */
    {
        return $this->collection->map(function ($user) use ($request) {
            // ...
        });
    }

    // ...
}
```

Using this ResourceCollection throws `BadMethodCallException with message 'Call to undefined method App\Models\User::jsonOptions()'`.

Should I not use model classes in `$collects` property anymore? In this resource collection I want it NOT to make each row into a `UserResource` instance.

I've been following [this](https://laravel.com/docs/8.x/eloquent-resources#customizing-the-underlying-resource-class) guide to make it work the way I want.

driesvints:
@bastien-phi can you fix the issue from above? Otherwise it'll be best to revert this PR.

bastien-phi:
@driesvints It looks like an unexpected usage of ResourceCollection. As documentation states, `$collects` should be a class-string of JsonResource
image

As it worked with that, I will open a PR to fix this.

driesvints:
@bastien-phi you're correct. @ninjami-matti it seems you're using a model here but `$collects` is intended for resources only. That's also explained in the guide you're linking to.

ninjami-matti:
Fair enough 👍

I would still argue that it is slightly confusing that you can still make `ResourceCollection` classes without a corresponding `JsonResource` class existing, which then makes the `ResourceCollection` collect the actual `Model` class as the resource.

I set the `$collects` property with a `Model` class based on the assumption that it should work since using the `Model` does work when not explicitly defined. And it _did_ work before this addition.

But indeed I see now that setting a `Model` class to the `$collects` property is the wrong way go about this. Thanks for the help!

bastien-phi:
@ninjami-matti If you want to bypass auto-discovery of underlying resource, you can just rename your resource collection in order to make it not ending with `Collection` and remove `$collects`

```php
class UserCollectionResource extends ResourceCollection
{
    /**
     * Transform the resource collection into an array.
     *
     * @return array
     */
    {
        return $this->collection->map(function (\App\Models\User $user) use ($request) {
            // ...
        });
    }

    // ...
}
```
will work

ninjami-matti:
That's true! However, I'm just too much of a stickler on following "correct" naming conventions 😄 So I went ahead and changed my collections and resources to reflect the correct usage and the problem is now gone.

Thanks for the tip!

PR #40210, opened 2021-12-30
[8.x] Add `shuffle()` method to `Str` and `Stringable`

ash-jc-allen:
Hi! This PR adds a new `shuffle()` method to the `Str` and `Stringable` classes that just shuffles a string and then returns it.

I've taken a similar approach to how the `Arr` class implements `shuffle()` and allowed a seed to be passed in if needed.

Here's a few ways you could use the methods:

```php
// Without seed...
$result = Str::shuffle('hello123'); // $result is: "h132eoll"
$result = (new Stringable('hello123'))->shuffle(); // $result is: "l1h32loe"

// With seed...
$result = Str::shuffle('hello123', 1234); // $result is always: "lle21ho3"
$result = (new Stringable('hello123'))->shuffle(1234); // $result is always: "lle21ho3"
```

If this is something you might consider pulling in, please let me know if there's anything that needs changing on it 😄

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -687,6 +687,26 @@ class Str
     }
 
     /**
+     * Shuffle the given string and return the result.
+     *
+     * @param  string  $value
+     * @param  int|null  $seed
+     * @return string
+     */
+    public static function shuffle(string $value, $seed = null)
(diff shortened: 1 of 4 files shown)


taylorotwell:

PR #40211, opened 2021-12-30
[9.x] Adds `soundex` search to Mysql 

mateusjunges:
This PR adds  implements [soundex search for mysql](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_soundex), adding `whereSoundex` and `orWhereSoundex` methods to the `Illuminate\Database\Query\Builder` class.

In mysql, the `SOUNDEX()` function in MySQL is used to return a phonetic representation of a string, which represents the way the string will sound.

```php
    $table->id();
    $table->string('email', 200);
    $table->string('first_name');
    $table->string('last_name');
});
```

```php
// Search users where their first name sounds like 'tailor'
$users = DB::table('users')->whereSoundex('first_name', 'tailor')->get();

// Search users where their last name sounds like 'otuel':
$users = DB::table('users')->whereSoundex('last_name', 'Otuel')->get();
```

If this PR gets merged, we can also add `soundex` support for other database engines:

- https://docs.microsoft.com/pt-br/sql/t-sql/functions/soundex-transact-sql?view=sql-server-ver15
- https://wiki.postgresql.org/wiki/Soundex

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1218,6 +1218,27 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Adds a "where soundex" clause to the query.
+     *
+     * @param  string  $columns
+     * @param  string  $value
+     * @param  string  $boolean
+     * @return $this
+     */
+    public function whereSoundex($columns, $value, $boolean = 'and')
+    {
+        $type = 'Soundex';
+
+        $columns = Arr::wrap($columns);
+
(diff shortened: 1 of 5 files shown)


driesvints:
This is non breaking afaik and can go to 8.x

mateusjunges:
@driesvints should i close this and send to 8.x instead?

mateusjunges:
@driesvints i cherry picked to #40214. Since there is no BC, i think we can close this one

PR #40214, opened 2021-12-30
[8.x] Adds `soundex` search to Mysql 

mateusjunges:
This PR is the same as #40211, since it doesn't contain breaking changes.

---

This PR implements [soundex search for mysql](https://dev.mysql.com/doc/refman/8.0/en/string-functions.html#function_soundex), adding `whereSoundex` and `orWhereSoundex` methods to the `Illuminate\Database\Query\Builder` class.

In mysql, the `SOUNDEX()` function in MySQL is used to return a phonetic representation of a string, which represents the way the string will sound.

```php
    $table->id();
    $table->string('email', 200);
    $table->string('first_name');
    $table->string('last_name');
});
```

```php
// Search users where their first name sounds like 'tailor'
$users = DB::table('users')->whereSoundex('first_name', 'tailor')->get();

// Search users where their last name sounds like 'otuel':
$users = DB::table('users')->whereSoundex('last_name', 'Otuel')->get();
```

If this PR gets merged, we can also add `soundex` support for other database engines:

- https://docs.microsoft.com/pt-br/sql/t-sql/functions/soundex-transact-sql?view=sql-server-ver15
- https://wiki.postgresql.org/wiki/Soundex

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1518,6 +1518,27 @@ class Builder
     }
 
     /**
+     * Adds a "where soundex" clause to the query.
+     *
+     * @param  string  $columns
+     * @param  string  $value
+     * @param  string  $boolean
+     * @return $this
+     */
+    public function whereSoundex($columns, $value, $boolean = 'and')
+    {
+        $type = 'Soundex';
+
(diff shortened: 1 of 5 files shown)


taylorotwell:

PR #40220, opened 2022-01-01
[8.x] Make use of vendor:publish for stubs, so packages can hook into it

benjam-es:
Make use of the vendor publish code in the same way exceptions are published. This allows packages to hook into the publishing of stubs. T

The use case would be for packages with custom make commands. Packages can publish relevant stubs when the 'laravel-stubs' tag is used with ```vendor:publish```

Changing the existing stub:publish command to call the vendor:publish command, allows the command to work in exactly the same way it did previously, with the benefit of being able to use the laravel-stubs tag.

### DIFF
--- a/src/Illuminate/Foundation/Console/StubPublishCommand.php
+++ b/src/Illuminate/Foundation/Console/StubPublishCommand.php
@@ -3,7 +3,6 @@
 namespace Illuminate\Foundation\Console;
 
 use Illuminate\Console\Command;
-use Illuminate\Filesystem\Filesystem;
 
 class StubPublishCommand extends Command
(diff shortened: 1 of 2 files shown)


benjam-es:
Considered just publishing the 'stubs' directory, but kept the files as separate lines so we keep the functionality of not overwriting existing files (without --force)

GrahamCampbell:
I think this is going to cause merge conflicts. Possibly master is the best branch for this. Laravel 9 is only weeks away from release anyway. :)

taylorotwell:
Maybe it would be better to just have an event or something?

PR #40226, opened 2022-01-02
[8.x] Use first host if multiple

### LINKED ISSUE #40186: `php artisan db` command does not work when separate read and write hosts are configures in db config
- Laravel Version: 8.77.1
- PHP Version: 7.4.27
- Database Driver & Version:
mysql  Ver 8.0.27-0ubuntu0.20.04.1 for Linux on x86_64 ((Ubuntu))

### Description:
I have configured separate read and write connections in database config file, like below

Screenshot 2021-12-28 153939

And when running the `php artisan db` it cannot find the host and throw exception.

Screenshot 2021-12-28 154159

### Steps To Reproduce:
1. Setup the separate read write connections for mysql in database config
`'read' => [
                'host' => [
                    env('DB_HOST_SLAVE', '127.0.0.1')
                ],
            ],
            'write' => [
                'host' => [
                    env('DB_HOST_MASTER', '127.0.0.1')
                ],
            ],`
2. Run `php artisan db` or `php artisan db mysql`

jelib3an:
According to the [docs](https://laravel.com/docs/8.x/database#read-and-write-connections), the host option can be an array of hosts. 

If the `php artisan db` is called with either `--read` or `--write`, this adds a check to use the first host from the array. 

This is a limited use case so impact should be low. I don't know how to write tests for this but if it is necessary and someone can kindly show me how, I'd be happy to add tests.

---

Fixes #40186.

### DIFF
--- a/src/Illuminate/Database/Console/DbCommand.php
+++ b/src/Illuminate/Database/Console/DbCommand.php
@@ -67,8 +67,14 @@ class DbCommand extends Command
         }
 
         if ($this->option('read')) {
+            if (is_array($connection['read']['host'])) {
+                $connection['read']['host'] = $connection['read']['host'][0];
+            }
             $connection = array_merge($connection, $connection['read']);
         } elseif ($this->option('write')) {
+            if (is_array($connection['write']['host'])) {
+                $connection['write']['host'] = $connection['write']['host'][0];
+            }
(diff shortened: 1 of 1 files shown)


NickSdot:
@jelib3an don't you think it would be better to allow "--read" to optionally accept a value (array key) instead of doing this fully hard coded?

PR #40232, opened 2022-01-03
[8.x] Adds support for paginating relations at query-time

DarkGhostHunter:
## What?

Allows query relationships as paginated results at query-time.

```php
use App\Models\Post;

$post = Post::withPaged('comments')->find(1);

$post->comments->currentPage(); // 1

// https://myapp.com/post/1/?comments_page=1
```` 

It also works for simple and cursor pagination.

```php
Post::withSimplePaged('comments')->find(2);
Post::withCursorPaged('comments')->find(3);
```` 

This avoids a setting the relation, which is counter-intuitive, or creating a new query in other places.

```php
use App\Models\Post;

$post = Post::find(1);

$post->setRelation('comments', $post->comments()->paginate());
```` 

## Flexibility

You can use a short declaration to set the items per page, the page/cursor, and page name.

```php
Post::withPaged('comments', 10, 2, 'post_comments')->find(4);
```

It supports using a function to modify the query and the pagination parameters separately.

```php
use Illuminate\Pagination\Cursor;

Post::withPaged('comments', function ($query, $pagination) {
    $pagination->page(2)->perPage(10)->pageName('approved_comments');
    
    $query->where('is_approved', true);
})->find(5);

Post::withCursorPaged('comments', function ($query, $pagination) {
    $pagination->cursor(['comment.id' => 10])->perPage(10)->cursorName('approved_comments');
    
    $query->where('is_approved', true);
})->find(6);
```

Additionally, it works like `with()` if you use multiple relations,

```php
Post::withPaged('comments', 'drafts')->find(7);

// https://myapp.com/post/1/?comments_page=1&drafts_page=1
```

... or an array with callbacks.

```php
Post::withPaged([
    'comments' => function ($query) {
        $query->where('is_approved', true);
    },
    'drafts' => function ($query, $pagination) {
        $pagination->perPage(5);
    }
])->find(8);
```

### First level relation only

This approach avoids the shenanigans of using nested relations. If this happened, then multiple children would have multiple pages with the same name, forcing the developer to identify each by an ID, which would be more crazy if someone decides to use UUID.

```php
Post::withPaged('galleries.images')->find(9);

// https://myapp.com/post/1/?galleries_page=1&galleries_images_page=1&galleries_images_page=1...
```

Instead, the pagination works only at the first level, while eager-loading the nested relations.

### No Load

Some users may prefer to use something like `loadPaged()`. Instead, they should use the pagination helpers directly.

## How?

First, I separated the pagination procedure into two: modifying the query to paginate, and wrapping the result into a paginator. This is (was) done opaquely by the Query Builder.

Second, I blatantly copied the code of both tasks to the `Pagination` class. It exposes the last task as a static function.

This instance also outputs some convenient methods to change how the pagination should proceed, like changing the items per page. Is passed down as a second argument to the query callback.

Lastly, I created a function to handle eager-loading the relation if it was marked as paginated. This data is inside the Eloquent Builder.

PS: Happy new year.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -52,6 +52,13 @@ class Builder
     protected $eagerLoad = [];
 
     /**
+     * The relationships that should be eager loaded with pagination.
+     *
+     * @var array|\Illuminate\Database\Eloquent\Pagination[]
+     */
+    protected $paginatedEagerLoad = [];
+
+    /**
      * All of the globally registered builder macros.
      *
      * @var array
@@ -116,7 +123,6 @@ class Builder
      * @var array
      */
     protected $scopes = [];
-
     /**
      * Removed global scopes.
      *
@@ -663,6 +669,13 @@ class Builder
 
         $relation->addEagerConstraints($models);
 
+        // If the relationship should be paginated, we will use the Pagination instance
+        // to build the page and then wrap it into its own Paginator. If that is not
+        // the case, we can proceed to normally get these results and match them.
+        if (isset($this->paginatedEagerLoad[$name])) {
+            return $this->paginateRelation($models, $name, $relation, $constraints);
+        }
+
         $constraints($relation);
 
         // Once we have the results, we just match those back up to their parent models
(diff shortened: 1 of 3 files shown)


src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -116,7 +123,6 @@ class Builder
      * @var array
      */
     protected $scopes = [];
-
```
driesvints, line 119: Undo this please.

driesvints (changes requested):

driesvints:
Can you explain why this is needed?

taylorotwell:

claudiodekker:
Do note this can also be accomplished without any modifications to Laravel's core:

```php

Builder::macro('withPaged', function ($arguments) {
    // your logic for when e.g. User::withPaged(...) is called.
});
```

dammy001:
Yeah and probably convert it to a package or something

PR #40234, opened 2022-01-03
[8.x] Enable `only-to-others` functionality when using Ably broadcast driver

leeovery:
### TL;DR

This PR enables the `only to others` broadcast functionality when using the Ably broadcast driver and the AblyJs frontend client. 

### Issue

When using the Ably broadcast driver the [`only to others` functionality provided by Laravel](https://laravel.com/docs/8.x/broadcasting#only-to-others) does not work when using the AblyJs client on the frontend.

This is because Laravel includes the socketId in the `payload` array but Ably requires it to be a top-level property called `connectionKey`.

(source: https://faqs.ably.com/is-it-possible-to-prevent-messages-published-being-echoed-back-to-the-publishing-client)

### Fix

This PR refactors the Ably broadcaster to pass a fully-formed Ably `Message` object to the Ably `publish` method, where we can also include the socketId as the `connectionKey`. 

Currently,  we pass a string event and an array payload to the `publish` method, and the Ably Rest client builds the `Message` object for us, but there is no way to include the `connectionKey` as a top-level property. (source: https://github.com/ably/ably-php/blob/main/src/Channel.php#L111)

To ensure backwards compatibility, the `socket` property will remain in the `payload` array too. It's possible people are working around this missing feature by using that property on the frontend.

With this PR merged, to enable the `only to others` functionality with Ably, you then only need to set the broadcast event to include the socket ID as per the Laravel docs, and then set `echoMessages=false` on the frontend when using the AblyJs client. This will prevent echoing messages back to originator.

Thanks
Lee

### DIFF
--- a/src/Illuminate/Broadcasting/Broadcasters/AblyBroadcaster.php
+++ b/src/Illuminate/Broadcasting/Broadcasters/AblyBroadcaster.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Broadcasting\Broadcasters;
 
 use Ably\AblyRest;
+use Ably\Models\Message as AblyMessage;
 use Illuminate\Support\Str;
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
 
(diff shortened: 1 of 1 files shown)


leeovery:
To add, you also need to configure the axios interceptor. Echo isnt currently fully compatible with Ably. But thats a frontend issue and easily added:

eg:

```javascript
const broadcaster = new Ably.Realtime({
  ...options,
  echoMessages: false,
});

let interceptorId;
broadcaster.connection.on('connected', () => {
 interceptorId = $axios.interceptors.request.use((config) => {
    const socketId = broadcaster.connection.key;
    if (socketId) {
      config.headers['X-Socket-Id'] = socketId;
    }

    return config;
  });
});
broadcaster.connection.on('suspended', () => {
  if (interceptorId) {
    $axios.interceptors.request.eject(interceptorId);
  }
});
```

taylorotwell:
Don't we currently document using Ably via the Pusher JS client and Pusher protocol support enabled on the Ably control panel?

leeovery:
Yea we do. But by using that you miss out on some Ably features in seems. They suggest not doing this too in their docs (obviously lol).

Point 3 under the General header: https://faqs.ably.com/using-the-ably-pusher-protocol-adapter

This PR wont break that functionality though as its purely additive. Just enables the feature for those that choose to go the Ably way in order to get the extra features and security.

PR #40236, opened 2022-01-03
[9.x] Add a default user agent to the HTTP client

mattkingshott:
This PR adds a default user agent to the HTTP client (`PendingRequest` class).

## Reasoning

It is generally a considered a good idea to add a custom user agent when sending HTTP requests to external sites. This allows the maintainers of these sites to identify requests coming from your application and provides them the means to get in touch with you should there be any issues e.g. a need to use rate limiting, caching etc.

## Status Quo

Out of the box, Laravel doesn't use a user agent. This means, that the HTTP client defers to Guzzle's default user agent, which is simply 'Guzzle' followed by a version number.

## Solution

The PR adds a `withDefaultUserAgent` method, which is called from the `PendingRequest` constructor. This method generates and sets a user agent derived from the application's name, url and version e.g.

```
Example/1.0 - https://example.com
```

The developer is of course free to override this by using the already present `withUserAgent` method.

## Breaking Changes

In all likelihood, the chance of this causing a breaking change in actual applications is low. However, since the PR does alter the constructor of `PendingRequest`, and I had to update some validation tests that used the HTTP client, I've decided to submit it to the `master` branch.

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -397,7 +397,7 @@ class PendingRequest
      */
     public function withDefaultUserAgent()
     {
-        $agent = config('app.name') . '/' . app()->version() . ' - ' . config('app.url');
+        $agent = config('app.name').'/'.app()->version().' - '.config('app.url');
 
         return $this->withUserAgent($agent);
     }
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -173,6 +173,8 @@ class PendingRequest
             'timeout' => 30,
         ];
 
+        $this->withDefaultUserAgent();
+
         $this->beforeSendingCallbacks = collect([function (Request $request, array $options, PendingRequest $pendingRequest) {
             $pendingRequest->request = $request;
             $pendingRequest->cookies = $options['cookies'];
@@ -386,6 +388,18 @@ class PendingRequest
         return tap($this, function ($request) use ($userAgent) {
             return $this->options['headers']['User-Agent'] = trim($userAgent);
         });
+    }
+
+    /**
+     * Assign the default user agent for the request.
+     *
+     * @return $this
+     */
+    public function withDefaultUserAgent()
+    {
+        $agent = config('app.name') . '/' . app()->version() . ' - ' . config('app.url');
+
+        return $this->withUserAgent($agent);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


SjorsO:
Not sure how I feel about sending out my application name and url every time I make an HTTP request. Especially since this is a default, I probably won't even realise I'm sending out this information. I think I'd be more comfortable with the current anonymous "Guzzle" default.

mattkingshott:
> I think I'd be more comfortable with the current anonymous "Guzzle" default.

I'm not sure why you would need your HTTP user agent to be anonymous? You could change it to something generic if you wanted to. However, doing so doesn't alter the fact that your IP address (and perhaps other details) are still being sent. So, you haven't really given yourself true anonymity.

Regardless, per [RFC 7231](http://svn.tools.ietf.org/svn/wg/httpbis/specs/rfc7231.html#rfc.section.5.5.3), you should really include one:

The "User-Agent" header field contains information about the user agent originating the request, which is often used by servers to help identify the scope of reported interoperability problems, to work around or tailor responses to avoid particular user agent limitations, and for analytics regarding browser or operating system use. A user agent SHOULD send a User-Agent field in each request unless specifically configured not to do so.

While the RFC doesn't explicitly mention sending something unique, generic user agents such as `Guzzle/7` do run the risk of being flagged. They might be interpreted as bots and be banned, or they may become subject to less harsh, but still problematic restrictions.

driesvints:
Bit torn, I can definitely see this being useful. But at the same time also feel for @SjorsO' plea. Another thing is that I don't think using Laravel's version number makes sense as it's the version of Laravel, not the app performing the HTTP request.

Therefor, I believe this should always be better set in user land because it'll always be different/custom for each app.

Regardless of the above opinions, we can't just use the `app` and `config` helpers inside the `PendingRequest` class.

mattkingshott:
It's definitely a valid point. My reasoning behind this was that many developers probably wouldn't think to set one, which isn't a good practice. What would you say to adding a config value to `app.php` e.g.

```php
'user_agent' => env('APP_NAME') . '/' . env('APP_VERSION', '1.0') . ' - ' . env('APP_URL'),
```

driesvints:
@mattkingshott I'll let @taylorotwell review this PR further 👍

GrahamCampbell:
Big 👎 for making this default. This is a borderline security issue. NB Guzzle now shares even less information by default, since 6, and again since 7.

mattkingshott:
May I ask why you consider it a security issue?

mattkingshott:
Seems like this is gonna be controversial, so I’m going to close this.

PR #40238, opened 2022-01-03
[8.x] add possibility to pass relation name of morphToMany relations

schaeftide:
This PR adds the possibility to specify a custom name for the morphedByMany relation.

The primary added benefit comes from using the Dynamic relations (resolveRelationUsing). 

**Old behavior:**

```
Tag::resolveRelationUsing('posts', function(Tag $model) {
    return $model->morphedByMany(
        Post::class,
        'taggable',
    );
});
```

If you subsequently call ```$tag->posts()->getRelationName();``` you will get ```...{closure}```.

**New behavior:**

```
Tag::resolveRelationUsing('posts', function(Tag $model) {
    return $model->morphedByMany(
        Post::class,
        'taggable',
        null,
        null,
        null,
        null,
        null,
        false,
        'posts' // pass my relation name
    );
});
```

If you subsequently call ```$tag->posts()->getRelationName();``` you now get the ```posts``` instead of ```...{closure}```.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
@@ -536,13 +536,14 @@ trait HasRelationships
      * @param  string|null  $parentKey
      * @param  string|null  $relatedKey
      * @param  bool  $inverse
+     * @param  string|null  $relationName
      * @return \Illuminate\Database\Eloquent\Relations\MorphToMany
      */
     public function morphToMany($related, $name, $table = null, $foreignPivotKey = null,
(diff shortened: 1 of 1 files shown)


driesvints:
These are breaking changes to the method signatures. Please send this to master. Thanks

schaeftide:
Thanks for the feedback, but can you explain why this is considered as a breaking change?
There is a new parameter added, but there is no change in the actual logic/call does not change.

mateusjunges:
@schaeftide if you change a public method signature, according to [SemVer](https://semver.org) it's considered a BC.

PR #40240, opened 2022-01-03
[8.x] The Queue Failed_Job Table be compatible with Job Table

DKepov:
## The Queue Failed_Job Table be compatible with Job Table

The commit consists of two small improvements.

1. It makes no sense to have a field type greater than what is actually used. For example, forwarding a failure task from jobs of database.

https://github.com/laravel/framework/blob/7ef38c8f25f0f243d4512661ac0ec649236d738c/src/Illuminate/Queue/Console/stubs/jobs.stub#L18

2. It makes sense to index this field. This may be useful in future tasks. And very often projects use the queue of failure tasks for analysis. Other packages also often work with an error queue by selecting a certain type of task.


taylorotwell:
Feel free to change this in your own application if you wish.

PR #40243, opened 2022-01-04
[8.x] Add 'getRouter' method to RouteRegistrar

iamgergo:
This PR adds the `getRouter` method to the `RouteRegistrar` class.

In some cases, it is would be very useful to get access to the router from the registrar.

```php
$registrar = Route::prefix('foo')->as('foo.');

$router = $registrar->getRouter();

// Do some stuff

$registrar->group(...);
```

### DIFF
--- a/src/Illuminate/Routing/RouteRegistrar.php
+++ b/src/Illuminate/Routing/RouteRegistrar.php
@@ -91,6 +91,16 @@ class RouteRegistrar
     }
 
     /**
+     * Get the router instance.
+     *
+     * @return \Illuminate\Routing\Router
+     */
+    public function getRouter()
+    {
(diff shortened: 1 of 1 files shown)


driesvints:
Why do you need this?

iamgergo:
In some of my methods I want to pass the router directly, after I prepared a specific group. Now I can do it only pass the router from the group callback:

```php
Route::prefix('foo')->as('foo.')->group(function ($router) use ($widget) {
    $widget->routes($router);
});
```

This would be perfect in most of the cases, but in some cases I have to do some different logic where accessing the router not from a closure would be easier and better (I think).

taylorotwell:

PR #40244, opened 2022-01-04
[8.x] Fixed behavior when unsaved *soft delete* models are `restore()` before `save()`

KentarouTakeda:
If we perform `restore()` on a newly created instance of "soft delete" models, subsequent `save()` will not save the model to the database.

## Example of Issue

```php
$model = new Foo(); // Foo is a "soft delete" model.
$model->restore();
$model->save();

$model->refresh();
// Illuminate\Database\Eloquent\ModelNotFoundException with message 'No query results for model [Foo].'

DB::select("select * from foos");
// []
```

## Expected behavior

The model should be saved.

`restore()` for a non-soft-deleted model simply does nothing if it exists in the database, and subsequent `save()` works fine. So, `save()` should work the same even if it doesn't exist in the database.

## Solution

At the time of `restore()`, `$model->exists`  is [always set to true](https://github.com/laravel/framework/blob/v8.77.1/src/Illuminate/Database/Eloquent/SoftDeletes.php#L122) . Therefore, the cause is that the subsequent `save()` queries the database for update instead of insert. 

Fixed to not change `$model->exists` when `restore()`, and do not do `save()` because the record does not exist from the beginning, that is, do nothing.

### DIFF
--- a/src/Illuminate/Database/Eloquent/SoftDeletes.php
+++ b/src/Illuminate/Database/Eloquent/SoftDeletes.php
@@ -119,9 +119,11 @@ trait SoftDeletes
         // Once we have saved the model, we will fire the "restored" event so this
         // developer will do anything they need to after a restore operation is
         // totally finished. Then we will return the result of the save call.
-        $this->exists = true;
+        $result = null;
+        if ($this->exists) {
+            $result = $this->save();
+        }
 
-        $result = $this->save();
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't really want to mess with this logic on a patch release, especially on something as highly used as Eloquent. Just don't call `restore` on a model that doesn't exist?

PR #40245, opened 2022-01-04
[8.x] Fix attribute casting

### LINKED ISSUE #40230: The issue is a new pattern of Attribute Cast
I'm using the new version of `Attribute Cast` in the Laravel version: 8.77 but showed this error when logged in authorization with guard admin 

- Laravel Version: 8.77.1
- PHP Version: 8.0

### Error:
image

### Model Admin:
```php
<?php

namespace App\Models\Tenant;

class Admin extends Authenticatable
{
    use Notifiable, HasRoles, SoftDeletes;

    /**
     * The authentication guard name.
     *
     * @var string
     */
    public $guard = 'admin';

    /**
     *
     * @var array
     */
    protected $fillable = [
        'name',
        'email',
        'password',
    ];

    /**
     * The attributes that should be hidden for arrays.
     *
     * @var array
     */
    protected $hidden = [
        'password',
        'remember_token',
    ];

    /**
     * The attributes that should be cast to native types.
     *
     * @var array
     */
    protected $casts = [
        'email_verified_at' => 'datetime',
    ];
    
    /**
     * Set the admin's Hash password.
     *
     * @return Attribute
     */
    {
        return new Attribute(
            set: fn($value) => Hash::make($value),
        );
    }
}

### PR DESCRIPTION
**Pull request by driesvints**
Fixes an oversight in https://github.com/laravel/framework/pull/40022.

At the moment when you omit a get cast with the new attribute casting, the attribute call will propagate through the getAttribute method and end up to getRelationValue which will throw an exception that the attribute isn't a relationship instance. This is because the way attribute calls work is to fallback on relationships when they're not available. So the way the checks are done on get/set callables with the new attribute casting isn't entirely correct. We should omit the callable checks and just check if the method that corresponds with the requested attribute is returning an Attribute cast instance. If that's the case we can assume the user wanted to use attribute casting, regardless of a getter/setter being available or not. 

A cavaet of this is that this will prevent the call to continue to class casts and the old date casting. I do not know how to solve this and would appreciate any insights you have @taylorotwell.

Fixes #40230

### DIFF
--- a/tests/Integration/Database/DatabaseEloquentModelAttributeCastingTest.php
+++ b/tests/Integration/Database/DatabaseEloquentModelAttributeCastingTest.php
@@ -142,6 +142,8 @@ class DatabaseEloquentModelAttributeCastingTest extends DatabaseTestCase
     {
         $model = new TestEloquentModelWithAttributeCast;
 
+        $this->assertEquals(null, $model->password);
+
         $model->password = 'secret';
 
         $this->assertEquals(hash('sha256', 'secret'), $model->password);
--- a/tests/Integration/Database/DatabaseEloquentModelAttributeCastingTest.php
+++ b/tests/Integration/Database/DatabaseEloquentModelAttributeCastingTest.php
@@ -142,7 +142,7 @@ class DatabaseEloquentModelAttributeCastingTest extends DatabaseTestCase
     {
         $model = new TestEloquentModelWithAttributeCast;
 
-        $this->assertEquals(null, $model->password);
+        $this->assertNull($model->password);
 
         $model->password = 'secret';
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by driesvints** [comment]
If there's specific scenarios for class casting/date casting in combination with the new attribute casting then we can add those to the existing test case and work until we get all the tests to pass.

File: tests/Integration/Database/DatabaseEloquentModelAttributeCastingTest.php [hunk 1]
```diff
@@ -142,6 +142,8 @@ public function testOneWayCasting()
     {
         $model = new TestEloquentModelWithAttributeCast;
 
+        $this->assertEquals(null, $model->password);
```
BrandonSurowiec, line 145: Can use `$this->assertNull($model->password);`

driesvints: Thanks!

PR #40247, opened 2022-01-04
[8.x] Add `htmlString` helper method

Larsklopstra:
**Motivation:**

I often use the `HtmlString` to render `Stringable` data without using `{!! $content !!}`.

**Before:**

```php
return new HtmlString(Str::of($content)->markdown());

return new HtmlString(Str::markdown($content));
```

**After:**

```php
return Str::of($content)
    ->markdown()
    ->html();
```

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -822,6 +822,16 @@ class Stringable implements JsonSerializable
     }
 
     /**
+     * Convert the string into a `HtmlString`.
+     *
+     * @return HtmlString
+     */
+    public function htmlString()
+    {
+        return new HtmlString($this->value);
+    }
+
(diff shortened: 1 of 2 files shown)


driesvints:
@selcukcukur instead of downvoting please provide feedback why you don't like this addition.

driesvints:
@Larsklopstra maybe just `->html()`?

src/Illuminate/Support/Stringable.php
```diff
@@ -821,6 +821,16 @@ public function wordCount()
         return str_word_count($this->value);
     }
 
+    /**
+     * Convert the string into a `HtmlString`.
+     *
+     * @return HtmlString
```
driesvints, line 827: Use a FQN here.

PR #40249, opened 2022-01-04
[9.x] Add support for psr/simple-cache v2 and v3

driesvints:
Add support for `psr/simple-cache` v2 and v3 in Laravel v9.

### DIFF
--- a/src/Illuminate/Cache/composer.json
+++ b/src/Illuminate/Cache/composer.json
@@ -21,7 +21,7 @@
         "illuminate/support": "^9.0"
     },
     "provide": {
-        "psr/simple-cache-implementation": "1.0"
+        "psr/simple-cache-implementation": "1.0|2.0|3.0"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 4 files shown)


src/Illuminate/Cache/composer.json
```diff
@@ -21,7 +21,7 @@
         "illuminate/support": "^9.0"
     },
     "provide": {
-        "psr/simple-cache-implementation": "1.0"
+        "psr/simple-cache-implementation": "1.0|2.0|3.0"
```
GrahamCampbell, line 24: I'm not sure that syntax is supported? Don't we have to specify "the version" we provide? What do symfony do here?

GrahamCampbell (changes requested):

src/Illuminate/Cache/composer.json
driesvints: It is: https://github.com/symfony/cache/blob/5.4/composer.json#L20

PR #26747, opened 2018-12-04
[5.8] Add ArrayAccess to config repository contract

X-Coder264:
This is the same as it's been done with the container in #26378. Basically a bunch of code throughout the framework uses `ArrayAccess` for getting configuration values even though that is not on the contract at the moment which means it'd break with any other implementation thus rendering the contract pointless without this change.

PR #26378, opened 2018-11-03
[5.8] Add ArrayAccess to container contract

X-Coder264:
This PR adds `ArrayAccess` to the Container contract as Taylor suggested in #26364.

PR #26364, opened 2018-11-02
[5.8] Fix container contract violations

X-Coder264:
The Container contract concrete implementation also implements `ArrayAccess` (it uses the methods which are on the contract to implement `ArrayAccess`). As the Container contract itself doesn't extend `ArrayAccess` we cannot use `ArrayAccess` when resolving services from the container as a `Illuminate\Contracts\Container\Container` contract implementation isn't forced to implement it and would break when we'd attempt to resolve the services from it that way. Some parts of the framework are using `ArrayAccess` which is a clear violation of the contract while on some other places the proper contract methods are used so not only is the contract violated in a lot of places, the entire code is pretty inconsistent. This PR aims to fix that. I also added some type-hints for better IDE auto-completion in Closures.

PR #40252, opened 2022-01-04
[9.x] Move array access to container contract

axlon:
Laravel's container (and by extension application) allows for its services to be retrieved via array access, e.g.: 

```php
$app = app();
$cache = $app['cache'];
```

The problem however, is that most of the time code doesn't actually "know" if the container supports being accessed as an array, because neither the container nor the application contract enforces it.

The framework itself uses this feature, primarily in service providers. Which is a context where it cannot be sure array access is supported, because the application is type-hinted as the application contract, which does not implement the array access interface.

Example:
https://github.com/laravel/framework/blob/b86e57e062fea77cc388c66b93fd006717f28c0f/src/Illuminate/Queue/QueueServiceProvider.php#L143

## What does this PR do?
It moves `ArrayAccess` from the framework's container implementation, to the container contract. Meaning all code will now be sure that array access is supported when interacting with an application or a container.

## Breaking change
This is a breaking change for any custom applications or containers that do not extend from Laravel's container or application. But these custom containers and applications would likely already have implemented `ArrayAccess` to be compatible with framework code. Because of this I estimate the impact to be very minimal.

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -2,7 +2,6 @@
 
 namespace Illuminate\Container;
 
-use ArrayAccess;
 use Closure;
 use Exception;
 use Illuminate\Contracts\Container\BindingResolutionException;
@@ -14,7 +13,7 @@ use ReflectionException;
 use ReflectionParameter;
 use TypeError;
 
-class Container implements ArrayAccess, ContainerContract
+class Container implements ContainerContract
 {
     /**
      * The current globally available container (if any).
--- a/src/Illuminate/Contracts/Container/Container.php
+++ b/src/Illuminate/Contracts/Container/Container.php
@@ -2,10 +2,11 @@
 
 namespace Illuminate\Contracts\Container;
 
+use ArrayAccess;
 use Closure;
 use Psr\Container\ContainerInterface;
 
-interface Container extends ContainerInterface
+interface Container extends ContainerInterface, ArrayAccess
 {
     /**
      * Determine if the given abstract type has been bound.


driesvints:
No we shouldn't do this. The container contract should be separate.

taylorotwell:
I don't think this is a bad change given how the container is often used. This does not make it match reality a bit more. @driesvints can you explain your objections a bit more?

driesvints:
@taylorotwell there's a lengthy discussion with lots of reason on why to not do this in https://github.com/laravel/framework/pull/26747. Attempts have been made for this in the past: https://github.com/laravel/framework/pull/26747 & https://github.com/laravel/framework/pull/26378 but they've all been reverted.

Interfaces should stay as small as possible. An implementation can implement multiple interfaces but they shouldn't depend on each other. A container doesn't depends on `ArrayAccess` to function properly. A container needs to fetch and store bindings. The fact that it can do that with the behaviour of an array is a separate concern that isn't a requirement for a container to function properly. I strongly recommend we don't do this.

Here's also an excellent read on the subject: https://jamison.dance/12-13-2019/keep-interfaces-small

axlon:
> A container needs to fetch and store bindings. The fact that it can do that with the behaviour of an array is a separate concern that isn't a requirement for a container to function properly. 

@driesvints while I agree with what you're saying here, fact of the matter is that currently Laravel does require array-like behaviour from its container contact, because multiple components are accessing it as such.

The way I see it there are two ways to fix this issue:

The first solution is what this PR is trying to do (and #26378 previously, later reverted in 0b9abba058d4fb8f9025ab8e28b437ee6ca94435). I would argue for this change, because it's a way smaller change than the alternative. On top of that, having `ArrayAccess` only exist on the container implementation (and not the interface) limits its usefulness as we can only treat the object like an array when we're sure which implementation it is, which defeats the whole purpose of having an interface in the first place.

The other solution would be to make the framework adhere to the current container and application interfaces as was previously attempted in #26364. This does not address the fact that in reality one would almost never be able to treat the container like an array, unless doing a type check first. (Which isn't the biggest deal, because the `make` method exists)

driesvints:
I also believe #26364 to be a solution but I want to strongly recommend not to go ahead with the current PR.

mfn:
FWIW, also vote for a direction like #26364 but **not** to change the contract, this feels definitely the wrong way.

PR #30542, opened 2019-11-08
[7.x] Allow Contextual Binding outside of constructors

nmokkenstorm:
This PR fixes [a long standing bug](https://github.com/laravel/framework/issues/23602) where [Contextual Binding](https://laravel.com/docs/5.8/container#contextual-binding) wouldn't work outside of the constructor of a class.

Notable relevant userland examples include the `handle` method of a Queue Job, and the injected services for a Controller.

As for the exact implementation here, I just added a new method which directly modifies the 'build stack', as that seemed to be the least intrusive way of getting stuff done. Initially I just passed along the extra parameter in `makeFor`, but that became messy very quickly.

I'm not sure wat this should be targeted to. It's a backwards compatible bugfix, but it also adds a 'new feature' in the sense that there's a new public method on the container. Any advice here would be welcome.

PR #40255, opened 2022-01-04
[8.x] Allow method typed variadics dependencies

LeoColomb:
This reasonably simple change would enable to use variadic for DI when calling a `Callable`, in addition to class resolution.

Here is an example reusing the objects from the [documentation](https://laravel.com/docs/8.x/container#binding-typed-variadics):

```php
$app->bind(Filter::class, function ($app) {
    return [
        $app->make(NullFilter::class),
        $app->make(ProfanityFilter::class),
        $app->make(TooLongFilter::class),
    ];
});

$app->call(function (Logger $logger, Filter ...$filters) {
    //                               ⏫ this is currently unsupported
});
```

I think this can be beneficial for the following reasons:
* you may have a callable or a class method that receives an array of typed objects, the same way [documentation](https://laravel.com/docs/8.x/container#binding-typed-variadics) explains for constructors;
* there are cases when resolution for constructors only is not sufficient unless over objectifying things.
  * see my example below.
* this addition partially mitigates the need of contextual binding for methods (#30542):
  * for variadic, the enforced specific context is no more required, they can be use with classical bindings.
  * improve consistency across DI
* backward compatible, as it only match variadic when a type is specified (maybe array of arrays type can conflit? not even sure it is natively supported)
* it is not very easy to implement the suggested behavior via a package

<details>
<summary>Example of a case where a constructor cannot handle it</summary>

* Not allowed:
  ```php
  public function __construct(public Logger $logger, Filter ...$filters, Whatever ...$objects)
  { //                                                                   ⏫ this is illegal
      $this->filters = $filters;
      $this->objects = $objects;
      $this->method();
  }  

  public function method()
  {
      // attributes consumer
  }
  ```

* Naive remedy:
  ```php
  public function __construct(public Logger $logger, Filter ...$filters)
  {
      $this->filters = $filters;
      App::call([$this, 'method']);
  }
  
  public function method(Whatever ...$objects)
  {
      //
  }
  ```

</details>

### DIFF
--- a/src/Illuminate/Container/BoundMethod.php
+++ b/src/Illuminate/Container/BoundMethod.php
@@ -173,7 +173,11 @@ class BoundMethod
                 unset($parameters[$className]);
             } else {
                 if ($parameter->isVariadic()) {
-                    $dependencies = array_merge($dependencies, $container->make($className));
+                    $variadicDependencies = $container->make($className);
+
+                    $dependencies = array_merge($dependencies, is_array($variadicDependencies)
+                                ? $variadicDependencies
+                                : [$variadicDependencies]);
                 } else {
                     $dependencies[] = $container->make($className);
                 }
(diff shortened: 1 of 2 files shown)


taylorotwell:
Had to make one small change for backward compatibility with existing code.

PR #40257, opened 2022-01-04
[8.x] Run `clearstatcache` after deleting file and asserting `Storage` using exists/missing

crynobone:
PHP by default would cache `is_file()` and `file_exists()` output and this would potentially cause an issue when asserting file changes if changes occurred from a different process. 

1. https://www.php.net/manual/en/function.clearstatcache.php
2. https://github.com/composer/composer/blob/711f436b2431ebf2845ee89a556ffde619212405/src/Composer/Util/Filesystem.php#L125-L126


driesvints (approved):

driesvints:
Great work @crynobone 👍

PR #40003, opened 2021-12-12
[8.x] Support x-forwarded-prefix in TrustProxies.php

ppodds:
#39996 

If we want to deploy laravel project in a subfolder of the domain like the following Nginx setting. We need to use the x-forwarded-prefix header.

```
server {
  listen 80;
  listen [::]:80;

  server_name _;
  location /to/project/ {
    proxy_set_header X-Forwarded-Prefix /to/project;
    proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
    proxy_set_header X-Forwarded-Host my.domain;
    proxy_set_header X-Forwarded-Port 443;
    # laravel backend at localhost:8080
    proxy_pass http://localhost:8080/;
}
```

Symfony has already supported the x-forwarded-prefix header. We can deploy laravel in the subfolder of domain with this feature.

PR #40260, opened 2022-01-05
[8.x] Handle X-Forwarded-Prefix HTTP header

gguichard:
Symfony HttpFoundation supports X-Forwarded-Prefix HTTP header since version 5.3.
This header is used by Traefik reverse proxy for example.

> This ensures that the links rendered by _Laravel_ application deployed behind LB are valid even if the application is deployed via prefix URL.
> 
> Example routing setup:
> route /admin/(.*) => _Laravel_ backend /$1
> in this case links rendered by _Laravel_ backend must start with /admin/

c.f. https://github.com/symfony/symfony/pull/37734

### DIFF
--- a/src/Illuminate/Http/Middleware/TrustProxies.php
+++ b/src/Illuminate/Http/Middleware/TrustProxies.php
@@ -19,7 +19,7 @@ class TrustProxies
      *
      * @var int
      */
-    protected $headers = Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_HOST | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PROTO | Request::HEADER_X_FORWARDED_AWS_ELB;
+    protected $headers = Request::HEADER_X_FORWARDED_FOR | Request::HEADER_X_FORWARDED_HOST | Request::HEADER_X_FORWARDED_PORT | Request::HEADER_X_FORWARDED_PREFIX | Request::HEADER_X_FORWARDED_PROTO | Request::HEADER_X_FORWARDED_AWS_ELB;
 
     /**
      * Handle an incoming request.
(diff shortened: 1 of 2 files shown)


derekmd:
This change has been made for 9.x (https://github.com/laravel/framework/pull/40014) being released this month since it was deemed a breaking change (https://github.com/laravel/framework/pull/40003).

Although having your test additions on the `master` branch might be nice since that previous PR didn't touch the test suite.

GrahamCampbell:
This is a breaking change, and cannot be made to 8.x.

gguichard:
Okay sorry I didn't see the PR for 9.x.

After this patch (and the other patch for 9.x which is almost the same), URLs generated by the router (RouteUrlGenerator) are still not correctly prefixed with the value provided by _X-Forwarded-Prefix_.

The prefix provided by _X-Forwarded-Prefix_ is handled in getBaseUrl() function from Symfony HttpFondation (see https://github.com/symfony/http-foundation/blob/ce952af52877eaf3eab5d0c08cc0ea865ed37313/Request.php#L903)
but in the following code: https://github.com/laravel/framework/blob/283971c37c32fae6283e70c3a53fcb1a926d6350/src/Illuminate/Routing/RouteUrlGenerator.php#L103 the base url is removed from generated URI with a _preg_replace_ call. So after this call, the generated URL becomes invalid.

For the moment, the simplest fix I see is to add a little bit more code to https://github.com/laravel/framework/blob/283971c37c32fae6283e70c3a53fcb1a926d6350/src/Illuminate/Routing/RouteUrlGenerator.php#L106 (we could prepend the prefix here).

gguichard:
Okay sorry I didn't see the PR for 9.x.

After this patch (and the other patch for 9.x which is almost the same), URLs generated by the router (RouteUrlGenerator) are still not correctly prefixed with the value provided by _X-Forwarded-Prefix_.

The prefix provided by _X-Forwarded-Prefix_ is handled in getBaseUrl() function from Symfony HttpFoundation (see https://github.com/symfony/http-foundation/blob/ce952af52877eaf3eab5d0c08cc0ea865ed37313/Request.php#L903)
but in the following code: https://github.com/laravel/framework/blob/283971c37c32fae6283e70c3a53fcb1a926d6350/src/Illuminate/Routing/RouteUrlGenerator.php#L103
The base url is removed from generated URI with a _preg_replace_ call. After this call the generated URI becomes invalid.

taylorotwell:
Closing this on 8.x since we are attempting to solve this on the `master` branch (which is just a few weeks from release). @gguichard are you able to fix the other problem you mentioned?

PR #40265, opened 2022-01-05
[8.x] Add onLastPage method to the Paginator

johanvanhelden:
**In short:**
This PR adds a convenience method to the paginator to determine if we are on the last page.

**Why:**
There is a useful `onFirstPage` method but it lacks the opposite `onLastPage` method.

I was just writing some blade code for a Tailwind pagination component and without reading a any documentation `onFirstPage` worked.

Meaning, you can have the following code:

```blade
@if ($paginator->onFirstPage())
```

But for the last page, you need to call an unexpected `hasMorePages` method and reverse the logic, so it becomes:
```blade
@if (!$paginator->hasMorePages())
```
So not only did I encounter a blade error, because `onLastPage` did not exist, but also I had to reverse the logic.

This PR keeps things consistent by adding a `onLastPage()` so you can simple do:

```blade
@if ($paginator->onLastPage())
```

Which, in my mind, is a very logical thing to try after using `onFirstPage` without reading any documentation.

### DIFF
--- a/src/Illuminate/Pagination/AbstractPaginator.php
+++ b/src/Illuminate/Pagination/AbstractPaginator.php
@@ -385,7 +385,7 @@ abstract class AbstractPaginator implements Htmlable
      */
     public function onLastPage()
     {
-        return !$this->hasMorePages();
+        return ! $this->hasMorePages();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #40266, opened 2022-01-05
update DB.php

celaraze:
be more developer friendly.


taylorotwell:
There is no method with that argument name.

PR #40272, opened 2022-01-06
[7.x] Update PHPDoc of Collection::forget()

liquid207:
The PHPDoc of Collection::forget() is
```
/**
 * Remove an item from the collection by key.
 *
 * @param  string|array  $keys
 * @return $this
 */
public function forget($keys)
```
Available types of `$keys` is `string|array`, missing `int`, this PR add it.


driesvints:
7.x isn't maintained anymore sorry.

PR #40273, opened 2022-01-06
[9.x] Generic Type for `Illuminate\Support\Fluent`

crynobone:

### DIFF
--- a/src/Illuminate/Support/Fluent.php
+++ b/src/Illuminate/Support/Fluent.php
@@ -7,19 +7,23 @@ use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Contracts\Support\Jsonable;
 use JsonSerializable;
 
+/**
+ * @template TKey of array-key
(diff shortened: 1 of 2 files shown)


src/Illuminate/Support/Fluent.php
```diff
@@ -7,19 +7,23 @@
...
-     * @var array
+     * @var array<TKey, TValue>
      */
     protected $attributes = [];
 
     /**
      * Create a new fluent instance.
      *
-     * @param  array|object  $attributes
+     * @param  array<TKey, TValue>|iterable<TKey, TValue>|object  $attributes
```
devfrey, line 26: `iterable` is a supertype of `array`, so wouldn't `iterable<TKey, TValue>|object` be sufficient?

nunomaduro: Yes - I've fixed that.

nunomaduro (approved):

PR #40275, opened 2022-01-06
[8.x] Added contains Method to Arr Helper

hmennen90:
This PR adds a contains Method to Arr helper which simply Checks if a given Value is Inside a Array - supporting Multi Dimensional Arrays.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -155,7 +155,10 @@
         "wildbit/swiftmailer-postmark": "Required to use Postmark mail driver (^3.0)."
     },
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "composer/package-versions-deprecated": true
+        }
     },
     "minimum-stability": "dev",
     "prefer-stable": true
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -155,7 +155,10 @@
         "wildbit/swiftmailer-postmark": "Required to use Postmark mail driver (^3.0)."
     },
     "config": {
-        "sort-packages": true
+        "sort-packages": true,
+        "allow-plugins": {
+            "composer/package-versions-deprecated": true
+        }
```
driesvints, line 161: Can you revert this? It's already added.

taylorotwell:

Wiruu:
@taylorotwell Hello, I was just about to write an issue about this, its not a bug but an enhancement since the Arr helper methods of has and hasAny check for array keys and not values, trying to write as less code as possible and not manipulate original array, php function in_array() serves well... I believe this would be a great addition to the Arr helper...

aurawindsurfing:
I second this request. It feel unnatural to Laravel ecosystem that this helper is missing.

PR #40281, opened 2022-01-06
[9.x] Adds implicit `Enum` route binding

nunomaduro:
This pull request introduces implicit `Enum` route binding for Laravel v9.x. So let's see an example:

```php
// Assuming the following `Enum`
enum Category: string
{
    case Fruits = 'fruits';
    case People = 'people';
}

// And assuming the following route;
Route::get('/categories/{category}', function (Category $category) {
    return $category->value;
});

// GET `/categories/fruits` returns `fruits`...
// GET `/categories/people` returns `people`...
// GET `/categories/cars` returns `404 Not Found`...
```

In other words, when the given argument matches one of the `Enum` cases, the code in this pull request will resolve the associated `Enum` for the route method / closure. Otherwise, will respond `404 Not Found`, similar to Model route binding.

```php
Route::get('/categories/{category}', function (Category $category) {
    dd($category);

    // ^ Illuminate\Tests\Integration\Routing\Category^ {#1220
    //  +name: "Fruits"
    //  +value: "fruits"
    // }
});
```

Now, it's important to keep in mind, that the given `Enum` needs to be a [Backed Enum](https://www.php.net/manual/en/language.enumerations.backed.php) with a `string` backing type. Meaning that, the `Enum` header needs to look like this:

```php
enum {{ EnumName }}: string
```

### DIFF
--- a/src/Illuminate/Foundation/Exceptions/Handler.php
+++ b/src/Illuminate/Foundation/Exceptions/Handler.php
@@ -17,6 +17,7 @@ use Illuminate\Http\Exceptions\HttpResponseException;
 use Illuminate\Http\JsonResponse;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Response;
+use Illuminate\Routing\Exceptions\BackedEnumCaseNotFoundException;
 use Illuminate\Routing\Router;
 use Illuminate\Session\TokenMismatchException;
 use Illuminate\Support\Arr;
@@ -85,6 +86,7 @@ class Handler implements ExceptionHandlerContract
     protected $internalDontReport = [
         AuthenticationException::class,
         AuthorizationException::class,
+        BackedEnumCaseNotFoundException::class,
         HttpException::class,
         HttpResponseException::class,
         ModelNotFoundException::class,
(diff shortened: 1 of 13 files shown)


timmartin19:
@nunomaduro is there a specific reason that an integer backed enum type can't be used?

andrzejkupczyk:
@timmartin19 maybe this is because enums must not include the `__toString` method?

lupinitylabs:
> @nunomaduro is there a specific reason that an integer backed enum type can't be used?

I was thinking the same. And patching the code to allow for int-backed Enums works like a charm, because `tryFrom()` automatically casts string numbers to integers. 

*However*, if the user tries to pass a string like `'test'` as a route parameter, it will cause a `TypeError`. Of course you could catch the error and fail route resolution, or have the user take care of that by making sure only numerics are allowed for the parameter, e.g. by using `Route::get(...)->where(['enumParam' => '[0-9]+'])`, but I suppose for the sake of simplicity it was decided not to go down that rabbit hole.

However, I would love to see the docs being more explicit about this. It should be stated that `Laravel allows you to type-hint a *string*-backed Enum` only.

grafik

PR #40283, opened 2022-01-07
[8.x] Update PHPDoc of Collection::forget()

liquid207:
The PHPDoc of Collection::forget() is

```
/**
 * Remove an item from the collection by key.
 *
 * @param  string|array  $keys
 * @return $this
 */
public function forget($keys)
```

Available types of $keys is string|array, missing int, this PR add it.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -1628,7 +1628,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     /**
      * Unset the item at a given offset.
      *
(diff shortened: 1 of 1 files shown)


src/Illuminate/Collections/Collection.php
```diff
@@ -1628,7 +1628,7 @@ public function offsetSet($key, $value)
     /**
      * Unset the item at a given offset.
      *
-     * @param  string  $key
+     * @param  string|int  $key
```
driesvints, line 1631: This should be `mixed` actually: https://www.php.net/manual/en/arrayaccess.offsetunset.php

PR #40285, opened 2022-01-07
[8.x] Added whenContains and whenContainsAll to Stringable

telkins:
I mainly did this because I wanted to combine `when()` and `contains()`: `whenContains()`.  I added `whenContainsAll()` since it was quite similar.

I would also have taken the time to implement `whenXyz()` for all of the "check-like" methods on the class:
* `endsWith()`
* `exactly()`
* `is()`
* `isAscii()`
* `isUuid()`
* `test()`
* `startsWith()`

Please let me know if it's worthwhile or not.  I can definitely see it being valuable, so I'm happy to add these as well. 🤓

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -766,6 +766,52 @@ class Stringable implements JsonSerializable
     }
 
     /**
+     * Execute the given callback if the string contains a given substring.
+     *
+     * @param  string|array  $needles
+     * @param  callable  $callback
+     * @param  callable|null  $default
+     * @return static
+     */
+    public function whenContains($needles, $callback, $default = null)
+    {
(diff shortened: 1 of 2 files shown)


telkins:
I realized that I ought to maybe make an example to demonstrate how this can help.

Unless I'm missing something, here's an example of how things might have to be done before this PR:
```php
$stringable = Str::of('some important announcement');

if ($stringable->contains(['important', 'emergency'])) {
    $stringable->upper();
}

return (string) $stringable;

// SOME IMPORTANT ANNOUNCEMENT
```

After:
```php
return (string) Str::of('some important announcement')
    ->whenContains(
        ['important', 'emergency'],
        fn (Stringable $stringable) => $stringable->upper(),
    );
}

// SOME IMPORTANT ANNOUNCEMENT
```

PR #40286, opened 2022-01-07
[8.x] Automatically add event description when scheduling a command

maartenpaauw:
# Description

This code will use the command description as event description. This could be useful when listing all scheduled commands.

**Before**

```php
$schedule->command(Command::class)
    ->description((new Command())->getDescription())
    ->daily();
```

**After**
```php
$schedule->command(Command::class)
    ->daily();
```

The code above will output the same result when running `php artisan schedule:list`.


taylorotwell:

PR #40292, opened 2022-01-07
Fulltext sqlserver

otilor:


driesvints:
I think it might be best to create a new branch and copy/paste your code onto it.

PR #40293, opened 2022-01-07
[8.x] Fulltext search for sql server

otilor:
This adds implementation of fulltext search for SQL server

### DIFF
--- a/src/Illuminate/Database/Query/Grammars/SqlServerGrammar.php
+++ b/src/Illuminate/Database/Query/Grammars/SqlServerGrammar.php
@@ -472,6 +472,25 @@ class SqlServerGrammar extends Grammar
     }
 
     /**
+     * Compile a "where fulltext" clause.
+     *
+     * @param  \Illuminate\Database\Query\Builder  $query
+     * @param  array  $where
+     * @return string
+     */
+    public function whereFulltext(Builder $query, $where)
+    {
+        $columns = $this->columnize($where['columns']);
+        $value = $this->parameter($where['value']);
+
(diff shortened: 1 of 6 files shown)


humaneguy:
@driesvints The test for SQL server is failing because I cannot, create full text indexes on `master` database which seems to be the database configured in the workflows

Jubeki:
Maybe this resource for updating the workflow is helpful: https://github.community/t/ms-sql-service-with-full-text-search/137345

driesvints:
Seems that the SQL Server tests indeed need a separate database to be set up to test this. I can look at that when I have some time, otherwise I'd appreciate if anyone could figure that out.

driesvints:
@Jubeki from the error message it seems to me that it's possible to use full text search on the current image but that some databases aren't allowed to use it. If not, then the error message is misleading imo.

humaneguy:
> @Jubeki from the error message it seems to me that it's possible to use full text search on the current image but that some databases aren't allowed to use it. If not, then the error message is misleading imo.

Yes. you can use the full text search on the image but on a different database aside `master, tempdb, or model database`

humaneguy:
@driesvints I'm kind of in another issue here. I have to create a catalog before creating the index. Creating a catalog is another query entirely. Should I still include it in the where ```compileFulltext()``` method?

driesvints:
I have no idea. I have no experience with SQL Server at all.

driesvints:
@humaneguy if we don't have the knowledge to wrap up this PR it's best that we close it for now.

humaneguy:
Yeah. I'd just close it.

PR #40303, opened 2022-01-07
[8.x] Fix Doctrine type mappings creating too many connections

### LINKED ISSUE #40297: #40119 broke how `dbal` array in `config/database.php` is handled.
- Laravel Version: 8.78.1
- PHP Version: 8.1.0
- Database Driver & Version: MySQL 8

### Description:

#40119 introduced an improvement for handling custom Doctrine types. But for some people it broke the test suites. (https://github.com/laravel/framework/pull/40119#issuecomment-1005854737 and https://github.com/laravel/framework/pull/40119#issuecomment-1006906061)

For sequential test runs issue happens when there is only the `dbal` array in `config/database.php` And the issue is gone when adding `DB::registerDoctrineType call` to `AppServiceProvider`. Which is not a good solution because it causes the application to make DB connection while the app is booting.

For parallel test runs issue happens in both cases.

### Steps To Reproduce:

[This](https://github.com/canvural/doctrine-type-bug) repo shows the bug.

bakerkretzmar:
**TLDR:** this PR closes #40297 by only registering custom Doctrine type mappings on a Doctrine connection after that connection is created (as opposed to creating one to register them).

### Context

My and @TomHAnderson's original PR, #40119, created an issue with tests where way too many database connections would be created. We were manually creating a database connection to register the custom type mappings, which caused database connections to be created every time the app boots, and dozens of unnecessary connections to be created when tests are run in parallel.

I'm still not entirely sure why all those connections stick around, I tried manually closing/destroying them through Doctrine and that didn't seem to work.

### DIFF
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -178,6 +178,13 @@ class Connection implements ConnectionInterface
     protected $doctrineConnection;
 
     /**
+     * Type mappings that should be registered with new Doctrine connections.
+     *
+     * @var array
+     */
+    protected $doctrineTypeMappings = [];
+
+    /**
      * The connection resolvers.
      *
      * @var array
@@ -1007,6 +1014,10 @@ class Connection implements ConnectionInterface
                 'driver' => method_exists($driver, 'getName') ? $driver->getName() : null,
                 'serverVersion' => $this->getConfig('server_version'),
             ]), $driver);
+
+            foreach ($this->doctrineTypeMappings as $name => $type) {
+                $this->doctrineConnection->getDatabasePlatform()->registerDoctrineTypeMapping($type, $name);
+            }
         }
 
         return $this->doctrineConnection;
@@ -1035,9 +1046,7 @@ class Connection implements ConnectionInterface
             Type::addType($name, $class);
         }
 
-        $this->getDoctrineSchemaManager()
-            ->getDatabasePlatform()
-            ->registerDoctrineTypeMapping($type, $name);
+        $this->doctrineTypeMappings[$name] = $type;
     }
 
     /**
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -1016,7 +1016,9 @@ class Connection implements ConnectionInterface
             ]), $driver);
 
             foreach ($this->doctrineTypeMappings as $name => $type) {
-                $this->doctrineConnection->getDatabasePlatform()->registerDoctrineTypeMapping($type, $name);
+                $this->doctrineConnection
+                    ->getDatabasePlatform()
+                    ->registerDoctrineTypeMapping($type, $name);
             }
         }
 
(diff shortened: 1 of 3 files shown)


bakerkretzmar:
@canvural @skollro if either of you have feedback about this or are able to test it that would be greatly appreciated 🙏🏻

canvural:
@bakerkretzmar I confirm this fixes it! Thank you! Couldn't test with parallel though.

skollro:
@bakerkretzmar I've just tried your proposed solution.
Registering a custom type via `database.dbal.types` works in sequential and parallel tests now.
`DB::registerDoctrineType()` in the `AppServiceProvider`'s `register` method works for sequential tests, but not for parallel tests.

bakerkretzmar:
> `DB::registerDoctrineType()` in the `AppServiceProvider`'s `register` method works for sequential tests, but not for parallel tests.

@skollro does it work if you do it in the `boot` method?

skollro:
@bakerkretzmar unfortunately, no. It seems like `getDoctrineConnection` isn't even called in parallel test environments 🤔

bakerkretzmar:
@skollro thanks. It does get called but not where I thought it did, working on fixing this now.

src/Illuminate/Database/Connection.php
```diff
@@ -1035,9 +1048,7 @@ public function registerDoctrineType(string $class, string $name, string $type):
             Type::addType($name, $class);
         }
 
-        $this->getDoctrineSchemaManager()
-            ->getDatabasePlatform()
-            ->registerDoctrineTypeMapping($type, $name);
+        $this->doctrineTypeMappings[$name] = $type;
```
bakerkretzmar, line 1051: I've kept this whole `registerDoctrineType` method for backwards compatibility but it can be removed in Laravel 9 because the same method on the DatabaseManager class replaces it. I'll PR that today too.

bakerkretzmar:
I'm pretty sure this fixes the issue now in all the cases mentioned here and in various comments: registering custom types either via `database.dbal.types` or in a service provider, and testing either normally or in parallel.

There is some code duplication here to maintain complete backwards compatibility, I'll submit another PR asap to clean this up a bit for Laravel 9. Basically, the type registration has to happen through the `DatabaseManager` so that it persists across test runs properly in parallel testing, but every single new connection that gets created also has to be provided with the types. If there's a better way to do that than to store them in an array, I'm all ears 😅

driesvints:
Thanks @bakerkretzmar & @TomHAnderson !

skollro:
Thanks! Can confirm `DB::registerDoctrineType()` and `database.dbal.types` also works with parallel testing now.
But the backwards-compatible `DB::connection()->registerDoctrineType()` still breaks parallel tests.

bakerkretzmar:
Yeah use `DB::registerDoctrineType()` or the config. I'm going to remove `DB::connection()->registerDoctrineType()` if possible.

PR #40307, opened 2022-01-07
[8.x] Resolve Blueprint class out of the container

imanghafoori1:
Currently, it is almost impossible to swap the `Blueprint $table` implementation for migrations with a fake one from within a "package space".
Because the `Schema` facade returns a new object each time from the `getFacadeAccessor` method (hence there is no way of mocking or hot swapping) and non of the other objects are resolved out of the container. (Connection objects or Builder objects)
There is also no change of using the `resolver` property on the `Builder` class from a package space because if you set it before the migration starts it will be thrown away after the `Schema` is called in the migration.

Currently "illuminate/database" depends on "illuminate/container" so we can use the Container class directly. (if not wrong)
This way it is possible to swap the Blueprint class upon the `registration phase`.

### DIFF
--- a/src/Illuminate/Database/Schema/Builder.php
+++ b/src/Illuminate/Database/Schema/Builder.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Database\Schema;
 
 use Closure;
+use Illuminate\Container\Container;
 use Illuminate\Database\Connection;
 use InvalidArgumentException;
 use LogicException;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Why do you want to write a custom Blueprint?

imanghafoori1:
This is part of the migrator UI package, we need to have a spy blueprint to see what is going to be done in migrations without actually performing them. Currently, the package is reading the migration code by regex, which is crappy.

https://github.com/rezaamini-ir/migrator

Issue #40308, opened 2022-01-08
Gte/Lte/Lt/Gt incorrectly treat strings as numeric?


sparxooo:
- Laravel Version: 8.75.
- PHP Version: 8.1.1

### Description:
When validating strings using Gte/Gt/Lte/Lt, as per the Laravel docs, it should count characters in strings, as per the size validator.

When validating "password" using rule "gte:8" it will fail
When validating "8" using rule "gte:8" it will pass
When validating "7" using rule "gte:8" it will fail
When validating "9" using rule "gte:8" it will pass

Size works correctly: 

When validating "password" using rule "size:8" it will pass
When validating "8" using rule "size:8" it will fail

If i'm doing it wrong, I apologise!

### Steps To Reproduce:

Validate a string with lte/lt/gt/gte to match lengths

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

However, this issue will not be locked and everyone is still free to discuss solutions to your problem!

Thanks.

sparxooo:
Hi driesvints,

Either the documentation is wrong, or there is a bug.

[Validation Rule GTE](https://laravel.com/docs/8.x/validation#rule-gte)

```
Rule GTE
The field under validation must be greater than or equal to the given field. The two fields must be of the same type. Strings, numerics, arrays, and files are evaluated using the same conventions as the size rule.
```

The problem is that it _does not_ evaluate it the same as the size rule. There is some more logic in the code before it uses the getSize function (and I believe part of the issue is that it checks the type of the parameter, which is always going to be a number, and so it never actually gets to calling getSize).

sparxooo:
Furthermore to this issue, i've managed (finally! :) ) to set up PHP debugging, and i've stepped through the code:

I believe that Issue #33711 and #28174 have both contributed to this change in behaviour (and this was commented about in the latter!)

It immediately falls into the below, as $parameters[0] is 8: 

[...]

PR #40309, opened 2022-01-08
Fix for the GT, GTE, LT, LTE comparison operators not comparing strings, arrays, or files to a numeric parameter

sparxooo:
Fix for the GT, GTE, LT, LTE operators not comparing to a numeric value (i.e. characters in strings, file length or array length). Fixes my issue #40308 

For proof of bug, run the tests included in this PR without the changes to ValidatesAttributes.php file.

Removes the numeric check for value (which may not be numeric if comparing strings, files or arrays). This is the case if you do not pass another attribute to compare to (i.e. just passing in a number to compare to). getSize has the logic required to deal with what type it is.

This means that the operators work correctly, as per the Laravel documentation.

Tests fully pass on my machine.

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1000,7 +1000,7 @@ trait ValidatesAttributes
 
         $this->shouldBeNumeric($attribute, 'Gt');
 
-        if (is_null($comparedToValue) && (is_numeric($value) && is_numeric($parameters[0]))) {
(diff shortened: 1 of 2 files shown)


driesvints:
You misunderstand how this works. These rules don't accept numbers but references to other fields.

sparxooo:
> You misunderstand how this works. These rules don't accept numbers but references to other fields.

You replied to my issue, but never actually said this. I didn't realise I'd misread the docs (and even then, it will still work with numeric values when it shouldn't?).

Is this likely to break any existing logic if it were to be implemented as a new feature rather than a "bug fix" as it were?

PR #40312, opened 2022-01-09
[8.x] Fix: Str::substrReplace() doesn't support multibyte

NewEXE:
Because of https://github.com/laravel/framework/pull/39988, Str::substrReplace() method doesn't support multibyte encoding.

```php
$expected = 'Ларавел – PHP-фреймворк';
$actual = \Illuminate\Support\Str::substrReplace('Ларавел фреймворк', '– PHP-фреймворк', 8);
dump($expected, $actual);

// Expected result:
// 'Ларавел – PHP-фреймворк'
// Actual result:
// 'Лара– PHP-фреймворк'
```

For now this was fixed. Thanks to https://github.com/sallaizalan/mb-substr-replace/blob/master/mbsubstrreplace.php

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -891,6 +891,7 @@ class Str
         }
 
         $start = static::substr($string, 0, $offset);
+
         $end = static::substr($string, $offset + $length, static::length($string, 'UTF-8'));
 
         return $start.$replace.$end;
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -893,7 +893,7 @@ class Str
         $start = static::substr($string, 0, $offset);
         $end = static::substr($string, $offset + $length, static::length($string, 'UTF-8'));
 
-        return $start . $replace . $end;
+        return $start.$replace.$end;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


ralphjsmit:
Isn't this more of a PHP-issue? The `Str::substrReplace()` function is just a wrapper around PHP's native `substr_replace()` function, which [doesn't support multibytes](https://www.php.net/manual/en/function.substr-replace.php#48066)..

taylorotwell:
True `substr_replace` has array support @NewEXE - which this PR doesn't account for. Overall idea is fine though.

NewEXE:
> Isn't this more of a PHP-issue? The `Str::substrReplace()` function is just a wrapper around PHP's native `substr_replace()` function, which [doesn't support multibytes](https://www.php.net/manual/en/function.substr-replace.php#48066)..

Of course this is PHP issue (and mb_* functions set is big kludge), but please see on `length()`, `title()`, `lower()`, `upper()` methods in `Support\Str` - this ones are wrappers for mb_* functions for providing multibyte strings support

NewEXE:
@taylorotwell my bad. I don't see easy way to handle arrays in params, but in **voku/portable-utf8** this issue was solved: https://github.com/voku/portable-utf8/blob/master/src/voku/helper/UTF8.php#L11890
Maybe we need to use this library as we use **voku/portable-ascii** for ASCII-related functions?..

PR #40313, opened 2022-01-09
[8.x] default value on join

abbasudo:
this pr will add a  default second value on join so you dont have to specify secod parameter


GrahamCampbell:
This is a major breaking change, and cannot go to 8.x.

PR #40320, opened 2022-01-10
[8.x] Added whenEndsWith(), whenExactly(), whenStartsWith(), etc to Stringable

telkins:
Completing the changes noted in #40285, I implemented the following (remaining) "check-like" methods on the `Stringable` class:
- `endsWith()`
- `exactly()`
- `is()`
- `isAscii()`
- `isUuid()`
- `test()`
- `startsWith()`

Please let me know if there are any questions/concerns. 🤓

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -775,17 +775,7 @@ class Stringable implements JsonSerializable
      */
     public function whenContains($needles, $callback, $default = null)
     {
-        if ($this->contains($needles)) {
-            $result = $callback($this);
-
-            return $result ?? $this;
-        } elseif ($default) {
-            $result = $default($this);
-
-            return $result ?? $this;
-        }
-
-        return $this;
+        return $this->when($this->contains($needles), $callback, $default);
     }
 
     /**
@@ -798,17 +788,7 @@ class Stringable implements JsonSerializable
      */
     public function whenContainsAll(array $needles, $callback, $default = null)
     {
-        if ($this->containsAll($needles)) {
-            $result = $callback($this);
-
-            return $result ?? $this;
-        } elseif ($default) {
-            $result = $default($this);
-
-            return $result ?? $this;
-        }
-
-        return $this;
+        return $this->when($this->containsAll($needles), $callback, $default);
     }
 
     /**
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -766,6 +766,52 @@ class Stringable implements JsonSerializable
     }
 
     /**
+     * Execute the given callback if the string contains a given substring.
+     *
+     * @param  string|array  $needles
+     * @param  callable  $callback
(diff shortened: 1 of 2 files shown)


huangdijia:
Can implement them by __call()?

telkins:
> Can implement them by __call()? like:
> 
> ```
>     public function __call($name, $args)
>     {
>         $method = substr($name, 4);
>         if ($this->$method(...$args)) {
>             $result = $callback($this);
> 
>             return $result ?? $this;
>         } elseif ($default) {
>             $result = $default($this);
> 
>             return $result ?? $this;
>         }
> 
>         return $this;
>     }
> ```

Perhaps.  I'd be a little reluctant to do it that way, though, because I don't know what problems might come about by using `__call()` for these methods.

It's a good suggestion, otherwise, though, as it's clear that all the methods have the exact same structure.  I was a bit annoyed with that but hadn't taken the time to find a good way to DRY things up a bit while still keeping things straightforward, clear, and obvious.

I did consider (re)using `when()`, which might be a nice step toward cleaning a lot of the stuff up.  I might have to play around with some of these ideas.  🤓

telkins:
I quickly tossed out all of the repetitive elements and re-used `when()`.  (I also applied this same change the the relatively new methods, `whenContains()` and `whenContainsAll()`.)

I'm not comfortable using `__call()` to further minimize the repetition.  I feel this is quite clean as-is.  I'm happy to hear from others, however, who may feel differently.  🤓

huangdijia:
```php
public function __call($name, $args)
{
    if (Str::startsWith($name, 'when')) {
        $method = lcfirst(substr($name, 4));
        if (in_array($method, ['isUuid', 'isAscii', ...])) {
            $refMethod = new \ReflectionMethod(self::class, $method);
            $argc = $refMethod->getNumberOfParameters();
            $params = array_slice($args, 0, $argc);
            $callack = $args[$argc];
            $default = $args[$argc + 1] ?? null;

            return $this->when($this->$method(...$params), $callback, $default);
        }
    }
}
```

@telkins

telkins:
> ```
> public function __call($name, $args)
> {
>     if (Str::startsWith($name, 'when')) {
>         $method = lcfirst(substr($name, 4));
>         if (in_array($method, ['isUuid', 'isAscii', ...])) {
>             $refMethod = new \ReflectionMethod(self::class, $method);
>             $argc = $refMethod->getNumberOfParameters();
>             $params = array_slice($args, 0, $argc);
>             $callack = $args[$argc];
>             $default = $args[$argc + 1] ?? null;
> 
>             return $this->when($this->$method(...$params), $callback, $default);
>         }
>     }
> }
> ```
> 
> @telkins

Hey, @huangdijia .  Thanks for the updated suggestion.  It looks pretty slick, but I'm still not sure I'm comfortable using `__call()` for this.

I'm sure you can create your own PR that changes the implementation for all these methods to this...and then see what Taylor and Crew say.  🤓 

Take care... 🚀

PR #40322, opened 2022-01-10
add whereLike function to query Builder

GeoffreySanchez:
This is a helper function that greatly simplifies writing a 'where' condition with the 'like' operator. The function
accepts 3 parameters:

-  `column`
-  `value`
- `position`

By default the value passed as a `position` parameter wrap the `value` with `%`. 

This `position` parameter can be
modified to `before` or `after` to specify where the `%` should be placed in relation to the `value`.

````php
$email = '@laravel.com';

// Default position case
User::where('email', 'like', '%'.$mail.'%');
User::whereLike('email', $email);

// Before position case
User::where('email', 'like', '%'.$mail);
User::whereLike('email', $email, 'before');

// After position case
User::where('email', 'like', $mail.'%');
User::whereLike('email', $email, 'after');
````

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1745,10 +1745,10 @@ class Builder
     /**
      * Simple and clean way to write 'where' with like operator.
      *
-     * @param string $column
-     * @param string $value
-     * @param string $position
-     * @return $this
+     * @param  string  $column
+     * @param  string  $value
+     * @param  string  $position
+     * @return  $this
      */
     public function whereLike($column, $value, $position = 'both')
     {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Illuminate/Database/Query/Builder.php [hunk 1]
```diff
@@ -1742,6 +1742,30 @@ public function whereJsonLength($column, $operator, $value = null, $boolean = 'a
...
+                $value = '%'.$value;
+                break;
+            case 'after':
+                $value .= '%';
+                break;
+            default:
+                $value = '%'.$value.'%';
+        }
+
+        return $this->where($column, 'like', $value);
```
crynobone, line 1766: not all database driver support `like` operator

taylorotwell:

PR #40323, opened 2022-01-10
[8.x] Removes PHP deprecations from tests

nunomaduro:
This pull request enforces the `E_ALL & ~E_DEPRECATED` error level in our framework tests. In other words, this ensures PHP deprecations are not reported, fixing this way a few deprecations we were seeing while running PHPUnit:

Screenshot 2022-01-10 at 12 48 47

Note that the `E_USER_DEPRECATED` level are still enabled, as we need user level deprecations to test our deprecations log channel: 

- [tests/Testing/Concerns/InteractsWithDeprecationHandlingTest.php](https://github.com/laravel/framework/blob/8.x/tests/Testing/Concerns/InteractsWithDeprecationHandlingTest.php)

This should be merged to `master` - that have a lot more warnings.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -2,6 +2,7 @@
 <phpunit backupGlobals="false"
          backupStaticAttributes="false"
          beStrictAboutTestsThatDoNotTestAnything="false"
+         bootstrap="tests/bootstrap.php"
          colors="true"
          convertDeprecationsToExceptions="true"
          convertErrorsToExceptions="true"
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -1,0 +1,5 @@
+<?php
+
+error_reporting(E_ALL & ~E_DEPRECATED);
+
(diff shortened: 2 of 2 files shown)


driesvints:
Can we use https://github.com/shivammathur/setup-php#ini-values-optional instead here? That's gonna be a bit cleaner than fiddling with code.

GrahamCampbell:
I'm not sure we actually should silence deprecations? Would it not be better to fix them, so we don't get bitten later down the line?

GrahamCampbell:
Or at least, explicitly list the deprecations we are OK with, rather than a blanket allow all.

nunomaduro:
@GrahamCampbell I am ok on having the deprecations appearing on the CLI tbh, just wanted to show Dries and Taylor a way of silence all of them.

driesvints:
@GrahamCampbell please hold on until I'm ready to PR something

GrahamCampbell:
:)

PR #40324, opened 2022-01-10
[8.x] Throw deprecation errors in CI

driesvints:
This PR enables the CI environment to also throw errors for deprecations. By doing this we both ensure we're on top of changes as soon as possible and make sure that the test suite can also be run in different environments (for example, locally) when deprecation throwing is enabled in PHP.

As soon as this is merged and 8.x is merged into master I'll also fix the tests on master.


GrahamCampbell:
Noice. 🚀

PR #36846, opened 2021-04-01
[8.x] Add tests for `LazyCollection` constructor laziness

JosephSilber:
This adds tests ensuring that:

1. Constructing a new `LazyCollection` instance from a generator function _is_ lazy.
2. Constructing a new `LazyCollection` instance from another `LazyCollection` instance _is_ lazy.
3. Constructing a new `LazyCollection` instance from a `Generator` object _is not_ lazy.

See here for reference: https://github.com/laravel/framework/issues/36837#issuecomment-811926022

Issue #36837, opened 2021-03-31
Cannot traverse already closed generator - v8.35 change broke our LazyCollection


jryd:
- Laravel Version: 8.35.1
- PHP Version: 7.4.14
- Database Driver & Version: MySQL 5.6

### Description:
We've been using LazyCollections in conjunction with a generator function to page from our APIs.

We've been doing this for about 6 months now.

In a recent upgrade from v8.34 to v8.35 some previously working code is now erroring out with:
> Cannot traverse an already closed generator

I saw in a recent PR (#36738) that some changes were made to LazyCollections. I'm not entirely sure how the changes cause our code to break, but downgrading to v8.34 (before the change was made) resolves the issue.

### Steps To Reproduce:

We have a service class that looks roughly like this:

```
class ExternalData
{
    private PendingRequest $client;

    public function __construct(PendingRequest $client)
    {
        $this->client = $client;
    }

    public function all()
    {
        return new LazyCollection($this->fetchData());
    }

    private function fetchData($page = 1)
    {
        do {
            $response = $this->client
                ->get(sprintf('/api/to/external/service?page=%s', $page++))
                ->json();

            foreach ($response['data'] as $result) { // $response['data'] is an array of objects/models
                yield $result;
            }

        } while ($response['next_page_url'] !== null);
    }
}
```

When we use this code, we've been using it like so:

```
$externalData = $this->externalData // $this->externalData is the injected ExternalData class
    ->all()
    ->filter(fn ($data) => true) // we perform actual filtering here but its enough to still cause the error
    ->count()
```

As mentioned, this code works in v8.34 but breaks in v8.35.

Let me know if you need any further information.

driesvints:
Ping @victorlap ^

@JosephSilber do you agree we should revert the PR that broke this?

victorlap:
@jryd I think this was working before because the lazy collection was not "truly" lazy. Upon initialization it would page through the whole api before continuing with the filter step (and are accessing the data more than once), that is exactly the reason I made the change. 

The problem lies in the fact that the generator is used for a second time. If you would like this code to work, you could try and instantiate a regular collection and make it lazy:
```php
Collection::make($this->fetchData())->lazy()
```

Or use the remember method:
```php
new LazyCollection($this->fetchData())->remember()
```

I believe the code works in the intended way currently, as the following test demonstrates:
```php
$iterable = function () {
            yield 'a' => 1;
            yield 'b' => 2;
            yield 'c' => 3;
        };
[...]

PR #40326, opened 2022-01-10
[8.x] Add generator support to LazyCollection

neemspees:
LazyCollection supports generator closures but not generators themselves, if you pass in a generator it gets eager loaded (turned into an array). My PR wraps the generator in a closure so it maintains the generator behavior.

```php
$source = [1, 2, 3, 4, 5];

$generatorClosure = function () use (&$source) {
    yield from $source;
};

$collection = LazyCollection::make($generatorClosure());

$source[] = 6;

$collection->all(); // before: [1, 2, 3, 4, 5]
                    // after: [1, 2, 3, 4, 5, 6]
```

An example use case would be the `cursor` method in the `Illuminate\Database\Query` class:

```php
public function cursor()
{
    if (is_null($this->columns)) {
        $this->columns = ['*'];
    }

    return new LazyCollection(function () {
        yield from $this->connection->cursor(
            $this->toSql(), $this->getBindings(), ! $this->useWritePdo
        );
    });
}
```
Becomes:
```php
public function cursor()
{
    if (is_null($this->columns)) {
        $this->columns = ['*'];
    }

    return new LazyCollection($this->connection->cursor(
        $this->toSql(), $this->getBindings(), ! $this->useWritePdo
    ));
}
```

### DIFF
--- a/src/Illuminate/Collections/LazyCollection.php
+++ b/src/Illuminate/Collections/LazyCollection.php
@@ -32,6 +32,10 @@ class LazyCollection implements CanBeEscapedWhenCastToString, Enumerable
     {
         if ($source instanceof Closure || $source instanceof self) {
             $this->source = $source;
+        } elseif ($source instanceof \Generator) {
+            $this->source = function () use ($source) {
+                yield from $source;
+            };
         } elseif (is_null($source)) {
             $this->source = static::empty();
         } else {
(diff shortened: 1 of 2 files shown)


JosephSilber (changes requested):
[It works this way by design](https://github.com/laravel/framework/issues/36837#issuecomment-811926022).

There's even [an explicit test](https://github.com/laravel/framework/pull/36846) ensuring that passing a `Generator` directly _is not_ lazy.

It should really throw TBH.

PR #40328, opened 2022-01-10
[8.x] Add generator support to LazyCollection

neemspees:
LazyCollection fix

LazyCollection supports generator closures but not generators themselves, if you pass in a generator it gets eager loaded (turned into an array). My PR wraps the generator in a closure so it maintains the generator behavior.

```php
$source = [1, 2, 3, 4, 5];

$generatorClosure = function () use (&$source) {
    yield from $source;
};

$collection = LazyCollection::make($generatorClosure());

$source[] = 6;

$collection->all(); // before: [1, 2, 3, 4, 5]
                    // after: [1, 2, 3, 4, 5, 6]
```

An example use case would be the `cursor` method in the `Illuminate\Database\Query` class:

```php
public function cursor()
{
    if (is_null($this->columns)) {
        $this->columns = ['*'];
    }

    return new LazyCollection(function () {
        yield from $this->connection->cursor(
            $this->toSql(), $this->getBindings(), ! $this->useWritePdo
        );
    });
}
```
Becomes:
```php
public function cursor()
{
    if (is_null($this->columns)) {
        $this->columns = ['*'];
    }

    return new LazyCollection($this->connection->cursor(
        $this->toSql(), $this->getBindings(), ! $this->useWritePdo
    ));
}
```

Note: I am aware of [this PR](https://github.com/laravel/framework/discussions/33297), but I don't think the reasoning is valid. If you pass in a closure that returns a Generator there is no difference to us wrapping a generator into a closure.

Please share your thoughts about this below, I'm trying to find a reason why it shouldn't be supported.

### DIFF
--- a/src/Illuminate/Collections/LazyCollection.php
+++ b/src/Illuminate/Collections/LazyCollection.php
@@ -32,6 +32,10 @@ class LazyCollection implements CanBeEscapedWhenCastToString, Enumerable
     {
         if ($source instanceof Closure || $source instanceof self) {
             $this->source = $source;
+        } elseif ($source instanceof \Generator) {
+            $this->source = function () use ($source) {
+                yield from $source;
+            };
         } elseif (is_null($source)) {
             $this->source = static::empty();
         } else {
--- a/tests/Support/SupportLazyCollectionIsLazyTest.php
+++ b/tests/Support/SupportLazyCollectionIsLazyTest.php
@@ -29,13 +29,13 @@ class SupportLazyCollectionIsLazyTest extends TestCase
         });
     }
 
-    public function testMakeWithGeneratorIsNotLazy()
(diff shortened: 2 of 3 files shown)


driesvints:
We should not change this if @JosephSilber specifically notes that this isn't wanted.

taylorotwell:

neemspees:
@driesvints, I think we should reconsider this. @JosephSilber, could you elaborate on why this is a bad idea seeing that this use case is what you would logically want to do with something called a lazy collection?

JosephSilber:
@neemspees if you follow the links I've posted, you'll see the reasoning many times over.

If nothing else, this on its own is reason enough:

```php
$collection->first() == $collection->first(); // false :(
```

driesvints:
@neemspees we won't be reconsidering this, sorry.

PR #40329, opened 2022-01-10
[9.x] Fix master CI

driesvints:


src/Illuminate/Database/Query/Grammars/Grammar.php
```diff
@@ -358,9 +358,9 @@ protected function whereBetween(Builder $query, $where)
     {
         $between = $where['not'] ? 'not between' : 'between';
 
-        $min = $this->parameter(reset($where['values']));
+        $min = $this->parameter(is_array($where['values']) ? reset($where['values']) : $where['values']);
```
driesvints, line 361: Before, `reset` would be called on both objects and arrays equally. However, this is deprecated as of PHP 8.1. I don't believe this makes sense for objects? All tests still pass so I believe this works properly?

PR #40332, opened 2022-01-10
[8.x] Add `Http::configure()` method

ryangjchandler:
This pull request introduces a new `Http::configure()` method that can be used to modify all `PendingRequest` objects globally.

```php
{
    public function boot() {
        Http::configure(function (PendingRequest $request) {
            $request->withHeaders(['custom' => 'yes!']);
        });
    }
}
```

In this scenario, all requests made from the app will have this custom header applied.

### DIFF
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -140,7 +140,7 @@ class Factory
      *
      * @param  \Closure  $callback
      * @return $this
-    */
+     */
     public function configure(Closure $callback)
     {
         $this->configureCallback = $callback;
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -72,6 +72,13 @@ class Factory
     protected $stubCallbacks;
 
     /**
+     * The callback used to configure all requests.
+     *
+     * @var \Closure
+     */
+    protected $configureCallback;
+
+    /**
      * Indicates if the factory is recording requests and responses.
      *
      * @var bool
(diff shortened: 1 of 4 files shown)


taylorotwell:
Does it make sense for this to be invoked right before the request is actually sent? That way you could inspect other information about the request before configuring it further - for example, what if you only wanted to add some headers if the URI matched a given value?

ryangjchandler:
@taylorotwell Yeah, fair point. My personal use case was for all requests, regardless of URL but you're right, that would make a lot more sense.

taylorotwell:
OK

taylorotwell:
Hmm - I dunno, something about this still kinda clunky to use it feels like. How do I inspect the request for a header or look at its final URI? How can I see what data it has?

ryangjchandler:
Happy to close this and leave it be if you think it's not worth it. There are user-land workarounds like constructing a base client that has the correct options already.

Without dedicated methods on the request object to inspect the Guzzle options, there's no easy way to do the things you mentioned.

taylorotwell:
Yeah - here's how I would do this:

In service provider:

```php
Http::macro('github', function () {
    return Http::withHeaders([
        'X-First' => 'foo',
    ])->baseUrl('https://github.com');
});
```

In code:

```php
Http::github()->get('/');
```

ryangjchandler:
@taylorotwell Yeah, this is an alternative approach. Thanks for taking a look though! :)

taylorotwell:
Embrace the dynamic language ❤️

dammy001:
🔥🔥🔥

PR #40333, opened 2022-01-10
[8.x] Makes it easy to add additional options to `PendingBatch`

orkhanahmadov:
Useful when extra identifier or parameter needed on Batch.

Use-case:
When a batch is dispatched by a user we can set `user_id` as an extra option. When authorizing the batch model we can use the value of this option.

### DIFF
--- a/src/Illuminate/Bus/PendingBatch.php
+++ b/src/Illuminate/Bus/PendingBatch.php
@@ -215,7 +215,7 @@ class PendingBatch
 
         return $this;
     }
-    
+
     /**
      * Specify additional option.
      *
@@ -226,7 +226,7 @@ class PendingBatch
     public function addOption(string $key, $value)
     {
         $this->options[$key] = $value;
-        
+
         return $this;
     }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Can you show me how you would actually use this in your own application? And where you would read the option back out?

orkhanahmadov:
In our application when a batch is dispatched we assign it to the user, then broadcast batch progress from the broadcasting channel. When the user tries to join the broadcasting channel we authorize if the user "owns" the batch:
```php
public function join(User $user, string $batchId): bool
{
    $bus = Bus::findBatch($batchId);

    return $bus?->options['user_id'] == $user->id;
}
```
Reading options is easy, since it is a public property on `\Illuminate\Bus\Batch`. But setting additional options isn't straightforward. Currently we do it like this:
```php
$batch = new PendingBatch($this->container, $jobs);
$batch->options = ['user_id' => $user->id];

$batch->then(fn (Batch $batch) => ...)->dispatch();
```

Having `addOption` method will simplify the code to:
```php
Bus::batch($jobs)
    ->then(fn (Batch $batch) => ...)
    ->addOption('user_id', $user->id)
    ->dispatch();
```

PR #40335, opened 2022-01-10
[8.x] Make assertPath() check if $path exists

paulandroshchuk:
This PR updates the behavior of the `assertJsonPath()` method to make it fail when `$expect` is `null` and `$path` is not a real path.

This change will prevent from creating assertions that will always pass:
```php
// Response
[
  data: [
    [id: int, first: null],
    ...
  ]
]

// Before
->get('/users')
  ->assertJsonPath('data.first', null) // This will always pass.

// After
->get('/notes')
  ->assertJsonPath('data.0.first', null) // This will pass.
  ->assertJsonPath('data.first', null) // This will fail: Unable to find path: user within JSON: {data: ...}
```

Notes:
* It will potentially make some tests fail in users' projects but only if their `$expected` is `null` and `$path` does not exist in the response.
* I'll add tests if this idea is approved.

### DIFF
--- a/src/Illuminate/Testing/AssertableJsonString.php
+++ b/src/Illuminate/Testing/AssertableJsonString.php
@@ -219,6 +219,19 @@ class AssertableJsonString implements ArrayAccess, Countable
      */
     public function assertPath($path, $expect)
     {
+        if (! Arr::has($this->json(), $path)) {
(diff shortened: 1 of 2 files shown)


taylorotwell:
Breaking change as you note.

PR #40336, opened 2022-01-10
[8.x] Update the Pluralizer Inflector instanciator

cesarep:
Uses the InflectorFactory::createForLanguage function to generate the instance

Should make it easier to change to other available languages rulesets such as French, Norwegian Bokmal, Portuguese, Spanish and Turkish

### DIFF
--- a/src/Illuminate/Support/Pluralizer.php
+++ b/src/Illuminate/Support/Pluralizer.php
@@ -2,9 +2,8 @@
 
 namespace Illuminate\Support;
 
-use Doctrine\Inflector\CachedWordInflector;
+use Doctrine\Inflector\InflectorFactory;
 use Doctrine\Inflector\Inflector;
-use Doctrine\Inflector\Rules\English;
 use Doctrine\Inflector\RulesetInflector;
 
 class Pluralizer
@@ -136,14 +135,7 @@ class Pluralizer
         static $inflector;
 
         if (is_null($inflector)) {
-            $inflector = new Inflector(
-                new CachedWordInflector(new RulesetInflector(
-                    English\Rules::getSingularRuleset()
-                )),
-                new CachedWordInflector(new RulesetInflector(
-                    English\Rules::getPluralRuleset()
-                ))
-            );
+            $inflector = InflectorFactory::createForLanguage('english')->build();
         }
 
         return $inflector;


src/Illuminate/Support/Pluralizer.php
```diff
@@ -136,14 +135,7 @@ public static function inflector()
...
         if (is_null($inflector)) {
-            $inflector = new Inflector(
-                new CachedWordInflector(new RulesetInflector(
-                    English\Rules::getSingularRuleset()
-                )),
-                new CachedWordInflector(new RulesetInflector(
-                    English\Rules::getPluralRuleset()
-                ))
-            );
+            $inflector = InflectorFactory::createForLanguage('english')->build();
```
seriquynh, line 138: I think we should use their language constants.

```php
<?php

use Doctrine\Inflector\InflectorFactory;
use Doctrine\Inflector\Language;

$inflector = InflectorFactory::createForLanguage(Language::ENLISH)->build();

```

cesarep: My main goal actually was to make it easier to change the language ruleset, just like changing the locale in the `config/app.php` file, adding another configuration for the pluralizer ruleset. I was thinking something like:
```php
$inflector = InflectorFactory::createForLanguage(  $app['config']->get('app.pluralizer_ruleset', 'english')  )->build();
```
But I'm not sure how to fully implement it...
Other classes appear to get the config values from `$this->app`, however class Pluralizer does not extends any other class, so `$this` shoudn't have access to the app right?

cesarep:
I included the `use Illuminate\Support\Facades\Config;` on top, and
`$inflector = InflectorFactory::createForLanguage(Config::get('app.pluralizer_ruleset', 'english'))->build();`
 
and added `'pluralizer_ruleset' => 'portuguese',` to the `config/app.php` to test on my machine, it appers to be working fine!

I just had to mannualy set the users table name on the `User` Model

Should i make another pull request or is there a way to reopen it? I didn't expect it to be closed so quickly 😅

For me it's a very useful feature, not having to manually set a lot of parameters and table names because of different language plurals

driesvints:
@cesarep a facade can't be mixed here. This class may be used outside Laravel. 

If you aren't ready with a PR then please use draft PR's to mark them as in progress.

PR #40341, opened 2022-01-11
[8.x] Add a method to sort keys in a collection using a callback

redelschaap:
I added a method to the Collection classes to sort the keys of the items using a callback. For example, this helps you to sort item keys case insensitive.

With this method, the following code:

```php
$items = $collection->all();

uksort($items, 'strnatcasecmp');

$items = collect($items);
```

can be replaced with this code:

```php
$collection->sortKeysUsing('strnatcasecmp');
```

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -1382,6 +1382,21 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
+     * Sort the collection keys using a callback.
+     *
+     * @param  callable  $callback
+     * @return static
+     */
+    public function sortKeysUsing(callable $callback)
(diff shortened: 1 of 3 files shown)


taylorotwell:
I can already do this via `sortBy`?

CleanShot 2022-01-11 at 09 32 49@2x

redelschaap:
@taylorotwell Not entirely. The new method provides much more flexibility. Take this example to sort items where keys represent version number strings:

image

Your example does not sort this correctly. However,

```php
collect([...])->sortKeysUsing('version_compare');
```

would do the trick.

redelschaap:
@taylorotwell @driesvints With the argument in my last reply, would this feature be worth reconsidering?

driesvints:
@redelschaap best to just attempt a new PR.

PR #40343, opened 2022-01-11
[8.x] Update Logger for multi-driver/multi-channels at once

Sunhat:
When the logger is interacted with, it can be beneficial for these messages to go to multiple places.

An example is I work as a PHP contractor, and connect logging into my Company Slack, but my client may also wish to know when an error occurs.

Multi-channel/Multi-driver support enables multiple logs being called without a copy/paste

### DIFF
--- a/src/Illuminate/Log/LogManager.php
+++ b/src/Illuminate/Log/LogManager.php
@@ -112,6 +112,7 @@ class LogManager implements LoggerInterface
         if (is_array($driver)) {
             return new MultiChannelLogger($driver, $this);
         }
+
         return $this->get($this->parseDriver($driver));
     }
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
This is not necessary. You can already use the stack logger.

GrahamCampbell:
https://laravel.com/docs/8.x/logging#building-log-stacks

Sunhat:
@GrahamCampbell Thanks - I didn't know that was an option. 

I can see how it can be a duplicate feature, but there's a million ways to do the same things in Laravel that are completely unnecessary and bloat the framework. I don't see how this can be considered unnecessary, when Laravel has a lot of unnecessary additions.

In addition to this, the stack logger is hidden in documentation, whereas this feature is visible when you're typing from your IDE.

GrahamCampbell:
The stack logger is used in every app, out of the box, and is the default when you go into the logging configuration. I wouldn't call that hidden away. https://github.com/laravel/laravel/blob/8.x/config/logging.php#L20

Sunhat:
I think one can consider string-based config files to be overused and obscure in the Laravel framework.

PR #40344, opened 2022-01-11
[8.x] Prevents `TypeError` when validating int backed enum with string value

orkhanahmadov:
Let's say we have the following enum:
```php
enum Status: int
{
    case pending = 1;
    case done = 2;
}
```

We want to validate the incoming request against the values in this enum with the following ruleset:
```php
['status' => ['required', new \Illuminate\Validation\Rules\Enum(Status::class)]]
```

Currently if from application receives `status` as anything other than an integer, like a string of "abc", the application throws `TypeError` exception, because of the enum's `tryFrom()` usage. For string enums, this method expects a string, for integer enums it expects integer only.

It is possible to prevent `TypeError` by adding the `integer` and `bail` rules to the ruleset
```php
['status' => ['bail', 'required', 'integer', new \Illuminate\Validation\Rules\Enum(Status::class)]]
```
This way validation will not go into enum value check if anything other than an integer is passed.

This PR offers a solution to prevent the `TypeError` error completely by using the `in_array` check against enum values, instead of using the `tryFrom()` method. Even when a string is passed to integer-backed enum, validation will fail without throwing an exception, no need to use `bail`.

### DIFF
--- a/src/Illuminate/Validation/Rules/Enum.php
+++ b/src/Illuminate/Validation/Rules/Enum.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Validation\Rules;
 
+use BackedEnum;
 use Illuminate\Contracts\Validation\Rule;
 
 class Enum implements Rule
@@ -37,7 +38,7 @@ class Enum implements Rule
             return false;
         }
 
-        return ! is_null($this->type::tryFrom($value));
+        return in_array($value, array_map(fn (BackedEnum $enum) => $enum->value, $this->type::cases()));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
But this means validation passes if you receive a string `"1"` and the Enum value is an integer `1`. I'm not sure that is desired or expected?

taylorotwell:
Can we just catch the TypeError and fail validation?

orkhanahmadov:
> But this means validation passes if you receive a string `"1"` and the Enum value is an integer `1`. I'm not sure that is desired or expected?

@taylorotwell Good point. Initially I wanted to use `in_array` with strict mode but from test coverage I see that `ValidationEnumRuleTest::testValidationPassesWhenProvidingDifferentTypeThatIsCastableToTheEnumType` expects type casting and loose comparison.

Another option would be adding an optional `$strict` argument to `\Illuminate\Validation\Rules\Enum` constructor. Making it `false` by default will be backward compatible. When set to `true` it will run `in_array` in strict mode.

taylorotwell:
Fixed by @themsaid

Issue #40345, opened 2022-01-11
Artisan command make:policy generates invalid policy code with a non-default user provider configuration


slavicd:
- Laravel Version: 8.78.1
- PHP Version: 7.4, 8.0

### Description:

`php artisan make:policy` will produce invalid output, similar to what's been reported [here](https://github.com/laravel/framework/issues/32647), [here](https://github.com/laravel/framework/issues/32397) and [here](https://github.com/laravel/framework/issues/21117) (missing user class) if a non-default `auth.providers.users` configuration is in place.

### Steps To Reproduce:

1. `composer create-project laravel/laravel`
2. Change the user provider configuration. E.G: 
```php
// config/auth.php
// ...
    'providers' => [
        'users' => [
            'driver' => 'database',
            'table' => 'users',
        ],
    ],
// ...
```
3. `php artisan make:policy CommentPolicy --model=Comment`

Note the generated code in Policies: it is missing the User class and is invalid. 

I realized this happens when the provider configuration is missing the `model` key, which was my case too (except I had a custom driver, and I never needed the `model` key).

### Expected behaviour

The `make:policy` command throws an error if it needs explicit user class configuration or otherwise inserts a default user class.

hmennen90:
Just tested it with a new Laravel Project:

```
'providers' => [
        //'users' => [
        //    'driver' => 'eloquent',
        //    'model' => App\Models\User::class,
        //],

         'users' => [
             'driver' => 'database',
             'table' => 'users',
         ],
    ],
```

Output:
```

❯ php artisan make:policy CommentPolicy --model=Comment
Policy created successfully.
```

slavicd:
@hmennen90 check the generated policy code.

hmennen90:
The Target Models are in use Statements using App\Models Namespace in both cases: -m and --model

If no Model is specified, there is a blank Policy created

```
use App\Models\Comment;
use App\Models\User;
use Illuminate\Auth\Access\HandlesAuthorization;
```

hmennen90:
hmm i try it on php 8.0 - i'm currently on 8.1.1

hmennen90:
Results on PHP8.0 and PHP7.4 are same as at PHP8.1.1

Maybe anyone else can reproduce your issue?

hmennen90:
@slavicd command back ;-) Can now confirm your issue also on PHP 8.1.1 - had the config cached...

Did you mean the following Output?

```
namespace App\Policies;

use App\Models\Comment;
[...]

PR #40346, opened 2022-01-11
[8.x] Return Default User::class FQCN if no model key is defined in config

hmennen90:
As issued in #40345 this is a first idea to solve this Problem by return the default value of model Key in auth.php['providers']['users']['model'] using database instead of eloquent.


driesvints:
As I said in #40345, there is no real bug here. You should add a model key if you want to generate policies.

hmennen90:
Alright :-)

slavicd:
> You should add a model key if you want to generate policies.

And one should learn about this fact where? In this very pull request comment section?

PR #40350, opened 2022-01-12
[8.x] Adds select columns for Pivot

DarkGhostHunter:
## What?

Allows to override pivot table selection. Useful when a relationship defines `withPivot()`, and later these can be discarded.

```php
public function tags() {
    return $this->belongsToMany(Tag::class)->withPivot('name', 'visible');
}

// Don't get the visible property
$post->tags()->selectPivot('name');

// Or don't get anything at all.
$post->tags()->withoutPivotColumns();
```

## Why?

Because there is no way to override the columns to get. `withPivot()` is additive.

## BC?

Nope, only additive.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
@@ -587,6 +587,19 @@ trait InteractsWithPivotTable
     }
 
     /**
+     * Override the columns on the pivot table to retrieve.
+     *
+     * @param  array|mixed  $columns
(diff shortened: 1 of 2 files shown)


DarkGhostHunter:
Halp. It fails _because_.

devfrey:
This is already possible, so I don't think we really need more methods:

```php
public function tags(array $columns = ['name', 'visible']) {
    return $this->belongsToMany(Tag::class)->withPivot($columns);
}

// Don't get the visible property
$post->tags(['name']);

// Or don't get anything at all.
$post->tags([]);
```

taylorotwell:

PR #40361, opened 2022-01-12
[8.x] Fixes stub publishing

jspekken:
This fixes when using custom listener stubs in a Laravel application.


jspekken:
Published stub files where not used when running the `php artisan make:listener ...` command. Other generator commands that derive from the `GeneratorCommand` class do have this functionality (look at classes like `ModelMakeCommand` or `MailMakeCommand` or `PolicyMakeCommand` etc...). They usually have a `resolveStubPath()` method. The `ListenerMakeCommand` does not have this functionality so running `php artisan make:listener` will use stubs in the framework folder instead of the published stubs in the application (like other generator commands do). 

This PR fixes this functionality and copies the `resolveStubPath()` used in other generators and mirrors the functionality like those generators. You'll now be able to use your customized stubs like other generators.

PR #40362, opened 2022-01-12
[8.x] Handle type mismatch in the enum validation rule

themsaid:


src/Illuminate/Validation/Rules/Enum.php
```diff
@@ -37,7 +38,11 @@ public function passes($attribute, $value)
             return false;
         }
 
-        return ! is_null($this->type::tryFrom($value));
+        try {
+            return ! is_null($this->type::tryFrom($value));
+        } catch (Throwable $e) {
```
nunomaduro, line 43: You can be a little bit more specific here, using [TypeError](https://www.php.net/manual/en/class.typeerror.php).

PR #40363, opened 2022-01-12
[8.x] Add forget method to config repository

LasseRafn:
This adds a `forget($key)` method to the config repository, which can be called like so: `config()->forget($key)`. It removes the config key from the config (similarly to how we can `set` config values at runtime).

### DIFF
--- a/tests/Config/RepositoryTest.php
+++ b/tests/Config/RepositoryTest.php
@@ -131,6 +131,12 @@ class RepositoryTest extends TestCase
         $this->assertSame('value2', $this->repository->get('key2'));
     }
 
+    public function testCanForget()
+    {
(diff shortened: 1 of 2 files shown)


tests/Config/RepositoryTest.php
```diff
@@ -131,6 +131,12 @@ public function testSetArray()
         $this->assertSame('value2', $this->repository->get('key2'));
     }
 
+    public function testCanForget()
+    {
+        $this->repository->forget('foo');
```
BrandonSurowiec, line 136: Shouldn't this test also set `foo` first?

```
$this->repository->set('foo', 'value');
```

taylorotwell:
Interesting - what was your use case here?

PR #40368, opened 2022-01-12
[9.x] Change default behavior of unvalidated array keys

taylorotwell:
Makes [this](https://laravel.com/docs/8.x/validation#excluding-unvalidated-array-keys) the default behavior - new method added to opt-in to old behavior for easy backwards compatibility.


ankurk91:
Off the topic but 

```php
Model::preventLazyLoading(! app()->isProduction());
```

Should be added to Skelton as well

driesvints (approved):

PR #40369, opened 2022-01-12
New: Email Column Shorcut Creation

devajmeireles:
This PR introduces the ability to **create email columns via a direct shortcut**, optionally setting the column to `unique` and also controlling the size via the `length` parameter, this saving time in writing email columns using the `string` method (the method known and used until today)

### Use Cases

**Before:**
```
$table->string('email')->unique();
```

**With the PR proposal:**
```
$table->email();
```

---

### Alternatives _(based on PHP 8+)_:

1. Setting a different name for the column:
```
$table->email('destination_email');
```

2. Disabling the `unique`:
```
$table->email(unique: false);
```

3. Controling the length to the column:
```
$table->email(length: 50);
```

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -608,6 +608,27 @@ class Blueprint
     }
 
     /**
+     * Create a new email column on the table.
+     *
+     * @param  string  $column
+     * @param  bool  $unique
+     * @param  int|null  $length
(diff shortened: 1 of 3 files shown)


driesvints:
I don't think we should be mixing in user land concepts with database schema's. Email isn't a real database column type in any engine afaik? Maybe I'm wrong.

taylorotwell:

PR #40375, opened 2022-01-12
[8.x] Allow enums as `entity_type` in morphs

damek24:
Laravel allows casting model properties to PHP native enums. However currently it is not possible that morph [entity type with custom polimorphic type](https://laravel.com/docs/8.x/eloquent-relationships#custom-polymorphic-types) can be a native PHP enum. This PR allow that behavior

## Steps to reproduce (Laravel 8.78, PHP 8.1)

1. create-project laravel/laravel
2. php artisan make:model Player
3. php artisan make:model Event
4. create php enum in app/Enums directory
```
enum EventType
{
    case player = 'Player';
}
``` 
5. modify models as follows:
```
/**
 * @property string Name
 */
class Player extends Model
{
    use HasFactory;
}
```

```
<?php 

use App\Enums\EventType;
use Illuminate\Database\Eloquent\Relations\MorphTo;

/**
 * @property Team|Player $entity
 * @property EventType $entity_type
 */
class Event extends Model
{
    use HasFactory;

    public $casts = [
        'entity_type' => EventType::class
    ];

    public function entity(): MorphTo
    {
        return $this->morphTo();
    }
}
```
6. add to AppServiceProvider::boot following code:
```
use Illuminate\Database\Eloquent\Relations\Relation;
        Relation::morphMap(
            [
                'Player' => Player::class,
            ]
        );
```
7. insert some data to database
8. try to get first Event model with entity relation
Instead of model the exception is thrown:
`Model attribute value is an object but does not have a __toString method.`

## Important note
We cannot use in the enum magic "__toString" method because this was not allowed by [enum rfc](https://wiki.php.net/rfc/enumerations)

## Behavior after PR is merged
the enum file changes to
```
use Illuminate\Contracts\Database\Eloquent\StringableAttribute;

enum EventType: string implements StringableAttribute
{
    case player = 'Player';

    public function toString(): string
    {
        return $this->value;
    }
}
```
and the code above works :)


taylorotwell:
Why can you not just check if it is an Enum class and call ->value on it directly? Why do you need a new interface?

damek24:
I cannot call value directly because the exception is thrown in core Laravel's trait 
`Illuminate\Database\Eloquent\Relations\Concerns\InteractsWithDictionary`
enums cannot have __toString and an object cannot be an array key. Because of this I created a new interface that requires "custom" toString method

taylorotwell:
Personally I would assume it is a backed enum and use it's value. I don't want to introduce a new interface for this.

damek24:
@taylorotwell agreed
however I did not find any official clue how to know whether if this is a native enum. I've found that PHP storm uses `BackedEnum` interface however I didn't found anything like that official.
Do you have any clues how can I detect if an attribute is normal class or a new php8.1 enum ?

damek24:
I removed a new interface and as you suggested used enum's value

taylorotwell:
Tests broken on lower PHP versions because the entire test file can't parse. See failures.

damek24:
tests passed however CI failed. @taylorotwell  I am developping on windows 11. what line endings should be set (notepad ++ shows LS in both files) ?
could you give me any clue what should I set in my github settings to make the last check working?

taylorotwell:

src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithDictionary.php
```diff
@@ -21,9 +21,13 @@ protected function getDictionaryKey($attribute)
                 return $attribute->__toString();
             }
 
+            if (isset($attribute->value) && (is_string($attribute->value) || is_numeric($attribute->value))) {
+                return $attribute->value;
+            }
```
crynobone, line 26: Should just check if `$attribute instanceof \BackedEnum`

PR #40382, opened 2022-01-13
[10.x] Prepare Laravel 10

driesvints:


GrahamCampbell:
Is Laravel 10 going to stay PHP `^8.0.2`, or go to `^8.1`?

driesvints:
@GrahamCampbell 8.0.2 because EoL is only at the end of 2023 so Laravel 11 will drop it: https://www.php.net/supported-versions.php

GrahamCampbell:
OK, cool. I guess we'll see what happens with Symfony 7. Our hand may be forced to go higher, lol.

GrahamCampbell (approved):

Issue #40364, opened 2022-01-12
Custom attribute accessor yields "column does not exist" error


phanan:
- Laravel Version: 8.78.1
- PHP Version: 8.0
- Database Driver & Version: Irrelevant

### Description:

With [reference](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L622-L638) to the method `mutateAttributeMarkedAttribute` found in the `HasAttributes` trait [introduced](https://github.com/laravel/framework/pull/40022) with the new attribute casting feature in v8.77:

```php
    protected function mutateAttributeMarkedAttribute($key, $value)
    {
        if (isset($this->attributeCastCache[$key])) {
            return $this->attributeCastCache[$key];
        }

        $value = call_user_func($this->{Str::camel($key)}()->get ?: function ($value) {
            return $value;
        }, $value, $this->attributes);

        if (! is_object($value)) {
            unset($this->attributeCastCache[$key]);
        } else {
            $this->attributeCastCache[$key] = $value;
        }

        return $value;
    }
```

It seems to me that the calculated `$value` is only cached if it's an object, which essentially means any `callable` Attribute that returns scalar values (string, boolean etc.) will _not_ be cached. I'm not sure of the reason behind this, but even with this aside, it seems the logic here introduces a bug. Considering the method `mergeAttributesFromAttributeCasts()` in the same trait, which is called upon saving/updating a model:

```php
    protected function mergeAttributesFromAttributeCasts()
    {
        foreach ($this->attributeCastCache as $key => $value) {
            $callback = $this->{Str::camel($key)}()->set ?: function ($value) use ($key) {
                $this->attributes[$key] = $value;
            };

            $this->attributes = array_merge(
                $this->attributes,
                $this->normalizeCastClassResponse(
                    $key, call_user_func($callback, $value, $this->attributes)
                )
            );
        }
    }
```

If I read it correctly, this method will merge `$attributeCastCache` back to the model before performing the query. Now if in the model I have an attribute accessor like this:

```php
// User.php
protected function foo(): Attribute
{
   return Attribute::get(fn () => new Foo());
}
```

Once `$user->foo` is accessed, a `'foo'` key will be stored into `$attributeCastCache` and merged into `$attributes`. If my table doesn't have a `foo` column, however, saving the model will yield an error saying the column `foo` doesn't exist.

[...]

PR #40386, opened 2022-01-13
[8.x] Models with constructor parameters throw an error when attributes are casted

thijsvdanker:
When your model has constructor parameters, the new attribute casting logic produces an error.
It tries to create a new class but it does not provide the constructor parameters.

```php
protected static function getAttributeMarkedMutatorMethods($class)
{
   $instance = is_object($class) ? $class : new $class;
```

The error
```
1) Illuminate\Tests\Integration\Database\DatabaseEloquentModelAttributeCastingTest::testCastingModelWithConstructorParameters
ArgumentCountError: Too few arguments to function Illuminate\Tests\Integration\Database\TestEloquentModelWithConstructor::__construct(), 0 passed in /code/framework/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php on line 2099 and exactly 1 expected
```

This pull request only contains a failing test.

### DIFF
--- a/tests/Integration/Database/DatabaseEloquentModelAttributeCastingTest.php
+++ b/tests/Integration/Database/DatabaseEloquentModelAttributeCastingTest.php
@@ -188,6 +188,16 @@ class DatabaseEloquentModelAttributeCastingTest extends DatabaseTestCase
 
         $this->assertTrue(empty($model->getDirty()));
     }
+
+    public function testCastingModelWithConstructorParameters()
+    {
+        $model = new TestEloquentModelWithConstructor('__test__');
+        $model->uppercase = 'taylor';
+
+        $this->assertSame('TAYLOR', $model->uppercase);
+        $this->assertSame('TAYLOR', $model->getAttributes()['uppercase']);
+        $this->assertSame('TAYLOR', $model->toArray()['uppercase']);
+    }
 }
 
 class TestEloquentModelWithAttributeCast extends Model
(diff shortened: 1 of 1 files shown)


driesvints:
Is this for https://github.com/laravel/framework/issues/40364 ?

thijsvdanker:
No, I have not created an issue for this and that issue looks unrelated.
Do you want me to create an issue?

driesvints:
@thijsvdanker no it's fine.

driesvints:
I tried to find a solution to this but I think there isn't any. I think model's shouldn't have custom constructors.

taylorotwell:
We have always recommended models do not have custom constructors since inception of Laravel.

SjorsO:
> We have always recommended models do not have custom constructors since inception of Laravel.

Unfortunately there are a lot of developers who did not get this memo

thijsvdanker:
@SjorsO it doesn't matter if you use the new accessors or not. The models fail regardless.
I will work around it by adding

```php
 protected static function getAttributeMarkedMutatorMethods($class)
    {
        return [];
    }
```

to the models that have their own constructor.

thijsvdanker:
@driesvints I agree, and can see there isn't a good solution to this issue.

In my case we've implemented the HasAttributes trait on something other than an eloquent model, which worked fine until the new accessors.

Anyway, it's case closed. Thanks for having a look!

SjorsO:
> it doesn't matter if you use the new accessors or not. The models fail regardless.

I guess that's a good excuse to finally refactor those models 😄 

Would probably be a good idea to document this in the 9.x upgrade guide

driesvints:
I don't think we're gonna make any changes here sorry. Please do not use custom constructors on models.

PR #40388, opened 2022-01-13
[8.x] Arr::get contains a logic error

stijn1989:
image
The $array[$key] there isn't logic because $key simply doesn't exists in $array on this line. If it did exist, it is returned in the previous if statement.


stijn1989:
Why is this closed? Is it wrong?

PR #40393, opened 2022-01-13
[9.x] Uses `static` instead `static<TKey, TValue>`

canvural:
This PR changes the `static<TKey, TValue>` syntax to `static`. Some tools do not support generic static PHPDocs. And since they both have the same meaning, there is no braking change. 

Static analysis tool results or IDE completion is also not effected.


nunomaduro (approved):
This pull request is correct. ✅

PR #40396, opened 2022-01-13
[8.x] Add complimentary `Str::doesntContain`

jasonmccreary:
Similar to the recent additions to the `Collection` class, this PR adds a complimentary `doesntContain` method. This may be used to test a string does not contain a substring or set of substrings, which allow the developer to write more expressive code.

**Before**
```php
if (! Str::contains('foo', 'bar')) {
    // ...
}

if (! Str::containsAll('foo', ['bar', 'baz'])) {
    // ...
}
```

**After**
```php
if (Str::doesntContain('foo', 'bar')) {
    // ...
}

if (Str::doesntContain('foo', ['bar', 'baz'])) {
    // ...
}
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -208,6 +208,30 @@ class Str
     }
 
     /**
+     * Determine if a given string does not contain a given substring.
+     *
+     * @param  string  $haystack
+     * @param  string|string[]  $needles
+     * @return bool
+     */
+    public static function doesntContain($haystack, $needles)
+    {
+        return ! static::contains(...func_get_args());
+    }
+
+    /**
+     * Determine if a given string does not contain any array values.
+     *
+     * @param  string  $haystack
+     * @param  string[]  $needles
+     * @return bool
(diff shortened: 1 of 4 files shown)


foremtehan:
Can you also add those to Stringable class ? https://github.com/laravel/framework/blob/8.x/src/Illuminate/Support/Stringable.php

jasonmccreary:
@foremtehan, you got it.

taylorotwell:
Going to hold off on this for now - mainly because the wording around it confuses my brain when I start thinking about possible methods for a method that would determine if a string doesn't contain at least one of the given strings.

jasonmccreary:
No problem. Negated methods can be rather mind bending. Especially with parameter overloading like this one. I really only chose this name because you used it [for the same method on the Collection class](https://github.com/laravel/framework/commit/3e3cbcf4cb4b8116f504a1e8363c7c958067b49a). I'd probably call have called it `missing`, `doesntContainAny`, or `doesntInclude`.

PR #40398, opened 2022-01-13
Add `is_environment` helper

jasonmccreary:
While a similar method may be called through the `App` facade, I find I initially reach for an `environment` helper method. In addition, the overloading nature of the `App::environment()` method makes me forget how exactly to call it.

I thought this might be a welcome addition to the core helpers to clarify a common use case during development/testing.

**Before**
```php
if (App::environment('staging')) {
    // ...
}
```

**After**
```php
if (is_environment('staging')) {
    // ...
}
```

### DIFF
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -146,6 +146,20 @@ if (! function_exists('filled')) {
     }
 }
 
+if (! function_exists('is_environment')) {
+    /**
+     * Determine if the current environment is
+     * one of the passed environments.
+     *
+     * @param  mixed  $environments
+     * @return bool
+     */
+    function is_environment(...$environments)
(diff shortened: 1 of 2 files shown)


BrandonSurowiec:
For Blade there are `@production` and `@env()` helpers already:
https://laravel.com/docs/8.x/blade#environment-directives

```
@production
    // Production specific content...
@endproduction
```

```
@env('staging')
    // The application is running in "staging"...
@endenv

@env(['staging', 'production'])
    // The application is running in "staging" or "production"...
@endenv
```

jasonmccreary:
@BrandonSurowiec, yes, this was a poor off-the-cuff example, I'll update it.

taylorotwell:
Think I'll leave this one out for now.

PR #40404, opened 2022-01-14
[8.x] `Rule::requiredIf` should accept callable string

dinhquochan:
Dear,

Currently, the `$condition` param accepts `callable` or `boolean` type but it only checks `is_string` and it ignore the callable string.
This PR make it works with callable string (function name).

Thanks.

### DIFF
--- a/tests/Validation/ValidationRequiredIfTest.php
+++ b/tests/Validation/ValidationRequiredIfTest.php
@@ -21,6 +21,14 @@ class ValidationRequiredIfTest extends TestCase
 
         $this->assertSame('', (string) $rule);
 
+        $rule = new RequiredIf('Illuminate\Tests\Validation\example_required_if_true_test');
+
+        $this->assertSame('required', (string) $rule);
+
+        $rule = new RequiredIf('Illuminate\Tests\Validation\example_required_if_false_test');
+
+        $this->assertSame('', (string) $rule);
+
         $rule = new RequiredIf(true);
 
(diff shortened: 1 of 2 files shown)


tests/Validation/ValidationRequiredIfTest.php
```diff
@@ -38,7 +46,7 @@ public function testItOnlyCallableAndBooleanAreAcceptableArgumentsOfTheRule()
 
         $this->expectException(\InvalidArgumentException::class);
 
-        $rule = new RequiredIf('phpinfo');
+        $rule = new RequiredIf('Taylor Otwell');
```
driesvints, line 49: This is a change in behaviour. I'm not sure if it's a good idea to send this in to 8.x

dinhquochan: Yeah. But I think it isn't B.C

bangnokia: I think this could be a security problem when we let user passes callable from the input
```php
'name' => new RequireIf(['rmdir', '/some/directory'])
```

driesvints: It'll now invoke any string that matches a native or non-native function.

dinhquochan: @bangnokia It's accepted callable array, closure before. The end-user can not send this to the param.

taylorotwell:
No plans to change this.

PR #40405, opened 2022-01-14
[8.x] DatabaseUuidFailedJobProvider: switch docblock returntype to string for uuid

danielsimkus:
Currently the docblock for the log method on the DatabaseUuidFailedJobProvider states the return as `int|null` when in-fact it returns a string `$uuid`.
I've just updated that so it's more accurate 👍


driesvints (approved):

danielsimkus:
Thanks @driesvints that was so fast! 👏 

There's a test failing, but I don't think it's anything to do with what I've changed 😕

driesvints:
@danielsimkus that's a small hiccup that sometimes occurs in our CI. I've re-triggered the build for you 👍

PR #40411, opened 2022-01-14
[9.x] Flysystem v3 support

driesvints:
This PR introduces Flysystem v3 support for Laravel 9. I've implemented the new `has` method that checks for both a file and directory. I've updated the subsequent methods docblock descriptions and assertion messages to make that more clear. It's how the behaviour is now in Laravel 8.

I've also added complementary `fileExists`, `directoryExists` and their `missing` equivalents. I didn't add `assertExists` equivalents because that felt a little overkill.


src/Illuminate/Filesystem/FilesystemAdapter.php
```diff
@@ -141,22 +141,22 @@ public function assertMissing($path)
...
-     * Determine if a file exists.
+     * Determine if a file or directory exists.
      *
      * @param  string  $path
      * @return bool
      */
     public function exists($path)
     {
-        return $this->driver->fileExists($path);
+        return $this->driver->has($path);
```
driesvints, line 159: This was how it's currently on 8.x:

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Filesystem/FilesystemAdapter.php#L127

PR #40413, opened 2022-01-14
[9.x] Add backoff functionality to retry helper

hebinet:
This PR enables the retry helper function to also use an array as first argument in order to implement a backoff strategy

After this PR you can use the helper like this
```php
return retry([50, 100, 200], function () {
    // ...
});
```

As it is now, you have to use the following implementation
```php
return retry(4, function () {
    // ...
}, fn($attempt) => [50, 100, 200][$attempt - 1]);
```
I like the backoff approach of Queues Jobs and I think this would also be a good addition to the framework without breaking anything.


taylorotwell:

PR #37941, opened 2021-07-07
[9.x] Update to Symfony v6 and drop PHP 7.4 support

driesvints:

PR #40415, opened 2022-01-14
[9.x] Fixes `Command::setHidden()`

nunomaduro:
This pull request fixes the `Command::setHidden()` method that was not actually hidden the command from the `list` command.

This regression was introduced on the https://github.com/laravel/framework/pull/37941 pull request, and I've added a few tests to ensure this don't happen in the future.


owenvoke:
Thanks for adding this back Nuno. 👍🏻

driesvints:
hmm yeah that could indeed be it.

PR #40416, opened 2022-01-14
[9.x] Fixes `containsStrict` and `modelKeys` collection method typehints

canvural:
- `modelKeys` will return integer or string (for example UUID keys) So it makes sense to change `mixed` to `array-key`
- `containsStrict` can also accept integer in it's key. `collect([[1]])->containsStrict(0, 1)` for example would check if `0` keyed item contains value `1`

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -195,7 +195,7 @@ trait EnumeratesValues
     /**
      * Determine if an item exists, using strict comparison.
      *
-     * @param  (callable(TValue): bool)|TValue|string  $key
+     * @param  (callable(TValue): bool)|TValue|array-key  $key
      * @param  TValue|null  $value
      * @return bool
      */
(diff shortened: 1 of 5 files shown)


src/Illuminate/Collections/Traits/EnumeratesValues.php
```diff
@@ -195,7 +195,7 @@ public function some($key, $operator = null, $value = null)
     /**
      * Determine if an item exists, using strict comparison.
      *
-     * @param  (callable(TValue): bool)|TValue|string  $key
+     * @param  (callable(TValue): bool)|TValue|array-key  $key
```
nunomaduro, line 198: This change is covered by our tests @canvural?

nunomaduro (changes requested):

src/Illuminate/Collections/Traits/EnumeratesValues.php
canvural: Yes. This two lines are covering this change: https://github.com/laravel/framework/pull/40416/files#diff-61f68115cbe917593bd9c2d4e38338bb6ef9e485739fdff0e989936695c504efR94-R95

PR #38066, opened 2021-07-19
[8.x] Allow for named arguments via dispatchable trait

sleicester:
I've had a situation where it would be nice to use named arguments in dispatching a job to the queue as it allows me to leave out arguments:

```php
RegisterResponse::dispatch(
    uuid: $uuid,
    question: $request->question,
    answer: $request->answer,
    referer: $request->input('referer', $request->headers->get('referer'))
);
```

Unfortunately this will now work as `dispatch`, `dispatchSync` and `dispatchAfterRequest` don't receive any arguments.

Obviously it's easy enough to go about this a different way:

```php
dispatch(new RegisterResponse(
    question: $request->question,
    answer: $request->answer,
));
```

Are there any consequences to these simple changes that I'm not foreseeing?

PR #40421, opened 2022-01-14
[8.x] Fix forwarded call with named arguments

bakerkretzmar:
This PR fixes one instance where passing named arguments is currently broken.

```php
// Doesn't work before this PR, ends up calling the base `listContents` method
// with arguments `true, false` instead of `'', true`

app('filesystem')->listContents(recursive: true);
```

[Laravel's support policy](https://laravel.com/docs/8.x/releases#named-arguments) is very clear that named arguments are not guaranteed to be backwards-compatible:

> PHP's named arguments functionality are not covered by Laravel's backwards compatibility guidelines [...] using named arguments when calling Laravel methods should be done cautiously.

However, the `array_values` call that this PR removes explicitly prevents named arguments from being used at all, even if the developer is using them intentionally and understands the implications of doing so.

I understand that there are a lot of strong opinions about whether named arguments are a good idea, and personally I agree with the decision to exclude them from any backwards-compatibility guarantees, but they are part of the language so it makes sense for it to at least be **possible** to use them.

See also #38066.

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemAdapter.php
+++ b/src/Illuminate/Filesystem/FilesystemAdapter.php
@@ -824,6 +824,6 @@ class FilesystemAdapter implements CloudFilesystemContract
             return $this->macroCall($method, $parameters);
         }
 
-        return $this->driver->{$method}(...array_values($parameters));
+        return $this->driver->{$method}(...$parameters);
     }
 }


taylorotwell:
Is it possible this `array_values` call was present for some other reason?

bakerkretzmar:
It was added as a replacement for `call_user_func_array` to make Laravel 6 compatible with PHP 8, here: https://github.com/laravel/framework/commit/82ffe3cd4a72887ba3e6f8f4bf8fa28fda1f3691#diff-4f23ce415eb581db9639b4aba229a0c5df474b436a7dd46a075a06dcbb4532db

There are several other very similar places in the framework doing this kind of thing with `__call()` and it's inconsistent, some of them use `...array_values($parameters)` and some just use `...$parameters`. Even in that commit, it was done differently in different places.

Honestly I think `array_values` is there just to discourage/prevent use of named arguments (and I totally understand _discouraging_ using them I just don't want it to be prevented entirely). @GrahamCampbell would probably know for sure though.

taylorotwell:
I don't think you can even unpack an array with string keys until PHP 8.1?

bakerkretzmar:
I've tested this on PHP 8 and 8.1 and it works in function arguments (but not in something like `['one', ...['two' => 'three']]`). I'll try on 7.4.

bakerkretzmar:
Technically it doesn't work on PHP 7.4 but you'd never be able to get that far because there aren't named arguments in PHP 7.4, so the `$parameters` array would never have string keys.

bakerkretzmar:
Okay cool, I just tried it with `call_user_func_array` too and named arguments actually work fine there, it can handle an associative array. So even though PHP 8 wasn't out/supported, technically this used to work before it was updated to use argument unpacking with `...` instead.

bakerkretzmar:
Last comment I swear 😅 after looking through every `array_values` call in the framework, I think this is actually the one and only place this is broken, specifically because it's inside a `__call()`. Other magic methods in the framework don't call `array_values` on their parameters like this, and everywhere else `array_values` is used it's on user input, which makes sense because you don't know if it has string keys or not. Inside magic methods though it's safe not to.

PR #40425, opened 2022-01-15
[8.x] Add a BladeCompiler::render() method to render a string with Blade

jasonlbeggs:
After @aarondfrancis' [tweet](https://twitter.com/aarondfrancis/status/1481747670322622464) detailing how to compile a string with Blade got some traction, we decided to submit a PR to core to make this easier for everyone else in the future.

This new method allows you to call `BladeCompiler::render($stringWithBladeInside, $data)` and get the string with all the Blade compiled returned. Alternatively you may call the `render` method from the `Blade` facade: `Blade::render($string, $data)`.

Examples:

```php
Blade::render('Hello, {{ $name }}', ['name' => 'Claire']);

// Returns 'Hello, Claire'
```

```php
Blade::render('@if($foo) Foo @else Bar @endif', ['foo' => true]);

// Returns 'Foo '
```

By default, the method will cleanup the cached view that's created, but there's a third optional argument to disable that part.


stancl:
@jasonlbeggs Does this support using components (anonymous & class-based) in the Blade string? The existing ways of rendering Blade strings had issues with complex use cases like these.

mattstauffer:
This is suuuuuper hot! Nice work!

jasonlbeggs:
@stancl I believe it supports all Blade syntax but I could be wrong.

stancl:
I think so too, though it's worth trying it out to be sure. iirc the issue with using the compiler directly was that it didn't support including compiled component files, but since this approach creates a compiled .php file it's probably fine.

jasonlbeggs:
I just tested it with an anonymous component and it works great!

> resources/views/components/test.blade.php

```blade
Hello, {{ $name }}
```

> Usage

```php
Blade::render('<x-test name="Taylor" />');

// Returns 'Hello, Taylor'
```

Sharifur:
great addition

cwilby:
Side note for anyone interested - We were able to use Twig to generate other XML-based formats.  Curious to see if it would work with this!

Great work.

browner12:
What is the intended use case for this, and where should it be called?

Calling it in a controller works, but calling it in a view file does not. Not sure if that's just a limitation of how it works, or if I'm doing something wrong.

For a long time I've been toying with something like this to try and allow users to write Blade into a field stored in the database.  Theoretically this would allow functionality similar to Wordpress's shortcodes. Currently I'm using a package that does this.  Wondering if this would be a good way to handle this. Thoughts?

PR #40434, opened 2022-01-15
[8.x] Generate docblocks for real-time facades

imanghafoori1:
Realtime facades are introduced way back in laravel 5.4 I think, but they are not used that often.
One of the main reasons that people get peace off is that they immediately lose IDE auto-completion when switching to a real-time facade.

This PR makes docblocks available for public non-static methods on the generated facade class in the cache folder.

- We can have an artisan command to remove the generated facades to refresh the docblocks when the underlying class evolves over time.

image


taylorotwell:

PR #40435, opened 2022-01-15
View named arguments

tenmajkl:
# View named arguments

Ever since i started with PHP I was wondering if there is better way of doing `view('view', ['foo' => 'bar'])`. And since Laravel switches to PHP 8 I wanted to add ability to use named arguments in views. So now its possible not just 
```php
return view('view', ['foo' => 'bar'])
```

but also 

```php
return view('view', foo: 'bar')
```

Since its my first pull request I know its not the best commit here, but I hope it will somehow help.


driesvints:
Seems like something went wrong with branching here

PR #40436, opened 2022-01-15
[9.x] View named arguments

tenmajkl:
# View named arguments

Ever since i started with PHP I was wondering if there is better way of doing `view('view', ['foo' => 'bar'])`. And since Laravel 9 switches to PHP 8 I wanted to add ability to use named arguments in views. So now its possible not just 
```php
return view('view', ['foo' => 'bar'])
```

but also 

```php
return view('view', foo: 'bar')
```

For merge data there is
```php
return view('view', foo: 'bar', mergeData: $someData)
```

Since its my first pull request I know its not the best commit here, but I hope it will somehow help.

### DIFF
--- a/src/Illuminate/Contracts/View/Factory.php
+++ b/src/Illuminate/Contracts/View/Factory.php
@@ -26,11 +26,10 @@ interface Factory
      * Get the evaluated view contents for the given view.
      *
      * @param  string  $view
-     * @param  \Illuminate\Contracts\Support\Arrayable|array  $data
-     * @param  array  $mergeData
+     * @param  mixed  ...$data
      * @return \Illuminate\Contracts\View\View
      */
-    public function make($view, $data = [], $mergeData = []);
+    public function make($view, ...$data);
 
     /**
      * Add a piece of shared data to the environment.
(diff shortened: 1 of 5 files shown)


SjorsO:
> Ever since i started with PHP I was wondering if there is better way of doing `view('view', ['foo' => 'bar'])`. 

What's wrong with `view('view', ['foo' => 'bar'])` though? 😄

igorgaming:
Personally for me `['foo' => 'bar']` more readable

TENMAJKL:
Maybe its just the thing that before laravel I've done flask which has this feature. I thing its faster to write.

dennisprudlo:
Wouldn't there be issues when using a name like "view" or "data" because those names are already used?

TENMAJKL:
> Wouldn't there be issues when using a name like "view" or "data" because those names are already used?

No, because it takes only the arguments you pass after view name, so this shouldn't be problem

taylorotwell:

TENMAJKL:
@taylorotwell its not possible to make it as separate package, because it rewrites existing view function. I mean I can make package but it won't work well

PR #40438, opened 2022-01-16
[8.x] simplify ``whenEmpty`` & ``whenNotEmpty`` methods

isaeken:

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -795,34 +795,24 @@ class Stringable implements JsonSerializable
      * Execute the given callback if the string is empty.
      *
      * @param  callable  $callback
+     * @param  callable|null  $default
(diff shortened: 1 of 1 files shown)


src/Illuminate/Support/Stringable.php
```diff
@@ -795,34 +795,24 @@ public function whenContainsAll(array $needles, $callback, $default = null)
...
+    public function whenNotEmpty($callback, $default = null)
     {
-        if ($this->isNotEmpty()) {
-            $result = $callback($this);
-
-            return is_null($result) ? $this : $result;
-        }
-
-        return $this;
+        return $this->when(! $this->ieEmpty(), $callback, $default);
```
bogdankharchenko, line 815:
```suggestion
        return $this->when(! $this->isEmpty(), $callback, $default);
```

PR #40442, opened 2022-01-16
Fix casting attribute to object

ben221199:
This pull request fixes the casting to an object.

When using `array` as cast type, `[]` would be casted to an array and back, no matter using associative arrays or not.
However, for `object` this was not the case. The `{}` was casted to an object, but the way back didn't work.


taylorotwell:

Feel free to re-submit your change with a **thorough** explanation of the feature and **tests** - integration tests are preferred over unit tests. Please include it's benefit to end users; the reasons it does not break any existing features; how it makes building web applications easier, etc.

Thanks!

PR #40444, opened 2022-01-16
Add validation for object

ben221199:
This pull request adds a validation rule for an object. It is similar to the `array` rule.

```php
$array = []; //The rule 'array' will succeed and the rule `object` will fail.

$object = (object) []; //The rule 'object' will succeed and the rule `array` will fail.
```

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1429,6 +1429,18 @@ trait ValidatesAttributes
     }
 
     /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
What would you be showing the user in this case?

ben221199:
In case of JSON, an array isn't the same as an object. PHP also has an array type and an object type. When using `json_decode` with the second parameter being `false`, the function returns data where JSON objects are converted to PHP objects and JSON arrays are converted to PHP arrays. No associative arrays are used. In that case, it is important to test if an attribute is an object or an array.

Issue #20406, opened 2017-08-03
Cache::flush does not work with redis clusters


Dadibom:
- Laravel Version: 5.4.31
- PHP Version: 7.1
- Database Driver & Version: phpredis (3.1.1-1), cluster using redis 3.2.9

### Description:
(1/1) ErrorException RedisCluster::flushdb() expects exactly 1 parameter, 0 given

Illuminate\Cache\RedisStore::flush does not accept any parameters and will not pass anything in the call to $this->connection()->flushdb

### Steps To Reproduce:
Set cache driver to redis, set it to use a cluster, call \Cache::flush(0 /* or whatever, won't be passed anyway */);

themsaid:
@alibo, do you have any idea how to solve this?

alibo:
@Dadibom @themsaid PhpRedis, unlike Predis, needs a key to execute this command:

> In the case of all commands which need to be directed at a node, the calling convention is identical to the Redis call, except that they require an additional (first) argument in order to deliver the command. Following is a list of each of these commands:
>1. SAVE
>2. BGSAVE
>3. FLUSHDB
>4. FLUSHALL
>5. DBSIZE
>6. BGREWRITEAOF
>7. LASTSAVE
>8. INFO
>9. CLIENT
>10. CLUSTER
>11. CONFIG
>12. PUBSUB
>13. SLOWLOG
>14. RANDOMKEY
>15. PING

https://github.com/phpredis/phpredis/blob/develop/cluster.markdown#directed-node-commands


```php
    public function testDBSize() {
        for ($i = 0; $i < 10; $i++) {
            $str_key = "key:$i";
            $this->assertTrue($this->redis->flushdb($str_key));
            $this->redis->set($str_key, "val:$i");
            $this->assertEquals(1, $this->redis->dbsize($str_key));
        }
    }
```
https://github.com/phpredis/phpredis/blob/master/tests/RedisClusterTest.php#L113


If you connect to one of master nodes and run `flushdb` command, only data stored in that node will be removed and the rest of the data in other master nodes will remain. 

In addition, if you use `Predis`, it throws an exception:

```
>>> Cache::store('redis')->flush();
Predis\NotSupportedException with message 'Cannot use 'FLUSHDB' with redis-cluster.'
```

As it doesn't remove the whole data stored in all master nodes and Predis doesn't support it, I think we can throw an exception for `Cache::flush()` and `PhpRedisClusterConnection::flushdb($key = null)` if the client is Redis Cluster.

[...]

PR #40446, opened 2022-01-17
[8.x] Fix flushdb for predis cluster

bernardwiesner:
This is to fix this issue: https://github.com/laravel/framework/issues/20406

`Redis::flushdb()` or `Cache::flush` does not work currently with predis cluster, this PR fixes that.

@tillkruss asked me to mention him

### DIFF
--- a/src/Illuminate/Redis/Connections/PredisClusterConnection.php
+++ b/src/Illuminate/Redis/Connections/PredisClusterConnection.php
@@ -14,7 +14,7 @@ class PredisClusterConnection extends PredisConnection
     public function flushdb()
     {
         $this->client->executeCommandOnNodes(
-            tap(new ServerFlushDatabase)->setArguments(func_get_args());
+            tap(new ServerFlushDatabase)->setArguments(func_get_args())
         );
     }
 }
--- a/src/Illuminate/Redis/Connections/PredisClusterConnection.php
+++ b/src/Illuminate/Redis/Connections/PredisClusterConnection.php
@@ -13,7 +13,7 @@ class PredisClusterConnection extends PredisConnection
      */
     public function flushdb()
     {
-        foreach ($this->getConnection()->getIterator() as $node) {
+        foreach ($this->client->getIterator() as $node) {
             $node->executeCommand(new ServerFlushDatabase);
         }
     }
--- a/src/Illuminate/Redis/Connections/PredisClusterConnection.php
+++ b/src/Illuminate/Redis/Connections/PredisClusterConnection.php
@@ -13,8 +13,8 @@ class PredisClusterConnection extends PredisConnection
      */
     public function flushdb()
     {
-        foreach ($this->client->getIterator() as $node) {
-            $node->executeCommand(new ServerFlushDatabase);
-        }
+        $this->client->executeCommandOnNodes(
+            tap(new ServerFlushDatabase)->setArguments(func_get_args());
+        );
     }
 }
--- a/src/Illuminate/Redis/Connections/PredisClusterConnection.php
+++ b/src/Illuminate/Redis/Connections/PredisClusterConnection.php
@@ -7,7 +7,7 @@ use Predis\Command\ServerFlushDatabase;
 class PredisClusterConnection extends PredisConnection
 {
     /**
-     * Flush the selected Redis database on all custer nodes.
+     * Flush the selected Redis database on all cluster nodes.
      *
      * @return void
      */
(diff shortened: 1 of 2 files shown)


tillkruss:
Thanks @bernardwiesner!

Let's move that logic into the `PredisClusterConnection` class.

bernardwiesner:
@tillkruss 
thanks I moved it, let me know if its ok now

src/Illuminate/Redis/Connections/PredisConnection.php
```diff
@@ -52,20 +50,4 @@ public function createSubscription($channels, Closure $callback, $method = 'subs
...
-    public function flushdb()
-    {
-        if (! $this->client->getConnection() instanceof ClusterInterface) {
-            return $this->command('flushdb');
-        }
-
-        foreach ($this->getConnection() as $node) {
-            $node->executeCommand(new ServerFlushDatabase);
-        }
-    }
```
driesvints, line 70: Removing this here is a breaking change.

bernardwiesner: why? The flushdb will go to the client, I moved it as recommended by @tillkruss

driesvints: There's no longer a flushdb method on PredisConnection. Any app calling it will break.

bernardwiesner: https://github.com/laravel/framework/blob/45995483913655b7061cdec58421539096afb7b9/src/Illuminate/Redis/Connections/Connection.php#L214

Won't it get called here? I believe I tested it and there was no issues. Maybe I am misunderstanding something

driesvints: Yes but the behaviour won't be applied anymore. It'll just propagate to `command`. Maybe I'm missing something as well but I'm just seeing removed functionality 😅

tillkruss: I meant to only extract part of the method @bernardwiesner. This must stay:

```php
/**
      * Flush the selected Redis database.
      *
      * @return void
      */
     public function flushdb()
     {
         return $this->command('flushdb');
     }
```

taylorotwell:
Please mark as ready for review when all issues have been resolved.

tillkruss (commented):
LGTM 👍

src/Illuminate/Redis/Connections/PredisClusterConnection.php
```diff
@@ -2,7 +2,19 @@
...
 {
-    //
+    /**
+     * Flush the selected Redis database.
+     *
+     * @return void
+     */
+    public function flushdb()
+    {
+        foreach ($this->getConnection()->getIterator() as $node) {
```
tillkruss, line 16: As of Redis v1.1.0, we can iterate `Predis\Client` directly. @bernardwiesner What Predis version are you running?

Let's update this to:

```php
foreach ($this->client as $node) {
    $node->executeCommand(new ServerFlushDatabase);
}
```

src/Illuminate/Redis/Connections/PredisClusterConnection.php
```diff
@@ -2,7 +2,19 @@
...
+            $node->executeCommand(new ServerFlushDatabase);
```
tillkruss, line 17: We are bypassing the `Illuminate\Redis\Connections\Connection::command()` here, which always has been the case, but I wonder if we can fix that in this PR as well.

tillkruss:
@taylorotwell: The fix here will be going from `Client::getConnection()` to `Client::getIterator()` which loops over several `Predis\Client` instances.

src/Illuminate/Redis/Connections/PredisClusterConnection.php
tillkruss: @driesvints: I'm tempted to suggest this to capture the `CommandExecuted` events. Any thoughts?

```php
foreach ($this->client->getIterator() as $node) {
    tap(new PhpRedisConnection($node))->
        ->setEventDispatcher($this->getEventDispatcher())
        ->command('flushdb');
}
```

driesvints:
Please mark this as ready for review when this can be reviewed again.

src/Illuminate/Redis/Connections/PredisClusterConnection.php
tillkruss: I'll open a separate PR for this.

tillkruss:
@dries: I removed `PredisConnection::flushdb()` altogether, which allows new modifiers like:

```php
$redis->flushdb('ASYNC');
```

PR #40451, opened 2022-01-17
[9.x] Use Faker within seeders

otilor:
This exposes the Faker API within seeders.

### Example

```php
<?php

class TestSeeder extends Seeder {
    public function run()
    {
         Test::create([
             'title' => $this->faker->title
         ]);
    }
}

```


mateusjunges:
Can't this PR go to 8.x instead?

taylorotwell:
I suggest putting this in your factory.

PR #40457, opened 2022-01-17
[8.x] Fix 'strstr' function usage based on its signature

alirezasaharkhiz:
Fixed invalid usage of `strstr` function of php in `InteractsWithInput` trait.
Here is the reference for the `strstr` function usage: 
https://www.php.net/manual/en/function.strstr.php


nunomaduro:
@alirezasaharkhiz Can you add a test that previously was failing, but now - with your pull request - is passing?

alirezasaharkhiz:
I executed default tests with no changes on what it was and some of the tests are failing does it have any specific env file or config?

taylorotwell:
Added test.

PR #40461, opened 2022-01-17
[8.x] Allow configuration of the CSRF cookie name

lhsazevedo:
## Intro
Allow users to change the CSRF cookie name using a new `session.csrf_cookie` config. If no name is set, the previously hardcoded `XSRF-TOKEN` name is used.

I didn't include the token name in this proposal, but I could implement if you find it useful.

## Motivation
In our use case we need to change the CSRF cookie name during runtime. Currently we override the entire `addCookieToResponse()` method, because the cookie name is hardcoded:
https://github.com/laravel/framework/blob/45995483913655b7061cdec58421539096afb7b9/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php#L191-L196

For reference, Axios [exposes](https://github.com/axios/axios-docs/blob/21930fea939cdb7f8b6ab616b8eaf18a1a2ffa91/posts/en/req_config.md?plain=1#L108) a setting for the cookie name (and header as well), so it was really simple to change this on the client side.
```js
// `xsrfCookieName` is the name of the cookie to use as a value for xsrf token
xsrfCookieName: 'XSRF-TOKEN', // default
```

If this is accepted, please let me know if it should be added to the app or docs.

### DIFF
--- a/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php
+++ b/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php
@@ -190,7 +190,7 @@ class VerifyCsrfToken
 
         $response->headers->setCookie(
             new Cookie(
-                'XSRF-TOKEN', $request->session()->token(), $this->availableAt(60 * $config['lifetime']),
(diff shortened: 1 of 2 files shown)


taylorotwell:

lhsazevedo:
Sure, I understand that it is not a common use case.
Thanks for your time though

PR #40471, opened 2022-01-18
[9.x] Remove return type will change annotations

X-Coder264:
The return type will change annotation was just a temporary workaround so that those methods do not throw deprecations on PHP 8.1. The proper fix is to add the real return types to those methods which couldn't be done on the Laravel 8 branch as those changes could technically be breaking changes (plus the `mixed` type is only a PHP 8.0+ thing)  so I'm sending this PR for Laravel 9.

### DIFF
--- a/src/Illuminate/Bus/Batch.php
+++ b/src/Illuminate/Bus/Batch.php
@@ -462,8 +462,7 @@ class Batch implements Arrayable, JsonSerializable
      *
      * @return array
      */
-    #[\ReturnTypeWillChange]
-    public function jsonSerialize()
+    public function jsonSerialize(): array
     {
         return $this->toArray();
     }
--- a/src/Illuminate/Cache/Repository.php
+++ b/src/Illuminate/Cache/Repository.php
@@ -615,8 +615,7 @@ class Repository implements ArrayAccess, CacheContract
      * @param  string  $key
      * @return bool
      */
-    #[\ReturnTypeWillChange]
-    public function offsetExists($key)
+    public function offsetExists($key): bool
     {
         return $this->has($key);
     }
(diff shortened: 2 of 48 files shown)


driesvints:
We shouldn't do this anymore now. It's way too late and it'll impose some significant breaking changes on users migrating to Laravel 9.

Jubeki:
@driesvints Since Laravel 9 is a LTS Version I think it should be changed.

foremtehan:
@driesvints they should excepting breaking changes why to be worried ?

driesvints:
Merging this will have a significant impact on the entire ecosystem. For example, we'll have to re-review each first party library to make sure there's no breakages.

X-Coder264:
@driesvints I doubt that end users override all those array access/iterator methods so the impact for end users should be minimal at worst. First party packages etc. have good CI coverage and I doubt many (if any) override those methods too so it shouldn't be too much of a problem to fix those even if that's the case. But I leave the final decision to you. Personally I don't think it makes too much sense to wait an entire year for Laravel 10 for these changes, especially since 9 is an LTS but I can understand if you think these changes are too late for it.

taylorotwell:
Discussed this with @driesvints and decided to merge in case PHP 8.2 requires these.

LukeTowers:
This did trip us up at @WinterCMS but at least it's mentioned in the upgrade guide. Thanks!

driesvints:
@X-Coder264 did we add the wrong return type for `jsonSerialize` here? It should be `mixed`: https://www.php.net/manual/en/jsonserializable.jsonserialize.php

X-Coder264:
@driesvints It's not incorrect. PHP allows [covariant return types](https://wiki.php.net/rfc/covariant-returns-and-contravariant-parameters) and as `mixed` means that anything can be returned it's allowed to narrow it down to any other type.

For example, the collection (EnumeratesValues trait) already had `@return array<TKey, mixed>` so the `array` return typehint was put there instead of `mixed` as it doesn't make sense to allow everything to be returned when the docblock already documents that the return value must be an array. Same goes for example for the Stringable class too where the return type got specified as `string` etc.

Unless there's a concrete valid use case for which the types got narrowed too much I would leave this as is.

driesvints:
Okay, sounds good 👍

PR #40475, opened 2022-01-18
[9.x] Use the model defined connection when present before using the fallback connection when testing

thibodelanghe:
When an application has multiple DB connections and the tests use the database assertions for a certain connection and in that connection, for example, a prefix is used, then you must always include that connection in your assertion otherwise the table name will not be found.

It would be easier that if a connection is defined in a model, the test will use that connection and not the fallback connection. 

Example:

```php
class User extends Model
{
    protected $connection = 'custom-connection';
}
```

Before
```php
$this->assertDatabaseCount(User::class, 1, 'custom-connection');
```

After
```php
$this->assertDatabaseCount(User::class, 1);
```

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php
@@ -26,7 +26,7 @@ trait InteractsWithDatabase
     protected function assertDatabaseHas($table, array $data, $connection = null)
     {
         $this->assertThat(
-            $this->getTable($table), new HasInDatabase($this->getConnection($table, $connection), $data)
+            $this->getTable($table), new HasInDatabase($this->getConnection($connection, $table), $data)
         );
 
         return $this;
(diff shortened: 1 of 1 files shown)


src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php
```diff
@@ -206,7 +206,21 @@ protected function getConnection($connection = null)
...
+
+        return $entity ? $entity->getTable() : $table;
+    }
+
+    protected function getTableConnection($table)
+    {
+        $entity = $this->getModelEntity($table);
+
+        return $entity ? $entity->getConnectionName() : null;
+    }
```
derekmd, line 219: 9.x targets PHP 8.0+ so the nullsafe operator can be used in these two methods to avoid temporary variables:

```
return $this->getModelEntity($table)?->getConnectionName();
```

taylorotwell:
PR does not have doc blocks like the rest of the framework.

PR #40480, opened 2022-01-18
[8.x] Add render method to Blade Facade

4m1n0s:
this was introduced here #40425


stancl:
Shouldn't the second and third args have default values even in the annotation?

driesvints:
@stancl fixed, thanks

PR #40481, opened 2022-01-18
[8.x] Remove `else` block

faissaloux:


GrahamCampbell:
These changes are unnecessary. The code already works fine.

lupinitylabs:
> These changes are unnecessary. The code already works fine.

Also, I don't see a readability issue that would justify a change.

PR #40489, opened 2022-01-19
[8.x] Add support for Http request defaults

JaZo:
This PR adds support for Http request defaults.

## Context
I'd like to add middleware to all Http requests, but don't want to scatter this across my entire codebase. Example middleware I'd like to add is a [debugbar middleware](https://www.csrhymes.com/2020/07/01/track-http-client-in-laravel-debugbar.html) on local environments or a Bugsnag breadcrumb middleware on production environments.

## Solution
I've added a `defaults` method to the request factory. This allows you to define callbacks that can set defaults on pending requests. This way you can easily add middleware, or set other options such as retry, on all requests.

```php
// In a service provider
Http::defaults(function (PendingRequest $request) {
    $request->middleware(new Middleware());
    $request->withHeaders(['X-Test-Header' => 'foo']);
    $request->retry(3, 100, fn ($exception) => $exception instanceof ConnectionException);
    // etc.
});

// Somewhere else
Http::get('http://example.com');
```

Now everytime you perform an Http request, these defaults are set.

## Alternatives
You could use a [macro](https://laravel.com/docs/8.x/http-client#macros) for this, but if you have multiple macros, you still have to add these defaults to all macros as you can't chain macros. Unless you add all other macros to the pending request itself, but I don't think that's an ideal solution.

## Breaking changes
This is only additive, but - as with all macroable classes - if someone has defined a 'defaults' macro, it would be a breaking change for them.

### DIFF
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -72,6 +72,13 @@ class Factory
     protected $stubCallbacks;
 
     /**
+     * The callbacks that should execute when a pending request is created.
+     *
+     * @var \Illuminate\Support\Collection
+     */
+    protected $defaultsCallbacks;
+
+    /**
      * Indicates if the factory is recording requests and responses.
      *
      * @var bool
@@ -103,6 +110,7 @@ class Factory
         $this->dispatcher = $dispatcher;
 
         $this->stubCallbacks = collect();
+        $this->defaultsCallbacks = collect();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


nunomaduro:
While this pull request seems interesting, vendor code may assume that the Laravel's HTTP client is stateless — this can be dangerous, when performing HTTP requests on APIs with sensitive information.

stevebauman:
Going off @nunomaduro's comment -- wouldn't this impact everywhere else in the application that may be using the `Http` client depending on where/when this is called? Setting defaults globally on the `Http` client definitely seems problematic.

> I'd like to add middleware to all Http requests, but don't want to scatter this across my entire codebase.

I'd personally create a class to generate a pending request with all these defaults:

```php
class Api
{
    public static function client()
    {
        return Http:withMiddleware()
            ->withHeaders(['X-Test-Header' => 'foo'])
            ->retry(3, 100, fn ($exception) => $exception instanceof ConnectionException);
    }
}
```

taylorotwell:

JaZo:
Thanks for the feedback @nunomaduro and @stevebauman, I understand your concerns!

PR #40498, opened 2022-01-19
[9.x] Add ability to generate unique validation rules per nested array element (repush)

stevebauman:
## Description

This PR introduces the ability to generate unique validation rules per nested array element, granting you direct access to each value supplied the request, the attribute name, and the whole data set in case other attribute values are needed.

This will be done using a new `Rule` static method: `Rule::nested($callback)`.

## Purpose

Generation of unique validation rules in nested arrays is currently pretty difficult, especially if you use any sort of custom validation rules. There is no built in way to *generate* validation rules taking into account the current iteration's value, and you'll have to set the indexes and such manually inside of a `$rules` array prior to sending it to the validator.

This also becomes quite problematic because we need to trust the data structure that was sent to us in the request and then iterate over it to dynamically generate our rules prior to actual validation taking place.

## Example

**Before:**
```php
$rules = [
    'companies.*.id' => [Rule::exists(Company::class, 'id')],
];

foreach ($request->companies as $index => $company) {
    $rules["companies.{$index}.id"] = [new HasPermission('manage-company', $company['id'])];
}

Validator::validate($request->all(), $rules);
```

**After:**
```php
$rules = [
    'companies.*.id' => Rule::nested(function ($attribute, $value) {
        return [
            Rule::exists(Company::class, 'id'),
            new HasPermission('manage-company', $value)
        ];
    }),
];

Validator::validate($request->all(), $rules);
```

---

Another example where this is extremely handy is the validation of complex nested relationships:

```php
$rules = [
    'companies.*.id' => Rule::nested(function ($attribute, $value) {
        return [
            Rule::exists(Company::class, 'id'),
            new HasPermission('manage-company', $value)
        ];
    })
    'companies.*.locations.*.id' => Rule::nested(function ($attribute, $value, $data) {
        // $attribute = 'companies.0.locations.0.id'
        $company = Str::before($attribute, '.locations');

        return Rule::exists(Locations::class, 'id')->where('company_id', $data["$company.id"]);
    }),
];
````

You may also supply an array of rules as you would normally with `nested` instances, as well as supply multiple `nested` instances:

```php
$rules = [
    'companies.*.id' => [
        Rule::exists(Company::class, 'id'),
        Rule::nested(function ($attribute, $value) {
            return [new HasPermission('manage-company', $value)];
        }),
        Rule::nested(function ($attribute, $value) {
            // ...
        }),
    ],
];
```

Validation of arrays is super tricky stuff. I think this greatly simplifies the developer experience, allowing us not to think too much about its complexities.

---

Please let me know if you have any questions/comments/concerns. Thanks so much for your time! ❤️

---

Apologies for having to [re-push](https://github.com/laravel/framework/pull/40464) this -- my git skills are not the greatest 😅


taylorotwell:
In this example:

```php
    'items.*' => Rule::nested(function () {
        return ['discounts.*.id' => 'distinct'];
    }),
```

What if you wanted to do something like `'discounts.*.id' => ['distinct', 'another-rule-here']`?

stevebauman:
> What if you wanted to do something like `'discounts.*.id' => ['distinct', 'another-rule-here']`?

```php
'items.*' => Rule::nested(function () {
    return ['discounts.*.id' => ['distinct', 'another-rule]];
}),
```

```php
'items.*' => Rule::nested(function () {
    return ['discounts.*.id' => 'distinct|another-rule'];
}),
```

https://github.com/laravel/framework/blob/8ef0613eda9097fdfbc21d1242dd80bddf22c317/tests/Validation/ValidationNestedRulesTest.php#L139

src/Illuminate/Validation/ValidationRuleParser.php
```diff
@@ -130,10 +144,22 @@ protected function explodeWildcardRules($results, $attribute, $rules)
...
-                    $results = $this->mergeRules($results, $key, $rule);
+                foreach (Arr::flatten((array) $rules) as $rule) {
+                    if ($rule instanceof NestedRules) {
+                        $compiled = $rule->compile($key, $value, $data);
+
+                        $this->implicitAttributes = array_merge_recursive(
+                            $compiled->implicitAttributes,
+                            $this->implicitAttributes,
+                            [$attribute => [$key]]
+                        );
```
taylorotwell, line 155: Can you explain this statement and why it is needed?

src/Illuminate/Validation/ValidationRuleParser.php
```diff
@@ -79,26 +79,34 @@ protected function explodeRules($rules)
...
 
-        return array_map([$this, 'prepareRule'], $rule);
+        $attributes = array_fill(
+            array_key_first($rule), count($rule), $attribute
+        );
+
+        return array_map(
+            [$this, 'prepareRule'], $rule, $attributes
+        );
     }
```
taylorotwell, line 100: Can you explain the `array_fill` and `array_map` above and explain why they are needed?

stevebauman: Yes absolutely.

I use `array_fill` here to generate an array filled with the exact same attribute, which will be equal to the number of rules that are being iterated over. This allows `array_map` to pass the attribute into the second parameter of the `parepareRule` method.

This is mainly for convenience because `array_map` doesn't support passing secondary arguments into the map method.

src/Illuminate/Validation/ValidationRuleParser.php
stevebauman: Yes absolutely:

I'm using `array_merge_recursive` here so the `implicitAttributes` arrays (along with their sub arrays) are merged together properly per iteration. Any additional implicit attribute *absolute* dot paths will be appended properly to each *asterisk* dot path.

I also make sure to merge the current implicit attribute dot path using: `[$attribute => [$key]]` which is done in the `else` statement below. This "segments" the implicit attributes into their own sections -- and from what I understand, the validator will perform validation on these independently, allowing us to use `distinct` for example, to validate across the current array and not the entire array:

Example:

```php
$rules = [
    'items.*' => Rule::nested(function () {
        return [
            'discounts.*.id' => Rule::nested(function () {
                return 'distinct';
            }),
        ];
    }),
];

$result = (new ValidationRuleParser(['...']))->explode($rules);

{#65 ▼
//   +"rules": array:5 [▼
//     "items.0.discounts.0.id" => array:1 [▼
//       0 => "distinct"
//     ]
//     "items.0.discounts.1.id" => array:1 [▼
//       0 => "distinct"
//     ]
//     "items.0.discounts.2.id" => array:1 [▼
//       0 => "distinct"
//     ]
//     "items.1.discounts.0.id" => array:1 [▼
//       0 => "distinct"
//     ]
//     "items.1.discounts.1.id" => array:1 [▼
//       0 => "distinct"
//     ]
//   ]
//   +"implicitAttributes": array:3 [▼
//     "items.1.discounts.*.id" => array:2 [▼
//       0 => "items.1.discounts.0.id"
//       1 => "items.1.discounts.1.id"
//     ]
//     "items.0.discounts.*.id" => array:3 [▼
//       0 => "items.0.discounts.0.id"
//       1 => "items.0.discounts.1.id"
//       2 => "items.0.discounts.2.id"
//     ]
//     "items.*" => array:2 [▼
//       0 => "items.0"
//       1 => "items.1"
//     ]
//   ]
// }
```

taylorotwell:
Just thinking out loud - what do you think about `Rule::forEach` instead of `Rule::nested`?

stevebauman:
I like `Rule::forEach()`, that definitely sounds better than `nested()`. Much clearer to what is taking place in the validator 👍

stevebauman:
Also I forgot to mention @taylorotwell, should we be converting the `$value` argument to a `Fluent` instance, similarly to the `$validator->sometimes()` method?

**Sometimes**:
```php
$validator->sometimes('...', ['...'], function (Fluent $input) {
     // ...
});
```

**For Each**:
```php
Rule::forEach(function ($attribute, Fluent $value, $data = null) {
    // ...
});
```

taylorotwell:
@stevebauman doesn't Fluent kinda seem weird in situations like this where the $value is a single element?

```php
    $validator = Validator::make([
        'users' => [
            ['id' => 1, 'email' => 'taylor@laravel.com'],
            ['id' => 2, 'email' => 'foo bar'],
        ],
    ], [
        'users.*.email' => Rule::forEach(function ($attribute, $value, $data) {
            ray($value);

            return ['email'];
        }),
    ]);
```

I think passing `Fluent` would maybe make more sense if the last part of the element being nested was a `*`, like here:

```php
    $validator = Validator::make([
        'users' => [
            ['id' => 1, 'email' => 'taylor@laravel.com'],
            ['id' => 2, 'email' => 'foo bar'],
        ],
    ], [
        'users.*' => Rule::nested(function ($attribute, $value, $data) {
            ray($value);

            return ['email' => 'email'];
        }),
    ]);
```

stevebauman:
Ah yea you're right. Guess we can't really convert it to anything since the data type will change

taylorotwell:
Do you have any way to detect if the attribute Rule::foreEach is being applied to ends with a *?

stevebauman:
We could detect it via the `$attribute` parameter in `explodeWildcardRules()`:

```php
$rules = [
    'items.*' => Rule::nested(function () {
        return ['discounts.*.id' => 'distinct'];
    }),
];

$results = (new ValidationRuleParser)->explode($rules);

// ...

// explodeWildcardRules()...

if ($rule instanceof NestedRules) {
    $arrayable = Str::afterLast($attribute, '.') === '*';

    $compiled = $rule->compile($key, $arrayable ? new Fluent($value) : $value, $data);

    // ...
} else {
```

taylorotwell:
What do you think about that? Good idea?

stevebauman:
Ah, not too sure after some thought. This wouldn't make sense to use a `Fluent` instance if `items` is a list of numbers/ID's:

```php
$data = ['items' => [1,2,3]];

$rules = [
    'items.*' => Rule::nested(function ($attribute, Fluent $value) {
        return 'distinct';
    }),
];
```

Though we could allow developers to provide a secondary "transformation" function to the `Rule::nested()` method:

```php
$data = ['items' => [1,2,3]];
Rule::nested(function ($attribute, Collection $value) {

}, function ($value) {
    return collect($value);
});
```

stevebauman:
I think we should swap the argument order though since the `$value` is probably going to be needed more than the `$attribute`. What do you think?

```php
Rule::forEach(function ($value, $attribute) {
    // ...
});

// vs

Rule::forEach(function ($attribute, $value) {
    // ...
});
```

taylorotwell:
Yeah, I'm fine with swapping the argument order.

siarheipashkevich:
@stevebauman why you did not rename this rule from `nested` to `forEach`?

stevebauman:
@siarheipashkevich https://github.com/laravel/framework/pull/40586

Issue #40468, opened 2022-01-18
Str::headline does not work with umlauts


cemrich:
- Laravel Version: 8.78.1
- PHP Version: 8.1.1
- Database Driver & Version:

### Description:
`Illuminate\Support\Str::headline` does not support umlauts (like ö,ä,ü). There are two related bugs I can see:
1. Uppercase umlauts are not detected as uppercase.
2. Umlauts at the beginning of a word do not get capitalized.

#### Umlauts at the beginning of a word do not get capitalized

    public function test_make_umlauts_uppercase() {
        $input = 'ärmere-öffentliche-überraschungen-sind-öde';
        $expected = 'Ärmere Öffentliche Überraschungen Sind Öde';
        $output = Str::headline($input);
        $this->assertEquals($expected, $output);
    }

Yields the output: 

    Failed asserting that two strings are equal.
    Expected :'Ärmere Öffentliche Überraschungen Sind Öde'
    Actual   :'ärmereöffentlicheüberraschungen Sindöde'

#### Uppercase umlauts are not detected as uppercase

    public function test_detect_uppercase_umlauts() {
        $input = 'sindÖdeUndSo';
        $expected = 'Sind Öde Und So';
        $output = Str::headline($input);
        $this->assertEquals($expected, $output);
    }

Yields the output: 

    Failed asserting that two strings are equal.
    Expected :'Sind Öde Und So'
    Actual   :'SindÖde Und So'

foremtehan:
@stevebauman

driesvints:
We'd welcome PR's for this, thank you.

stevebauman:
This is because `ucwords()` doesn't handle unicode characters:

```php
// ärmere öffentliche
echo ucwords('ärmere öffentliche');
```

This means `Str::studly()` or anything relying on that method also does not work with unicode characters.

We'll have to be crafty and use `mb_convert_case()`. I'll see if I can come up with something 👍

PR #40499, opened 2022-01-19
[8.x] Fix Str::headline/Str::studly with unicode and add Str::ucsplit method

stevebauman:
## Description

This PR resolves issues with unicode characters mentioned in issue https://github.com/laravel/framework/issues/40468, fixing `Str::headline()` and `Str::studly()`.

I've also added another `Str` helper function `ucsplit()` to split a string based on casing, which is compatible with english and unicode characters. This new function is used by `Str::headline()`, to help with parsing strings that do not contain spaces.

All previous tests pass and there are no BC breaks 🎉 

Let me know if you have any concerns/questions, thanks for your time! ❤️

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -730,17 +730,15 @@ class Str
      */
     public static function headline($value)
     {
-        $parts = explode('_', static::replace(' ', '_', $value));
+        $parts = explode(' ', $value);
 
-        if (count($parts) > 1) {
-            $parts = array_map([static::class, 'title'], $parts);
-        }
+        $parts = count($parts) > 1
+            ? $parts = array_map([static::class, 'title'], $parts)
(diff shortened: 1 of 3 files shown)


nunomaduro:
@taylorotwell I feel that this should target `9.x`.

foremtehan:
@stevebauman i Found a issue with headline method:

```php
Str::headline('Year 2022')
```

the result is `Year2022`, Is this ok ?

stevebauman:
@foremtehan I'm not able to reproduce this with this new patch:

```
```

foremtehan:
Look like it will be fix via this PR, I'm on v8.77.1 and this method has that issue.

stevebauman:
You're right I'm able to reproduce that as well on v8.x. I've just added some extra assertions so we can ensure this bug is resolved 👍

PR #40504, opened 2022-01-19
[9.x] Fixes `mergeRecursive` and `countBy` return types

canvural:
This PR fixes `mergeRecursive` and `countBy` return types.

`mergeRecursive` does not return an `array` as the collection items. An example:
image

`countBy` was using `TValue` as the key type, but that was wrong because anything other than `int` or `string` would be casted to `int` or `string` in runtime. So `array-key` suits better. I added a test for that kind of case `$collection->make([new User])->countBy('email'));`


nunomaduro (approved):
Perfect, thank you!

PR #40505, opened 2022-01-20
[9.x] Add multitap helper for continuous chaining

dakira:
This PR adds a multitap helper that allows chaining as many methods as needed on an object that doesn't support chaining.

Credit for the idea and implementation goes to @DarkGhostHunter and [his blog post](https://itnext.io/laravel-multitap-multitap-multitap-2496f2d58174) on the topic.

Imagine using the Goutte scraper client:

```php
$client = new Goutte\Client();
$client->request('GET', 'https://github.com');
$client->clickLink('Sign in');
$crawler = $client->submitForm(
    'Sign in',
    ['login' => '...', 'password' => '...'],
);
```

Continuous chaining is not possible because it will always return a `DomCrawler`. With the multitap helper you can do this:

```php
$crawler = multitap(new Goutte\Client())
    ->request('GET', 'https://github.com')
    ->clickLink('Sign in')
    ->finally()
    ->submitForm('Sign in', ['login' => '...', 'password' => '...'])
```

The `finally()` method breaks the chain and in this case returns the real `Goutte\Client` object, so that the following call of `submitForm()` will return the `DomCrawler`.

This is still missing tests and corresponding documentation. Just wanted to put the draft here for discussion and ideas. Also feel free to just close this if you think it's a bad idea, I won't be mad. :-)


driesvints:
@dakira please note that PR's in draft aren't reviewed.

nunomaduro:
I actually have the need for this pull request - as, in the past, have done stuff like `tap(tap(...` to perform continuous chaining. Yet, the `multitap` name does not seem `cool` to me.

dakira:
@driesvints as it should be... the PR was still missing tests, which are now added. ;-)

@nunomaduro Maybe tapOn? deepTap? tapMore?

dammy001:
deepTap maybe 🤔

nunomaduro:
We can also mutate the current `HigherOrderTapProxy` instance. I think is best to hear @taylorotwell thoughts in all proposals so far.

Screenshot 2022-01-20 at 16 47 43

stevebauman:
`deepTap()` is a weird name imo. It's not immediately clear what's going on.

Maybe simply `chain()` to "begin a chain" of methods?

```php
chain($object)
    ->methodOne()
    ->methodTwo()
    ->methodThree();
```

Or maybe even `on()`, which is simpler and reads a bit like english:

```php
on($object)
    ->methodOne()
    ->methodTwo()
    ->methodThree();
```

DarkGhostHunter:
What if a class already has a deeply or finally  method? I think that a LogicException should be thrown because the dev may not expect the object as a result. 

Same case as using the Tapable trait.

Italo Baeza C.

dakira:
My current favourite is a mixture of all. 🤷‍♂️ Repurpose the current `HigherOrderTapProxy` as Nuno suggested, use a `chain()` method instead of `multitap()` and make sure an Exception is thrown when the object has its own `finally()` method.

stevebauman:
We could also use `thenReturn()` which is from the Laravel pipeline class, instead of `finally()` to keep a sort of "convention" for this style of chaining.

```php
chain($object)
    ->methodOne()
    ->methodTwo()
    ->thenReturn();
```

taylorotwell:
@nunomaduro how does your example terminate the chain?

nedwors:
Just throwing in that `access()` might also convey what's happening?

`access($foo)->...->...`

derekmd:
`tap($var)->deeply()` for the best innuendo vote.

PHP allows `break()` as a method name so a chain being broken metaphor may read better than `finally()` that doesn't relate much to exception `finally {}` blocks:

```php
tap($var)
    ->chain() // or chained() to reduce name collision probability?
    ->foo()
    ->bar()
    ->break()
    ->baz();
```

They still have the `chain()` / `break()` name collision issue and I see [a few examples](https://github.com/search?q=%22function+chain%22+language%3APHP&type=Code&l=PHP) for the former using Github's code search although the likelihood of that in a web app is minimal.

nunomaduro:
Added an alternative implementation here: https://github.com/laravel/framework/pull/40538. Let's see what Taylor thinks.

taylorotwell:
Going to close this one so I only have PR on this topic open at a time.

PR #40507, opened 2022-01-20
[9.x] Skip image create tests if gd library is not supported

imdhemy:
Skipping test cases by using `@requires` is okay, but it is not enough in the `gd extension` case. 

The following error appears on running the test suite. 

```
1) Illuminate\Tests\Http\HttpTestingFileFactoryTest::testImageWebp
TypeError: call_user_func(): Argument #1 ($callback) must be a valid callback, function "imagewebp" not found or invalid function name
```

This is because `WebP Support` is not supported. Below is the `gd_info()` result.

```
array:13 [
  "GD Version" => "bundled (2.1.0 compatible)"
  "FreeType Support" => false
  "GIF Read Support" => true
  "GIF Create Support" => true
  "JPEG Support" => true
  "PNG Support" => true
  "WBMP Support" => true
  "XPM Support" => false
  "XBM Support" => true
  "WebP Support" => false // <-- Is not supported
  "BMP Support" => true
  "TGA Read Support" => true
  "JIS-mapped Japanese Font Support" => false
]
```

Note: I'm using the `bin/test.sh` to run tests.

**Solution**
Skip test case if the required library is not supported.

### DIFF
--- a/tests/Http/HttpTestingFileFactoryTest.php
+++ b/tests/Http/HttpTestingFileFactoryTest.php
@@ -7,6 +7,7 @@ use PHPUnit\Framework\TestCase;
 
 /**
  * @requires extension gd
+ *
  * @link https://www.php.net/manual/en/function.gd-info.php
  */
 class HttpTestingFileFactoryTest extends TestCase
@@ -20,7 +21,7 @@ class HttpTestingFileFactoryTest extends TestCase
 
     public function testImagePng()
     {
-        if (!$this->isGDSupported(self::GD_PNG_SUPPORT)) {
+        if (! $this->isGDSupported(self::GD_PNG_SUPPORT)) {
             $this->markTestSkipped('Requires PNG support.');
         }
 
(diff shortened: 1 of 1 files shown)


tests/Http/HttpTestingFileFactoryTest.php
```diff
@@ -7,11 +7,24 @@
...
+ * @link https://www.php.net/manual/en/function.gd-info.php
  */
 class HttpTestingFileFactoryTest extends TestCase
 {
+    private const GD_VERSION = 'GD Version';
+    private const GD_PNG_SUPPORT = 'PNG Support';
+    private const GD_JPEG_SUPPORT = 'JPEG Support';
+    private const GD_GIF_CREATE_SUPPORT = 'GIF Create Support';
+    private const GD_WEBP_SUPPORT = 'WebP Support';
+    private const GD_WBMP_SUPPORT = 'WBMP Support';
```
driesvints, line 20: Since these are only used once, please just use them directly at their used location.

imdhemy: @driesvints resolved.

PR #40494, opened 2022-01-19
[9.x] Adds internal static analysis 

nunomaduro:
This pull request adds an internal static analysis job to the CI.

No worries, it's simply PHPStan at **level `0`**. This level, **is not intrusive, and it does not get in the way of developing code on the framework**. Yet, it's very powerful as it detects:

- Bad merges: https://github.com/laravel/framework/commit/ac6fc32c2933d5868a2c0d2aea4af9405d710378.
- Invalid PHP code, like missing commas, syntax errors, and this kind of things.

Screenshot 2022-01-19 at 16 00 46

PR #40509, opened 2022-01-20
[9.x] Rename workflows

driesvints:
This renames the workflows introduced in #40494 to match the directory names they're covering. I had some difficulty to see what covered what at first glance and this unifies the naming so it's very easy to see which config is for which directory.


.github/workflows/static-analysis-src.yml
```diff
@@ -1,4 +1,4 @@
-name: static analysis api
+name: static analysis src
```
driesvints, line 1: Btw, ignore the fact that this seems like I renamed `api` to `src` and visa versa. It's because git can't resolve this properly in the diff. The files are still correct.

Issue #40478, opened 2022-01-18
[8.x] Custom channels for model broadcasting are ignored


wokes:
- Laravel Version: 8.80.0
- PHP Version: 7.4.24
- Database Driver & Version: not pertinent

### Description:
Model broadcasting seemed like a very convenient way for real-time updates on the frontend, so I gave it a try.
Ran into an issue when I wanted to change the channel the events are broadcasted on.
As per docs, I added `BroadcastsEvents` trait on the model along with
```
    public function broadcastOn($event)
    {
        return [new PrivateChannel('foo-bar-baz')];
    }
```

This did not yield expected results, as the events kept broadcasting only on the default model channel.

I dug around for a bit and the culprit seems to be in https://github.com/laravel/framework/blob/8940ff3c9c643f8abcca026a6341fdfcfc83e983/src/Illuminate/Database/Eloquent/BroadcastsEvents.php#L112

We check `if (! empty($this->broadcastOn($event)) || ! empty($channels)) {` to ensure that there are channels to broadcast on, however:
`return broadcast($instance->onChannels(Arr::wrap($channels)));`
we completely ignore the channels we defined in `broadcastOn`. It's only broadcasted on `$channels`.

Changing https://github.com/laravel/framework/blob/ac6fc32c2933d5868a2c0d2aea4af9405d710378/src/Illuminate/Database/Eloquent/BroadcastsEvents.php#L119
to `return broadcast($instance->onChannels(array_merge(Arr::wrap($channels), $this->broadcastOn($event))));` rectifies the issue.

### Steps To Reproduce:

1. Set up model broadcasting
2. Add `broadcastOn` to the model definition as shown in the description above
3. Create/update/delete the model
4. Model will be sent on the default auto-generated channel instead of the provided one

driesvints:
See the closed PR by Taylor.

PR #40511, opened 2022-01-20
[8.x] Fix custom channels for model broadcasting being ignored

wokes:
Fix for issue #40478

Model broadcasting seemed like a very convenient way for real-time updates on the frontend, so I gave it a try.
Ran into an issue when I wanted to change the channel the events are broadcasted on.
As per docs, I added `BroadcastsEvents` trait on the model along with
```
    public function broadcastOn($event)
    {
        return [new PrivateChannel('foo-bar-baz')];
    }
```

This did not yield expected results, as the events kept broadcasting only on the default model channel.

I dug around for a bit and the culprit seems to be in https://github.com/laravel/framework/blob/8940ff3c9c643f8abcca026a6341fdfcfc83e983/src/Illuminate/Database/Eloquent/BroadcastsEvents.php#L112

We check `if (! empty($this->broadcastOn($event)) || ! empty($channels)) {` to ensure that there are channels to broadcast on, however:
`return broadcast($instance->onChannels(Arr::wrap($channels)));`
we completely ignore the channels we defined in `broadcastOn`. It's only broadcasted on `$channels`.

Changing https://github.com/laravel/framework/blob/ac6fc32c2933d5868a2c0d2aea4af9405d710378/src/Illuminate/Database/Eloquent/BroadcastsEvents.php#L119
to `return broadcast($instance->onChannels(array_merge(Arr::wrap($channels), $this->broadcastOn($event))));` rectifies the issue.

### DIFF
--- a/src/Illuminate/Database/Eloquent/BroadcastsEvents.php
+++ b/src/Illuminate/Database/Eloquent/BroadcastsEvents.php
@@ -116,7 +116,7 @@ trait BroadcastsEvents
         }
 
         if (! empty($this->broadcastOn($event)) || ! empty($channels)) {
-            return broadcast($instance->onChannels(Arr::wrap($channels)));
+            return broadcast($instance->onChannels(array_merge(Arr::wrap($channels), $this->broadcastOn($event))));
         }
     }
 


taylorotwell:
This shouldn't be necessary. Typically, channels would be `null` in that line you sent. Then, later in `BroadcastableModelEventOccurred` we check if it is empty and call `broadcastOn` like you suggest. 

Your change would break passing specific channels as an override directly to a model method like `broadcastCreated`.

jasonburn:
Hi. I encountered the same issue using the `BroadcastsEvents` for my models.
I used something like this on the `broadcastOn` method:
```
public function broadcastOn($event)
{
    return [new PrivateChannel('organization.gova.members')];
}
```
I was using [clockwork](https://underground.works/clockwork) to monitor the broadcast and I can confirm that the custom channels are ignored. I tried @wokes solution and it worked.
Is there another workaround for this one aside from modifying the vendor files?
I really like this feature implemented on our new build.

PR #40513, opened 2022-01-20
[8.x] Allow drop / rename multiple Columns in one Statement using SQLite

hmennen90:
Currently it is not possible using SQLite, to rename and create / drop and create multiple columns in one Schema::table() statement.

This PR splits the Migration if SQLite is in use to first execute the commands (rename and drop), then creating new columns.

So there is no need to split the Schema::table() Statement in one Migration File for each dropColumn / renameColumn and adding new Columns.

Before (need in Migration File):

```
Schema::table('users', function(Blueprint $table) {
$table->renameColumn('name', 'first_name');
});

Schema::table('users', function(Blueprint $table) {
$table->string('last_name')->after('first_name');
});
```

After (splitter during Migration):

```
Schema::table('users', function(Blueprint $table) {
$table->renameColumn('name', 'first_name');
$table->string('last_name')->after('first_name');
});
```

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -1667,6 +1667,7 @@ class Blueprint
                 unset($this->commands[$index]);
             }
         }
+
         return $this;
     }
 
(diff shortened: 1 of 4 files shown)


taylorotwell:

hmennen90:
@taylorotwell as documented it is not supported by SQLite Currently. This PR makes it possible to change columns (rename and drop) in One Schema::table statement instead spreading it in multiple Schema::table Statements.

It is not a Bug - but improves the functionality of SQLite together with Laravel.

PR #40515, opened 2022-01-20
[8.x] Memoize date format

stancl:
Right now, timestamp attributes cause `$model->toArray()` to fail if the model's connection doesn't exist. I think serialization of attributes should always work, regardless of the environment. This small change appears to fix the issue completely.

The reason why `toArray()` fails is that it calls the `getDateFormat()` which always fetches the model's connection to grab the format from the grammar.

However, there are valid use cases for using models without their connections existing. One such example is multi-tenancy, where you may briefly introduce a DB connection for a few queries and clean it up afterward. It seems like simply avoiding calling `toArray()` or making sure to call it while the connection still exists prevents this issue in most cases. But there are a few scenarios when `toArray()` is called in the background.

For example, Telescope will call `toArray()` on any model passed to a view (see https://github.com/archtechx/tenancy/issues/774#issuecomment-1010038755).

## Used date formats

The change seems small and harmless, but to be sure, I went through the relevant code to see how this is used.

I originally assumed the `getQueryGrammar()->getDateFormat()` call uses PDO to get the user-defined date format from the connection, but apparently, this is purely in the PHP land.

- All connections use `Illuminate\Database\Grammar::getDateFormat()` which returns `'Y-m-d H:i:s'`
- SQL server uses `Illuminate\Database\Query\Grammars\SqlServerGrammar::getDateFormat()` which returns `'Y-m-d H:i:s.v'`
- The docs instruct developers to set the `$dateFormat` property on the model to customize the timestamp formats

This makes me think that the value is never dynamic, and therefore memoizing it is safe. And there shouldn't be any breaking changes either, so 8.x is the appropriate target.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1386,7 +1386,7 @@ trait HasAttributes
      */
     public function getDateFormat()
     {
-        return $this->dateFormat ?: $this->getConnection()->getQueryGrammar()->getDateFormat();
+        return $this->dateFormat = $this->dateFormat ?: $this->getConnection()->getQueryGrammar()->getDateFormat();
     }
 
     /**


stancl:
Hmm, seems like this breaks a test:

```
1) Illuminate\Tests\Database\DatabaseEloquentSoftDeletesIntegrationTest::testUpdateModelAfterSoftDeleting
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
     'classCastCache' => Array ()
     'attributeCastCache' => Array ()
     'dates' => Array ()
-    'dateFormat' => 'Y-m-d H:i:s'
+    'dateFormat' => null
     'appends' => Array ()
     'dispatchesEvents' => Array ()
     'observables' => Array ()
/home/runner/work/framework/framework/tests/Database/DatabaseEloquentSoftDeletesIntegrationTest.php:319
```

The code being this:
```php
$userModel = SoftDeletesTestUser::find(2);
$userModel->delete();
$this->assertEquals($now->toDateTimeString(), $userModel->getOriginal('deleted_at'));
$this->assertNull(SoftDeletesTestUser::find(2));
$this->assertEquals($userModel, SoftDeletesTestUser::withTrashed()->find(2));
```

The `getOriginal('deleted_at')` call probably sets the value for `$dateFormat`, which then breaks `assertEquals()` since `$dateFormat` is included in the compared properties.

Excluding the property from the compared properties would be the best option (since it's never instance-specific), but I'm not sure if there's a straightforward BC way to do this.

taylorotwell:
Hmm, I'm not sure the best path forward here. Mark as active if you come up with something.

src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
```diff
@@ -1386,7 +1386,7 @@ public function getDates()
      */
     public function getDateFormat()
     {
-        return $this->dateFormat ?: $this->getConnection()->getQueryGrammar()->getDateFormat();
+        return $this->dateFormat = $this->dateFormat ?: $this->getConnection()->getQueryGrammar()->getDateFormat();
```
johanrosenson, line 1389: This should fix your failing test.

```php
public function getDateFormat()
{
    static $dateFormat = null;
    
    return $this->dateFormat ?: ($dateFormat ??= $this->getConnection()->getQueryGrammar()->getDateFormat());
}
```

If you memoize the `date format` inside the method instead you don't have to touch the class property, changing that property during runtime seems like an unexpected side effect anyway so best to aviod it.

GrahamCampbell: that will not work. it will cache it for ALL instances, the same

GrahamCampbell (changes requested):

src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
johanrosenson: you are very much correct, good catch, i did not take into account that this method is on the base model, I was too focused of a single model.

stancl:
Yeah. I think the best solution would be making it *somehow* disappear from the properties used for comparison, but I'm not sure if PHP offers anything for that.

I wonder if `private` properties are used for comparison. Here the `dateFormat` is `protected` and it is used when comparing instances, so I guess `private` properties are used as well, since they're both functionally the same — non-public. And if `private` properties would be used as well, we can't even move the `dateFormat` cache to another property.

An alternative approach would be to access the grammar *without* the connect somehow. But I still don't see how we could implement that without changing the properties on the model.

src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
johanrosenson: @GrahamCampbell can you clarify what you mean? 

The solution by @stancl would allow ModelA to have it's own dateFormat defined and ModelB can still use default with caching to prevent multiple lookups.

see example, or maybe i misunderstand your comment.

```php
class BaseModel 
{
    protected $dateFormat = null;
    
    public function getDateFormat() 
    {
        static $dateFormat = null;
        
        return $this->dateFormat ?: $dateFormat ??= 'Y-m-d H:i:s';
    }
}

class ModelA extends BaseModel 
{
    protected $dateFormat = 'Ymd His';
}

class ModelB extends BaseModel {}

$instanceA = new ModelA;
$instanceB = new ModelB;

dd(
    $instanceA->getDateFormat(), // "Ymd His"
    $instanceB->getDateFormat(), // "Y-m-d H:i:s"
    $instanceA->getDateFormat(), // "Ymd His"
    $instanceB->getDateFormat(), // "Y-m-d H:i:s"
);
```

stancl:
Also, we could load the dataFormat straight away the moment we load the connection/grammar 

That way in the passing test the "expected" value would be the same as the model instance after toArray() is called

tpetry:
Couldn't you just write a trait to import in the model replacing the `getDateFormat()` method with a static value if called like this:

```php
Model::withDateFormat('Y-m-d', function() {
  // ....
});
```

But to be honest i still don't understand why db connections should be built and removed in a multi-tenancy context. Except some very strange one-db-per-tenant implementation which is not really one-db-per-tenant because it needs to switch between multiple dbs in a single request - hence one-db-for-all-tenants is used.
Or did i miss some critical part?

stancl:
@tpetry We create a `tenant` connection when tenancy is "initialized". This happens either on tenant requests (e.g. requests to the tenant's subdomain), or when `$tenant->run(function () { ... })` is used in the central context.

The connection doesn't always exist, since central requests don't need it and different tenants can have different DB configurations.

We let developers provide a *template connection* which is used to build the base of the `tenant` connection, but the `tenant` connection itself is always created by the package and removed afterwards if the app reverts to a central context.

stancl:
I'd like to finish this PR so that there isn't an incomplete draft open, but I also can't think of any good solution 🤔

It's bizarre because the change is extremely simple. The object comparison is causing problems due to how PHP works with properties.

From what I wrote above, this seems like the best option:
> Also, we could load the dateFormat straight away the moment we load the connection/grammar

It'd probably look like this:
```diff
public function getDateFormat()
{
-    $this->dateFormat ?: $this->getConnection()->getQueryGrammar()->getDateFormat();
+    return $this->dateFormat;
}
```

with `$this->dateFormat = $this->getConnection()->getQueryGrammar()->getDateFormat()` at some point in the lifecycle.

This means that **all** models would have a `dateFormat` value upon interacting with a database.

There might be some edge cases with that approach, but it should probably be fine. I'll add it to my tasks to explore the code and update the PR with the new implementation this week.

driesvints:
@stancl probably best sent to 9.x regardless since we're on the verge of releasing now.

stancl:
@driesvints I can either try changing the target, or you can close this and I'll resubmit. I'm fine with the latter if you don't want the draft hanging in open PRs

driesvints:
@stancl might be best yeah, thanks!

Issue #40497, opened 2022-01-19
Inconsistent caching behaviour with virtual attributes using new Attribute cast


decnorton:
- Laravel Version: 8.79.0
- PHP Version: 7.4.27
- Database Driver & Version: Postgres 13

### Description:
Unsure whether this would be considered a bug, but there's definitely some inconsistency in behaviour.

When returning an object from a virtual attribute, the results are always cached. However, primitives are not. The previous (`get{Name}Attribute()`) did perform _any_ caching. 

We discovered this as we were using this as a shortcut for fetching a model from a relation, which fails if the result of `getCheckIn()` changes:
```php
public function getCheckInAttribute(): ?CheckIn
{
    return $this->booking->getCheckIn($this->carer);
}

public function checkIn(): Attribute
{
    return Attribute::get(fn () => $this->booking->getCheckIn($this->carer));
}
```

We also use it to construct new DateTime instances, which also is inhibited by caching if `start` or `duration` are updated, which is likely a more common use:
```php
public function getFinishAttribute(): CarbonImmutable
{
    return $this->start->addSeconds($this->duration);
}

public function finish(): Attribute
{
    return Attribute::get(fn () => $this->start->addSeconds($this->duration)));
}
```

The relevant lines in `HasAttributes` are here: https://github.com/laravel/framework/blob/29a6692fb0f0d14e5109ae5f02ed70065f10e966/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L632-L636

I understand that caching was put in place in order to facilitate casting value objects, however if this is to be new recommended approach for attribute accessors this could be an edge case that catches a lot of people out. 

I'm not really sure what a potential solution would be. Perhaps a flag on `Attribute` to disable caching explicitly? Happy to have a crack at a PR if a solution can be agreed.

### Steps To Reproduce:

I've added a couple of failing tests in a fork here:
https://github.com/laravel/framework/compare/8.x...decnorton:virtual-attribute-cast-objects

taylorotwell:
Caching is definitely the intended behavior (similar to custom cast classes), but if you want to take a stab at a way to optionally disable it for a given attribute feel free! ❤️

PR #40519, opened 2022-01-20
[8.x] Allow caching to be disabled for virtual attributes accessors that return an object

decnorton:
Context: https://github.com/laravel/framework/issues/40497

This PR allows caching to be disabled for virtual attributes being accessed used the new `Attribute` class method.

I often find myself adding virtual accessors to construct objects based on other attributes within the model, most commonly `DateTime` instances. Currently, the result of an accessor is cached if it's an object and only invalidated if the attribute's mutator is called. Scalar types are unaffected.

This PR adds a flag that will disable this behaviour. Mutating multiple attributes & value object casts will be unaffected by default.

## Example

```php
public function finish(): Attribute
{
    return Attribute::getWithoutCaching(fn () => $this->start->addSeconds($this->duration)));
}
```
or

```php
public function finish(): Attribute
{
    return Attribute::get(fn () => $this->start->addSeconds($this->duration)))->disableObjectCaching();
}
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Casts/Attribute.php
+++ b/src/Illuminate/Database/Eloquent/Casts/Attribute.php
@@ -19,7 +19,7 @@ class Attribute
     public $set;
 
     /**
-     * Indicates if caching of objects should be disabled for this attribute.
+     * Indicates if caching of objects is enabled for this attribute.
      *
      * @var bool
      */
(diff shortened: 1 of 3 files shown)


taylorotwell:

RoodFruit:
@decnorton @taylorotwell shouldn't this be the default behavior of accessors (virtual attributes)? Why are they cached in the first place?

PR #31707, opened 2020-03-03
[7.x] Add str() helper

calebporzio:
It's no secret, I love me a good string helper.

It was a sad day for me when they were all gone from the core...

BUT! HERE is the one helper to rule them all: `str()`

This will be the last string helper ever pulled into Laravel. The helper to end all helpers.

Witness its beauty:
```php
str('how majestic')->finish(' is the str helper');
str('watch it dance')->between('watch ', 'dance');
str('with such grace')->explode(' ')->implode(' ');
```

We have `collect()` and love it, `str()` is a natural companion.

PR #40520, opened 2022-01-20
[9.x] Adds `str()` helper

nunomaduro:
This pull request proposes the addition of the `str` helper exactly as proposed here: https://github.com/laravel/framework/pull/31707.


dennisprudlo:
@nunomaduro Any thoughts on something like this?

```php
use Illuminate\Support\Str;

function str (?string $string = null) : mixed {
    if (! isset($string)) {
        return new class {
            public function __call (string $name, array $arguments) : string|\Illuminate\Support\Stringable {
                return Str::$name(...$arguments);
            }
        };
    }

    return Str::of($string);
}
```

This would allow for `str('hello')->upper()` as well as `str()->random(64)` or `str()->uuid()`.

PR #40524, opened 2022-01-20
[9.x] Extend str() helper to provide access to the Str class

dennisprudlo:
### What?

Recently @nunomaduro added a `str()`-helper in https://github.com/laravel/framework/pull/40520. On of the reasons for a `str()`-helper in the original PR (https://github.com/laravel/framework/pull/31707) was that this helper would reduce the imports of the `Illuminate\Support\Str` class.

This PR extends the recently added string helper to provide access to all other methods of the Str class.
```php
str('some-string'); // \Illuminate\Support\Stringable as per the referenced PR
str()->random(16); // QDae31W1nFshBV97
str()->uuid(); // 85d57da7-12d0-4594-ad00-ff1eb1bcfff9
```

### Why?
With the current state of this helper you will still need to use `Illuminate\Support\Str` in your file when you use a function different from `Str::of`. With this PR there is no need to use the Str class whenever you want to generate a random string or use other functions of the Str class.

### Return type
I added `mixed` as a return type because the string helper can now return an anonymous class. I know that strictly speaking mixed includes Stringable, but without explicitly having Stringable in the doc, IDE autocomplete wouldn't work

### DIFF
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -261,11 +261,21 @@ if (! function_exists('str')) {
     /**
      * Get a new stringable object from the given string.
      *
-     * @param  string  $string
-     * @return \Illuminate\Support\Stringable
+     * @param  string|null  $string
+     * @return \Illuminate\Support\Stringable|mixed
      */
-    function str($string)
+    function str($string = null)
     {
+        if (! isset($string)) {
+            return new class
+            {
+                public function __call($method, $parameters)
+                {
+                    return Str::$method(...$parameters);
+                }
+            };
+        }
+
         return Str::of($string);
     }
 }
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -266,7 +266,7 @@ if (! function_exists('str')) {
      */
     function str($string = null)
     {
-        if (! isset($string)) {
+        if (func_num_args() === 0) {
             return new class
             {
                 public function __call($method, $parameters)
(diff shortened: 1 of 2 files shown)


src/Illuminate/Support/helpers.php
```diff
@@ -261,11 +261,21 @@ function retry($times, callable $callback, $sleepMilliseconds = 0, $when = null)
...
      *
-     * @param  string  $string
-     * @return \Illuminate\Support\Stringable
+     * @param  string|null  $string
+     * @return \Illuminate\Support\Stringable|mixed
      */
-    function str($string)
+    function str($string = null)
     {
+        if (! isset($string)) {
```
nunomaduro, line 269: This should be `if (func_num_args() == 0) {` - as null is equivalent to `''` when coerced to string:

```php
// str($post->title)->isEmpty();
// etc...
```

tests/Support/SupportHelpersTest.php
```diff
@@ -371,6 +371,10 @@ public function testStr()
 
         $this->assertInstanceOf(Stringable::class, $stringable);
         $this->assertSame('string-value', (string) $stringable);
+
```
nunomaduro, line 374: This test should also pass:

```
        $stringable = str($name = null);
        $this->assertInstanceOf(Stringable::class, $stringable);
        $this->assertTrue($stringable->isEmpty());
```

nunomaduro (changes requested):
Added a few comments. Note, **I am not sure if this pull request is a good idea**.

src/Illuminate/Support/helpers.php
dennisprudlo: Right, didn't think of that. Will fix that in a moment.

PR #40526, opened 2022-01-20
[8.x] Match Redirect Facade DocBlock to actual class

kemp:
The current Redirect Facade DocBlocks show `$parameters` as an array when in fact `mixed` is accepted (according to Redirector.php), causing an IDE warning (at least in PHPStorm).

Before:

```php
$postId = 3;

Redirect::route('posts.show', $postId); // IDE warning, but works perfectly
```

After:

```php
$postId = 3;

Redirect::route('posts.show', $postId); // No IDE warning
```


kemp:
Thanks! :+1:

Issue #40484, opened 2022-01-19
Bitwise operators not working properly for PostgreSQL


Arzaroth:
- Laravel Version: 8.79.0
- PHP Version: 7.4.27
- Database Driver & Version: MariaDB 10.6.5, PostgreSQL 13.5

### Description:

MariaDB supports a bitwise AND operator as follows:

~~~sql
SELECT 'a' WHERE (3 & 1);
~~~

PostgreSQL also does, but requires a bool cast like so:

~~~sql
SELECT 'a' WHERE (3 & 1):::bool;
~~~

The expression itself produces an integer, so I understand the necessity to cast in raw SQL.
However, given this operator is in the list of available ones, alongside other bitwise operators (e.g. |, ^, &~, etc...) I fail to see how it is supposed to be used.

### Steps To Reproduce:

Let's pretend you have a Model named `Order`, with a `status` column. Since the status could be a combination of many, we'll use a bitmask in order to "combine" multiple status in one column.

Then, we should be able to query the table like this :

~~~sql
$query = Order::query();
// doing some stuffs
$query->where('status', '&', 1)->get();
~~~

This yields a `PDOException`.

What we currently have to do :

~~~sql
$query = Order::query();
// doing some stuffs
$query->where($query->raw('(status & 1)'), '!=', 0)->get();
~~~

This could cause some problems, since the column name is no longer escaped.

driesvints:
This seems like a feature request? I don't think Laravel is capable of handling these right now.

Arzaroth:
I would beg to differ.
Laraval boasts compatibility with the following databases :
 - MySQL
 - PostgreSQL
 - SQLite
 - SQL Server

Thus, I would expect a level of abstraction so my code would run using any databases previously mentionned (unless raw is involved of course).
JSON expressions are a perfect example of this.

Binary operators already are in the list of availables operators as shown here :
https://github.com/laravel/framework/blob/d11d431ff271daa9e170567a04a5679e9197523f/src/Illuminate/Database/Query/Builder.php#L197
They are, in my understanding, solely for WHERE and HAVING purposes.

From what I've gathered, $query->where('field', '&', $value) works on MySQL/MariaDB & SQLite and doesn't on PostgreSQL & SQL Server.
An approach that works everywhere is '(field & $value) != 0'.

By no means I'm trying to implying that such a change would be easy to implement.

I've snooped around and I have a basic idea of what should be done for this to work.
I'll try to submit a PR soonish.

[...]

PR #40529, opened 2022-01-21
[8.x] Add better bitwise operators support

Arzaroth:
### Impetus

This addresses the problem mentioned in issue #40484

### Changes

Since the bitwise operators are fundamentally different from other, in the sense that they do not produce boolean expression but integer based expressions, I elected to treat them differently from others.

Currently, both MySQL and SQLite support "implicit casting" of integer to boolean, but SQL Server and PostgreSQL do not.
The idea is to have a uniform behavior across all supported databases.

Previously this would yield a PDOException on PostgreSQL for instance :
~~~php
$builder->select('*')->from('foo')->where('bar', '&', 1)->get(); // MySQL is cool with that, PostgreSQL looks upon you and declares that argument of WHERE must be type boolean, not type integer
~~~

The proposed implementation is database agnostic (as far as I'm aware at least). This is done by forcing a comparison, thus transforming the expression into a boolean one.

### What next

This could open the door to support for basic arithmetic operators (e.g. `+`, `-`, etc...), although I would argue those appear even more elusively in WHERE clauses.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -204,6 +204,15 @@ class Builder
     ];
 
     /**
+     * All of the available binary operators.
+     *
+     * @var string[]
+     */
+    public $binaryOperators = [
+        '&', '|', '^', '<<', '>>', '&~',
+    ];
+
+    /**
      * Whether to use write pdo for the select.
      *
      * @var bool
@@ -754,6 +763,10 @@ class Builder
             }
         }
 
+        if ($this->isBinaryOperator($operator)) {
+            $type = 'Binary';
+        }
+
         // Now that we are working with just a simple query we can put the elements
         // in our array and add the query binding to our array of bindings that
         // will be bound to each SQL statements when it is finally executed.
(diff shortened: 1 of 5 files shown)


src/Illuminate/Database/Query/Builder.php
```diff
@@ -837,6 +850,12 @@ protected function invalidOperator($operator)
                ! in_array(strtolower($operator), $this->grammar->getOperators(), true);
     }
 
+    protected function isBinaryOperator($operator)
+    {
+        return in_array(strtolower($operator), $this->binaryOperators, true) ||
+               in_array(strtolower($operator), $this->grammar->getBinaryOperators(), true);
```
Jubeki, line 856: What is your reason for calling `strtolower($operator)`? In my opinion the operator has now lower or upper case.

```suggestion
        return in_array($operator, $this->binaryOperators, true) ||
               in_array($operator, $this->grammar->getBinaryOperators(), true);
```

Arzaroth: None in particular, I copied what I saw in the `invalidOperator`.

PR #40534, opened 2022-01-21
[9.x] Precise return type for `Enumerable::partition()`

spawnia:
The return type is actually `static`, not `array`.

See https://github.com/nunomaduro/larastan/pull/1044


driesvints:
Seems like there's conflicts here.

driesvints:
Ping @canvural @nunomaduro

PR #40535, opened 2022-01-21
[8.x] Added getOrPut on Collection

eusonlito:
On often changing collections we need to check every time if a key value exists or add if not:

```php
/**
 * @param string $key
 * @param array $data
 *
 * @return mixed
 */
protected function getValueFromCollection(string $key, array $data): mixed
{
    if ($this->collection->has($key) === false) {
        $this->collection->put($key, $this->create($data));
    }

    return $this->collection->get($key);
}
```

We want to avoid this common check adding a `getOrPut` method on collections and allow to simplify as:

```php
/**
 * @param string $key
 * @param array $data
 *
 * @return mixed
 */
protected function getValueFromCollection(string $key, array $data): mixed
{
    return $this->collection->getOrPut($key, fn () => $this->create($data));
}
```

Also fixed values are valid:

```php
/**
 * @param string $key
 * @param mixed $value
 *
 * @return mixed
 */
protected function getValueFromCollection(string $key, mixed $value): mixed
{
    return $this->collection->getOrPut($key, $value);
}
```

On batch processes with collections caching models creation, this change help us a lot with code simplification.

What do you think?

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -430,16 +430,16 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
      * or add to collection if not exists.
      *
      * @param  mixed  $key
-     * @param  callable  $callback
+     * @param  mixed  $value
      * @return mixed
      */
-    public function getOrPut($key, callable $callback)
+    public function getOrPut($key, mixed $value)
     {
         if (array_key_exists($key, $this->items)) {
             return $this->items[$key];
         }
 
-        $this->offsetSet($key, $value = $callback());
+        $this->offsetSet($key, $value = value($value));
 
         return $value;
     }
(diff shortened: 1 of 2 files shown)


src/Illuminate/Collections/Collection.php
```diff
@@ -425,6 +425,24 @@ public function get($key, $default = null)
...
+    /**
+     * Get an item from the collection by key or add it to collection if it does not exist.
+     *
+     * @param  mixed  $key
+     * @param  mixed  $value
+     * @return mixed
+     */
+    public function getOrPut($key, $value)
+    {
+        if (array_key_exists($key, $this->items)) {
```
siarheipashkevich, line 437: Why you don't use 'has` method here?

eusonlito: Because `has` allow an array as `$key` but `get` and `put` not.

PR #40538, opened 2022-01-21
[9.x] Adds support for "chainable" taps

nunomaduro:
Originally posted https://github.com/laravel/framework/pull/40505, this pull request is an alternative implementation making the `HigherOrderTapProxy` immutable, and contains slightly different method / property names.

```php
$slug = tap($user)
    ->chain()
    ->save()
    ->withoutBroadcasting(fn () => /** ... */ })
    ->then(fn ($user) => $user->slug); // breaks the chain, and returns the callback result

$slug = tap($user)
    ->chain()
    ->save()
    ->withoutBroadcasting(fn () => /** ... */ })
    ->then() // breaks the chain, uses the $user object now...
    ->slug;
```

### DIFF
--- a/src/Illuminate/Support/HigherOrderTapProxy.php
+++ b/src/Illuminate/Support/HigherOrderTapProxy.php
@@ -12,14 +12,23 @@ class HigherOrderTapProxy
     public $target;
 
     /**
+     * If the tap proxy is chainable.
+     *
+     * @var bool
+     */
+    public $chainable;
+
+    /**
      * Create a new tap proxy instance.
      *
      * @param  mixed  $target
+     * @param  bool  $chainable
      * @return void
      */
-    public function __construct($target)
+    public function __construct($target, $chainable = false)
     {
         $this->target = $target;
+        $this->chainable = $chainable;
     }
 
     /**
@@ -33,6 +42,26 @@ class HigherOrderTapProxy
     {
         $this->target->{$method}(...$parameters);
 
+        return $this->chainable ? $this : $this->target;
+    }
+
+    /**
+     * Create a new chainable tap proxy.
+     *
+     * @return static
+     */
+    public function chain()
+    {
+        return new static($this->target, true);
+    }
+
+    /**
+     * Return the target being tapped.
+     *
+     * @return mixed
+     */
+    public function then()
+    {
         return $this->target;
     }
 }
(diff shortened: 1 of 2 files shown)


foremtehan:
using promoted properties is prohibited in laravel ?

dakira:
I like the optional callback on then(), great idea.

chu121su12:
How about `when`/`unless` support?

DarkGhostHunter:
> How about `when`/`unless` support?

That could collide if the class uses `Conditionable`.

chu121su12:
Is it just like when the target has already defined `chain`/`then` method?

Jubeki:
> That could collide if the class uses `Conditionable`.

@DarkGhostHunter if it already supports Conditional isn't it likely that this Class is chainable and would not need to be tapped?

DarkGhostHunter:
That’s right lol

Italo Baeza C.

timacdonald:
Just on naming conventions, the pipeline, which has a similar vibe to this, has a `thenReturn()` method.

Might be good to either support that here, or alternatively support the argument-less `then()` method on the pipeline as well - just for consistency across similar APIs...or both on both.

JosephSilber:
Wouldn't it make more sense to have a standalone `chain` helper? The way it is here, you can no longer call `chain` on the underlying object. That's unexpected!

DarkGhostHunter:
That was my original implementation long time ago, with different name. 

I think it’s simpler to do chain, and use “andThen” to exit the chain and return the underlying object. 

Add a Conditionable trait (since using an IF block would mean to break the chain) and that’s it. I believe this doesn’t need to be part of the tap helper at all, but the tap helper class could add the “chain” method to pass the object to the “chain” helper. 

This would work wonders on objects witch results are void or you want to discard. 

    return chain($message)->save()->rebuildCache();

Italo Baeza C.

chu121su12:
Should we leave `tap` as is and add new helper instead as in #40505 ? Maybe under different name. Because currently my assumption for `tap` is that any method passed in will be attempted on the original target.

@DarkGhostHunter's `chain` is good too

One fun thing I could see is if you change the `target` property to `then` (or add magic accessor), you can do `chain(...)->a()->b()->then->callTargetFn()` to simulate the `then()` call.

PR #40539, opened 2022-01-21
[8.x] Pruning Models: Get the default path for the models from a method instead

sander3:
Hi there,

We want to use the "Pruning Models" feature but use a custom model namespace. The Artisan make command allow for overriding the used namespace by creating your own `getDefaultNamespace` method.

This PR adds the same functionality to the `PruneCommand`.


sander3:
Thanks!

PR #40540, opened 2022-01-21
Improve types for Collection::groupBy()

spawnia:
Currently only a failing test case. Given `Illuminate\Support\Collection::groupBy()` returns `static<array-key, static<array-key, TValue>>`, I would expect PHPStan to respect late static binding.

See https://github.com/nunomaduro/larastan/pull/1054

@canvural @nunomaduro

### DIFF
--- a/types/Database/Eloquent/Collection.php
+++ b/types/Database/Eloquent/Collection.php
@@ -5,11 +5,11 @@ use function PHPStan\Testing\assertType;
 $collection = User::all();
 assertType('Illuminate\Database\Eloquent\Collection<int, User>', $collection);
 
-// @phpstan-ignore-next-line TODO remove ignore when https://github.com/phpstan/phpstan/issues/5512 is fixed
-assertType(
-    'Illuminate\Database\Eloquent\Collection<(int|string), Illuminate\Database\Eloquent\Collection<(int|string), User>>',
-    $collection->groupBy('string')
-);
+// TODO reinclude when https://github.com/phpstan/phpstan/issues/5512 is fixed
+//assertType(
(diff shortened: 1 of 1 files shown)


canvural:
Yes, this is expected. Support for `static<>` types in PHPStan is really basic. So this is solved by an extension in Larastan. You can see [here](https://github.com/nunomaduro/larastan/blob/master/tests/Type/data/collection-generic-static-methods.php#L26) similar test that is passing.

spawnia:
I think PHPStan is just plain wrong here, isn't it? I see you have invested quite a bit of effort to work around this issue, is there a bug report or relevant discussion in PHPStan's issue tracker?

canvural:
> I think PHPStan is just plain wrong here, isn't it? I see you have invested quite a bit of effort to work around this issue, is there a bug report or relevant discussion in PHPStan's issue tracker?

It's not wrong. It's not implemented yet. [Here](https://github.com/phpstan/phpstan/issues/5512) is the feature request.

spawnia:
Thanks for your insight @canvural. I propose to explicitly acknowledge this fault for now and hopefully include the test once PHPStan is fixed. I would have preferred to instruct PHPStan to ignore the error, as that would notify us when it is fixed, but that does not seem possible with `assertType()`.

taylorotwell:
Can you resubmit this when PHPStan is fixed?

spawnia:
I was hoping to get the reminder into the test code so that anyone can add it back in once it is fixed. I did subscribe to the PHPStan issue for now, perhaps I will get notified and remember this issue.

PR #40541, opened 2022-01-21
[8.x] Add only and except features to MorphTo

damiencriado:
### Why

We have a use case that cannot be solved actually.
We have a lot of `Comment` and we want to eager load only specific morphable types. Actually, all types are eager loaded.

#### Example:

The `Comment` model can morph to `Post` or `Video`.
We want to fetch all `Comment` with `commentable` only if the `commentable_type` is type of `Post::class`.
We don't want to eager load `Video::class` firstly because the relation is not used and secondly to save memory and database queries.

### Proposal

Adding 2 features to `MorphTo`:

- `only(array $types)`
  - Eager load only given morph types
- `except(array $types)`
  - Prevent eager load of given morph types

To make this behavior easily understandable by developers, it seems that the best place to call these functions is in the `with()` callback:

```php
$comments = Comment::with(['commentable' => function (MorphTo $morphTo) {
                $morphTo->only([Post::class]);
                // $morphTo->except([Video::class]);
            }])
            ->get();
```

The main challenge here is that the dictionary containing relations to eager load is built before we can interact with it. In other words, the dictionary is already built in the `with` callback.

Then the `getResultsByType()` function will eager load all morph types present in the dictionary. And this is exactly what we want to avoid here. To prevent undesirable eager loads by the `getResultsByType()` function, morph types that must not be eager loaded are removed from `$dictionary`.

But due to the `eagerLoadRelation()` behavior in the `Builder` class, the relation even if it should not be loaded, is initialized with `null` value. It's a bad thing because the relation is now considered as loaded by the `relationLoaded()` function. And therefore can't be lazy-loaded. To keep a valid behavior of the eager load mechanism we need to prevent relation initialization by checking `$dictionary`.

# Update
- I have found a better way to prevent relation initialization directly in `initRelation()` function.
- Method renamed to `only` and `except`

----

- This can solve some heavy performances issues on morphable eager loads
- I have added tests that cover this behavior
- Functions can be renamed if they seem ambiguous to you
- I did my best to add these features by limiting changes as much as possible.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/MorphTo.php
+++ b/src/Illuminate/Database/Eloquent/Relations/MorphTo.php
@@ -424,10 +424,11 @@ class MorphTo extends BelongsTo
     /**
      * Check if morphable type is excluded from dictionary.
      *
-     * @param  string $type
+     * @param  string  $type
      * @return bool
      */
-    protected function isMorphTypeExcluded($type) {
+    protected function isMorphTypeExcluded($type)
+    {
         $morphTypeKey = $this->getDictionaryKey($type);
 
         return array_key_exists($morphTypeKey, $this->exclusionDictionary);
--- a/src/Illuminate/Database/Eloquent/Relations/MorphTo.php
+++ b/src/Illuminate/Database/Eloquent/Relations/MorphTo.php
@@ -401,7 +401,7 @@ class MorphTo extends BelongsTo
             // Prevent relation initialization for excluded morph types.
 
             $morphTypeKey = $this->getDictionaryKey($model->{$this->morphType});
-            $relationNotExcluded = !array_key_exists($morphTypeKey, $this->exclusionDictionary);
+            $relationNotExcluded = ! array_key_exists($morphTypeKey, $this->exclusionDictionary);
 
             if ($relationNotExcluded) {
                 $model->setRelation($relation, $this->getDefaultFor($model));
--- a/src/Illuminate/Database/Eloquent/Relations/MorphTo.php
+++ b/src/Illuminate/Database/Eloquent/Relations/MorphTo.php
@@ -398,12 +398,7 @@ class MorphTo extends BelongsTo
     public function initRelation(array $models, $relation)
     {
         foreach ($models as $model) {
-            // Prevent relation initialization for excluded morph types.
-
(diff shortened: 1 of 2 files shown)


damiencriado:
#### Update
I have found a better way to prevent relation initialization directly in `initRelation()` function. It's a lot cleaner now.

morloderex:
@damiencriado How is this different from [constraining eager loading of morphTo relationships](https://laravel.com/docs/8.x/eloquent-relationships#constraining-eager-loading-of-morph-to-relationships)?

damiencriado:
> @damiencriado How is this different from [constraining eager loading of morphTo relationships](https://laravel.com/docs/8.x/eloquent-relationships#constraining-eager-loading-of-morph-to-relationships)?

This is different because `constrain` adds constraints to the morphable type root query.

In this example, the `Post` query is executed with the eager constraints + given constrain `whereNull`.

```php
$comments = Comment::with(['commentable' => function (MorphTo $morphTo) {
    $morphTo->constrain([Post::class => function (Builder $query) {
        $query->whereNull('foo');
    }]);
}])->get();
```

In my use case and using `morphExcept` the `Post` query is not executed at all. And the `commentable` relation of `Comment` is not loaded. Giving the ability to lazy load it further.

damiencriado:
> We've fixed the build on 8.x so please rebase and mark as ready when tests pass. Thanks!

@driesvints It seems that the CI is still broken 😕

taylorotwell:
I'm sure I'm missing something but why not just:

```php
$comments = Comment::with(['commentable'])->where('commentable_type', Post::class)->get();
```

damiencriado:
@taylorotwell It will doesn't work for this use case because this query will only fetch `Comment` where type is `Post`.
What I want to achieve here is to fetch all `Comment` (regardless of type) while only eagerloading the `commentable` relation when the type is `Post`. This will save us unnecessary queries and the `Comment` collection will be much lighter.

taylorotwell:
This?

```php
    $comments = Comment::all()
        ->tap(fn ($comments) => $comments
                ->where('commentable_type', Post::class)
                ->each
                ->load(['commentable']));
```

damiencriado:
Thanks for the trick @taylorotwell, it seem to works as expected !

taylorotwell:
Glad to hear it!

PR #40544, opened 2022-01-21
[8.x] Improve PhpRedis flushing

tillkruss:
Followup to #40446 to make PhpRedis connections behave like Predis as usual.

- Move cluster flush logic into `PhpRedisClusterConnection`
- Added support for `ASYNC` modifier to both PhpRedis `flushdb()` methods
- `PhpRedisClusterConnection::flushdb()` will now fire `CommandExecuted` events as well

### DIFF
--- a/src/Illuminate/Redis/Connections/PhpRedisClusterConnection.php
+++ b/src/Illuminate/Redis/Connections/PhpRedisClusterConnection.php
@@ -12,6 +12,7 @@ class PhpRedisClusterConnection extends PhpRedisConnection
     public function flushdb()
     {
         $arguments = func_get_args();
+
         $async = strtoupper((string) $arguments[0] ?? null) === 'ASYNC';
 
         foreach ($this->client->_masters() as $master) {
--- a/src/Illuminate/Redis/Connections/PhpRedisClusterConnection.php
+++ b/src/Illuminate/Redis/Connections/PhpRedisClusterConnection.php
@@ -7,12 +7,17 @@ class PhpRedisClusterConnection extends PhpRedisConnection
     /**
      * Flush the selected Redis database.
      *
+     * @param  string|null  $modifier
      * @return mixed
      */
-    public function flushdb()
+    public function flushdb($modifier = null)
     {
+        $async = strtoupper($modifier) === 'ASYNC';
+
         foreach ($this->client->_masters() as $master) {
-            $this->client->flushdb($master);
+            $async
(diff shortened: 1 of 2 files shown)


src/Illuminate/Redis/Connections/PhpRedisClusterConnection.php
```diff
@@ -4,5 +4,20 @@
...
-    //
+    /**
+     * Flush the selected Redis database.
+     *
+     * @param  string|null  $modifier
+     * @return mixed
+     */
+    public function flushdb($modifier = null)
+    {
+        $async = strtoupper($modifier) === 'ASYNC';
```
derekmd, line 15: You may want to `strtoupper($modifier ?? '')` for PHP 8.1+ because standard library string functions no longer allowing `null`  arguments was a feature devs have been absolutely clamoring for.

```php
strtoupper(null);
// PHP Deprecated:  strtoupper(): Passing null to parameter #1 ($string) of type
// string is deprecated in eval()'d code on line 1
```

driesvints:
I know nothing of the internals here so I'm not the right person to review this.

src/Illuminate/Redis/Connections/PhpRedisConnection.php
```diff
@@ -495,17 +494,17 @@ public function createSubscription($channels, Closure $callback, $method = 'subs
...
     public function flushdb()
     {
-        if (! $this->client instanceof RedisCluster) {
-            return $this->command('flushdb');
-        }
+        $arguments = func_get_args();
 
-        foreach ($this->client->_masters() as $master) {
-            $this->client->flushDb($master);
+        if (strtoupper((string) $arguments[0] ?? null) === 'ASYNC') {
```
TheLevti, line 503: This causes a lot of tests to fail for me.

It should be `if (strtoupper((string) ($arguments[0] ?? null)) === 'ASYNC') {`

TheLevti: As sometimes there are no arguments at all.

driesvints: @tillkruss ^

tillkruss: I'll make a PR with tests and the changes.

tillkruss: See https://github.com/laravel/framework/pull/40571

PR #40545, opened 2022-01-21
[9.x] Adds `package` argument to `vendor:publish` command

nunomaduro:
This pull request introduces a few changes to the `vendor:publish` command.

Screenshot 2022-01-25 at 14 52 34

---

So, if this pull request gets merged, to get started, when users type `php artisan vendor:publish`, they will be asked to select a package, instead of selecting a provider class:
<p align="center">

</p>
As you can see, I've added a small description next to the package name, and the description contains details of the files that will be published. In addition, the user can enjoy nice autocompletion while selecting the package:
<p align="center">

</p>

Next, once the user selects the package, he will be able to select which files (tags) he wish to publish:
<p align="center">

</p>

But of course, if there is 1 or 0 tags, we will publish everything of that package.
<p align="center">

</p>

Now, this pull request also adds the new argument `package` that users may use to specify the package directly. Yet, we will question the user about the tags, if the user don't specify it via the `--tag` option:
<p align="center">

</p>

Next, the user may combine the argument `package` with the `--all` option, to publish everything related to the given package:
<p align="center">

</p>

Finally, it's important to keep in mind that there is no breaking changes - and previous options work on the same way:

<p align="center">

</p>

### DIFF
--- a/src/Illuminate/Foundation/Console/VendorPublishCommand.php
+++ b/src/Illuminate/Foundation/Console/VendorPublishCommand.php
@@ -28,7 +28,7 @@ class VendorPublishCommand extends Command
         'docker' => 'docker files',
         'public' => 'assets',
         'resources/views' => 'views',
-        'resources/lang' => 'resources',
+        'resources/lang' => 'translations',
         'sail' => 'binary',
     ];
 
--- a/src/Illuminate/Foundation/Console/VendorPublishCommand.php
+++ b/src/Illuminate/Foundation/Console/VendorPublishCommand.php
@@ -168,7 +168,7 @@ class VendorPublishCommand extends Command
         $providerPackage = null;
 
         while (true) {
-            if ($currentPackageFolder == dirname(base_path())) {
+            if ($currentPackageFolder == '/' || $currentPackageFolder == dirname(base_path())) {
                 return;
             }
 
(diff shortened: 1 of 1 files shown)


dillingham:
Looks very nice, what if..

[1] **spark-config** config/spark.php

If numeric use index, else use autocomplete

ousid:
Great addition @nunomaduro 

What if, instead of creating the package name, you can just choose the number based on the package index?

image

negoziator:
Really great work once again @nunomaduro 🙏

robsontenorio:
How about “checkboxes” for multiple selection ?

dillingham:
> How about “checkboxes” for multiple selection ?

@renoki-co mentioned comma separate on twitter.. thought that would be pretty cool. config, views

rennokki:
@dillingham @robsontenorio @nunomaduro Yeah, pretty much selecting more would be useful. It works when deploying Sail assets 🤔 I think this can be a good way to specify more comma-separated values to avoid re-running the command for each number.

src/Illuminate/Foundation/Console/VendorPublishCommand.php
```diff
@@ -5,15 +5,33 @@
...
+     */
+    protected $directoryDescriptions = [
+        'app/Providers' => 'providers',
+        'config' => 'config',
+        'database/migrations' => 'migrations',
+        'database' => 'database files',
+        'docker' => 'docker files',
+        'public' => 'assets',
+        'resources/views' => 'views',
+        'resources/lang' => 'translations',
```
taylorotwell, line 31: `lang` is in the root application directory in L9. May want to have both here.

nunomaduro:
@taylorotwell done.

PR #40550, opened 2022-01-22
query filter feature using pipeline

alirezaGAlizade:
This pull request proposes a brand new `make:filter` command - regarding the way we can apply filters with queries after we received the request contains some parameters what should filter our data.

###  What problem does this pull request solves?
<hr>
Imagine, you have a form and this form contains some parameters like `name` and `age`
now if you want to filter your data there are multiple ways for example one of the usual ways is:

1. checking incoming request
2. see if the request contain `name` then apply a query on your model

now, what if you have multiple forms and you want to apply those forms on multiple models?
it will be **messy** and **frustrating**

### what will you earn with using this new feature?

- reusable queries
- apply queries using pipeline and simple helper
- high customizable classes
- clean code
- checking the request automatically
- saving time

### Let me propose it by some screenshots
<hr>

**make a reusable query:**

Screenshot from 2022-01-22 04-47-24

**A new class will be generated:**

here we have a class that include `$filterParameter` and  `apply` method
so, we can customize the filter parameter incoming from request and also change the query from the `apply` method
here we have access to the request by `$this->request`

Screenshot from 2022-01-22 04-48-08

**now let's use it on User model**

there is a helper `filter_pipes` incoming with this feature and it takes 2 parameter 
the first parameter is the model or the repository
and the second parameter is an array of generated classes

Screenshot from 2022-01-22 05-06-26

That's it **clean** and **easy** :1234:  :eagle:


driesvints:
Seems something went wrong here with the commit history

PR #40551, opened 2022-01-22
[8.x] Fixed the appends property to work with the new Accessor method

nickakitch:
When using the new attribute casting method from PR #40022 you are currently unable to include the method in the 'appends' property as you would when using the older method.

Currently:
```
$appends = ['title'];

public function title(): Attribute
{
    return ucfirst($this->name);
}

$model->toArray(); // throws a BadMethodCallException
```

This PR allows methods that return the Attribute object to be included in the 'appends' property

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -613,7 +613,7 @@ trait HasAttributes
     {
         try {
             return $this->{'get'.Str::studly($key).'Attribute'}($value);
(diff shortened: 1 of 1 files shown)


taylorotwell:
Works fine for me - your attribute is not defined correctly?

```php
public function title(): Attribute
{
    return new Attribute(
        fn ($value, $attributes) => ucfirst($attributes['name']),
    );
}
```

nickakitch:
From the looks of things the case matters:

```
$appends = ['firstName']; // won't work
$appends = ['first_name']; // works
```

With the old method camel case also worked

PR #40555, opened 2022-01-22
[10.x] Return a `Stringable` from the Translator instead of a regular string

ralphjsmit:
This pull request adds the ability to return a `Stringable` from the Translator instead of normal strings. This allows the developer to easily manipulate strings before being displayed, for example to change their case or to concatenate translations.

## Example

Consider the following translation file:

```php
return [
    'file' => 'File',
    
    'phrases' => [
        'select' => 'Select',
        'search' => 'Search',
    ]
];
```

Now, the developer wants to display the text 'Select file'. In the old situation, the developer would have to do something like this:

```blade
{{ Str::of(__('phrases.select'))->append(' ')->append(__('file'))->lower()->ucfirst() }}

// or

{{ Str::ucfirst(__('phrases.select')) }} {{ Str::lower(__('file')) }}
```

When returning a `Stringable`, the developer will now be able to do something like this:

```blade
{{ __('phrases.select')->append(' ')->append(__('file'))->lower()->ucfirst() }}
```

Another example:
```blade
{{ Str::upper(__('phrases.search')) }}

// becomes

{{ __('phrases.search')->upper() }}
```

The beauty is that PHP will always coerce a `Stringable` to a string when needed, so it will not be a big backwards incompatible change, whilst still providing developers with a lot of fluent methods on a translation. 

*I noticed that there are several validation errors failing in the testsuite. I suspect that is because of it trying to do an operation on a `Stringable` that is only allowed on a string. I've really looked into this, but I can't find the place where the problem is. I would be grateful if someone with more experience can explain where the problem lies, so that I can fix it, or if that person would be willing to fix it himself. Most likely it's only a small change somewhere in the Validator or MessageBag. Thanks!*

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -823,7 +823,7 @@ if (! function_exists('trans')) {
      *
      * @param  string|null  $key
      * @param  array  $replace
-     * @param  string|null  $locale
+     * @param  Stringable|null  $locale
      * @return \Illuminate\Contracts\Translation\Translator|string|array|null
      */
     function trans($key = null, $replace = [], $locale = null)
@@ -858,7 +858,7 @@ if (! function_exists('__')) {
      *
      * @param  string|null  $key
      * @param  array  $replace
-     * @param  string|null  $locale
+     * @param  Stringable|null  $locale
      * @return string|array|null
      */
     function __($key = null, $replace = [], $locale = null)
--- a/src/Illuminate/Translation/Translator.php
+++ b/src/Illuminate/Translation/Translator.php
@@ -8,6 +8,7 @@ use Illuminate\Contracts\Translation\Translator as TranslatorContract;
 use Illuminate\Support\Arr;
 use Illuminate\Support\NamespacedItemResolver;
 use Illuminate\Support\Str;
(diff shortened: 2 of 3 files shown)


Jubeki:
It seems like the failure comes from the Validator:
```diff
78) Illuminate\Tests\Validation\ValidationValidatorTest::testFailOnFirstError
Failed asserting that two arrays are equal.
--- Expected
+++ Actual
@@ @@
 Array (
     'baz' => Array (
-        0 => 'validation.required'
+        0 => 'validation.custom.baz.required'
     )
     'age' => Array (
-        0 => 'validation.min.string'
+        0 => 'validation.custom.age.min.string'
     )
 )

D:\a\framework\framework\tests\Validation\ValidationValidatorTest.php:6914
```

(This is only one of the many failed tests)

ralphjsmit:
@Jubeki Thank you! Yes, I saw that it was coming from the Validator (and the MessageBag). It isn't removing a part like 'custom.XXXXX', but I meant more the internal part of the Validator that's responsible for this.

I noticed that if I changed only one line of the Translator (line 155) back to `return $line`, all tests work again.

taylorotwell:
Too much of a breaking change IMO.

PR #35901, opened 2021-01-15
[9.x] Remove reduceWithKeys and merge the functionality with reduce

mokhosh:
This moves `reduce` out of `Collection` and `LazyCollection` and puts it into `EnumeratesValues` as suggested by @JosephSilber.

The existing test for `reduce` has be adapted to contain asserts for both situations, with and without `$key`.

I've also moved `tap` one method down to be alphabetically correct.

@JosephSilber while I was checking the code in `EnumeratesValues` I got an idea. Can we make `reduce` higher proxy-able as well? I don't know how to do that now, but I'm willing to do it if you give me a little guidance.

PR #35878, opened 2021-01-14
[8.x] Pass $key to closure in Collection and LazyCollection's reduce method as well

mokhosh:
Following the discussion we had here https://github.com/laravel/framework/pull/35839 I've removed `reduceWithKeys` and changed the `reduce` method to pass `$key` to its closure as well as  `$value`.

The previous test is there. I've just added another test for closures that accept the key, since I thought you might want to have one assertion per test. If you prefer to have one test per method and you don't care how many assertions are in one test, we can merge the two tests.

PR #35839, opened 2021-01-10
[8.x] Add reduce with keys to collections and lazy collections

mokhosh:
This adds the method `reduceWithKeys` to `Collection`s and `LazyCollection`s.
Similar to `map` and `mapWithKeys`, this augments `reduce` to pass associative arrays' keys to its callback. We can't do that currently because php's `array_reduce` doesn't pass array keys to its callback.

So, you can do this:
```php
$data = collect([
    'name' => 'Mo Khosh',
    'username' => 'mokhosh',
]);

return $data->reduceWithKeys(function($carry, $value, $key) {
    return $carry . $key . ': ' . $value . PHP_EOL;
});
```

I tried to mirror what's already been done with `map`, `mapWithKeys`, `reduce`, and respective tests.

PR #40566, opened 2022-01-23
[9.x] Remove useless `reduceWithKeys` method

JosephSilber:
The original introduction of `reduceWithKeys` was kind of not ideal [for two reasons](https://github.com/laravel/framework/pull/35839#issuecomment-759551607): the name is inconsistent with the other `*WithKeys` method, and it really should've never been a separate method - `reduce` itself could handle keys too.

...and [that's what we ultimately did](https://github.com/laravel/framework/pull/35878): we added `$key` support to the regular `reduce` method.

[We decided](https://github.com/laravel/framework/pull/35878#discussion_r557434475) to keep `reduceWithKeys` in 8.x, and then remove it completely in 9.x. I think #35901 was trying to do that, but evidently it's still around. So this PR removes it.

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -803,21 +803,6 @@ trait EnumeratesValues
     }
 
     /**
-     * Reduce an associative collection to a single value.
-     *
-     * @template TReduceWithKeysInitial
-     * @template TReduceWithKeysReturnType
-     *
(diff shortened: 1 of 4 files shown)


taylorotwell:
What is the migration path for those using this method already when upgrading rrom 8.x? Please mark as ready for review when provided.

JosephSilber:
Simply switch their uses of `reduceWithKeys` to `reduce`, as they both do the exact same thing.

PR #40571, opened 2022-01-23
[8.x] Avoid `undefined array key 0` error

tillkruss:
Followup to #40544 to avoid `Undefined array key 0` notices.

Added tests to ensure `ASYNC` parameter is accepted.


driesvints:
Thanks @tillkruss 👍

Issue #7620, opened 2015-02-26
Session + Third Party Packages


chadtomkiss:
Hi,

The fact that Session data isn't written until after the request has returned means that it does not get saved when using some third party libraries such as:

https://github.com/thephpleague/oauth1-client

This is due to the way they redirect to Twitter etc:

    header('Location: '.$url);
    exit;

I'm currently calling Session::save() before the library redirects which fixes the issue but thought I would give a heads up.

GrahamCampbell:
```
header('Location: '.$url);
exit;
```

That's going to screw up everything.

Sorry we can't process support requests here. Try the forums.

PR #40572, opened 2022-01-23
[8.x] Mail - Add missing test coverage for global "reply_to" and "to"

clemblanco:
This is just adding some test coverage for something which feels like should be covered.

Before
```
Mail Mailer (Illuminate\Tests\Mail\MailMailer)
 ✔ Global from is respected on all messages  2 ms
```

After
```
Mail Mailer (Illuminate\Tests\Mail\MailMailer)
 ✔ Global addresses are respected on all messages with from  2 ms
 ✔ Global addresses are respected on all messages with reply_to  1 ms
 ✔ Global addresses are respected on all messages with to  1 ms
```

See [laravel/docs #7620](https://github.com/laravel/docs/pull/7620) for context.

### DIFF
--- a/tests/Mail/MailMailerTest.php
+++ b/tests/Mail/MailMailerTest.php
@@ -128,7 +128,17 @@ class MailMailerTest extends TestCase
         unset($_SERVER['__mailer.test']);
     }
 
-    public function testGlobalFromIsRespectedOnAllMessages()
+    public function globalAddressesDataProvider()
+    {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Honestly the data provider makes this more confusing than just calling methods directly.

clemblanco:
Fair enough, was keen to avoid repetition but happy to create separate tests ✌️

PR #40586, opened 2022-01-24
[9.x] Rename `Rule::nested` to `Rule::forEach`

stevebauman:
## Description

Renames `Rule::nested()` to `Rule::forEach()` as discussed in https://github.com/laravel/framework/pull/40498.


stevebauman:
All set, thanks @driesvints!

PR #39261, opened 2021-10-19
[8.x] Adds `withoutDeprecationHandling`

nunomaduro:
This pull request adds  `withoutDeprecationHandling` to the base Laravel's Test Case class so people can opt-in to have deprecations as exceptions in their test suites. Here is an example:

```php
// routes/web.php
Route::get('/', function () {
    str_contains(null, null);

    return view('welcome');
});

// Test A
it('has a home page', function () {
    $response = $this->get('/');

    $response->assertStatus(200); // passes ✅
});

// Test B
it('has a home page', function () {
    $this->withoutDeprecationHandling();

    $response = $this->get('/');

    $response->assertStatus(200); // 🛑 Expected response status code [200] but received 500. The following exception occurred during the request: ErrorException: str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated in /Users/nunomaduro/code/laravel/laravel/routes/web.php:17
});
```

@taylorotwell If this pull request gets merged, let me know if this deserves a section on the docs.

PR #39219, opened 2021-10-15
[8.x] Logs deprecations instead of treating them as exceptions

nunomaduro:
This pull request changes the way Laravel's exception handler treats PHP deprecations.

Until today, Laravel always turn `E_DEPRECATION` or `E_USER_DEPRECATION` level errors into exceptions. In other words, if we were to use `str_contains('', null)` in PHP 8.1, Laravel would convert that deprecation into an Error Exception that gets reported on the Exception Handler, and rendered on the browser.

This pull request - just like Symfony - proposes that we start to ignore PHP deprecations and userland deprecations. Meaning that deprecations like `str_contains('', null)` in PHP 8.1 would not cause the program to stop. So, by default, if this pull request gets merged, no error exceptions appear on the browser or on the logger.

This pull request also introduces a new `deprecations` log channel that can be configured by the user. This channel, uses the `null` driver by default,  yet the user can configure a real driver like so:

```php
<?php

return [

    /*
    |--------------------------------------------------------------------------
    | Default Log Channel
    |--------------------------------------------------------------------------
    |
    | This option defines the default log channel that gets used when writing
    | messages to the logs. The name specified in this option should match
    | one of the channels defined in the "channels" configuration array.
    |
    */

    'default' => env('LOG_CHANNEL', 'stack'),

    /*
    |--------------------------------------------------------------------------
    | Deprecations Log Channel
    |--------------------------------------------------------------------------
    |
    | Lorem ipsum dolor sit amet, consectetur adipiscing elit. Eiuro,
    | adridens, iniquum, hac quidem de re; Nihil enim hoc differt.
    | Quodsi ipsam honestatem undique pertectam atque absolutam.
    |
    */

    'deprecations' => env('LOG_DEPRECATIONS_CHANNEL', 'stack'),

    // ..
]
```

Of course, this pull request also configures PHPUnit differently, using the `convertDeprecationsToExceptions` flag so deprecations don't get treated as exceptions on tests.

If this pull request gets merged, changes will be necessary on:

- [ ] Laravel Skeleton: `config/logging.php`, bump framework version, and phpunit version.
- [ ] Lumen Framework: Copy/Past the code of this pull request
- [ ] Lumen Skeleton: `config/logging.php`, bump framework version, and phpunit version.

PR #40592, opened 2022-01-24
[8.x] fix HandleExceptions memory leak in tests

### LINKED ISSUE #40554: Memory leak in tests because HandleExceptions never cleans up
- Laravel Version: 8.x (though any version with the same logic has this issue)
- PHP Version: 8.0.6/8.1.1

### Description:
Theres a problem of [HandleExceptions](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php) keeping the handler itself and the old [Application](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Foundation/Application.php) instances throughout the whole test run, this is because of how `set_error_handler`, `set_exception_handler` and in addition `register_shutdown_function` work. All of these functions keep all the handlers in a memory stack, meaning they all reside in memory until the equivalent handler is restored or the application shuts down - which only happens when all tests have finished running.

To prove it, we can add:
```php
dump('cleanup', spl_object_id($this), spl_object_id($this->app));
```

to the the `__destruct`  handler or `handleShutdown` function of the `HandleExceptions` and it will only be executed after all tests are done. In addition, it will also execute n times, where n is the number of tests executed and all their ids are different meaning the handler and the application object it's holding to is no the same as previous execution, resulting in memory leak.

We can also test that this is the case using
```php
$prevHandler = set_error_handler([$this, 'handleError']);

// Make sure it's not the PHPUnit handler but our registered array
if (is_array($prevHandler)) {
    $isThis = $prevHandler[0] === $this; // always false
    $isSameApp = $prevHandler[0]->app === $this->app; // always false
}
```

### Steps To Reproduce:

While super basic and barebones test, I created 1000 tests with just bool assertion:

```php
protected function doAction()
{
    $this->assertTrue(true);
}

public function test_example()
{
    $this->doAction();
}

public function test_example2()
{
    $this->doAction();
}

// ... test_example1000
```

just commenting out or storing the handler as a static class variable to keep it single and just switching the $app variable in bootstrap
```php
set_error_handler([$this, 'handleError']);

register_shutdown_function([$this, 'handleShutdown']);
```

saves 6 MB memory in total, tested with forced GC to avoid any memory fluctuations. Now if we have a `__destruct`, it will be called after each test and not after all tests are done, meaning it can now clean the object instantly without waiting for the process to finish. With more complex tests the memory difference can be greatly increased depending on the state the Application is holding on to.

I could make a PR but I have no idea what would be the best approach to keep a single exception handler instead of initiating it every time so knowledge about internals would help here. The main problem I see here is that if the HandleExceptions is to be instantiated every time, it needs to call `restore_error_handler` and `restore_exception_handler` before registering new handlers and `register_shutdown_function` should be registered only once no matter how many instances there are because it will prevent it from cleaning up.

donnysim:
Fix for https://github.com/laravel/framework/issues/40554 . By introducing a separate global handler we can avoid keeping every HandleExceptions with its own Application instance in memory and it can be cleaned up after each test instead of after all tests are done.

Couldn't think of any other way to deal with it without introducing a separate class.

### DIFF
--- a/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php
+++ b/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php
@@ -42,11 +42,7 @@ class HandleExceptions
 
         error_reporting(-1);
 
-        set_error_handler([$this, 'handleError']);
-
-        set_exception_handler([$this, 'handleException']);
-
-        register_shutdown_function([$this, 'handleShutdown']);
+        HandleExceptionsGlobal::register($this);
 
         if (! $app->environment('testing')) {
             ini_set('display_errors', 'Off');
--- a/src/Illuminate/Foundation/Bootstrap/HandleExceptionsGlobal.php
+++ b/src/Illuminate/Foundation/Bootstrap/HandleExceptionsGlobal.php
@@ -1,0 +1,79 @@
+<?php
+
+namespace Illuminate\Foundation\Bootstrap;
+
+use Throwable;
+
+class HandleExceptionsGlobal
+{
+    protected static $instance;
+
+    /**
+     * @var \Illuminate\Foundation\Bootstrap\HandleExceptions
+     */
+    protected $handler;
+
+    protected function __construct()
+    {
+        set_error_handler([$this, 'handleError']);
+        set_exception_handler([$this, 'handleException']);
+        register_shutdown_function([$this, 'handleShutdown']);
+    }
+
+    /**
+     * Register a global exception handler.
+     *
+     * @param  \Illuminate\Foundation\Bootstrap\HandleExceptions  $handler
+     * @return void
+     */
+    public static function register(HandleExceptions $handler)
+    {
+        if (! self::$instance) {
+            self::$instance = new self;
+        }
+
+        self::$instance->handler = $handler;
+    }
+
+    /**
+     * Report PHP deprecations, or convert PHP errors to ErrorException instances.
+     *
+     * @param  int  $level
+     * @param  string  $message
+     * @param  string  $file
+     * @param  int  $line
+     * @param  array  $context
+     * @return void
+     *
+     * @throws \ErrorException
+     */
+    public function handleError($level, $message, $file = '', $line = 0, $context = [])
+    {
+        $this->handler->handleError($level, $message, $file, $line, $context);
+    }
+
+    /**
+     * Handle an uncaught exception from the application.
+     *
+     * Note: Most exceptions can be handled via the try / catch block in
+     * the HTTP and Console kernels. But, fatal error exceptions must
+     * be handled differently since they are not normal exceptions.
+     *
+     * @param  \Throwable  $e
+     * @return void
+     */
+    public function handleException(Throwable $e)
+    {
+        $this->handler->handleException($e);
+    }
+
+    /**
+     * Handle the PHP shutdown event.
+     *
+     * @return void
+     */
+    public function handleShutdown()
+    {
+        $this->handler->handleShutdown();
+    }
+}
--- a/tests/Support/SupportJsTest.php
+++ b/tests/Support/SupportJsTest.php
@@ -53,7 +53,7 @@ class SupportJsTest extends TestCase
 
             public $bar = 'not world';
 
-            public function jsonSerialize()
+            public function jsonSerialize(): array
             {
                 return ['foo' => 'hello', 'bar' => 'world'];
             }
@@ -85,7 +85,7 @@ class SupportJsTest extends TestCase
                 return json_encode(['foo' => 'hello', 'bar' => 'world'], $options);
             }
 
-            public function jsonSerialize()
+            public function jsonSerialize(): array
             {
                 return ['foo' => 'not hello', 'bar' => 'not world'];
             }


taylorotwell:
When you say that currently it is not cleaned up until all tests are done. Do you mean it is cleaned up after each test class is finished or not until the entire suite is done running?

donnysim:
Test suite. The previous handlers will be kept as long as the PHP process is alive and increase with each test, or we could say, every application instantiation that executes the HandleException@bootstrap. This change makes sure that PHP doesn't hold it in the error, exception or in shutdown handlers stack, preventing them from being GC.

donnysim:
Failing PHP 8.1 tests are not related to my change but can fix them too if need be.

driesvints:
@donnysim I think these failures are on your branch since they're not happening on 8.x so can you rebase maybe?

donnysim:
@driesvints don't really get it, the 8.x branch does have this issue in tests even locally running with 8.1 https://github.com/laravel/framework/blob/8.x/tests/Support/SupportJsTest.php#L56 

```
Testing started at 5:40 PM ...
PHPUnit 9.5.13 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.1.1
Configuration: \phpunit.xml.dist


Fatal error: During inheritance of JsonSerializable: Uncaught Return type of JsonSerializable@anonymous::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

\tests\Support\SupportJsTest.php:56
\tests\Support\SupportJsTest.php:50
 in \tests\Support\SupportJsTest.php on line 50

Process finished with exit code 255
```

It doesn't matter if I run it with or without the change, I still get this error.

X-Coder264:
I think something was wrong with the error/deprecation handler prior to this PR and that this PR exposed that bug. The CI pipeline should be failing on the 8.x branch too (which currently does not happen) as there's no return type hint nor the return type will change annotation on that `jsonSerialize` method. You can add the `array` return typehint to those methods to fix this deprecation.

donnysim:
@X-Coder264 that would make sense, all other `jsonSerialize` definitions have return types defined except the ones that fail here.

X-Coder264:
@driesvints Yeah, they should fail on 8.x but they don't which is weird. After fixing the return type new deprecation exceptions popped up on the latest pipeline:

```
There were 3 errors:

1) Illuminate\Tests\Integration\Auth\ForgotPasswordTest::it_can_send_forgot_password_email
strncmp(): Passing null to parameter #1 ($string1) of type string is deprecated

/home/runner/work/framework/framework/src/Illuminate/Support/Str.php:818
/home/runner/work/framework/framework/src/Illuminate/Auth/Passwords/PasswordBrokerManager.php:87
/home/runner/work/framework/framework/src/Illuminate/Auth/Passwords/PasswordBrokerManager.php:72
/home/runner/work/framework/framework/src/Illuminate/Auth/Passwords/PasswordBrokerManager.php:49
/home/runner/work/framework/framework/src/Illuminate/Support/Facades/Facade.php:261
/home/runner/work/framework/framework/tests/Integration/Auth/ForgotPasswordTest.php:45

2) Illuminate\Tests\Integration\Auth\ForgotPasswordTest::it_can_send_forgot_password_email_via_create_url_using
strncmp(): Passing null to parameter #1 ($string1) of type string is deprecated

/home/runner/work/framework/framework/src/Illuminate/Support/Str.php:818
/home/runner/work/framework/framework/src/Illuminate/Auth/Passwords/PasswordBrokerManager.php:87
/home/runner/work/framework/framework/src/Illuminate/Auth/Passwords/PasswordBrokerManager.php:72
/home/runner/work/framework/framework/src/Illuminate/Auth/Passwords/PasswordBrokerManager.php:49
/home/runner/work/framework/framework/src/Illuminate/Support/Facades/Facade.php:261
/home/runner/work/framework/framework/tests/Integration/Auth/ForgotPasswordTest.php:73

3) Illuminate\Tests\Integration\Auth\ForgotPasswordTest::it_can_send_forgot_password_email_via_to_mail_using
strncmp(): Passing null to parameter #1 ($string1) of type string is deprecated

/home/runner/work/framework/framework/src/Illuminate/Support/Str.php:818
/home/runner/work/framework/framework/src/Illuminate/Auth/Passwords/PasswordBrokerManager.php:87
/home/runner/work/framework/framework/src/Illuminate/Auth/Passwords/PasswordBrokerManager.php:72
/home/runner/work/framework/framework/src/Illuminate/Auth/Passwords/PasswordBrokerManager.php:49
/home/runner/work/framework/framework/src/Illuminate/Support/Facades/Facade.php:261
/home/runner/work/framework/framework/tests/Integration/Auth/ForgotPasswordTest.php:105

--
```

donnysim:
@driesvints it has something to do with using Orchestra\Testbench\TestCase, as soon as this is used, deprecation warnings are gone.

donnysim:
@nunomaduro it seems like this https://github.com/laravel/framework/commit/52e2376203020e8ae8053c625c0ef5e17521e36f commit introduced a bug where the deprecation errors are swallowed because no LogManager is registered when used with `Orchestra\Testbench\TestCase`. When running plain `PHPUnit\Framework\TestCase` the application is not instantiated and HandleExceptions never registered so a pure PHP deprecation is thrown. Also, because of the nature where the handler is never unregistered, it will always swallow the deprecation exception for further tests. Even with current pull request, the last registered HandleExceptions will be used until process exits.

driesvints:
Odd because I sent in a PR recently to explicitly surface deprecation errors in CI: https://github.com/laravel/framework/pull/40324 so I'm not sure why these were being repressed. I pinged @nunomaduro so hopefully he can shed some light here.

driesvints:
Placing this in draft until @nunomaduro has had a chance to reply and we've resolved all issues.

donnysim:
The main problem is that the handler just exists silently - null/void handler result is held as user handled event. If it was supposed to fallback to the default PHP exception handling in case of the deprecation error it should return false or throw an error itself:

```php
if (! class_exists(LogManager::class)) {
    return false;
}

try {
    $logger = $this->app->make(LogManager::class);
} catch (Exception $e) {
    return false;
}
```

otherwise no matter the `error_reporting` level, it will never reach it. What I'm afraid is that this change might become a breaking change where suddenly deprecated errors are thrown in use applications, depending if the LogManager exists.

X-Coder264:
@driesvints @nunomaduro  I took a quick look at the state on the 8.x branch.

After https://github.com/laravel/framework/pull/39219 all deprecations for integration tests were silenced (both for user apps as well as for Laravel's own test suite).

`withoutDeprecationHandling` was added in https://github.com/laravel/framework/pull/39261 but it was never applied to all existing integration tests. If those tests do not explicitly call `withoutDeprecationHandling` the new Laravel error handler will only log those errors, but since no exception is thrown the tests do not fail. The entire `InteractsWithDeprecationHandling` trait should be first added to the `Orchestra\Testbench\TestCase` class in order to be able to call that method.

So that solves the issue of why some deprecations were left unnoticed in integration tests (you can confirm this is the cause of the problem by calling `withoutDeprecationHandling` on 8.x branch code in the `Illuminate\Tests\Integration\Auth\ForgotPasswordTest::it_can_send_forgot_password_email` test and seeing that the test is red as opposed to being green before adding that method call).

Now onto why some deprecations were not noticed in unit tests (like the one in `SupportJsTest` which was fixed in this PR) - the issue here is the order of execution.

PHPUnit is properly configured to convert deprecations to exceptions which happens in the PHPUnit ErrorHandler which is registered on the test run, in order for that to happen the PHPUnit error handler needs to be registered, but PHPUnit does not register its error handler if a previous error handler was already configured -> https://github.com/sebastianbergmann/phpunit/blob/9.5.13/src/Util/ErrorHandler.php#L131-L146

This means that if an integration test runs before an unit test it sets the Laravel error handler and then when the unit test runs PHPUnit does not register its own error handler and because Laravel's error handler does not throw an exception anymore the test turns out green instead of red.

The fix here would be to always clear the error handler to the built-in PHP handler after the request is handled.

X-Coder264:
Alternate fix which should resolve all problems -> https://github.com/laravel/framework/pull/40639

@driesvints @nunomaduro Any feedback on that PR is greatly appreciated.

@donnysim Can you also please confirm that it solves your memory leak problem? Thank you.

donnysim:
@X-Coder264 @driesvints with one small change the new pull is looking good. Closing this in favor of [#40639](https://github.com/laravel/framework/pull/40639).

PR #40595, opened 2022-01-24
[9.x] JSON Session Option

taylorotwell:
This pull requests allows developers to choose to serialize session data in storage using PHP or JSON. Over the years, several people have expressed concerns with deserializing sessions using PHP's `unserialize` function, as this could lead to security problems (including remote code execution) if the application's encryption key is leaked.

However, all known ways to exploit this require the attacker to know the application's encryption key, which if known, could already be used to craft authenticated cookies for the application, etc. Nevertheless, offering this an option to those who would prefer to use JSON to serialize sessions in storage.

### DIFF
--- a/src/Illuminate/Session/SessionManager.php
+++ b/src/Illuminate/Session/SessionManager.php
@@ -186,7 +186,12 @@ class SessionManager extends Manager
     {
         return $this->config->get('session.encrypt')
                 ? $this->buildEncryptedSession($handler)
-                : new Store($this->config->get('session.cookie'), $handler);
+                : new Store(
+                    $this->config->get('session.cookie'),
+                    $handler,
+                    $id = null,
+                    $this->config->get('session.serialization', 'php')
+                );
     }
 
     /**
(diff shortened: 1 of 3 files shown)


Krisell:
This is very interesting, both from a security perspective and since it makes the payload around 10–15 % smaller (measured both on a clean install and on a larger project).

I am running an app which uses the cookie driver and I am probably going to switch to the `json` setting if/when merged. However I am going to have to do the migration transparently, meaning that no session data should be lost. That is achievable by overriding `readFromHandler()` to fallback to `unserialize` in case `json_decode` returns null. If/when this is merged and I can successfully verify that transparent migration, I might publish a tiny package, or just a gist, for that code (which only needs to run > one session lifetime). It _could_ of course also be solved with a `php-to-json` setting in the core, but that is perhaps unnecessary for the narrow use case.

src/Illuminate/Session/SessionManager.php
```diff
@@ -186,7 +186,12 @@ protected function buildSession($handler)
     {
         return $this->config->get('session.encrypt')
                 ? $this->buildEncryptedSession($handler)
-                : new Store($this->config->get('session.cookie'), $handler);
+                : new Store(
+                    $this->config->get('session.cookie'),
+                    $handler,
+                    $id = null,
```
Krisell, line 192: Is this assignment just to clarify the parameter name? If so, what about named parameters instead?

PR #38117, opened 2021-07-23
[9.x] TrimString can now remove NBSP

allowing:
126768557-024bada4-0082-48cb-a553-c7f0b70742b1

PR #40596, opened 2022-01-25
Handle unicode characters on TrimStrings middleware

### LINKED ISSUE #40577: Multi-byte word problem with TrimStrings Middleware.
- Laravel Version: 9.0.0-beta.3
- PHP Version: 8.1.0

#40593 # Description:
In 9.x, the trimString middleware removes NBSP. #38117
This is causing the problem. When dealing with multi-byte words (in my case Japanese), we have some problems. A few words are garbled.
When I put `だ` or `ム` in the end of text, those words will be garbled.

## Steps to Reproduce (quick version)

Add below to the welcome.blade.php.
```php
<h1>Hello {{ request()->name }}</h1>
```

Access the url.
http://localhost/?name=やま
http://localhost/?name=やまだ

(Replace `localhost` to your domain. やま or やまだ may be encoded in the URL field of your browser.）

You can see やま is ok. But if you add だ, it's not working.

2022-01-25_09h46_35

## Steps To Reproduce: (Original version)
[Caution]
You cannot just copy and paste the below. NBSP is replaced with the normal space.
Please use NBSP as the second argument of the trim function.
You can copy NBSP from the real source code. (Please don't copy from the github website.)

```php
$str = 'あいうえおかきくけこさしすせそたちつてとなにぬねのはひふへほまみむめもやゆよらをわがぎぐげござしずぜぞだぢづでどばびぶべぼぱぴぷぺぽ'
    . 'アイウエオカキクケコサシスセソタチツテトナニヌネノハヒフヘホマミムメモヤユヨラヲワガギグゲゴザジズゼゾダヂヅデドバビブベボパピプペポ';

$words = preg_split('//u', $str); // split a string into each words. (This part is not related with the problem)

foreach ($words as $word) {
    echo $word.' '.bin2hex($word).' '.trim($word, " ")."<br>";
}
```
2022-01-25_09h16_51

I guess the reason is that `だ` is like `e381a0` and `ム` is like `e383a0` and the NBSP is like U+00A0 in Unicode.
So If I put `だ` in the end of text, the last part of word `a0` is trimmed and the word is garbled.
We (Japanese) also use chinese characters which I didn't looked into.

2022-01-24_15h06_27

2022-01-24 15 37 02 trim719fac98c688

Thank you for reading.

<!-- laravel new bug-report --github="--public" --

>

rodrigopedra:
PR #38117 introduced the ability to other space unicode characters to the `TrimStrings` middleware.

As issue #40577 stated the current implementation conflicts with some other unicode characters which end with the same escape sequence as the NBSP character.

In particular these japanese characters:

- だ (unicode `E381A0`)
- ム (unicode `E383A0`)

The implementation from PR #38117 causes the `A0` sequence to be trimmed, which converts the sequence to an invalid one.

This PR:

- Changes the implementation to use `preg_replace` to trim the spaces using the proper modifiers
- Add additional test cases that failed with the previous implementation and passed with the proposed one

Closes #40577


rodrigopedra:
sending to 9.x

PR #40605, opened 2022-01-25
[8.x] arrayable enum support

makroxyz:
This PR allows enum classes implementing Arrayable interface to be casted to array

### Why
We have an `enum` class that implements `Arrayable` interface and we write `toArray()` method in order to get an array when casted to array.
In a `Model` we cast a column to the above `enum`.
When we call `Model::toArray()` enum is casted to string instead the expected array.

### Example

Enum:
```php
enum ArrayableStatus: string implements Arrayable
{
    case pending = 'pending';
    case done = 'done';

    public function description(): string
    {
        return match ($this) {
            self::pending => 'pending status description',
            self::done => 'done status description'
        };
    }

    public function toArray()
    {
        return [
            'name' => $this->name,
            'value' => $this->value,
            'description' => $this->description(),
        ];
    }
}
```

Model:
```php
class MyModel extends Model
{
    public $casts = [
        'arrayable_status' => ArrayableStatus::class,
    ];
}
```

#### What I get:
```php
MyModel::first()->toArray()
/**
[
   'id' => 2,
   ....
   'arrayable_status' => 'pending'
]
*/
```

#### What I expect:
```php
MyModel::first()->toArray()
/**
[
   'id' => 2,
   ....
   'arrayable_status' => [
       'name' => 'pending',
       'value' => 'pending',
       'description' => 'pending status description',
   ]
]
*/

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -296,7 +296,7 @@ trait HasAttributes
                 $attributes[$key] = $this->serializeClassCastableAttribute($key, $attributes[$key]);
             }
 
-            if ($this->isEnumCastable($key)) {
+            if ($this->isEnumCastable($key) && (! $attributes[$key] instanceof Arrayable)) {
                 $attributes[$key] = isset($attributes[$key]) ? $attributes[$key]->value : null;
             }
 
--- a/tests/Integration/Database/EloquentModelEnumCastingTest.php
+++ b/tests/Integration/Database/EloquentModelEnumCastingTest.php
@@ -70,7 +70,7 @@ class EloquentModelEnumCastingTest extends DatabaseTestCase
             'arrayable_status' => [
                 'name' => 'pending',
                 'value' => 'pending',
-                'description' => 'pending status description'
+                'description' => 'pending status description',
             ],
         ], $model->toArray());
     }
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by driesvints** [comment]
https://github.com/laravel/framework/pull/40584 got closed because of a lack of explanation. Please add a **thorough** explanation to your initial PR description.

**Comment by taylorotwell** [comment]
I dunno - I'm a little concerned about the breaking change here on a patch release. Instead, you can add accessor attributes for this extra stuff and append them to the model array output using the "appends" feature. Alternatively, you could just write your own custom cast class that handles Enums in this way for you?

**Comment by diegotibi** [comment]
@taylorotwell sorry but I don't see where the breaking change is. Any Enum non implementing Arrayable will simply be cast as usual and I'm pretty sure that anyone that implements arrayable on a class is expecting that his class will be casted as in the toArray() method.

Now this DONT happens cause:

```php
            //isEnumCastable always return true if the object is an enum...
            if ($this->isEnumCastable($key)) {
                $attributes[$key] = isset($attributes[$key]) ? $attributes[$key]->value : null;
            }
            //...so the toArray() method will never be call on enums, even when Enum is instance of Arrayable
            if ($attributes[$key] instanceof Arrayable) {
                $attributes[$key] = $attributes[$key]->toArray();
            }
```

makroxyz:
@taylorotwell default enum cast works fine, I don't need to override. What you suggest is to append another attribute which is an accessor that calls `$this->arrayable_enum->toArray()` but if you look at @diegotibi post this is what I expect implementing Arrayable in a class (in this case enum class)

PR #37943, opened 2021-07-07
[8.x] Add Validator::excludeArrays() to exclude array keys that aren't included in the validation rules

themsaid:
Given the following rules:

```php
return request()->validate([
    'users' => 'required|array',
    'users.*.name' => 'required',
]);
```

The validator will return the `users` attribute with all its children:

```
array:1 [▼
  "users" => array:1 [▼
    1 => array:2 [▼
      "name" => "mohamed"
      "job" => "dev"
    ]
  ]
]
```

You can see that the `job` key was returned even though it wasn't included in the validator rules.

This PR adds a new `Validator::excludeArrays()` that when added to a `boot` method of a service provider, will instruct the validator to exclude any array keys that weren't validated:

```
array:1 [▼
  "users" => array:1 [▼
    1 => array:1 [▼
      "name" => "mohamed"
    ]
  ]
]
```

PR #40611, opened 2022-01-25
[8.x] Document `excludeUnvalidatedArrayKeys()` on `Validator` facade

devfrey:
`excludeUnvalidatedArrayKeys()` was added in #37943, but the facade doesn't document this method. I've added this method to the facade's PHPDoc to help with IDE autocompletion and static analysis.


devfrey:
The check failure seems unrelated.

driesvints:
@devfrey retriggered those for ya.

PR #40619, opened 2022-01-25
[9.x] No need to use Str helpers

lucasmichot:
No need to use helpers instead of the native `str_contains`, `str_starts_with` and `str_ends_with`  functions.

### DIFF
--- a/src/Illuminate/Auth/Access/Gate.php
+++ b/src/Illuminate/Auth/Access/Gate.php
@@ -239,7 +239,7 @@ class Gate implements GateContract
     protected function buildAbilityCallback($ability, $callback)
     {
         return function () use ($ability, $callback) {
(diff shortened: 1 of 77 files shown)


GrahamCampbell:
We need to be careful. These functions are not identical, in particular in the null or empty string cases.

lucasmichot:
> We need to be careful. These functions are not identical, in particular in the null or empty string cases.

This has been taken in consideration, same for the string as a 2nd argument, and the absence of `$ignoreCase` argument

taylorotwell:
@lucasmichot can you fix the conflicts here?

lucasmichot:
> @lucasmichot can you fix the conflicts here?

Conflicts fixed @taylorotwell 👌

PR #40582, opened 2022-01-24
[8.x] Add test coverage for Mail::alwaysTo() and Mail::alwaysReplyTo()

clemblanco:
Just adding some test coverage.

Before
```
Mail Mailer (Illuminate\Tests\Mail\MailMailer)
 ✔ Global from is respected on all messages  2 ms
 ✔ Global return path is respected on all messages  2 ms
```

After
```
Mail Mailer (Illuminate\Tests\Mail\MailMailer)
 ✔ Global from is respected on all messages  2 ms
 ✔ Global reply to is respected on all messages  1 ms
 ✔ Global to is respected on all messages  1 ms
 ✔ Global return path is respected on all messages  2 ms
```

Linked to [laravel/docs #7620](https://github.com/laravel/docs/pull/7620) and #40561

PR #40623, opened 2022-01-25
[9.x] Fix broken tests and clearing cc and bcc

driesvints:
This PR adjusts the tests that were added in https://github.com/laravel/framework/pull/40582 for 9.x and additionally fix a bug that was uncovered by them. We're now clearing CC and BCC by clearing all addresses from the respective message headers.


GrahamCampbell:
Noice. 🎉

PR #40629, opened 2022-01-25
[8.x] Add ability to generate URL's with query strings using `url()->query()`

stevebauman:
## Description

This PR adds the ability to generate absolute URL's with query strings using `url()->query()`.

## Purpose

Currently, there is no built-in way of generating an absolute URL with query string parameters. You will have to assemble it yourself. It is currently possible using the `route()` and `action()` helpers, but not with `url()`.

**Before**:
```php
$query = Arr::query(['brand' => 'Samsung', 'model' => 'Galaxy']);

// http://localhost/products?brand=Samsung&model=Galaxy
url()->to('/products?'.$query);
```

**After**:

```php
// http://localhost/products?brand=Samsung&model=Galaxy
url()->query('/products', ['brand' => 'Samsung', 'model' => 'Galaxy']));
```

---

**Extra Parameters**:

You can also add extra parameters as you would with `url()->to()`:

```php
// http://www.localhost.com/products/samsung?model=Galaxy
url()->query('/products', ['model' => 'Galaxy'], ['samsung']);
```

**Secure URL**:

As with `url()->to()`, we may also provide a `$secure` parameter to generate an HTTPS URL:

```php
// https://www.localhost.com/products/samsung?model=Galaxy
url()->query('/products', ['model' => 'Galaxy'], ['samsung'], $secure = true);
```

**Developer Shortcut**:

We may also provide a boolean into the `$extra` parameter argument as a shortcut to generate a secure URL:

```php
// https://www.localhost.com/products?model=Galaxy
url()->query('/products', ['model' => 'Galaxy'], $secure = true)
``` 

---

Thanks for your time! ❤️  Let me know your thoughts 🙏

### DIFF
--- a/src/Illuminate/Routing/GeneratesQueryStrings.php
+++ b/src/Illuminate/Routing/GeneratesQueryStrings.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace Illuminate\Routing;
+
+use Illuminate\Support\Arr;
+
+trait GeneratesQueryStrings
+{
+    /**
+     * Generate a query string with the given parameters.
+     *
+     * @param  array  $parameters
+     * @return string
+     */
+    protected function makeQueryString(array $parameters)
+    {
+        if (count($parameters) === 0) {
+            return '';
+        }
+
+        $query = Arr::query(
+            $keyed = $this->getStringParameters($parameters)
+        );
+
+        // If there happens to be more parameters remaining, we will fetch the numeric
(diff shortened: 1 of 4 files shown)


taylorotwell:
I dunno - tbh the before doesn't seem that bad to me? 😅 It looks a little worse because you have split it across two lines.

stevebauman:
Ok no worries! 😄

PR #40637, opened 2022-01-25
[9.x] Add support for nested having in queries

krve:
## Description

This PR implements nested having queries for the query builder, in the same way that wheres work. This means you can now group together "having" queries like the example below. Currently you are forced to make the query using `havingRaw` which could potentially introduce the possibility for SQL injections.

This shouldn't have an impact on existing applications as closure's where never supported in the first place.

## Example
```php
DB::table('product_data')
    ->having(function (Builder $query) {
        $query->having('column_one', 'foo')
            ->having('column_two', 'bar');
    })
    ->orHaving('column_three', 'biz')
    ->get();
```

Translates to: `select * from "product_data" having ("column_one" = "foo" and "column_two" = "bar") or "column_three" = "biz"`

This follows the current way of creating nested wheres.
```php
DB::table('product_data')
    ->where(function (Builder $query) {
        $query->where('column_one', 'foo')
            ->where('column_two', 'bar');
    })
    ->orWhere('column_three', 'biz')
    ->get();
```

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1549,6 +1549,33 @@ class Builder
         return $this;
     }
 
+    public function havingNested(Closure $callback, $boolean = 'and')
+    {
+        call_user_func($callback, $query = $this->forNestedWhere());
+
+        return $this->addNestedHavingQuery($query, $boolean);
+    }
+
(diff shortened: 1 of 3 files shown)


krve:
Wrong branch, closing in favor of #40638

PR #40639, opened 2022-01-25
[8.x] Fix error handling

### LINKED ISSUE #40554: Memory leak in tests because HandleExceptions never cleans up

X-Coder264:
Fixes problems described in https://github.com/laravel/framework/pull/40592#issuecomment-1021554516

TLDR; Laravel never cleaned up its handlers, resulting in the PHPUnit error handler never firing for unit tests that came after Laravel's integration tests, plus since the handlers were never cleaned it resulted in a memory leak.

In order to avoid having to call `$this->withoutDeprecationHandling()` in every Laravel integration test I've added that call in the `protected function deprecationHandling()` hook method which is now called from the base test class `setUp` method (users can override it in their base class if they want to change the behavior).

The changes made in the base test class here should also be ported over to Testbench, after which I can fix all the deprecation exceptions that will pop up on CI, before this PR can be merged.

Also note that this PR makes use of a terminating callback to restore the handlers, which also meant having to manually call `$this->app->terminate()` in the test teardown (it's weird that this was not already called as `terminate` is always called in a real HTTP request).

Fixes https://github.com/laravel/framework/issues/40554

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDeprecationHandling.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDeprecationHandling.php
@@ -21,9 +21,17 @@ trait InteractsWithDeprecationHandling
     protected function withDeprecationHandling()
     {
         if ($this->originalDeprecationHandler) {
-            set_error_handler(tap($this->originalDeprecationHandler, function () {
+            $previousHandler = set_error_handler(tap($this->originalDeprecationHandler, function () {
                 $this->originalDeprecationHandler = null;
             }));
+
+            if (method_exists($this, 'beforeApplicationDestroyed')) {
+                $this->beforeApplicationDestroyed(function () use ($previousHandler) {
+                    if (null !== $previousHandler) {
+                        restore_error_handler();
+                    }
+                });
+            }
         }
 
         return $this;
@@ -42,6 +50,14 @@ trait InteractsWithDeprecationHandling
                     throw new ErrorException($message, 0, $level, $file, $line);
                 }
             });
+
+            if (method_exists($this, 'beforeApplicationDestroyed')) {
+                $this->beforeApplicationDestroyed(function () {
+                    if (null !== $this->originalDeprecationHandler) {
+                        restore_error_handler();
+                    }
+                });
+            }
         }
 
         return $this;
--- a/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDeprecationHandling.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDeprecationHandling.php
@@ -29,6 +29,7 @@ trait InteractsWithDeprecationHandling
                 $this->beforeApplicationDestroyed(function () use ($previousHandler) {
                     if (null !== $previousHandler) {
                         restore_error_handler();
+                        $this->originalDeprecationHandler = null;
                     }
                 });
             }
@@ -55,6 +56,7 @@ trait InteractsWithDeprecationHandling
                 $this->beforeApplicationDestroyed(function () {
                     if (null !== $this->originalDeprecationHandler) {
                         restore_error_handler();
+                        $this->originalDeprecationHandler = null;
                     }
                 });
             }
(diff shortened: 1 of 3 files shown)


src/Illuminate/Foundation/Testing/Concerns/InteractsWithDeprecationHandling.php
```diff
@@ -42,6 +50,14 @@ protected function withoutDeprecationHandling()
                     throw new ErrorException($message, 0, $level, $file, $line);
                 }
             });
+
+            if (method_exists($this, 'beforeApplicationDestroyed')) {
+                $this->beforeApplicationDestroyed(function () {
+                    if (null !== $this->originalDeprecationHandler) {
+                        restore_error_handler();
```
donnysim, line 57: We need to reset `$this->originalDeprecationHandler = null;` otherwise the memory leak persists because PHPUnit TestSuiteIterator creates a new class instance for each test method in the class and keeps everything in memory until test suite is done. After this I think it's looking good.

donnysim:
There might still be something wrong, the deprecation errors are still not popping up in full test runs.

src/Illuminate/Foundation/Testing/Concerns/InteractsWithDeprecationHandling.php
X-Coder264: Done :+1:

X-Coder264:
Yeah, the Testbench changes are needed before the deprecation exceptions pop up on CI.

Thank you for testing.

nunomaduro:
Converting this to draft while I try to understand the reasoning behind this pull request.

nunomaduro:
@X-Coder264 @donnysim Let me know what do you think about this alternative: https://github.com/laravel/framework/pull/40656.

PR #40643, opened 2022-01-26
[8.x] Get item of array as int and float.

bdaler:
Often in projects I see: `(int) Arr::get($item, 'brand.id');` or `(float) Arr::get($item, 'discountTotal')` I suggest doing this instead `Arr::getInt($item, 'brand.id')` or `Arr::getFloat($item, 'discountTotal')`

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -744,4 +744,32 @@ class Arr
 
         return is_array($value) ? $value : [$value];
     }
+
+    /**
+     * Returns the item value converted to integer.
+     *
+     * @param      $array
+     * @param      $key
+     * @param null $default
+     *
+     * @return int
+     */
+    public static function getInt($array, $key, $default = null): int
+    {
+        return (int) self::get($array, $key, $default);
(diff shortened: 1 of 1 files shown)


src/Illuminate/Collections/Arr.php
```diff
@@ -744,4 +744,30 @@ public static function wrap($value)
...
+
+    /**
+     * Returns the item value converted to integer.
+     *
+     * @param  \ArrayAccess|array  $array
+     * @param  string|int|null  $key
+     * @param  mixed  $default
+     * @return int
+     */
+    public static function getInt($array, $key, $default = null): int
```
koenhoeijmakers, line 756: I'd expect `$default` to be returned when the array key doesn't exist, yet it will return `0` instead of `null`

X-Coder264: ` $default = null` should IMO be ` int $default = 0` as the whole point of this method is to always return an integer so allowing the `mixed` type for `$default` does not make sense IMO.

taylorotwell:

PR #40644, opened 2022-01-26
[8.x] Fix Not override actionList

fagai:
If you define the same action for multiple routes, I think the route you created earlier will take precedence.
The reality is different: the non-cached RouteCollection overwrites the ActionList key, giving priority to routes added later.
This fix corrects the differences with the CompiledRouteCollection.

### DIFF
--- a/src/Illuminate/Routing/RouteCollection.php
+++ b/src/Illuminate/Routing/RouteCollection.php
@@ -102,7 +102,11 @@ class RouteCollection extends AbstractRouteCollection
      */
     protected function addToActionList($action, $route)
     {
-        $this->actionList[trim($action['controller'], '\\')] = $route;
+        $key = trim($action['controller'], '\\');
+
+        if (! isset($this->actionList[$key])) {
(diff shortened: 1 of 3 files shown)


taylorotwell:

Thanks!

fagai:
@taylorotwell 

Thank you for your reply.
Here is a simple example.
Please let me know if there is something missing.
I will do everything I can.

# Example

web.php
```php
Route::resource('products', 'ProductController');
Route::get('products.index.csv', 'ProductController@index');
```

ProductController
```php
public function index()
{
    // search products

    if (request()->extension() === 'csv') {
        // If the extension is csv, change it to csv and download it.
    }
}
```

view (show)
```php
<a href="{{ action(controller_name(), 'index') }}">index</a>
```

For codes like the one above, the following differences occur:
## no cache
```php
<a href="http://localhost/products.index.csv">index</a>
```

## cached (`php artisan route:cache`)
```php
<a href="http://localhost/products">index</a>
```

PR #40649, opened 2022-01-26
[9.x] Support AWS SES with IAM Assume Role

deleugpn:
On PR https://github.com/laravel/framework/pull/38481, the file https://github.com/laravel/framework/blob/3234a8dfecbb12140e1ce980e415fcd0363f320a/tests/Mail/MailSesTransportTest.php was removed and with it 3 features were lost:

- Ability to let SES Assume Role through AWS SDK
- Ability to configure SES ConfigurationSet
- Ability to define Email tags

This PR addresses point 1. Upon closer look at item 2), it looks quite complex to get that functionality back because it must be configured at Mail Sending time (see https://github.com/symfony/symfony/pull/37897) and during the Mailable build we would end up needing some sort of `if (... instanceof Ses)`.

Item 3) seem to not even be supported by Symfony and is not clearly documented by AsyncAws (see https://async-aws.com/clients/ses.html).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -143,6 +143,9 @@ jobs:
 
       - name: Execute tests
         run: vendor/bin/phpunit --verbose
+        env:
+          AWS_ACCESS_KEY_ID: random_key
+          AWS_SECRET_ACCESS_KEY: random_secret
 
       - name: Store artifacts
         uses: actions/upload-artifact@v2
--- a/composer.json
+++ b/composer.json
@@ -92,8 +92,8 @@
         "phpstan/phpstan": "^1.0",
         "predis/predis": "^1.1.9",
         "phpunit/phpunit": "^9.5.8",
-        "symfony/cache": "^6.0",
-        "symfony/amazon-mailer": "^6.0"
+        "symfony/amazon-mailer": "^6.0",
+        "symfony/cache": "^6.0"
     },
     "provide": {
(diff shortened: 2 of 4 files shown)


deleugpn:
Looks like I can either get it to work on Windows or Linux, but not both 👀

Jubeki:
It seems like you are right. Symfony only supports settings these options via the Header:
https://github.com/symfony/amazon-mailer/blob/99a9a86ae5d3475300b63834c91175bf29a7f19d/Transport/SesApiAsyncAwsTransport.php#L92-L94
And the ConfigurationSetName and Tags are not set while creating via the Factory.

There are four possible solutions going forward:
1. Let the Developer set them using Tags (not ideal)
2. Make a PR to `symfony/amazon-mailer` to support it with the DSN-Object (not sure if this goes through)
3. Make a custom SesTransport which includes the support.
4. Create a callback which will be called for all E-Mails which are send using SES and set the correct tags.

@driesvints @taylorotwell what are your opinions on that?

driesvints:
I'll have a look at this today. Thanks

driesvints (approved):

driesvints:
This can already be merged to address 1. and I'll have a look at the other two. Thanks @deleugpn!

driesvints:
I took a new approach by re-introducing the old transport. @deleugpn can you try it out? https://github.com/laravel/framework/pull/40696

PR #40654, opened 2022-01-26
[8.x] support for closure count in Eloquent factory

timrspratt:
I have the following scenario where I wish to seed (using factories) several posts each with a varying comment count (via a comment relationship).

```php
Post::factory()->count(5)->has(Comment::factory()->count(3))->create();
```

The above would create 5 posts, each with 3 comments.

This pull requests means that for each of the 5 posts, a closure is used to determine the count for that particular iteration:

```php
Post::factory()->count(5)->has(Comment::factory()->count(fn() => rand(1, 9)))->create();
```

This may be something for 9.x?

Thanks

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/Factory.php
+++ b/src/Illuminate/Database/Eloquent/Factories/Factory.php
@@ -31,7 +31,7 @@ abstract class Factory
     /**
      * The number of models that should be generated.
      *
-     * @var int|null
+     * @var int|\Closure|null
      */
     protected $count;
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you send this to `9.x` branch?

igorgaming:
@timrspratt Hey! Is there any reason why you didn't send these changes to 9.x? Maybe they would have been merged.

timrspratt:
Hey, no reason other than being super busy and I think a test should accompany this

PR #40660, opened 2022-01-27
Prevent application Requests that extend FormRequest.php from unnecessarily defining authorize method unless overriding with some logic

thomasmokuaomweri2011:
Currently, app Requests extending FormRequest.php often define the `authorize` method even when it explicitly returns `true`;
```
public function authorize(): bool
{
   return true;
}
```
To avoid the compulsion of unnecessarily doing so, the `authorize` method can now be declare in the parent `FormRequest` class so that child classes can override it only when there's compelling logic to do so.
This change will not break things because the child classes are already overriding it

### DIFF
--- a/src/Illuminate/Foundation/Http/FormRequest.php
+++ b/src/Illuminate/Foundation/Http/FormRequest.php
@@ -271,4 +271,14 @@ class FormRequest extends Request implements ValidatesWhenResolved
 
         return $this;
     }
+
+    /**
+     *  Determine if the user is authorized to make this request.
+     *
+     * @return bool
+     */
+    public function authorize()
+    {
+        return true;
+    }
 }


igorgaming:
Actually you dont need to define this method that just return `true` in child Requests.

Look at this code in `FormRequest`: https://github.com/laravel/framework/blob/0f31ca4a11d99e01f22e3807c03e793625f1857b/src/Illuminate/Foundation/Http/FormRequest.php#L170-L179

So, if `authorize` method doesnt exists than it just return `true`.

driesvints:
Thanks but this indeed isn't necessary like @igorgaming explains.

thomasmokuaomweri2011:
Sure, that had escaped me. You are right, thanks

PR #40662, opened 2022-01-27
[8.x] Default schema value accepts Enum

pionl:
👋

In our migration we can now use PHP 8.1 Enum.

**Instead of using**

```php
$table->tinyInteger('x')->default(MyEnum::WAITING->value);
```

**we can use**

```php
$table->tinyInteger('x')->default(MyEnum::WAITING);
```

Also I've added tests to cover other values than Enum. I've not found a method that could tests if variable is enum. I've implemented a different approach. 

**FYI:**

As my first contribution it was little bit challenging to find out:

- can I use PHP 8.1(8.0) syntax in tests? __answer is no - we need to run tests agains 7.3 and lower -> maybe update docs with tests section?__
- how can I setup full tests suite (to run redis / mysql tests), i can see docker-compose.yml but it does not contain PHP runtime.

Probably we can make docker-compose.yml that contains PHP runtime + php-unit.xml that can be ready for to use by anyone with full test suite support. Probably it is possible to run multiple PHP runtimes?

Thanks for everything,
Martin.

### DIFF
--- a/src/Illuminate/Database/Schema/Grammars/Grammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/Grammar.php
@@ -297,6 +297,14 @@ abstract class Grammar extends BaseGrammar
             return $value;
         }
 
+        // Test if the object is "enum" - at this moment we have not found a method to
+        // check if a valid enum
+        if (is_object($value) &&
+            method_exists($value, '__toString') === false
+            && property_exists($value, 'value')) {
+            $value = $value->value;
+        }
+
         return is_bool($value)
                     ? "'".(int) $value."'"
                     : "'".(string) $value."'";
--- a/tests/Database/DatabaseAbstractSchemaGrammarDefaultValueTest.php
+++ b/tests/Database/DatabaseAbstractSchemaGrammarDefaultValueTest.php
@@ -1,0 +1,87 @@
+<?php
+
+namespace Illuminate\Tests\Database;
+
+use Carbon\Carbon;
+use Illuminate\Database\Query\Expression;
+use Illuminate\Database\Schema\Grammars\Grammar;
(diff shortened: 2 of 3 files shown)


driesvints:
I don't think we should encourage people to use ENUM's in migrations. Migrations are moments in time which should have fixed values that don't change anymore after being deployed to production. App code is interchangeable and can change after the migration has already been run. This risks your previous migrations to become out of sync with what has actually been run on production.

pionl:
> I don't think we should encourage people to use ENUM's in migrations. Migrations are moments in time which should have fixed values that don't change anymore after being deployed to production. App code is interchangeable and can change after the migration has already been run. This risks your previous migrations to become out of sync with what has actually been run on production.

Interesting thought. But imho, i think in real usage we always want to use constants etc. If the migration wasn't triggered on production and I will change in a another commit enum value, i do not need to upgrade the migration. Still I should prune the migration in some time to prevent this. I can see it used on new project where the `create` migrations are edited until "first" production ready code.

It was quite a quick think to do so im open to close it if you think it should not be supported. Thanks for your time.

driesvints:
@pionl yeah I think we have a different view on how migrations should be written. Imo the migration should always reflect the values it used at the time of the migration so you can always reference the file to know what was run. Mixing in app code should be avoided imo.

nunomaduro:
This proposal has been discussed already, and we don't plan to add support for Enums on migrations.

pionl:
> This proposal has been discussed already, and we don't plan to add support for Enums on migrations. Thank you for this pull request!

NP. maybe we can remove enum and leave the tests?

pionl:
@nunomaduro I've pushed only the tests if you are interesest. Have a great day and thanks all

PR #40672, opened 2022-01-27
[9.x] Resolving Paths using basePath function.

4m1n0s:
Resolving paths using basePath is more cleaner.


taylorotwell:
If the code works I'm not changing it atm.

PR #40676, opened 2022-01-27
[9.x]: Update doctype on `Cache` facade  and fix `putMany` argument in Cache `Repository` class

dammy001:
- Currently, `putMany` method in the `Repository` class expect `TTL` as an argument but the `value` argument is being passed.
- Updated the `pull` method to retrieve an array of items from the cache and delete it.


dammy001:
@driesvints kindly check this PR.

dammy001:
@driesvints 

kindly check `put` method in Cache/Repository class

Previous
```
if (is_array($key)) {
      return $this->putMany($key, $value);
}
```
updated
```
if (is_array($key)) {
     return $this->putMany($key, $ttl);
}
```

```public function putMany(array $values, $ttl = null)```

driesvints:
the current implementation is the correct one.

PR #40678, opened 2022-01-27
[9.x] Add `orderBy` property for default model sorting

jasonmccreary:
This adds an optional `orderBy` property to Eloquent models which allows the developer to easily add **default sorting** to any model.

**Why**
I believe this is a common enough use case where having something built-in adds to that "comes with everything right out-of-the-box" feel Laravel has.

While this may be done currently, it's a rather cumbersome experience: overwriting the `boot` method, registering a global scope, writing your own logic.

**How**
Instead, you may simply set the `orderBy` property with the column (or columns) and direction (optional) you want your Eloquent queries to be ordered by. For example:

```php
class Post extends Model
{
    public $orderBy = [
        'published_at' => 'DESC',
        'title',
    ];
}
```

**Note:** To avoid this feature being a "foot gun" with the complexity of merging default and runtime orders, this is only applied when `orderBy` was not called in the Eloquent query. Calling `orderBy` in an Eloquent query assumes the developer wants full control over the sorting.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -602,7 +602,7 @@ class Builder implements BuilderContract
      */
     public function get($columns = ['*'])
     {
-        $builder = $this->applyScopes();
+        $builder = $this->applyOrderBy()->applyScopes();
 
         // If we actually found models we will also eager load any relationships that
         // have been specified as needing to be eager loaded, which will solve the
@@ -747,7 +747,7 @@ class Builder implements BuilderContract
      */
     public function cursor()
     {
-        return $this->applyScopes()->query->cursor()->map(function ($record) {
+        return $this->applyOrderBy()->applyScopes()->query->cursor()->map(function ($record) {
             return $this->newModelInstance()->newFromBuilder($record);
         });
     }
@@ -1133,6 +1133,37 @@ class Builder implements BuilderContract
             $builder = $builder->callNamedScope(
                 $scope, Arr::wrap($parameters)
             );
+        }
+
+        return $builder;
+    }
+
+    /**
+     * Apply the default sorting to the Eloquent builder instance and return it.
+     *
+     * @return static
+     */
+    public function applyOrderBy()
+    {
+        if (! empty($this->orders)) {
+            return $this;
+        }
+
+        if (empty($this->model->orderBy)) {
+            return $this;
+        }
+
+        $builder = clone $this;
+
+        foreach ($builder->model->orderBy as $column => $direction) {
+            // If the column is an integer, the user has not defined a key/value pair
+            // of column/direction. Treat direction as the column name using the
+            // default ascending sort to apply to the new builder instance.
+            if (is_int($column)) {
+                [$column, $direction] = [$direction, 'asc'];
+            }
+
+            $builder->orderBy($column, $direction);
         }
 
         return $builder;
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -84,6 +84,13 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     protected $withCount = [];
 
     /**
+     * The columns to order by when none were specified in the query.
+     *
+     * @var string[]|null
+     */
+    public $orderBy;
+
+    /**
      * Indicates whether lazy loading will be prevented on this model.
      *
      * @var bool
--- a/tests/Integration/Database/EloquentModelHasOrderTest.php
+++ b/tests/Integration/Database/EloquentModelHasOrderTest.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace Illuminate\Tests\Integration\Database;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class EloquentModelHasOrderTest extends DatabaseTestCase
+{
+    protected function defineDatabaseMigrationsAfterDatabaseRefreshed()
+    {
+        Schema::create('posts', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('title');
+            $table->string('slug');
+        });
+    }
+
+    public function testUsesDefaultOrderByWhenNoOrderHasNotBeenSet()
+    {
+        $post1 = PostOrdered::create(['title' => 'Post 1', 'slug' => 'post-1']);
+        $post2 = PostOrdered::create(['title' => 'Post 2', 'slug' => 'post-2']);
+        $post3 = PostOrdered::create(['title' => 'Post 2', 'slug' => 'post-22']);
+
+        $posts = PostOrdered::all();
+
+        $this->assertCount(3, $posts);
+        $this->assertTrue($posts->get(0)->is($post1));
+        $this->assertTrue($posts->get(1)->is($post3));
+        $this->assertTrue($posts->get(2)->is($post2));
+    }
+
+    public function testItDoesNotUseDefaultOrderByWhenOrderHasBeenSet()
+    {
+        $post1 = PostOrdered::create(['title' => 'Post 1', 'slug' => 'post-1']);
+        $post2 = PostOrdered::create(['title' => 'Post 2', 'slug' => 'post-2']);
+        $post3 = PostOrdered::create(['title' => 'Post 2', 'slug' => 'post-22']);
+
+        $posts = PostOrdered::orderBy('id')->get();
+
+        $this->assertCount(3, $posts);
+        $this->assertTrue($posts->get(0)->is($post1));
+        $this->assertTrue($posts->get(1)->is($post2));
+        $this->assertTrue($posts->get(2)->is($post3));
+    }
+
+    public function testCursorRespectDefaultOrderBy()
+    {
+        $post1 = PostOrdered::create(['title' => 'Post 1', 'slug' => 'post-1']);
+        $post2 = PostOrdered::create(['title' => 'Post 2', 'slug' => 'post-2']);
+        $post3 = PostOrdered::create(['title' => 'Post 2', 'slug' => 'post-22']);
+
+        $this->assertSame([1, 3, 2], PostOrdered::cursor()->pluck('id')->toArray());
+    }
+
+    public function testFirstRespectDefaultOrderBy()
+    {
+        PostOrdered::create(['title' => 'Test Post', 'slug' => 'post-1']);
+        $expected = PostOrdered::create(['title' => 'Test Post', 'slug' => 'post-2']);
+
+        $this->assertTrue(PostOrdered::first()->is($expected));
+    }
+
+    public function testValueRespectsDefaultOrderBy()
+    {
+        PostOrdered::create(['title' => 'Test Post', 'slug' => 'post-1']);
+        PostOrdered::create(['title' => 'Test Post', 'slug' => 'post-2']);
+
+        $this->assertSame('post-2', PostOrdered::value('slug'));
+    }
+}
+
+class PostOrdered extends Model
+{
+    public $table = 'posts';
+    public $timestamps = false;
+    public $orderBy = [
+        'title',
+        'slug' => 'DESC',
+    ];
+    protected $guarded = [];
+}


jonnott:
@jasonmccreary Great stuff. My only reservation is whether an 'always-on' Concern trait like this might be needlessly filling up the global-scopes namespace with global scope names that get applied to every model.

What if this was an opt-in trait, like SoftDeletes?

jonnott:
...i.e.

```
namespace Illuminate\Database\Eloquent;

use DefaultOrderby;

class User extends Model {

protected $orderBy = [ .. ];
}
```

jonnott:
> In fairness, nothing is "needlessly filled". I took care to not add anything if the model does not set the `$orderBy` property. So either solution would apply the same amount of scopes/macros.

Oh yes, `if (empty($this->orderBy)) { return; }` is a nice touch.

jonnott:
> @jonnott, sure. could go that way. The DX of simply setting a property better keeps with the underlying goal - less code. Once you start adding traits, setting properties, defining methods it quickly becomes the same amount of code for the user to write.

I think because none of the other `Concerns/Has*` trait define named global scopes, it feels a bit like by doing so this trait is infringing on something which feels like the 'domain' of the app's developer - i.e. the global-scope-names-namespace. This is in contrast to say the SoftDeletes trait, which uses a global scope *class* instead. What if the 'defaultOrderBy' scope was defined in a class? Am I being too pedantic?

jonnott:
..and where would it live .. `Database/Eloquent/Concerns/Scopes/DefaultOrderBy.php` ?

jasonmccreary:
@jonnott, all fair points. Let's see what the Laravel team has to say. Glad to make tweaks as I feel this is a worthy addition. But I'd like to make sure they think so too first. 😅

jonnott:
> @jonnott, all fair points. Let's see what the Laravel team has to say. Glad to make tweaks as I feel this is a worthy addition. But I'd like to make sure they think so too first. 😅

Fully agree about the worthiness of the addition. I'd use it loads!

But also wanting to be wary of the possibility of:

- Models having too many `Concern`s by default - is that a slippery slope?
- potential developer pain if trying name their own global scopes something which the default Eloquent Model has already 'gazumped' ;) .. if the core framework utilises a global scope name that a developer may have already used in their code .. is that a breaking change?!

Getting towards an almost philosophical level here, I'm aware 🤓  Looks like one for @taylorotwell et al to review the exact implementation detail for maybe..

jonnott:
> @jonnott, indeed, that's why it's targeted for Laravel 9. 👍

I'd be interested to know if the core framework actually uses _any_ named global scopes itself, or if the implementation of all framework-level global scoping is done through classes & FQCNs. I'd not be surprised if this is the case.

jonnott:
@jasonmccreary I branched and added a commit to put the global scope into a class instead .. see what you think @ https://github.com/laravel/framework/compare/9.x...jonnott:model-orderby-property

jasonmccreary:
@jasonmccreary, yeah, more closely follows the `SoftDeletes`. The only reason I didn't do this was the scope is now aware of the property. That felt strange to me. But weighed against the "magic string", it is better.

I also thought about making the scope class generic, so it could be used directly. But then realized there isn't really a strong use case for that. Again, the streamlined DX is ultimately what makes this worthwhile.

taylorotwell:
One concern I might have with this being implemented as a global scope is there are certain times in (maybe in the framework, maybe in packages, etc.) where we explicitly retrieve models with all global scopes disabled in order to be sure we can definitely retrieve them without any soft delete constraints getting in the way, etc. In those cases, this ordering would be removed - which I find... odd?

I think the reason it starts to feel odd is this is not a "scope" at all, where I typically think of "scope" to mean something that "constrains" a query's results into some smaller result set via some where clause condition. This is a pure sorting operation.

jasonmccreary:
@taylorotwell, agree. This was a bit of a proof of concept. If you feel this is a worthwhile addition, an implementation directly within Eloquent or the Builder may avoid the oddities you mentioned.

taylorotwell:
I think there is something worthwhile here in general if we nail out how it is implemented. I've definitely had situations where models are very chronological (think blog posts, comments, etc.) and I almost always want them in a given order.

jasonmccreary:
@taylorotwell, cool. I also marked this was `WIP`. Let me see if I can ingrain this a bit more within Eloquent/Builder so it has less reaching implications you and @jonnott have noted.

jonnott:
> @taylorotwell, cool. I also marked this was `WIP`. Let me see if I can ingrain this a bit more within Eloquent/Builder so it has less reaching implications you and @jonnott have noted.

I think maybe this functionality could just be worked right into where the query builder applies other normal ordering from `->orderBy()`s to the query. Without having looked at that code in detail, not entirely sure how that would happen or if it's trivial to do. I'd imagine that's more like what @taylorotwell is getting at though.

The more I think about it, the more what Taylor's saying makes sense. Using global scopes to do things like this doesn't feel quite right, _unless_ they're opt-in traits like SoftDeletes.

jonnott:
> I think the reason it starts to feel odd is this is not a "scope" at all, where I typically think of "scope" to mean something that "constrains" a query's results into some smaller result set via some where clause condition. This is a pure sorting operation.

@taylorotwell What you're saying here reminds me of issues I've faced in larger Laravel codebases where we've ended up with scope methods which mix 'constraining' things like joins and wheres with ordering too, and it all can get very messy and out of hand..

jonnott:
@jasonmccreary @taylorotwell Could this be baked into `Illuminate\Database\Query\Grammars\Grammar::compileOrders()` ?

jonnott:
> @jasonmccreary @taylorotwell Could this be baked into `Illuminate\Database\Query\Grammars\Grammar::compileOrders()` ?

This could be done by looking for a `defaultOrder` property on the model maybe, at the time the query is compiled? ..but only if `$this->orders` is otherwise empty at that point..

```
// before compiling orders during compilation of query..

      if (empty($this->orders) and empty($this->orderBy)) {
            foreach ($this->orderBy as $column => $direction) {
                is_int($column) ? $this->orderBy($direction) : $this->orderBy($column, $direction);
            }
        }
        
// carry on compiling query
```

Things that are apparent when look at the existing Eloquent code/API...

- there's not really a straightforward way within this array property syntax to define a 'raw' order by clause, and I can foresee there being use cases for that.
- possible need to consider the interaction with the `reorder()` method. I guess it can ignore `reorder()` completely if using the methodology shown above, as we'll still apply the default order when no other orders are defined, even if `orderBy()` and then `reorder()` may have both been called on the builder  .. 
- I guess the other potential problem with this is that there's no way to remove the default ordering, or override it. Which does kinda support the argument that applying a global scope to the model is maybe a better way of achieving this..

src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -1138,6 +1147,37 @@ public function scopes($scopes)
...
+     *
+     * @return static
+     */
+    public function applyOrderBy()
+    {
+        if (! empty($this->orders)) {
+            return $this;
+        }
+
+        if (empty($this->model->orderBy)) {
```
jonnott, line 1161: Shouldn't this line also be checking $this->defaultOrdering?

jasonmccreary:
@taylorotwell, this has been updated to push everything down to the Eloquent Builder. The sorting from this property is applied before the query is executed.

However, given the infinite ways developers may write Eloquent queries and the interaction with the underlying Database Query Builder, I am not sure if this is applied everywhere it should be.

As noted in the updated PR description, I only apply this sorting if no other ordering has been set for the query. I believe this helps avoid situations where queries with explicit calls to `orderBy` sort unexpectedly because it's being merged with the sorting from this new property.

Put simply, this property allows the user to _set a default sorting when not otherwise specified_.

src/Illuminate/Database/Eloquent/Builder.php
jasonmccreary: No. It has been removed.

jonnott: Otherwise @jasonmccreary it looks like a great new approach.

Just a thought - are there any other places where applyOrderBy()->applyScopes() need chaining? e.g. `toBase()` ?

jonnott: > No. It has been removed.

So what will cover the situation where a query needs to be run on the model without any ordering (including the default ordering)?

jonnott:
> @taylorotwell, this has been updated to push everything down to the Eloquent Builder. The sorting from this property is applied before the query is executed.

I started looking at this kind of approach myself .. but then I thought, there seems to be so many different places where a query is compiled (i.e. calls to `applyScopes()`, etc) .. I didn't know where to start! Was hitting the limits of my understanding of Eloquent internals at that point. Chaining applyOrderBy() before applyScopes() might be a great starting point though which @taylorotwell might know how to take forward?

src/Illuminate/Database/Eloquent/Builder.php
jasonmccreary: Potentially. I commented on this but there are probably more places this could be applied. I tried to draw the line between the Eloquent and Database query builders.

Since `toBase` returns a database query builder, I figured the user knows what they're doing and would not want this Eloquent sorting applied to their query. 🤷 

Pretty easy to add though, just makes change reach farther.

jonnott:
@jasonmccreary I still think calling this whole thing 'defaultOrderBy' rather than 'orderBy' might be less confusing for developers..

If it's just called 'orderBy' then people might think that's how you always define ordering for queries, rather than the `$query->orderBy()` method?

driesvints:
Remember that draft PR's aren't reviewed. Please mark as ready if you want a new review.

jasonmccreary:
@driesvints, sorry, I had sent this to Taylor as I worried it might need more work. I have marked it for review so we can make those tweaks and hopefully get it into L9. 🤞

jonnott:
> @driesvints, sorry, I had sent this to Taylor as I worried it might need more work. I have marked it for review so we can make those tweaks and hopefully get it into L9. 🤞

@jasonmccreary @taylorotwell  Still I think the main issue with the re-done implementation is that there really needs to be a way to *remove* the default ordering for those edge-cases where you really need the resulting query (SQL) to have no ORDER BY clause *at all* (i.e. natural table order) even though you've set this property on the model. Examples might be within tests, for instance.

jasonmccreary:
@jonnott, you should be able to call `reorder` just as you would currently.

jonnott:
> @jonnott, you should be able to call `reorder` just as you would currently.

Ah, ok. Looking at your code, I couldn't make out how calling reorder() would remove the default ordering too, as it only clears what's in the ->orders property on the builder, rather than removing the ->orderBy property of the model, which gets handled later in the query building process. I know you had the `withoutDefaultOrder()` method to do this, but then removed it.

Can you clarify if/how the default ordering would also get removed by ->reorder() ?

jasonmccreary:
@jonnott, I'm not sure it does, tbh.

Thinking about it more, I don't see _removing all `orderBy` including the default_ as a common use case if you were defaulting the order for a model. At least not without explicitly setting the desired order - which does overwrite the default.

That could be a breaking change. Which is why this is targeted for 9.0. But even so, you have to opt-in.

In your case, I would change your test cases to have an explicit order. Not try to remove all ordering.

jonnott:
> @jonnott, I'm not sure it does, tbh.

I think what I'm really getting at here is consistency with how the rest of the query builder and Eloquent work.

Essentially, anything else that you add into a query, even things added to a model 'as default' (like global scopes) can be removed .. e.g. ORDER BYs with reorder(), global scopes with withoutGlobalScope(), etc .. therefore whatever things you've configured your model with, you can always, by calling the required methods, get back to a completely 'plain' SELECT query, i.e. `SELECT * FROM {table}`. Period. *** Nothing takes away your ability to get back to that basic query.

..apart from this new `orderBy` property. How it is now in this PR, once you've defined that as a property of your model class, there *no way* to get back to that pure `SELECT * FROM {table}` anymore. *Whatever your reason for needing to do so is..* And someone, somewhere, will for sure arrive at a situation where they need to do that. That's why I feel not having that 'way out' isn't very Laravel-y, or consistent within Builder/Eloquent.

*** Correct me if I'm wrong on that one!

Really not trying to just pick holes here, genuinely want this to be a killer Eloquent feature.

jasonmccreary:
@jonnott, at this point, it's pretty unlikely this will make it into the L9 release being 3 days away. Especially with all our back and forth.

I appreciate the feedback. It's likely there's more work to do here. But I'd really like to let the Laravel Team weigh in first to help guide us in the right direction.

If you have your own ideas of how this should work, I encourage you to use what I have here to open your own PR. 👍

jonnott:
> If you have your own ideas of how this should work, I encourage you to use what I have here to open your own PR. 👍

I think what you've (re)-done is great. My suggested changes would be really simple, and essentially the same as what you already had in place before, but removed..

- I'd called the model property `defaultOrder` so it doesn't get confused with the builder `->orderBy()` method, and call the applying method in Eloquent/Builder `applyDefaultOrder()` to match.

- Add a `protected boolean $withoutDefaultOrder = false` property to Eloquent/Builder, and a method `withoutDefaultOrder()` to Eloquent/Builder which sets `$withoutDefaultOrder` to true, and return early from `applyDefaultOrder()` if **either** `empty($this->orders)` or `$this->withoutDefaultOrder` are true.

jonnott:
@jasonmccreary I think that what you've done so far, combined with my suggestions/changes, ..

- fully round this out as a neat & complete much-needed new feature
- provide clear, unambiguous naming & clarity for devs as to what it does
- make it a no-brainer to merge
- causes _zero_ BC breaks
- could totally be targeted at both Laravel 8.x and 9.x without any issue

jonnott:
@jasonmccreary Seems like `$defaultOrderBy` / `applyDefaultOrderBy()` / `$withoutDefaultOrderBy` / withoutDefaultOrderBy() works equally well as naming, and fits with how you've named the tests.

jonnott:
@jasonmccreary I've created a new commit with my suggested changes, which you could merge if you'd like .. https://github.com/jonnott/framework/commit/93e3ad3084346d567c1a587fb1ca2802a8620398

I went for `$useDefaultOrderBy` as the Eloquent/Builder boolean property name in the end, so it doesn't have the same name as either of the new **methods** added to Eloquent/Builder (`applyDefaultOrderBy()` and `withoutDefaultOrderBy()`), which could make the naming confusing.

I guess the only thing needed further to this might be an additional test for `withoutDefaultOrderBy()` perhaps?

jonnott:
..and here's that test https://github.com/jonnott/framework/commit/0d0637fda6dfac139f2c54db52f49482282ac5c4

jonnott:
What I don't know, as you'll see from the above new test, or how (or whether) I'd need to configure things such that `withoutDefaultOrderBy()` can be called directly in the model just like e.g. `ModelName::withoutGlobalScopes()` can, without having to call `ModelName::query()` first and chain, or indeed if that's even something that must be explicitly configured at all, or will just work 'out of the box'?

Perhaps I can look into that a bit with tinker..

jasonmccreary:
@jonnott, nope. Waiting to see what the Laravel Team thinks.

src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -602,7 +602,7 @@ public function valueOrFail($column)
      */
     public function get($columns = ['*'])
     {
-        $builder = $this->applyScopes();
+        $builder = $this->applyOrderBy()->applyScopes();
```
crynobone, line 605: What's if order by is applied via scope, and based on the code scope only applied after order by?

jonnott: @crynobone That's a very good point. Maybe @jasonmccreary these chained calls need to be the other way around?

jonnott: @crynobone @jasonmccreary Fix for the method ordering committed to my fork https://github.com/jonnott/framework/commit/b835133c682a57ded5cf70029c76a0e46ba41b86

driesvints:
Wait for Taylor to confirm this but I think it's best that we retarget to 10.x since we're releasing today. It's a bit tricky to still merge this one in.

src/Illuminate/Database/Eloquent/Builder.php
jonnott: @jasonmccreary I've also added a test for @crynobone's mentioned situation where an orderBy() is applied via a scope: https://github.com/jonnott/framework/commit/231d4f6aacc8896c63ccdcf392dd93996aca9196 (test would've failed without the switching of `applyDefaultOrderBy()->applyScopes()` to `applyScopes()->applyDefaultOrderBy()`)

jonnott:
> Wait for Taylor to confirm this but I think it's best that we retarget to 10.x since we're releasing today. It's a bit tricky to still merge this one in.

100% agree this still isn't quite ready for today's 9.0 release. Even this morning it's needed further fixes and tests adding, thanks to @crynobone's helpful observation.

jasonmccreary:
@driesvints, no worries. Things got a bit noisy here and I didn't really know which final tweaks this needed. This does work, but agree there are a lot of boundary cases to consider.

I'll close this for now. I'm hopeful we won't need to wait a year until 10.x to get this added. As a new property it would be opt-in. And to the point above, it shouldn't have any breaking changes once the right tweaks are made.

PR #40680, opened 2022-01-27
[9.x] Use null safe operator in cache Repository class

imanghafoori1:
As laravel 9.x requires PHP 8.0.x I think we can refactor the situation to a more sugary one if there is not a subtle difference there.

### DIFF
--- a/src/Illuminate/Cache/Repository.php
+++ b/src/Illuminate/Cache/Repository.php
@@ -583,9 +583,7 @@ class Repository implements ArrayAccess, CacheContract
      */
     protected function event($event)
     {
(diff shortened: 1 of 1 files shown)


src/Illuminate/Cache/Repository.php
```diff
@@ -583,9 +583,7 @@ public function getStore()
      */
     protected function event($event)
     {
-        if (isset($this->events)) {
-            $this->events->dispatch($event);
-        }
+        $this->events?->dispatch($event);
```
dammy001, line 586: I think the `event` method is not expecting null so the previous is correct

PR #40681, opened 2022-01-27
[8.x] Add Transliterate shortcut to the Str helper

liamh101:
### Description

Add a transliterate shortcut to the String helper. 

I've added the by recommendation on [a recent PR](https://github.com/laravel/ui/pull/216) around fixing security exploit in the UI package. 

In short, MYSQL and other DB hosts will transliterate special characters in queries, making it possible to bypass security checks and throttles.

In the case of the UI security exploit, by default, Laravel throttles login attempts by the applications email/username. However, if I use special characters to replace standard characters in my email when the email gets to the Database, the query parameters will transliterate. This means that `ⓣⓔⓢⓣ@ⓛⓐⓡⓐⓥⓔⓛ.ⓒⓞⓜ` will be read as `test@laravel.com` by the Database. This made it was possible for a user to brute force a password by bypassing the login throttle.

Adding this shortcut will make it easier to implement a more elegant solution to the current UI fix while also making it easier for users to improve the security of their application or similar functionality by adding transliteration directly from the helper without needing prior knowledge of the ASCII package.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -100,6 +100,19 @@ class Str
     }
 
     /**
+     * Transliterate a string to its closest ASCII representation.
+     *
+     * @param  string  $string
+     * @param  string|null  $unknown
+     * @param  bool|null  $strict
+     * @return string
+     */
+    public static function transliterate($string, $unknown = '?', $strict = false)
+    {
+        return ASCII::to_transliterate($string, $unknown, $strict);
+    }
+
+    /**
      * Get the portion of a string before the first occurrence of a given value.
      *
      * @param  string  $subject
@@ -984,19 +997,6 @@ class Str
     public static function createUuidsNormally()
     {
         static::$uuidFactory = null;
-    }
-
-    /**
-     * Transliterate a string into it's closest ASCII representation.
-     *
-     * @param  string  $string
-     * @param  string|null  $unknown
-     * @param  bool|null  $strict
-     * @return string
-     */
-    public static function transliterate($string, $unknown = '?', $strict = false)
-    {
-        return ASCII::to_transliterate($string, $unknown, $strict);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
Some email addresses could genuinely have UTF-8 characters in them though, right? Or is that not allowed? 😅

src/Illuminate/Support/Str.php
```diff
@@ -986,6 +986,19 @@ public static function createUuidsNormally()
...
 
+    /**
+     * Transliterate a string into it's closest ASCII representation.
+     *
+     * @param  string  $string
+     * @param  string|null  $unkown
+     * @param  bool|null  $strict
+     * @return string
+     */
+    public static function transliterate($string, $unkown = '?', $strict = false)
```
derekmd, line 997: Should the `$unkown` argument be `$unknown`? https://github.com/voku/portable-ascii/blob/f1c1181c85e099b66713c7a36a6d65814795d149/src/voku/helper/ASCII.php#L1178

If accepted, most `Str` methods also get added to the `Stringable` class.

liamh101:
> Some email addresses could genuinely have UTF-8 characters in them though, right? Or is that not allowed? 😅

Currently, the only allowed special characters are `~!$%^&*_=+}{'?-.` however for the exploit it doesn't really matter as that email is just generating a key to store against. So thankfully if emoji emails do become a thing, that fix will still hold up... Mostly 😛

src/Illuminate/Support/Str.php
liamh101: I debated changing this and the strict flag. As both on their own are nondescript. However, decided to keep their original names for consistency. 

I could change `$unkown` to `$unkownFallback` and `$strict` to `$useIntl` as it's a bit more descriptive.

derekmd: Sorry, I meant that there's a typo missing the second 'n' in unk_n_own.

liamh101: D'oh! Two secs

PR #40683, opened 2022-01-27
[9.x] API Mode

taylorotwell:
This command can automatically be invoked via an upcoming new `--api` switch when creating a new Laravel application.

It performs several actions that prepare the application to be used primarily as a backend API. First, a new middleware is added that inject the `Accept` header into all incoming requests as `application/json` - this has numerous affects throughout the framework such as forcing the exception handler to always return JSON - disabling redirects in various places, etc.

- Unnecessary middleware is removed.
- Frontend asset related files are removed such as `package.json` and `webpack.mix.js`.
- The `resources/css` and `resources/js` directories are removed.
- The `routes/web.php` file is removed.
- The default `welcome.blade.php` file is removed.
- The HTTP Kernel is updated to remove unnecessary middleware the "web" middleware stack (still considering this).
- The `RouteServiceProvider` is updated accordingly.


smm993:
Add Cors middleware

src/Illuminate/Foundation/Console/stubs/convert-to-api/AlwaysAcceptJsonResponses.stub
```diff
@@ -0,0 +1,23 @@
...
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure(\Illuminate\Http\Request): (\Illuminate\Http\Response|\Illuminate\Http\RedirectResponse)  $next
+     * @return \Illuminate\Http\Response|\Illuminate\Http\RedirectResponse
+     */
+    public function handle(Request $request, Closure $next)
+    {
+        return $next(tap($request, function ($request) {
+            $request->headers->set('Accept', 'application/json');
+        }));
```
SjorsO, line 21: How about just:

```php
$request->headers->set('Accept', 'application/json');

return $next($request);
```

Easier to comprehend and 29 characters shorter.

src/Illuminate/Foundation/Console/stubs/convert-to-api/api.stub
```diff
@@ -0,0 +1,24 @@
...
+| API Routes
+|--------------------------------------------------------------------------
+|
+| Here is where you can register API routes for your application. These
+| routes are loaded by the RouteServiceProvider within a group which
+| is assigned the "api" middleware group. Enjoy building your API!
+|
+*/
+
+Route::get('/', function (Request $request) {
```
dammy001, line 18: I think we can use arrow function for this

src/Illuminate/Foundation/Console/stubs/convert-to-api/AlwaysAcceptJsonResponses.stub
```diff
@@ -0,0 +1,23 @@
...
+    }
+}
```
driesvints, line 23: This can be an actual class in the framework that can be copied over. No need to make it a stub.

driesvints:
Wild thought: what if we created a `laravel/api` skeleton repo instead? It would keep all of this from the core framework and we could more easily update the installer.

iAxel:
> Wild thought: what if we created a `laravel/api` skeleton repo instead? It would keep all of this from the core framework and we could more easily update the installer.

Good idea!

timothepearce:
It's time to deprecate Laravel Lumen 😄

driesvints:
@smortexa instead of downvoting me, can you explain why that wouldn't seem a good option to you?

smortexa:
@driesvints IMO, the backend API skeleton differences are not so much or fundamental that a separate repository is needed. Furthermore, a `laravel/laravel` repo as a full-stack skeleton and a `laravel/api` repo as a backend API skeleton beside `laravel/lumen` doesn't make sense.

driesvints:
@smortexa those are two very different things. `laravel/api` would be a skeleton repo for the Laravel framework while Lumen is a totally separate framework. Having different repos can help disentangle the skeleton repo from the core framework and make it much easier for us to perform updates.

ajmeireles:
With this PR can't we start thinking about isolating Blade in a separate package?  It's true that we don't need it at various times, especially now with this new feature.

dennisprudlo:
@ajmeireles Emails sent using an API will still require blade to render.

ajmeireles:
> @ajmeireles Emails sent using an API will still require blade to render.

I understand your position, but it is not in any case that an API created with Laravel will need to send emails, for example. The idea was to abstract Blade into a separate package: `composer require laravel/blade`.

In other words: whoever is creating an API and needs the Blade, just include it. Being in a separate package, this PR would no longer have to worry about anything related to it.

It's a simple suggestion.

mfn:
> Emails sent using an API will still require blade to render.

Drive-by observation: "usually" in modern apps, all email sending is deferred to background jobs.

Thus it remains the question of such an API is also the same codebase for background jobs / running worker / Horizon or if its actually dispatched somewhere else.

Laravel doesn't make it easy to randomly dispatch a code somewhere else (=out of the box) because it heavily relies on class serialization and de-serialization (but it is possible when being careful).

novalramdhani:
very amazing!!!!

src/Illuminate/Foundation/Console/stubs/convert-to-api/RouteServiceProvider.stub
```diff
@@ -0,0 +1,61 @@
...
+class RouteServiceProvider extends ServiceProvider
+{
+    /**
+     * The path to the "home" route for your application.
+     *
+     * This is used by Laravel authentication to redirect users after login.
+     *
+     * @var string
+     */
+    public const HOME = '/home';
```
siarheipashkevich, line 20: Can we remove it?

src/Illuminate/Foundation/Console/ConvertToApiCommand.php
```diff
@@ -0,0 +1,70 @@
...
+        $files->put(resource_path('views/.gitkeep'), PHP_EOL);
+
+        // Fix tests...
+        $files->put(
+            base_path('tests/Feature/ExampleTest.php'),
+            str_replace("'/'", "'/api'", $files->get(base_path('tests/Feature/ExampleTest.php')))
+        );
+
+        // Install stubs...
+        $files->copy(__DIR__.'/stubs/convert-to-api/AlwaysAcceptJsonResponses.stub', app_path('Http/Middleware/AlwaysAcceptJsonResponses.php'));
```
deleugpn, line 65: In my applications I call this EnforceJsonResponse

aneesdev: or how about `AlwaysExpectsJsonResponses`

taylorotwell:
Closing this for now as I do think it makes sense for this to maybe just be a separate skeleton in the future.

DarkGhostHunter:
Is this the end of Lumen then?

PR #40685, opened 2022-01-28
[9.x] Add SesTransport decorator

driesvints:


deleugpn:
Hey @driesvints which implementation you prefer to maintain long-term on Laravel? I see this decorator as a small tech footprint as it just adapts Laravel config to Symfony Mailer while the other PR bring back the entire AWS SDK and go back to handling email within Laravel instead of through Symfony.

Which one should I try to test first?

src/Illuminate/Mail/Transport/SesTransport.php
```diff
@@ -0,0 +1,101 @@
...
+    protected $options;
+
+    /**
+     * Create a new SES transport instance.
+     *
+     * @param  \Symfony\Component\Mailer\Bridge\Amazon\Transport\SesApiAsyncAwsTransport|\Symfony\Component\Mailer\Bridge\Amazon\Transport\SesHttpAsyncAwsTransport  $transport
+     * @param  array  $options
+     * @return void
+     */
+    public function __construct(TransportInterface $transport, array $options = [])
```
deleugpn, line 37: Since Laravel 9 requires PHP 8, couldn't the type-hint here be `SesApiAsyncAwsTransport|SesHttpAsyncAwsTransport` instead and remove the exception below?

driesvints:
@deleugpn heya, the re-implementation of the old SesTransport: https://github.com/laravel/framework/pull/40696

This is because that one already fully offers every feature for SES while the Amazon Mailer by Symfony is more limited. Plus that way we don't need to wait until Symfony does a new minor release to re-implement missing features.

deleugpn:
This implementation works correctly and the ConfigurationSet is properly configured. However, when adding custom headers to the Email, they are being lost somewhere along the execution stack. I will not investigate this further because of the preference of the other PR, so I will just jump to testing that one instead. I can dig deeper into the headers being lost if there is desire to merge this one.

Here is how I'm adding custom headers

```
$closure = function(Email $message) use ($tenant) {
    $message->getHeaders()->addTextHeader('X-ORG-TENANT', $tenant);
};
```

PR #40687, opened 2022-01-28
[8.x] Throw an event when a relation is loaded

ahmes:
Right now, when you use eloquent models you have to be mindful on what are the relations you need to eager load. 

Code is a living thing, continuously changing. As a result, one can easily forget to add/remove relations that are/aren't needed. 

By listening and counting, on a per model basis, what relations were called and how many times we can detect:

- relations that are used but not eager loaded
- relations that are eager loaded but not used

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -14,6 +14,7 @@ use Illuminate\Database\Eloquent\Casts\Attribute;
 use Illuminate\Database\Eloquent\InvalidCastException;
 use Illuminate\Database\Eloquent\JsonEncodingException;
 use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Database\Events\RelationLoadedEvent;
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
A few problems here:

* You cannot call the foundation event helper function from this package. This package doesn't know about that function.
* The event is very confusing in name. It is not raised every time a relation is loaded, it is only raised in certain specific cases.

olivernybroe:
A solution for these issues could be to use the methods from `HasEvents` trait instead of using the event dispatcher directly.
Would prob. make sense to add it as a new observable event on the model in `HasEvents`, that way the event class is also not needed.

ahmes:
> A few problems here:
> 
> * You cannot call the foundation event helper function from this package. This package doesn't know about that function.
> * The event is very confusing in name. It is not raised every time a relation is loaded, it is only raised in certain specific cases.

@GrahamCampbell 
Where would you recommend this event be thrown?
How can such an event be thrown and have it include the model that requested a relation and the relation name?

taylorotwell:

PR #40694, opened 2022-01-28
[8.x] Fix Stringable ucsplit

tomirons:
This was passing an array into the constructor which causes an exception. Also, this should be the last method you call on the chain since `Str::ucsplit` returns an array.


foremtehan:
What is wrong with returning Collection ?

PR #40696, opened 2022-01-28
[9.x] Re-implement original SesTransport

driesvints:
Replaces https://github.com/laravel/framework/pull/40685

This PR re-implements the old SesTransport again. All features should work again except the swift mailer specific ones.

Here's the original SesTransport: https://github.com/laravel/framework/blob/8.x/src/Illuminate/Mail/Transport/SesTransport.php

I've also removed the addition of Message ID's to the headers of the sent message afterwards since Symfony mailer now takes care of generating this.

I have not been able to manually test this out yet.


driesvints:
Ping @deleugpn ^

taylorotwell:
@driesvints  Can you mark this as ready when you are able to test it manually?

deleugpn:
I will try this out on Monday and provide feedback. Thank you so much for the work!

deleugpn:
This implementation works correctly and the ConfigurationSet is properly configured. However, when adding custom headers to the Email, they are being lost somewhere along the execution stack. I will not investigate this further because of the preference of the other PR, so I will just jump to testing that one instead. I can dig deeper into the headers being lost if there is desire to merge this one.

Here is how I'm adding custom headers

```
$closure = function(Email $message) use ($tenant) {
            $message->getHeaders()->addTextHeader('X-ORG-TENANT', $tenant);
        };
```

src/Illuminate/Mail/Transport/SesTransport.php
```diff
@@ -0,0 +1,104 @@
...
+                    $this->options, [
+                        'Source' => $message->getEnvelope()->getSender()->toString(),
+                        'RawMessage' => [
+                            'Data' => $message->toString(),
+                        ],
+                    ]
+                )
+            );
+        } catch (AwsException $e) {
+            throw new Exception('Request to AWS SES API failed.', $e->getCode(), $e);
```
deleugpn, line 60:
```suggestion
            throw new Exception('Request to AWS SES API failed. ' . $e->getAwsErrorMessage(), $e->getCode(), $e);
```

When the develper have a try/catch wrapping this error, the final error message is unclear on what happened. By including AWS Error Message here the developer is able to identify what went wrong.
In my case, I was missing `ses:SendRawEmail` permission, but I was only able to detect that after modifying this line.

driesvints: @deleugpn can you maybe tackle that in a separate PR? Since this has always been the behaviour from 8.x and before. Just trying to get this working again. I do agree with you on that.

deleugpn: I guess I will need to wait for this to get merged before asking to addres this because this file does not exist on the current stable branch so I can't propose a change yet

PR #40699, opened 2022-01-28
[8.x] Return collection from ucsplit

tomirons:
As suggested in my previous PR (#40694), other methods like `explode` return a collection, so why not this one?


buismaarten:
Isn't this a breaking change on 8.x?

driesvints:
@buismaarten this method was broken (see the other PR) so this wasn't working in the first place.

PR #40704, opened 2022-01-29
[8.x] Add getLineLoaded Function

AmrElabsy:
JSON Provides a hierarchy style for the key value. This means that value can be an array too.
For some organizing work, you would need to write __("foo.bar") and load it from the JSON file.
This function provides you with the ability to assign the value of the key "foo" to be an array
that contains a key called "bar" with a string value.
It doesn't affect the old style of writing and processing the function.


taylorotwell:

PR #40711, opened 2022-01-29
[8.x] Fix LazyCollection::collapse() not passing along keys

axlon:
Currently `LazyCollection::collapse()` is not consistent with `Collection::collapse()`, because the lazy collection does not pass along the keys within its items, while the normal collection does.

Example:

```php
>>> collect([['one' => 1], ['two' => 2]])->collapse()->all()
=> [
     "one" => 1,
     "two" => 2,
   ]
>>> collect([['one' => 1], ['two' => 2]])->lazy()->collapse()->all()
=> [
     1,
     2,
   ]

```

### DIFF
--- a/src/Illuminate/Collections/LazyCollection.php
+++ b/src/Illuminate/Collections/LazyCollection.php
@@ -166,8 +166,8 @@ class LazyCollection implements CanBeEscapedWhenCastToString, Enumerable
         return new static(function () {
             foreach ($this as $values) {
                 if (is_array($values) || $values instanceof Enumerable) {
-                    foreach ($values as $value) {
-                        yield $value;
+                    foreach ($values as $key => $value) {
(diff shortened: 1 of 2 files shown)


axlon:
Tests are failing, marking this PR as a draft for now

axlon:
Closing this as I don't think there is a way to make `LazyCollection::collapse()` work the same way as the normal collection's counterpart. This is because the normal collection makes use of `array_merge()` which will only overwrite keys if they are not sequential, while yielding will always result in overwritten keys.

For future reference, `LazyCollection::mapWithKeys()` may be (ab)used to create the structure you would expect from collapse when working with string keys.

E.g.:
```php
collect([['one' => 1], ['two' => 2]])->lazy()->mapWithKeys(fn (array $items) => $items)->all()
```

PR #40712, opened 2022-01-29
[8.x] Define route redirection when model is missing

cerbero90:
This PR introduces a papercut to define the route that a user should be redirected to when a model is missing:

```php
// before
Route::get('posts/{post}', [PostsController::class, 'show'])->missing(function() {
    return Redirect::route('posts.index', status: 301);
});

// after
Route::get('posts/{post}', [PostsController::class, 'show'])->redirectMissing('posts.index');
```

### DIFF
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -12,6 +12,7 @@ use Illuminate\Routing\Matching\MethodValidator;
 use Illuminate\Routing\Matching\SchemeValidator;
 use Illuminate\Routing\Matching\UriValidator;
 use Illuminate\Support\Arr;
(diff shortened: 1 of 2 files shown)


foremtehan:
Also add this to Illuminate\Support\Facades\Route docblock, thanks

cerbero90:
Hi @foremtehan, that is not needed as the `Illuminate\Support\Facades\Route` facade is linked to the `Illuminate\Routing\Router`, not the `Illuminate\Routing\Route`

taylorotwell:
Hmm, I think it's not too bad with existing functionality on Laravel 9 with new `to_route` helper:

`Route::get(...)->missing(fn () => to_route('posts.index', status: 301));`

PR #40713, opened 2022-01-29
[8.x] `getOrFail` for the Translator

foremtehan:
Almost all of my projects need `Translator`, Its important to me to 'translate' but sometimes i forget to translate  some keys in some languages, Translator return string no matter the key was defined or not So here it is.

If merged i add tests and facades method


taylorotwell:

PR #40717, opened 2022-01-30
Fix phpdoc on handle method

squiaios:
This PR fix the phpdoc of handle method by adding `bool` type


GrahamCampbell:
I think the mistake here that false is returned. We definitely shouldn't change the phpdoc, and we probably shouldn't even fix this on 8.x.

driesvints:
A return type can never be void and something else at the same time. Most likely, it needs to be `return;` like @GrahamCampbell suggests.

PR #40720, opened 2022-01-31
[8.x] Add array_keys validation rule

Drewdan:
I have recently been finding it more common to have to validate that certain items are available in an array. However, the items that are required are usually dynamically driven and then stored as JSON in a DB column.

A really simple example of this might be shifts for a team. Each team has working days which are pre-defined and stored in a database. There is then a dynamically generated form which asks for the shift times for these days.

The required data for each shift would be the same, a start time and an end time and would be required. Typically, to validate this dynamic data, you might do something such as:

```php
public function rules() {

$dayRules = BusinessHourDay::sameCompany()->get()->map(function ($businessHourDay) {  
  return [  
  'days.' . $businessHourDay->day . '.start_time' => 'required',  
  'days.' . $businessHourDay->day . '.end_time' => 'required',  
  // more rules  
  ];  
});  
  
return [  
  'name' => 'required',  
  'description' => 'required'
  'days' => 'array',
  ...$dayRules,  
];
```

This allows us to validate the days that the user has configured in the database are present in the dynamically generated form.

This proposed change would allow this instead:

```php
public function rules() {

$dayRules = BusinessHourDay::sameCompany()->get()->pluck('day');  
return [  
  'name' => 'required',  
  'description' => 'required',  
  'days' => [  
  'bail',
  'array',  
  'contains_all,' . $dayRules->implode(','),  
  ],  
  'days.*.start_time' => 'required',  
  'days.*.end_time' => 'required',
  // more rules  
];
```

This would simplify the validation in the cases where mapping a list of validation rules, which would have the same rules for each entry, but we still want to assert that all the required data was present.

The rule could also be applied to a string field to make sure a part or parts of it given match a particular structure, for example, a form where you might enter an invoice "number" something like "INV:12345-WEB"

```php
public function rules() {

return [
	'invoice_id' = [
		'required',
		'contains_all:INV,WEB',
	],
];
```

Would be grateful for any feedback or thought you have on this :)

Thanks guys.

### DIFF
--- a/src/Illuminate/Validation/Concerns/ReplacesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ReplacesAttributes.php
@@ -613,13 +613,13 @@ trait ReplacesAttributes
     /**
      * Replace all place-holders for the contains_all rule.
      *
-     * @param string  $message
-     * @param string  $attribute
-     * @param string  $rule
-     * @param array  $parameters
-     * @return string
+     * @param  string  $message
+     * @param  string  $attribute
+     * @param  string  $rule
+     * @param  array  $parameters
+     * @return  string
      */
-    protected function replaceContainsAll($message, $attribute, $rule, $parameters)
+    protected function replaceArrayKeys($message, $attribute, $rule, $parameters)
     {
         foreach ($parameters as &$parameter) {
             $parameter = $this->getDisplayableValue($attribute, $parameter);
--- a/src/Illuminate/Validation/Concerns/ReplacesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ReplacesAttributes.php
@@ -201,6 +201,24 @@ trait ReplacesAttributes
     }
 
     /**
+     * Replace all place-holders for the array_keys rule.
+     *
+     * @param  string  $message
+     * @param  string  $attribute
+     * @param  string  $rule
+     * @param  array  $parameters
+     * @return string
+     */
+    protected function replaceArrayKeys($message, $attribute, $rule, $parameters)
+    {
+        foreach ($parameters as &$parameter) {
+            $parameter = $this->getDisplayableValue($attribute, $parameter);
+        }
+
+        return str_replace(':values', implode(', ', $parameters), $message);
+    }
+
+    /**
      * Replace all place-holders for the mimetypes rule.
      *
      * @param  string  $message
@@ -602,24 +620,6 @@ trait ReplacesAttributes
      * @return string
      */
     protected function replaceStartsWith($message, $attribute, $rule, $parameters)
-    {
-        foreach ($parameters as &$parameter) {
-            $parameter = $this->getDisplayableValue($attribute, $parameter);
-        }
-
-        return str_replace(':values', implode(', ', $parameters), $message);
-    }
-
-    /**
-     * Replace all place-holders for the contains_all rule.
-     *
-     * @param  string  $message
-     * @param  string  $attribute
-     * @param  string  $rule
-     * @param  array  $parameters
-     * @return string
-     */
-    protected function replaceArrayKeys($message, $attribute, $rule, $parameters)
     {
         foreach ($parameters as &$parameter) {
             $parameter = $this->getDisplayableValue($attribute, $parameter);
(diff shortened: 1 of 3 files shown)


taylorotwell:
I'm curious if we need any additional functionality in the framework for this. It seems you could do:

```php
return [
  'name' => 'required',
  'description' => 'required',
  'days' => 'bail|array',
  ...BusinessHourDay::sameCompany()->get()->flatMap(function ($day) {
      return [
          'days.'.$day->day.'.start_time' => 'required',
          'days.'.$day->day.'.end_time' => 'required',
      ];
  }),
];
```

taylorotwell:
I think I'm a having a bit of a hard time following the rule. Can you give another example of using it - you also use `contains_all` and `contains` in your examples. I'm not sure if those are for the same rule?

> My intended approach here is to handle the times when you might have a some simple validation requirements, like just making sure it has some value using the required validation rule, without having to add any "complexity" to the code. Has the array got all of these keys? Yes, great, if not fail.

Can you give an example of this? Ideally without any loops - just hard-coded strings to make it very clear.

Drewdan:
Apologies, it should be `contains_all` in my examples - I originally named the rule `contains` and then realized that this was not clear, as it is checking to see if all keys are present, so changed it to `contains_all` and forgot to update my write-up.

I am just going to cook my kids dinner, and then I will write up a clearer example for you to review :)

Drewdan:
If you had an array of questions which you wanted to be asked in a form, you might have something like this:

```php
$questions = [
	'favourite_color' => 'What is your favorite color?',
	'favourite_animal' => 'What is your favorite animal?',
	'maiden_name' => 'What is your Mother\'s maiden name?',
	'favorite_php_framework' => 'What is your favorite PHP framework?',
];
```

(This would likely be database driven, or a static array on a model, but the method of which we get it is not really relevant to this.)

You could use these to generate a form in a blade file using the key as the input name in an array.

```blade
@foreach($questions as $question => $key)
<input type="text" name="questions[{{ $key }}] value="{{ old('questions.' . {{ $key}}) }} ">
@endforeach
```

When it comes to validation, with these, I don't have to care much about the content, only that the user has answered something for them.

I would build a rule set in a request validator by looping through the array which would end up looking something like this:


```php
public function rules() 
{
	return [
		'questions' => 'array',
		'questions.favorite_color' => 'required',
		'questions.favourite_animal' => 'required',
		'questions.maiden_name' => 'required',
		'questions.favorite_php_framework' => 'required',
	];
}
```

The rules are the same, and I would have had to flatMap or some other form of loop to build up this rule set. Whereas my approach with the contains_all would be to have a rule set looking like this:

```php
public function rules()
{
	return [
		'questions' => [
			'array',
			'contains_all:favorite_color,favorite_animal,maiden_name,favorite_php_framework',
		],
		'questions.*' => 'required',
	];
}
```
The contents of the `contains_all` would be imploded from the array.

My belief here is this approach is easier when you are trying to keep validation really simple.

taylorotwell:
Gotcha. I wonder if something like `contains_keys` would be a clearer name? Or maybe `requires_keys` so it's clear that all of them are required?

Drewdan:
I feel like array_keys would make sense. I had it like this so the behaviour could be applied to a string as well, as most other rules seem to work across data types, but I am not sold on my string implementation. I feel pulling this back to just work with arrays seems like a sensible move.

Drewdan:
I have pushed some new commits to rename the rule to `array_keys` and also makes the validation fail if the subject is not an array, also removing the validation rules for a string.

morloderex:
@Drewdan @taylorotwell just to make sure you have not forgotten that the “array” rule can actually already take in parameters and if I am remembering correctly already check for precedence of keys in the array?

Could it be that the array rule could be helpful in this case as well, sense you already know the specific keys?

Drewdan:
@morloderex - as I understand, the array validation method accepts a list of valid keys, but this does not validate that they are present. This rule differs in that it validates all keys are present in the array. Ideally, you would use these in tandem with each other.

taylorotwell:
Renamed to `required_array_keys`.

taylorotwell:
@morloderex ... @Drewdan is correct that `array` rule only accepts a list of valid keys but does not validation that they are present.

Drewdan:
@taylorotwell renaming to `required_array_keys` makes sense, it is a much clearer rule name :)

PR #40721, opened 2022-01-31
[8.x] Allows enums as morphed model type in QueryBuilder

damek24:
Consider following model and an enum
```
enum MyTypes: string
{
    case type1 = 'Type1';
    case type2 = 'Type2';
}
class MyModel extends Model
{
    public $casts [
         'item_type' => MyTypes::class
    ];

    public function item(): MorphTo
    {
        return $this->morphTo();
    }
}
```
In AppServiceProvider custom morph types are set (#40375 allowed this)
```
    public function boot()
    {
        Relation::morphMap(
            [
                'Type1' => \App\Models\ModelTypeOne::class,
                'Type2' => \App\Models\ModelTypeTwo::class
            ]
        );
    }
```
and when I try to get models with the query
`MyModel::query()->whereHasMorph('item', [MyTypes::type2 ])->get()`

instead of models I get
```
TypeError
Illegal offset type
```
This happens because a BakedEnum cannot be an array key

This pr allows BakedEnum to be used in `whereHasMorph` method and the key is either string key or enum's value

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Relation.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Relation.php
@@ -462,7 +462,7 @@ abstract class Relation
      */
     public static function getMorphedModel($alias)
     {
-        $key = is_string($alias) ? $alias : (isset($alias->value) ? $alias->value : '');
+        $key = $alias instanceof \BackedEnum ? $alias->value : $alias;
         return static::$morphMap[$key] ?? null;
     }
 
--- a/src/Illuminate/Database/Eloquent/Relations/Relation.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Relation.php
@@ -457,12 +457,13 @@ abstract class Relation
     /**
      * Get the model associated with a custom polymorphic type.
      *
-     * @param  string  $alias
+     * @param  string|object  $alias
      * @return string|null
      */
     public static function getMorphedModel($alias)
     {
-        return static::$morphMap[$alias] ?? null;
+        $key = is_string($alias) ? $alias : (isset($alias->value) ? $alias->value : '');
+        return static::$morphMap[$key] ?? null;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tests/Database/DatabaseEloquentRelationTest.php
```diff
@@ -19,6 +20,18 @@ protected function tearDown(): void
         m::close();
     }
 
+    public function testGetMorphedModelWithEnum()
+    {
+        if (version_compare(PHP_VERSION, '8.1') < 0) {
```
driesvints, line 25: Please use annotations for this instead.

damek24: @driesvints like this?
```
    /**
     * @requires PHP >= 8.1
     */
```

driesvints: Yep

damek24: done

src/Illuminate/Database/Eloquent/Relations/Relation.php
```diff
@@ -457,12 +457,13 @@ protected static function buildMorphMapFromModels(array $models = null)
...
      * Get the model associated with a custom polymorphic type.
      *
-     * @param  string  $alias
+     * @param  string|object  $alias
      * @return string|null
      */
     public static function getMorphedModel($alias)
     {
-        return static::$morphMap[$alias] ?? null;
+        $key = is_string($alias) ? $alias : (isset($alias->value) ? $alias->value : '');
```
crynobone, line 465:
```suggestion
        $key = $alias instanceof \BackedEnum ? $alias->value : $alias;
```

damek24:
I've updated the description. Is it sufficient enough for now? should I create an integration test where I simulate the behavior from the decription?

Issue #40724, opened 2022-01-31
Illuminate/view has hidden dependency on illuminate/conditionable not specified in composer.json


calvinalkan:
- Laravel Version: 8.81
- PHP Version:  7.4.27
- Database Driver & Version: standalone blade package.

### Description:

I'm using Blade as a standalone templating engine without the full framework.
There is a hidden dependency in the illuminate/view package that will cause a Fatal Error At runtime.

`Fatal error: Trait 'Illuminate\Support\Traits\Conditionable'`

The [ComponentAttributeBag](https://github.com/laravel/framework/blob/v8.81.0/src/Illuminate/View/ComponentAttributeBag.php) requires `Illuminate\Support\Traits\Conditionable` but that package is not specified in [composer.json](https://github.com/laravel/framework/blob/v8.81.0/src/Illuminate/View/composer.json)

driesvints:
This trait is present in the support component which is a dependency of the view component in 8.x so this can't happen under the version you specified. However, it seems this dependency is indeed missing in 9.x so I've sent in a fix for that: https://github.com/laravel/framework/pull/40725

Thanks

driesvints:
I also think this already works in 9.x because conditionable is a dependency of support which is a dependency of view.

calvinalkan:
I double-checked again why this happens. It happens when you run composer install with the --prefer-lowest option, which I need in CI/CD. And since the Conditionable trait was added in illuminate/support 8.45 ( I think ) it blows up because it will install 8.0.0

calvinalkan:
@driesvints So basically the first tagged version of `illuminate\view` that uses the Conditionable trait should explicitly set the `illuminate/support` version to `^8.45"`. [First introduced in 8.45](https://github.com/laravel/framework/commit/45ff23c6174416f63ea7dbd77bc7fe8aafced86b#diff-e2f597fdca8b411b08956d2bc32a6ddbe376b9f7edb83c4f26b55932456ef03e)

driesvints:
@calvinalkan I see. Yeah we can do that. Feel free to send in a PR 👍

calvinalkan:
@driesvints How would you go about this?

The problem also exists for other dependencies. For example, illuminate view will also break with any version of collections lower than 8.32 due to the value() helper not accepting args.
[...]

PR #40725, opened 2022-01-31
[9.x] Fix conditionable dependencies for Illuminate components

driesvints:
Fixes https://github.com/laravel/framework/issues/40724 and a couple of other components that depend on this trait


driesvints:
Actually, this isn't needed because conditionable is already a dependency of support.

PR #40729, opened 2022-01-31
[8.x] Define multiple foreign model IDs in table Blueprint at once

VaguelyOnline:
The PR exposes a convenience method of defining multiple foreign ID columns for related models. 

### Backwards compatability

Since it sits on top of the existing $table->foreignIdFor(...) method and does not modify the existing behaviour of that method, it does not present any breaking changes or backwards incompatability.

### Benefit to users

Currently, users are able to define foreign Ids tied to models using the existing 'foreignIdFor(Model::class)' method. This new method enables users to more succinctly define tables that relate to multiple foreign models. All resulting ForeignIdColumnDefinitions remain accessible for additional individual configuration or handling. Furthermore, since the method returns a Collection, the use of HigherOrderCollectionProxy methods, or 'each' Collection callbacks can also be used to apply settings to the entire collection.

### Unit tests

Tests accompany this PR. See the additional tests in tests/Database/DatabaseSchemaBlueprintTest.php. Since the tests are named with a similar base, they can all be ran with the following command: 
`php .\vendor\bin\phpunit --filter='testGenerateMultipleRelationshipColumns'`

### Example usages

Typically from within a migration:

Conventional column names will be used (user_id, project_id, state_id):
```php
Schema::table('uploads', function (Blueprint $table) {
    $table->foreignIdsFor([
        User::class,
        Project::class,
        State::class
    ]);
});
```

Conventional column names will be used (user_id, project_id, state_id), all columns made nullable.
```php
Schema::table('uploads', function (Blueprint $table) {
    $table->foreignIdsFor([
        User::class,
        Project::class,
        State::class
    ])->each->nullable();
});

// Or without the use of HigherOrderCollectionProxy
Schema::table('uploads', function (Blueprint $table) {
    $table->foreignIdsFor([
        User::class,
        Project::class,
        State::class
    ])->each( fn($def) => $def->nullable() );
});
```

Note - to express the same schema with the existing API, the following code is necessary - 215 chars vs 160 chars with the proposed change.
```php
Schema::table('uploads', function (Blueprint $table) {
    $table->foreignIdFor(User::class)->nullable();
    $table->foreignIdFor(Project::class)->nullable();
    $table->foreignIdFor(State::class)->nullable();
});
```

To provide a specific column name, or handle the ForeignIdColumnDefinition, the $table may configured in the following way. Assuming no other conventions are changed in model class definitions, the resulting columns created will be: 'post_id', 'uploader_id', 'project_id', and 'status_id;'
```php
Schema::table('uploads', function (Blueprint $table) {
    $table->foreignIdsFor([
        Post::class,
        User::class => 'uploader_id',
        Project::class => fn(ForeignIdColumnDefinition $definition) => $definition->nullable(),
        State::class => ['status_id', fn(ForeignIdColumnDefinition $definition) => $definition->after('id')]
    ]);
});
```

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -886,6 +886,55 @@ class Blueprint
         ]));
     }
 
+    /**
+     * Creates a collection of foreign ID columns for the given array of models. A convenient shorthand for when you
+     * need to reference a number of foreign models from this table. Where necessary, you may also specify the
+     * column name, and / or a closure to process the resulting ForeignIdColumnDefinition.
+     *
+     * E.g. Array of models - conventional column names will be used.
+     * $table->foreignIdsFor([
+     *      User::model,
+     *      Post::model
+     * ]);
+     *
+     * E.g. Since the method returns a Collection, you can use the HigherOrderCollectionProxy methods to set defaults
+     * that will apply to all columns (individual column definitions can also be configured - see below).
+     * $table->foreignIdsFor([
+     *      User::model,
+     *      Post::model
+     * ])->each->nullable();
+     *
+     * E.g. Use the array key to specify the model, and the value for the column name.
+     *
+     * $table->foreignIdsFor([
+     *      User::model => 'author_id',
+     *      Post::model
+     * ]);
+     *
+     * E.g. Use the array value as a callback function to modify the column definition.
+     *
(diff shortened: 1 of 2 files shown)


taylorotwell:

nwoodhouse:
Thanks for considering my little PR. No problem - I appreciate the need to be conservative about additions. I enjoyed learning the process of putting together a contribution :-).

PR #40731, opened 2022-01-31
[9.x] Remove redundant Doctrine method on schema builder

bakerkretzmar:
Follow-up to #40119 and #40303. Removes `Schema::registerCustomDoctrineType()`, the correct way to register custom types is with `DB::registerDoctrineType()` or in the config, and if for some reason someone still wants to do it this way they can use `Schema::getConnection()->registerDoctrineType()`.

### DIFF
--- a/src/Illuminate/Database/Schema/Builder.php
+++ b/src/Illuminate/Database/Schema/Builder.php
@@ -385,19 +385,6 @@ class Builder
     }
 
     /**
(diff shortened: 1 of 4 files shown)


driesvints (approved):

taylorotwell:
Do we need a PR to the 9.x upgrade guide for this? (`laravel/docs` master branch)

PR #40732, opened 2022-02-01
[10.x] Fix ensure that the visibility of the local driver is taken into account

nshiro:
When uploading files, the permissions of the directory to be created have changed after updating laravel 9.x.
I'm setting `visibility` to `public`.
I assume this is a bug.

2022-02-01_15h22_32

I was not so sure how to make tests, so I didn't.

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemManager.php
+++ b/src/Illuminate/Filesystem/FilesystemManager.php
@@ -169,7 +169,8 @@ class FilesystemManager implements FactoryContract
     public function createLocalDriver(array $config)
     {
         $visibility = PortableVisibilityConverter::fromArray(
-            $config['permissions'] ?? []
+            $config['permissions'] ?? [],
+            $config['visibility'] ?? Visibility::PRIVATE
         );
 
         $links = ($config['links'] ?? null) === 'skip'


GrahamCampbell:
Did you mean to send this to 9.x. You have targetted master, which is 10.x.

driesvints:
@nshiro please sent this to 9.x

dyanakiev:
@driesvints i'm having the same issue, even after adding the below config to the disk, the folder is still created with wrong permissions with is strange its created with 755 instead of 775.

```php
            'permissions' => [
                'file' => [
                    'public' => 0664,
                    'private' => 0664,
                ],
                'dir' => [
                    'public' => 0775,
                    'private' => 0775,
                ],
            ],
```

driesvints:
Please send this to 9.x

driesvints:
@dyanakiev actually this is already fixed on 9.x

dyanakiev:
> @dyanakiev actually this is already fixed on 9.x

Yes i noticed that now, but still it is creating the folders/files with other permissions than the defined ones.

dyanakiev:
@driesvints sorry, i'm on beta-5 and i guess the fix is not included

driesvints:
Will be in Tuesday's Laravel 9 release.

PR #40733, opened 2022-02-01
[9.x] Add .gitattributes to components

driesvints:
When I sent in #39318 I forgot to add a .gitattributes file to each component to filter out the .github directory when installing the components separately.


GrahamCampbell:
I don't think the .github folder is required in the components?

GrahamCampbell (approved):

PR #40734, opened 2022-02-01
[9.x] Respect the visibility of the local driver

nshiro:
When uploading files, the permissions of the directory to be created have changed after updating to laravel 9.x.
I'm setting `visibility` to `public`.
I assume this is a bug.

2022-02-01_15h22_32

I was not so sure how to make tests, so I didn't.

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemManager.php
+++ b/src/Illuminate/Filesystem/FilesystemManager.php
@@ -169,7 +169,8 @@ class FilesystemManager implements FactoryContract
     public function createLocalDriver(array $config)
     {
         $visibility = PortableVisibilityConverter::fromArray(
-            $config['permissions'] ?? []
+            $config['permissions'] ?? [],
+            $config['visibility'] ?? Visibility::PRIVATE
         );
 
         $links = ($config['links'] ?? null) === 'skip'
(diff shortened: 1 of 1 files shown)


driesvints (approved):
Okay, I think this seems correct. Thanks

src/Illuminate/Filesystem/FilesystemManager.php
```diff
@@ -169,7 +169,8 @@ protected function callCustomCreator(array $config)
     public function createLocalDriver(array $config)
     {
         $visibility = PortableVisibilityConverter::fromArray(
-            $config['permissions'] ?? []
+            $config['permissions'] ?? [],
+            $config['visibility'] ?? Visibility::PRIVATE
```
driesvints, line 173: Can you add a new `directory_visibility` config option as well that falls back on `visibility`? That way people can fine-grain permissions. `directory_visibility` conforms with Flysystem's own new option for it: https://github.com/thephpleague/flysystem/blob/3.x/src/Config.php#L12

```suggestion
            $config['directory_visibility'] ?? $config['visibility'] ?? Visibility::PRIVATE
```

driesvints:
Feel free to mark as ready when done.

driesvints:
I've added a new `directory_visibility` as well that can be used to fine-grain permissions for directories if people want that. This PR will still make it so it's BC with Laravel 8.

src/Illuminate/Filesystem/FilesystemManager.php
driesvints: Nvm, added this myself.

PR #40735, opened 2022-02-01
[9.x] Support `callables` in `value()` helper

stevebauman:
## Description

This PR provides the ability to provide callables in the `value()` helper. Ex:

```php
value([$object, 'method'], 'arg');
```

This adds some additional simplicity when allowing various data types in method parameters:

```php
value($var); // $var
value(fn () => true); // true
value([$object, 'method']); // method return
```

### DIFF
--- a/src/Illuminate/Collections/helpers.php
+++ b/src/Illuminate/Collections/helpers.php
@@ -184,6 +184,14 @@ if (! function_exists('value')) {
      */
     function value($value, ...$args)
     {
-        return $value instanceof Closure ? $value(...$args) : $value;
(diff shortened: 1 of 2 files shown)


nunomaduro:
I don't think is a good idea to add a breaking change to this function. In addition, you can now do in PHP stuff like this:

```php
value($object->method(...), 'arg');
```

taylorotwell:
Hmm - yeah little scared to break BC on this particular function since it's so widely used.

stevebauman:
Ok thanks for your time! 🙏

PR #38257, opened 2021-08-06
[8.x] Improves type definitions by using generics, generic arrays, callable types, and more

nunomaduro:
This pull request attempts to show what were the pros and cons if we were to introduce **serious strong type definitions** into Laravel's core. Now, so people can have an idea of what these represent in terms of work, this pull request **only** includes:

- Partially migrates the `Enumerable` interface.
- Test `Enumerable` interface type definitions.
- Adds CI job that to run type definitions tests.

In terms of advantages, we would of course offer much better **native** support for [Psalm](https://psalm.dev/) and [PHPStan](https://phpstan.org/). And also, we would have better **native** auto-completion in editors like [PHPStorm](https://www.jetbrains.com/phpstorm):

**Before:**
before

**After:**
after

In addition, functions like `tap`, `collect`, `retry`,  and more, would **100% static analyzable** and understood by static analysis tools.

Now, in terms of disadvantages, it seems that PHPStorm support for generics is very **primitive**, so there are a few cases where type inference does not work as expected - like VSCode and TypeScript:
```php
$users->each(function ($user) {
    $user->; // $user is mixed...
});
```

Another disadvantage is the fact that not everyone is comfortable with this type of annotations. So doing it right - for new/existing contributors - **may be a friction point**.

Finally, the goal of this pull request is not to have this pull request merged, but rather debate if we plan to have this **improved type system** across the entire code base or not.

PR #40736, opened 2022-02-01
Add generic hints to Cache

AJenbo:


driesvints:
Please send this to 9.x, thanks.

AJenbo:
Here you go: https://github.com/laravel/framework/pull/40759

I could also just have updated this PR, maybe a tad aggressive to close it, but I guess it keeps a clean house :)

PR #40738, opened 2022-02-01
[9.x] Add @pushOnce and @prependOnce

calebporzio:
## Problem
I almost always combine `@push` and `@once` and writing both out feels verbose:

```html
@push('scripts')
@once
    <script src="https://cdn.jsdelivr.net/npm/pikaday/pikaday.js"></script>
@endonce
@endpush
```

## Solution
A new directive: `@pushOnce` to combine them both:

```html
@pushOnce('scripts')
    <script src="https://cdn.jsdelivr.net/npm/pikaday/pikaday.js"></script>
@endpushOnce
```

(I added `@prependOnce` as well, though I personally never use it)

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesStacks.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesStacks.php
@@ -36,10 +36,9 @@ trait CompilesStacks
      */
     protected function compilePushOnce($expression)
     {
-        $parts = explode(',', $this->stripParentheses($expression));
+        $parts = explode(',', $this->stripParentheses($expression), 2);
 
-        $stack = $parts[0];
-        $id = $parts[1] ?? null;
+        [$stack, $id] = [$parts[0], $parts[1] ?? null];
 
         $id = trim($id) ?: "'".(string) Str::uuid()."'";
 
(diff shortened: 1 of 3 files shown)


Jubeki:
Isn't it better to do it the other way around?
Because `$__env->startPush()` and `$__env->endPush()` wouldn't need to be called if it was already used once.

```html
@once
@push('scripts')
    <script src="https://cdn.jsdelivr.net/npm/pikaday/pikaday.js"></script>
@endpush
@endonce
```
instead of
```html
@push('scripts')
@once
    <script src="https://cdn.jsdelivr.net/npm/pikaday/pikaday.js"></script>
@endonce
@endpush
```

calebporzio:
Yeah, that makes sense

marufsharia:
```php
\\ In your Service Provider PHP file, boot() method:

Blade::directive('pushonce', function ($expression) {
    $var = '$__env->{"__pushonce_" . md5(__FILE__ . ":" . __LINE__)}';

    return "<?php if(!isset({$var})): {$var} = true; \$__env->startPush({$expression}); ?>";
});

Blade::directive('endpushonce', function ($expression) {
    return '<?php $__env->stopPush(); endif; ?>';
});


```

PR #40740, opened 2022-02-01
[9.x] change artisan migrate command option --database to --connection

emyu01:
Benefit to end users
This change better describes the option as referring to the connection to be used rather than the rather confusing name of database.

This will not break existing laravel features but may break some third party packages.


taylorotwell:
No plans to make this breaking change.

PR #40741, opened 2022-02-01
[9.x] Add --view flag to make:component command

calebporzio:
I almost exclusively use anonymous Blade components, so because the `artisan make:component` command generates both a class and a view, I never use it.

This PR adds a `--view` flag to opt-out of generating a class for a component:
Screen Shot 2022-02-01 at 3 59 03 PM

Here is the command when a view already exists:
Screen Shot 2022-02-01 at 3 59 11 PM

Here is the definition of the command (with the `--view` flag added):
Screen Shot 2022-02-01 at 4 01 27 PM

I chose `--view` over `--anonymous`, because it's shorter and easier to type therefore I'm more likely to use it. If writing out the full `--anonymous` was required, I might not use this command at all as generating simple view is fairly easy anyways.

I'm open to opinions.

### DIFF
--- a/src/Illuminate/Foundation/Console/ComponentMakeCommand.php
+++ b/src/Illuminate/Foundation/Console/ComponentMakeCommand.php
@@ -46,6 +46,14 @@ class ComponentMakeCommand extends GeneratorCommand
      */
     public function handle()
     {
+        if ($this->option('view')) {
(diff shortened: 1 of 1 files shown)


SjorsO:
If you only need a view file, copy-pasting it is probably faster than typing this command

PR #40750, opened 2022-02-02
[8.x] Local providers

mhidea:
I've added support for local only providers support.
**local_providers** key will be in **config/app.php** and framework checks if there is any provider there, registers them. Something like require-dev for development.

It happens a lot to me that in development I need to make some changes to framework. For all local only things I need to ckeck isLocal, then register. Here we have in one place and they will not be registered in production.


mhidea:
Along this pr I added [this](https://github.com/laravel/laravel/pull/5793), but I can't fix the style.

PR #40759, opened 2022-02-02
Add generic hints to Cache

AJenbo:

### DIFF
--- a/src/Illuminate/Cache/Repository.php
+++ b/src/Illuminate/Cache/Repository.php
@@ -84,9 +84,11 @@ class Repository implements ArrayAccess, CacheContract
     /**
      * Retrieve an item from the cache by key.
      *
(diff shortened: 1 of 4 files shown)


nunomaduro:
@AJenbo Can you tests to those types, as these ones? https://github.com/laravel/framework/tree/9.x/types.

nunomaduro:
I will check this tomorrow.

nunomaduro:
We are not currently accepting generic annotations on the cache component of Laravel.

AJenbo:
> Sorry for noticing just now, but we are not currently accepting generic annotations on the cache component of Laravel.

why? when?

AJenbo:
@nunomaduro could you please let me know which components it is ok to work on, or how I can find this information? Also do you have any idea of when you would be accepting generics for the cache component?

PR #40765, opened 2022-02-02
[9.x] Close illuminate pull-requests.

lucasmichot:
#39318 and #40733 have introduced `.github/workflows` folders and `.gitattributes` files per sub-repository.

The goal is to auto-close any open PR in one of the associated `illuminate` repositories.

I feel there is something simpler with less maintenance could be possible here.

Each Github action comes for free with the `gh` cli, so we can use it.
In the script file, the repositories for which PR should be close have a `read-only` topic, but this could be anything else.
Of course a `GH_TOKEN` needs to be set with write access to the `illuminate` organisation

The script would run every hour, which is more than enough

cc @driesvints on this one

---

Side note, Github also promotes the org wide setting public repository called `.github`: 
https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/creating-a-default-community-health-file

You could add the script and the workflow fiels in this repository, and you could skip setting the `GH_TOKEN` token; I also feel this would make more sense:

- https://github.com/illuminate/.github/.github/workflows/close-pull-requests.yml
- https://github.com/illuminate/.github/bin/close-pull-requests.sh

And you can also add a bunch of community health files and a global readme, which is great 

cf. https://github.com/laravel/framework/discussions/40766


.github/workflows/close-pull-requests.yml
```diff
@@ -0,0 +1,18 @@
+name: close-pull-requests
+
+on:
+  schedule:
+    - cron: '0 * * * *'
```
lucasmichot, line 5: Running every hour which is more than sufficient

bin/close-pull-requests.sh
```diff
@@ -0,0 +1,26 @@
...
+
+set -e
+
+COMMENT='Thank you for your pull request.
+However, you have submitted this PR on the Illuminate organization which is a read-only sub split of \`laravel/framework\`.
+Please submit your PR on the https://github.com/laravel/framework repository.
+
+Thanks!'
+
+for REPO in $(gh repo list illuminate --public --topic="read-only" --json "nameWithOwner" --jq '.[].nameWithOwner'); do
```
lucasmichot, line 11: `read-only` topic is just an example here, it could be something else

bin/close-pull-requests.sh
```diff
@@ -0,0 +1,26 @@
...
+    for PR in $(gh pr list -R $REPO --json "number" --jq '.[].number'); do
+
+        # Add a comment
+        gh pr comment $PR -R $REPO -b $COMMENT
+
+        # Close the pull request
+        gh pr close $PR -R $REPO
+
+        # Lock the pull request too?
+        # gh api -X PUT repos/$REPO/issues/$PR/lock
```
lucasmichot, line 22: Could nice to have PR locked too here

bin/close-pull-requests.sh
driesvints: What does this mean/do? Is a topic a label?

driesvints: Nvm got it.

driesvints:
I think this is a good idea @lucasmichot but I believe having the separate files is still better to have. With the new way of doing things, people could maybe be waiting an hour on that feedback while right now they'd get the message instantly. I know we could adjust the cron to run every minute or so but to be honest I think having the separate action files is more easily to maintain than a single bash script (I know!). I feel like a dedicated action script per Illuminate repo makes sense. 

Moving the action/script to a separate `.github` repo isn't something we should do I think as that detaches it too much from its purpose.

PR #40768, opened 2022-02-02
refactor(Facade): If authorize method is present in the child class a…

hosseinsheikh-web:
If authorize method is present in the child class and its return false , it will not allow the user access


taylorotwell:

hosseinsheikh-web:
Dear @taylorotwell 
    When we use Facade in our classes, especially controllers, outside the main class, we can control user access, when we use the policy or gate or for any other reason  allow the user to receive the appropriate exception if they do not have access,
  And I'm working on this part of Facade, too, if the user does not access a part of the code, this piece of code will be ignored automatically.

PR #40770, opened 2022-02-02
[8.x] Update Lock.php with comments

imanghafoori1:
The `block` on the `Lock` class method was commented since I found this method a bit hard to understand.


taylorotwell:
I think it's fine as is.

PR #40774, opened 2022-02-03
[9.x] Adds `carry()` method to Enumerable

DarkGhostHunter:
## What?

Allows retrieving items from a Collection until reaching a numeric/countable threshold.

Remember [the `reduceSpread()` example?](https://laravel.com/docs/8.x/collections#method-reduce-spread). Here is with `carry()` instead:

```php
$batch = Image::where('status', 'unprocessed')
    ->lazy()
    ->carry(function ($creditsRemaining, $image) {
        if ($image->creditsRequired() > $creditsRemaining) {
            return false;
        }
        
        return $creditsRemaining - $image->creditsRequired();
    }, $creditsRemaining);
```

And if it `creditsRequired()` was a dynamic property, much, much better:

```php
$batch = Image::where('status', 'unprocessed')->lazy()->carry('credits_required', $creditsRemaining);
```

- It will stop once the threshold is reached, or `false` is returned by the function, avoiding looping over an entire collection (or database).
- It allows to use a function or a simple _dot.value_ and a limit.
- It also works over countable values, like arrays or Collections.
- It supports using an initial value on both syntaxes.

## Why?

It's useful for lazy collections where you want to retrieve data but stop once a certain threshold is reached. Without these methods, some verbose code is needed to replicate the same example.

The easy way is to name the property to carry over, and the threshold to stop.

```php
$images = collect([
    ['name' => 'avatar.jpg', 'kb' => 112], 
    ['name' => 'matrix.jpg', 'kb' => 245], 
    ['name' => 'antman.jpg', 'kb' => 548], 
]) 

$limited = $images->carry('kb', 500);
//    [ 
//        ['name' => 'avatar.jpg', 'kb' => 112], 
//        ['name' => 'matrix.jpg', 'kb' => 245]
//    ]
```

And waaaay better than using even `takeWhile()`;

```php
$cumulative = 0;

$limited = $images->takeWhile(function ($image) use (&$cumulative) {
    $cumulative += $image['kb'];
    
    return $cumulative < 500;
});
```

Additionally, you can set a starting initial value as third parameter.

```php
$images = collect([
    ['name' => 'avatar.jpg', 'kb' => 112], 
    ['name' => 'matrix.jpg', 'kb' => 245], 
    ['name' => 'antman.jpg', 'kb' => 548], 
]);

$cumulative = 0; 

$limited = $images->carry('kb', 500, 100));
```

Finally, it works over countable items, like arrays or other collections.

```php
$numbers = collect([
    ['name' => 'maria', 'lucky_numbers' => [8, 2]],
    ['name' => 'adam',  'lucky_numbers' => [4, 1, 2]],
    ['name' => 'eve',   'lucky_numbers' => [5, 8, 9, 3]],
]);

$limited = $numbers->carry('lucky_numbers', 5));

//    [ 
//        ['name' => 'maria', 'lucky_numbers' => [8, 2]],
//        ['name' => 'adam',  'lucky_numbers' => [4, 1, 2]],
//    ]
```

## Caveats

Since by default it expects data to carry, if the value is `null` or `false` it will stop automatically. For that reason, the developer may want to use a function instead.

```php
$items = collect([1, 2, 3, false , 5]);

$carry = $items->carry(11);

//    [1, 2, 3]
```

Second, it will always get N+1 items. This is because it always need to check the next item value. For example, the fifth value is retrieved since it's the item value that goes above the threshold, signaling to stop.

```php
$items = collect([1, 2, 3, 0, 4, 5]);

$carry = $items->carry(6);

// [1, 2, 3, 0];
```` 

## BC?

Nope, only additive.

## Naming suggestions?

I tried `carryUntil()`, `pour()`, `takeBefore()`, `below()`, and `reduceUntil()`. Suggestions are welcomed.

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -180,31 +180,59 @@ trait EnumeratesValues
     }
 
     /**
-     * Reduce the collection using an aggregate value.
+     * Retrieves items from the collection using an aggregate value.
      *
-     * @param  callable(int|float|\Countable, TValue, TKey): \Countable|int|float|bool|null|void  $callback
-     * @param  \Countable|int|float||null  $initial
+     * @param  callable(mixed, TValue, TKey): mixed  $callback
+     * @param  mixed  $initial
      * @return \Illuminate\Support\Collection<TKey, TValue>
      */
     public function carry(callable $callback, $initial = null)
     {
-        $cumulative = $initial;
-
         $collection = new Collection();
 
         foreach ($this as $key => $value) {
-            $result = $callback($cumulative, $value, $key);
+            $result = $callback($initial, $value, $key);
 
-            if ($result === null || $result === false) {
+            if ($result === false) {
                 break;
             }
 
-            $cumulative = $result;
+            $initial = $result;
 
             $collection->put($key, $value);
         }
 
         return $collection;
+    }
+
+    /**
+     * Retrieves items from the collection until an aggregate value is reached.
+     *
+     * @param  string  $key
+     * @param  int|float|null  $value
+     * @return \Illuminate\Support\Collection<TKey, TValue>
+     */
+    public function carryUntil($key, $value = null)
+    {
+        return $this->carry(function ($carry, $item) use ($key, $value) {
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #40781, opened 2022-02-03
Fix UndefinedDocblockClass error

GeorgII-web:
On a PSALM check, there is an error:
UndefinedDocblockClass - *** - Docblock-defined class, interface or enum named Doctrine\DBAL\Driver\PDOConnection does not exist (see https://psalm.dev/200) DB::getPdo()->lastInsertId());

Fixed to the current Doctrine structure

### DIFF
--- a/src/Illuminate/Support/Facades/DB.php
+++ b/src/Illuminate/Support/Facades/DB.php
@@ -3,7 +3,7 @@
 namespace Illuminate\Support\Facades;
 
 /**
- * @method static \Doctrine\DBAL\Driver\PDOConnection getPdo()
+ * @method static \Doctrine\DBAL\Driver\PDO\Connection getPdo()
  * @method static \Illuminate\Database\ConnectionInterface connection(string $name = null)
  * @method static \Illuminate\Database\Query\Builder table(string $table, string $as = null)
  * @method static \Illuminate\Database\Query\Expression raw($value)


X-Coder264:
[Laravel 8 can work with both Doctrine DBAL 2.13+ and DBAL 3.1+](https://github.com/laravel/framework/blob/v8.82.0/composer.json#L84) so `\Doctrine\DBAL\Driver\PDOConnection` and `\Doctrine\DBAL\Driver\PDO\Connection` are correct return values depending on the DBAL version used. So I guess the docblock should be `\Doctrine\DBAL\Driver\PDOConnection|\Doctrine\DBAL\Driver\PDO\Connection`

GeorgII-web:
> [Laravel 8 can work with both Doctrine DBAL 2.13+ and DBAL 3.1+](https://github.com/laravel/framework/blob/v8.82.0/composer.json#L84) so `\Doctrine\DBAL\Driver\PDOConnection` and `\Doctrine\DBAL\Driver\PDO\Connection` are correct return values depending on the DBAL version used. So I guess the docblock should be `\Doctrine\DBAL\Driver\PDOConnection|\Doctrine\DBAL\Driver\PDO\Connection`

With that kind of docblock:
`@method static \Doctrine\DBAL\Driver\PDOConnection|\Doctrine\DBAL\Driver\PDO\Connection getPdo()`
 PSALM throws an `UndefinedDocblockClass` ERROR again, as if there is only `\Doctrine\DBAL\Driver\PDOConnection`

dennisprudlo:
> as if there is only

and thats correct. You will either have v2 or v3 of doctrine/dbal installed. So you will either have `\Doctrine\DBAL\Driver\PDOConnection` or `\Doctrine\DBAL\Driver\PDO\Connection` but not both. So one of those classes will not exist for your static analysis tool. Still, both versions are supported so its best for you to suppress this warning in your PSALM config

PR #40783, opened 2022-02-03
[9.x] ability to add tags/metadata to Emails/Notifications

kbond:
`symfony/mailer` has a special `TagHeader` and `MetadataHeader` for [adding tags/metadata to emails](https://symfony.com/doc/current/mailer.html#adding-tags-and-metadata-to-emails) in a generic way. All the transports/drivers you document support these with the slight exception of SES: it doesn't have tags in the way Symfony defines them but does have metadata. SES calls these tags so it's a bit confusing. Symfony's SES transport will [support metadata in 6.1](https://github.com/symfony/symfony/pull/45222) but I see you created your own [SESTransport](https://github.com/laravel/framework/pull/40696). I can add support to your transport if desired.

Here's how [tags](https://postmarkapp.com/blog/tags-support-for-smtp#what-are-tags) and [metadata](https://postmarkapp.com/blog/feature-announcement-custom-metadata-for-your-emails) are used with Postmark - all services are similar I believe.

I thought it would be nice to have helpers to add this data in `Mailable`/`MailMessage`:

```php
// in your mailable:
public function build()
{
    return $this->from('example@example.com', 'Example')
                ->tag('shipment')
                ->metadata('order_id', $this->order->id)
                ->view('emails.orders.shipped');
}

// in your notification:
{
    return (new MailMessage)
                ->from('barrett@example.com', 'Barrett Blair')
                ->tag('upvote')
                ->metadata('comment_id', $this->comment->id)
                ->line('...');
}
```

If this is of interest, I can add tests/docs. Maybe you'd want the methods in a trait or done a different way - let me know!

### DIFF
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -874,6 +874,33 @@ class Mailable implements MailableContract, Renderable
     }
 
     /**
+     * Add a tag header to the message when supported by the underlying transport.
+     *
+     * @param  string  $value
+     * @return $this
+     */
+    public function tag($value)
+    {
+        return $this->withSymfonyMessage(function (Email $message) use ($value) {
+            $message->getHeaders()->add(new TagHeader($value));
+        });
+    }
+
+    /**
+     * Add a metadata header to the message when supported by the underlying transport.
+     *
+     * @param  string  $key
+     * @param  string  $value
+     * @return $this
+     */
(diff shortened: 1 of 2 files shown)


driesvints (approved):
Cool, thanks @kbond!

driesvints:
@taylorotwell I think this could indeed be useful for people now that we're integrating directly with Symfony Mailer.

PR #40784, opened 2022-02-03
[8.x] Accept 'on' as a valid boolean

jochensengier:
This PR will accept ``on`` as a valid boolean when the validation rule [boolean](https://laravel.com/docs/8.x/validation#rule-boolean) is set for an input field.
Browsers will sent ``on`` if a checkbox is checked that doesn't have a defined value.

``on`` is already accepted as a valid boolean in the [accepted](https://laravel.com/docs/8.x/validation#rule-accepted) and [acceptedIf](https://laravel.com/docs/8.x/validation#rule-accepted-if) validation rules.

It's also accepted in ``Illuminate\Http\Concerns\InteractsWithInput`` when you do ``request()->boolean('demo')``, as the underlying PHP function ``filter_var($this->input($key, $default), FILTER_VALIDATE_BOOLEAN)`` is called.

According to https://www.php.net/manual/en/filter.filters.validate.php, even ``yes`` should be accepted.

> Returns true for "1", "true", "on" and "yes". Returns false otherwise. 

If accepted, the documentation needs to be updated as well:
https://github.com/laravel/docs/blob/8.x/validation.md#rule-boolean

Looking forward to any feedback and suggestions regarding this PR.
Perhaps ``on`` was left out intentionally in the past for some (edge) cases?

Thanks!

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -420,7 +420,7 @@ trait ValidatesAttributes
      */
     public function validateBoolean($attribute, $value)
     {
-        $acceptable = [true, false, 0, 1, '0', '1'];
+        $acceptable = [true, false, 0, 1, '0', '1', 'on'];
 
         return in_array($value, $acceptable, true);
     }
--- a/tests/Validation/ValidationValidatorTest.php
+++ b/tests/Validation/ValidationValidatorTest.php
@@ -2170,6 +2170,9 @@ class ValidationValidatorTest extends TestCase
 
         $v = new Validator($trans, ['foo' => 0], ['foo' => 'Boolean']);
         $this->assertTrue($v->passes());
+
+        $v = new Validator($trans, ['foo' => 'on'], ['foo' => 'Boolean']);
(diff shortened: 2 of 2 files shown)


driesvints:
I don't think we should merge this. "on" is a specific concept for checkboxes for which `accepted` and acceptedIf` directly were designed for. `boolean` purely handles truthy values, meaning true, false, 1 and 0.

driesvints:
> What if you have an opt-in checkbox (newsletter, gdpr related, ...) that can be both true or false?

Accepted is only designed to check for true. If it's not checked then you should handle the input accordingly.

> Are the validation rules not designed specifically for incoming http requests/input?

They are generic, meaning they can be used for simple data validation or HTTP requests.

jochensengier:
Like you said, they are only designed to check for true, not false. 
Optional checkboxes (without a value) can't be validated with these rules.

After some consideration, I'm going to close this PR.
Saving the validated data directly in a MySQL database would probably throw an error as `on` would not be recognized as a valid `tinyInt`.

It's also a pretty easy workaround, just make sure the checkbox has a value 😅.

Sorry for wasting your time Dries!

driesvints:
@jochensengier don't worry about it 👍

PR #40787, opened 2022-02-03
[9.x] No need to implement an interface twice.

lucasmichot:
There is no need mentioning that a class implements an interface, when its parent class already implements it.


driesvints (approved):

driesvints:
@taylorotwell we could simplify the docs for custom transports as well.

taylorotwell:
@driesvints will do

PR #40792, opened 2022-02-03
[9.x] Remove useless variables

lucasmichot:
In many places in the code are useless variables set.
It sometimes make sense for understanding, but very often it feels like refactoring where those variables have been left untouched.

This PR fixes that.

### DIFF
--- a/src/Illuminate/Auth/Access/Gate.php
+++ b/src/Illuminate/Auth/Access/Gate.php
@@ -481,7 +481,7 @@ class Gate implements GateContract
             $reflection = new ReflectionClass($class);
 
             $method = $reflection->getMethod($method);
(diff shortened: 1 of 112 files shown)


GrahamCampbell:
This adds little value, and is going to introduce more merge conflicts. I don't think we should do this.

rodrigopedra:
Some of those were used to clarify usage before named arguments were added to PHP. Maybe in those cases, change it to named arguments?

I know named arguments are not promised to be consistent in Laravel, but in that case I think these one at least should be kept.

Unused exception variables I am totally in favor for removal.

imanghafoori1:
Nice but a bit too big.

PR #40794, opened 2022-02-03
[9.x] Limit morphTo columns when relationship is added by with() and morph eager loading is used.

ioan-cristea:
Hi, 

When we use with to load a morph relation, and we use eagle relationship loading to relations we receive all columns and we can't limit that, this feature will do the job.

Tests are written.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/MorphTo.php
+++ b/src/Illuminate/Database/Eloquent/Relations/MorphTo.php
@@ -48,6 +48,13 @@ class MorphTo extends BelongsTo
     protected $morphableEagerLoads = [];
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:

ioan-cristea:
Hi, 
This contribution have no place as a package, because it is essential to limit the properties that are been returned when we use "with" and also when we use on polymorphic relationship. 
When we have a simple belongsTo relation we can do something like "with('relation:id,name')", but on polymorphic we need to have diferent properties based on class we will have. Thats when $morphTo->morphColumns(..) is doing the job, where we cand specify different columns to be returned based on the class of polymorphic relation.

PR #40799, opened 2022-02-04
[8.x] Fix ResourceCollection::jsonOptions when it collects Model

bastien-phi:
It appears that some developers are defining `$collects` of `ResourceCollection` with a model class to bypass the auto-discovery of the underlying resource class.

PR #40208 breaks this usage as the model does not have the `jsonOptions` method. (see https://github.com/laravel/framework/pull/40208#issuecomment-1027833288)

This PR fixes it returning 0 as json options when the collected class is not a `JsonResource`.

### DIFF
--- a/src/Illuminate/Http/Resources/CollectsResources.php
+++ b/src/Illuminate/Http/Resources/CollectsResources.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Http\Resources;
 
+use Illuminate\Http\Resources\Json\JsonResource;
 use Illuminate\Pagination\AbstractCursorPaginator;
 use Illuminate\Pagination\AbstractPaginator;
 use Illuminate\Support\Collection;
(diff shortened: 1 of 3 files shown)


driesvints:
@taylorotwell this is actually based on misusage of the `$collects` property where the OP of the linked PR used a model instead of a resource. Should we look into throwing an exception to warn people against this?

ninjami-matti:
In the `CollectsResource` trait, that `ResourceCollection` class uses, `collectResource` method does this:

```php
$collects = $this->collects();

$this->collection = $collects && ! $resource->first() instanceof $collects
    ? $resource->mapInto($collects)
    : $resource->toBase();
```

It checks and allows a `Model` class to go through, even when it is defined explicitly in the `$collects`. Maybe an exception of some kind should be thrown here?

taylorotwell:
IMO throwing an error is fine - hiding it via this PR could just create more confusion down the line.

driesvints:
@bastien-phi @ninjami-matti any of you two willing to provide that PR? Otherwise I can check in on that somewhere next week.

Issue #31345, opened 2020-02-04
Redis cache store issues on numeric values


halaei:
- Laravel Version:  6.13.1
- PHP Version: 7.2.27
- Database Driver & Version: Redis

### Description:
`RedisStore::serialize()` and `unserialize()` don't do right when it comes to caching numbers. I think the code has some considerations that Redis stores strings that are integers more efficient than normal integers. But the implementation is wrong.
From https://redis.io/commands/INCR:
> Redis stores integers in their integer representation, so for string values that actually hold an integer, there is no overhead for storing the string representation of the integer.

### Steps To Reproduce:
Set cache driver to redis and try the followings:
```php
Cache::put('test', '1');
Cache::get('test');; // expected = actual = "1"
Cache::put('test', 1);
Cache::get('test'); // expected: 1, actual "1"
Cache::put('test', 1.0);
Cache::get('test'); // expected: 1.0, actual "1"
Cache::put('test', INF)
Cache::get('test'); // expected INF, actual: PHP Notice:  unserialize(): Error at offset 0 of 3 bytes
Cache::put('test', -INF)
Cache::get('test'); // expected -INF, actual: PHP Notice:  unserialize(): Error at offset 0 of 4 bytes
Cache::put('test', NAN);
Cache::get('test');// expected: NAN, actual: PHP Notice:  unserialize(): Error at offset 0 of 3 bytes 
```
I don't know if other drivers have similar issues.

driesvints:
Managed to reproduce this. Sending in a fix. Also not really sure about other stores.

halaei:
Should the complete fix be for the next major release, as considering returning 1 instead of "1" be a possible breaking change?

driesvints:
@halaei I guess this is now completely broken atm and never worked? I sent in a fix here: https://github.com/laravel/framework/pull/31348

Let me know what you think.

carestad:
Came across this today when we tried switching to redis. Some of the stuff in our cache is type sensitive, so when some of the cache keys went from `1` to `"1"` the `===` checks failed.

It seems this is sort of broken for all numeric strings, since `is_numeric()` is being used, and then skips serializing it. Why not just consistently serialize everything before storing it, like in the file and memcached stores? A quick test shows me that the outcome of `Cache::put('test', 123)` with the cache drivers `file`, `array`, `database`, `apc` and `memcached` all return an integer, but for `redis` I get a string.

halaei:
[...]

PR #40801, opened 2022-02-04
[10.x] Fix Redis store set int value, but get returns string

twinh:
Fix the issues in #31345

```php
Cache::put('test', 1);
Cache::get('test'); // expected: 1, actual: "1"
Cache::put('test', 1.0);
Cache::get('test'); // expected: 1.0, actual: "1"

//  throws an error on the second call
function foo(): int
{
    return Cache::remember('foo', 60, function() {
        return 1;
    });
}
```

## Implement 

- Store raw value of int and float to Redis
- Use "." to distinguish int and float
- Serialize other type(including string) before store to Redis

## BREAKING CHANGES

- Reading existing numeric values in Redis will be converted from string to int or float
- Increment new stored string value wont work, use int instead

```php
Cache::set('test', '1'); // store "s:1:"1";" in Redis
Cache::increment('test');  // (error) ERR value is not an integer or out of range
```

To avoid breaking changes, we can add option like `keepNumeric`/`keepType` ? to enable this feature.

### DIFF
--- a/src/Illuminate/Cache/RedisStore.php
+++ b/src/Illuminate/Cache/RedisStore.php
@@ -335,11 +335,11 @@ class RedisStore extends TaggableStore implements LockProvider
             return $value;
         }
 
-        if (is_float($value) && !in_array($value, [\INF, -\INF], true) && !is_nan($value)) {
+        if (is_float($value) && ! in_array($value, [\INF, -\INF], true) && ! is_nan($value)) {
             // Note that float like "1.0" will be converted to string "1"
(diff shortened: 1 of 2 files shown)


taylorotwell:

AndrewMast:
Then you can open a GitHub [discussion](https://github.com/laravel/framework/discussions) or maybe start a conversation on the laravel discord (https://discord.com/invite/laravel). Discussions on closed PRs rarely are responded to by the maintainers.

PR #40802, opened 2022-02-04
[8.x] Add tests for put method of FileStore

imanghafoori1:
This adds a test for some important aspects of the "put" method:
1- it can handle space and slash characters in the cache key.
2- it locks the file exclusively while storing the data.
3- it Considers 0 seconds as forever.

I did not build the file path myself, since I believe the way it is built is an implementation detail and is irrelevant to the test or at least this particular test.
Being able to accept spaces and slashes in the key without affecting the path is what really matters.

Currently, the details of the path method are copy/pasted in many tests and that does not make sense. If this is accepted we can remove the implementation details of $path calculation from the other tests without degrading the tests.

Replace:
```php
        $hash = sha1('foo');
        $cache_dir = substr($hash, 0, 2).'/'.substr($hash, 2, 2);
...
       ->with($this->equalTo(__DIR__.'/'.$cache_dir.'/'.$hash), ...

```
with:
```php
       ->with($this->matchesRegularExpression('/.*/'), ...
```
So that other tests can focus on what they are really testing.

- This way you may switch to a faster hashing algorithm in PHP 8.1 from `sha1` without tampering with the all tests.

### DIFF
--- a/tests/Cache/CacheFileStoreTest.php
+++ b/tests/Cache/CacheFileStoreTest.php
@@ -62,6 +62,49 @@ class CacheFileStoreTest extends TestCase
         $this->assertTrue($result);
     }
 
+    public function testGoodPathForFilesIsCalculated()
+    {
+        $files = new Filesystem();
+        $cache = new class($files, __DIR__) extends FileStore
+        {
+            public function path($key)
+            {
+                return parent::path($key);
+            }
+        };
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think it's fine.

imanghafoori1:
@taylorotwell You said that you do not want to accept more mocked tests anymore.

PR #40803, opened 2022-02-04
[9.x] Return status code on `view:cache` error

JayBizzle:
Hi,

We have run into a few situations where deployments have completed successfully, even though one of the `artisan` commands has failed.

I don't know if there is a reason these commands don't already return a status (`artisan:up` and `down` already return status codes), hence I am submitting this PR as a `draft` just to get feedback on why this may be the case.

I have only edited the `view:cache` command at the moment, but other artisan commands that potentially should return a status code are...

- config:cache
- config:clear
- event:cache
- event:clear
- optimize:cache
- optimize:clear
- route:cache
- route:clear
- view:cache
- view:clear

Looking forward to hearing your thoughts 👍

### DIFF
--- a/src/Illuminate/Foundation/Console/ViewCacheCommand.php
+++ b/src/Illuminate/Foundation/Console/ViewCacheCommand.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Foundation\Console;
 
+use Exception;
 use Illuminate\Console\Command;
 use Illuminate\Support\Collection;
 use Symfony\Component\Finder\Finder;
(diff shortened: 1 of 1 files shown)


driesvints:
@JayBizzle seems like tests are failing here. Please also remember that draft PR's aren't reviewed.

JayBizzle:
Im pretty sure that failing test is not to do with this PR 😕 

How should I proceed to get feedback, complete all the work I deem required, then remove the draft status?

I was cautious about submitting this PR as it seems (to me) an obvious addition, so much so, it leads me to believe there would be some reason it hasn't been implemented in the past 🤔

axlon:
@JayBizzle shouldn't any artisan command already return a faulty status code when it throws an exception that is not caught? 

I tested this on a fresh Laravel 8 application (by making sure it couldn't write to the view cache directory) and it returned unsuccessfully as expected.

The code responsible for this is found here:
https://github.com/laravel/framework/blob/c7f4c23e17013b9c6f86a8fd1d37a590e2ad29b4/src/Illuminate/Foundation/Console/Kernel.php#L124-L137

JayBizzle:
@axlon thank for that info. I'll have to do more testing. Our deployment script didn't fail after an unsuccessful view:cache call. Will have to do more digging 👍🏻

PR #40805, opened 2022-02-04
[9.x] Add isolatedTransaction() method for transactions

reziamini:
Now we can't use a specific isolation level for our transaction.
We can just set a default isolation level in our configs.
With this new `isolatedTransaction()` method, we can easily pass our isolation level without touching config files and customize them.

These new methods don't have any breaking changes, and they work separately.

Also, SQLite cannot support ISOLATION LEVEL unlike SQL Server, MySQL, and PostgreSQL.

Because of that, we skip the isolation level for SQLite, but if we use `SQLite` in our test environment everything works properly!

Example: 
```php

DB::isolatedTransaction('SERIALIZABLE', function(){
    DB::table('users')->update([
        'transaction_count' => $user->transaction_count++,
    ]);
    
    DB::table('transaction')->insert([
        'user_id' => 1,
        'price' => 999.99
    ]);

    return DB::table('transactions')->count('id');
});

```

There are some documentations to set the isolation level for the current transaction:
[SQL Server](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-transaction-isolation-level-transact-sql?view=sql-server-ver15) 
[MySQL](https://dev.mysql.com/doc/refman/5.6/en/set-transaction.html)
[PostgreSQL](https://www.postgresql.org/docs/9.3/sql-set-transaction.html)

### DIFF
--- a/src/Illuminate/Database/Concerns/ManagesTransactions.php
+++ b/src/Illuminate/Database/Concerns/ManagesTransactions.php
@@ -65,6 +65,23 @@ trait ManagesTransactions
     }
 
     /**
+     * Execute a transaction using isolation level
+     *
+     * @param  string  $isolationLevel
+     * @param  \Closure  $callback
+     * @param  int  $attempts
+     * @return mixed
+     *
+     * @throws \Throwable
+     */
+    public function isolatedTransaction($isolationLevel, Closure $callback, $attempts = 1)
+    {
+        $this->prepareIsolationLevel($isolationLevel);
+
+        return $this->transaction($callback, $attempts);
+    }
+
+    /**
      * Handle an exception encountered when running a transacted statement.
      *
      * @param  \Throwable  $e
@@ -330,5 +347,23 @@ trait ManagesTransactions
         }
 
         throw new RuntimeException('Transactions Manager has not been set.');
+    }
+
+
+    /**
+     * Set isolation level for current transaction
+     *
+     * @param string $isolationLevel
+     * @return void
+     */
+    protected function prepareIsolationLevel($isolationLevel)
+    {
+        if($this->getDriverName() == 'sqlite'){
+            return;
+        }
+
+        $query = $this->getQueryGrammar()->compileIsolationLevel($isolationLevel);
+
+        $this->getPdo()->prepare($query)->execute();
     }
 }
--- a/src/Illuminate/Database/Concerns/ManagesTransactions.php
+++ b/src/Illuminate/Database/Concerns/ManagesTransactions.php
@@ -65,7 +65,7 @@ trait ManagesTransactions
     }
 
     /**
-     * Execute a transaction using isolation level
+     * Execute a transaction using isolation level.
      *
      * @param  string  $isolationLevel
      * @param  \Closure  $callback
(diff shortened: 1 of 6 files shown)


imanghafoori1:
Provide a sample usage of new methods

rezaamini-ir:
@imanghafoori1 Edited.

imanghafoori1 (commented):
Nice

src/Illuminate/Database/Concerns/ManagesTransactions.php
```diff
@@ -331,4 +348,21 @@ public function afterCommit($callback)
...
+
+    /**
+     * Set isolation level for current transaction.
+     *
+     * @param  string  $isolationLevel
+     * @return void
+     */
+    protected function prepareIsolationLevel($isolationLevel)
+    {
+        if ($this->getDriverName() === 'sqlite') {
```
coderkoala, line 360: Instead of bailing only in case of sqlite, wouldn't it be smarter to implement a switch case, with fallback also returning in case of an unknown database driver being used?

I may be wrong, but some other database driver could pose a problem that doesn't support the feature might break functionality(although it'd entirely be the implementer's fault). I'm well aware Laravel Officially only support MySQL, PG, MSSQL and SQLite. Within official support, though, the argument is moot.

What are your thoughts on it?

rezaamini-ir:
image

@taylorotwell It will just affect on current transaction.

The photo comes from Postgres doc, but in other databases, it's the same as Postgres.

src/Illuminate/Database/Concerns/ManagesTransactions.php
rezaamini-ir: I guess with the new commit, it got fixed.

taylorotwell:
But isn't this statement run before the transaction has started in your PR? What transaction is it affecting?

rezaamini-ir:
@taylorotwell 
Here is an example of how does it work:
When we want to change the isolation level we have to first execute this SQL code:
```sql
SET TRANSACTION ISOLATION LEVEL SERIALIZABLE;
```

Now, we can begin our transaction:
```sql
BEGIN;

   SELECT `whatever` FROM `anywhere`

COMMIT;
```

And when we call `isolatedTransaction` method, it will first execute the first query, then start the transaction.
And the first query only affect the coming transaction.

taylorotwell:
From Postgres docs:

> If SET TRANSACTION is executed without a prior START TRANSACTION or BEGIN, it emits a warning and otherwise has no effect.

rezaamini-ir:
Oh, that's right.

It was my mistake with Postgres.

We need to use `start` instead of `set` to change the incoming transaction isolation in Postgres to not affect other transactions.

Here is an example with Postgres queries:

```sql

START TRANSACTION ISOLATION LEVEL SERIALIZABLE;

BEGIN;
	SHOW TRANSACTION ISOLATION LEVEL; // serializable
COMMIT;

START TRANSACTION ISOLATION LEVEL READ UNCOMMITTED;

BEGIN;
	SHOW TRANSACTION ISOLATION LEVEL; // read uncommited
COMMIT;

BEGIN;
	SHOW TRANSACTION ISOLATION LEVEL; // default level: read committed
COMMIT;
```

I will send new commits to fix this out.

taylorotwell:
This still isn't quite right. You are now starting 2 transactions on Postgres.

PR #40813, opened 2022-02-04
[10.x] Improves `Support\Collection` reduce method type definition

fdalcin:
Following the improvements made by @nunomaduro on #38538, this pull request updates the type definition of `reduce` and `reduceWithKeys` methods to consider the `$callback` key parameter.

Whenever the reduce callback expects the third parameter static analysis fail because the type is not defined which produces the following error.
image

### DIFF
--- a/src/Illuminate/Collections/Enumerable.php
+++ b/src/Illuminate/Collections/Enumerable.php
@@ -805,7 +805,7 @@ interface Enumerable extends Arrayable, Countable, IteratorAggregate, Jsonable, 
      * @template TReduceInitial
      * @template TReduceReturnType
      *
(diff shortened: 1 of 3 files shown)


jamespavett:
Sorry if this isn't the right place to ask. I noticed this was merged into master, was this not a candidate for inclusion in the latest 9.x release?

driesvints:
@fdalcin it seems you indeed sent this to master instead of 9.x. Can you send this in to 9.x as well?

fdalcin:
Hi, yes I can make a new PR sending this to 9.x. Thought I've done that, sorry about it.

PR #40569, opened 2022-01-23
[8.x] Add phpredis serialization and compression support to the cache component

TheLevti:
This is a follow up of https://github.com/laravel/framework/pull/36412 and https://github.com/laravel/framework/pull/40282 which added phpredis serialization and compression support to the `redis` component.

### The Problem

Laravel's Cache component is built on top of the Redis component. The Cache component serializes on its own before sending data to the Redis component, which is a problem when we use `phpredis` with serialization enabled.

### The Solution

When `phpredis` is used together with serialization enabled, skip the serialization part on the Cache component.

### Details

This pull request adds these capabilities now to the `cache` component as there is some custom logic sitting on top of the `redis` component. The changes include prevention of double serialization/deserialization, proper eval argument packing and a reworked phpredis test suite, which lets all test that use redis run with all kind of different connection setups using data providers.

**Note: I explicitly skipped also upgrading the `queue` component in this pull request. A `queue` component upgrade pull request will follow after this one. Until then redis queues do not support phpredis compression and serialization functionality yet. I was almost done with also the queue adjustments, but noticed that we do some LUA json decoding/encoding to increment the attempt counter, which I would need to move to a different level as phpredis will serialize/compress with different set of algorithms. Can you please suggest someone who has in depth knowledge of the php redis queue implementation in laravel as I have some concrete questions regarding doing changes there.**

PR #40816, opened 2022-02-04
[8.x] Improve redis test suite

TheLevti:
This is a follow up of https://github.com/laravel/framework/pull/40569, splitting the original pull request up. This pull request is about refining the existing redis test suite making it easy to test future changes related to redis.

### What changed

- Added data providers with named data sets to allow easy testing of different redis connection configurations.
- All tests in all components that use redis are now using the new data providers to test all supported redis connection settings separately making it easy to pinpoint which redis setting causes an issue with an existing or new component.

If any test related to redis fails, it will now output the exact data set that caused the failure:

image

### How to use

Now when you need to test something where redis is used, you can simply add the helper trait in the beginning like so:

```php
use InteractsWithRedis;
```

Then to let a test run through all possible redis variations, add one of the following doc blocks:

```php
/**
 * Runs this test with only two basic redis connections where one is using the predis driver and
 * the other the phpredis driver.
 *
 * @dataProvider redisConnectionDataProvider
 */
public function testSomethingRelatedToRedis($connection)
{
    $this->app['redis'] = $this->getRedisManager($connection);
    ...
}
```

```php
/**
 * Runs this test with the basic connection set and in addition advanced connection configurations like
 * persistent connections, prefix, scan options, serialization, compression etc.
 *
 * @dataProvider extendedRedisConnectionDataProvider
 */
public function testSomethingRelatedToRedis($connection)
{
    $this->app['redis'] = $this->getRedisManager($connection);
    ...
}
```

To get a custom redis connection during a test you can also pass two additional parameters to the `getRedisManager` helper method where the second parameter is the driver and third parameter an connection config array:

```php
$redisManager = $this->getRedisManager('predis_custom', 'predis', [
    'cluster' => false,
    'options' => [
        'prefix' => 'test_',
    ],
    'default' => [
        'url' => "redis://{$host}:{$port}",
        'database' => 5,
        'timeout' => 0.5,
    ],
]);
```

To properly clean up after every test provide the following method call in your tearDown method:

```php
protected function tearDown(): void
{
    $this->tearDownRedis();

    parent::tearDown();
}
```

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/InteractsWithRedis.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/InteractsWithRedis.php
@@ -2,15 +2,18 @@
 
 namespace Illuminate\Foundation\Testing\Concerns;
 
-use Exception;
 use Illuminate\Foundation\Application;
 use Illuminate\Redis\RedisManager;
+use Illuminate\Support\Collection;
 use Illuminate\Support\Env;
+use Redis;
+use Throwable;
+use UnexpectedValueException;
 
 trait InteractsWithRedis
 {
     /**
-     * Indicate connection failed if redis is not available.
+     * Indicates connection failed if redis is not available.
      *
      * @var bool
      */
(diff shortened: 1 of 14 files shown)


src/Illuminate/Foundation/Testing/Concerns/InteractsWithRedis.php
```diff
@@ -21,93 +24,344 @@ trait InteractsWithRedis
...
+    /**
+     * Initializes a new RedisManager with the given driver and config.
+     *
+     * @param  string  $driver
+     * @param  array  $config
+     * @return \Illuminate\Redis\RedisManager
+     */
+    private function initializeRedisManager($driver, $config)
+    {
+        $app = $this->app ?? new Application();
```
crynobone, line 341: Why do we need `new Application()` here?

TheLevti: First of all it was like that before my changes and second, the manager requires an app instance.

TheLevti: Why we create a new one? Some Integration tests do not use the Laravel test case and there we need to create one as the class prop is not set in that case.

taylorotwell:
Can you send this to `9.x` branch?

PR #40821, opened 2022-02-05
Command for dump current config

Davidnadejdin:
This command can be useful when debugging, for example, to get the current config values from the cache

Usage:
```
php artisan config:dump 
```
```
+-----------------------------------------------------------------+------------------------------------------------------------------+
| Key                                                             | Value                                                            |
+-----------------------------------------------------------------+------------------------------------------------------------------+
| app.name                                                        | Laravel                                                          |
| app.env                                                         | local                                                            |
| app.debug                                                       | 1                                                                |
....
| auth.passwords.users.provider                                   | users                                                            |
| auth.passwords.users.table                                      | password_resets                                                  |
| auth.passwords.users.expire                                     | 60                                                               |
| auth.passwords.users.throttle                                   | 60                                                               |
| auth.password_timeout                                           | 10800                                                            |

```
```
php artisan config:dump app
```
```
+-----------------------------------------------------------------+------------------------------------------------------------------+
| Key                                                             | Value                                                            |
+-----------------------------------------------------------------+------------------------------------------------------------------+
| app.name                                                        | Laravel                                                          |
| app.env                                                         | local                                                            |
| app.debug                                                       | 1                                                                |

```
```
php artisan config:dump app.name
```
```
Laravel
```

PR #40824, opened 2022-02-05
[9.x] Command for dump current config values

Davidnadejdin:
#40821

### DIFF
--- a/src/Illuminate/Foundation/Console/ConfigDumpCommand.php
+++ b/src/Illuminate/Foundation/Console/ConfigDumpCommand.php
@@ -12,7 +12,7 @@ class ConfigDumpCommand extends Command
     /**
      * @var string
      */
-    protected $signature = 'config:dump {id?}';
+    protected $signature = 'config:dump {key?}';
 
     /**
      * @var string
@@ -30,7 +30,7 @@ class ConfigDumpCommand extends Command
 
     public function handle(): int
     {
-        $rawConfig = $this->argument('id') ? $this->config->get($this->argument('id')) : $this->config->all();
+        $rawConfig = $this->argument('key') ? $this->config->get($this->argument('key')) : $this->config->all();
 
         if (is_array($rawConfig)) {
             $rows = Collection::make(Arr::dot($rawConfig));
--- a/src/Illuminate/Foundation/Console/ConfigDumpCommand.php
+++ b/src/Illuminate/Foundation/Console/ConfigDumpCommand.php
@@ -1,0 +1,54 @@
+<?php
+
+namespace Illuminate\Foundation\Console;
+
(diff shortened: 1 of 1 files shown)


src/Illuminate/Foundation/Console/ConfigDumpCommand.php
```diff
@@ -0,0 +1,54 @@
...
+                    $key,
+                    is_array($value) ? '' : $value,
+                ];
+            });
+
+            $this->table(
+                ['Key', 'Value'],
+                $rows,
+            );
+        } elseif (is_string($rawConfig)) {
```
rezaamini-ir, line 48: I guess it's better to refactor in this way:

```php
if(is_string($rawConfig)){
    $this->info($rawConfig);
    return 0;
}

// Handle if it is an array
```

It's much more cleaner.

Davidnadejdin: but then we will duplicate the same return. This is a too small command

driesvints:
I think we shouldn't do this. Dumping config values onto the CLI leaves you to the risk of exposing sensitive info in logs like the APP_KEY env variable.

Davidnadejdin:
> I think we shouldn't do this. Dumping config values onto the CLI leaves you to the risk of exposing sensitive info in logs like the APP_KEY env variable.

command output can get into the log?

src/Illuminate/Foundation/Console/ConfigDumpCommand.php
```diff
@@ -0,0 +1,54 @@
...
+use Illuminate\Console\Command;
+use Illuminate\Support\Arr;
+use Illuminate\Support\Collection;
+
+class ConfigDumpCommand extends Command
+{
+    /**
+     * @var string
+     */
+    protected $signature = 'config:dump {id?}';
```
lkmadushan, line 15: maybe we should use `key` is here?

Davidnadejdin: Maybe. But i focused on it:
https://github.com/php-fig/container/blob/master/src/ContainerInterface.php#L22

axlon:
My 2 cents: I don't think this adds a lot of value since Laravel already ships with tinker and its not very difficult or cubersome to access your config through tinker when debugging.

Davidnadejdin:
> My 2 cents: I don't think this adds a lot of value since Laravel already ships with tinker and its not very difficult or cubersome to access your config through tinker when debugging.

you can also get a list of routes
```
Illuminate\Support\Facades\Route::getRoutes();
```

config:dump is small helper command, not a big feature 🙂

taylorotwell:

PR #40825, opened 2022-02-05
[8.x] Add test for tags method on cache Repository

imanghafoori1:
The `tags` method on the `Repository` class is not tested very well.
These tests avoid mocking to be easy to understand and avoid mockery magics.


GrahamCampbell:
I think 9.x would be a better target branch for this.

PR #40829, opened 2022-02-06
[8.x] Don't add duplicate callbacks

telkins:
While troubleshooting a semi-related problem with tests, Livewire, and the `TrimStrings` middleware, I noticed that while running a batch of tests, callbacks will "stack up"....that is, they will keep being added.

I'm not sure that this behavior is necessarily bad, and it may actually be a good thing under some circumstances...I'm not sure.  But, it seemed to me that it would generally be considered something worth avoiding.

To test this, I simply dumped the count of `static::$skipCallbacks` at the end of `skipWhen()` and, when running my tests normally, the value just continues to increase.

So...I put this little change in anticipating that it might make tests a little more efficient.  🤓 

Please let me know if there are any questions/concerns...or let me know if I've missed something.  😜 

Thx.  🚀

### DIFF
--- a/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
+++ b/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
@@ -64,6 +64,8 @@ class TrimStrings extends TransformsRequest
      */
     public static function skipWhen(Closure $callback)
     {
-        static::$skipCallbacks[] = $callback;
+        if (! in_array($callback, static::$skipCallbacks)) {
+            static::$skipCallbacks[] = $callback;
+        }
     }
 }


taylorotwell:
I don't think this is going to work if each time we pass a callback it's actually a new object instance though?

CleanShot 2022-02-07 at 10 34 26@2x

Please mark as ready for review again if I'm missing something.

telkins:
> I don't think this is going to work if each time we pass a callback it's actually a new object instance though?
> 
> CleanShot 2022-02-07 at 10 34 26@2x
> 
> Please mark as ready for review again if I'm missing something.

I actually tested it differently 😅 :
image

But, I can see that it's more likely to be used the way you tested it, so I'm not sure how valuable this change actually is.  🤔 

I'm going to mark it as Ready for Review **just in case** you see value for this after all.  It's probably not going to make much of a difference and I think the bigger problem is when developers use `skipWhen()` incorrectly...stacking duplicate callbacks.

Thanks for your time/attention.  🚀

X-Coder264:
The proper fix for this would probably be to add a static method in that class which empties out that static array which can be called as part of the `tearDown` method of `\Illuminate\Foundation\Testing\TestCase`.

taylorotwell:
Agree with @X-Coder264

dennisprudlo:
@telkins I think the addressed class is `TrimStrings`. If we had a `TrimStrings::clearSkipCallbacks()` function that would clear `TrimStrings::$skipCallbacks`, we could clear them after each test.

telkins:
> @telkins I think the addressed class is `TrimStrings`. If we had a `TrimStrings::clearSkipCallbacks()` function that would clear `TrimStrings::$skipCallbacks`, we could clear them after each test.

Thx, @dennisprudlo .  That's what I thought, initially, but I wasn't 100% sure because of the way it was worded and also because I thought Taylor might suggest that a separate PR with such a solution be contributed if that was what was meant...but he didn't.  And...I didn't want to second-guess Taylor.  😅 

So...when I get a few minutes, I 'll contribute such a method and then see about getting Livewire to cleanup after itself with it.  🤓

PR #40831, opened 2022-02-06
[8.x] Add a string helper to swap multiple keywords in a string.

amitmerchant1990:
[This tweet](https://twitter.com/aarondfrancis/status/1488298201937588229) by Aaron Francis inspired me to open this PR. 

Essentially, it adds a new string helper called `Str::swap` that lets you swap multiple keywords, that you can specify in form of an array, in a string. This can come in handy in certain scenarios and I think it would be pretty useful if it's available in the framework natively.

Here are some examples.

```php
echo Str::swap([
    'PHP' => 'PHP 8',
    'awesome' => 'fantastic'
], 'PHP is awesome');

// outputs: 'PHP 8 is fantastic'

echo Str::swap([
    'ⓐⓑ' => 'baz',
], 'foo bar ⓐⓑ');

// outputs: 'foo bar baz'
```

I explained this in a little more detail in [this article](https://www.amitmerchant.com/swapping-keywords-in-a-string-in-laravel/).


mabdullahsari:
What is different from [strtr](https://www.php.net/manual/en/function.strtr.php)?

aarondfrancis:
@mabdullahsari I think it's a lot easier to read! 😂  Laravel has a lot of functions that help make native PHP more clear.

Also they are not functionally the same, as the implementation here (which I prefer) follow your `map`s order, whereas strtr re-orders by key length.

```php
$subject = 'test';

$map = [
    'te' => 'be',
    'test' => 'example',
];

dump('swap: ' . Str::swap($map, $subject));
// "swap: best"
dump('strtr: ' . strtr($subject, $map));
// "strtr: example"


$map = [
    'test' => 'example',
    'te' => 'be',
];
dump('swap: ' . Str::swap($map, $subject));
// "swap: example"
dump('strtr: ' . strtr($subject, $map));
// "strtr: example"
```

amitmerchant1990:
On top of @aarondfrancis' comment, I think it has a meaningful name from which you instantly get to know what it's intended to do.

PR #40833, opened 2022-02-06
findOrFail with array of ids throws ModelNotFound with all ids instead of only the missing ones

juampi92:
```php
try {
  Model::findOrFail([1, 2, 3]);
} catch (ModelNotFoundException $e) {
  dd($e->ids);
}
```

If only id=3 is missing, this will throw a `ModelNotFound` with the `ids = [1, 2, 3]`.

The ideal behaviour should be using only the missing models `ids = [3]`.

*Note:* I did not know if this is a proper bug or what, so I didn't pick the LTS just in case. Please let me know how to proceed.


driesvints:
Please send this to 9.x

PR #40840, opened 2022-02-07
Add `whereDistinctFrom` and `whereDistinctFromColumn`

ekvedaras:
Postgresql has `is distinct from` operator which will treat `null` values as known - meaning if you say

```sql
select * from "posts" where "og"  is distinct from 'foo'
```

it will return all values where `og` is not equal to `'foo'` or where `og` is `null`.
If we simply use `!=` operator `null` values will not be returned.

Mysql works the same way, except there is no `is distinct from` operator. So for it and other grammars, we need to simulate this by building the `!= or null` query ourselves.

Mysql example: https://www.db-fiddle.com/f/q74q6Q5RQ6vM6N3cMdNYfW/3
Postgresql example: https://www.db-fiddle.com/f/s88JcDQn6wg8FZJH6KoAMU/0

This should ease building queries to cater for this case and maybe even bring more awareness that `!=` operator works that way.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1473,6 +1473,44 @@ class Builder
         return $this;
     }
 
+    public function whereDistinctFrom($column, $value, $boolean = 'and')
+    {
+        $type = 'distinctFrom';
+
+        $this->wheres[] = compact('type', 'column', 'value', 'boolean');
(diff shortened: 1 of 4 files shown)


dammy001:
I think this PR should go to 9.x
@driesvints

driesvints:
Please send this to 9.x

PR #40844, opened 2022-02-07
[9.x] Throw exception when ResourceCollection collects something else than …

bastien-phi:
…JsonResource

This PR is a follow up of https://github.com/laravel/framework/pull/40799 and discussions on https://github.com/laravel/framework/pull/40208

I set taget to 9.x as to me it is a breaking change but feel free to change it if you want.

Not sure about the implementation nor the exception, I let you adjust this the way you prefer.

### DIFF
--- a/src/Illuminate/Http/Resources/CollectsResources.php
+++ b/src/Illuminate/Http/Resources/CollectsResources.php
@@ -2,11 +2,13 @@
 
 namespace Illuminate\Http\Resources;
 
+use Illuminate\Http\Resources\Json\JsonResource;
 use Illuminate\Pagination\AbstractCursorPaginator;
 use Illuminate\Pagination\AbstractPaginator;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
 use ReflectionClass;
+use RuntimeException;
 use Traversable;
 
 trait CollectsResources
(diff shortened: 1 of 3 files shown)


src/Illuminate/Http/Resources/CollectsResources.php
```diff
@@ -45,15 +47,21 @@ protected function collectResource($resource)
...
              class_exists($class = Str::replaceLast('Collection', 'Resource', get_class($this))))) {
-            return $class;
+            $collects = $class;
         }
+
+        if (! $collects || is_subclass_of($collects, JsonResource::class)) {
+            return $collects;
+        }
+
+        throw new RuntimeException('ResourceCollection must collect JsonResources.');
```
axlon, line 64: Wouldn't it make more sense to throw a `InvalidArgumentException` here?

driesvints: @axlon it's not really an argument for this method that's used, more like a preset value on the object's property. In this case it can be a LogicException: https://www.php.net/manual/en/class.logicexception.php as the dev is able to fix this directly in their code.

axlon: Yeah I was debating between the 2, but you're right `LogicException` makes more sense 👍

driesvints: @bastien-phi can you adjust that?

bastien-phi: done !

driesvints (approved):

PR #40845, opened 2022-02-07
[9.x] findOrFail with array of ids throws ModelNotFound with all ids instead of only the missing ones

juampi92:
```php
try {
  Model::findOrFail([1, 2, 3]);
} catch (ModelNotFoundException $e) {
  dd($e->ids);
}
```

If only id=3 is missing, this will throw a `ModelNotFound` with the `ids = [1, 2, 3]`.

The ideal behavior should be using only the missing models `ids = [3]`.


juampi92:
I think there might be a flaky test: https://github.com/laravel/framework/runs/5091994408?check_suite_focus=true#step:7:134

PR #40846, opened 2022-02-07
Eager-load Eloquent relationships via getter functions, allowing for more dynamic optimisations

jackwh:
Eloquent allows us to eager-load relationships by defining the `protected $with = [];` property on a Model. This works great if you always want to eager-load the same relationships globally — but for more complex apps, it would be good to have a bit more control over this.

For example, the `User` models in my app have many different relationships defined for several distinct areas of the platform. I want to be able to eager-load different relations by default, depending on which route/controller they're visiting. Whilst we could just use `User::with([...])`, this becomes hard to keep track of and maintain in larger projects.

This PR adds new getter/setter functions for `$with` and `$withCount`. The only place these property values are loaded is in `newQueryWithoutScopes()`, so this should be nicely backwards-compatible. Developers are then free to override the `getWith()` function for their own purposes, if needed.

Hopefully this helps some people with more complex Eloquent use cases. Thanks in advance!

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1997,7 +1997,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     {
         return static::$modelsShouldPreventLazyLoading;
     }
-    
+
     /**
      * Get the relations to eager load on every query.
      *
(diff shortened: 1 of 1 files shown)


ankurk91:
Laravel model is becoming a God object

https://en.m.wikipedia.org/wiki/God_object

JackWH:
> Laravel model is becoming a God object

Maybe... but it's also a fundamental part of the framework, and I think it's important developers should be able to control it.

This PR just makes `$with` consistent with how other properties can be dynamically overridden — like `getIncrementing()`, `getKeyName()`, `preventsLazyLoading()`, and more.

taylorotwell:

PR #40853, opened 2022-02-07
[8.x] Make `Str::swap()` first argument the string value

derekmd:
Follow-up to https://github.com/laravel/framework/pull/40831 before new method `Str::swap()` is release-tagged tomorrow.

`Str::is()` is the only method in its class that doesn't accept the string subject as the first argument. Ensure consistency with other `Str` methods and PHP standard library `strtr()` that this aliases with a better name.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -911,11 +911,11 @@ class Str
     /**
      * Swap multiple keywords in a string with other keywords.
      *
-     * @param  array  $map
      * @param  string  $subject
+     * @param  array  $map
      * @return string
      */
-    public static function swap(array $map, $subject)
+    public static function swap($subject, array $map)
     {
         return str_replace(array_keys($map), array_values($map), $subject);
     }
--- a/tests/Support/SupportStrTest.php
+++ b/tests/Support/SupportStrTest.php
@@ -627,17 +627,17 @@ class SupportStrTest extends TestCase
     {
         $this->assertSame(
             'PHP 8 is fantastic',
-            Str::swap([
+            Str::swap('PHP is awesome', [
                 'PHP' => 'PHP 8',
                 'awesome' => 'fantastic',
-            ], 'PHP is awesome')
+            ])
         );
 
         $this->assertSame(
             'foo bar baz',
-            Str::swap([
+            Str::swap('foo bar ⓐⓑ', [
                 'ⓐⓑ' => 'baz',
-            ], 'foo bar ⓐⓑ')
+            ])
         );
     }
 


taylorotwell:
Note that `strtr` has different behavior than the original implementation of this - which I think was intentional based on the PR comments.

derekmd:
OK, I can revert that part in this PR although without `strtr()` you get the replacement leak problem of  https://github.com/laravel/framework/issues/39258:

```php
$subject = 'foo bar';
$replacements = [
    'foo' => 'bar',
    'bar' => 'baz',
];

str_replace(array_keys($replacements), array_values($replacements), $subject);
// "baz baz"

strtr($subject, $replacements);
// "bar baz"
```

aarondfrancis:
I don't think there's an obvious right or wrong way here, but when I wrote my macro I wrote it in that order to match the other `Str` replacement functions.

```php
public static function replaceArray($search, array $replace, $subject)

public static function replace($search, $replace, $subject)

public static function replaceFirst($search, $replace, $subject)

public static function replaceLast($search, $replace, $subject)

public static function remove($search, $subject, $caseSensitive = true)
```

So doing `Str::swap($map, $subject)` seemed most inline with the rest of Laravel's Str to me.

taylorotwell:
I do agree that `strtr` may be the preferred way to do that - as not using it has created bugs for us in the past for sure.

taylorotwell:
I also do think there @aarondfrancis has a point that the subject usually comes last in PHP replacement functions. Thoughts @derekmd?

derekmd:
Ugh, yeah PHP API inconsistency strikes again. Those `replace*()` methods mimic `str_replace()` but probably should have put `$subject` first for the `Str` class' API design. Too late now - closing PR.

taylorotwell:
@derekmd I still think there may be some validity to using `strtr`?

taylorotwell:
Went ahead and pushed a change to use `strtr`.

derekmd:
@taylorotwell: `Stringable::swap()` also needs the `strtr()` change because it isn't calling `Str::swap()`.

According the other PR both implementations have a replacement leak problem with the `strtr()` issue being more obscure:

```php
$subject = 'foo bar';
$replacements = [
    'foo' => 'bar', // searching substring of the next searched phrase
    'foo bar' => 'baz',
];

str_replace(array_keys($replacements), array_values($replacements), $subject);
// "bar bar"

strtr($subject, $replacements);
// "baz"
```

Both are edge cases but this is why the translator uses string variable prefix `:` to prevent false-replacements. I'm unsure how this macro was used in real apps - untranslated natural language phrases for UI and API endpoint feedback? Apps can also use `__()` / `trans()` on arbitrary phrases:

```php
__('Hello :name, the day of the week is :day.', ['name' => 'Taylor', 'day' => date('l')]);
// "Hello Taylor, the day of the week is Monday."
```

Or go the unnamed variable `sprintf()` / `vsprintf()` path.

PR #40859, opened 2022-02-08
[8.x] Retrieve the underlying string value in Stringable

cerbero90:
This PR adds the method `get()` to `Illuminate/Support/Stringable`. This can be useful to quickly retrieve the underlying string value without the need of casting:

```php
// before
return (string) Str::of('foo')->uppercase()->append('!');

// after
return Str::of('foo')->uppercase()->append('!')->get();
```

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -32,6 +32,16 @@ class Stringable implements JsonSerializable
     }
 
     /**
+     * Retrieve the underlying string value.
+     *
+     * @return string
+     */
+    public function get()
+    {
+        return $this->value;
+    }
+
+    /**
      * Return the remainder of a string after the first occurrence of a given value.
      *
      * @param  string  $search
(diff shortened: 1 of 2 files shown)


huangdijia:
`Str::of('foo')->uppercase()->append('!')->__toString()`

cerbero90:
@huangdijia definitely possible but not as elegant and consistent, in my opinion

SjorsO:
I like this PR, but I'd call the method `toString` instead.

A downside to having to use `__toString()` is that it doesn't get auto-completed by PHPStorm:

image

And having a dedicated method for getting the string looks nicer:
```php
$string = Str::of('test')->__toString();

$string = (string) Str::of('test');

$string = Str::of('test')->toString();
```

driesvints:
Please send this to 9.x

cerbero90:
New PR: https://github.com/laravel/framework/pull/40861

PR #40861, opened 2022-02-08
[9.x] Retrieve the underlying string value in Stringable

cerbero90:
This PR adds the method `get()` to `Illuminate/Support/Stringable`. This can be useful to quickly retrieve the underlying string value without the need of casting:

```php
// before
return (string) Str::of('foo')->uppercase()->append('!');

// after
return Str::of('foo')->uppercase()->append('!')->get();
```

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -32,6 +32,16 @@ class Stringable implements JsonSerializable
     }
 
     /**
+     * Retrieve the underlying string value.
+     *
+     * @return string
+     */
+    public function get()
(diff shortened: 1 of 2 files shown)


nunomaduro:
I rather prefer a `->value()`, but lets see what the team thinks.

SjorsO:
I'd prefer `->toString()`. It aligns nicely with `->toHtmlString()` (and other methods like `->toArray()`). We know the value is always a string, `->value()` doesn't reflect that.

taylorotwell:
Added `value()` and `toString()`.

PR #40864, opened 2022-02-08
[9.x] Collection find first key from a collection by testing the values

ivandokov:
This PR adds support to find the first key from a collection by testing the values.
I couldn't find a good way to achieve this with the currently available methods and I thought this could be a nice addition to the framework.

Example:

```php
$data = new Collection(['john' => 'doe', 'foo' => 'bar']);
$result = $data->firstKeyWhereValue(fn ($value) => $value === 'bar', 'default');
// foo
```

I am not sure if this method name is the best but it's a descriptive one and I think it doesn't raise any doubts of what it does. You don't need to go to the docs to understand how to use it.
I also considered `firstKeyWhere` as it is shorter but you don't know what you test with the callback.

Additionally I think adding the key to the callback is probably useful if you want to test both the value and the key but I am not sure about that. 
```php
if ($callback($value, $key)) {...
```

I'm looking for feedback.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -390,6 +390,26 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
+     * Get the first key from the collection where the value passes the given truth test.
+     *
+     * @template TFirstDefault
+     *
+     * @param  (callable(TValue, TKey): bool)  $callback
+     * @param  TFirstDefault|(\Closure(): TFirstDefault)  $default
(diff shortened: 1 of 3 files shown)


devfrey:
I think [`Collection::search()`](https://laravel.com/docs/8.x/collections#method-search) is what you're looking for. The only difference is that it returns `false` when no matching value can be found and does not support a default.

ivandokov:
Hey @devfrey, it looks like this method will do the trick indeed. The name though is really not what I expect it to do. `search` actually sounds like an alias for `where`.

driesvints:
Thanks @devfrey

PR #40865, opened 2022-02-08
[9.x] Add `whereDistinctFrom` and `whereDistinctFromColumn`

ekvedaras:
Second attempt of #40841 with added comments.

---

Postgresql has `is distinct from` operator which will treat `null` values as known - meaning if you say

```sql
select * from "posts" where "og"  is distinct from 'foo'
```

it will return all values where `og` is not equal to `'foo'` or where `og` is `null`.
If we simply use `!=` operator `null` values will not be returned.

Mysql works the same way, except there is no `is distinct from` operator (`not <=>` was suggested in the comments). So for it and other grammars, we need to simulate this by building the `!= or null` query ourselves.

Mysql example: https://www.db-fiddle.com/f/q74q6Q5RQ6vM6N3cMdNYfW/6
Postgresql example: https://www.db-fiddle.com/f/s88JcDQn6wg8FZJH6KoAMU/1
MsSQL example: http://sqlfiddle.com/#!18/d0af7a2/9
MariaDB example: https://dbfiddle.uk/?rdbms=mariadb_10.3&fiddle=9c1bbbbe05776334a448cf9a05ded7c7
Sqlite example: https://www.db-fiddle.com/f/hc4RKupTufvuUs3evz3RpN/1

This should ease building queries of such cases and maybe bring more awareness to this behavior of `!=` operator.

### DIFF
--- a/src/Illuminate/Database/Query/Grammars/Grammar.php
+++ b/src/Illuminate/Database/Query/Grammars/Grammar.php
@@ -600,6 +600,34 @@ class Grammar extends BaseGrammar
     }
 
     /**
+     * Compile an "is distinct from" clause.
+     *
+     * @param  \Illuminate\Database\Query\Builder  $query
+     * @param  array  $where
+     * @return string
+     */
+    protected function whereDistinctFrom(Builder $query, $where)
+    {
+        $column = $this->wrap($where['column']);
+
+        return '('.$column.' != '.$this->parameter($where['value']).' or '.$column.' is null)';
+    }
+
+    /**
+     * Compile an "is distinct from column" clause.
+     *
+     * @param  \Illuminate\Database\Query\Builder  $query
+     * @param  array  $where
+     * @return string
+     */
+    protected function whereDistinctFromColumn(Builder $query, $where)
+    {
+        $first = $this->wrap($where['first']);
+
+        return '('.$first.' != '.$this->wrap($where['second']).' or '.$first.' is null)';
+    }
+
+    /**
      * Compile a "JSON contains" statement into SQL.
      *
      * @param  string  $column
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1507,6 +1507,65 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Add an "is distinct from" (not equals or is null) statement to the query.
+     *
+     * @param  string  $column
+     * @param  mixed  $value
+     * @param  string  $boolean
+     * @return $this
+     */
+    public function whereDistinctFrom($column, $value, $boolean = 'and')
+    {
+        $type = 'distinctFrom';
+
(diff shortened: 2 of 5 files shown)


src/Illuminate/Database/Query/Grammars/Grammar.php
```diff
@@ -599,6 +599,34 @@ protected function whereJsonContains(Builder $query, $where)
...
+     *
+     * @param  \Illuminate\Database\Query\Builder  $query
+     * @param  array  $where
+     * @return string
+     */
+    protected function whereDistinctFrom(Builder $query, $where)
+    {
+        $column = $this->wrap($where['column']);
+
+        return '('.$column.' != '.$this->parameter($where['value']).' or '.$column.' is null)';
```
timmartin19, line 613: I  think this should probably use the [null safe equals operator](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#operator_equal-to).

> The [<=>](https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html#operator_equal-to) operator is equivalent to the standard SQL IS NOT DISTINCT FROM operator

So effectively, using `NOT col1 <=> col2` is equivalent to using `IS DISTINCT FROM` in other SQL dialects.

ekvedaras: Interesting 👀

timmartin19: I just realized that this is the default grammar and not MySQL specific.  The MySQL docs make me think `IS DISTINCT FROM` may be the SQL standard and MySQL is a little weird with different syntax.  Maybe we could move the PostgreSQL version here and update the `MySqlGrammar` instead of `PostgresGrammar` since the postgres  mechanism is more in line with the SQL standard.  Or alternatively, we could update `MySqlGrammar` and leave this and the `PostgresGrammar`  in place as is?

ekvedaras: Seems like `is distinct from` only works on PostgreSQL, `not <=>` only on MySQL and MariaDB and the rest need to use `!= or is null` 🤷

rodrigopedra:
Shouldn't this operator also be added here:

https://github.com/laravel/framework/blob/5317720fc6953d7f22a22a20dee0c0b4746c407a/src/Illuminate/Database/Query/Grammars/MySqlGrammar.php#L10-L15

So someone can also use it manually in a where clause?

ekvedaras:
> Shouldn't this operator also be added here:
> 
> https://github.com/laravel/framework/blob/5317720fc6953d7f22a22a20dee0c0b4746c407a/src/Illuminate/Database/Query/Grammars/MySqlGrammar.php#L10-L15
> 
> So someone can also use it manually in a where clause?

Seems like that operator is already in general list

taylorotwell:
Hmm - I dunno - it feels fairly easy to achieve this with existing methods. Will hold off for the time being. Thank you.

PR #40867, opened 2022-02-08
[9.x] Model::loadCounts are incorrect when the relationship being counted contains a join

troy-whitespark:
This problem stemmed from a real issue I was having building relationships that are scoped to particular "locations" in my application.

The issue is, `loadCounts` and `withCounts` both create a new subquery and copy the `where`s, but they do not copy the join constraints - in the case where we're doing something like inner joining to a pivot table to get permutations of content and location pairs, that means that the `relationship_count` and `relationship()->count()` will return two different values and cause bugs.

I've submitted a failing test encapsulating the behaviour, but this seemed like too big of a fix to something core to have an outside maintainer solve - I suspect you'd end up rewriting my contributions anyway. Let me know if you want me to actually develop a solution to this problem as well.

### DIFF
--- a/tests/Database/DatabaseEloquentWithCountJoinTest.php
+++ b/tests/Database/DatabaseEloquentWithCountJoinTest.php
@@ -1,0 +1,142 @@
+<?php
+
+namespace Illuminate\Tests\Database;
+
+use Illuminate\Database\Capsule\Manager as DB;
+use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Is this a Laravel 9 specific issue, or is this the same in Laravel 6/8?

troy-whitespark:
That's a good point - I haven't tested in 6&7, but it definitely affects 8
as well.

PR #40868, opened 2022-02-08
[9.x] Update Mailable.php

rentalhost:
Allow `->from(new Address($address, $name))` support again (was `->from($address, $name)` in Laravel 8).

### DIFF
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -22,6 +22,7 @@ use ReflectionClass;
 use ReflectionProperty;
 use Symfony\Component\Mailer\Header\MetadataHeader;
 use Symfony\Component\Mailer\Header\TagHeader;
+use Symfony\Component\Mime\Address;
 use Symfony\Component\Mime\Email;
 
 class Mailable implements MailableContract, Renderable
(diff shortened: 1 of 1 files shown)


rentalhost:
Just to be clear, with the **Laravel 9**, the call to `->from()` and similar methods stopped supporting _email address_ and _name_ and started to support only the _address_ itself (without the _name_). However, it still allows receiving an `Address` object from _Symfony Mailer_, which contains both information. However, the normalization process throws an exception when using `Address`, even though the method says accept.

In this case, I still suggest an update to the guide:

Instead of using `->from($address, $name)` use `->from(new Address($address, $name))`.

taylorotwell:
Are we unable to support `->from($address, $name)` in Laravel 9? Why do we need to force people to create an `Address` instance?

rentalhost:
@taylorotwell I do not see any problem to keep support to old way (that I personally prefer).

But the current code is:

```php
    public function from(Address|string ...$addresses): static
    {
        return $this->setListAddressHeaderBody('From', $addresses);
    }
```

To support old way it need be modified.

The PR is only a fix to supports also `Address`, as this method parameter indicates.

rentalhost:
Ah! It seems these methods are provided directly by Symfony Mailer. Anyway, the Laravel Message will create an Address instance and the exception will be throw anyway: 

https://github.com/laravel/framework/blob/fbe7ed828819eadabb3cb9f95847a2ffe8846fd6/src/Illuminate/Mail/Message.php#L42-L56

taylorotwell:
Hmm, so does `->from($address, $name)` work? I got a little lost in the last message. What do we need to fix?

rentalhost:
Not in fact, because it will throw an exception when you set the `$name`. As the `normalizeRecipient()` will not handle the `instanceof Address`, it will just return the own instance of `Address` that does not supports `$email` property (or `$name`). Then the PR fixes that.

image

taylorotwell:
Thanks

PR #40870, opened 2022-02-08
add controller helper to help us when looking for action of a specific route

al1r374:
Hey @taylorotwell 
Suppose we deal with an application that has many `route()` helpers all over Tests and source code. 
When we face `route()` helpers in our code, it's hard to detect the controllers that are related to these helpers, especially when we deal with some dirty code or some huge application, it gets harder.
At the moment we must use `php artisan route:list | grep "ROUTENAME"` if we want to find the related controller but this way has more complex and even many developers don't know what they can do.
As a solution, I think if we have a `controller()` helper this help us overcome this situation.

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -820,6 +820,19 @@ if (! function_exists('to_route')) {
     }
 }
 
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #40880, opened 2022-02-09
[9.x] fix: bring back old behaviour 

ankurk91:
Hi,

This PR bring back old behavior, since this can be a breaking change for many.

```php
$notifiable->routeNotificationFor('MethodWhichIsNotImplemented', $notification);

// Larval 8.x and before => null
// Larval v9.x => throws UnhandledMatchError
```

The routeNotificationFor() method should return `null` if the method is not implemented on Notifiable.

Tip: the `match` expression does not return `null` when there is no match.

PS - This PR adds `default` case to some more `match` expressions.

### DIFF
--- a/src/Illuminate/Notifications/RoutesNotifications.php
+++ b/src/Illuminate/Notifications/RoutesNotifications.php
@@ -46,6 +46,7 @@ trait RoutesNotifications
         return match ($driver) {
             'database' => $this->notifications(),
             'mail' => $this->email,
+            default => null,
         };
     }
 }
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -869,6 +869,7 @@ class Connection implements ConnectionInterface
             'beganTransaction' => new TransactionBeginning($this),
             'committed' => new TransactionCommitted($this),
             'rollingBack' => new TransactionRolledBack($this),
+            default => null,
         });
     }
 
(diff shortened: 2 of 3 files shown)


src/Illuminate/Notifications/RoutesNotifications.php
```diff
@@ -46,6 +46,7 @@ public function routeNotificationFor($driver, $notification = null)
         return match ($driver) {
             'database' => $this->notifications(),
             'mail' => $this->email,
+            'default' => null,
```
axlon, line 49:
```suggestion
            default => null,
```

Match's default should not have quotes

ankurk91: @axlon Fixed, thanks

driesvints:
@ankurk91 would it be possible to go over https://github.com/laravel/framework/pull/39583 and add the other places where this is needed to this PR as well? Seems like there's more of them.

ankurk91:
@driesvints 

I can do that, but looks like all of other `match` expression has a `default` case in the mentioned PR. 
Only `src/Illuminate/Notifications/RoutesNotifications.php` was missed.

crynobone:
`src/Illuminate/Database/Connection.php`
`src/Illuminate/Redis/RedisManager.php`

ankurk91:
@crynobone

Done!

driesvints (approved):

Issue #40693, opened 2022-01-28
enum arrayable casting


makroxyz:
- Laravel Version: 8.81.0
- PHP Version: 8.1.1

### Description:
Different behaviour when casting to array an Enum implementing Arrayable class from collection and from eloquent model attribute.

### Steps To Reproduce:

I have this enum
```php
enum Status: string implements Arrayable
{
    case pending = 'pending';
    case done = 'done';
    
    /**
     * @inheritDoc
     */
    public function toArray()
    {
        return [
            'name' => $this->name,
            'value' => $this->value,
            'description' => "description of {$this->name}",
        ];
    }
}
```

In collection:
```php
collect(Status::pending)->toArray();
/**
[
     "name" => "pending",
     "value" => "pending",
     "description" => "description of pending",
   ]
*/
```

in Model:
```php
class MyModel extends Model
{
    protected $casts = [
        'status' => Status::class
    ];
}


MyModel::first()->toArray();
/**
[
    'id' => 1,
    'status' => 'pending'
]
*/
```

Even without Arrayable interface on enum the result is different.

Is it right? 

How can I cast MyModel to array and have enum attribute (automatically) casted following the Arrayable interface?

This is related to a (closed) PR #40605. Should it be considered for 9.x?

driesvints:
Heya, this seems to be rejected. You can indeed always attempt a new PR for 9.x but it's up to Taylor in the end to accept or not.

Issue #40580, opened 2022-01-24
Enum casts skips Arrayable implementation when calling Model::toArray()


makroxyz:
- Laravel Version: 8.80
- PHP Version: 8.1.1
- Database Driver & Version: MariaDB 10.6.5

### Description:
Enum casts skips Arrayable when calling Model::toArray()

### Steps To Reproduce:
- I have native PHP 8.1 Enum backed class implementing Illuminate\Contracts\Support\Arrayable.
- My Model casts an attribute (enum) to the above Enum class
- Calling Model::first()->enum->toArray() gives me what I'm expecting
- Calling Model::first()->toArray() always casts attribute to string (enum value)

Could the issue be in Illuminate\Database\Eloquent\Concerns\HasAttributes line 299?
```php
if ($this->isEnumCastable($key)) {
    $attributes[$key] = isset($attributes[$key]) ? $attributes[$key]->value : null; // <- this returns string instead of Enum class 
}

if ($attributes[$key] instanceof Arrayable) { <- this will be skipped
    $attributes[$key] = $attributes[$key]->toArray();
}
```

themsaid:
I don't understand your issue. Please submit code that I can test.

makroxyz:
My Enum.php
```php
namespace App\Enums;

use Illuminate\Contracts\Support\Arrayable;


enum MyEnum: string implements Arrayable
{
   
    case COMPANY = 'company';
    case PERSON = 'person';

    public function label(): string 
    {
        return static::getLabel($this);
    }

    public static function getLabel(self $value): string 
    {
        return match ($value) {
            self::COMPANY => 'Company Label',
            self::PERSON => 'Person Label',
        };
    }

    public function toArray(): array
    {
        return [
            'name' => $this->name,
            'value' => $this->value,
            'label' => $this->label()
        ];
    }
}
```

MyModel.php
```php
namespace App\Models;

use App\Enums\MyEnum;


class MyModel extends \Illuminate\Database\Eloquent\Model
{
    protected $casts = [
       'enum' => MyEnum::class,
 
    ]
}
```

```php
MyModel::first()->enum; // return MyEnum class -> OK

MyModel::first()->enum->toArray(); // array with keys [name,value,label] -> OK 

MyModel::first()->toArray(); // string -> KO I suppose to get array with keys [name,value,label]
```

themsaid:
This is by design yes so toArray always returns scalar types.

makroxyz:
> 

[...]

PR #40885, opened 2022-02-09
[9.x] Fix enum casts arrayable behaviour

diegotibi:
This PR fix a wrong behavoiur when an attribute is casted to an enum implementig Arrayable in comparision to the behavoiur when the same thing is done on a collection object.

Clarifying example:

Let's take this enum:
```php
enum Status: string implements Arrayable
{
    case pending = 'pending';
    case done = 'done';
    
    /**
     * @inheritDoc
     */
    public function toArray()
    {
        return [
            'name' => $this->name,
            'value' => $this->value,
            'description' => "description of {$this->name}",
        ];
    }
}
```

Behaviour in a collection object:
```php
collect(Status::pending)->toArray();
/**
[
     "name" => "pending",
     "value" => "pending",
     "description" => "description of pending",
]
*/
```

Behaviour in a casted model attribute:
```php
class MyModel extends Model
{
    protected $casts = [
        'status' => Status::class
    ];
}


MyModel::first()->toArray();
/**
[
    'id' => 1,
    'status' => 'pending'
]
*/
```

Expected behaviour:
```php
MyModel::first()->toArray();
/**
[
    'id' => 1,
    'status' => [
         "name" => "pending",
         "value" => "pending",
         "description" => "description of pending",
   ]
]
*/
```

Related issues: #40580 and #40693

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -297,7 +297,7 @@ trait HasAttributes
                 $attributes[$key] = $this->serializeClassCastableAttribute($key, $attributes[$key]);
             }
 
-            if ($this->isEnumCastable($key)) {
+            if ($this->isEnumCastable($key) && (! $attributes[$key] instanceof Arrayable)) {
                 $attributes[$key] = isset($attributes[$key]) ? $attributes[$key]->value : null;
             }
 
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -297,7 +297,7 @@ trait HasAttributes
                 $attributes[$key] = $this->serializeClassCastableAttribute($key, $attributes[$key]);
             }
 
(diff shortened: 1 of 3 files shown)


driesvints:
I definitely wouldn't expect enums to be converted to their array version. Their value should be used. The current behavior is expected imo.

X-Coder264:
Without this change there's no point in an enum implementing `Arrayable` as it does nothing for enums currently - which is inconsistent with how all other `Arrayable` implementations are treated by Eloquent.

On the other hand I don't see much of an use case in transforming a scalar value into an array.

I guess there's no harm in letting enums behave like all other `Arrayable` implementations as I don't see any other use case as to why users would implement `Arrayable` on their enums.

driesvints:
@X-Coder264 hmm I guess you're right about that.

GrahamCampbell:
If this is considered a bug fix, this should probably land on Laravel 8, not 9.

diegotibi:
@GrahamCampbell yes, you're probably right, this PR can ported to 8.x branch with no work at all. However enums are a php 8.1 feature so it's probably more usefull on 9.x branch

diegotibi:
@taylorotwell Should I make a PR on 8.x branch too, like @GrahamCampbell suggested?

PR #40892, opened 2022-02-09
[9.x] Add `defaultOrderBy` property to model

jonnott:
Builds on @jasonmccreary's previous PR at https://github.com/laravel/framework/pull/40678 .. with some tweaks/additions:

### Naming

Feels as though simply `$orderBy` as the name of the property could be confusing for devs, as they may think that either (a) this is the place to always specify ordering for your queries, instead of using `orderBy()` within queries/scopes directly, or (b) that this default ordering always gets applied _in addition_ to any ordering from `orderBy()` calls, which it does not - the default ordering only gets applied when there is no other ordering specified for the query.

It's a little more verbose, but hopefully much clearer and unambiguous.

### Applying in query builder methods

Previously, the call to `applyOrderBy()` (now `applyDefaultOrderBy()`) was chained _before_ `applyScopes()` in the Eloquent builder methods. This meant that any scopes which themselves included an `orderBy()` call would have been applied to the query _in addition to_ and _after_ the default ordering, which would be incorrect/ambiguous. This method-chaining order has been corrected. Thanks to @crynobone who spotted this in the previous PR thread.

### Removal of the default ordering

Eloquent allows any modification or constraint applied to queries via model configuration to also be _removed_, enabling a query to return to simply `select * from {table}` as the compiled SQL. For example, even if a `boot()` method applies a global scope, this can be removed for a certain query scenario via `withoutGlobalScope()` or `withoutGlobalScopes()`. This new default ordering functionality should also be removable in the same way, so I've added a `$useDefaultOrderBy = true` property to the Eloquent builder and a `withoutDefaultOrderBy()` method which prevents the default ordering from being applied by setting this property to `false`. This allows the model to still be queried without the default ordering when required, and I feel this provides consistency with other similar mechanisms the query builders provide.

### Tests

I've added additional tests. One to verify that chaining `withoutDefaultOrderBy()` into a query prevents the default ordering from being applied, and one to verify that the default ordering does not get applied if a local scope method added to the query contains an `orderBy()` call. All tests passing.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -126,6 +126,13 @@ class Builder implements BuilderContract
     protected $removedScopes = [];
 
     /**
+     * Indicates whether to apply any default ordering.
+     *
+     * @var bool
+     */
+    protected $useDefaultOrderBy = true;
+
+    /**
      * Create a new Eloquent query builder instance.
      *
      * @param  \Illuminate\Database\Query\Builder  $query
@@ -211,6 +218,18 @@ class Builder implements BuilderContract
     public function removedScopes()
     {
         return $this->removedScopes;
+    }
+
+    /**
+     * Remove any default ordering.
+     *
+     * @return $this
+     */
+    public function withoutDefaultOrderBy()
+    {
+        $this->useDefaultOrderBy = false;
+
+        return $this;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jonnott:
@taylorotwell Also, I'd be more than happy to PR the docs to document this.

taylorotwell:

jonnott:
@taylorotwell It already exists as a community package, which is linked above: https://github.com/stephenjude/default-model-sorting

jonnott:
- - [ ] Also, @taylorotwell do you have any specific thoughts on what issues there might be with this addition? You seemed to want to give some extra thought to the implementation detail at the time of Jason's original PR ..?

PR #40894, opened 2022-02-09
[10.x] Add redirectToRoute to router

shaedrich:
so that you can do `Route::redirectToRoute('route.name')` in `routes/web.php` for instance instead of
```
Route::any('uri', function() {
    return redirect()->route('route.name');
});
```
or
`Route::redirect('uri', 'destination-uri');`


derekmd:
Previous PR submission https://github.com/laravel/framework/pull/40878 was already declined.

driesvints:
Thanks but looks like we won't be merging this for now.

PR #40896, opened 2022-02-09
[9.x] Allow base JsonResource class to be collected

jwohlfert23:
In some cases, my app is just using the base JsonResource like so:

```php
return JsonResource::collection(User::take(10)->get());
```

This fails as the `collects()` method is only checking if the class is a subclass. It does not allow the base JsonResource class.

### DIFF
--- a/src/Illuminate/Http/Resources/CollectsResources.php
+++ b/src/Illuminate/Http/Resources/CollectsResources.php
@@ -57,7 +57,7 @@ trait CollectsResources
             $collects = $class;
         }
 
-        if (! $collects || is_subclass_of($collects, JsonResource::class)) {
+        if (! $collects || $collects === JsonResource::class || is_subclass_of($collects, JsonResource::class)) {
             return $collects;
         }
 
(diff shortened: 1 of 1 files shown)


src/Illuminate/Http/Resources/CollectsResources.php
```diff
@@ -57,7 +57,7 @@ class_exists($class = Str::replaceLast('Collection', 'Resource', get_class($this
             $collects = $class;
         }
 
-        if (! $collects || is_subclass_of($collects, JsonResource::class)) {
+        if (! $collects || $collects === JsonResource::class || is_subclass_of($collects, JsonResource::class)) {
```
rodrigopedra, line 60: Maybe change `is_subclass_of()` by `is_a()` or the `instanceof` operator would be cleaner?

https://www.php.net/manual/en/function.is-a.php

> Checks if the given object_or_class is of this class or has this class as one of its parents. 

https://www.php.net/manual/en/language.operators.type.php

driesvints: `instanceof` is only for actual instances for actual instances of the object so `is_a` with the `$allow_string` parameter set to `true` is what's needed here.

```suggestion
        if (! $collects || is_a($collects, JsonResource::class, true)) {
```

PR #40900, opened 2022-02-09
[9.x] Add disabled directive

belzaaron:
This PR provides the use of the common HTML directive `disabled` through a shorthand syntax via a Blade directive. Allowing developers to have a more enjoyable method of disabling a button, for example, could improve the smooth flow that we Laravel developers are consistently being treated too. Simple yet, effective.

```blade
<button type="submit" @disabled($errors->isNotEmpty())>Submit</button>
```

Editing for documentation PR: https://github.com/laravel/docs/pull/7690

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php
@@ -326,4 +326,15 @@ trait CompilesConditionals
     {
         return "<?php if{$condition}: echo 'checked'; endif; ?>";
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


bogdankharchenko:
I feel like I've been seeing a few of these new blade helpers -- but sometimes I wonder is this ultimately becoming a form builder?  

I'm not strongly opposed to this but I wonder if we need these additional blade helpers... `@readonly` next week?

This is a form builder for example, which uses components/functions to define the form structure in an elegant way instead - maybe be the solution for some people and perhaps could prevent a dozen more PR's for form related directives.
https://github.com/glhd/aire#blade-components

PR #40907, opened 2022-02-09
[9.x] Add support for getters, setters and magic methods on collections objects

sidux:
Using Symfony property accessor component to make collections detect setters, getters on objects


driesvints:
There needs to be way more explanation about this and tests are also failing it seems.

PR #40911, opened 2022-02-10
[9.x] Add the langPath() function to `Illuminate/Contracts/Foundation/Application`

wilsenhc:
Now that the `lang` directory has changed, many developers will be updating their packages to publish translations in the correct path it would help to have that in the Contract so it is type-hinted for a better experience.


crynobone:
Laravel never add new interface on patch release. This has to go to Laravel 10.

driesvints:
Please send this to master, thanks.

PR #40913, opened 2022-02-10
Allow enums in migrations

assertchris:
This change allows enums to be used during column creation, in migrations:

```php
enum Difficulty: string
{
    case Easy = 'easy';
    case Hard = 'hard';
}

$table->enum('difficulty', Difficulty::class);
```


assertchris:
Sigh. https://twitter.com/jeffrey_way/status/1489609913588461571

Issue #40914, opened 2022-02-10
Possible removal of Google Font usage from default templates


bashgeek:
First of all, thanks for your hard work and congratulations on the 9.0 release!

Although it's not a bug, it's definitely something that should be considered for a future removal/replacement: Removing Google Font from the default exception templates and welcome page:
- https://github.com/laravel/framework/blob/b9203fca96960ef9cd8860cb4ec99d1279353a8d/src/Illuminate/Foundation/Exceptions/views/layout.blade.php#L11
- https://github.com/laravel/framework/blob/b9203fca96960ef9cd8860cb4ec99d1279353a8d/src/Illuminate/Auth/Console/stubs/make/views/layouts/app.stub#L17
- https://github.com/laravel/framework/blob/b9203fca96960ef9cd8860cb4ec99d1279353a8d/src/Illuminate/Foundation/Exceptions/views/minimal.blade.php#L11
- https://github.com/laravel/laravel/blob/1d9e53f90d484e636dfca1812e38953befd97158/resources/views/welcome.blade.php#L10

The reasoning behind this, German courts very recently ruled the integration of Google Fonts without the previous consent of the user illegal in compliance with GDPR for all users in the EU:
- https://www.theregister.com/2022/01/31/website_fine_google_fonts_gdpr/
- https://nakedsecurity.sophos.com/2022/01/31/website-operator-fined-for-using-google-fonts-the-cloudy-way/
- https://news.ycombinator.com/item?id=30135264

I'd assume the same goes for other public CDN's like jsDelivr etc. in the future and for other EU countries, because it's essentially the same service and reasoning behind, but I'm (obviously) not a lawyer.

I understand that this is entirely the responsibility of the maintainer of the actual website and not Laravel, but having default Exception templates (which probably won't get changed or removed in most installations) including Google Fonts is a bit of a pain, especially because it seems completely unnecessary to have those in there. It might be a good idea to remove this entirely to avoid any legal issues for websites having users in the EU just by installing a default Laravel application.

Thank you!

driesvints:
Can you attempt a PR?

PR #40926, opened 2022-02-10
[9.x] Removal of Google Font integration from default exception templates

bashgeek:
Follow-up PR regarding issue #40914:

This PR removes all occurrences of using Google Fonts in the Exception templates.

The reasoning behind this, German courts very recently ruled the integration of Google Fonts, without the previous consent of the user, illegal in compliance with GDPR for all users in the EU:
- https://www.theregister.com/2022/01/31/website_fine_google_fonts_gdpr/
- https://nakedsecurity.sophos.com/2022/01/31/website-operator-fined-for-using-google-fonts-the-cloudy-way/
- https://news.ycombinator.com/item?id=30135264

I'd assume the same goes for other public mass-hosting CDN's by US companies in the future and for other EU countries, because it's essentially the same service and reasoning behind, but I'm (obviously) not a lawyer, same thing happened for Google Analytics in the past few weeks though.

I understand that this is entirely the responsibility of the maintainer of the actual website and not Laravel, but having default Exception templates (which most likely won't get changed or removed in most installations) including Google Fonts is a bit of a pain, especially because it seems completely unnecessary to have those in there. I think it might be a good idea to remove this entirely to avoid any legal issues for websites having users in the EU just by installing a default Laravel application.

Alternatively, we could pull the font files from Google and deliver them together with Laravel and host them locally, but that sounds even more unnecessary.

Thanks!


bashgeek:
I'd rule out using other public CDNs, because I'd assume they'll eventually run into the same legal issue in the EU, if they process any kind of user-data outside the EU. If you really want to use a custom font and want to be safe from this, the only way is to package those font files together with Laravel (and obviously make sure the copyright allows to do so), which seems total overkill for a default exception page that doesn't do or show much besides an error code and title anyway, for more people will probably have integrated custom error pages in their apps. In my opinion it's more than sufficient to use the systems default sans-serif font for this, which is what the changes would do now.

Essentially, every public Laravel app that has not implemented custom error pages, or removed the default ones shipped with Laravel, is now, thanks to the current German court ruling, effectively in breach of the GDPR if it's available to users in the EU. I'd also assume that most maintainers (incl. us) are most likely not even aware that these pages are using Google Fonts, I certainly wouldn't have expected this, but a legal check of a big client came up with this issue.

I'm all ears if anyone has a suggestion on how to make this legally safe and yet look similar to what the font looks now and I'm happy to update the PR accordingly.

jimmitjoo:
Possibly a flag during installation where the font is downloaded into resources/fonts if a gdpr flag is provided? Or the other way around, if a "cdn "flag is provided, just go the current way.

driesvints:
@bashgeek definitely don't want to downplay your concerns. Just saying that we can't just strip the fonts and let these pages be unstyled.

bashgeek:
Just a follow up posts with some screenshots to visualize what would change:

Current default 404 page with the Nunito Google Font:
CleanShot 2022-02-10 at 20 36 54

Default 404 page after this PR and with default sans-serif font on my Chrome on macOS:
CleanShot 2022-02-10 at 20 38 30

There is a small difference, yes, but I certainly wouldn't say it's now "unstyled" :-)

Jubeki:
> Possibly a flag during installation where the font is downloaded into resources/fonts if a gdpr flag is provided? Or the other way around, if a "cdn "flag is provided, just go the current way.

I think a gdpr flag behind the installer would be great, but how many people would actually forget or even know about the flag?

> Alternatively, we could pull the font files from Google and deliver them together with Laravel and host them locally, but that sounds even more unnecessary.

I am also no fan of bundling a font together with Laravel, but still here is a package from spathe which allows exactly that: https://github.com/spatie/laravel-google-fonts

> There is a small difference, yes, but I certainly wouldn't say it's now "unstyled" :-)

I think this is looks okay for me. But still not sure this is the right way to remove it completely.

dennisprudlo:
> I think a gdpr flag behind the installer would be great, but how many people would actually forget or even know about the flag?

Yeah I think most people wouldn't know about the flag and will step into this "trap" not knowing that their app is in breach of the GDPR.

I agree with @bashgeek that using the systems sans-serif font is not much of a difference, however, makes sure that the app complies with the GDPR.

GrahamCampbell:
All of these are just defaults, right? People don't have to use this. It's not up to Laravel to manage your compliance requirements.

bashgeek:
> All of these are just defaults, right? People don't have to use this. It's not up to Laravel to manage your compliance requirements.

While this is of course very true, that doesn't mean that Laravel should ship with a default that would break a law for probably the vast majority of its users by default, especially if including Google Fonts here has no apparent reason or advantage.

Jubeki:
> All of these are just defaults, right? People don't have to use this. It's not up to Laravel to manage your compliance requirements.

It doesn't affect only Germany, but probably the whole EU.
I think a Privacy-First approach would be recommended, because every Application which interacts with EU-Customers will be affected in the future.

taylorotwell:
Updated to Tailwind's system font stack.

PR #40936, opened 2022-02-10
[9.x] Makes `ExceptionHandler::renderForConsole` internal

### LINKED ISSUE #40935: App\Exceptions\Handler::renderForConsole() is ignored
Please reopen ticket #23515 or at least update the Laravel documentation to explain why renderForConsole no longer works properly.

nunomaduro:
This pull request makes the `ExceptionHandler::renderForConsole` internal. **For years**, Collision overrides the method with its own logic, and as consequence, `App\Exceptions\ExceptionHandler::renderForConsole` never gets called.

Adding the flag `@internal` will instruct people that this method is not meant to be overwritten.

Closes https://github.com/laravel/framework/issues/40935.

### DIFF
--- a/src/Illuminate/Foundation/Exceptions/Handler.php
+++ b/src/Illuminate/Foundation/Exceptions/Handler.php
@@ -692,6 +692,8 @@ class Handler implements ExceptionHandlerContract
      * @param  \Symfony\Component\Console\Output\OutputInterface  $output
      * @param  \Throwable  $e
      * @return void
+     *
+     * @internal This method is not meant to be used or overwritten outside the framework itself.
      */
     public function renderForConsole($output, Throwable $e)
     {
--- a/src/Illuminate/Foundation/Exceptions/Handler.php
+++ b/src/Illuminate/Foundation/Exceptions/Handler.php
@@ -693,7 +693,7 @@ class Handler implements ExceptionHandlerContract
      * @param  \Throwable  $e
      * @return void
      *
-     * @internal This method is not meant to be used or overwritten outside the framework itself.
+     * @internal This method is not meant to be used or overwritten outside the framework.
      */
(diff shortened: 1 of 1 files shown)


kieransimkin:
okay, so how do you do it then?! How do you catch exceptions thrown within commands, there must be a way, other than wrapping every command runner in a try..catch?! 

Why isn't this in the error reporting documentation? when you search for anything relating to "catch errors in commands in laravel", you get old articles telling you to overwrite the renderForConsole, and no official documentation from Laravel on how to achieve the same thing. This isn't good, if there is a solution perhaps you would point me to it, so that I can update the forum posts that are giving out of date information?

nunomaduro:
@kieransimkin Because we use GitHub hooks, issues get automatically closed once we merged the pull request that solves a particular issue. To catch exceptions from the console, you may use the `report` method of your application's exception handler.

Of course, you may use `app()->runningInConsole()` to see if the error is coming from the console.

kieransimkin:
My App's report function doesn't get called when an exception happens in console.

src/Illuminate/Foundation/Exceptions/Handler.php
```diff
@@ -692,6 +692,8 @@ protected function convertExceptionToArray(Throwable $e)
      * @param  \Symfony\Component\Console\Output\OutputInterface  $output
      * @param  \Throwable  $e
      * @return void
+     *
+     * @internal This method is not meant to be used or overwritten outside the framework.
```
mfn, line 696: This PR marks a method internal which is part of an interface?

`class Handler implements ExceptionHandlerContract` -> https://github.com/laravel/framework/blob/05b785d7d24375b5437ea115a363f6a63364f84a/src/Illuminate/Contracts/Debug/ExceptionHandler.php#L38-L45

PR #40941, opened 2022-02-10
[9.x] Patch regex rule parsing due to `Rule::forEach()`

### LINKED ISSUE #40924: [9.x] Regex validation rules for array items containing 'or' seperators cause ErrorException on input validation
- Laravel Version: 9.0.1
- PHP Version: 8.1.2
- Database Driver & Version: irrelevant

### Description:
Using an or sign (`|`) within a regex validation rule for an array item breaks on Laravel 9.x. This works fine on Laravel 8.x.

It seems somewhere the rules are parsed and the `|` is interpreted as start of another validation rule instead of as being part of the regex rule, causing the parsed regex rule to be `regex:/^(typeA|`. 

Most probably related to https://github.com/laravel/framework/pull/40498.

### Steps To Reproduce:

See https://github.com/jnoordsij/laravel/tree/regex-validation-bug-8.x for the (working) 8.x version and https://github.com/jnoordsij/laravel/tree/regex-validation-bug-9.x for the test breaking on 9.x.

1. clone repo
2. `composer install`
3. `cp .env.example .env.testing`
4. `php artisan key:generate --env=testing`
5. `php artisan test tests/Feature/SimpleRequestTest.php`

Resulting exception:
`ErrorException: preg_match(): No ending delimiter '/' found in ...`

stevebauman:
## Description

This PR fixes https://github.com/laravel/framework/issues/40924.

This also allows the ability to supply a regex validation rule inside of a single string-based rule, instead of requiring it to be an array.

**Before**:

This would previously fail in Laravel:

```php
$data = ['items' => [['type' => 'admin']]]

// Fails. preg_match throws malformed regex exception.
$rules = ['items.*' => 'regex:/^(super|admin)$/i'];

// Passes.
$rules = ['items.*' => ['regex:/^(super|admin)$/i']];

```

**After**:

```php
$data = ['items' => [['type' => 'admin']]]

// Passes.
$rules = ['items.*' => 'regex:/^(super|admin)$/i'];

// Passes.
$rules = ['items.*' => ['regex:/^(super|admin)$/i']];

```

## Notes

I've added various test cases to ensure parsing is done properly. However, maybe we should introduce an exception when `preg_match()` throws an exception due to a malformed regex which would indicate to the developer they must wrap the regex rule in an array when using multiple rules?

Let me know your thoughts. If you like the idea, I will update my PR to include this exception.

### DIFF
--- a/src/Illuminate/Validation/ValidationRuleParser.php
+++ b/src/Illuminate/Validation/ValidationRuleParser.php
@@ -88,7 +88,9 @@ class ValidationRuleParser
             [$name] = static::parseStringRule($rule);
 
             return explode('|', $rule, ...array_filter([static::ruleIsRegex($name)]));
-        } elseif (is_object($rule)) {
+        }
+
+        if (is_object($rule)) {
             return Arr::wrap($this->prepareRule($rule, $attribute));
         }
 
--- a/src/Illuminate/Validation/ValidationRuleParser.php
+++ b/src/Illuminate/Validation/ValidationRuleParser.php
@@ -87,7 +87,7 @@ class ValidationRuleParser
         if (is_string($rule)) {
             [$name] = static::parseStringRule($rule);
 
-            return explode('|', $rule, ...array_filter([static::ruleIsRegex($name)]));
+            return static::ruleIsRegex($name) ? [$rule] : explode('|', $rule);
         }
 
         if (is_object($rule)) {
--- a/src/Illuminate/Validation/ValidationRuleParser.php
+++ b/src/Illuminate/Validation/ValidationRuleParser.php
@@ -276,11 +276,7 @@ class ValidationRuleParser
      */
     protected static function parseParameters($rule, $parameter)
     {
-        if (static::ruleIsRegex($rule)) {
-            return [$parameter];
-        }
-
-        return str_getcsv($parameter);
+        return static::ruleIsRegex($rule) ? [$parameter] : str_getcsv($parameter);
     }
 
     /**
--- a/src/Illuminate/Validation/ValidationRuleParser.php
+++ b/src/Illuminate/Validation/ValidationRuleParser.php
@@ -156,7 +156,7 @@ class ValidationRuleParser
                     } else {
                         $this->implicitAttributes[$attribute][] = $key;
 
-                        $results = $this->mergeRules($results, $key, $rule);
+                        $results = $this->mergeRules($results, $key, Arr::wrap($rule));
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


src/Illuminate/Validation/ValidationRuleParser.php
```diff
@@ -85,8 +85,12 @@ protected function explodeRules($rules)
     protected function explodeExplicitRule($rule, $attribute)
     {
         if (is_string($rule)) {
-            return explode('|', $rule);
-        } elseif (is_object($rule)) {
+            [$name] = static::parseStringRule($rule);
+
+            return explode('|', $rule, ...array_filter([static::ruleIsRegex($name)]));
```
stevebauman, line 90: I've used the spread operator to completely omit the third (`$limit`) parameter if the rule **is not** regex.

I've added this, because `explode()` will take a `null` or `0` parameter and set the `$limit` to `1`.

We want `$limit` to be `1` when the rule is a regular expression, or `$limit` to be completely omitted when the rule is not.

stevebauman:
Hmm, a random test is failing. If we execute another run, we should be green 👍

taylorotwell:
I'm guessing this still doesn't work?

```php
'attribute' => 'required|string|regex:/^(super|admin)$/i|another_rule|final_rule`,

**Comment by stevebauman** [comment]
Yes that's correct -- the rule will fail to parse properly. I've added a test case so we have some visibility on this:

https://github.com/stevebauman/framework/blob/ae6b101abbf4665396a860e52537c2ea7c3fd6c8/tests/Validation/ValidationRuleParserTest.php#L127-L138

**Comment by taylorotwell** [comment]
@stevebauman can you explain how your original PR actually introduced this bug? I'm not sure I quite understand how and what broke.

**Comment by stevebauman** [comment]
@taylorotwell

I introduced this bug by using `Arr::flatten()` on this line:

https://github.com/laravel/framework/blob/4a474ced7889979f50e794fff9091eace57dc5bd/src/Illuminate/Validation/ValidationRuleParser.php#L145

The structure of `$rules` given to `explodeExplicitRule()` was changed:

```php
// Previously:
^ array:1 [
  0 => array:2 [
    0 => "in:foo"
    1 => "regex:/^(foo|bar)$/i"
  ]
]

// With Arr::flatten()
^ array:2 [
  0 => "in:foo"
  1 => "regex:/^(foo|bar)$/i"
]
```

The arrays were previously multi-dimensional and are now single dimensional.

This means, the first `is_string()` condition is now being hit on `explodeExplicitRule()`, instead of the last `prepareRule` array map, since this method will be called for each rule in the array:

https://github.com/laravel/framework/blob/4a474ced7889979f50e794fff9091eace57dc5bd/src/Illuminate/Validation/ValidationRuleParser.php#L85-L97

This causes the explode to inadvertently target the regex rules, when it did not previously.

For `Rule::forEach()` to work, I had to flatten the attributes `$rules` so that I could check each rule if it's an instance of `NestedRules`. Without `flatten()`, we would have to cycle over the rules again to determine if they contain any `Rule::forEach()` instances.

Issue #40854, opened 2022-02-07
Laravel quietly ignores Symfony's deprecations: @trigger_error()


rudiedirkx:
- Laravel Version: 8.73.2
- PHP Version: 7.4.26

### Description:

The Symfony ecosystem triggers deprecations like this:

    @trigger_error("Some deprecation", E_USER_DEPRECATED);

with the `@` to suppress error, **to make deprecations opt-in**. That's a decent idea, IF the app/framework's error handler CAN opt-in. The first thing Laravel's error handler does is:

    if (error_reporting() & $level) {

which means all of Symfony's deprecation triggers are always completely ignored. No opt-in possible.

I think Laravel shouldn't check for error suppression if it's a deprecation, because that's how Symfony does deprecations on purpose. This could work beautifully with `isDeprecation()` and `handleDeprecation()`, but it doesn't, and it's too early for the app's `Handler` to do/catch anything.

### Steps To Reproduce:

Trigger a deprecation error, anywhere in your app code, with any Laravel config:

    @trigger_error("Some deprecation", E_USER_DEPRECATED);

Without the `@` it's reported to my `deprecations.log`, but with the `@` it's quietly ignored. Symfony's standard is to do it with the `@`, so I feel that should work.

rudiedirkx:
Some opt-in details on https://github.com/twigphp/Twig/issues/2082

rudiedirkx:
No, you can't, because they never ever get past Laravel's error handler:

    if (error_reporting() & $level) {

That's the problem.....

X-Coder264:
I can confirm this issue.

My feature tests use `$this->withoutDeprecationHandling();` and I use the Faker library which recently deprecated calling the property name in favour of the method itself -> https://github.com/FakerPHP/Faker/blob/v1.19.0/src/Faker/Generator.php#L942

Since [Symfony silences those deprecations](https://github.com/symfony/deprecation-contracts/blob/v3.0.0/function.php#L25) the test doesn't fail while my expectation is for it to fail.

So TLDR;

```php
use Illuminate\Foundation\Testing\WithFaker;
use Tests\TestCase;

final class FooTest extends TestCase
{
    use WithFaker;

    public function testFoo(): void
    {
        $this->withoutDeprecationHandling();

        $this->faker->name;
    }
}
```

Expected result is for the test to fail as Faker emits `Since fakerphp/faker 1.14: Accessing property "name" is deprecated, use "name()" instead.` deprecation.
[...]

PR #40942, opened 2022-02-10
[8.x] Catch suppressed deprecation logs

nunomaduro:
This pull requests closes https://github.com/laravel/framework/issues/40854 by making Laravel's deprecation handler catch suppressed deprecations.

The motivation behind this pull request is the way Symfony manages deprecations, as they use the `@` operator like so:

```
@trigger_error('idn_to_utf8(): INTL_IDNA_VARIANT_2003 is deprecated', \E_USER_DEPRECATED);
```

### DIFF
--- a/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php
+++ b/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php
@@ -67,11 +67,11 @@ class HandleExceptions
      */
     public function handleError($level, $message, $file = '', $line = 0, $context = [])
     {
-        if (error_reporting() & $level) {
-            if ($this->isDeprecation($level)) {
-                return $this->handleDeprecation($message, $file, $line);
-            }
+        if ($this->isDeprecation($level)) {
+            return $this->handleDeprecation($message, $file, $line);
+        }
 
+        if (error_reporting() & $level) {
             throw new ErrorException($message, 0, $level, $file, $line);
         }
     }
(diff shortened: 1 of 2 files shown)


madpilot78:
I'm seeing a problem with this approach. Since it uses the Log facade, now it makes mocking that much less useful, since I have to expect various calls to it due to the framework trying to use the Log APIs.

For example I get tests using `Log::shouldReceive('warning')->times(2);` with:

```
Mockery\Exception\BadMethodCallException: Received Mockery_2_Illuminate_Log_LogManager::error(), but no expectations were specified
```

Investigating this it has multiple calls failing and being catched. First one is actually this:

```
Received Mockery_2_Illuminate_Log_LogManager::channel(), but no expectations were specified
```

from https://github.com/laravel/framework/blob/ff9ca06ecc67e7f665e823223bb638092b20ae4a/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php#L101

Not sure if I'm missing some basic way to fix the problem without having to rewrite a lot of tests.

Any suggestions?

madpilot78:
For information, at present I found a workaround by overriding the symfony `trigger_deprecation()` function in the autoloader before everything else. Quite horrible, but makes things work since I don't have time right now to properly address this.

I'll try to file a but report or pull request for this as soon as I have a good idea on the best way to address it.
Right now my best idea is to wrap the `handleDeprecation()` call in a conditional checking a configuration variable or env variable, to allow opting out of it, restoring old behavior if wanted.

nunomaduro:
@madpilot78 The https://github.com/laravel/framework/pull/41057 pull request should address your issue.

PR #40944, opened 2022-02-10
[9.x] Add "getChunk" method to Filesystem

imanghafoori1:
The `get` method is a wrapper around the `file_get_contents` but it does not give the option of reading a portion of a big file.
So the getChuck method can be useful.

### Backward Compatibility:
Adding more parameters to `get` method is not backward compatible in case someone overrides the method in a subclass.
so I decided to implement a new method. This is still theoretically a breaking change in case someone has implemented a `getChunck` method in a subclass with a different signature, which is very unlikely.

- The `get` method reuses the new method to avoid duplication of logic.

### DIFF
--- a/src/Illuminate/Filesystem/Filesystem.php
+++ b/src/Illuminate/Filesystem/Filesystem.php
@@ -50,8 +50,24 @@ class Filesystem
      */
     public function get($path, $lock = false)
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:

imanghafoori1:
My implementation is incorrect.

PR #28418, opened 2019-05-05
Create route through an iterator

aronpc:
Create a route group from an iterator.

So we can separate our routes into more files avoiding that they get too big:

./routes
├── api.php
├── api.posts.php
├── api.users.php
├── channels.php
├── web.php
├── web.posts.php
└── web.users.php

```
    protected function mapWebRoutes()
    {
        $directory = new \RecursiveDirectoryIterator(base_path('routes/'));
        $files = new \RegexIterator($directory, '/web(.*).php$/');

        Route::middleware('web')
            ->namespace($this->namespace)
            ->group($files);
    }
```

All the tests in the framework are passing and did not have any major modifications.

PR #40945, opened 2022-02-10
[9.x] Add support for passing array as the second parameter for the group method.

hossein-zare:
Hi.

**What's this PR for? (example):**
in the `RouteServiceProvider` file:
```php
Route::middleware('web')
   ->namespace($this->namespace)
      ->group(base_path('routes/account/web.php'));

Route::middleware('web')
   ->namespace($this->namespace)
      ->group(base_path('routes/account/staff.php'));
```
can be written as follows
```php
Route::middleware('web')
   ->namespace($this->namespace)
      ->group([
         base_path('routes/account/web.php'),
         base_path('routes/account/staff.php')
       ]);
```
Both of them with the same middleware and namespace in just an array.

**Reasons why this PR won't break other features:**
1. The original `group` method only accepts `\Closure|string` as the second parameter which is `$routes`, so passing an array won't mess around.

    **The original group:**
    ```php
    /**
      * Create a route group with shared attributes.
     *
     * @param  array  $attributes
     * @param  \Closure|string  $routes
     * @return void
     */
    public function group(array $attributes, $routes)
    ```

2. It's only one simple commit, I also tested it on a real/ready project.

I ❤ Laravel

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -369,10 +369,28 @@ class Router implements BindingRegistrar, RegistrarContract
      * Create a route group with shared attributes.
      *
      * @param  array  $attributes
-     * @param  \Closure|string  $routes
+     * @param  \Closure|array|string  $routes
      * @return void
      */
     public function group(array $attributes, $routes)
+    { 
+        if (is_array($routes)) {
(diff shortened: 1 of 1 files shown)


aronpc:
#28418

PR #18127, opened 2017-02-26
[5.4] Use fromRawAttributes when creating a newPivot

PhiloNL:
Casting attributes stored as JSON don't work properly when using a pivot model and setting the `$casts` property.

In 5.2 fromRawAttributes method was added to manually resolve this issue. https://github.com/laravel/framework/pull/13294

When using the new "using" method, for example: `$this->belongsToMany(Team::class, 'user_teams', 'user_id', 'team_id')->using(UserTeamPivot::class)`, the JSON is still encoded twice. It would be great if this could be changed so the `fromRawAttributes` method is used.

```php
class UserTeamPivot extends Pivot
{
    protected $casts = ['permissions' => 'collection'];
}
```

```php
// App\User::find(1)->teams->first()
 App\Team {#1437
     id: 1,
     name: "Team Name",
     pivot: App\UserTeamPivot {#1436
       user_id: 1,
       team_id: 1,
       permissions: ""[\"create_campaign\"]"",
     },
   }
```

When using the `fromRawAttributes` method this is resolved.

```php
// App\User::find(1)->teams->first()
 App\Team {#1437
     id: 1,
     name: "Team Name",
     pivot: App\UserTeamPivot {#1436
       user_id: 1,
       team_id: 1,
       permissions: "["create_campaign"]",
     },
   }
```

PR #40947, opened 2022-02-10
[9.x] Fix default pivot attributes

iamgergo:
This PR attempts to fix an issue, when a custom pivot model is being used with default attributes. Related PR: https://github.com/laravel/framework/pull/18127

----

Let's say I have the same pivot setup as the linked PR, but also, my pivot has default attributes:

```php
class UserTeamPivot extends Pivot
{
    protected $attributes = ['permissions' => '["create_campaign"]'];

    protected $casts = ['permissions' => 'json'];
}
```

It works well, when just creating a new instance:

```php
(new UserTeamPivot())->toArray(); // ['permissions' => ['create_campaign']]
```

But fails when creating a new pivot via the relation instance, it just simply overrides the default attributes that are defined in the pivot model's `$attributes` property:

```php
$user->teams()->newPivot()->toArray(); // []
```
----

In my opinion, this is a bug, because if I want the default attributes to be empty, then I won't define them, but if they are defined, then they should be present in a fresh pivot instance.

Casting should be working as before, and I think this is not a BC.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php
@@ -76,7 +76,9 @@ trait AsPivot
 
         $instance->timestamps = $instance->hasTimestampAttributes($attributes);
 
-        $instance->setRawAttributes($attributes, $exists);
+        $instance->setRawAttributes(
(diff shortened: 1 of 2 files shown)


crynobone:
This should definitely includes some tests

iamgergo:
I've added a test that covers the described issue. Let me know if something else should be added.

PR #40950, opened 2022-02-11
[9.x] Added dropForeignIdFor method to match foreignIdFor method

bretto36:
Hi All

So i recently started using these foreignIdFor methods to make my migrations look so much nicer. But i found that the down() method of the migration didn't look as nice

Currently i'm writing

`$table->foreignIdFor(\App\Models\Item::class);`

and then to drop 

`$table->dropForeign('item_id')`

This change introduces 

`$table->dropForeignIdFor(\App\Models\Item::class);`

And also the constrained scenario

`$table->foreignIdFor(\App\Models\Item::class)->constrained();`

This change introduces 

`$table->dropConstrainedForeignIdFor(\App\Models\Item::class);`

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -401,6 +401,22 @@ class Blueprint
     }
 
     /**
+     * Indicate that the given foreign key should be dropped.
+     *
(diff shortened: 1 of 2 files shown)


driesvints:
@bretto36 I believe the handling of `$column` in `dropForeignIdFor` is incorrect here. Please see https://github.com/laravel/framework/issues/43753

PR #40955, opened 2022-02-11
[9.x] Fix in rule for values with trailing backslash

stefket:
The In Rule fails, if the value has a trailing backslash.


driesvints:
If this is broken in 8.x as well please send this to 8.x instead then.

stefket:
> If this is broken in 8.x as well please send this to 8.x instead then.

Sorry, I will close this pull request and create another one for 8.x instead.

PR #40957, opened 2022-02-11
[8.x] Fix in rule

stefket:
The In Rule fails, if the value has a trailing backslash.


stefket:
I'll try to fix this for 7.3 and will move the test to the right position.

PR #40958, opened 2022-02-11
[10.x] Bump to Symfony 6.1

driesvints:
Already bump Symfony to v6.1 to test their dev version to catch bugs early on.


GrahamCampbell (approved):

camerontucker:
If Symfony 6.1 requires PHP 8.1 then Laravel probably should too?

https://symfony.com/blog/symfony-6-1-will-require-php-8-1
https://laravel.com/docs/9.x/upgrade

camerontucker:
Thanks @GrahamCampbell.

I'm developing with Sail (thus 8.1) but deploying to 8.0, so this caught me.

PR #40960, opened 2022-02-11
[10.x] Implement Responsable on query builder

kemp:
This is a draft PR. I will add a description, tests and the required information before I mark it as ready for review.


kemp:
Moved to `9.x`, because I had a misconception about how middleware worked.

PR #40961, opened 2022-02-11
Minor corrections to docblocks / comments

u01jmg3:
Cascading comments have been maintained 😄

### DIFF
--- a/src/Illuminate/Auth/DatabaseUserProvider.php
+++ b/src/Illuminate/Auth/DatabaseUserProvider.php
@@ -124,7 +124,7 @@ class DatabaseUserProvider implements UserProvider
             }
         }
 
-        // Now we are ready to execute the query to see if we have an user matching
+        // Now we are ready to execute the query to see if we have a user matching
         // the given credentials. If not, we will just return nulls and indicate
         // that there are no matching users for these given credential arrays.
         $user = $query->first();
--- a/src/Illuminate/Auth/SessionGuard.php
+++ b/src/Illuminate/Auth/SessionGuard.php
@@ -143,7 +143,7 @@ class SessionGuard implements StatefulGuard, SupportsBasicAuth
         $id = $this->session->get($this->getName());
 
         // First we will try to load the user using the identifier in the session if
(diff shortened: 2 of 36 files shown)


driesvints:
This unfortunately breaks the uniform styling of these commend blocks.

u01jmg3:
@driesvints: I made sure not to break the uniformity though... have you viewed the diff? Many times it's the same number of letters that's changed so with a monospaced font there is no change to the styling.

u01jmg3:
@driesvints: what happens when this is not possible? For example, some of the corrections involve adding or removing a single character (a array → a**n** array, an user → a~~n~~ user).

Does comment styling take precedent?

driesvints:
> Does styling take precedence over having a comment that uses the correct grammar? 😕

Yes but we're open to improve grammar as long as the styling is preserved.

u01jmg3:
> _Yes but we're open to improve grammar as long as the styling is preserved._

- How do you suggest this is achieved?
- Do you mean the whole comment must be rewritten to: 1) correct the grammar (most of the time a single character) and 2) comply with _"three spaces for the second line and two for the third line"_?
- The commas can be removed if you would prefer to omit them.

---

A handful of examples (to me, the styling is intact):

```diff
- // Now we are ready to execute the query to see if we have an user matching
+ // Now we are ready to execute the query to see if we have a user matching
  // the given credentials. If not, we will just return nulls and indicate
  // that there are no matching users for these given credential arrays.

- // Here we'll check if the given subset of where clauses contains any "or"
+ // Here we'll check if the given subset of where clauses contain any "or"
  // booleans and in this case create a nested where expression. That way
  // we don't add any unnecessary nesting thus keeping the query clean.

- // If the values implements the Arrayable interface we can just call this
+ // If the values implement the Arrayable interface we can just call this
  // toArray method on the instances which will convert both models and
  // collections to their proper array form and we'll set the values.

  // of the relation so that we will easily and quickly match them to their
- // parents without having a possibly slow inner loops for every models.
+ // parents without having a possibly slow inner loop for every model.
```

Issue #40959, opened 2022-02-11
MessageSent Event return wrong class


henryavila:
- Laravel Version: 9.0.2
- PHP Version: 8.0.12

### Description:
The Event `Illuminate\Mail\Events\MessageSent` return a `Symfony\Component\Mime\Email` as message, but the corrected should be `Symfony\Component\Mailer\SentMessage`.

This object (`Symfony\Component\Mailer\SentMessage`) contains informations about the sent mail that can't be retrieved with  `\Symfony\Component\Mime\Email`.

One example is the MessageId. In `SwiftMailer`, we could define the messageId, but now it's no more possible with `Symfony Mailer`.
The only way to the the message ID now is from `Symfony\Component\Mailer\SentMessage` that is not returned by `MessageSent` event.

One problem with the current class returned is when we use the mailer webhooks to log the mail delivery/open/etc.
If we sent the message using queue, we don't have access to `SentMessage` and the MessageId.

This simple code change solve the problem
```diff
public function send($view, array $data = [], $callback = null)
{
...
if ($this->shouldSendMessage($symfonyMessage, $data)) {
          $sentMessage = $this->sendSymfonyMessage($symfonyMessage);
-         $this->dispatchSentEvent($message, $data);
+         $this->dispatchSentEvent($sentMessage , $data);
            ...
        }
}
```

*ps: In this code sample, I'm considering just the class returned, not the validations required, the chanfe on `MessageSent` class, etc

It's not just more semantical, but give a lot of usefull data that can't be get from another way

driesvints:
We're checking if https://github.com/laravel/framework/pull/40963 could fix things.

driesvints:
Thanks. We'll try to patch this soon!

henryavila:
Amazing. Thanks @driesvints

PR #40964, opened 2022-02-11
[9.x] fix MessageSent Event return appropriate class

henryavila:
This PR implement the changs reported on https://github.com/laravel/framework/issues/40959
This only affect Laravel 9 and provide much more control and access to sent mail data


driesvints:
We're already working on this. Thanks

PR #40965, opened 2022-02-11
[9.x] Add auto discovery controller to API Resource Routes

thalissonbarbosas:
This PR allows users to add the API Resource Routes inside the Controller Route Groups without adding the controller.

E.g.

```php
Route::controller(GoalsController::class)->group(function () {
    Route::apiResource('/goals')->only('index');
});
```

This is usefull when you want to add a different route name for an api resource method, but keeping inside the controller group.

Tests added


taylorotwell:

PR #40977, opened 2022-02-12
[9.x] Add `assertClientError` to `TestResponse`

lucaxue:
I have added support for asserting that a test response has a client side error status code (>= 400, <500). Similar to how `assertSuccessful` uses the Symfony response class' `isSuccessful`, this assertion uses `isClientError`.

This allows for asserting that responses are caused by a client error, not necessarily needing to test which specific status code it has returned.

```php
$response->assertClientError();
```


taylorotwell:

PR #40982, opened 2022-02-12
[9.x] Fix loading env for serve command

### LINKED ISSUE #40978: Laravel 9 command serve does not use custom .env file
- Laravel Version: 9.0.1
- PHP Version: 8.0.15
- Database Driver & Version: php-mysqlnd-8.0.15-1.fc35 connected to mysql:8 (docker)  / sqlite3

### Description:

command `php artisan serve --env=testing` seems not to be working since it is still using default `.env` file instead.

### Steps To Reproduce:

With mysql running using default .env file (configured against mysql)
image

With mysql stopped using custom .env.testing file (configured against sqlite)
image

With mysql stopped moving .env.testing to .env file
image

Another easy test is not having a `.env` file at all and specify custom one:

image

This time results into a err 500 

```
local.ERROR: No application encryption key has been specified. {"exception":"[object] (Illuminate\\Encryption\\MissingAppKeyException(code: 0): No application encryption key has been specified. at /tmp/laravel-serve-bug/vendor/laravel/framework/src/Illuminate/Encryption/EncryptionServiceProvider.php:79)
```

Thanks

dehbka:
Fix issue [40978](https://github.com/laravel/framework/issues/40978)

### DIFF
--- a/src/Illuminate/Foundation/Console/ServeCommand.php
+++ b/src/Illuminate/Foundation/Console/ServeCommand.php
@@ -103,7 +103,7 @@ class ServeCommand extends Command
     protected function startProcess($hasEnvironment)
     {
         $process = new Process($this->serverCommand(), public_path(), collect($_ENV)->mapWithKeys(function ($value, $key) use ($hasEnvironment) {
-            if ($this->option('no-reload') || ! $hasEnvironment) {
+            if ($this->option('no-reload') || $hasEnvironment) {
                 return [$key => $value];
             }
 


driesvints:
Seems like https://github.com/laravel/framework/pull/37142 is the cause here.

taylorotwell:
I'd like some feedback from the original author of this check tbh. It feels weird to totally reverse a check without an explanation and understanding on why we are doing so.

ugorur:
I put this check because we are developed our laravel apps with docker compose file without env file. Without env file but with docker compose, artisan serve command didn't work.

ugorur:
> Still feels like we don't (or at least I don't) have a good understanding with what we're doing here? Does anyone?

this condition exist because if we don't have .env file, we should fill our env variables with global Environment variables. Before this control, serve command couldn't use global Environment variables

taylorotwell:
So how do you suggest we fix #40978 @ugorur

taylorotwell:
I actually can't recreate the issue. Does your `.env.testing` file actually contain `APP_ENV=testing`. If not - add it.

taylorotwell:
Please resubmit if my solution does not work for you.

PR #40984, opened 2022-02-13
[9.x] Adds `expectsOutputToContain` to the `PendingCommand`.

xiCO2k:
Hey guys,

This PR adds a method to the `PendingCommand` to be used for testing.

The main reason is to test if the output contains a text, I found this really helpful, to use for the CLI output, its like the `assertSee` method from the `Http` response.

## Example

```php

$this->artisan('Hello World')
      ->expectsOutputToContain('Hello');
```

Hope that helps,

Thanks,
Francisco.

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/InteractsWithConsole.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/InteractsWithConsole.php
@@ -23,6 +23,13 @@ trait InteractsWithConsole
     public $expectedOutput = [];
 
     /**
+     * All of the expected text to be present on the output.
+     *
+     * @var array
+     */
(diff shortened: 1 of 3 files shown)


driesvints:

xiCO2k:
sure thing! Thanks @driesvints

stancl:
@xiCO2k I noticed that the behavior is a bit different than `expectsOutput()` which may be confusing to the user. It seems that with `expectsOutputToContain()`, you need to actually trigger the command with something else, such as `assertExitCode(0)` in the test you PR'd here.

But if you simply swap `expectsOutput()` to `expectsOutputToContain()`, the test will make no assertions.

PR #35147, opened 2020-11-08
[8.x] Use array_key_first instead of foreach

J2TEAM:
Starting from PHP 7.3, there is a new built-in function called array_key_first() which will retrieve the first key from the given array without resetting the internal pointer. Check out the [documentation](https://www.php.net/manual/en/function.array-key-first.php) for more info.

I checked [composer.json](https://github.com/laravel/framework/blob/8.x/src/Illuminate/Auth/composer.json#L17) and see that we can use this function.

image

PR #32550, opened 2020-04-26
[7.x] array_key_first function used to get first credential key

sehmbimanvir:
When login is attempted and only password key is passed then a function `firstCredentialKey` is called in `src\Illuminate\Auth\EloquentUserProvider.php` which is used to get first key from passed credentials array. 

In this PR i have removed that **foreach** loop which returns the first key from credentials array. Instead of that i used `array_key_first` function.

PR #40989, opened 2022-02-13
[9.x] Use array_key_first

iamgergo:
Since PHP 7.3 there is a native function for retreiving the first key of the given array: https://www.php.net/manual/en/function.array-key-first.php

So a small refactoring can be done here. The behavior is exactly the same, actually the polyfill for this function is the current implementation.

Related PRs: #35147 and #32550. 

I don't think this has any feelable affect on the performance in real Laravel applications, as it was discussed. My point of view is just, why not to use native stuff where we can.

### DIFF
--- a/src/Illuminate/Auth/EloquentUserProvider.php
+++ b/src/Illuminate/Auth/EloquentUserProvider.php
@@ -141,9 +141,7 @@ class EloquentUserProvider implements UserProvider
      */
     protected function firstCredentialKey(array $credentials)
     {
-        foreach ($credentials as $key => $value) {
-            return $key;
-        }
+        return array_key_first($credentials);
     }
 
     /**


iamgergo:
The tests were failing (not only here, for the PRs submetted before this one as well). The reason was because the tests were not checking for the default `timeout => 30` option, also the `assertSame` is sensitive for the keys I belive, so I've updated the order as well, just to make the tests passing.

crynobone:
> The tests were failing (not only here, for the PRs submetted before this one as well)

Probably should create a separate PR for that?

iamgergo:
I thought about that, but since this is a very small change, I thought it's okay.

Let me know if I still should submit a separate PR.

iamgergo:
I created a separate PR for the tests, since this PR might not be merged. https://github.com/laravel/framework/pull/40995

alecpl:
firstCredentialKey() method should probably be removed as it is used in one place only.

iamgergo:
I think it can be useful, if I want to extend the class and override that method. I would not remove it, however it's true, it's used only once.

driesvints:

PR #40991, opened 2022-02-13
[9.x] Add new `ConvertStringBooleansToBoolean` middleware

slashequip:
## What

This PR adds a new `ConvertStringBooleansToBoolean` middleware to the frameworks. This middleware converts string-like booleans `"true"` or `"false"` into their true `boolean` type.

Upon acceptance, and merge, of this PR I would want to add a follow up PR to `laravel/laravel` to include this new middleware as part of the default stack - like `ConvertEmptyStringsToNull`.

## Why

In my recent Laravel adventures I was building a filtering system to a GET endpoint and sending these filters via a query string. I found that when validating these inputs that `booleans` were treated as strings, this prevented me from using the 'boolean' validation rule.

e.g; `/invoices?approved=true`

```php
// ...
public function rules()
{
    return [
        'approved' => ['boolean'],
    ];
}
// ...
```

The above pseudo-code  would result in a validation error being thrown as within the request object `'approved' => 'true'`. 

This new middleware resolves what I believe to be a developer inconvenience and normalises expectations when passing booleans in a query string.

### DIFF
--- a/src/Illuminate/Foundation/Http/Middleware/ConvertStringBooleansToBoolean.php
+++ b/src/Illuminate/Foundation/Http/Middleware/ConvertStringBooleansToBoolean.php
@@ -40,7 +40,7 @@ class ConvertStringBooleansToBoolean extends TransformsRequest
      */
     protected function transform($key, $value)
     {
-        return match($value) {
+        return match ($value) {
             'true' => true,
             'false' => false,
             default => $value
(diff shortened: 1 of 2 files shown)


driesvints:

slashequip:
Thanks @driesvints 🤙

igorgaming:
Converting 'true' | 'false' in `bool` by default can be a very bad solution for some projects. For example, if a user leaves a comment with the text 'true', his comment will be added to the database as '1'

punyflash:
I think this is edge case when you want to cast `true` and `false` strings and this is risky to add this by default.

taylorotwell:

PR #40995, opened 2022-02-14
[9.x] Fix HttpClientTest

iamgergo:
This PR fixes the failing HttpClientTest.

The reason was because the tests were not checking for the default 'timeout' => 30 option, also the assertSame is sensitive for the keys I belive, so I've updated the order as well.


iamgergo:
Ironically another test failed, but I belive it's not related to the current PR, it looks quite random. Can someone re-run the tests, please?

driesvints:

PR #41000, opened 2022-02-14
[9.x] Adds `Str::excerpt`

nunomaduro:
This pull request adds the `Str::excerpt` method to Laravel 9.

The original idea comes from Rails [`ActionView::Helpers::TextHelper::excerpt`](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-excerpt), and extracts an excerpt from the given `$text` that matches the first instance of `$phrase`.

```php
public static function excerpt($text, $phrase = '', $options = [])
```

In other words, it just creates a relevant excerpt that contains the given `$phrase`:

```php
// Description: Laravel is a web application framework with expressive, elegant syntax.
// Search: Framework
$article = Article::search($request->search)->get()->first();

// Excerpt: ...plication framework with expr...
$excerpt = str($article->description)->excerpt($request->search, [
    'radius' => 10,
]);
```

The `radius` option expands the excerpt on each side of the first occurrence of `$phrase` by the number of characters defined in `radius` (which defaults to 100). If the excerpt radius overflows the beginning or end of the `$text`, then the `$omission` option (which defaults to "...") will be prepended/appended accordingly:
```php
// Excerpt: ...on framework wi...
// The left side with radius is: "on ".
// The right side with radius is: " wi".
$excerpt = str($article->description)->excerpt($request->search, [
    'radius' => 3,
]);

// Excerpt: ???on framework wi???
$excerpt = str($article->description)->excerpt($request->search, [
    'radius' => 3,
    'omission' => '?'
]);
```

Of course, if the excerpt radius does not overflows the beginning or end of the `$text`, the no omission is applied:
```
// Excerpt: '......el  Framework'
$excerpt = str('Laravel Framework')->excerpt('Framework', [
    'radius' => 3,
    'omission' => '......'
]);
```

If the `$phrase` isn't found, `null` is returned:

```php
// Excerpt: null
$excerpt = str($article->description)->excerpt('This does not exist on text');
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -266,7 +266,7 @@ class Str
         $radius = $options['radius'] ?? 100;
         $omission = $options['omission'] ?? '...';
 
-        preg_match('/^(.*?)('.preg_quote((string) $phrase).')(.*)$/i', (string) $text, $matches);
+        preg_match('/^(.*?)('.preg_quote((string) $phrase).')(.*)$/iu', (string) $text, $matches);
 
         if (empty($matches)) {
             return null;
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -266,7 +266,7 @@ class Str
         $radius = $options['radius'] ?? 100;
         $omission = $options['omission'] ?? '...';
 
-        preg_match("/^(.*?)(".preg_quote((string) $phrase).")(.*)$/i", (string) $text, $matches);
+        preg_match('/^(.*?)('.preg_quote((string) $phrase).')(.*)$/i', (string) $text, $matches);
 
         if (empty($matches)) {
             return null;
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -254,6 +254,42 @@ class Str
     }
 
     /**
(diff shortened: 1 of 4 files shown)


martindilling:
So `highligt` next? ;)
https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-highlight

src/Illuminate/Support/Str.php
```diff
@@ -253,6 +253,42 @@ public static function endsWith($haystack, $needles)
...
+     * @param  string  $phrase
+     * @param  array  $options
+     * @return string|null
+     */
+    public static function excerpt($text, $phrase = '', $options = [])
+    {
+        $radius = $options['radius'] ?? 100;
+        $omission = $options['omission'] ?? '...';
+
+        preg_match_all("/^(\W*|.*)(".preg_quote((string) $phrase).")(.*\W*)$/i", (string) $text, $matches);
```
olivernybroe, line 269:
```suggestion
        preg_match("/^(.*?)(".preg_quote((string) $phrase).")(.*)$/i", (string) $text, $matches);
```

Regex can be simplified to get rid of `\W` and I added a lazy token to the first match, to remove the need for `preg_match_all`.

This also means that the following statements that uses `$matches`  don't be wrapped in a `collect#first`.

PR #41001, opened 2022-02-14
[9.x] Expose next and previous cursor of cursor paginator

gdebrauwer:
The json response of a `CursorPaginator` does currently not expose the raw values of the next cursor and previous cursor (it only exposes full URLs), while that is something that should be exposed by default in my opinion. This is similar to the `LengthAwarePaginator`, where the values of the current / next / last page are also exposed.


taylorotwell:
Think it makes sense to serialize them as `next_cursor` and `prev_cursor` instead?

gdebrauwer:
I chose to use `_page` suffix because of the existing `next_page_url` / `prev_page_url` fields, but you are probably right that it is better if we use the `_cursor` suffix for these new fields 👍

PR #41003, opened 2022-02-14
[9.x] Cursor pagination: convert original column to expression

gdebrauwer:
Imagine you write the following query:
```php
User::query()
    ->select('*')
    ->selectSub('CONCAT(firstname, \' \', lastname)', 'fullname')
    ->orderBy('fullname')
    ->orderBy('id')
    ->cursorPaginate()
```
According to the [documentation](https://laravel.com/docs/8.x/pagination#cursor-vs-offset-pagination), this should work:
> Query expressions in "order by" clauses are supported only if they are aliased and added to the "select" clause as well.

But this query will only work when fetching the first page. When you fetch the users of the second page / next cursor, you will get the following error:
```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'CONCAT(firstname, ' ', lastname)' in 'where clause'
```
This is caused by the fact that the raw output of the `getOriginalColumnNameForCursorPagination` method is provided to a `where` method. This means that the original column value is interpreted as a real column name. That is not the case when aliasing a dynamic column.

By wrapping the result of the `getOriginalColumnNameForCursorPagination` method in an `\Illuminate\Database\Query\Expression` object, we ensure that the original column value is correctly interpreted.

### DIFF
--- a/src/Illuminate/Database/Concerns/BuildsQueries.php
+++ b/src/Illuminate/Database/Concerns/BuildsQueries.php
@@ -13,6 +13,7 @@ use Illuminate\Pagination\LengthAwarePaginator;
 use Illuminate\Pagination\Paginator;
 use Illuminate\Support\Collection;
 use Illuminate\Support\LazyCollection;
+use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Conditionable;
 use InvalidArgumentException;
 use RuntimeException;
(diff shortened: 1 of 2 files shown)


driesvints:
Seems like tests are still failing here. Please fix the failing tests and mark as ready.

taylorotwell:
Won't this break columns that actually need to be quoted?

gdebrauwer:
Oh I did not know that in certain scenario's the columns need to be quoted 🤔  I changed the logic so it checks if the original column is an expression. It does this by checking if the original column starts with a `(` and ends with a `)`.

gdebrauwer:
To do this I also needed to use the grammar of the builder in the `getOriginalColumnNameForCursorPagination`. This is because the alias may already be wrapped in quotes and the current check `$parameter === $alias` would not pass in that case, because `$parameter` is not quoted while the `$alias` might be.

I'm not sure if it is allowed to use the grammar in that place 🤔

GrahamCampbell:
> It does this by checking if the original column starts with a ( and ends with a ).

That logic looks wrong to me. Different sql engines have different rules.

gdebrauwer:
I chose to use that logic because the `selectSub` method also just wraps the subquery in parentheses, regardless of the sql engine.

https://github.com/laravel/framework/blob/3a9d6097613561fa92315807631e8d5dc523d655/src/Illuminate/Database/Query/Builder.php#L269-L271

PR #41008, opened 2022-02-14
[9.x] Fix dispatch_now behavior for fail call

driesvints:
At the moment when you call dispatch_now on a job and that job contains a manual `->fail()` call [(documented here)](https://laravel.com/docs/9.x/queues#manually-failing-a-job), the app will break and an Undefined array key "job" error is thrown.

This is because of this line that passes an empty array as the job payload: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Bus/Dispatcher.php#L115

I'm not sure what the correct solution here is... Haven't investigated yet if this is also the case on Laravel v8.

See https://github.com/laravel/lumen-framework/issues/1217


taylorotwell:
We dug into this a while back (maybe in 2020?). I believe it is the reason we introduced and now only document the `dispatchSync` method and no longer document `dispatch_now` at all.

PR #41011, opened 2022-02-14
[9.x] Make:model --morph flag to generate MorphPivot model

michael-rubel:
- This PR adds the possibility to generate `MorphPivot` model using `make:model` command in Laravel 9.

### Why?
Had to create a lot of models for polymorphic tables, and I always miss the option to include the `MorphPivot` instead of the basic one on `make:model` immediately, so I think it will be helpful for those who work with polymorphic tables on a daily basis.


taylorotwell:
Rename option to `morph-pivot`.

PR #41012, opened 2022-02-14
Default options for Http requests

TiiFuchs:
This adds a defaultOptions() method to the Http\Client\Factory class, that can be called in the AppServiceProvider to set some Guzzle options that should be applied to each and every request that follow.

The most relatable use-case that comes to my mind are proxy options that should apply to all upcoming requests.

You can use it in your AppServiceProvider like this
```php
Http::defaultOptions([
	'proxy'  => 'localhost:9090',
	'verify' => false
]);
```

And every request created through the Factory should get the same options automatically.

`Http::get('http://foo.com')` gets automatically send through the proxy.

If you need another setting for a specific request you still can overwrite the options with the `Http::withOptions(['verify' => true])->get('http://bar.com')`.

This does not break any existing features, because in `newPendingRequest()` we pass an empty array to `withOptions()` when `defaultOptions()` is never called before. With the current array_merge_recursively implementation in withOptions nothing is changed.

It's backwards compatible too, because it's one additional function that has no effect if not called.

This implementation makes it a little easier to have an app specific proxy setting or other Guzzle settings, that should apply to every request. Before you would have to create a macro and call it on every request. But 3rd-party libraries using the Http Factory would not be affected by this since they would not use the user-created Macro. Another solution would be to overwrite the Factory class to implement this functionality on a project basis. A direct implementation in the Factory class has the benefit that everyone can use it out of the box.

### DIFF
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -7,6 +7,7 @@ use GuzzleHttp\Promise\PromiseInterface;
 use GuzzleHttp\Psr7\Response as Psr7Response;
 use GuzzleHttp\TransferStats;
 use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
 use PHPUnit\Framework\Assert as PHPUnit;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Avoiding add global options like this - use a macro or a helper method within some class that is responsible for calling a given service.

PR #41014, opened 2022-02-14
[9.x] Adds `Attribute::with`

bvtterfly:
- This PR adds the possibility to create `Attribute` using static `with` method to Laravel 9.

**Why?**

Laravel has a very expressive syntax, but I think the current syntax of `Attributes` isn't like other parts of laravel. So I made this PR with a cleaner and more readable way to create them, significantly when disabling object caching.

Current approach:

```

    public function uppercase(): Attribute
    {
        return (new Attribute(
            get: fn ($value) => strtoupper($value),
            set: fn ($value) => strtoupper($value)
        ))->withoutObjectCaching();
    }


```
New approach:

```

    public function uppercase(): Attribute
    {
        return Attribute::with(
            get: fn ($value) => strtoupper($value),
            set: fn ($value) => strtoupper($value)
        )->withoutObjectCaching();
    }


```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Casts/Attribute.php
+++ b/src/Illuminate/Database/Eloquent/Casts/Attribute.php
@@ -47,10 +47,7 @@ class Attribute
      */
     public static function make(callable $get = null, callable $set = null): static
     {
-        return new static(
-            get: $get,
-            set: $set
-        );
+        return new static($get, $set);
     }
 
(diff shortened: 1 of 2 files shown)


edemots:
Hey! Since it mimics other parts of Laravel behavior for instantiation, why not use `make` as the method name?

PR #41017, opened 2022-02-15
[9.x] Fix the error message when no routes are detected

LukeTowers:
This fixes the error message that's displayed when Laravel can't find any routes to list. Router->getRoutes() returns a RouteCollectionInterface, not an array.

### DIFF
--- a/src/Illuminate/Foundation/Console/RouteListCommand.php
+++ b/src/Illuminate/Foundation/Console/RouteListCommand.php
@@ -98,7 +98,7 @@ class RouteListCommand extends Command
     {
         $this->router->flushMiddlewareGroups();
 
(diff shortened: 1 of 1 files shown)


src/Illuminate/Foundation/Console/RouteListCommand.php
```diff
@@ -98,7 +98,7 @@ public function handle()
     {
         $this->router->flushMiddlewareGroups();
 
-        if (empty($this->router->getRoutes())) {
+        if (!$this->router->getRoutes()->count()) {
```
LukeTowers, line 101:
```suggestion
        if (! $this->router->getRoutes()->count()) {
```

PR #41021, opened 2022-02-15
[9.x] Fix Duplicate Route Namespace

moisish:
When you are using group namespaces you cannot use the group controller method or single action controllers. 
Also you cannot use the controller class as a parameter on your route

```
Route::namespace($module->namespace().'\App\Http\Controllers')->group($module->getRoutes());
```

This works

```
Route::get('/', 'FrameworkController@index')->name('index');
```

The following scenarios are not working

Controller group
```
use Modules/Foo/App/Http/Controllers/FooController;
Route::prefix('foo')
    ->controller(FooController::class)
    ->group(function () {
        Route::get('/foo', 'foo');
    });
```

Single Action Controller
```
use Modules/Foo/App/Http/Controllers/InvokableController;
Route::prefix('foo')->group(function () {
        Route::get('/foo', InvokableController::class);
    });
```

```
use Modules/Foo/App/Http/Controllers/FooController;

Route::prefix('foo')->group(function () {
        Route::get('/', [FooController::class, 'index']);
    });
```

The solution I came up with is to not prepend the group namespace if the controller class starts with the same namespace

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -544,7 +544,7 @@ class Router implements BindingRegistrar, RegistrarContract
     {
         $group = end($this->groupStack);
 
-        return isset($group['namespace']) && ! str_starts_with($class, '\\')
+        return isset($group['namespace']) && ! str_starts_with($class, '\\') && ! str_starts_with($class, $group['namespace'])
                 ? $group['namespace'].'\\'.$class : $class;
     }
 


driesvints:
If you use controller classes in your route it's intended to not set the namespace. That's why it's disabled by default in the skeleton. You either use one or the other.

moisish:
I get that it's disabled but is there a reason not being flexible no matter if you set a namespace?

moisish:
When you have namespaced groups you cannot use the new controller method or even invokable controllers which I think we should allow to.

Main reason we have namespaced groups is when used with modules/domains which means including multiple route files.
Then we can check the namespace of the current route ```Route::getCurrentRoute()->getAction('namespace')```

Invokable controllers should not be affected by group namespaces tho

driesvints:
> When you have namespaced groups

That's what I'm trying to say: I think we removed this from the docs to no longer promote namespaced groups.

PR #41024, opened 2022-02-15
[9.x] Fix loadAggregate not correctly applying casts

ahme-d:
Currently "loadExists" only applies the "bool cast" on the first model in a collection and skips the rest.
 
This should be merged to 8.x as well.


driesvints:
Please send this to 8.x

PR #41026, opened 2022-02-15
Use PHP enums in migrations

levu42:
Currently, when I have a migration, I can create an enum column by writing:

```php
$table->enum('user_role', ['admin', 'member']);
```

Since PHP 8.1 supports now native enums, it would be great if I could do the following:

```php
enum UserRole: string
{
    case Admin = 'admin';
    case Member = 'member';
}
```
```php
$table->enum('user_role', UserRole::class);
```

This pull requests adds this functionality.


GrahamCampbell:
I think this is actually a bad idea. It means you can never edit the enum.

driesvints:
See previous attempts

PR #41028, opened 2022-02-15
[9.x] Fix mails with tags and metadata are not queuable

joostdebruijn:
In PR #40783 functionality is added to mails to add tags en metadata to supported mail drivers. However, this is not working when queuing mails, because `Serialization of 'Closure' is not allowed`. This PR fixes that problem and is adding some unit tests for this functionality.

### DIFF
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -172,11 +172,11 @@ class Mailable implements MailableContract, Renderable
     protected $metadata = [];
 
     /**
-     * The tag for the message.
+     * The tags for the message.
      *
-     * @var string
+     * @var array
      */
-    protected $tag;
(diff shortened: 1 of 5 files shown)


src/Illuminate/Mail/Mailable.php
```diff
@@ -164,6 +164,20 @@ class Mailable implements MailableContract, Renderable
...
+     * @var array
+     */
+    protected $metadata = [];
+
+    /**
+     * The tag for the message.
+     *
+     * @var string
+     */
+    protected $tag;
```
driesvints, line 179: This isn't correct. You should be able to add multiple tags.

joostdebruijn: Corrected.

driesvints (approved):

PR #41029, opened 2022-02-15
[9.x] Add keyBy() method to Arr helpers

medeirosinacio:
Hi guys, this is my first PR. Please let me know if you have anything to change or improve. This method was inspired by the [ArrayHelper::index of the Yii2 framework](https://www.yiiframework.com/doc/api/2.0/yii-helpers-basearrayhelper#index()-detail). I used this method a lot to format arrays and I would really like laravel to have it too

Indexes the array according to a specified key.

use case
```
$array = [
    ['id' => '123', 'data' => 'abc', 'device' => 'laptop'],
    ['id' => '345', 'data' => 'def', 'device' => 'tablet'],
    ['id' => '345', 'data' => 'hgi', 'device' => 'smartphone'],
];
$result = Arr::index($array, 'id');
```

The result will be an associative array, where the key is the value of id attribute
```
[
    '123' => ['id' => '123', 'data' => 'abc', 'device' => 'laptop'],
    '345' => ['id' => '345', 'data' => 'hgi', 'device' => 'smartphone']
    // The second element of an original array is overwritten by the last element because of the same id
]
```

- Renamed Arr:index to Arr:keyBy

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -744,4 +744,34 @@ class Arr
 
         return is_array($value) ? $value : [$value];
     }
+
+    /**
+     * Indexes the array according to a specified key.
+     *
+     * @param iterable $array
+     * @param string|int|null $key
+     * @return array the indexed array
+     */
+    public static function index($array, $key)
+    {
+        $results = [];
+
+        foreach ($array as $item) {
+
+            $keyValue = data_get($item, $key);
+
+            if (is_null($keyValue)) {
+                $results[] = $item;
+                continue;
+            }
+
+            if (is_object($keyValue) && method_exists($keyValue, '__toString')) {
+                $keyValue = (string)$keyValue;
+            }
+
+            $results[$keyValue] = $item;
+        }
+
+        return $results;
+    }
 }
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -757,7 +757,6 @@ class Arr
         $results = [];
 
         foreach ($array as $item) {
-
             $keyValue = data_get($item, $key);
 
             if (is_null($keyValue)) {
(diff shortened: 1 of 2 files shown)


pyrou:
About naming, same behavior exists on Collection, its named [`keyBy`](https://github.com/laravel/framework/blob/master/src/Illuminate/Collections/Collection.php#L516)

medeirosinacio:
> Sobre a nomeação, o mesmo comportamento existe na coleção, seu nome[`keyBy`](https://github.com/laravel/framework/blob/master/src/Illuminate/Collections/Collection.php#L516)



> 

Yes true! I can change the name of the method to leave both with the same behavior. It would be interesting for the Arr class to also have this method.

medeirosinacio:
> About naming, same behavior exists on Collection, its named [`keyBy`](https://github.com/laravel/framework/blob/master/src/Illuminate/Collections/Collection.php#L516)

Yes true! I can change the name of the method to leave both with the same behavior. It would be interesting for the Arr class to also have this method.

dennisprudlo:
If this array helper and the collection function are supposed to do the same we maybe should simply use 
```php
collect($array)->keyBy('id')
```
or if merged update the `keyBy` function of the collection to use the `Arr::index()` helper. Otherwise we'd end up with two different version of this kind of function.

taylorotwell:
If we already have a collection method that does the **exact** same thing then this method should be named the same as well.

medeirosinacio:
> If we already have a collection method that does the **exact** same thing then this method should be named the same as well.

Hello, I refactored the method to use the function that already exists in the collection.

Let Me know if anything is missing.

medeirosinacio:
> 

Yes, I made the change so that the method uses the existing function in the collection.

medeirosinacio:
> If this array helper and the collection function are supposed to do the same we maybe should simply use
> 
> ```
> collect($array)->keyBy('id')
> ```
> 
> or if merged update the `keyBy` function of the collection to use the `Arr::index()` helper. Otherwise we'd end up with two different versions of this kind of function.

Yes, I made the change so that the method uses the existing function in the collection

PR #41036, opened 2022-02-15
simplifying getAlias method in Container class

navidhasanitabar:
getAlias in Container class is unnecessarily complex; i make it simpler; there was a recursive-call in the method that IMO won't add any-logic to the method; so i deleted it and replaced it with a more intuitive expression;


driesvints:
Heya, this seems to break the test suite. Please resubmit this with passing tests.

Issue #38828, opened 2021-09-15
SQLSTATE[01000]: Warning: 1265 Data truncated for column 'date_field' at row 1


EgorGruzdev:
- Laravel Version: 8.61.0
- PHP Version: 8.0.10
- Database Driver & Version: MySQL 8 (strict mode)

### Description:

Colleagues, when working in strict mode with a database, we get an error, because after the implementation of this functionality #38720, the date fields are saved to the database in the format specified in the grammar, ignoring the caster parameters, i.e. in case of mysql format "Y-m-d H:i:s", although the database for the date type expects "Y-m-d".

In version 8.60, the format coincided with the format specified in the caste.

### Steps To Reproduce:

```php
class TestModel1 extends Model
{
    public $table = 'test_model1';
    protected $guarded = [];

    public $casts = [
        'date_field' => 'date:Y-m-d',
    ];
}

TestModel1::create([
          'date_field' => '2019-10-01',
]);
```

EgorGruzdev:
Also, if the $dateFormat property is specified, then in version 8.60, if there was a caste on the field, then this format was ignored and taken from the caste attributes. In version 8.61, the caste attributes are ignored and the $dateFormat property is always used.

daniser:
I agree, since 8.61 $dateFormat or grammar's default date format have priority over custom caster format. In fact, caster format is never used anymore.

derekmd:
Last week's pull request fixed `Model@setAttribute()` falsely marking an attribute as dirty when assigning  `Carbon` objects. However the new test case added didn't have coverage for assigning _strings_ which is the common use case when `$request->validated()` is filled into the model.

This is a `setAttribute()` change that could potentially fix it: https://github.com/laravel/framework/compare/8.x...derekmd:fix-setattribute-custom-cast-date-format?expand=1

However I need to figure out a more ~~eloquent~~ graceful solution to conditionally invoke `fromDateTime()` with the custom date format without introducing a new argument in 8.x. That commit introduces some duplicate code with `addCastAttributesToArray()` so the casting lifecycle is now a bit weird.

Ilyes512:
I also came across this issue because of an failing test case. I have a model "Company" with an date field 'founded_at':

```php
class Company extends Model
{
    protected $casts = [
        'founded_at' => 'date:Y-m-d',
    ];
}
```

In my tests I use `assertDatabaseHas` and I need to set different `founded_at` values for different versions of Laravel.

[...]

PR #41040, opened 2022-02-15
[9.x] MySQL driver might modify custom casted date

Naoray:
This PR only shows that the custom date formatting is not working in all cases. **it does not implement a fix**, because I am not 100% sure where to fix this issue and why it occurs in the first place or whether it should be fixed on the config level of the MySQL driver.

## Explanation
Setting a custom date format which does not follow the American approach and list the year first, will be altered when the related model is saved to the MySQL DB driver.

```php
class TestModel extends Model
{
    protected $casts = [
        'custom_date' => 'date:d.m.Y'
    ];
}

TestModel::create(['custom_date' => '26.01.2022']);

//  custom_date is set to 20.01.2026 instead of the expected date
TestModel::first()->custom_date->format('d.m.Y') === '26.01.2022'; // false
```

## Attempted fixes
```php
// Illuminate\Database\Eloquent\Concerns\HasAttributes

public function setAttribute($key, $value)
{
    //...

    elseif ($value && ($this->isDateAttribute($key)  || $this->isDateCastableWithCustomFormat($key))) { // added last statement
        $value = $this->fromDateTime($value);
    }

    //...
}
```
**or**
```php
// Illuminate\Database\Eloquent\Concerns\HasAttributes

protected function isDateAttribute($key)
{
    return in_array($key, $this->getDates(), true) ||
        $this->isDateCastable($key) ||
        $this->isDateCastableWithCustomFormat($key); // added that line
}
```

Both attempts fixed the issue, however they resulted in the failing of `EloquentModelDateCastingTest::testDatesFormattedAttributeBindings` test

```
1) Illuminate\Tests\Integration\Database\EloquentModelDateCastingTest\EloquentModelDateCastingTest::testDatesFormattedAttributeBindings
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
-    0 => '2019-10-01'
+    0 => '2019-10-01 00:00:00'
     1 => '2019-10-01 10:15:20'
-    2 => '2019-10-01'
-    3 => '2019-10-01 10:15'
+    2 => '2019-10-01 00:00:00'
+    3 => '2019-10-01 10:15:00'
 )
```

### DIFF
--- a/tests/Integration/Database/MySql/EloquentCustomDateCastingRetrivalTest.php
+++ b/tests/Integration/Database/MySql/EloquentCustomDateCastingRetrivalTest.php
@@ -1,0 +1,40 @@
+<?php
+
+namespace Illuminate\Tests\Integration\Database\MySql;
+
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+class EloquentCustomDateCastingRetrivalTest extends MySqlTestCase
+{
+    protected function defineDatabaseMigrationsAfterDatabaseRefreshed()
+    {
+        Schema::create('test_models', function (Blueprint $table) {
+            $table->id('id');
+            $table->date('custom_date');
+        });
+    }
+
+    protected function destroyDatabaseMigrations()
+    {
+        Schema::drop('test_models');
+    }
+
+    public function testCustomDateMatchesGivenDateBeforeSave()
+    {
+        TestModel::create(['custom_date' => '26.01.2022']);
+
+        $this->assertEquals('26.01.2022', TestModel::first()->custom_date->format('d.m.Y'));
+    }
+}
+
+class TestModel extends Model
+{
+    public $timestamps = false;
+    protected $guarded = [];
+
+    protected $casts = [
+        'custom_date' => 'date:d.m.Y'
+    ];
+}


derekmd:
What is the output of below, ignoring Eloquent on database retrival?

```php
TestModel::create(['custom_date' => '26.01.2022']);
DB::table('test_models')->value('custom_date');
```

I assume this will return "2026-01-20 00:00:00" and the `INSERT` query has the expected parameter binding of `'26.01.2022'`.

---

Is this an issue with MySQL not supporting the given `d.m.Y` date format?

I just attempted the `TestModel` example on MariaDB and received this fatal error:

```
Illuminate\Database\QueryException with message 'SQLSTATE[22007]:
Invalid datetime format: 1292 Incorrect date value: '26.01.2022' for column `forge`.`test_models`.`custom_date`
```

It may be related to database strict mode settings in my local environment.

```sql
SELECT @@GLOBAL.sql_mode;
-- "NO_ZERO_IN_DATE,NO_ZERO_DATE,NO_ENGINE_SUBSTITUTION"
```

The GitHub Actions environment might have looser configuration that attempts to interpret date substitutions?

---

Currently any Eloquent model custom date format allows raw attribute values to be set without going through `Carbon` re-formatting:

```php
TestModel::make(['custom_date' => '26.01.2022'])->getAttributes();
// [
//     "custom_date" => "26.01.2022",
// ]

TestModel::make(['custom_date' => '26.01.2022 any text'])->getAttributes();
// [
//     "custom_date" => "26.01.2022 any text",
// ]
```

So Eloquent trusts the dev is providing a payload supported for the given database driver. Updating `setAttribute()` behavior would be a pretty big breaking change on a patch release. You can see what happened with the first PR attempted to fix https://github.com/laravel/framework/issues/38828 on 8.x.

---

For supporting this date format in the UI and presentation layer, you may need to use [attribute set mutators](https://laravel.com/docs/eloquent-mutators#defining-an-accessor) or [custom cast attributes](https://laravel.com/docs/9.x/eloquent-mutators#custom-casts). That `Model::$cast` / `Model::$date` format is really only useful for Eloquent checks on when model attributes are "dirty" and then for API endpoints returning `Model::toArray()` or `JsonResource::toArray()` results.

Naoray:
> What is the output of below, ignoring Eloquent on database retrival?
> 
> ```
> TestModel::create(['custom_date' => '26.01.2022']);
> DB::table('test_models')->value('custom_date');
> ```
> 
> I assume this will return "2026-01-20 00:00:00" and the `INSERT` query has the expected parameter binding of `'26.01.2022'`.

I get `"2026-10-20"` from `DB::table('test_models')->value('custom_date');`.

> Is this an issue with MySQL not supporting the given `d.m.Y` date format?

Could be, but I'd expect that the framework recognizes that the attribute `custom_date` is a date value with a custom format and parses it to a DB accepted format. In fact the modification of the `isDateAttribute()` method would do exactly that.

```php
// Illuminate\Database\Eloquent\Concerns\HasAttributes

protected function isDateAttribute($key)
{
    return in_array($key, $this->getDates(), true) ||
        $this->isDateCastable($key) ||
        $this->isDateCastableWithCustomFormat($key); // added that line
}
```

> I just attempted the `TestModel` example on MariaDB and received this fatal error:
> 
> ```
> Illuminate\Database\QueryException with message 'SQLSTATE[22007]:
> Invalid datetime format: 1292 Incorrect date value: '26.01.2022' for column `forge`.`test_models`.`custom_date`
> ```

That's the same error as in the [MariaDB Github Action Test Run](https://github.com/laravel/framework/runs/5202616402?check_suite_focus=true).


> So Eloquent trusts the dev is providing a payload supported for the given database driver. Updating `setAttribute()` behavior would be a pretty big breaking change on a patch release. You can see what happened with the first PR attempted to fix #38828 on 8.x.
> 
> For supporting this date format in the UI and presentation layer, you may need to use [attribute set mutators](https://laravel.com/docs/eloquent-mutators#defining-an-accessor) or [custom cast attributes](https://laravel.com/docs/9.x/eloquent-mutators#custom-casts). That `Model::$cast` / `Model::$date` format is really only useful for Eloquent checks on when model attributes are "dirty" and then for API endpoints returning `Model::toArray()` or `JsonResource::toArray()` results.

From my understanding that is exactly what the date cast with custom format allows you to do. At least that is how I read the [documentation](https://laravel.com/docs/9.x/eloquent-mutators#date-casting). I'd expect the date attribute to be casted to a DateTime Object when the attribute is set. Currently however, custom date formats are not casted to DateTime/Carbon objects and therefore it's up for the DB driver to interpret the given value.

As a workaround I currently have a trait `DetectsCastableDatesWithCustomFormats` that overrides the `isDateAttribute()` which is called in the `setAttribute()` method and basically decides whether a date attribute should be casted to a DateTime object or not.

```php
trait DetectsCastableDatesWithCustomFormats
{
    protected function isDateAttribute($key)
    {
        return in_array($key, $this->getDates(), true) ||
            $this->isDateCastable($key)
            || $this->isDateCastableWithCustomFormat($key);
    }
}
```

derekmd:
cast-date-string

"date-time string" implicitly means a date format supported by the database driver. Last sentence "The date's value will be correctly converted and stored in your database." is a bit ambiguous as I think it means assigning `Carbon` instances to an attribute will convert that object to a `Y-m-d h:i:s` string (via `Carbon::__toString()`) when the database query is run. That line may need to be rewritten for clarity.

AFAIK when a custom `$casts` date format is configured, Laravel models have _never_ parsed and auto-corrected strings to a common date format to prepare it for the database driver. And such an approach would have its own issues. e.g., for `d.m.Y` how would Eloquent know to convert that to `Y-m-d` and not include complete timestamp ` h:i:s` as `Carbon` resolves to by default?

```php
class TestModel extends \Illuminate\Database\Eloquent\Model
{
    protected $guarded = [];
    protected $casts = [
        'custom_date' => 'date:d.m.Y'
    ];
}

$date = new Carbon;
$model = new TestModel(['custom_date' => $date]);

$model->getAttributes()['custom_date'] === $date;
// true, original Carbon instance stored inside TestModel

$model->save();
```

```sql
insert into `test_models`
(`custom_date`, `updated_at`, `created_at`)
values
('2022-02-16 18:14:38', '2022-02-16 18:15:19', '2022-02-16 18:15:19')
```

---

This custom cast class will probably fix your database storage issue:

```php
namespace App\Casts;
 
use Carbon\CarbonInterface;
 
class DateCustomFormat implements CastsAttributes
{
    {
        return $value === null ? null : Carbon::parse($value);
    }
 
    {
        if ($value === null) {
            return;
        }

        if ($value instanceof CarbonInterface) {
            return $value->format('Y-m-d');
        }

        return Carbon::createFromFormat('d.m.Y', $value)->format('Y-m-d');
    }
}
```

```php
class TestModel extends \Illuminate\Database\Eloquent\Model
{
    protected $guarded = [];
    protected $casts = [
        'custom_date' => \App\Casts\DateCustomFormat::class,
    ];
}
```

```php
// <input name="custom_date"> can be in d.m.Y format
TestModel::create(request(['custom_date']));
```

---

This is really the only use of the `$casts` date format feature:

json

taylorotwell:
I'm with @derekmd that I think this is mainly a documentation issue. This particular line needs to be more specific that these are the only accepted ways to set a date: 

> When a column is cast as a date, you may set the corresponding model attribute value to a UNIX timestamp, date string (Y-m-d), date-time string, or a DateTime / Carbon instance. The date's value will be correctly converted and stored in your database.

The ability to define a format on the `datetime` cast is really only about changing the format of a model's dates when the model is serialized to an array or JSON. From the docs:

> When defining a date or datetime cast, you may also specify the date's format. This format will be used when the [model is serialized to an array or JSON](https://laravel.com/docs/9.x/eloquent-serialization):

It does not affect how a date may be **set** when interacting with a model.

taylorotwell:
For reference - the actual logic for how an incoming "date" value is turned into a DateTime instance before being turned into a date string for storage can be found in `HasAttributes@asDateTime`...

```php
// If this value is already a Carbon instance, we shall just return it as is.
// This prevents us having to re-instantiate a Carbon instance when we know
// it already is one, which wouldn't be fulfilled by the DateTime check.
if ($value instanceof CarbonInterface) {
    return Date::instance($value);
}

// If the value is already a DateTime instance, we will just skip the rest of
// these checks since they will be a waste of time, and hinder performance
// when checking the field. We will just return the DateTime right away.
if ($value instanceof DateTimeInterface) {
    return Date::parse(
        $value->format('Y-m-d H:i:s.u'), $value->getTimezone()
    );
}

// If this value is an integer, we will assume it is a UNIX timestamp's value
// and format a Carbon object from this timestamp. This allows flexibility
// when defining your date fields as they might be UNIX timestamps here.
if (is_numeric($value)) {
    return Date::createFromTimestamp($value);
}

// If the value is in simply year, month, day format, we will instantiate the
// Carbon instances from that format. Again, this provides for simple date
// fields on the database, while still supporting Carbonized conversion.
if ($this->isStandardDateFormat($value)) {
    return Date::instance(Carbon::createFromFormat('Y-m-d', $value)->startOfDay());
}

$format = $this->getDateFormat();

// Finally, we will just assume this date is in the format used by default on
// the database connection and use that format to create the Carbon object
// that is returned back out to the developers after we convert it here.
try {
    $date = Date::createFromFormat($format, $value);
} catch (InvalidArgumentException $e) {
    $date = false;
}

return $date ?: Date::parse($value);
```

PR #41043, opened 2022-02-15
[9.x] Add test for the cacheManager purge and etc

imanghafoori1:
These methods on the `CacheManager` still have no tests at all:
- `setDefaultDriver`
- `refreshEventDispatcher`
- `purge`

Memoization behavior also has no tests.

It also removes a redundant extra argument passed to the `forever` method, seemingly copy/pasted from the previous test.


taylorotwell:
Tests failing.

PR #41045, opened 2022-02-15
[9.x] Add instantiate method to collections

yoeriboven:
I often add classes to an array which I then later want to instantiate using a collection.

For example, I want to scrape these news websites.

```php
$sites = [
    NYTimes::class,
    BuzzFeed::class,
    TheTelegraph::class,
];

// Before
collect($sites)->map(function($class) {
    return new $class;
});

// Now
collect($sites)->instantiate();
```

The new collections now have an object of each class.

**For reviewers:**
- Please check if the DocBlocks are correct .
- Maybe you know an easier name for the method.


taylorotwell:

PR #41046, opened 2022-02-15
[9.x] Add test for cache manager

imanghafoori1:
These methods on the `CacheManager` still have no tests at all:
- `setDefaultDriver`
- `refreshEventDispatcher`
- `purge`

Memoization behavior also has no tests.

It also removes a redundant extra argument passed to the `forever` method, seemingly copy/pasted from the previous test.

- The `Container::getInstance` method was giving a shared singleton and was messing up with other tests, hence replaced with `new Container`


imanghafoori1:
I am investigating the problem of failing tests.

taylorotwell:
I think we are OK on tests for now.

PR #41047, opened 2022-02-15
[9.x] Enhance Container flush method

imanghafoori1:
I did not find a precise definition for the `flush` method of the container.
But it seems that when we remove all the binding, we should also remove the `contextual binding` and the `tags` for its bindings as well because they are very close to each other.

These three properties are flushed nowhere. neither in the `Container` flush nor in the `Application` flush method.

- Let me mention that this is important for `swoole` or `road-runner` style, since we may want to simulate a restart from scratch.

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -1365,6 +1365,9 @@ class Container implements ArrayAccess, ContainerContract
         $this->aliases = [];
         $this->resolved = [];
         $this->bindings = [];
+        $this->contextual = [];
+        $this->tags = [];
+        $this->with = [];
         $this->instances = [];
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Either this is a bug fix or it's a breaking new feature. If this a bug fix, it needs to go to 6.x. If it's a breaking new feature, it needs to go to 10.x.

driesvints:
@GrahamCampbell 6.x doesn't receives big fixes anymore

imanghafoori1:
I think this is an old edge case bug,
which is so rare that it had remained unreported up until now.

taylorotwell:
I think unless we are fixing a very specific bug or problem I'm not going to change this for now.

PR #41050, opened 2022-02-16
[9.x] Fix loadAggregate not correctly applying casts

ahme-d:
Reopening #41025 with tests and better explanation about the bug. 

There's a bug in `loadAggregate` in Eloquent Collections specefically when running `LoadExists` where running something like

`$user->articles->loadExists("comments")`  will add `comments_exists` to all existing articles but only the first model will have bool `casts` automatically added to it. Resulting in something similar to:

    .. "comments_exists" => true
    .. "comments_exists" => '1'
    .. "comments_exists" => '1'  

This PR fixes that and adds casts to all items instead of only the first one. I highly doubt it'll be breaking change to anyone.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Collection.php
+++ b/src/Illuminate/Database/Eloquent/Collection.php
@@ -92,7 +92,7 @@ class Collection extends BaseCollection implements QueueableCollection
         $this->each(function ($model) use ($models, $attributes) {
             $extraAttributes = Arr::only($models->get($model->getKey())->getAttributes(), $attributes);
 
-            $model->forceFill($extraAttributes)->syncOriginalAttributes($attributes);
+            $model->forceFill($extraAttributes)->syncOriginalAttributes($attributes)->mergeCasts($models->get($model->getKey())->getCasts());
         });
 
         return $this;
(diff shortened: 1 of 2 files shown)


aalyusuf:
Hey @taylorotwell 

Here's a simple code example ( Also the PR contains a test which would fail on a fresh laravel installation ):

**Models**

```PHP
{

    public function posts()
    {
        return $this->hasMany(Post::class);
    }

}
```

```PHP
class Post extends Model
{
    use HasFactory;
    protected $guarded = [];

    public function comments()
    {
        return $this->hasMany(Comment::class);
    }
}
```

```PHP
class Comment extends Model
{
    use HasFactory;
}
```


**Some data**

```PHP
    $user = \App\Models\User::find(1);
    $user->posts()->createMany([
        ['content' => 'Test'],
        ['content' => 'Test'],
        ['content' => 'Test'],
    ]);
```

**The test**
Now, running `loadExists` on comments

```PHP
    $user = \App\Models\User::with('posts')->find(1);
    $user->posts->loadExists("comments");
    return $user->posts;
```

This will result in the following ( notice ``comments_exists`` )  , loadExists here will only cast the first Post we created 
image

and 
```PHP
 dd($user->posts->pluck('comments_exists')); 
```
image

taylorotwell:
I created a user with 3 posts and each posts has 3 comments... this is what I get:

CleanShot 2022-02-16 at 14 12 32@2x

aalyusuf:
I believe that's the response because of tinker & it's not casting there. But after executing loadExists the next line you can try can be:
```PHP
$user->posts->pluck('comments_exists');
```

and this should reproduce the issue

image


or even directly 
```PHP
$user->posts->map->getCasts();
```

should give you:

image

taylorotwell:
Thanks

PR #41054, opened 2022-02-16
[9.x] Add "assertRelationLoaded"

SjorsO:
I have some big projects that were created before `Model::preventLazyLoading()` existed. I'm not mentally prepared yet to fix every instance of lazy loading in the whole application, so instead, for new code that I write, I do something like this:

```php
public function perfectCommentsCount(): Attribute
{
    throw_unless($this->relationLoaded('comments'), '"perfectCommentsCount" used, but "comments" relation is not loaded.');

    return new Attribute(
        get: fn () => $this->comments->where('is_perfect', true)->count(),
    );
}
```

This PR adds a `assertRelationLoaded` method so that I can do this instead:

```php
public function perfectCommentsCount(): Attribute
{
    $this->assertRelationLoaded('comments');

    return new Attribute(
        get: fn () => $this->comments->where('is_perfect', true)->count(),
    );
}
```

It seemed fitting to throw the existing `LazyLoadingViolationException` exception, but the exception message isn't exactly correct:

https://github.com/laravel/framework/blob/4b478a60d629861244aa8242abc2d376d9ece6e7/src/Illuminate/Database/LazyLoadingViolationException.php#L34

I think changing the exception message might be a breaking change? We could leave it be for now and change the message in Laravel 10, or we could throw a `LogicException` instead.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
@@ -18,6 +18,7 @@ use Illuminate\Database\Eloquent\Relations\MorphOne;
 use Illuminate\Database\Eloquent\Relations\MorphTo;
 use Illuminate\Database\Eloquent\Relations\MorphToMany;
 use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Database\LazyLoadingViolationException;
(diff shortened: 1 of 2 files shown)


taylorotwell:
I suggest just adding this to your own projects.

PR #41063, opened 2022-02-16
[9.x] Add Str::highlight()

BramVanBerkel:
After seeing #41000, I got inspired to also add the highlight feature, the original idea also coming from [Rails](https://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-highlight)

This pull requests adds the ability to easily highlight words in a string:
```php
public static function highlight($text, $highlight, $tag = 'mark')
```

Usage:
```php
Str::highlight('Laravel, The PHP Framework for Web Artisans', 'Laravel');
// <mark>Laravel</mark>, The PHP Framework for Web Artisans

Str::highlight('Laravel, The PHP Framework for Web Artisans', ['Laravel', 'Framework', 'Artisans']);
// <mark>Laravel</mark>, The PHP <mark>Framework</mark> for Web <mark>Artisans</mark>
```
Or with custom tags:
```php
Str::highlight('Laravel, The PHP Framework for Web Artisans', ['Laravel', 'Framework', 'Artisans'], 'b');
// <b>Laravel</b>, The PHP <b>Framework</b> for Web <b>Artisans</b>
```
If no `$highlight` or an empty `$tag` is given, the string is returned as-is, because there is nothing to highlight
```php
Str::highlight('Laravel, The PHP Framework for Web Artisans', '');
Str::highlight('Laravel, The PHP Framework for Web Artisans', []);
Str::highlight('Laravel, The PHP Framework for Web Artisans', 'Laravel', '');
// Laravel, The PHP Framework for Web Artisans
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -303,7 +303,7 @@ class Str
 
         $highlights = array_map(fn ($value) => preg_quote($value, '/'), (array) $highlights);
 
-        return preg_replace('/(?<![\w+])('.implode('|', $highlights).')(?![\w?+:.*^$()\[{\\}\]|])/', "<$tag>$0</$tag>", $text);
+        return preg_replace('/(?<![\w?+:.*^$()\[\{\}\]|\\])('.implode('|', $highlights).')(?![\w?+:.*^$()\[{\\}\]|])/', "<$tag>$0</$tag>", $text);
     }
 
     /**
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -301,7 +301,7 @@ class Str
             return $text;
         }
 
-        $highlights = array_map(fn($value) => preg_quote($value, '/'), (array)$highlights);
+        $highlights = array_map(fn ($value) => preg_quote($value, '/'), (array) $highlights);
 
         return preg_replace('/(?<![\w+])('.implode('|', $highlights).')(?![\w?+:.*^$()\[{\\}\]|])/', "<$tag>$0</$tag>", $text);
     }
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -303,7 +303,7 @@ class Str
 
         $highlights = array_map(fn ($value) => preg_quote($value, '/'), (array) $highlights);
 
-        return preg_replace('/(?<![\w?])('.implode('|', $highlights).')(?![\w?+:.*^$()\[{\\}\]|])/', "<$tag>$0</$tag>", $text);
+        return preg_replace('/(?<![\w?])('.implode('|', $highlights).')(?![\w?+:.\/*^$()\[{\\}\]|])/', "<$tag>$0</$tag>", $text);
     }
 
     /**
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -291,17 +291,19 @@ class Str
 
     /**
      * @param  string  $text
-     * @param  string|array  $highlight
+     * @param  string|array  $highlights
      * @param  string  $tag
      * @return string
      */
-    public static function highlight($text, $highlight, $tag = 'mark')
+    public static function highlight($text, $highlights, $tag = 'mark')
     {
-        if (empty($highlight) || $tag === '') {
+        if (empty($highlights) || $tag === '') {
             return $text;
         }
 
-        return preg_replace('/(?:'.implode('|', (array) $highlight).')/', "<$tag>$0</$tag>", $text);
+        $highlights = array_map(fn($value) => preg_quote($value, '/'), (array)$highlights);
+
+        return preg_replace('/(?<![\w+])('.implode('|', $highlights).')(?![\w?+:.*^$()\[{\\}\]|])/', "<$tag>$0</$tag>", $text);
     }
 
     /**
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -290,6 +290,21 @@ class Str
     }
 
     /**
+     * @param  string $text
+     * @param  string|array $highlight
+     * @param  string $tag
+     * @return string
+     */
+    public static function highlight($text, $highlight, $tag = 'mark')
+    {
+        if(empty($highlight) || $tag === '') {
+            return $text;
+        }
+
+        return preg_replace('/(?:'.implode('|', (array)$highlight).')/', "<$tag>$0</$tag>", $text);
+    }
+
+    /**
      * Cap a string with a single instance of a given value.
      *
      * @param  string  $value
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hey there - is this the exact same implementation as Rails? What if the string to highlight contains some characters that have meaning in regex?

JayBizzle:
Yeah, as Taylor says, this wouldn't work...
```php
Str::highlight('Laravel PHP framework?', '?');
// returns null
```

You probably need to use `preg_quote()`

```php
return preg_replace('/(?:'.preg_quote(implode('|', (array) $highlight)).')/', "<$tag>$0</$tag>", $text);
```

BramVanBerkel:
Good catch, thanks!

I've fixed it by `preg_quote()`'ing every highlight variable, to prevent it from escaping the `|` when `implode`'ing.

I've also changed the regex to include characters that have meaning in regex, the default word boundary in regex doesn't take them into account. This way it'll be more intuitive to use.
Old:
```php
//old:
Str::highlight('Laravel?', 'Laravel); // '<mark>Laravel</mark>?'
//new:
Str::highlight('Laravel?', 'Laravel); // '<mark>Laravel?</mark>'
```

JayBizzle:
> Old:
> 
> ```php
> 
> //old:
> 
> Str::highlight('Laravel?', 'Laravel); // '<mark>Laravel</mark>?'
> 
> //new:
> 
> Str::highlight('Laravel?', 'Laravel); // '<mark>Laravel?</mark>'
> 
> ```

Not sure I would expect that new output 🤔

JayBizzle:
Ah, that's better 👍🏻

BramVanBerkel:
I have also opened a PR for the docs, for if we want to merge this: https://github.com/laravel/docs/pull/7733

tests/Support/SupportStrTest.php
```diff
@@ -202,6 +202,38 @@ public function testStrExcerpt()
...
+
+        $this->assertSame('Laravel PHP framework', Str::highlight('Laravel PHP framework', ''));
+        $this->assertSame('Laravel PHP framework', Str::highlight('Laravel PHP framework', []));
+        $this->assertSame('', Str::highlight('', 'Laravel'));
+
+        $this->assertSame('Laravel PHP framework', Str::highlight('Laravel PHP framework', 'Laravél'));
+        $this->assertSame('<mark>Laravél</mark> PHP framework', Str::highlight('Laravél PHP framework', 'Laravél'));
+
+        $this->assertSame('Laravel PHP <mark>framework?</mark>', Str::highlight('Laravel PHP framework?', 'framework?'));
+        $this->assertSame('Laravel PHP framework?', Str::highlight('Laravel PHP framework?', 'framework'));
```
bert-w, line 226: This second line should resolve to `'Laravel PHP <mark>framework</mark>?'` right?

tests/Support/SupportStrTest.php
```diff
@@ -202,6 +202,38 @@ public function testStrExcerpt()
...
+        $this->assertSame('<mark>Laravel:</mark> <mark>PHP</mark> <mark>framework!</mark>', Str::highlight('Laravel: PHP framework!', ['Laravel:', 'PHP', 'framework!']));
+        $this->assertSame('Laravel: a PHP framework', Str::highlight('Laravel: a PHP framework', 'Laravel'));
```
bert-w, line 228: `'<mark>Laravel</mark>: a PHP framework'`

tests/Support/SupportStrTest.php
```diff
@@ -202,6 +202,38 @@ public function testStrExcerpt()
...
+        $this->assertSame('<mark>Laravel:</mark> a PHP framework', Str::highlight('Laravel: a PHP framework', 'Laravel:'));
+        $this->assertSame('?Laravel: a PHP framework', Str::highlight('?Laravel: a PHP framework', 'Laravel'));
```
bert-w, line 230: `'?<mark>Laravel</mark>: a PHP framework'`

tests/Support/SupportStrTest.php
```diff
@@ -202,6 +202,38 @@ public function testStrExcerpt()
...
+        $this->assertSame('<mark>?Laravel:</mark> a PHP framework', Str::highlight('?Laravel: a PHP framework', '?Laravel:'));
+
+        $this->assertSame('Laravel PHP/framework', Str::highlight('Laravel PHP/framework', 'PHP'));
```
bert-w, line 233: `'Laravel <mark>PHP</mark>/framework'`

bert-w (commented):
These are some unexpected outcomes imho

(P.S. the review refers to the last line in each marked code segment)

src/Illuminate/Support/Str.php
```diff
@@ -289,6 +289,23 @@ public static function excerpt($text, $phrase = '', $options = [])
...
+     */
+    public static function highlight($text, $highlights, $tag = 'mark')
+    {
+        if (empty($highlights) || $tag === '') {
+            return $text;
+        }
+
+        $highlights = array_map(fn ($value) => preg_quote($value, '/'), (array) $highlights);
+
+        return preg_replace('/(?<![\w?])('.implode('|', $highlights).')(?![\w?+:.\/*^$()\[{\\}\]|])/', "<$tag>$0</$tag>", $text);
```
JayBizzle, line 306: Why are all these characters in the negative look-ahead?

```\w?+:.\/*^$()\[{\\}\]|```

Comment (BramVanBerkel) [reply src/Illuminate/Support/Str.php:306]: They ensure that for example `Laravel?` doesn't match with `Laravel`, but only with `Laravel?`. Imho this way it's more intuitive use.

Comment (JayBizzle) [reply src/Illuminate/Support/Str.php:306]: I agree with your example, but it has the undesired effect (in my opinion) of the following output

```php
Str::highlight('Laravel?', 'Laravel');
// Actual output: Laravel?
// Expected output: <mark>Laravel</mark>?
```

JayBizzle: > They ensure that for example `Laravel?` doesn't match with `Laravel`, but only with `Laravel?`. Imho this way it's more intuitive use.

Also, i'm pretty sure it wouldn't actually do this, as the `?` is passed through `preg_quote`

JayBizzle: I think your original implementation was better (and simpler), but just needed to add the `preg_quote`

```php
$highlights = array_map(fn ($value) => preg_quote($value, '/'), (array) $highlights);

return preg_replace('/(?:'.implode('|', $highlights).')/', "<$tag>$0</$tag>", $text);
```

taylorotwell:

BrandonSurowiec:
I made this as a macro. It's case insensitive `/i` to match the browser's native "Find" functionality for my use case.

```php
        /**
         * Highlight the given words in a string.
         * 
         * @param  string  $text
         * @param  string|array  $highlights
         * @param  string  $tag
         * @return string
         */
        Str::macro('highlight', function($text, $highlights, $tag = 'mark') {
            if (empty($highlights) || $tag === '') {
                return $text;
            }

            $highlights = array_map(fn ($value) => preg_quote($value, '/'), (array) $highlights);

            return preg_replace('/('.implode('|', $highlights).')/i', "<$tag>$0</$tag>", $text);
        });
```

PR #41067, opened 2022-02-17
[9.x] Fix static return type of Collection sort function callback parameter

Krisell:
Unless my morning brain is playing tricks with me, the return type of the `Illuminate\Support\Collection` `sort` method is incorrectly set to `bool` but should be `int`. Larastan gave me a warning about my return type being incorrect.

The callback is passed to the php builtin `uasort` which has the signature `callback(mixed $a, mixed $b): int`, and it's the parity (positive, negative or 0) of the return value that determines the sort order.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -1283,7 +1283,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     /**
      * Sort through each item with a callback.
      *
-     * @param  (callable(TValue, TValue): bool)|null|int  $callback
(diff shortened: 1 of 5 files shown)


Krisell:
I'm trying to fix the type test but I'm not sure how to run the static analysis locally

canvural:
This also needs to be changed in `Enumerable` interface and in `LazyCollection`

Krisell:
> This also needs to be changed in `Enumerable` interface and in `LazyCollection`

Thanks, I have updated that now.

Note that the type test for `LazyCollection` did _not_ fail when running `phpstan` despite only changing the implementation and not the test. Is that expected? I did update the test as well in my commit, but the test passed before that.

PR #41068, opened 2022-02-17
Improve type hints in ModelNotFoundException

spawnia:
I am using this class in my unit tests to build up an exception I expect
to be thrown in the code under test. My model is using string keys, so
both my IDE and static analysis tooling complain about passing its ID.

Additional work on this could be done by introducing generic types for both the model and its ID. However, this would most likely require moving the hydration of the exception into `__construct()` and thus breaking the interface. I kept it light and simple for now and only changed the type hints.


spawnia:
I would love to grant edit access to the PR, but apparently it is not possible yet for organization repositories - see https://github.com/isaacs/github/issues/1681

driesvints:
Please re-create this PR from your personal account. Thanks

PR #41069, opened 2022-02-17
[9.x] Improve type hints in ModelNotFoundException

spawnia:
I am using this class in my unit tests to build up an exception I expect
to be thrown in the code under test. My model is using string keys, so
both my IDE and static analysis tooling complain about passing its ID.

Additional work on this could be done by:
- introducing generic types for both the model and its ID
- using strong type hints
I kept it light and simple for now and only changed the type hints.

### DIFF
--- a/src/Illuminate/Database/Eloquent/ModelNotFoundException.php
+++ b/src/Illuminate/Database/Eloquent/ModelNotFoundException.php
@@ -10,22 +10,22 @@ class ModelNotFoundException extends RecordsNotFoundException
     /**
      * Name of the affected Eloquent model.
      *
-     * @var string
+     * @var class-string<\Illuminate\Database\Eloquent\Model>
      */
     protected $model;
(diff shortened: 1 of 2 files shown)


nunomaduro:
@spawnia Can you add tests for this? On the folder `types`.

spawnia:
I thought about it, but did not since there are no generics or other advanced type shenanigans involved. If you still think it is necessary, I can add some.

nunomaduro:
Yes, thanks.

nunomaduro:
Replaced by https://github.com/laravel/framework/pull/41082. I've added you as co-author.

zachfraizer:
On Thursday, February 17, 2022, 05:44:44 AM CST, Nuno Maduro ***@***.***> wrote:
 
 
 


@spawnia Can you add tests for this? On the folder types.

Issue #40757, opened 2022-02-02
Getting wrong SQL generated with "!= 0" in statement since v8.81.0 


martin-pfeiffer-dibooq:
- Laravel Version: > 8.81
- PHP Version: used > 7.4.16 
- Database Driver & Version: pgsql & postgresql v12-14 

### Description:

Since this commit https://github.com/laravel/framework/commit/a9c50095980e6f24ca23c811fa55752e0f50dd6c 
we get unnecessary "!= 0" in SQL statements. 

### Steps To Reproduce:

In Tinker: 
`App\Models\Occupation::query()->where('real_estate_id',59)->where('date_range', '>>','[2022-01-08 00:00:00,2022-01-09 00:00:00)')->toSql();`

SQL output with >v8.81: 

=> "select * from "occupations" where "real_estate_id" = ? and ("date_range" >> ?) **!= 0** and "occupations"."deleted_at" is null"

SQL output with <v8.81: 

=> "select * from "occupations" where "real_estate_id" = ? and "date_range" >> ? and "occupations"."deleted_at" is null"

### Error:

Illuminate\Database\QueryException with message 'SQLSTATE[42883]: Undefined function: 7 ERROR:  operator does not exist: boolean <> integer
LINE 1: ...re "real_estate_id" = $1 and ("date_range" >> $2) **!= 0** and "...

HINT:  No operator matches the given name and argument types. You might need to add explicit type casts. (SQL: select * from "occupations" where "real_estate_id" = 226 and ("date_range" >> [2022-01-08 00:00:00,2022-01-09 00:00:00)) != 0 and "occupations"."deleted_at" is null limit 1)'

Why changes are made on https://github.com/laravel/framework/commit/a9c50095980e6f24ca23c811fa55752e0f50dd6c#r65853324 e.G. ?

BrandonSurowiec:
@martin-pfeiffer-dibooq  I know nothing about the differences between PgSQL and PostgresSQL when compared to MySql. Is it normal to open the `date_rangse` argument with a square bracket "[" and then close it with parenthesis ")" `'[2022-01-08 00:00:00,2022-01-09 00:00:00)'` ? Should it close with a square bracket? `'[2022-01-08 00:00:00,2022-01-09 00:00:00]'`

martin-pfeiffer-dibooq:
@BrandonSurowiec its not about the date range. Its just about the "!= 0" what is added afterwords. 

This causes an SQL error like shown.

[...]

PR #41072, opened 2022-02-17
[8.x] Add better bitwise operators support

Arzaroth:
### Impetus

This addresses the problem mentioned in issue #40484, while avoiding pitfalls brought up by issue #40757.
This PR was originally submitted as #40529, by reverted by #40791.

### Changes

The problem with previous implementation was assuming that bitwise operators always produce an integer (or at least a number).

I failed to see that these operators (notably `>>` and `<<`) where not only bitwise arithmetic operators for PostgreSQL but also range operators (as described [here](https://www.postgresql.org/docs/13/functions-range.html)).
In the issue #40757, we fall exactly in this case.

Thus, the approach I took this time was to force a cast to bool in PostgreSQL, which solves both the initial problem and ones caused by the previous implementation.

Since both MySQL and SQLite support "implicit casting" to boolean, it is essentially a noop for those grammars.

SQL Server still requires the "old trick" of `!= 0`. [The documentation](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/bitwise-operators-transact-sql?view=sql-server-ver15) states that it "converts the result to an integer.", so the comparison is safe.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -204,6 +204,15 @@ class Builder
     ];
 
     /**
+     * All of the available bitwise operators.
+     *
+     * @var string[]
+     */
+    public $bitwiseOperators = [
+        '&', '|', '^', '<<', '>>', '&~',
+    ];
+
+    /**
      * Whether to use write pdo for the select.
      *
      * @var bool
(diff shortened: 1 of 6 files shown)


taylorotwell:
@Arzaroth I can't edit your fork. Please grant me access or I can't merge this.

taylorotwell:
Put it here so I can edit it: https://github.com/laravel/framework/pull/41112

PR #41073, opened 2022-02-17
[9.x] Cast $perPage to integer on Paginator

gdebrauwer:
The `$perPage` parameter of `LengthAwarePaginator`/`CursorPaginator` must be an integer according to docblocks, but when the parameter is an integer in a string, it is never cast to an integer. Such casting is already done on `$lastPage` and `$currentPage`, but not yet on `$perPage`. This PR fixes that.

### DIFF
--- a/src/Illuminate/Pagination/CursorPaginator.php
+++ b/src/Illuminate/Pagination/CursorPaginator.php
@@ -37,7 +37,7 @@ class CursorPaginator extends AbstractCursorPaginator implements Arrayable, Arra
             $this->{$key} = $value;
         }
 
-        $this->perPage = $perPage;
+        $this->perPage = (int) $perPage;
         $this->cursor = $cursor;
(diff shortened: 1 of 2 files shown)


driesvints:
I don't think we should cast this. If you send in something else than an integer then there's a bug in your app somewhere and I think you should be aware of that.

gdebrauwer:
When you use the paginate-method with a "per page" value from a querystring parameter, then that value is an integer in a string when you provide it to the `paginate`. This is similar to the default `page` parameter in the querystring that is used by the paginator. The `currentPage` value is, because of that, also an integer as a string by default, but is cast in the constructor of the `LenghtAwarePaginator`. So I think it is logical to do the same with the `perPage` value.

gdebrauwer:
When you provided an actual string as `perPage`, not an integer in a string, you will still get an exception because the query will fail when using a string as a limit amount

driesvints (approved):

PR #41082, opened 2022-02-17
[9.x] Makes `ModelNotFoundException` generic

nunomaduro:
The contributor of this pull request is @spawnia, I've just made a few tweaks.

This pull request makes the `ModelNotFoundException` generic. So, in theory, once Eloquent is generic too, we will be able to do stuff like this:

```php
try {
    User::findOrFail(1);
} catch (ModelNotFound $e) {
    $e->getModel()::myUserMethod(); // --> autocompletion, and static analysis support here.
}
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/ModelNotFoundException.php
+++ b/src/Illuminate/Database/Eloquent/ModelNotFoundException.php
@@ -5,27 +5,30 @@ namespace Illuminate\Database\Eloquent;
 use Illuminate\Database\RecordsNotFoundException;
 use Illuminate\Support\Arr;
 
+/**
+ * @template TModel of \Illuminate\Database\Eloquent\Model
+ */
 class ModelNotFoundException extends RecordsNotFoundException
 {
     /**
      * Name of the affected Eloquent model.
      *
-     * @var string
+     * @var class-string<TModel>
      */
     protected $model;
 
     /**
      * The affected model IDs.
      *
-     * @var int|array
+     * @var array<int, int|string>
      */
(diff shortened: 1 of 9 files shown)


types/Database/Eloquent/ModelNotFoundException.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+use Illuminate\Database\Eloquent\ModelNotFoundException;
+use function PHPStan\Testing\assertType;
+
+/** @var ModelNotFoundException<User> $exception */
+$exception = new ModelNotFoundException();
```
spawnia, line 7: Constructing this class generically is always going to be awkward, since the generic type can not be derived from the constructor arguments. Perhaps it can be considered to move the arguments that are currently passed to `setModel()` into the constructor for the next major version of Laravel?

src/Illuminate/Database/Eloquent/ModelNotFoundException.php
```diff
@@ -5,27 +5,30 @@
 use Illuminate\Database\RecordsNotFoundException;
 use Illuminate\Support\Arr;
 
+/**
+ * @template TModel of \Illuminate\Database\Eloquent\Model
```
spawnia, line 9: The ID type could be made generic too.

spawnia:
Thanks @nunomaduro for getting this merged. Perhaps we can iterate on it some more, although the biggest possible improvement to this would allowing inference of the generics through constructor arguments - this would require breaking changes.

PR #41085, opened 2022-02-17
[9.x] Update facade.stub to revive IDE auto-completion

imanghafoori1:
### Revive IDE auto-completion (The simple way)

With this simple change, real-time facades will have IDE auto-completion with no backward compatibility issues or introducing any fancy logic.

When I click on the method call, my phpstorm takes me not to the facade file but directly to the underlying implementation and that is wonderful !!!

 The hello method in this case is NOT a static method but is suggested.

([my previous PR](https://github.com/laravel/framework/pull/40434) with the same goal had a lot of logic to generate docblocks but this time it is dead simple.)

image

## initial class:
image

## Generated cache file:

image

### DIFF
--- a/src/Illuminate/Foundation/stubs/facade.stub
+++ b/src/Illuminate/Foundation/stubs/facade.stub
@@ -5,7 +5,7 @@ namespace DummyNamespace;
 use Illuminate\Support\Facades\Facade;
 
 /**
- * @see \DummyTarget
+ * @mixin \DummyTarget
  */
 class DummyClass extends Facade
 {
@@ -16,6 +16,6 @@ class DummyClass extends Facade
      */
     protected static function getFacadeAccessor()
     {
-        return 'DummyTarget';
+        return \DummyTarget::class;
     }
 }


src/Illuminate/Foundation/stubs/facade.stub
```diff
@@ -5,7 +5,7 @@ namespace DummyNamespace;
 use Illuminate\Support\Facades\Facade;
 
 /**
- * @see \DummyTarget
+ * @mixin \DummyTarget
```
GrahamCampbell, line 8: This is incorrect. The facade methods are static, not instance.

GrahamCampbell (changes requested):

src/Illuminate/Foundation/stubs/facade.stub
imanghafoori1: I did not find a standard definition of mixin but it worked very well in the latest version of phpstorm in practice.

taylorotwell:

PR #41088, opened 2022-02-17
[9.x] Fix parsing `config('database.connections.pgsql.search_path')`

### LINKED ISSUE #41062: Laravel 9, Postgres can't change search_path to valid schema
- Laravel Version: 9.1.0
- PHP Version: 8.1.2
- Database Driver & Version: PostgreSQL 14.1 (Debian 14.1-1.pgdg110+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit

### Description:
I am trying to change the search_path to the current schema and unfortunately neither queries nor migrations work. Laravel fails to connect to the database on the defined schema.

### Steps To Reproduce:
- create new laravel 9 project
- change driver to pgsql
- create test-db schema
- change database.php, postgres setting to `'search_path' => 'test-db',`
- try to run migration or select query from database

`dd(DB::getPdo()->query('select current_schemas(true)')->fetchAll());`

```
^ array:1 [[▼]()
  0 => array:2 [[▼]()
    "current_schemas" => "{pg_catalog}"
    0 => "{pg_catalog}"
  ]
]
```

```bash
php artisan migrate

   Illuminate\Database\QueryException

  SQLSTATE[3F000]: Invalid schema name: 7 ERROR:  no schema has been selected to create in at character 14 (SQL: create table "migrations" ("id" serial primary key not null, "migration" varchar(255) not null, "batch" integer not null))

  at vendor/laravel/framework/src/Illuminate/Database/Connection.php:712
    708▕         // If an exception occurs when attempting to run a query, we'll format the error
    709▕         // message to include the bindings with SQL, which will make this exception a
    710▕         // lot more helpful to the developer instead of just the database's errors.
    711▕         catch (Exception $e) {
  ➜ 712▕             throw new QueryException(
    713▕                 $query, $this->prepareBindings($bindings), $e
    714▕             );
    715▕         }
    716▕     }

      +37 vendor frames
  38  artisan:37
      Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
```

derekmd:
Fixes https://github.com/laravel/framework/issues/41062

e.g., Config value `'test-db'` was being parsed as `set search_path to "test", "db"` instead of `set search_path to "test-db"`

9.x renamed `config('database.connections.pgsql.schema')` to `config('database.connections.pgsql.search_path')` and introduced special handling if the config value is a string containing many segments. The [originating PR](https://github.com/laravel/framework/pull/35463#issuecomment-767263747) has some examples. However the given `PostgresConnector` regex doesn't consider the full range of characters allowed in a schema or variable name - specifically `'-'` and accented characters.

Replace the `'search_path'` regex alphanumeric-only allowlist of characters with a blocklist of  delimiters when  `config('database.connections.pgsql.search_path')` is a string value. Technically Postgres _does_ allow our config delimiter characters (spaces, comma, quotes) in symbols so an array configuration can instead be used for such schema names. (However non-wrapping single/double quote characters in such array configs still aren't supported.)

```php
return [
    'default' => env('DB_CONNECTION', 'mysql'),
    'connections' => [
        'pgsql' => [
            'search_path' => [
                "'public'",
                '"$user"',
                'containing commas, and spaces'
                // 'can\'t name schema with "inner quotes"'
            ],
```

```sql
set search_path to "public", "$user", "containing commas, and spaces"
```

---

* Roll methods `testPostgresSearchPathCommaSeparatedValueSupported()` & `testPostgresSearchPathVariablesSupported()` into `testPostgresSearchPathIsSet()` with the fuller coverage `provideSearchPaths()` data set.
* `testPostgresSearchPathArraySupported()` is repurposed to show `config('database.connections.pgsql.schema')` from versions < 9.x is used when the `'search_path'` config key is absent.
* Fix `PostgresConnector::quoteSearchPath()` parameter docblock since passing in a string value would throw an exception for being un-`Countable`. Its only  use is being passed `parseSearchPath()`'s return value which is an array.

### DIFF
--- a/src/Illuminate/Database/Connectors/PostgresConnector.php
+++ b/src/Illuminate/Database/Connectors/PostgresConnector.php
@@ -127,7 +127,7 @@ class PostgresConnector extends Connector implements ConnectorInterface
     protected function parseSearchPath($searchPath)
     {
         if (is_string($searchPath)) {
-            preg_match_all('/[a-zA-z0-9$]{1,}/i', $searchPath, $matches);
+            preg_match_all('/[^\s,"\']+/', $searchPath, $matches);
 
             $searchPath = $matches[0];
         }
@@ -144,7 +144,7 @@ class PostgresConnector extends Connector implements ConnectorInterface
     /**
      * Format the search path for the DSN.
      *
-     * @param  array|string  $searchPath
+     * @param  array  $searchPath
      * @return string
      */
     protected function quoteSearchPath($searchPath)
--- a/tests/Database/DatabaseConnectorTest.php
+++ b/tests/Database/DatabaseConnectorTest.php
@@ -89,27 +89,103 @@ class DatabaseConnectorTest extends TestCase
         $this->assertSame($result, $connection);
     }
 
-    public function testPostgresSearchPathIsSet()
+    /**
+     * @dataProvider provideSearchPaths
+     *
+     * @param  string  $searchPath
+     * @param  string  $expectedSql
+     */
+    public function testPostgresSearchPathIsSet($searchPath, $expectedSql)
     {
         $dsn = 'pgsql:host=foo;dbname=\'bar\'';
-        $config = ['host' => 'foo', 'database' => 'bar', 'search_path' => 'public', 'charset' => 'utf8'];
+        $config = ['host' => 'foo', 'database' => 'bar', 'search_path' => $searchPath, 'charset' => 'utf8'];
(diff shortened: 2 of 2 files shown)


taylorotwell:
Thanks!

PR #41093, opened 2022-02-18
Add isWithParameters method and an alias named isWith

namvarii:
With using Route::isWith() we can extend Route::is() functionality to check named routes with parameter.
like below :
Route::isWith('without.parameter', ['parameters.removed'], ['with.parameter', 'john'], ['with.parameter', ['name' => 'john']]) ) {
   //
}

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -1198,7 +1198,7 @@ class Router implements BindingRegistrar, RegistrarContract
     {
         return $this->current() && $this->current()->named(...$patterns);
     }
-    
+
     /**
      * Alias for the "isWithParameters" method.
      *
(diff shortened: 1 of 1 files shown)


src/Illuminate/Routing/Router.php
```diff
@@ -1198,6 +1198,39 @@ public function currentRouteNamed(...$patterns)
...
+                if ($this->currentRouteNamed([$pattern])) {
+                    return true;
+                }
+            } else {
+                if (url()->current() == route($pattern[0], $pattern[1] ?? [])) {
+                    return true;
+                }
+            }
+        }
+        return false;
```
gguney, line 1232:
```suggestion

        return false;
```

PR #41099, opened 2022-02-18
[9.x] Eloquent: firstWhere returns Object instead of NULL

### LINKED ISSUE #41098: Eloquent: firstWhere returns Object instead of NULL
- Laravel Version: 8.83.x, 9.0.x
- PHP Version: 8.1.1
- Database Driver & Version: Postgresql 10

### Description:

When calling the firstWhere() method of a Model with the second Parameter being "-" and when there is no database entry with "-" it returns a seemingly random Model object.

Probably it assumes "-" is the operator and value is "null". Which is not correct.

### Steps To Reproduce:

Failing Test:

```
        User::factory()->create([
            "name" => 'Foo'
        ]);

        $user = User::firstWhere('name', '-');

        $this->assertNull($user);
```

This test is using the User model. But every other model will do.

When explicitly specifying an operator it works as expected:

```php
$user = User::query()->firstWhere('name', '=', '-');
```

marvinrabe:
Fixes #41098

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -296,7 +296,7 @@ class Builder implements BuilderContract
      */
     public function firstWhere($column, $operator = null, $value = null, $boolean = 'and')
     {
-        return $this->where($column, $operator, $value, $boolean)->first();
+        return $this->where(...func_get_args())->first();
     }
 
     /**


driesvints:
This definitely isn't wanted and changes the behavior of this method to accept more parameters than needed.

driesvints (approved):
Can confirm this fixes the issue for me as well. @marvinrabe can you maybe also add a test?

PR #41100, opened 2022-02-18
[9.x] Add `isWithParameters` method and an alias named `isWith`

Namvarii:
With using Route::isWith() we can extend Route::is() functionality to check named routes with parameters.

Usage :
```
Route::isWith(
    'without.parameter', 
    ['parameters.removed'], 
    ['with.parameter', 'john'],
    ['with.parameter', ['name' => 'john']]
) {
    //
}
```

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -1198,7 +1198,7 @@ class Router implements BindingRegistrar, RegistrarContract
     {
         return $this->current() && $this->current()->named(...$patterns);
     }
-    
+
     /**
      * Alias for the "isWithParameters" method.
      *
@@ -1229,6 +1229,7 @@ class Router implements BindingRegistrar, RegistrarContract
                 }
             }
         }
+
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


src/Illuminate/Routing/Router.php
```diff
@@ -1199,6 +1199,40 @@ public function currentRouteNamed(...$patterns)
...
+     */
+    public function isWithParameters(...$patterns)
+    {
+        foreach ($patterns as $pattern) {
+            if (is_string($pattern)) {
+                if ($this->currentRouteNamed([$pattern])) {
+                    return true;
+                }
+            } else {
+                if (url()->current() == route($pattern[0], $pattern[1] ?? [])) {
```
GrahamCampbell, line 1227: Can't make call to foundation helper function here.

GrahamCampbell (changes requested):

src/Illuminate/Routing/Router.php
namvarii: Thanks, We need full URI of site, what is your suggestion to get that without foundation helper?

namvarii: Now using `$this->getCurrentRequest()->url()` rather than `url()->current()`

taylorotwell:

PR #41101, opened 2022-02-18
[9.x] Add `X` headers when using `Mail::alwaysTo`

morrislaptop:
When using `Mail::alwaysTo` in development environments, the original To, Cc and Bcc are lost. This makes it difficult to determine where the email was going to during testing. 

This PR adds the original to, cc and bcc into X-Headers in the email so this information can be retrieved, while still preventing the email being sent to these recipients. 

The below screenshot of Helo shows `alwaysTo` being set, an attempt to send an email with To, Cc, Bcc and the resultant email with the added headers. 

image

### DIFF
--- a/tests/Mail/MailMailerTest.php
+++ b/tests/Mail/MailMailerTest.php
@@ -210,9 +210,11 @@ class MailMailerTest extends TestCase
         });
 
         $this->assertSame('taylor@laravel.com', $sentMessage->getEnvelope()->getRecipients()[0]->getAddress());
-        $this->assertStringContainsString('X-To: nuno@laravel.com', $sentMessage->toString());
-        $this->assertStringContainsString('X-Cc: dries@laravel.com', $sentMessage->toString());
-        $this->assertStringContainsString('X-Bcc: james@laravel.com', $sentMessage->toString());
+        $this->assertDoesNotMatchRegularExpression('/^To: nuno@laravel.com/m', $sentMessage->toString());
+        $this->assertDoesNotMatchRegularExpression('/^Cc: dries@laravel.com/m', $sentMessage->toString());
+        $this->assertMatchesRegularExpression('/^X-To: nuno@laravel.com/m', $sentMessage->toString());
+        $this->assertMatchesRegularExpression('/^X-Cc: dries@laravel.com/m', $sentMessage->toString());
+        $this->assertMatchesRegularExpression('/^X-Bcc: james@laravel.com/m', $sentMessage->toString());
         $this->assertFalse($recipients->contains('nuno@laravel.com'));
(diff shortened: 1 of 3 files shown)


tests/Mail/MailMailerTest.php
```diff
@@ -209,8 +210,11 @@ public function testGlobalToIsRespectedOnAllMessages()
         });
 
         $this->assertSame('taylor@laravel.com', $sentMessage->getEnvelope()->getRecipients()[0]->getAddress());
-        $this->assertStringNotContainsString('dries@laravel.com', $sentMessage->toString());
-        $this->assertStringNotContainsString('james@laravel.com', $sentMessage->toString());
+        $this->assertStringContainsString('X-To: nuno@laravel.com', $sentMessage->toString());
+        $this->assertStringContainsString('X-Cc: dries@laravel.com', $sentMessage->toString());
+        $this->assertStringContainsString('X-Bcc: james@laravel.com', $sentMessage->toString());
+        $this->assertFalse($recipients->contains('nuno@laravel.com'));
+        $this->assertFalse($recipients->contains('dries@laravel.com'));
```
driesvints, line 217: We should be a bit carefully here because we're not actually checking anymore if the CC: and BCC: headers are not present anymore (something the original tests implicitly did). I don't think BCC recipients, for example, are listed in the recipients of an envelope to begin with?

tests/Mail/MailMailerTest.php
```diff
@@ -209,8 +210,13 @@ public function testGlobalToIsRespectedOnAllMessages()
         });
 
         $this->assertSame('taylor@laravel.com', $sentMessage->getEnvelope()->getRecipients()[0]->getAddress());
-        $this->assertStringNotContainsString('dries@laravel.com', $sentMessage->toString());
-        $this->assertStringNotContainsString('james@laravel.com', $sentMessage->toString());
```
morrislaptop, line 213: This assertion was irrelevant as BCC addresses would not be in the string

tests/Mail/MailMailerTest.php
```diff
@@ -200,6 +200,7 @@ public function testGlobalToIsRespectedOnAllMessages()
 
         $sentMessage = $mailer->send('foo', ['data'], function (Message $message) {
             $message->from('hello@laravel.com');
+            $message->to('nuno@laravel.com');
```
morrislaptop, line 203: Added this for extra measure

tests/Mail/MailMailerTest.php
driesvints: That was the whole point to prove with this assertion.

driesvints: But current tests seem all good to me 👍

driesvints (approved):

PR #41103, opened 2022-02-18
Update ResetPassword.php

krusche-sensetence:
Setting 
`ResetPassword::createUrlUsing` 
was useless if also
 `ResetPassword::toMailUsing`
was set. 

The resetUrl wasn't calculated by the given ResetPassword::createUrlUsing callback and wasn't passed to the  ResetPassword::toMailUsing callback.

### DIFF
--- a/src/Illuminate/Auth/Notifications/ResetPassword.php
+++ b/src/Illuminate/Auth/Notifications/ResetPassword.php
@@ -59,11 +59,13 @@ class ResetPassword extends Notification
      */
     public function toMail($notifiable)
     {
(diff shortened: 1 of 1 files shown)


driesvints:
We're adding a test that makes the current behavior a requirement. With it, we cannot merge this PR. Please see https://github.com/laravel/framework/pull/41104

zeitlerquintet:
Okey, so you don't think this is a bug?

zeitlerquintet:
thanks, I'll do so 👍

PR #41109, opened 2022-02-18
[10.x] Update ResetPassword.php

krusche-sensetence:
Setting 
`ResetPassword::createUrlUsing` 
was useless if also
 `ResetPassword::toMailUsing`
was set. 

The resetUrl wasn't calculated by the given ResetPassword::createUrlUsing callback and wasn't passed to the  ResetPassword::toMailUsing callback.

### DIFF
--- a/src/Illuminate/Auth/Notifications/ResetPassword.php
+++ b/src/Illuminate/Auth/Notifications/ResetPassword.php
@@ -59,11 +59,13 @@ class ResetPassword extends Notification
      */
     public function toMail($notifiable)
     {
+        $resetUrl = $this->resetUrl($notifiable);
+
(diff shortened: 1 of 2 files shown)


driesvints:
@taylorotwell this will fail on 9.x after the test @crynobone sent in earlier today.

taylorotwell:
https://github.com/laravel/framework/pull/41104

Putting this here for reference.

zeitlerquintet:
@crynobone I adjusted the test

taylorotwell:
I don't think I'm going to make this change. I think if you are using toMailUsing it is assumed you will create your own URL within that callback.

PR #41114, opened 2022-02-19
[9.x] Add attributesToDatabaseArray() function to HasAttributes trait

YungLin-CHU:
When testing model value updated or not using assertDatabaseHas(), if model has boolean attributes would make test failed due to database store boolean value as "0", "1", while model casts the attributes to be boolean.

For example:

```
// Model
class User extends Model
{
    protected $casts = [
        'bool_col' => 'boolean',
    ];
}

// Factory
class UserFactory extends Factory
{
    public function definition()
    {
        return [
            'bool_col' => "1", // true
        ];
    }
}

// Test
class UserTest extends TestCase
{
    public function test_user_can_be_updated()
    {

        $response = $this->put('/users/' . $user->id, $user->attributesToArray());

        $response->assertSessionHasNoErrors();
        $this->assertDatabaseHas('users', $user->attributesToArray()); // Failed here
    }
}

// Error message
Failed asserting that a row in the table [users] matches the attributes {
    "bool_col": false,
    "id": 1
}.

Found similar results: [
    {
        "bool_col": "0",
        "id": 1
    }
]
```

Improvement of attributesToDatabaseArray() function: 
```
// Output
{
    "bool_col": "0",
    "id": 1
}

// Output is the same as the retrieved records.
```

P.S.
I made this change when I use sqlite as test database, this may no be worked when using other database such as MongoDB. Therefore, this may need further improvements.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -206,7 +206,7 @@ trait HasAttributes
     }
 
     /**
-     * Convert the model's attributes to an databse result array.
+     * Convert the model's attributes to a database result array.
      *
      * @return array
      */
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -204,6 +204,26 @@ trait HasAttributes
 
         return $attributes;
     }
+    
+    /**
+     * Convert the model's attributes to an databse result array.
+     *
+     * @return array
+     */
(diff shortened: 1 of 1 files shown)


src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
```diff
@@ -205,6 +205,26 @@ public function attributesToArray()
         return $attributes;
     }
 
+    /**
+     * Convert the model's attributes to an databse result array.
```
gguney, line 209:
```suggestion
     * Convert the model's attributes to a database result array.
```

taylorotwell:

PR #41118, opened 2022-02-19
[9.x] Rollback a specific batch of migrations

PHPGuus:
- Add a new option `batch` to the RollbackCommand.
- Add a new function `getMigrationsByBatch` to the MigrationRepositoryInterface.
- Implement `getMigrationsByBatch` in DatabaseMigrationRepository.

This new option allows the developer to rollback a specific batch of migrations. This can be helpful
in situations where a production bugfix introduced database changes after the current feature branch introduced database changes. The develop may not be able to rollback the production bugfix batch, but still wants to revert the batch from the current feature branch to make slight changes to the database migrations introduced in the current feature branch.

A PR to laravel/docs will also be submitted.

### DIFF
--- a/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
@@ -60,6 +60,7 @@ class RollbackCommand extends BaseCommand
                 $this->getMigrationPaths(), [
                     'pretend' => $this->option('pretend'),
                     'step' => (int) $this->option('step'),
+                    'batch' => (int) $this->option('batch'),
                 ]
             );
         });
@@ -86,6 +87,8 @@ class RollbackCommand extends BaseCommand
             ['pretend', null, InputOption::VALUE_NONE, 'Dump the SQL queries that would be run'],
 
             ['step', null, InputOption::VALUE_OPTIONAL, 'The number of migrations to be reverted'],
+
+            ['batch', null, InputOption::VALUE_REQUIRED, 'The batch of migrations (identified by their batch number) to be reverted'],
         ];
     }
 }
--- a/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
@@ -88,7 +88,7 @@ class RollbackCommand extends BaseCommand
 
             ['step', null, InputOption::VALUE_OPTIONAL, 'The number of migrations to be reverted'],
 
-            ['batch', null, InputOption::VALUE_REQUIRED, 'The batch of migrations (identified by their batch number) to be reverted'],
+            ['batch', null, InputOption::VALUE_OPTIONAL, 'The batch of migrations (identified by their batch number) to be reverted'],
         ];
     }
 }
--- a/src/Illuminate/Database/Migrations/DatabaseMigrationRepository.php
+++ b/src/Illuminate/Database/Migrations/DatabaseMigrationRepository.php
@@ -69,6 +69,21 @@ class DatabaseMigrationRepository implements MigrationRepositoryInterface
     }
 
     /**
+     * Get the list of the migrations by batch number.
+     *
+     * @param  int  $batchNumber
+     * @return array
+     */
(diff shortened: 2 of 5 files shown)


antonkomarev:
It is not a good practice to rollback on production. Production migrations should go forward only.

PHPGuus:
It is intended for rolling back a specific batch of migrations. The example is that a migration is introduced in your working copy as part of a production bug fix (main -> bugfix -> merge into main *and* develop). But it holds for any migration that comes in to your working copy from any branch.

The point is that once you have a migration younger than the latest migrations in your working copy, you may not want to roll back and redo those younger migrations, but you still may want to roll back a specific batch of migrations from before. Meaning, executing `php artisan migrate:rollback --step 2` would also roll back the youngest migration (the one you don't control), but executing `php artisan migrate:rollback --batch 5` would roll back the migrations from batch 5, and leave the migrations from batch 6 alone and in tact.

Akin to cherry picking which set of migrations you want to roll back.

dennisprudlo:
I agree with @antonkomarev, wouldn't it be better to pull the bugfix changes into the feature branch to make adjustments on that branch?

The rollback could also throw exceptions when it expects a database structure that doesn't exist anymore because of later migrations:

```php
// Migration A
public function down () : void {
        $table->string('nickname')->after('lastname');
    });
}
```

```php
// Migration B (already run)
public function up () : void {
        $table->dropColumn('lastname');
    });
}
```
When trying to rollback migration A the `lastname` field won't exist, because it was dropped by migration B. Of course, in the end it's the developers choice and one could argue that the developer has to make sure the rollback will work, but I'm not sure if this is something that should be supported by the core? I'm seeing a lot of "Error when using migrate:rollback" issues because people tried to randomly rollback a migration to simply remove a column they added a few months back, instead of creating a new migration that drops the column.

PHPGuus:
Yes absolutely correct. Those arguments would ultimately lead to dropping the entire roll back functionality. I'm not sure if that's what's wanted either.

The way I see it is that a co-developer created a very valid but rather complex migration, and that migration ended up being merged into your current feature branch. You apply the migration *in your working copy* (as you do, because hey why not that particular feature is now in production), and about a day or later you figure that you need to change some migration that is part of the feature set that you are working on.

So your migrations are: `2022_02_17_064050_something_or_other.php` and `2022_02_17_070812_other_than_that.php` and your co-developer's migration is `2022_02_18_120000_lets_not_have_a_down_function_here.php`.

This leads to a situation where you cannot roll back the last migration, because there's no `down` function. But you still want to roll back your misconstrued migrations and actually create proper migrations (since you're still working on a feature branch, you might as well, right? Right!) You're stuck.

Intro cherry picking which migration set to roll back, and you're good to go!

PHPGuus:
or maybe the down function wasn't tested properly, and you want to be able to by pass it...

or maybe ... whatever the reason you have for not wanting to roll back the younger migrations.

And since I refer to working copy / feature branch fairly often, it would be kind of clear that this is a possible command to execute intended for your working copy, rather then for production.

But then again, if we don't want functions / features that can break production, then we'd be looking at stripping all sorts of features out of Laravel :-)

antonkomarev:
> Those arguments would ultimately lead to dropping the entire roll back functionality. 

It is right. It is common no to write rollback migrations at all. Because there are many cases when a rollback is no longer possible. And no one ever tests rollback migrations, you can't be sure that it really works as expected. If you want to revert changes, you are writing new migration which will revert all the changes back.

I recently created a small library to check that migrations cannot be reverted:
https://github.com/antonkomarev/php-db-migration-validator

PHPGuus:
@antonkomarev Let's not get into a biased or even opinionated discussion about "no one" and "common" and "many cases".

I am merely proposing a way to cherry pick the roll back of a particular set of migrations without the need to also roll back the last migrations (which is what `--step` offers).

driesvints:
@PHPGuus Please enable maintainer edit access

driesvints:
@PHPGuus you're submitting from an organisation account. Please resubmit from your personal account because maintainer edit access cannot be granted on organisation PR's.

PR #41123, opened 2022-02-19
[9.x] Added support for index and position placeholders in array validation messages

Bird87ZA:
Adds a backward compatible change that enables a developer to use `:index` or `:pos` in the message for array inputs. Assumes 0 or 1 respectively if it's not an array input.

Discussed in #32977

### DIFF
--- a/src/Illuminate/Validation/Concerns/FormatsMessages.php
+++ b/src/Illuminate/Validation/Concerns/FormatsMessages.php
@@ -229,9 +229,7 @@ trait FormatsMessages
         );
 
         $message = $this->replaceInputPlaceholder($message, $attribute);
-
         $message = $this->replaceIndexPlaceholder($message, $attribute);
-
         $message = $this->replacePositionPlaceholder($message, $attribute);
 
         if (isset($this->replacers[Str::snake($rule)])) {
@@ -320,11 +318,13 @@ trait FormatsMessages
      */
     protected function replaceIndexPlaceholder($message, $attribute)
     {
-        return str_ireplace(':index', explode('.', $attribute)[1] ?? 0, $message);
+        return $this->replaceIndexOrPositionPlaceholder(
+            $message, $attribute, 'index'
+        );
     }
 
     /**
-     * Replace the :pos placeholder in the given message.
+     * Replace the :position placeholder in the given message.
      *
      * @param  string  $message
      * @param  string  $attribute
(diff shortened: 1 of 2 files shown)


Bird87ZA:
Not sure why the builds are failing. Like I can read why, but can't see if and what I should change to fix it.

src/Illuminate/Validation/Concerns/FormatsMessages.php
```diff
@@ -307,6 +311,35 @@ protected function replaceAttributePlaceholder($message, $value)
...
+     * Replace the :pos placeholder in the given message.
+     *
+     * @param  string  $message
+     * @param  string  $attribute
+     * @return string
+     */
+    protected function replacePositionPlaceholder($message, $attribute)
+    {
+        $index = explode('.', $attribute);
+        $i = ($index[1] ?? 0) + 1;
```
gguney, line 338: if the $attribute is string then exploding it should return array of strings then you try to increase a string so it will throw an error. You may need to cast it to int. And you don't need $i variable you can write it directly inside irreplace

src/Illuminate/Validation/Concerns/FormatsMessages.php
```diff
@@ -307,6 +311,35 @@ protected function replaceAttributePlaceholder($message, $value)
...
+     *
+     * @param  string  $message
+     * @param  string  $attribute
+     * @return string
+     */
+    protected function replaceIndexPlaceholder($message, $attribute)
+    {
+        $index = explode('.', $attribute);
+
+        return str_ireplace(':index', $index[1] ?? 0, $message);
```
gguney, line 325:
```suggestion
        return str_ireplace(':index', explode('.', $attribute) ?? 0, $message);
```

Bird87ZA:
@gguney both changes are made. Thank you. Any idea why the PHP testing fails?

Bird87ZA:
Seems to be all passing now so no issues there

taylorotwell:
What if the item is more deeply nested? `foo.*.bar.*.baz`, etc.?

Bird87ZA:
Perhaps something like this?

```
    // $message would be something like 'Item :key-index is required'

    protected function replaceIndexPlaceholder($message, $attribute)
    {
        $attribute = explode('.', $attribute);

        if (count($attribute) > 1) {
            $message = str_ireplace(sprintf(':%s-index', $attribute[0]), $attribute[1] ?? 0, $message);
            unset($attribute[0], $attribute[1]);
            if (count($attribute) > 1) {
                return $this->replaceIndexPlaceholder($message, implode('.', $attribute));
            } else {
                return $message;
            }
        } else {
            return str_ireplace(':index',  0, $message);
        }
    }
```
    
 I'm all ears if you have a different approach.

ionutantohi:
This has been attempted before in #35754. You might want to check it out.

Bird87ZA:
> This has been attempted before in #35754. You might want to check it out.

Thanks, will have a read through it and see what I can come up with.

taylorotwell:
I've updated this so that more deeply nested indexes / positions can be accessed via `second-index, third-index, fourth-index`, etc.

```php
    $data = [
        'users' => [
            [
                'name' => 'Taylor Otwell',
                'attributes' => [
                    'foo',
                ],
            ],
            [
                'name' => '',
                'attributes' => [
                    'foo',
                    'bar',
                    1,
                ],
            ],
        ]
    ];

    $validator = Validator::make($data, [
        'users.*.name' => 'required|string',
        'users.*.attributes.*' => 'string',
    ], [
        'users.*.name.required' => ':index :position',
        'users.*.attributes.*.string' => ':first-index :first-position :second-index :second-position',
    ]);
```

PR #41126, opened 2022-02-19
[9.x] Add selectAllExcept() method to query builder

reziamini:
In some conditions, we need to pull all columns from the table except one which is heavy and takes lots of memory.
For example, we have a table for our articles and it has a column like `body` and we need to pull all tables except one of them (or more).

From now on we can do this:
```php
Article::query()->selectAllExcept('body')->get()
```

Or with multiple columns:

```php
Article::query()->selectAllExcept(['body', 'picture'])->get()
```

Also, I know that a query will be executed to get the tables, but in some conditions, it does make sense to execute a query instead of passing a huge array by yourself to make it happen and just filter a column.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -25,6 +25,7 @@ use Illuminate\Support\Traits\Macroable;
 use InvalidArgumentException;
 use LogicException;
 use RuntimeException;
+use Illuminate\Support\Facades\Schema;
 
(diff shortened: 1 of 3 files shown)


dennisprudlo:
I'm not sure how to feel about the fact, that the query "builder" runs a query while being built. Wouldn't it be better to execute the query on `information_schema` when the query builder is executed? Just like relationships.

taylorotwell:

PR #41129, opened 2022-02-20
[9.x] RateLimiting: Fix tooManyAttempts method

ismailnakkar:
The current usage of tooManyAttempts has some issues. In this pull request I tried fixing it
Currently, tooManyAttempts never resets when the timer finishes

### DIFF
--- a/src/Illuminate/Cache/RateLimiter.php
+++ b/src/Illuminate/Cache/RateLimiter.php
@@ -92,11 +92,11 @@ class RateLimiter
         $key = $this->cleanRateLimiterKey($key);
 
         if ($this->attempts($key) >= $maxAttempts) {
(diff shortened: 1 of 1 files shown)


driesvints:
Please provide more info why the current way doesn't work for you.

ismailnakkar:
As you can see, the function does not take into consideration the time limit, which means that it never resets when 1 minute passes. This makes RateLimiting unintuitive and I need to clear the cache key every time with the help availableAt method as a work around this

ismailnakkar:
The method "attempt" does not work also because it relies on tooManyattemps method, and thus, when the check fails because too many attempts one time, it never resets

taylorotwell:

Thanks!

PR #41130, opened 2022-02-20
Add faker() helper

ghost:
This pull request adds a faker() helper, primarily to use in blade views. 

This makes it ease to get fake data in your Laravel blade view, without having to make a factory based on your eloquent model.

```php
<div>{{ faker()->name() }}</div>
```

Get a collection of faker instances
````php
<ul>
    @foreach(faker(10) as $person)
        <li>{{ $person->name() }}</li>
    @endforeach
</ul>
````

Get a random image 
````php
    <img src="{{ faker()->image() }}" alt="{{ faker()->alt() }}" />
````

### DIFF
--- a/src/Illuminate/Support/Faker.php
+++ b/src/Illuminate/Support/Faker.php
@@ -50,6 +50,7 @@ class Faker extends Factory
 
     /**
      * Return a faker sentence to use for our random img.
+     * 
      * @return string
      */
     public function alt(): string
--- a/src/Illuminate/Support/Faker.php
+++ b/src/Illuminate/Support/Faker.php
@@ -1,0 +1,59 @@
+<?php
+
+namespace Illuminate\Support;
+
+use Exception;
+use Faker\Factory;
+use Faker\Generator;
+
+class Faker extends Factory
+{
+    public Generator $faker;
+
+    public function __construct()
+    {
+        $this->faker = self::create();
+    }
+
+    /**
+     * Check the given method name.
+     * Here we can choose to use our own method,
+     * Or call a faker method.
+     *
+     * @param string $name
+     * @param array $arguments
+     * @return mixed
+     */
+    public function __call(string $name, array $arguments)
+    {
+        return match ($name) {
+            'image', 'alt' => $this->faker->$name(),
+            default => $this->faker->$name(),
+        };
+    }
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/Illuminate/Support/Faker.php [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+     * Or call a faker method.
+     *
+     * @param  string  $name
+     * @param  array  $arguments
+     * @return mixed
+     */
+    public function __call(string $name, array $arguments)
+    {
+        return match ($name) {
+            'image', 'alt' => $this->faker->$name(),
```
rodrigopedra, line 30: Shouldn't here be  `$this->$name()` instead?

Otherwise it is equal to the `default` arm

ankurk91:
Faker is a dev dependency, and would fail to run in production.
I afraid that developer may start using this method beyond seeders

JohannesCleve:
> Faker is a dev dependency, and would fail to run in production.
I afraid that developer may start using this method beyond seeders

@ankurk91 That's a good point. Although this helper would only make sense for development, there isn't any way to stop a developer for mistakingly use it in production.

One solution might be to check the environment in the Faker magic __cal method and throw an exception when not in development mode. This might make it more clear for anyone using this helper. Although not ideal.

src/Illuminate/Support/Faker.php
JohannesCleve: @rodrigopedra I think you are correcrt, code could be simplified to:

```php
    public function __call(string $name, array $arguments)
    {
        return $this->faker->$name();
    }
```

Then it would still use the custom image and alt method, and pass the other calls to faker.

dennisprudlo:
@JohannesCleve An exception will be thrown in production either way, because the classes for faker won't exist in production. I think @ankurk91 concerns are that people use the faker helper in a dev environment because it works and then deploy the code to production, where it will fail.

src/Illuminate/Support/Faker.php
rodrigopedra: Oh sure… I missed that the magic method will only be called for missing methods… might be the late hours from yesterday :)

taylorotwell:
Feel free to add this helper to your own application!

PR #41132, opened 2022-02-21
[9.x] Improve factory stub type annotation

rodrigopedra:
https://github.com/laravel/laravel Ships with a `UserFactory` with an improved type annotation

https://github.com/laravel/laravel/blob/871ff9e65fea5588b1018d8770db6fc6ad37069d/database/factories/UserFactory.php#L13-L27

But when creating new factories with the `make:factory` or with `make:model` and the `-f` option, the factory definition does not include this improved type annotation.

This PR:

- Improves the type annotation on the `factory.stub` file

### DIFF
--- a/src/Illuminate/Database/Console/Factories/stubs/factory.stub
+++ b/src/Illuminate/Database/Console/Factories/stubs/factory.stub
@@ -12,7 +12,7 @@ class {{ factory }}Factory extends Factory
     /**
      * Define the model's default state.
      *
-     * @return array
+     * @return array<string, mixed>
      */
     public function definition()
     {


src/Illuminate/Database/Console/Factories/stubs/factory.stub
```diff
@@ -12,7 +12,7 @@ class {{ factory }}Factory extends Factory
     /**
      * Define the model's default state.
      *
-     * @return array
+     * @return array<string, mixed>
```
gguney, line 15:
```suggestion
     * @return array|string|mixed
```

actually mixed reflects all of them.

rodrigopedra: I based on the type annotation from here:

https://github.com/laravel/laravel/blob/871ff9e65fea5588b1018d8770db6fc6ad37069d/database/factories/UserFactory.php#L16

From what I know, these  "generic" style type definitions were introduced to improve PHPStan and other static analysis tool usage.

They actually reads like this: "The return type will be an array with string keys and any kind of values"

Also your suggestion suggests a `string` or any other type (`mixed`) could be returned, which is not true, as this method should always return an array.

PR #41134, opened 2022-02-21
[9.x] Rollback a specific batch of migrations

PHPGuus:
- Add a new option `batch` to the RollbackCommand.
- Add a new function `getMigrationsByBatch` to the MigrationRepositoryInterface.
- Implement `getMigrationsByBatch` in DatabaseMigrationRepository.

This new option allows the developer to rollback a specific batch of migrations. This can be helpful
in situations where a production bugfix introduced database changes after the current feature branch introduced database changes. The develop may not be able to rollback the production bugfix batch, but still wants to revert the batch from the current feature branch to make slight changes to the database migrations introduced in the current feature branch.

A PR to laravel/docs will also be submitted.

### DIFF
--- a/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
@@ -60,6 +60,7 @@ class RollbackCommand extends BaseCommand
                 $this->getMigrationPaths(), [
                     'pretend' => $this->option('pretend'),
                     'step' => (int) $this->option('step'),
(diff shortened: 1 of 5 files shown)


PHPGuus:
laravel/docs PR #7735

PHPGuus:
#41118 was closed, because I submitted it from an org account, which cannot allow maintainer edits... Sorry for double work!

taylorotwell:
Unfortunately I can't add a method to that interface on a patch release... will hold off on this feature for now.

PR #41137, opened 2022-02-21
[9.x] Integrate Laravel CORS into framework

driesvints:
This PR ports the middleware from https://github.com/fruitcake/laravel-cors over to the core Laravel framework. The main reason is that we want to remove a circular dependency we rely on additionally to the fact that we eliminate another dependency of the skeleton. 

All credits for the code go to @barryvdh of @fruitcake. Thanks for maintaining that package for so long!

After this PR has been merged and tagged I'll undraft the PR that contains the changes needed for the Laravel skeleton: https://github.com/laravel/laravel/pull/5825

### DIFF
--- a/src/Illuminate/Http/Middleware/TrustHosts.php
+++ b/src/Illuminate/Http/Middleware/TrustHosts.php
@@ -2,7 +2,6 @@
 
 namespace Illuminate\Http\Middleware;
 
-use Closure;
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Http\Request;
 
@@ -40,7 +39,7 @@ abstract class TrustHosts
      * @param  \Closure  $next
      * @return \Illuminate\Http\Response
      */
-    public function handle(Request $request, Closure $next)
+    public function handle(Request $request, $next)
     {
         if ($this->shouldSpecifyTrustedHosts()) {
             Request::setTrustedHosts(array_filter($this->hosts()));
--- a/src/Illuminate/Http/Middleware/TrustHosts.php
+++ b/src/Illuminate/Http/Middleware/TrustHosts.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Http\Middleware;
 
+use Closure;
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Http\Request;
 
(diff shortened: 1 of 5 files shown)


src/Illuminate/Http/Middleware/TrustHosts.php
```diff
@@ -36,10 +37,10 @@ abstract public function hosts();
      * Handle the incoming request.
      *
      * @param  \Illuminate\Http\Request  $request
-     * @param  callable  $next
+     * @param  \Closure  $next
      * @return \Illuminate\Http\Response
      */
-    public function handle(Request $request, $next)
+    public function handle(Request $request, Closure $next)
```
driesvints, line 43: I spotted that the method signature didn't contain the `Closure` typehint here even though all other middleware do. Thought I'd fix this one up here as well while I'm at it.

browner12:
Should we remove the `fruitcake/laravel-cors` dependency in our projects, and update our kernel middleware?

Is the intent that people upgrade this now, or wait for v10?

barryvdh:
You can just apply the changes listed here directly, if you are on 9 already: https://github.com/laravel/laravel/pull/5825/files

browner12:
Thanks @barryvdh.  In addition to those changes I'm assuming we'll have to remove our local composer requirement of `fruitcake/laravel-cors` as well.

I guess the question is more if and how the team is going to disseminate this change to the public, since it's occurring on a minor version.  I only came across this by chance after running a `composer outdated`.  Or if the thought is people are fine as is until v10.

barryvdh:
It's fine either way, but you can remove the local package, if you also replace the middleware (same as the changes in Laravel/laravel)

driesvints:
There's no breaking change. You're totally free to leave the current package in place or move to the new one.

jonnott:
@driesvints Should this possibly now be mentioned in the Laravel 9 upgrade guide, as an optional change to the skeleton?

driesvints:
@jonnott not really since, like you say, it's optional. It's not required to upgrade.

jonnott:
> @jonnott not really since, like you say, it's optional. It's not required to upgrade.

True, but a lot of these kind of changes have at least been mentioned-in-passing before in upgrade guides, I guess with the intention to help people maintain as 'clean' a codebase as possible as they upgrade through the versions..?

driesvints:
We already have a changelog for the skeleton: https://github.com/laravel/laravel/blob/9.x/CHANGELOG.md

And often announce these on Twitter: https://twitter.com/driesvints/status/1497146821491281922

The upgrade guide really is meant to only contain the bare minimum to upgrade to a new version. A better place would be the release notes maybe: https://laravel.com/docs/9.x/releases#laravel-9. But the feature is so small I don't think it's worth adding.

PR #41147, opened 2022-02-21
[9.x] Make `Enumerable` and `Arrayable` covariance-aware

ryangjchandler:
Since the templates are marked with `@template`, PHPStan won't adhere to covariance rules.

An example would be:

```php
class Parent {}
class Child extends Parent {}

/** @return Collection<int, Parent> */
function test(): Collection
{
	return collect([new Child]);
}
```

In an application of mine, a similar scenario is currently causing PHPStan to fail. This PR just updates Collection types to use the `@template-covariant` tag instead.

Running PHPStan with this branch as the framework target resolves those problems. The only thing I'm not super sure about is testing this inside of the `types` test files, wonder if @nunomaduro could provide some pointers.

This is supported by both PHPStan and Psalm, fwiw.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -12,7 +12,7 @@ use Traversable;
 
 /**
  * @template TKey of array-key
- * @template TValue
+ * @template-covariant TValue
  *
  * @implements \ArrayAccess<TKey, TValue>
  * @implements \Illuminate\Support\Enumerable<TKey, TValue>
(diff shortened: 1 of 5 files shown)


nunomaduro:
@ryangjchandler Can you add a failing to test to https://github.com/laravel/framework/tree/9.x/types/Support, so we can understand the problem you are trying to solve?

ryangjchandler:
@nunomaduro Sure, I'll do this tomorrow.

nunomaduro:
Fell free to re-open this pull request once you have added tests.

PR #41160, opened 2022-02-22
[9.x] Dump and die after X times

amirHossein5:
Dumps and dies after the given time.

```php
foreach (['a','b','c','d'] as $item) {
  ddt(3, $item, 'test');

  // will dump variables and dies after( when ) 3 times.
}
```

```php
foreach (['a','b','c','d'] as $item) {
  ddt(1, $item, 'test');

  // will dump variables and dies after( when ) 1 time. (like dd())
}
```

```php
foreach (['a','b','c','d'] as $item) {
  ddt(0, $item, 'test');
}

  // just dies.
```

it also can be useful when don't have size of array, or just want to dump more items and die.


taylorotwell:

Issue #41153, opened 2022-02-21
$connection property not working in Migration. 


Yszty:
- Laravel Version: 8.75
- PHP Version: 8.1
- Database Driver & Version: mariadb

### Description:

After declaring `protected $connection = 'mysql_external';`  Schema facade still using default connection, but should declared in `$connection` property. 

So, fix the framework or change documentation... https://laravel.com/docs/8.x/migrations#setting-the-migration-connection

I see the problem exist form laravel 5, and still arent fixed, but issue closed? https://github.com/laravel/framework/issues/31060

Resulting problems:
-php artisan mirgrate:fresh not droping external db tables

### Steps To Reproduce:

![image](https://user-images.githubusercontent.com/23386525/155025931-ce3b94fd-4b88-4dbe-b21a-7d2b0961f90c.png)

derekmd:
Mar 2021 comment from Taylor about the state of things: https://github.com/laravel/framework/issues/36596#issuecomment-799784094

`$connection` alone can't be used to configure a secondary database. `Schema::connection()` is required too:

```php
protected $connection = 'mysql_external';

function down()
{
    Schema::connection('mysql_external')->drop('user_settings');
}
```

or:

```php
function down()
{
    Schema::drop('user_settings');
}
```

```
php artisan migrate:rollback --database=mysql_external
```

---

This docs section was added last year: https://github.com/laravel/docs/commit/f3f5a24aa228290d69404f4f4fbdb41a4b0b7d1b

![connect-me-not](https://user-images.githubusercontent.com/823566/155074967-69a6e1b6-3386-4fc2-8f6d-70bcc66db4eb.png)

I'm unsure if the framework behaved this way after looking at branches 7.x, 6.x, 5.8, 5.7, then back to 5.4. I can't find anytime `Illuminate\Database\Migrations\Migration@getConnection()` was explicitly switched into per migration before methods `up()` or `down()` are called. That `$connection` is only used by `DB::transaction()` and `DB::pretend()` but the queries would still run using the default connection (or the `--database` command override.)

A bug fix PR was attempted last year (https://github.com/laravel/framework/pull/35841 https://github.com/laravel/framework/pull/35840) but it lacked important additional tests on a critical feature (:rotating_light:) so it was closed.

[...]

Issue #40097, opened 2021-12-18
php artisan migrate:reset not reseting other database connections


Nimaw:
- Laravel Version: 8.28.1
- PHP Version: 8.0.9
- Database Driver & Version: MYSQL 10.4.20-MariaDB

### Description:
When i've running command `php artisan migrate:reset` it reseting created tables but it isn't reset other database connections. 

### Steps To Reproduce:
Create some migrations with different database connections then run `php artisan migrate`, it runs all migrations on multiple connections, but  after if you run `php artisan migrate:reset` it isn't reset all tables.

driesvints:
Hey there,

Can you first please try one of the support channels below? If you can actually identify this as a bug, feel free to open up a new issue with a link to the original one and we'll gladly help you out.

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

Thanks!

Issue #38777, opened 2021-09-12
$connection property ignored in migrations


JohnnyWalkerDesign:
- Laravel Version: 8.60.0
- PHP Version: 7.4.11
- Database Driver & Version: MySQL 8.0.25

### Description:
The [documentation](https://laravel.com/docs/8.x/migrations#setting-the-migration-connection) states that you can create a migration connection by using the $connection property. I have done this with 70 migrations in my project and can confirm that the default connection was used for all migrations. 

### Steps To Reproduce:
1. Create a second database
2. Create a new migration that you wish to use with the second database
3. Add `protected $connection = 'second_database';` to the top of your migration.
4. Run migration.

The migration should use the connection specified in the `$connection` property, but it will ignore it and use the default connection.

derekmd:
Commit that changed the behavior starting in laravel/framework 7.0.0: https://github.com/laravel/framework/commit/f4bd05fc1015a1a49dd6850d6817a10d129143c8

Pull request https://github.com/laravel/framework/pull/35842 was rejected earlier this year but I think only because it lacked a bug report description and the addition of tests to cover documented behavior.

lk77:
I've tested it and it work :

```
class TestMigration extends Migration
{
    protected $connection = 'does_not_exist';
    [...]
}
```

```
Migrating: 2021_09_13_090345_test_migration

   InvalidArgumentException 

  Database connection [does_not_exist] not configured.

  at vendor/laravel/framework/src/Illuminate/Database/DatabaseManager.php:152
    148▕         // If the configuration doesn't exist, we'll throw an exception and bail.
    149▕         $connections = $this->app['config']['database.connections'];
    150▕ 
    151▕         if (is_null($config = Arr::get($connections, $name))) {
  ➜ 152▕             throw new InvalidArgumentException("Database connection [{$name}] not configured.");
    153▕         }
    154▕ 
    155▕         return (new ConfigurationUrlParser)
    156▕                     ->parseConfiguration($config);

      +23 vendor frames 
  24  artisan:37
      Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
```

Issue #36596, opened 2021-03-15
artisan migrate with --pretend flag doesn't work with multiple connections


dshafik:
- Laravel Version: 8.22.1
- PHP Version: 7.4.13
- Database Driver & Version: pdo_mysql, MySQL 5.6

### Description:

When using the `--pretend` flag with `php artisan migrate` and using multiple connections, the schema changes are actually applied to the non-default DB.

### Steps To Reproduce:

Add a second DB to `config/database.php` e.g. `testing`:

```php
     'testing' => [
            'host' => env('DB_TESTING_HOST', '127.0.0.1'),
            'driver' => 'mysql',
            'port' => env('DB_TESTING_PORT', '3306'),
            'database' => env('DB_TESTING_DATABASE', 'analytics'),
            'username' => env('DB_TESTING_USERNAME', 'root'),
            'password' => env('DB_TESTING_PASSWORD', ''),
            'unix_socket' => '',
            'charset' => 'utf8mb4',
            'collation' => 'utf8mb4_unicode_ci',
            'strict' => true,
            'engine' => null,
            'options' => extension_loaded('pdo_mysql') ? array_filter([
                PDO::MYSQL_ATTR_SSL_CA => env('MYSQL_ATTR_SSL_CA'),
            ]) : [],
        ],
```

Create a fresh migration:

```sh
php artisan make:migration create_testing_table
```

Update the `up()` method to use the new connection:

```php
    public function up()
    {
        Schema::connect('testing')->create('testing', function (Blueprint $table) {
            $table->id();
            $table->timestamps();
        });
    }
```

Run migrate with pretend:

```sh
php artisan migrate --pretend
```

When doing this, I see the standard `--pretend` SQL output for migrations using the default DB connection, but this migration runs no matter what (and no `--pretend` SQL output), e.g. if you also `php artisan make:migration create_foo_table` and don't modify it:

```
$ php artisan migrate --pretend
CreateFooTable: create table `foo` (`id` bigint unsigned not null auto_increment primary key, `created_at` timestamp null, `updated_at` timestamp null) default character set utf8mb4 collate 'utf8mb4_unicode_ci'
```

**Additionally** the `migrations` table isn't updated (as expected), so If you run it _again_, it'll fail because the `testing` table already exists, and if you try to rollback, it'll rollback the _**previous**_ migration, which could be disastrous.

taylorotwell:
[...]

PR #41161, opened 2022-02-22
[8.x] Fix database migrations `$connection` property

derekmd:
Fixes https://github.com/laravel/framework/issues/41153 https://github.com/laravel/framework/issues/40097 https://github.com/laravel/framework/issues/38777 https://github.com/laravel/framework/issues/36596 https://github.com/laravel/framework/issues/31060 

The [docs](https://laravel.com/docs/8.x/migrations#setting-the-migration-connection) show the below migration will be run on the named database connection:

```php
{
    protected $connection = 'pgsql';

    function up()
    {
            // ...
        });
    }
}
```

However `Schema` facade-built queries will run on the app's default database connection or the `--database` command line option's named connection. The migration also requires `Schema::connection('pgsql')`:

```php
{
    protected $connection = 'pgsql';

    function up()
    {
        Schema::connection('pgsql')->create('users', function (Blueprint $table) {
            // ...
        });
    }
}
```

This fix allows the first example to work as documented for the `up()` & `down()` methods. The selected connection priority is:

1. `Schema::connection()`
2. migration class property `$connection`
3. `php artisan migrate*` command line option `--database`
4. `config('database.default')`

I also added tests for the `Migrator@usingConnection()` method added in 7.x for the migration commands.

### DIFF
--- a/src/Illuminate/Database/Migrations/Migrator.php
+++ b/src/Illuminate/Database/Migrations/Migrator.php
@@ -387,11 +387,11 @@ class Migrator
             $migration->getConnection()
         );
 
-        $callback = function () use ($migration, $method) {
+        $callback = function () use ($connection, $migration, $method) {
             if (method_exists($migration, $method)) {
                 $this->fireMigrationEvent(new MigrationStarted($migration, $method));
 
-                $migration->{$method}();
+                $this->runMethod($connection, $migration, $method);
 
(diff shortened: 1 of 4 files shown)


rawmarius:
Thank you so much for the fix! It would be nice if the priority would be added to the documentation.
Regards!

PR #40587, opened 2022-01-24
[9.x] Use new CommonMark 2.2 API

GrahamCampbell:

PR #41162, opened 2022-02-22
[9.x] Update deprecated method on Str Support

wit3:
Since league/commonmark 2.2.0: Calling "convertToHtml()" on a League\CommonMark\MarkdownConverter class is deprecated, use "convert()" instead.

Update depreacted method


GrahamCampbell:
Looks like this was missed in https://github.com/laravel/framework/pull/40587.

PR #41164, opened 2022-02-22
[9.x] Let Multiple* exceptions hold the number of records and items found

lucasmichot:
It can be pretty handy to get access to the numbers of items or records found within `MultipleItemsFoundException` and `MultipleRecordsFoundException` exceptions.

This PR ensures that those numbers are always accessible.

### DIFF
--- a/src/Illuminate/Collections/MultipleItemsFoundException.php
+++ b/src/Illuminate/Collections/MultipleItemsFoundException.php
@@ -6,4 +6,34 @@ use RuntimeException;
 
 class MultipleItemsFoundException extends RuntimeException
 {
+    /**
+     * The number of items found.
+     *
+     * @var int
+     */
+    protected $count;
+
+    /**
+     * MultipleItemsFoundException constructor.
+     *
+     * @param  int  $count
+     * @param  int  $code
+     * @param  \Throwable|null  $previous
+     * @return void
+     */
+    public function __construct($count, $code = 0, $previous = null)
+    {
+        $this->count = $count;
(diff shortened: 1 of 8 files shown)


src/Illuminate/Collections/MultipleItemsFoundException.php
```diff
@@ -6,4 +6,34 @@
...
+
+    /**
+     * MultipleItemsFoundException constructor.
+     *
+     * @param  int  $count
+     * @param  int  $code
+     * @param  \Throwable|null  $previous
+     * @return void
+     */
+    public function __construct($count, $code = 0, $previous = null)
```
driesvints, line 24: This sounds like a breaking change? This exception now has a new type of required dependency which wasn't there before.

src/Illuminate/Database/MultipleRecordsFoundException.php
```diff
@@ -6,5 +6,34 @@
...
+
+    /**
+     * MultipleRecordsFoundException constructor.
+     *
+     * @param  int  $count
+     * @param  int  $code
+     * @param  \Throwable|null  $previous
+     * @return void
+     */
+    public function __construct($count, $code = 0, $previous = null)
```
driesvints, line 24: Same here.

src/Illuminate/Collections/MultipleItemsFoundException.php
driesvints: Actually, it's not so bad I think. This should be good 👍

PR #41173, opened 2022-02-22
[9.x] Enforce Mockery m aliasing consistency

lucasmichot:
This PR enforces `Mockery` `m` aliasing consistency.

### DIFF
--- a/tests/Database/DatabaseEloquentCollectionQueueableTest.php
+++ b/tests/Database/DatabaseEloquentCollectionQueueableTest.php
@@ -5,19 +5,19 @@ namespace Illuminate\Tests\Database;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Pivot;
-use Mockery;
+use Mockery as m;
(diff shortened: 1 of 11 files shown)


driesvints (approved):

tests/Database/DatabaseEloquentCollectionQueueableTest.php
```diff
@@ -5,19 +5,19 @@
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Pivot;
-use Mockery;
+use Mockery as m;
```
gguney, line 8: I think it is not clear enough to just use "m" as an alias. I had to look at to the top to understand what is "m"

lucasmichot: `m` is used everywhere in the tests; those were the only 11 files where it was not the case.
And within the tests `m::mock` are pretty self-explanatory (imo)

driesvints: `m` is a standard across PHP projects to alias Mockery.

PR #41178, opened 2022-02-22
[9.x] Fix setUp and tearDown method visibility and casing.

lucasmichot:
This PR fix `setUp` and `tearDown` method visibility and casing.


driesvints:
Thanks!

PR #41179, opened 2022-02-22
[9.x] Ensure that assertions methods are not called statically.

lucasmichot:
This PR ensures that assertions methods are not called statically.


driesvints:
Thanks!

PR #41186, opened 2022-02-23
[9.x] Include missing PHPDoc signature for includeUnvalidatedArrayKeys()

silverjason:
Laravel 9 includes the option to `includeUnvalidatedArrayKeys` to opt-in to previous Laravel 8 behaviour with regards to [unvalidated array keys](https://laravel.com/docs/9.x/upgrade#unvalidated-array-keys) validation.

When adding this on within the boot method of one of the service providers I noticed it was undocumented in the Laravel Validator Facade.

I'm requesting to add the declaration to the signature for transparency.


nhanvu327:
LGTM

PR #41190, opened 2022-02-23
[9.x] Update Builder.php

celaraze:
Add a basic find_in_set query to the query.
This method can realize the function of `find_in_set()` in SQL.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1724,6 +1724,18 @@ class Builder implements BuilderContract
             }
         }
     }
+    
+    /**
+     * Add a basic find_in_set query to the query.
+     *
+     * @param mixed $column
+     * @param mixed $value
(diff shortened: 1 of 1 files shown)


celaraze:
in Laravel:
`User::query()->findInSet('role_ids','1')->first();`

in SQL:
| name | role_ids |
|------|----------|
| Tom  | 1,2      |
| Jack | 2        |

rodrigopedra:
Some concerns:

- `FIND_IN_SET(...)` is a function exclusive to MySQL, as such it should be added to the MySQL Grammar
- What you propose is a `WHERE` clause, so implementation should match other `WHERE` clauses (`whereNull`, `whereIn`, ...), and support `OR` for example
- There should be tests added to ensure expected output

As the Builder is macroable, I think you are better off adding this on your own app. 

As any added code results in more maintenance work to maintainers, imagine if they add every function each database supports and address any differences between them? It would also be very hard to keep up to date with changes between database versions.

taylorotwell:

PR #41199, opened 2022-02-23
[9.x] Refactor paginators

iamgergo:
This PR refactors the paginator classes (`Paginator`, `LengthAwarePaginator`, `CusrorPaginator`).

-------

I realized there were a lots of code that are exactly the same in the 3 paginator classes or in their parent abstract classes. 

Also, it was interesting to see that in some cases the child class was implementing an interface, but it was its parent who held the method for that interface. I know it's not a bug (and it's possible there is a reason behind it), but it looked a bit wierd for me.

The point is, I created an abstract base paginator that holds all the common methods for the paginators also the common interfaces they implement.

--------

Note: I made the `render` method as an abstract, since it is called in the `AbstractBasePaginator` class but its implementation (not the method signature) is different in all three classes.

--------

Why? Besides code duplication, I face with a situation where I use cursor and length aware pagination. I'm planning to sumbit a PR where we can easily add custom data for the array representation of the pagination classes (like custom meta stuff, in some case it can be handy).

There is no point to add the same code to all the 3 classes, so as a first step I thought it's a good idea to refactor a bit to prevent future duplication.

--------

This should not be a BC, because yes, the base paginator has the interfaces now, but also it holds all the relevant methods as well. For the child classes nothing has changed because they inherit the same code that they has before.

### DIFF
--- a/src/Illuminate/Pagination/AbstractCursorPaginator.php
+++ b/src/Illuminate/Pagination/AbstractCursorPaginator.php
@@ -504,7 +504,7 @@ abstract class AbstractCursorPaginator extends AbstractBasePaginator
         ]));
     }
 
-        /**
+    /**
      * Get the instance as an array.
      *
      * @return array
--- a/src/Illuminate/Pagination/AbstractCursorPaginator.php
+++ b/src/Illuminate/Pagination/AbstractCursorPaginator.php
@@ -5,31 +5,17 @@ namespace Illuminate\Pagination;
 use ArrayAccess;
 use Closure;
 use Exception;
-use Illuminate\Contracts\Support\Htmlable;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Pivot;
 use Illuminate\Http\Resources\Json\JsonResource;
 use Illuminate\Support\Arr;
-use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
-use Illuminate\Support\Traits\ForwardsCalls;
-use Illuminate\Support\Traits\Tappable;
-use Traversable;
 
 /**
  * @mixin \Illuminate\Support\Collection
  */
-abstract class AbstractCursorPaginator implements Htmlable
+abstract class AbstractCursorPaginator extends AbstractBasePaginator
 {
-    use ForwardsCalls, Tappable;
-
-    /**
-     * All of the items being paginated.
-     *
-     * @var \Illuminate\Support\Collection
-     */
-    protected $items;
-
     /**
      * The number of items to be shown per page.
      *
@@ -78,13 +64,6 @@ abstract class AbstractCursorPaginator implements Htmlable
      * @var array
      */
     protected $parameters;
-
-    /**
-     * The paginator options.
-     *
-     * @var array
-     */
-    protected $options;
 
     /**
      * The current cursor resolver callback.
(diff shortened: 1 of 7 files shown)


iamgergo:
Soon I'll fix the tests, please don't close until. Thanks!

src/Illuminate/Pagination/AbstractCursorPaginator.php
```diff
@@ -512,152 +491,34 @@ public static function viewFactory()
...
-
-    /**
-     * Render the contents of the paginator when casting to a string.
-     *
-     * @return string
-     */
-    public function __toString()
-    {
-        return (string) $this->render();
+    public function toArray()
```
iamgergo, line 512: I moved up the `toArray` method here from the CursorPaginator child class. Changes nothing, but makes sure the tests are passing.

src/Illuminate/Pagination/AbstractCursorPaginator.php
```diff
@@ -512,152 +491,34 @@ public static function viewFactory()
...
-     * Determine if the list of items is not empty.
+     * Render the paginator using the given view.
      *
-     * @return bool
+     * @param  string|null  $view
+     * @param  array  $data
+     * @return \Illuminate\Contracts\Support\Htmlable
      */
-    public function isNotEmpty()
+    public function render($view = null, $data = [])
```
iamgergo, line 500: I moved up the `render` method here, from the `CursorPaginator` child class. Changes nothing, but makes sure the tests are passing.

src/Illuminate/Pagination/AbstractPaginator.php
```diff
@@ -646,152 +625,35 @@ public static function useBootstrapFive()
...
-
-    /**
-     * Render the contents of the paginator when casting to a string.
-     *
-     * @return string
-     */
-    public function __toString()
-    {
-        return (string) $this->render();
+    public function toArray()
```
iamgergo, line 646: The `Paginator` and the `LengthAwarePaginator` have some matching attributes. So I moved up the `toArray` method to their parent class with their common array attributes. So from the child classes it's enough just add their own data.

src/Illuminate/Pagination/AbstractPaginator.php
```diff
@@ -646,152 +625,35 @@ public static function useBootstrapFive()
...
-     * Determine if the list of items is not empty.
+     * Render the paginator using the given view.
      *
-     * @return bool
+     * @param  string|null  $view
+     * @param  array  $data
+     * @return \Illuminate\Contracts\Support\Htmlable
      */
-    public function isNotEmpty()
+    public function render($view = null, $data = [])
```
iamgergo, line 634: I moved the `Paginator`'s render method to the parent class, because it was simpler and all the required prop and attribute was available in the parent class as well. The `LengthAwarePaginator` still has its own render method that overrides this one here.

src/Illuminate/Pagination/LengthAwarePaginator.php
```diff
@@ -191,41 +185,12 @@ public function lastPage()
      */
     public function toArray()
     {
-        return [
-            'current_page' => $this->currentPage(),
-            'data' => $this->items->toArray(),
-            'first_page_url' => $this->url(1),
-            'from' => $this->firstItem(),
+        return array_merge(parent::toArray(), [
```
iamgergo, line 188: Here it's enough just to add the custom data related to `LengtAwarePaginator` class, the common ones come from the parent.

src/Illuminate/Pagination/Paginator.php
```diff
@@ -140,37 +120,8 @@ public function hasMorePages()
      */
     public function toArray()
     {
-        return [
-            'current_page' => $this->currentPage(),
-            'data' => $this->items->toArray(),
-            'first_page_url' => $this->url(1),
-            'from' => $this->firstItem(),
+        return array_merge(parent::toArray(), [
             'next_page_url' => $this->nextPageUrl(),
```
iamgergo, line 124: Here it's enough just to add the custom data related to `Paginator` class, the common ones come from the parent.

tests/Integration/Http/ResourceTest.php
```diff
@@ -553,7 +553,7 @@ public function testResourcesMayCustomizeJsonOptionsOnPaginatedResponse()
         );
 
         $this->assertEquals(
-            '{"data":[{"id":5,"title":"Test Title","reading_time":3.0}],"links":{"first":"\/?page=1","last":"\/?page=1","prev":null,"next":null},"meta":{"current_page":1,"from":1,"last_page":1,"links":[{"url":null,"label":"&laquo; Previous","active":false},{"url":"\/?page=1","label":"1","active":true},{"url":null,"label":"Next &raquo;","active":false}],"path":"\/","per_page":15,"to":1,"total":10}}',
+            '{"data":[{"id":5,"title":"Test Title","reading_time":3.0}],"links":{"first":"\/?page=1","last":"\/?page=1","prev":null,"next":null},"meta":{"current_page":1,"from":1,"path":"\/","per_page":15,"to":1,"last_page":1,"links":[{"url":null,"label":"&laquo; Previous","active":false},{"url":"\/?page=1","label":"1","active":true},{"url":null,"label":"Next &raquo;","active":false}],"total":10}}',
```
iamgergo, line 556: Because we are using array merge in the paginators to add custom data:

```php
public function toArray()
{
    return array_merge(parent::toArray(), [
        'next_page_url' => $this->nextPageUrl(),
     ]);
}
```

The keys can have different orders. So when encoding to JSON and the comparing it with a string, they won't match because some values can change place. This is what happened here.

iamgergo:
Okay, I might target 10.x then.

PR #41204, opened 2022-02-23
[10.x] Generalize paginator classes

iamgergo:
This is a retry of #41199 but now I'm trageting the master branch.

I thought give it another shot, since it's on the table for me.

My former PR was closed becase: 

> Too risky for me to merge such a large "refactoring" on a patch release that I didn't write myself.

I agree, but still I think this refactoring would be beneficial. Nothing disappears, just a bit of reorganization makes possible to keep the code DRY.

----------

Currently the paginator classes have quite much code that is exactly the same, I guess it was just copy-pasted. I think it would be a good idea, to have a base paginator class that holds all the common code that is being used by the paginator implementations.

Also, this way we can prevent future duplications.

### DIFF
--- a/src/Illuminate/Pagination/AbstractBasePaginator.php
+++ b/src/Illuminate/Pagination/AbstractBasePaginator.php
@@ -1,0 +1,217 @@
+<?php
+
+namespace Illuminate\Pagination;
+
+use ArrayAccess;
+use Countable;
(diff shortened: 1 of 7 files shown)


taylorotwell:
I'm sorry but there just isn't a ton of ROI for me on this. I just risk introducing bugs.

PR #41209, opened 2022-02-23
[9.x] Pass listener arguments to middleware method

adriaanzon:
In #38128, support for job middleware was added to queued event listensers.

This pull request passes the same arguments that are passed to handle(), to the middleware() method. Doing so allows us to construct the middleware with certain context. In my use case, I am using implementing a rate limiter that is scoped to a specific user:

```php
class SyncToExternalService implements ShouldQueue
{
    public function handle(Updated $event)
    {
    }

    public function middleware(Updated $event)
    {
        return [new RateLimitMiddleware($event->user->external_service_id)];
    }
}
```

This is backwards compatible because the argument list on middleware() can still remain empty.

### DIFF
--- a/src/Illuminate/Events/Dispatcher.php
+++ b/src/Illuminate/Events/Dispatcher.php
@@ -612,7 +612,7 @@ class Dispatcher implements DispatcherContract
      * Propagate listener options to the job.
      *
      * @param  mixed  $listener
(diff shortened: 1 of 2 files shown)


adriaanzon:
I suppose the same can be done for the listener's backoff() and retryUntil() methods.

PR #41210, opened 2022-02-23
[9.x] Allow passing query parameters to test requests

tontonsb:
This allows passing parameters as array to `get`, `getJson` and `json` (because it's used by `getJson`).

```php
// Now
$this->get($uri, $headers, $params);

// Before
$this->get($uri.'?'.http_build_query($params), $headers);
// or you had to hardcode a $uri with all the query string
```

This is especially useful when reusing the `$params` where manipulations on array keys are more obvious than reassigning a whole string.

```php
$params = [
    'search' => 'keyword',
    'perPage' => 10,
    'order' => 'default',
];

$this->getJson($uri, $headers, $params)->assertDefaultOrderIsCool();

$params['order'] = 'views';
$this->getJson($uri, $headers, $params)->assertIsOrderedByViews();
```

This is not a breaking change, because it's an optional parameter at the end of the list. If one doesn't need `$headers`, they can either `get($uri, [], $params)'` or `get($uri, parameters: $params)`.

In addition to testing the new feature, I wrote a more generic test that verifies the behaviour of `json()` method (and by proxy `call()`) as those appeared to not have any direct tests.

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
@@ -285,12 +285,12 @@ trait MakesHttpRequests
      * @param  array  $headers
      * @return \Illuminate\Testing\TestResponse
      */
-    public function get($uri, array $headers = [])
+    public function get($uri, array $headers = [], array $parameters = [])
     {
         $server = $this->transformHeadersToServerVars($headers);
         $cookies = $this->prepareCookiesForRequest();
 
-        return $this->call('GET', $uri, [], $cookies, [], $server);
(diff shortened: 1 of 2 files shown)


u01jmg3:
Docblocks need to be updated

driesvints:
This is still a breaking change for anyone overwriting these methods. Please send this to master, thanks.

tontonsb:
@driesvints sure I can, but does it really break some cases? I have a hard time coming up with such a case. If someone is overriding `get` or `getJson`, they would just keep their override. If someone is overriding `json` but not `getJson`, the new parameter of the proposed `getJson` won't work, but everything that did work would keep on working. Am I missing something?

driesvints:
Yes it can. Anyone who's overwriting this in their tests will see their test suite break.

PR #41216, opened 2022-02-24
[9.x]  Unset Connection Resolver extended callback

emrancu:
Unset Connection Resolver extended callback. Specially for Octane with multitenancy application.

By default, there is a method called 'extend' in DatabaseManager class for passing connection resolver to override default connection resolver. But there is no way to unset/remove callback to rollback to default connection resolver. 

Can be used: 
`app('db').forgetExtend($name)`

### DIFF
--- a/src/Illuminate/Database/DatabaseManager.php
+++ b/src/Illuminate/Database/DatabaseManager.php
@@ -397,12 +397,12 @@ class DatabaseManager implements ConnectionResolverInterface
     }
 
     /**
-     * unset an extension connection resolver.
+     * Remove an extension connection resolver.
      *
      * @param  string  $name
      * @return void
      */
-    public function forgetExtend($name)
+    public function forgetExtension($name)
     {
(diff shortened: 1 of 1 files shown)


src/Illuminate/Database/DatabaseManager.php
```diff
@@ -396,6 +396,19 @@ public function extend($name, callable $resolver)
...
+     * unset an extension connection resolver.
+     *
+     * @param  string  $name
+     * @return void
+     */
+    public function forgetExtend($name)
+    {
+        if (isset($this->extensions[$name])) {
+            unset($this->extensions[$name]);
+        }
```
derekmd, line 409:
```suggestion
        unset($this->extensions[$name]);
```

You wouldn't have to worry about the array key already existing or not.

alemran-xs: @derekmd, Thank you. I have changed.

PR #41221, opened 2022-02-24
[9.x] No need to declare a variable which is returned right after

lucasmichot:
This PR ensures that no variables are declared, to be returned just after.


driesvints (approved):

taylorotwell:
Personally I find those view ones to be helpful - situations where there are multiple transformations happening on a variable I find it readable to list them all and then return the variable.

PR #41226, opened 2022-02-24
[WIP] Always call parent setUp and tearDown.

lucasmichot:
Always ensure that testCase always call parent setUp and tearDown


driesvints:
This isn't needed imo if there's nothing in parent to resolve.

lucasmichot:
> This isn't needed imo if there's nothing in parent to resolve.

Unless if, at some point, a global AbstractTestCase would be introduced that would handle all the cleanup tasks, Mockery, Carbon and TZ reset, etc., etc...

PR #41228, opened 2022-02-24
[9.x] Refactoring Attribute.php to use php8 features 

mehdirajabi59:
Laravel 9  requires a minimum PHP version of 8.0 but not use PHP8 feature such as named argument and return type

To do it i refactor casts/Attribute.php  to uses named argument and return type


driesvints:
This makes it less readable in my 😅

driesvints:
Also, adding return types is a breaking change to the method signature.

Issue #41180, opened 2022-02-22
Unable to insert resource (stream) into table for MySQL, or SQLite database


aedart:
- Laravel Version: 8.x | 9.x
- PHP Version: 7.4.x | 8.x
- Database Driver & Version: MySql, SQLite

### Description:

When attempting to store a large file into a database table, e.g. a large photo, the database `Connection` does not bind a `resource` correctly, for MySQL and SQLite. Yet, for Postgresql it works?

### Steps To Reproduce:

```php
use Illuminate\Support\Facades\DB;

$file = fopen('/pictures/large_image.jpg', 'rb');

// Insert large file into table...
DB::table('files')
    ->insert([
        'file' => $file
    ]);

// Select inserted record (...not shown here...)

echo $record->file; // "Resource id #1801" - expected file's content,...
```

The above example works when using Postgres connection.

### Cause

The `\Illuminate\Database\Connection::bindValues()` method does not handle values of the type `resource`. Yet, the `\Illuminate\Database\PostgresConnection` does.

Code from `PostgresConnection` 

```php
// Inside PostgresConnection::bindValues()
foreach ($bindings as $key => $value) {
    if (is_int($value)) {
        $pdoParam = PDO::PARAM_INT;
    } elseif (is_resource($value)) {    // --> handling of resource type
        $pdoParam = PDO::PARAM_LOB;
    } else {
        $pdoParam = PDO::PARAM_STR;
    }

    // ...remaining not shown ...
}
```

### Possible Solution

Update the `\Illuminate\Database\Connection::bindValues()` to handle values of type `resource`.

### Alternative Solution

Perhaps allow specifying a callback that allows developers to define how values should be bound to PDO statement, in `Connection`.

E.g. 
```php
// Inside Connection class

public function bindUsing($callback) {
    $this->bindCallback = $callback;
}

public function bindValues($statement, $bindings)
{
    // Use custom callback, if one is provided...
    if (isset($this->bindCallback)) {
        return $this->bindCallback($statement, $bindings);
    }
    
    // Default to current implementation
    foreach ($bindings as $key => $value) {
        $statement->bindValue(
            is_string($key) ? $key : $key + 1,
            $value,
            is_int($value) ? PDO::PARAM_INT : PDO::PARAM_STR
        );
    }
}
```

If this solution should be chosen, then developers are free to resolve their PDO bindings, however they see fit :)

driesvints:
[...]

PR #41233, opened 2022-02-24
[9.x] Add resource binding

aedart:
Adds support for binding resources values to PDO statement, in `Connection`. Removed the duplicate code from `PostgresConnection` as it is now handled in base class.

See #41180 for additional details.


driesvints:
Thanks @aedart

PR #41236, opened 2022-02-24
Feature/d helper

zsoltjanes:
I think this alias would be helpful during development or debuging. If we have dd() why don’t we have d()?


driesvints:
Seems like something went wrong with branching

PR #41243, opened 2022-02-25
Became possible pass enum obj to default value

mlk9:
we have to pass 'string' default value in migration but this changes help to us be easiest use enum obj. for example : 

$table->string('status')->default(StatusEnum::Draft->value); **===>** $table->string('status')->default(StatusEnum::Draft);


driesvints:
Please see previous attempts.

PR #41247, opened 2022-02-25
[9.x] improve resolve method in cacheManger for early return.

hAz5:
"**resolve**" method of class **CacheManager** Refactored like as **LogManager** "**resolve**" method in early return pattern and remove some extra indent.


taylorotwell:
If this is not fixing a bug no plans to merge.

PR #41254, opened 2022-02-25
[9.x] Add option to filter out routes defined in vendor packages in `route:list` command

amiranagram:
This PR adds the ability for `route:list` command to filter out routes defined in vendor (third-party) packages.

```sh
php artisan route:list --except-vendor
```

IMHO, very useful option in projects that have tons of registered routes via third-party packages (Horizon, Nova, Debugbar, etc.).

Tests are missing because I can't figure out a way to mock routes so they show another path other than where I defined them (in the test class). So maybe someone can help me out.

Also let me know if `--except-vendor` is not a good option name for the command.

Cheers.


nunomaduro:
I like this feature - just not sure about the `--except-vendor` naming. Lets see what Taylor thinks.

amiranagram:
Agree @nunomaduro, open to suggestions.

dammy001:
@amiranagram how about --ignore-vendor or --skip-vendor

nunomaduro:
Nevermind. Agree that `--except-vendor` is the best naming.

amiranagram:
Alright. What do you think about tests? @nunomaduro 

>Tests are missing because I can't figure out a way to mock routes so they show another path other than where I defined them (in the test class). So maybe someone can help me out.
>P.S. Would mocking the whole RouteListCommand class be a good way to modify the desired behavior so I can add some tests?

PR #41256, opened 2022-02-25
[9.x] Add ability to push additional pipes onto a pipeline via `chain($pipes)`

stevebauman:
## Description

This PR adds the ability to push additional pipes onto a pipeline via a `chain()` method.

This allows us to create a pipeline instance and pass it around the application, appending pipes onto the chain, rather than having to define all of the pipes in a single place:

```php
$pipeline = new Pipeline();

$pipeline->chain(fn ($object, $next) => $next($object));
$pipeline->chain(fn ($object, $next) => $next($object));

$pipeline->send($object)->thenReturn();
```

Let me know if you would like anything adjusted. Thanks for your time! ❤️

### DIFF
--- a/src/Illuminate/Pipeline/Pipeline.php
+++ b/src/Illuminate/Pipeline/Pipeline.php
@@ -81,7 +81,7 @@ class Pipeline implements PipelineContract
      * @param  array|mixed  $pipes
      * @return $this
      */
-    public function chain($pipes)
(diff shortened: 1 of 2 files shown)


taylorotwell:
Any thoughts on naming this method `pipe` instead of `chain`?

stevebauman:
Yea, I'm cool with that name 👍

Issue #41261, opened 2022-02-26
Email notifications sent to anonymous users fail silently when using MailChannel FCQN


thejager:
- Laravel Version: 8.79.0
- PHP Version: 8.1.2
- Database Driver & Version: N/A

### Description:
When you try to send an email notification to an anonymous user, and use `MailChannel::class` instead of `'mail'` for the route, it will not send anything nor notify you of any issues.

### Steps To Reproduce:

Create a notification as such:

```php
class WelcomeNotification extends Notification implements ShouldQueue
{
    use Queueable;

    public function via(): array
    {
        return [MailChannel::class];
    }

    public function toMail(mixed $notifiable): MailMessage
    {
        // Omitted for brevity, not relevant
    }
}
```

Now sent the notification as such:

```php
(new AnonymousNotifiable())
    ->route(MailChannel::class, 'johan@example.com')
    ->notifyNow(new WelcomeNotification());
```

Intuitively to me, this should work. But nothing happens, it will "fail" silently. I only figured out what was wrong after diving into the whole notification system and found this:

```php
public function send($notifiable, Notification $notification)
{
    $message = $notification->toMail($notifiable);

    if (! $notifiable->routeNotificationFor('mail', $notification) &&
        ! $message instanceof Mailable) {
        return;
    }
    
    // ...
}
```

It just returns if there's no `'mail'` key defined in the `AnonymousNotifiable::$routes`. I feel like this function should:

1.  Check `routeNotificationFor(MailChannel::class)`
2. Throw an error instead of return if there is no `'mail'` or `MailChannel::class` route

or am I missing something? Apologies if this expected behavior.

derekmd:
As [documented](https://laravel.com/docs/8.x/notifications#on-demand-notifications), the `'mail'` alias should be used:

```php
(new AnonymousNotifiable)
    ->route('mail', 'johan@example.com')
    ->notifyNow(new WelcomeNotification);
```

You could also change `WelcomeNotification` from a `MailMessage` into `Mailable` and set the `route(MailChannel::class, ...)` recipient using this example: https://laravel.com/docs/8.x/notifications#mailables-and-on-demand-notifications

---

A PR on `MailChannel` could be submitted to make either work, adding a `MailChannel::class` fallback if no `'mail'` recipients are found:

```php
class MailChannel
{
    // ...

    /**
     * Get the notification routing information for the mail driver.
     *
     * @param  mixed  $notifiable
[...]

PR #41263, opened 2022-02-27
Allow MailChannel::class to be used for routing information in anonymous notification

dejagersh:
PR fixing issue #41261.

## Problem
Consider the following code:

```php
class InvoicePaid extends Notification
{
    public function via() {
        return [MailChannel::class];
    }
    
    // ...
}
```

```php
Notification::route(MailChannel::class, 'taylor@example.com')
            ->notify(new InvoicePaid($invoice));
```

This will not send anything nor notify anyone of it not sending anything

You have to use `'mail'` in `Notification::route()`:

```php
Notification::route('mail', 'taylor@example.com')
            ->notify(new InvoicePaid($invoice));
```

To me, this is very unintuitive. Why can I use `MailChannel::class` in `via()`:

```php
class TestMailNotification extends Notification
{
    {
        return [MailChannel::class];
    }
}
```

but then I can't use `MailChannel` in the following context?

```php
Notification::route(MailChannel::class, 'taylor@example.com')
            ->notify(new InvoicePaid($invoice));
```

## Solution
This PR adds support for `MailChannel::class` in `Notification::route(MailChannel::class, 'taylor@example.com');`. `MailChannel` will check `'mail'` first and as fallback it will check `MailChannel::class`. The implementation is taken from the suggested solution that Derek commented on my issue: https://github.com/laravel/framework/issues/41261#issuecomment-1052468893 Credits to him :)

### DIFF
--- a/src/Illuminate/Notifications/Channels/MailChannel.php
+++ b/src/Illuminate/Notifications/Channels/MailChannel.php
@@ -52,13 +52,12 @@ class MailChannel
     {
         $message = $notification->toMail($notifiable);
 
-        if (! $notifiable->routeNotificationFor('mail', $notification) &&
-            ! $message instanceof Mailable) {
-            return;
-        }
-
         if ($message instanceof Mailable) {
             return $message->send($this->mailer);
+        }
+
+        if (! $this->routeNotificationFor($notifiable, $notification)) {
+            return;
         }
 
         $this->mailer->mailer($message->mailer ?? null)->send(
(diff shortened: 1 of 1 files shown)


ankurk91:
So you changes will only allow to use `MailChannel::class`, if yes, what about other channels like
`SlackChannel::class`, `SMSChannel::class`

Can we make other channel work in the same way?

thejager:
Hi @ankurk91 thanks for your message.

I've thought about this, but `laravel/framework` does not contain these channels, so I can't make those changes in this PR.

ankurk91:
The `via()` method on a notification class can accept FCQN already, there must be a way to use FCQN for anonymous notifications too

```php
    public function via($notifiable): array
    {
        return [FCMChannel::class, SlackChannel::class, NexemoChannel::class,  'database', 'mail'];
    }
```

taylorotwell:
Right we just document strings for first-party channels and FQCN are used for third-party channels. If we are going to do this I agree we would probably want to support it across every first-party channel.

PR #41265, opened 2022-02-27
fix a bug in InteractWithInput.php

amirtavakolian:
in InteractWithInput.php we have a method called filled which is for determining if the request contains a non-empty value for an input item.

filled method uses isEmptyString method for doing the job but in filled method we send value of an input for isEmptyString 
not the name of input


derekmd:
The failing tests show it was behaving as expected. The variable `$value` is an array key; it wasn't named `$key` to avoid conflicting with the method argument.

PR #41267, opened 2022-02-27
[9.x] improve some test of CacheManager

hAz5:
the test should be like real Laravel in up working.
CacheManager expect Application Contract Not Array

```
    /**
    * Create a new Cache manager instance.
     *
     * @return void
     */
    public function __construct($app)
    {
        $this->app = $app;
    }
```


taylorotwell:

Thanks!

PR #41270, opened 2022-02-28
[9.x] Fix Conditional::when and Conditional::unless when called with invokable

bastien-phi:
This PR fixes the PHPStan analysis when Conditional::when and Conditional::unless are called with invokable object

### DIFF
--- a/src/Illuminate/Conditionable/Traits/Conditionable.php
+++ b/src/Illuminate/Conditionable/Traits/Conditionable.php
@@ -13,7 +13,7 @@ trait Conditionable
      * @template TWhenParameter
      * @template TWhenReturnType
      *
-     * @param  (Closure($this): TWhenParameter)|TWhenParameter  $value
+     * @param  (\Closure($this): TWhenParameter)|TWhenParameter  $value
      * @param  (callable($this, TWhenParameter): TWhenReturnType)|null  $callback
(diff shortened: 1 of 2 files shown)


src/Illuminate/Conditionable/Traits/Conditionable.php
```diff
@@ -13,7 +13,7 @@ trait Conditionable
      * @template TWhenParameter
      * @template TWhenReturnType
      *
-     * @param  (callable($this): TWhenParameter)|TWhenParameter  $value
+     * @param  (Closure($this): TWhenParameter)|TWhenParameter  $value
```
nunomaduro, line 16: Can you use FQFN?

src/Illuminate/Conditionable/Traits/Conditionable.php
```diff
@@ -41,7 +41,7 @@ public function when($value, callable $callback = null, callable $default = null
      * @template TUnlessParameter
      * @template TUnlessReturnType
      *
-     * @param  (callable($this): TUnlessParameter)|TUnlessParameter  $value
+     * @param  (Closure($this): TUnlessParameter)|TUnlessParameter  $value
```
nunomaduro, line 44: Can you use FQFN?

nunomaduro (changes requested):

PR #41276, opened 2022-02-28
[9.x] Remove `static` return types

browner12:
I would definitely like a second set of eyes on this, but I'm pretty sure these methods will only return instances of the Eloquent or Query Builder.  It was a little difficult to find how these methods are called, but it's in the `\Illuminate\Database\Eloquent\SoftDeletingScope`.

The reason I came across this was that I was trying to select trashed Eloquent models on a specific database connection.

```php
$user = User::withTrashed()->on($connection);
```

This doesn't actually work, but the IDE thinks it's okay, because it sees `withTrashed()` returning `static`, which is `\App\User` which extends the `\Illuminate\Database\Eloquent\Model`, which *does* have the `on()` method.

In reality, it returns the `\Illuminate\Database\Eloquent\Builder` which **does not** have the `on()` method.

### DIFF
--- a/src/Illuminate/Database/Eloquent/SoftDeletes.php
+++ b/src/Illuminate/Database/Eloquent/SoftDeletes.php
@@ -3,9 +3,9 @@
 namespace Illuminate\Database\Eloquent;
 
 /**
- * @method static static|\Illuminate\Database\Eloquent\Builder|\Illuminate\Database\Query\Builder withTrashed(bool $withTrashed = true)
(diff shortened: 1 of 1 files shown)


browner12:
originally from commit https://github.com/laravel/framework/commit/7b91285cf7c100b07bf3e277bc0acb3e0598b7aa

browner12:
On a related note, any ideas on how to get all models, including trashed, on a specific connection?

PR #41280, opened 2022-03-01
[9.x] Update resolveFacadeInstance method docblock

imanghafoori1:
Since the $app variable is documented to be an instance of the Application class, this method is definitely a throwish one in case of a missing key on the container.

I think the public `getFacadeRoot` also needs this.


taylorotwell:
No plans to add this.

PR #41282, opened 2022-03-01
[9.x] Fix implicitBinding and withTrashed route with child with no SoftDeletes trait

stein-j:
This fixes an issue caused when using _implicitBinding_ and _withTrashed_ on a route and where the parent model **has** the SoftDelete trait and the child **doesn't have** the SoftDelete trait.

Ex:

```php
class User extends Model
{
    use SoftDeletes;

    public function posts()
    {
        return $this->hasMany(Post::class);
    }

}

class Posts extends Model
{
    //
}

Route::scopeBindings()->get('users/{user}/posts/{post}', /*...*/)->withTrashed();
```

The following route call `/users/1/posts/2` will throw this error `Call to undefined method Illuminate\\Database\\Eloquent\\Relations\\HasMany::withTrashed()`.

____

This PR adds an extra check to verify if the child model instance has the SoftDeletes trait.

### DIFF
--- a/src/Illuminate/Routing/ImplicitRouteBinding.php
+++ b/src/Illuminate/Routing/ImplicitRouteBinding.php
@@ -47,7 +47,7 @@ class ImplicitRouteBinding
                         : 'resolveRouteBinding';
 
             if ($parent instanceof UrlRoutable && ($route->enforcesScopedBindings() || array_key_exists($parameterName, $route->bindingFields()))) {
-                $childRouteBindingMethod = $route->allowsTrashedBindings()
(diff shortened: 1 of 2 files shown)


BrandonSurowiec:
@taylorotwell Can this fix be backported to 8.x? The bug is present there as well.

driesvints:
@BrandonSurowiec I think we can. Can you try a PR?

BrandonSurowiec:
@driesvints Sure. Just won't get to it until the weekend. There are probably a few other PRs submitted to 9.x like the recent `Http::retry()` fix that could also be backported as a bug fix.

driesvints:
@BrandonSurowiec Nothing is urgent : )

PR #41284, opened 2022-03-01
[10.x] Add ability to purge after resolved callbacks

hAz5:
sometimes we need to purge all callbacks that are chained on resolving an abstract.

like we need to resolve a FormRequest to do/check something but we see afterResolve method chained on ValidatesWhenResolved Trait in FormRequestServiceProvider.

I know we can replace our Custom Application class with Laravel Application Class and our functions but it seemed to me maybe it is useful.


taylorotwell:

PR #41286, opened 2022-03-01
[9.x] Add PHPStan-specific tags to improve analysis on model factories

devfrey:
Nuno made model factories generic in [this PR](https://github.com/laravel/framework/pull/39169). To preserve compatibility with PhpStorm the return types were defined as a union of `\Illuminate\Database\Eloquent\Model|TModel`. [PHPStan normalises this](https://phpstan.org/developing-extensions/type-system#type-normalization) into `\Illuminate\Database\Eloquent\Model` (because `TModel` is a subtype of `Model`), resulting in a bunch of unknown property/method errors when running PHPStan on tests that use model factories.

This PR adds specific tags for PHPStan that overrule the standard PHPDoc tags, to preserve the specific model types provided by the generics. PhpStorm's understanding of return types is unaffected (because it ignores these PHPStan-specific tags), but PHPStan gains a lot of benefits from these specific tags as can be seen in the updated tests.


nunomaduro:
In the future, we will be replacing `\Illuminate\Database\Eloquent\Model` by `TModel`. But we need to wait for IDEs to follow up. Only PHPStorm seems to be following.

And, we don't plan to add specific phpstan annotations for now.

devfrey:
Thanks for your reply @nunomaduro.

As an alternative solution I'd like to suggest to remove `Model` from the unions completely. You mention wanting to wait for IDE support, but Eloquent collections don't have this 'fallback' either ([example here](https://github.com/laravel/framework/pull/38538/files#diff-0807e2b2cd561231f1b2d07add1347003d4deebd2e098c041f78a588746c3b23R28)) and I couldn't find any complaints. In its current form the generic annotations are useless to PHPStan – which I think is a missed opportunity.

I'd be happy to send an updated PR.

PR #41289, opened 2022-03-02
[10.x] Use Guzzle to resolve the URL against the base URL

JaZo:
## Problem
I've created an HTTP macro to easily work with an API, just like the [documentation suggests](https://laravel.com/docs/9.x/http-client#macros). The response of this API is paginated and the URL of the next page is returned in the response. This URL is absolute and making a request with a base URL and an absolute URL currently results in requests like `https://api.github.com/https://api.github.com/organizations/laravel`. Guzzle allows you to set a [base URL](https://docs.guzzlephp.org/en/stable/quickstart.html) and use absolute URL's in conjuction, so I'd expect this client to support that too.

## Solution
Guzzle has quite some [logic](https://github.com/guzzle/psr7/blob/c0dcda9f54d145bd4d062a6d15f54931a67732f9/src/UriResolver.php#L58) to resolve the URL against the base URL, so I've chosen to use the same [logic](https://github.com/guzzle/guzzle/blob/master/src/Client.php#L213). If you prefer not to use an external dependency and would like to use some other logic, please let me know!

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -8,6 +8,7 @@ use GuzzleHttp\Exception\ConnectException;
 use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Exception\TransferException;
 use GuzzleHttp\HandlerStack;
+use GuzzleHttp\Psr7;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Http\Client\Events\ConnectionFailed;
 use Illuminate\Http\Client\Events\RequestSending;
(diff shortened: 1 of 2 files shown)


JaZo:
I've converted it to draft as I've found some breaking changes;

baseUrl | Url | Current result | New result
-- | -- | -- | --
http://foo.com | /bar | http://foo.com/bar | http://foo.com/bar
http://foo.com/foo | /bar | http://foo.com/foo/bar | http://foo.com/bar
http://foo.com/foo | bar | http://foo.com/foo/bar | http://foo.com/bar
http://foo.com/foo/ | bar | http://foo.com/foo/bar | http://foo.com/foo/bar
http://foo.com | http://baz.com | http://foo.com/http://baz.com | http://baz.com
http://foo.com/?bar | bar | http://foo.com/?bar/bar | http://foo.com/bar

Guzzle follows the rules described in [RFC 3986](https://tools.ietf.org/html/rfc3986#section-5.2), but as seen above that will be a breaking change. We can focus on fixing the issue with absolute URL's now and move this change to 10.x?

JaZo:
Retargeted to 10.x and created #41307 as a fix for the base URL + absolute URL issue.

PR #41296, opened 2022-03-02
[9.x] Allow eloquent whereNot() and orWhereNot() to work on column and value

shadoWalker89:
Hi,

This PR https://github.com/laravel/framework/pull/41096 added `whereNot()` and `orWhereNot()` to the eloquent query builder and query builder.

The PR allows to apply a where not on a group of wheres but does not allow to do something like this

```php
$query->whereNot('role', 'admin'); // WHERE NOT role = 'admin'
``` 

This PR adds that (for both whereNot() and orWhereNot() ) while it also keeps the ability to use a where not on grouped wheres as well.

It will keep the same method signature developers are used to when using `where()` and `orWhere()` by allowing to specify and `$operator` and `$column` variables

If this PR is merged now, it will be a good thing since the Original PR was not yet tagged in any new release.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -317,26 +317,30 @@ class Builder implements BuilderContract
     }
 
     /**
-     * Add a "where not" clause to the query.
+     * Add a basic "where not" clause to the query.
      *
-     * @param  \Closure  $callback
+     * @param  \Closure|string|array|\Illuminate\Database\Query\Expression  $column
+     * @param  mixed  $operator
+     * @param  mixed  $value
      * @param  string  $boolean
      * @return $this
      */
-    public function whereNot(Closure $callback, $boolean = 'and')
+    public function whereNot($column, $operator = null, $value = null, $boolean = 'and')
     {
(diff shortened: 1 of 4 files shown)


taylorotwell:
Ah - didn't see this until I tagged - but still probably not a big deal to merge it into 9.x since it just came out.

deleugpn:
Is `whereNot('role', 'admin')` the same as `where('role', '<>', 'admin')`? I really like that `whereNot`

shadoWalker89:
@deleugpn it will use the NOT keyword to negate the condition. See the code comment in the code example in the PR description

taylorotwell:
How does this address: https://github.com/laravel/framework/pull/41096#issuecomment-1055700273

shadoWalker89:
> How does this address: [#41096 (comment)](https://github.com/laravel/framework/pull/41096#issuecomment-1055700273)

As mentioned in the MySQL documentation https://dev.mysql.com/doc/refman/8.0/en/operator-precedence.html the comparison operators have higher precedence then the NOT keyword which means we don't need to wrap `a {operator} b` within parentheses. 

This is a very simple example to demonstrate that.

https://www.db-fiddle.com/f/hZTgE1x2hyT1ksx4CNDqnM/2

PR #41301, opened 2022-03-02
Add more inspiring quotes to celebrate Origins above 11,333 ft.

onlime:
> Life is like a cigarette, smoke it to the butt. - Harvie Krumpet
> Life is now. There was never a time when your life was not now, nor will there ever be. - Eckhart Tolle
> Here I sit broken hearted, came to shit but only farted. Then one day I took a chance, went to fart, but shit my pants. - William Shakespeare
> The primary cause of unhappiness is never the situation, but your thoughts about it. - Eckhart Tolle

... my all-time favorites, quotes that guide me through life, and let it live alive!

Cheers, Pipo


taylorotwell:
Hmm - not sure about a couple of these 😅

PR #41303, opened 2022-03-03
[9.x] Update changelog

demiurge-ash:
Update changelog for v9.3.0


driesvints:
Thanks!

Also, an easter egg isn't really an easter egg if it's revealed 😉

PR #41304, opened 2022-03-03
[9.x] Change how Laravel handles strict morph map with pivot classes

crynobone:
* Generic Pivot class should always be ignored as it doesn't have any knowledge of the exact usage and is generated automatically by Laravel.
* Custom Pivot class should work the same as a normal model class since it is defined manually by the developer.

In Laravel Nova, when we attach, update attached, and detach resources we
would store the pivot model in `action_events` table and this can cause
issue when the developer attempt to use `Relation::requireMorphMap()`.

### DIFF
--- a/tests/Database/DatabaseEloquentStrictMorphsTest.php
+++ b/tests/Database/DatabaseEloquentStrictMorphsTest.php
@@ -4,6 +4,7 @@ namespace Illuminate\Tests\Database;
 
 use Illuminate\Database\ClassMorphViolationException;
 use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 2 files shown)


tests/Database/DatabaseEloquentStrictMorphsTest.php
```diff
@@ -39,7 +40,7 @@ public function testStrictModeDoesNotThrowExceptionWhenMorphMap()
 
     public function testMapsCanBeEnforcedInOneMethod()
     {
-        $model = TestModel::make();
+        $model = new TestModel;
```
crynobone, line 43: faced issue while running test on single file:

image

Change to `new TestModel` fixed that.

PR #40902, opened 2022-02-09
[9.x] Improve types on model factory methods

axlon:
This fixes the following PHPStan error found in the application boilerplate (tested at level 8):

```
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   database/factories/UserFactory.php                                                                                                                                                       
 ------ ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- 
  36     Parameter #1 $state of method Illuminate\Database\Eloquent\Factories\Factory<App\Models\User>::state() expects array<string, mixed>|(callable(): array<string, mixed>), Closure(array):  
         array{email_verified_at: null} given.                                                                                                                                                    
 ------ -----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

I also updated the types for `create`, `createOne`, `createQuietly`, `make`, `makeOne`, `raw` and `new`. I wanted to update `lazy` as well but its currently hardcoded as an array

PR #41305, opened 2022-03-03
[9.x] Add extra type checks for model factories

axlon:
This PR is a followup to #40902, previously some type assertions were not possible due to a limitation in PHPStan, this has now been fixed (since PHPStan 1.4.7, in phpstan/phpstan-src#1019) so these assertions can be added in.


driesvints:
@axlon can you also bump PHPStan to this new minimum?

axlon:
@driesvints done!

driesvints:
Thanks 👍

Issue #41290, opened 2022-03-02
whereIn() returns 0 results with 1000 values or more with mariadb 10.3.34


sventendo:
- Laravel Version: v8.83.0
- PHP Version: 8.0
- Database Driver & Version: mariadb 10.3.34

### Description:

With the latest version of mariadb there seems to be an issue with the underlying `WHERE IN()` optimization when the clause has many, many values. The default threshold is 1000+ values. This can be changed with the system variable `in_predicate_conversion_threshold`:
https://mariadb.com/docs/reference/mdb/system-variables/in_predicate_conversion_threshold/

### Steps To Reproduce:

Works:
- use mariadb 10.3.32
- run tinker like `\DB::table('foo')->whereIn('bar', $arrayOf999Values)->count();`
- return 1234 results

Works:
- use mariadb 10.3.34
- run tinker like `\DB::table('foo')->whereIn('bar', $arrayOf999Values)->count();`
- return 1234 results

Works:
- use mariadb 10.3.32
- run tinker like `\DB::table('foo')->whereIn('bar', $arrayOf1000Values)->count();`
- return 1234 results

Doesn't work:
- use mariadb 10.3.34
- run tinker like `\DB::table('foo')->whereIn('bar', $arrayOf1000Values)->count();`
- return 0 results

Is this something that can be fixed on the Laravel side?
I tried implementing a global scope to split the "whereIn" into a chunked "where(whereIn($chunk1)->orWhereIn($chunk2)->orWhereIn(...))" statement but this only works well for arrays of values. If the parameter is another query instance it gets a lot more complex.

driesvints:
This seems like a problem with MariaDB itself? Most likely you'll need to set that system variable higher? Not sure if there's anything we can do here ourselves...

RoodFruit:
@sventendo I just had the same problem and was about to create a ticket. It seems that `addBinding()` is the culprit in `Illuminate/Database/Query/Builder.php`:

```
public function whereIn($column, $values, $boolean = 'and', $not = false)
{
    ...

    $this->addBinding($this->cleanBindings($values), 'where'); 

    return $this;
}
```

You can bypass using the bindings with `whereIntegerInRaw()` or `whereIntegerNotInRaw()`, in case your dataset are integers.

@driesvints why does Laravel only has `whereIntegerInRaw()`? I would think a `whereInRaw()` or `whereNotInRaw()` is pretty nice as well. What do you think?

sventendo:
[...]

PR #41306, opened 2022-03-03
[9.x] Added whereInRaw() and whereNotInRaw()

RoodFruit:
Currently we have a `whereIntegerInRaw()` and a `whereIntegerNotInRaw()`. These functions do not apply PDO bindings, which make the use of these raw queries really fast. The downside is that it's for integers only (it casts every value to an int).

Due to a bug in MariaDB (see https://jira.mariadb.org/browse/MDEV-27937), where `whereIn()` cannot have more than a 999 bindings or it will return 0 records (see https://github.com/laravel/framework/issues/41290), I found the Laravel function `whereIntegerInRaw()`. Unfortunately, my array has strings. So I came up with these `whereInRaw()` and `whereNotInRaw()` functions.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1064,6 +1064,41 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Add a "where in raw" clause to the query.
+     *
+     * @param  string  $column
+     * @param  \Illuminate\Contracts\Support\Arrayable|array  $values
+     * @param  string  $boolean
+     * @param  bool  $not
+     * @return $this
+     */
+    public function whereInRaw($column, $values, $boolean = 'and', $not = false)
+    {
+        $type = $not ? 'NotInRaw' : 'InRaw';
+
+        if ($values instanceof Arrayable) {
+            $values = $values->toArray();
+        }
+
+        $this->wheres[] = compact('type', 'column', 'values', 'boolean');
+
+        return $this;
+    }
+
+    /**
+     * Add a "where not in raw" clause to the query.
+     *
+     * @param  string  $column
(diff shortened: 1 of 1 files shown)


driesvints:
@RoodFruit please add a description so taylor knows why this PR is needed.

RoodFruit:
> @RoodFruit please add a description so taylor knows why this PR is needed.

Note the "See #41290". But I'll put it here as well.

@taylorotwell Currently we have a `whereIntegerInRaw()` and a `whereIntegerNotInRaw()`. These functions do not apply PDO bindings, which make the use of these raw queries really fast. The downside is that it's for integers only (it casts every value to an int). 

Due to a bug in MariaDB (see https://jira.mariadb.org/browse/MDEV-27937), where `whereIn()` cannot have more than a 999 bindings or it will return 0 records (see https://github.com/laravel/framework/issues/41290), I found the Laravel function `whereIntegerInRaw()`. Unfortunately, my array are strings. So I came up with these `whereInRaw()` and `whereNotInRaw()` functions.

driesvints:
> Note the "See https://github.com/laravel/framework/issues/41290".

Please don't link to an issue and provide the description in the main PR comment. That saves Taylor from having to dig through comments to find the explanation. We handle dozens of PR's daily and it's just not feasible for us to dig through issues for every PR.

RoodFruit:
> > Note the "See #41290".
> 
> Please don't link to an issue and provide the description in the main PR comment. That saves Taylor from having to dig through comments to find the explanation. We handle dozens of PR's daily and it's just not feasible for us to dig through issues for every PR.

Fair enough.

taylorotwell:

Thanks!

driesvints:
@RoodFruit if you need this, it's best that you macro it in. Tbvh, I think the real issue is that you shouldn't use so many values for that method.

PR #41315, opened 2022-03-03
[9.x] Fix MySqlSchemaState does not add --ssl-ca to mysql cli 

DeepDiver1975:
…gured

fixes #41300

PR #41317, opened 2022-03-03
[9.x] Fix MySqlSchemaState does not add --ssl-ca to mysql cli + tests

DeepDiver1975:
https://github.com/laravel/framework/pull/41315 + tests


DeepDiver1975:
sure - wanted to double check how ci behaves without influencing the original pr at first

PR #41320, opened 2022-03-03
[9.x] Support modifying a char column type

hafezdivandari:
This PR provides support for modifying a `char` column type.

```php
    $table->char('name', 50)->nullable()->change();
});
```

Before this PR, modifying a `char` column led to an error:

> `Unknown column type \"char\" requested. Any Doctrine type that you use has to be registered with \\Doctrine\\DBAL\\Types\\Type::addType(). You can get a list of all the known types with \\Doctrine\\DBAL\\Types\\Type::getTypesMap(). If this error occurs during database introspection then you might have forgotten to register all database types for a Doctrine Type. Use AbstractPlatform#registerDoctrineTypeMapping() or have your custom types implement Type#getMappedDatabaseTypes(). If the type name is empty you might have a problem with the cache or forgot some mapping information`

But `doctrine/dbal` package actually supports modifying `char` column types as [`StringType::class`](https://github.com/doctrine/dbal/blob/e839cecc812fd04ec6c39fc865302b6a322d967d/src/Types/StringType.php) by setting `fixed` option to `true`.

So this PR, maps Laravel `char` to its Doctrine equivalent `string` type and set `fixed` option to `true` that finally [gets the SQL snippet to declare a CHAR column](https://github.com/doctrine/dbal/blob/e839cecc812fd04ec6c39fc865302b6a322d967d/src/Platforms/AbstractMySQLPlatform.php#L199).

### DIFF
--- a/src/Illuminate/Database/Schema/Grammars/ChangeColumn.php
+++ b/src/Illuminate/Database/Schema/Grammars/ChangeColumn.php
@@ -125,6 +125,10 @@ class ChangeColumn
             $options['length'] = static::calculateDoctrineTextLength($fluent['type']);
         }
 
+        if ($fluent['type'] === 'char') {
+            $options['fixed'] = true;
+        }
+
         if (static::doesntNeedCharacterOptions($fluent['type'])) {
             $options['customSchemaOptions'] = [
                 'collation' => '',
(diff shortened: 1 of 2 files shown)


hafezdivandari:
@taylorotwell sure, I updated my comment above.

PR #41321, opened 2022-03-03
[9.x] Add temporary file helpers

inxilpro:
This adds three new functions to the `Filesystem` class:

- `temporaryFilename()` — returns the path to a temporary file that will be removed when the script ends
- `putTemporary()` — write contents to a temporary file that will be removed when the script ends
- `closeOpenTemporaryFiles()` — a static method to close all open temporary file handles (useful for testing and probably should be added to Octane if this PR is merged)

### DIFF
--- a/src/Illuminate/Filesystem/Filesystem.php
+++ b/src/Illuminate/Filesystem/Filesystem.php
@@ -18,6 +18,25 @@ class Filesystem
     use Macroable;
 
     /**
+     * The currently open temporary files.
+     *
+     * @var array
+     */
+    protected static $openTemporaryFiles = [];
+
+    /**
+     * Close all teh currently open temporary files.
+     *
+     * @return void
+     */
+    public static function closeOpenTemporaryFiles()
+    {
+        while ($pointer = array_pop(self::$openTemporaryFiles)) {
+            fclose($pointer);
+        }
+    }
+
+    /**
      * Determine if a file or directory exists.
      *
      * @param  string  $path
@@ -746,7 +765,11 @@ class Filesystem
      */
     public function temporaryFilename()
     {
-        return stream_get_meta_data(tmpfile())['uri'];
+        $handle = tmpfile();
+
+        self::$openTemporaryFiles[] = $handle;
+
+        return stream_get_meta_data($handle)['uri'];
     }
 
     /**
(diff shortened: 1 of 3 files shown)


taylorotwell:
What do you see as the primary use case for this functionality?

inxilpro:
I just searched our codebase and here are a few examples:

- Building a very large CSV file that will eventually be attached to an email or uploaded to S3
- Downloading images from an old CDN to migrate them to a new media library solution
- Testing an artisan command that accepts a file as an input

It's also cumbersome, because either you have to use `sys_get_temp_dir()` combined with `tempnam()` and then do your own cleanup, or you need to know about using `tmpfile()` with `stream_get_meta_data()` (which is a pretty unintuitive combination of functions).

Honestly, every time I find myself writing `sys_get_temp_dir()` I think, ”Wait, there's probably a Laravel method to do this for me.” And then I'm surprised when there isn't. So I thought I'd add one!

inxilpro:
> using laravel scheduler each day I delete old files from that disk

Unfortunately, that's not something you can rely on at a package level. For example, `spatie/image` needs to handle temporary file creation and cleanup when performing image manipulations.

taylorotwell:
One thing I don't love about this is the static variable of current temporary files that need to be cleaned up, etc. I wonder if something like this would satisfy your use case:

```php
$filesystem->withTemporaryFile(function ($path) {
    // Interact with path... temporary file automatically deleted after callable execution...
});

$filesystem->withTemporaryFile($contents, function ($path) {
    // Interact with path (contents already inserted)... temporary file automatically deleted after callable execution...
});

$filesystem->withTemporaryFiles(2, function ($pathOne, $pathTwo) {
    // Interact with paths... files deleted after execution...
});
```

inxilpro:
@taylorotwell I agree, the cleanup stuff is a bummer. I did consider a Closure-based solution, but it honestly doesn't feel nearly as nice to work with because usually the context for these types of calls aren't particularly well suited to running inside a callback.

I'd be OK with removing the resource tracking code (which really is only there to ensure Laravel Octane support) and just implement `temporaryFilename` and `putTemporary`. Either that, or I can just keep this in userland and close the PR.

taylorotwell:
I think I'll go ahead and hold off on this one then and let you do it in user-land.

PR #41322, opened 2022-03-03
[8.x] Fixes getting the trusted proxies IPs from the configuration file

nunomaduro:
This pull request allows getting the trusted proxies IPs from the configuration file, like back in the days when we had Fideloper's package.


amenk:
@nunomaduro Did you consider re-adding a default `env('TRUSTED_PROXIES')` ? This would be helpful in situations where Laravel is hosted in Docker for example and would not required to change the default installation anymore.

We could as well make a pull request if @taylorotwell sees changes to get that accepted?

See also https://github.com/laravel/framework/discussions/38974

PR #41323, opened 2022-03-03
[9.x] Test FilesystemAdapter::put() exceptions in Windows environment

derekmd:
Follow-up to https://github.com/laravel/framework/pull/41308

Instead of attempting a write to a read-only directory to force a `UnableToWriteFile` exception only in Unix filesystems, target a read-only file which is also supported on Windows OS.


driesvints:
Thanks @derekmd!

PR #41325, opened 2022-03-03
[9.x] Remove constructor from console command stub

browner12:
I don't believe there's really any reason a constructor would be used in 99% of console commands. Any dependencies can be injected directly into the `handle()` method, and if we drop the constructor, the parent constructor will be called automatically.

I know we can generate our own stubs, but I believe this applies to the large majority of users. Looking through some of my projects, not a single console command does anything in the constructor.


nunomaduro (approved):
Agree

driesvints (approved):

PR #41328, opened 2022-03-03
[9.x] Set SES MessageId to the sent message ID

leandro-hermes:
This PR sets the given SES message ID as the `SentMessage`'s message ID.

As mentioned in #40696, the addition of Message ID's to the headers of the just-sent message was removed because Symfony is generating one. But the actual SES message ID may be needed, in my case, to manage the status of the sent e-mail, that may be afterwards sent by the SES service, referencing the previously given ID.

### DIFF
--- a/src/Illuminate/Mail/Transport/SesTransport.php
+++ b/src/Illuminate/Mail/Transport/SesTransport.php
@@ -45,7 +45,7 @@ class SesTransport extends AbstractTransport
     protected function doSend(SentMessage $message): void
     {
         try {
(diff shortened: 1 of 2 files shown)


taylorotwell:
Would this be a breaking change for existing applications? Why or why not?

leandro-hermes:
After you asked me, I dug deeper and realized that this behavior is actually wrong, because the generated ID has already sent in the message headers. I'll try to find another approach to attach the returned ID. Sorry.

Issue #37961, opened 2021-07-09
Eloquent dirty check for collections


tpetry:
- Laravel Version: 8.x
- PHP Version: all

### Description:
Currently, the eloquent model dirty checks are not working correctly for collection attributes if a value changes between `boolean` and `string`.

### Steps To Reproduce:
```php
use Illuminate\Database\Eloquent\Model;

class Test extends Model
{
    protected $casts = [
        'collection' => 'collection',
    ];
    protected $fillable = [
        'collection',
    ];
}

$model = new Test(['collection' => ['key' => true]]);
$model->syncOriginal();
$model->fill(['collection' => ['key' => 'value']]);

$model->isDirty(); // false but should be true
```

### Reason for this behaviour:
The attributes are casted within the `HasAttributes` trait to a collection and checked with php's equality logic:
1. `$this->castAttribute($key, $attribute) == $this->castAttribute($key, $original);`
2. When using the comparison operator (==), object variables are compared in a simple manner, namely: Two object instances are equal if they have the same attributes and values (values are compared with ==), and are instances of the same class. [[PHP: Comparing Objects]](https://www.php.net/manual/de/language.oop5.object-comparison.php)
3. $a == $b | Equality | true if $a and $b have the same key/value pairs. [[PHP: Array Operators]](https://www.php.net/manual/en/language.operators.array.php)

So, the values are checked with PHP's simple equality check, which is now showing the typical unsafe type checks with strings. In my opinion, the model should be marked as dirty. Depending on the opinion whether this is a bug or expected behavior, I would propose doing strict dirty checks for next major release.

driesvints:
Hey @tpetry, this is more of a feature request I think. Feel free to attempt a PR to see if Taylor would accept it 👍

tpetry:
I wouldn't call it a feature request if my models at the moment don't save their updated values to the database because of the actual behavior of the dirty check algorithm.

driesvints:
@tpetry we could maybe add another `strict-collection` besides the current one. Please try a PR if you're willing. If it's not accepted you can always use if in your own code base.

tpetry:
The same problem is happening for `array`, `object`, maybe `AsArrayObject::class` and `AsCollection ::class`. @taylorotwell Is this expected behavior, or a bug? If it's a bug, i can write a PR for `8.x` or `9.x` if it is declared a breaking change.

krnsptr:
I think this is a bug. I had unknown bug for years because of this.

[...]

PR #41337, opened 2022-03-04
[9.x] Loose comparison causes the value not to be saved

JBtje:
The issue has been addressed in #37961.

Given an attribute with the cast `object` or `collection`, the `$model->isDirty()` method returns the wrong value due to the use of loose comparison.

Example from issue #37961:
```php

class Test extends Model
{
    protected $casts = [
        'collection' => 'collection',
    ];
    protected $fillable = [
        'collection',
    ];
}

$model = new Test(['collection' => ['key' => true]]);
$model->syncOriginal();
$model->fill(['collection' => ['key' => 'value']]);

$model->isDirty(); // false but should be true

// or
$model = new Test(['collection' => ['key' => null]]);
$model->syncOriginal();
$model->fill(['collection' => ['key' => false]]);

$model->isDirty(); // false but should be true
```

The solution proposed by @krnsptr [link](https://github.com/laravel/framework/issues/37961#issuecomment-1017184911) is in my opinion the best solution.

Furthermore, casting the attributes to object/collection prior to checking the values, makes no sense to me. We are interested in knowing if the content differs, the cast prior to this check does not provide useful additional information (only causes that you can't use strict comparison due to object being always different even with the same content). The cast also adds overhead.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1939,8 +1939,8 @@ trait HasAttributes
             return $this->fromDateTime($attribute) ===
                 $this->fromDateTime($original);
         } elseif ($this->hasCast($key, ['object', 'collection'])) {
-            return $this->castAttribute($key, $attribute) ==
-                $this->castAttribute($key, $original);
+            return $this->fromJson($attribute) ===
+                $this->fromJson($original);
         } elseif ($this->hasCast($key, ['real', 'float', 'double'])) {
             if ($original === null) {
                 return false;


taylorotwell:
Here is what I think people will complain about here, and I believe this has been attempted in the past. An array with the same key / value pairs but the key / value pairs are in a different "order" is considered equal when using `==`, but not equal when using `===`:

CleanShot 2022-03-04 at 08 54 56@2x

JBtje:
Personally, I stand with PHP on this, and think a change in order means that the value is no longer the same and should be updated.

Implementing it as is, the worst-case scenario would be an extra `UPDATE` and the `updated_at` attribute obtains a new timestamp due to the order change. Though, this can easily be solved by keeping the same order in the array… whereas the issue at hand only has less nice solutions.

If the order thing is a big deal, it would be possible to add a recursive `ksort()` in the check if the values don't match using strict comparison.

taylorotwell:
I agree that the current situation seems a little worse than the change in behavior introduced by this PR. Will think about it over the weekend.

iamgergo:
I've also met with this problem. Could it be a solution to use `Arr::dot()` and keep loose comparison at the same time? In that case nested array will be flattened and it should work, I guess.

```php
Arr::dot($this->fromJson($attribute)) == Arr::dot($this->fromJson($original));
```

JBtje:
> I've also met with this problem. Could it be a solution to use Arr::dot() and keep loose comparison at the same time? In that case nested array will be flattened and it should work, I guess.

`Arr::dot` does not solve the problem at hand.

```php
$attribute = [
  'b' => [
    'c' => 1
  ]
];
$original = [
  'b' => [
    'c' => true
  ]
];

Arr::dot($attribute) == Arr::dot($original);  // Returns true, should be false.
```

Since loose comparison is the cause of the problem

iamgergo:
I see. Another solution that crossed my mind:

```php
$a = Arr::dot($attrubite);
$o = Arr:.dot($original);

count($a) === count($o) && empty(array_diff_assoc($a, $o));
```

Note, comparing the count of the array elements is necessary, because let's say if `$a` is just a subset of `$o` (because some items were removed, but others were not changed), the `array_diff_assoc` will return an empty array.

What do you think about that solution?

JBtje:
> Okay, I see what do you mean. This solution won't work because `1` will be handled as `true` while it should not be.

Yup, same for all other "loose" vs "strict" comparisons differences https://www.php.net/manual/en/types.comparisons.php

DarkGhostHunter:
Can this be fixed by retrieving the items as an array, using `ksort()`, and then compare strictly?

JBtje:
> Can this be fixed by retrieving the items as an array, using `ksort()`, and then compare strictly?

Yes that would be possible (when using recursive `ksort()`). But it would also give extra overhead and the behaviour would deviate from [L1957-L1958](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1957-L1958).

tpetry:
I had to resolve this problem in my application too by patching the dirty logic. 

> Here is what I think people will complain about here, and I believe this has been attempted in the past. An array with the same key / value pairs but the key / value pairs are in a different "order" is considered equal when using `==`, but not equal when using `===`:
> 
> CleanShot 2022-03-04 at 08 54 56@2x
For PHP a different ordering will result in a `===` check being false, but the database acknowledges that these two JSON objects are the same despite it's ordering:
```sql
select CAST('{"a": 1, "b":2}' as JSON) = CAST('{"b": 2, "a":1}' as JSON)
```

I've solved the problem in a backwards-compatible way by recursively sorting a collection/json array when `array_is_list` is false, which means the resulting variable will be a json object in the hierarchy.

DarkGhostHunter:
May be this will need a flag or sort in the Model:

```php
protected $strictComparison = ['my-array'].
```

This way users can opt-in when dealing with comparing model attributes.

MasterRO94:
This PR breaks BC. If original $value is `null` then `PHP Deprecated:  json_decode(): Passing null to parameter #1 ($json) of type string is deprecated` is thrown

PR #41339, opened 2022-03-04
[9.x] Fix collection typehint templates

netpok:
The diffUsing methods have invalid params defined, this PR fixes it.


nunomaduro (changes requested):
Can you add / update the tests under the `types` directory?

netpok:
Done

PR #41340, opened 2022-03-04
[9.x] Adds partial queue faking

yoeriboven:
This PR allows you to only fake certain jobs.

Useful if you want to test if one job dispatches another job.

```php
<?php

namespace App\Jobs;

use ...

class ParentJob
{
    public function handle() 
    {
        ChildJob::dispatch();
    }
{
```

```php
/** @test */
public function it_dispatches_another_job()
{
    Queue::fake(ChildJob::class);

    ParentJob::dispatch();

    Queue::assertPushed(ChildJob::class);
}
```

When using `Queue::fake()` all jobs will be faked. Therefore the `ParentJob` won't run and will never dispatch the `ChildJob`.

With this addition you can choose which job to fake. Not passing a parameter will fake all jobs like before.

**Example:** I run a web scraper which scrapes the homepage of a news site in a job. That job will then dispatch a job for every article it found.

### DIFF
--- a/src/Illuminate/Support/Testing/Fakes/QueueFake.php
+++ b/src/Illuminate/Support/Testing/Fakes/QueueFake.php
@@ -415,7 +415,7 @@ class QueueFake extends QueueManager implements Queue
             return true;
         }
 
-        return $this->jobsToFake->contains(function ($jobToFake) use ($job){
+        return $this->jobsToFake->contains(function ($jobToFake) use ($job) {
             return get_class($job) === $jobToFake;
         });
     }
--- a/src/Illuminate/Support/Testing/Fakes/QueueFake.php
+++ b/src/Illuminate/Support/Testing/Fakes/QueueFake.php
@@ -6,6 +6,7 @@ use BadMethodCallException;
 use Closure;
 use Illuminate\Contracts\Queue\Queue;
 use Illuminate\Queue\QueueManager;
+use Illuminate\Support\Collection;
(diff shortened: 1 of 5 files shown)


yoeriboven:
There's one thing that needs to be fixed before it gets merged.

`QueueFake` is used on [this line](https://github.com/laravel/framework/blob/04e4edce3ecfc3c3d0388409d99f7d147382128a/src/Illuminate/Support/Testing/Fakes/BatchRepositoryFake.php#L48).

I didn't catch this during testing because there is no test for it. 
Am I right in assuming it should be implemented as below?

```php
// Before
new QueueFake(Facade::getFacadeApplication())

// After
new QueueFake($app = Facade::getFacadeApplication(), new QueueManager($app))
```

netpok:
This PR changes the public interface so it should target master

src/Illuminate/Support/Testing/Fakes/QueueFake.php
```diff
@@ -6,20 +6,44 @@
...
+    protected $jobsToFake;
+
     /**
      * All of the jobs that have been pushed.
      *
      * @var array
      */
     protected $jobs = [];
 
+    public function __construct($app, QueueManager $queue, $jobsToFake = [])
```
driesvints, line 38: This is a major breaking change as you're adding a new dependency to this class.

yoeriboven: Should I send it to master?

PR #41344, opened 2022-03-04
[9.x] Adds partial queue faking

yoeriboven:
This PR allows you to only fake certain jobs.

Useful if you want to test if one job dispatches another job.

```php
<?php

namespace App\Jobs;

use ...

class ParentJob
{
    public function handle() 
    {
        ChildJob::dispatch();
    }
}
```

```php
/** @test */
public function it_dispatches_another_job()
{
    Queue::fake(ChildJob::class);

    ParentJob::dispatch();

    Queue::assertPushed(ChildJob::class);
}
```

When using `Queue::fake()` all jobs will be faked. Therefore the `ParentJob` won't run and will never dispatch the `ChildJob`.

With this addition you can choose which job to fake. Not passing a parameter will fake all jobs like before.

**Example:** I run a web scraper which scrapes the homepage of a news site in a job. That job will then dispatch a job for every article it found.

### DIFF
--- a/src/Illuminate/Bus/BusServiceProvider.php
+++ b/src/Illuminate/Bus/BusServiceProvider.php
@@ -18,8 +18,8 @@ class BusServiceProvider extends ServiceProvider implements DeferrableProvider
     public function register()
     {
         $this->app->singleton(Dispatcher::class, function ($app) {
-            return new Dispatcher($app, function ($connection = null) use ($app) {
-                return $app[QueueFactoryContract::class]->connection($connection);
+            return new Dispatcher($app, function ($connection = null, $job = null) use ($app) {
+                return $app[QueueFactoryContract::class]->connection($connection, $job);
             });
         });
 
--- a/src/Illuminate/Bus/Dispatcher.php
+++ b/src/Illuminate/Bus/Dispatcher.php
@@ -216,7 +216,7 @@ class Dispatcher implements QueueingDispatcher
     {
         $connection = $command->connection ?? null;
 
-        $queue = call_user_func($this->queueResolver, $connection);
+        $queue = call_user_func($this->queueResolver, $connection, $command);
 
         if (! $queue instanceof Queue) {
             throw new RuntimeException('Queue resolver did not return a Queue implementation.');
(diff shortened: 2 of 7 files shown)


taylorotwell:
One potential problem I see here is the following situation:

```php
Queue::fake(ChildJob::class);

ParentJob::dispatch()->onConnection('non-default-connection');
```

The `ParentJob` would be dispatched, but to the wrong connection, because `QueueFake::connection` just returns `$this`.

yoeriboven:
I'm not sure I understand what's the problem.

Don't we want to run `ParentJob` on the queue driver defined in `phpunit.xml` in that case? (usually `sync`) Else you'll have to have a queue worker running as well, right?

taylorotwell:
The queue driver in the `phpunit.xml` is just the default driver. It is possible applications are actually running a queue driver during some integration tests... just thinking of all the possible issues. 🤷‍♂️

taylorotwell:
OK - just thinking through this from the very beginning.

I would test your original example like so:

```php
Queue::fake();

$parentJob = new ParentJob;
$parentJob->handle();

Queue::assertPushed(ChildJob::class);
```

Is there another example you are trying to test that isn't currently already possible?

yoeriboven:
In that example you'd lose automatic dependency injection in the `handle` method. I guess you could do it in the constructor but you'll quickly get a lot of parameters in your constructor that way.

The example that made me attempt this PR is when using batches. When you want to test that a job is added to a batch.

```php
class ParentJob
{
    use Batchable, Dispatchable, InteractsWithQueue, Queueable, SerializesModels;

    public function handle() 
    {
        $this->batch()->add(new ChildJob);
    }
{
```
```php
Queue::fake();

Bus::batch(new ParentJob)->dispatch();

Queue::assertPushed(ChildJob::class);
```

You can't do `(new ParentJob)->handle()` because it is not part of a batch. When only faking `ChildJob` and executing ParentJob you can assert that it was pushed to the queue.

Thinking of it we could also add this:

```php
Queue::fake(ChildJob::class);

$batch = Bus::batch(new ParentJob)->dispatch();

Queue::assertAddedToBatch($batch, ChildJob::class);
```

taylorotwell:
Added here: https://github.com/laravel/framework/pull/41425

Thanks

PR #41347, opened 2022-03-04
[10.x] Use native php 8.1 array_is_list function

imanghafoori1:
As laravel 10 requires php 8.1, we can use its new features.
I also added a little bit more tests copied from its counterpart.

https://php.watch/versions/8.1/array_is_list

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -405,9 +405,7 @@ class Arr
      */
     public static function isAssoc(array $array)
     {
-        $keys = array_keys($array);
-
-        return array_keys($keys) !== $keys;
+        return ! array_is_list($array);
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Illuminate/Collections/Arr.php
```diff
@@ -420,7 +418,7 @@ public static function isAssoc(array $array)
      */
     public static function isList($array)
```
antonkomarev, line 419: @taylorotwell do we really need to keep maintaining this method? Maybe mark it as deprecated and remove in next major release?

driesvints: @antonkomarev don't think it's worth it but feel free to attempt a pr

imanghafoori1: @driesvints @antonkomarev I think they are needed for package developers since if you are writing a package you do not know which laravel and PHP version the package will be installed on.
So we can not deprecate and remove them.

PR #41349, opened 2022-03-04
[9.x] Add `insensitive_in` validation rule

ash-jc-allen:
Hi! This PR adds a new `insensitive_in` rule that's similar to `in` but is case-insensitive.

I've had a need for this in a couple of projects in the past, so I thought I'd propose it just in case anyone else has run into the same situation. Sometimes, I want to be able to validate that a value exists in a given list but don't need to necessarily care about the string's case. In my own projects, I'd convert the item and the allowable values to lower case before passing them into the `in` rule. This new rule could remove the need to change any data for validating it.

Examples where the rule would fail using `in` but pass with `insensitive_in`:

```php
Validator::make(['foo' => 'Bar'], ['foo' => 'required|string|insensitive_in:bar']);
Validator::make(['foo' => 'Bar'], ['foo' => 'required|string|insensitive_in:BAR']);
Validator::make(['foo' => 'Bar'], ['foo' => ['required', 'string', Rule::insensitiveIn('bar')]);
```

There's a bit of duplication between the `in` and `insensitive_in` rules, so I think there's probably room for me to clean it up a bit more if you'd like?

If it's something that you think might be worthwhile but needs some changes, please let me know and I'll get it sorted ASAP 😄

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1186,6 +1186,31 @@ trait ValidatesAttributes
     }
 
     /**
+     * Validate an attribute is contained within a list of values without
+     * regardless of the attribute's case.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @param  array  $parameters
+     * @return bool
+     */
+    public function validateInsensitiveIn($attribute, $value, $parameters)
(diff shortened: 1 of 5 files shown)


ash-jc-allen:
@taylorotwell Yeah, to be fair, that's what I was more or less doing before, so I'll stick to that. Thanks 🙂

PR #41356, opened 2022-03-05
[9.x] Add `injectOnlyActionParameters` method to Route

Peterragheb:
This merge request adds `injectOnlyActionParameters` method to Route class.
## Why?

In some cases the developer may want to only receive the parameters he defined in the controller action even if the route changed later and got new bindings added to it.
For example:
Imagine a route like this 
```php
Route::get('hello/{id}', 'App\Http\Controllers\HelloController')
```
and the controller defined like this:
```php
class HelloController extends Controller
{
    public function __invoke(int $id)
    {
       return $id;
    }
}
```
If the route changed later on to be like so:
```php
Route::prefix('{lang}')->get('hello/{id}', 'App\Http\Controllers\HelloController')
```

The controller action defined earlier will either fail with TypeError exception  if there's a type mismatch or it'll get the `$lang`'s value as the `$id` which is an unexpected behavior.

## Solution?

I added `injectOnlyActionParameters` method that can be chained to a defined route. This function will ensure that the controller action will only receive the parameters it expects even if the route changed later.
So after applying this method to the route above, it'll look like this:
```php
Route::prefix('{lang}')->get('hello/{id}', 'App\Http\Controllers\HelloController')
    ->injectOnlyActionParameters();
 ```

### DIFF
--- a/src/Illuminate/Routing/ControllerDispatcher.php
+++ b/src/Illuminate/Routing/ControllerDispatcher.php
@@ -38,7 +38,7 @@ class ControllerDispatcher implements ControllerDispatcherContract
     public function dispatch(Route $route, $controller, $method)
     {
         $parameters = $this->resolveClassMethodDependencies(
-            $route->parametersWithoutNulls(), $controller, $method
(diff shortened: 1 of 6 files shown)


taylorotwell:

Issue #41042, opened 2022-02-15
Model::loadCounts does not work correctly when relationship contains a join


troy-whitespark:
(recreating this as an issue, as requested in my pull with a failing test)

- Laravel Version: 8+
- PHP Version: 8+
- Database Driver & Version: 

### Description:

This problem stemmed from a real issue I was having building relationships that are scoped to particular "locations" in my application.

The issue is, loadCounts and withCounts both create a new subquery and copy the wheres, but they do not copy the join constraints - in the case where we're doing something like inner joining to a pivot table to get permutations of content and location pairs, that means that the relationship_count and relationship()->count() will return two different values and cause bugs.

### Steps To Reproduce:

Existing pull request with failing test is here: https://github.com/laravel/framework/pull/40867

derekmd:
TLDR below; I'm unsure if this is worth investigating since `has()`/`doesntHave()` + `with()` eager loading do play nice with the relationship aggregate methods.

---

From your [given example test](https://github.com/laravel/framework/pull/40867/files), instead of table joins I'd suggest using `has()` / `whereHas()` when defining the relationship, then `with()`/`load()` to access nested `Location` models:

### App\Models\User

```php
class User extends Model
{
    public function locations()
    {
        return $this->hasMany(Location::class);
    }

    public function posts()
    {
        return $this->hasMany(Post::class);
    }

    public function postsOfManyLocations()
    {
        return $this->posts()
            ->where('global', '=', false)
            ->has('locations'); // WHERE EXISTS subquery is run
    }
}
```

### App\Models\Post

```php
class Post extends Model
{
    public function locations()
    {
        return $this->belongsToMany(Location::class, 'post_locations');
    }

    public function user()
    {
        return $this->belongsTo(User::class);
    }
}
```

### App\Models\Location

```php
class Location extends Model
{
    public function posts()
    {
        return $this->belongsToMany(Post::class);
    }

    public function user()
    {
        return $this->belongsTo(User::class)
    }
}
```

```php
// $user = User::withCount('postsOfManyLocations')->first();
$user = User::first()->loadCount('postsOfManyLocations');

// this will be accurate
$user->posts_of_many_locations_count;

$user->postsOfManyLocations->load('locations');
```

---

The docs may need to discourage Eloquent models using `join()` (only covered in the [general query builder section](https://laravel.com/docs/queries#joins)) since there are a few other active record issues I'm aware of:

[...]

PR #41359, opened 2022-03-05
[8.x] Added `withCountAsStated()` and `loadCountAsStated()` methods

theriddleofenigma:
These two new methods `withCountAsStated()` and `loadCountAsStated()` allows us to customize the relationship queries while loading the counts as sub-select.

Example:

```
Class User extends Model {
    public function commentedUsers() {
        return $this->hasMany(Post::class)->join('comments', 'posts.id', 'comments.post_id')->selectRaw('distinct(user_id)');
    }
}


// Before

User::withCount('commentedUsers')->first();

// Query: select users.*, (select count(*) from posts where posts.user_id = users.id) as commented_users_count from users limit 1;

// After

User::withCountAsStated('commentedUsers')->first();

// Query: select users.*, (select count(*) from (select distinct(user_id) from posts join comments on posts.id = comments.post_id where posts.user_id = users.id) as commented_users_aggregate) as commented_users_count from users limit 1;
```

And the `withCount()` and `loadCount()` method still behave the same.

This behaviour allows one to customize the count query for the sub-select and avoiding the need for writing raw sub-select because of the default behaviour.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Collection.php
+++ b/src/Illuminate/Database/Eloquent/Collection.php
@@ -69,9 +69,10 @@ class Collection extends BaseCollection implements QueueableCollection
      * @param  array|string  $relations
      * @param  string  $column
      * @param  string  $function
+     * @param  bool  $preserveRelationState
      * @return $this
      */
-    public function loadAggregate($relations, $column, $function = null)
+    public function loadAggregate($relations, $column, $function = null, $preserveRelationState = false)
     {
         if ($this->isEmpty()) {
             return $this;
@@ -80,7 +81,7 @@ class Collection extends BaseCollection implements QueueableCollection
         $models = $this->first()->newModelQuery()
             ->whereKey($this->modelKeys())
             ->select($this->first()->getKeyName())
-            ->withAggregate($relations, $column, $function)
+            ->withAggregate($relations, $column, $function, $preserveRelationState)
             ->get()
             ->keyBy($this->first()->getKeyName());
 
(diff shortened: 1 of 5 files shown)


GrahamCampbell:
I think you meant to target 9.x. 8.x is only open to bug fixes, now.

theriddleofenigma:
@GrahamCampbell Yeah, but the behavior of `withCount()` producing the count result without the joins and custom selected column seems to be a bug. So made a PR to 8.x.

In order to solve this issue #41042

theriddleofenigma:
@GrahamCampbell Also, The newly added helper produces 2 levels of nested sub-select. But the **MySQL 5.7** and **MariaDB 10** was only supported to pass the table reference to 1 level of sub-select. Hence the 2 database checks are failing. I'm not sure whether it's fine or not to release these methods without supporting **MySQL 5.7** and **MariaDB 10**.

src/Illuminate/Database/Eloquent/Collection.php
```diff
@@ -69,9 +69,10 @@ public function load($relations)
      * @param  array|string  $relations
      * @param  string  $column
      * @param  string  $function
+     * @param  bool  $preserveRelationState
      * @return $this
      */
-    public function loadAggregate($relations, $column, $function = null)
+    public function loadAggregate($relations, $column, $function = null, $preserveRelationState = false)
```
driesvints, line 75: This is a breaking change and cannot be done on a patch release. Please re-send this without this change. Thanks

src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
```diff
@@ -556,16 +557,26 @@ public function withAggregate($relations, $column, $function = null)
...
+                $query = $relation->getRelationExistenceQuery(
+                    $relation->getRelated()->newQuery(), $this, new Expression($expression)
+                )->setBindings([], 'select');
+            }
 
             $query->callScope($constraints);
 
-            $query = $query->mergeConstraintsFrom($relation->getQuery())->toBase();
+            if ($preserveRelationState) {
+                $query = $query->getConnection()->query()->from($query, $name.'_aggregate')->select(new Expression('count(*)'));
```
derekmd, line 576: `count(*)` is hardcoded here but this method also supports `exists`, `max`, `min`, `avg`, or `sum` as indicated by the `$function` argument.

taylorotwell:
Closing pending inactivity.

PR #41361, opened 2022-03-06
[8.x] Backport dynamically access batch options

crynobone:
This was added to the project documentation and it would be better to have it available for all supported versions.


taylorotwell:
@crynobone isn't this already on 9.x? I'm confused:

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Bus/Batch.php#L476

PR #41362, opened 2022-03-06
[9.x] BCC broken on 9.x with AWS SES Mail Transport

### LINKED ISSUE #41357: Blind copies are not sent by sendmail transport
- Laravel Version: 9.3.1
- PHP Version: 8.0.14

### Description:
Using a mail controller like this:

```php
class SendmailController extends Controller
{
	public function post_sendmail(Request $request)
	{
		$email = new Sendmail($request->subject, $request->msg);

		Mail::to($request->to)
			->bcc($request->bcc)
			->send($email);

		return "Email sent";
	}
}
```

If MAIL_MAILER=smtp in the .env file, and a suitable SMTP server is configured, the blind copy is sent, but when MAIL_MAILER=sendmail it isn't.

### Steps To Reproduce:
1. Extract the sendmail app from the attached archive to a linux system with a sendmail binary.
2. cd sendmail
3. composer update
4. cd public
5. php -S 127.0.0.1:8080
6. In a web browser, browse to http://127.0.0.1:8080/
7. Complete the fields in the form.
8. Click Send.

[sendmail.tar.gz](https://github.com/laravel/framework/files/8191129/sendmail.tar.gz)

cuppett:

I've corrected both of those and now can receive To/Cc/Bcc emails.

### DIFF
--- a/src/Illuminate/Mail/Transport/SesTransport.php
+++ b/src/Illuminate/Mail/Transport/SesTransport.php
@@ -11,6 +11,12 @@ use Symfony\Component\Mailer\Transport\AbstractTransport;
 class SesTransport extends AbstractTransport
 {
     /**
+     * Amazon SES has a limit on the total number of recipients per message.
+     * See Also: https://docs.aws.amazon.com/ses/latest/dg/quotas.html.
+     */
+    protected const RECIPIENT_LIMIT = 50;
+
+    /**
      * The Amazon SES instance.
      *
      * @var \Aws\Ses\SesClient
@@ -45,16 +51,24 @@ class SesTransport extends AbstractTransport
     protected function doSend(SentMessage $message): void
     {
         try {
-            $this->ses->sendRawEmail(
-                array_merge(
-                    $this->options, [
-                        'Source' => $message->getEnvelope()->getSender()->toString(),
-                        'RawMessage' => [
-                            'Data' => $message->toString(),
-                        ],
-                    ]
-                )
-            );
+            /**
+             * Batch up API calls to adhere to the recipient limit of the service.
+             * See Also: https://docs.aws.amazon.com/aws-sdk-php/v2/api/class-Aws.Ses.SesClient.html#_sendRawEmail.
+             */
+            $recipientList = collect($message->getEnvelope()->getRecipients());
+            foreach ($recipientList->chunk(self::RECIPIENT_LIMIT) as $recipients) {
+                $this->ses->sendRawEmail(
+                    array_merge(
+                        $this->options, [
+                            'Source' => $message->getEnvelope()->getSender()->toString(),
+                            'RawMessage' => [
+                                'Data' => $message->toString(),
+                            ],
+                            'Destinations' => $recipients,
+                        ]
+                    )
+                );
+            }
         } catch (AwsException $e) {
             throw new Exception('Request to AWS SES API failed.', $e->getCode(), $e);
         }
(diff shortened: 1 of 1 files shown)


src/Illuminate/Mail/Transport/SesTransport.php
```diff
@@ -10,6 +10,12 @@
 
 class SesTransport extends AbstractTransport
 {
+    /**
+     * Amazon SES has a limit on the total number of recipients per message.
+     * See Also: https://docs.aws.amazon.com/ses/latest/dg/quotas.html.
+     */
+    const recipientLimit = 50;
```
seriquynh, line 17: This constant must be RECIPIENT_LIMIT because of coding convention.

cuppett: Fixed, thanks!

src/Illuminate/Mail/Transport/SesTransport.php
```diff
@@ -45,21 +51,43 @@ public function __construct(SesClient $ses, $options = [])
...
-                array_merge(
-                    $this->options, [
-                        'Source' => $message->getEnvelope()->getSender()->toString(),
-                        'RawMessage' => [
-                            'Data' => $message->toString(),
-                        ],
-                    ]
-                )
-            );
+            foreach ($this->getRecipients($message) as $recipients) {
```
GrahamCampbell, line 54: this looks wrong. we should chunk into batches of 50, instead.

GrahamCampbell (changes requested):

src/Illuminate/Mail/Transport/SesTransport.php
GrahamCampbell, line 54: your method re-implements chunk. you should just use collection chunk.

GrahamCampbell (changes requested):

src/Illuminate/Mail/Transport/SesTransport.php
```diff
@@ -10,6 +10,12 @@
 
 class SesTransport extends AbstractTransport
 {
+    /**
+     * Amazon SES has a limit on the total number of recipients per message.
+     * See Also: https://docs.aws.amazon.com/ses/latest/dg/quotas.html.
+     */
+    const RECIPIENT_LIMIT = 50;
```
GrahamCampbell, line 17: should be protected

GrahamCampbell (changes requested):

src/Illuminate/Mail/Transport/SesTransport.php
cuppett: Done. Thanks!

cuppett:
> Does this mean it's also broken in 8.x but hidden? If so we need to target 8.x

Correct. BCC worked, but I didn't see any indication of adherence to the 50 recipient limit.

driesvints:
Please send this to 8.x then, thanks.

cuppett:
I can definitely open another. Won't we still need two PRs? The code in question wasn't exactly identical between the branches.

driesvints:
We'll try to merge this into 9.x

cuppett:
I can tell the PR will be different for 8.x. there are some headers added
for the SES return IDs and a recipient count returned from the function.

Do we want 2 PRs?

driesvints:
@cuppett just send in whatever you think is best 👍

cuppett:
Can do. I may open both just so we have all the right raw material to work with easily.

PR #41367, opened 2022-03-06
Eloquent\Builder $query->value() clear return type on not finding a model

vanyamil:
Theoretically should not change anything (I am uncertain of how PHP handles evaluating a no-return, but maybe it auto returns null), but should clearly indicate to readers that in case of not being able to find a model through the query, value() returns null. This matches the behaviour in Query\Builder.


driesvints:
The current method already returns this. In general, we don't explicitly add `return null`.

PR #41374, opened 2022-03-07
PHP 8.0: switch to match

mrrazvi:


SjorsO:
`preg_match` returns 0 or 1, and `match` is a strict comparison (`===`), so the last two cases will never match.

driesvints:
Please resubmit this with passing tests, thanks.

PR #41376, opened 2022-03-07
[9.x] PHP 8, switch to match

mrrazvi:
much shorter and nicer way

### DIFF
--- a/src/Illuminate/Notifications/resources/views/email.blade.php
+++ b/src/Illuminate/Notifications/resources/views/email.blade.php
@@ -19,7 +19,7 @@
 {{-- Action Button --}}
 @isset($actionText)
 <?php
-$color = match ((string)$level) {
+$color = match ((string) $level) {
     'success', 'error' => $level,
     default => 'primary',
 };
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This is actually a breaking change. The old code used to work with stringable objects. The new code does not.

rizimore:
So, I close the pull request?

src/Illuminate/Notifications/resources/views/email.blade.php
```diff
@@ -19,14 +19,10 @@
...
 <?php
-    switch ($level) {
-        case 'success':
-        case 'error':
-            $color = $level;
-            break;
-        default:
-            $color = 'primary';
-    }
+$color = match ($level) {
```
netpok, line 22: To fix the BC:
```suggestion
$color = match ((string)$level) {
```

src/Illuminate/Notifications/resources/views/email.blade.php
```diff
@@ -19,14 +19,10 @@
...
 <?php
-    switch ($level) {
-        case 'success':
-        case 'error':
-            $color = $level;
-            break;
-        default:
-            $color = 'primary';
-    }
+$color = match ((string)$level) {
```
GrahamCampbell, line 22: missing space after `(string)`

GrahamCampbell (changes requested):

Issue #16570, opened 2016-11-28
Namespace received from composer.json in \Illuminate\Foundation\Application


HDsign:
- Laravel Version: 5.3.*
- PHP Version: 7.0.12

### Description:
I have a problem where the composer.json file is not placed in the same folder as the project it self.

My folder structure:

```
/composer.json
/lara/app/*
/lara/bootstrap/*
/lara/config/*
...
```

So when the `getNamespace()` methode is used on the `\Illumintate\Foundation\Application` class it will not find my composer.json file:

```php
$composer = json_decode(file_get_contents(base_path('composer.json')), true);

foreach ((array) data_get($composer, 'autoload.psr-4') as $namespace => $path) {
    foreach ((array) $path as $pathChoice) {
        if (realpath(app_path()) == realpath(base_path().'/'.$pathChoice)) {
            return $this->namespace = $namespace;
        }
    }
}
```

This can of course be solved easily by adding a `setNamespace` methode to the application class or perhaps add the possibility to add it to the config files. Right now the only way to fix this without changing everything drastically is by adding a `composer.json` file to the 'lara' folder.

themsaid:
Feel free to propose a PR.

PR #41378, opened 2022-03-07
[9.x] Add `setNamespace` method

istiak-tridip:
This PR adds a new function to the `Foundation\Application` class:
- `setNamespace($namespace)` — Set the application namespace

**Motivation / Why:**
- If you don't want to add `composer.json` to the production bundle (my motivation)
- If your `composer.json` isn't placed in the project root (issue: #16570)

By setting namespace manually we can prevent the `Application` class from searching for `composer.json`.

### DIFF
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -1401,7 +1401,7 @@ class Application extends Container implements ApplicationContract, CachesConfig
     /**
      * Set the application namespace.
      *
-     * @param string $namespace
+     * @param  string  $namespace
      */
     public function setNamespace($namespace)
     {
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -1399,6 +1399,16 @@ class Application extends Container implements ApplicationContract, CachesConfig
     }
 
     /**
+     * Set the application namespace.
+     *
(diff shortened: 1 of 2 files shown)


driesvints:
You should always include your `composer.json` file with your project and it should always be placed in the project root.

istiak-tridip:
@driesvints I generally archive the production build (composer vendor & pre-generated frontend assets) in a `.zip` file and send them to my client and they just extract it and it's all ready to go. 

They're not developers and `composer` is not even installed in their machine. Why do they really need the `composer.json` file?
I'm just trying to understand.

Thanks.

driesvints:
Why do you need to remove it? You can still zip it with vendor files. They don't need to use composer. The file's location is the important part.

istiak-tridip:
My main motivation was to find a way to ship without the `composer.json` or `composer.lock` file. My `composer.json` is stored in the project root, and I just linked a related issue. 

Anyway, thank you @driesvints for the discussion. I will go ahead and close the PR in a few minutes.

driesvints:
If there's no real reason to remove it then I don't feel we should add this.

driesvints:
You should never add credentials directly to your composer.json. Instead you should use separate auth files like an `auth.json` file. See https://getcomposer.org/doc/articles/authentication-for-private-packages.md for more info.

PR #41380, opened 2022-03-07
[10.x] Replace password reset string constants with enum

netpok:
With dropping php8.0 support it makes sense to use an enum to represent a limited set of responses.

This PR replaces the password reset string constants with an enum. I have used a string backed enum so it can be used with minimal migration to support both Laravel9 and 10:

```php
$responseString = is_object($response) ? $response->value : $response;
```

If this deemed unnecessary I will remove the constants from the interface and the test.

### DIFF
--- a/src/Illuminate/Auth/Passwords/PasswordBroker.php
+++ b/src/Illuminate/Auth/Passwords/PasswordBroker.php
@@ -5,6 +5,7 @@ namespace Illuminate\Auth\Passwords;
 use Closure;
 use Illuminate\Contracts\Auth\CanResetPassword as CanResetPasswordContract;
 use Illuminate\Contracts\Auth\PasswordBroker as PasswordBrokerContract;
+use Illuminate\Contracts\Auth\PasswordResetResponse;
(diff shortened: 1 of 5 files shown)


driesvints:
Heya, I don't really see any big benefit of doing this? The current way works fine and this is only a breaking change to users.

netpok:
It basically gives an interface for the response and thus prevents errors.

For example random netizen Joe watched a tutorial that says error states should be handled first and writes the following (very bad code):

```php
    public function reset(Request $request)
    {
        $status = $this->passwordBroker->reset(...);

        if ($status == Password::INVALID_USER){
            return 'invalid user page';
        }
        
        if ($status == Password::INVALID_TOKEN){
            return 'invalid token page';
        }
        
        Auth::login('user resolved request data');
        
        return 'home page';
    }
```

And he also uses the RandomPasswordResettingPackageNo57 where the maintainer decided to return `passwords.service-unavailable` whenever the upstream service is unavailable. Now when the upstream service is unavailable the user will be logged in after a failed password reset.

PR #41381, opened 2022-03-07
[9.x] Make DatabaseManager macroable

telkins:
I'd like to propose this change primarily to make my life a little easier.

I'm working with a project that has multiple database connections and was hoping that I could introduce via `Macroable` methods where I could access the various connections more easily and in a way that is a little nicer to look at.

For example, instead of this:
```php
DB::connection('foo_connection')->whatever();
```

I could do this:
```php
DB::foo()->whatever();
```

As I noted, I'm not sure if there's a different/better way to accomplish this.  I'm also not sure if there's a good reason to **not** do this.  As such, I'm open to any suggestions, links, whatever that will help.  🤓 

I'm also aware that there may be other, better reasons to make `DatabaseManager` macroable...but I can't think of any at this moment.  😜 

That's it.  🚀

### DIFF
--- a/src/Illuminate/Database/DatabaseManager.php
+++ b/src/Illuminate/Database/DatabaseManager.php
@@ -7,6 +7,7 @@ use Illuminate\Database\Connectors\ConnectionFactory;
 use Illuminate\Support\Arr;
 use Illuminate\Support\ConfigurationUrlParser;
 use Illuminate\Support\Str;
+use Illuminate\Support\Traits\Macroable;
 use InvalidArgumentException;
 use PDO;
 use RuntimeException;
@@ -16,6 +17,8 @@ use RuntimeException;
  */
 class DatabaseManager implements ConnectionResolverInterface
 {
+    use Macroable;
+
     /**
      * The application instance.
      *


taylorotwell:
Usually classes that already have a `__call` method need some additional logic to make Macroable work. Is that not the case here?

telkins:
> Usually classes that already have a `__call` method need some additional logic to make Macroable work. Is that not the case here?

That appears to be the case here, as you suggested.  I will look for some examples of other classes that are macroable and that already have a `__call()` method and see what might be necessary to make this work...and then update the PR.

Thx.

src/Illuminate/Database/DatabaseManager.php
```diff
@@ -16,6 +17,8 @@
  */
 class DatabaseManager implements ConnectionResolverInterface
 {
+    use Macroable;
```
sulemaanhamza, line 20:
```suggestion
    use Macroable {
        __call as macroCall;
    }
```

sulemaanhamza: As @taylorotwell suggests,  classes that already have a __call method need some additional logic to make Macroable work. You also need to update the __call() method accordingly,




`
__call($method, $parameters) { 
return static::hasMacro($method)
                    ? $this->macroCall($method, $parameters)
                    : $this->connection()->$method(...$parameters); }
`

telkins: @sulemaanhamza Thx...!  🤓 I was hoping to have time to look into finishing this up this weekend and it looks like you've already done all the hard work.  😜 I definitely appreciate it.

I'll take some time later today or tomorrow to look more closely at your suggestion and try to get this thing wrapped up.  Thanks again.  🚀

driesvints:
@telkins feel free to resubmit this when you have time.

PR #41388, opened 2022-03-07
[9.x] BCC broken on 9.x with AWS SES Mail Transport

cuppett:
Redo of #41362
8.x version of same is #41389 

I've corrected both of those and now can receive To/Cc/Bcc emails.

### DIFF
--- a/src/Illuminate/Mail/Transport/SesTransport.php
+++ b/src/Illuminate/Mail/Transport/SesTransport.php
@@ -11,6 +11,12 @@ use Symfony\Component\Mailer\Transport\AbstractTransport;
 class SesTransport extends AbstractTransport
 {
     /**
+     * Amazon SES has a limit on the total number of recipients per message.
+     * See Also: https://docs.aws.amazon.com/ses/latest/dg/quotas.html.
+     */
+    protected const RECIPIENT_LIMIT = 50;
+
+    /**
      * The Amazon SES instance.
      *
      * @var \Aws\Ses\SesClient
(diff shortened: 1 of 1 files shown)


taylorotwell:
This seems to attempt to address the recipient limit. How is BCC broken and how is it addressed in this PR?

taylorotwell:
The code in 8.x did not seem to have this chunk or the `Destinations` key either though? So, how did this ever work on 8.x?

cuppett:
> The code in 8.x did not seem to have this chunk or the `Destinations` key either though? So, how did this ever work on 8.x?

I didn't try 8.x in a debugger (just 9.x), but in 9.x I saw the $message->toString() report back:

```
From: somebody
To: somebody_else
// No BCC
Subject: Whatever

content
```

I bet if I did the same in 8.x, I'd see this:

```
From: somebody
To: somebody_else
Bcc: hidden_person
Subject: Whatever

content
```

And then the mail relay sniffs it out and strips that header when it's doing the relaying.

This difference in ->toString() is probably why Sendmail stopped working too (https://github.com/laravel/framework/issues/41357)

cuppett:
(Ideally, we'd restore the old ->toString() output behavior; however, it's still dangerous to rely on it and continue to ignore the SES recipient limits)

src/Illuminate/Mail/Transport/SesTransport.php
```diff
@@ -45,16 +51,21 @@ public function __construct(SesClient $ses, $options = [])
...
+            // See also: https://docs.aws.amazon.com/aws-sdk-php/v2/api/class-Aws.Ses.SesClient.html#_sendRawEmail.
+            foreach (collect($message->getEnvelope()->getRecipients())->chunk(self::RECIPIENT_LIMIT) as $recipients) {
+                $this->ses->sendRawEmail(
+                    array_merge(
+                        $this->options, [
+                            'Source' => $message->getEnvelope()->getSender()->toString(),
+                            'RawMessage' => [
+                                'Data' => $message->toString(),
+                            ],
+                            'Destinations' => $recipients,
```
taylorotwell, line 64: How is this array supposed to be formatted?

taylorotwell:
Fixed here: https://github.com/laravel/framework/commit/8ca43f4c2a531ff9d28b86a7e366eef8adf8de84

PR #41389, opened 2022-03-07
[8.x] Honor recipient limit with AWS SES Mail Transport

cuppett:

I've corrected both of those and this is the merge for 8.x

### DIFF
--- a/src/Illuminate/Mail/Transport/SesTransport.php
+++ b/src/Illuminate/Mail/Transport/SesTransport.php
@@ -10,6 +10,12 @@ use Swift_TransportException;
 class SesTransport extends Transport
 {
     /**
+     * Amazon SES has a limit on the total number of recipients per message.
+     * See Also: https://docs.aws.amazon.com/ses/latest/dg/quotas.html.
+     */
+    protected const RECIPIENT_LIMIT = 50;
+
+    /**
      * The Amazon SES instance.
      *
      * @var \Aws\Ses\SesClient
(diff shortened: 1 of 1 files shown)


cuppett:
Looks like my naive array_merge attempt wasn't good.

cuppett:
(Hang on here for a day, I'll find the answer and tack on a commit)

src/Illuminate/Mail/Transport/SesTransport.php
```diff
@@ -46,25 +52,32 @@ public function send(Swift_Mime_SimpleMessage $message, &$failedRecipients = nul
...
+                                'Data' => $message->toString(),
+                            ],
+                            'Destinations' => $recipients,
+                        ]
+                    )
+                );
+
+                $messageId = $result->get('MessageId');
+
+                $message->getHeaders()->addTextHeader('X-Message-ID', $messageId);
```
GrahamCampbell, line 74: This will only add the message ID of the last message, right, throwing out the earlier IDs?

GrahamCampbell (changes requested):

GrahamCampbell:
Probably the array merge didn't work because of keys. Try jamming it through array_values. When the input is a list, array_merge concats all the things.

GrahamCampbell:
I'm honestly not even sure this is a bug. We probably shouldn't artificially try to support this stuff, otherwise people could get cced in emails with no to address. Very confusing. We should maybe leave this up to application code to decide how to chunk up the recipients, instead of the framework taking this decision for them.

cuppett:
> Probably the array merge didn't work because of keys. Try jamming each arg through array_values, first. When the input is a list, array_merge concats all the things.

The error in the tests seemed to indicate that $message->cc was 'null'. I'll try to work with it and find the right coordinates or generator to get real addresses.

cuppett:
> I'm honestly not even sure this is a bug. We probably shouldn't artificially try to support this stuff, otherwise people could get cced in emails with no to address. Very confusing. We should maybe leave this up to application code to decide how to chunk up the recipients, instead of the framework taking this decision for them.

I doubt this is the MUA experience on the end, the payload is the same in SMTP, it just overrides their relay how to proceed. I'll run some SES tests though to confirm (setting chunk to 1 with a To and a Bcc).

cuppett:
> We should maybe leave this up to application code to decide how to chunk up the recipients, instead of the framework taking this decision for them.

I'd be leery of having my application code condition based on transport and know/behave different based on the inner workings of the APIs/implementations (sendmail versus mailgun versus SES).

GrahamCampbell:
Does mailgun actually support a large number of ccs, or does it just silently drop them on the floor?

cuppett:
> Does mailgun actually support a large number of ccs, or does it just silently drop them on the floor?

It was an example only to illustrate how hairy it could get, I only know SES, sorry.

src/Illuminate/Mail/Transport/SesTransport.php
cuppett: If it has to be a singular value and most RFC docs specify it will be when present, what do we want to capture here? 

I could aggregate them up and add a CSV of the values? I'm not sure who may inspect/capture this or what processing they may or may not do today. I could also just capture the first one and drop any subsequent batches (to keep it behaving exactly like today, but sending > 50 correctly and losing those extra message IDs).

A single value (first, last, any) is the same "API" as today. First represents identical. Last would be a slight change. A CSV is a change in the API. In 9.x, this code is gone altogether, but I didn't want to change the "API" .

cuppett:
> I'm honestly not even sure this is a bug. We probably shouldn't artificially try to support this stuff, otherwise people could get cced in emails with no to address. Very confusing. We should maybe leave this up to application code to decide how to chunk up the recipients, instead of the framework taking this decision for them.

There's some interesting commentary here to this effect:

https://community.sophos.com/utm-firewall/f/mail-protection-smtp-pop3-antispam-and-antivirus/85124/smtp-proxy---sending-email-to-50-recipients-split-into-several-smtp-envelopes

That, in general, mailers should support up to 100 RCPT and it's just unique that SES sets limit to 50. Without extra testing, I do think we'd see some awkward behavior.

I actually think I'd want to defer this 8.x fix altogether and focus on the BCC breakage in 9.x.

driesvints:
Closing this one as the 9.x equivalent was rejected as well. It's best to solve this one in userland I think.

PR #41392, opened 2022-03-08
[8.x] Fix option overrides

### LINKED ISSUE #41382: HTTP Client query parameters not merged as expected after upgrading
- Laravel Version: v8.83.3
- PHP Version: 8.1.1
- Database Driver & Version: PostgreSQL 13

### Description:

I just upgraded Laravel from v8.83.0 to 8.83.3 and started experiencing issues with query parameters not being merged as expected via 
```
public function __construct()
{
    $this->client = Http::baseUrl('https://api.hubapi.com')
        ->withOptions([
            'query' => [
                'hapikey' => config('services.hubspot.api_key'),
            ],
        ]);
}
```
```
public function getPipelines(): array
{
    return $this->client
        ->asJson()
        ->get('/crm/v3/pipelines/deals')
        ->throw()
        ->json();
}

query: ""
```
I am aware that the get() method takes an array as a second argument for populating the query parameters. But I would expect the above to work anyway, since this would work using Guzzle directly. If I add an empty array `->get('https://google.com', [])` it works as expected and correctly appends `?test=test` to the url. I encountered this issue, since I am declaring a class property of type PendingRequest that I reuse for different purposes. It did not break any ->post() or ->patch() methods, but it started failing for all the ->get() methods. 

I could add the placeholder array to all occurrences, but at the moment I have around 100 places where I use this. I also suspect that other people might encounter this issue.

### Steps To Reproduce:
Does not add query parameters

```
Http::withOptions([
    'query' => [
      'test' => 'test',
    ]
])->get('https://google.com')

query: ""
```

Correctly adds query parameters

```
Http::withOptions([
    'query' => [
      'test' => 'test',
    ]
])->get('https://google.com', [])

query: "?test=test"
```

driesvints:
This PR fixes an issue with option overriding introduced in https://github.com/laravel/framework/pull/40954. The problem is that by default a value of `'options' => null` gets added when the default value of a HTTP call (`get`, `post`,...) is left empty.  [Because `array_merge_recursive` was changed to `array_replace_recursive`](https://github.com/laravel/framework/pull/40954/files#diff-1ac9aa0ab55e04c899a86ef074016e5ebc3683e3e90101f03284305301771067L990-R990), the default value removed the value of `query` set with `withOptions`.

The way I fixed it is to filter out any `null` value of options before sending the request. It doesn't seems you want to add `null` values to a request? However, if users are explicitly removing options set with `withOptions` by assigning `null` values, this will no longer work. *But* all tests pass right now and the docs don't note anything about that behavior so I don't think that's expected.

Related: https://github.com/laravel/framework/pull/40954 & https://github.com/laravel/framework/commit/30e341b7fe4e4d9019df42b7eff6c7dfa5ea30e5#

Fixes https://github.com/laravel/framework/issues/41382


taylorotwell:
Fixed https://github.com/laravel/framework/commit/a54f48102deea2864071e510172fe0b22a1c1d5a

taylorotwell:
Fixed https://github.com/laravel/framework/commit/a54f48102deea2864071e510172fe0b22a1c1d5a

PR #41395, opened 2022-03-08
Simpify `Request::is()` and `Request::fullUrlIs()`

sebdesign:
https://github.com/laravel/framework/pull/39857 replaced a foreach loop with a `Collection::contains()`, but the `Str::is()` method can already handle an array of patterns to find a single match.

So it makes sense to pass the the variadic arguments directly to this method instead of a foreach loop or a collection method.


taylorotwell:
If the code is not broken I don't want to change it.

sebdesign:
It wasn't broken before #39857 either. I'm only suggesting this change because `Collection::contains()` is redundant here and does not provide anything that `Str::is()` cannot do.

Issue #41394, opened 2022-03-08
Issue with `afterResolving` when the class had already been resolved.


pxpm:
Hello guys! 

I have a question regarding the `afterResolving` callbacks.  First of all sorry if I miss something basic here, I will try my best to explain it clearly. The question first and I will give the context after to don't biase the answer to a concrete example:

- **What is expected to happen when you define a `afterResolving` callback to some class that had already been resolved in the container ? Shoudn't that callback be imediatelly run on the already resolved class ?** 

https://github.com/laravel/framework/blob/22c2cee98a229d8da6ae02caadda6848565dfd94/src/Illuminate/Container/Container.php#L1154

#### Context of my use case: 

So the documented way to add custom blade directives is by going into the ServiceProvider `boot()` method and use the `Blade::directive()`. All fine. 

For long long time I've been using the following to add my own directives: 

```php

// NOTE: this on the register() method of my provider

$this->app->afterResolving('blade.compiler', function (BladeCompiler $bladeCompiler) {
    $bladeCompiler->directive(someDirective', function ($parameter) {
        // ....
    });
});
```

Everything works fine as long as  **MyProvider is loaded before** other providers that define the blade directives with `Blade::directive()`, otherwise it would resolve the class first, and my registered callback would never run. I can see it in the `afterResolvingCallbacks` but the `blade.compiler` is already resolved when I register my callback if any other ServiceProvider is registering the diirectives without using `afterResolving()`!

![image](https://user-images.githubusercontent.com/7188159/157238139-dc388970-15a4-4437-8ee7-4b8a766b2da6.png)

My expectation would be that an `afterResolving` callback would be called if the class is already resolved, probably it would be impossible to do that for a `beforeResolving` since at that time it's already resolved, but for a `afterResolving` can't we work something out ? 

The reason why I asked first is because I don't expect answers like `change your directive definition`, I want the focus to be, what happens when I define a `afterResolving` callback to some class that had already been resolved. 

Thanks for your time  and patience! Let me know if I wasn't clear enough!

Pedro

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
[...]

PR #41397, opened 2022-03-08
[8.x] Run `afterResolving` callbacks if the type is resolved

pxpm:
Hello guys. 

I am sorry if my explanation on issue https://github.com/laravel/framework/issues/41394 was not enough. I hope with this PR the things I tried to explain become more clear for what I meant.

**Adding callbacks to `afterResolvingCallbacks` array have no effect at all if the type is already resolved in the container.** 

This proposal aims to check if the type is already resolved when someone tries `$this->app->afterResolving('something', function() {}))` and `something` is already resolved. That callback would never be run, will just sit in the `afterResolvingCallbacks` array forever.

I haven't mess much with Laravel core, so I hope I am not attempting to do something eregious, if I am, please disregard, but it does not make sense to me that I define an `afterResolving` callback on something and it does not run because it is already resolved, if it is resolved, fine, just run what I want since the resolving part is done. I guess this is the mindset to look at this `afterResolving` callback? 

I look at it as a issue from Laravel side since it accepts my `afterResolving` callbacks in the ServiceProvider so I expect that whenever that type is resolved, It would run those callbacks, even if previously resolved.

Once again, sorry for insisting in this issue, I think there is something more we can do here, and I am here to help on what I can. 👍 

Even if this is not the proper or most adequate way to solve this, with some guidance I can try to work something out. 

I know Laravel eco-system is huge, but this `afterResolving` callbacks have taken hours of sleep from many developers across many packages like:

https://github.com/spatie/laravel-permission/pull/2048
https://github.com/Laravel-Backpack/CRUD/issues/4251
https://github.com/spatie/laravel-markdown/pull/35

And atleast 3 or 4 more that I am aware of. 

Debugging this have been an hell, help us here please 😅 

Thanks for everything, 🙏

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -1160,7 +1160,14 @@ class Container implements ArrayAccess, ContainerContract
         if ($abstract instanceof Closure && is_null($callback)) {
             $this->globalAfterResolvingCallbacks[] = $abstract;
         } else {
-            $this->afterResolvingCallbacks[$abstract][] = $callback;
+            // if the type is resolved, we can already run the callback on the object.
+            if($this->resolved($abstract)) {
+                $object = $this->get($abstract);
+                $this->fireCallbackArray($object, [$callback]);    
+            }else {
+                // if not resolved save the callbacks for later execution.
+                $this->afterResolvingCallbacks[$abstract][] = $callback;
+            }
         }
     }
 


src/Illuminate/Container/Container.php
```diff
@@ -1160,7 +1160,14 @@ public function afterResolving($abstract, Closure $callback = null)
         if ($abstract instanceof Closure && is_null($callback)) {
             $this->globalAfterResolvingCallbacks[] = $abstract;
         } else {
-            $this->afterResolvingCallbacks[$abstract][] = $callback;
+            // if the type is resolved, we can already run the callback on the object.
+            if($this->resolved($abstract)) {
+                $object = $this->get($abstract);
+                $this->fireCallbackArray($object, [$callback]);    
+            }else {
```
pxpm, line 1167: If the callback should still be needed for later execution, we should remove this `else`, so it fires the callback once (since it's already resolved) and store the callback for further executions (if resolved again ?). I was not sure how to do it here 👍

erikn69:
This is a real headache, I hope that it will be solved once and for all

taylorotwell:
Maybe I'm missing something - but I looked through the issues and don't see how this PR would have solved them? Two of the issues were fixed by simply using `afterResolving` where it wasn't being used at all?

GrahamCampbell:
This is not a bug fix. This is a breaking change. You need to check if it's resolved yet.

GrahamCampbell:
Here's the evidence: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/ServiceProvider.php#L254-L268.

GrahamCampbell:
Looks like you are calling the wrong method in  your service provider. You need `$this->callAfterResolving`.

pxpm:
Hey @taylorotwell . Thanks for the answer. 

I don't like to give concrete examples because the focus is moved to those examples and gets away from the main problem. But if following the examples is easier for you guys it is as the following:
- if all providers that register the directives using `afterResolving` are loaded before the ones that use `Blade:: directive ()` everything is fine. 

If one register the directives using the Blade facade that forces the BladeCompiler to be resolved no other providers could use afterResolving for that class after that directive is loaded.

It's not a register/boot method problem, it's a Container problem IMO. 

The after resolving is not truly a after resolving but it's a: after resolving if not resolved. And that's bad, no reason at all to be like it at all. 

Did I make myself clear about the problem here? Blade compiler was just how we found it, but the underlaying problem is that afterResolving does not respect the rules provided in some specific scenarios. 

Let me know, thanks for sticking with us developers here. This issue has ages!

pxpm:
> Looks like you are calling the wrong method in your service provider. You need `$this->callAfterResolving`.

I get that I might not get the method names correctly. I excused myself plenty of times. But the things is: when the callback is registered the class is already resolved, so it will store the callback in the array, but nothing will happen.

taylorotwell:
@pxpm but the core issue is you are registering Blade extensions in the service provider `register` method. There is no guarantee that Blade is even available in the container at that point in the application lifecycle. If the order of service provider registration were to change internally - your application would break.

The real solution to this is to only use `register` for *registering* services in the container. Do not resolve or interact with any bound services otherwise you are going to have these problems. Maybe I can make that clearer in the docs. Your Blade extensions should be registered in the `boot` method of your service provider. Calls to `afterResolving` should be made in `register`.

pxpm:
@GrahamCampbell just a note, why would the caller need to know if it's already resolved or not? It's the container responsibility to know that, if it's resolved apply the callback since it's an: **afterResolving** callback.

pxpm:
> @pxpm but the core issue is you are registering Blade extensions in the service provider `register` method. There is no guarantee that Blade is even available in the container at that point in the application lifecycle. If the order of service provider registration were to change internally within Laravel - your application would break.
> 
> The real solution to this is to only use `register` for _registering_ services in the container. Do not resolve or interact with any bound services otherwise you are going to have these problems. Maybe I can make that clearer in the docs. Your Blade extensions should be registered in the `boot` method of your service provider. Calls to `afterResolving` should be made in `register`. Is there a reason you can't register your Blade extensions in `boot`?

Thanks again @taylorotwell . Moving to boot() has the same effect. It's not respect if already resolved!

erikn69:
I think, if one of the packages has that on register method, it breaks for others, no matter they have it in the correct method

pxpm:
> I think, if one of the packages has that on register method, it breaks for others, no matter they have it in the correct method

Just to be 100% clear here: Does not matter where you add your afterResolving boot/register. It's not respect internally in the container. 

And there is no reason for that in my opinion, since it's an afterResolving, we can safely run it if it's already resolved. 

Thanks, I know you also spend ALOT of time in this issue. 🙏

GrahamCampbell:
@pxpm You are calling the wrong method. You need to call `$this->callAfterResolving`. If we merge your PR, then we break everything for people doing this right, including the framework code itself. Notice the `loadViewComponentsAs` method calls this. If we merge your PR, the callbacks run twice now, instead of once.

taylorotwell:
@GrahamCampbell is correct that `$this->callAfterResolving(fn)` already does exactly what you want. But, again, don't register Blade extensions in the `register` method.

pxpm:
> @pxpm You are calling the wrong method. You need to call `$this->callAfterResolving`. If we merge your PR, then we break everything for people doing this right, including the framework code itself. Notice the `loadViewComponentsAs` method calls this. If we merge your PR, the callbacks run twice now, instead of once.

Didn't test yet, but seems you are right. I will be back to you after dinner and test this. If `callAfterResolving` does the same I am attempting here, sure, I need to apologize again, and move on with my life. 

Also, IF that method fixes this, it should be properly document to use that method, because the `afterResolving` is not doing what it's supposed to. 

I guess that if you are right, someone step in the same problem that we did, but probably decided to fix it in an "unbreakable" way, and just glued a new method that does what afterResolving should do. Otherwise it should not be afterResolving at all. 

When you want to setup your package services and interactions you do it in the provider. You don't need to know if some other provider is resolving the class you want to add a callback to, if you are adding an afterResolving callback you just expect it to be run. 

Back to you in a second cause food is getting cold. 

🙏 Thanks

pxpm:
> @pxpm You are calling the wrong method. You need to call `$this->callAfterResolving`. If we merge your PR, then we break everything for people doing this right, including the framework code itself. Notice the `loadViewComponentsAs` method calls this. If we merge your PR, the callbacks run twice now, instead of once.

I guess this is wrong from the beggining. The solution found to fix the problem in 6.X is not really a fix, but a workaround for some problem people were having at that time. `afterResolving` callback should be bound to Container not to a ServiceProvider helper method. The functionality introduced in the helper method for ServiceProviders should be introduced in the container callback register methods, more specifically in this case `afterResolving`.

I can't see where is the confusion here, if the ServiceProvider helper method is already doing that, why the container does not do that by default ? 

This is what makes me confusion, can you please help me to understand this bit ? 

Why is `callAfterResolving` (call this, if resolved call in resolved instance) bound to Provider, and `afterResolving` (that should be the same as callAfterResolving) bound to Container ? 

I know you are a busy person, but if you can spare some time to answer me this, I would really appreciate you. 

Thanks in advance 🙏

PR #41398, opened 2022-03-08
[9.x] Add empty directory assertion to `Storage::fake()`

JayBizzle:
I was trying to assert that a given directory was empty in some tests and was surprised to see that there was no bulit-in assertion to handle this.

This PR proposes adding an `assertDirectoryEmpty($path)` method to the Storage fake.

Obviously this is just a cleaner version of doing...
```php
$this->assertEmpty(Storage::disk('temp')->allFiles('/foo'));
```
vs

```php
Storage::disk('temp')->assertDirectoryEmpty('/foo');
```

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemAdapter.php
+++ b/src/Illuminate/Filesystem/FilesystemAdapter.php
@@ -149,6 +149,21 @@ class FilesystemAdapter implements CloudFilesystemContract
     }
 
     /**
+     * Determine if a directory is empty.
+     *
+     * @param  string  $path
+     * @return $this
+     */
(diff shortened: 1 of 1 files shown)


taylorotwell:
What is the behavior if there are no files within the directory but there are other sub-directories?

JayBizzle:
> What is the behavior if there are no files within the directory but there are other sub-directories?

The assertion will fail, correctly IMO as the directory is not empty if it contains other folders

```php
Storage::fake('temp');

Storage::disk('temp')->put('/foo/bar.txt', 'string');

Storage::disk('temp')->assertDirectoryEmpty('/'); // fail
```

PR #41401, opened 2022-03-09
[9.x] Update AuthServiceProvider variable names and docs

imanghafoori1:
A tiny bit of modernazation of AuthServiceProvider for doc-blocks and variable names.


driesvints:
It's not necessary imo to do this everywhere. just in the places where it makes sense.

PR #41409, opened 2022-03-09
[9.x] Make assertPath() accepts Closure

villfa:
Hi,

I propose to modify `AssertableJsonString::assertPath()` in order to ease the way to write tests:
```php
$this
    ->getJson('/api/whatever')
    ->assertOK()
    ->assertJsonPath('data.created_at', fn($date) => $date !== null)
    ->assertJsonPath('data.name', fn($name) => strlen($name) > 0)
;
```

At first I thought of adding a new method, but after struggling to find a right name, I realized that accepting closures in `assertPath` would not cause any BC break.

### DIFF
--- a/src/Illuminate/Testing/AssertableJsonString.php
+++ b/src/Illuminate/Testing/AssertableJsonString.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Testing;
 
 use ArrayAccess;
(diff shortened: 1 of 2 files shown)


BrandonSurowiec:
I think it is an awesome addition.

PR #41412, opened 2022-03-09
[9.x] Fixes HTTP::pool response when a URL returns a null status code

### LINKED ISSUE #41360: HTTP::pool response returned without the PSR response when accessing a URL with an expired SSL cert.
- Laravel Version: v9.3.1
- PHP Version: 8.0.12

### Description:
When using the HTTP::pool to access a site with an expired SSL cert, I was expecting a GuzzleHttp\Exception\RequestException to be returned, but instead we get a Illuminate\Http\Client\Response but it is missing the internal PSR response, so accessing any of the response details fails.

### Steps To Reproduce:

```
use Illuminate\Http\Client\Pool;
	
    $pool->get('https://expired.badssl.com')
]);

$responses[0]->status();
```

I would expect the $response[0] to be a type of GuzzleHttp\Exception\RequestException but we get a valid looking Illuminate\Http\Client\Response

The call to ->status() results in an error because the Illuminate\Http\Client\Response internal PSR response property is null:
**Call to a member function getStatusCode() on null in ....**

At present my solution to detect this failure is to check for a null return on the response->toPsrResponse() and then check for a non zero in the handlerStats ssl_verify_result, but this is fragile as it depends on using the curl supplied ssl_verify_result.

```
...

if ($responses[0]->toPsrResponse() === null && $responses[0]->handlerStats()['ssl_verify_result']) {
   // SSL error occurred during request
}
```

Any help or pointers to more detail would be greatly appreciated!

sulemaanhamza:
This PR fixes the bug reported by [sfilzek](https://github.com/sfilzek).

When we try to use HTTP::pool with a URL having an expired certificate, Guzzle returns a RequestException with a null response. I've added an extra check to make sure if the guzzle response is not null then populateResponse otherwise returns the exception.

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -790,7 +790,7 @@ class PendingRequest
                 });
             })
             ->otherwise(function (TransferException $e) {
-                return $e instanceof RequestException ? $this->populateResponse(new Response($e->getResponse())) : $e;
+                return $e instanceof RequestException && $e->getResponse() ? $this->populateResponse(new Response($e->getResponse())) : $e;
             });
     }
 
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -790,7 +790,7 @@ class PendingRequest
                 });
             })
             ->otherwise(function (TransferException $e) {
-                return $e instanceof RequestException && $e->getResponse() ? $this->populateResponse(new Response($e->getResponse())) : $e;
+                return $e instanceof RequestException && $e->hasResponse() ? $this->populateResponse(new Response($e->getResponse())) : $e;
             });
     }
 


src/Illuminate/Http/Client/PendingRequest.php
```diff
@@ -790,7 +790,7 @@ protected function makePromise(string $method, string $url, array $options = [])
                 });
             })
             ->otherwise(function (TransferException $e) {
-                return $e instanceof RequestException ? $this->populateResponse(new Response($e->getResponse())) : $e;
+                return $e instanceof RequestException && $e->getResponse() ? $this->populateResponse(new Response($e->getResponse())) : $e;
```
bogdankharchenko, line 793:
```suggestion
                return $e instanceof RequestException && $e->hasResponse() ? $this->populateResponse(new Response($e->getResponse())) : $e;
```

I think this should also work

sulemaanhamza: Thank you @bogdankharchenko, Good catch!

shinsenter:
@taylorotwell 
I encountered a similar error as above, but the returned result was not a `RequestException` but a `ConnectException`. I think it would be best to replace the `RequestException` of the  following code snippet with `TransferException`:
https://github.com/sulemaanhamza/framework/blob/20f8658355c5b1c843a61c1cdfa249f7da515841/src/Illuminate/Http/Client/PendingRequest.php#L791

PR #41414, opened 2022-03-09
[9.x] Added callable support to operatorForWhere on Collection

eusonlito:
Added `callable` support to `operatorForWhere`.

Currently only Collection methods `where` and `whereFirst` are used without `callable` support.

Adding callable to `operatorForWhere` allow complex usages as:

```php
$city = $cities->firstWhere(fn ($city) => $city->state->available && ($city->state->id === $stateId));
```
or
```php
$state = $states->firstWhere(fn ($state) => (bool)$state->cities->firstWhere('id', $cityId));
```

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -311,7 +311,7 @@ trait EnumeratesValues
     /**
      * Get the first item by the given key value pair.
      *
-     * @param  string  $key
+     * @param  callable|string  $key
      * @param  mixed  $operator
      * @param  mixed  $value
      * @return TValue|null
@@ -548,7 +548,7 @@ trait EnumeratesValues
     /**
      * Filter items by the given key value pair.
      *
-     * @param  string  $key
+     * @param  callable|string  $key
      * @param  mixed  $operator
      * @param  mixed  $value
      * @return static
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hmm, what all methods does this affect?

eusonlito:
Only `where` and `firstWhere`.

Method `every` has a previous parameter check when only one argument:

https://github.com/laravel/framework/blob/1a1c140b91acaad0caf0dc5cf0eecc649373ebcb/src/Illuminate/Collections/Traits/EnumeratesValues.php#L294-L298

Same as `partition`:

https://github.com/laravel/framework/blob/1a1c140b91acaad0caf0dc5cf0eecc649373ebcb/src/Illuminate/Collections/Traits/EnumeratesValues.php#L455-L462

Same `sole` method on `LazyCollection`:

https://github.com/laravel/framework/blob/bf9f8f702f8f8847d95e7a6c79771ccf93370344/src/Illuminate/Collections/LazyCollection.php#L1149-L1153

`firstOrFail` on `LazyCollection`:

https://github.com/laravel/framework/blob/bf9f8f702f8f8847d95e7a6c79771ccf93370344/src/Illuminate/Collections/LazyCollection.php#L1173-L1177

And `contains` on `LazyCollection`:

https://github.com/laravel/framework/blob/bf9f8f702f8f8847d95e7a6c79771ccf93370344/src/Illuminate/Collections/LazyCollection.php#L215-L218

taylorotwell:
Would it be safer to just modify `where` and `firstWhere` directly instead of this method just to be sure?

eusonlito:
@taylorotwell I will add a new check, because `is_callable` can get a string (like `trim`) and return `true`.

I will send a new commit changing `if (is_callable($key)) {` with `if ($this->useAsCallable($key))`.

Also a new test will be added using a callable string.

sebdesign:
What's the difference from the `first` method?

```php
$city = $cities->first(
    fn ($city) => $city->state->available && ($city->state->id === $stateId)
);
```

PR #41415, opened 2022-03-09
[9.x] Allow to run a number of migrations at a time

vanyamil:
This would make the `php artisan migrate` command be complementary to `php artisan migrate:rollback` by allowing to control the maximum number of migrations ran at a time. 
This change is (almost) backwards compatible - not supplying the --step option runs it as usual, supplying --step true runs as --step does currently.
What it allows is doing --step N, where N is an integer; then, this number of migrations runs at most.
This is useful for releases where multiple migrations may be present but testing shows there may be issues to fix before reaching production. Particularly so given that by default, some migrator grammars do not run migrations as transactions.
The current changes modify the command and the migrator behaviour, but NOT the tests. I'm not certain what tests to write to verify the possible options, and to verify that it in fact can run only some migrations vs all.

### DIFF
--- a/src/Illuminate/Database/Migrations/Migrator.php
+++ b/src/Illuminate/Database/Migrations/Migrator.php
@@ -156,7 +156,14 @@ class Migrator
 
         $pretend = $options['pretend'] ?? false;
 
-        $step = $options['step'] ?? false;
+        // We will also decide whether a limited number of migrations should run,
+        // based on the "step" option passed. Expected allowed values are strings
+        // "false", "true" or integer strings.
+        $step = $options['step'] ?? "false";
+        $increment = $step !== "false";
+        $batch_limit = $increment && is_numeric($step) && ($step = intval($step)) > 0 
+            ? $batch + $step
+            : PHP_INT_MAX;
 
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Migrations/Migrator.php
```diff
@@ -156,17 +156,28 @@ public function runPending(array $migrations, array $options = [])
 
         $pretend = $options['pretend'] ?? false;
 
-        $step = $options['step'] ?? false;
+        // We will also decide whether a limited number of migrations should run,
+        // based on the "step" option passed. Expected allowed values are strings
+        // "false", "true" or integer strings.
+        $step = $options['step'] ?? "false";
+        $increment = $step !== "false";
```
bogdankharchenko, line 163: 🤔

vanyamil:
That is ugly, I agree :-D unfortunately couldn't figure out how to
guarantee type safety on the options, to cast to bool or int, and fail
otherwise. I know config or env helpers do it, but wasn't sure how to best
approach it. And all option values seemed to be strings.

taylorotwell:

PR #41417, opened 2022-03-10
Adds Str::end() adds the ability to end a string with a single instance of a given value.

josezenem:
This feature is the inverse of Str:start( https://laravel.com/docs/9.x/helpers#method-str-start, however it does it at the end.

ie: 

```php
Str::end('test/string', '/') 
// outputs test/string/

str('test/string')->end( '/')->value()
// outputs test/string/
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -271,6 +271,20 @@ class Str
     }
 
     /**
+     * End a string with a single instance of a given value.
+     *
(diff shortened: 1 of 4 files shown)


SjorsO:
This method already exists:
```
Str::finish($value, $cap);
```

josezenem:
@SjorsO thank you, I missed it due to the naming convention, noticed startWith, endsWidth, figured this would be end().

Closing this!

PR #41418, opened 2022-03-10
[9.x] Add null typing to connection property in \Illuminate\Database\Eloquent\Factories\Factory

jnoordsij:
When extending the Factory in my application, using a null coalescing operator on `$this->connection` and running static analysis using Larastan, I got an error `Property Illuminate\Database\Eloquent\Factories\Factory<Illuminate\Database\Eloquent\Model>::$connection (string) in isset() is not nullable.` 

However, the constructor by defaults sets `$this->connection = null;`. Similarly, there are some other nullable properties. This PR fixes the docstring typehints on these properties.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/Factory.php
+++ b/src/Illuminate/Database/Eloquent/Factories/Factory.php
@@ -41,35 +41,35 @@ abstract class Factory
     /**
      * The state transformations that will be applied to the model.
      *
-     * @var \Illuminate\Support\Collection|null
+     * @var \Illuminate\Support\Collection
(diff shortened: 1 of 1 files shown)


driesvints:
Most of these can never be `null`. They're always filled in through the object's constructor. I don't think we should adjust this for extending classes.

jnoordsij:
Ah nevermind, I was being too quick to judge about most of them. Only the `connection` can really be `null` after the constructor. I'll adjust them. 

Might as well use null coalescing operators in the constructor for the others?

PR #41419, opened 2022-03-10
[9.x] Handle when $haystack is `null` in Str::contains

hmazter:
Fixes this deprecation message from `str_contains`:
```
str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated
```

The deprecation warning appears when running this in a fresh install:
```
LOG_DEPRECATIONS_CHANNEL=stderr php artisan route:list --name foo
```

I opted to place it first in the method since there is no need for anything else to run.
This also has the bonus to fix the potential deprecation warning of passing `null` to `mb_strtolower`

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -211,6 +211,10 @@ class Str
      */
     public static function contains($haystack, $needles, $ignoreCase = false)
     {
+        if ($haystack === null) {
+            return false;
+        }
+
         if ($ignoreCase) {
             $haystack = mb_strtolower($haystack);
             $needles = array_map('mb_strtolower', (array) $needles);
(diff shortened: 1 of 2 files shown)


hmazter:
Okay. The problem, in this case, is these `Str::contains` checks.

https://github.com/laravel/framework/blob/692a634cba2c4b402734e7d0549bdf5328f0ccac/src/Illuminate/Foundation/Console/RouteListCommand.php#L243-L249

Not really sure what the best approach is to fix that

GrahamCampbell:
Cast those to a string, probably, before calling.

GrahamCampbell:
Likely this bug is also present on 8.x, so needs fixing there too?

driesvints:
This method does not allow passing a string. Typecast the value before passing.

hmazter:
I have a failing test for it here at least https://github.com/hmazter/framework/commit/efc8bc8ffb5831e4506bb2a0dddd513ee882d94d
Not sure what to do with it though

PR #41422, opened 2022-03-10
[9.x] Implement email "metadata" for SesTransport

kbond:
This adds metadata support for the SesTransport. This was added in Symfony core (https://github.com/symfony/symfony/pull/45222) but won't be available until 6.1.

SES calls these "tags" but by our definition, they are metadata. SES does not have the concept of "tags" by our definition.


driesvints (approved):
Lgtm!

taylorotwell:
@kbond can you point me to where "tags" and "metadata" are clearly defined from Symfony mailer's perspective? What is the purpose of a tag and what is the purpose of metadata? It feels like it would be great to have documentation on what each type (tag / metadata) does for each driver.

PR #41427, opened 2022-03-10
[9.x] Added `@return` statement to Blade

eusonlito:
This statement adds a `<?php return; ?>` on compiled view to allow to stop render execution on some point and avoid nested code and simplify templates.

For example, `pagination` templates can be easily updated with:

```php
@unless ($paginator->hasPages())
    @return
@endunless

<nav>
    <ul class="pagination">
```

Another addition (not in this commit) could to be add two new statements: `@returnUnless` and `@returnIf`

### DIFF
--- a/tests/View/Blade/BladePhpStatementsTest.php
+++ b/tests/View/Blade/BladePhpStatementsTest.php
@@ -22,6 +22,13 @@ class BladePhpStatementsTest extends AbstractBladeTestCase
         $this->assertEquals($expected, $this->compiler->compileString($string));
     }
 
+    public function testReturnStatements()
+    {
+        $string = '@return';
+        $expected = '<?php return; ?>';
+        $this->assertEquals($expected, $this->compiler->compileString($string));
+    }
+
     public function testPhpStatementsDontParseBladeCode()
     {
         $string = '@php echo "{{ This is a blade tag }}" @endphp';
--- a/src/Illuminate/View/Compilers/Concerns/CompilesRawPhp.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesRawPhp.php
@@ -20,6 +20,16 @@ trait CompilesRawPhp
     }
 
     /**
+     * Compile the return statements into valid PHP.
+     *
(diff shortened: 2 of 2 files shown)


dennisprudlo:
Is this a common practice? I see a lot of downsides when stopping script execution whilst rendering a view. This could only really be useful when you have a view without any nested elements.

```html
<!-- some-view.blade -->
<div>
    <strong>
        @return
    </strong> <!-- never gets rendered, will have to rely on the browser to close this -->
</div> <!-- never gets rendered, will have to rely on the browser to close this -->
```
This would result in all of the following being bold, because the `strong` tag never got closed.

I know you said it's supposed to remove nesting, but IMO an if-wrapped element has better readability:
```html
@if ($paginator->hasPages())
    <nav> ... </nav>
@endif
```

And when there's really a use case where it's necessary to stop further rendering there's always:
```html
@unless ($paginator->hasPages())
    @php return @endphp
@endunless

<nav>
    <ul class="pagination">
```

tests/View/Blade/BladePhpStatementsTest.php
```diff
@@ -22,6 +22,13 @@ public function testPhpStatementsWithoutExpressionAreIgnored()
         $this->assertEquals($expected, $this->compiler->compileString($string));
     }
 
+    public function testReturnStatements()
+    {
+        $string = '@return';
+        $expected = '<?php return; ?>';
+        $this->assertEquals($expected, $this->compiler->compileString($string));
```
gguney, line 29:
```suggestion

        $this->assertEquals($expected, $this->compiler->compileString($string));
```

For assertion readability, below codes did that too.

eusonlito: Has the same format as existing methods.

eusonlito:
@dennisprudlo Yes, you can write bad code anyway, and yes, you can write PHP code everywhere.

eusonlito:
Closing due low popularity 😅😅

I will continue using my custom `return` blade directive 😛

PR #41432, opened 2022-03-11
[9.x] Use parents to resolve middleware priority

netpok:
This PR was already merged in 8.x but was later reverted because it broke the laravel-impersonate package. In commit https://github.com/laravel/framework/commit/50b46db563e11ba52a53e3046c23e92878aed395 an interface was added for the AuthenticateSession middleware and it looks like that behavior isn't considered a problem anymore so I opened this MR again.

Basically this allows extending any middleware while preserving the sort order. For example extending `\Illuminate\View\Middleware\ShareErrorsFromSession` will still execute it before authenticating the requests.

While this PR makes https://github.com/laravel/framework/commit/50b46db563e11ba52a53e3046c23e92878aed395 obsolete I don't think that the interface should be removed.

---

Original PR message:

Currently when middlewares are ordered it uses the class name and its interfaces for sorting but not it's parents.

This is a problem for example in Jetstream because:

Jetstream extends the AuthenticateSession middleware with it's own.
Since it doesn't have a contract it will run before any Authenticate middleware
Resulting in it's always being initialized via the default guard instead of the authenticated one
This prevents using two guards simultaniously (for example a user and an admin guard)
This PR fixes this behaviour by using parent classes for ordering so the Jetstream AuthenticateSession middleware will be in the correct place because it extends the Laravel AuthenticateSession middleware.

I also added test for the parent sorting and the already present interface sorting.

### DIFF
--- a/src/Illuminate/Routing/SortedMiddleware.php
+++ b/src/Illuminate/Routing/SortedMiddleware.php
@@ -101,6 +101,14 @@ class SortedMiddleware extends Collection
                 yield $interface;
             }
         }
+
+        $parents = @class_parents($stripped);
+
+        if ($parents !== false) {
+            foreach ($parents as $parent) {
+                yield $parent;
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
Since we fixed the problem in Jetstream I think I'll hold off on merging this in 9.x since it could break packages like before.

PR #41436, opened 2022-03-11
[9.x] Resolve Markdown converter from container instead of directly instantiating it

ralphjsmit:
I'm putting the last hand at converting my blog from WordPress to Laravel (yeah!). To render blog content, I'm using the `Str::markdown()` function. 

However, I ran into the use case where I wanted to add a renderer (or extension, or whatever) to the League Commonmark environment.

Currently this is not possible, as the implementation only allowed the passing of an `$options` array to the converter. The `$options` array doesn't support adding renderers or extensions.

The solution for this is quite simple: instead of instantiating the `GithubFlavoredMarkdownConverter` directly, we can also resolve it from the container. By resolving it from the container (or was it called service location?), we opened up the opportunity to use the `->afterResolving()` function in a service provider.

This makes it very easy for developers to extend the Commonmark package or add an extension, without having to install and configure the complete Commonmark package themselves:

```php
$this->app->afterResolving(GithubFlavoredMarkdownConverter::class, function (GithubFlavoredMarkdownConverter $converter) {
    $environment = $converter->getEnvironment();

    $environment->addRenderer(Link::class, new ExternalLinkRenderer());
    // Do other things with the environment
});
```

(The use case here was to determine for every URL whether to add a `target="_blank"` attribute, based on whether the URL contains the `Request::root()`).

Thanks!

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -471,7 +471,11 @@ class Str
      */
     public static function markdown($string, array $options = [])
     {
-        $converter = app(GithubFlavoredMarkdownConverter::class, $options);
+        if (function_exists('app')) {
+            $converter = app(GithubFlavoredMarkdownConverter::class, $options);
+        } else {
+            $converter = new GithubFlavoredMarkdownConverter($options);
+        }
 
         return (string) $converter->convert($string);
     }
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -471,7 +471,7 @@ class Str
      */
     public static function markdown($string, array $options = [])
     {
-        $converter = new GithubFlavoredMarkdownConverter($options);
+        $converter = app(GithubFlavoredMarkdownConverter::class, $options);
 
         return (string) $converter->convert($string);
     }


GrahamCampbell:
This don't look right to me. The support package doesn't know about the app foundation function.

GrahamCampbell:
If you need a much more custom markdown wiring, a package exists (which actually pre-dates markdown being in the framework): https://github.com/GrahamCampbell/Laravel-Markdown/tree/14.0.

ralphjsmit:
> This don't look right to me. The support package doesn't know about the app foundation function.

Because you can get also get the `illuminate/support` in a non-Laravel project?

ralphjsmit:
> If you need a much more custom markdown wiring, a package exists (which actually pre-dates markdown being in the framework): https://github.com/GrahamCampbell/Laravel-Markdown/tree/14.0.

Thanks for sharing it! The thing is that I already knew about your package and the underlying Commonmark package. I tried both, but it felt somewhat overkill and hard to setup. Allowing this would make it much easier for the casual markdown-user, who does not need to setup a full environment

Plus, this won't work well in this situation, because the `Str::markdown()` is used in an underlying package and not the main application itself.


You're the markdown-expert, so would there be an other way to retrieve the environment? Is that a singleton in the Commonmark package or does the Commonmark package emit events / hooks?

taylorotwell:
Graham's correct that we can't call the global container easily from here. If you need a custom Markdown I recommend just interacting with CommonMark directly. Thanks! 🙂

PR #41442, opened 2022-03-11
[9.x] Add ability to validate route parameters via callback

stevebauman:
## Description

This PR provides the ability to add callbacks to validate route parameters, instead of regex only.

For people (like myself) who haven't dug deep into writing regex, this helps tremendously in validating complex parameters in pure PHP, and will help newcomers to Laravel avoid writing confusing and possibly error prone regex.

## Usage

```php
Route::get('/products/{brand}', '...')->where('brand', function ($brand) {
    return in_array($brand, ['samsung', 'apple']);
});
```

```php
Route::get('/invitation/{token}', '...')->where('token', function ($token) {
    return strlen($token) === 30;
});
```

```php
Route::get('/license/{key}', '...')->where('key', function ($key) {
    $str = Str::of($key);

    return $str->length() === 32
        && $str->startsWith('1234-')
        && $str->endsWith('-4567');
});
```

## Important

This PR is not set in stone. I'm not sure if `Route::$callbacks` is the best property name to contain these "validators" (which has already been taken unfortunately).

Please let me know your thoughts, thanks for your time! 🙏 

No hard feelings on closure ❤️

### DIFF
--- a/src/Illuminate/Routing/AbstractRouteCollection.php
+++ b/src/Illuminate/Routing/AbstractRouteCollection.php
@@ -28,7 +28,7 @@ abstract class AbstractRouteCollection implements Countable, IteratorAggregate, 
      */
     protected function handleMatchedRoute(Request $request, $route)
     {
-        if (! is_null($route)) {
+        if (! is_null($route) && $route->matches($request)) {
             return $route->bind($request);
         }
 
@@ -144,6 +144,7 @@ abstract class AbstractRouteCollection implements Countable, IteratorAggregate, 
                 'fallback' => $route->isFallback,
                 'defaults' => $route->defaults,
                 'wheres' => $route->wheres,
+                'callbacks' => $route->callbacks,
                 'bindingFields' => $route->bindingFields(),
                 'lockSeconds' => $route->locksFor(),
                 'waitSeconds' => $route->waitsFor(),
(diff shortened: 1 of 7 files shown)


amaelftah:
❤️

taylorotwell:
Hey @stevebauman ...

This will properly return a 404; however, after running `route:cache` it will be ignored and the user can access the route:

image

stevebauman:
Ah okay, let me make sure it works as intended after caching. I’ll update this tomorrow and mark ready for review once complete 👍

src/Illuminate/Routing/AbstractRouteCollection.php
```diff
@@ -28,7 +28,7 @@ abstract class AbstractRouteCollection implements Countable, IteratorAggregate,
      */
     protected function handleMatchedRoute(Request $request, $route)
     {
-        if (! is_null($route)) {
+        if (! is_null($route) && $route->matches($request)) {
```
stevebauman, line 31: This may be performing unnecessary "re-validations" for route matching.

I'm not sure if we should add something like a `$route->passes($request)` method for executing *only* the callback validator? Let me know if you have an opinion on this 👍

stevebauman: It was necessary to execute the `CallbackValidator` to deny the route and throw the `NotFound` HTTP exception. Otherwise, the route matches and bypasses the validator.

driesvints:
@stevebauman feel free to resubmit this when you have time.

stevebauman:
Will do, thanks Dries!

PR #41443, opened 2022-03-11
[9.x] Allow passing a method name in mapInto()

denniseilander:
This pr adds an extra argument to the `mapInto()` helper to pass in a method name. 
This allows you to map data into a class via the given method name.

### Use case
For example, when you are dealing with transformers that have a method instead of a constructor that accepts certain data. You receive some data from an API and want to convert that data into a valueObject.

Current way to do this is with the `map()` helper or  the `array_map` function:
```php
$data->map(fn(array $item) => MyTransformer::transform($item));

array_map(fn($item) => MyTransformer::transform($item), $data);
```
This is fine, but it would be easier to use the `mapInto()` method which is meant to map data into an object.
So with this addition, you can now set a second argument which contains the method name of the class to map the data into.

It works both with static and non-static methods

### Example
```php
$data->mapInto(MyTransfomer::class, 'transform');
```

```php
class MyTransfomer
{
    public function transform($data)
    {
        return [
            ...
        ]
    }
}
```

### Breaking changes
The added argument is optional and `null` by default, so there should be no breaking changes.

### DIFF
--- a/src/Illuminate/Collections/Enumerable.php
+++ b/src/Illuminate/Collections/Enumerable.php
@@ -689,9 +689,10 @@ interface Enumerable extends Arrayable, Countable, IteratorAggregate, Jsonable, 
      * @template TMapIntoValue
      *
      * @param  class-string<TMapIntoValue>  $class
+     * @param  string|null  $method
      * @return static<TKey, TMapIntoValue>
      */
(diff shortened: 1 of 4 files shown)


taylorotwell:
See above.

PR #41444, opened 2022-03-11
[9.x] Define a new method to create a through model

hafezdivandari:
It is helpful to have a customizable method for creating a new model instance for a through model without overriding the whole [`hasOneThrough`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L116-L118) or [`hasManyThrough`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L388-L390) methods. It is just as useful as [`newRelatedInstance`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L753-L760) method.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
@@ -115,7 +115,7 @@ trait HasRelationships
      */
     public function hasOneThrough($related, $through, $firstKey = null, $secondKey = null, $localKey = null, $secondLocalKey = null)
     {
-        $through = new $through;
+        $through = $this->newThroughInstance($through);
 
         $firstKey = $firstKey ?: $this->getForeignKey();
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you describe your use case for this?

hafezdivandari:
@taylorotwell sure, I wanted to pass the table name instead of the class and set the through model's table dynamically, something like this:

```php
protected function newThroughInstance($resource)
{
    return (new \App\Models\Entity)->setTable($resource);
}
```

As I said in the comment above, the use case can be setting the model's connection or table conditionally/dynamically just like [`newRelatedInstance`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php#L753-L760) method.

PR #41445, opened 2022-03-11
[9.x] Redesign `php artisan schedule:list` Command.

xiCO2k:
This PR Improves the look for the `php artisan schedule:list` command.

The design was heavily inspired by the new version of the `route:list`.

##  Before:

image

## After

image

Also there is available a verbose mode that will show the **full cli path**, **raw dates* and **description**.

### Verbose Mode

image

Cheers,
Francisco

### DIFF
--- a/src/Illuminate/Console/Scheduling/ScheduleListCommand.php
+++ b/src/Illuminate/Console/Scheduling/ScheduleListCommand.php
@@ -98,7 +98,7 @@ class ScheduleListCommand extends Command
         });
 
         if ($events->isEmpty()) {
-            return $this->comment('No tasks have been scheduled.');
+            return $this->comment('No scheduled tasks have been defined.');
         }
 
         $this->output->writeln(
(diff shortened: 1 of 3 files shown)


xiCO2k:
Looks like the `$this->artisan(ScheduleListCommand::class, ['-v' => true])` is interfering with `RouteListCommandTest`. maybe it need to be set as default to `OutputInterface::VERBOSITY_NORMAL`.

Thoughts?

taylorotwell:
@xiCO2k ready now?

xiCO2k:
Yep

taylorotwell:
Thanks for your work on this! ❤️

flightsupport:
I think scheduled jobs should be visible without having to set the -v flag.
With many scheduled jobs, the output is not meaningful.

Output without -v
`*/5 * * * *  ................................................................. Next Due: 3 minutes from now`

Output with -v
```
*/5 * * * *  .................................................. Next Due: 2022-03-15 20:35:00 +00:00
               ⇁ App\Jobs\ShiftSupervisor

```

driesvints:
@flightsupport please see https://github.com/laravel/framework/pull/41494

xiCO2k:
It is not that much clear a Job, but I will think on a way to show the job correctly like a command schedule. Thanks for pointing that out @flightsupport

PR #41447, opened 2022-03-11
[9.x] Fix the guard instance used.

selcukcukur:


driesvints:
We reverted this because it broke Jetstream: https://github.com/laravel/jetstream/issues/1000

PR #41450, opened 2022-03-12
[9.x] Adds collection key first/last.

DarkGhostHunter:
## What?

Adds methods to retrieve the first and last key of the `Collection` instance.

```php
// Before
array_key_first($collection->all());

// After
$collection->firstKey();
```

## Why?

Sometimes you want to get the last key of the collection, or get which key the last item added has. Or when unshifting (prepending) the collection.

## BC?

None, not expected the user to have a macro of the same name.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -390,9 +390,9 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
-     * Returns the first key of the collection.
+     * Get the first key of the collection.
      *
-     * @return int|string|void|null
+     * @return mixed
      */
     public function firstKey()
     {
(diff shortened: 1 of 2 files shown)


taylorotwell:
Reverted this since it didn't take into account LazyCollection and the collection interface.

DarkGhostHunter:
> Reverted this since it didn't take into account LazyCollection and the collection interface.

Should this execute the Lazy Collection if implemented there? I didn't consider it right because using this would resolve the Lazy Collection.

mikeydevelops:
Yes that would resolve it, but that's what `first()` and `last()` methods already do on the `LazyCollection` .

axlon:
Whats wrong with simply doing `->keys()->first()`, this would work on a lazy collection as well, and if you're using a normal collection the array likely isn't large enough where calling `->keys()` first matters anyway. Don't really see the benefit of adding 2 more functions for existing functionality (just my 2 cents)

DarkGhostHunter:
IMO shouldn’t be for Lazy, because that would resolve all the keys. Is that an accepted behavior?

Italo Baeza C.

Issue #41437, opened 2022-03-11
Route list sort option "precedence" does not work anymore


maarsson:
- Laravel Version: 9.4.1
- PHP Version: 8.1.0
- Database Driver & Version: irrelevant

### Description:
The very handy feature sorting routes by their precedence by the `artisan route:list --sort=precedence` does not work anymore.

### Steps To Reproduce:
1. Install a fresh Laravel project
2. Run command `php artisan route:list --sort=precedence`
3. See error exception

sedattcom:
Accepted values are host, method, uri, name, action or middleware.

epixian:
Is `precedence` spelled correctly on the commandline?

selcukcukur:
Sorting by precedence is currently not available.

maarsson:
I spelled with an extra "n" consistently here, thanks @epixian. Issue is updated to the correct `precedence`.
Anyways the problem is still exists. I also took a closer look: it was working with framework 9.0.2 and stopped working from 9.1.0

driesvints:
This value isn't an accepted one.

PR #41451, opened 2022-03-12
Fix sorting by precedence

maarsson:
This PR provides a real fix of the issue https://github.com/laravel/framework/issues/41437
The PR https://github.com/laravel/framework/pull/41448 only "validates" the disappearing of the function. 

I think the route list should have the sorting method by `precedence`.
This update keeps the default sorting by `uri` (if no sort option was given) and also prevents `Undefined array key` errors (if not a valid sort option was given)


taylorotwell:
Isn't precedence the default?

maarsson:
No, I don´t think so

PR #41455, opened 2022-03-12
[9.x] Add new wrap string helper

hebinet:
This PR adds a new `wrap` string helper.

**Usage**
```php
Str:wrap('value')->wrap('"');
Str::of('value')->wrap('"');
str('value')->wrap('"');

// Outputs: "value"

Str:wrap('is', 'This ', ' me!');
Str::of('is')->wrap('This ', ' me!');
str('is')->wrap('This ', ' me!');

// Outputs: This is me!
```

**Why do I think this is usefull?**
Lately I had a Stringable class and modified its value like this:
```php
$string = Str::of($path)
  ->lower()
  ->replaceLast('/')
  ->replaceFirst($host);
```
And I couldn't incorporate the wrapping fluently and had to do something like this:
```php
$string = '"'.$string.'"';
```
...to further modify the string.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -463,23 +463,6 @@ class Str
     }
 
     /**
-     * Wraps a string with the given wrappers.
-     *
-     * @param  string  $value
-     * @param  string  $before
-     * @param  string|null  $after
-     * @return string
-     */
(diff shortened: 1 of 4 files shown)


TechyMS:
I think colon(':') is missing from Str:wrap.

PR #41456, opened 2022-03-12
[9.x] Remove unnecessary empty keys in encrypted data

Krisell:
The encryption payload contains either a `mac` or a `tag` for integrity check, and the name is different depending on the encryption mode being used (`CBC` or `GCM` mode). Currently, one of `mac/tag` is always empty (when the other is used), but still present in the encrypted data.

Example of data encrypted with `GCM` mode
`{"iv":"4bSa...,"value":"lPaA...","tag":"yRrQa...","mac":""}` (note the empty `mac`)

Example of data encrypted with `CBC` mode
`{"iv":"JIIW...","value":"W/VFc...,"mac":"b53c...","tag":""}` (note the empty `tag`)

This PR removes this extra key with the empty value.

This is BC since the chosen encryption mode determines if `mac` or `tag` is read. However data encrypted with this PR will not be decryptable on older versions since the presence of the `mac` key was necessary in `validPayload`. I don't know how you usually handle such "non-rollbackable" changes, which might cause problems with aspects like queue workers reading new data on an older version (if not being restarted correctly).  I'm absolutely fine with accepting the empty keys if that is an unnecessary risk, or re-targeting this at 10.x if that is better.

An alternative approach could of course be to use the same key for both, however that would require more changes and could lead to confusion. The fact that the encryption payload reveals the method (which it already does) is not a security problem since there is a fixed set of methods to choose from (currently 2), i.e. it only reduces brute-force complexity by a low constant factor.

### DIFF
--- a/src/Illuminate/Encryption/Encrypter.php
+++ b/src/Illuminate/Encryption/Encrypter.php
@@ -111,11 +111,11 @@ class Encrypter implements EncrypterContract, StringEncrypter
         $iv = base64_encode($iv);
         $tag = base64_encode($tag ?? '');
 
-        $mac = self::$supportedCiphers[strtolower($this->cipher)]['aead']
+        $mac = $this->isUsingAEAD()
             ? '' // For AEAD-algoritms, the tag / MAC is returned by openssl_encrypt...
             : $this->hash($iv, $value);
 
-        $json = json_encode(compact('iv', 'value', 'mac', 'tag'), JSON_UNESCAPED_SLASHES);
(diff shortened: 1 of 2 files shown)


Krisell:
I did not think if BC problems with any custom class reading the raw encryption data, expecting both keys to be present. Does this mean we should close this and that we have to "live with" these extra empty keys forever?

GrahamCampbell:
10.x may be a better target for this?

taylorotwell:
I think it's fine as is for now as long as there are no security issues.

PR #41459, opened 2022-03-13
[9.x] Don't report renderable exception when $dontReport is true

avosalmon:
Currently, if we don't want to report a [renderable exception](https://laravel.com/docs/9.x/errors#renderable-exceptions), we need to define an empty `report()` method like below.

```php
class CustomException extends Exception
{
    /**
     * Report the exception.
     *
     * @return bool|null
     */
    public function report()
    {
        //
    }
 
    /**
     * Render the exception into an HTTP response.
     *
     */
    public function render($request)
    {
        return response(...);
    }
}
```

Alternatively, we can add the exception to the `$dontReport` array in the exception handler. However, as the application grows, this array can become very long.

This PR will allow us to mark a renderable exception "unreportable" by setting the exception's `$dontReport` property to `true` without having to define an empty `report()` method.

```php
class CustomException extends Exception
{
    // This exception won't be reported.
    public $dontReport = true;
 
    /**
     * Render the exception into an HTTP response.
     *
     */
    public function render($request)
    {
        return response(...);
    }
}
```

### DIFF
--- a/src/Illuminate/Foundation/Exceptions/Handler.php
+++ b/src/Illuminate/Foundation/Exceptions/Handler.php
@@ -270,6 +270,10 @@ class Handler implements ExceptionHandlerContract
      */
     protected function shouldntReport(Throwable $e)
     {
+        if (property_exists($e, 'dontReport') && $e->dontReport === true) {
+            return true;
+        }
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hmm, I don't mind the empty report method too much tbh. Thanks for the suggestion though!

PR #41463, opened 2022-03-13
[9.x] Update PackageManifest::$vendorPath initialisation for cases, when composer vendor dir is not in project directory

vasymus:
It's possible, that for some reason, vendor dir could be in another directory (https://getcomposer.org/doc/03-cli.md#composer-vendor-dir).

For example, when you change vendor dir in docker image for easier app bind mounting (https://github.com/BretFisher/php-docker-good-defaults/blob/master/Dockerfile).

In that case, if you run `php artisan package:discover` generated bootstrap/cache/packages.php file will contain empty array.

That's why those changes should be accepted, so even if COMPOSER_VENDOR_DIR is changed package discover working correct.

Those changes are backward compatible, because it's assumed that nobody would set or change global environment variable `COMPOSER_VENDOR_DIR` without reason.

### DIFF
--- a/src/Illuminate/Foundation/PackageManifest.php
+++ b/src/Illuminate/Foundation/PackageManifest.php
@@ -55,7 +55,7 @@ class PackageManifest
         $this->files = $files;
         $this->basePath = $basePath;
         $this->manifestPath = $manifestPath;
-        $this->vendorPath = getenv('COMPOSER_VENDOR_DIR') ?: $basePath.'/vendor';
+        $this->vendorPath = env('COMPOSER_VENDOR_DIR') ?: $basePath.'/vendor';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Illuminate/Foundation/PackageManifest.php
```diff
@@ -55,7 +55,7 @@ public function __construct(Filesystem $files, $basePath, $manifestPath)
         $this->files = $files;
         $this->basePath = $basePath;
         $this->manifestPath = $manifestPath;
-        $this->vendorPath = $basePath.'/vendor';
+        $this->vendorPath = getenv('COMPOSER_VENDOR_DIR') ?: $basePath.'/vendor';
```
GrahamCampbell, line 58: `getenv` is not a good function to use (because it's not threadsafe, so can leak variables across different websites running on the same server, or resolve to nothing, even after `putenv` is called). We should use `env` or just `$_SERVER`.

GrahamCampbell (changes requested):

src/Illuminate/Foundation/PackageManifest.php
vasymus: Thanks! Made changes.

PR #41405, opened 2022-03-09
[9.x] Added callback support on implode Collection method.

eusonlito:
Allow to use a callback as `$value` on `implode` Collection method to simplify -`>map()->implode()` calls:

Before:

```php
<span>{{ $user->cities->map(fn ($city) => $city->name.' ('.$city->state->name.')')->implode(', ') }}</span>
```

After:

```php
<span>{{ $user->cities->implode(fn ($city) => $city->name.' ('.$city->state->name.')', ', ') }}</span>
```

PR #41468, opened 2022-03-14
[9.x] Fix `Collection::implode()` string argument

### LINKED ISSUE #41466: [9.x] calling Collection::implode('_') throws error
- Laravel Version: 9.x-dev
- PHP Version: 8.0.16

### Description:
The following pull request broke parts of our code.
https://github.com/laravel/framework/pull/41405

### Steps To Reproduce:

```php
collect([1,2,3])->implode('_')
```

Throws the following error.

```
_() expects exactly 1 argument, 2 given 
```

This is because '_' is a function defined by the gettext extension.

Checking with `is_callable` is probably not suitable here.

derekmd:
Fixes https://github.com/laravel/framework/issues/41466 introduced by https://github.com/laravel/framework/pull/41405

`collect(['foo', 'bar'])->implode('_')` should be `'foo_bar'` but it instead attempts to call translation helper `_()`.

It needs the same fix as https://github.com/laravel/framework/pull/41424

For a callable arg, don't support global helpers. Only allow a Closure, invokable class, or [$object, 'method'] array pair.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -583,7 +583,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
      */
     public function implode($value, $glue = null)
     {
-        if (is_callable($value)) {
+        if ($this->useAsCallable($value)) {
             return implode($glue ?? '', $this->map($value)->all());
         }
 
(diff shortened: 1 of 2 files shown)


seriquynh:
Is this a typo mistake? You mean collect(['foo', 'bar'])->implode('_') should be 'foo_bar', don't you?
WXWorkCapture_16472344322968

driesvints (approved):

PR #41474, opened 2022-03-14
[10.x] Collection::containsManyItems

lucasmichot:
This PR introduces a new class method `containsManyItems`, very similar to `containsOneItem`.
`containsManyItems` checks that a collection contains more than one item.


src/Illuminate/Collections/Enumerable.php
```diff
@@ -591,6 +591,13 @@ public function isNotEmpty();
      */
     public function containsOneItem();
 
+    /**
+     * Determine if the collection contains many items.
+     *
+     * @return bool
+     */
+    public function containsManyItems();
```
driesvints, line 599: We can't add new methods to interfaces on minor/patch releases.

taylorotwell:

PR #41475, opened 2022-03-14
[9.x] Allows object instead of array when adding to `PendingBatch`

yoeriboven:
Currently you need to pass an array when you want to add to a batch.

```php
$batch = Bus::batch([]);
$batch->add([new Job]);
$batch->dispatch();
```
With this PR you can do this:
```php
$batch = Bus::batch([]);
$batch->add(new Job);
$batch->dispatch();
```

This is already possible if you want to add to an [already dispatched batch](https://github.com/laravel/framework/blob/9e7d31dcd4ba39376418d4cda425fb90809a7c2e/src/Illuminate/Bus/Batch.php#L165):
```php
$this->batch()->add(new Job);
```

The `add` function on `PendingBatch` wasn't tested yet either so I added a test.

### DIFF
--- a/src/Illuminate/Bus/PendingBatch.php
+++ b/src/Illuminate/Bus/PendingBatch.php
@@ -57,12 +57,12 @@ class PendingBatch
     /**
      * Add jobs to the batch.
      *
-     * @param  iterable  $jobs
+     * @param  \Illuminate\Support\Enumerable|object|array  $jobs
      * @return $this
      */
     public function add($jobs)
     {
-        foreach ($jobs as $job) {
+        foreach (Arr::wrap($jobs) as $job) {
             $this->jobs->push($job);
         }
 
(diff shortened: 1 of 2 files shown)


src/Illuminate/Bus/PendingBatch.php
```diff
@@ -57,12 +57,12 @@ public function __construct(Container $container, Collection $jobs)
     /**
      * Add jobs to the batch.
      *
-     * @param  iterable  $jobs
+     * @param  \Illuminate\Support\Enumerable|object|array  $jobs
```
GrahamCampbell, line 60: `\Illuminate\Support\Enumerable|object` is the same type as `object`. Why do you need `object` here?

GrahamCampbell (changes requested):

src/Illuminate/Bus/PendingBatch.php
yoeriboven: I copied [the docblock](https://github.com/laravel/framework/blob/9e7d31dcd4ba39376418d4cda425fb90809a7c2e/src/Illuminate/Bus/Batch.php#L158) from the `Batch` class. I can update that one too if you have a better suggestion.

PR #41485, opened 2022-03-15
[9.x] Add ability to mark a class as singleton with interface

pionl:
> Add ability to mark a class as singleton with interface instead of using service provider and.

## Why?

When we create a lot of services / actions and we are heavily using dependency injection it is better to mark the classes as singleton to speed things up. 

It takes some time to register all actions / services in service provider and the file is getting bigger. This would make marking singleton easy to write and even enables to find which classes are singletons via using IDE and find usages.

## Checklist

- [x] Unit test
- [x] Tested on real app - I have not found any performance issues (no notable difference in ms).
- [ ] Does it math Laravel ecosystem? **I think yes, there is already a lot of "magic" using interfaces**

```
class SharedStub implements IsSingleton {
    //
}
```

Thank you for your time

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -8,6 +8,7 @@ use Exception;
 use Illuminate\Contracts\Container\BindingResolutionException;
 use Illuminate\Contracts\Container\CircularDependencyException;
 use Illuminate\Contracts\Container\Container as ContainerContract;
+use Illuminate\Contracts\Container\IsSingleton;
 use LogicException;
 use ReflectionClass;
 use ReflectionException;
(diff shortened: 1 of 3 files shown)


driesvints:
> It takes some time to register all actions / services in service provider and the file is getting bigger.

Really? What use case do you have to register so much services that it slows your app down? I don't see how this is possible.

> This would make marking singleton easy to write and even enables to find which classes are singletons via using IDE and find usages.

This isn't an exact truth tbh. You could still register the classes as non-singletons. Tbh, you can already create your own interface in userland if you wanted to to achieve this.

pionl:
1. It is some time ago, but in quite large code base with heavy dependency injection and using events system the app on , while responding on events, had some performance issues. There were bigger loops (not so big) that would trigger events and Laravel would create listeners. We did a profile of the application and found most of the bottleneck was in `make` method of container. We did use singleton and this resolved the issues.
2. I've no idea how you mean this. Without extending and changing container implementation. 

Of course I can do it in service provider, but if I want to move services / actions that are heavily used then I want to mark then as singleton to prevent issues above :) There is minimal memory cost on this.

If you do not like it, we can close the PR, I've no problem with that. It was an idea how to simplify my life (and maybe others).

I don't mind and don't want to waste your time if you don't see it matching teams vision 👍

Thanks

PR #41487, opened 2022-03-15
[9.x] Add an argument to `morphToMany` and `morphedByMany` methods to pass the desired relationship name

hafezdivandari:
All `belongsTo`, `belongsToMany`, `morphTo` relationship methods have an argument to pass the desired relationship name, but `morphToMany` and `morphByMany` always use backtraces to get the name of the calling function and there is no way to pass a different name. This makes it impossible to implement a custom ways to define dynamic relationships using these 2 methods.

This PR adds an argument to both `morphToMany` and `morphedByMany` methods to pass the desired relationship name or `null` to get the caller function name as default.


GrahamCampbell:
This is a breaking change, so probably needs to go to 10.x.

driesvints:
Please send to master.

PR #41500, opened 2022-03-15
[9.x] Add some tests for singular method on Str and Stringable

reziamini:
Also, there was not any kind of test for the singular method too.


taylorotwell:
It is not our code. We use the Doctrine Inflector - therefore we don't have tests for that.

PR #41502, opened 2022-03-16
Bugfix #41302 - Add an explicit reconnect for failed Redis connections

### LINKED ISSUE #41302: Redis Connection does not reconnect when failed, causing Queue Workers to loop
- Laravel Version: v8.83.2 (but has been an issue for a long time)
- PHP Version: 8.1
- Database Driver & Version: Redis

### Description:

Originally reported in #30081 and #29969

If the Redis connection fails, workers do not realise this, and error with:

    Trying to access array offset on value of type null {"exception":"[object] (ErrorException(code: 0): Trying to access array offset on value of type null at /usr/local/app/vendor/laravel/framework/src/Illuminate/Queue/Jobs/Job.php:315)

and 

     Undefined offset: 1 {"exception":"[object] (ErrorException(code: 0): Undefined offset: 1 at /usr/local/app/vendor/laravel/framework/src/Illuminate/Queue/RedisQueue.php:269)

The code flow basically ends up calling https://github.com/laravel/framework/blob/9.x/src/Illuminate/Redis/RedisManager.php#L82 which doesn't check if the connection is still valid. As the connection NAME is still valid, but the connection itself is closed, there is no reconnect attempt (as there is with other drivers).

This is triggered when there is a socket error:

    socket error on read socket {"exception":"[object] (RedisException(code: 0): socket error on read socket at /usr/local/app/vendor/laravel/framework/src/Illuminate/Redis/Connections/Connection.php:116)

which probably THERE should unset itself, or at least mark itself as failed.

### Steps To Reproduce:

* Create a a simple job (eg, sleep 10)
* Dispatch a pile of those jobs using Redis queue
* Shut down redis, and wait for the `socket error` to occur
* Restart redis

At this point, all jobs will immediately fail with the null/undefined offset, and the only way to fix them is to kill and restart the worker.

xrobau:
When a \Redis connection has a socket/connection error, there's no
reliable way to re-establish the connection. This adds a simple
public 'shouldRestart' bool (based on the existing Queue\Worker
'stopWorkerIfLostConnection' code) to the Redis Connection, which
is checked in RedisManager and recreated if needed there.

Signed-Off-By: Rob Thomas <xrobau@gmail.com>

### DIFF
--- a/src/Illuminate/Redis/Connections/Connection.php
+++ b/src/Illuminate/Redis/Connections/Connection.php
@@ -36,8 +36,8 @@ abstract class Connection
      */
     protected $events;
 
-     /**
-     * Indicates if the connection is invalid and needs to reconnect
+    /**
+     * Indicates if the connection is invalid and needs to reconnect.
      *
      * @var bool
      */
@@ -115,6 +115,7 @@ abstract class Connection
      * @param  string  $method
      * @param  array  $parameters
      * @return mixed
+     *
      * @throws \Exception Throws an exception if this connection should not be used
      */
     public function command($method, array $parameters = [])
@@ -126,7 +127,7 @@ abstract class Connection
 
             // Basically - if you see this, this is a bug in whatever is NOT using RedisManager
             // to get Redis objects.
-            throw new \Exception("The Redis connection " . $this->name . " has been marked as failed and needs to be recreated.");
+            throw new \Exception('The Redis connection '.$this->name.' has been marked as failed and needs to be recreated.');
         }
         $start = microtime(true);
 
@@ -136,6 +137,7 @@ abstract class Connection
             // If there is a RedisException, this connection is no longer usable.
             // Mark the connection as shouldReconnect, and return an empty array.
             $this->shouldReconnect = true;
+
             return [];
         }
 
--- a/src/Illuminate/Redis/Connections/Connection.php
+++ b/src/Illuminate/Redis/Connections/Connection.php
@@ -135,10 +135,10 @@ abstract class Connection
             $result = $this->client->{$method}(...$parameters);
         } catch (\RedisException $e) {
             // If there is a RedisException, this connection is no longer usable.
-            // Mark the connection as shouldReconnect, and return an empty array.
+            // Mark the connection as shouldReconnect, and re-throw the Exception
             $this->shouldReconnect = true;
 
-            return [];
+            throw $e;
         }
 
         $time = round((microtime(true) - $start) * 1000, 2);
--- a/src/Illuminate/Redis/Connections/Connection.php
+++ b/src/Illuminate/Redis/Connections/Connection.php
@@ -36,6 +36,13 @@ abstract class Connection
      */
     protected $events;
 
+     /**
+     * Indicates if the connection is invalid and needs to reconnect
+     *
+     * @var bool
+     */
+    public $shouldReconnect = false;
+
     /**
      * Subscribe to a set of given channels for messages.
      *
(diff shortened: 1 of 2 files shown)


src/Illuminate/Redis/RedisManager.php
```diff
@@ -84,11 +84,19 @@ public function connection($name = null)
...
         if (isset($this->connections[$name])) {
-            return $this->connections[$name];
+            // If the connection has had a fatal socket error, the \Redis object
+            // in this connection is no longer valid, and needs to be destroyed
+            // and recreated.
+            if (! $this->connections[$name]->shouldReconnect) {
+                return $this->connections[$name];
+            }
+            // Unset and destroy the Connection object to ensure it's cleaned up.
+            unset($this->connections[$name]);
```
xrobau, line 94: Note the unset is here to make sure that the isset always returns false, even if the connection below fails.

driesvints:
Thanks @xrobau

taylorotwell:
I don't understand why we would return an empty array on failure. Wouldn't we just want to rethrow the exception and let it be logged, etc?

xrobau:
Updated PR @taylorotwell  - I was returning an empty array because the noise was irritating to me, but that's not going to help anyone else 8)

src/Illuminate/Redis/Connections/Connection.php
```diff
@@ -108,12 +115,31 @@ public function psubscribe($channels, Closure $callback)
...
+            // Basically - if you see this, this is a bug in whatever is NOT using RedisManager
+            // to get Redis objects.
+            throw new \Exception('The Redis connection '.$this->name.' has been marked as failed and needs to be recreated.');
+        }
         $start = microtime(true);
 
-        $result = $this->client->{$method}(...$parameters);
+        try {
+            $result = $this->client->{$method}(...$parameters);
+        } catch (\RedisException $e) {
```
driesvints, line 136: Can you import all classes?

src/Illuminate/Redis/RedisManager.php
```diff
@@ -169,8 +177,8 @@ protected function connector()
         }
 
         return match ($this->driver) {
-            'predis' => new PredisConnector,
-            'phpredis' => new PhpRedisConnector,
+            'predis' => new PredisConnector(),
+            'phpredis' => new PhpRedisConnector(),
```
driesvints, line 181: Please revert all extra `()` that were added, thanks.

driesvints (changes requested):

taylorotwell:
Did you examine the code for `PhpRedisConnection`? It already has reconnection logic within it's command method. Why can that not be used / modified?

Furthermore, `RedisException` is specific to the phpredis C extension - not Predis. So, these changes don't make sense in the base `Connection` class. They are phpredis specific.

xrobau:
> They are phpredis specific.

@taylorotwell I'm unsure if the Predis connection can survive after a network error. Looking at that code, it looks like it can. I'm certain that Phpredis can not survive, which is why I was specific about only catching that one exception, and then explicitly destroying and recreating the object.

src/Illuminate/Redis/Connections/Connection.php
xrobau: As per my reply, this is specific only to Phpredis, and this deliberately only handling \RedisException

xrobau:
Oh, I misread that screenshot. Sorry, I think I see what you're getting at.

xrobau:
Screenshot above references https://github.com/laravel/framework/blob/9.x/src/Illuminate/Redis/Connections/PhpRedisConnection.php#L533 for anyone with visual issues.

@taylorotwell - What would you prefer? That code is currently ONLY catching 'has gone away', which appears to be triggered by a controlled shutdown of Redis.  The error that I'm attempting to fix is when an UNCONTROLLED shutdown happens, or a network issue.

Anecdotally, I've never had any exception thrown by Phpredis that is recoverable at all.  Was there any reason that only one edge case is handled?

src/Illuminate/Redis/Connections/Connection.php
xrobau: @driesvints this may be resolved simply by ALWAYS reconnecting redis here, rather than only on one specific edge case. Waiting for feedback below, which would simplify this to simply removing the `if`.

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Redis/Connections/PhpRedisConnection.php#L534

xrobau:
Alternative PR https://github.com/laravel/framework/pull/41546 - as I mention in the PR, I don't think this is suitable, and would prefer to remove the `if` totally.

PR #41504, opened 2022-03-16
[9.x] Generic doc for make method

HalfBottleOfMind:
With this update, static analyzer will know the return type of `app()->make(Foo::class)`.

```php
$foo = app()->make(Foo::class);
// Now static analyzer know that $foo is an instance of Foo::class

$foo->...
// Autocomplete will suggest property and method names
```


nunomaduro:
We are not supporting generics outside of Laravel's collections for now.

PR #41505, opened 2022-03-16
[9.x] Add some tests for lcfirst on Stringable

reziamini:
lcfirst method on Stringable class doesn’t have any tests.


taylorotwell:
It is tested on Str class.

PR #41514, opened 2022-03-16
[9.x] handle toMailUsing and createUrlUsing for ResetPassword the same as in VerifyEmail

dseidl:
I noticed that **createUrlUsing** callback for the **ResetPassword** notification was not called anymore, when also using the **toMailUsing** callback. 

As it did work for the **VerifyEmail** notification i expected the same behaviour.

This would be a breaking change i guess, when people expect the token as the second parameter in the toMailUsing callback.

### DIFF
--- a/tests/Integration/Auth/ForgotPasswordWithoutDefaultRoutesTest.php
+++ b/tests/Integration/Auth/ForgotPasswordWithoutDefaultRoutesTest.php
@@ -9,6 +9,7 @@ use Illuminate\Support\Facades\Password;
 use Illuminate\Tests\Integration\Auth\Fixtures\AuthenticationTestUser;
 use Orchestra\Testbench\Factories\UserFactory;
 use Orchestra\Testbench\TestCase;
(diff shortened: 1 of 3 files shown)


tests/Integration/Auth/ForgotPasswordWithoutDefaultRoutesTest.php
```diff
@@ -93,15 +93,18 @@ function (ResetPassword $notification, $channels) use ($user) {
     }
 
     /** @test */
-    public function it_can_send_forgot_password_email_via_to_mail_using()
+    public function it_cannot_send_forgot_password_email_via_to_mail_using_without_default_route()
     {
+        $this->expectException('Symfony\Component\Routing\Exception\RouteNotFoundException');
```
driesvints, line 98: Import this class here please.

driesvints (changes requested):

taylorotwell:
We can't make this breaking change on 9.x. It has been surfaced before - see here: https://github.com/laravel/framework/pull/41109

PR #41516, opened 2022-03-16
[9.x] Make throttle lock acquisition retry time configurable

timmartin19:
Depending on context, 750ms can be a long time to wait between retries
for a throttle.  We are making the retry time configurable so that users
can tune the throttling further.

All changes are backwards compatible.

I set the unit as milliseconds because this seemed like the natural time
unit to use.  Seconds would require floats which felt dirty.  That being
said, it's not consistent with the unit used for `->every` and `->block`
on the builder.  So I'm open to changing it for consistency.

I was additionally unsure what useful tests could be written for this
configuration.  Mocking the `usleep` function felt silly. Any suggestions 
are welcome.

```php
Redis::throttle('my-key')
    ->allow(5)
    ->every(10)
    ->retryWait(100)
    ->then(static fn () => echo "hello!");
```

### DIFF
--- a/src/Illuminate/Redis/Limiters/DurationLimiter.php
+++ b/src/Illuminate/Redis/Limiters/DurationLimiter.php
@@ -70,12 +70,12 @@ class DurationLimiter
      *
      * @param  int  $timeout
      * @param  callable|null  $callback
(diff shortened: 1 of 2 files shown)


taylorotwell:
Renamed method to `sleep`.

PR #41519, opened 2022-03-16
[9.x] Update dispatch function declaration

kaydenvanrijn:


driesvints:
There's no explanation here.

kaydenvanrijn:
The explanation is that it doesn't use a variadic parameter like the rest of the functions within the trait. All I did was add a parameter so that an IDE can display that arguments are accepted. 

Take a look here: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Foundation/Bus/Dispatchable.php
Versus the change I was making to: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Foundation/Events/Dispatchable.php

PR #41524, opened 2022-03-17
[9.x] Add Arr::pad

Lukasss93:
Added missing pad method to Arr class.

```php
Arr::pad([1], 3, 0);     // [1, 0, 0]
Arr::pad([1], -3, 0);    // [0, 0, 1]
Arr::pad([], 3, 0);      // [0, 0, 0]
Arr::pad([1], 0, 0);     // [1]
```

Use case example:
```php
$string = 'A-1';
[$char, $int] = Arr::pad(explode('-', $array), 2, null); // fix a possible exception if string is "A" or "B-"
```

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -504,6 +504,19 @@ class Arr
     }
 
     /**
+     * Pad an array to the specified length with a value.
+     *
+     * @param  array  $array
+     * @param  int  $length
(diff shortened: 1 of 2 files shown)


SjorsO:
What is the advantage of calling this method on the `Arr` class instead of just calling `array_pad`?

driesvints:
There's indeed no tangible benefit here. Thanks

PR #41527, opened 2022-03-17
[9.x] returned template in container make()

krzysztofrewak:
This is a little change, but I think it could be very helpful for many developers.

After adding a template to phpdocs of Container contract we'd be able to get all hints about resolved class:

obraz

### DIFF
--- a/src/Illuminate/Contracts/Container/Container.php
+++ b/src/Illuminate/Contracts/Container/Container.php
@@ -157,7 +157,7 @@ interface Container extends ContainerInterface
      *
      * @template T
      *
-     * @param  class-string<T> $abstract
+     * @param  class-string<T>  $abstract
      * @param  array  $parameters
      * @return T
      *
--- a/src/Illuminate/Contracts/Container/Container.php
+++ b/src/Illuminate/Contracts/Container/Container.php
@@ -155,9 +155,11 @@ interface Container extends ContainerInterface
     /**
      * Resolve the given type from the container.
      *
-     * @param  string  $abstract
+     * @template T
+     *
+     * @param  class-string<T> $abstract
      * @param  array  $parameters
-     * @return mixed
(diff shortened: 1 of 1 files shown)


driesvints:
Please see https://github.com/laravel/framework/pull/41504

krzysztofrewak:
> We are not supporting generics outside of Laravel's collections for now.

Could you or @nunomaduro explain why Laravel is closed for changes like this one? It would really improve my and my team's (and a lot of other people's too) development process, it's compatible with everything else in current codebase and it's not hurting anyone. Would you like to conduct some bigger analysis "What should we improve in this file in one bigger task"?

pionl:
@krzysztofrewak I've hold these changes for a week and made PR today (did not notice other did it to, my bad), not sure why its not supported, closed to. Not sure why. Probably there is some reason that we are unaware off, we will see.

 I've tried the changes it with PHPstan and no warnings.

pionl:
@driesvints Is it possible to revisit this?

I understand that `get` and `make` can use the type hint. I think there could be a solution by providing new method, something like `makeClass`.

```php

/**
 * Get the available container instance.
 *
 * @template T of object
 *
 * @param class-string<T>   $class
 * @param array<int, mixed> $parameters
 *
 * @return T
 */
function makeClass(string $class, array $parameters = []): object
{
    $instance = $this->make($class, $parameters);

    assert(assertion: $instance instanceof $class, description: 'Instance is not of type ' . $class);

    return $instance;
}
```

This would save quite a of code I have to do (to make PHPStan and IDE happy - without any plugins):

```php
$dateFormats = $this->app->get(DateFormatterService::class);
assert($dateFormats instanceof DateFormatterService);
```

I can make the PR.

func0der:
@driesvints Why? Is there some public reasoning/statement for this?

PR #41528, opened 2022-03-17
[9.x] Exactly match scheduled command --name in schedule:test

ziadoz:
I've stripped the PHP and Artisan binary paths off of the command names here. This means instead of looking at the end of the string, which could accidentally match the wrong command, an exact match can be done.

### DIFF
--- a/src/Illuminate/Console/Scheduling/ScheduleTestCommand.php
+++ b/src/Illuminate/Console/Scheduling/ScheduleTestCommand.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Console\Scheduling;
 
+use Illuminate\Console\Application;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can you provide further explanation of scenarios where this improves things and before / after examples? Mark as ready for review when you have done so please.

ziadoz:
Sure, I can explain. 

I used `Str::endsWith()` when I added `--name` to this command because I hadn't realised I could remove the PHP and Artisan binaries from the command names. With them gone an exact match on the command name is possible.

So before if you had a command named `cmd:apple` then `--name=e` would match it because it ends in `e`. After `--name=e` will match nothing, and you'll have to explicitly do `--name=cmd:apple` instead.

PR #41529, opened 2022-03-17
[9.x] Improves `ResetPassword` Notification docblock

crynobone:

### DIFF
--- a/src/Illuminate/Auth/Notifications/ResetPassword.php
+++ b/src/Illuminate/Auth/Notifications/ResetPassword.php
@@ -18,14 +18,14 @@ class ResetPassword extends Notification
     /**
      * The callback that should be used to create the reset password URL.
      *
-     * @var \Closure|null
+     * @var (\Closure(mixed, string): string)|null
      */
     public static $createUrlCallback;
 
     /**
      * The callback that should be used to build the mail message.
      *
-     * @var \Closure|null
+     * @var (\Closure(mixed, string): \Illuminate\Notifications\Messages\MailMessage)|null
      */
     public static $toMailCallback;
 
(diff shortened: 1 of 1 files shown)


src/Illuminate/Auth/Notifications/ResetPassword.php
```diff
@@ -18,14 +18,14 @@ class ResetPassword extends Notification
...
-     * @var \Closure|null
+     * @var (\Closure(mixed, string): string)|null
      */
     public static $createUrlCallback;
 
     /**
      * The callback that should be used to build the mail message.
      *
-     * @var \Closure|null
+     * @var (\Closure(mixed, string): \Illuminate\Notifications\Messages\MailMessage)|null
```
Jubeki, line 28: It should also be possible to return a Mailable (I have never tried that).
```suggestion
     * @var (\Closure(mixed, string): (\Illuminate\Notifications\Messages\MailMessage|\Illuminate\Mail\Mailable))|null
```

src/Illuminate/Auth/Notifications/ResetPassword.php
```diff
@@ -114,7 +114,7 @@ public static function createUrlUsing($callback)
     /**
      * Set a callback that should be used when building the notification mail message.
      *
-     * @param  \Closure  $callback
+     * @param  \Closure(mixed, string): \Illuminate\Notifications\Messages\MailMessage  $callback
```
Jubeki, line 117: It should also be possible to return a Mailable (I have never tried that).
```suggestion
     * @param  \Closure(mixed, string): (\Illuminate\Notifications\Messages\MailMessage|\Illuminate\Mail\Mailable)  $callback
```

src/Illuminate/Auth/Notifications/ResetPassword.php
crynobone: The usage class doesn't return what you suggesting. Maybe create different PR. I'm not adding something I can't confirm right now :-)

PR #41537, opened 2022-03-17
[9.x] Add tests for Pluralizer::inflector() method to verify object

reziamini:
There was no test to identify the `Inflector` object.
This object should be created one time (kind of singleton) and it should be an instance of the `\Doctrine\Inflector\Inflector::class` class.

The first and second event n created object should be equal to the first object and it should not create an object per each method call.


taylorotwell:
It's really fine.

PR #41540, opened 2022-03-17
[9.x] Fix make:factory command to work with nested namespaces and fix guessModelName to work with new stucutre

m-mohammadi1:
If you run this
```php
php artisan make:factory Test/Sub1/Sub2/User
```
then yout would have this result:
```php
<?php

namespace Database\Factories;

use Illuminate\Database\Eloquent\Factories\Factory;

/**
 * @extends \Illuminate\Database\Eloquent\Factories\Factory<\App\Models\Model>
 */
class UserFactory extends Factory
{
    /**
     * Define the model's default state.
     *
     * @return array<string, mixed>
     */
    public function definition()
    {
        return [
            //
        ];
    }
}
```
as you can see the namespace is not correct also notice <\App\Models\Model> that should be <\App\Models\User> because I have User Model, and actually I think guessing model names had some problems
but with new changes all the problems was fixed and the result would be like this:
```php
<?php

namespace Database\Factories\Test\Sub1\Sub2;

use Illuminate\Database\Eloquent\Factories\Factory;

/**
 * @extends \Illuminate\Database\Eloquent\Factories\Factory<\App\Models\User>
 */
class UserFactory extends Factory
{
    /**
     * Define the model's default state.
     *
     * @return array<string, mixed>
     */
    public function definition()
    {
        return [
            //
        ];
    }
}
```
as you can see the namespace and model name are correct now.
also guessing model name is correct now. 
I also removed old syntax and used laravel Str and Arr facades.

### DIFF
--- a/src/Illuminate/Database/Console/Factories/FactoryMakeCommand.php
+++ b/src/Illuminate/Database/Console/Factories/FactoryMakeCommand.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Database\Console\Factories;
 
 use Illuminate\Console\GeneratorCommand;
+use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Revert all unnecessary spacing / style changes. 

Also, doing `artisan make:model Foo/Bar/Baz -f` already namespaces correctly. Do you know why that works and `make:factory` doesn't?

luckyboy1001:
I need more time to fix that actully guessing modal name was not correct I think.
And when we use --model=SomeModel it just replace it if that exists in app/ or app/Models.
Replacing model name was not so smart.
I will fix issues and maybe make gussing model name smarter.

luckyboy1001:
When we build `artisan make:model Foo/Bar/Baz -f` in `Illuminate\Foundation\Console\ModelMakeCommand` you can see this
```php
    protected function createFactory()
    {
        $factory = Str::studly($this->argument('name'));

        $this->call('make:factory', [
            'name' => "{$factory}Factory",
            '--model' => $this->qualifyClass($this->getNameInput()),
        ]);
    }
```
as you can see we use `qualifyClass` method that gives the `make:factory` full namespace of the model but when we use --model ourselves we just give model name and if it exists in `app/` or `app/Models` it works I think the problem was inside the `make:model` command and now it is fixed with new changes.

driesvints:
Seems like there's still quite a bit of styling changes. Please revert those first.

taylorotwell:
This PR breaks creating factories for models outside of the App models directory. For example:

```
php artisan make:factory AdminFlightFactory --model=App/Admin/Entities/Flight
```

m-mohammadi1:
> This PR breaks creating factories for models outside of the App models directory. For example:
> 
> ```
> php artisan make:factory AdminFlightFactory --model=App/Admin/Entities/Flight
> ```
> 
> After this PR:
> 
> CleanShot 2022-03-18 at 09 15 07@2x
I fixed it but now it works like this:

- if you give --model= it will be replaced for every folder in app/ and outside of app folder even the model does not exists
- if you don't give --model option based on factory name it will be searched in app/ and app/Models and if it not found the factory's model namespace becomes `App/Models/Model` and if Models/ folder not exists it becomes `App/Model`

when giving model name as an option it won't check for model existence maybe we could make it better and ask for creating model but make:model just support creating models for app/Models folder now.

taylorotwell:
Now if I do: `artisan make:model App/Admin/Flight --factory`

I get the following factory, which to me isn't correct:

CleanShot 2022-03-25 at 16 28 06@2x

The app namespace shouldn't be part of it.

taylorotwell:
@m-mohammadi1 can you tell me if this simple PR fixes your problems: https://github.com/laravel/framework/pull/41688

taylorotwell:
Closing pending inactivity.

m-mohammadi1:
> @m-mohammadi1 can you tell me if this simple PR fixes your problems: #41688

Yes it works fine now I think I edit wrong place

PR #41542, opened 2022-03-17
[9.x] Enable cache tags for DynamoDB driver

Robertbaelde:
This PR enables cache tags for the dynamoDB cache driver.

One known caveat is that when a user flushes tags (`Cache::tags('authors')->flush();`), the cache tag keys are refreshed but the original cache item still remains in the DynamoDB table, now ignored as should, but still taking up some space. I believe this is the same for all other cache methods though, except for the Redis implementation.

### DIFF
--- a/src/Illuminate/Cache/DynamoDbStore.php
+++ b/src/Illuminate/Cache/DynamoDbStore.php
@@ -5,13 +5,12 @@ namespace Illuminate\Cache;
 use Aws\DynamoDb\DynamoDbClient;
 use Aws\DynamoDb\Exception\DynamoDbException;
 use Illuminate\Contracts\Cache\LockProvider;
-use Illuminate\Contracts\Cache\Store;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\InteractsWithTime;
 use Illuminate\Support\Str;
 use RuntimeException;
 
-class DynamoDbStore implements LockProvider, Store
+class DynamoDbStore extends TaggableStore implements LockProvider
 {
     use InteractsWithTime;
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
I think we would have to resolve the items staying out there forever. Memcache has a max memory setting (if memory serves me right) so stale items would eventually be purged out even before expiration. Dynamo would be similar to Redis in that we would need to take care of removing those orphaned forever entries.

Robertbaelde:
Hmm, I'm not sure if the amount of code/queries needed to delete the keys outweighs the costs of storage for those extra few rows. (Of-course this trade off is different per use-case). Right now expired cache items in dynamo are not purged as well (Only when you configure ttl in your table configuration, but the docs don't mention it). 

If we'd have to delete the orphaned (forever) items on tag flush, the following work needs to be done: 
* Keep track of tag - tem relation (so for each tag of an item we push we'd need an extra row in the table). 
* On delete of a tag query the tag - item relation and flush the items accordingly

I'm happy to write the code for this. But I'm afraid that this will result in the PR being closed since the amount of code/complexity needed (and the maintenance that comes with it) doesn't outweighs the benefit it brings. 

I think this is a case of: "Storage is cheap, writing code is expensive". I'd suggest to put a small note in the docs that items won't be physically deleted in Dynamo and keep it at that. 

So decision for You / the Laravel team would be: 

1. Give green light for the added complexity that deleting items in Dynamo brings. Happy to write the code once there's green light. 
2. Keep the PR as is because deleting keys would be too complex to maintain, with optional adding a small note in the docs. 
3. Keep tag keys out of Dynamo (wouldn't be too happy with that. But would also understand that).

taylorotwell:
I guess one question I have with this is if we got this route, why disable tagging support for *any* cache drivers? Database, file, etc.

However, I still have concerns about leaving data in the cache indefinitely with no reasonable way to reclaim that space.

taylorotwell:
I don't think I'm going to pursue this for now due to concerns above.

PR #41543, opened 2022-03-17
[9.x] DynamoDB don't set expired at value on forever

Robertbaelde:
Currently when using `Cache::driver('dynamodb')->forever('name', 'Taylor')` the item expires at time will actually be set. 
It will be set to expire in "the timestamp of 5 years in the future" seconds from now. Obviously this practically hasn't caused (and won't cause) any bugs.

However it is a bit odd to see "expiredAt" being set when looking at an item, when you don't expect it to expire. This PR has a small (backwards compatible) refactor that will omit expiredAt to be stored when using forever as expiresAt value.


taylorotwell:
Just going to leave it for now.

PR #41545, opened 2022-03-17
[9.x] Collection select()

roberto-butti:
As explained here: https://github.com/laravel/framework/discussions/41521 , I open a PR for introducing _select()_ method for Collection.
_select()_ get data from bi-dimensional array by keys.

I created also tests in tests/Support/SupportCollectionTest.php adding _testSelect()_ method with some test case (not existent columns, null columns, string column.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -598,7 +598,7 @@ class Arr
     }
 
     /**
-     * Select and "pick" columns from a bi-dimensional array
+     * Select and "pick" columns from a bi-dimensional array.
      *
      * @param  array  $array
      * @param  array  $columns
@@ -607,11 +607,12 @@ class Arr
     public static function select(&$array, $columns)
     {
         if (is_string($columns)) {
-            $columns = array($columns);
+            $columns = [$columns];
         }
         if (is_null($columns)) {
             $columns = [];
         }
+
         return array_map(fn ($item) => array_intersect_key($item, array_flip($columns)), $array);
     }
 
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -606,12 +606,7 @@ class Arr
      */
     public static function select(&$array, $columns)
     {
-        if (is_string($columns)) {
-            $columns = [$columns];
-        }
-        if (is_null($columns)) {
-            $columns = [];
-        }
+        $columns = (array) $columns;
 
         return array_map(fn ($item) => array_intersect_key($item, array_flip($columns)), $array);
     }
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -601,14 +601,14 @@ class Arr
      * Select and "pick" columns from a bi-dimensional array.
      *
      * @param  array  $array
-     * @param  array  $columns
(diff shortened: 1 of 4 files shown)


src/Illuminate/Collections/Arr.php
```diff
@@ -597,6 +597,25 @@ public static function random($array, $number = null, $preserveKeys = false)
...
+    /**
+     * Select and "pick" columns from a bi-dimensional array.
+     *
+     * @param  array  $array
+     * @param  array  $columns
+     * @return array
+     */
+    public static function select(&$array, $columns)
+    {
+        if (is_string($columns)) {
```
hAz5, line 609: I think You can use 
```
$columns = (array) $columns;
```

instead of two If conditions.

roberto-butti: > I think You can use
> 
> ```
> $columns = (array) $columns;
> ```
> 
> instead of two If conditions.

Thank you for the feedback.
Yes i think that your suggestion is great.
```
>>> (array) "some"
=> [
     "some",
   ]
>>> (array) null
=> []

```

src/Illuminate/Collections/Arr.php
```diff
@@ -597,6 +597,20 @@ public static function random($array, $number = null, $preserveKeys = false)
         return $results;
     }
 
+    /**
+     * Select and "pick" columns from a bi-dimensional array.
+     *
+     * @param  array  $array
+     * @param  array  $columns
```
dennisprudlo, line 604: Shouldn't this be `@param  array|string  $columns` then, if we accept strings as well?

src/Illuminate/Collections/Collection.php
```diff
@@ -1046,6 +1046,17 @@ public function search($value, $strict = false)
         return false;
     }
 
+    /**
+     * Select and "pick" columns from a bi-dimensional array.
+     *
+     * @param  array  $columns
```
dennisprudlo, line 1052: Same thing here

roberto-butti:
> 
> 

ok , thank you for the suggestion.
I'm building a package for this

```php
use HiFolks\DataType\Table;

$table = Table::make([
    ['product' => 'Desk', 'price' => 200, 'active' => true],
    ['product' => 'Chair', 'price' => 100, 'active' => true],
    ['product' => 'Door', 'price' => 300, 'active' => false],
    ['product' => 'Bookcase', 'price' => 150, 'active' => true],
    ['product' => 'Door', 'price' => 100, 'active' => true],
]);
/**
 * I would like to **filter** the rows with price greater than 100,
 * select only "product" and "price" fields,
 * and for each rows **create a new field** named "new_filed"
 * that is a calculated field (doubling the price)
 */
var_export(
    $table
        ->select(['product' , 'price'])
        ->where('price', 100, '>')
        ->calc("new_field", fn ($item) => $item['price'] * 2)
        ->arr()
);
/*
array (
  0 => array (
    'product' => 'Desk',
    'price' => 200,
    'new_field' => 400,
  ),
  2 => array (
    'product' => 'Door',
    'price' => 300,
    'new_field' => 600,
  ),
  3 => array (
    'product' => 'Bookcase',
    'price' => 150,
    'new_field' => 300,
  ),
)
 */
```

The packages is https://github.com/Hi-Folks/array , the class is **Table**

PR #41551, opened 2022-03-18
[9.x] Add "order by" with field concatenation clause to the query

lee-to:
This PR adds convenient sorting with field concatenation

```php
User::query()->orderByConcat(['first_name', 'last_name'], 'desc')->get()
```

```sql
select *, CONCAT(first_name, last_name) as first_name_last_name from "users" order by "first_name_last_name" desc
```

especially useful in the presence of Accessors, reducing the logic

```php
    {
        return Attribute::make(
            get: fn ($value) => $this->first_name . ' ' . $this->last_name,
        );
    }
```

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -2243,6 +2243,30 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Add "order by" with field concatenation clause to the query.
+     *
+     * @param array $columns
+     * @param string $direction
+     * @return $this
+     */
+    public function orderByConcat(array $columns, $direction = 'asc')
+    {
+        if (empty($columns)) {
+            return $this;
+        }
+
+        if (is_null($this->columns)) {
+            $this->select([$this->from.'.*']);
+        }
+
+        $alias ??= Str::snake(implode(' ', $columns));
+
+        $this->addSelect(new Expression('CONCAT('.implode(',', $columns).') as '.$alias));
+
+        return $this->orderBy($alias, $direction);
+    }
+
+    /**
      * Alias to set the "offset" value of the query.
      *
      * @param  int  $value
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -2247,9 +2247,10 @@ class Builder implements BuilderContract
      *
      * @param array $columns
      * @param string $direction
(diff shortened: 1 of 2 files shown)


BrandonSurowiec:
If only for sorting, it seems inferior to:

```
User::query->latest('first_name')->latest('last_name');
```

With the `orderByConcat()` version, there will be an unexpected order as there is no "separator" character between the columns to differentiate. An example would be given this data:

```
Name    | Email
------------------------
Bray    |  z@email.com
Braylon | a@email.com
```

With double order by like `->latest('name')->latest('email')` you get the expected results.
```
Bray z@email.com
Braylon a@email.com
```

With `->orderByConcat(['name', 'email'])`'s current implementation, it smashes everything together without a separator and sorts out of order:
```
Braylona@email.com
Brayz@email.com
```

I think having some sort of `selectConcat` method could be beneficial, but not sure that having it as ordering syntax is the best idea. I'm not a DB expert, but it seems that two `orderBy()` statements with the proper indexes will be much faster than the DB having to concatenate the columns together before sorting with a larger dataset.

lee-to:
> If only for sorting, it seems inferior to:
> 
> 
> 
> ```
> 
> User::query->latest('first_name')->latest('last_name');
> 
> ```
> 
> 
> 
> With the `orderByConcat()` version, there will be an unexpected order as there is no "separator" character between the columns to differentiate. An example would be given this data:
> 
> 
> 
> ```
> 
> Name    | Email
> 
> ------------------------
> 
> Bray    |  z@email.com
> 
> Braylon | a@email.com
> 
> ```
> 
> 
> 
> With double order by like `->orderBy('name')->orderBy('email')` you get the expected results.
> 
> ```
> 
> Bray z@email.com
> 
> Braylon a@email.com
> 
> ```
> 
> 
> 
> With `->orderByConcat(['name', 'email'])`'s current implementation, it smashes everything together without a separator and sorts out of order:
> 
> ```
> 
> Braylona@email.com
> 
> Brayz@email.com
> 
> ```
> 
> 
> 
> I think having some sort of `selectConcat` method could be beneficial, but not sure that having it as ordering syntax is the best idea. I'm not a DB expert, but it seems that two `orderBy()` statements with the proper indexes will be much faster than the DB having to concatenate the columns together before sorting with a larger dataset.

thanks for the answer, sounds right, I think you're right

taylorotwell:

PR #41555, opened 2022-03-18
[9.x] Handle Symfony defaultName deprecation

driesvints:
This PR adds handling for Symfony's deprecated `$defaultName ` property on console commands in the upcoming Symfony 6.1 release. To mitigate deprecation notices we'll need to implement the new `AsCommand` attribute that ships with Symfony v6. 

See the related Symfony PR: https://github.com/symfony/symfony/pull/45361
And issue: https://github.com/symfony/symfony/issues/43600

Merging this into master will fix our failing tests there. After this is merged on master, I'll take the additional step to remove all `$defaultName` properties.

We'll also need to do this for all first party libraries that make use of the `$defaultName` property. Additionally, all apps and packages that make use of the `$defaultName` property will need to migrate to this new syntax.

### DIFF
--- a/src/Illuminate/Database/Console/WipeCommand.php
+++ b/src/Illuminate/Database/Console/WipeCommand.php
@@ -18,15 +18,6 @@ class WipeCommand extends Command
     protected $name = 'db:wipe';
 
     /**
-     * The name of the console command.
-     *
-     * This name is used to identify the command during lazy loading.
-     *
-     * @var string|null
-     */
-    protected static $defaultName = 'db:wipe';
-
-    /**
      * The console command description.
      *
      * @var string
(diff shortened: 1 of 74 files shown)


src/Illuminate/Database/Console/WipeCommand.php
```diff
@@ -17,15 +17,6 @@ class WipeCommand extends Command
...
 
-    /**
-     * The name of the console command.
-     *
-     * This name is used to identify the command during lazy loading.
-     *
-     * @var string|null
-     */
-    protected static $defaultName = 'db:wipe';
-
```
Jubeki, line 28: Was there a reason you deleted that instead of add `@deprecated` and `AsCommand` Attribute?

Jubeki: I checked all other changed commands and I found no missing `@deprecated` or `AsCommand`. I also checked if all command names are correctly copied.

driesvints: @Jubeki nope, I re-added that. Thanks for spotting and reviewing!

adideas: I don't know why I did the same. ))

PR #41557, opened 2022-03-18
[9.x] Add @template typehints for make / instance / app to improve type hints in code

pionl:
👋

Just a small life improvements for IDE users.

- This enables resolving type of instance that container resolves without any IDE plugins.
- Helps phpstan / psalm / other tools

Screenshots from PHPstorm while testing the annotations.

Snímek obrazovky 2022-03-15 v 16 49 52

DispatchJobAction

PR #41562, opened 2022-03-18
[9.x] Revert #41557

nunomaduro:


caendesilva:
@nunomaduro Hey, sorry for the ping, but I was wondering if we could get some insight in why this was reverted? This seems to have been a popular feature and I find it odd that it was removed. See https://github.com/laravel/framework/discussions/41597

PR #41568, opened 2022-03-19
[9.x] Use arrow functions to bind singletons

imanghafoori1:
Uses php 7.4 arrow function to bind singleton in the service providers.
It also remove a little bit of extra whitespace from `Connection.php`

- Conversion process is done with the help of IDE to minimize human error.


taylorotwell:
I think the code is fine. Thanks.

PR #41569, opened 2022-03-19
[9.x] use null coalescing operator and null safe operator

masouddarvishi:
minor refactor in three methods to use null safe and null coalescing operators instead of if and isset.


taylorotwell:
If the code is working I do not like to change it much.

PR #41570, opened 2022-03-19
[9.x] Remove useless if statement in Str

hAz5:
because min requirement of Laravel9 PHP version is 8 and above
 we don't need to check this  
```
  if (is_null($length) && PHP_MAJOR_VERSION < 8) {
      $length = mb_strlen($string, $encoding);
  }
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -492,10 +492,6 @@ class Str
             return $string;
         }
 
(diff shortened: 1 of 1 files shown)


imanghafoori1:
Please explain why it is useless.

crynobone:
> Please explain why it is useless.

Laravel 9 doesn't support PHP 7

PR #41571, opened 2022-03-19
[9.x] Use Null coalescing assignment operator ??=

imanghafoori1:
It uses php 7.4 Null coalescing assignment operator (??=).


taylorotwell:
Honestly I don't like to merge big formatting PRs like this because of the risk of subtle changes just isn't worth it. Thanks.

PR #41572, opened 2022-03-19
[9.x] compatible contract return type and method body

hAz5:
return type listen method was void but in the method body, we got return collection.

```
/**
     * Register an event listener with the dispatcher.
     *
     * @param  \Closure|string|array  $events
     * @param  \Closure|string|array|null  $listener
     *
     * @return **void**
     */
    public function listen($events, $listener = null)
    {
        if ($events instanceof Closure) {
            **return** collect($this->firstClosureParameterTypes($events))
                ->each(function ($event) use ($events) {
                    $this->listen($event, $events);
                });

```

### DIFF
--- a/src/Illuminate/Events/Dispatcher.php
+++ b/src/Illuminate/Events/Dispatcher.php
@@ -72,7 +72,7 @@ class Dispatcher implements DispatcherContract
      *
      * @param  \Closure|string|array  $events
      * @param  \Closure|string|array|null  $listener
-     * @return void
+     * @return void|\Illuminate\Support\Collection
      */
     public function listen($events, $listener = null)
     {


src/Illuminate/Events/Dispatcher.php
```diff
@@ -72,7 +72,7 @@ public function __construct(ContainerContract $container = null)
      *
      * @param  \Closure|string|array  $events
      * @param  \Closure|string|array|null  $listener
-     * @return void
+     * @return void|\Illuminate\Support\Collection
```
driesvints, line 75: Void can never be combined with something else. This method follows its contract which has void. The only reason why there are returns is to early terminate the method call.

hAz5: thank you for your advice.
I think if we want to be on contract its better to use this :
```
 if ($events instanceof Closure) {
            collect($this->firstClosureParameterTypes($events))
                ->each(
                    fn($event) => $this->listen($event, $events)
                );

            return;
        } elseif ($events instanceof QueuedClosure) {
```

I reopen this again and changed the PR

driesvints: Changing it has no tangible benefit.

hAz5: IDE shows this code doc block as a warning, and the behavior of this method is different from its return type and its contact.

PR #41573, opened 2022-03-19
[9.x] Collections: Allow pluck as higher order message

levu42:
Additionally to the few other higher order message functions I just assumed that `$collection->pluck->id` was a thing but it turned out that it wasn't implemented yet. I don't see any reason why it shouldn't be implemented, so here's the pull request for it :)

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -697,12 +697,16 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     /**
      * Get the values of a given key.
      *
-     * @param  string|array<array-key, string>  $value
+     * @param  string|array<array-key, string>|callable  $value
      * @param  string|null  $key
      * @return static<int, mixed>
      */
     public function pluck($value, $key = null)
     {
+        if (is_callable($value)) {
+            return $this->map($value);
+        }
+        
         return new static(Arr::pluck($this->items, $value, $key));
     }
 
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -706,7 +706,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
         if (is_callable($value)) {
             return $this->map($value);
         }
-        
+
         return new static(Arr::pluck($this->items, $value, $key));
     }
 
--- a/src/Illuminate/Collections/LazyCollection.php
+++ b/src/Illuminate/Collections/LazyCollection.php
@@ -697,7 +697,7 @@ class LazyCollection implements CanBeEscapedWhenCastToString, Enumerable
         if (is_callable($value)) {
             return $this->map($value);
         }
-        
+
         return new static(function () use ($value, $key) {
             [$value, $key] = $this->explodePluckParameters($value, $key);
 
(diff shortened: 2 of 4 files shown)


levu42:
Seems that there's some test breaking for PHP8 regarding Lazy Collections and some Closure (where does the Closure come from? There's no Closure in the test I wrote). I don't understand, why the test is failing, so it would be great if someone who does understand it could have a quick look and point me in the right direction to fix it.

In my understanding `$collection->pluck->id` should work in all cases that `$collection->map->id` works as well, shouldn't it?

mikeydevelops:
It is not only `LazyCollection` that is affected, it is `Collection` too.

See:
```log
There was 1 error:

1) Illuminate\Tests\Support\SupportCollectionTest::testHigherOrderCollectionPluck with data set #0 ('Illuminate\Support\Collection')
TypeError: explode(): Argument #2 ($string) must be of type string, Closure given

/home/runner/work/framework/framework/src/Illuminate/Collections/helpers.php:52
/home/runner/work/framework/framework/src/Illuminate/Collections/Arr.php:469
/home/runner/work/framework/framework/src/Illuminate/Collections/Collection.php:706
/home/runner/work/framework/framework/src/Illuminate/Collections/HigherOrderCollectionProxy.php:47
/home/runner/work/framework/framework/tests/Support/SupportCollectionTest.php:893

--

There was 1 failure:

1) Illuminate\Tests\Support\SupportCollectionTest::testHigherOrderCollectionPluck with data set #1 ('Illuminate\Support\LazyCollection')
Failed asserting that Illuminate\Support\LazyCollection Object &000000005a50c8f5000000007e4719ee (
    'source' => Closure Object &000000005a50ca2c000000007e4719ee (
        0 => Closure Object &000000005a50ca2c000000007e4719ee
    )
    'escapeWhenCastingToString' => false
) is identical to Array &0 (
    0 => 23
    1 => 42
    2 => 1
).
```

You see, all methods listed in the `$proxies` array must support a closure as an argument, if you go through the list and see the definitions of each of those methods in the Collection classes you will see that. Also the way the `HigherOrderCollectionProxy` works, it calls the proxied method with a closure as an argument. See magic method [`__get`](https://github.com/laravel/framework/blob/066eb53db841aa275c0afab4549a58bf581e14f4/src/Illuminate/Collections/Traits/EnumeratesValues.php#L961) in `HigherOrderCollectionProxy`.

Also as an aside, PHPUnit has these [`dataProviders`](https://phpunit.readthedocs.io/en/9.5/writing-tests-for-phpunit.html#data-providers] which basically call the test as many times as there are items in the array returned by the dataProvider method specified in the doc in this case the method is `collectionClassProvider` and as you can see it returns an array with 2 items - array of parameters for the each test iteration, so it calls your method `testHigherOrderCollectionPluck` twice, first with `Collection` as argument and second time with `LazyCollection` as an argument.

levu42:
@mikeydevelops Thank you so much for the explanation! I'll amend the PR accordingly :)

mikeydevelops:
I guess you could add a check to see if the `$value` is callable, if so just return the result of the `map` method of the collection.

Like so:
```php
    public function pluck($value, $key = null)
    {
        if (is_callable($value)) {
            return $this->map($value);
        }

        return new static(Arr::pluck($this->items, $value, $key));
    }
```
It should be similar for the `LazyCollection` too.

levu42:
@mikeydevelops thanks for all your help, it's working now

taylorotwell:

PR #41574, opened 2022-03-19
[8.x] Added test for unsetting relations on model refresh

Patabugen:
[The Docs](https://laravel.com/docs/8.x/eloquent#refreshing-models) say that when calling `$model->refresh()` the models relationships will also be refreshed:

> In addition, all of its loaded relationships will be refreshed as well

I expected this to un-set a relation which no longer exists in the database.

However I noticed that if the relation is already loaded - it does not get unset.

I've tested this on branch 8.x and 9.x

### Example
I've added a failing test to demonstrate the problem.

Adding `$child->unsetRelation('parent');` right after `$child->refresh();` gives the result I expected.

### Fix
I'm not sure the best way or place to fix - happy to submit a PR to make the test pass with a bit of guidance if this is desired behaviour.

One thought was that the refresh() method should eitehr call `$model->unsetRelations()` or individually `$model->unsetRelation()` right before it calls `$model->load()`:

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/Model.php#L1550

### DIFF
--- a/tests/Integration/Database/EloquentModelRefreshTest.php
+++ b/tests/Integration/Database/EloquentModelRefreshTest.php
@@ -71,6 +71,36 @@ class EloquentModelRefreshTest extends DatabaseTestCase
 
         $post->children->first()->refresh();
     }
+
+    public function testItUnsetsRelations()
+    {
+        Schema::create('post_children', function (Blueprint $table) {
+            $table->increments('id');
+            $table->string('title');
+            $table->bigInteger('parent_id')->nullable();
+            $table->timestamps();
+            $table->softDeletes();
+        });
+
+        $parent = Post::create(['title' => 'parent']);
+        $child = ChildPost::create(['title' => 'child']);
+
+        $child->parent()->associate($parent)->save();
+
+        $this->assertInstanceOf(
+            Post::class,
+            $child->parent
+        );
+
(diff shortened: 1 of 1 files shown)


driesvints:
Heya, you submitted this from an organization which doesn't allows us to edit the PR. Please submit this PR from your personal account. Thanks

Patabugen:
@driesvints Ah sorry about that, I can do that.

A quick question if I may first:

Is there a way to do it from the organisation account at all (I'm trying to organise things better and do them as the Org) or is the only way to re-fork with the personal account and start a PR from there?

Patabugen:
Done, I've transferred the Fork and PR to my personal account and ticked the box to give you access!

Can this PR be re-opened or shall I create a new one?

driesvints:
Cool, thanks 👍

taylorotwell:
I'm not sure if we will be taking any action on this issue to be honest.

If someone wanted to take a stab at a PR I could review it but if it was a complex change it likely would not be merged.

Patabugen:
Thanks @taylorotwell .

I've worked around this by overriding the refresh() method on my models so will probably not look into this further at the moment.

```
    public function refresh()
    {
        $this->unsetRelations();
        return parent::refresh();
    }
```

PR #41576, opened 2022-03-19
[9.x] Add support for Rule::enum()

lbausch:
This adds support for using `Rule::enum()`, mainly to make enum validation also available via the `Rule` class.


taylorotwell:
Seems unnecessary I think?

lbausch:
I think it would provide more consistency when all the available validation rules can be used via the Rule class. Actually, I was surprised to find that `Password` and `Enum` are not available via static `Rule` methods.

PR #41578, opened 2022-03-20
[9.x] remove useless if statement from SupportReflectorTest

hAz5:
because of Laravel9 min PHP requirement is 8 and above , this if statement is useless

```
if (PHP_MAJOR_VERSION >= 8) {
    eval('
namespace Illuminate\Tests\Support;
```

### DIFF
--- a/tests/Support/SupportReflectorTest.php
+++ b/tests/Support/SupportReflectorTest.php
@@ -89,17 +89,12 @@ class B extends A
     }
 }
 
-if (PHP_MAJOR_VERSION >= 8) {
-    eval('
-namespace Illuminate\Tests\Support;
-
 class C
 {
     public function f(A|Model $x)
     {
         //
     }
-}');
 }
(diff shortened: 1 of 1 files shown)


tests/Support/SupportReflectorTest.php
```diff
@@ -89,8 +89,7 @@ public function f(parent $x)
     }
 }
 
-if (PHP_MAJOR_VERSION >= 8) {
-    eval('
+eval('
```
GrahamCampbell, line 92: We don't need this eval, either.

GrahamCampbell (changes requested):

tests/Support/SupportReflectorTest.php
hAz5: thank you for your review :)
i fix it

driesvints:
Thanks!

PR #41579, opened 2022-03-20
[9.x] Add startingAt and endingAt to console scheduler

jwpage:
This PR adds the `startingAt()` and `endingAt()` helper functions to the Artisan scheduler, for the situations where you want to start running a command only after or before a given date. The functions accept a DateTime or string, which is parsed to a Carbon instance.

Tests are attached to check that:

* startingAt in the past passes filters
* startingAt in the future does not pass filters
* endingAt in the future passes filters
* endingAt in the past does not pass filters

If this PR is merged I can make a pull request to the docs repo to add relevant documentation.

### DIFF
--- a/src/Illuminate/Console/Scheduling/ManagesFrequencies.php
+++ b/src/Illuminate/Console/Scheduling/ManagesFrequencies.php
@@ -20,6 +20,32 @@ trait ManagesFrequencies
     }
 
     /**
+     * Schedule the event to start running at a specific datetime.
+     *
+     * @param DateTime|string $end
+     * @return $this
+     */
+    public function startingAt(DateTime $start)
+    {
+        $start = Carbon::parse($start)->timezone($this->timezone);
+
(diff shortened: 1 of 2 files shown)


jwpage:
Just realised this doesn't play nicely with `artisan schedule:list` because it's using `Event::when()` and will show the next run date earlier than they actually start running.

Marking this as draft for now.

src/Illuminate/Console/Scheduling/ManagesFrequencies.php
```diff
@@ -19,6 +19,32 @@ public function cron($expression)
         return $this;
     }
 
+    /**
+     * Schedule the event to start running at a specific datetime.
+     *
+     * @param \DateTime|string $end
```
seriquynh, line 25: This param docblock is wrong. It must be $start.

driesvints:
Going to close this due to inactivity. Feel free to attempt a new PR if you decide to pick this up again. Thanks

PR #20340, opened 2017-07-31
[5.4] Fix 'Too Many Connections' error during testing using `$connectionsToTransact`

twig1337:
Terminate user defined database connections after rollback while using the DatabaseTransactions trait.

_Note: This bug fix is pointed to the 5.4 branch as the corresponding code doesn't exist in the 5.1 release._

Issue #18471, opened 2017-03-24
PDO Too many connections when running test suite


ryrobbo:
- Laravel Version: 5.4.16
- PHP Version: 7.1
- Database Driver & Version: MySQL

### Description:

I have a test suite which now has 140 tests. Each of my tests uses the DatabaseMigrations trait along with a dedicated database solely for tests to run against.

When PHP Unit gets to the last test it fails with the following error:

    Caused by
    Doctrine\DBAL\Driver\PDOException: SQLSTATE[08004] [1040] Too many connections

I can run this test individually and it passes fine.

After finding this issue (https://github.com/laravel/framework/issues/10619#issuecomment-151075567) I tried the following but the problem still persists:

- Added the `PDO::ATTR_PERSISTENT => false` option to my test database connection entry
- Added the following code to the DatabaseMigrations trait:

```php
$this->beforeApplicationDestroyed(function () {
    $this->artisan('migrate:rollback');

    foreach ($this->app->make('db')->getConnections() as $connection) {
        $connection->disconnect();
    }
});
```

### Steps To Reproduce:

N/A

Kyslik:
Its problem with `max_connections`.

Increase it using (run query)

```
set global max_connections = 200;
```

---

To check current `max_connections`

```
show variables like "max_connections";
```

---

I tested `max_connections = 10` and run 60 tests and it failed with 
```
Caused by
PDOException: SQLSTATE[HY000] [1040] Too many connections
```

sisve:
I believe the problem is that the connection should be reused (or a few of them pooled), not that the database has a too low max_connections value.

Kyslik:
What exactly do you mean reused? Connection is closed after each test, and until its closed new one opens up...

I tested this using Innotop on my local machine (installed from homebrew), I have 69 test total, and each test class uses `DatabaseMigrations` trait and Innotop showed 69 connections. I guess PHPUnit is fast so connections are not closed ASAP.

I found this [post](https://www.neontsunami.com/posts/too-many-connections-using-phpunit-for-testing-laravel-51) about this issue.

```
public function tearDown()
{
    $this->beforeApplicationDestroyed(function () {
        DB::disconnect();
    });

    parent::tearDown();
}
```

After that `max_connection` stats were ±3.

patinthehat:
I had a similar problem not too long ago, and solved it with:
```php
    public function tearDown()
    {
        \DB::connection()->setPdo(null);
    }```

arubacao:
[...]

Issue #18056, opened 2017-02-22
Postgre "too many clients already" on running tests


mariomka:
- Laravel Version: 5.4.12
- PHP Version: PHP 7.1.0-2+deb.sury.org~xenial+1 (cli) ( NTS )
- Database Driver & Version: psql (PostgreSQL) 9.5.5

(Running on Homestead)

### Description:
On running more than 100 tests that use ``DatabaseMigrations`` the following error appears:

```
Illuminate\Database\QueryException: SQLSTATE[08006] [7] FATAL:  sorry, too many clients already
```

### Steps To Reproduce:
Run more than 100 tests (methods) using ``DatabaseMigrations`` on Postgree (default max connections is 100).

I have created a repository to ease reproduce it: https://github.com/mariomka/laravel-pgsql-too-many-clients-test

- git clone https://github.com/mariomka/laravel-pgsql-too-many-clients-test.git
- composer install
- php vendor/bin/homestead make
- vagrant up
- vagrant ssh
- cd laravel-pgsql-too-many-clients-test
- ./vendor/bin/phpunit

While tests are running you can see Postgre connections count with:

```
select count(*) from pg_stat_activity;
```

Or full info with:
```
select * from pg_stat_activity;
```

apstyx:
I am experiencing the same error with one difference, not making use of DatabaseMigrations in my tests.

wins1908:
I have the fix for lumen above

wins1908:
here is PR for Laravel #18752

RobvH:
Yep, same here, see my post on the PR for details.

mfn:
I believe this was fixed (for Laravel) in 5.5 and at least backported to 5.4 (or it was introduced in 5.4, don't remember); I know this issue well because I had the same problem in the past.

It's all connected to this trait `\Illuminate\Foundation\Testing\DatabaseTransactions`.

See the code in 5.3:
```php
    public function beginDatabaseTransaction()
    {
        $database = $this->app->make('db');

        foreach ($this->connectionsToTransact() as $name) {
            $database->connection($name)->beginTransaction();
        }

        $this->beforeApplicationDestroyed(function () use ($database) {
            foreach ($this->connectionsToTransact() as $name) {
                $database->connection($name)->rollBack();
            }
        });
    }
```

Now compare it to 5.4/5.5
```php
    public function beginDatabaseTransaction()
    {
        $database = $this->app->make('db');

        foreach ($this->connectionsToTransact() as $name) {
[...]

PR #41584, opened 2022-03-20
[8.x] Close doctrineConnection on disconnect

lcharette:
Consider this test:

```php 
class ExampleTest extends TestCase
{
    use DatabaseMigrations; 

    protected function tearDown(): void
    {
        // parent::tearDown(); <-- Omitted on purposed (or not)

        DB::disconnect();
    }

    public function test_1()
    {
        $this->assertTrue(true);
    }

    public function test_2()
    {
        $this->assertTrue(true);
    }

    // Repeat test 102x times...

    public function test_105()
    {
        $this->assertTrue(true);
    }
}
```

With a migration that modify a column: 
```php 
public function up()
{
        $table->increments('id');
        $table->string('user_name', 50);
    });

        $table->string('user_name', 120)->change();
    });
}
```

On PostgreSQL, after the 100th iteration, this error will pop up:
```txt
Illuminate\Database\QueryException: SQLSTATE[08006] [7] connection to server at "127.0.0.1", port 5432 failed: FATAL:  sorry, too many clients already (SQL: create table "migrations" ("id" serial primary key not null, "migration" varchar(255) not null, "batch" integer not null))
```

While disconnect [sets the “main” PDO object as `null`](https://github.com/laravel/framework/blob/b9203fca96960ef9cd8860cb4ec99d1279353a8d/src/Illuminate/Database/Connection.php#L825-L828), the [Doctrine connection created by the “change”](https://github.com/laravel/framework/blob/b9203fca96960ef9cd8860cb4ec99d1279353a8d/src/Illuminate/Database/Connection.php#L1008) action is not. The actual connection to the database is not actually closed in this specific case as the underlying PDO object is not fully destroyed as per PHP Documentation: https://www.php.net/manual/en/pdo.connections.php. 

This bug won't affect any migration that are not modifying a column, as the Doctrine Connection is only created when needed, eg. using `->change()`. 

I understand this fix won’t have much impact on a normal Laravel install, as :
1. Laravel TestCase undefined the whole app on `tearDown()`, Connection included, closing the Doctrine connection. 
1. This test is pointless and should call `parent::tearDown();`

But in a situation where `disconnect()` is called outside of a test scenario, or anyone not using the whole Laravel TestCase _might_ [encounter](https://github.com/userfrosting/sprinkle-account/runs/5579295094?check_suite_focus=true#step:12:23) [this issue](https://bbqsoftwares.com/blog/postgre-too-many-connection). 

References : 
- https://github.com/laravel/framework/issues/18056
- https://github.com/laravel/framework/issues/18471
- https://github.com/laravel/framework/pull/20340

### DIFF
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -825,6 +825,7 @@ class Connection implements ConnectionInterface
     public function disconnect()
     {
         $this->setPdo(null)->setReadPdo(null);
+        $this->doctrineConnection = null;
     }
 
     /**
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -825,6 +825,7 @@ class Connection implements ConnectionInterface
     public function disconnect()
     {
         $this->setPdo(null)->setReadPdo(null);
+
         $this->doctrineConnection = null;
     }
 


jonaszkadziela:
@lcharette @taylorotwell Are you sure that setting `$this->doctrineConnection` to `null` actually closes the PDO connection that is wrapped inside the `Doctrine\DBAL\Connection` class? I'm not sure if the PHP garbage collector will clean up the PDO connection because it is still referenced by the `$_conn` property in the `Doctrine\DBAL\Connection`.

PHP GC Cleanup Problems: https://www.php.net/manual/en/features.gc.refcounting-basics.php#features.gc.cleanup-problems

PR #41586, opened 2022-03-21
[9.x] Adds `isEqual` to compare file hashes in filesystem

DarkGhostHunter:
## What?

Allows to compare two files if these exists and their hashes are the same.

```php
use Illuminate\Support\Facades\File;

if (File::isEqual('foo.txt', 'bar.txt')) {
   return 'Both files are equal.'
}
```

It will always return `false` if one of the compared files doesn't exist.

## Why?

Useful to check if a file copied was later edited without using `exists()` and `hash()` damn every time.

## BC?

Is not expected the developer to have a macro of the same name for different functionality.

### DIFF
--- a/src/Illuminate/Filesystem/Filesystem.php
+++ b/src/Illuminate/Filesystem/Filesystem.php
@@ -503,14 +503,14 @@ class Filesystem
      * Check if both files are the same by comparing their hash.
      *
      * @param  string  $file
-     * @param  string  $comparable
+     * @param  string  $compared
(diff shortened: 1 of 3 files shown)


vlakoff:
Notes:
* The method has been named `hasSameHash()`, instead of the initially proposed `isEqual()`.
* It has been followed up by #49721.

vlakoff:
Now that we can replace MD5 with xxh128, I have submitted PR #54690.

PR #41591, opened 2022-03-21
[9.x] Make reconnectIfMissingConnection public

mfn:
## Summary
I've use-cases where I need to get the PDO from a connection but also
need to respect, in case a disconnect happened, to perform a reconnect.

Currently, I've to:
- get pdo
- check if null
- call reconnect

everywhere I need it. Or I could just call this one on the connection
object, if it were public, and then get the PDO.

### DIFF
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -814,7 +814,7 @@ class Connection implements ConnectionInterface
      *
      * @return void
      */
-    protected function reconnectIfMissingConnection()
+    public function reconnectIfMissiegConnection()
     {
         if (is_null($this->pdo)) {
             $this->reconnect();


src/Illuminate/Database/Connection.php
```diff
@@ -814,7 +814,7 @@ public function reconnect()
      *
      * @return void
      */
-    protected function reconnectIfMissingConnection()
+    public function reconnectIfMissiegConnection()
```
chrysanthos, line 817:
```suggestion
    public function reconnectIfMissingConnection()
```

driesvints:
Changing the visibility is unfortunately a breaking change: https://3v4l.org/vppve

Best to send this to master.

driesvints:
Please resubmit this to master and with passing tests. Thanks

src/Illuminate/Database/Connection.php
mfn: 🤦🏼 

thanks, no idea how that happened

mfn:
Re-submitted: https://github.com/laravel/framework/pull/41594

PR #41603, opened 2022-03-22
[9.x] remove extra useless parameter for method serialize in EncryptCookie

hAz5:
in serialize method we got $name parameter but we doesn't use it anywhere in method body

```
public static function serialized($name)
    {
        return static::$serialize;
    }
```

### DIFF
--- a/src/Illuminate/Cookie/Middleware/EncryptCookies.php
+++ b/src/Illuminate/Cookie/Middleware/EncryptCookies.php
@@ -135,7 +135,7 @@ class EncryptCookies
     {
         return is_array($cookie)
                         ? $this->decryptArray($cookie)
-                        : $this->encrypter->decrypt($cookie, static::serialized($name));
+                        : $this->encrypter->decrypt($cookie, static::serialized());
     }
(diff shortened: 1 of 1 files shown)


src/Illuminate/Cookie/Middleware/EncryptCookies.php
```diff
@@ -216,10 +216,9 @@ public function isDisabled($name)
     /**
      * Determine if the cookie contents should be serialized.
      *
-     * @param  string  $name
      * @return bool
      */
-    public static function serialized($name)
+    public static function serialized()
```
driesvints, line 221: Removing this is a breaking change for anyone overwriting this method.

hAz5: thank you for your review, 
so we just need a test for cookie serialization?

driesvints: No, I'm saying if you remove this people's apps that override this method will break. Just leave this be.

PR #41608, opened 2022-03-22
[9.x] Display how long the command took to complete if timer is set to true

b4oshany:
Display how long the command took to complete if the timer is true

### DIFF
--- a/src/Illuminate/Console/Command.php
+++ b/src/Illuminate/Console/Command.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Console;
 
+use Illuminate\Support\Carbon;
 use Illuminate\Support\Traits\Macroable;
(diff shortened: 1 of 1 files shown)


src/Illuminate/Console/Command.php
```diff
@@ -166,9 +177,15 @@ public function run(InputInterface $input, OutputInterface $output): int
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        $start_time = Carbon::now();
```
GrahamCampbell, line 180: This is not a good way to time things in PHP. use `hrtime`.

GrahamCampbell (changes requested):
I don't think this makes sense to add. Just run your command with `time` in front.

PR #41609, opened 2022-03-22
[9.x] Improve types on factory state() method

mdpoulter:
As per the Laravel docs (https://laravel.com/docs/9.x/database-testing#has-many-relationships), the `state()` method can accept a callable with 2 parameters, e.g. 

```php
$user = User::factory()
        ->has(
            Post::factory()
                    ->count(3)
                    ->state(function (array $attributes, User $user) {
                        return ['user_type' => $user->type];
                    })
        )
        ->create();
```

Currently, the type-hinting on the `state()` method doesn't support this (and throws an error when using static analysis). This PR adds the optional second parameter to the type hint.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/Factory.php
+++ b/src/Illuminate/Database/Eloquent/Factories/Factory.php
@@ -469,7 +469,7 @@ abstract class Factory
     /**
      * Add a new state transformation to the model definition.
      *
-     * @param  (callable(array<string, mixed>): array<string, mixed>)|array<string, mixed>  $state
+     * @param  (callable(array<string, mixed>, \Illuminate\Database\Eloquent\Model|null=): array<string, mixed>)|array<string, mixed>  $state
      * @return static
      */
     public function state($state)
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Eloquent/Factories/Factory.php
```diff
@@ -469,7 +469,7 @@ protected function expandAttributes(array $definition)
     /**
      * Add a new state transformation to the model definition.
      *
-     * @param  (callable(array<string, mixed>): array<string, mixed>)|array<string, mixed>  $state
+     * @param  (callable(array<string, mixed>, \Illuminate\Database\Eloquent\Model|null=): array<string, mixed>)|array<string, mixed>  $state
```
GrahamCampbell, line 472: This is invalid syntax

GrahamCampbell (changes requested):

src/Illuminate/Database/Eloquent/Factories/Factory.php
mdpoulter: According to PHPStan (https://phpstan.org/writing-php-code/phpdoc-types#callables), the equals represents an optional parameter, but I'm not sure if that's what you're referring to.

GrahamCampbell: Ah, nice. Thanks. :)

PR #41614, opened 2022-03-22
[8.x] Fix require fails if is_file cached by opcache

### LINKED ISSUE #41613: php artisan config:cache error
- Laravel Version: 9.5.1
- PHP Version: 8.0.16
- Database Driver & Version: Mariadb but doesn't matter

### Steps To Reproduce:

Just run `php artisan config:cache` a bunch of times, around 1 out of 2 times (if it was already cached) the following error appears

```txt
Configuration cache cleared successfully.

   ErrorException

  require(releases/55a338ab2cb555eb6a03cbcdd176d1f1473d89ef/bootstrap/cache/config.php): Failed to open stream: No such file or directory

  at vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/LoadConfiguration.php:28
     24▕         // First we will see if we have a cache configuration file. If we do, we'll load
     25▕         // the configuration items from that file so that it is very quick. Otherwise
     26▕         // we will need to spin through every configuration file and load them all.
     27▕         if (is_file($cached = $app->getCachedConfigPath())) {
  ➜  28▕             $items = require $cached;
     29▕
     30▕             $loadedFromCache = true;
     31▕         }
     32▕

      +18 vendor frames
  19  artisan:37
      Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
```

This seems to be a problem with `opcache.enable_file_override` which caches the result of is_file

Tofandel:
Fixes #41613 

In the case where `opcache.enable_file_override` is enabled, the `is_file` call might return `true` because it's cached until the end of the script so the require will fail if it was unlinked during the execution which is the case when running the command `artisan config:cache`
This just changes `is_file` to `file_exists` which is the method to use before a require and is not affected by this issue

### DIFF
--- a/src/Illuminate/Foundation/Bootstrap/LoadConfiguration.php
+++ b/src/Illuminate/Foundation/Bootstrap/LoadConfiguration.php
@@ -24,7 +24,7 @@ class LoadConfiguration
         // First we will see if we have a cache configuration file. If we do, we'll load
         // the configuration items from that file so that it is very quick. Otherwise
         // we will need to spin through every configuration file and load them all.
-        if (is_file($cached = $app->getCachedConfigPath())) {
+        if (file_exists($cached = $app->getCachedConfigPath())) {
             $items = require $cached;
 
             $loadedFromCache = true;


driesvints:
It's best that you explain what you're doing here or your PR will be closed (see the PR template when submitting PR's).

driesvints:
Also if this also exists in 8.x it should be sent in to that branch: https://laravel.com/docs/9.x/contributions#which-branch

Tofandel:
It does exist in 8.x so, I'll rebase the PR

driesvints:
The main issue here probably is that you're enabling opcache on the CLI which I don't see a use case for.

Tofandel:
I know but it's not always in our power to change this with some hosts

driesvints:
Why did you change the branch again? Can you also update the PR title when you do?

Tofandel:
After some testing, no need for try catch, file_exists fixes the issue entirely

PR #41615, opened 2022-03-22
[9.x] Store SES message ID in X-SES-Message-ID header

jdavidbakr:
Restoring previous functionality with SES mail driver to store the SES message ID in X-SES-Message-ID header.  Needed to support SES message tracking in https://github.com/jdavidbakr/mail-tracker/issues/155

### DIFF
--- a/src/Illuminate/Mail/Transport/SesTransport.php
+++ b/src/Illuminate/Mail/Transport/SesTransport.php
@@ -75,7 +75,7 @@ class SesTransport extends AbstractTransport
         } catch (AwsException $e) {
             throw new Exception('Request to AWS SES API failed.', $e->getCode(), $e);
         }
-        
+
         $messageId = $result->get('MessageId');
         $message->getOriginalMessage()->getHeaders()->addHeader('X-Message-ID', $messageId);
         $message->getOriginalMessage()->getHeaders()->addHeader('X-SES-Message-ID', $messageId);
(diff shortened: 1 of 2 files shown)


driesvints:
Yes I think this is a good solution for now. It's unfortunate that the header can't be part of SentMessage where it should be imo but this might be good enough for now.

@taylorotwell this will allow us to revert part of the upgrade guide PR I sent in yesterday.

src/Illuminate/Mail/Transport/SesTransport.php
```diff
@@ -72,6 +72,10 @@ protected function doSend(SentMessage $message): void
                     ]
                 )
             );
+            $messageId = $result->get('MessageId');
+
+            $message->getOriginalMessage()->getHeaders()->addHeader('X-Message-ID', $messageId);
+            $message->getOriginalMessage()->getHeaders()->addHeader('X-SES-Message-ID', $messageId);
```
driesvints, line 78: Move this block below the `catch` part.

driesvints (changes requested):

src/Illuminate/Mail/Transport/SesTransport.php
jdavidbakr: 1effd1ba0 has this change

driesvints (approved):

PR #41616, opened 2022-03-22
[9.x] Adds class discoverer `Discover`

DarkGhostHunter:
## What?

Introduces a class specialized in discovering files as PHP Classes. Rework of the `DiscoverEvents`, but generalized.

```php
use Illuminate\Filesystem\Discover;

$events = Discover::inside('Events')->all();
```

The main idea is to detach the _discover_ mechanism so other services or packages can also discover files from within a path, like Policies. As with the `DiscoverEvents`, it only collects instantiable classes.

```php
use Illuminate\Filesystem\Discover;
use ReflectionClass;

$policies = Discover::inside('Policies')
    ->all()
    ->filter(function (ReflectionClass $policy) {
        return str_ends_with($policy->name, 'Policy');
    });
```

- It receives an `Application` instance, so it can be mocked and tested.
- It's multi-tenancy friendly, as it supports changing the root dir and root namespace.
- Can be set to recursive.
- Returns a `Collection`, so further filters can be applied over it.

```php
use Illuminate\Filesystem\Discover;

// Find all classes in `/www/projects/laravel/myProject/Components`.
$policies = Discover::inside('Components')
    ->recursively()
    ->atNamespace('myProject', 'Project\\')
    ->all()
```

## Why?

Basically, centralize the discovery of Classes. This can be done for Policies, Events and what not. May be Components can benefit from this approach as a way to cache them.

## BC?

None, is only additive. May be in next version we could get rid of `DiscoverEvents` and whatever the Gate does to find Policies (it doesn't work recursively last time I checked).

> PS: Should I PR this to `master` instead?

### DIFF
--- a/src/Illuminate/Filesystem/Discover.php
+++ b/src/Illuminate/Filesystem/Discover.php
@@ -6,7 +6,6 @@ use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Support\Str;
 use ReflectionClass;
 use ReflectionException;
-use const DIRECTORY_SEPARATOR;
 
 class Discover
 {
@@ -117,7 +116,7 @@ class Discover
                 continue;
             }
 
-            if (!$reflection->isInstantiable()) {
+            if (! $reflection->isInstantiable()) {
                 continue;
             }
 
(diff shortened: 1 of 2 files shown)


DarkGhostHunter:
I’ll take that as a PR too master. 

Italo Baeza C.

PR #41625, opened 2022-03-22
[9.x] Change switch case statement to match expression in write output

al1r374:
maybe we can take advantage of match expression for increasing readability

### DIFF
--- a/src/Illuminate/Queue/Console/WorkCommand.php
+++ b/src/Illuminate/Queue/Console/WorkCommand.php
@@ -182,14 +182,11 @@ class WorkCommand extends Command
      */
     protected function writeOutput(Job $job, $status)
     {
-        switch ($status) {
-            case 'starting':
-                return $this->writeStatus($job, 'Processing', 'comment');
-            case 'success':
-                return $this->writeStatus($job, 'Processed', 'info');
-            case 'failed':
-                return $this->writeStatus($job, 'Failed', 'error');
-        }
+        match ($status) {
+            'starting' => $this->writeStatus($job, 'Processing', 'comment'),
+            'success' => $this->writeStatus($job, 'Processed', 'info'),
+            'failed' => $this->writeStatus($job, 'Failed', 'error')
+        };
     }
 
     /**


src/Illuminate/Queue/Console/WorkCommand.php
```diff
@@ -182,14 +182,11 @@ protected function listenForEvents()
...
     {
-        switch ($status) {
-            case 'starting':
-                return $this->writeStatus($job, 'Processing', 'comment');
-            case 'success':
-                return $this->writeStatus($job, 'Processed', 'info');
-            case 'failed':
-                return $this->writeStatus($job, 'Failed', 'error');
-        }
+        match ($status) {
```
browner12, line 185: you need to `return` the result of the `match` to mimic the previous functionality

alirezadp10: the writeStatus method has void type and does not return anything already

browner12: correct, but you need to match the behavior of the existing `writeOutput` method, which `return`s the result of the `writeStatus` method. `writeStatus()` returning `void` is irrelevant.

ankurk91:
match will throw an exception when none of values matches, while switch case will silently ignore it, beware.

alirezadp10:
> match will throw an exception when none of values matches, while switch case will silently ignore it, beware. It is suggested have a `default => throw new \RuntimeException("Message")`

Thank you for the warning, but it will not be something other than starting, success, failed. so maybe if we keep it simple is better.

src/Illuminate/Queue/Console/WorkCommand.php
GrahamCampbell: nothing to change here.

GrahamCampbell (approved):

alirezadp10:
> match will throw an exception when none of values matches, while switch case will silently ignore it, beware. It is suggested have a `default => throw new \RuntimeException("Message")`

Thank you for the warning, but it will not be something other than starting, success, failed. so maybe if we keep it simple is better.

GrahamCampbell (changes requested):
👎 for merging. match is meant to be used with expressions with a return value returning void from lines in a match is silly. match is not a swap out for switch, just like fn is not for function. The code was already fine as is, and shouldn't be changed.

X-Coder264:
> -1 for merging. **match is meant to be used with expressions with a return value** returning void from lines in a match is silly. match is not a swap out for switch, just like fn is not for function. The code was already fine as is, and shouldn't be changed.

No, it's not. Just because `match` has the ability to return a value it does not mean it must return a value.

An example from the `match` RFC itself:

```php
match ($pressedKey) {
    Key::RETURN_ => save(),
    Key::DELETE => delete(),
};
```

https://wiki.php.net/rfc/match_expression_v2

IMO there's nothing wrong with improving the (readability of the) code so I don't understand the "code was already fine as is and shouldn't be changed" argument, especially since Taylor already merged a bunch of similar PRs recently.

src/Illuminate/Queue/Console/WorkCommand.php
browner12: strongly disagree. if you didn't already have the docblock, or where able to look into the `writeStatus` method, you would have no clue that it currently returns `void`. Therefore, if you want to match functionality in this PR targeting the current major version, you **need** to `return` the result of the call, just like we did prior.

If you want to change functionality, which is what dropping `return` would be, target 10.x.

X-Coder264: I strongly disagree with your strong disagreement :smile:

The `writeStatus` method is clearly documented as `void`. As Laravel does not have (especially scalar) typehints in most of its codebase the docblock is what defined the API contract of the method. If someone overrode the `writeStatus` method and made it actually return anything is in breach of that contract. So it should not be in Laravel backwards compatibility policy to care about "breaking" use cases that were never officially supported/allowed to begin with.

The only reason why there's a return in that switch statement is to not have to add an additional `break;` line, not to actually return some value.

PR #41626, opened 2022-03-22
[10.x] Adds Discover class

DarkGhostHunter:
## What?

Introduces a class specialized in discovering files as PHP Classes. Rework of the `DiscoverEvents`, but generalized.

```php
use Illuminate\Filesystem\Discover;

$events = Discover::inside('Events')->all();
```

The main idea is to detach the _discover_ mechanism so other services or packages can also discover files from within a path, like Policies. As with the `DiscoverEvents`, it only collects instantiable classes.

```php
use Illuminate\Filesystem\Discover;
use ReflectionClass;

$policies = Discover::inside('Policies')
    ->all()
    ->filter(function (ReflectionClass $policy) {
        return str_ends_with($policy->name, 'Policy');
    });
```

- It receives an `Application` instance, so it can be mocked and tested.
- It's multi-tenancy friendly, as it supports changing the root dir and root namespace.
- Can be set to recursive.
- Returns a `Collection`, so further filters can be applied over it.

```php
use Illuminate\Filesystem\Discover;

// Find all classes in `/www/projects/laravel/myProject/Components`.
$policies = Discover::inside('Components')
    ->recursively()
    ->atNamespace('myProject', 'Project\\')
    ->all()
```

## Why?

Basically, centralize the discovery of Classes. This can be done for Policies, Events and what not. May be Components can benefit from this approach as a way to cache them.

## Why draft?

Waiting on how is received to replace the `DiscoverEvents` class for this implementation.

## BC?

Assuming this goes through, it will break all services that rely on their own "discoverer" because it will be replaced by this. It also adds the `illuminate/filesystem` dependency to these packages.

### DIFF
--- a/src/Illuminate/Filesystem/Discover.php
+++ b/src/Illuminate/Filesystem/Discover.php
@@ -96,7 +96,7 @@ class Discover
                 continue;
             }
 
-            if (!$reflection->isInstantiable()) {
+            if (! $reflection->isInstantiable()) {
                 continue;
             }
 
--- a/src/Illuminate/Filesystem/Discover.php
+++ b/src/Illuminate/Filesystem/Discover.php
@@ -1,0 +1,170 @@
+<?php
+
+namespace Illuminate\Filesystem;
+
+use Illuminate\Contracts\Foundation\Application;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\File;
+use Illuminate\Support\Str;
+use Illuminate\Support\Stringable;
+use ReflectionClass;
+use ReflectionException;
+use Symfony\Component\Finder\SplFileInfo;
+use function lcfirst;
+use function ucfirst;
+use const DIRECTORY_SEPARATOR;
+
+class Discover
+{
+    /**
+     * Project path where all discoveries will be done.
+     *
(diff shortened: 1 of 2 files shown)


driesvints:
Taylor doesn't sees draft prs

DarkGhostHunter:
> Taylor doesn't sees draft prs

What about the rest? Nobody interested in changing the discovery behaviour?

GrahamCampbell:
Why is this targetting 10.x, and not 9.x?

DarkGhostHunter:
> Why is this targetting 10.x, and not 9.x?

Headshotted #41616

GrahamCampbell:
What. Taylor didn't say re-submit to master. He said he didn't want it?

DarkGhostHunter:
> What. Taylor didn't say re-submit to master. He said he didn't want it?

Just used the copy-paste response that I've seen around for anything.

I though it could have a chance here considering it was made to break those services that use auto-discovery on their own (I'm looking at you Events), while exposing to make it reusable externally.

driesvints:
It seems Taylor indeed didn't want to merge this so closing this sorry.

PR #41635, opened 2022-03-23
[9.x] Add contract for notification channels

al1r374:
As we call send method on notification channel [(NotificationSender.php)](https://github.com/laravel/framework/blob/19b0abf57abc74d09aa059057a5b6576f127c7dc/src/Illuminate/Notifications/NotificationSender.php#L148), maybe we could use contracts to implement it every notification channel.


taylorotwell:

PR #41636, opened 2022-03-23
[9.x] [Collection] Add retrieve of nested keys

daniele-sarnari:
**Title**: Add retrieve of nested keys

**Description**: Allow retrieve of nested keys using get method of Collection class. Important factor is that keys must be delimited by dot (.)

**Tests**: tested using existing and not existing keys

**Benefits**: I was trying to access deep keys using get method but without success. I think it is a useful feature and can avoid writing more code that what it is needed.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -440,6 +440,19 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
             return $this->items[$key];
         }
 
+        if (str_contains((string) $key, ".")) {
+            $keys = explode('.', (string) $key);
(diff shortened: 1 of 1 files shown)


taylorotwell:

DrewLiberty:
> Thanks for your pull request to Laravel!
> 
> Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include.
> 
> 
> If you feel absolutely certain that this code corrects a bug in the framework, please "**@**" mention me in a follow-up comment with further explanation so that GitHub will send me a notification of your response.

Ok i understand, thanks for the reply 👍

PR #31363, opened 2020-02-05
[7.x] Implement anonymous components

driesvints:
This PR introduces class-less components à la Blade X. What it does under the hood is make use of a `ClassLessComponent` class (wait whut?) to pipe through all the attributes as data to the component. This way you can do this:

**welcome.blade.php**

```
<x-avatar size="100" />
```

**components/avatar.blade.php**

```
@php($user = $user ?? Auth::user())
@php($size = $size ?? 50)

<img
    class="inline-block rounded-full"
    src="{{ $user->gravatarUrl($size) }}"
    width="{{ $size }}"
    height="{{ $size }}"
/>
```

All without a component class. This will be super useful for very simple small components and composition where you don't always need a class for.

### Update

Decided to rename `ClassLessCompnent` to `AnonymousComponent`.

PR #41637, opened 2022-03-23
[9.x] Discover anonymous Blade components in other folders

ralphjsmit:
Do you like anonymous components? (I like them a lot) And did you ever want to also 'enable' anonymous components for directories other than `resources/views/components`? 

For example, `resources/views/admin` and `resources/view/frontend`? And autodiscover the components like `<x-admin::component.name`?

This PR introduces exactly that. It allows the developer to register a directory and a prefix, which would allow to use the files in those directories as anonymous components. This would get rid of things like `@include` and provide developers with the ability to use the attribute bag everywhere.

## Example

In your service provider, you can register the directories you want to enable discovery for:

```php
// Prefix <x-admin::your.component> and folder: resources/views/admin 
Blade::anonymousComponentNamespace('admin', 'admin'); 

// Second parameter is nullable and defaults to first parameter, so this means the same:
Blade::anonymousComponentNamespace('admin'); 

// Prefix <x-frontend.auth::your.component> and folder: resources/views/public/auth
Blade::anonymousComponentNamespace('public/auth', 'frontend.auth');
```

Now you can use this in your Blade files like this:

```blade
<!-- Looks for resources/views/admin/dashboard/widget.blade.php OR resources/views/admin/dashboard/widget/index.blade.php -->
<x-admin::dashboard.widget>

<!-- Looks for resources/views/public/auth/login/form.blade.php OR resources/views/public/auth/login/form/index.blade.php -->
<x-frontend::login.form>
```

I added some additional inline comments to the compiler-file just for clarity, but feel free to trim this down however you want. 

I took inspiration from the original PR that introduced anonymous components (#31363). I'm a slightly unsure how this should be tested, perhaps @driesvints could give me a hint?

This is a feature I wanted for a long time and I hope this will be useful to many developers! The syntax is in line with package components.

### DIFF
--- a/src/Illuminate/View/Compilers/ComponentTagCompiler.php
+++ b/src/Illuminate/View/Compilers/ComponentTagCompiler.php
@@ -271,12 +271,39 @@ class ComponentTagCompiler
             return $class;
         }
 
-        if ($viewFactory->exists($view = $this->guessViewName($component))) {
-            return $view;
-        }
+        // The following code serves to guess the view name for this component.
+        // First, we'll create a collection with the custom component paths provided
+        // by the developer. We can use these folders to guess the correct view name.
+        $guess = collect($this->blade->getAnonymousComponentNamespaces())
+            // Next, we'll filter out the component paths that cannot be used here, based on their prefix.
+            // For example, for the anonymous component "<x-admin:dashboard>", we should look for the 'dashboard'
+            // view in the folder associated with the 'admin' prefix.
+            ->filter(function ($directory, $prefix) use ($component): bool {
+                return Str::startsWith($component, $prefix.'::');
+            })
+            // Next, we'll prepend the default components directory and our full component name. This is convention.
+            ->prepend('components', $component)
+            // Finally, we'll go over the components and prefixes and check whether we can find a matching view name.
+            ->reduce(function($carry, $directory, $prefix) use ($component, $viewFactory) {
+                $componentName = Str::after($component, $prefix.'::');
 
-        if ($viewFactory->exists($view = $this->guessViewName($component).'.index')) {
-            return $view;
+                if ($carry !== null) {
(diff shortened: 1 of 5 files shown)


src/Illuminate/View/Compilers/ComponentTagCompiler.php
```diff
@@ -341,11 +369,14 @@ public function formatClassName(string $component)
      * Guess the view name for the given component.
      *
      * @param  string  $name
+     * @param  string  $prefix
      * @return string
      */
-    public function guessViewName($name)
+    public function guessViewName($name, $prefix = 'components.')
```
ahinkle, line 375: Wouldn't this be a breaking change? 🙂

ralphjsmit: Thanks! I think that theoretically it would be a breaking change in the sense of people who override this class. However, when I do a search for ComponentTagCompiler, I see that this class is always instantiated like `new ComponentTagCompiler(/** */)`, so as far as I know people are technically not able to override this. So in that sense it wouldn't be a breaking change.

taylorotwell:
Do you have the ability to write tests for this?

ralphjsmit:
@taylorotwell Yes, if you're okay with the idea.

ralphjsmit:
I just added test coverage for this and applied the code style!

taylorotwell:
I think one thing that gives me pause about this how it behaves when used via a package. 

I could see package maintainer's using anonymous components and wanting to use this feature, but how would they use it? It seems to always assume the "directory" is a dot path relative to the application's main view directory.

taylorotwell:
Thanks!

I decided to swap the order of directory and prefix - I felt that was a bit more consistent with the componentNamespace method which accepts "location" first and then "prefix".

PR #41622, opened 2022-03-22
[9.x] Add newLine to Stringable

paulandroshchuk:
This PR adds a new method to Stringable to easily add new lines to the string.

```php
str('Laravel')
  ->newLine()
  ->append('The PHP Framework For Web Artisans');

// Laravel
// The PHP Framework for Web Artisans
```

PR #41638, opened 2022-03-23
[9.x] Add support for multiple new lines

dennisprudlo:
Based on https://github.com/laravel/framework/pull/41622 this PR adds additional support for multiple new lines like for the console output in Artisan commands:

https://github.com/laravel/framework/blob/6f743e236a49df03dd579933e2360a3304b55c5b/src/Illuminate/Console/Concerns/InteractsWithIO.php#L371-L382

So for the use case of @foremtehan in https://github.com/laravel/framework/pull/41622#issuecomment-1076301259 the code could look like this:

```php
return str($site)
    ->newLime(2)
    ->append($this-›sub-›name)
    ->newLine()
    ->append("<b>$newTitle</b>")
    ->newLine(2)
    -> ... you get the idea
```

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -67,11 +67,12 @@ class Stringable implements JsonSerializable
     /**
      * Append a new line to the string.
      *
+     * @param  int  $count
      * @return $this
      */
-    public function newLine()
+    public function newLine($count = 1)
     {
-        return $this->append(PHP_EOL);
+        return $this->append(str_repeat(PHP_EOL, $count));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


JawadR1:
I like the idea, but could it be better this way?

`public function newLine(){...}`
`public function newLines($count){...}`

src/Illuminate/Support/Stringable.php
```diff
@@ -67,11 +67,12 @@ public function append(...$values)
     /**
      * Append a new line to the string.
      *
+     * @param  int  $count
      * @return $this
      */
-    public function newLine()
+    public function newLine($count = 1)
```
GrahamCampbell, line 73: this is a major breaking change to a non-private method in a non-final class.

GrahamCampbell (changes requested):

src/Illuminate/Support/Stringable.php
dennisprudlo: This is a new method that was added a few hours ago. If I am not mistaken there wasn't any release since then so there should be nothing to break here.

dennisprudlo:
@JawadR1 yeah I thought about it like in carbon methods for example `addMonth()` and `addMonths($count)`. But I didn't see similar alias-methods in Laravel itself so I'll wait for the review

PR #41644, opened 2022-03-23
[9.x] Add database test for unique validation

masouddarvishi:
this PR adds database tests for unique validation. there wasn't any.


taylorotwell:
Should be an Integration test case in that folder and using Orchestra.

PR #41657, opened 2022-03-24
[9.x] Improve types for `Collection::whereInstanceOf` method

canvural:
This PR improves the types for `whereInstanceOf` collection method. So that we can know after a `->whereInstanceOf(User::class)` call, the collection will only have `User` as the values.


taylorotwell:
Marking as draft pending until @nunomaduro reviews.

PR #41661, opened 2022-03-24
[9.x] Add previousPath (standalone) to UrlGenerator

josephmiclaus:
This is the "standalone" version of the idea outlined in https://github.com/laravel/framework/discussions/41612 ("standalone" because `previousPath` doesn't depend on `previousWithoutQuery` as in PR https://github.com/laravel/framework/pull/41662)

In short this would enable:

Submit form when URL is e.g. `https://website.dev/foo?bar=1` then in the controller:

```php
return redirect(url()->previousPath()); // /foo
```

This will redirect to URL: `https://website.dev/foo` (without the query string)

---

PR for the "dependent" version: https://github.com/laravel/framework/pull/41662

### DIFF
--- a/src/Illuminate/Contracts/Routing/UrlGenerator.php
+++ b/src/Illuminate/Contracts/Routing/UrlGenerator.php
@@ -20,14 +20,6 @@ interface UrlGenerator
     public function previous($fallback = false);
 
     /**
-     * Get the previous path info for the request.
-     *
-     * @param  mixed  $fallback
-     * @return string
-     */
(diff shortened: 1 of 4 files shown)


seriquynh:
This is a breaking change and it can not be merged into 9.x branch.

miclaus:
@seriquynh Thanks for noticing! I removed the method from the contract. Should be good now.

taylorotwell:
Could you add some tests for this?

miclaus:
@taylorotwell Test added.

PR #41662, opened 2022-03-24
[9.x] Add previousWithoutQuery and previousPath to UrlGenerator

josephmiclaus:
This is the "dependent" version of the idea outlined in https://github.com/laravel/framework/discussions/41612 ("dependent" because `previousPath` depends on `previousWithoutQuery`)

In short this would enable:

Submit form when URL is e.g. `https://website.dev/foo?bar=1` then in the controller:

```php
return redirect(url()->previousWithoutQuery()); // https://website.dev/foo
// or
return redirect(url()->previousPath()); // /foo
```

This will redirect to URL: `https://website.dev/foo` (without the query string)

---

PR for the "standalone" version: https://github.com/laravel/framework/pull/41661

There is also a branch with implementation of solely `previousWithoutQuery` [here](https://github.com/miclaus/laravel-framework/tree/url-previous-without-query).

### DIFF
--- a/src/Illuminate/Contracts/Routing/UrlGenerator.php
+++ b/src/Illuminate/Contracts/Routing/UrlGenerator.php
@@ -20,6 +20,14 @@ interface UrlGenerator
     public function previous($fallback = false);
 
     /**
+     * Get the URL for the previous request without the query string.
+     *
+     * @param  mixed  $fallback
+     * @return string
+     */
+    public function previousWithoutQuery($fallback = false);
+
+    /**
      * Generate an absolute URL to the given path.
      *
      * @param  string  $path
--- a/src/Illuminate/Contracts/Routing/UrlGenerator.php
+++ b/src/Illuminate/Contracts/Routing/UrlGenerator.php
@@ -28,6 +28,14 @@ interface UrlGenerator
     public function previousWithoutQuery($fallback = false);
 
     /**
+     * Get the previous path info for the request.
+     *
+     * @param  mixed  $fallback
+     * @return string
+     */
+    public function previousPath($fallback = false);
+
+    /**
      * Generate an absolute URL to the given path.
      *
      * @param  string  $path
--- a/src/Illuminate/Contracts/Routing/UrlGenerator.php
+++ b/src/Illuminate/Contracts/Routing/UrlGenerator.php
@@ -20,22 +20,6 @@ interface UrlGenerator
     public function previous($fallback = false);
 
     /**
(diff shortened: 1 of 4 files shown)


src/Illuminate/Contracts/Routing/UrlGenerator.php
```diff
@@ -19,6 +19,22 @@ public function current();
      */
     public function previous($fallback = false);
 
+    /**
+     * Get the URL for the previous request without the query string.
+     *
+     * @param  mixed  $fallback
+     * @return string
+     */
+    public function previousWithoutQuery($fallback = false);
```
GrahamCampbell, line 28: adding methods to a non-internal interface is a major breaking change

GrahamCampbell (changes requested):

src/Illuminate/Contracts/Routing/UrlGenerator.php
miclaus: Oh.. I'm unsure now, but isn't the `UrlGenerator` interface internal?

GrahamCampbell: no. it's in the contracts repo. and not marked with `@internal`

miclaus: Do I understand correctly that other projects might use `laravel/framework` and thus the `UrlGenerator` contract have induce breaking changes?

miclaus: Could we simply add `@internal This method is not meant to be used or overwritten outside the framework.` for `previousWithoutQuery` and `previousPath` then? 

Or would it be better to simply not add these methods to the contract?

dennisprudlo: @miclaus It doesn't matter if the functions are marked with `@internal` because anyone who implements this interface will still have to add those two functions to their own implementation. If they don't their app will break.

miclaus: @dennisprudlo Got it now, thanks! I removed the methods from the contract.

taylorotwell:
Can you add tests to this?

miclaus:
@taylorotwell Added the tests.

Aside, I should mention that the "standalone" version of this PR would suffice and looks cleaner for the mentioned use-case.

dennisprudlo:
@miclaus Taylor doesn't review draft PRs so you should mark at least one of both PRs to open if you want them reviewed.

PR #41664, opened 2022-03-24
[9.x] Add function for testing HEAD requests

axlon:
Laravel already provides testing helpers for various request types such as `GET`, `POST`, `PUT` and `DELETE`. This PR adds a helper for `HEAD` request that works similarly to existing request helpers.

Example usage:

```php
use Tests\TestCase;

class MyTest extends TestCase
{
    public function testSomeApplicationEndpoint()
    {
        $this->head('/url-being-tested')->assertOk();
    }
}
```

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
@@ -451,6 +451,21 @@ trait MakesHttpRequests
     }
 
     /**
+     * Visit the given URI with a HEAD request.
+     *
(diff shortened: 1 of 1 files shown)


ahinkle:
You should be able to add this to your own application via macro, right? There are [so many other request types](https://www.iana.org/assignments/http-methods/http-methods.xhtml) that are not included. This would be a pretty low use-case.

axlon:
I could, but since the framework already takes this responsibility for other HTTP methods (including relative uncommon ones such as OPTIONS and PATCH), I felt it makes more sense to add it here.

PR #41668, opened 2022-03-25
[9.x] Adding additional search methods to the Arr class

michaelr0:
Hello,

This aims to add an alias to `array_search` to the Arr class and also explores introducing a binary search method.

The implementation of the binary search might not be optimal, however I'm open to discussions and adjustments.

A core implementation of binary searching on ordered arrays would be handy to have, as it can be much more performant than traditional (next, next, next) searching methods, where supported by the datasets.

Thanks.


taylorotwell:

PR #41680, opened 2022-03-25
[8.x] fix: VerifyCsrfToken - http_only config

peterfliszar:
VerifyCsrfToken - newCookie() is did not used the session config "http_only" value.


GrahamCampbell:
I don't think this is a bug. The http_only value is **only** for the session, and not csrf tokens?

PR #41689, opened 2022-03-25
[9.x] Add column helper to allow easier use of database functions

krve:
## Description

This pull request implements a new column helper to allow the use of certain database functions in the query builder. This means you can now easily use aggregate functions like `COUNT` or `MIN` when selecting, ordering or grouping any data.

Currently, if you want to use any of the existing database functions you are forced to use `DB::raw()` helper and write the query yourself.

This shouldn't have an impact on existing applications as this is an entirely new class, and it's split up into seperate functions in the Grammar class.

## Example

Say you want to fetch the daily views of all categories on your shop. Today you would have to do as follows:

```php
DB::table('category_views')
    ->select(
        DB::raw('DATE(created_at) as date'),
        DB::raw('SUM(views) as total_views'),
    )
    ->groupBy('date')
    ->get();
```

This can now be done using a helper class called `Column`

```php
DB::table('category_views')
    ->select(
        Column::date('created_at')->as('date'),
        Column::sum('views')->as('total_views'),
    )
    ->groupBy('date')
    ->get();
```

Translates to: `select DATE("created_at") as "date", SUM("views") as "total_views" from "category_views" group by "date"`

This can also easily be used with DB::raw() if you want to.

```php
DB::table('scores')
    ->select(
        Column::date('created_at')->as('date'),
        Column::avg(DB::raw('score - 1'))->as('avg_score'),
    )
    ->groupBy('date')
    ->get();
```

Translates to: `select DATE("created_at") as "date", AVG(score - 1) as "avg_score" from "scores" group by "date"`

There are currently two functions accepting multiple parameters: `CONCAT` and `COALESCE`.

```php
User::select(
    Column::concat(Column::name('first_name'), ' ', Column::name('last_name')),
)->get()
```

```php
User::select(
    Column::coalesce(Column::name('primary_contact'), Column::name('default_contact'), 'no contact'),
)->get()
```

## Advantages
- **No more writing raw queries:** instead of having to manually write every function using `DB::raw` you can now use the new column class for this.
- **Extendability:** This opens up for extending the column class to handle advanced nested queries by using a fluent builder instead of manually creating the query.
  - `SUM(total_price - discount) as discount` could potentially be allowed in the future using some sort of fluent syntax for the new `Column`. I have some ideas for this if the pull request is accepted.

## Caveats

I had to implement the `Column::name()` method to allow the grammar builder to determine if the parameter is a string, or an actual column in the database.
This causes the functions allow multiple parameters to look cluttered.  

If you have any idea of how else to handle this I'm all ears, as I don't think this is quite as clean as I'd like.

## Note

There are a lot more database functions than present in this pull request, but I wanted to integrate the most used ones and keep it simple to see if there is any interest in this. New ones should be easy to implement in the future.

Let me know if there is anything that stands out and needs a bit of explaining.

image

### DIFF
--- a/src/Illuminate/Database/Grammar.php
+++ b/src/Illuminate/Database/Grammar.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Database;
 
+use Illuminate\Database\Query\Column;
 use Illuminate\Database\Query\Expression;
 use Illuminate\Support\Traits\Macroable;
 use RuntimeException;
@@ -46,12 +47,16 @@ abstract class Grammar
     /**
      * Wrap a value in keyword identifiers.
      *
-     * @param  \Illuminate\Database\Query\Expression|string  $value
+     * @param  \Illuminate\Database\Query\Expression|\Illuminate\Database\Query\Column|string  $value
      * @param  bool  $prefixAlias
      * @return string
      */
     public function wrap($value, $prefixAlias = false)
     {
+        if ($value instanceof Column) {
+            return $this->wrapColumn($value, $prefixAlias);
+        }
+
         if ($this->isExpression($value)) {
             return $this->getValue($value);
         }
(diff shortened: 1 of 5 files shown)


morloderex:
I'd really like to see this expanded into https://github.com/laravel/framework/pull/40115. But this would be a great start. As for the implementation i have no comments tho. 👍

taylorotwell:

PR #41700, opened 2022-03-27
[10.x] Ask for confirmation when running optimze command on local environment

mhmdomer:
I noticed a lot of people fall into the trap of running the optimize command locally and then making changes to their configuration or routes files and not being able to see their changes take effect.

This PR will ask for confirmation when running the `optimize` artisan command on `local` environment with the message:
```
Further configuration and routes changes will not take effect unless you run the command again or clear the cache, are you sure you want to run this command?
```

Also, added a `--force` flag to the `optimize` artisan command to ignore the environment and run the command anyways.

### DIFF
--- a/src/Illuminate/Foundation/Console/OptimizeCommand.php
+++ b/src/Illuminate/Foundation/Console/OptimizeCommand.php
@@ -29,7 +29,7 @@ class OptimizeCommand extends Command
     public function handle()
     {
         $confirmationMessage = 'Further configuration and routes changes will not take effect unless you run the command again or clear the cache, are you sure you want to run this command?';
-        if (App::environment('local') && ! $this->confirm($confirmationMessage)) {
+        if (App::isLocal() && ! $this->confirm($confirmationMessage)) {
             return 1;
         }
 
--- a/src/Illuminate/Foundation/Console/OptimizeCommand.php
+++ b/src/Illuminate/Foundation/Console/OptimizeCommand.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Foundation\Console;
 
 use Illuminate\Console\Command;
+use Illuminate\Support\Facades\App;
 
 class OptimizeCommand extends Command
 {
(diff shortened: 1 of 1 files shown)


ankurk91:
```php
use Illuminate\Console\ConfirmableTrait;

if (! $this->confirmToProceed()) {
            return Command::SUCCESS;
        }

```

Laravel has inbuilt support for such confirmation

mhmdomer:
@ankurk91  This inbuilt confirmation is only triggered in production. In this case, I want it to be triggered when the command is run on local

ankurk91:
Oh, gotcha

mikeydevelops:
I am sorry but my opinion is that this is not needed, even with confirmation in a local environment you will just accept it as nothing happened and forget it. I don't see how this helps you remember you ran the command.
I am embarrassed to say this, but route caching and config caching have bitten me before and I don't think it would have helped if there was a confirmation. You just have to keep it in mind when you run it.

mhmdomer:
@mikeydevelops I agree with you in the point that it doesn't help you remember you ran the command, but in development, you shouldn't run this command, and if you do want to run it for some reason, this message will remind you of the consequences if you do so, and there are a lot of beginners who don't know what will happen if you run it locally, so it could be helpful info for them.

mikeydevelops:
@mhmdomer I'm sorry I thought I read the issue but it appears I am an idiot. Idiotic thing is also running commands you don't know what they do. (like I have done and probably will still do because I will forget I cached something and will bang my head in the cache rabbit hole).
Still, the command just needs a better description (that it caches routes and config only).
Or you could do the same thing to the other direct commands like `route:cache` and `config:cache` since all the things `optimize` does.

mhmdomer:
@mikeydevelops totally.
I am just waiting to see what others think about this change and if I should make the same changes to the `config:cache` and `route:cache` commands. Thanks for the inputs!

src/Illuminate/Foundation/Console/OptimizeCommand.php
```diff
@@ -13,6 +14,13 @@ class OptimizeCommand extends Command
      */
     protected $name = 'optimize';
```
driesvints, line 15: Remove this if you're adding a `$signature` property.

taylorotwell:

PR #41701, opened 2022-03-27
[9.x] Fix Postgres driver not dropping all tables & views

### LINKED ISSUE #41483: db:wipe doesn't remove migrations table from second schema

derekmd:
Fixes https://github.com/laravel/framework/issues/41483
Replaces https://github.com/laravel/framework/pull/41541

This fixes `php artisan db:wipe` not dropping all tables & views across multiple Postgres schemas. `DROP` queries didn't schema-qualify names so only the first object found is dropped. e.g., if a "migrations" table exists in two schemas, `Schema::dropAllTables()` leaves one of those tables.

Start schema-qualifying these names to ensure each object is explicitly removed.

```sql
-- assuming config('database.connections.pgsql.search_path') === 'public,laravel'

-- before
drop table "migrations","users" cascade

-- after
drop table "public"."migrations","public"."users","laravel"."migrations","laravel"."users" cascade
```

---

This implementation is almost functionality the same as the other draft PR except it:

* simplifies quote-wrapping table/view/type names
* `Schema::getAllTables()` returns an already-quoted `$row['qualifiedname']` column. e.g,
  ```php
  [
    [
      'tablename' => 'migrations',
      'qualifiedname' => '"laravel"."migrations"',
    ],
    [
      'tablename' => 'users',
      'qualifiedname' => '"laravel"."users"',
    ],
  ]
  ```
* adds integration tests to wipe tables & views defined in two separate schemas
* replaces `array_walk()` expressions from previous PRs with simpler `array_map()`.
* clarifies up some mock `andReturn()` values from previous PRs

### DIFF
--- a/src/Illuminate/Database/Concerns/ParsesSearchPath.php
+++ b/src/Illuminate/Database/Concerns/ParsesSearchPath.php
@@ -18,12 +18,8 @@ trait ParsesSearchPath
             $searchPath = $matches[0];
         }
 
-        $searchPath ??= [];
-
-        array_walk($searchPath, static function (&$schema) {
-            $schema = trim($schema, '\'"');
-        });
-
-        return $searchPath;
+        return array_map(function ($schema) {
+            return trim($schema, '\'"');
+        }, $searchPath ?? []);
     }
 }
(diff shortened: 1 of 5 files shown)


driesvints (approved):

driesvints:
Thanks @derekmd

PR #41703, opened 2022-03-27
[9.x] Add new test for jobs which dispatched sync in BusDispatcher

al1r374:
This PR adds a new test for dispatch a queueable job with `dispatchSync` command that is not covered yet.


taylorotwell:
I don't like adding over-mocked tests.

PR #41705, opened 2022-03-28
[9.x] Fix  expected parameter is inconsistent with given parameter in MailManager

kayw-geek:


driesvints:
The transport created by `EsmtpTransportFactory` can only be an `EsmtpTransport` so the current types are correct.

Issue #41665, opened 2022-03-24
lazy() skips and duplicates rows


tp971:
- Laravel Version: 9.5.1
- PHP Version: 8.1.3
- Database Driver & Version:
    - Server: MariaDB 10.3.34-MariaDB-0ubuntu0.20.04.1 - Ubuntu 20.04
    - Database client version: libmysql - mysqlnd 8.0.17 

### Description:

When fetching models with `...->orderBy(...)->lazy()`, the resulting collection potentially skips rows or returns the same row multiple times.

### Steps To Reproduce:

Iterate over `SomeModel::orderBy('some_non_unique_column')->lazy()` with a big table (e.g. 10000-100000 rows). I provided an example project here: https://github.com/tp971/laravel-bug-lazy-collection

Since this bug is probably a result of ORDER BY on non-unique columns not being deterministic, this bug may or may not occur depending on the used DBMS, the DBMS configuration and the number of inserted rows. The example project above should insert enough rows to cause the bug pretty consistently in an Ubuntu 20.04 VM with MariaDB.

crissi:
It is because it uses page/limit sql, that can cause skipped rows.

Would probably be nice for lazy to implement cursor pagination as the internals instead of page/limit logic this will prevent the skipped rows and make it faster

bernardwiesner:
I suppose this issue also happens with chunk() as it also uses offset and limit.

tp971:
I don't know, but this doesn't seem very "undocumented": https://laravel.com/docs/9.x/queries#streaming-results-lazily

driesvints:
@tp971 sorry, I mixed this up with the lazy method of collections.

To answer your question, I don't think you can use `->lazy()` like this: https://github.com/tp971/laravel-bug-lazy-collection/blob/master/app/Console/Commands/BugTrigger.php#L34. Can you try to use the `->each()` with it as documented?

tp971:
@driesvints using `->each()` results in the same issue. I just tested it and updated the repo

driesvints:
@tp971 sorry I didn't see this before but it's indeed probably because you're ordering by a non-unique column. Cursor pagination also has this limitation:

<img width="699" alt="Screenshot 2022-03-28 at 13 18 16" src="https://user-images.githubusercontent.com/594614/160387091-29e70c27-8bf8-4507-b969-779f90a6abf3.png">

PR #41710, opened 2022-03-28
[9.x] Mark lazy method as internal

driesvints:
People keep trying to use this method while it's only meant to be used internally.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -68,6 +68,8 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
      * Get a lazy collection for the items in this collection.
      *
      * @return \Illuminate\Support\LazyCollection<TKey, TValue>
+     *
+     * @internal This method is only meant to be used internally and not in userland.
      */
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Isn't this a BC break?

driesvints:
@GrahamCampbell no

ionutantohi:
@driesvints I think is a misunderstanding here, the issue #41665 refers to `lazy` method from BuildsQueries not Collection.

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Concerns/BuildsQueries.php#L187

tp971:
@ionutantohi @driesvints yes, I used the lazy method on the query builder, not on any collection, just like here: https://laravel.com/docs/9.x/queries#streaming-results-lazily

driesvints:
@ionutantohi @tp971 thanks. I answered on the issue again. I'm gonna leave this PR open as it's still valid.

mikeydevelops:
Why is it meant to be used only internally? I think it is nice to have it available.

driesvints:
@mikeydevelops there are no use cases afaik.

PR #41713, opened 2022-03-28
[10.x] Added `assertClosurePushed` and `assertClosureNotPushed`

ash-jc-allen:
Hey! This PR proposes 2 new methods that can be used in tests: `assertClosurePushed` and `assertClosureNotPushed`.

Say we have either of these snippets of code in our apps:

```php
dispatch(function () {
    // Do something here...
});

// Or...
Queue::push(function () {
    // Do something here...
});
```

If we wanted to assert in our test that a closure was queued (or not queued), and we used the `dispatch` helper we'd need to write something like this:

```php
use Illuminate\Queue\CallQueuedClosure;

Queue::fake();

Queue::assertPushed(CallQueuedClosure::class);

Queue::assertNotPushed(CallQueuedClosure::class);
```

If we wanted to assert in our test that a closure was queued(or not queued), and we used the `Queue::push` helper we'd need to write something like this:

```php
Queue::fake();

Queue::assertPushed(\Closure::class);

Queue::assertNotPushed(\Closure::class);
```

So, I've added a small change to the `QueueFake` class that creates a `CallQueuedClosure` class for consistency, so that both `Queue::push` and `dispatch` act the same way. I guess that this would be a breaking change because if anyone is using `Queue::assertPushed(\Closure::class)` with the `Queue::push()` method, the assertions would now fail.

I've also added two new methods I think will make the code a little bit easier to read like so:

```php
use Illuminate\Queue\CallQueuedClosure;

Queue::fake();

Queue::assertClosurePushed();

Queue::assertClosureNotPushed();
```

The `assertClosurePushed` method will also support callbacks and checking the times that a closure was pushed too.

These methods aren't really huge additions and might not be needed, but I just thought it was worth putting the idea out there. My main reason is that I don't think it's immediately obvious how we can test that a closure is pushed onto the queue and can be a bit confusing for newer devs because they'll need to use different assertions based on if they'd used `dispatch` or `Queue::fake()` (maybe an addition to the docs could be more suitable instead?). So, hopefully, this change would make it a bit easier for devs to use in tests.

Apologies if there's already an easier way of doing this that I hadn't noticed.

If it's something you think might be useful, please let me know if there's anything that I might need to change 😄

### DIFF
--- a/src/Illuminate/Support/Facades/Queue.php
+++ b/src/Illuminate/Support/Facades/Queue.php
@@ -17,8 +17,10 @@ use Illuminate\Support\Testing\Fakes\QueueFake;
  * @method static mixed pushRaw(string $payload, string $queue = null, array $options = [])
  * @method static string getConnectionName()
  * @method static void assertNotPushed(string|\Closure $job, callable $callback = null)
+ * @method static void assertClosureNotPushed(callable $callback = null)
  * @method static void assertNothingPushed()
  * @method static void assertPushed(string|\Closure $job, callable|int $callback = null)
+ * @method static void assertClosurePushed(callable|int $callback = null)
  * @method static void assertPushedOn(string $queue, string|\Closure $job, callable $callback = null)
  * @method static void assertPushedWithChain(string $job, array $expectedChain = [], callable $callback = null)
  *
--- a/src/Illuminate/Support/Testing/Fakes/QueueFake.php
+++ b/src/Illuminate/Support/Testing/Fakes/QueueFake.php
@@ -5,6 +5,7 @@ namespace Illuminate\Support\Testing\Fakes;
 use BadMethodCallException;
 use Closure;
 use Illuminate\Contracts\Queue\Queue;
+use Illuminate\Queue\CallQueuedClosure;
 use Illuminate\Queue\QueueManager;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Traits\ReflectsClosures;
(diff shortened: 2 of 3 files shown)


taylorotwell:
Does this allow you to test queueing Closures using `dispatch`?

```php
dispatch(fn () => true);
```

taylorotwell:
Does this allow you to test queueing closures via the `dispatch` helper?

```php
dispatch(fn () => true);
```

ash-jc-allen:
As far as I'm aware, I think it should with closures that have been queued with either `dispatch` or `Queue::push()`. I've just thrown together two quick tests in a fresh(ish) Laravel app using my changes and they appear to be working :)

Screenshot 2022-03-28 at 21 30 43
Screenshot 2022-03-28 at 21 31 08

driesvints:
@ash-jc-allen why did you send this in to master and not 9.x?

ash-jc-allen:
@driesvints I do have another branch (https://github.com/laravel/framework/compare/9.x...ash-jc-allen:feature/assert-queued-closure) that I was going to use originally to target 9.x. But, I think there's a breaking change though (I might be wrong though).

If anyone is currently using `Queue::push()` to queue a closure, they'll need to use `Queue::assertPushed(\Closure::class)` in there test.

But, with my new change, that assertion would now fail because the closure is getting changed into a `CallQueuedClosure` instead. So, they'd need to update their assertion to: `Queue::assertPushed(CallQueuedClosure::class)`.

If there's a way that you think I could change this though so that I could target it to 9.x, I'd be more than happy to make the changes :)

driesvints:
Right makes sense. Keep it here 👍

Issue #41714, opened 2022-03-28
Allow localization of Validation Error summary


dianfishekqi:
Hello there,

It would be great if we added localization support to the summeriza part of laravel validation. 

![Screen Shot 2022-03-28 at 14 42 39](https://user-images.githubusercontent.com/4225509/160400226-84fafe00-72cc-4a5b-8f6d-ba33ef836ee4.png)

https://github.com/laravel/framework/blob/f6915333c21b9c656e98f211cf4510115d19c6bd/src/Illuminate/Validation/ValidationException.php#L99

driesvints:
Heya, thanks for submitting this.

This seems like a feature request or an improvement so I'm converting this to a discussion instead. It's best to post these in the discussions section in the future to get support for your idea. After that you may send a PR to the framework. Please only use the `laravel/framework` issue tracker to report bugs and issues with the framework.

Thanks!

PR #41717, opened 2022-03-28
[9.0] Localization of validation error summary

dianfishekqi:
Localizes validation error summaries with fallback to english.

Related #41714 , #41715


driesvints:
Internal exception messages shouldn't be translated or shown to actual users. Please see many previous attempts.

GrahamCampbell:
Indeed, it's very important for security that you don't forward exception messages, in general, to end users.

Pablo-Camara:
Please check my question out:
https://stackoverflow.com/questions/73898408/how-to-translate-laravel-hardcoded-string-and-1-more-error-using-dynamic

can you help ?

Issue #41704, opened 2022-03-27
Can't use camelCase naming with new Eloquent attribute syntax


timhoeppner:
- Laravel Version: 9.5.1
- PHP Version: 8.1.3
- Database Driver & Version: MySQL 8.0.28-1debian10

### Description:
I cannot use camelCase attribute names with the new Eloquent attribute syntax.

### Steps To Reproduce:

The following code results in `Call to undefined method App\Models\MyModel::getTestFooAttribute()`. Modifying the $appends to use 'test_foo' correctly calls the custom attribute.

Camel case attributes are supported using the get...Attribute() and set...Attribute() syntax which is why I believe this to be an error.

```php
class MyModel extends Model
{
  protected $appends = ['testFoo'];

  protected function testFoo(): Attribute
  {
    return Attribute::make(
      get: fn () => 'test'
    );
  }
}
```

driesvints:
`$appends` needs to be `test_foo`

timhoeppner:
I respect that's how it works right now I just thought this may be a bug since you can use camel case attributes with the older get...Attribute() and set...Attribute() syntax.

PR #41725, opened 2022-03-28
[9.x] Adding the ability to use camelCase or snake_case attribute marked attributes

timhoeppner:
See #41704 for reference.

The old style of get...Attribute() and set...Attribute supported using either snake case or camel case attribute names. The new Attribute marked attributes only support snake case. This PR adds the ability to use camel case attribute marked attributes.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -660,8 +660,10 @@ trait HasAttributes
     {
         if ($this->isClassCastable($key)) {
             $value = $this->getClassCastableAttributeValue($key, $value);
-        } elseif (isset(static::$getAttributeMutatorCache[get_class($this)][$key]) &&
-                  static::$getAttributeMutatorCache[get_class($this)][$key] === true) {
+        } elseif ((isset(static::$getAttributeMutatorCache[get_class($this)][$key]) &&
+                static::$getAttributeMutatorCache[get_class($this)][$key] === true) ||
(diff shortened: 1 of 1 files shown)


taylorotwell:
No plans to add support for this for now.

timhoeppner:
@taylorotwell Is there more I could do to help get this added? I would be happy to add tests or whatever to make this as simple as possible on your end.

driesvints:
@timhoeppner I think you just have to wrap your head around the fact that attributes are snake_cased (like database columns) while mutators/accessors are camelCased. This has always been how it worked since the beginning of Eloquent. We won't be changing this, sorry.

timhoeppner:
@driesvints You know that you can use camelcase attributes with the get...Attribute() syntax though right? Anyway, all the best.

driesvints:
@timhoeppner yeah I realize that and realize that's a bit confusing. But attributes were in the first place designed to be snake cased.

timhoeppner:
@driesvints that's totally fair.  The thing I find a bit awkward is with snake case attributes when I'm consuming these on a front end and I'm having to put snake case in my typescript interfaces whereas everywhere else in my code base in camel case.

driesvints:
@timhoeppner definitely. Something that can help there are API Resources. So you have control over how the keys look like consuming on the front-end: https://laravel.com/docs/9.x/eloquent-resources

guoyunhe:
camelCase is the most popular practice of front-end programming with JS. Especially when you are using React. Eloquent resources are helpful for receiving response, but you still have to do some manual conversion when sending requests...

[Adonis supports this via naming strategy](https://docs.adonisjs.com/reference/orm/naming-strategy).

PR #41726, opened 2022-03-28
[10.x] Adds a new `doesntHave` method to the session store

mateusjunges:
## What?
This PR adds a `doesntHave` method to the session store, which allows to check if a given key is missing from the session of if it's value is `null`.

## Why? 
Before this PR, you have to use `! session()->has('key')`. If this PR gets merged, you can use `session()->doesntHave('key')`, which IMO is much more readable.

```php
//Before
if (! session()->has('key')) {
    // Do something...
}

// After
if (session()->doesntHave('key')) {
    // Do something...
}
```

PS..: I'm sending this to master because i think adding a new method to a non internal interface is a major breaking change accordingly to semver.

### DIFF
--- a/src/Illuminate/Contracts/Session/Session.php
+++ b/src/Illuminate/Contracts/Session/Session.php
@@ -72,6 +72,14 @@ interface Session
     public function has($key);
 
     /**
+     * Checks if a key is missing or is null.
+     *
+     * @param  string|array  $key
+     * @return bool
+     */
+    public function doesntHave($key);
+
+    /**
      * Get an item from the session.
      *
      * @param  string  $key
--- a/src/Illuminate/Session/Store.php
+++ b/src/Illuminate/Session/Store.php
@@ -231,6 +231,17 @@ class Store implements Session
     }
 
     /**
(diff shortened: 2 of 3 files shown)


src/Illuminate/Contracts/Session/Session.php
```diff
@@ -71,6 +71,14 @@ public function exists($key);
      */
     public function has($key);
 
+    /**
+     * Checks if a key is missing or is null.
+     *
+     * @param  string|array  $key
+     * @return bool
+     */
+    public function doesntHave($key);
```
driesvints, line 80: I don't think we should add this method to the contract. Seems unnecessary since the opposite one already exists.

a-bashtannik: Agree. `doesntHave` on Eloquent creates a custom query, here it's just a negation of truth.

mateusjunges: @driesvints thanks, so i'll send this to 9.x instead.

mateusjunges:
@driesvints I'm closing this, please check #41738.

PR #38927, opened 2021-09-23
[8.x] Add `whereBelongsTo()` Eloquent builder method

danharrin:
Hey 👋 

When retrieving `belongsTo` related records, you can usually use the inverse (`hasMany`) of the relationship, like so:

```php
$author->posts()
```

However, sometimes this is not possible, and you must filter an existing query by its parent record:

```php
$query->where('author_id', $author->id)
```

This works, but explicitly depends on the name of foreign key (`author_id`), and the name of the owner key (`id`) in most cases. This is a maintenance burden, as these key names can be changed inside the relationship method and will subsequently become outdated across your entire app.

This PR introduces a `whereBelongsTo()` method to the Eloquent builder. You may use it like so:

```php
$query->whereBelongsTo($author)
```

It will automatically retrieve the foreign key name from the relationship (`author`), and the correct owner key from the related model (`$author`).

By default, the name of the relationship is guessed based on the class of the related model. Sometimes, this is not appropriate. If the `$author` is an instance of `App\Models\User`, `whereBelongsTo()` will search for a `user()` relationship that does not exist. In this case, you may manually specify the relationship name:

```php
$query->whereBelongsTo($author, 'author')
```

PR #41733, opened 2022-03-29
[9.x] Make `whereBelongsTo` accept Collection

erikgaal:
Hi there!

Extending upon #38927, I'm submitting this PR that allows you to retrieve records using a collection of related records. 

Previously, if you wanted to retrieve, for example, all Posts that belong to one of multiple Categories, you would need to perform the following query:

```php
$query->whereBelongsTo($category[0])
    ->orWhereBelongsTo($category[1])
    // etc...

// Or:
$query->whereIn('category_id', $categories->modelKeys());
```

Same as in the aforementioned PR, you would need to know the key names when building and maintain these if they change. This PR adds support for Collections passed to the `whereBelongsTo` method:

```php
$query->whereBelongsTo($categories);
```

```php
$query->whereBelongsTo($categories, 'category');
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -509,6 +509,59 @@ trait QueriesRelationships
     }
 
     /**
+     * Add a "belongs to one of" relationship where clause to the query.
+     *
+     * @param  \Illuminate\Database\Eloquent\Collection<\Illuminate\Database\Eloquent\Model>  $values
+     * @param  string|null  $relationshipName
+     * @param  string  $boolean
+     * @return $this
+     *
+     * @throws \Illuminate\Database\Eloquent\RelationNotFoundException
+     */
+    public function whereBelongsToOneOf($values, $relationshipName = null, $boolean = 'and')
+    {
+        if (! $related = $values->first()) {
+            return $this->whereNull($this->getModel()->getKeyName(), $boolean);
+        }
+
+        if ($relationshipName === null) {
+            $relationshipName = Str::camel(class_basename($related));
+        }
+
+        try {
+            $relationship = $this->model->{$relationshipName}();
+        } catch (BadMethodCallException $exception) {
+            throw RelationNotFoundException::make($this->model, $relationshipName);
+        }
+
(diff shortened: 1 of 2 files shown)


a-bashtannik:
Could somebody explain or point to the docs, why [whereIn](https://github.com/laravel/framework/blob/b9a34d01bf19347aef3f04b11709f1ca8c9deddf/src/Illuminate/Database/Query/Builder.php#L1001) accepts a flag `boolean` supposing 2 values: `and` and `or` while the SQL IN operator is a shorthand for multiple `or` conditions? Thanks!

erikgaal:
@a-bashtannik It allows you to make the distinction between adding `and "column" in  ?` and `OR "column" IN ?` to the SQL query. It's the same for all `where` methods in the Builder.

a-bashtannik:
@erikgaal argghh so simple true, thanks!

taylorotwell:
@erikgaal did you give any thoughts to just baking this functionality into `whereBelongsTo`?

erikgaal:
> @erikgaal did you give any thoughts to just baking this functionality into `whereBelongsTo`? 

No, I haven't. Seems logical to support both a single `Model` and a `Collection` as argument to the function. If that works for you too, I'm happy to implement it that way!

taylorotwell:
@erikgaal yeah I think it's fine to bake this behavior into the existing method.

erikgaal:
@taylorotwell Done!

src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
```diff
@@ -470,6 +471,17 @@ public function orWhereMorphedTo($relation, $model)
      */
     public function whereBelongsTo($related, $relationshipName = null, $boolean = 'and')
     {
+        if (! $related instanceof Collection) {
+            $values = $related->newCollection([$related]);
+        } else {
+            $values = $related;
+            $related = $values->first();
```
Jubeki, line 478: What happens if you pass in an empty collection? I would expect to get no models back after the query is run.

Bosphoramus:
Hi @erikgaal would it be possible for empty collections to be passed?

Id expect it to work the same as if i pass [] to whereIn(), no records would be returned, this would be handy to avoid some conditional checks.

Issue #41533, opened 2022-03-17
 Mutator called 9 times


mo20130123:
- Laravel Version: 9.2
- PHP Version: 8.0.16 

### Description:
 
```
class Category extends Model
{
    public function banner(): Attribute
    {
        return Attribute::make (
            get: fn ($value) => asset('images/category/Banner/'.$value),
            set: function ($value) {
                if(is_file($value))
                {
                    Log::info('count');
                    $fileName = 'Banner_'.rand(11111,99999).'.'.$value->getClientOriginalExtension();
                    $value->storeAs('images/category/Banner', $fileName );
                    return $fileName;
                }
            }
        );
    }
}
```

when creating a new Category - 
the Log count called 9 times 
which cause the image to be uploaded 9 times

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

However, this issue will not be locked and everyone is still free to discuss solutions to your problem!

Thanks.

mo20130123:
Hello Dries 
please let me if I am wrong - 
when I use the old way SetBannerAttribute() It worked fine - but with the new way - it has been called 9 times  - 
so I assumed that this is a problem in the library

zupolgec:
@driesvints I think it's an issue. It seems that to allow things such as `$model->address->lineOne = '123 High St.'` the `set` function of the Attribute is called on save (which is correct) and on every subsequent call of *any* accessor.

i.e. if the model has a `secondaryAddress` Attribute doing this

```
$model->address->lineOne = '123 High St.';
echo $model->secondaryAddress->lineOne;
```

triggers the set function on the `address` Attribute.

driesvints:
You should not perform file operations in an attribute. The attribute is purely to transform the attribute's value and shouldn't perform any side effects. It shouldn't matter how many times it's invoked.

zupolgec:
[...]

PR #41736, opened 2022-03-29
[9.x] Fix multiple calls to class cast mutators

zupolgec:
Solves #41533. Class casts (`Attribute` included) allow the developer to use a more expressive syntax like `$model->address->lineOne = '117 Spencer St.'` avoiding assigning the object to a variable and then updating the attribute to trigger the mutator.

What the framework currently does to allow this to work is "if a mutated attribute has already been accessed once, every time I access any attribute (through `getAttributes`), call the mutator to sync the changes made to the object back to the model attributes".

My tentative solution to this issue is: every time an attribute or class mutated attribute is accessed, consider that it could potentially also lead to an object modification. So flag this attribute as "potentially altered" and call the mutator on first occasion, the remove the flag until next access.

**EDIT**

This performance check to avoid multiple calls doesn't allow anymore to modify the object by itself but you need to access the object through the model.

So given
```
$address = new AttributeCastAddress('110 Kingsbrook St.', 'My Childhood House');
$model->address = $address
```
you should do
```
$model->address->lineOne = '117 Spencer St.';
```
not
```
$address->lineOne = '117 Spencer St.';
```

I guess that this introduces a breaking change and then this PR should target the next major version.

### DIFF
--- a/tests/Integration/Database/DatabaseEloquentModelAttributeCastingTest.php
+++ b/tests/Integration/Database/DatabaseEloquentModelAttributeCastingTest.php
@@ -331,6 +331,17 @@ class DatabaseEloquentModelAttributeCastingTest extends DatabaseTestCase
             $this->assertNotSame($previous, $previous = $model->virtualDateTimeWithoutCachingFluent);
         }
     }
+
+    public function testMutatorsAreCalledOnlyWhenNeeded()
+    {
+        $model = new TestEloquentModelWithAttributeCast;
+
+        $model->address = new AttributeCastAddress('110 Kingsbrook St.', 'My Childhood House');
+        $model->address->lineOne = '117 Spencer St.';
+        $this->assertSame('117 Spencer St.', $model->getAttributes()['address_line_one']);
+
+        $this->assertEquals(2, $model->countAddressMutatorCalls);        
+    }
 }
 
 class TestEloquentModelWithAttributeCast extends Model
@@ -341,6 +352,8 @@ class TestEloquentModelWithAttributeCast extends Model
      * @var string[]
      */
     protected $guarded = [];
+
+    public $countAddressMutatorCalls = 0;
 
     public function uppercase(): Attribute
     {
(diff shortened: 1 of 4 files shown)


linaspasv:
Unfortunately this does not solve the main problem. The main cause of too many getAttributes calls is because of irrational and expensive multiple uses of getDirty method during the save process.

taylorotwell:

tests/Integration/Database/DatabaseEloquentModelAttributeCastingTest.php
```diff
@@ -49,8 +49,8 @@ public function testBasicCustomCasting()
 
         $model = new TestEloquentModelWithAttributeCast;
 
-        $model->address = $address = new AttributeCastAddress('110 Kingsbrook St.', 'My Childhood House');
-        $address->lineOne = '117 Spencer St.';
+        $model->address = new AttributeCastAddress('110 Kingsbrook St.', 'My Childhood House');
+        $model->address->lineOne = '117 Spencer St.';
```
taylorotwell, line 53: Why did you change this test?

zupolgec: Apologies, I was sure I had added an edit to the PR description highlighting that by introducing this performance check to avoid multiple calls you then can't modify the object by itself but you need to access the object through the model.

So given
```
$address = new AttributeCastAddress('110 Kingsbrook St.', 'My Childhood House');
$model->address = $address
```
you should do
```
$model->address->lineOne = '117 Spencer St.';
```
not
```
$address->lineOne = '117 Spencer St.';
```

I guess that this introduces a breaking change and then this PR should target the next major version.

zupolgec:
Yep, you're right (as I clarified in the PR description now). It's a use case I didn't think of at first, since I normally access attributes through the model. Probably just flagging every attribute with `withoutObjectCaching()` is enough if you want to save a mutator call every time an attribute is accessed.

zupolgec:
@taylorotwell could perhaps this still be valid for AttributeMutated attributes? No need to update those every time, since they can only be accessed through the model.

taylorotwell:
No - I think it causes the same problem with those as well. I don't think there is a guaranteed way to solve this other than the current implementation to be honest.

PR #41738, opened 2022-03-29
[9.x] Adds a new `doesntHave` method to the session store

mateusjunges:
## What?
This PR is a follow up on #41726, and adds a `doesntHave` method to the session store, which allows to check if a given key is missing from the session of if it's value is `null`.

## Why? 
Before this PR, you have to use `! session()->has('key')`. If this PR gets merged, you can use `session()->doesntHave('key')`, which IMO is much more readable.

```php
//Before
if (! session()->has('key')) {
    // Do something...
}

// After
if (session()->doesntHave('key')) {
    // Do something...
}
```

### DIFF
--- a/src/Illuminate/Session/Store.php
+++ b/src/Illuminate/Session/Store.php
@@ -231,6 +231,17 @@ class Store implements Session
     }
 
     /**
+     * Checks if a key is missing or is null.
+     *
+     * @param  string|array  $key
+     * @return bool
+     */
+    public function doesntHave($key)
+    {
+        return ! $this->has($key);
+    }
+
+    /**
      * Get an item from the session.
      *
      * @param  string  $key
--- a/src/Illuminate/Session/Store.php
+++ b/src/Illuminate/Session/Store.php
@@ -218,7 +218,7 @@ class Store implements Session
     }
 
     /**
-     * Checks if a key is present and not null.
+     * Determine if a key is present and not null.
      *
      * @param  string|array  $key
      * @return bool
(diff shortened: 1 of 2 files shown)


taylorotwell:
I suggest using `if (session()->missing('key'))`

mateusjunges:
@taylorotwell @driesvints the `missing` method returns false when the key exists but it's value is `null`. The proposed `doesntHave` will return true in this case. This method does not have the same results. You can check it by looking at the `missing` method test:

```php
public function testKeyMissing()
{
    $session = $this->getSession();
    $session->put('foo', 'bar');
    $this->assertFalse($session->missing('foo'));
    $session->put('baz', null);
    $session->put('hulk', ['one' => true]);
    $this->assertFalse($session->has('baz'));
    $this->assertFalse($session->missing('baz')); // Returns false even if the key is set to null
    $this->assertTrue($session->missing('bogus'));
    $this->assertFalse($session->missing(['foo', 'baz']));
    $this->assertTrue($session->missing(['foo', 'baz', 'bogus']));
    $this->assertFalse($session->missing(['hulk.one']));
    $this->assertTrue($session->missing(['hulk.two']));
}
```

And the tests for `doesntHave`:

```php
public function testDoesntHaveKey()
{
    $session = $this->getSession();
    $session->put('foo', 'bar');
    $this->assertTrue($session->doesntHave('baz'));
    $session->put('baz', null);
    $this->assertTrue($session->doesntHave('baz')); // Returns true when the key exists but is set to null
}
```

PR #41740, opened 2022-03-29
[9.x] Added boolean to console argument function's return type docblock

jamesabell:
## What
`Console/Concerns/InteractsWithIO.php@argument()` should include bool (`@return string|array|null|bool`) in its return type docblock.

## Why
By calling `Artisan::call([command], [ 'fooBool' => true ])` the value returned from `$this->argument('fooBool')` in laravel commands would be a boolean.

## Reasoning
This will prevent IDEs, phpStan etc. from flagging boolean comparisons.  ie: `$this->argument('fooBool') === false`

The workaround to this is to just pass `0` or `1`, but I much prefer to use booleans to be more explicit and don't see any reason why we wouldn't include it as an option.

### DIFF
--- a/src/Illuminate/Console/Concerns/InteractsWithIO.php
+++ b/src/Illuminate/Console/Concerns/InteractsWithIO.php
@@ -64,7 +64,7 @@ trait InteractsWithIO
      * Get the value of a command argument.
      *
      * @param  string|null  $key
-     * @return string|array|null|bool
+     * @return array|string|bool|null
      */
     public function argument($key = null)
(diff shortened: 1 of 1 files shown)


jamesabell:
Through the artisan CLI I don't think there's a way to pass a boolean value. 

But for validity, yes I think so. It is valid as mentioned in the docs here: https://laravel.com/docs/9.x/artisan#programmatically-executing-commands.

May not be a very typical use case to pass a boolean value, but I ran into it.

PR #41742, opened 2022-03-29
Assert that a response matches the specified errors

pacoorozco:
When you're asserting validation errors you can use `assertSessionHasErrors()` and `assertSessionDoesNotHaveErrors()`. You need to use both to ensure that just only a set of validations have failed.

But I've created a new assertion that could be interesting to be shared: `assertSessionMatchesErrors()`. 

The assertion will pass **if the provided errors and only these ones are present in the session**. 

Let's imagine a form request with 4 fields that are validated: `field1`, `field2`, `field3` and `field4`.

If you have a testcase to ensure that `field1` validation fails you need to do:

```
// To ensure that only `field1` is failing.

$this->assertSessionHasErrors(['field1']);
$this->assertSessionDoesNotHaveErrors(['field2', 'field3', 'field4']);

// Instead you can do

$this->assertSessionMatchesErrors(['field1']);
```
I've added a discussion in the [GitHub discussion board](https://github.com/laravel/framework/discussions/41741).

**WDYT?**

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -1346,6 +1346,29 @@ EOF;
     }
 
     /**
+     * Assert that the session matches the given errors.
+     *
+     * @param  string|array  $keys
+     * @param  mixed  $format
+     * @param  string  $errorBag
+     * @return $this
+     */
+    public function assertSessionMatchesErrors($keys = [], $errorBag = 'default')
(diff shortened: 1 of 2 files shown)


taylorotwell:
Personally I'm not very interested in assertions that are specifically for session based errors or JSON errors. I would prefer things like `assertInvalid` and `assertValid` that work for either.

PR #41743, opened 2022-03-30
[10.x] Added command to generate globally unique app-id

allowing:
I found that writing prefixes depends on `app_name`, which may cause data confusion, such as cache prefixes. If the user does not change `app_name`, then multiple programs may use the same prefix.

So it is recommended to use `app_id` where `unique` is required.

### DIFF
--- a/src/Illuminate/Foundation/Console/AppIdGenerateCommand.php
+++ b/src/Illuminate/Foundation/Console/AppIdGenerateCommand.php
@@ -1,0 +1,118 @@
+<?php
+
+namespace Illuminate\Foundation\Console;
+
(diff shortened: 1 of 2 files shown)


allowing:
image

allowing:
image

dennisprudlo:
All this does is adding an `app-id:generate` command that will put a string into the .env, right? Why not use this command in your own project? I mean it doesn't change anything about how laravel works, developers would still need to use that new `APP_ID` environment variable or the `app.app_id` config value.

As for the cache naming you can overwrite the prefix in your .env:
```
CACHE_PREFIX=my_unique_cache_prefix
```

PR #41746, opened 2022-03-30
[9.x] Custom Cast serialize castable attributes

rrpadilla:
[Two JSON objects are equal if they have the same set of keys, and each key has the same value in both objects](https://dev.mysql.com/doc/refman/8.0/en/json.html#json-comparison).
```php
{"a": 1, "b": 2} = {"b": 2, "a": 1}
```

Now if you want to casts your JSON attribute to a [Custom Cast](https://laravel.com/docs/9.x/eloquent-mutators#attribute-casting) and the json keys do not match the **serialized** representation of the value, Laravel determines the given attribute(s) have been modified ("is dirty"). 

We don't have this issue with the "AsCollection" and "AsArrayObject" casts because Laravel will decode first to compare the values.

Thanks.


taylorotwell:
I'm not sure I feel comfortable making a behavior change like this on a patch release that should not include breaking changes.

rrpadilla:
Hi @taylorotwell, are you ok if I submit a PR with a test coverage only to show that the order of json keys does not affect "AsCollection" and "AsArrayObject" casts but could affect custom casts? Thanks.

PR #41747, opened 2022-03-30
[9.x] Directly return the condition result.

lucasmichot:
In many cases, conditions can be simplified: there is no need returning true or false, when we just have to return the condition result.

eg.

```php
// before
if (! Arr::has($this->data, $parameters[0])) {
    return true;
 }
 return false;

// after
return ! Arr::has($this->data, $parameters[0]);
```


taylorotwell:
Honestly I find some of these harder to read.

PR #41748, opened 2022-03-30
[9.x] Added a superior & inferior clauses to the query builder

Julian-Louis:
Hello,
In the course of my work, I had to implement multiples times numeric comparison between values in where clauses.
I thought it might be useful to others and I would love to contribute to the Laravel community.
Have a great day!

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -346,6 +346,30 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Add an "superior to" clause to the query.
+     *
+     * @param  \Closure|array|string|\Illuminate\Database\Query\Expression  $column
+     * @param  mixed  $value
+     * @return $this
+     */
+    public function whereSuperiorTo($column, $value = null)
+    {
+        return $this->where($column, '>', $value);
+    }
+
+    /**
+     * Add an "inferior to" clause to the query.
+     *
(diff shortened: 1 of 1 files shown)


dennisprudlo:
Maybe a macro would be better for this case?

If this gets merged (I would suggest to rename the functions to `whereGreaterThan` and `whereLessThan`) there will be PRs for every other case like what about those:
- `orWhereGreaterThan` (`or column > value`)
- `orWhereLessThan` (`or column < value`)
- `whereGreaterThanOrEquals` (`and column >= value`)
- `whereLessThanOrEquals` (`and column <= value`)
- `orWhereGreaterThanOrEquals` (`or column >= value`)
- `orWhereLessThanOrEquals` (`or column <= value`)

Julian-Louis:
Thank you for your macro tip.
I agree with whereGreaterThan but regarding whereLessThan, how about whereLowerThan instead ?
Let's see what the community feels about it !

dennisprudlo:
> how about whereLowerThan instead?

I would still use "less" in that case, since it is the name for the `<` comparison in mathematics and is widely adopted in programming (See the MySQL docs for reference: https://dev.mysql.com/doc/refman/8.0/en/comparison-operators.html)

Julian-Louis:
Okay, sorry, you're right

taylorotwell:

taylorotwell:

PR #41749, opened 2022-03-30
[9.x] Set correct PHP file permissions

peter279k:
# Changed log

- Set the correct PHP file permissions because they're not executable files.


driesvints:
No plans to merge this.

PR #41753, opened 2022-03-30
Override variadic parameter

altdovydas:
Issue creating object with variadic parameters
Example:
```
class ClassWithVariadicParameters
{
    public function __construct(TestInterface ...$objects)
}

$container = new Container;
$container->make(
    ClassWithVariadicParameters::class,
    [
        'objects' => [
            $container->make(TestInterface::class), 
            $container->make(TestInterface::class)
        ]
    ]
);

```
Error: 
`Argument 1 passed to ClassWithVariadicParameters::__construct() must implement interface TestInterface, array given`

@driesvints


driesvints:
There's no explanation here.

PR #41756, opened 2022-03-30
[9.x] Add `whereJsonContainsKey` to query builder

amir9480:
This PR adds a new function to the query builder to check JSON column type has the given key or not.
The syntax for keys is exactly like `whereJsonContains`.

Sample codes:
```php
$users = DB::table('users')
                ->whereJsonContainsKey('options->languages')
                ->get();
```
```php
$users = DB::table('users')
                ->whereJsonDoesntContainKey('options->language->primary')
                ->get();
```

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1777,6 +1777,57 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Add a "where JSON contains key" clause to the query.
+     *
+     * @param  string  $column
+     * @param  string  $boolean
+     * @param  bool  $not
+     * @return $this
+     */
+    public function whereJsonContainsKey($column, $boolean = 'and', $not = false)
+    {
+        $type = 'JsonContainsKey';
+
+        $this->wheres[] = compact('type', 'column', 'boolean', 'not');
+
+        return $this;
+    }
+
+    /**
+     * Add an "or where JSON contains key" clause to the query.
+     *
+     * @param  string  $column
+     * @return $this
+     */
+    public function orWhereJsonContainsKey($column)
+    {
+        return $this->whereJsonContainsKey($column, 'or');
+    }
+
+    /**
+     * Add a "where JSON not contains key" clause to the query.
+     *
+     * @param  string  $column
+     * @param  string  $boolean
+     * @return $this
+     */
+    public function whereJsonDoesntContainKey($column, $boolean = 'and')
+    {
+        return $this->whereJsonContainsKey($column, $boolean, true);
+    }
+
+    /**
+     * Add an "or where JSON not contains key" clause to the query.
+     *
+     * @param  string  $column
+     * @return $this
+     */
+    public function orWhereJsonDoesntContainKey($column)
+    {
+        return $this->whereJsonDoesntContainKey($column, 'or');
+    }
+
+    /**
      * Add a "where JSON length" clause to the query.
      *
      * @param  string  $column
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1777,7 +1777,7 @@ class Builder implements BuilderContract
     }
 
     /**
-     * Add a "where JSON contains key" clause to the query.
+     * Add a clause that determines if a JSON path exists to the query.
      *
      * @param  string  $column
      * @param  string  $boolean
@@ -1794,7 +1794,7 @@ class Builder implements BuilderContract
     }
 
     /**
-     * Add an "or where JSON contains key" clause to the query.
+     * Add an "or" clause that determines if a JSON path exists to the query.
      *
      * @param  string  $column
      * @return $this
@@ -1805,7 +1805,7 @@ class Builder implements BuilderContract
     }
 
     /**
-     * Add a "where JSON not contains key" clause to the query.
+     * Add a clause that determines if a JSON path does not exist to the query.
      *
      * @param  string  $column
      * @param  string  $boolean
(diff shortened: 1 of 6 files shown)


taylorotwell:
What if the user does `whereJsonContainsKey('options->foo->0')` ... where `options->foo` is an array like `['first', 'second', 'third']`?

amir9480:
@taylorotwell I'm confused about implementing this feature for the arrays as `whereJsonContains` doesn't support array keys on all database drivers.

derekmd:
Here's a sample integration test for Postgres where 3 of the 6 data sets currently don't pass: https://github.com/derekmd/framework/commit/ccf43d2979815f445dcd79ac0764c69264d95ad6

```php
/**
 * @dataProvider jsonContainsKeyDataProvider
 */
public function testWhereJsonContainsKey($count, $column)
{
    DB::table('json_table')->insert([
        ['json_col' => '{"foo":{"bar":["baz"]}}'],
        ['json_col' => '{"foo":{"bar":false}}'],
        ['json_col' => '{"foo":{}}'],
    ]);

    $this->assertSame($count, DB::table('json_table')->whereJsonContainsKey($column)->count());
}

public function jsonContainsKeyDataProvider()
{
    return [
        'key exists' => [3, 'json_col->foo'],
        'nested key exists' => [2, 'json_col->foo->bar'],
        'array key exists with arrow' => [1, 'json_col->foo->bar->0'],
        'array key exists with braces' => [1, 'json_col->foo->bar[0]'],
        'array key not exists' => [0, 'json_col->foo->bar[1]'],
        'key not exists' => [0, 'json_col->none'],
    ];
}
```

---

1. ✔️ Test passes
   ```php
   DB::table('json_table')->whereJsonContainsKey('json_col->foo')->count();
   ```
   ```sql
   where coalesce(("json_col")::jsonb ?? 'foo', false)
   ```
2. ✔️ Test passes
   ```php
   DB::table('json_table')->whereJsonContainsKey('json_col->foo->bar')->count();
   ```
   ```sql
   where coalesce(("json_col"->'foo')::jsonb ?? 'bar', false)
   ```
3. ❌ Test fails, must be ` ?? 0` with no wrapping quotes
   ```php
   DB::table('json_table')->whereJsonContainsKey('json_col->foo->bar->0')->count();
   ```
   ```sql
   where coalesce(("json_col"->'foo'->'bar')::jsonb ?? '0', false)
   ```
4. ❌ Test fails, array index must be parsed to match example 3's SQL: `("json_col"->'foo'->'bar')::jsonb ?? 0`
   ```php
   DB::table('json_table')->whereJsonContainsKey('json_col->foo->bar[0]')->count();
   ```
   ```sql
   where coalesce(("json_col"->'foo')::jsonb ?? 'bar[0]', false)
   ```
5. ❌ Test passes but for a different missing key. SQL must compile the same as example 4.
   ```php
   DB::table('json_table')->whereJsonContainsKey('json_col->foo->bar[1]')->count();
   ```
   ```sql
   where coalesce(("json_col"->'foo')::jsonb ?? 'bar[1]', false)
   ```
6. ✔️ Test passes
   ```php
   DB::table('json_table')->whereJsonContainsKey('json_col->none')->count();
   ```
   ```sql
   where coalesce(("json_col")::jsonb ?? 'none', false)
   ```

For `PostgresGrammar`, SQL in this new method may need to use parts of `wrapJsonPathAttributes()`.

1. avoid wrapping integer keys in quotes
2. PHP string `json_col->key[0][1]` must be compiled to SQL `("json_col"->'key'->0)::jsonb ?? 1`. The `$this->wrapJsonPathAttributes(["json_col", "key[0][1]"])` collection pipeline with `collapse()` reformats the JSON segments array like this:
   ```php
   ["json_col", "key[0][1]"]
   [["json_col", ["key", "0", "1"]]
   ["json_col", "key", "0", "1"]
   ["'json_col'", "'key'", 0, 1] // or ['"json_col"', '"key"', 0, 1] with arg $quote === "\""
   ```

taylorotwell:
Is it even possible to support numeric keys?

derekmd:
Experimental commits with integration tests covering array integer keys: https://github.com/derekmd/framework/commits/query-where-json-contains-key

### MySQL/MariaDB

👍 Supported and very straightforward with MySQL's `json_contains_path()` function.

### SQL Server

👍 Method `SqlServerGrammar@compileJsonContainsKey()` needs additional handling to extract the array index from the column string and avoid wrapping it in quotes. However `select [key] from openjson()` does support array integer keys.

### PostgreSQL

👍-ish, this is the most complicated implementation since Postgres functions differentiate between JSONB objects and arrays. Like SQL Server, the array index must be extracted from the column string. Also:
* checking for object string keys use the `?` operator (this PR's implementation)
* array integer keys call functions `jsonb_typeof()` and `jsonb_array_length()`
   * this checks the given index is within the (0 to n-1) the range.
   * Postgres also supports negative array indexes, counting from the array end
   * conditional `case when ... then ... else false end` must used to avoid calling `jsonb_array_length()` on a non-array

### SQLite

👎 This driver doesn't yet even implement existing method `whereJsonContains()` although I think that's now possible with [json_type()](https://www.sqlite.org/json1.html#the_json_type_function) `'WHERE json_type('.$field.$path.') IS NOT NULL'`).

From briefly looking at the docs, it may be possible to implement this using a mix of `json_type()`, `json_extract()`, and `json_each()`/key? I'd need to spend more time investigating this.

taylorotwell:
Thanks @derekmd 

Getting a bit lost though, so this PR is not ready to merge?

derekmd:
This PR is 80% there, I can finish the last 20% in another commit that can be merged into this.

PR #41758, opened 2022-03-31
[9.x] Enable batch jobs delay for database queue

civilcoder55:
[before]
- batch jobs was ignoring delay time when inserting into database.

[after]
- batch jobs delay time will be considered and stored into database.

```php
use App\Jobs\ImportCsv;
use Illuminate\Bus\Batch;
 
$batch = Bus::batch([
    (new ImportCsv(1, 100))->delay($delay),
    (new ImportCsv(101, 200))->delay($delay)
])->dispatch();

```

### DIFF
--- a/src/Illuminate/Queue/DatabaseQueue.php
+++ b/src/Illuminate/Queue/DatabaseQueue.php
@@ -145,10 +145,9 @@ class DatabaseQueue extends Queue implements QueueContract, ClearableQueue
     {
         $queue = $this->getQueue($queue);
 
-        $availableAt = $this->availableAt();
-
         return $this->database->table($this->table)->insert(collect((array) $jobs)->map(
-            function ($job) use ($queue, $data, $availableAt) {
+            function ($job) use ($queue, $data) {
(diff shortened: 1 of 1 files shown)


civilcoder55:
> @civilcoder55 thanks for this update. Does this work with setting the delay in the job constructor as well? I've no been able to bump to Laravel 9 yet, but this will be a nice improvement.

Hi, @georanma 
sorry for the delay in replying!

if you mean delay in the job constructor like this, yes you can do it 
```php
<?php

namespace App\Jobs;

use Illuminate\Bus\Queueable;

class SampleJob implements ShouldQueue
{

    private $data;

    /**
     * Create a new job instance.
     *
     * @return void
     */
    public function __construct(array $data)
    {
        $this->data = $data;
        $this->delay(now()->addMinute(1));
    }

    /**
     * Execute the job.
     *
     * @return void
     */
    public function handle()
    {
        //
    }
}
```

georanma:
@civilcoder55 awesome. Thank you again.

PR #41765, opened 2022-03-31
[9.x] Fire event before route matched

timroberson:
Fire a new RouteMatching event before the router attempts to find a matching route. This event is a companion to the existing RouteMatched event and allows a developer access to work with the request immediately before and after routing. Since this is simply adding a new event that existing applications will not be subscribed to, I believe it is fully backward compatible. Thank you for your consideration.


AdiechaHK:
Hey, @timroberson I don't get your idea, we can achieve the same using middleware so what is the point to raise this event. I just wanted to understand the reason why you added this.

Thanks

PR #38391, opened 2021-08-16
[9.x] Database queries containing JSON paths support array index references

derekmd:
Fixes: https://github.com/laravel/framework/issues/26415

When running queries that update JSON paths including an array index, Laravel compiles an SQL statement that escapes the array index as part of a JSON object key.

e.g.,

Database column `owner`.`packages` row with JSON payload:

```json
[
    {
        "name":"laravel/dusk",
        "versions":["6.17.0", "6.16.0", "6.15.1", "6.15.0"]
    },
    {
        "name":"laravel/framework",
        "versions":["8.54.0", "8.53.1", "8.53.0"]
    }
]
```

```php
DB::table('owner')
    ->where('packages[1]->name', 'laravel/framework')
    ->update(['packages[1]->versions[0]' => '9.0.0']);
```

For MySQL this compiles to:

```sql
UPDATE `owner`
SET `packages` = JSON_SET(`packages`, $"[1]"."versions[0]", '9.0.0')
WHERE json_unquote(json_extract(`packages[1]`, '$."name"')) = 'laravel/framework';
```

Which would instead append to the above payload with a new nested JSON object key:

```json
{
    "0": {
        "name":"laravel/dusk",
        "versions":["6.17.0", "6.16.0", "6.15.1", "6.15.0"]
    },
    "1": {
        "name":"laravel/framework",
        "versions":["8.54.0", "8.53.1", "8.53.0"]
    },
    "[1]": {
        "versions[0]": "9.0.0"
    }
}
```

When escaping JSON path segments, avoid enclosing array dereference characters `[` and `]`:

```sql
UPDATE `owner`
SET `framework` = JSON_SET(`framework`, $[1]."versions"[0], '9.0.0')
WHERE json_unquote(json_extract(`packages[1]`, '$."name"')) = 'laravel/framework';
```

---

### Database documentation for JSON path support

Each of Laravel's supported database drivers use the same syntax for JSON path array indices.

* MySQL: https://dev.mysql.com/doc/refman/8.0/en/json.html#json-path-syntax
* MariaDB: https://mariadb.com/kb/en/jsonpath-expressions/
* PostgreSQL: https://www.postgresql.org/docs/12/functions-json.html#FUNCTIONS-SQLJSON-PATH
* SQLite: https://www.sqlite.org/json1.html (scroll down for examples as the prose is bad readin')
* SQL Server: https://docs.microsoft.com/en-us/sql/relational-databases/json/json-path-expressions-sql-server?view=sql-server-ver15

[...]

PR #41767, opened 2022-03-31
[9.x] Database queries containing JSON paths support array index braces, part 2

### LINKED ISSUE #26415: mysql json field cannot update nested array item through nested array's index
- Laravel Version: 5.7.12
- PHP Version: 7.1.14
- Database Driver & Version:  MySQL 5.7.17

### Description:
Laravel framework cannot update array item through array index like: `'$.test[0].name'`, the sql generate by `\Illuminate\Database\Query\Grammars\Grammar::wrapJsonPath` seem like this `'$."test[0]".name'`,  but with double quotation mark won't update MySQL json field,  at the same time, the update statement return the wrong result without updating MySQL's record, database record not updated but get the true result.

### source code
```
    /**
     * Wrap the given JSON path.
     *
     * @param  string  $value
     * @return string
     */
    protected function wrapJsonPath($value)
    {
        return '\'$."'.str_replace('->', '"."', $value).'"\'';
    }
```

### Steps To Reproduce:
* MySQL test
  * `set @json = '{"test": [{"name": "laravel"}, {"name": "symfony"}]}';`
  * run `select json_set(@json, '$.test[0].name', "lumen");`,  no double quotation mark
    * result: `{"test": [{"name": "lumen"}, {"name": "symfony"}]} `, update successfully
  * run `select json_set(@json, '$."test[0]".name', "lumen");`, with double quotation mark
    * result: `{"test": [{"name": "laravel"}, {"name": "symfony"}]} `, update failed
* Laravel test
  *  The sql generate by MySQL grammer: 
~~~~
      update `xx` set `a` = json_set(`a`, '$."test[0]"."name"', lumen)
~~~~
  * This sql has double quotes wrap up `test[0]`, but it will not work for array in MySQL json field. Unless remove the double quotes.

derekmd:
Follow-up to https://github.com/laravel/framework/pull/38391 meant to fix https://github.com/laravel/framework/issues/26415
Related to today's PR https://github.com/laravel/framework/pull/41756

1. ```php
   DB::table('json_table')
       ->where('column->json_option[0]', 'foo')
       ->update(['column->json_option[0]', => 'bar']);
   ```

   The previous PR allowed the above query to match by JSON array index. But the fixes didn't apply to the Postgres database driver that overrides `PostgresGrammer@wrapJsonSelector()`.

   Postgres does already support this:

   ```php
   DB::table('json_table')
      ->where('column->json_option->0', 'foo')
      ->update(['column->json_option->0', => 'bar']);
   ```

   This fix allows the first PHP snippet to also work, bringing it in line with the other database drivers. SQL:

   ```sql
   -- before
   update "json_table"
   set "column" = jsonb_set("column"::jsonb, '{"json_option[0]"}', '"bar"'::jsonb)
   where ("column"->'json_option[0]')::jsonb = '"foo"'::jsonb;
   
   -- after, array index is extracted and braces are stripped
   update "json_table"
   set "column" = jsonb_set("column"::jsonb, '{"json_option",0}', '"bar"'::jsonb)
   where ("column"->'json_option'->0)::jsonb = '"foo"'::jsonb;
   ```

   Now the GitHub Actions config has a Postgres container I've added some integration tests.
2. `Illuminate\Database\Schema\Grammars\Grammar@wrapJsonFieldAndPath()` was a copy and paste duplicate from `Illuminate\Database\Query\Grammars\Grammar` and didn't receive the same fixes from the previous PR. It's a far edge case for column methods `virtualAsJson()` and `storedAsJson()` to use an array index but it probably makes sense for these classes to share a trait for compiling SQL containing JSON paths.

### DIFF
--- a/src/Illuminate/Database/Concerns/CompilesJsonPath.php
+++ b/src/Illuminate/Database/Concerns/CompilesJsonPath.php
@@ -1,0 +1,64 @@
+<?php
+
+namespace Illuminate\Database\Concerns;
+
+use Illuminate\Support\Str;
+
+trait CompilesJsonPath
+{
+    /**
+     * Split the given JSON selector into the field and the optional path and wrap them separately.
+     *
+     * @param  string  $column
+     * @return array
+     */
+    protected function wrapJsonFieldAndPath($column)
+    {
+        $parts = explode('->', $column, 2);
+
+        $field = $this->wrap($parts[0]);
+
+        $path = count($parts) > 1 ? ', '.$this->wrapJsonPath($parts[1], '->') : '';
+
+        return [$field, $path];
+    }
+
+    /**
+     * Wrap the given JSON path.
+     *
+     * @param  string  $value
+     * @param  string  $delimiter
+     * @return string
+     */
+    protected function wrapJsonPath($value, $delimiter = '->')
+    {
+        $value = preg_replace("/([\\\\]+)?\\'/", "''", $value);
+
+        $jsonPath = collect(explode($delimiter, $value))
+            ->map(fn ($segment) =>  $this->wrapJsonPathSegment($segment))
+            ->join('.');
+
+        return "'$".(str_starts_with($jsonPath, '[') ? '' : '.').$jsonPath."'";
+    }
+
+    /**
+     * Wrap the given JSON path segment.
+     *
+     * @param  string  $segment
+     * @return string
+     */
+    protected function wrapJsonPathSegment($segment)
+    {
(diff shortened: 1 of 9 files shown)


taylorotwell:
Thanks @derekmd! We would be pretty lost without you 😅

robsontenorio:
@derekmd 

I notice that `->update()` from Eloquent  does not use `jsonb_set`  syntax, but a default SQL  `SET "a" = '{b: 12}' `

```
$person->update([
  'a->b' => 12
]);
```

Is that expected, can we have in any way the same `jsonb_set` syntax  for Eloquent?

PR #41769, opened 2022-03-31
[9.x] Add `whenNotNull` method

smokills:
This PR adds a new whenSelectedMethod to the ConditionallyLoadsAttributes trait. This is basically a syntactic sugar around the when method. Working with the spatie query builder, I found that I was iterating the same checks for every resource, so maybe this little wrapper could be helpful to maintain the logic inside the Resource as clean as possible.

### DIFF
--- a/src/Illuminate/Http/Resources/ConditionallyLoadsAttributes.php
+++ b/src/Illuminate/Http/Resources/ConditionallyLoadsAttributes.php
@@ -147,8 +147,8 @@ trait ConditionallyLoadsAttributes
     /**
      * Retrieve a model attribute if it has been selected.
      *
-     * @param mixed $value
-     * @param mixed $default
+     * @param  mixed  $value
+     * @param  mixed  $default
      * @return \Illuminate\Http\Resources\MissingValue|mixed
      */
     protected function whenSelected($value, $default = null)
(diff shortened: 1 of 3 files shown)


smokills:
Another thing that I realized during the development, is that maybe could be handy to include the Macroable trait in the JsonResource class. In this way, developers could make packages in order to easily extend them. In my case I create another JsonResource base class, that extends the original one, with the trait included, so I was able to reuse the method above through all my Resources:

```php
JsonResource::macro('whenSelected', function ($column, $default = null) {
    func_num_args() == 1 ?
        $args = [$column] :
        $args = [$column, $default];

    return $this->when(!is_null($column), ...$args);
});
```

taylorotwell:
Should it be a bit more explicitly named? Like just `whenNotNull`?

smokills:
Sure thing! I'll do it right now. What about the Macroble trait? I think that the community would benefit from that.

taylorotwell:
JsonResource is already macroable via the DelegatesToResource trait (which uses Macroable).

smokills:
@taylorotwell first of all thanks for having merged my PR. I've just seen the DelegatesToResource trait, and in version 9.x I can't see the Macroble trait. Do you still think that it could be a good idea to have it? If so, I can provide a PR for both the framework and the doc.

PR #41770, opened 2022-03-31
Laravel is not respecting time zone settings in the php.ini

haydenk:
Currently, the `date.timezone` setting in the php.ini is ignored.

Removing "timezone" from the `app/config.php` will always default to UTC no matter what but `date_default_timezone_get()` will default to UTC if the `date.timezone` setting is not set but use the timezone specified if it is set. 

Developers can still set a default timezone in the php.ini, have it be respected, and override it in the application if necessary.

I have a follow-up change for `laravel/laravel` to comment the timezone config by default.

### DIFF
--- a/src/Illuminate/Foundation/Bootstrap/LoadConfiguration.php
+++ b/src/Illuminate/Foundation/Bootstrap/LoadConfiguration.php
@@ -46,7 +46,7 @@ class LoadConfiguration
             return $config->get('app.env', 'production');
         });
 
-        date_default_timezone_set($config->get('app.timezone', 'UTC'));
+        date_default_timezone_set($config->get('app.timezone', date_default_timezone_get()));
 
         mb_internal_encoding('UTF-8');
     }


driesvints:
This is massive breaking change and something we should absolutely avoid. I think UTC as a default is the right choice and this has been the default for many years already.

haydenk:
> This is massive breaking change and something we should absolutely avoid. I think UTC as a default is the right choice and this has been the default for many years already.

How is it a breaking change? If `date.timezone` is not set in the php.ini, `date_default_timezone_get` returns UTC which is what is currently hard coded. https://www.php.net/date_default_timezone_get

driesvints:
@haydenk if people have `date.timezone` set to something else than `UTC` in their php.ini they'll suddenly get a different timezone.

haydenk:
@driesvints That would mean that they explicitly set the timezone in PHP since it is not set by default. The `laravel/laravel` project could still keep setting the timezone to UTC by default as it could still override the setting.

GrahamCampbell:
Agree with Dries here. Choosing a timezone that is not UTC is almost always the wrong option, and even then, this is a huge breaking change.

PR #41773, opened 2022-04-01
[9.x] Ignore empty redis username string in `PhpRedisConnector`

seanmtaylor:
A username option was added to the `PhpRedisConnector` in https://github.com/laravel/framework/pull/41683.

However, some services, such as Heroku, produce a `REDIS_URL` connection string that has an empty username part.

This is causing the following exception:

```
RedisException: WRONGPASS invalid username-password pair or user is disabled.
```

So given the following `REDIS_URL` format:

```
redis://username:password@host:port
```

The following url string, with a blank username, will thrown an error.

```
redis://:abc123@redishost.example:12345
```

## Solution

I have added an extra check to ensure that the provided username isn't an empty string before attempting to use it.

## Considerations

I have used a `$config['username'] !== ''` check rather than `! empty($config['username'])` in the small off chance that the username value is falsey (eg `0`).

You might just say why not update the `REDIS_URL` and replace the empty username part with `null`. 

That's possible, but the issue is services such as Heroku will generate these `REDIS_*` variables at any time, such as when upgrading the redis instance, so it would need to be manually changed each time the credentials were generated.

## Tests

Sorry, I couldn't see any existing tests for this class so wasn't quite sure where to begin.

Feel free to point me in the right direction if a test is required for this.

### DIFF
--- a/src/Illuminate/Redis/Connectors/PhpRedisConnector.php
+++ b/src/Illuminate/Redis/Connectors/PhpRedisConnector.php
@@ -83,8 +83,8 @@ class PhpRedisConnector implements Connector
 
             $this->establishConnection($client, $config);
 
-            if (! empty($config['password'])) {
-                if (isset($config['username']) && is_string($config['password'])) {
+            if (!empty($config['password'])) {
+                if (isset($config['username']) && $config['username'] !== '' && is_string($config['password'])) {
                     $client->auth([$config['username'], $config['password']]);
                 } else {
                     $client->auth($config['password']);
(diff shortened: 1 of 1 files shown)


taylorotwell:
Thanks!

dwightwatson:
@seanmtaylor did this fix the issue for you? I just upgraded to v9.7.0 which appears to have this fix but I'm still getting a NOAUTH error on Heroku. 

Dropping back to 9.5.1 where this issue wasn't present.

dwightwatson:
If it's worth noting - I have 2 Redis instances connected to my Laravel app on Heroku. One has no username like above, and the other has the username `h`. No idea why - perhaps Heroku changed the auth scheme at some point.

However, it worked fine on 9.5.1 and not beyond.

seanmtaylor:
> @seanmtaylor did this fix the issue for you? I just upgraded to v9.7.0 which appears to have this fix but I'm still getting a NOAUTH error on Heroku.
> 
> Dropping back to 9.5.1 where this issue wasn't present.

Yeah it has sorted it for me @dwightwatson

PR #41776, opened 2022-04-01
Custom key support for implicit enum route binding

Propaganistas:
This PR introduces "custom key" support for implicitly bound enums in routes, similarly to Eloquent models.

```php
Route::get('{category:slug}', ...);
```

```php
enum Category: int {
    case People = 1;
    case Fruits = 2;

    public function slug(): string
    {
        // Up to the developer to define this...
        return match($this) {
            self::People => 'people',
            self::Fruits => 'fruits',
        };
    }
}
```

**Rationale**
An enum is often used as a model attribute to simplify DX while still efficiently persisting integers in the database. There are ample of situations where a model's enum would be included in a route as well. Currently Laravel is only able to implicitly resolve string-backed enum route bindings, but that sort of contradicts the aforementioned use case.

One could of course go with explicitly defining such an enum route binding. But since implicit model binding features a custom route key I figured a similar feature would fit the enum binding as well.

**Considerations**
The implementation of the desired "key" method is not airtight. It's up to the developer to define a value for every enum case. However, this can also be seen as an opportunity: a specific case can simply be omitted to make the binding return a 404 for that case...

**Original PR for implicit enum binding**
https://github.com/laravel/framework/pull/40281

### DIFF
--- a/src/Illuminate/Routing/Exceptions/BackedEnumImplicitKeyMethodNotFoundException.php
+++ b/src/Illuminate/Routing/Exceptions/BackedEnumImplicitKeyMethodNotFoundException.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Illuminate\Routing\Exceptions;
+
+use RuntimeException;
+
+class BackedEnumImplicitKeyMethodNotFoundException extends RuntimeException
+{
+    /**
+     * Create a new exception instance.
+     *
+     * @param  string  $backedEnumClass
+     * @param  string  $key
+     * @return void
+     */
+    public function __construct($backedEnumClass, $key)
+    {
+        parent::__construct("Implicit key method [{$key}] not found on Backed Enum [{$backedEnumClass}].");
+    }
+}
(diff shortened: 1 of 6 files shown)


taylorotwell:
Personally I wonder if this would be better solve by the addition of a `whereIn` regex helpers on routes (like `whereAlpha`), etc.

```php
Route::get('/foo/{bar}')->whereIn('bar', Category::slugs());
```

Where `slugs` returns an array of valid values. This feels like a more flexible approach that can solve a variety of use cases, including this one. Feel like taking a stab at that?

taylorotwell:
`CreatesRegularExpressionRouteConstraints` is the file you would be interested in.

Propaganistas:
@taylorotwell On a further thought, this implementation might still be worthwhile to consider in light of parameter injection:

If the enum is string-backed, implicit route binding would return an instantiated enum:
```php
Route::get('/foo/{bar}', function($bar) {
    // $bar is an instantiated enum
});
```

Using the implementation put forward in this PR, binding to a custom key would also return an instantiated enum:
```php
Route::get('/foo/{bar:slug}', function($bar) {
    // $bar is an instantiated enum
})->whereIn('bar', Category::slugs());
```

While the newly added `whereIn` constraint would return the plain parameter string:
```php
Route::get('/foo/{bar}', function($bar) {
    // $bar is a string
})->whereIn('bar', Category::slugs());
```

PR #41779, opened 2022-04-01
[9.x] Http Client: provide pending request to retry callback

gdebrauwer:
In a couple of our projects, we are using the `retry` method of the Http Client to handle authentication errors. If a 401 error is thrown, we request a new token and then retry.
To achieve this, we currently have to manually create a request variable and pass the variable inside the closure in order to modify the request and set a new token on the request.
```php
($request = Http::baseUrl(...))
  ->withToken($this->getToken())
  ->retry(2, 0, function ($exception) use ($request) {
      if ($exception->response->status() !== 401) {
          return false;
      }
  
      $request->withToken($this->getNewToken());
  
      return true;
  })
  ->get('...');
```
This PR allows you to access the request as a parameter of the closure:
```php
Http::baseUrl(...)
  ->withToken($this->getToken())
  ->retry(2, 0, function ($exception, $request) { 
      if ($exception->response->status() !== 401) {
          return false;
      }
  
      $request->withToken($this->getNewToken());
  
      return true;
  })
  ->get('...');
```

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -715,7 +715,7 @@ class PendingRequest
                     $this->populateResponse($response);
 
                     if (! $response->successful()) {
-                        $shouldRetry = $this->retryWhenCallback ? call_user_func($this->retryWhenCallback, $response->toException()) : true;
+                        $shouldRetry = $this->retryWhenCallback ? call_user_func($this->retryWhenCallback, $response->toException(), $this) : true;
 
                         if ($attempt < $this->tries && $shouldRetry) {
                             $response->throw();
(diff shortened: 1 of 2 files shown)


taylorotwell:
@gdebrauwer  It seems surprising to me that this works. Aren't the options given to `sendRequest` already built? I wouldn't expect modifying the request to actually change them.

gdebrauwer:
The `sendRequest` method is called every time the request is retries. The `$options` array paramer only contains querystring data or request body data. It merges that `$options` array with `$this->options` in the `sendRequest` method. Methods on the `PendingRequest`, like the `withToken` method, make changes to that `$this->options` array, so that's why it works 🙂

taylorotwell:
@gdebrauwer can you resolve the conflicts on this PR?

PR #41782, opened 2022-04-01
[9.x] Fix afterCommit and RefreshDatabase

### LINKED ISSUE #35857: `DB::afterCommit` callbacks aren't run in tests when `RefreshDatabase` trait is used
- Laravel Version: v8.21.0
- PHP Version: 7.4.11
- Database Driver & Version: psql (PostgreSQL) 12.4 (Ubuntu 12.4-1.pgdg20.04+1) - It doesn't depend on DB driver.

### Description:

`DB::afterCommit` callbacks aren't run when `RefreshDatabase` trait is used because `ManagesTransactions` trait only commits when all transactions are executed, but `RefreshDatabase` uses a transaction. Then the callbacks aren't called.

### Steps To Reproduce:

I've created a reproduction repo: https://github.com/mariomka/laravel-after-commit-repro

There are two tests, one with `RefreshDatabase` (it fails) and one without `RefreshDatabase` (it passes): https://github.com/mariomka/laravel-after-commit-repro/tree/master/tests

Both tests run the same "action": https://github.com/mariomka/laravel-after-commit-repro/blob/master/app/AfterCommitAction.php

Step by step reproduction:

```sh
git clone git@github.com:mariomka/laravel-after-commit-repro.git
cd laravel-after-commit-repro
cp .env.example .env
composer install
php vendor/bin/homestead make
# Enable postgresql in Homestead.yaml
vagrant up
vagrant ssh
cd code
php vendor/bin/phpunit
```

taylorotwell:
Previously, using the `RefreshDatabase` trait was fundamentally incompatible with `afterCommit`. This is because the trait is based around the fact that it opens a transaction and rolls it back at the end of the test, never committing it. 

This slight modification updates the `DatabaseTransactionManager` to "ignore" a given transaction when determining if a callback should be executed, since we want the outer transaction to essentially be invisible.

Would appreciate some feedback on if this solves people's problems and / or creates new problems.

Fix #35857

### DIFF
--- a/src/Illuminate/Database/Concerns/ManagesTransactions.php
+++ b/src/Illuminate/Database/Concerns/ManagesTransactions.php
@@ -47,9 +47,7 @@ trait ManagesTransactions
 
                 $this->transactions = max(0, $this->transactions - 1);
 
-                if ($this->transactions == 0 ||
-                   ($this->transactionsManager &&
-                    $this->transactionsManager->callbackApplicableTransactions()->count() === 1)) {
+                if ($this->afterCommitCallbacksShouldBeExecuted()) {
                     $this->transactionsManager?->commit($this->getName());
                 }
             } catch (Throwable $e) {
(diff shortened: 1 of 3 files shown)


BrandonSurowiec:
Shouldn't this target 8.x?

driesvints:
@taylorotwell I think @BrandonSurowiec is correct and we should target 8.x

taylorotwell:
Hmm, I'm torn - existing users can just use `DatabaseMigrations` trait instead of `RefreshDatabase` on 8.x. Let's see if this actually works out on 9.x first.

BrandonSurowiec:
Hey @taylorotwell I use the `LazilyRefreshDatabase` trait on my base TestCase, so in my 8.x project it is difficult to switch to `DatabaseMigrations`. This PR seems to have "worked out" so would you accept a PR to backport this and #42502 into 8.x? I ran into this issue today and would love to delete my workaround.

reza305z:
@taylorotwell still have this issue on laravel 9.42.2 using postgres as database for testing.

mpyw:
Still reproduces if the transactions are nested. Please check #48451

PR #41784, opened 2022-04-01
[9.x] Add `htmlContains` and `textContains` to `Mailable` class

xiCO2k:
This PR adds `htmlContains` and `textContains` to the `Mailable` class.

This way the user can check with the `Mail::assertSent` method:

## Example:

```php
Mail::fake();

Mail::assertSent(NewUserMail::class, fn ($mail) =>
    $mail->hasTo('john@doe.com') &&
    $mail->hasSubject('Login Information') &&
    $mail->htmlContains($password) &&
    $mail->textContains($password)
);
```

I also notice that are methods like the `assertSeeInHtml` but that does not work well with the `Mail::fake()`.

We can also rename it to be called like `->seeInHtml` and `->seeInText` if you think that is better.

Hope this is a cool addition,
Thanks,
Francisco

### DIFF
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -877,7 +877,7 @@ class Mailable implements MailableContract, Renderable
         $view = $this->buildView();
 
         return str_contains(
-            is_array($view) ? $view['text'] ?? $view[1] ?? '' : $view,
+            is_array($view) ? $view['text'] ?? $view[1] ?? '' : '',
             $string
         );
     }
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -839,14 +839,16 @@ class Mailable implements MailableContract, Renderable
      */
     public function htmlContains(string $string)
     {
-        Container::getInstance()->call([$this, 'build']);
+        return $this->withLocale($this->locale, function () use ($string) {
+            Container::getInstance()->call([$this, 'build']);
 
-        $view = $this->buildView();
+            $view = $this->buildView();
 
-        return str_contains(
-            is_array($view) ? $view['html'] ?? $view[0] ?? '' : $view,
-            $string
-        );
+            return str_contains(
+                is_array($view) ? $view['html'] ?? $view[0] ?? '' : $view,
+                $string
+            );
+        });
     }
 
     /**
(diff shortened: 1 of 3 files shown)


taylorotwell:
I kinda prefer people test content of mailables under certain circumstances separate from sending with given arguments:

https://laravel.com/docs/9.x/mail#testing-mailables

Thoughts?

xiCO2k:
I think both ways are good to have.

This is really useful for this feature tests like this example:

```php
it('allows an admin to create a new user', function () {
    Mail::fake();
    signInAsAdmin();

    $this->post(route('users.store'), [
        'name' => $name = 'John Doe',
        'email' => $email = 'john.doe@example.com',
        'password' => $password = 'SomePasswordOverEightChars12345@',
    ])->assertRedirect(route('users.index'));

    $user = User::whereEmail($email)->first();

    expect($user)
        ->name->toEqual($name)
        ->password->not->toBeEmpty();

    Mail::assertSent(NewUserMail::class, fn ($mail) =>
        $mail->hasTo($email) && 
        $mail->hasSubject('Login Information') &&
        $mail->htmlContains($password)
    );
});
```

GC-Mark:
I agree these can be useful to have, but not having these methods kinda forced us in our company to test the _entire_ email HTML under varying circumstances, with help from https://github.com/spatie/phpunit-snapshot-assertions which I think gives a more solid test and proves your email will render correctly as well as having the correct content.

It was initially surprising that you couldn't already easily do what this PR is proposing, so I think it is a 'nice to have' 👍

taylorotwell:
What about when locales are used? Note `render` method.

CleanShot 2022-04-06 at 10 16 02@2x

JayBizzle:
@xiCO2k I don't think draft PRs get reviewed, so if this is ready, you need to remove the `Draft` status 👍

xiCO2k:
Thanks @JayBizzle, I'm aware, is not ready to be reviewed, I will need to add a test and figure one issue that I found.

xiCO2k:
Hey @taylorotwell sorry for the delay

Found one issue that probably invalidates this PR.

If the user uses the `->view('name')` method, the `buildView` does not contain the html, and it only contains the view name `string`, for that is needed to run the:

```php
return Container::getInstance()->make('mailer')->render(
    $this->buildView(), $this->buildViewData()
);
```

And that does not work with the `Mail::fake()`.

I guess that may invalidate this PR, but will leave it here, if you have any ideas.

Thanks.

taylorotwell:
Will close for now.

PR #40790, opened 2022-02-03
[9.x] Fix Carbon::setTestNow usage

lucasmichot:
In many place in the tests, the usage of Carbon::setTestNow does not make sense or is not used properly.
This PR fixes that.

See PR code comments.

---

In a ideal world, a way to go would be to have an abstract test class that every other test (excepted integration) class could  extend, this abstract class would close Mockery and reset Carbon in its tearDown method.

PR #41791, opened 2022-04-02
[9.x] Add String::squish() helper

dwightwatson:
This proposes a new string helper method called `squish`, which is named similarly to the one in Rails. I know there's a lot of fair scrutiny given to adding new helpers but felt it was worth putting it out there in case you saw the usefulness that I do.

My use-cases are both when outputting user input that had additional unnecessary whitespace, and also when creating interpolated strings that have optional values (i.e. `"A {$description} listing has been created."`). In both of these instances using `squish` would make the content appear more professional.

If you do like this addition I can follow-up and add it to `Stringable` as well.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -882,7 +882,7 @@ class Str
     }
 
     /**
-     * Squish all blank space in a string.
+     * Squish all blank space from the given string.
      *
      * @param  string  $value
      * @return string
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -882,7 +882,7 @@ class Str
     }
 
     /**
-     * Squish all blank space from the given string.
+     * Remove all "extra" blank space from the given string.
      *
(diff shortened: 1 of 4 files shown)


dwightwatson:
Looks like that failing test case may not be my fault 🤞

derekmd:
You can try force-pushing your branch to re-trigger GitHub Actions and get all-green.

A recent merged PR can cause `CacheFileStoreTest@testIncrementDoesNotExtendCacheLife()` to intermittently fail if the second clock ticks over mid-case: https://github.com/laravel/framework/pull/40790#discussion_r841080542

taylorotwell:
Can you also add it to fluent strings?

dwightwatson:
Of course, added to `Stringable` as well.

rcerljenko:
@dwightwatson could the replacement param be passed as second argument like this:
```php
public static function squish($value, $replacement = ' ')
{
    return preg_replace('/\s+/', $replacement, trim($value));
}
```

PR #36526, opened 2021-03-09
[8.x] Allow nullable columns for AsArrayObject/AsCollection casts

tomirons:
When using either of these casts and allowing the column to be `null` it throws an undefined error whenever you access the attribute.

This fixes that issue.

PR #41797, opened 2022-04-02
[9.x] Fix support of nullable type for AsArrayObject/AsCollection

a-bashtannik:
Previous fix https://github.com/laravel/framework/pull/36526 of these 2 classes not covered the behavior of php `json_encode` / `json_decode` functions when casting null.

```php
// Bugly fragment
{
    // if $attributes[$key] === "null" it is assumed as "set" but ArrayObject can't accept null values.
    return isset($attributes[$key]) ? new ArrayObject(json_decode($attributes[$key], true)) : null;
}
```

If you have tried to set `AsArrayObject` attribute value to null (or had it by default), `json_encode` processed it as sting `"null"` which is correct from the point of JSON, and it was stored successfully. Nothing wrong here.

But next time you tried to access the attribute, the error `ArrayObject::__construct(): Argument #1 ($array) must be of type array, null given` was raised because it was already converted to a string and `json_decode` returns `null` taking `"null"` as an argument, which is not acceptable by [ArrayObject](https://php.net/manual/en/class.arrayobject.php). `Collection` class casts null to an empty array under the hood, there was no problem on practice, but I consider better to keep the same predictable behavior of `get` methods for both cast attribute classes. 

This PR also fix potential issues with `json_encode` / `json_decode` casting boolean values.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Casts/AsArrayObject.php
+++ b/src/Illuminate/Database/Eloquent/Casts/AsArrayObject.php
@@ -21,7 +21,7 @@ class AsArrayObject implements Castable
             {
                 $data = json_decode($attributes[$key], true);
 
-                return is_array($data) ? new ArrayObject($data) : new ArrayObject();
+                return is_array($data) ? new ArrayObject($data) : null;
             }
 
             public function set($model, $key, $value, $attributes)
--- a/src/Illuminate/Database/Eloquent/Casts/AsArrayObject.php
+++ b/src/Illuminate/Database/Eloquent/Casts/AsArrayObject.php
@@ -19,7 +19,9 @@ class AsArrayObject implements Castable
         {
             public function get($model, $key, $value, $attributes)
             {
-                return isset($attributes[$key]) ? new ArrayObject(json_decode($attributes[$key], true)) : null;
+                $data = json_decode($attributes[$key], true);
+
+                return is_array($data) ? new ArrayObject($data) : new ArrayObject();
             }
 
             public function set($model, $key, $value, $attributes)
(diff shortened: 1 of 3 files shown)


taylorotwell:
Why would it return an empty array object vs. null? It should just return null.

a-bashtannik:
@taylorotwell it was a doubt, but I concluded it should be an empty ArrayObject because otherwise, it ruins the main purpose, and idea of `AsArrayObject` — casting the attribute to the object and allowing it to behave like an array.

You can't just call 
```
$model->array_object['meta']['title'] = 'Developer';
```
Because `array_object` attribute is `null`. 

You need to create something new 
```
$model->array_object = ['meta' => ['title' => 'Developer']];
``` 
instead, but that's a bit against `AsArrayObject` idea, no?

Anyway, both variants are acceptable and close the bug.

foremtehan:
https://github.com/laravel/framework/issues/38570

a-bashtannik:
@foremtehan you are right, [docs are misleading](https://laravel.com/docs/9.x/eloquent-mutators#array-object-and-collection-casting), I followed them in the wrong direction designing the behavior of `AsArrayObject`/`AsCollection`. 

Ok then, I will update this PR and docs. Should contain also one-two examples more. 

```php
$user = new User();
$user->options = ['foo' => ['key' => $value]];
$user->options['foo']['key'] = $value;
```

AyubM:
Taylor, why should it return null instead of an empty ArrayObject? The whole point of casting a property as such is to allow subproperty modification.

As things are, `$model->info['foo'] = 'bar'` will fail unless the value had previously been set. Which, for new objects, isn't going to be the case.

Major inconvenience if it can return null -- devs either need to check and instantiate as an empty array before modifying any subproperties, or they will have to set the json column to default as {} in the migration as a hack.

PR #41806, opened 2022-04-03
[9.x] Allow cleanup of databases when using parallel tests

abenerd:
The number of created database can get very high when having several laravel projects that use parallel testing. This option allows laravel to clean up after itself when the process is finished dropping all created databases used for the tests.


nunomaduro:
@abenerd Do you mind of now do the necessary changes to the Collision package? 

https://github.com/nunomaduro/collision/blob/stable/src/Adapters/Laravel/Commands/TestCommand.php

abenerd:
Yeah, just created the PR @nunomaduro.

PR #41807, opened 2022-04-03
[9.x] Add String isHex() helper method & validation rule

adamczykpiotr:
This PR proposes a new Str and Stringable helper method `isHex` that validates whether given value is a valid hex string - characters in range of a-f, A-F, 0-9. Method optionally takes a second parameter (`$prefix`) to simplify validating certain strings (like HTML colors `#FFFFFF` or literal string values like `0x1E5`). I've also included a new `hex` validator rule.

Examples:
**True:**
```php
// Ordinary value
Str::isHex('abcdef123456') 

// Prefixed with '0x'
Str::isHex('0x1E5', '0x')

// HTML colour
Str::isHex('#FFFFFF', '#')

// Stringable interface
Str::of('0x1E5')
    ->isHex('0x')

// Validator
Validator::validate([
  'value' => 'abcdef123456'
], [
  'value' => 'hex'
]);

// Validator with optional param
Validator::validate([
  'value' => '0x123'
], [
  'value' => 'hex:0x'
]);
```

**False:**

```php
// Not a string
Str::isHex([])

// Character out of range (g)
Str::isHex('abcdefg123456')

// Character out of range (x), missing prefix
Str::isHex('0x1E5')

// Character out of range (#), missing prefix
Str::isHex('#FFFFFF')
```

Issue #9988, opened 2015-08-19
[Proposal] Validation rule for hexadecimal values 


prady00:
This will be useful for checking hexadecimal values, for instance the color hex like #000000 (black) and #ffffff (white)

bmitch:
@prady00  Are you thinking about submitting a pull request? If not I can give it a go.

lukasgeiter:
Couldn't this easily be accomplished using the `regex` validation rule?

bmitch:
Yeah that's a good point @lukasgeiter

prady00:
@bmitch : I've already coded it for one of my current project, just need to move it to Validator.php and do a simple PR.

prady00:
@lukasgeiter using regex all types of validations can be performed. IP, alpha, numbers, URLs etc.

We include easy subsystem for something fast and non-time consuming in terms of development.

lukasgeiter:
@prady00 Maybe not all types, but in most cases that's true. I just wanted to point out the possibility of using the `regex` validation rule, in case someone didn't know about it. I'm not particularly against it.

Just send the PR and we'll see if Taylor is up for it.

PR #41813, opened 2022-04-03
[9.x] Add 'hex' validation rule

adamczykpiotr:
This is a proposal of a single new `hex` validation rule that checks whether a string has all characters in range of a-f, A-F, 0-9. Second optional parameter is a prefix to add a simple way of handling html colours or literal hex values.

It's an implementation of the proposal https://github.com/laravel/framework/issues/9988 and also solution to a real problem i've had while working on a project. 

**Examples:**
```php
// Simple value
Validator::validate([
  'value' => 'abcdef123456'
], [
  'value' => 'hex'
]);

// Optional param - literal hex value
Validator::validate([
  'value' => '0x123'
], [
  'value' => 'hex:0x'
]);

// Optional param - HTML colour
Validator::validate([
  'value' => '#FFFFFF'
], [
  'value' => 'hex:#'
]);
```

This PR is also a followup to my previous https://github.com/laravel/framework/pull/41807. I agree that particular string helper might not have been the most frequently used method but a simple validation rule surely will be quite handy.

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -2039,6 +2039,31 @@ trait ValidatesAttributes
     }
 
     /**
+     * Validate that an attribute is a hex string.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @param  array  $parameters
+     * @return bool
+     */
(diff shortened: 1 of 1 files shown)


a-bashtannik:
Not really sure it's a common need, since it's just a simple regexp validation rule, but at least some tests are needed.

taylorotwell:

PR #41824, opened 2022-04-04
[9.x] Fix time mocking in CacheRepositoryTest

kylekatarnls:
`setTestNow()` should not be done in `dataProvider`. dataProvider in general shouldn't modify any state as they are triggered by PHPUnit in early steps and not related to the tests done with this provider.

For this case (`CacheRepositoryTest`) the time should be always mocked, else stuff like `$repo->getStore()->shouldReceive('put')->once()->with('foo', 'bar', 602)` can't be guaranteed because few microseconds elapsed since `Carbon::now()->addMinutes(10)->addSeconds(2)` was called.

Relates to #41810, #40790

### DIFF
--- a/tests/Cache/CacheRepositoryTest.php
+++ b/tests/Cache/CacheRepositoryTest.php
@@ -264,7 +264,7 @@ class CacheRepositoryTest extends TestCase
     public function dataProviderTestGetSeconds()
     {
         return [
-            [Carbon::now()->addMinutes(5)],
+            [Carbon::parse($this->getTestDate())->addMinutes(5)],
             [(new DateTime($this->getTestDate()))->modify('+5 minutes')],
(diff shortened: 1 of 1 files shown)


tests/Cache/CacheRepositoryTest.php
```diff
@@ -256,8 +263,6 @@ public function testAddWithDatetimeInPastOrZeroSecondsReturnsImmediately()
 
     public function dataProviderTestGetSeconds()
     {
-        Carbon::setTestNow(Carbon::parse($this->getTestDate()));
-
         return [
             [Carbon::now()->addMinutes(5)],
```
kylekatarnls, line 267:
```suggestion
            [Carbon::parse($this->getTestDate())->addMinutes(5)],
```

PR #41827, opened 2022-04-04
Get value of backed enums in schema builder default value

paulhennell:
A very small scratching my own itch PR - but working with enums it bugged me that I have to reference the enum value when it's used as a schema default:
`$table->string('status')->default(Status::PENDING->value);`

This PR makes this nicer looking version work:
`$table->string('status')->default(Status::PENDING);`

Test would be logical to inline with the similar tests above, but I think it has to be separated to only run under PHP 8.1 .


driesvints:
Please see previous attempts. Schemas should not reference enums like this.

PR #41839, opened 2022-04-05
[9.x] Update maintainer action

driesvints:
These changes do three things:
1. Update `actions/github-script` to the latest version
2. Explain in the posted comments that PR's sent from an organization cannot be edited
3. Automatically close the PR and ask to resubmit with maintainer permissions

I mainly submit this because atm there's no actual action taken against PRs without maintainer permissions so I feel we should close them until they're sent in with proper permissions. Otherwise we can't handle them anyway. 

Other than that I see quite a bit of PR's sent in from organizations. Atm GitHub doesn't allows those PR's to be edited by maintainers so we should proactively inform the OP about that.

### DIFF
--- a/.github/workflows/pull-requests.yml
+++ b/.github/workflows/pull-requests.yml
@@ -13,9 +13,11 @@ permissions:
 jobs:
   uneditable:
     if: github.repository == 'laravel/framework'
+
     runs-on: ubuntu-latest
+
     steps:
-      - uses: actions/github-script@2b34a689ec86a68d8ab9478298f91d5401337b7d
+      - uses: actions/github-script@v6
         with:
           script: |
             const query = `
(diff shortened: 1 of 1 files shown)


Jubeki:
Just an idea: What about adding a label to the PR that no maintainer access was granted?

I think it is also not possible to grant maintainer access if you create a PR from the default branch (`9.x` in this case, though I am not sure about that). Maybe this should also be mentioned.

driesvints:
> Just an idea: What about adding a label to the PR that no maintainer access was granted?

It's a good idea but I'd rather avoid polluting our labels too much. I'm probably gonna look into adding labels to auto generate changelogs soon and I don't want to overextend on too much extra ones.

> I think it is also not possible to grant maintainer access if you create a PR from the default branch (9.x in this case, though I am not sure about that). Maybe this should also be mentioned.

I don't think that's true? You should always be able to grant maintainer access to any branch you send in a PR (except from an organization account - which is now mentioned).

PR #41846, opened 2022-04-06
[9.x] Stringable exactly not match with other Stringable variable

medeiroz:
- Laravel Version: 9.5.1
- PHP Version: 8.1

### Description:
When need compare two Stringable variables with exactly method. Then false is returned

### Steps To Reproduce:

```php
<?php

$first = \Str::of('name');
$second = \Str::of('name');
$first->exactly($second);
//  => false

// true is expected but false is returned

```

image

## Solution

- Cast argument to string when argument is Stringable

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -214,7 +214,7 @@ class Stringable implements JsonSerializable
      */
     public function exactly($value)
     {
-        return $this->value === $value;
+        return $this->value === (string) $value;
     }
 
     /**
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -215,7 +215,7 @@ class Stringable implements JsonSerializable
     public function exactly($value)
     {
         if ($value instanceof Stringable) {
-          $value = $value->toString();
+            $value = $value->toString();
         }
 
         return $this->value === $value;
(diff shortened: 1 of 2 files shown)


SjorsO:
Doesn't make a lot of sense to have a method called `exactly` that does a loose comparison.

Might be a better idea to do this instead:
```php
if ($value instanceof Stringable) {
    $value = $value->toString();
}

return $this->value === $value;
```

driesvints:
The method dockblock clearly indicates you can only pass a string.

imanghafoori1:
```php
if (is_object($value) && method_exists('__toString', $value)) {
    $value = (string) $value;
}
```

So that there is no loose comparison will happen.

dennisprudlo:
@imanghafoori1 I'm not sure any object that has a `__toString()` should be a valid parameter.
```php
$model = new \App\Models\SomeModel;
str('[]')->exactly($model); // true
```

The type in the docblock says only `string` can be passed, so maybe just add `Stringable` (the class not the global interface) to the docblock and check like @SjorsO did.

GrahamCampbell (changes requested):
👎 You are incorrectly calling this function if you do not pass a string.

GrahamCampbell:
If you want this function to support stringable objects, then we would have to widen its param type.

driesvints:
@GrahamCampbell that's what this PR tries to do 😅  I also misjudged this PR at first for wrong usage but I can see the use case here.

pedroaff:
The two itens has the exact values, so it should not return false.
It's makes sense when we need to compare two itens w/ the same value.

PR #41849, opened 2022-04-06
[9.x] Set the foreign ID for a related model

sander3:
Set the foreign ID for related models when using `$model->push()`. Related to #4641.

This allows you to create a new model and its relations in one go:

```php
$checkout = $user->checkouts()->make();

$checkout->carts->push($checkout->carts()->make(['amount' => '1000', 'currency' => 'EUR']));
$checkout->carts->push($checkout->carts()->make(['amount' => '750', 'currency' => 'EUR']));

$checkout->calculate(); // Use the set relations to calculate the checkout totals for example

$checkout->push();
```


taylorotwell:
Likely a breaking change.

PR #41850, opened 2022-04-06
[9.x] Add tests for session JSON serialization

Krisell:
This adds tests for the new session JSON serialization option added in Laravel 9 (https://github.com/laravel/framework/pull/40595).

The added tests does not duplicate all test cases for `serialize/unserialize`, but they do test both reading and writing to the session.

### DIFF
--- a/tests/Session/SessionStoreTest.php
+++ b/tests/Session/SessionStoreTest.php
@@ -35,6 +35,22 @@ class SessionStoreTest extends TestCase
         $this->assertTrue($session->has('baz'));
     }
 
+    public function testSessionDataCanBeSerializedAsJson()
+    {
+        $session = $this->getSession('json');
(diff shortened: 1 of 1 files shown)


Krisell:
Should the new JSON serialization option be added to the documentation (e.g. here https://laravel.com/docs/9.x/session#configuration) or should the `serialization` option perhaps be mentioned in the `config/session.php` template?

taylorotwell:
@Krisell JSON serialization is totally broken for view error bag sharing and should probably be removed.

Krisell:
@taylorotwell I did see your comment here now, and the related issue: https://github.com/laravel/laravel/pull/5796#issuecomment-1030098327

Has your suggest solution there been attempted?

I think using JSON is preferred for several reasons, as described by you in https://github.com/laravel/laravel/pull/5787 and also for the smaller payload size.

PR #41859, opened 2022-04-07
[9.x] Add option to disable cached view

bernardwiesner:
Especially during development its very helpful to disable the cached views. Here are a few use cases:

- Switching domains on local from 127.0.0.1 to local-your-domain.com on local (such as when using ssl certificates)
When switching domains the same compiled view is returned using the old domain in JS, CSS, etc.

- Custom implementation that add modify contents on compile time, such as custom cache busters for JS.
Adding a query string to JS for cache busting done with PHP at compile time.

And many other use cases where view caching might not be desirable on local.

Relates to: https://github.com/laravel/framework/issues/2501

Would enable inside `config/view.php` using

```php
'cache' => App::environment('local') ? false : true
```

### DIFF
--- a/src/Illuminate/View/Compilers/Compiler.php
+++ b/src/Illuminate/View/Compilers/Compiler.php
@@ -30,16 +30,24 @@ abstract class Compiler
     protected $basePath;
 
     /**
+     * Determines if cached view should be used.
+     *
+     * @var bool
+     */
+    protected $useCache;
+
+    /**
      * Create a new compiler instance.
      *
      * @param  \Illuminate\Filesystem\Filesystem  $files
      * @param  string  $cachePath
      * @param  string  $basePath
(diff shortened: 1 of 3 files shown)


xxdalexx:
This comes with a big caveat of test speed.
For the project I tried this on, running the full suite not in parallel:
With View Caching: 4.8 Seconds
Without: 1 Minute, 19 Seconds

bernardwiesner:
@xxdalexx 
Thanks for pointing out, however I think most folks won't have an issue since by default phpunit.xml is already set to testing environment.
https://github.com/laravel/laravel/blob/d0603437cbbb478586979a3792d49e0d157ce554/phpunit.xml#L21

joelmellon:
@bernardwiesner Thanks! Great feature.

PR #41862, opened 2022-04-07
[9.x] Display CallbackEvent description in schedule:list

ehamrin:
Before the redesign of `php artisan schedule:list`, the description for anonymous closures was displayed - giving the developer a possibility to differentiate between callbacks. After the redesign the description was moved to `-v` and an addition was made to display the closure location as command name.

Having a list of closure locations doesn't help much if you have a lot of closures:
```
 * * * * *  Closure at: app\Console\Kernel.php:32 ............... Next Due: 48 minutes from now
 * * * * *  Closure at: app\Console\Kernel.php:48 ............... Next Due: 48 minutes from now
 * * * * *  Closure at: app\Console\Kernel.php:70 ............... Next Due: 48 minutes from now
 * * * * *  Closure at: app\Console\Kernel.php:83 ............... Next Due: 48 minutes from now
 * * * * *  Closure at: app\Console\Kernel.php:102 ............ Next Due: 48 minutes from now
 * * * * *  Closure at: app\Console\Kernel.php:123 ............ Next Due: 48 minutes from now
 * * * * *  Closure at: app\Console\Kernel.php:168 ............ Next Due: 48 minutes from now
```

# What changed?
Calling `$schedule->call()` without a description produces the closure location, as before.

If you set a description it will take priority over the location and the location will be pushed to verbose output `-v`
```php
$this->schedule->call(fn () => '')->everyMinute()->name('Foo closure description');
```

`php artisan schedule:list` produces:
```
 * * * * *  Foo closure description ............... Next Due: 48 minutes from now
```

And `php artisan schedule:list -v` output is:
```
 * * * * *  Foo closure description ............... Next Due: 2022-04-08 03:00:00 +00:00
             ⇁ Closure at: app\Console\Kernel.php:32
```

This adds a lot more readbility to the first example:
```
 * * * * *  Foo closure description ............................ Next Due: 48 minutes from now
 * * * * *  Perform foo bar .................................... Next Due: 48 minutes from now
 * * * * *  Perform some foo job................................. Next Due: 48 minutes from now
 * * * * *  Perform some foo job 2 ............................ Next Due: 48 minutes from now
 * * * * *  Perform some foo job 3 ............................ Next Due: 48 minutes from now
 * * * * *  Perform some foo job 4 ............................ Next Due: 48 minutes from now
 * * * * *  Perform some foo job 5 ............................ Next Due: 48 minutes from now
```

# Reason for PR
The expected output should be the most readable form of the commands, that is probably the description if it exists. The verbose output should contain more detailed information that will guide the developer to troubleshooting - that would be the location of the closure.

### DIFF
--- a/src/Illuminate/Console/Scheduling/ScheduleListCommand.php
+++ b/src/Illuminate/Console/Scheduling/ScheduleListCommand.php
@@ -75,6 +75,9 @@ class ScheduleListCommand extends Command
                 if (class_exists($event->description)) {
                     $command = $event->description;
                     $description = '';
+                } else if($event->description) {
+                    $command = $event->description;
+                    $description = 'Closure at: '.$this->getClosureLocation($event);
                 } else {
                     $command = 'Closure at: '.$this->getClosureLocation($event);
                 }
--- a/src/Illuminate/Console/Scheduling/ScheduleListCommand.php
+++ b/src/Illuminate/Console/Scheduling/ScheduleListCommand.php
@@ -75,7 +75,7 @@ class ScheduleListCommand extends Command
                 if (class_exists($event->description)) {
                     $command = $event->description;
                     $description = '';
-                } else if($event->description) {
+                } else if ($event->description) {
                     $command = $event->description;
                     $description = 'Closure at: '.$this->getClosureLocation($event);
                 } else {
--- a/src/Illuminate/Console/Scheduling/ScheduleListCommand.php
+++ b/src/Illuminate/Console/Scheduling/ScheduleListCommand.php
@@ -75,7 +75,7 @@ class ScheduleListCommand extends Command
                 if (class_exists($event->description)) {
                     $command = $event->description;
                     $description = '';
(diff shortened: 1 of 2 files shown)


taylorotwell:
Honestly I feel like the current behavior is more consistent where the description is always shown in verbose mode.

taylorotwell:
Honestly I feel like the current behavior is more consistent where the description is always shown in verbose mode.

ehamrin:
It might be more consistent with other parts of Laravel, but not the scheduling eco system. 

All of the other schedule commands use description as the primary display text:

`php artisan schedule:test` 
`php artisan schedule:run`
`php artisan schedule:work`  

If this PR stays closed, someone should look at changing those commands as well so that is in fact is consistent.

It would be useful to split the `->name()` and `->description()` methods on CallbackEvent so that they don't write to the same internal `$description` prop. That way developers could name the callbacks and have a different description in verbose output

PR #41864, opened 2022-04-07
docs: Fix Php Doc

squiaios:
Fix the php doc on `object` method from object to array.


driesvints:
This isn't correct. The return type is an object because the second param of `json_decode` is set to false.

squiaios:
I agree with you, but finally the method can return objects and array.

`json_decode('{"a": "b"}', false) => object`

`json_decode('[{"a": "b"}]', false) => array of object`

PR #41866, opened 2022-04-07
[9.x] docs: add array on object method

squiaios:
Fix the php doc on object method

json_decode with false on second parameter can return an array of object.


driesvints:
Now this one is indeed correct 👍

driesvints (approved):

PR #41871, opened 2022-04-07
[9.x] Catch permission exception when creating directory

sebdesign:
Flysystem v3 throws an exception when attempting to create a directory that already exists, and the correct permissions cannot be set.

I have added the `['throws' => false]` in my disks on the `config/filesystems.php` file.

For example, running `Storage::makeDirectory('test')` from artisan (console kernel) creates the `test` directory with the console user and appropriate permissions.
Then, running `Storage::makeDirectory('test')` from the web (http kernel), throws an exception because the directory exists, and attemps to change the permissions, while the web user is not the same as the console user, so it fails.

```
League\Flysystem\UnableToSetVisibility Unable to set visibility for file test. 
    vendor/league/flysystem/src/UnableToSetVisibility.php:33 League\Flysystem\UnableToSetVisibility::atLocation
    vendor/league/flysystem/src/Local/LocalFilesystemAdapter.php:422 League\Flysystem\Local\LocalFilesystemAdapter::setPermissions
    vendor/league/flysystem/src/Local/LocalFilesystemAdapter.php:320 League\Flysystem\Local\LocalFilesystemAdapter::createDirectory
    vendor/league/flysystem/src/Filesystem.php:96 League\Flysystem\Filesystem::createDirectory
    vendor/laravel/framework/src/Illuminate/Filesystem/FilesystemAdapter.php:781 Illuminate\Filesystem\FilesystemAdapter::makeDirectory
```

Since Laravel allows to return false instead of throwing exceptions, this exception should be handled.

The same exception is already handled in the `Storage::setVisibility()` method.

Also, it should probably be handled in the `writeStream` and `put` methods?

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemAdapter.php
+++ b/src/Illuminate/Filesystem/FilesystemAdapter.php
@@ -779,7 +779,7 @@ class FilesystemAdapter implements CloudFilesystemContract
     {
         try {
             $this->driver->createDirectory($path);
-        } catch (UnableToCreateDirectory $e) {
+        } catch (UnableToCreateDirectory | UnableToSetVisibility $e) {
             throw_if($this->throwsExceptions(), $e);
 
             return false;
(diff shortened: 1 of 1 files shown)


driesvints:
> Also, it should probably be handled in the writeStream and put methods?

Yes

taylorotwell:
Thanks!

PR #41878, opened 2022-04-08
[9.x] Add last() method to Query Builder

VictorMLima98:
This pull request adds `last()` method to Eloquent Query Builder classes.

People were using `latest()->first()` and I think it could be helpful.

Also, the use of `latest()->first()` is actually a bad practice when you want the last record of a given table, because it could return unexpected results on records created at same timestamp (literally returning the first of the last).

This PR also solves this, adding an optional parameter to `latest()` method called `primaryKey`. When it is set to true, Eloquent will `getKeyName()` before use `created_at`, while default Query Builder will always use `defaultKeyName()`.

It also works for model relationships, of course.

Examples:

image
image
image

### DIFF
--- a/src/Illuminate/Database/Concerns/BuildsQueries.php
+++ b/src/Illuminate/Database/Concerns/BuildsQueries.php
@@ -304,7 +304,7 @@ trait BuildsQueries
      */
     public function last($columns = ['*'])
     {
-        return $this->latest(primaryKey: true)->take(1)->get($columns)->first();
+        return $this->latest(defaultKey: true)->take(1)->get($columns)->first();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


huangdijia:
`User::latest()->first()` or `DB::table('users')->latest()->first()` ??

rodrigopedra:
Overall I think this is a good addition, but adding a parameter to a public function is a breaking change, so I guess you should target this PR to 10.x.

driesvints:
This indeed will need to target master

VictorMLima98:
Oh, okay... Thank you guys! I'll target this PR to 10.x and then re-open it.

PR #41881, opened 2022-04-08
[9.x] Fix bug in BelongsToMany where non-related rows are returned

marcovo:
While working on a [new package](https://gitlab.com/marcovo/laravel-dag-model) for laravel, I encountered a failing test on laravel 9.x with no apparent reason. After digging a bit deeper, I came to understand that `firstOrNew()`, `firstOrCreate()` and `updateOrCreate()` were not doing what I expected.

**Expected behaviour**
These three methods only return or affect related rows that are in fact related to the current model.

**Actual behaviour**
The `firstOr*` methods will return any row in the related table, not subject to the relation constraint. Likewise, the `updateOrCreate` method updates any one row in the related table, not subject to the relation constraint. In all cases, any passed `$attributes` are correctly used, it is the relation constraint that is missing.

**Fix**
In commit 3e66eb75ae379f14eb760a19071a07134de797e3 a change in laravel 9 was introduced where the lookup would be done using `$this->related` instead of `$this`. This PR changes this to `$this->clone()`. Three tests are added which fail on the current 9.x version, and are fixed using this PR.

Targetting 9.x because 8.x does not use the `$this->related` approach.

**Considerations**
In commit 3e66eb75ae379f14eb760a19071a07134de797e3, no tests are added backing the change to use `$this->related` instead of `$this`. I can imagine it has something to do with not wanting to pollute the current query builder with the additional `where()` method? In that case I'm confident this PR is still compatible, otherwise you might want to double check.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php
@@ -600,7 +600,7 @@ class BelongsToMany extends Relation
      */
     public function firstOrNew(array $attributes = [], array $values = [])
     {
-        if (is_null($instance = $this->related->where($attributes)->first())) {
+        if (is_null($instance = $this->clone()->where($attributes)->first())) {
             $instance = $this->related->newInstance(array_merge($attributes, $values));
         }
 
@@ -618,7 +618,7 @@ class BelongsToMany extends Relation
      */
     public function firstOrCreate(array $attributes = [], array $values = [], array $joining = [], $touch = true)
     {
-        if (is_null($instance = $this->related->where($attributes)->first())) {
+        if (is_null($instance = $this->clone()->where($attributes)->first())) {
             $instance = $this->create(array_merge($attributes, $values), $joining, $touch);
         }
 
@@ -636,7 +636,7 @@ class BelongsToMany extends Relation
      */
     public function updateOrCreate(array $attributes, array $values = [], array $joining = [], $touch = true)
     {
-        if (is_null($instance = $this->related->where($attributes)->first())) {
+        if (is_null($instance = $this->clone()->where($attributes)->first())) {
             return $this->create(array_merge($attributes, $values), $joining, $touch);
         }
 
--- a/tests/Integration/Database/EloquentBelongsToManyTest.php
+++ b/tests/Integration/Database/EloquentBelongsToManyTest.php
@@ -427,6 +427,18 @@ class EloquentBelongsToManyTest extends DatabaseTestCase
         $this->assertInstanceOf(Tag::class, $post->tags()->firstOrNew(['id' => 666]));
     }
 
+    public function testFirstOrNewWithoutExistingRelated()
+    {
+        $post = Post::create(['title' => Str::random()]);
+
+        $name = Str::random();
+        $tag = Tag::create(['name' => $name]);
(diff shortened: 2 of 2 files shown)


taylorotwell:
I think this would be a breaking change now. It basically reverts it back to the original behavior that some others were complaining about. Typically, you don't want duplicate models in a many to many relationship. For example, you want one tag named "Technology" and you want to share that same Tag record across all tagged Posts with that Tag.

I don't think it's possible to please everyone with these methods. If they aren't behavior how you want them to, I suggest just making more manual queries to check for the tags existence yourself, etc.

marcovo:
@taylorotwell Do you have a reference to the complaints? I could not find any in the related commit.

Your observation sounds solid, but also in that case I think the implementation is misleading. Consider that any method on the `BelongsToMany` relationship returning a model, returns a model that is related to the model operated upon. Except that the two `firstOr*` methods may return 'related' models that are not related at all, and may not be related throughout the entire lifetime of the request.

My understanding of `firstOrCreate` is that you will always receive a related model instance that already was related ('first') or is newly related ('create') to the model operated upon. In case a new related record is created, `attach()` is explicitly called. Would hence something along these lines not be better? (Rough outline, may be missing some details)
```
    public function firstOrCreate(array $attributes = [], array $values = [], array $joining = [], $touch = true)
    {
        if (!is_null($instance = $this->clone()->where($attributes)->first())) {
            return $instance;
        } elseif (!is_null($instance = $this->related->where($attributes)->first())) {
            $this->attach($instance); // Make sure this model is related
            return $instance;
        }

        return $this->create(array_merge($attributes, $values), $joining, $touch);
    }
```

Similar arguments hold for `firstOrNew`, but you cannot automatically attach in that case due to the nature of the method. What you have here is really a kind of undefined behaviour, since a model returned can either be:
 1. A genuine related record
 2. A record from the related table, that is present (so `exists = true`), but is not related
 3. A new instance (`exists = false`)
While in the old situation you can distinguish between the two possible situations, you cannot distinguish cases 1. and 2. here. To me it seems that in a fetching setting this would render this method useless (you cannot know its exact state, right?), and in a mutating setting (in which you want to store the related model) this would imply that you would always have to `attach()` the related model as you cannot tell whether it is already attached or not?
Also consider that `findOrNew()` does only return related records, while `firstOrNew()` doesn't.

`updateOrCreate` is different story. Consider that `$model->relation()->update()` will always only update rows that are related. Changing it into `$model->relation()->updateOrCreate()` introduces the possibility that (without clear indication) a row is being updated that is not related to the model at all. One could argue that also in this case (similar to `firstOrCreate()`), the implementation should automatically `attach()` a related model if it is not related already.


I am not entirely convinced that the `->attach()` approach in above rough outline would be the best approach. If I understand the past issue correctly, maybe a better alternative would have been to leave these three methods as they originally were, and introduce new methods `firstOrCreateAndAttach()` and `updateOrCreateAndAttach()`. But probably that is a bit of a hindsight.


Long story short, I am concerned that these three methods violate the invariant that all similar BelongsToMany methods always act on related data, and these three now do not behave in this same way. I can see what the intention behind the change might have been, but maybe that should have required a different solution? Currently I feel the behaviour can be rather unexpected, especially in some edge cases that likely will occur in applications.

I would love to hear what others think of this issue.

PR #41889, opened 2022-04-08
[10.x] Eloquent & Query Builder last() methods

VictorMLima98:
Now it is in the right branch, I expect! 😄 

#### This pull request adds `last()` method to Eloquent & Query Builder.

People were using `latest()->first()` and I think it could be helpful.

Also, the use of `latest()->first()` is actually a bad practice when you want the last record of a given table, because it could return unexpected results on records created at same timestamp (literally returning the first of the last).

#### This PR also solves this, adding an optional parameter to `latest()` method called `primaryKey`. When it is set to true, Eloquent will `getKeyName()` instead of use `created_at`, while default Query Builder will always use `defaultKeyName()`.

It also works for model relationships, of course.

Examples:

image
image
image

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -346,15 +346,15 @@ class Builder implements BuilderContract
     }
 
     /**
-     * Add an "order by" clause for a timestamp to the query.
+     * Add an "order by" clause for a timestamp or default key to the query.
      *
      * @param  string|\Illuminate\Database\Query\Expression  $column
      * @return $this
      */
-    public function latest($column = null)
+    public function latest($column = null, $defaultKey = false)
     {
         if (is_null($column)) {
-            $column = $this->model->getCreatedAtColumn() ?? 'created_at';
+            $column = $defaultKey ? $this->model->defaultKeyName() : $this->model->getCreatedAtColumn() ?? 'created_at';
         }
 
         $this->query->latest($column);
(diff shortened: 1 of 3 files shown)


src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -346,15 +346,15 @@ public function orWhereNot($column, $operator = null, $value = null)
...
 
     /**
-     * Add an "order by" clause for a timestamp to the query.
+     * Add an "order by" clause for a timestamp or default key to the query.
      *
      * @param  string|\Illuminate\Database\Query\Expression  $column
      * @return $this
      */
-    public function latest($column = null)
+    public function latest($column = null, $defaultKey = false)
```
GrahamCampbell, line 354: needs phpdoc

src/Illuminate/Database/Query/Builder.php
```diff
@@ -2192,14 +2192,14 @@ public function orderByDesc($column)
...
 
     /**
-     * Add an "order by" clause for a timestamp to the query.
+     * Add an "order by" clause for a timestamp or default key to the query.
      *
      * @param  \Closure|\Illuminate\Database\Query\Builder|\Illuminate\Database\Query\Expression|string  $column
      * @return $this
      */
-    public function latest($column = 'created_at')
+    public function latest($column = 'created_at', $defaultKey = false)
```
GrahamCampbell, line 2200: needs phpdoc

GrahamCampbell (changes requested):

taylorotwell:

PR #41891, opened 2022-04-08
[9.x] Removes redundant uncountable words and change language for pluralizer ruleset

cesarep:
Hi!

As seen in https://github.com/doctrine/inflector/blob/2.0.x/lib/Doctrine/Inflector/RulesetInflector.php#L37
when inflecting a word, either pluralizing or singularizing, Inflector calls $ruleset->getUninflected()->matches($word)
to check if the inflected word is in https://github.com/doctrine/inflector/blob/2.0.x/lib/Doctrine/Inflector/Rules/English/Uninflected.php
or the corresponding language Uninflected file, if it is, it returns the very same word, unchanged.

There are only 7 words not directly listed there, those beeing: `'cattle', 'firmware', 'hardware', 'kin', 'recommended', 'related', 'software'`, from those, `yield new Pattern('\w+ware$')` on [line 42](https://github.com/doctrine/inflector/blob/2.0.x/lib/Doctrine/Inflector/Rules/English/Uninflected.php#L42) matches firm... hard... and soft...ware.

I already created a pull request to add those there https://github.com/doctrine/inflector/pull/194

There's really no reason to keep them. New words should be requested on that Uninflected file, for all the available languages.

Now, as for the language ruleset, I'm not really sure if it's allowed to call `Container::getInstance()` inside Illuminate/Support, the only other file that does this is [Traits/Localizable.php](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/Traits/Localizable.php) but it works fine locally on my machine.

With that, we can add a 'pluralizer_ruleset' config on app.php to choose between `'english', 'french', 'norwegian-bokmal', 'portuguese', 'spanish', 'turkish'`

I believe this change is backward compatible with older versions, i tested on my 8.x version

### DIFF
--- a/src/Illuminate/Support/Pluralizer.php
+++ b/src/Illuminate/Support/Pluralizer.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Support;
 
+use Illuminate\Container\Container;
 use Doctrine\Inflector\InflectorFactory;
 
 class Pluralizer
@@ -77,7 +78,8 @@ class Pluralizer
         static $inflector;
 
         if (is_null($inflector)) {
-            $inflector = InflectorFactory::createForLanguage('english')->build();
+            $app = Container::getInstance();
+            $inflector = InflectorFactory::createForLanguage( $app['config']->get('app.pluralizer_ruleset', 'english') )->build();
         }
 
         return $inflector;
--- a/src/Illuminate/Support/Pluralizer.php
+++ b/src/Illuminate/Support/Pluralizer.php
@@ -9,53 +9,8 @@ class Pluralizer
     /**
      * Uncountable word forms.
      *
-     * @var string[]
+     * New uncuntable words should be added to: https://github.com/doctrine/inflector/blob/2.0.x/lib/Doctrine/Inflector/Rules/English/Uninflected.php
      */
-    public static $uncountable = [
-        'audio',
-        'bison',
-        'cattle',
-        'chassis',
-        'compensation',
-        'coreopsis',
-        'data',
-        'deer',
-        'education',
-        'emoji',
-        'equipment',
-        'evidence',
-        'feedback',
-        'firmware',
-        'fish',
-        'furniture',
-        'gold',
-        'hardware',
-        'information',
-        'jedi',
-        'kin',
-        'knowledge',
-        'love',
-        'metadata',
-        'money',
-        'moose',
-        'news',
-        'nutrition',
-        'offspring',
-        'plankton',
-        'pokemon',
-        'police',
-        'rain',
-        'recommended',
-        'related',
-        'rice',
-        'series',
-        'sheep',
-        'software',
-        'species',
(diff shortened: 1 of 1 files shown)


willrowe:
May want to spellcheck the PR name...

cesarep:
Oh my... Well the checks sure don't liked it

driesvints:
Given the failing tests, I don't think we should go ahead with this.

cesarep:
How can i try these tests locally?
When i run phpunit on my console i get
`PHP Fatal error:  Trait 'Illuminate\Auth\Access\HandlesAuthorization' not found in 'tests/Auth/AuthAccessGateTest.php' on line 1203`

src/Illuminate/Support/Pluralizer.php
```diff
@@ -2,60 +2,16 @@
...
+use Illuminate\Container\Container;
 use Doctrine\Inflector\InflectorFactory;
 
 class Pluralizer
 {
     /**
      * Uncountable word forms.
      *
-     * @var string[]
+     * New uncuntable words should be added to: https://github.com/doctrine/inflector/blob/2.0.x/lib/Doctrine/Inflector/Rules/English/Uninflected.php
```
liamkeily, line 13:
```suggestion
     * New uncountable words should be added to: https://github.com/doctrine/inflector/blob/2.0.x/lib/Doctrine/Inflector/Rules/English/Uninflected.php
```

driesvints:
I'm going to close this one for now. You're free to attempt again with passing tests but honestly I think we should not go ahead with this. Thanks

src/Illuminate/Support/Pluralizer.php
cesarep: Actually, as said https://github.com/doctrine/inflector/pull/194#discussion_r846107000, "related" and "recommended" are not nouns, perhaps the uncountable array and function should remain, but just for these edge cases, not nouns, nor words that already exist on the Uninflected file, with the recommendation to add new nouns to the inflector repository

driesvints:
Not sure why it doesn't work for you sorry

cesarep:
Hi!
So, i managed to get the tests working! I changed `SupportPluralizerTest.php` itself adding the `getEnvironmentSetUp($app)` function, that was present on all the other tests on Classes that depend on configs, setting the required config there. But ultimately (and it feels like cheating), I checked if the Container instance had the config class with `$app->has('config')`, setting the language to `'english'` if it didn't, effectively bypassing the `Target class [config] does not exist.` error. Is it allowed tho?

cesarep:
Hey @driesvints!

Should a try a new PR now with the tests working? https://github.com/cesarep/framework/commit/74639ebe0e92c3c3cff0f43424f19e5fef4b74a1
Or this is not how nor where `$app->has()` should be used? Because i haven't seen it like that anywhere else, but the tests didn't complaint

PR #41898, opened 2022-04-09
Add multibyte support to string padding helper functions

clarkewing:
This PR adds support for multibyte strings to the `Str::padBoth`, `Str::padLeft`, and `Str::padRight` helper functions.

This is especially useful for localized strings, where accented characters or other multibyte glyphs were incorrectly padded before.

PHP has yet to implement a native `mb_str_pad` function so this PR adds a `Str::mbStrPad` helper function adapted from [a comment on the PHP docs](https://www.php.net/manual/en/ref.mbstring.php#90611). There is an [ongoing 19 year-old PHP request](http://bugs.php.net/bug.php?id=21317) for this native function, but as it is yet to see any progress, it seems appropriate to implement a workaround in the framework.


clarkewing:
Closing because I messed up my commit

PR #41899, opened 2022-04-09
[8.x] Add multibyte support to string padding helper functions

clarkewing:
This PR adds support for multibyte strings to the `Str::padBoth`, `Str::padLeft`, and `Str::padRight` helper functions.

This is especially useful for localized strings, where accented characters or other multibyte glyphs were incorrectly padded before.

PHP has yet to implement a native `mb_str_pad` function so this PR adds a `Str::mbStrPad` helper function adapted from [a comment on the PHP docs](https://www.php.net/manual/en/ref.mbstring.php#90611). There is an [ongoing 19 year-old PHP request](http://bugs.php.net/bug.php?id=21317) for this native function, but as it is yet to see any progress, it seems appropriate to implement a workaround in the framework.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -489,7 +489,7 @@ class Str
      */
     public static function padBoth($value, $length, $pad = ' ')
     {
-        return static::mbStrPad($value, $length, $pad, STR_PAD_BOTH);
+        return str_pad($value, strlen($value) - mb_strlen($value) + $length, $pad, STR_PAD_BOTH);
     }
 
     /**
@@ -502,7 +502,7 @@ class Str
      */
     public static function padLeft($value, $length, $pad = ' ')
     {
-        return static::mbStrPad($value, $length, $pad, STR_PAD_LEFT);
+        return str_pad($value, strlen($value) - mb_strlen($value) + $length, $pad, STR_PAD_LEFT);
     }
 
     /**
@@ -515,7 +515,7 @@ class Str
      */
     public static function padRight($value, $length, $pad = ' ')
     {
-        return static::mbStrPad($value, $length, $pad, STR_PAD_RIGHT);
+        return str_pad($value, strlen($value) - mb_strlen($value) + $length, $pad, STR_PAD_RIGHT);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
Laravel 8 is closed to new features. This PR adds a new non-private method.

clarkewing:
Thanks for the quick reply @GrahamCampbell! Any chance this could be considered for Laravel 9?

clarkewing:
Thanks for the speedy reply @GrahamCampbell! I made the `Str::mbStrPad` method protected. As it's my first time contributing to the framework, I'm open to any feedback.

BrandonSurowiec:
It would probably be better to simply inline the change in every method instead of adding a protected helper. There are no separate "helper" method for the other string methods that have multibyte support.

**Only the second parameter is modified :**

```php
    public static function padRight($value, $length, $pad = ' ')
    {
        return str_pad($value, strlen($value) - mb_strlen($value, $encoding) + $length, $pad, STR_PAD_RIGHT);
    }
```

clarkewing:
Thanks for the helpful feedback @BrandonSurowiec! I committed a fix 😊

BrandonSurowiec:
Looks good to me. I think it's ready to go :)

src/Illuminate/Support/Str.php
```diff
@@ -1034,7 +1034,7 @@ public static function flushCache()
      * @param  string|null  $encoding
      * @return string
      */
-    public static function mbStrPad($value, $length, $pad = ' ', $pad_type = STR_PAD_RIGHT, $encoding = null)
+    protected static function mbStrPad($value, $length, $pad = ' ', $pad_type = STR_PAD_RIGHT, $encoding = null)
```
GrahamCampbell: protected is still a new addition to the non-private API

GrahamCampbell: this whole PR is really a new feature anyway, so should target 9.x.

GrahamCampbell (changes requested):

src/Illuminate/Support/Str.php
BrandonSurowiec: Seems you're looking at an older version? He inlined that method after my comment.

clarkewing: I'm marking this as resolved since the function was previously inlined. Don't hesitate to reopen it if need be.

PR #41903, opened 2022-04-09
[10.x] leverage native php 8.1 array_is_list

hAz5:
check array is list 

from:
```
array_values($value->data) === $value->data
```
to:
```
array_is_list($value->data)
```
https://php.watch/versions/8.1/array_is_list


taylorotwell:
If the code is already working I don't see a big reason to change it.

PR #41904, opened 2022-04-09
[9.x] Refactor switch to match

masouddarvishi:
refactor switch to match. makes code cleaner and simpler.


taylorotwell:
If this code is working correctly we don't need to change it.

PR #41907, opened 2022-04-10
[9.x] Facade accessor auto resolve

ahmedash95:
Hello, 

I've hit this many times in my personal projects and also have seen it in the core and common packages. for most Facades, the accessor name is the same as the class name. it would be good if the facade can resolve that automatically instead of defining the `getFacadeAccessor` just to return the class name in lower case. also, detecting from the class name is very common in laravel as it's used in Factory to detect the model name. and in Eloquent to detect the table name.

Before
```php
class Hash extends Facade
{
    /**
     * Get the registered name of the component.
     *
     * @return string
     */
    protected static function getFacadeAccessor()
    {
        return 'hash';
    }
}
```

After
```php
class Hash extends Facade
{
    //
}
```

### DIFF
--- a/src/Illuminate/Support/Facades/App.php
+++ b/src/Illuminate/Support/Facades/App.php
@@ -52,13 +52,5 @@ namespace Illuminate\Support\Facades;
  */
 class App extends Facade
 {
-    /**
-     * Get the registered name of the component.
-     *
-     * @return string
-     */
-    protected static function getFacadeAccessor()
-    {
-        return 'app';
-    }
+    //
 }
--- a/src/Illuminate/Support/Facades/Auth.php
+++ b/src/Illuminate/Support/Facades/Auth.php
@@ -37,16 +37,6 @@ use RuntimeException;
 class Auth extends Facade
 {
     /**
-     * Get the registered name of the component.
-     *
-     * @return string
-     */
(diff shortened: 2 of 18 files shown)


Kocal:
What about performances?

ahmedash95:
> What about performances? The facade accessor name is not supposed to change while running the code, so we can keep it hardcoded instead of calling 3 functions each time we call `Facade::getFacadeAccessor()` no?

you are correct. the performance on 1M call is 2x with dynamic resolving. I've added a `static:::$resolvedNames` to cache resolved names. and the performance now is the same.

driesvints:
I don't think letting class names reference the string alias is a wise thing to do. Right now they're uncoupled, meaning they can be easily changes. But after this PR they're hardwired.

ahmedash95:
> I don't think letting class names reference the string alias is a wise thing to do. Right now they're uncoupled, meaning they can be easily changes. But after this PR they're hardwired.

There is still a possibility to keep it uncoupled by just overriding the `getFacadeAccessor`. the purpose of my PR is that it makes the Facade class cleaner. which is something I see it would be good to have like getting table names from Models without having to define them manually.

taylorotwell:

PR #41908, opened 2022-04-10
[9.x] Add return type declarations to migration stubs

andrejsmelnikovs:
PHP 7 added support for return type declarations, but Laravel stubs still don't use them. 

Return type declarations specify the type of the value that will be returned from a function. The returned value must be of the correct type, otherwise a TypeError will be thrown.

I think this is necessary change in all other Laravel stubs as well.

### DIFF
--- a/src/Illuminate/Database/Migrations/stubs/migration.create.stub
+++ b/src/Illuminate/Database/Migrations/stubs/migration.create.stub
@@ -11,7 +11,7 @@ return new class extends Migration
      *
      * @return void
      */
-    public function up()
+    public function up(): void
     {
         Schema::create('{{ table }}', function (Blueprint $table) {
             $table->id();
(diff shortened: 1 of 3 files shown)


src/Illuminate/Database/Migrations/stubs/migration.create.stub
```diff
@@ -11,7 +11,7 @@ return new class extends Migration
      *
      * @return void
      */
-    public function up()
+    public function up(): void
```
dammy001, line 14: There's doctype already @andrejsmelnikovs

andrejsmelnikovs: Doctype doesn't throw an error, if returned value is not of the correct type.

driesvints:
Please see many previous attempts

src/Illuminate/Database/Migrations/stubs/migration.create.stub
GrahamCampbell: That's what a static analyzer is for. :)

PR #41911, opened 2022-04-10
[9.x] Add AWS S3 delete multiple files using deleteObjects method

aneeskhan47:
Hi,

I needed to delete 100+ files on AWS s3. the `Storage::disk('s3')->delete()` delete a single file so if i loop through the 100+ files and do the `Storage::disk('s3')->delete($file)` it will do 100+ request to s3 for 100+ files.

So I searched and I find out that AWS s3 allows you to delete multiple files using the AWS client `deleteObjects` method which accepts an array of object paths e.g `['path/to/file1', 'path/to/file2']`. using this it just took 2 secs to delete 100+ files. 

more info: https://docs.aws.amazon.com/AmazonS3/latest/userguide/example_s3_DeleteObjects_section.html

I first implemented this as a macro on the Storage facade in my projects. but I think this should be in laravel by default :)

Thanks,
Anees

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemAdapter.php
+++ b/src/Illuminate/Filesystem/FilesystemAdapter.php
@@ -91,8 +91,7 @@ class FilesystemAdapter implements CloudFilesystemContract
         $this->config = $config;
 
         $this->prefixer = new PathPrefixer(
-            $config['root'] ?? '',
-            $config['directory_separator'] ?? DIRECTORY_SEPARATOR
+            $config['root'] ?? '', $config['directory_separator'] ?? DIRECTORY_SEPARATOR
         );
     }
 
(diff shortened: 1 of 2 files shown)


ankurk91:
I think this should added here instead
https://github.com/thephpleague/flysystem-aws-s3-v3

src/Illuminate/Filesystem/FilesystemAdapter.php
```diff
@@ -110,7 +111,8 @@ public function assertExists($path, $content = null)
 
         foreach ($paths as $path) {
             PHPUnit::assertTrue(
-                $this->exists($path), "Unable to find a file or directory at path [{$path}]."
+                $this->exists($path),
+                "Unable to find a file or directory at path [{$path}]."
```
driesvints, line 115: Can you please undo all of these formatting changes?

driesvints (changes requested):

src/Illuminate/Filesystem/FilesystemAdapter.php
aneeskhan47: Sorry about that. it was my php-cs-fixer doing this. I undid all of the formatting changes 👍

ankurk91:
We should wrap the api call in try/catch and all
`throw_if($this->throwsExceptions(), $e);`

PR #41914, opened 2022-04-11
[9.x] Standardize `withCount()` & `withExists()` eager loading aggregates.

r-kujawa:
## What does this PR do?

- Standardizes the way you can eager load your `withCount()` & `withExists()` aggregates.
- Adds asserts to existing tests to support this functionality☝🏼.

### Before this PR
There are 3 different ways to pass parameters in `with...()` functions:
1. Passing relation strings (internally will check `is_array($relations) ? $relations : func_get_args()`).
2. Passing an array of relations (with or without a closure as the value).
3. Passing the first parameter as a relation string and second argument as a closure.

If we compare the `with()`, `withCount()` & `withExists` functions, this is what each of them currently support:
- `with()`: 1, 2 & 3.
- `withCount()`: 1 & 2.
- `withExists()`: 1 (with only 1 relation 😞) & 2.

The reason why I am writing this PR is because I ran into a situation where I supposed that since I was able to do it with one of these methods, why shouldn't I be able to do the same with any of the other 2 methods (that's where I got disappointed 😞).

I did notice that it would probably be best practice and most efficient to have way 1 & 3 removed and only support way 2. But that would result in some breaking changes (LMK if it sounds interesting and I will make a PR for 10.x).

### With this PR
The 3 functions mentioned above will support all 3 different ways to pass the params.
🔴 The only caveat I might see is that the `with()` function includes a second parameter in it's signature (`@param  string|\Closure|null  $callback`) while the other 2 don't.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -622,7 +622,13 @@ trait QueriesRelationships
      */
     public function withCount($relations)
     {
-        return $this->withAggregate(is_array($relations) ? $relations : func_get_args(), '*', 'count');
+        $relations = is_array($relations) ? $relations : func_get_args();
+
+        if (count($relations) === 2 && $relations[1] instanceof Closure) {
+            $relations = [$relations[0] => $relations[1]];
+        }
+
+        return $this->withAggregate($relations, '*', 'count');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


nuernbergerA:
@taylorotwell This is actually a breaking change because the [example from the docs](https://laravel.com/docs/9.x/eloquent-relationships#counting-related-models) no longer works now

```php
 
$posts = Post::withCount(['votes', 'comments' => function (Builder $query) {
    $query->where('content', 'like', 'code%');
}])->get();
 
echo $posts[0]->votes_count;
echo $posts[0]->comments_count;
```

will throw `Undefined array key 1`

PR #41918, opened 2022-04-11
[9.x] Fix base query for relationships

### LINKED ISSUE #41885: Calling toBase() on relationships removes soft-deleted condition in Laravel 9 compared to Laravel 8
- Laravel Version: 9.7.0
- PHP Version: 8.1.1

### Description:
In Laravel 9 the `toBase()` function behaves differently than in Laravel 9 when used on a relationship: The `deleted_at IS NULL` condition of the `SoftDeletes` scope was still applied to relationship queries in Laravel 8 when calling `toBase()`, but in Laravel 9 the condition is missing. However calling `toBase()` on a normal Eloquent query still retains the `deleted_at IS NULL` condition of the `SoftDeletes` trait in Laravel 9.

### Steps To Reproduce:

```php
use Illuminate\Database\Eloquent\Relations\HasMany;
use Illuminate\Database\Eloquent\SoftDeletes;

class ModelA extends Model
{
    use SoftDeletes;
    
    public function modelsB(): HasMany
    {
        return $this->hasMany(ModelB::class);
    }
}

class ModelB extends Model
{
    use SoftDeletes;
}

Model::unguard();
$modelA = new ModelA(['id' => 1]);

// Laravel 8: [...] where [...] and "model_b_s"."model_a_id" is not null and "model_b_s"."deleted_at" is null
// Laravel 9: [...] where [...] and "model_b_s"."model_a_id" is not null and "model_b_s"."deleted_at" is null
$modelA->modelsB()->dump();

// Laravel 8: [...] where [...] and "model_b_s"."model_a_id" is not null and "model_b_s"."deleted_at" is null
// Laravel 9: [...] where [...] and "model_b_s"."model_a_id" is not null and "model_b_s"."deleted_at" is null
$modelA->modelsB()->getQuery()->dump();

// Laravel 8: [...] where [...] and "model_b_s"."model_a_id" is not null and "model_b_s"."deleted_at" is null
// Laravel 9: [...] where [...] and "model_b_s"."model_a_id" is not null
$modelA->modelsB()->toBase()->dump();

// Laravel 8: [...] where "model_a_s"."deleted_at" is null
// Laravel 9: [...] where "model_a_s"."deleted_at" is null
$modelA->newQuery()->toBase()->dump();
```

driesvints:
This PR fixes an issue where scopes (like soft deletes) weren't applied when the base query of a relationship was retrieved. Previously, in Laravel v8, this was still applied properly.

Fixes https://github.com/laravel/framework/issues/41885

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Relation.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Relation.php
@@ -317,7 +317,7 @@ abstract class Relation implements BuilderContract
      */
     public function toBase()
     {
-        return $this->query->getQuery();
+        return $this->query->applyScopes()->getQuery();
     }
 
     /**
--- a/tests/Database/DatabaseEloquentSoftDeletesIntegrationTest.php
+++ b/tests/Database/DatabaseEloquentSoftDeletesIntegrationTest.php
@@ -123,6 +123,19 @@ class DatabaseEloquentSoftDeletesIntegrationTest extends TestCase
         $this->assertCount(1, $query->get());
     }
 
+    public function testSoftDeletesAreNotRetrievedFromRelationshipBaseQuery()
+    {
+        [, $abigail] = $this->createUsers();
+
+        $abigail->posts()->create(['title' => 'Foo']);
+        $abigail->posts()->create(['title' => 'Bar'])->delete();
+
+        $query = $abigail->posts()->toBase();
+
+        $this->assertInstanceOf(Builder::class, $query);
+        $this->assertCount(1, $query->get());
+    }
+
     public function testSoftDeletesAreNotRetrievedFromBuilderHelpers()
     {
         $this->createUsers();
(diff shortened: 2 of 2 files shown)


taylorotwell:
We need to dig into why this method was changed for Laravel 9.x - there was likely a reason.

tpetry:
The `toBase` method was not implemented in `Relation` for Laravel 8. As of the query builder's mixin and forwarding principle the method call was forwarded by `__call()` to `Eloquent\Builder::toBase()` which is executing `$this->applyScopes()->getQuery();` - the same as this PR.

So, my guess is that this is not an intended change but simply an oversight in the implementation because with the query builder interface the `toBase` method needed to be implemented.

driesvints:
@tpetry ah right, can you send in a PR for that?

tpetry:
Give me some minutes 😉

PR #32770, opened 2020-05-12
[7.x] Allow setting the event handler queue via a getQueue() method

lupinitylabs:
Using queued event listeners, it is possible to set the queue the resulting job should be executed on using the `$queue` attribute. This attribute is then adopted by the dispatcher. 

However, it is not possible to set the queue name from the config, the environment or otherwise programatically, because the listener instance is created without invoking the constructor in `Dispatcher::createListenerAndJob()`. Thus, trying to set the queue name in the constructor or anywhere else in the Listener has no effect.

This pull request proposes a non-breaking change to the Dispatcher class that allows a user to set the queue name by defining a `getQueue()` method within the Listener, as it is already done with `retryAfter` and `retryUntil` methods. 

If no getQueue method is defined, the code falls back to use the `$queue` attribute or null, as it was before.

The benefit of this change is to provide developers with a flexible way to influence the target queues instead of having to hard code them.

PR #41922, opened 2022-04-11
[9.x] Allow Setting Delay Via withDealy Method In Queued Events

haider00125:
This PR will fix setting $delay property in queued handlers with Queueable trait. Unable to set delay property when using Queueable trait in queued event handler. 
Similar issue was fixed in [this](https://github.com/laravel/framework/pull/32770) for queue property.


taylorotwell:

PR #41925, opened 2022-04-11
[9.x] Allow to use custom log level in exception handler reporting

Gummibeer:
This PR adds a new property `$reportLogLevels` and corresponding method `useLogLevel()` to the exception handler. These allow the user to configure different log levels for any exception. This will make it easier to report `PDOException`s as _critical_ and known/common user exceptions as _debug_ or whatever you want.

solves #41921

### DIFF
--- a/src/Illuminate/Foundation/Exceptions/Handler.php
+++ b/src/Illuminate/Foundation/Exceptions/Handler.php
@@ -28,6 +28,7 @@ use Illuminate\Support\ViewErrorBag;
 use Illuminate\Validation\ValidationException;
 use InvalidArgumentException;
 use Psr\Log\LoggerInterface;
+use Psr\Log\LogLevel;
 use Symfony\Component\Console\Application as ConsoleApplication;
 use Symfony\Component\ErrorHandler\ErrorRenderer\HtmlErrorRenderer;
 use Symfony\Component\HttpFoundation\Exception\SuspiciousOperationException;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Renamed `$reportLogLevels` to `$levels` and `useLogLevel` to `level`.

Nice feature - thanks!

Gummibeer:
Thanks @taylorotwell ! 🥳
Was pretty sure that the naming wasn't the perfect Laravel-way - but wanted to still provide the PR. 🙂

src/Illuminate/Foundation/Exceptions/Handler.php
```diff
@@ -64,6 +65,13 @@ class Handler implements ExceptionHandlerContract
      */
     protected $reportCallbacks = [];
 
+    /**
+     * A map of exceptions with their corresponding custom log levels.
+     *
+     * @var array<string, string>
```
GrahamCampbell, line 71: this should be `array<class-string<\Throwable>, LogLevel::*>`, really.

src/Illuminate/Foundation/Exceptions/Handler.php
```diff
@@ -208,6 +216,20 @@ public function ignore(string $class)
         return $this;
     }
 
+    /**
+     * Set the log level for the given exception type.
+     *
+     * @param  class-string<\Throwable>  $type
+     * @param  string  $level
```
GrahamCampbell, line 223: `string` could be replaced by `LogLevel::*`

Gummibeer: #41935

src/Illuminate/Foundation/Exceptions/Handler.php
```diff
@@ -241,7 +263,12 @@ public function report(Throwable $e)
             throw $e;
         }
 
-        $logger->error(
+        $level = Arr::first($this->levels, function ($level, $type) use ($e) {
+            return $e instanceof $type;
+        }, LogLevel::ERROR);
+
+        $logger->log(
+            $level,
```
michael-rubel, line 271: @taylorotwell this is a breaking change for tests. 😕

PR #41926, opened 2022-04-11
[9.x] Support nested relationships in whenLoaded

krenor:
I believe the reason this wasn't already implemented is because it only makes sense for relationships that return a Model instead of a Collection. But in the first case, e.g. `Post::with(['user.profile])`, I figured it would be handy. Unless there's a relation returning a collection in between, e.g. `Post::with(['comments.user'])`,  it would still work for a deeply nested Collection relationship value.

### DIFF
--- a/src/Illuminate/Http/Resources/ConditionallyLoadsAttributes.php
+++ b/src/Illuminate/Http/Resources/ConditionallyLoadsAttributes.php
@@ -189,16 +189,23 @@ trait ConditionallyLoadsAttributes
             $default = new MissingValue;
         }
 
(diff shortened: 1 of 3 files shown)


taylorotwell:

kingyyy:
@taylorotwell I think this feature will be helpful to the users by extending whenLoaded method for nest relations
like this
```
'directorate' => $this->whenLoaded("directorate", $this->directorate->name),
'governorate' => $this->whenLoaded("directorate.governorate", $this->directorate->governorate->name),
```

PR #41927, opened 2022-04-11
[9.x] Add implicit CSRF tokens

andriihorpenko:
This PR adds implicit CSRF tokens compilation by the Blade compiler. This allows developers to not include `@csrf` helper to every single form on the page. Additionally, this feature can be disabled on per-form basis by adding the provided `@nocsrf` helper to the form. This can be helpful if you are dealing with merchant's payment forms, which have nothing to do with the project's CSRF token.

This PR can be potentially improved. For instance, don't add duplicate `@csrf` if one already exists in the form. This was not implemented in order to keep code simple, and also because the latter duplicate input field overrides the former one, so there are no actions required.

## Examples

**Implicit CSRF token compilation:**
```html
<!-- Before -->
<form method="POST">
    <input type="text">
</form>

<!-- After (not compiled) -->
<form method="POST">
    @csrf
    <input type="text">
</form>
```

**Forbidding the compiler to add CSRF token:**
```html
<!-- Before -->
<form method="POST">
    @nocsrf
    <input type="text">
</form>

<!-- After -->
<form method="POST">
    <input type="text">
</form>
```


taylorotwell:
Can't do this on a patch release. Might be a good package or something.

PR #41939, opened 2022-04-12
[9.x] 'current_password' validation rule will use the guard UserProvider

shadoWalker89:
When an application uses a different logic then hashing for storing the passwords for example when upgrading a legacy app to laravel and the two apps must coexist while the migration is in progress (i did this before) or in my case the client wants the passwords to be reversible so in my case I'm encrypting the passwords instead of hashing them the `current_password` validation rule will fail because it assumes that the passwords are hashed.

This PR updates the password validation to use the `validateCredentials()` method on the `UserProvider` used by the guard instead of checking the hash.

This fully backward compatible.


taylorotwell:
I would just make your own validation rule. 👍

PR #41942, opened 2022-04-12
[9.x] withCount() add failing test for example code from docs

nuernbergerA:
Since #41914 the [example from the docs](https://laravel.com/docs/9.x/eloquent-relationships#counting-related-models) no longer works.

```php
 
$posts = Post::withCount(['votes', 'comments' => function (Builder $query) {
    $query->where('content', 'like', 'code%');
}])->get();
 
echo $posts[0]->votes_count;
echo $posts[0]->comments_count;
```

will throw `Undefined array key 1`

This PR adds a failing test to cover the expected behavior.

A possible fix would be to check if the given array is a list `array_values($relations) === $relations`
```php
    public function withCount($relations)
    {
        $relations = is_array($relations) ? $relations : func_get_args();

        if (count($relations) === 2 && array_values($relations) === $relations && $relations[1] instanceof Closure) {
            $relations = [$relations[0] => $relations[1]];
        }

        return $this->withAggregate($relations, '*', 'count');
    }
```

### DIFF
--- a/tests/Database/DatabaseEloquentBuilderTest.php
+++ b/tests/Database/DatabaseEloquentBuilderTest.php
@@ -1110,7 +1110,7 @@ class DatabaseEloquentBuilderTest extends TestCase
     {
         $model = new EloquentBuilderTestModelParentStub;
 
-        $builder = $model->withCount(['address', 'foo' => function($query) {
+        $builder = $model->withCount(['address', 'foo' => function ($query) {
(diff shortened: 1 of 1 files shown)


nuernbergerA:
@taylorotwell merged master test passes now.

you can reopen and merge this PR for test hardening?

driesvints:
@nuernbergerA best that you send in a new pr

PR #41944, opened 2022-04-12
[9.x] Fix breaking change introduced in #41914

r-kujawa:
Fixes breaking changes introduced in #41914 .

As mentioned by @nuernbergerA

@taylorotwell  Now I matched the `withCount()` & `withExists()` function signature to the `with()` function signature. It is no longer a breaking change.

Sorry about that.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -618,10 +618,15 @@ trait QueriesRelationships
      * Add subselect queries to count the relations.
      *
      * @param  mixed  $relations
+     * @param  string|\Closure|null  $callback
      * @return $this
      */
-    public function withCount($relations)
+    public function withCount($relations, $callback = null)
     {
+        if ($callback instanceof Closure) {
+            $relations = [$relations => $callback];
+        }
+
         return $this->withAggregate(is_array($relations) ? $relations : func_get_args(), '*', 'count');
     }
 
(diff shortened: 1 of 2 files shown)


r-kujawa:
I added the same tests as in the original PR (#41914) since it was reverted. It should also work with the tests being submitted in #41945.

derekmd:
Adding method arguments is also a breaking change so this solution would need to target 10.x. A non-breaking solution likely exists using your previous PR but adding a `func_num_args() === 2 && func_get_arg(1) instanceof Closure` check to detect the extra callback arg.

r-kujawa:
> Adding method arguments is also a breaking change so this solution would need to target 10.x. A non-breaking solution likely exists using your previous PR but adding a `func_num_args() === 2 && func_get_arg(1) instanceof Closure` check to detect the extra callback arg.

Could you help me understand why it is considered a breaking change if the added argument is `null` by default? I understand it is a change in the api, but in function could it actually break something?

derekmd:
1. In a Laravel app, update file vendor\laravel\framework\src\Illuminate\Database\Eloquent\Concerns\QueriesRelationships.php with new method signature `public function withCount($relations, $callback = null)`
2. In a Laravel Tinker prompt, define this new class:
   ```php
   class CustomBuilder extends \Illuminate\Database\Eloquent\Builder
   {
       public function withCount($relations)
       {
           //
       }
   }
   ```

Exception:

> Declaration of CustomBuilder::withCount($relations) must be compatible with Illuminate\Database\Eloquent\Builder::withCount($relations, $callback = null)

Laravel allows packages or userland app to extend the base framework with custom implementation overrides. Since any method signature change can break that behavior in a point release, such concrete argument changes are delayed until the next major release (10.x).

However in the case of Eloquent query builder methods `withCount()`, `withExists()`, and `with()`, they contain dynamic handling of method arguments so you can get around PHP's hard enforcement of the method signature without breaking existing packages.

taylorotwell:
Honestly not sure it's worth making breaking changes for this. It doesn't seem to be in very popular demand?

r-kujawa:
I understand @taylorotwell, would this be something you would consider if it was targeted at '10.x'?

PR #41947, opened 2022-04-12
[9.x] Ability to get method bindings from the container

michael-rubel:
### Why?
There might be a case when you may want to get method bindings from the container, in the same way as basic bindings. Additionally, it probably makes sense to flush them as other variables.


michael-rubel:
@taylorotwell May I ask why it was rejected? Does the PR break something?

PR #41949, opened 2022-04-13
[9.x] Trim zwnbsp

allowing:
"ZWNBSP" mean "Zero Width No-Break SPace"

The '\s' does not include it

image


rodrigopedra:
Hi @allowing would you send the same change to the recently added  `Str::squish()`?

I basically ported the same fix I used here before on PR #41877 

I guess it would be nice to keep both places consistent. Unfortunately this week I won't have time to work on this.

allowing:
@rodrigopedra :)
I may forget to reply to you, which has been implemented here. Supplement the reply for the convenience of other netizens. #41971

Issue #41745, opened 2022-03-30
file_exists(): File name is longer than the maximum allowed path length on this platform (4096)


fangmuke:
- Laravel Version: 9.5.1
- PHP Version: 8.0.11
- Database Driver & Version: MySQL 8

### Description: 
Calling `Blade::render($mailTemplateHtml)` can't work properly. 
it can't run normally in **php-fpm**.

### Steps To Reproduce:

```php
$mailTemplateHtml = str_repeat('a', 4097);
Blade::render($mailTemplateHtml .'{{ $name }}', ['name' => 'Taylor']);
```

fangmuke:
src/Illuminate/View/Component.php
```php
    /**
     * Resolve the Blade view or view file that should be used when rendering the component.
     *
     * @return \Illuminate\Contracts\View\View|\Illuminate\Contracts\Support\Htmlable|\Closure|string
     */
    public function resolveView()
    {
        $view = $this->render();

        if ($view instanceof ViewContract) {
            return $view;
        }

        if ($view instanceof Htmlable) {
            return $view;
        }

        $resolver = function ($view) {
            $factory = Container::getInstance()->make('view');

            return $factory->exists($view)
                        ? $view
                        : $this->createBladeViewFromString($factory, $view);
        };

        return $view instanceof Closure ? function (array $data = []) use ($view, $resolver) {
            return $resolver($view($data));
        }
        : $resolver($view);
    }
```

I think the problem is here. 
```php
return $factory->exists($view)
                ? $view
                : $this->createBladeViewFromString($factory, $view);
```

Using `Blade::render()` should not be judged whether the view exists, it will generate a view after calling `createbladeViewFromString()`.

a-bashtannik:
Could you try to run the test, please? 
```php
use Illuminate\Support\Facades\Blade;

public function test_rendering_long_blade_string() {
    $longString = str_repeat('a', 4097);
    $this->assertSame($longString . 'Taylor', Blade::render($longString .'{{ $name }}', ['name' => 'Taylor']));
}
```
Passes for me.
PHP 8.0.16, latest 9.x branch

fangmuke:
> Could you try to run the test, please?
> 
> ```
> use Illuminate\Support\Facades\Blade;
> 
> public function test_rendering_long_blade_string() {
>     $longString = str_repeat('a', 4097);
>     $this->assertSame($longString . 'Taylor', Blade::render($longString .'{{ $name }}', ['name' => 'Taylor']));
[...]

Issue #32254, opened 2020-04-06
Returning rendered blade component causes file_exists to fail on some systems


raitisg:
- Laravel Version: 7.4.0
- PHP Version: 7.2.28
- Database Driver & Version:

### Description:

When Component class returns string (html), `resolveView()` checks if such file exists before creating view from string. If component's content is large enough, exception is thrown: `file_exists(): File name is longer than the maximum allowed path length on this platform (4096): /path/to/html-string-...`

### Steps To Reproduce:

Create component class:

```
public function render() {
    return str_repeat('a', 99999);
    // or (more realistically):
    // return view('components.my-component')->render();
}
```

driesvints:
Can you please share the exact code that's causing this?

raitisg:
[src/Illuminate/View/Component.php:64](../blob/7.x/src/Illuminate/View/Component.php#L64)

```
return $factory->exists($view)
            ? $view
            : $this->createBladeViewFromString($factory, $view);
```

`$view` in this case would be `aaa....aaa` (or my more realistic scenario `<div>html of a component</div>`).

driesvints:
Please share your own code that's causing this.

raitisg:
Create blade component:
```
php artisan make:component Foobar
```

`app/View/Components/Foobar.php`:

```php
<?php

namespace App\View\Components;

use Illuminate\View\Component;

class Foobar extends Component
{
	public function render() {
		return str_repeat('a', 9999);
	}
}
```

`routes/web.php`:

```php
Route::get('/component-test', function(){
	return view('component');
});
```

`resources/views/component.blade.php`:

```html
<h1>Component</h1>

<x-foobar/>
```

Opening `/component-test` throws error:
```
file_exists(): File name is longer than the maximum allowed path length on this platform (4096): /absolute/path/to/resources/views/aaaaaaaaaaaaaaaaaaaa...
```

This error only occurs on production server (Ubuntu). On my local dev machine (MacOS) everything works fine, but the problem remains: Laravel calls `file_exists` with component's content as first argument before trying to compile it.

driesvints:
Is there any reason at all why you can't place the contents in a view?

raitisg:
I am placing it in a view, this is a simplified example. My render function looks more like this:

```
public function render() {
	return view('components.foobar')->render();
}
```

[...]

PR #41956, opened 2022-04-13
[9.x] Fix PHP warnings when rendering long blade string

a-bashtannik:
`\Illuminate\View\Component::resolveView` checks the string for being a file
```php
$factory->exists($view) ? $view : $this->createBladeViewFromString($factory, $view);
```
by calling `\Illuminate\Filesystem\Filesystem::exists` method, which is a pure standard `file_exists` function. 

PHP has a hard limit on path length as defined by the constant [PHP_MAXPATHLEN](https://www.php.net/manual/en/reserved.constants.php) which depends on the operating system (Windows 260 characters, Unix: as specified in the c-runtime - on Linux this is `PATH_MAX` in limits.h being 4096).

PHP detects the program is trying to check plain file existence and calls `php_check_open_basedir`

1. https://github.com/php/php-src/blob/master/ext/standard/filestat.c#L757

If the PHP configuration has no `open_basedir` value, then another method `virtual_file_ex` will be called:

1. https://github.com/php/php-src/blob/master/main/fopen_wrappers.c#L282
2. https://github.com/php/php-src/blob/master/ext/standard/filestat.c#L767
3. https://github.com/php/php-src/blob/master/Zend/zend_virtual_cwd.c#L1008

But `virtual_file_ex` doesn't throw warnings, just returns `false`.

If `open_basedir` is configured, `php_check_open_basedir_ex` is called, it cares about `open_basedir` and [throws a warning](https://github.com/php/php-src/blob/master/main/fopen_wrappers.c#L290), otherwise `file_exists` just returns false. The program behavior depends of `open_basedir` setting.

Although it's OK in the PHP design paradigm, users meet problems from time to time having this warning, see https://github.com/laravel/framework/issues/32254 https://github.com/laravel/framework/issues/41745 https://github.com/livewire/livewire/issues/918

This PR adds PHP_MAXPATHLEN check before calling `file_exists`.

Although I can add a test case for different open_basedir states, it will affect other tests running in parallel, and considering the issue is rare, I don't think there is a sense to make a custom php.ini environment for the case.

### DIFF
--- a/src/Illuminate/View/Component.php
+++ b/src/Illuminate/View/Component.php
@@ -74,7 +74,7 @@ abstract class Component
         $resolver = function ($view) {
             $factory = Container::getInstance()->make('view');
 
-            return $factory->exists($view)
+            return strlen($view) <= PHP_MAXPATHLEN && $factory->exists($view)
                         ? $view
                         : $this->createBladeViewFromString($factory, $view);
         };
--- a/tests/Integration/View/BladeTest.php
+++ b/tests/Integration/View/BladeTest.php
@@ -14,12 +14,13 @@ class BladeTest extends TestCase
         $this->assertSame('Hello Taylor', Blade::render('Hello {{ $name }}', ['name' => 'Taylor']));
     }
 
-    public function test_rendering_blade_long_maxpathlen_string() {
+    public function test_rendering_blade_long_maxpathlen_string()
+    {
         $longString = str_repeat('a', PHP_MAXPATHLEN);
 
-        $result = Blade::render($longString .'{{ $name }}', ['name' => 'a']);
+        $result = Blade::render($longString.'{{ $name }}', ['name' => 'a']);
 
-        $this->assertSame($longString . 'a', $result);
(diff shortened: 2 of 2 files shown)


taylorotwell:
Thanks for the PR, but can you succinctly describe the before / after behavior of this PR?

a-bashtannik:
@taylorotwell Sure

```php
Blade::render(string $string, array $data = [], bool $deleteCachedView = false)
```

❌ **Before:**

`$string` argument may be a raw blade template string (e.g. from the database) and a file path. If the length > PHP_MAXPATHLEN `Blade::render` fails because `file_exists` function throws a warning and `createBladeViewFromString` method is never called even if `$string` argument is a valid, but long blade template string. 

✅ **After** 

`$string` argument may have length > PHP_MAXPATHLEN, and considering it's impossible to have a file on the path with length > PHP_MAXPATHLEN, `file_exists` call is omitted, `createBladeViewFromString` is called and `Blade::render` returns compiled blade template as expected.

taylorotwell:
Thanks!

PR #41957, opened 2022-04-13
[9.x] Fix bug where runInBackground would cause commands not to run.

Boorinio:
When adding runInBackground and withoutOverlapping the mutex name is based on description when running the Scheduler, and when the schedule:finish command starts we are basing the mutex on expression+command signature, this causes commands not to be executed. 

This behaviour is reproducible is you try to modify a scheduled event 

foreach ($schedule->events() as $event) {
        if ($event instanceof Event && !$event instanceof CallbackEvent) {
            $event->runInBackground()->withoutOverlapping();
        }
    }

I think mutexNames should be based on the same thing, and imo expression+command makes more sense.

### DIFF
--- a/src/Illuminate/Console/Scheduling/CallbackEvent.php
+++ b/src/Illuminate/Console/Scheduling/CallbackEvent.php
@@ -184,7 +184,7 @@ class CallbackEvent extends Event
      */
     public function mutexName()
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:
No reproducible example was included in your PR that I can copy and paste into an application. If you would like to include that you can submit another PR.

PR #41958, opened 2022-04-13
[9.x] Add Eloquent `->toggle()` function for Boolean Attributes

ralphjsmit:
This PR adds a little function called `->toggle()` to Eloquent models. The use case is toggling boolean values in an expressive way. 

There's not a lot to explain, so here are a few examples:

```php
$user = User::create([
    'is_active' => false,
]);

// Before
$user->is_active = true; 

// Or:
$user->is_active = ! $user->is_active;

// After:
$user->toggle('is_active');

// You can directly save the model:
$user->toggle('is_active')->save();
```

(Regarding the position of the function, I now put it into the base `Model` class, but if you think that is more fitting, you can of course also put it in the HasAttributes trait.)

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1488,6 +1488,19 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     }
 
     /**
+     * Toggle a boolean attribute.
+     *
+     * @param  string $attribute
+     * @return static
+     */
+    public function toggle($attribute)
+    {
+        $this->{$attribute} = ! $this->{$attribute};
+
+        return $this;
+    }
+
+    /**
      * Convert the model instance to JSON.
      *
      * @param  int  $options
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1495,7 +1495,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
      */
     public function toggle($attribute)
     {
-        $this->{$attribute} = ! $this->{$attribute};
+        $this->setAttribute($attribute, ! $this->getAttribute($attribute));
 
         return $this;
     }
--- a/tests/Database/DatabaseEloquentModelTest.php
+++ b/tests/Database/DatabaseEloquentModelTest.php
@@ -69,6 +69,19 @@ class DatabaseEloquentModelTest extends TestCase
         $this->assertSame(json_encode(['name' => 'taylor']), $attributes['list_items']);
     }
 
+    public function testToggleBooleanAttributes()
+    {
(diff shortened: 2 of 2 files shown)


antonkomarev:
I've added toggle methods in [Laravel Love](https://github.com/cybercog/laravel-love) package in early releases, but later decided to remove it and avoid this pattern. The main reason, to make idempotent requests.

Use case:
- Post is inactive and user want to activate it
- user is making 2 clicks on activate button in SPA UI because of internet lag or something (response is not received between clicks)
- backend executes `Post::toggle('is_active')` 2 times
- Post is still inactive

If we will send exact status which should be set, then 2nd call might be ignored.

ralphjsmit:
> I've added toggle methods in [Laravel Love](https://github.com/cybercog/laravel-love) package in early releases, but later decided to remove it and avoid this pattern. The main reason, to make idempotent requests.
> 
> Use case:
> 
> * Post is inactive and user want to activate it
> * user is making 2 clicks on activate button in SPA UI because of internet lag or something (response is not received between clicks)
> * backend executes `Post::toggle('is_active')` 2 times
> * Post is still inactive
> 
> If we will send exact status which should be set, then 2nd call might be ignored.

Thanks for describing this use case! To me, this personally feels more like a “UX-problem”, which should be solved (or not)  and not by Eloquent. You could even argue that it is bad UX for the user that the backend isn’t doing anything with the second click. Plus, after processing those two clicks, the user will still _see_ the toggle as inactive, thus knowing that it clicked the toggle too much and/or should click another time.

IMO, if you’re using an SPA and are worried about internet lag, you will be able to gracefully handle this yourself. Otherwise, the ->toggle() helper will still be useful for making tests and other business logic more expressive. That’s just my point of view, but thanks for adding your considerations!

dennisprudlo:
In any way this will have to target 10.x because it may break applications where models already have a toggle function.

driesvints:
@dennisprudlo in general, we don't consider overriding classes to be part of the public API.

dennisprudlo:
@driesvints I see but as of right now there is no `toggle` function in a model. If I was to create one in my `User` model like so:
```php
public function toggle(bool $value) : bool
{
    return !$value;
}
```

and then update Laravel 9.x with this PR merged my app will break because both of those methods have a different signature.
```
App\Models\User::toggle(bool $value): bool must be compatible with Illuminate\Database\Eloquent\Model::toggle($attribute)
```

Yet I didn't override something I just created a simple method like any other method in my model. I mean `toggle` is a very generic name for a method.

driesvints:
@dennisprudlo we don't consider that to be a breaking change. If we had to take that into account we'd need to release a new major version of Laravel for every public method we add. This situation is also specifically mentioned in the SemVer spec: https://semver.org/#spec-item-7

dennisprudlo:
@driesvints Got it 👍🏼 Then good for me I saw this PR and know to update my toggle method if this gets merged.

Man do I hate that PHP doesn't support method overloading 😄

taylorotwell:

PR #41963, opened 2022-04-13
[9.x] Use arrow functions

imanghafoori1:
Replaces:
```php
function () use ($variable) {
    return $variable;
}
```
with:
```php
fn () => $variable;
```

### DIFF
--- a/src/Illuminate/Bus/DatabaseBatchRepository.php
+++ b/src/Illuminate/Bus/DatabaseBatchRepository.php
@@ -286,9 +286,7 @@ class DatabaseBatchRepository implements PrunableBatchRepository
      */
     public function transaction(Closure $callback)
     {
-        return $this->connection->transaction(function () use ($callback) {
-            return $callback();
-        });
+        return $this->connection->transaction(fn () => $callback());
     }
 
     /**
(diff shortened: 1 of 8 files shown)


src/Illuminate/Bus/DatabaseBatchRepository.php
```diff
@@ -286,9 +286,7 @@ public function pruneUnfinished(DateTimeInterface $before)
      */
     public function transaction(Closure $callback)
     {
-        return $this->connection->transaction(function () use ($callback) {
-            return $callback();
-        });
+        return $this->connection->transaction(fn () => $callback());
```
derekmd, line 289: This line can be shortened further, removing the extra stack call:

```php
return $this->connection->transaction($callback);
```

which would also allow `$callback` to optionally accept the `Illuminate\Database\Connection` instance as an argument.

https://github.com/laravel/framework/blob/76a158efd9a760da708bf957c7df1377bfce1315/src/Illuminate/Database/Concerns/ManagesTransactions.php#L28-L30

imanghafoori1: That requires your contribution. I do not dare to go for it.

imanghafoori1:
Database integration tests are failing for unrelated reasons to this PR.

PR #41967, opened 2022-04-13
[9.x] Fix database events PHPDoc so it uses ConnectionInterface instead of Connection

taka-oyama:
I was getting PHPDoc warnings in PhpStorm when trying to create a new class that extends `Illuminate\Database\Events\ConnectionEvent` because I was trying to pass `Illuminate\Database\ConnectionInterface` 
in the first argument of the constructor when the event was expecting a `Illuminate\Database\Connection`. 

This is probably incorrect and `Illuminate\Database\Events\ConnectionEvent` should be expecting a `Illuminate\Database\ConnectionInterface`.

I have fixed other events in `Illuminate\Database\Events` as well for consistency.

### DIFF
--- a/src/Illuminate/Database/Events/ConnectionEvent.php
+++ b/src/Illuminate/Database/Events/ConnectionEvent.php
@@ -14,14 +14,14 @@ abstract class ConnectionEvent
     /**
      * The database connection instance.
      *
-     * @var \Illuminate\Database\Connection
+     * @var \Illuminate\Database\ConnectionInterface
      */
     public $connection;
 
     /**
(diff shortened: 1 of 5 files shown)


driesvints:
This is technically a breaking change since people expect a database connection instance right now rather than any implementation of a connection interface.

taka-oyama:
Hi, thanks for the feedback. 

I understand the concern now.

Would it be better if the pull request was targeted for 10.0 branch?

driesvints:
I'll let @taylorotwell decide.

taylorotwell:

PR #41971, opened 2022-04-14
[9.x] Str::squish trim ZWNBSP

allowing:
#41949

### DIFF
--- a/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
+++ b/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
@@ -53,7 +53,7 @@ class TrimStrings extends TransformsRequest
             return $value;
         }
 
-        return is_string($value) ? preg_replace('~^[\s﻿]+|[\s﻿]+$~iu', '', $value) : $value;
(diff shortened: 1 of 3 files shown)


nshiro:
@allowing
Hello! I saw your comment today.

Nice PR, but I am a little confused.
I thought this `squish` is supposed to cut spaces from the string in the middle, too.
If so, don't we need "ZWNBSP" in the first preg_replace() too for consistency ?

#41791
#41877
#41924

allowing:
@nshiro 
嗯对的，第一个也应该加上。感谢你的提醒。

其实通过枚举的方式来实现，有点Hack的味道。没办法，'\s'并不完全包含不可见字符。

Well yes, the first one should also be added. Thank you for your reminder.

In fact, it is implemented by enumeration, which is a bit like Hack. No way, '\s' doesn't exactly contain invisible characters.

allowing:
Oh!

The `ZWNBSP` cannot be replaced with spaces. Because it's called `Zero Width No-Break SPace`. It doesn't take up space.

nshiro:
Ah, things are not simple.. 😢

PR #41980, opened 2022-04-14
[9.x] Allow appending query strings to paginators globally

innocenzi:
This PR adds a new `Paginator::appendQueryString()` method that appends query strings by default to all `Paginator` instances. This method can be added in a service provider. 

I needed this because I call `->withQueryString` on basically all my paginators. My main use case is paginating and filtering with query strings using Inertia.

### DIFF
--- a/src/Illuminate/Pagination/AbstractPaginator.php
+++ b/src/Illuminate/Pagination/AbstractPaginator.php
@@ -127,7 +127,7 @@ abstract class AbstractPaginator implements Htmlable
      *
      * @var bool
      */
-    protected static bool $appendsQueryString = false;
+    protected static $appendsQueryString = false;
 
     /**
      * Determine if the given value is a valid page number.
@@ -655,7 +655,7 @@ abstract class AbstractPaginator implements Htmlable
     /**
      * Indicate if all query string values should be appended to the paginator.
      *
-     * @param  bool|\Closure  $value
+     * @param  \Closure|bool  $value
      * @return void
      */
     public static function appendQueryString($value = true)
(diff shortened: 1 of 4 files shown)


src/Illuminate/Pagination/AbstractPaginator.php
```diff
@@ -122,6 +122,13 @@ abstract class AbstractPaginator implements Htmlable
      */
     public static $defaultSimpleView = 'pagination::simple-tailwind';
 
+    /**
+     * Indicates if all query string values should be appended to the paginator.
+     *
+     * @var bool
+     */
+    protected static bool $appendsQueryString = false;
```
driesvints, line 130: Remove the typehint here.

src/Illuminate/Pagination/AbstractPaginator.php
```diff
@@ -645,6 +652,17 @@ public static function useBootstrapFive()
         static::defaultSimpleView('pagination::simple-bootstrap-5');
     }
 
+    /**
+     * Indicate if all query string values should be appended to the paginator.
+     *
+     * @param  bool|\Closure  $value
```
driesvints, line 658:
```suggestion
     * @param  \Closure|bool  $value
```

driesvints (changes requested):

src/Illuminate/Pagination/AbstractPaginator.php
innocenzi: Done!

taylorotwell:

innocenzi:
@taylorotwell Is it an implementation issue? Honestly, I think this would be pretty useful. Having to call `->withQueryString()` along with every single `->paginate()` call is cumbersome. 

Happy to discuss another way to solve this.

PR #41981, opened 2022-04-14
Added exists method to collection

debuqer:
This PR will add exists() method to Collection and LazyCollection like Query/Builder

```php
$collection = collect([1, 2, 3, 4]);
 
$filtered = $collection->filter(function ($value, $key) {
    return $value > 2;
})->exists();  // will return TRUE

$filtered = $collection->filter(function ($value, $key) {
    return $value > 5;
})->exists();  // will return FALSE

```

It also can be used on result of Eloquent as collection

```php
$users = Subscription::where('is_active', 1)->get();

if( $users->exists() ) {
      $activeUsers = $users;
     // do something
} else {
      throw new NoActiveUserFoundException;
}

```
Instead of
```php
$users = Subscription::where('is_active', 1)->get();

if( $users->count() > 0 ) {
      $activeUsers = $users;
     // do something
} else {
      throw new NoActiveUserFoundException;
}

```

The main purpose is to having same existence check method for both Collection and Eloquent

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -361,6 +361,16 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
+     * Determine if at least one item exists
+     *
+     * @return bool
+     */
+    public function exists()
+    {
+        return $this->count() > 0;
+    }
+
+    /**
      * Run a filter over each of the items.
      *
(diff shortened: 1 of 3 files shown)


marvin-wtt:
Isn't that the same as [isNotEmpty()](https://laravel.com/docs/9.x/helpers#method-fluent-str-is-not-empty)?

taylorotwell:

PR #41982, opened 2022-04-14
[9.x] Revert lazy command forcing

taylorotwell:


etyson17:
I was wondering if someone could help me understand a piece of code related to this commit [here](https://github.com/laravel/framework/pull/41982/files#diff-12e294b5c3dce364d3becb97174de0d28b4cfc888058b7e593f9f755a6eeaf30R7-R31).

I am working on a Laravel 0 app and I am unable to hide the `db:wipe` command the by adding it to the hidden section like with other commands, however by removing these 3 line it works as expected.

`use Symfony\Component\Console\Attribute\AsCommand;`
`#[AsCommand(name: 'db:wipe')]`
`protected static $defaultName = 'db:wipe';`

I was hoping you could explain why you added the first 2 back and how or what is using the 3rd.

PR #41983, opened 2022-04-14
[9.x] Nested Resource routes shorter URL method

cesarep:
Hi!

This PR proposes a new method to a resource route, making nested resources URLs shorter
Use cases might be something like:
For a `Route::resource('states.cities')` for example, instead of a `/states/{state}/cities/{city}` URL, one might want something like `cities/{state:abbreviation}/{city:name}` (Eg. `cities/CA/losAngeles`), without the need to separately define all the action routes:
```
index => (GET) cities/{state:abbreviation}
create => (GET) cities/{state:abbreviation}/create
store =>(POST) cities/{state:abbreviation}
show => (GET) cities/{state:abbreviation}/{city:name}
edit => (GET) cities/{state:abbreviation}/{city:name}/edit
[...]
```
With this, it's possible to set all the resource routes with a single:
`Route::resource('cities/states.cities')->attributeOnly(['state' => 'abbreviation', 'city' => 'name'])`

As this URL format probably only makes sense if used with slugs on the URL, it also calls the scoped function inside

I couldn't think of a better name for the method, it's kinda of a shallow URL i guess, but Shallow is already in use so...
Please feel free to think of a better name for the method.

Should be fully compatible with 8.x too

### DIFF
--- a/src/Illuminate/Routing/PendingResourceRegistration.php
+++ b/src/Illuminate/Routing/PendingResourceRegistration.php
@@ -202,6 +202,19 @@ class PendingResourceRegistration
     }
 
     /**
+     * Indicate that a nested resource routes should have the attributes only.
+     *
+     * @param  bool  $attronly
+     * @return \Illuminate\Routing\PendingResourceRegistration
+     */
+    public function attributeOnly($attronly = true)
+    {
+        $this->options['attronly'] = $attronly;
+
+        return $this;
+    }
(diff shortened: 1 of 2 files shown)


driesvints:
> Should be fully compatible with 8.x too

8.x doesn't receives new features anymore.

taylorotwell:

PR #41896, opened 2022-04-09
[9.x] Add getAllTables support for SQLite and SQLServer schema builders

jhavenz:
This PR adds support for calling `\Illuminate\Support\Facades\Schema::getAllTables()` on SQLite and SQLServer connections.

I'm returning the `name` and `type` columns to maintain consistency with what's already being returned when using the MySQL or Postgres connections.

P.S. 
This being my 1st Laravel contribution, please go easy on me if I've done something incorrectly here. 😉

PR #41984, opened 2022-04-15
[9.x] fix SQLServer test

jhavenz:
This fixes the failing SQLServer builder test submitted in #41896, referenced here:
https://github.com/laravel/framework/runs/6003485596?check_suite_focus=true


driesvints:
I feel like we should go ahead with https://github.com/laravel/framework/pull/41985 as that's a better alternative. Thanks!

PR #41985, opened 2022-04-15
[9.x] Fix failing SQL Server integration tests

derekmd:
Replaces https://github.com/laravel/framework/pull/41984

* Avoid the other PR's approach where test setup depends on calling `Schema::getAllTables()` / `SqlServerBuilder::getAllTables()` since these are the methods the integration tests are covering.
* Simplify the test cases (that copied `DatabaseSqliteSchemaBuilderTest` / `DatabaseMySqlSchemaBuilderAlterTableWithEnumTest`) which were cramming 2 or 3 different cases into each test method.
* `DB::connection('sqlsrv')` / `Schema::connection('sqlsrv')` don't need explicit calls as this is already the default database connection when any test class extending `SqlServerTestCase` is run.

### DIFF
--- a/tests/Integration/Database/SqlServer/DatabaseSqlServerSchemaBuilderTest.php
+++ b/tests/Integration/Database/SqlServer/DatabaseSqlServerSchemaBuilderTest.php
@@ -9,15 +9,6 @@ use stdClass;
 
 class DatabaseSqlServerSchemaBuilderTest extends SqlServerTestCase
 {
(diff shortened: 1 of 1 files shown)


driesvints (approved):
This sounds like a better alternative to me as well 👍

taylorotwell:
Thanks!

Sourcefli:
fair enough,  thanks @derekmd

PR #41989, opened 2022-04-15
Deep copy functionality on Collection and LazyCollection

fmeccanici:
This PR adds the functionality of creating a deep copy of a Collection instance. This means that new references are created for the objects inside the Collection. A deep copy is useful when the Collection contains non built-in types, since `clone` will only make new references of built-int types. So for example here:
```
$testObject = new TestObject();
$testObject->quantity = 1;
$collection = new Collection(array($testObject));
$clonedCollection = clone $collection;
$clonedCollection->changeQuantity(2);

class TestObject 
{
     public int $quantity;
}
```
What happens is that both the quantity of the `$clonedCollection` and the `$collection` instances have been changed to 2, as they share the same reference of `$testObject`. To solve this, you can map over the instances yourself and return a cloned version:

```
$collection->map(function (TestObject $testObject) 
{
      return clone $testObject;
});
```
But this will not work if `TestObject` has instances of non built-in types. So therefore, a better solution in my opinion is to have a deepCopy function on a Collection instance. A deep copy will make sure that new references are made for every instance, even for non built-in types and works for nested objects. I did not implement the deep copy logic myself, but use the library from myclabs/deep-copy. 

The end user can then use the following:
```
$deepCopiedCollection = $collection->deepCopy();
```
Let me know what you think!

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
         "league/commonmark": "^2.2",
         "league/flysystem": "^3.0",
         "monolog/monolog": "^2.0",
+        "myclabs/deep-copy": "^1.11",
         "nesbot/carbon": "^2.53.1",
         "psr/container": "^1.1.1|^2.0.1",
         "psr/log": "^1.0|^2.0|^3.0",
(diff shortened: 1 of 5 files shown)


driesvints:
There's no explanation here.

Issue #15286, opened 2016-09-05
Method to Clone Collections


jaschweder:
- Laravel Version: 5.2.8
- PHP Version: 5.6.25
- Database Driver & Version: not applied

### Description:
One method that is important do Illuminate/Collection have is ```clone()```, this method generate a copy of the current collection.

Can anyone make some comments here why or why not this is a bad idea ?

Thanks.

### Steps To Reproduce:
```php
$collection = collect();
$clone = $collection->clone();
// or
$clone = clone $collection;
```

alexbowers:
Why are you not able to use the `$clone = clone $collection;` example?

For me that works:

```php
$collection = collect([1, 2, 3]);

    dd(
        $collection,
        clone $collection
    );
```

![image](https://cloud.githubusercontent.com/assets/842974/18269088/8e8216b8-741e-11e6-972a-5e5bb644b560.png)

Or is this a feature request?

If it is, its pretty simple to complete with a macro;

```php
\Illuminate\Support\Collection::macro('clone', function() {
        return clone $this;
    });
```

jaschweder:
@alexbowers this works with simple internals, but test this with objects that are passed has references, the clone method obviously don't copy the objects, only the references, is correct to copy all object's in the collection or mantain the references to them original objects ?

alexbowers:
I'd say if you were cloning a reference, you'd expect it to still reference the same thing. A clone should be an exact replica.

GrahamCampbell:
> I'd say if you were cloning a reference, you'd expect it to still reference the same thing

That's not the definition of cloning.

GrahamCampbell:
Cloning a collection will give a shallow clone. Are you asking for a way to do a deep clone?

alexbowers:
You could always traverse the object and clone each item individually if it is itself an object.  Doing this recursively would be a deep clone I guess

jaschweder:
@GrahamCampbell @alexbowers yes deep-copy is what I mean, any chance to have some native stuff in Laravel for this ? or is more application like feature ?

themsaid:
Well I'm going to close this Issue since it's not a bug report, I suggest that you open a PR with the suggested changes or open an issue on the internals repository, it's where we keep all our discussions & feature requests.

Thanks for the suggestion though :)

PR #41994, opened 2022-04-15
[9.x] Deep copy functionality on Collection and LazyCollection

fmeccanici:
This PR adds the functionality of creating a deep copy of a Collection instance. This means that new references are created for the objects inside the Collection. A deep copy is useful when the Collection contains non built-in types, since `clone` will only make new references of built-int types (shallow copy). So for example here:
```
$testObject = new TestObject();
$testObject->quantity = 1;
$collection = new Collection(array($testObject));
$clonedCollection = clone $collection;
$clonedCollection->changeQuantity(2);

class TestObject 
{
     public int $quantity;
}
```
What happens is that both the quantity of the `$clonedCollection` and the `$collection` instances have been changed to 2, as they share the same reference of `$testObject`. To solve this, you can map over the instances yourself and return a cloned version:

```
$collection->map(function (TestObject $testObject) 
{
      return clone $testObject;
});
```
But this will not work if `TestObject` has instances of non built-in types, then you need to loop through them too. So therefore, a better solution in my opinion is to have a deepCopy function on a Collection instance, such that the end user does not need to loop through the nested objects and clone them. A deep copy will make sure that new references are made for every instance, even for non built-in types and works for nested objects. I did not implement the deep copy logic myself, but I use the library from myclabs/deep-copy. 

The end user can then use the following:
```
$deepCopiedCollection = $collection->deepCopy();
```

Let me know what you think! Also, this feature was proposed in a closed issue (https://github.com/laravel/framework/issues/15286). But no new PR was ever made to add this feature.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,7 @@
         "league/commonmark": "^2.2",
         "league/flysystem": "^3.0",
         "monolog/monolog": "^2.0",
+        "myclabs/deep-copy": "^1.11",
         "nesbot/carbon": "^2.53.1",
         "psr/container": "^1.1.1|^2.0.1",
         "psr/log": "^1.0|^2.0|^3.0",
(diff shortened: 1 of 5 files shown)


ankurk91:
One can install `myclabs/deep-copy` package in their project and deep copy the objects.

taylorotwell:

PR #41998, opened 2022-04-15
[9.x] Removes core service providers from the skeleton

andriihorpenko:
This PR allows removing core service providers from the `config/app.php` from the skeleton. It implements same design decisions Nuno and Taylor used in the similar PR regarding core class aliases (https://github.com/laravel/framework/pull/40062).

The motivation is exactly the same — these classes are rarely modified by the majority of users, so they can find their new home in the core itself.

An example below shows what the future iteration of the skeleton may look like:
```php
'providers' => ServiceProvider::defaultProviders()->merge([
    App\Providers\AppServiceProvider::class,
    App\Providers\AuthServiceProvider::class,
    // App\Providers\BroadcastServiceProvider::class,
    App\Providers\EventServiceProvider::class,
    App\Providers\RouteServiceProvider::class,

    // ...
])->toArray(),
```

As this is presumably not a breaking change (existing applications may opt for this way of defining service providers only by manually changing the config file), this PR is targeting `9.x` branch, but will gladly switch to `master` if required.

### DIFF
--- a/src/Illuminate/Support/ServiceProvider.php
+++ b/src/Illuminate/Support/ServiceProvider.php
@@ -3,12 +3,34 @@
 namespace Illuminate\Support;
 
 use Closure;
+use Illuminate\Auth\AuthServiceProvider;
+use Illuminate\Auth\Passwords\PasswordResetServiceProvider;
+use Illuminate\Broadcasting\BroadcastServiceProvider;
+use Illuminate\Bus\BusServiceProvider;
+use Illuminate\Cache\CacheServiceProvider;
 use Illuminate\Console\Application as Artisan;
 use Illuminate\Contracts\Foundation\CachesConfiguration;
 use Illuminate\Contracts\Foundation\CachesRoutes;
 use Illuminate\Contracts\Support\DeferrableProvider;
+use Illuminate\Cookie\CookieServiceProvider;
+use Illuminate\Database\DatabaseServiceProvider;
 use Illuminate\Database\Eloquent\Factory as ModelFactory;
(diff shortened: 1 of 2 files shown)


crynobone:
- How do you swap any of core service providers with custom class?
- How do you remove any of core service providers?

andriihorpenko:
@crynobone, as I mentioned, this change is great for the majority of users which rarely/never touch this portion of the config. If you need more gradual control over this — you can either copy the whole default section or use provided examples below.

> How do you swap any of core service providers with custom class?

Here we have two options: use the full power of collections, or just copy the default list of service providers to our own configuration file. This is how it may look like by using collections magic:
```php
...
->transform(fn($provider) => match ($provider) {
    CookieServiceProvider::class => FooServiceProvider::class,
    default => $provider,
})->toArray(),
```
> How do you remove any of core service providers?

We can use the `diff` method which will exclude specified classes from the final collection. I was initially using it to exclude `BroadcastServiceProvider` as an example, but the current variant looks cleaner as for a skeleton.
```php
...
->diff([
    CookieServiceProvider::class,
])->toArray(),
```

GrahamCampbell:
I prefer what we already have. It is much better DX than what is proposed here.

andriihorpenko:
I wonder why Nuno and Taylor applied that cleanup for core class aliases only, as the service providers section is right above aliases. Maybe there’s indeed a reason behind that decision. 🤷🏻‍♂️

taylorotwell:
I think because aliases are more rarely changed.

PR #42001, opened 2022-04-16
[9.x]: add `isProduction` & `isLocal` to `Application` contract

dammy001:


driesvints:
Adding methods to a contract on a minor or patch release is a breaking change. I also don't think we should pollute the contract with these methods.

PR #42003, opened 2022-04-16
[9.x] Add common base class for paginators

iamgergo:
This is a sort of retry of #41204 and #41199.

-------

Currently the paginator classes does not have a common base class. Because of this, there are a big portion of code duplication that could be prevented by using a base class.

My former PRs were closed, because the refacoring was too big and might had some potentional bugs.

In this PR I just added the empty base class, and if it's merged, then we can slowly start a step-by-step refactoring without the risk that a big code chage holds.

### DIFF
--- a/src/Illuminate/Pagination/AbstractBasePaginator.php
+++ b/src/Illuminate/Pagination/AbstractBasePaginator.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Illuminate\Pagination;
+
+abstract class AbstractBasePaginator
+{
+    //
+}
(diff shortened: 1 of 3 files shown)


driesvints:
I don't see the use case of this? Why do you need this?

iamgergo:
As I wrote in my prevous PRs, the paginator classes implement mostly the same functionality with exactly or at least nearly the same code. Also, they don't have any common root/base class, and that feels a bit odd to me.

Reorgaizing a bit would clean up the current classes' code base and would prevent future duplications. Also, if in the future, by any reason a new paginator type would be added, it also could use the base class.

driesvints:
I understand reusage but an empty base class doesn't serves any purpose

GrahamCampbell:
Agree with Dries here.

taylorotwell:

PR #42005, opened 2022-04-16
[9.x] Use increment method realize decrement method

laradocs:
Follow the DRY principle


GrahamCampbell:
This is a major breaking change for people extending these classes.

PR #42012, opened 2022-04-16
[9.x] Delete replication withoutTrashed method 

ghaninia:
The withoutTrashed method seems to be replicated in the following two classes:

1. Illuminate/Validation/Rules/Exists
2. Illuminate/Validation/Rules/Unique

It is better to transfer this method to a DatabaseRule trait .


GrahamCampbell:
Thanks for the PR, but since this is a breaking change, it should target the 10.x, the master branch.

PR #42013, opened 2022-04-16
[9.x] Better handling for DateTime classes when using `mapInto()`

HajMo:
This PR will resolve an issue that occur when mapping any collection values to date classes, for example:

```php
$mappedMonths = collect(['2022-01-01', '2022-02-01', '2022-03-01'])->mapInto(Carbon::class);
```

when debugging the `$mappedMonths` variable, it will show the timezone for every date is increased by each date key in `$mappedMonths` array, for example:
```php
// Croped from the Carbon objects...
date: 2022-01-01 00:00:00.0 +00:00 
date: 2022-01-01 00:00:00.0 +01:00 
date: 2022-01-01 00:00:00.0 +02:00 
```
And when mapping to `DateTime` class, it will throw an exception: `Argument #2 ($timezone) must be of type ?DateTimeZone, int given`

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -388,6 +388,10 @@ trait EnumeratesValues
     public function mapInto($class)
     {
         return $this->map(function ($value, $key) use ($class) {
+            if ($class == 'DateTime' || in_array('DateTime', class_parents($class))) {
+                return new $class($value);
+            }
+
             return new $class($value, $key);
         });
     }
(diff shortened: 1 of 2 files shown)


derekmd:
[`Carbon::make()`](https://github.com/briannesbitt/Carbon/blob/3226de554007c79a4f81b29320b2bdcbe4462908/src/Carbon/CarbonInterface.php#L3233-L3245) can be used with `collect()->map()` since the former only supports one argument:

```php
collect(['2022-01-01', '2022-02-01', '2022-03-01'])->map([Carbon::class, 'make'])->all();

// [
//   Carbon\Carbon @1640995200 {
//     date: 2022-01-01 00:00:00.0 UTC (+00:00),
//   },
//   Carbon\Carbon @1643673600 {
//     date: 2022-02-01 00:00:00.0 UTC (+00:00),
//   },
//   Carbon\Carbon @1646092800 {
//     date: 2022-03-01 00:00:00.0 UTC (+00:00),
//   }
// ]
```

taylorotwell:
See above.

PR #42016, opened 2022-04-17
[9.x] Record the place that call log method

allowing:
image

### DIFF
--- a/src/Illuminate/Log/Logger.php
+++ b/src/Illuminate/Log/Logger.php
@@ -178,9 +178,13 @@ class Logger implements LoggerInterface
      */
     protected function writeLog($level, $message, $context): void
     {
+        $backtrace = debug_backtrace(\DEBUG_BACKTRACE_PROVIDE_OBJECT, 4)[3];
+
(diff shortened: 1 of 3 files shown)


allowing:
通过回溯定位调用的地方并不总是有效，像是一种hack，而且还会影响其他单元测试。

Locating where the call is made by backtracking doesn't always work, it's a hack, and it affects other unit tests.

mfn:
There exists already a solution for this in monolog itself, it's called [`IntrospectionProcessor`](https://github.com/Seldaek/monolog/blob/4192345e260f1d51b365536199744b987e160edc/src/Monolog/Processor/IntrospectionProcessor.php#L17-L31)

In `config/logging.php` you need use the `via` key so your own code is called to instantiate the logger and then you add the processes, e.g.
```php
    'via' => MyApplicationLogger::class,
```
```php
class MyApplicationLogger {
  public function __invoke(array $config) {
    …
    $processors = [
      new IntrospectionProcessor(…),
    ];
    return new Logger(…, …, $processors);
  }
```

PR #42017, opened 2022-04-17
[9.x] Correct php doc setUser method

ghaninia:
I do not understand the reason for using alias for **Authenticatable** interface !

### DIFF
--- a/src/Illuminate/Contracts/Auth/Guard.php
+++ b/src/Illuminate/Contracts/Auth/Guard.php
@@ -51,7 +51,7 @@ interface Guard
      * Set the current user.
      *
      * @param  \Illuminate\Contracts\Auth\Authenticatable  $user
-     * @return void
+     * @return $this
      */
     public function setUser(Authenticatable $user);
 }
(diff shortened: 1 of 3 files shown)


src/Illuminate/Contracts/Auth/Guard.php
```diff
@@ -51,7 +51,7 @@ public function hasUser();
      * Set the current user.
      *
      * @param  \Illuminate\Contracts\Auth\Authenticatable  $user
-     * @return void
+     * @return $this
```
driesvints, line 54: This is a breaking change

ghaninia: In the **TokenGuard** class, a trait called **GuardHelpers** is used ,and in that there is a method called setUser. Note the output of this method

GuardHelpers

- Illuminate\Auth\GuardHelpers (line 91)

driesvints: The `setUser` methods that implement this contract should be `void`. We don't use the return anywhere in the framework.

driesvints:
I've sent in an alternative to this here: https://github.com/laravel/framework/pull/42027.

PR #42027, opened 2022-04-18
[10.x] Remove instance return from setUser

driesvints:
Alternative for #42017. The return here isn't needed. The contract already indicates this should be a void return. Right now IDE's warn about this so it's best that we match the type of the contract.


taylorotwell:
If we keep this for 10.x and don't end up reverting it we probably will want to search through our first-party packages to make sure none of them expect a return here.

driesvints:
@taylorotwell will do

PR #42034, opened 2022-04-18
use arrow function in broadcasting folder

ghaninia:
In BroadcastController file, I think we can use condition inline
arrow function was used in this folder as much as possible


taylorotwell:
Some of these changes are less readable... like the change to BroadcastController.

PR #42035, opened 2022-04-19
[9.x] Fixed PendingMail send() method

akhan619:
As per the laravel 9.x documentation `Mail::send()` should return the `Illuminate\Mail\SentMessage` instance. However because of method chaining, after a call to methods like `to()`, `cc()`, etc we are working with a `PendingMail` object. The `send()` method on the `PendingMail` class correctly routes the call to the underlying mailer but doesn't return the result of the call, hence breaking the return type. This PR attempts to fix this issue.

Relevant documentation: `https://laravel.com/docs/9.x/upgrade#symfony-mailer`
>  The send, html, text, and plain methods no longer return the number of recipients that received the message. Instead, an instance of Illuminate\Mail\SentMessage is returned.

### DIFF
--- a/src/Illuminate/Mail/PendingMail.php
+++ b/src/Illuminate/Mail/PendingMail.php
@@ -117,11 +117,11 @@ class PendingMail
      * Send a new mailable message instance.
      *
      * @param  \Illuminate\Contracts\Mail\Mailable  $mailable
-     * @return void
+     * @return \Illuminate\Mail\SentMessage|null
      */
     public function send(MailableContract $mailable)
     {
-        $this->mailer->send($this->fill($mailable));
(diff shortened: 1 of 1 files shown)


driesvints:
So, that reference in the docs was for the Mailer class, not the PendingMail class. I've updated the docs to better reflect what was actually meant: https://github.com/laravel/docs/pull/7883. The PendingMail class should be left unaltered.

akhan619:
Ok. That makes it clear. Cheers.

akhan619:
@driesvints I had another look at the `Mailer` class. The bug is still there or at least the functionality of `Mailer` is reduced. Please refer to PR #42048 for details.

driesvints:
Replied on the related PR.

PR #42036, opened 2022-04-19
Allow to pass a method to call for `mapInto`

ekvedaras:
Adds a possibility to use `mapInto` with a static method instead of a constructor.

You may want to have classes where the constructor uses promoted properties or mapping needs some special logic, but still keep the constructor with all the properties instead of just accepting value as a single param. 

This would allow for code like this:

```php
class A {
    public function __construct(
        public string $id,
        public int $amount,
        public Carbon $createdAt,
    ) {
    }

    public static function fromDbRow(array $row): static
    {
        return new static(
            id: $row['id'],
            amount: $row['amount'],
            createdAt: Carbon::parse($row['created_at']),
        );
    }
}

DB::from('table')->get()->mapInto(A::class, 'fromDbRow');
```

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -383,11 +383,16 @@ trait EnumeratesValues
      * @template TMapIntoValue
      *
      * @param  class-string<TMapIntoValue>  $class
+     * @param  string|null  $method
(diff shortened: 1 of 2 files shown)


ekvedaras:
Realised you can just use `->map([A::class, 'fromDbRow'])` with feels same. Closeing.

Issue #42031, opened 2022-04-18
Grouping the result of a query by an enum casted attribute


C14r:
- Laravel Version: 9.8.1
- PHP Version: 8.1.4
- Database Driver & Version: mysql & 10.4.13-MariaDB

### Description:

I have a Customer model, whicht casts the status attribute to an php 8.1 enum.

```php

namespace App\Models;

use App\Enums\CustomerStatus;

class Customer extends \Illuminate\Database\Eloquent\Model
{
    protected $casts = [
        'status' => CustomerStatus::class,
    ];
}
```

and I'm trying to group the result of a query by status

```php
Customer::all()->groupBy('status'));
```

I'm getting the following error: `array_key_exists(): Argument #1 ($key) must be a valid array offset type` in `..vendor\laravel\framework\src\Illuminate\Collections\Collection.php: 497`.

```php
if (! array_key_exists($groupKey, $results)) {
    $results[$groupKey] = new static;
}
```

driesvints:
Heya, thanks for reporting.

We'll need more info and/or code to debug this further. Can you please create a repository with the command below, commit the code that reproduces the issue as separate commits on the main/master branch and share the repository here? Please make sure that you have the [latest version of the Laravel installer](https://github.com/laravel/installer) in order to run this command. Please also make sure you have both Git & [the GitHub CLI tool](https://cli.github.com) properly set up.

```bash
laravel new bug-report --github="--public"
```

Please do not amend and create a separate commit with your custom changes. After you've posted the repository, we'll try to reproduce the issue.

Thanks!

C14r:
Hey, thanks for the answer.

If created the repository `C14r/bug-report` with the laravel/installer 4.2.10 and made a commit with everything to reproduce the bug.

### Steps To Reproduce:

- gh repo clone C14r/bug-report
- php artisan migrate --seed
- php artisan serve
- start the browser and call http://localhost:8000/

agus24:
i think the problem within `valueRetriever` of `EnumeratesValues` trait it should be check if the value is enum or not. cmiiw.
```
    protected function valueRetriever($value)
    {
        if ($this->useAsCallable($value)) {
            return $value;
        }

        return function ($item) use ($value) {
            $outputValue = data_get($item, $value);
            if ($outputValue instanceof UnitEnum) {
                return $outputValue->value;
            }

            return $outputValue;
        };
    }
```

driesvints:
[...]

PR #42043, opened 2022-04-19
add support enum parameters for groupby and keyby

agus24:
this pull request add enum support for `groupBy` and `keyBy` based on issue #42031
this would be breaking for php 8.0
any suggestions?

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -5,6 +5,7 @@ namespace Illuminate\Support\Traits;
 use CachingIterator;
 use Closure;
 use Exception;
+use UnitEnum;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Contracts\Support\Jsonable;
 use Illuminate\Support\Arr;
@@ -1071,7 +1072,12 @@ trait EnumeratesValues
         }
 
         return function ($item) use ($value) {
-            return data_get($item, $value);
+            $outputValue = data_get($item, $value);
+            if ($outputValue instanceof UnitEnum) {
+                return $outputValue->value;
+            }
+
+            return $outputValue;
         };
     }
 
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -5,7 +5,6 @@ namespace Illuminate\Support\Traits;
 use CachingIterator;
 use Closure;
 use Exception;
-use UnitEnum;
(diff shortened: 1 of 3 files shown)


BrandonSurowiec:
Look at this test for enums:
https://github.com/laravel/framework/pull/39315/files

You can conditionally include an Enums.php file with your declared enums, and then skip the tests for lower versions of php.

```
public function testGroupByWithEnumValue($collection)
{
        if (version_compare(PHP_VERSION, '8.1') < 0) {
            $this->markTestSkipped('PHP 8.1 is required');
        } 
```

driesvints:
This is a new feature I think, not a bug fix.

agus24:
> Look at this test for enums: https://github.com/laravel/framework/pull/39315/files
> 
> You can conditionally include an Enums.php file with your declared enums, and then skip the tests for lower versions of php.
> 
> ```
> public function testGroupByWithEnumValue($collection)
> {
>         if (version_compare(PHP_VERSION, '8.1') < 0) {
>             $this->markTestSkipped('PHP 8.1 is required');
>         } 
> ```
> 
> Note: It seems like this should target 8.x as enum support was first added there and this appears to be a bug fix.

is this good enough? or should i conditionally include a file?

BrandonSurowiec:
@agus24 Looks good to me, tests are passing :)

taylorotwell:

agus24:
@taylorotwell i think it's a bug fix since we cant directly use groupby to enum field from database.
for example
```php
enum CustomerStatus: string
{
    case Bussines = 'bussines';
    case Private = 'private';
}
```
```php
class Customer {
    protected $casts = [
    ];
}
```

then when we do 
```
Customer::all()->groupBy('status')
``` 

it will cause an error `array_key_exists(): Argument #1 ($key) must be a valid array offset type`

so we need to workaround something like `Customer::all()->groupBy('status.value')`
this PR will make groupBy and keyBy to directly without using the `.value` suffix

sjorsvanleeuwen:
> This is a new feature I think, not a bug fix.

How is this a feature request and not a bug fix? When using enums on models you are no longer able to groupby that column, that is a feature break in my book. This PR actually fixes it with minimal code. Hoping it can be merged soon.

BrandonSurowiec:
@agus24 You can try resubmitting this PR. Just include the full explanation of why it is needed when you submit it. (Feel free to use reasons that other people mentioned in this thread plus your full explanation after it was closed.)

You'll have to resubmit this to target 9.x and it will probably be accepted there. I do think it should be considered a bug fix for 8.x as ENUM support was expanded a bunch in 8.x.

However, target 9.x and maybe mention that you think it belongs in 8.x and see what happens.

PR #42051, opened 2022-04-19
Make the order ascending parameter case insensitive

nyamsprod:
When using the collection sorting method the following is not supported and default to a descending ordering

```php
$result = LazyCollection::make($csv)
    ->sortBy([['annee', 'ASC'], ['nombre', 'ASC']])
    ->slice(3, 2)
```

The current PR enable support for case insensitve `asc` value. 

Of note, `desc` is already case insensitive as the current code only check for the scrict lowecased `asc` value. Any other value is treated as `desc`.


taylorotwell:
I suggest just passing lower case.

PR #42052, opened 2022-04-19
[9.x] Provide a method to store a cached result permanently, periodically refreshing it

27pchrisl:
I'm proposing this addition as a feature to meet the following use case:

Where there is a service such as an HTTP endpoint that our application relies on caching a result for, but the remote service is unreliable or slow. In this case we don't want to clear the cached result after a TTL in case we can't reliably replace it, causing our application to have problems. We do want to try periodically refreshing the result, but if we can't then we prefer to keep the cached version.

This feature caches the result permanently, alongside a separate expiring value. When the expiring value expires we immediately mark the expiring value as valid again, and re-run the closure. The closure may safely fail until the TTL expires again.

A bonus advantage of using this technique is to prevent a stampede of workers all attempting to re-cache an expired value.

### DIFF
--- a/src/Illuminate/Cache/Repository.php
+++ b/src/Illuminate/Cache/Repository.php
@@ -396,18 +396,18 @@ class Repository implements ArrayAccess, CacheContract
     /**
      * Get an item from the cache, or execute the given Closure and refresh the result when the TTL expires.
      *
-     * @param  string                                              $key
+     * @param  string  $key
      * @param  \Closure|\DateTimeInterface|\DateInterval|int|null  $ttl
-     * @param  Closure                                             $callback
+     * @param  Closure  $callback
      * @return mixed
      */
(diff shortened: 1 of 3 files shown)


taylorotwell:
I'm not sure I'm following the differences between this and the `remember` method.

27pchrisl:
If you `remember` for 30 seconds, the cached result is gone after 30 seconds and your callback may not be able to restore it immediately (if the remote service is unreliable).

You can `rememberForever`, but then there's not a built-in option to periodically refresh if the cached value can become outdated. This function sits between these two options.

taylorotwell:
I'm not totally in love with the naming here. `refresh` sounds like it's going to refresh every N seconds automatically or something. For now, until we can determine better naming, I suggest implementing this yourself in your own application behind some repository class or something.

PR #42057, opened 2022-04-20
[9.x] Fixes return type of request function

DriverCat:
Adds \Illuminate\Contracts\Foundation\Application|\Illuminate\Http\Request to PHPDoc in order IDE to suggest autocompletion, for example: ``` request()->ip()```

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -636,7 +636,7 @@ if (! function_exists('request')) {
      *
      * @param  array|string|null  $key
      * @param  mixed  $default
-     * @return mixed
+     * @return mixed|\Illuminate\Contracts\Foundation\Application|\Illuminate\Http\Request
      */
     function request($key = null, $default = null)
     {


driesvints:
Mixed can never be combined with something else. It already involves "everything".

DriverCat:
> Mixed can never be combined with something else. It already involves "everything".

I agree here in case of function return type declaration, but isn't PHPDoc for reflecting actual behaviour of function? 
I didn't find any references in PHP-FIG standards where it is not allowed to use so. Even PHPStorm doesnt "complain" about using this. And we have similar cases in ``` src/Illuminate/Foundation/helpers.php ```, for example:

```
 /**
     * Get / set the specified session value.
     *
     * If an array is passed as the key, we will assume you want to set an array of values.
     *
     * @param  array|string|null  $key
     * @param  mixed  $default
     * @return mixed|\Illuminate\Session\Store|\Illuminate\Session\SessionManager
     */
    function session($key = null, $default = null)
```

PR #42059, opened 2022-04-20
[8.0] Fix SFTP file system in queue or long running tasks

joelharkes:
Making the file system scoped makes sure that the file system gets cleaned up and reset after a queue job.
This forces for example the SFTP connection to close and reset so that you don't experience: `Message: Connection closed prematurely`

Possibly adds overhead for users who are using the queue extensively with cloud filesystems as they might need to reconnect on every Job. But it will make the framework behave more consistently.

There is only 1 major problem: `Storage::` facade is globally cached and wont be refreshed, so you can't use this or this should also be cleared with:

```
        Storage::clearResolvedInstances();
```

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemServiceProvider.php
+++ b/src/Illuminate/Filesystem/FilesystemServiceProvider.php
@@ -25,7 +25,7 @@ class FilesystemServiceProvider extends ServiceProvider
      */
     protected function registerNativeFilesystem()
     {
-        $this->app->singleton('files', function () {
+        $this->app->scoped('files', function () {
             return new Filesystem;
         });
     }
@@ -39,11 +39,11 @@ class FilesystemServiceProvider extends ServiceProvider
     {
         $this->registerManager();
 
-        $this->app->singleton('filesystem.disk', function ($app) {
+        $this->app->scoped('filesystem.disk', function ($app) {
             return $app['filesystem']->disk($this->getDefaultDriver());
         });
 
-        $this->app->singleton('filesystem.cloud', function ($app) {
+        $this->app->scoped('filesystem.cloud', function ($app) {
             return $app['filesystem']->disk($this->getCloudDriver());
         });
     }
@@ -55,7 +55,7 @@ class FilesystemServiceProvider extends ServiceProvider
      */
     protected function registerManager()
     {
-        $this->app->singleton('filesystem', function ($app) {
+        $this->app->scoped('filesystem', function ($app) {
             return new FilesystemManager($app);
         });
     }


taylorotwell:
I wonder if the fix we could make specifically for SFTP that isn't quite such a big change for all other drivers?

iBotPeaches:
This is a good find, since I don't believe `scoped` existed when I ran into this awhile back. I believe you are experiencing this in the same way I was since if you use SFTP, you are probably using Flysystem. If you are using Flysystem you are probably using the [sftp-v3 driver](https://github.com/thephpleague/flysystem-sftp-v3/blob/2.x/composer.json)

Which uses [phpseclib](https://github.com/phpseclib/phpseclib) under the hood, which the SFTP Class wraps SSH2 which only closes the connection on destruction of the class, which only happens in either `queue:listen` mode or when a  timeout occurs - https://github.com/phpseclib/phpseclib/blob/master/phpseclib/Net/SSH2.php#L3167

joelharkes:
We actually fixed it by forgetting the sftp driver after each job and our other long running Cron scheduler (custom build)

driesvints:
@joelharkes if the PR isn't needed anymore can you please close it?

ikudosi:
> We actually fixed it by forgetting the sftp driver after each job and our other long running Cron scheduler (custom build)

Can you explain how you did this? I am running into this issue where I am extensively using queues to fetch files via SFTP.

ericsssan:
I am experiencing the same problem. I have a job dispatched to Horizon queue to upload csv file via SFTP. Only successfully uploaded SOME of the times. Very dodgy. 

> We actually fixed it by forgetting the sftp driver after each job and our other long running Cron scheduler (custom build)

@joelharkes how do you "forgetting" the sftp driver?

ericsssan:
Using scoped method doesn't work. Have to resort to using reflection to get access to sftpConnectionProvider and connection object and call disconnect method manually after put file.

joelharkes:
@ericsssan sorry for the late reply:

run:
```php
Storage::purge('sftp');
```

for each sftp disk you want to purge.
You can do in the `AppServiceProvider`

```

        $this->app->terminating(function () {
            BudgetChangeRecorder::flushChanges();
            Storage::purge('sftp');
        });

        \Illuminate\Support\Facades\Queue::after(function () {
            BudgetChangeRecorder::flushChanges();
            Storage::purge('sftp');
        });
```

Im not sure if this is still needed in Laravel 9?

joelharkes:
@ikudosi  see comment above

ericsssan:
> 

Thanks @joelharkes

PR #42060, opened 2022-04-20
[9.x] Make password rule errors translatable

reziamini:
Right now, the only way to translate password errors is to set messages into a JSON file.
And we can't use lang/{language}/validation.php in our system to translate error messages.

With this change, we will be able to use PHP files to use translation.

For example, we can change the lang/en/validation.php file into:
```php

return [
    'password' => [
         'mixedCase' => 'The :attribute must contain at least one uppercase and one lowercase letter.',
         'letters' => 'The :attribute must contain at least one letter.',
         'symbols' => 'The :attribute must contain at least one symbol.',
         'numbers' => 'The :attribute must contain at least one number.',
         'uncompromised' => 'The given :attribute has appeared in a data leak. Please choose a different :attribute.',
    ],
];
```

These changes do not break any old changes, and old users still can use JSON files to translate this message.

If it will be merged, I'll add these messages to `lang/en/validation.php` in laravel/laravel repo.

And we don't need to create `en.json` file anymore.

### DIFF
--- a/src/Illuminate/Validation/Rules/Password.php
+++ b/src/Illuminate/Validation/Rules/Password.php
@@ -305,7 +305,7 @@ class Password implements Rule, DataAwareRule, ValidatorAwareRule
             if ($this->mixedCase && ! preg_match('/(\p{Ll}+.*\p{Lu})|(\p{Lu}+.*\p{Ll})/u', $value)) {
                 $validator->errors()->add(
                     $attribute,
-                    $this->getErrorMessage('validation.password.mixedCase')
+                    $this->getErrorMessage('validation.password.mixed')
                 );
             }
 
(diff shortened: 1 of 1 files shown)


src/Illuminate/Validation/Rules/Password.php
```diff
@@ -361,4 +371,27 @@ protected function fail($messages)
...
+            'validation.password.symbols' => 'The :attribute must contain at least one symbol.',
+            'validation.password.numbers' => 'The :attribute must contain at least one number.',
+            'validation.password.uncompromised' => 'The given :attribute has appeared in a data leak. Please choose a different :attribute.',
+        ];
+
+        if (($message = $this->validator->getTranslator()->get($key)) !== $key) {
+            return $message;
+        }
+
+        return $messages[$key];
```
seriquynh, line 395: What about directly return instead of defining $messages variable on the top of this method?
```php
return [
    'validation.password.mixedCase' => 'The :attribute must contain at least one uppercase and one lowercase letter.',
    'validation.password.letters' => 'The :attribute must contain at least one letter.',
    'validation.password.symbols' => 'The :attribute must contain at least one symbol.',
    'validation.password.numbers' => 'The :attribute must contain at least one number.',
    'validation.password.uncompromised' => 'The given :attribute has appeared in a data leak. Please choose a different :attribute.',
][$key];
```

src/Illuminate/Validation/Rules/Password.php
```diff
@@ -303,19 +303,31 @@ public function passes($attribute, $value)
             $value = (string) $value;
 
             if ($this->mixedCase && ! preg_match('/(\p{Ll}+.*\p{Lu})|(\p{Lu}+.*\p{Ll})/u', $value)) {
-                $validator->errors()->add($attribute, 'The :attribute must contain at least one uppercase and one lowercase letter.');
+                $validator->errors()->add(
```
seriquynh, line 306: I think Laravel team prefers putting parameters in a single line like this.
```
<?php

$validator->errors()->add(
    $attribute, $this->getErrorMessage('validation.password.mixedCase')
);
```

seriquynh: I do too.

PR #42065, opened 2022-04-20
[9.x] Fix TrimStrings middleware with non-UTF8 characters

MrMicky-FR:
Currently, the `TrimStrings` middleware sets to `null` all requests inputs with a unicode character not in UTF-8, which can be really frustrating to debug

It can be reproduced by using a `POST` request with a body like `foo=abcd%E9`: the `foo` input will just be `null` if the `TrimString` middleware is used.

This issue was introduced in #40600, because `preg_replace` with `u` flag will return `null` for a non-UTF8 input (with `Malformed UTF-8 characters, possibly incorrectly encoded` as `preg_last_error_msg()`)

As a fix, when `preg_replace` return a `null` value, it's simply pass the value to the PHP `trim` function, which was the behavior used by Laravel for years. This doesn’t change the current behavior for valid UTF-8 inputs.

### DIFF
--- a/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
+++ b/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
@@ -49,11 +49,11 @@ class TrimStrings extends TransformsRequest
      */
     protected function transform($key, $value)
     {
-        if (in_array($key, $this->except, true)) {
+        if (in_array($key, $this->except, true) || ! is_string($value)) {
             return $value;
         }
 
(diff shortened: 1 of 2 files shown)


MrMicky-FR:
No sure why the tests are failing, as the failing test (`Mockery\Exception\NoMatchingExpectationException` in `ThrottlesExceptionsWithRedisTest`) seems unrelated to the changes in this PR ? And the tests was passing for the commit on my fork: https://github.com/MrMicky-FR/laravel-framework/commit/5076a1da69bc465797ff6ebe7957aca369df0f9f

PR #42066, opened 2022-04-20
[9.x] Prevent direct invocation of compiled files under docroot

will0:
When the template compiler stores .php files, many setups allow those .php files to be invoked by external traffic. This may be provably non-problematic here beyond errors in logs, but at least [one managed host](https://wpengine.com/support/wp-engines-security-environment/#Disk_Write_Protection) prevents writes to .php files at runtime because of the large number of vulnerabilities in the category.  Many WordPress themes using the https://roots.io/sage/ starter theme encounter this issue because of the popularity of this component.

This change could be reasonably targeted at 8.x rather than 9.x, but the patch is not literally the same. Happy to retarget.

### DIFF
--- a/src/Illuminate/View/Compilers/Compiler.php
+++ b/src/Illuminate/View/Compilers/Compiler.php
@@ -67,7 +67,7 @@ abstract class Compiler
      */
     public function getCompiledPath($path)
     {
-        return $this->cachePath.'/'.sha1('v2'.Str::after($path, $this->basePath)).'.php';
+        return $this->cachePath.'/'.sha1('v2'.Str::after($path, $this->basePath)).'.compiled';
     }
 
     /**
--- a/tests/View/ViewBladeCompilerTest.php
+++ b/tests/View/ViewBladeCompilerTest.php
@@ -18,7 +18,7 @@ class ViewBladeCompilerTest extends TestCase
     public function testIsExpiredReturnsTrueIfCompiledFileDoesntExist()
     {
         $compiler = new BladeCompiler($files = $this->getFiles(), __DIR__);
-        $files->shouldReceive('exists')->once()->with(__DIR__.'/'.sha1('v2foo').'.php')->andReturn(false);
+        $files->shouldReceive('exists')->once()->with(__DIR__.'/'.sha1('v2foo').'.compiled')->andReturn(false);
         $this->assertTrue($compiler->isExpired('foo'));
     }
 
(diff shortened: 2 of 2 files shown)


Krisell:
I don't quite see why these php files, located in `storage/framework/views` would be exposed externally, however if they are and the extension is changed, wouldn't that expose the raw contents of the files instead which could contain sensitive information.

In addition, if `storage/framework/views` are served, chances are other even more sensitive files are also served. Do you have an example environment/setup where you are facing the problem you describe?

I'm also curious how OpCache would work with other extensions, but perhaps that doesn't matter as long as the file is included? I did test quickly that using `require` on a non-php file did in fact work and execute the code.

will0:
I'll get an example set up to share, but in the meantime offer some more text :)

What happens is a WordPress extension author creates a plugin or theme, and bundles in the templating system in a `vendor` subdirectory of the extension. WordPress is deployed as a bunch of files under the document root, placing assets under `/wp-content/themes/my-theme/...`. Mostly these php files are not externally accessed in the plugin design, but are often enough that it cannot be blanket denied.

taylorotwell:
Hey @will0 - I'm a bit nervous to change this for existing applications. What do you think if I added a configuration option for the compiled view extension so that people using this type of setup on WP Engine could just tweak that one configuration setting in their config/view.php file?

GrahamCampbell:
Setups that have their storage directory mounted on a publicly accessible server have their stuff configured in an insecure way. There is no reason to configure things in that way. I'm happy to work with with whatever open source project has this misconfigured to help them find the best way to configure this in a secure way - reach out to me please.

taylorotwell:
@GrahamCampbell the problem in question is some hosts (WP Engine it seems) prevent ANY writes that write files with a ".php" extension.

will0:
Yes @GrahamCampbell @taylorotwell I don't mean to raise this as a security issue, but a compatibility issue with a platform security measure. @taylorotwell 's point about existing applications makes sense to me. I'll have to expand my understanding of the framework/codebase to make a reasonable submission along those lines. Is that directionally agreeable? If so, my turnaround time may be a few weeks - is that a new submission or iterating within this PR?

taylorotwell:
@will0 if you are ok with the configuration option - I can probably handle that on my end for you.

will0:
@taylorotwell configuration would absolutely allow downstream or end users to solve the problem, and I can't but agree with you on responsible change management.

taylorotwell:
https://github.com/laravel/framework/commit/68e41fd3691b9aa5548e07c5caf38696c4082513

Done.

will0:
Thanks so much, finding that last file in your change would have taken me a while :)

PR #42072, opened 2022-04-21
Feature/add save and refresh to eloquent model

alexanderkroneis:
# Type: Improvement

This pull request adds `saveAndRefresh` to `Model` by using `save()` and `refresh()` of the model itself.

## Use Case
I use a lot of actions to encapsulate my application logic, when I assign a relation to a model I need to save it and then manually refresh it. It would be easier to use `saveAndRefresh()` instead of the example below.

```
<?php

namespace Domain\Project\Actions;

use Domain\Customer\Models\Customer;
use Domain\Project\Models\Project;

class AssignCustomerToProjectAction
{
    public function execute(Project $project, Customer|int $model)
    {
        $project->customer()->associate($model);
        $project->save();

        return $project->refresh();
    }
}
```

It's fully backward compatible since it only uses already existing functions of `Model`.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -23,8 +23,7 @@ use Illuminate\Support\Traits\ForwardsCalls;
 use JsonSerializable;
 use LogicException;
 
-abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToString, HasBroadcastChannel, Jsonable,
-                                JsonSerializable, QueueableEntity, UrlRoutable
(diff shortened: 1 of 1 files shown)


ankurk91:
```php
return $project->save()->refresh();
```

You can already do this

ralphjsmit:
> ```
> return $project->save()->refresh();
> ```

Are you sure? I thought `->save()` returns a boolean.

BrandonSurowiec:
You can use `tap()`.

```php
return tap($project)->save()->refresh();
```

https://laravel.com/docs/8.x/helpers#method-tap

alexgaal:
I think this PR can be closed due to `tap`. Thank you @BrandonSurowiec for this need aspect of Laravel.

PR #35714, opened 2020-12-25
[8.x] Fix OPTIONS method bug with use same path and diff domain when cache route

MilesChou:
When set the same path and diff domain route and cache route, the first route with OPTIONS method will return 404.

## reproduce step

* PHP version: 7.4
* Laravel version: v8.5.5
* Laravel Framework version: v8.20.1

1. edit `routes/web.php` file:

```php
Route::domain('foo.localhost')->get('/login', 'FooController@index'); // Will return 404 when use OPTIONS method
Route::domain('bar.localhost')->get('/login', 'BarController@index');
```

Notice: same path, but different domain.

Route list:

```
$ php artisan route:list
+---------------+----------+-------+------+------------------------------------------+------------+
| Domain        | Method   | URI   | Name | Action                                   | Middleware |
+---------------+----------+-------+------+------------------------------------------+------------+
|               | GET|HEAD | /     |      | Closure                                  | web        |
| foo.localhost | GET|HEAD | login |      | App\Http\Controllers\FooController@index | web        |
| bar.localhost | GET|HEAD | login |      | App\Http\Controllers\BarController@index | web        |
+---------------+----------+-------+------+------------------------------------------+------------+
```

2. cache route

```bash
php artisan route:cache
# or 
php artisan optimize
```

3. start server and curl route with OPTIONS method:

```
$ curl -v -X OPTIONS http://foo.localhost:8000/login
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to foo.localhost (127.0.0.1) port 8000 (#0)
> OPTIONS /login HTTP/1.1
> Host: foo.localhost:8000
> User-Agent: curl/7.64.1
> Accept: */*
>
* HTTP 1.0, assume close after body
< HTTP/1.0 404 Not Found
< Host: foo.localhost:8000
< Date: Fri, 25 Dec 2020 08:55:00 GMT
< Connection: close
< X-Powered-By: PHP/7.4.13
< Cache-Control: no-cache, private
< date: Fri, 25 Dec 2020 08:55:00 GMT
< Content-type: text/html; charset=UTF-8
<
<!DOCTYPE html>
<html lang="en">
    <head>
        <meta charset="utf-8">
        <meta name="viewport" content="width=device-width, initial-scale=1">

        <title>Not Found</title>

        <!-- Fonts -->
        <link rel="dns-prefetch" href="//fonts.gstatic.com">
        <link href="https://fonts.googleapis.com/css?family=Nunito" rel="stylesheet">

        <style>
[...]

PR #42078, opened 2022-04-21
[8.x] Fix array keys from cached routes

glamorous:
Introduced by https://github.com/laravel/framework/pull/35714.
Removed the slash to match the same array key as in
src/Illuminate/Routing/RouteCollection.php:60

We noticed a different result when we used route caching for in this code piece:

```
$url = parse_url($requestUri, PHP_URL_PATH);

 if (!is_string($url)) {
        return true;
 }

$routeKey = '{tenant}.' . trim(config('app.domain')) . trim($url, '/');

$routes = Route::getRoutes()->get($requestMethod);

return array_key_exists($routeKey, $routes);
```

Only routes with a specified domain where affected and a the result array of "->get('GET')" would not have the same array keys. Check next example with uri "test" and domain "{tenant}.laravel.test", this would give these two different results if you use cached router or not:

 "{tenant}.laravel.test/test" => cached
 "{tenant}.laravel.testtest" => not cached

The check for the existing of a domain is not necessary because the "->getDomain" function can return `null` as well and wil result in empty string (just like it behaves in the non compiled RouteCollection).

If approved & merged, I can cherry-pick it for 9.x release as well.

### DIFF
--- a/src/Illuminate/Routing/CompiledRouteCollection.php
+++ b/src/Illuminate/Routing/CompiledRouteCollection.php
@@ -252,11 +252,7 @@ class CompiledRouteCollection extends AbstractRouteCollection
             })
             ->map(function (Collection $routes) {
                 return $routes->mapWithKeys(function (Route $route) {
-                    if ($domain = $route->getDomain()) {
-                        return [$domain.'/'.$route->uri => $route];
-                    }
-
-                    return [$route->uri => $route];
+                    return [$route->getDomain().$route->uri => $route];
                 })->all();
             })
             ->all();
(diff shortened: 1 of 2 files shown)


driesvints:
I don't think this is valid? Seems like existing tests are breaking because of this PR.

driesvints:
It indeed seems that this is the correct fix. We'll need to adjust the tests and also version the routes file anew to make sure existing cached route files aren't conflicting with this change.

glamorous:
I'll try to find some time this weekend to check which tested needs an update

taylorotwell:
Can we get a test on this?

glamorous:
There was just a test that needed the fix as well to match the changed code. If merged, I can cherry-pick for 9.x

driesvints (approved):

driesvints:
@taylorotwell there were tests for this, they've now been changed to the new behavior 👍

PR #42080, opened 2022-04-21
[9.x] Copy locale and defaultLocale from original request

juukie:
When a `FormRequest` is created from the original request, the values `locale` and `defaultLocale` won't be copied over.
Hopefully this will be a correct pull request to solve the problem.


taylorotwell:
Those are protected properties?

juukie:
@taylorotwell Sorry those are protected indeed. I've changed it to getters 👍 .

PR #42081, opened 2022-04-21
[9.x] Add expressive, relative date methods to builder

jasonmccreary:
This adds expressive query builder methods for checking relative dates. Date math can be mind bending. Hopefully these continue to improve the developer experience for Laravel artisans by removing the need to think about the logic and pass the current date.

**Before**
```php
$previousPosts = Post::where('published_at', '<', now())->get();
$currentPosts = Post::whereDate('published_at', '=', now())->get();
$pendingPosts = Post::where('published_at', '>', now())->get();
```

**After**
```php
$previousPosts = Post::wherePast('published_at')->get();
$currentPosts = Post::whereToday('published_at')->get();
$pendingPosts = Post::whereFuture('published_at')->get();
```

**Note:** Currently the implementation uses Carbon to create the date value. This could be changed to use a _raw_ expression, such as `NOW()`. However, this may increase the complexity of the code given the varying RDBMS support for such methods.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1312,7 +1312,7 @@ class Builder implements BuilderContract
     {
         $type = 'Basic';
         $operator = $not ? '>=' : '<';
-        $value = Carbon::now()->format('Y-m-d h:i:s');
+        $value = Carbon::now()->format('Y-m-d H:i:s');
 
         foreach (Arr::wrap($columns) as $column) {
             $this->wheres[] = compact('type', 'column', 'boolean', 'operator', 'value');
@@ -1420,7 +1420,7 @@ class Builder implements BuilderContract
     {
         $type = 'Basic';
         $operator = $not ? '<=' : '>';
-        $value = Carbon::now()->format('Y-m-d h:i:s');
+        $value = Carbon::now()->format('Y-m-d H:i:s');
 
         foreach (Arr::wrap($columns) as $column) {
             $this->wheres[] = compact('type', 'column', 'boolean', 'operator', 'value');
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1452,7 +1452,7 @@ class Builder implements BuilderContract
      */
     public function orWhereFuture($columns, $now = null)
     {
-        return $this->whereFuture($columns, $now , 'or');
+        return $this->whereFuture($columns, $now, 'or');
     }
 
     /**
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1304,15 +1304,20 @@ class Builder implements BuilderContract
      * Add a where clause to determine if "date" column is in the past to the query.
      *
      * @param  string|array  $columns
+     * @param  \DateTimeInterface|string|null  $now
      * @param  string  $boolean
      * @param  bool  $not
      * @return $this
      */
-    public function wherePast($columns, $boolean = 'and', $not = false)
+    public function wherePast($columns, $now = null, $boolean = 'and', $not = false)
     {
         $type = 'Basic';
         $operator = $not ? '>=' : '<';
-        $value = Carbon::now()->format('Y-m-d H:i:s.u');
+        $value = $now ?? Carbon::now();
+
+        if ($value instanceof DateTimeInterface) {
+            $value = $value->format('Y-m-d H:i:s.u');
+        }
 
         foreach (Arr::wrap($columns) as $column) {
             $this->wheres[] = compact('type', 'column', 'boolean', 'operator', 'value');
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Query/Builder.php
```diff
@@ -1299,6 +1300,169 @@ public function orWhereNotNull($column)
...
+     * @param  string|array  $columns
+     * @param  string  $boolean
+     * @param  bool  $not
+     * @return $this
+     */
+    public function wherePast($columns, $boolean = 'and', $not = false)
+    {
+        $type = 'Basic';
+        $operator = $not ? '>=' : '<';
+        $value = Carbon::now()->format('Y-m-d h:i:s');
```
adevade, line 1315: `Y-m-d h:i:s` should be `Y-m-d H:i:s` with a capital H, I believe.

Small h is 12 hour clock: https://www.php.net/manual/en/datetime.format.php

src/Illuminate/Database/Query/Builder.php
```diff
@@ -1299,6 +1300,169 @@ public function orWhereNotNull($column)
...
+     * @param  string|array  $columns
+     * @param  string  $boolean
+     * @param  bool  $not
+     * @return $this
+     */
+    public function whereFuture($columns, $boolean = 'and', $not = false)
+    {
+        $type = 'Basic';
+        $operator = $not ? '<=' : '>';
+        $value = Carbon::now()->format('Y-m-d h:i:s');
```
adevade, line 1423: Same here. I think this should be 24 hour clock, using `Y-m-d H:i:s`.

src/Illuminate/Database/Query/Builder.php
jasonmccreary: Thanks. Just pushed a patch which should fix it. As well as add some test variance.

src/Illuminate/Database/Query/Builder.php
```diff
@@ -1299,6 +1300,169 @@ public function orWhereNotNull($column)
...
+     * @param  string|array  $columns
+     * @param  string  $boolean
+     * @param  bool  $not
+     * @return $this
+     */
+    public function wherePast($columns, $boolean = 'and', $not = false)
+    {
+        $type = 'Basic';
+        $operator = $not ? '>=' : '<';
+        $value = Carbon::now()->format('Y-m-d H:i:s');
```
juancho48, line 1315: Do you need the format(...?

jasonmccreary: No, not for a full date. But I like the explicitness. Plus, it is needed for `whereToday`.

jasonmccreary:
Some points that have been brought up outside of GitHub include:

- Microsecond precision
- Timezone support

The former could be solved by simply adding microseconds (`.u`) the "now" timestamp for `wherePast` and `whereFuture`. However, the latter is not so simple. Timezone could depend on a lot of things (app settings, DB settings, individual column values, etc). As such, the easiest way I could think of supporting timezones would be to allow the user to pass in a timezone to these methods. Use a default of the app settings to hopefully keep these methods streamlined for most developers.

x7ryan:
@jasonmccreary While watching you stream building this I thought a more generic comparison like `whereDateBefore()` or `whereDateAfter()` or even `whereDateBetween()` would be useful.

I think it would make sense to PR that and I almost did but thinking about it I think that and this would be too similar because as far as I can tell, unless I'm missing something, the only difference between that and this is that would require an explicit datetime for comparison and this uses an implied now.

Maybe it would make sense to roll both up into one set of methods where the datetime is optional and defaults to now? What do you think?

jasonmccreary:
@x7ryan, yes, I think that may be the best path. It would allow me to avoid the "timezone issue" if a user could pass in the date which represents _now_.

I'll make that tweak over the weekend.

imanghafoori1:
```php
$previousPosts = Post::whereInPast('published_at')->get();
$previousPosts = Post::whereInPastHours('published_at', 24)->get();

$pendingPosts = Post::whereInFuture('published_at')->get();
$pendingPosts = Post::whereInFutureHours('published_at', 24)->get();
```

jasonmccreary:
I've added microsecond precision for `wherePast` and `whereFuture`. I also allow passing in an optional `now` argument to control the "current timestamp". This should provide ultimate flexibility for columns which may not match the configured timezone or to use these as more readable date comparison methods.

This is satisfies my original vision for this PR. If others are interested in a `whereDateBetween`, `whereInFutureHours`, etc, they can PR them separately. 👍

bert-w:
what about creating a Trait instead of adding code to an already 3700+ line file? I see it hasnt been done much in the past for the `Illuminate\Database\Query\Builder` file but it would be nice to do so, especially since the new functions are somewhat grouped. `Illuminate\Database\Concerns\QueriesExtendedDates` perhaps?

DarkGhostHunter:
> what about creating a Trait instead of adding code to an already 3700+ line file? I see it hasnt been done much in the past for the `Illuminate\Database\Query\Builder` file but it would be nice to do so, especially since the new functions are somewhat grouped. `Illuminate\Database\Concerns\QueriesExtendedDates` perhaps?

Could be done using a separate PR, but dunno if would be breaking.

DarkGhostHunter:
Let me throw a monkey wrench here and ask... How its handles timezones?

driesvints:
@DarkGhostHunter it's discussed in the comments

PR #42088, opened 2022-04-22
[9.x] Re-order messages variable in Password::getErrorMessage() method

oliverquynh:
In Password::getErrorMessage() method, the $messages variable is only necessary if developers don't define their custom messages in locale files. So, it shouldn't be initialized on the top of this method block.

### DIFF
--- a/src/Illuminate/Validation/Rules/Password.php
+++ b/src/Illuminate/Validation/Rules/Password.php
@@ -300,8 +300,6 @@ class Password implements Rule, DataAwareRule, ValidatorAwareRule
                 return;
             }
 
-            $value = (string) $value;
-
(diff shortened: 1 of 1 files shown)


src/Illuminate/Validation/Rules/Password.php
```diff
@@ -300,8 +300,6 @@ public function passes($attribute, $value)
                 return;
             }
 
-            $value = (string) $value;
```
seriquynh, line 303: This cast statement is useless. Because after the first return of the check ! is_string($value), $value is absolutely a string.

seriquynh:
A failure test isn't related this change.

PR #36035, opened 2021-01-25
[8.x] Add Route::missing()

hotmeteor:
This PR introduces a new `->missing()` method to `Route`.

Here's the intention...

Let's say you have a route that's using implicit route-model binding (in this case, bound to a particular attribute):
```php
Route::get('/locations/{location:slug}', [LocationsController::class, 'show'])
    ->name('locations.view');
```
Now, if that record doesn't exist, or has been removed, it'll throw a 404 if someone tries to access it. Currently to mitigate that you'd need to customize the controller method to check for it's existence and then handle it accordingly.

Instead, it would be great if there was an `->missing()` method that accepted a callback and handled the situation, ie:
```php
Route::get('/locations/{location:slug}', [LocationsController::class, 'show'])
    ->name('locations.view')
    ->missing(fn($request) => Redirect::route('locations.index', null, 301));
```

To do this we'd now be catching a `ModelNotFoundException` in the `SubstituteBindings` middleware, but this seems acceptable since it'll only be doing this if it detects an `missing` fallback.

Open to suggestions!

PR #42092, opened 2022-04-22
[9.x] Add a `findOr` method to Eloquent

jessarcher:
This PR adds a `findOr()` method to the Eloquent Builder and Relations. This matches the existing `firstOr()` methods.

A typical scenario where this method is useful is when a resource route (e.g. `/posts/1`) cannot find a model that is *not* part of the URL, and so a `findOrFail()` returning a 404 is not appropriate.

Example usage:

```php
User::findOr(1, fn () => throw new RuntimeException);
User::findOr(1, fn () => abort(403));
User::findOr(1, fn () => 'return something');
```

The method also supports passing columns as the second argument:

```php
User::findOr(1, ['columns'], fn () => '...');
```

It also works on `HasMany`, `HasManyThrough`, `BelongsToMany`, `MorphMany`, and `MorphToMany` relations:

```php
$user->posts()->findOr(1, fn () => '...');
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -499,6 +499,29 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Find a model by its primary key or call a callback.
+     *
+     * @param  mixed  $id
+     * @param  \Closure|array  $columns
+     * @param  \Closure|null  $callback
+     * @return \Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Collection|static[]|static|mixed
+     */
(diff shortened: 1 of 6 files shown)


BrandonSurowiec:
Neat idea. Has some overlap use-case wise with `Route::missing()` that was added in 8.x: https://github.com/laravel/framework/pull/36035

mreduar:
Nice!

novalramdhani:
Great feature!!!

Technoyer:
This method adds some flexibility instead of findOrFail() we can now handle the fail output using whatever callback.
@jessarcher congratulations for the great work.

PR #42094, opened 2022-04-22
[9.x] Add new ucwords string helper

hebinet:
This PR adds a new ucwords string helper.

Usage

```php
Str:ucwords('laravel php framework')
Str::of('laravel php framework')->ucwords();
str('laravel php framework')->ucwords();
```
`// Outputs: "Laravel Php Framework"`

Why do I think this is usefull?
Lately I had a Stringable class and modified its value like this:

```php
$string = str($name)->lower();
$string = ucwords($string);
$parts = str($string)->ucsplit();
```
It would have been much "sexier" to just use the fluent interface like this.
```php
$parts = str($name)->lower()->ucwords()->ucsplit();
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -1028,6 +1028,18 @@ class Str
     }
 
     /**
+     * Uppercase the first character of each word in a string.
+     *
+     * @param  string  $string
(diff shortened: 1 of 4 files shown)


taylorotwell:
I suggest using `Str::title`

Issue #42075, opened 2022-04-21
'whereHas' using wrong table name when relationship refers to self


tiagof:
- Laravel Version: 9.1.0
- PHP Version: 8.1
- Database Driver & Version: Mysql 

### Description:

When using `whereHas` on a Model that implements SoftDeletes and has a relationship to itself, the generated query is wrong.

### Steps To Reproduce:
Create 2 models like this

```php
class Client extends Model {
  use SoftDeletes;
}

class Contract extends Model {
  use SoftDeletes;

  public function relationSelf(): HasMany
  {
      return $this->hasMany(static::class)->onlyTrashed();
  }

  public function relationOther(): HasMany
   {
      return $this->hasMany(Client::class)->onlyTrashed();
   }
}
```

````php
➜ art --version
Laravel Framework 9.1.0

➜ art tinker
Psy Shell v0.11.1 (PHP 8.1.4 — cli) by Justin Hileman

>>> Contract::whereHas('relationSelf')->toSql();
[!] Aliasing 'Contract' to 'Domains\Contracts\Models\Contract' for this Tinker session.
=> "select * from `contracts` where exists (select * from `contracts` as `laravel_reserved_0` where `contracts`.`id` = `laravel_reserved_0`.`contract_id` and `contracts`.`deleted_at` is not null) and `contracts`.`deleted_at` is null"

>>> Contract::whereHas('relationOther')->toSql();
=> "select * from `contracts` where exists (select * from `clients` where `contracts`.`id` = `clients`.`contract_id` and `clients`.`deleted_at` is not null) and `contracts`.`deleted_at` is null"
>>>
````

Notice that in the first instruction, we have `contracts.deleted_at is not null` when I believe we should get `laravel_reserved_0.deleted_at is not null`.
The complete correct sql query would be
````select * from `contracts` where exists (select * from `contracts` as `laravel_reserved_0` where `contracts`.`id` = `laravel_reserved_0`.`contract_id` and `laravel_reserved_0`.`deleted_at` is not null) and `contracts`.`deleted_at` is null````

In the second instruction, because this relationship refers to another Model, the query is generated correctly.

driesvints:
Can you try the very latest Laravel version? This is already fixed I think.

tiagof:
Don't think it is ...

````php
site on  upg/laravel9 [!] via ⬢ v17.7.2 via 🐘 v8.1.4 took 11s
➜ art --version
Laravel Framework 9.9.0

[...]

PR #42100, opened 2022-04-22
[9.x] Add wherehas soft deleting scopes

tiagof:
An attempt to fix/Improve mergeConstraintsFrom to use correct qualified column names.

Related issue --> https://github.com/laravel/framework/issues/42075

### DIFF
--- a/tests/Database/DatabaseEloquentSoftDeletesIntegrationTest.php
+++ b/tests/Database/DatabaseEloquentSoftDeletesIntegrationTest.php
@@ -44,6 +44,7 @@ class DatabaseEloquentSoftDeletesIntegrationTest extends TestCase
     {
         $this->schema()->create('users', function ($table) {
             $table->increments('id');
+            $table->integer('user_id')->nullable(); // circular reference to parent User
             $table->integer('group_id')->nullable();
             $table->string('email')->unique();
             $table->timestamps();
@@ -878,6 +879,22 @@ class DatabaseEloquentSoftDeletesIntegrationTest extends TestCase
         $this->assertNull($comment->owner);
     }
 
+    public function testSelfReferencingRelationshipWithSoftDeletes()
+    {
+        /*
+         * https://github.com/laravel/framework/issues/42075
+         */
+        [$taylor, $abigail] = $this->createUsers();
+
+        $versionsA = $abigail->versions;
+        $versionsT = $taylor->versions;
+
+        self::assertCount(1, $versionsA);
+        self::assertTrue($versionsA->first()->is($taylor));
+
+        self::assertCount(0, $versionsT);
+    }
+
     /**
      * Helpers...
      *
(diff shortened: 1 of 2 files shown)


tests/Database/DatabaseEloquentSoftDeletesIntegrationTest.php
```diff
@@ -878,14 +879,32 @@ public function testMorphToNonSoftDeletingModel()
...
+        $versionsA = $abigail->versions;
+        $versionsT = $taylor->versions;
+
+        self::assertCount(1, $versionsA);
+        self::assertTrue($versionsA->first()->is($taylor));
+
+        self::assertCount(0, $versionsT);
+
+        self::assertEquals(1, SoftDeletesTestUser::whereHas('versions')->count());
+    }
```
BrandonSurowiec, line 898: Would match the code style of the rest of the tests (by using `$this` over `self::` and inline your variables `$versionsA` and `$versionsT`. More like this:

```php
    public function testSelfReferencingRelationshipWithSoftDeletes()
    {
        [$taylor, $abigail] = $this->createUsers();

        $this->assertCount(1, $abigail->versions);
        $this->assertTrue($abigail->versions->first()->is($taylor));

        $this->assertCount(0, $taylor->versions);
        $this->assertEquals(1, SoftDeletesTestUser::whereHas('versions')->count());
    }
```

There also may be a better name than `version` for this "test" relationship. Maybe `self_referencing_relationships()` to make the test relation match the test case, making it more clear?

    public function testSelfReferencingRelationshipWithSoftDeletes()
    {
        [$taylor, $abigail] = $this->createUsers();

        $this->assertCount(1, $abigail->self_referencing_relationships);
        $this->assertTrue($abigail->self_referencing_relationships->first()->is($taylor));

        $this->assertCount(0, $taylor->self_referencing_relationships);
        $this->assertEquals(1, SoftDeletesTestUser::whereHas('self_referencing_relationships')->count());
    }

tiagof:
Agree and done :)

it-can:
I suddenly get these errors after this commit

```
str_starts_with(): Argument #1 ($haystack) must be of type string, App\Models\Enums\ProductType given
```

Seems to be caused by this line, https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php#L786=

I have a product model, with a product_type with an enum. The problem appears when doing something like this


--
This is located in my Product model

```
/**
     * Get all sub products
     *
     * @return \Illuminate\Database\Eloquent\Relations\hasMany
     */
    public function simpleProducts()
    {
        return $this->hasMany(self::class, 'parent_id', 'id')
            ->where('active', 1)
            ->where('product_type', ProductType::SIMPLE)
            ->orderBy('sort');
    }
```

ProductType enum
```
namespace App\Models\Enums;

enum ProductType: string
{
    case SIMPLE = 'simple';
    case GROUPED = 'grouped';
    case CONFIGURABLE = 'configurable';
}
```

```
return Product::withCount('simpleProducts')
    ->get();
```

Calling this works correct....

```
return Product::where('product_type', ProductType::SIMPLE)
    ->get();
```

PR #42103, opened 2022-04-22
[9.x] Set the cause of lost connection

imanghafoori1:
Now that the #42102 is accepted, I think it is also needed to have the root cause of the losing connection in the report logs to properly debug the situation.
Since there is a huge list of various reasons mentioned in the `causedByLostConnection` method.
But currently, the framework is swallowing it.
```
            'server has gone away',
            'no connection to the server',
            'Lost connection',
            'is dead or not enabled',
            'Error while sending',
            'decryption failed or bad record mac',
            'server closed the connection unexpectedly',
            'SSL connection has been closed unexpectedly',
            'Error writing data to the connection',
       .........

```
- unfortunately, this change is not backward-compatible in theory since the signature of the method is changing and may cause trouble in case the method is overridden in a subclass.

- Callers of the `reconnect` method are not affected since the $message parameter has a default value.

### DIFF
--- a/src/Illuminate/Database/Concerns/ManagesTransactions.php
+++ b/src/Illuminate/Database/Concerns/ManagesTransactions.php
@@ -170,7 +170,7 @@ trait ManagesTransactions
     protected function handleBeginTransactionException(Throwable $e)
     {
         if ($this->causedByLostConnection($e)) {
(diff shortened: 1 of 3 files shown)


taylorotwell:
It's not backwards compatible.

PR #42104, opened 2022-04-22
[9.x] Added new email validation rule messages

SuperDJ:
Added Email validation rule. This makes it possible to give different error messages for different errors.

Using the following e-mailadress: `Кириfoö()@example.com`.
**Current behavior**
With the current email validation:
```php
'email' => 'email:rfc,strict,dns,spoof,filter'
```
When it fails it will only return
>The :attribute must be a valid email address.

**New behavior**
With the new email validation:
```php
'email' => [Email::rfc()->strict()->dns()->spoof()->filter()]
```
When it fails it wil return:
```php
[
       'The email must be a valid email address.',
       'The email does not have a valid domain.',
       'The email appears to be spoofed.',
       'The email must pass the filter.',
]
```

The behavior is taken from the [Password rule](https://laravel.com/docs/9.x/validation#validating-passwords).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -49,7 +49,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: ${{ matrix.php }}
-          extensions: dom, curl, libxml, mbstring, zip, pcntl, pdo, sqlite, pdo_sqlite, gd, redis-phpredis/phpredis@5.3.5, igbinary, msgpack, lzf, zstd, lz4, memcached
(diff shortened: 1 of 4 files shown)


SjorsO:
I think someone who fills in `Кириfoö()@example.com` as their email probably doesn't need a specific explanation to figure out why their email isn't considered valid.

driesvints:
I also think a generic message is probably enough. Also, there's now quite a bit of duplication between the new functionality and the validation trait.

SuperDJ:
@driesvints I agree that the functionality is the same. In the validation trait I wasn't able to find a way of adding multiple validation messages for the same validation rule. 

My main reason for this PR is not add a new validation rule but to have specific validation messages for each email validation rule option (rfc, strict, dns etc.). I would like this to give better validation information to our end users.

taylorotwell:

tbalint1995:
@SuperDJ 
Where is the Email class located in Laravel 9, as I understand there is no such class and methods like that?

PR #42105, opened 2022-04-23
[9.x] Allow objects to be passed to the database query builder

lkmadushan:
This pull request allows users to pass an object to the database query builder. This might be helpful to encapsulate a complex query logic into an object using predefined classes.

Something similar to: https://www.martinfowler.com/eaaCatalog/queryObject.html

We can implement similar functionality using Eloquent scope functions. But scopes are only available in Eloquent models. The intention of this approach is to move business logic from a query builder to a business class. (Create query by referring to classes and fields rather than tables and columns)

Example of filtering user records:

```php
DB::table('users')
    ->applyCriteria(new FilterCriteria(auth()->user(), new FilterData('taylor')))
    ->get()

// or

User::applyCriteria(new FilterCriteria(auth()->user(), new FilterData('taylor')))->get()

class FilterData {
    public $name;
}

class FilterCriteria implements Criteria
{
    public function __construct(
       protected User $user, 
       protected FilterData $filters
    ) {
    }

    public function apply($q)
    {
        $q->where('name', 'like', $this->filters->name.'%');
        
        if ($this->user->isAdmin()) {
            $q->applyCriteria(new VisibleToAdminCriteria($this->user));
            
            return $q;
        }
        
        return $q;
    }
}
```

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -3571,6 +3571,17 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Apply the criteria to a given query builder.
+     *
+     * @param  \Illuminate\Database\Query\Criteria  $criteria
+     * @return \Illuminate\Contracts\Database\Query\Builder
+     */
+    public function applyCriteria(Criteria $criteria)
+    {
+        return $criteria->apply($this);
+    }
+
+    /**
      * Get a scalar type value from an unknown type of input.
      *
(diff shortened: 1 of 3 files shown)


dennisprudlo:
I think that's what scopes are supposed to be

lkmadushan:
> I think that's what scopes are supposed to be

@dennisprudlo I have updated the description with my concerns.

taylorotwell:

PR #42118, opened 2022-04-24
[9.x] Add str() and string() to request object

cwilby:
This PR adds `str()` and `string()` methods to the Request class.

## Usage

```php
// use Illuminate\Http\Request;
public function index(Request $request)
{
	// instead of 
	$names = str($request->input('name'))->explode(',');

	// we can write
	$names = $request->str('name')->explode(',');
}
```

## Real world example

```php
// use App\Models\User;
// use Illuminate\Http\Request;

public function index(Request $request)
{
	User::query()->when($request->has('filters'), function ($query) use ($request) {
		
		// instead of
		$filters = $request->input('filters');

		foreach (explode(',' $filters) as $filter) {
			[$column, $value] = str($filter)->explode(':');

			$query->where($this->sanitize($column), $value);
		}

		// we can write
		$request->str('filters')->explode(',')->each(function ($filter) use ($query) {
			[$column, $value] = str($filter)->explode(':');

			$query->where($this->sanitize($column), $value);
		});

	});
}
```

It's a bit shorter, and an expressive way to interact with `Illuminate\Support\Stringable` from a request.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -542,4 +542,28 @@ trait InteractsWithInput
 
         return $this;
     }
+
+    /**
+     * Retrieve input from the request as a stringable.
+     *
+     * @param  string  $key
+     * @param  mixed  $default
+     * @return \Illuminate\Support\Stringable
+     */
+    public function str($key, $default = null)
+    {
+        return $this->string($key, $default);
+    }
+
+    /**
+     * Retrieve input from the request as a stringable.
+     *
+     * @param  string  $key
+     * @param  mixed  $default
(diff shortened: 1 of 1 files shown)


cwilby:
@BrandonSurowiec Thank you!  I've updated the "Real Life" example.

BrandonSurowiec:
@cwilby Hah. I didn't want any new developer to copy and paste your nice looking code and use it blindly. 😉

taylorotwell:
Merged. Dunno why it's acting like I didn't.

shaffe-fr:
IMO, str should return a Stringable instance (make sense because of the str helper), but string should return an actual string.
I know it's maybe to late for the change, but I think it really make more sense and avoid some string casting.

PR #42132, opened 2022-04-25
[9.x] Add incrementQuietly and decrementQuietly database methods

jonhassall:
Add incrementQuietly and decrementQuietly database methods to complement the existing saveQuietly method.

saveQuietly updates a model without triggering events, and is documented here:
https://laravel.com/docs/9.x/eloquent#saving-a-single-model-without-events

Complementary incrementQuietly and decrementQuietly methods were not available so I have added them. Also added tests.

Does not break any existing features.


jonhassall:
I made this pull request that was accepted. However it doesn't appear to be in the release.

PR #42133, opened 2022-04-26
[9.x] Allow jsonSerialize() to return other types

mikewuu:
This PR updates `jsonSerialize()`'s return type to `mixed`, matching PHP's definition. 

## Use case

Allows models to override `jsonSerialize`, and serialize into strings rather than always having to be an JSON object.

## Example

We have a simple model, `Permission`, who's only attribute is `name`:

```php
class Permission extends Model
{
    protected $fillable = [
        'name',
    ];
}
```

When we fetch all permissions, the collection is automatically serialized into JSON objects.

### Controller

```php
class PermissionsController
{
  public function list()
  {
     return Permission::all();
  }
}
```

Would currently return

```
[
  {name: "foo"},
  {name: "bar"},
  {name: "baz"},
]
```

With this change, `Permission`, can override `jsonSerialize`, and return a `string`. 

Updated response after this PR:

```
["foo", "bar", "baz"]
```

### DIFF
--- a/src/Illuminate/Bus/Queueable.php
+++ b/src/Illuminate/Bus/Queueable.php
@@ -192,6 +192,32 @@ trait Queueable
     }
 
     /**
+     * Prepend a job to the current chain so that it is run after the currently running job.
+     *
+     * @param  mixed  $job
+     * @return $this
+     */
+    public function prependToChain($job)
+    {
+        $this->chained = Arr::prepend($this->chained, $this->serializeJob($job));
+
+        return $this;
+    }
+
+    /**
+     * Append a job to the end of the current chain.
+     *
+     * @param  mixed  $job
+     * @return $this
+     */
+    public function appendToChain($job)
+    {
+        $this->chained = array_merge($this->chained, [$this->serializeJob($job)]);
+
+        return $this;
+    }
+
+    /**
      * Serialize a job for queuing.
      *
      * @param  mixed  $job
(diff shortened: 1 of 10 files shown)


driesvints:
Ping @X-Coder264 this is what I meant on that pr your sent in ^

GrahamCampbell (changes requested):
This change looks wrong to me. Having an implementation return a more specific type than that required by the interface is totally OK, and actually desirable in this case. https://3v4l.org/CYkJl

mike3u:
Updated it to be a union type with `string` instead. `JsonSerializable` only specifies that the return type has to be compatible with `json_encode`, which a `string` is.

dennisprudlo:
With `array|string` it won't be possible to override `jsonSerialize` and return an integer, a float or a boolean, right? `json_encode`'s $value can have any type but a resource.

driesvints:
@X-Coder264 my ping was more about you being okay with this change?

X-Coder264:
@driesvints There's nothing intrinsically wrong with this change. From my perspective it's a bit weird for a collection (which is just arrays on steroids) to be JSON serialized as anything else but an array. But if people want to do that I guess there's no reason from a framework perspective to block that so on that level I'm fine with the proposed change.

mike3u:
@X-Coder264 

> it's a bit weird for a collection (which is just arrays on steroids) to be JSON serialized as anything else but an array.

It's actually the Collection item's `jsonSerialize` that I'm hoping to override. The collection can (and should?) still expect an array to be returned.

I've updated it to still return an `array` for collection in: https://github.com/laravel/framework/pull/42133/commits

driesvints:
If we do this then I suggest we move to the mixed return type like the PHP interface and not bother with union types.

driesvints (approved):

PR #42130, opened 2022-04-25
[9.x] Fix HasAttributes::mutateAttributeForArray when accessing non-cached attribute

aijorgenson:
Currently there is no check for `hasAttributeGetMutator` in the `HasAttributes::mutateAttributeForArray` method. 

This causes serializing a model to an array to fail when an attribute is specified in the appends array, as it falls back trying to retrieve the value via `mutateAttribute`. `mutateAttribute` utilizes the previous standard `getFirstNameAttribute` vs the new `firstName(): Attribute` approach referred to in 
https://laravel.com/docs/9.x/eloquent-mutators.

PR #42140, opened 2022-04-26
[9.x] Add integration test for PR #42130

aijorgenson:
Adding integration test for https://github.com/laravel/framework/pull/42130

@taylorotwell When writing the test, I found out its only for attributes using camel case and an append specified. The fix still seems to work great, but maybe this was a bandaid over a deeper underlying problem?

### DIFF
--- a/src/Illuminate/Bus/Queueable.php
+++ b/src/Illuminate/Bus/Queueable.php
@@ -192,6 +192,32 @@ trait Queueable
     }
 
     /**
+     * Prepend a job to the current chain so that it is run after the currently running job.
+     *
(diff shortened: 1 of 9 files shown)


taylorotwell:
Hmm, in this case I may revert the original PR in case it causes other problems. I do note that the documentation on appending uses snake case as well, since attributes are usually referenced in snake case.

taylorotwell:
https://laravel.com/docs/9.x/eloquent-serialization#appending-values-to-json

aijorgenson:
Oh, my bad! Thank you for catching that

PR #42147, opened 2022-04-27
Upgrade PHPStan to 1.6.1

kayw-geek:
Upgrading PHPStan to 1.6.x can reduce memory usage by 50-70%
https://phpstan.org/blog/phpstan-1-6-0-with-conditional-return-types#:~:text=Lower%20memory%20consumption
opening-file-in-an-editor It can quickly locate problems in PHPStorm.
https://phpstan.org/user-guide/output-format#opening-file-in-an-editor

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -90,7 +90,7 @@
         "mockery/mockery": "^1.4.4",
         "orchestra/testbench-core": "^7.1",
         "pda/pheanstalk": "^4.0",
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
This is not necessary. We already will be installing 1.6.1.

driesvints:
This is unnecessary and not everyone uses PhpStorm. Thanks

PR #33294, opened 2020-06-21
[7.x] Remove misreading of method parameters

restuff:
Swap confusing method parameters names for alias() method. An alias actually comes first, and real class name second. The logical order of arguments remains the same.

PR #42149, opened 2022-04-27
[9.x] Update Container.php

robotsniper:
Some parameter names are semantically inaccurate

We should correct the naming to increase code readability

For example:

For the `alias()` method as below

It's easy to think that `$alias` stands for alias name

But `$abstract` actually stands for alias name

The reading experience was terrible, So we need to correct it

```php
/**
     * Alias a type to a different name.
     *
     * @param  string  $abstract
     * @param  string  $alias
     * @return void
     *
     * @throws \LogicException
     */
    public function alias($abstract, $alias)
    {
        if ($alias === $abstract) {
            throw new LogicException("[{$abstract}] is aliased to itself.");
        }

        $this->aliases[$alias] = $abstract;

        $this->abstractAliases[$abstract][] = $alias;
    }
```

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -59,14 +59,14 @@ class Container implements ArrayAccess, ContainerContract
     protected $scopedInstances = [];
 
     /**
-     * The registered type aliases.
+     * The registered abstract type aliases.
      *
      * @var string[]
      */
     protected $aliases = [];
 
     /**
-     * The registered aliases keyed by the abstract name.
+     * The registered abstract type keyed by the alias name.
      *
      * @var array[]
      */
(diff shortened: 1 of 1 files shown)


src/Illuminate/Container/Container.php
```diff
@@ -548,21 +548,21 @@ public function tagged($tag)
...
      *
-     * @param  string  $abstract
      * @param  string  $alias
+     * @param  string  $abstract
      * @return void
      *
      * @throws \LogicException
      */
-    public function alias($abstract, $alias)
+    public function alias($alias, $abstract)
```
GrahamCampbell, line 557: Didn't we recently swap these around?

GrahamCampbell (changes requested):

src/Illuminate/Container/Container.php
robotsniper: There are?  
Where can I see the changes you are talking about ？

BrandonSurowiec: Good memory, but the `Container.php` has always been in this order since the first commit. That was the AliasLoader: https://github.com/laravel/framework/pull/33294

If you change the parameter order here, you'll have to make the same change in the Container Contract. https://github.com/laravel/framework/blob/9.x/src/Illuminate/Contracts/Container/Container.php#L27

PR #42160, opened 2022-04-28
Fixes helpers request return PHPDoc

Coloynle:
Adds \Illuminate\Http\Request to request() in PHPDoc return type.
because i need my IDE to suggest autocompletion.
eg:  request()->ip()

Reference documentation:
[PHPDoc Definition of a 'Type'](https://docs.phpdoc.org/3.0/guide/references/phpdoc/types.html#multiple-types)
Many tags use a Type as part of their definition (such as the @return tag). These types differ from the official PHP definition to be able to represent all kinds of data.
Please note that mixed is also a single type and with this keyword it is possible to indicate that each array element contains any possible type.

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -636,7 +636,7 @@ if (! function_exists('request')) {
      *
      * @param  array|string|null  $key
      * @param  mixed  $default
-     * @return mixed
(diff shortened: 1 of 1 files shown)


driesvints:
Mixed can never be combined with something else.

Coloynle:
> Mixed can never be combined with something else.

but this is PHPDoc

PR #42004, opened 2022-04-16
[9.x] Fix pushMiddlewareToGroup not working on Provider in My Package

ilyutkin:
- Laravel Version: 9.8.1
- PHP Version: 8.0.5
- Database Driver & Version: mariadb

### Description:
In continuation of bug https://github.com/laravel/framework/issues/36604. I found the reason for this bug.
We want to dynamically connect Middleware in our package in the Service Provider:

    public function boot(Router $router)
    {
        // Push middleware to web group
        $router->pushMiddlewareToGroup('web', MyMiddleware::class);
    }

But the Middleware is not executed if the package name starts with the letters a-k.

### Reason:

In the /Illuminate/Foundation/Http/Kernel.php file, the syncMiddlewareToRouter method overwrites the Middleware list in the Router

    protected function syncMiddlewareToRouter()
    {
        $this->router->middlewarePriority = $this->middlewarePriority;

        foreach ($this->middlewareGroups as $key => $middleware) {
            $this->router->middlewareGroup($key, $middleware);
        }

        foreach ($this->routeMiddleware as $key => $middleware) {
            $this->router->aliasMiddleware($key, $middleware);
        }
    }

In the /Illuminate/Routing/Router.php file the middlewareGroup method

    public function middlewareGroup($name, array $middleware)
    {
        $this->middlewareGroups[$name] = $middleware;

        return $this;
    }

The syncMiddlewareToRouter method is called in prependToMiddlewarePriority which calls /laravel/sanctum/src/SanctumServiceProvider.php after my ServiceProvider

Service providers are named in alphabetical order. If the package name starts with letters a-k Middleware Group is overwritten. If the package name begins with the letters m-z, everything works.

### Solution:

If rewriting the middleware group is not required, then you can change the middlewareGroup method in /Illuminate/Routing/Router.php

    public function middlewareGroup($name, array $middlewares)
    {
        foreach ($middlewares as $middleware)
            $this->pushMiddlewareToGroup($name, $middleware);

        return $this;
    }

Please add to the framework if I'm right.

PR #42161, opened 2022-04-28
[9.x] Ensure middleware group exists

### LINKED ISSUE #42159: Change in middlewareGroup function behaviour introduces a breaking change in v9.10.0
- Laravel Version: 9.10.0
- PHP Version: 8.1.3

### Description:

Since the `v9.10.0`, I encounter many issues on a large project. I can only say "many issues" for now, because it's "just" failling tests here and there that doesn't seem related to each other, even if everything works fine when I roll back to `v9.9.0`. 

I'm still in the process of digging the problem and to figure out why and when it happens.

However, I've determined this commit  https://github.com/laravel/framework/commit/05365f8186044af51fc55fb1e2b5b97caaac59e8 (part of `v9.10.0`) is the cause of these problems. If I checkout on it, it still doesn't work, the moment I manually revert the changes in the `middlewareGroup` public function, it works. With this reverted change, `v9.10.0` works fine too.

In the previous version, prior to https://github.com/laravel/framework/commit/05365f8186044af51fc55fb1e2b5b97caaac59e8, the entire middleware group was replaced by a new set of middlewares. In the new version, after https://github.com/laravel/framework/commit/05365f8186044af51fc55fb1e2b5b97caaac59e8, only the missing ones are added. At least it's what I understand for now.

Even if it's a small change, it's a breaking change that should be reverted.

What do you think?

### Steps To Reproduce:

Not yet but the change in behaviour is clear enough to be discussed.

rodrigopedra:
Closes #42159 

After PR #42004 the behavior of method `Illuminate\Routing\Router@middlewareGroup` was slightly changed, as before it would define a middleware group even if the group was empty, e.g. the group had an empty array of middleware.

After the aforementioned PR, a middleware group with an empty array does not actually add a corresponding middleware group to the stack, thus when running a route with that empty middleware group, the dispatcher would try to instantiate a middleware named as the group which will fail, as reported in #42159 .

This PR:

- Adds a new protected method named `ensureMiddlewareGroup`, which will check if a middleware group is defined, and if not will add it.
- Modify the methods: `middlewareGroup`, `prependMiddlewareToGroup` and `pushMiddlewareToGroup` to use the method above

To reproduce the error described in issue #42159 one could use the following steps:

1. Create a new laravel app
2. Ensure it uses Laravel version 9.10.0
3. Add an empty middleware group to the app's HTTP kernel (for example, named `public`)
4. Define a route that uses the middleware defined as above
5. Access the defined route through the browser

After applying the changes proposed in this PR, the route will work.

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -946,6 +946,19 @@ class Router implements BindingRegistrar, RegistrarContract
     }
 
     /**
+     * Ensure a middleware group is defined.
+     *
+     * @param  string  $name
+     * @return void
+     */
+    protected function ensureMiddlewareGroup($name)
+    {
+        if (! $this->hasMiddlewareGroup($name)) {
+            $this->middlewareGroups[$name] = [];
+        }
+    }
+
+    /**
      * Register a group of middleware.
      *
      * @param  string  $name
@@ -954,6 +967,8 @@ class Router implements BindingRegistrar, RegistrarContract
      */
     public function middlewareGroup($name, array $middleware)
     {
+        $this->ensureMiddlewareGroup($name);
+
         foreach ($middleware as $m) {
             $this->pushMiddlewareToGroup($name, $m);
         }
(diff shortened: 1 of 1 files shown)


toyi:
I'm afraid it doesn't solves https://github.com/laravel/framework/issues/42159 because a middleware group is still not entirely replaced by using `middlewareGroup`, which was the case before https://github.com/laravel/framework/pull/42004:

```php
public function middlewareGroup($name, array $middleware)
    {
        $this->middlewareGroups[$name] = $middleware;

        return $this;
    }
```
_Notice that `$middleware` is an array even if it's singular, that typo was fixed in https://github.com/laravel/framework/pull/42004_
Allowed a package to litterally replace, let's say, the `api` group, by another set of middlewares.

taylorotwell:
Reverting the original PR.

PR #42164, opened 2022-04-28
[9.x] Add default value to RedisManager connections

vvval:
Everywhere in the codebase RedisManager's `$connections` property is used as an array. But it doesn't even have a default value (so it's null). The getter `getConnections()` has an array return-type hint, but still this method may return uninitialised null value

### DIFF
--- a/src/Illuminate/Redis/RedisManager.php
+++ b/src/Illuminate/Redis/RedisManager.php
@@ -47,9 +47,9 @@ class RedisManager implements Factory
     /**
      * The Redis connections.
      *
-     * @var mixed
+     * @var array
      */
-    protected $connections;
+    protected $connections = [];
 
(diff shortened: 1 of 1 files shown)


vvval:
If approved, I'd like to create the same PR to `8.x` branch

laradocs:
Hi @vvval 

PHP8 version was using in Laravel 9.x, you can define hints by type.

driesvints:
@laradocs we don't use property type hints anywhere in the framework atm.

vvval:
Hi, @laradocs
I'd love to, but this is gonna be a BC-break. Also, none of the other properties have a type-hint, so at least now the code-style looks consistent

taylorotwell:
The code isn't broken atm.

PR #42170, opened 2022-04-28
[9.x] Assert redirect parameters

ohnotnow:
This PR adds a couple of extra assertions to the test response to check a query parameter in a redirect response is set to a given value or is null.

My reason for this is I have a few places where I do something like redirect to :

`https://example.com/?s=123&e=23iud8u13249w9djsdf`

Where the `e=...` will be a random-ish value (or at least not one I can be sure of).  At the moment I can manually do the test, or macro it in.  But I mentioned to a few people on Discord and they seemed to think it was something they'd wanted to do as well.

So this allows :
```
$response = $this->get('https://page-that-redirects-with/?s=123&e=baz');

$response->assertRedirectQueryParameterNotNull('e');
$response->assertRedirectQueryParameterEquals('123', 's');
```

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -333,10 +333,11 @@ EOF;
     /**
      * Assert whether the redirect response has a query parameter with a given value
      *
+     * @param  mixed   $value
      * @param  string  $parameterName
      * @return $this
      */
-    public function assertRedirectQueryParameterEquals($parameterName, $value)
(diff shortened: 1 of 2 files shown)


taylorotwell:
I find this methods to be a bit of a mouthful. Maybe there is a shorter name? Otherwise I suggest just macroing them into your project.

ohnotnow:
Yeah - the names felt a bit long to me too.  No worries - I was happy to macro them in before, just a few other people said 'yeah - I'd use that'.  But I'll package them up another way.

PR #42177, opened 2022-04-29
fix: Incorrect mask position

lscar:
- fix Str::mask Incorrect mask position due to repeated substrings


lscar:
https://github.com/laravel/lumen-framework/issues/1235

For example, want to use "*" to mask the 5th to 8th characters of "123412341234".

"Illuminate\Support\Str::mask('123412341234', '*', 4, 4)" output "****12341234"

PR #42184, opened 2022-04-29
[9.x] Added `has` and `missing` methods to ValidatedInput

Sammyjo20:
This PR introduces two new methods to the `Illuminate\Support\ValidatedInput` class.

- `has($keys)`
- `missing($keys)`

These two methods are used to check the existence of validated input easily. For example, now with the `$request->safe()` method, I could now do:

```php
$validatedName = $request->safe()->has('name'); // True
$validatedAge = $request->safe()->has('age') // False

$validatedName = $request->safe()->missing('name'); // False
$validatedAge = $request->safe()->missing('age') // True
```

You can also provide an array, and it will validate each key...

```php
$validatedName = $request->safe()->has(['name', 'age']);
```

### DIFF
--- a/src/Illuminate/Support/ValidatedInput.php
+++ b/src/Illuminate/Support/ValidatedInput.php
@@ -103,7 +103,7 @@ class ValidatedInput implements ValidatedData
     /**
      * Check if the validated input contains one or many keys.
      *
-     * @param  array|mixed  $keys
+     * @param  mixed  $keys
      * @return bool
      */
     public function has($keys)
--- a/src/Illuminate/Support/ValidatedInput.php
+++ b/src/Illuminate/Support/ValidatedInput.php
@@ -103,7 +103,7 @@ class ValidatedInput implements ValidatedData
     /**
      * Check if the validated input contains one or many keys.
      *
-     * @param $keys
+     * @param  array|mixed  $keys
      * @return bool
      */
     public function has($keys)
@@ -122,7 +122,7 @@ class ValidatedInput implements ValidatedData
     /**
      * Check if the validated input does not contain one or many keys.
      *
-     * @param $keys
+     * @param  array|mixed  $keys
      * @return bool
      */
     public function doesntHave($keys)
--- a/src/Illuminate/Support/ValidatedInput.php
+++ b/src/Illuminate/Support/ValidatedInput.php
@@ -101,7 +101,7 @@ class ValidatedInput implements ValidatedData
     }
 
     /**
-     * Check if the validated input contains one or many keys.
+     * Determine if the validated input has one or more keys.
      *
      * @param  mixed  $keys
      * @return bool
@@ -120,12 +120,12 @@ class ValidatedInput implements ValidatedData
     }
 
     /**
-     * Check if the validated input does not contain one or many keys.
+     * Determine if the validated input is missing one or more keys.
      *
      * @param  mixed  $keys
      * @return bool
      */
-    public function doesntHave($keys)
+    public function missing($keys)
     {
         return ! $this->has($keys);
     }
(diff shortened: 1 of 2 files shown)


src/Illuminate/Support/ValidatedInput.php
```diff
@@ -100,6 +100,36 @@ public function all()
         return $this->input;
     }
 
+    /**
+     * Check if the validated input contains one or many keys.
+     *
+     * @param $keys
```
driesvints, line 106: Can you update the docblocks so they match the rest of the framework?

Sammyjo20 (commented):
Self review complete

src/Illuminate/Support/ValidatedInput.php
```diff
@@ -100,6 +100,36 @@ public function all()
...
+        return true;
+    }
+
+    /**
+     * Check if the validated input does not contain one or many keys.
+     *
+     * @param  mixed  $keys
+     * @return bool
+     */
+    public function doesntHave($keys)
```
Neol3108, line 128: I feel like `missing` is more Laravel-ish

BrandonSurowiec: I had the same thought, `missing()` is used as the method name when interacting with a plain `request()` instance.

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Http/Concerns/InteractsWithInput.php#L220

Sammyjo20: I'll get that updated 👍🏼

Sammyjo20 (commented):
I have changed the `doesntHave` to `missing` as I do agree it sounds more like Laravel's language, so I will update the description of the PR.

Sammyjo20:
Hey guys, I'm just thinking - does this PR actually add value to the framework? I could have sworn I recently had a use-case where I wanted to check if something was validated, but surely if it was provided, it has to be validated. This does double check that it was provided AND valid, but not sure if that's really helping.

taylorotwell:
@Sammyjo20 one thing I notice is the there could be a slight behavior difference between request->has and validatedInput->has... you use array_key_exists while request->has uses Arr::has under the hood. Thoughts?

taylorotwell:
Please mark as ready for review after responding. Thanks.

Sammyjo20:
Ah yes, good spot. I have switched it to use `Arr::has()` like the `InteractsWithInput` trait has.

matiaslauriti:
May I ask why is Laravel not requiring a type-hinted return?

For example, the `has` and `missing` could easily have `: bool` at the end:
```php
public function has($key): bool

public function missing($key): bool
```

Why I am saying so? I see other tests and author's tests are using `assertEquals` for `booleans`, which means that if `has` or `missing` starts returning something that is not `true` or `false`, the tests would not fail nor the method itself would throw a `PHP Fatal` (allowing the Laravel team to fix this on a test).

```php
public function has($keys)
{
    return null; // Example
}

$this->assertEquals(false, $inputA->missing('name')); // This passes because false == null
```

The fix would be to start using `assertSame` (`valueA === valueB`) and type-hint the return value (because we can in this case and we are allowing PHP 8.0+ on composer for Laravel 9.x).
```php
public function has($keys): bool

$this->assertSame(false, $inputA->missing('name'));
```

So, is there a _why_ we are not doing so? I am asking so when I contribute in the near future I am already aware of this 👍

matiaslauriti:
@driesvints

Sammyjo20:
@matiaslauriti I was just keeping it consistent with the rest of the test. I assume it was written a while ago. I’d rather keep it consistent than make lots more changes to have to be reviewed. I agree though I prefer “assetTrue” and useful methods like that.

driesvints:
We don't use return types right now in the codebase anywhere. We have no plans right now to add any.

PR #42185, opened 2022-04-29
[9.x] Return success constant instead of 0

MortenDHansen:
I believe returning self::SUCCESS instead of 0 communicates better to the developer and helps to enforce a correct use of output integers (0 meaning 'good' instead of the php falsey 0).


GrahamCampbell (changes requested):
Thanks for the PR, but I don't think this will be merged. Please see previous attempts to make this change. :)

PR #42186, opened 2022-04-29
[9.x] Add assertJsonKeyEquals() to response assertions

reziamini:
There is a method in the JSON assertions called `assertPath` which in my opinion it is not clear for developers.
`assertJsonKeyEquals()` may be a good alias for this method to make it more readable and more understandable.
As a developer, when I saw `assertPath` first I didn't have any idea about this method.
I guess this alias has a better developer experience.


taylorotwell:
I don't think `assertPath` is really meant to be used by end users?

https://laravel.com/docs/9.x/http-tests#fluent-json-testing

PR #42194, opened 2022-04-29
[9.x] Add whereBetweenDate() method to filter two dates with a different formats

Haythamasalama:
PR: Add whereBetweenDate() method to filter two dates with different formats.

The use case to use the `whereBetweenDate()` to filter two dates with this format `Y-m-d -> 2022-04-29` and the column on the database is a timestamp or DateTime ` 2022-04-29 16:05:00 `  if using `whereBetween()` that will return inaccurate data because will match with two dates with different format to solve this need to cast the column on the database to data using function date.

### using `whereBetween()`

Query Builder : 

```php
 $checklists= DB::table('checklists')
            ->select('id','created_at')
            ->whereBetween('created_at', ['2022-04-17','2022-04-18'])
            ->get();
```

Eloquent : 
```php 
 $checklists= Checklist::select('id','created_at')
        ->whereBetween('created_at', ['2022-04-17','2022-04-18'])
        ->get()
```

SQL :
```sql 
SELECT 
   id ,created_at
FROM 
   checklists
where
   created_at between '2022-04-17' and '2022-04-18';   
```

Output : 

image

### using `whereBetweenDate()`

Query Builder : 
```php
 $checklists= DB::table('checklists')
        ->select('id','created_at')
        ->whereBetweenDate('created_at', ['2022-04-17','2022-04-18'])
        ->get();
```

Eloquent  : 
```php 
 $checklists= Checklist::select('id','created_at')
        ->whereBetweenDate('created_at', ['2022-04-17','2022-04-18'])
        ->get()
```

SQL : 
```sql 
SELECT 
  id ,created_at
FROM 
  checklists
where
  date(created_at) between '2022-04-17' and '2022-04-18';   
```

Output  : 
image

So I made this PR and added a new method `whereBetweenDate()` that casts the Timestamp or DateTime column to date and shows the exact values. `whereBetweenDate()` is useful when filtering data between two dates in different or the same formats.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1333,7 +1333,7 @@ class Builder implements BuilderContract
      */
     public function whereBetweenDate($column, iterable $values, $boolean = 'and')
     {
-        return $this->whereBetween( "date($column)", $values, $boolean);
+        return $this->whereBetween("date($column)", $values, $boolean);
     }
 
     /**
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1324,6 +1324,19 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Add a where between for two dates with different formats.
+     *
+     * @param  string  $column
+     * @param  iterable  $values
+     * @param  string  $boolean
+     * @return $this
+     */
+    public function whereBetweenDate($column, iterable $values, $boolean = 'and')
+    {
+        return $this->whereBetween( "date($column)", $values, $boolean);
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


dammy001 (commented):
Can you add to eloquent builder as well?

taylorotwell:
This bakes SQL grammar  (the `date` function) directly into the query builder, which is not our usual approach.

HaythamaSalama:
> This bakes SQL grammar (the `date` function) directly into the query builder, which is not our usual approach.

I think the new way more readable than passing `date` function SQL to `whereBetween()` like this

```php 
   DB::table('checklists')
    ->select('id','created_at')
    ->whereBetween(DB::raw('DATE(created_at)'),  ['2022-04-17','2022-04-18'])
    ->get();
```

PR #42197, opened 2022-04-30
[9.x] Adds Arr::join()

ecrmnn:
This PR adds `join` method from `Collections` to `Arr`.

```php
$stack = ['Tailwind', 'Alpine', 'Laravel', 'Livewire'];

Arr::join($stack, ', ', ' and ');

// Tailwind, Alpine, Laravel and Livewire
```


DarkGhostHunter:
Should the `Collection` use then `Arr::join()` instead?

PR #42209, opened 2022-05-01
[9.x] We can pass route names as well in except array

manadinho:
We can pass route names as well in except array in VerifyCsrfToken middleware.


driesvints:
Seems like this breaks the test suite. There's also a few unrelated style changes.

Issue #42204, opened 2022-04-30
Deprecation warning when trying to trans() with a null value in replacements


oprypkhantc:
- Laravel Version: 9.10.0
- PHP Version: 8.1.5

### Description:
Trying to feed `trans()` a second argument that is an array with a key and a null value results in a deprecation warning:
`local.WARNING: ErrorException: mb_substr(): Passing null to parameter #1 ($string) of type string is deprecated in /app/vendor/laravel/framework/src/Illuminate/Support/Str.php:941`

While that case isn't described in the Laravel documentation, it used to work and still does. I also believe it's reasonably common to occur, especially when just feeding the values from the database into the translator.

### Steps To Reproduce:

1. Configure the deprecations log as per docs
2. Run the following:
```php
trans('validation.after', [
	'date' => null,
]);
```
3. See a deprecation warning:
```log
[2022-04-30 20:42:31] local.WARNING: ErrorException: mb_substr(): Passing null to parameter #1 ($string) of type string is deprecated in /app/vendor/laravel/framework/src/Illuminate/Support/Str.php:941
Stack trace:
#0 /app/vendor/laravel/framework/src/Illuminate/Support/helpers.php(418): Illuminate\Foundation\Bootstrap\HandleExceptions->Illuminate\Foundation\Bootstrap\{closure}()
#1 /app/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(106): with()
#2 /app/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(71): Illuminate\Foundation\Bootstrap\HandleExceptions->handleDeprecation()
#3 /app/vendor/laravel/framework/src/Illuminate/Foundation/Bootstrap/HandleExceptions.php(229): Illuminate\Foundation\Bootstrap\HandleExceptions->handleError()
#4 [internal function]: Illuminate\Foundation\Bootstrap\HandleExceptions->Illuminate\Foundation\Bootstrap\{closure}()
#5 /app/vendor/laravel/framework/src/Illuminate/Support/Str.php(941): mb_substr()
#6 /app/vendor/laravel/framework/src/Illuminate/Support/Str.php(1011): Illuminate\Support\Str::substr()
#7 /app/vendor/laravel/framework/src/Illuminate/Translation/Translator.php(220): Illuminate\Support\Str::ucfirst()
[...]

PR #42216, opened 2022-05-02
[8.x] Fix deprecation issue with translator

driesvints:
This fixes an issue where a translation param was null and triggers a deprecation warning in PHP 8.1. On lower PHP versions, the parameter is simple added as an empty string.

See #42204

### DIFF
--- a/src/Illuminate/Translation/Translator.php
+++ b/src/Illuminate/Translation/Translator.php
@@ -218,8 +218,8 @@ class Translator extends NamespacedItemResolver implements TranslatorContract
         $shouldReplace = [];
 
         foreach ($replace as $key => $value) {
-            $shouldReplace[':'.Str::ucfirst($key)] = Str::ucfirst($value);
-            $shouldReplace[':'.Str::upper($key)] = Str::upper($value);
+            $shouldReplace[':'.Str::ucfirst($key ?? '')] = Str::ucfirst($value ?? '');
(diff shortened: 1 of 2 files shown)


src/Illuminate/Translation/Translator.php
```diff
@@ -218,8 +218,8 @@ protected function makeReplacements($line, array $replace)
         $shouldReplace = [];
 
         foreach ($replace as $key => $value) {
-            $shouldReplace[':'.Str::ucfirst($key)] = Str::ucfirst($value);
-            $shouldReplace[':'.Str::upper($key)] = Str::upper($value);
+            $shouldReplace[':'.Str::ucfirst($key ?? '')] = Str::ucfirst($value ?? '');
```
GrahamCampbell, line 221: I don't think keys can ever be null?

GrahamCampbell (changes requested):

src/Illuminate/Translation/Translator.php
driesvints: I only fixed the behavior that was previously in place. In any PHP version before 8.1, if `null` was used as a replacement it'd be casted to a string. Since PHP 8.1 throws a deprecation for this, we'll need to make sure to do this ourselves.

We can't change the behavior of handling `null` on 8.x or 9.x as that'd be a breaking change.

PR #42217, opened 2022-05-02
[8.x] Fix refresh during down

### LINKED ISSUE #42208: `php artisan down --refresh=5` does not work in combination with ` --render="maintenance.index"`
- Laravel Version: 9.9.0
- PHP Version: 8.1.5
- Database Driver & Version:

### Description:
`php artisan down --refresh=5` does not work in combination with ` --render="maintenance.index"`   

### Steps To Reproduce:

create a custom `maintenance.index` view

run `php artisan down --refresh=5 --render="maintenance.index"`

#### Expected behavior:
`maintenance.index` view will be rendered and refreshed every 5 seconds

#### Actual behavior:
`maintenance.index` view will be rendered but **NOT** refreshed every 5 seconds

I found this issue on [stackoverflow ](https://stackoverflow.com/questions/72004966/laravel-maintenance-mode-refresh-and-render-parameter)and proposed the below workaround

### current workaround:

Parsing the `--refresh=5` argument from `$_SERVER['argv']` if available and conditionally add  a refresh meta tag in the blade
```template
$result = array_values(preg_grep('/(--refresh)/', array_values( $_SERVER['argv'] ) ));
if(!empty($result)){
  $refresh = substr($result[0], strpos($result[0], "=") + 1);
}
```

```php
@if(isset($refresh))
   <meta http-equiv="refresh" content="{{$refresh}}">
@endif
```

driesvints:
This PR fixes an issue where the Refresh header was not set when an application is down with a specific template. There was a missing header at the bottom of the maintenance file.

Fixes #42208

### DIFF
--- a/src/Illuminate/Foundation/Console/stubs/maintenance-mode.stub
+++ b/src/Illuminate/Foundation/Console/stubs/maintenance-mode.stub
@@ -69,6 +69,10 @@ if (isset($data['retry'])) {
     header('Retry-After: '.$data['retry']);
 }
 
+if (isset($data['refresh'])) {
+    header('Refresh: '.$data['refresh']);
+}
+
 echo $data['template'];
 
 exit;


X-Coder264:
https://github.com/laravel/framework/blob/v9.10.1/src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php#L74

Shouldn't the `$this->getHeaders($data)` call already add this header?

```php
    /**
     * Get the headers that should be sent with the response.
     *
     * @param  array  $data
     * @return array
     */
    protected function getHeaders($data)
    {
        $headers = isset($data['retry']) ? ['Retry-After' => $data['retry']] : [];

        if (isset($data['refresh'])) {
            $headers['Refresh'] = $data['refresh'];
        }

        return $headers;
    }
```

driesvints:
@X-Coder264 maintenance mode only reaches that code if there's no custom template. If there's is, the request bubbles down to the lines I added.

X-Coder264:
Are you saying that this block which is supposed to be reached when there is a custom template is never reached, thus being dead code?

```php
            if (isset($data['template'])) {
                return response(
                    $data['template'],
                    $data['status'] ?? 503,
                    $this->getHeaders($data)
                );
            }
```

driesvints:
@X-Coder264 I think you misunderstand how this works. The `maintenance.php` file is checked first, see https://github.com/laravel/laravel/blob/9.x/public/index.php#L19. Only if all the conditions in that file fail, the request is propagated to the app, otherwise it bubbles down to the lines I changed.

X-Coder264:
Ah okay, yeah, I was missing that part, thanks for the explanation.

PR #42006, opened 2022-04-16
[9.x] Fixes return type of request function

DriverCat:
This PR offers changing return type of request function to ```mixed``` in order to avoid IDE warning on incompatible types. 

While making route model binding, for example:
```
Route::get('/{form}', [IndexController::class, 'show'])
    ->name('project.show');
```
and accessing it: ``` $form = request('form'); ``` IDE shows warning:

```
Incompatible types: Expected property of type '\App\Models\Form', 'array|\Illuminate\Contracts\Foundation\Application|\Illuminate\Http\Request|null|string' provided
```

PR #42219, opened 2022-05-02
[9.x] Revert "Fixes return type of request function (#42006)"

stasadev:
This PR returns autocompletion for `\Illuminate\Http\Request` in the `request()` helper function.

The addition of `mixed` to the previous return type complements the previous change in #42006.

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -636,7 +636,7 @@ if (! function_exists('request')) {
      *
      * @param  array|string|null  $key
      * @param  mixed  $default
-     * @return mixed
+     * @return mixed|\Illuminate\Http\Request|string|array|null
      */
     function request($key = null, $default = null)
     {


src/Illuminate/Foundation/helpers.php
```diff
@@ -636,7 +636,7 @@ function report($exception)
      *
      * @param  array|string|null  $key
      * @param  mixed  $default
-     * @return mixed
+     * @return mixed|\Illuminate\Http\Request|string|array|null
```
driesvints, line 639: `mixed` can never be combined with something else.

stasadev: Look in the same file, you will see literally four `mixed` in combination with other types:

- [@return mixed|\Illuminate\Contracts\Foundation\Application](https://github.com/laravel/framework/blob/ab1506091b9f166b312b3990d07b2e21d971f2e6/src/Illuminate/Foundation/helpers.php#L111)
- [@return mixed|\Illuminate\Cache\CacheManager](https://github.com/laravel/framework/blob/ab1506091b9f166b312b3990d07b2e21d971f2e6/src/Illuminate/Foundation/helpers.php#L229)
- [@return mixed|\Illuminate\Config\Repository](https://github.com/laravel/framework/blob/ab1506091b9f166b312b3990d07b2e21d971f2e6/src/Illuminate/Foundation/helpers.php#L263)
- [@return mixed|\Illuminate\Session\Store|\Illuminate\Session\SessionManager](https://github.com/laravel/framework/blob/ab1506091b9f166b312b3990d07b2e21d971f2e6/src/Illuminate/Foundation/helpers.php#L778)

GrahamCampbell: All those places are wrong, too.

stasadev: But isn't the purpose of PHPDoc to help the IDE do the autocomplete?
Because here `mixed` is not a type from PHP 8.0 and `barryvdh/laravel-ide-helper` doesn't handle such global functions.

driesvints: You're saying "everything but also this" which doesn't makes sense. I get the autocompletion part but from a documentation perspective this isn't correct.

driesvints: > barryvdh/laravel-ide-helper doesn't handle such global functions.

I think you can request a feature then on that library.

stasadev: Okay, I understand, it's completely wrong to write such hints on the one hand.

But there is no real standard: "You should never combine `mixed` with anything else in PHPDoc". This type hint is used even in the one of the [PSR packages](https://github.com/php-fig/http-message/blob/efd67d1dc14a7ef4fc4e518e7dee91c271d524e4/src/StreamInterface.php#L153).

In any case, are you okay with just reverting the changes without adding the `mixed` one? Because `request()` is one of the most commonly used helpers, and I don't want to lose autocomplete out of the box.

driesvints: Hey @stasadev. Let's let @taylorotwell decide on that.

PR #42220, opened 2022-05-02
[9.x] Make Validation exception messages translatable

schonhoff:
Hello Laravel team,

I'm currently testing my application after I upgraded to Laravel 9. I found a "new" behavior on the validation exception message. It included `(and x more errors)`. This confused me a lot because my applications validation errors where fully translated. I dug deeper into the issue and found the following code added to Laravel 9.

```php
protected static function summarize($validator)
{
  $messages = $validator->errors()->all();

  if (! count($messages)) {
    return 'The given data was invalid.';
  }

  $message = array_shift($messages);

  if ($additional = count($messages)) {
    $pluralized = $additional === 1 ? 'error' : 'errors';

    $message .= " (and {$additional} more {$pluralized})";
  }

  return $message;
}
```

I always thought about the `The given data was invalid.` part to be translatable but it didn't appeared in my application. Now the message is the first error with the addition of ` (and {$additional} more {$pluralized})` I got a mix of German and English in my validation exception messages that I return to my client. That isn't a nice experience.

With my PR I'm changing the message to be translatable in the singular and pluralized form. I changed the way the message is build because in my opinion it is much better to have 2 translatable sentences for this use case instead translating `error` and `errors` before including it in the translation as a parameter.

I really hope I did the right thing for the test part, because I never used the container part. I looked it up on the `tests/Validation/ValidationEnumRuleTest.php` class which only includes another translation.

### DIFF
--- a/src/Illuminate/Validation/ValidationException.php
+++ b/src/Illuminate/Validation/ValidationException.php
@@ -88,15 +88,14 @@ class ValidationException extends Exception
         $messages = $validator->errors()->all();
 
         if (! count($messages)) {
-            return 'The given data was invalid.';
+            return trans('The given data was invalid.');
         }
 
         $message = array_shift($messages);
-
         if ($additional = count($messages)) {
-            $pluralized = $additional === 1 ? 'error' : 'errors';
-
-            $message .= " (and {$additional} more {$pluralized})";
+            $message .= $additional === 1
+                ? trans(' (and :additional more error)', ['additional' => $additional])
+                : trans(' (and :additional more errors)', ['additional' => $additional]);
         }
 
         return $message;
--- a/tests/Validation/ValidationExceptionTest.php
+++ b/tests/Validation/ValidationExceptionTest.php
@@ -61,4 +61,13 @@ class ValidationExceptionTest extends TestCase
 
         return new ValidationException($validator);
     }
+
+    protected function tearDown(): void
+    {
+        Container::setInstance(null);
+
+        Facade::clearResolvedInstances();
+
+        Facade::setFacadeApplication(null);
+    }
 }
--- a/tests/Validation/ValidationExceptionTest.php
+++ b/tests/Validation/ValidationExceptionTest.php
@@ -2,9 +2,12 @@
 
 namespace Illuminate\Tests\Validation;
 
+use Illuminate\Container\Container;
+use Illuminate\Support\Facades\Facade;
 use Illuminate\Translation\ArrayLoader;
 use Illuminate\Translation\Translator;
 use Illuminate\Validation\ValidationException;
+use Illuminate\Validation\ValidationServiceProvider;
 use Illuminate\Validation\Validator;
 use PHPUnit\Framework\TestCase;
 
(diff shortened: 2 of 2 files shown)


driesvints:
These errors are never meant to surface to end users. Please see many previous attempts.

schonhoff:
Thanks for the answer. I saw other attempts after I committed the PR. I saw a lot of discussion around it and my question wasn't answered. Why is the validation message translated and the ` (and {$additional} more {$pluralized})` should not be translated. In my opinion it should be translatable for the devs (even if it is not for the client). A mix of languages in an exception message isn't nice to see even as a dev. I would still see this as a bug instead of a feature request. Someone invented a change in the validation exception, what is good, but it isn't good if there is a mix of languages because of that. What is the reason behind not translating the second part of the validation exception message but the first part? I except that it is not meant to be surfaced to the end user but why not changing it for the developers? Is it a performance reason?

driesvints:
An exception message is a system level message, not an end user message. This is the same concept across many languages.

schonhoff:
At the moment in laravel it is a mix of 2 languages if you use another language than English.

In the `de\validation.php` for example is:
```php
'password' => 'Das Password ist nicht korrekt.',
```

Now the `message` from a validation exception that has more errors included it shows now:
```php
Das Password ist nicht korrekt. (and 3 more errors)
```
Even it is not used for an end user a dev could imagine that the translation is missing, the translation has an error or other confusions that I can not imagine. I would say that even it is a system level message and is only for Devs it isn't intuitive to mix 2 languages in message. The message is already translated in the summarize call.

Maybe I'm not expressing myself well enough because English isn't my native language. I hope my point on this topic is now clearer. Maybe you can explain if the mix of 2 languages in 1 message is wanted. I should have implemented a test for that but I don't know how to fake another language for the translator. This isn't tested at the moment.

bert-w:
I think you are supposed to use the properties on the `ValidationException` like the errorbag or the `errors()` function, instead of outputting the actual `$exception->getMessage()`, which as you can see is just a summary.

schonhoff:
Hi @bert-w,

even if I'm using the `$exception->getMessage()` as a summary for logging it still has 2 languages inside what might be confusing for other devs. I would not use the message itself for display or output. For me it doesn't feel right to see a "summary" in multiple languages in the same expression. The mixing of languages is what in my opinion is wrong with the "summary". Either the summary should all be English or the content added by the `summarize` method should be translatable. A mix doesn't feel natural and right for me. I saw a lot of bugs related to translation and something like this calls my attention even it isn't shown to the end user (client/customer).

I hope you get my point :-) Have a nice day/night!

bert-w:
Hm yea, I think that makes sense to have it as one consistent language. In that light, it doesnt seem that bad to translate the additional `'(and :additional more error(s))'` as well, since the other option would be to _always_ translate the validator messages as English just for this `summarize()` function (under the motto "all exception messages should be English"), but that seems not so easy to implement since the validator is already translated when it is assigned to the exception.

Pablo-Camara:
Please check: https://stackoverflow.com/questions/73898408/how-to-translate-laravel-hardcoded-string-and-1-more-error-using-dynamic

@derekmd let me know how to solve this.

schonhoff:
Hey!

This should be fixed with:
https://github.com/laravel/framework/pull/43837

Pablo-Camara:
I tried adding the keys in my lang/pt/validation.php and also tried in lang/en/validation.php,
tried adding it with the placeholder, and with the number directly, and I wasn't able to get the translation to work..
So for my specific case I was able to solve the issue by removing that string completely with "->stopOnFirstFailure(true)"

please see question for my details on my attempts:
https://stackoverflow.com/questions/73898408/how-to-translate-laravel-hardcoded-string-and-1-more-error-using-dynamic/73898713#73898713

@schonhoff

PR #42225, opened 2022-05-02
[10.x] Add `or` & `and` methods to query builder

halaei:
This is just something fun I had in my mind that wanted to share:) It lets us write eloquent queries more similar to how we write sql and it is closer to English.

```php
$model->where('foo', 1)
      ->or('bar', '=', 2)
      ->and('baz', 3);
```
Thanks for your time.


ankurk91:
Too much magic to save some keystroke.

PR #42226, opened 2022-05-02
localization nested json language file when lang key called

vildanbina:
In my case I have nested json in translations:

```
{
    "form": {
        "create": "Create project",
        "edit": "Edit project",
        "show": "Show project",
        "edit_specific": "Edit Project: #{modelId}",
        "go_back": "Back to projects"
    }
}
```

but when I call `__('form.create')` returns `form.create`, after these changes will return current translation (`Create Project`).

Also works correctly when we have translation string with dot notation

### DIFF
--- a/src/Illuminate/Translation/Translator.php
+++ b/src/Illuminate/Translation/Translator.php
@@ -106,7 +106,7 @@ class Translator extends NamespacedItemResolver implements TranslatorContract
         // only one level deep so we do not need to do any fancy searching through it.
         $this->load('*', '*', $locale);
 
-        $line = $this->loaded['*']['*'][$locale][$key] ?? null;
+        $line = data_get($this->loaded['*']['*'][$locale], $key);
 
         // If we can't find a translation for the JSON key, we will attempt to translate it
         // using the typical translation file. This way developers can always just use a
--- a/src/Illuminate/Translation/Translator.php
+++ b/src/Illuminate/Translation/Translator.php
@@ -106,7 +106,7 @@ class Translator extends NamespacedItemResolver implements TranslatorContract
         // only one level deep so we do not need to do any fancy searching through it.
         $this->load('*', '*', $locale);
 
-        $line = data_get($this->loaded['*']['*'][$locale], $key);
+        $line = $this->loaded['*']['*'][$locale][$key] ?? data_get($this->loaded['*']['*'][$locale], $key);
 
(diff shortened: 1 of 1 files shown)


francoisauclair911:
@taylorotwell could you share how it would be organized then ? Should all the keys should be in one `fr.json` file and never nested ?


Thank you

dennisprudlo:
@francoisauclair911 exactly. The JSON translations are supposed to be a direct mapping between the default language and the translated language (https://laravel.com/docs/9.x/localization#using-translation-strings-as-keys)

In your example you use a more _semantic_ way of translating, that's what short keys are for (https://laravel.com/docs/9.x/localization#using-short-keys)

francoisauclair911:
Thank you @dennisprudlo, do you happen to know of any open source project where I could see a real world example of that ? Just trying to wrap my head around this when there is tons and tons of strings

dennisprudlo:
@francoisauclair911  The `validation.php` in laravel uses these translations:
https://github.com/laravel/laravel/blob/c0b60c0ac74e599122a6a417171798f609f16814/lang/en/validation.php#L154-L169

Here is a snapshot from my project:
```php
<?php

return [

    'title' => 'Settings',

    'general' => [
        'title' => 'General',

        'profile' => [
            'title' => 'Profile',
            'description' => 'Your profile picture and name are only visible to you.',
            'picture' => [
                'select' => 'Select image',
                'validation' => [
                    'required' => 'Please select JPG or PNG files only.',
                    'max' => 'The image must not be larger than 200 KB.'
                ]
            ],
            'name' => [
                'label' => 'Name',
                'validation' => [
                    'required' => 'Please enter a name.',
                    'min' => 'The name must consist of at least two characters.',
                    'max' => 'The name may consist of a maximum of 100 characters.'
                ]
            ],
    ...
```

Then for the label of the name input field in the profile settings:
```php
trans('settings.general.profile.name.label');
```

For the validation messages of the name field:
```php
trans('settings.general.profile.name.validation.{xyz}');
```

But you can structure it the way you want of course. You could have a `validations.php` with all validation messages for all field or a `fields.php` with all field names/placeholders etc. for all fields.

You won't see the exact text in your blade or php files anymore just the "translation key" however it prevents issues between languages, like the word "book".
```json
de.json

{
    "Book": "Buch",
    "Book": "buchen"
}
```

This is not possible with JSON translations. The german words for "(the) book" and "(to) book" are different. But when using english as the base language the words are the same.

PR #42227, opened 2022-05-02
[10.x] New keyword in default parameter value 

imanghafoori1:
Uses the new feature of PHP 8.1 for the default parameter values in case they are new simple objects.


GrahamCampbell:
👎 here. I strongly prefer null as a default, since it makes it easy to skip over defaults by passing null. Especially useful given Laravel does not include param names in its BC policy.

imanghafoori1:
It also seems that mockery does not support the new syntax.

PR #42229, opened 2022-05-03
[9.x] Add attemptWith for authenticating a user instance

ambengers:
Hello Team!

This PR adds an `attemptWith` method in `Illuminate\Auth\SessionGuard` for authenticating a pre-existing authenticatable instance. This feature is really helpful when we already have an existing user instance before we even attempt an authentication.

Use case: Imagine a `User` model has one `Profile` model which has a unique column `employee_number` that is used to authenticate instead of the typical username/email field:

```
$user = User::whereHas('profile', 
    fn ($query) => $query->where('employee_number', request()->employee_number)
)->firstOrFail();

if (Hash::check(request()->password, $user->password)) {
    Auth::login($user, request()->remember);
    // ...
}
```

Instead of manually checking against the hash on the password of an existing user, I thought it would be good to be able to attempt authentication, just like a typical attempt with credentials:
```
$user = User::whereHas('profile', 
    fn ($query) => $query->where('employee_number', request()->employee_number)
)->firstOrFail();

if (Auth::attemptWith($user, request()->only('password'), request()->remember)) {
    // Do something here...
}
```

### DIFF
--- a/src/Illuminate/Auth/SessionGuard.php
+++ b/src/Illuminate/Auth/SessionGuard.php
@@ -424,7 +424,7 @@ class SessionGuard implements StatefulGuard, SupportsBasicAuth
      */
     public function attemptWith(AuthenticatableContract $user, array $credentials = [], $remember = false)
     {
-        $this->fireAttemptEvent(['user' => $user, ...$credentials], $remember);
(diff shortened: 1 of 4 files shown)


taylorotwell:

PR #42231, opened 2022-05-03
[9.x] Improve arithmetic operation in `eachById()`

jay-youngn:
There is no need to calculate paging offsets in every loop.


taylorotwell:
If the code is not broken I don't see a huge benefit in changing it.

ginnerpeace:
> If the code is not broken I don't see a huge benefit in changing it.

Okay, It's just a matter of programming philosophy. 
As you said, this change is negligible on modern computers.

BTW. 
In my opinion, since we regard Laravel as an `Web Artisan`, so we should accept some better coding style, isn't it?

PR #42232, opened 2022-05-03
[9.x] Helper data_set() can create new elements as objects

SkibidiProduction:
Sometimes you have an empty object that you want to add layered nested elements to as objects using helper data_set(). Currently, you can only get this structure as an associative array, like so:

```php
$data = new stdClass();

data_set($data, 'user.name', 'foo', true);

```

and $data will contain the structure like this:

```php
{#3506
    +"user": [
        "name" => "foo",
    ],
}


```

Changes allow you to get this structure as an object:

```php
$data = new stdClass();

data_set($data, 'user.name', 'foo', true, false);

```

and $data will contain the structure like this:

```php
{#3506
    +"user": {#3507
       +"name": "foo",
    },
}

### DIFF
--- a/src/Illuminate/Collections/helpers.php
+++ b/src/Illuminate/Collections/helpers.php
@@ -95,9 +95,10 @@ if (! function_exists('data_set')) {
      * @param  string|array  $key
      * @param  mixed  $value
      * @param  bool  $overwrite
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #42244, opened 2022-05-03
[8.x] Enable to modify HTTP Client request headers when using beforeSending() callback

ianriizky:
Basically, I'm trying to add a request header when sending an HTTP request using HTTP Client from Laravel. But I want to make this thing dynamically since it requires the targeted URL as the header data and it will be used in many further HTTP request. So I decide to use the `Http::beforeSending()` method and take the given ``Illuminate\Http\Client\Request`` parameter to get the URL as the example shown below.
```php
<?PHP

use Illuminate\Http\Client\PendingRequest;
use Illuminate\Http\Client\Request;

Http::beforeSending(function (Request $request, array $options, PendingRequest $pendingRequest) {
    $requestLine = sprintf(
        '%s %s HTTP/%s',
        $request->toPsrRequest()->getMethod(),
        $request->toPsrRequest()->getUri()->withScheme('')->withHost(''),
        $request->toPsrRequest()->getProtocolVersion()
    );

    // this way doesn't work as I mention below.
    $pendingRequest->withHeaders(['Authorization' => 'Bearer '.$requestLine]);

    // this way works, but need some changes in the HTTP client.
    return $request->toPsrRequest()->withHeader('Authorization', 'Bearer '.$requestLine);
})->get('http://foo.com/json');
```

The problem is the request header wasn't changed at all since the ``PendingRequest::runBeforeSendingCallbacks()`` used to run this callback doesn't bring that into the real HTTP request as I expect. Using ``PendingRequest::withHeaders()`` also doesn't work because the actual HTTP Request was already created, so any modification that comes from the ``PendingRequest`` instance is useless.

This PR will fix that problem, so every registered ``beforeSending`` callback on the HTTP Client that returns a ``GuzzleHttp\Psr7\RequestInterface`` or ``Illuminate\Http\Client\Request`` value will be applied to the HTTP request instance.

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -984,7 +984,7 @@ class PendingRequest
      *
      * @param  \GuzzleHttp\Psr7\RequestInterface  $request
      * @param  array  $options
-     * @return \Closure
+     * @return \GuzzleHttp\Psr7\RequestInterface
      */
     public function runBeforeSendingCallbacks($request, array $options)
     {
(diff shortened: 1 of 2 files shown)


ianriizky:
@taylorotwell Could you please also bring this fix into 9.x?

PR #42248, opened 2022-05-04
[9.x] Improve password checks

korkoshko:
It also fixes a possible collision when credentials contains only "password" values.
```
$provider->retrieveByCredentials(['password' => 'dayle', 'password2' => 'night']);
```
As a result of the code above, the first record will be retrieved from the table in the database.


driesvints:
This a major breaking change. You're trying to retrieve users by their unhashed passwords, something that should be avoided at all costs. Please hash your passwords.

driesvints:
@korkoshko I finally see what you meant now. Sorry, misjudged this. Thanks for your PR 👍

PR #42249, opened 2022-05-04
[9.x] Add the ability to steal atomic locks

AlexDanault:
This PR adds the ability to steal atomic locks, for all current cache stores.

A stolen lock is a lock that changes owner (instantly) and has a new expiration. The previous owner loses control of the lock and can't release it. Some may think we can already do that, by force releasing a lock and taking a new one, but this introduces race conditions where the lock stealer could force release the previous owner's lock but can't acquire a new lock if a process manages to acquire a new lock before it. Lock stealing is instantaneous.

Lock stealing has many uses in the wild, and in my case this is groundwork for my next PR, which will enable easy debouncing of queue jobs. Think like unique jobs, but instead of dropping duplicates, only the latest job will be kept, and includes a delay. This will help greatly reduce job repetition in some cases.

### DIFF
--- a/src/Illuminate/Cache/DatabaseLock.php
+++ b/src/Illuminate/Cache/DatabaseLock.php
@@ -135,9 +135,9 @@ class DatabaseLock extends Lock
     {
         if ($this->isOwnedByCurrentProcess()) {
             $this->connection->table($this->table)
-                             ->where('key', $this->name)
-                             ->where('owner', $this->owner)
-                             ->delete();
+                        ->where('key', $this->name)
+                        ->where('owner', $this->owner)
+                        ->delete();
 
             return true;
         }
@@ -153,8 +153,8 @@ class DatabaseLock extends Lock
     public function forceRelease()
     {
         $this->connection->table($this->table)
-                         ->where('key', $this->name)
-                         ->delete();
+                    ->where('key', $this->name)
+                    ->delete();
     }
 
     /**
--- a/src/Illuminate/Cache/DatabaseLock.php
+++ b/src/Illuminate/Cache/DatabaseLock.php
@@ -117,7 +117,7 @@ class DatabaseLock extends Lock
     /**
      * Clear expired locks from database table, if lottery hits.
      *
-     * @return int
+     * @return void
      */
     protected function clearExpiredLocks()
     {
--- a/src/Illuminate/Cache/DatabaseLock.php
+++ b/src/Illuminate/Cache/DatabaseLock.php
@@ -70,9 +70,9 @@ class DatabaseLock extends Lock
                 ->where(function ($query) {
                     return $query->where('owner', $this->owner)->orWhere('expiration', '<=', time());
                 })->update([
-                'owner' => $this->owner,
-                'expiration' => $this->expiresAt(),
-            ]);
+                    'owner' => $this->owner,
+                    'expiration' => $this->expiresAt(),
+                ]);
 
             $acquired = $updated >= 1;
         }
@@ -135,9 +135,9 @@ class DatabaseLock extends Lock
     {
         if ($this->isOwnedByCurrentProcess()) {
             $this->connection->table($this->table)
-                ->where('key', $this->name)
-                ->where('owner', $this->owner)
-                ->delete();
+                             ->where('key', $this->name)
+                             ->where('owner', $this->owner)
+                             ->delete();
 
             return true;
         }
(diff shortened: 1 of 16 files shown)


DarkGhostHunter:
How this is used? Just calling `steal()`?

AlexDanault:
@DarkGhostHunter Bascially, yes. Instead of calling `acquire()` you call `steal()`. Look at the unit tests I included, they can serve as examples. `tests/Integration/Cache/RedisCacheLockTest.php`

src/Illuminate/Cache/DatabaseLock.php
```diff
@@ -94,6 +114,18 @@ protected function expiresAt()
         return $this->seconds > 0 ? time() + $this->seconds : Carbon::now()->addDays(1)->getTimestamp();
     }
 
+    /**
+     * Clear expired locks from database table, if lottery hits.
+     *
+     * @return int
+     */
+    protected function clearExpiredLocks()
```
a-bashtannik, line 122: Defined as `@return int`, but returns nothing

src/Illuminate/Cache/DatabaseLock.php
```diff
@@ -77,13 +77,33 @@ public function acquire()
...
+            ->update([
+                'owner' => $this->owner,
+                'expiration' => $this->expiresAt(),
+            ]);
+
+        $stolen = $updated >= 1;
+
+        $this->clearExpiredLocks();
+
+        return $stolen;
```
a-bashtannik, line 104: `return $updated >= 1;` no?

src/Illuminate/Cache/RedisLock.php
```diff
@@ -41,6 +41,20 @@ public function acquire()
...
+     *
+     * @return bool
+     */
+    public function steal()
+    {
+        if ($this->seconds > 0) {
+            return $this->redis->set($this->name, $this->owner, 'EX', $this->seconds, 'XX') == true;
+        } else {
+            return $this->redis->set($this->name, $this->owner, null, null, 'XX') == true;
+        }
```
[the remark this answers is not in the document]
AlexDanault: While I agree that early return is better, I'd prefer to leave this as-is because this function (steal) has the same structure as the one just before it (acquire).

src/Illuminate/Cache/DatabaseLock.php
AlexDanault: Fixed ! Thanks for catching it :)

src/Illuminate/Cache/DatabaseLock.php
AlexDanault: It would be better, I'd prefer to leave this as-is because this function (steal) has the same structure as the one just before it (acquire).

AlexDanault:
Thanks for the review @a-bashtannik ! I hope my comments explain my reasoning.

taylorotwell:
I'm a bit concerned about race conditions in many of these situations, especially since you indicate your next PR involves queued jobs. There are few places in this PR where you check for the existence of something using `get` and then make a `put` call after inspecting the result of the `get`. As time passes (even though it's a tiny amount) the results of that `get` may now be different before `put` is actually called. It's not an atomic operation and has a race condition.

DarkGhostHunter:
Same for mine, resolved it using a Lock only if the store supports it, otherwise is a clean get-put operation.

The problem for using a lock for this and that is that, at least, you should offer some Lock control to the developer (like limiting the lock hold, waiting time, and owner), which adds _some code_ more to think.

AlexDanault:
Thanks for your input @taylorotwell , I can't disagree about the issue you bring up. You can close this PR and I'll go back to the drawing board.

@DarkGhostHunter is there anywhere I could see your take on this ?

DarkGhostHunter:
> @DarkGhostHunter is there anywhere I could see your take on this ?

https://github.com/laravel/framework/pull/42342/files#diff-d89c0fd3f8f941c5dac2be73182b2a0ba894eefc5f21196ec79241ea0a3b27ddR130-R142

Just basically wraps the operation into a lock if the store supports locking. May give you some ideas.

AlexDanault:
@DarkGhostHunter That looks clean. And It'd be dumb for me to re-do it, I'll wait for your PR to be merged (hopefully) and just consume your new stuff.

DarkGhostHunter:
> @DarkGhostHunter That looks clean. And It'd be dumb for me to re-do it, I'll wait for your PR to be merged (hopefully) and just consume your new stuff.

While you are it, vote for the naming. I think `refresh` is cool. Before it was `upsert` and then `getSet`.

PR #42253, opened 2022-05-04
[9.x] add `firstOrThrow()` and `findOrThrow()` methods to eloquent query builders

Gummibeer:
This PR adds `firstOrThrow()` and `findOrThrow()` methods to the eloquent query builders as an alternative to `firstOr(fn() => throw new Exception)` to throw any custom exception.
This is based on #42243 - as I believe that this custom exception method is the most flexible solution for the named "problem".

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -483,6 +483,31 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Find a model by its primary key or throw the given exception.
+     *
+     * @param  mixed  $id
+     * @param  array|\Exception  $columns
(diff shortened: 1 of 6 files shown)


tpetry:
How about also supporting the possibility to pass in an exception class name and the exception is automatically created?

```php
$builder->findOrThrow('bar', ['column'], MyCustomException::class);
```

Gummibeer:
Good idea, for exceptions that don't need any "payload" but are self initializing. Will adjust the PR accordingly.

tpetry:
You can just pass in a default message and error code and provide the previous exception for better debugging ;)

taylorotwell:
Hmm - I think I'll table this for now in preference for `firstOr(fn)`

PR #42256, opened 2022-05-04
[9.x] Add `containsItems` Collection method to determine if it contains a specific amount of items

bradyrenting:
This PR introduces the method `containsItems()` to Collection and LazyCollection. Example below.

Currently you can already determine if a collection contains a single item by calling the method `containsOneItem()`.

The reason of this implementation is because I think it would be useful, especially for lazy collections (and for consistency), to have a similar method to be called on a collection which contains of more than 1 item.

Because of the naming convention I decided to go with `containsItems()`. But `itemsCount()` could be an alternative, which describes the method better.

### LazyCollection

```php
// Instead of ..

if ($lazyCollection->take(3 + 1)->count() === 3) {
   // do something..
}

// You can..

if ($lazyCollection->containsItems(3)) {
   // do something..
}
```

### Collection

```php
// Intead of..

if ($collection->count() === 3) {
   // do something..
}

// You can..

if ($collection->containsItems(3)) {
   // do something..
}
```

Thanks!

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -645,6 +645,16 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
+     * Determine if the collection contains specified amount of items.
+     *
+     * @return bool
+     */
+    public function containsItems($amount)
+    {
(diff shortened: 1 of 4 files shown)


taylorotwell:

PR #42257, opened 2022-05-04
[9.x] Add `value()` collection method

stevebauman:
## Description

This PR adds the `collect($items)->value($key)` method, which operates similarly to Eloquent's `$query->value($column)` method, allowing you to fetch a single key's first matching value from a collection.

## Example

Let's say we have a `Book` model with a relationship named "metadata".

This `metadata()` relationship has a pivot table, containing the metadata's `value`.

In this application, users can attach specific metadata to books, and then assign their own value via the pivot table.

With the `value()` collection method, we can take the eager loaded metadata, filter by the metadata `key`, and then retrieve the first matching `pivot.value`:

```php
// John Doe
$author = $book->metadata->where('key', 'author')->value('pivot.value');
```

---

Let me know if you'd like anything adjusted or changed. Thanks for your time! 🙏 

No hard feelings on closure ❤️

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -322,6 +322,22 @@ trait EnumeratesValues
     }
 
     /**
+     * Get a single key's value from the first matching item in the collection.
+     *
+     * @param  string  $key
+     * @param  mixed  $default
+     * @return mixed
+     */
+    public function value($key, $default = null)
+    {
+        if ($value = $this->firstWhere($key)) {
+            return data_get($value, $key, $default);
+        }
+
+        return $default;
+    }
+
+    /**
      * Determine if the collection is not empty.
      *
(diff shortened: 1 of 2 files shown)


src/Illuminate/Collections/Traits/EnumeratesValues.php
```diff
@@ -321,6 +321,22 @@ public function firstWhere($key, $operator = null, $value = null)
...
+     * @param  mixed  $default
+     * @return mixed
+     */
+    public function value($key, $default = null)
+    {
+        if ($value = $this->firstWhere($key)) {
+            return data_get($value, $key, $default);
+        }
+
+        return $default;
```
taylorotwell, line 337: Is there a reason you didn't use `value($default)` here?

stevebauman: Nope, I just didn’t think of it 😄

PR #42258, opened 2022-05-04
[9.x] Add auto language fallback to less specific localization

emiliopedrollo:
Hi everyone,

A bit of context first: as my team is preparing to undertake a huge effort on localization we've been looking for specialized software and personal online and we figured out that the vast majority charge either by word or "target translation key" (as key x language). 

Here's our problem: we plan to support locales like `en_US` and `en_GB` that would have most of the translations keys exactly the same, having those fallback to `en` could reduce drastically the number of strings in the end. As this is currently possible using the current fallback config when we add `pt_BR` and `pt_PT` to the mix suddenly there's no way to have those fallback to `pt`. 

This proposal adds auto fallback when the locale is something like `en_US` or `en-US` to a less specific `en`. It could also fallback something like `en_US-custom` to `en_US` then `en`.

As a downside, I realize that this could add several CPU cycles to process the regular expression and even worse, a lot more of I/O.

### DIFF
--- a/src/Illuminate/Translation/Translator.php
+++ b/src/Illuminate/Translation/Translator.php
@@ -325,7 +325,19 @@ class Translator extends NamespacedItemResolver implements TranslatorContract
      */
     protected function localeArray($locale)
     {
-        return array_filter([$locale ?: $this->locale, $this->fallback]);
+        $locales = [];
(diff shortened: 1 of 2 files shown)


taylorotwell:
Instead, can you PR a method like `determineLocalesUsing(fn)` that accepts a Closure. That way you can invoke that from your own service provider and do whatever custom behavior you wish. Thanks!

PR #42262, opened 2022-05-04
[9.x] Add custom segments on "remember me" recaller

PaolaRuby:
Actually `Recaller` only admit 3 segments, `id`,`token`,`hash`,
If we create a custom session class extending `SessionGuard`, or we change the `remember me` cookie on login, we need a way to get the extra data from recaller
This is useful when you have a multi company login, or when needs an extra id/value to rebuild the session
https://github.com/laravel/framework/discussions/42067

### DIFF
--- a/src/Illuminate/Auth/Recaller.php
+++ b/src/Illuminate/Auth/Recaller.php
@@ -49,7 +49,20 @@ class Recaller
      */
     public function hash()
     {
-        return explode('|', $this->recaller, 3)[2];
+        return explode('|', $this->recaller, 4)[2];
+    }
+
+    /**
+     * Get the custom segments from the recaller.
+     *
+     * @return array
+     */
+    public function customSegments()
+    {
+        $segments = explode('|', $this->recaller);
+        array_splice($segments, 0, 3);
+
+        return $segments;
     }
 
     /**
@@ -81,6 +94,6 @@ class Recaller
     {
         $segments = explode('|', $this->recaller);
 
-        return count($segments) === 3 && trim($segments[0]) !== '' && trim($segments[1]) !== '';
+        return count($segments) >= 3 && trim($segments[0]) !== '' && trim($segments[1]) !== '';
     }
 }


taylorotwell:

PaolaRuby:
https://github.com/laravel/framework/issues/42271#issuecomment-1118611261
>How are you setting these custom segments?

Login or custom session guard

>Why do you need them at all?

Rebuild the picked login company on session with remember me

>Why not store them in a database table associated with the user?

User can be logged on two or more companies at the same time(using different browsers, or with incognito mode)

PaolaRuby:
@taylorotwell i am answering all your questions, please don't let me hanging without one 🙏
I know you must have a lot to handle, and that your time must be hard.
>Multi company login, company is a manyToMany relation with user, on login user pick wich company use for session, on "remember me" laravel rebuild the user session but not the company, of course i can create another cookie for rebuild the picked company, but would be great if the recaller can handle that extra id/ids
>
>keeps same functionality, but adds option for more segments, really small not breaking change

PaolaRuby:
@taylorotwell so, on a custom guard extending [SessionGuard](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Auth/SessionGuard.php) we can make something like this
```php
protected function queueRecallerCookie(AuthenticatableContract $user)  {
    $this->getCookieJar()->queue($this->createRecaller(
        $user->getAuthIdentifier().'|'.$user->getRememberToken().'|'.$user->getAuthPassword().
            '|'.$this->request->input('company_id')
    ));
}
protected function userFromRecaller($recaller){      
      $user = parent::userFromRecaller($recaller);
      if ($user && $this->viaRemember && !($recaller->customSegments()[0] ?? false)) {
            return;
      }
      $this->session->put(['company_id' => $recaller->customSegments()[0]]);
      return $user;
  }
```
Or with a middleware
```php
// on login authenticated or in a auth macro
// auth()->getCookieJar()->make(auth()->getRecallerName(), 'pipe_segments', auth()->getRememberDuration())
if (auth()->viaRemember()) {
    $segments = explode('|', auth()->getCookieJar()->queued(auth()->getRecallerName()) ?: '');
    session(['company_id' =>$segments[3]]);
}
```

PaolaRuby:
@taylorotwell still there, please don't let me hanging

PR #42263, opened 2022-05-05
[9.x] Fixing some tokens that came with url characters

williamsilveira45:
Some software as Insomnia, Postman, etc... used to send the cookie csrf token like this:

`eyJpdiI6Ikp6UDJoTnpJTTlkNGh5bnZnc0tvOUE9PSIsInZhbHVlIjoiUjkvbzUzeTNUQk5nWUJrQjNrSEZGQ1dmM3lOVUdPdkhaV1UxK3I1UC96Y2ZLbHB4VFZRS1RZQjd1SXoxMld0Ymo4KzlSVFRWOVlLU3BjL01BTmppWkxDc3JIeWhlRVViUERsZFZVcW5xY0laWnY3dnRBU05MZ1A0Rm5YOURCMEUiLCJtYWMiOiJkMDQ1NjdlM2E4NWU3MjBhZTk0YjNhYWQxNjc2OGVkNmJlMTc4ZjllNjZmYmE3ZWIxYjk5M2E3Y2ExYzE1NzNiIiwidGFnIjoiIn0%3D`

The problem is that the function can't decrypt the token because of URL characters as `%3D`, so I just added the urldecode function to prevent this kind of issue.

### DIFF
--- a/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php
+++ b/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php
@@ -153,6 +153,7 @@ class VerifyCsrfToken
         $token = $request->input('_token') ?: $request->header('X-CSRF-TOKEN');
 
         if (! $token && $header = $request->header('X-XSRF-TOKEN')) {
+            $header = urldecode($header);
             try {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Not sure I want to change CSRF behavior on a patch release.

PR #42265, opened 2022-05-05
[9.x] Added "withoutForeignKeyConstraints" method to Builder

fredtrentini:
Heavily inspired by "withoutEvents" method.

Meant to simplify the common pattern in which foreign keys are disabled, something is done and they are enabled back again.

**Before**
```php
Schema::disableForeignKeyConstraints();
myFunction();
Schema::enableForeignKeyConstraints();
```

**After**
```php
Schema::withoutForeignKeyConstraints(function () {
    myFunction();
});
```

### DIFF
--- a/src/Illuminate/Database/Schema/Builder.php
+++ b/src/Illuminate/Database/Schema/Builder.php
@@ -32,6 +32,13 @@ class Builder
     protected $resolver;
 
     /**
+     * Indicates if foreign key constraints are active.
+     * 
+     * @var bool
+     */
+    protected $hasForeignKeyConstraints = true;
+
+    /**
      * The default string length for migrations.
      *
      * @var int|null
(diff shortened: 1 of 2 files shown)


taylorotwell:
Line 376 and line 390 always evaluate to `true`. Am I missing something?

src/Illuminate/Database/Schema/Builder.php
```diff
@@ -378,9 +387,31 @@ public function enableForeignKeyConstraints()
      */
     public function disableForeignKeyConstraints()
     {
-        return $this->connection->statement(
+        $this->hasForeignKeyConstraints = $this->connection->statement(
```
stevebauman, line 390: Since you're assigning the value I feel like you'd might as well return in one line?

```php
return $this->hasForeignKeyConstraints = $this->connection->statement('...');
```

fredtrentini: Fair enough, that makes sense, however taylor stated it may be returning true all the time, so I need to investigate if this logic is correct because I know that `connection->statement` returns a boolean, however I'm not sure what that boolean means.

PR #42268, opened 2022-05-05
[9.x] Allow passing key/value arrays to getArguments and getOptions

jnoordsij:
Allow passing key/value arrays to getArguments and getOptions on Artisan commands for a slightly easier interface.


taylorotwell:
Demonstration of the easier interface?

jnoordsij:
Basically the example in the test: you can pass an array with key/values in an arbitrary order you find convenient, without having to remember a specific order of arguments. 

Also it is possible to skip those you don't want to use without providing/copying the default value, e.g. `['name' => 'argument', 'default' => 'default']` which previously would have been `['argument', null, '', 'default']`.

PR #42270, opened 2022-05-05
$table must be string

kl4ver:
Type of Illuminate\Notifications\DatabaseNotification::$table must be string (as in class Illuminate\Database\Eloquent\Model)


driesvints:
We don't use property types in the framework atm.

PR #42276, opened 2022-05-06
[9.x] Share logging context across channels and stacks

timacdonald:
## Purpose

This PR bring the ability to share context across all channels and stacks, rather than just one.

## Why this feature is useful

Currently when we specify context, it is on a per log or per channel basis.

```php
// per log...

Log::info('per log', ['log' => 'context']);

// per channel...

Log::withContext(['channel' => 'context']);

Log::channel('stderr')->withContext(['channel' => 'context']);
```

but there is no way to specify context to be shared across all channels and stacks.

This feature is useful in the example [currently shown in the docs](https://laravel.com/docs/9.x/logging#contextual-information) for the `withContext` feature of a channel.

The example in the docs being...

```php
class AssignRequestId
{
    /**
     * Handle an incoming request.
     *
     * @param  \Closure  $next
     * @return mixed
     */
    public function handle($request, Closure $next)
    {
        $requestId = (string) Str::uuid();
 
        Log::withContext([
            'request-id' => $requestId
        ]);
 
        return $next($request)->header('Request-Id', $requestId);
    }
}

```

The problem with this is that it only applies the context the the default channel, but it would be useful if the context was shared across ALL channels and stacks.

## Usage

This new feature could be used in a middleware as above, but also in a service provider to apply a unique ID to all invocations of the app, be it from a request or a job or an artisan command.

```php
class AppServiceProvider
{
    public function register()
    {
        //
    }


    public function boot()
    {
        Log::shareContext([
            'invocation-id' => (string) Str::uuid(),
        ]);
    }
}
```

Now in my application, no matter what logger I use, I'll have the invocation ID in my context.

```php
Log::info(/* ... */);

Log::channel('stderr')->info(/* ... */);

Log::channel('bugsnag')->info(/* ... */);
```

Other logging systems outside of the Laravel logging system can also hook in and get the shared context to utilise...

```php
$sharedContext = Log::sharedContext();
```

## Notes

- Not doing anything to the emergency logger. One I couldn't test it, and two if there is something funky happening with the context that is causing the initial logger to fail creation I wouldn't want that to also apply to the emergency logger.
- Stacks get the shared context at the time they are created, but do not get any additional shared context once resolved - where as channels will always have the latest shared context.
- Merging is performed in the same manner as `withContext` i.e. not recursive.

### DIFF
--- a/src/Illuminate/Log/LogManager.php
+++ b/src/Illuminate/Log/LogManager.php
@@ -544,6 +544,18 @@ class LogManager implements LoggerInterface
     }
 
     /**
+     * Flush the shared context.
+     *
+     * @return array
+     */
+    public function flushSharedContext()
+    {
+        $this->sharedContext = [];
+
+        return $this;
+    }
+
+    /**
      * Parse the driver name.
      *
      * @param  string|null  $driver
--- a/src/Illuminate/Log/LogManager.php
+++ b/src/Illuminate/Log/LogManager.php
@@ -52,6 +52,13 @@ class LogManager implements LoggerInterface
     protected $dateFormat = 'Y-m-d H:i:s';
 
     /**
+     * The context shared across channels and stacks.
+     *
+     * @var array
+     */
+    protected $sharedContext = [];
+
+    /**
      * Create a new Log manager instance.
      *
      * @param  \Illuminate\Contracts\Foundation\Application  $app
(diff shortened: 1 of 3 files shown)


ankurk91:
How can we reset this behavior?
How does it work with Octane?

timacdonald:
Regarding Octane, you would need to hook into events...

```php
$this->app['events']->listen([
    RequestReceived::class,
    TaskReceived::class,
    TickReceived::class,
], fn () => Log::shareContext(['invocation-id' => (string) Str::uuid()]));
```

We could potentially add some magic into Octane to handle this for you or document it on the Octane documentation.

timacdonald:
@taylorotwell see: https://github.com/laravel/octane/pull/513

note that the new test will fail until this PR is merged / tagged. I've tested locally and got the ✅

omitobi:
I just tested the sharedContext but the logs on my Job does not have the context.
Is this by design or there are other configuration needed?

PS: I tested this on Laravel 10.

PR #42277, opened 2022-05-06
[9.x] Update app() docblock

IwakamiYuki:
Update docblock for passing class-string to app().


driesvints:
Seems this breaks static analysis

PR #42280, opened 2022-05-06
[9.x] Set relation parent key when using `forceCreate` on `HasOne` and `HasMany` relations

### LINKED ISSUE #42273: ForceCreate through relationship
- Laravel Version: 9.11.0
- PHP Version: 8.1.1

### Description:
When creating a model through a relationship (e.g. `$book->reviews()->create(['text' => '...'])`), you expect that relation foreign attributes will be populated (in this example, `book_id` should be auto-populated when we create the Review instance through the book instance).

However, when using `forceCreate` instead of `create`, this behaviour does not occur as might be expected (at least by me!). This seems to be due to `create` being implemented for relationships to have this behaviour (https://github.com/laravel/framework/blob/3ec1e4ce0c10f6414b7ef3c4f42dd85cd5102d07/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php), but `forceCreate` is not and as such it gets forwarded to `Builder` where this setting of foreign attributes doesn't occur.

Either this behaviour is intended but misleading, or it has just been missed?

### Steps To Reproduce:

1. Create two models, one with a HasMany/HasOne relation to the other
2. Attempt to `forceCreate` a model through the HasMany/HasOne relation
3. Watch as either the query fails because the foreign id is not nullable, or the foreign id is set to null

MVP showing this behaviour: https://github.com/atoff/laravel-demo-forcecreate - run `php artisan bug:demo`
image

rodrigopedra:
Closes #42273

As explained in issue #42273 when calling several methods that either create or instantiate a related instance from a relation method, the relation parent's key is automatically filled, but not when calling `forceCreate`.

Take this model as an example:

~~~php
class Lesson extends Model
{
    public function questions()
    {
        return $this->hasMany(Question::class);
    }
}
~~~

When calling all these following methods the lesson primary key is set as the new question instance foreign key automatically:

~~~php
$lesson = Lesson::first();

$q1 = $lesson->questions()->create(...);
$q2 = $lesson->questions()->findOrNew(...);
$q3 = $lesson->questions()->firstOrNew(...);
$q4 = $lesson->questions()->make(...);
$q5 = $lesson->questions()->save(new Question(...));
~~~

This behavior is also present in other methods which use the ones listed above such as `updateOrCreate`, `saveMany`, and others.

But when calling `->forceCreate(...)` from a relation, the foreign key is not set, which I agree with issue's #42273 OP, that is not the expected behavior.

Reason it does not work as the other is due to `forceCreate` not being implemented on `HasOneOrMany`, thus it is deferred to the underlying `Builder` which, of course, would not set any additional attributes.

This PR:

- Adds a `forceCreate` method to the `HasOneOrMany` relation class, which adds the relation parent key before actually force creating a related instance
- Adds relevant tests

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
@@ -311,6 +311,19 @@ abstract class HasOneOrMany extends Relation
     }
 
     /**
+     * Create a new instance of the related model. Allow mass-assignment.
+     *
+     * @param  array  $attributes
+     * @return \Illuminate\Database\Eloquent\Model
+     */
+    public function forceCreate(array $attributes)
+    {
+        $attributes[$this->getForeignKeyName()] = $this->getParentKey();
+
+        return $this->related->forceCreate($attributes);
+    }
+
+    /**
      * Create a Collection of new instances of the related model.
      *
      * @param  iterable  $records
--- a/tests/Database/DatabaseEloquentHasManyTest.php
+++ b/tests/Database/DatabaseEloquentHasManyTest.php
@@ -55,6 +55,15 @@ class DatabaseEloquentHasManyTest extends TestCase
         $this->assertEquals($created, $relation->create(['name' => 'taylor']));
     }
 
+    public function testForceCreateMethodProperlyCreatesNewModel()
+    {
+        $relation = $this->getRelation();
+        $created = $this->expectForceCreatedModel($relation, ['name' => 'taylor']);
+
+        $this->assertEquals($created, $relation->forceCreate(['name' => 'taylor']));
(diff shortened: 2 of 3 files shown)


driesvints:
@rodrigopedra is this also present in 8.x? If so we'll need to direct the PR there.

rodrigopedra:
@driesvints it is. 

But as I considered it not to be a bug, but an improvement over an expected behavior I thought it would not fit the 8.x branch anymore.

But I can redirect it to 8.x if you think it is fine.

rodrigopedra:
@driesvints as I will be out for some hours, I sent PR #42281 targeted to the 8.x branch. Please close the one you think should be closed. Thanks =)

driesvints:
Thanks @rodrigopedra

PR #42282, opened 2022-05-06
Make additional error message translatable inside ValidationException class.

usanzadunje:
This PR only wraps plain string in translate function making it translatable inside project.

When an validation error occurs this exception is thrown and `message` filed is sent with it in an JSON response. I am using this message to display an alert to my end users. But I noticed that string that is being added by Laravel at the end of 1st validation error is not translatable. It is plain string and therefore I could not make it localized and display it differently for different languages my end user has selected.

Having all this in mind, and after some nice people at Laravel discord suggested that I should make PR, I made this PR in order to enable this "feature". 

I have never made an PR to any bigger open-source project, so forgive me if I missed any details. :)

Regards,

Dusan


driesvints:
See many many previous attempts.

Issue #36763, opened 2021-03-26
Http facade doesn't wrap Guzzle's RequestException


AdirYad:
- Laravel Version: 8.30.1
- PHP Version: 8.0.2
- Database Driver & Version: MySQL 8.0.23

### Description:
When I use Laravel's `Http` facade, I expect it to wrap any exception that Guzzle throws.
There's Laravel's exception: `Illuminate\Http\Client\RequestException`
And Guzzle's exception: `GuzzleHttp\Exception\RequestException`

In my code, I catch only Laravel's exception, and for some reason, a Guzzle's exception is thrown, I wasn't prepared for it and I believe it's not the intended behavior.

I'm using a proxy for HTTP, sometimes it fails and causes this exception to be thrown.

### Steps To Reproduce:

```php
try {
  Http::withOptions([
     'proxy' => config('app.proxy_url')
  ])->get('https://example.com/users')->throw();
} catch (Illuminate\Http\Client\RequestException $exception) {
  Log::error('Exception caught', ['message' => $exception->getMessage()]);
} catch (GuzzleHttp\Exception\RequestException $exception) {
  // For some reason this exception is throws instead of Laravel's exception
}
```

```
GuzzleHttp\Exception\RequestException
cURL error 56: Received HTTP code 502 from proxy after CONNECT (see https://curl.haxx.se/libcurl/c/libcurl-errors.html) for https://example.com
```

![image](https://user-images.githubusercontent.com/45428455/112613616-5aa16380-8e31-11eb-9725-6d23477584a2.png)

Is that the intended behavior? `Http` wraps Guzzle, I personally expect it to convert Guzzle's exceptions to Laravel's ones.

driesvints:
I get the laravel exception with your exact piece of code.

AdirYad:
> I get the laravel exception with your exact piece of code.
Are you sure that Laravel converts 100% guzzle exceptions into Laravel exceptions, did you check my stack trace?
I have this exception once to a few minutes using proxies, when I remove the proxies it works fine.

driesvints:
If you can someone share the exact code to reproduce this I'll try again 👍

AdirYad:
> If you can someone share the exact code to reproduce this I'll try again
You can see my example code above.

driesvints:
@AdirYad I can't reproduce it with your code.

PR #42286, opened 2022-05-06
[8.x] All guzzle errors should be wrapped

glamorous:
Guzzle Request exception that don't have a response should be wrapped
as well. Most logical exception is the ConnectionException from Laravel
to be thrown because RequestException expects a response (and there is
no response in these cases)

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -4,7 +4,6 @@ namespace Illuminate\Http\Client;
 
 use GuzzleHttp\Client;
 use GuzzleHttp\Cookie\CookieJar;
-use GuzzleHttp\Exception\ConnectException;
 use GuzzleHttp\Exception\RequestException;
 use GuzzleHttp\Exception\TransferException;
 use GuzzleHttp\HandlerStack;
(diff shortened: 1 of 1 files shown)


glamorous:
FYI:  this will close/fix the issue from https://github.com/laravel/framework/issues/36763 as well.

taylorotwell:
I am closing this pull request because it **lacks sufficient explanation**, tests, or both. It is difficult for us to merge pull requests without these things because the change may introduce breaking changes to the framework.

Thanks!

glamorous:
Hi Taylor, I opened an issue https://github.com/laravel/framework/issues/42286 that explains the issue. The linked issue in the comments https://github.com/laravel/framework/issues/36763 is another example. 

As an end user you only expect to catch HttpClientException from the `Illuminate\Http\Client` namespace or a more detailed exception like `ConnectionException` or `RequestException` from the same namespace. Without this change, in some usecases you receive a GuzzleException instead. The reason is that for guzzle will still throw exceptions even if the guzzle option http_errors is set to false. Only the ConnectException was catched but all other cases aren't. Replacing the connectException (from Guzzle) with the TransferException (the parent) will catch:
- BadResponseException (including Client 4xx & ServerException 5xx), but guzzle will not throw those with HTTP_ERRORS = false (by default by Laravel)
- ConnectException (which was already catched)
- RequestException (including TooManyRedirectsException and BadResponseException)

So it will catch all cases of curl that is failing without there's a response available to check for success.

Don't directly know how I could mock a failing curl call without response, can't find an example on that in the current tests.

PR #42287, opened 2022-05-06
[9.x] Fix too many channels with pusher broadcasting

### LINKED ISSUE #42269: `PusherBroadcaster` max channels per call limit
- Laravel Version: v8.83.11
- PHP Version: 8.1.5
- Database Driver & Version: MySQL 5.7.37

### Description:

Pusher limits publishing an event to 100 channels per call.
https://pusher.com/docs/channels/server_api/http-api/#publishing-events

In our application each user has it's own private channel where different events get published on.
Some events need to be published to a large number (> 100) of user channels.

Currently there is no way to work around this limitation.
Since this is a Pusher limitation, I think `PusherBroadcaster` should chunk the channels in batches with 100 channels each and broadcast the event in multiple calls.

### Steps To Reproduce:

1. Use `pusher` as `BROADCAST_DRIVER`
2. Define more than 100 channels in the `broadcastOn` method of an event
3. Broadcast the event
4. **Error:** `An event can be triggered on a maximum of 100 channels in a single call.`

driesvints:
This PR fixes an issue with the Pusher channel so it sends the events in batches of 100 as Pusher will not allow more than that amount of channels. 

I didn't sent this in to 8.x because the pusher implementation is different there and didn't want to introduce merge conflicts.

See https://pusher.com/docs/channels/server_api/http-api/#publishing-events

Fixes https://github.com/laravel/framework/issues/42269

### DIFF
--- a/src/Illuminate/Broadcasting/Broadcasters/PusherBroadcaster.php
+++ b/src/Illuminate/Broadcasting/Broadcasters/PusherBroadcaster.php
@@ -4,6 +4,7 @@ namespace Illuminate\Broadcasting\Broadcasters;
 
 use Illuminate\Broadcasting\BroadcastException;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Collection;
 use Pusher\ApiErrorException;
 use Pusher\Pusher;
 use Symfony\Component\HttpKernel\Exception\AccessDeniedHttpException;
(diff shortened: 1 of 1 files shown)


L3o-pold:
Using custom pusher server, there is no such limitation, so we are introducing a limitation / a breaking change.

PR #42290, opened 2022-05-06
[9.x] add findOr method to Query/Builder

mihaileu:
This PR adds a findOr() method to the Query Builder. It is similar with #42092 

Example usage:

``` 
$data = \DB::table('something')->findOr($id, fn () => throw new CustomException);
$data = \DB::table('something')->findOr($id, fn () => (object) ['id=>'0', 'name'=>'N/A']);
```

Edit: I updated the example as it was wrong.
Probably there are other usage cases.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -2555,6 +2555,29 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Execute a query for a single record by ID or call a callback.
+     *
+     * @param  mixed  $id
+     * @param  \Closure|array  $columns
+     * @param  \Closure|null  $callback
+     * @return mixed|static
+     */
+    public function findOr($id, $columns = ['*'], Closure $callback = null)
+    {
+        if ($columns instanceof Closure) {
+            $callback = $columns;
+
+            $columns = ['*'];
+        }
+
+        if (! is_null($data = $this->find($id, $columns))) {
+            return $data;
+        }
+
+        return $callback();
+    }
+
+    /**
      * Get a single column's value from the first result of a query.
      *
      * @param  string  $column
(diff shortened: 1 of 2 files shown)


eduance:
Another way of doing this is by using the the `?:` operator, but for readability I'd really like this addition.

bert-w:
No support because you can do the same thing with plain PHP:
```
$id = DB::table('users')->find(9876)?->id ?? throw new Exception('nope');
```
vs
``` 
$id = DB:table('users')->findOr(9876, throw new Exception('nope'));
```

(also the `$callback` parameter cannot be `null` in your implementation)

GrahamCampbell:
I'm :-1: here, for the reason provided by @bert-w. Since PHP 8.0, throw statements are also expressions.

davisngl:
And yet none of you had problems with #42092, when only here is an example of throwing exception?

bert-w:
> And yet none of you had problems with #42092, when the usage with throwing an exception was highlighted?

good catch, didnt see that PR since I only browse here occasionally. However that being said I'd take the same stance. Any of the `doSomethingOr` functions I dont really like since they can often times be achieved with other methods like stated above. So the functions are bloat.

Anyway that means that this function is already merged so RIP

davisngl:
> > And yet none of you had problems with #42092, when the usage with throwing an exception was highlighted?
> 
> good catch, didnt see that PR since I only browse here occasionally. However that being said I'd take the same stance. Any of the `doSomethingOr` functions I dont really like since they can often times be achieved with other methods like stated above. So the functions are bloat.
> 
> Anyway that means that this function is already merged so RIP

True.

Would say more, but this isn't the community for taking constructive criticism.

bert-w:
Apparently there's also https://github.com/laravel/framework/pull/42253 , no clue why these all pop up around the same time. It is kinda weird especially since Laravel 9.x now requires PHP 8+ which adds nice functionality to the language itself. I think it will take a couple of years (yes) for people to start using features frequently like the null safe (`$obj->user()?->id`) operators...

PR #42291, opened 2022-05-07
add ``cache_if`` and ``cache_unless`` helpers

isaeken:
If a developer wants a data to be cached only when the required comparison is true, they can write cleaner code with the ``cache_if`` or ``cache_unless`` function.

example usage:

```php
cache_if(app()->isProduction(), 'users', now()->addSeconds(30), function () {
    return User::all();
});
```

### DIFF
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -418,3 +418,37 @@ if (! function_exists('with')) {
         return is_null($callback) ? $value : $callback($value);
     }
 }
+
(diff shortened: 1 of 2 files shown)


DarkGhostHunter:
Usually there is no need to use `cache_if()`/`cache_unless8)` because a call to a cache is usually done in a idempotent way with `cache()->remember()`.

Even then, you can't get rid of the function to set the value, so it seems redundant IMHO.

driesvints:
Feel free to add these to your own app 👍

PR #42293, opened 2022-05-07
[9.x] Use View Contract instead of implementation

alies-dev:
Laravel allows to use custom implementations of `Illuminate\Contracts\View\View`. This PR is a simple fix for PHPDoc to underline this fact


taylorotwell:
I don't know if we always plan to stick to that interface in this class.

PR #42304, opened 2022-05-08
[9.x] Add Arr::cycle() method for cycling classes in a loop

ziadoz:
This adds an `Arr::cycle()` method that makes it easy to cycle through classes in a loop. It behaves identically to [Twig's cycle method](https://twig.symfony.com/doc/3.x/functions/cycle.html).

It can be used like this in a Blade template:

```
@foreach ($objects as $object)
    <div class="{{ Arr::cycle(['foo', 'bar', 'baz', 'qux'], $loop->index) }}"></div>
@endforeach
```

Which would output this:

```
<div class="foo"></div>
<div class="bar"></div>
<div class="baz"></div>
<div class="qux"></div>
<div class="foo"></div>
<div class="bar"></div>
<div class="baz"></div>
<div class="qux"></div>
...
```

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -91,8 +91,8 @@ class Arr
     /**
      * Cycle through items in an array.
      *
-     * @param array $array
-     * @param int $position
+     * @param  array  $array
+     * @param  int  $position
      * @return mixed
      */
     public static function cycle($array, $position)
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -89,6 +89,20 @@ class Arr
     }
 
     /**
+     * Cycle through items in an array.
+     *
+     * @param array $array
+     * @param int $position
+     * @return mixed
+     */
+    public static function cycle($array, $position)
+    {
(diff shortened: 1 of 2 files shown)


DarkGhostHunter:
Isn't `Arr::get()` the same? You can even use `optional()` for when you go out of bounds.

```blade
@foreach ($objects as $object)
    <div class="{{ Arr::get(['foo', 'bar', 'baz', 'qux'], $loop->index, optional()) }}"></div>
@endforeach
```

DarkGhostHunter:
In other words, this method allows to loop through an array/`Arrayable` using an index number and rewinding if it goes out of bounds.

```php
$array = ['foo', 'bar', 'baz', 'quz'];

Arr::cycle($array, 0); // 'foo'
Arr::cycle($array, 4); // 'foo', goes back to the beginning
```

ziadoz:
> I don't think it should work with non-numeric arrays, as if that's the case you're better using `Arr::get()` with `optional()`:

It won't work with string indexed arrays or numeric indexed arrays with gaps in. You'll get an undefined key error when you try to cycle them. That seemed like the right behaviour to me.

taylorotwell:

PR #42306, opened 2022-05-09
[9.x] Adds alphabetical word count validation

DarkGhostHunter:
## What?

Adds validation rules to count alphabetic words in a field for minimum, maximum and between two amounts.

```php
public function contact(Request $request)
{
    $request->validate([
        'title' => 'required|words_between:2,10',
        'body' => 'required|words_min:100',
        'subject' => 'required|words_max:1',
    ]);

    // ...
}
```

## How?

It's just a `str_word_count()` call underneath.

> For other languages, the `ext-intl` extension may be needed an a little rewrite as this [answer](https://stackoverflow.com/a/22465204) says, and Laravel doesn't require it. Could be expanded with an additional parameter to define the locale.

## BC?

None, I don't expect somebody using `words_*` as a rule.

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -2039,6 +2039,53 @@ trait ValidatesAttributes
     }
 
     /**
+     * Validate the number of alphabetical words is greater than a minimum value.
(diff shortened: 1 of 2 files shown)


taylorotwell:
I suggest a custom rule for this if needed.

PR #42316, opened 2022-05-09
[9.x] Add custom segments on "remember me" for session rebuild

### LINKED ISSUE #42271: How to add extra data on session guard recaller
### Discussed in https://github.com/laravel/framework/discussions/42067

<div type='discussions-op-text'>

<sup>Originally posted by **PaolaRuby** April 20, 2022</sup>
Only admit 3 segments, `id`,`token`,`hash`
https://github.com/laravel/framework/blob/b9203fca96960ef9cd8860cb4ec99d1279353a8d/src/Illuminate/Auth/Recaller.php#L80-L85
Is there a config for set a custom recaller with more segments?
Is there a way or a plugin for customize session recaller?</div>

If we create a custom session class extending `SessionGuard`, or we change the `remember me` cookie on login, we need a way to get the extra data from recaller
This is useful when you have a multi company login, or when needs an extra id/value to rebuild the session

PaolaRuby:
Closes #42271

Actually `Recaller` only admit 3 segments, `id`,`token`,`hash`,
If we create a custom session class extending `SessionGuard`, or we change the `remember me` cookie on login, we need a way to get the extra data from recaller, this is useful when you have a multi company login, or when needs an extra id/value to rebuild the session,

**Example:** company is a manyToMany relation with user, on login user pick wich company use for session, on "remember me" laravel rebuild the user session **but not the company**, of course i can create another cookie for rebuild the picked company or another company selector after session rebuild, but would be great if the recaller can handle that extra id/ids

**keeps same functionality, but adds option for more segments, really small not breaking change**

-----
https://github.com/laravel/framework/issues/42271#issuecomment-1118611261
>How are you setting these custom segments?

Login, custom session guard, or auth macro

>Why do you need them at all?

Rebuild the picked login company on session with remember me

>Why not store them in a database table associated with the user?

User can be logged on two or more companies at the same time(using different browsers, or with incognito mode, different computers, etc)

----

### DIFF
--- a/src/Illuminate/Auth/Recaller.php
+++ b/src/Illuminate/Auth/Recaller.php
@@ -49,7 +49,7 @@ class Recaller
      */
     public function hash()
     {
-        return explode('|', $this->recaller, 3)[2];
+        return explode('|', $this->recaller, 4)[2];
     }
 
     /**
--- a/src/Illuminate/Auth/Recaller.php
+++ b/src/Illuminate/Auth/Recaller.php
@@ -49,20 +49,7 @@ class Recaller
      */
     public function hash()
     {
-        return explode('|', $this->recaller, 4)[2];
-    }
-
-    /**
-     * Get the custom segments from the recaller.
-     *
-     * @return array
-     */
-    public function customSegments()
-    {
-        $segments = explode('|', $this->recaller);
-        array_splice($segments, 0, 3);
-
-        return $segments;
+        return explode('|', $this->recaller, 3)[2];
     }
 
     /**
@@ -95,5 +82,15 @@ class Recaller
         $segments = explode('|', $this->recaller);
 
         return count($segments) >= 3 && trim($segments[0]) !== '' && trim($segments[1]) !== '';
+    }
+
+    /**
+     * Get the recaller's segments.
+     *
+     * @return array
+     */
+    public function segments()
+    {
+        return explode('|', $this->recaller);
     }
 }
--- a/src/Illuminate/Auth/Recaller.php
+++ b/src/Illuminate/Auth/Recaller.php
@@ -49,7 +49,20 @@ class Recaller
      */
     public function hash()
     {
-        return explode('|', $this->recaller, 3)[2];
+        return explode('|', $this->recaller, 4)[2];
+    }
+
+    /**
+     * Get the custom segments from the recaller.
+     *
+     * @return array
+     */
+    public function customSegments()
+    {
+        $segments = explode('|', $this->recaller);
+        array_splice($segments, 0, 3);
+
+        return $segments;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
How are you actually setting the custom segments? Can you provide a code example?

PaolaRuby:
@taylorotwell thanks for answer
https://github.com/laravel/framework/pull/42262#issuecomment-1118749523
on a custom guard extending [SessionGuard](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Auth/SessionGuard.php) 
```php
protected function queueRecallerCookie(AuthenticatableContract $user)  {
    $this->getCookieJar()->queue($this->createRecaller(
        $user->getAuthIdentifier().'|'.$user->getRememberToken().'|'.$user->getAuthPassword().
            '|'.session('company_id', $this->request->input('company_id'))
    ));
}
protected function userFromRecaller($recaller){      
      $user = parent::userFromRecaller($recaller);
      if ($user && $this->viaRemember && !($recaller->customSegments()[0] ?? false)) {
            return;
      }
      $this->session->put(['company_id' => $recaller->customSegments()[0]]);
      return $user;
  }
```
And overwriting a lot for use the same recaller

**Maybe, is a better option for this?**

taylorotwell:
How can you remove all that code? This PR gives no way to define custom segments - only retrieve them?

PaolaRuby:
>This PR gives no way to define custom segments - only retrieve them?

yes, it is true, i think the same, but i'm trying to make the smaller change for the framework, i can remove the extending with an macro for renew the cookie, without this change it is imposible because if there is more than 3 segments "remember me" stop working

https://github.com/laravel/framework/blob/544ee0d2add5619ea85993c6462e86a1a250ecbe/src/Illuminate/Auth/Recaller.php#L80-L84

PaolaRuby:
@korkoshko yes, it is what i have now, but i'm trying to avoid all the extendingm recaller.php, sessionguard.php, adding a service provider for the new guard, and others

PaolaRuby:
@korkoshko you forget about overwriting also
https://github.com/laravel/framework/blob/544ee0d2add5619ea85993c6462e86a1a250ecbe/src/Illuminate/Auth/Recaller.php#L80-L84
`count($segments) === 3` makes stop "remember me" on more than 4 segments

angeljqv:
+1 for this
>This PR gives no way to define custom segments

Would be great some changes on SessionGuard class like a segments resolver on login for the custom segments

PaolaRuby:
@korkoshko again, yes, that's what i currently have, 
and the whole "extending" is what I'm trying to avoid

PaolaRuby:
>public static $segments = 4;

@korkoshko maybe, still, i'm waiting for Taylor, maybe he close this

PaolaRuby:
@angeljqv @taylorotwell i can add a "session keys setter" on [SessionGuard](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Auth/SessionGuard.php) for a define segments solution

PaolaRuby:
@korkoshko just refresh the cookie with the new id, look at this for example 
https://codeanddeploy.com/blog/laravel/how-to-implement-remember-me-with-custom-expiration-to-your-laravel-8-login

taylorotwell:
Refactored this to just a plain `segments` method - you can do whatever you want from there.

PaolaRuby:
@taylorotwell thank you very much, 
that works the same for me

Issue #42317, opened 2022-05-09
Exception::__construct(): Argument #2 ($code) must be of type int, string given at .../Illuminate/Database/Concerns/ManagesTransactions.php:94)


Brenneisen:
- Laravel Version: 9.11.0
- PHP Version: 8.0.18
- Database Driver & Version: 

### Description:

After updating to v9.11.0, we are conveniently seeing this error in our log:

`Exception::__construct(): Argument #2 ($code) must be of type int, string given at .../Illuminate/Database/Concerns/ManagesTransactions.php:94)`

It seems to be related to this PR https://github.com/laravel/framework/pull/42129.

rodrigopedra:
In theory `->getCode()` should return an `int`, as seen here:

https://www.php.net/manual/en/throwable.getcode.php

I know PHP is not yet enforcing method signature on extended classes, but I would say the exception being caught and wrapped is not conforming to the definition.

Can you share the full stack trace, so we can assess the original Exception? 

Or, if what originates this error is a custom exception defined in your app, check if its `getCode` is returning an `int`, or is being constructed with a `string` code instead of an `int`?

The `$code` argument for `Exception` descendants, should be an `int`, while not yet being enforced:

https://www.php.net/manual/en/exception.construct.php

rodrigopedra:
To illustrate what I meant:

~~~php
class OtherException extends RuntimeException {}

Artisan::command('local:test', function () {
    try {
        throw new RuntimeException('foo', 'bar'); // <<< code should be an int here
    } catch (Throwable $e) {
        throw new OtherException($e->getMessage(), $e->getCode(), $e);
    }
});
~~~

As the first `RuntimeException` thrown is not honoring the constructor signature, when this is wrapped into another inside the catch, it fails, as this wrapper exception defines the scalar type hint `int`.

In my opinion, the fault is from the inner exception which is not respecting the constructor signature, and not from the wrapping exception.

But if this signature mismatch is from a third-party package and not from code you own, a middle ground would be to propose to remove the typehints from the `DeadlockException` constructor signature.

driesvints:
I doubt this is an issue with Laravel itself for the reasons @rodrigopedra explained. Maybe if @rodrigopedra's PR gets merged you can see what the actual exception is.

Thanks for the help always @rodrigopedra, we appreciate it 👍

rodrigopedra:
Thank you @driesvints for your hard work and all Laravel team.

[...]

PR #42320, opened 2022-05-09
[9.x] Keep original exception to preserve the exception trace

rodrigopedra:
While commenting on issue #42317 I noted `ManagesTransactions@handleTransactionException` discards the exception being handled when wrapping it into the new `DeadlockException`, by passing `$e->previous()` instead of the handled exception.

Passing the actual exception as the `$previous` parameter to the new `DeadlockException` being thrown, would preserve the actual stack trace and allow to assess why the original exception was thrown.

This PR:

- Modifies `ManagesTransactions@handleTransactionException` to use the actual exception being handled instead of its `->getPrevious()` result.

### DIFF
--- a/src/Illuminate/Database/Concerns/ManagesTransactions.php
+++ b/src/Illuminate/Database/Concerns/ManagesTransactions.php
@@ -88,11 +88,7 @@ trait ManagesTransactions
                 $this->getName(), $this->transactions
             );
 
-            throw new DeadlockException(
-                $e->getMessage(),
-                $e->getCode(),
-                $e->getPrevious()
-            );
+            throw new DeadlockException($e->getMessage(), $e->getCode(), $e);
         }
 
         // If there was an exception we will rollback this transaction and then we


rodrigopedra:
One note: this PR does not tries to solve or handle issue #42317, I just noted this usage while commenting out the issue.

driesvints:
If this is a bug fix then we need to target 8.x

rodrigopedra:
Hey @driesvints , the `DeadlockException` wrapping was added to 9.x, so this change is not possible on 8.x.

Code in 9.x:

https://github.com/laravel/framework/blob/544ee0d2add5619ea85993c6462e86a1a250ecbe/src/Illuminate/Database/Concerns/ManagesTransactions.php#L78-L96

Code in 8.x:

https://github.com/laravel/framework/blob/abd5b3896fd82fe5177272b04552bc4ba1d933de/src/Illuminate/Database/Concerns/ManagesTransactions.php#L77-L104

driesvints:
@rodrigopedra thanks for clarifying that 👍

rodrigopedra:
@driesvints you're welcome! Thanks for looking into this, wish you a nice release day =)

Issue #42326, opened 2022-05-10
Non-digits allowed in digits_between


taai:
- Laravel Version: 9.11.0

### Description:

Meaning of **digit**: [any one of the ten numbers 0 to 9](https://dictionary.cambridge.org/dictionary/english/digit)

A dot (```.```) is **not a digit**.

The problem is that the validation rule ```digits_between``` accepts fractions (a dot), multiple dots and **even dots only** (```....```). :rofl:

Note that the validation rule ```digits``` does not accept fractions (that is good) and that makes an inconsistency between these two validation rules.

Please, let the digits **stay digits**! In my opinion, [this PR](https://github.com/laravel/framework/pull/40278) should be rolled back.

I suggest to make a separate **and a valid** validation rule for fractions. Name it ```floats``` and ```floats_between```, or something like that.

This is how validation rule ```digits_between``` looks like in [Laravel v9.11.0](https://github.com/illuminate/validation/blob/v9.11.0/Concerns/ValidatesAttributes.php#L578) (look at ```preg_match```):

```php
public function validateDigitsBetween($attribute, $value, $parameters)
{
    $this->requireParameterCount(2, $parameters, 'digits_between');

    $length = strlen((string) $value);

    return ! preg_match('/[^0-9.]/', $value)
                && $length >= $parameters[0] && $length <= $parameters[1];
}
```

I found that this validation rule looks like that in these versions: v6.20.44, v8.79.0 (and newer, including v9.x.x)

This was fun. No hard feelings about my presentation of the issue, I hope. :wink: 

### Steps To Reproduce:

```php
$data = [
    'pin' => '1234.78',
    // 'pin' => '1234..78', // ok, weird
    // 'pin' => '....',     // OMG
];

$validator = \Illuminate\Support\Facades\Validator::make($data, [
    'pin' => 'digits_between:4,8',
]);

if (!$validator->fails()) {
    echo 'These are not the digits you are looking for.';
} else {
    echo 'When you see this, the validation rule digits_between is now fixed.';
}
```

driesvints:
I don't think we're going to change this any time soon, sorry.

driesvints:
Also see the original issue for that: https://github.com/laravel/framework/issues/40264. This should be a numeric value. Maybe digit isn't an 100% word but the rule atm does what it's suppose to do.

driesvints:
Ah the multiple dots thing is a concern. I'll send in a PR for that.

driesvints:
[...]

PR #42330, opened 2022-05-10
[8.x] Fix multiple dots for digits_between rule

driesvints:
This makes the `digits_between` rule a bit more strict by validating for multiple dots in a number (`2.2.3`, `2..3`, `...`).

Fixes https://github.com/laravel/framework/issues/42326

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -582,6 +582,11 @@ trait ValidatesAttributes
 
         $length = strlen((string) $value);
 
+        // Make sure only one dot at max is present...
+        if (($length - strlen(str_replace('.', '', (string) $value))) > 1) {
+            return false;
+        }
+
         return ! preg_match('/[^0-9.]/', $value)
                     && $length >= $parameters[0] && $length <= $parameters[1];
     }
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -582,6 +582,10 @@ trait ValidatesAttributes
 
         $length = strlen((string) $value);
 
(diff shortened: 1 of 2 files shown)


taai:
Possibilities to consider (for example, ```digits_between:1,10```):

* single dot
* dot at the end (some may argue that it is valid, not me)
* dot at the beginning (some may argue that it is valid, not me)

taai:
* I suggest using ```substr_count($value, '.')``` to count the dots
* Why to convert/cast the value to string so many times?
* What if the value is not a string and is not a number, but is an array?
* I suggest doing simple, lightweight stuff (comparison operations) before doing hard stuff (like regex)
* I suggest using ```ctype_digit()``` which is significantly faster than regex


:warning: N.B. Should also allow dots in validation rule ```digits``` – for consistency...

My solution:

```php
public function validateDigitsBetween($attribute, $value, $parameters)
{
    $this->requireParameterCount(2, $parameters, 'digits_between');

    // make sure that the value is a number or a numeric string
    if (!is_numeric($value)) {
        return false;
    }

    if (is_string($value)) {
        // make sure that the value is not a dot and does not contain more than single dot
        if ($value === '.' || substr_count($value, '.') > 1) {
            return false;
        }
    } else {
        $value = strval($value);
    }

    $length = strlen($value);

    return $length >= $parameters[0] && $length <= $parameters[1]
        && ctype_digit(str_replace('.', '', $value));
}

public function validateDigits($attribute, $value, $parameters)
{
    $this->requireParameterCount(1, $parameters, 'digits');

    // make sure that the value is a number or a numeric string
    if (!is_numeric($value)) {
        return false;
    }

    if (is_string($value)) {
        // make sure that the value is not a dot and does not contain more than single dot
        if ($value === '.' || substr_count($value, '.') > 1) {
            return false;
        }
    } else {
        $value = strval($value);
    }

    return strlen($value) == $parameters[0]
        && ctype_digit(str_replace('.', '', $value));
}
```

PR #42331, opened 2022-05-10
Support type hinting when resolving manually

mewejo:
When it's impracticable to resolve a service via a constructor, using the `resolve()` function (or similar) must be used. This works, but makes it hard to find usage of the service and benefit from type hints without manually docblocking the service, which may be missed.

## Existing

```php 
// Having to manually docblock a service each time it's resolved to benefit from
// type hinting, clicking through in an IDE, finding usage etc.

/** @var MyServiceInterface $myService */
$myService = resolve(MyServiceInterface::class);

$myService->doThatThing();
```

## Proposed
```php
// $myService is inferred as an instance of MyServiceInterface automatically
$myService = resolve(MyServiceInterface::class);

// This also allows calls directly on the resolve() function result with type hinting
resolve(MyServiceInterface::class)->doThatThing();
```
Docs: https://phpstan.org/writing-php-code/phpdoc-types#class-string

This will not work if using string names and not class names for services in the container, for example `router`. It might be worth having another function which is exclusively used for services in the container which are bound using class names.

**_This does not change any PHP code, just docblock comments on the `resolve()` function._**

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -679,12 +679,13 @@ if (! function_exists('rescue')) {
 
 if (! function_exists('resolve')) {
     /**
-     * Resolve a service from the container.
-     *
-     * @param  string  $name
-     * @param  array  $parameters
-     * @return mixed
-     */
+    * Resolve a service from the container.
+    *
+    * @param    class-string<T>  $name
+    * @param    array            $parameters
+    * @return   T
+    * @template T
+    */
     function resolve($name, array $parameters = [])
     {
         return app($name, $parameters);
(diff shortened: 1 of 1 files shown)


mewejo:
I'm not going to spend time appeasing StyleCI unless this PR is something that would be considered

taylorotwell:
As you noted - not only class names are bound in the container. You can write and docblock your own helper for this.

src/Illuminate/Foundation/helpers.php
```diff
@@ -681,9 +681,10 @@ function rescue(callable $callback, $rescue = null, $report = true)
     /**
      * Resolve a service from the container.
      *
-     * @param  string  $name
-     * @param  array  $parameters
-     * @return mixed
+     * @param    class-string<T>  $name
```
mfn, line 684: > This will not work if using string names

Maybe using this would solve at least "half" of the problems, i.e. those cases indeed using a class?
```php
      * @param    class-string<T>|string  $name
```

PR #42339, opened 2022-05-10
Add strict validation style for validation rule digits_between

taai:
Today I learned that since PR #40278 the validation rule ```digits_between``` allows the value to contain also a dot (```.```). I think it is wrong, because a dot is not a digit. But what is done, is done…
I discussed it in issue #42326 but I failed to persuade Laravel core member to roll back that PR. Breaking changes must be prevented… At least, I was encouraged to make a PR to achieve my goal, so this is it.

I created a third parameter for the validation rule ```digits_between``` – the validation style ```strict```. This is a non-breaking feature and, as written in documentation, can be targeted for 9.x branch.

To validate by using ```strict``` validation style and ensure that the value contains digits only:
```php
$data = [
    'pin' => '1234',
];

$validator = \Illuminate\Support\Facades\Validator::make($data, [
    'pin' => 'digits_between:4,8,strict',
]);

if ($validator->passes()) {
    echo 'Success!';
} else {
    echo 'The PIN number must contain 4-8 digits.';
}
```

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -579,19 +579,30 @@ trait ValidatesAttributes
     {
         $this->requireParameterCount(2, $parameters, 'digits_between');
 
-        $length = strlen((string) $value);
-
-        if (((string) $value) === '.') {
+        if (! is_numeric($value)) {
             return false;
         }
 
-        // Make sure there is not more than one dot...
-        if (($length - strlen(str_replace('.', '', (string) $value))) > 1) {
+        if (is_string($value)) {
+            // value must not be a dot or contain more than single dot
+            if ($value === '.' || substr_count($value, '.') > 1) {
+                return false;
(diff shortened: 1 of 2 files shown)


taylorotwell:
If you need custom validation behavior just pass a Closure rule that does exactly what you want for your own application's needs. Or combine digits_between with "int" rule.

taai:
@taylorotwell But this is not a custom validation behavior that I'm talking about, I'm trying to fix what was broken 4 months ago. And a dot is not a digit. 🤷

I wrote many arguments about this in the issue #42326 – I suggest you to read it.

If you insist that the validation rule ```digits_between``` must allow value to contain a dot, then the validation rule ```digits``` should too – for consistency ™. I will create a PR for that, if you want.

By the way, you can't validate value ```0000``` with validation rule ```integer```, because this code (that is used internally in that validation rule) will return ```false```:
```php
$value = '0000';
$isValid = filter_var($value, FILTER_VALIDATE_INT) !== false; // false
```

PR #42340, opened 2022-05-10
[9.x] Fix type mismatch on Pusher::validate_channels()

lucasvscn:
- Laravel 9.12.0
 - Pusher 7.0

This PR fixes an issue with the most recent version of Pusher package.

After PR #42287, the argument passed on to `Pusher\Pusher::trigger()` was changed to a Collection. But since version 7.0 Pusher requires an argument of type array down the stack.

Trying to broadcast an event using Pusher 7.0 on Laravel 9.12.0 raises the following error:

`Pusher\\Pusher::validate_channels(): Argument #1 ($channels) must be of type array, Illuminate\\Support\\Collection given, called in /vendor/pusher/pusher-php-server/src/Pusher.php on line 341`


taylorotwell:
Released 9.12.1

PR #42341, opened 2022-05-10
[9.x] Add `is_string` check to `QueriesRelationships@requalifyWhereTables`

### LINKED ISSUE #42338: [9.12.0] Enum bug within relationship
- Laravel Version: 9.12.0
- PHP Version: 8.1.5
- Database Driver & Version: MariaDB 

### Description:
I suddenly get these errors after this commit (laravel v9.12.0), commit #42100.
v9.11.0 and earlier works correct...

```
str_starts_with(): Argument #1 ($haystack) must be of type string, App\Models\Enums\ProductType given
```

Seems to be caused by this line, https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php#L786=

### Steps To Reproduce:
I have a product model, with a product_type with an enum. The problem appears when doing something like this

--
This is located in my Product model

```
/**
     * Get all sub products
     *
     * @return \Illuminate\Database\Eloquent\Relations\hasMany
     */
    public function simpleProducts()
    {
        return $this->hasMany(self::class, 'parent_id', 'id')
            ->where('active', 1)
            ->where('product_type', ProductType::SIMPLE)
            ->orderBy('sort');
    }
```

ProductType enum
```
namespace App\Models\Enums;

enum ProductType: string
{
    case SIMPLE = 'simple';
    case GROUPED = 'grouped';
    case CONFIGURABLE = 'configurable';
}
```

```
return Product::withCount('simpleProducts')
    ->get();
```

Calling this works correct....

```
return Product::where('product_type', ProductType::SIMPLE)
    ->get();
```

rodrigopedra:
Closes #42338

This PR:

- Adds a `is_string()` check before a `str_starts_with()` on `QueriesRelationships@requalifyWhereTables`
- Adds a test case which would fail without this patch

Notes:

- The `QueriesRelationships@requalifyWhereTables` was added by PR #42100 which tried to fix table prefixing issues when using self-referencing relations
- There is no need to check non-string values, as the idea is to prefix values used in `whereColumn`, or other that could reference related tables
- The added test case used a `new stdClass` as enums are not available in PHP 8. As it tests the generated SQL statement, the `stdClass` instance never acutually gets bound to PDO.

EDIT: added information about added test case. Please read 3rd note about it.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -783,7 +783,7 @@ trait QueriesRelationships
     {
         return collect($wheres)->map(function ($where) use ($from, $to) {
             return collect($where)->map(function ($value) use ($from, $to) {
-                return str_starts_with($value, $from.'.')
+                return is_string($value) && str_starts_with($value, $from.'.')
                     ? $to.'.'.Str::afterLast($value, '.')
                     : $value;
             });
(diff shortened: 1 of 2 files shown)


rodrigopedra:
ping @driesvints , @taylorotwell 

P.S.: I didn't add any tests as this is more like a patch bug-fix. I can try adding any later.

PR #34998, opened 2020-10-27
[9.x] Added push method to Cache Repository 

foremtehan:
Hello laravel,

This PR added `push()` method to Cache Repository which allow to pushing items into exist cache key and return the items in collection instance.

```php
//before
Cache::put('foo', 'bar');

$key = Cache::get('foo');

$key[] = 'baz';

Cache::put('foo', $key);

Cache::get('foo');
// ['bar', 'baz]
```

```php
//after
Cache::push('items.key', 'bar');
// ['bar']

$items = Cache::push('items.key', 'baz');
// ['bar', 'baz]
```

PR #42342, opened 2022-05-10
[9.x] Adds cache `refresh()` (now its final)

DarkGhostHunter:
# What?

> This comment has been edited to reflect new development.

Retrieves an item from the Cache and stores back the new value, refreshing its lifetime.

```php
use Illuminate\Support\Facades\Cache;
use App\Models\Message;

public function storeMessage(Message $new)
{
    // Before
    Cache::lock('messages:lock')->block(10, function () use ($new) {
        $messages = Cache::get('messages', collect());
        
        $messages->put($new);
        
        Cache::put('messages', $messages, 60);
    });
    
    // After
    Cache::refresh('messages', fn ($messages) => collect($messages)->put($new), 60);
}
```

The expiration time can be computed within the callback using the second parameter.

```php
use Illuminate\Support\Facades\Cache;

Cache::refresh('token', function ($token, $expire) {
    $response = Http::post("https://oauth.service.com");

    // Compute your own TTL.
    $expire->at = Carbon::parse($response->json('expires_at'));

    // Store it forever
    $expire->never();

    // Remove it while returning the value.
    $expire->now();

    return $response;
});
```

Behind the scenes it uses the lock to avoid data races, with a waiting default of 10 seconds (same as the HTTP Client). Not issuing a callback returns a `RefreshOperation` instance to further change the lock configuration.

```php
use Illuminate\Support\Facades\Cache;

Cache::refresh('token')
    ->lock('my_lock_name', 15, 'i_own_this')
    ->waitFor(5)->push(function ($token) {
        // ...
    });
```

## Caveats?

Well, if the item doesn't exist, the callback argument is `null`. Since it's a callback, the developer has total control on what to do for a default value.

Works exclusively with cache locks.

## BC?

None, just adding a method... and a class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,31 +1,111 @@
 # Release Notes for 9.x
 
-## [Unreleased](https://github.com/laravel/framework/compare/v9.11.0...9.x)
+## [Unreleased](https://github.com/laravel/framework/compare/v9.13.0...9.x)
 
+
+## [v9.13.0](https://github.com/laravel/framework/compare/v9.12.2...v9.13.0) - 2022-05-17
+
+### Added
+- Added Illuminate/Collections/Traits/EnumeratesValues::value() ([#42257](https://github.com/laravel/framework/pull/42257))
+- Added new TestResponse helper: assertJsonMissingPath ([#42361](https://github.com/laravel/framework/pull/42361))
+- Added Illuminate/Support/Testing/Fakes/NotificationFake::assertCount() ([#42366](https://github.com/laravel/framework/pull/42366))
+- Added new DetectLostConnections ([#42377](https://github.com/laravel/framework/pull/42377), [#42382](https://github.com/laravel/framework/pull/42382))
+- Added Illuminate/Testing/TestResponse::collect() ([#42384](https://github.com/laravel/framework/pull/42384))
+- Added full callable support to schedule:list ([#42400](https://github.com/laravel/framework/pull/42400))
+- Added `Illuminate/Collections/Arr::map()` ([#42398](https://github.com/laravel/framework/pull/42398))
+
+### Fixed
+- Fixed PruneCommand finding its usage within other traits ([#42350](https://github.com/laravel/framework/pull/42350))
+- Fix assert that exception is thrown without message ([#42360](https://github.com/laravel/framework/pull/42360))
+
+### Changed
+- Skip parameter parsing for raw post body in HTTP Client ([#42364](https://github.com/laravel/framework/pull/42364))
+- Consistency between digits and digits_between validation rules ([#42358](https://github.com/laravel/framework/pull/42358))
+- Corrects the use of "failed_jobs" instead of "job_batches" in BatchedTableCommand ([#42389](https://github.com/laravel/framework/pull/42389))
+- Update email.blade.php ([#42388](https://github.com/laravel/framework/pull/42388))
+- Remove old monolog 1.x compat code ([#42392](https://github.com/laravel/framework/pull/42392))
+- SesTransport: use correct Tags argument ([#42390](https://github.com/laravel/framework/pull/42390))
+- Implement robust handling of forwarding of exception codes ([#42393](https://github.com/laravel/framework/pull/42393))
+
+
+## [v9.12.2](https://github.com/laravel/framework/compare/v9.12.1...v9.12.2) - 2022-05-11
+
+### Fixed
+- Factory fails to eval models and factories when returned from closure ([#42344](https://github.com/laravel/framework/pull/42344))
+
+### Changed
+- Added is_string check to QueriesRelationships@requalifyWhereTables ([#42341](https://github.com/laravel/framework/pull/42341))
+
+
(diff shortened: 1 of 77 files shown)


morloderex:
@DarkGhostHunter i really like this idea.

However, one use case I would use for this is to make a http call to retrieve and store a oAuth 2 client credential token. My only concern is that the ttl value cannot be made out of the actual value that got returned by the callback.

Many OAuth 2 servers comes with an `expires_at` which is usually in seconds which would need to be made into a datetime object out of the current time, and this would then also become the ttl value.

Can you think of a way to make this possible maybe?

DarkGhostHunter:
> Can this maybe be addressed in this pr?

Great idea. I can add an object that stores the item TTL, so you can change it, as second parameter. IMHO it's better than returning a an object with the item and a new TTL.

```php
use Illuminate\Cache\ItemExpiration;
use Illuminate\Support\Facades\Cache;

Cache::upsert('token', function ($token, ItemExpiration $expire) {
    $response = Http::post("https://oauth.service.com");

    $expire->at($response->json('expires_at'));

    return $response;
});
```

Using that object you could also add or remove the expiration time at will inside the function.

morloderex:
@DarkGhostHunter that would be a great addition! 👍

foremtehan:
I was trying to add some similar functionality called `push` but it got rejected https://github.com/laravel/framework/pull/34998

I think the 'updating existing items' is the missing feature in cache system 

https://stackoverflow.com/questions/48110089/pushing-value-to-existing-cache-value-in-laravel/48110145
https://github.com/laravel/ideas/issues/2606

DarkGhostHunter:
> I was trying to add some similar functionality called `push` but it got rejected https://github.com/laravel/framework/pull/34998
> 
> 
> 
> I think the 'updating existing items' is the missing feature in cache system 
> 
> 
> 
> https://stackoverflow.com/questions/48110089/pushing-value-to-existing-cache-value-in-laravel/48110145
> 
> https://github.com/laravel/ideas/issues/2606

I think mine has more chances of success because it covers more use cases than just pushing an array, and using `??=` on the item gives more _control_ con what the default value should be.

taylorotwell:
This does generally have a race condition since you are calling get and then put as two separate operations.

DarkGhostHunter:
> This does generally have a race condition since you are calling get and then put as two separate operations.

I think it's unavoidable unless I add a lock duration as third parameter, which could be 30 seconds by default.

```php
Cache::upsert('messages', function ($messages, $expire, $lock) {
    // ...
}, 60); // Feeling greedy.
```

This way the cache key gets locked, and with a `finally` we release it.

DarkGhostHunter:
> This does generally have a race condition since you are calling get and then put as two separate operations.

There is a way to use get & put in one operation? My other solution brings locking.

DarkGhostHunter:
What about this approach? Seems a little overkill but it works with locking only if the store supports it, otherwise it uses a simple get->put.

```php
$new = Message::latest()->first();

Cache::of('messages')->upsert(function ($messages, $expire) use ($new) {
    return Collection::wrap($messages)->add($new);
});
```

It supports managing the lock fluently.

```php
Cache::of('messages')->lockBy(20)->owner('my-app-1')->wait(10)->upsert( /* ... */ );
```

I updated this PR with this approach anyway. I suggest somebody checks the locking mechanism if it's okay.

taylorotwell:
Eh, on second thought I think it's mainly the method name that I think is a bit confusing. ALL Cache::put calls are effectively already "upserts" in the database sense. They update a key's value if it exists. If it doesn't, the key is created with the associated value. So, Cache::put is already "upsert". But, you are wanting a new method that conveniently lets you modify the retrieved value then put it back in the cache with one method call.

DarkGhostHunter:
> But, you are wanting a new method that conveniently lets you modify the retrieved value then put it back in the cache with one method call.

Exactly. _I hereby resign mine own liberty to nameth the method_, unless it's `getset()`.

I double down on the lock mechanism, if someone can check it's okay.

DarkGhostHunter:
Ready for review guys.

driesvints:
`Cache::refresh`?

DarkGhostHunter:
> `Cache::refresh`?

Noone else? Last call!

DarkGhostHunter:
It's `refresh()` then.

Since the developer **_expects_** an atomic operation (the value doesn't change while it's being manipulated), this will be done only if the store supports locking, otherwise an exception will be thrown.

✅ Doesn't calls the cache store if nothing to put.
✅ Doesn't calls the cache store if nothing to forget (expire = 0).
✅ Key lifetime can be computed within the callback.
✅ Lock wait defaults to 10 seconds.
✅ Lock and wait can be configured.
✅ Just a simple call away

Updated initial post to reflect changes. I 100% confident this is final.

driesvints:
@DarkGhostHunter Taylor will review this once he has time.

DarkGhostHunter:
May be “getset” is better. Well, naming things it’s not my forte.

I double down on the lock code review. 

Italo Baeza C.

PR #42352, opened 2022-05-11
[9.x] Fix casting nulls in AsCollection and AsArrayObject casts

bastien-phi:
These 2 casts actually set `"null"` in database instead of `null`.

Looking at https://laravel.com/docs/9.x/upgrade#custom-casts-and-null, it looks like these custom casts need to be updated.

This behaviour is actually already implemented in `AsEncryptedArrayObject` and `AsEncryptedCollection`


taylorotwell:
Actually I'm going to revert this. It could be a breaking change if the column is not nullable on existing applications.

bastien-phi:
I understand. Should it target 10.x branch instead ? The actual behavior of `AsCollection` and `AsArrayObject` feels weird...

PR #42353, opened 2022-05-11
[9.x] Translate "and X more errors" message for validator exception

bsaqqa:
# Problem

When building a multilingual website, the user gets an error message with the suffix "and 3 more errors" in the Form Request error messages. So I need to override the `ValidationException` class every time want to use the translated error messages.

# Solution
I have updated the `ValidationException` class in the `summarize` function with the below translatable keys. Therefore, I used the following two translation keys:
- `validation.invalid_data`  to the value message: `The given data was invalid.`
- `validation.additional` to the translation message: `and X more errors`.

You can use these keys in `languages/en.php` as below 

```
[
   ...,
  "validation.invalid_data" => "The given data was invalid.",
  "validation.additional" => "and :additional more :pluralized".
  ...
];
```

### DIFF
--- a/src/Illuminate/Validation/ValidationException.php
+++ b/src/Illuminate/Validation/ValidationException.php
@@ -88,15 +88,22 @@ class ValidationException extends Exception
         $messages = $validator->errors()->all();
 
         if (! count($messages)) {
-            return 'The given data was invalid.';
+            $message = trans('validation.invalid_data');
+
+            return $message === 'validation.invalid_data'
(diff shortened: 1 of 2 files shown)


driesvints:
Please see many many previous attempts

Pablo-Camara:
Left a question here in case anyone wants to help:
https://stackoverflow.com/questions/73898408/how-to-translate-laravel-hardcoded-string-and-1-more-error-using-dynamic

PR #42355, opened 2022-05-11
[9.x] Adds support to `HEAD` in route. `Route::head($uri, $action);`

medeirosinacio:
Hello, I need to create a route with the HEAD method to update a token, I saw that laravel did not have this support.

Here's a PR adding this feature. Please let me know if anything is missing and if this functionality can come in.

Thanks

` Route::head($uri, $action); `

### DIFF
--- a/src/Illuminate/Contracts/Routing/Registrar.php
+++ b/src/Illuminate/Contracts/Routing/Registrar.php
@@ -14,6 +14,15 @@ interface Registrar
     public function get($uri, $action);
 
     /**
+     * Register a new HEAD route with the router.
+     *
+     * @param  string  $uri
+     * @param  array|string|callable  $action
+     * @return \Illuminate\Routing\Route
+     */
+    public function head($uri, $action = null);
+
+    /**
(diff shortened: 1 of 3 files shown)


medeirosinacio:
> Reference to this new method would also need to be added to the `Route` facade:
> 
> https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/Facades/Route.php
> 
> I would also add test for sanity checking 👍

I will make the changes. thanks.

medeirosinacio:
> Reference to this new method would also need to be added to the `Route` facade:
> 
> https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/Facades/Route.php
> 
> I would also add test for sanity checking 👍

I included the method in the facade, but I couldn't find the place to create the tests. Do you have any references in a file?

taylorotwell:
I suggest just using the `get` method (which responds to head as well) and then if you really need specific HEAD handling to use the `$request->isMethod('head')` method inside your route.

PR #42356, opened 2022-05-11
Replace the magic number by a constant in the handle Command function…

melicerte:
The `php artisan make:command` maker builds a command class with a default `return 0` in the handle function.

Let's not use magic numbers and use the constants provided by Symfony Command instead.
Makes the code clearer, and points the develop to the existence of the other constants ::FAILURE and ::INVALID

As PHP 8 is now a pre-requesite to use Laravel, let's type-hint new created things.

### DIFF
--- a/src/Illuminate/Foundation/Console/stubs/console.stub
+++ b/src/Illuminate/Foundation/Console/stubs/console.stub
@@ -25,8 +25,8 @@ class {{ class }} extends Command
      *
      * @return int
      */
-    public function handle()
+    public function handle(): int
     {
-        return 0;
+        return self::SUCCESS;
     }
 }


sjokkateer:
Just being a messenger but the following PR seems similar [#42185](https://github.com/laravel/framework/pull/42185)

taai:
@melicerte A zero is not a magic number in CLI (command line interface) programs. Zero means that the program execution succeeded. This is [documented](https://www.php.net/exit) also in PHP manual.

melicerte:
> @melicerte A zero is not a magic number in CLI (command line interface) programs. Zero means that the program execution succeeded. This is [documented](https://www.php.net/exit) also in PHP manual.

Obviously, but from the PHP developer perspective, it is indeed a magic number.
Symfony understood that, and created some constants to be clear. I think it is best to avoid any ambiguity.

You could also say 403 and 421 are not magic numbers, they are well documented HTTP code responses. But from the PHP developer perspective, they are magic numbers.

browner12:
as stated above, this PR has been submitted and rejected multiple times already.

I would recommend closing yourself.

antonkomarev:
Is there any reason why all such PRs are closed?

taylorotwell:
I just don't think it makes the code "clearer". Most developers know 0 is a successful exit code for computer programs. I can appreciate constants being used throughout a codebase when there is some chance other underlying value changing and you want to be able to update that value in one place while the program can still use the constant. However, 0 will always have the same meaning in CLI programs. It is not changing.

melicerte:
It's a different philosophy from Symfony. Noted.

PR #42358, opened 2022-05-11
[8.x] Consistency between digits and digits_between validation rules

taai:
Bugfix. Makes validation rule ```digits``` consistent with the validation rule ```digits_between```.

### DIFF
--- a/tests/Validation/ValidationValidatorTest.php
+++ b/tests/Validation/ValidationValidatorTest.php
@@ -2272,6 +2272,30 @@ class ValidationValidatorTest extends TestCase
         $v = new Validator($trans, ['foo' => '2e7'], ['foo' => 'Digits:3']);
         $this->assertTrue($v->fails());
 
+        $v = new Validator($trans, ['foo' => '1.2'], ['foo' => 'digits:3']);
+        $this->assertTrue($v->passes());
+
+        $v = new Validator($trans, ['foo' => '0.9876'], ['foo' => 'digits:5']);
+        $this->assertTrue($v->fails());
+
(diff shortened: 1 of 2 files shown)


driesvints:
I guess if we want to make them the same that this is indeed the wtg yeah.

tests/Validation/ValidationValidatorTest.php
```diff
@@ -2272,6 +2272,30 @@ public function testValidateDigits()
...
+        $this->assertTrue($v->fails());
+
+        $v = new Validator($trans, ['foo' => '.'], ['foo' => 'digits:1']);
+        $this->assertTrue($v->fails());
+
+        $v = new Validator($trans, ['foo' => '.2'], ['foo' => 'digits:2']);
+        $this->assertTrue($v->passes());
+
+        $v = new Validator($trans, ['foo' => '2.'], ['foo' => 'digits:2']);
+        $this->assertTrue($v->passes());
```
driesvints, line 2297: Shouldn't this evaluate to `false`? The `.0` should basically be ignored I think and represent the integer `2`.

$v = new Validator($trans, ['foo' => '2.'], ['foo' => 'digits:2']);
$this->assertTrue($v->passes());

taai: If ```2.``` should be perceived as 1 character, shouldn't ```.2``` perceived as 3 characters?

As we are thinking about how these values would be evaluated (converted to floats), then other questions occur:
* what to do with ```0.0``` - should it be perceived as 1 character (```0```)?
* what to do with ```0000.0``` - should it be perceived as 1 character ('''0''')?

driesvints: You're right. Let's keep it as is.

PR #42362, opened 2022-05-11
[9.x] Adds password hash helper for Factories

DarkGhostHunter:
## What?

This small method will actually reuses the password hash to avoid creating it for each created model from the Factory. Should have major speedbumps when seeding 1.000+ models in a database. It uses `password` by default, and the application default hasher.

```php
public function definition()
{
    return [
        'name' => $this->faker->name,
        'password' => $this->password(), // "password"
        // ...
    ];
}
```

## How?

It saves the password (when called) into the Factory. This allows to hash a password only once when it makes/creates models.

## TODO?

The [`UserFactory.php` in laravel/laravel](https://github.com/laravel/laravel/blob/9.x/database/factories/UserFactory.php#L24) could be changed too to use the method instead of hard-coding the hash to _bcrypt_.

## BC?

Just adding a method and a property.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/Factory.php
+++ b/src/Illuminate/Database/Eloquent/Factories/Factory.php
@@ -9,7 +9,6 @@ use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Collection;
-use Illuminate\Support\Facades\Hash;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Conditionable;
 use Illuminate\Support\Traits\ForwardsCalls;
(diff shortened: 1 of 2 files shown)


driesvints:
How is this more performant than just using the hashed string like we do now?

Krisell:
I also think the current implementation is better from a security perspective. Of course, this should not be used in production but the hardcoded bcrypt hash stands out very clearly, whereas a `password()` helper, even with a very bad default password, might be missed in reviews.

taylorotwell:

DarkGhostHunter:
> How is this more performant than just using the hashed string like we do now?

For users or apps with different password requirements. For example, you may change `password` for `123asdAS#" or something if your app requires strong passwords with numbers or else.

Another way is to Artisan Tinker -> hash your new default password -> copy the hash, but feels less intuitive that using a password and saving the same hash in Factories.

I could have named it as `reuseHash('password')` tho.

Issue #36976, opened 2021-04-13
Http client parses request body with parse_str and runs into input variables exceeded limit.


briedis:
- Laravel Version:  8.37.0
- PHP Version: 8.0.2
- Database Driver & Version: Does not apply

### Description:

`\Illuminate\Http\Client\PendingRequest::parseRequestData` method uses `parse_str` if laravelData variable is a string, even if the contents are not a query string. This method is restricted by PHP ini setting `max_input_vars`. With some payloads (XML, e.g.) this can be easily exceeded by accident.

### Steps To Reproduce:

Generate a valid XML with "looks-like-query-params" contents.

```
$content = str_repeat("<tag>&amp;=</tag>", 1000);
$xml = '<?xml version="1.0" encoding="UTF-8"?><body>'. $content. '</body>';
ini_set('max_input_vars', '1000'); // This is usually the default setting
\Http::withBody($xml, 'application/octet-stream')->post('does-not-matter');
```

This code will trigger an exception: `parse_str(): Input variables exceeded 1000. To increase the limit change max_input_vars in php.ini.`

briedis:
I tried to pass a resource, since PHPDOC clearly stated that the withBody method accepts either resource or string.

```$content = str_repeat("<tag>&amp;=</tag>", 1000);
$xml = '<?xml version="1.0" encoding="UTF-8"?><body>'. $content. '</body>';

// Create a stream instead of a regular string
$xmlStream = fopen('php://memory', 'r+');
fwrite($xmlStream, $xml);
rewind($xmlStream);
        
\Http::withBody($xmlStream, 'application/octet-stream')->post('does-not-matter');
```
![image](https://user-images.githubusercontent.com/500491/114625534-10d2cd00-9cbb-11eb-872e-b2d3308e15d9.png)

But this dies with a different error somewhere really deep:

```TypeError : Illuminate\Http\Client\Request::withData(): Argument #1 ($data) must be of type array, resource given, called in /path/vendor/laravel/framework/src/Illuminate/Http/Client/PendingRequest.php on line 899
 /path/vendor/laravel/framework/src/Illuminate/Http/Client/Request.php:240
 /path/vendor/laravel/framework/src/Illuminate/Http/Client/PendingRequest.php:899
 /path/vendor/laravel/framework/src/Illuminate/Support/helpers.php:263
 /path/vendor/laravel/framework/src/Illuminate/Http/Client/PendingRequest.php:902
[...]

PR #42364, opened 2022-05-12
[9.x] Skip parameter parsing for raw post body in HTTP Client

### LINKED ISSUE #42349: HTTP Client unable to post raw body with more than a thousand ampersands
- Laravel Version: 9.12.1
- PHP Version: 8.1.5

### Description:
In a new Laravel install, running the following command fails with the error message below:
```php
Artisan::command('send', function () {
    Http::withBody(
        str_repeat('A text with a thousand ampersands &. ', 1000), 'text/html'
    )->post('https://some-url.com');
});
```

```
  ErrorException

  parse_str(): Input variables exceeded 1000. To increase the limit change max_input_vars in php.ini.

  at vendor/laravel/framework/src/Illuminate/Http/Client/PendingRequest.php:911
    907▕             $laravelData = (string) $urlString->after('?');
    908▕         }
    909▕
    910▕         if (is_string($laravelData)) {
  ➜ 911▕             parse_str($laravelData, $parsedData);
    912▕
    913▕             $laravelData = is_array($parsedData) ? $parsedData : [];
    914▕         }
    915▕
```

I understand that this could be solved by increasing `max_input_vars`, but I'm confused why the raw POST body is being parsed as query/form-data parameters at all. I'm not familiar enough with the HTTP Client to understand all details, but the problem might be related to testing-specific code (relevant when faking the client). 

By making the following change in `PendingRequest.php`, the request is successful and the correct data is sent.

```php
protected function parseRequestData($method, $url, array $options)
{
    $laravelData = $options[$this->bodyFormat] ?? $options['query'] ?? [];

>>   // This "solves" the problem, by ensuring the parse_str on line 911 is skipped.
>>    if ($this->bodyFormat === 'body') {
>>        return [];
>>   }

    ...

    if (is_string($laravelData)) {
        parse_str($laravelData, $parsedData);
```

The Client tests in the framework seem to pass with this change.

### Steps To Reproduce:

The example mentioned above is sufficient, and note that a thousand ampersands could commonly appear in HTML code being transmitted (as in our use case). Changing the content type to `text/plain` doesn't help.

```php
Http::withBody(
    str_repeat('A text with a thousand ampersands &. ', 1000), 'text/html'
)->post('https://some-url.com');
```

Krisell:
Fixes https://github.com/laravel/framework/issues/42349
Related to https://github.com/laravel/framework/issues/36976

When sending an `Http::withBody()` using the Http client, the raw body is parsed as query/form parameters. This seems wrong and is both unnecessary work and risks hitting the default limit of 1000 "input variables" in php.

```php
Http::withBody(str_repeat('A thousand &. ', 1000), 'text/plain')->post('https://some-url.com');
```

```
ErrorException

parse_str(): Input variables exceeded 1000. To increase the limit change max_input_vars in php.ini.
```

The limit of 1000 is quite easily hit by sending HTML, XML or similar text. 

The suggested solution here skips all parameter parsing of the body, also saving a fair bit of work. All tests pass, but is there a specific reason to parse the body this way, or might that have been introduced by mistake in order to parse query and form-data parameters?

Our use case is sending user-provided HTML to an external service (for text analysis) and while being very rare, we sometimes see the 1000 limit being hit via Bugsnag.

_Thanks to @pelmered for suggestion of simplified flow (early return)._

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -895,6 +895,10 @@ class PendingRequest
      */
     protected function parseRequestData($method, $url, array $options)
     {
+        if ($this->bodyFormat === 'body') {
+            return [];
+        }
+
         $laravelData = $options[$this->bodyFormat] ?? $options['query'] ?? [];
 
         $urlString = Str::of($url);
--- a/tests/Http/HttpClientTest.php
+++ b/tests/Http/HttpClientTest.php
@@ -152,6 +152,21 @@ class HttpClientTest extends TestCase
         $this->factory->withBody($body, 'application/json')->send('get', 'http://foo.com/api');
     }
 
+    public function testSendRequestBodyWithManyAmpersands()
+    {
+        $body = str_repeat('A thousand &. ', 10000);
+
+        $fakeRequest = function (Request $request) use ($body) {
+            self::assertSame($body, $request->body());
+
+            return ['my' => 'response'];
+        };
+
+        $this->factory->fake($fakeRequest);
+
+        $this->factory->withBody($body, 'text/plain')->send('post', 'http://foo.com/api');
+    }
+
     public function testUrlsCanBeStubbedByPath()
     {
         $this->factory->fake([
--- a/tests/Http/HttpClientTest.php
+++ b/tests/Http/HttpClientTest.php
@@ -154,7 +154,7 @@ class HttpClientTest extends TestCase
 
     public function testSendRequestBodyWithManyAmpersands()
     {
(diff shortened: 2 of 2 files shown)


driesvints:
Thanks @Krisell

nuernbergerA:
After this change, how am i supposed to send XML to a soap endpoint?

```php
Http::withBody($someXml, 'text/xml; charset=UTF8')->post('soap-endpoint-url');
```

Kuzuto:
Please add this update to Laravel 8.x

driesvints:
@Kuzuto 8.x isn't maintained anymore. Please upgrade to 9.x

PR #42365, opened 2022-05-12
[9.x] Avoid deprecated const DateTimeInterface::ISO8601

sneycampos:
This fixes the use of a deprecated const DateTimeInterface::ISO886 and switched to DateTimeInterface::ATOM

The const ISO8601 was left for backward compatibility and it's recommended to use ATOM const added since php 7.2


driesvints:
These are breaking changes.

PR #42370, opened 2022-05-12
[9.x] Update Encrypter.php

diadal:
enforce decrypt payload to string, fix for below error

`base64_decode(): Passing null to parameter #1 ($string) of type string is deprecated in vendor/laravel/framework/src/Illuminate/Encryption/Encrypter.php on line 208`

### DIFF
--- a/src/Illuminate/Encryption/Encrypter.php
+++ b/src/Illuminate/Encryption/Encrypter.php
@@ -148,6 +148,7 @@ class Encrypter implements EncrypterContract, StringEncrypter
      */
     public function decrypt($payload, $unserialize = true)
     {
+        $payload = strval($payload);
         $payload = $this->getJsonPayload($payload);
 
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
Not passing a string is not a valid way to call this function. One should expect it to crash if you pass null.

diadal:
> Not passing a string is not a valid way to call this function. One should expect it to crash if you pass null.

null value never crashes only keep logging PHP 8.1 I will advise conversion of `null to ''`

Screen Shot 2022-05-26 at 6 42 03 AM

diadal:
> Not passing a string is not a valid way to call this function. One should expect it to crash if you pass null.

you can check with PHP 8.0 with no logging but PHP 8.1 keep logging without crashes and is not a bad idea to convert `null to ''` in `Encrypter.php`

PR #42371, opened 2022-05-12
[9.x] Direct routes for resource controller methods

medilies:
Add methods to the `Illuminate\Routing\Router` class that target routes generated by the `resource` method individually.

 The `resource` method is very convenient since it makes for us conventional routes and avoid us some typos, but it feels like an overkill when the resource is touched by only one or two methods. I know about the `except` and `only` chainable methods but I still feel that this PR offers a terse syntax :)

```php
// Doing
Route::create('entity', EntityController::class);
Route::update('entity', EntityController::class);

// Instead of:
Route::resource('entity', EntityController::class)->only(['create', 'store']);
```

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -334,6 +334,90 @@ class Router implements BindingRegistrar, RegistrarContract
     }
 
     /**
+     * Route a resource to a controller index method.
+     *
+     * @param  string  $name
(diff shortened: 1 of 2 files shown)


ankurk91:
Too much magic to save some keystrokes. 
It does not solve any problem in framework.

Such PRs should be discouraged.

taylorotwell:

PR #42374, opened 2022-05-13
[9.x] PHPStan support for union methods

fagai:
The union method is supports EloquentBuilder, but an error occurs because there is no description.

```php
$first = User::where('id',8);

$users = User::where('id',2)->union($first)->get();
```

```
Parameter #1 $query of method Illuminate\Database\Eloquent\Builder<Illuminate\Database\Eloquent\Model>::union() expects Closure|Illuminate\Database\Query\Builder, Illuminate\Database\Eloquent\Builder given.
```

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -2438,7 +2438,7 @@ class Builder implements BuilderContract
     /**
      * Add a union statement to the query.
      *
-     * @param  \Illuminate\Database\Query\Builder|\Closure  $query
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder|\Closure  $query
      * @param  bool  $all
      * @return $this
      */
@@ -2458,7 +2458,7 @@ class Builder implements BuilderContract
     /**
      * Add a union all statement to the query.
      *
-     * @param  \Illuminate\Database\Query\Builder|\Closure  $query
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder|\Closure  $query
      * @return $this
      */
     public function unionAll($query)


GrahamCampbell:
This is not necessary. Including subtypes of types already in the union adds no extra information. It like having a the type `Person` and then having `Person|Man`. The types are already identical, and `|Man` should be removed.

fagai:
@GrahamCampbell 
`Illuminate\Database\Eloquent\Builder` does not inherit from `Illuminate\Database\Query\Builder`.
It didn't look like a subtype, but is it different?

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Builder.php#L30
https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Query/Builder.php#L29

src/Illuminate/Database/Query/Builder.php
```diff
@@ -2438,7 +2438,7 @@ protected function removeExistingOrdersFor($column)
     /**
      * Add a union statement to the query.
      *
-     * @param  \Illuminate\Database\Query\Builder|\Closure  $query
+     * @param  \Illuminate\Database\Query\Builder|\Illuminate\Database\Eloquent\Builder|\Closure  $query
```
nunomaduro, line 2441: This is not correct, for the reason @GrahamCampbell mentioned:

> This is not necessary. Including subtypes of types already in the union adds no extra information.

n1215:
`Illuminate\Database\Eloquent\Builder ` is not a subtype of `Illuminate\Database\Query\Builder`.

fagai:
@GrahamCampbell @nunomaduro 
Not a subtype...

src/Illuminate/Database/Query/Builder.php
zonuexe: @nunomaduro 
[`Illuminate\Database\Eloquent\Builder`](https://github.com/illuminate/database/blob/v9.12.2/Query/Builder.php#L29) and [`Illuminate\Database\Eloquent\Builder`](https://github.com/illuminate/database/blob/v9.12.2/Query/Builder.php#L29) are not subclassed to each other, so they are not what @GrahamCampbell calls `Person|Man`.

The code that @fagai exemplifies in the issue is valid code that actually works.

＊　https://phpsandbox.io/n/odd-queen-wkk4-prgxg?files=%2Froutes%2Fweb.php

Is there a reasonable reason to exclude `Illuminate\Database\Eloquent\Builder` from PHPDoc?

Issue #42375, opened 2022-05-13
TypeError when using Redis pipeline


timndus:
- Laravel Version: 9.1.8
- PHP Version: 8.1.2
- Redis Version: 6.2.6

### Description:
https://github.com/laravel/framework/blob/016b0c1801f4c6b0bb79b0d82791f4853ed19fc5/src/Illuminate/Redis/Connections/PhpRedisConnection.php#L117  

This will try to get values of a Redis object instead of an array if we use the HMGET in a PIPELINE 
 
`array_values(): Argument #1 ($array) must be of type array, Redis given`  

  -----
  
  
detail:  
```
#0 /var/www/html/vendor/laravel/framework/src/Illuminate/Redis/Connections/PhpRedisConnection.php(117): array_values()
#1 /var/www/html/vendor/laravel/framework/src/Illuminate/Redis/RedisManager.php(276): Illuminate\Redis\Connections\PhpRedisConnection-&gt;hmget()
#2 /var/www/html/vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php(337): Illuminate\Redis\RedisManager-&gt;__call()
#3 /var/www/html/routes/api.php(21): Illuminate\Support\Facades\Facade::__callStatic()  
```  

### Steps To Reproduce:
```
use Illuminate\Support\Facades\Redis;
use Illuminate\Support\Facades\Route;

Route::get('/test', function (){
    Redis::HMSET(
        'test',
        [
            'name' => 'timndus',
            'age' => '64',
            'color' => 'green'
        ]
    );

    Redis::PIPELINE();
    Redis::HMGET('test', ['name', 'age']);
    $test = Redis::EXEC();
    
    return $test;
});
```

driesvints:
Please fill out the issue template when submitting issues.

timndus:
> Please fill out the issue template when submitting issues.

Sorry about that, I updated the issue. Can u open it or should I submit a new one? @driesvints

eduance:
@driesvints I can have something ready by tonight!

timndus:
> Thanks. Can you maybe send in a PR for this?

Here is a PR, but honestly I'm not sure of my solution
https://github.com/laravel/framework/pull/42383

eduance:
@timndus This is not the correct usage of the pipeline method, after some studying and reading into the predis test suite, this is how a pipeline should be built:

``
    $pipe = Redis::pipeline();

[...]

PR #42383, opened 2022-05-13
[8.x] Fix/phpredis hmget

timndus:
First of all, there is a difference between the Redis HMGET command result and the PhpRedis client HMGET result.
Redis HMGET returns only the values of the hash:
```
redis:6379> HSET myhash field1 "Hello"
(integer) 1
redis:6379> HSET myhash field2 "World"
(integer) 1
redis:6379> HMGET myhash field1 field2 nofield
1) "Hello"
2) "World"
3) (nil)
```
But PhpRedis returns values **and keys**:
```
$redis->hSet('h', 'field1', 'value1');
$redis->hSet('h', 'field2', 'value2');
$redis->hMGet('h', ['field1', 'field2']); /* returns ['field1' => 'value1', 'field2' => 'value2'] */
```
To handle this, the current approach tries to manipulate the result of PhpRedis HMGET and extract only the values from it(using array_values function), which is works perfectly till we use pipeline (and of course transactions).
https://github.com/laravel/framework/blob/016b0c1801f4c6b0bb79b0d82791f4853ed19fc5/src/Illuminate/Redis/Connections/PhpRedisConnection.php#L117
When using pipeline the problem is the current approach tries to manipulate data before executing the EXEC command which results in the following error described in the issue https://github.com/laravel/framework/issues/42375

`array_values(): Argument #1 ($array) must be of type array, Redis given`

We can simply solve this by passing exact result we got from the PhpRedis client
----
My solution probably is wrong because I see the main developer tries to provide a identical interface for PhpRedis and Predis but my solution forced me to separate the HMGET tests for PhpRedis and Predis
I think they(PhpRedis) should be responsible for keeping the client compatible with the latest Redis changes, not us(Laravel).

### DIFF
--- a/src/Illuminate/Redis/Connections/PhpRedisConnection.php
+++ b/src/Illuminate/Redis/Connections/PhpRedisConnection.php
@@ -115,7 +115,7 @@ class PhpRedisConnection extends Connection implements ConnectionContract
             $dictionary = $dictionary[0];
         }
 
-        return array_values($this->command('hmget', [$key, $dictionary]));
+        return $this->command('hmget', [$key, $dictionary]);
     }
 
     /**
--- a/tests/Redis/RedisConnectionTest.php
+++ b/tests/Redis/RedisConnectionTest.php
@@ -434,21 +434,36 @@ class RedisConnectionTest extends TestCase
         }
     }
 
-    public function testItGetsMultipleHashFields()
+    public function testItGetsMultipleHashFieldsUsingPredis()
     {
-        foreach ($this->connections() as $redis) {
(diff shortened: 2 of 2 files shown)


timndus:
> You say your solution is probably wrong but the PR is marked as ready for review. What do you want me to do?

means I'm not sure of the solution and need someone else opinion on this. just take a look at it now that you are here, please :)

eduance:
According to redis, php-redis and predis, this method is currently functioning properly:

```
HMGET myhash field1 field2 nofield
1) "Hello"
2) "World"
3) (nil)`
```

I have explained the cause of the bug in #42375 as pipeline was not used properly.

PR #42385, opened 2022-05-13
[9.x] Add notNullableTimestamps method to Blueprint

Anton5360:
This method is actually useful for developers who prefer sometimes to use `DB facade` rather than `Eloquent`.
**Benefits:**
1. It prevents developer`s duty to pass timestamps manually
2. It prevents developer`s need to create Trait contained this method every single project


taylorotwell:

PR #42388, opened 2022-05-14
[9.x] Update email.blade.php

ghost:
Change switch to match

### DIFF
--- a/src/Illuminate/Notifications/resources/views/email.blade.php
+++ b/src/Illuminate/Notifications/resources/views/email.blade.php
@@ -20,7 +20,7 @@
 @isset($actionText)
 <?php
     $color = match ($level) {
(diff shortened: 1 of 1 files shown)


ankurk91:
`success` and `error` both must have  $level

src/Illuminate/Notifications/resources/views/email.blade.php
```diff
@@ -19,14 +19,11 @@
...
-        case 'success':
-        case 'error':
-            $color = $level;
-            break;
-        default:
-            $color = 'primary';
-    }
+    $color = match ($level) {
+        'success' => $level,
+        'error' => $level,
```
faissaloux, line 24:
```suggestion
        'success', 'error' => $level,
```

PR #42393, opened 2022-05-15
[9.x] Implement robust handling of forwarding of exception codes

GrahamCampbell:
`getCode(): string|int` but `__construct(..., int $code, ...)`. The most sensible thing to do is to attempt to only forward on integer exception codes, and just default back to zero, otherwise. String exception codes (at least ignoring janky 3rd party extensions), can only be found amongst the PHP database exceptions, and for those, we should consider the code to be zero for the purposes of forwarding - casting to `1` doesn't make any sense.

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -706,7 +706,7 @@ class Container implements ArrayAccess, ContainerContract
                 throw $e;
             }
 
-            throw new EntryNotFoundException($id, $e->getCode(), $e);
+            throw new EntryNotFoundException($id, is_int($e->getCode()) ? $e->getCode() : 0, $e);
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/Illuminate/Container/Container.php
```diff
@@ -706,7 +706,7 @@ public function get(string $id)
                 throw $e;
             }
 
-            throw new EntryNotFoundException($id, $e->getCode(), $e);
+            throw new EntryNotFoundException($id, is_int($e->getCode()) ? $e->getCode() : 0, $e);
```
simonhamp, line 709: Couldn't you assign the result of the function call to save calling it twice when it is an `int`?

```suggestion
            throw new EntryNotFoundException($id, is_int($code = $e->getCode()) ? $code : 0, $e);
```

Appreciate this is a _very_ minor optimisation in this case.

PR #42394, opened 2022-05-15
[9.x] Adds bootable traits to TestCase

pascalbaljet:
This PR brings the *bootable trait* feature from Eloquent over to the Foundation TestCase. 

Currently, the `setUpTraits` method checks several traits like `RefreshDatabase` and `DatabaseMigrations`, and then calls a method if the TestCase uses that trait. If you want to do this for custom traits, you must override the `setUp` or `setUpTraits` method. With this PR, it checks whether the trait has a *bootTraitName* method, and then calls that method.

```php
trait RefreshSomeService
{
    public function bootRefreshSomeService()
    {
        SomeService::refresh();
    }
}

class SomeServiceTest extends TestCase
{
    use RefreshDatabase;
    use RefreshSomeService;
}
```

### DIFF
--- a/src/Illuminate/Foundation/Testing/TestCase.php
+++ b/src/Illuminate/Foundation/Testing/TestCase.php
@@ -143,6 +143,14 @@ abstract class TestCase extends BaseTestCase
             $this->setUpFaker();
         }
 
+        foreach ($uses as $trait) {
+            $method = 'boot'.class_basename($trait);
+
+            if (method_exists($this, $method)) {
+                $this->{$method}();
+            }
+        }
+
         return $uses;
     }
 
--- a/src/Illuminate/Foundation/Testing/TestCase.php
+++ b/src/Illuminate/Foundation/Testing/TestCase.php
@@ -144,7 +144,7 @@ abstract class TestCase extends BaseTestCase
         }
 
         foreach ($uses as $trait) {
-            $method = 'boot'.class_basename($trait);
+            $method = 'setUp'.class_basename($trait);
 
             if (method_exists($this, $method)) {
                 $this->{$method}();
--- a/tests/Foundation/Testing/BootTraitsTest.php
+++ b/tests/Foundation/Testing/BootTraitsTest.php
@@ -10,7 +10,7 @@ trait TestTrait
 {
     public $booted = false;
 
-    public function bootTestTrait()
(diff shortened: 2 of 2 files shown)


src/Illuminate/Foundation/Testing/TestCase.php
```diff
@@ -143,6 +143,14 @@ protected function setUpTraits()
             $this->setUpFaker();
         }
 
+        foreach ($uses as $trait) {
+            $method = 'boot'.class_basename($trait);
```
markhuot, line 147: I’ve implemented something similar and I love this! Only critique: since this is getting called from the `setUp` method I feel like it should call `setUpTraitName`. Boot has a pretty specific meaning within the application lifecycle and this actually fires after the booting is done.

taylorotwell:
There was a similar PR by @DarkGhostHunter: https://github.com/laravel/framework/pull/39883/files

His supported setUp and tearDown in the traits, which may be useful? What are the use cases for this type of thing? What kind of traits are you writing?

DarkGhostHunter:
My PR tried to deal only with first level traits, because PHPUnit includes its own set of traits and there is no guarantee it uses / will use a random `bootSomething` or `initializeSomething`. I preferred only to use first-level traits and initialize these instead.

```php
foreach (class_uses(static::class) as $trait) {
    $method = "initialize$trait";

    if (method_exists($method)) {
        $this->{$method}();
    }
}
```

pascalbaljet:
I have Dusk tests that upload a file directly to an S3 bucket (similar to Vapor with `Vapor.store`). I can't swap the filesystem disk to a local one as I'm using signed multipart uploads requiring specific S3 API requests. I'm using a local S3-compatible [MinIO](https://min.io/) server to prevent uploading to a live service.

MinIO is fast and lightweight, so I've created a trait that boots up a new server for each test and destroys it when the test has finished. I don't have to think about starting the server manually before running the test suite, and it only boots on tests that use the trait.

I haven't thought about a tearDown method, as I register a callback using the `beforeApplicationDestroyed` method, just like most of the Laravel traits that interact with the database. If you want support for a tearDown method, I will gladly update the PR :)

taylorotwell:
Renamed to `setUpFoo`. Thanks.

PR #42395, opened 2022-05-16
[8.x] Retain the original attribute value during validation of an array key with a dot for correct failure message

chiefey:
During validation of an array key with a dot, using a `integer|min:1` rule - the package will incorrectly return `validation.min.string` instead of the expected `validation.min.numeric`.

Illuminate\Validation\Validator replaces the placeholder in the attribute a bit prematurely in the [addFailure](https://github.com/laravel/framework/blob/f3940f8cd42bae1fe38e21400c19126ba8ff030d/src/Illuminate/Validation/Validator.php#L870) function. 

Code will continue to [makeReplacements](https://github.com/laravel/framework/blob/f3940f8cd42bae1fe38e21400c19126ba8ff030d/src/Illuminate/Validation/Validator.php#L881), [getMessage](https://github.com/laravel/framework/blob/f3940f8cd42bae1fe38e21400c19126ba8ff030d/src/Illuminate/Validation/Concerns/FormatsMessages.php#L49), [getSizeMessage](https://github.com/laravel/framework/blob/f3940f8cd42bae1fe38e21400c19126ba8ff030d/src/Illuminate/Validation/Concerns/FormatsMessages.php#L175), [getAttributeType](https://github.com/laravel/framework/blob/f3940f8cd42bae1fe38e21400c19126ba8ff030d/src/Illuminate/Validation/Concerns/FormatsMessages.php#L193), [hasRule](https://github.com/laravel/framework/blob/f3940f8cd42bae1fe38e21400c19126ba8ff030d/src/Illuminate/Validation/Validator.php#L1007), [getRule](https://github.com/laravel/framework/blob/f3940f8cd42bae1fe38e21400c19126ba8ff030d/src/Illuminate/Validation/Validator.php#L1019)

You can see down the line you are searching a with-placeholder rules array for a without-placeholder key. When the numeric search fails, it will default to string.

My fix retains that original attribute value including the placeholder to return the expected failure message

### DIFF
--- a/src/Illuminate/Validation/Concerns/FormatsMessages.php
+++ b/src/Illuminate/Validation/Concerns/FormatsMessages.php
@@ -20,6 +20,10 @@ trait FormatsMessages
      */
     protected function getMessage($attribute, $rule)
     {
+        $attributeWithPlaceholders = $attribute;
+
+        $attribute = $this->replacePlaceholderInString($attribute);
+
         $inlineMessage = $this->getInlineMessage($attribute, $rule);
 
         // First we will retrieve the custom message for the validation rule if one
@@ -46,7 +50,7 @@ trait FormatsMessages
         // specific error message for the type of attribute being validated such
         // as a number, file or string which all have different message types.
         elseif (in_array($rule, $this->sizeRules)) {
-            return $this->getSizeMessage($attribute, $rule);
+            return $this->getSizeMessage($attributeWithPlaceholders, $rule);
         }
 
         // Finally, if no developer specified messages have been set, and no other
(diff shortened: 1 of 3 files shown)


taylorotwell:
I don't think it makes sense to keep the place-holders in place when passing the attribute name to `getMessage`? It then checks the custom attribute array for that attribute using the place-holders, which doesn't make sense to me.

chiefey:
@taylorotwell ok if we are worried about the custom attributes - I will isolate this change to the `getSizeMessage` path inside `getMessage` and resubmit.

chiefey:
With this update the custom attribute array is left as-is. Only the `getSizeMessage` path uses the attributeWithPlaceholders value.

In this way we prevent a call to [getRule](https://github.com/laravel/framework/blob/f3940f8cd42bae1fe38e21400c19126ba8ff030d/src/Illuminate/Validation/Validator.php#L1019) with a cleaned attribute value.

PR #42407, opened 2022-05-16
[9.x] Enhance docblocks of Validation Factory

imanghafoori1:
This mentions the strict types in arrays.


imanghafoori1:
@taylorotwell Yeah, there should be a plan to migrate the docblocks by the core team.
Since as a community member I can not do these neither in too many small PRs nor very big hefty ones.

Thanks.

PR #42411, opened 2022-05-17
[9.x] Added ShouldDispatch

vinnygambiny:
This change adds an optional `shouldDispatch` function to jobs. The function gives the ability to keep the condition for dispatching or not a job inside of the job itself. With this, you can dispatch a job and if the shouldDispatch is false, the job will not be dispatched to queue and reduce processing.

### DIFF
--- a/src/Illuminate/Bus/Dispatcher.php
+++ b/src/Illuminate/Bus/Dispatcher.php
@@ -107,6 +107,10 @@ class Dispatcher implements QueueingDispatcher
      */
     public function dispatchNow($command, $handler = null)
     {
+        if (method_exists($command, 'shouldDispatch') && ! $command->shouldDispatch()) {
+            return;
+        }
+
         $uses = class_uses_recursive($command);
 
         if (in_array(InteractsWithQueue::class, $uses) &&
@@ -214,6 +218,10 @@ class Dispatcher implements QueueingDispatcher
      */
     public function dispatchToQueue($command)
     {
+        if (method_exists($command, 'shouldDispatch') && ! $command->shouldDispatch()) {
+            return;
+        }
+
         $connection = $command->connection ?? null;
 
         $queue = call_user_func($this->queueResolver, $connection);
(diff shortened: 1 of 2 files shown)


DarkGhostHunter:
What about forcing the dispatch? If I add the method, I'm forever bound to what the method returns.

I think this could be fixed with a `forceDispatch` on the job itself that changes de job condition to true.

bert-w:
This adds unnecessary complexity imo.

vinnygambiny:
I see it useful for keeping condition if the job needs to be run inside, in one place. These conditions can be like services are enabled, a user has subscribed to. It removes the condition from the handler, so the model isn't fetched to only discard the job at the first line. When you have more than 1k jobs per minute and you can discard job dispatch that will not be executed, that will help.

BrandonSurowiec:
I like the strategy of adding my own static method to my job.

```php
public static function dispatchOnChange($model)
{
    return static::dispatchIf($somecheck, $model);
}
```

bert-w:
What about sending a PR for `dispatchIf` and `dispatchUnless` so you can provide a callback as the argument? i.e. 
```php
public static function dispatchIf($condition, ...$arguments)
    {
        return value($condition)
            ? new PendingDispatch(new static(...$arguments))
            : new Fluent;
    }
```

(`dispatchNow()` is deprecated btw)

vinnygambiny:
@bert-w like this, you need to repeat the condition everywhere you dispatch the job. With my solution, the job is responsible for knowing if we need to dispatch or not.

bert-w:
yea thats true. Nothing a simple inheritance cant fix :)

DarkGhostHunter:
```php
trait DispatchesConditionally
{
    protected $dispatchable = true;

    public function shouldDispatch()
    {
        return $this->dispatchable;
    }

    public function forceDispatch()
    {
        $this->dispatchable = true;

        return $this;
    }
}
```

```php
// Before
if ($message->isNotEmpty()) {
    $user->notify(new Notification($message));
}

// After
$user->notify(new Notification($message));

// Forcefully

$user->notify((new Notification($message))->forceDispatch());
```

Anyway, I don't think this is gonna pass as it's somewhat difficult tomake a "everyone should use this" case. I've been around quite a number of projects and haven't seen the need to.

PR #42412, opened 2022-05-17
[9.x] Add convenience method to set foreignIdFor to be constrained and inde…

jarrodtomholt:
This PR adds the ability to index and constrain by default when using the `foreignIdFor` method on `Illuminate\Database\Schema\Blueprint`.

```php
/** AppServiceProvider::boot() **/

Blueprint::constrainForeignFor();
```

This is a convenience method only and is opt-in by design, removing the necessity for a developer to chain `index()` and `constrained()` methods within migrations when using `foreignIdFor()`


taylorotwell:

PR #42414, opened 2022-05-17
[9.x] Add dynamic `trashed` factory state

jasonmccreary:
This adds dynamic support of a `trashed` factory state for models using the `SoftDeletes` trait. It has an optional argument of the date value. If not passed, the default is yesterday.

**Before**
```php
// in Factory class
public function trashed()
{
    return $this->state([
        'deleted_at' => now()->subDay(),
    ]);
}
```

If this is merged, you would no longer need to define such a state in factories for `SoftDeletes` models.

This was implemented within `__call`. As such, this should be fully backward compatible if a `trashed` method already exists. A test case also demonstrates this.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/Factory.php
+++ b/src/Illuminate/Database/Eloquent/Factories/Factory.php
@@ -9,6 +9,7 @@ use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\SoftDeletes;
+use Illuminate\Support\Carbon;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Conditionable;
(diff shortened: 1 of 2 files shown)


jasonmccreary:
Not sure about the failing tests. All the database test cases passed for me locally (PHP 8.1).

jasonmccreary:
@lukeraymonddowning, the only reason for a "past" date would be to avoid any foot-gun date comparisons in tests since they run so quickly. For example:

```php
$model = Model::factory()->trashed()->create();
$this->assertTrue($model->deleted_at->isPast())
```

Open to changing it. But that was my reasoning.

lukeraymonddowning:
@jasonmccreary good point 👍

jasonmccreary:
@taylorotwell, unfortunately, I don't get these failures locally on PHP 8 or PHP 8.1.

driesvints:
@jasonmccreary retriggered the tests for ya. Might be a hiccup.

jasonmccreary:
@driesvints, actually, seems like I do get the failure when I run the entire test suite. Strange, as when it's run independently it passes. Maybe some kind of test pollution. I'll take a closer look.

jasonmccreary:
@DarkGhostHunter, I'm open to changing it. Keep in mind, this does support passing in the `$deleted_at` timestamp.

jasonmccreary:
@taylorotwell + @driesvints, it was test pollution. `Carbon::setTestNow()` needed to be reset. I assumed a `tearDown` method did this somewhere.

Everything is passing now, so I'm marking this for review. Let me know what you think about the default timestamp (now versus yesterday). Glad to make any final changes.

DarkGhostHunter:
Does it handles intervals? For example, I need to compute messages that have been "online" for more than a minute. 

This way I can warn the user that deleting it may not remove it from the internerks.

PR #42418, opened 2022-05-17
[9.x] Added magic method for WhereRelation

mehdirajabi59:
This PR allows you to where on relation  in that way out of convenience, if a name of method start with `whereRelation` and after it mentions relation name then will proxy to `whereRelation` mehtod.

`User::whereRelationPosts('viewed', '>', 9)->get()` 
`User::whereRelationPosts('publushed',  true)->get()`


ankurk91:
No more magic methods to eloquent please

PR #42420, opened 2022-05-18
[8.x] Use duplicate instead of createFromBase to clone request when routes are cached

### LINKED ISSUE #42403: [9.x] memory exhausted because JSON payload is unwanted parsed twice by Illuminate\Http\Request::createFromBase
- Laravel Version: 9.12.2
- PHP Version: PHP 8.1.4 (cli) (built: Mar 16 2022 09:33:31) (ZTS Visual C++ 2019 x64)
- Database Driver & Version: MySql 8.0.16

### Description:

I have an api POST route with json contents. Very unexpected this route was giving  Allowed memory size of 134217728 bytes exhausted (tried to allocate ... bytes) errors.

To debug this issue:

1) I adjusted public/index.php so the request would run to the end and would not be aborted by memory exhausted errors.

```php
ini_set("memory_limit", "-1");
```

2) I attached XDEBUG in profile mode.

3) I replayed the request.

4) I used QCacheGrind to visualize the profile:

image

As the profile shows, ```Illuminate\Http\Request::createFromBase``` via ```Request->json()``` calls ```json_decode()``` twice. Which is unwanted, because I don't use the ```Request::json()``` to retrieve the json contents. I do my own ```json_decode((string) Request::getContents())```.

I did some digging in the Laravel History and found the commit introducing the json_decode behaviour in [9.x]: https://github.com/laravel/framework/commit/13e4a7ff1799ee51698faa10161b2d2af889b713 The commit message states: _... This commit solves the underlying issue without breaking compatibility with the original functionality. ..._

**This commit is breaking compatibility, because it does not take memory consumption into account. Neither does it take performance into account, decoding the same request body twice is performance degradation.** For large JSON (in my case Content-Length: 6428178) this can easily lead to memory exhausted errors. In Laravel 7.x there was no memory exhausted error.

rodrigopedra:
Closes #42403 

When routes are compiled, the `CompiledRouteCollection@requestWithoutTrailingSlash` method uses the static method `Request@createFromBase` to clone a `Illuminate\Http\Request`.

When the request is a JSON request, this call causes the request's JSON payload to be parsed twice:

- first when the request is captured
- and then when `CompiledRouteCollection@requestWithoutTrailingSlash` is called, which, as noticed on issue #42403 , can cause a noticeable increase on memory usage when processing large JSON payloads.

Actually, issue #42403 describes a memory exhaust scenario with default PHP configuration.

The doc block on `CompiledRouteCollection@requestWithoutTrailingSlash` states  this method `Get a cloned instance of the given request without any trailing slash on the URI`, so as the original `$request` is already a `Illuminate\Http\Request`, e.g. it is already bootstrapped, we can safely change this call to use the `Request@duplicate` method, which internally actually clones the request object, avoiding the double `json_encode` parsing.

This PR:

- Changes the method `CompiledRouteCollection@requestWithoutTrailingSlash` to call `Request@duplicate` instead of `Request@captureFromBase`, to clone the current request.

Notes:

This behavior only happens when routes are cached.

I didn't add any tests, as I would need to mock the `Illuminate\Http\Request` and assert its static method `createFromBase` was not called twice.

I don't know how to properly mock static methods. If someone can guide me on how to do it, I will appreciate.

### DIFF
--- a/src/Illuminate/Routing/CompiledRouteCollection.php
+++ b/src/Illuminate/Routing/CompiledRouteCollection.php
@@ -152,7 +152,7 @@ class CompiledRouteCollection extends AbstractRouteCollection
      */
     protected function requestWithoutTrailingSlash(Request $request)
     {
-        $trimmedRequest = Request::createFromBase($request);
+        $trimmedRequest = $request->duplicate();
 
         $parts = explode('?', $request->server->get('REQUEST_URI'), 2);
 


driesvints:
Thanks @rodrigopedra

PR #42428, opened 2022-05-18
[9.x] Assert on eloquent collections

jasonmccreary:
This solves a [long standing paper-cut of mine](https://github.com/laravel/ideas/issues/1914) when testing. It's best illustrated with a code sample.

The following assertion fails:

```php
public function testExample()
{
    $posts = Post::factory()->times(3)->create();

    $response = $this->get(route('posts.index'));

    $response->assertViewHas('posts', $posts);
}
```

This fails even if `$posts` is _refreshed_ or pulled directly from the database (i.e. `Post::all()`). While the latter comparison should possibly work, the above feels like a common enough testing path to add an explicit path for testing an `EloquentCollection` to improve the out-of-the-box developer experience.

**Note:** The current implementation assumes if you're pass an `EloquentCollection` to the assertion, then the view data is also an `EloquentCollection`. If this is an unsafe, a type check on the view data could also be added.


jasonmccreary:
One more _note_, this implementation performs an equality comparison between the array keys. This factors in things like the order of the collection items. As opposed to a loose `diff` comparison.

I feel this aligns with a strict testing philosophy, which should be the default (imo). If needed, the tester could manipulate the collection beforehand or continue to use a closure.

PR #42429, opened 2022-05-18
[9.x] Adds ability to have `paginate()` `$perPage` parameter as callable with access to `$total`

Niush:
Feature adds ability to have `$perPage` as a callback closure with access to `$total`.

```php
// Previously:
$total = DB::table('products')->count();
DB::table('products')->paginate( $total <= 110 ? 110 : 100 );

// Now (saves extra query and builder code):
DB::table('products')->paginate(function ($total) {
    return $total <= 110 ? 110 : 100;
});
```

There might be a scenario when you want to use different/dynamic per page depending on the total count of records.

Example, in a list of products, with above snippet, it will load all ~110 products if there is less then or equal to 110 products in total. Else, defaults to 100 as usual. So that, there won't be just a few items sitting around in second page.

Similarly, if we want to have exact 4 pages, we can provide dynamic per page value like such:

```php
Product::oldest()->paginate(function ($total) {
    return ceil($total / 4); // Will always have at most 4 pages in total
});
```

And, of course the normal `paginate(10)` will still work.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -864,7 +864,7 @@ class Builder implements BuilderContract
     /**
      * Paginate the given query.
      *
-     * @param  int|null  $perPage
+     * @param  int|null|callable  $perPage
      * @param  array  $columns
      * @param  string  $pageName
      * @param  int|null  $page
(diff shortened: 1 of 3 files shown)


johanrosenson:
Nitpicking, but your last example need to use `ceil` instead of `round` to get at most 4 pages, otherwise something like `total = 101` will be `5 pages`.

PR #42430, opened 2022-05-18
Fixed unavailability to drop primary key on PostgreSQL database

kmarkovych:


driesvints:
There's no explanation here. Please resubmit this with a thorough explanation why this is needed.

PR #42432, opened 2022-05-18
[9.x] Use `where` as `whereIn` if operator is an array

dunhamjared:
This update allows you to use the `where` clause as a `whereIn` if the where `$operator` is an array or instance of `Arrayable`. 

So instead of writing code like this:

```
$query = DB::table('users');

if(is_array($emails)){
    $query->whereIn('email', $emails);
} else {
    $query->where('email', $emails);
}
```

You could write it like this:

```
DB::table('users')->where('email', $emails);
```

This also works with `whereNot`:

```
DB::table('users')->whereNot('name', ['foo', 'bar']);
```

This is helps reduce the amount of code, especially if you write a lot of functions that accept either a single or multiple values.

----

_Very small side note:_

If you pass an array with one value as the operator for `where`:

```
DB::table('users')->where('email', ['foo']);
```

Query builder will use `where in` instead of `where =`. 
 
I confirmed this should not impact performance for MySQL, Postgresql, and SQL Server. 
Based on my research, they should convert single value `where in` clauses into a normal `where =`.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -712,7 +712,7 @@ class Builder implements BuilderContract
         // If operator is an array and has multiple values, we will assume the developer
         // wants a “where in” clause instead. We also assume the developer wants a
         // "where not in" if they use the "whereNot" function.
-        if(is_array($operator) && count($operator) > 1) {
+        if(is_array($operator) || $operator instanceof Arrayable) {
             return $this->whereIn($column, $operator, str_ireplace(' not', '', $boolean), str_contains($boolean, 'not'));
         }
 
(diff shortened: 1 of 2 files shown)


johanrosenson:
```
$query->whereIn('email', (array) $emails);
```

You could also simplify your code and remove your `if-else?` completely by casting your `$emails` to an array, if it's already an array, nothing will change, if it's a string it will be wrapped in an array.

fgaroby:
> You could also simplify your code and remove your `if/else` completely by casting your `$emails` to an array and always use `whereIn`, if `$emails` already is an array, nothing will change, if it's a string it will be wrapped in an array.
> 
> ```
> $query->whereIn('email', (array) $emails);
> ```

Or maybe using `Arr::wrap($emails)`?

taylorotwell:
I would just call `whereIn` explicitly.

dunhamjared:
Could I also get your input on my recent comment @driesvints, @staudenmeir, and @GrahamCampbell?

driesvints:
Seems like Taylor doesn't wants to merge this in, sorry.

PR #42441, opened 2022-05-19
fix: Incorrect mask position

lscar:
- Fix mask error when $index<-strlen($string)，for example "Str::mask('ssss','*', -6, 1);" the current output is "*sss"
- Use the original code style and fix the "offset" value of mb_strpos()

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -495,18 +495,15 @@ class Str
             return $string;
         }
 
-        $strlen = mb_strlen($string, $encoding);
-        $startIndex = $index;
-
-        if ($index < 0) {
-            $startIndex = $index < -$strlen ? 0 : $strlen + $index;
+        $posIndex = $index;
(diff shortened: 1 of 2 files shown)


taylorotwell:
I dunno - I'm kind of done messing with this method. I don't think your test case is intuitive to me.

lscar:
A month ago, I discovered the problem that if there are repeated substrings in the masked string, the mask position is wrong[#1235](https://github.com/laravel/lumen-framework/issues/1235)。

However "driesvints" said it wasn't a bug, so I didn't push my code. Didn't realize I missed a code merge opportunity until I saw this[#42295](https://github.com/laravel/framework/pull/42295)

The test cases I used might not be very realistic. My thinking is also wrong. I think subscripting of negative numbers is circular. Negative subscripts that exceed the length of the string itself start again from the end of the string. This logic may not conform to mainstream cognition.

After this incident it did make me a little frustrated, but thanks for your reply anyway. @taylorotwell

PR #42444, opened 2022-05-19
[9.x] Extract ServeCommand env list to static property

rasmuscnielsen:
This PR extracts the hardcoded list of allowed ENV variables from the ServeCommand to a public static property. 

**Background**

Currently the Artisan serve command filters out any host ENV variables that is not explicitly listed in a hardcoded array.

```
return in_array($key, [
    'APP_ENV',
    'LARAVEL_SAIL',
    'PHP_CLI_SERVER_WORKERS',
    'PHP_IDE_CONFIG',
    'SYSTEMROOT',
    'XDEBUG_CONFIG',
    'XDEBUG_MODE',
    'XDEBUG_SESSION',
]) ? [$key => $value] : [$key => false];
```

This makes it hard as a developer to add pass additional ENV to the server process - for instance from docker-compose.yml using Sail. The additional ENV can be accessed just fine from CLI but will never be accessible from the browser because it is filtered out from the serve command.

**Proposal**

Make it easier to allow extra ENV to be passed through to the php built-in server by extracting the above list to a public static property so it can be changed from a ServiceProvider@register/boot function without having to extend and customise the serve command.

Additionally it could be handy to automatically pass through any ENV variable starting with a prefix such as `PUBLIC_` or `LARAVEL_`. 
This is similar to what VUE CLI does by making any ENV variables accessible from client javascript if they start with `VUE_APP_`. This would give the best developer experience as no customisations are needed by the developer.
If there is support for this I'm happy to implement it too.

### DIFF
--- a/src/Illuminate/Foundation/Console/ServeCommand.php
+++ b/src/Illuminate/Foundation/Console/ServeCommand.php
@@ -45,11 +45,11 @@ class ServeCommand extends Command
     protected $portOffset = 0;
 
     /**
-     * The env variables that should be passed from host machine to the php server process.
+     * The environment variables that should be passed from host machine to the PHP server process.
      *
      * @var string[]
      */
-    public static $passthruEnv = [
+    public static $passthroughVariables = [
         'APP_ENV',
         'LARAVEL_SAIL',
         'PHP_CLI_SERVER_WORKERS',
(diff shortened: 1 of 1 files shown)


zolotov88:
Laravel Octane do the same things. It is a reason why do not use it)

PR #42449, opened 2022-05-19
[9.x] Adds expected behavior to `where` function in query builder

dunhamjared:
_Coming at this with a different angle -- original: https://github.com/laravel/framework/pull/42432_

If you pass an array of values as the operator, I expect it to return rows that include those values. 

For example:

```
DB::table('users')->where('name', ['foo', 'bar']);
```

Would return users that have the name `foo` and `bar`. 

However, it only returns users that have the name `foo`.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -712,7 +712,7 @@ class Builder implements BuilderContract
         // If operator is an array and has multiple values, we will assume the developer
         // wants a “where in” clause instead. We also assume the developer wants a
         // "where not in" if they use the "whereNot" function.
-        if(is_array($operator) && count($operator) > 1) {
+        if(is_array($operator) || $operator instanceof Arrayable) {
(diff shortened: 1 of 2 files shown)


lk77:
Well it's the expected behaviour, that's how the sql where clause works

lk77:
Well if you want you can use the equal sign :

```
DB::table('Customer')->where('City',  '=', 'Paris');
```

dunhamjared:
@lk77 Yes, I understand that you can pass a comparison operator under the operator argument.

lk77:
I don't like the idea of making too much assumptions on what the dev wants or does not wants,

like for example  :

```
DB::table('Customer')->where('City', $city);
```

is it a where or a where in ?

taylorotwell:
Just call `whereIn`.

lk77:
I understand your point of view, but your not gaining that much, you correct a potential mistake from new devs, and make the syntax a little bit shorter

dunhamjared:
@lk77 A bunch of little gains add up. 

If we always live by "but your not gaining that much" to dismiss a change, then Laravel, and many other things in life, wouldn't be where it is today.

PR #42455, opened 2022-05-19
[9.x] Add IgnoringCase variants for assertSee functions

supun-io:
The following variants are added to `TestResponse` class.

* `assertSee` -> `assertSeeIgnoringCase`
* `assertSeeInOrder` -> `assertSeeInOrderIgnoringCase`
* `assertSeeText` -> `assertSeeTextIgnoringCase`
* `assertSeeTextInOrder` -> `assertSeeTextInOrderIgnoringCase`
* `assertDontSee` -> `assertDontSeeIgnoringCase`
* `assertDontSeeText` -> `assertDontSeeTextIgnoringCase`

Under the hood, PHPUnit's [assertStringContainsStringIgnoringCase](https://phpunit.readthedocs.io/en/9.5/assertions.html#assertstringcontainsstringignoringcase) function and [mb_stripos](https://www.php.net/manual/en/function.mb-stripos.php) are used.

### Why?

I first noticed this is a handy feature to have while testing error messages in HTTP responses. For example, if the error message is "Subdomain is already taken", I have to write one of these:

```php
$response->assertSee('Subdomain is already taken');
$response->assertSee(['Subdomain', 'taken']);
```

While I prefer the second one, it is still a hassle to write it in the same case as the response. If I ever had to change the error message to "The subdomain is already taken", I would have to change the test too because of the case sensitivity. So, the following way would be better.

```php
$response->assertSeeIgnoringCase(['subdomain', 'taken']);
```

## Decisions Made

* **Dropped the $escape param** - I have not used the `$escape` param in the new functions. The reason is that HTML escaped entities are case-sensitive. There is no point in doing a case-insensitive if you want to match escaped HTML entities. Therefore, the newly added functions do not escape the given strings.
* The `IgnoringCase` suffix was inspired by PHPUnit.

### Other discussions

* A feature request in Dusk: https://github.com/laravel/dusk/issues/217
* A discussion on forums: https://laracasts.com/discuss/channels/testing/assertsee-assertseein-case-sensitive

Finally, if this idea is to be accepted, I can add similar variants to other classes that have `->assertSee` functions (Mailable, View, etc).

### DIFF
--- a/src/Illuminate/Testing/Constraints/SeeInOrder.php
+++ b/src/Illuminate/Testing/Constraints/SeeInOrder.php
@@ -15,6 +15,13 @@ class SeeInOrder extends Constraint
     protected $content;
 
     /**
+     * Whether the comparison is case-sensitive
+     *
+     * @var bool
+     */
+    protected $caseSensitive;
+
+    /**
      * The last value that failed to pass validation.
      *
      * @var string
@@ -27,9 +34,10 @@ class SeeInOrder extends Constraint
      * @param  string  $content
      * @return void
      */
-    public function __construct($content)
+    public function __construct($content, $caseSensitive = true)
     {
         $this->content = $content;
+        $this->caseSensitive = $caseSensitive;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #42457, opened 2022-05-20
[9.x]  Add method to make instance of Illuminate\Http\Request for testing

mehdirajabi59:
Sometimes maybe you should pass instance of request to test any class such as middleware. This PR adds a method for make instance of  `Illuminate\Http\Request` to inject any class.

**Before This PR** : 

```
$middleware = app(UserBannedMiddleware::class);

$request = Request::createFromBase(SymfonyRequest::create(
            '/panel',
            'GET',
        ));

$response = $middleware->handle(
            $request,
            fn () => new Response()
);
```

**After this PR**

```
$middleware = app(UserBannedMiddleware::class);

  $response = $middleware->handle(
            $this->createRequest('/panel', 'GET'),
            fn () => new Response()
);
```

**Method signature :**

```
createRequest(string $uri, string $method = 'GET', array $parameters = [], array $cookies = [], array $files = [], array $server = [], $content = null): Request
```

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
@@ -666,4 +666,31 @@ trait MakesHttpRequests
             );
         });
     }
+
+    /**
+     * Create the request instance.
+     *
+     * @param  string  $uri  The URI
+     * @param  string  $method  The HTTP method
+     * @param  array  $parameters  The query (GET) or request (POST) parameters
+     * @param  array  $cookies  The request cookies ($_COOKIE)
+     * @param  array  $files  The request files ($_FILES)
+     * @param  array  $server  The server parameters ($_SERVER)
+     * @param  string|resource|null  $content  The raw body data
+     * @return Illuminate\Http\Request
+     */
+    protected function createRequest(string $uri, string $method = 'GET', array $parameters = [], array $cookies = [], array $files = [], array $server = [], $content = null): Request
+    {
+        $symfonyRequest = SymfonyRequest::create(
+            $uri,
+            $method,
+            $parameters,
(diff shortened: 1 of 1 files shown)


crynobone:
Why `Request::createFromBase()` and not just `Request::create()`?

johanrosenson:
Just an idea:

Instead of adding this to `MakesHttpRequests.php`, it could be added to `Illuminate\Http\Request` instead and override the `Symfony Request create` method.

This would make it a breaking change and force the PR to be targeting 10.x instead, but I feel it's a longterm better solution.

Then you could always easily create a custom request by doing `Request::create()`, you can already do this today, but it will give you a base request (Symfony Request)

But with this change in 10.x it would give you a Laravel Request instead.

It would replace this code you have to write today to achieve the same result:

```
Request::createFromBase(
    SymfonyRequest::create(...),
);
```

johanrosenson:
> Why `Request::createFromBase()` and not just `Request::create()`?
> 
> Following should already working:
> 
> ```
> $middleware = app(UserBannedMiddleware::class);
> 
> $response = $middleware->handle(
>     Request::create('/panel', 'GET'),
>     fn () => new Response()
> );
> ```

I guess because that would create a Symfony Request and not a Laravel request.

crynobone:
> I guess because that would create a Symfony Request and not a Laravel request.

Have you tried it?


image

johanrosenson:
> > I guess because that would create a Symfony Request and not a Laravel request.
> 
> Have you tried it?
> 
> image

wow, mind blown! Thanks!

I just checked the code in Symfony\Request now and it uses `return new static` to create the new Request, this invalidates my whole comment above and i can stop doing this now:

```
Request::createFromBase(
    SymfonyRequest::create(...),
);
```

^^ not needed :)

PR #42460, opened 2022-05-20
Enforce complete Deferrable ServiceProviders

PrinsFrank:
In the DeferrableProvider interface, the 'provides' method is present. When implementing this interface, [the 'provides' method should be implemented](https://laravel.com/docs/9.x/providers#deferred-providers), otherwise the services in this provider are never available and are technically broken. 

The 'provides' method is also always defined on the abstract ServiceProvider, returning an empty array there. So when a serviceprovider implements the deferabbleProvider interface the presence of a 'provides' method is no longer enforced as it is already present in the parent class.

This PR moves the 'provides' method out of the abstract serviceprovider so Deferrable providers that don't implement the 'provides' method themselves are not silently broken anymore. 

As any existing serviceProvider that is deferrable without a 'provides' method now breaks (And is currently also broken but silently), this is a breaking change, hence the PR to master instead of 9.x.

### DIFF
--- a/src/Illuminate/Support/AggregateServiceProvider.php
+++ b/src/Illuminate/Support/AggregateServiceProvider.php
@@ -43,8 +43,9 @@ class AggregateServiceProvider extends ServiceProvider
 
         foreach ($this->providers as $provider) {
             $instance = $this->app->resolveProvider($provider);
-
-            $provides = array_merge($provides, $instance->provides());
+            if ($instance->isDeferred()) {
+                $provides = array_merge($provides, $instance->provides());
+            }
         }
 
         return $provides;
--- a/src/Illuminate/Support/ServiceProvider.php
+++ b/src/Illuminate/Support/ServiceProvider.php
@@ -388,16 +388,6 @@ abstract class ServiceProvider
     }
 
     /**
-     * Get the services provided by the provider.
-     *
-     * @return array
-     */
-    public function provides()
-    {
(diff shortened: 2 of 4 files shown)


ankurk91:
We should force developer to define `provides` method via `Illuminate\Contracts\Support\DeferrableProvider` interface

taylorotwell:
I don't see a big end-user benefit to this breaking change.

PrinsFrank:
@taylorotwell Maybe my garble of words was too complex, let my try again: 

We have an interface "DeferrableProvider". Object interfaces allow you to create code which specifies which methods a class must implement, without having to define how these methods are implemented. But at the same time, the 'provides' methods is implemented in the abstract ServiceProvider. A method that is only used for deferrableProviders, that's why it is in the interface. But the way it is implemented there is always wrong, so you HAVE to overwrite it, otherwise the deferrable provider doesn't work.

Instead of trusting that end-users read the documentation, we can enforce that the method should be implemented by the end-user.

I can speak from personal experience that I have implemented the interface multiple times, forgetting to overwrite the incomplete method as my IDE doesn't warn me about missing methods when they exist. And subsequently wondering why my service provider doesn't work. I can't be the only one frustrated by this.

PrinsFrank:
@taylorotwell I expect an error when I work with deferrableProviders and the provides method is missing;
image

But currently that is not the case:
image

PR #42461, opened 2022-05-20
[9.x] Use route parameters in view

yoeriboven:
```php
Route::view('route/{parameter}', 'view');
```

I expected `$parameter` to be available in the view.
There is also [a test](https://github.com/laravel/framework/blob/c73253e51a3365a93eed0fb745f51771b89aa53c/tests/Integration/Routing/RouteViewTest.php#L21) which makes it seem like it is the expected behavior.

This will actually throw an exception:

```
Undefined variable $parameter
```

This PR fixes this.

It takes all function parameters, removes the [default values](https://github.com/laravel/framework/blob/c73253e51a3365a93eed0fb745f51771b89aa53c/src/Illuminate/Routing/Router.php#L277-L286) and adds them to the view's data.

### DIFF
--- a/src/Illuminate/Routing/ViewController.php
+++ b/src/Illuminate/Routing/ViewController.php
@@ -32,13 +32,18 @@ class ViewController extends Controller
      */
     public function __invoke(...$args)
     {
-        $parameters = array_filter($args, function ($key) {
+        $routeParameters = array_filter($args, function ($key) {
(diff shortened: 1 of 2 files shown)


jasonmccreary:
@taylorotwell, since this has been merged, what do we think about revisiting #40192 (this, but for implicitly bound models)?

PR #42463, opened 2022-05-20
[9.x] Rename $sleep argument to $sleepMilliseconds in retry method on PendingRequest

denniseilander:
When using the `retry()` method on the `Http` facade, you have to set the `$sleep` as second argument.
Now it is unclear whether the $sleep is in seconds or in milliseconds, although it makes sense that it is in milliseconds.

To clear up this confusion and maintain consistency with the `retry()` helper function (`Support/helpers.php`), I'v opened this PR which renames the `$sleep` argument variable into `$sleepMilliseconds`.

Before:
Screenshot 2022-05-20 at 15 30 57

After:
Screenshot 2022-05-20 at 15 29 16


marvin-wtt:
I'm not 100% sure but doesn't this break calls with named arguments?

driesvints:
@marvin-wtt named arguments don't fall under our BC policy: https://laravel.com/docs/9.x/releases#named-arguments

PR #42464, opened 2022-05-20
Attribute cast to array of enum

Cellard:
Count this PR as just an idea.

As Laravel supports Enum casting, it seems reasonable to add support for arrays with Enums.

This PR brings support of array with enum (e.g. `set` database column type) to Eloquent.

This PR implements syntax for `set` database column type:

```php
protected $casts = [
    'roles' => \App\Enums\RoleEnum::class . ':set',
];
```

And for `text` or `json` database column type:

```php
protected $casts = [
    'roles' => \App\Enums\RoleEnum::class . ':json',
];
```


taylorotwell:

PR #38499, opened 2021-08-23
[8.x] Adds a simple `where` helper for querying relations

DarkGhostHunter:
## What?

This is sugar for `whereHas()` for simple `where` clauses.

```php
// Before:
User::whereHas('posts', function ($query) {
    $query->where('published_at', '>', now());
})->get();

// After
User::whereRelation('posts', 'published_at', '>', now())->get();
```

## How?

Basically creates the `whereRelation()` and `orWhereRelation()` helpers, and `whereMorphRelation()` and `orWhereMorphRelation()` for morph relations. Since these use the `where()` method underneath, you can do advanced things:

```php
Comment::whereMorphRelation('commentable', '*', [
    'is_public' => true,
    'is_vip' => false,
])->get();
```

## Why?

Removes code complexity by making the query more readable when dealing with simple `where` clauses over relations.

## Breaking Changes

None - it's merely and additive PR.

PR #42465, opened 2022-05-20
[9.x] Model::whereRelation ignores the callback function or does it wrong

adideas:
Eloquent
```php
$userSQL = User::whereRelation(
    "posts",
    function($query) {
        // THIS
        $query->selectRaw("id")->limit(1)->where("public", true);
    }
)->toSql();
```
Query
```php
$dbSQL = DB::table("users")->whereExists(
    function ($builder) {
        // relations simulation
        $builder->from("posts")->addSelect(['*'])->whereColumn(
            'users.id', '=', 'posts.user_id'
        );
        // THIS
        $builder->selectRaw("id")->limit(1)->where("public", true);
    }
)->toSql();
```

Eloquent output
```sql
select * 
from "users" 
where exists (
   select * 
   from "posts" 
   where "users"."id" = "posts"."user_id" and ("public" = ?)
)
```
Query output
```sql
select * 
from "users" 
where exists (
   select *, id 
   from "posts" 
   where "users"."id" = "posts"."user_id" and "public" = ? limit 1
)
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -365,7 +365,11 @@ trait QueriesRelationships
     public function whereRelation($relation, $column, $operator = null, $value = null)
     {
         return $this->whereHas($relation, function ($query) use ($column, $operator, $value) {
-            $query->where($column, $operator, $value);
+            if ($column instanceof Closure) {
+                $column($query);
+            } else {
+                $query->where($column, $operator, $value);
+            }
         });
     }
 
@@ -381,7 +385,11 @@ trait QueriesRelationships
     public function orWhereRelation($relation, $column, $operator = null, $value = null)
     {
         return $this->orWhereHas($relation, function ($query) use ($column, $operator, $value) {
-            $query->where($column, $operator, $value);
+            if ($column instanceof Closure) {
+                $column($query);
+            } else {
+                $query->where($column, $operator, $value);
+            }
         });
     }
 
--- a/tests/Integration/Database/EloquentWhereHasTest.php
+++ b/tests/Integration/Database/EloquentWhereHasTest.php
@@ -2,13 +2,13 @@
 
 namespace Illuminate\Tests\Integration\Database\EloquentWhereHasTest;
 
+use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
 use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Query\Builder as QueryBuilder;
 use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Facades\Schema;
 use Illuminate\Tests\Integration\Database\DatabaseTestCase;
-use Illuminate\Support\Facades\DB;
-use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
-use Illuminate\Database\Query\Builder as QueryBuilder;
 
 class EloquentWhereHasTest extends DatabaseTestCase
 {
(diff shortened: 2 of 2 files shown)


adideas:
The reason is not known to me. But the error is generated only from the model.
Once you study the question, everything becomes easy.
The problem is in this piece of code.
[QueriesRelationships::whereRelation](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php#L365)
```php
    public function whereRelation($relation, $column, $operator = null, $value = null)
    {
        return $this->whereHas($relation, function ($query) use ($column, $operator, $value) {
            $query->where($column, $operator, $value);
        });
    }
```
[QueriesRelationships::orWhereRelation](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php#L381)
```php
    public function orWhereRelation($relation, $column, $operator = null, $value = null)
    {
        return $this->orWhereHas($relation, function ($query) use ($column, $operator, $value) {
            $query->where($column, $operator, $value);
        });
    }
```
I realized that it works like levels. But at the last level, it cannot pass parameters such as "Limit" "Select" and so on. since it is only responsible for "where". After that, all parameters except "where" are excluded and we get an error.

adideas:
I agree I could have used a different method. For example "whereHas" instead of "whereRelation".

adideas:
As I understand it, the last one who changed this code was [Italo DarkGhostHunter](https://github.com/DarkGhostHunter).
Pull Request [[8.x] Adds a simple where helper for querying relations #38499](https://github.com/laravel/framework/pull/38499)

DarkGhostHunter:
(checking...)

DarkGhostHunter:
Well, it's expected:

> Add a basic where clause to a relationship query.

I don't remember why I didn't do that in the first place, I think it was because the callback was already handled by Eloquent `where()` and it doesn't replaces `whereHas()`, which is a convenience method for `has()` with a callback where you have more control.

adideas:
> Well, it's expected:
> 
> > Add a basic where clause to a relationship query.
> 
> I don't remember why I didn't do that in the first place, I think it was because the callback was already handled by Eloquent `where()` and it doesn't replaces `whereHas()`, which is a convenience method for `has()` with a callback where you have more control.

I'm sorry I commented on you. I thought you should see.

adideas:
@GrahamCampbell "Changed name" Thanks for correcting me!

taylorotwell:
We don't document using these methods this way.

adideas:
@taylorotwell I did an analogy.
where(callback)
...
has('relation', '>=', 1, "and", callback)
whereHas('relation', callback)
whereRelation() // no callback function
doesntHave('relation', operator, callback)
whereDoesntHave('relation', callback)
whereHasMorph('relation', [], callback)
whereDoesntHaveMorph('relation', [], callback)

I agree it's not in the documentation.
But as I understood by researching other methods, it is always possible to specify a callback function.
Also, not everything is described in the documentation.

Example 1 (method has)
There is no information that the 4th argument is a callback.
Example 2 (method doesntHave)
There is no information that the 3th argument is a callback.

In the last pull request you told me.
"All changes should follow the Laravel approach."
I believe that the current change does not violate the Laravel approach.

The developer who implemented this method said. [DarkGhostHunter: I don't remember why I didn't do that in the first place, I think it was because the callback was already handled by Eloquent where() and it doesn't replaces whereHas(), which is a convenience method for has() with a callback where you have more control.](https://github.com/laravel/framework/pull/42465#issuecomment-1133339833)

I don't know why you think the current addition is wrong.
I followed the Laravel approach.
This is not a critical addition.
You or I can add information to the documentation.

Tell me how it would be better. I'll fix it.

adideas:
@taylorotwell I think we're talking about a supplement. I'm not sure if this is a mistake. But many expect a different result using this method.

adideas:
please show [this](https://github.com/laravel/framework/pull/42491)

PR #42467, opened 2022-05-21
[9.x] Added the "for" method to the Eloquent Builder and Model

ash-jc-allen:
Hey! This PR proposes a new "for" method that can be used on models and in Eloquent queries.

I've marked this as a draft PR for the time being because I'm about half way through doing it. I still need to add the "for" method to the relationships (explained in more detail below). So, I didn't want to spend too much time working on it if you didn't think it would be worth merging in.

## Context

This PR is inspired by the `whereBelongsTo` method that you can use in queries like:

```php
Post::whereBelongsTo($user)->get();
```

I like how it makes the code a lot more human-readable and avoids writing column names (e.g. - `user_id`). So, I thought I'd give it a go and see if we could add something similar for creating/updating models.

I chose the "for" name based on the fact that the functionality is similar to the "for" model factories.

## Before

Excuse the super basic examples, but they should hopefully demonstrate my general point. Before my proposed method, you could have either of these example statements:

```php
Comment::create([
    'user_id' => $user->id,
    'post_id' => $post->id,
    'content' => 'abc',
]);
```

Or...

```php
$post->comments()->create([
    'user_id' => $user->id,
    'content' => 'abc',
]);
```

Or...

```php
$comment->update([
    'user_id' => $anotherUser->id,
    'content' => 'abc',
]);
```

## After

But the problem is that there's always still another ID field in the `create` method if you're trying to make a relationship to more than one model. So, we could rewrite the queries as:

```php
Comment::for($user)
    ->for(post)
    ->create([
        'content' => 'abc',
    ];
```

Or...

```php
$post->comments()
    ->for($user)
    ->create([
        'content' => 'abc',
    ]);
```

```php
$comment->for($anotherUser)->update([
    'content' => 'abc',
]);
```

At the moment, I've only got the `Model::for()` and `$model->for()` approaches working for all the different query methods (as far as I can see, but I might have missed something). I haven't made a start on the relationships (e.g. - `$model->relationship()->for()->create()`) yet just in case this PR wasn't something that would get merged. If you think that this is something that would be valuable to other devs though, I'll carry on and get the relationships working too.

Personally, I could see this being particularly useful feature and think it'd be a nice way of making our Laravel project codebases even cleaner! 😄

Apologies, by the way, if this sort of logic already exists and I've just completely missed it.

Hopefully, it's heading in a good direction, but it's no problem if you'd rather close the PR 😄

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1434,8 +1434,8 @@ class Builder implements BuilderContract
     /**
      * Set the foreign key for a relationship to another model.
      *
-     * @param  Model  $model
-     * @param  string  $relationship
+     * @param  \Illuminate\Database\Eloquent\Model  $model
+     * @param  string|null  $relationship
      * @return $this
      */
     public function for($model, $relationship = null)
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -681,7 +681,7 @@ class Builder implements BuilderContract
      */
     public function get($columns = ['*'])
     {
-        $builder = $this->applyScopes();
+        $builder = $this->where($this->mergeForeignKeys([]))->applyScopes();
 
         // If we actually found models we will also eager load any relationships that
         // have been specified as needing to be eager loaded, which will solve the
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -537,7 +537,7 @@ class Builder implements BuilderContract
      */
     public function firstOrNew(array $attributes = [], array $values = [])
     {
-        if (! is_null($instance = $this->where($attributes)->first())) {
+        if (! is_null($instance = $this->where($this->mergeForeignKeys($attributes))->first())) {
             return $instance;
         }
 
@@ -553,7 +553,7 @@ class Builder implements BuilderContract
      */
     public function firstOrCreate(array $attributes = [], array $values = [])
     {
-        if (! is_null($instance = $this->where($attributes)->first())) {
+        if (! is_null($instance = $this->where($this->mergeForeignKeys($attributes))->first())) {
             return $instance;
         }
 
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -681,7 +681,7 @@ class Builder implements BuilderContract
      */
     public function get($columns = ['*'])
     {
-        $builder = $this->where($this->mergeForeignKeys([]))->applyScopes();
+        $builder = $this->applyScopes();
 
         // If we actually found models we will also eager load any relationships that
         // have been specified as needing to be eager loaded, which will solve the
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1155,10 +1155,8 @@ class Builder implements BuilderContract
      */
     protected function mergeForeignKeys($attributes)
     {
-        if (count($this->for)) {
-            foreach ($this->for as $model) {
-                $attributes[$model['foreignKey']] = $model['model']->getKey();
-            }
+        foreach ($this->for as $model) {
+            $attributes[$model['foreignKey']] = $model['model']->getKey();
         }
 
         return $attributes;
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -128,6 +128,13 @@ class Builder implements BuilderContract
     protected $removedScopes = [];
 
     /**
+     * The models that the queried models will be related to via a foreign key.
+     *
+     * @var array
+     */
+    protected $for = [];
+
+    /**
      * Create a new Eloquent query builder instance.
      *
      * @param  \Illuminate\Database\Query\Builder  $query
@@ -146,7 +153,7 @@ class Builder implements BuilderContract
      */
     public function make(array $attributes = [])
     {
-        return $this->newModelInstance($attributes);
+        return $this->newModelInstance($this->mergeForeignKeys($attributes));
     }
 
     /**
@@ -534,7 +541,7 @@ class Builder implements BuilderContract
             return $instance;
         }
 
-        return $this->newModelInstance(array_merge($attributes, $values));
+        return $this->newModelInstance($this->mergeForeignKeys(array_merge($attributes, $values)));
     }
 
     /**
@@ -550,7 +557,7 @@ class Builder implements BuilderContract
             return $instance;
         }
 
-        return tap($this->newModelInstance(array_merge($attributes, $values)), function ($instance) {
+        return tap($this->newModelInstance($this->mergeForeignKeys(array_merge($attributes, $values))), function ($instance) {
             $instance->save();
         });
     }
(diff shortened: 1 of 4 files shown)


michaelnabil230:
Nice job 👍🏻

driesvints:
@ash-jc-allen remember that draft PRs aren't reviewed

ash-jc-allen:
Hey, @driesvints! Sorry, I completely forgot about that.

I know that you won't review the PR whilst it's a draft, but from looking at the code examples in the PR description, would you say it's worth me carrying on with it? I wouldn't want to spend too much time on it if you don't think it's something that could get merged :)

driesvints:
Hey @ash-jc-allen, I don't review PR's, only Taylor does. So please mark this as ready if you want a review from him.

ash-jc-allen:
I think that this is probably ready now for some kind of review. Apologies that this one has sat as a draft for so long! I've added some basic tests for each of the methods that I've updated too.

I fully expect that I've probably missed something here or overlooked something. But, hopefully, it should be headed in the right direction and might be something that you think would be helpful for other developers.

If it's something that you might consider merging, please let me know if you'd like me to make any changes 😄

src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -1140,6 +1147,23 @@ protected function addUpdatedAtToUpsertColumns(array $update)
...
 
+    /**
+     * Merge the attributes that will be used in the query with the specific foreign keys.
+     *
+     * @param array $attributes
+     * @return array
+     */
+    protected function mergeForeignKeys($attributes)
+    {
+        if (count($this->for)) {
```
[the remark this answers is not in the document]
ash-jc-allen: Thanks for picking up on that for me. It was something I'd put in when I was first tinkering with another idea. But, I've taken that out now :)

deleugpn:
When I saw the tweet about for on Eloquent I expected something to do with loops. Personally I prefer the explicit definition of the columns,  but if this were to be merged I think I would prefer `belongingTo(...)` than `for`

3x1io:
Awesome 👍

dillingham:
> When I saw the tweet about for on Eloquent I expected something to do with loops. Personally I prefer the explicit definition of the columns,  but if this were to be merged I think I would prefer `belongingTo(...)` than `for`

Disagree.. a model is a single entity.. not a collection. Also loop is not assumed for factories.  I've thought "for" so many times. Happy to see a PR for it

Haythamasalama:
I think it's a great idea but I think the name method it's good to be `by()` to avoid misunderstanding

src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
```diff
@@ -295,6 +320,27 @@ public function saveMany($models)
         return $models;
     }
 
+    /**
+     * Set the foreign key for a relationship to another model.
+     *
+     * @param Model $model
+     * @param string $relationship
+     * @return $this
+     */
```
driesvints, line 327: Please make sure your docblocs are in the same style as the rest of the framework (spaces & FQN).

ash-jc-allen: Apologies, I completely forgot to format them before I pushed up. I've just pushed an update to them, so they should hopefully be okay now? :)

lepikhinb:
IMO it would be great if the method also called `whereBelongsTo`, so then we would have a neat syntax for `firstOrCreate` and similar methods.

```php
$post->views()
    ->for($user)
    ->firstOrCreate([]);
```
```php
$video->templates()
    ->for($parentTemplate)
    ->firstOrCreate([], [
        'elements' => $parentTemplate->elements,
    ]);
```

taylorotwell:
I'm just thinking:

```php
Comment::for($user)
    ->for(post)
    ->where('status', 'published')
    ->firstOrCreate([
        'content' => 'abc',
    ]);
```

It feels a bit ambiguous to me what this query would be. It feels like I would expect it to search for comments that belong to a given user and post that also have a given status and content. But, right now, it seems with this PR in its current state it would search for ANY comment with the given status and content - regardless of the user and post it belongs to.

ash-jc-allen:
@taylorotwell Yeah the more that I look at your example, the more I agree with that. Based on that, do you want me to put together a quick change for the `firstOrCreate` method to try this out? :)

lepikhinb:
@taylorotwell that’s what I was thinking as well. It would probably be cool to have a combined method. But it also shouldn’t be limited to relations.

Say, we have a `having` method (it’s already reserved, but I can’t think of a better name right now), which would modify all `create`, `update` and `get` queries.

Then the following query:
```php
Comment::query()
    ->having($user)
    ->having('status', 'published')
    ->first();
```

Would be transformed into:
```php
Comment::query()
    ->where('user_id', $user->id)
    ->where('status', 'published')
    ->first();
```

And the query as:
```php
Comment::query()
    ->having($user)
    ->having('status', 'published')
    ->create(['content' => $content]);
```

Will get transformed into:
```php
Comment::query()
    ->create([
        'user_id' => $user->id,
        'status' => 'published',
        'content' => $content,
    ]);
```

So, this could become a pretty versatile helper for all create / update / firstOr / updateOr / get methods.

ash-jc-allen:
I've updated this now so that if you use `for()` in a query (like in your example), it'll add the "for"s as `where` constraints.

For example:

```php
Comment::for($user)
    ->for(post)
    ->where('status', 'published')
    ->firstOrCreate([
        'content' => 'abc',
    ]);
```

This would attempt to find a comment that belongs to `$user` and belongs to `$post`. If it can't find one, it'll create a post that belongs to `$user` and belongs to `$post`.

I started adding the "for" functionality to the `get()` and `first()` methods, but didn't know if they were needed, so I reverted them back for the time being.

Hopefully, this should be working as expected 🙂

taylorotwell:
@ash-jc-allen can you summarize where you can and can't use `for` based on the current state of this PR? I've gotten a bit lost on what methods support it, what relationships support it, etc.

ash-jc-allen:
Yeah, I've got a bit lost too because I've not looked at it in a few days. As far as I'm aware, it affects the following methods:

### Builder/Model

make
```php
Comment::for($user)->make(...);
```

create
```php
Comment::for($user)->create(...);
```

forceCreate
```php
Comment::for($user)->forceCreate(...);
```

update (the `for` relationship is not applied to the search. It's only used for setting data that will be stored in the DB)
```php
Comment::for($user)->update(...);
```

firstOrNew (the `for` relationship is also applied to the search for the "first")
```php
Comment::for($user)->firstOrNew([], []);
```

firstOrCreate (the `for` relationship is also applied to the search for the "first")
```php
Comment::for($user)->firstOrCreate([], []);
```

updateOrCreate (the `for` relationship is also applied to the search for the "update")
```php
Comment::for($user)->updateOrCreate([], []);
```

### HasOneOrMany

make
```php
$post->comments()->for($user)->make([...]);
```

create
```php
$post->comments()->for($user)->create([...]);
```

forceCreate
```php
$post->comments()->for($user)->forceCreate([...]);
```

update (the `for` relationship is not applied to the search. It's only used for setting data that will be stored in the DB)
```php
$post->comments()->for($user)->update([...]);
```

firstOrNew (the `for` relationship is also applied to the search for the "first")
```php
$post->comments()->for($user)->firstOrNew([...], [...]);
```

firstOrCreate (the `for` relationship is also applied to the search for the "first")
```php
$post->comments()->for($user)->firstOrCreate([...], [...]);
```

updateOrCreate (the `for` relationship is also applied to the search for the "update")
```php
$post->comments()->for($user)->updateOrCreate([...], [...]);
```

taylorotwell:
I kinda think I will hold off on this for now. I worry it will be a bit too implicit and magical as to what is going on and will be hard to remember exactly what is happening in each scenario vs. just actually setting where clauses explicitly in the code.

ash-jc-allen:
@taylorotwell That's a totally fair point, and I agree. It was a worth a shot and I'm glad I got chance to explore the idea and scratch an itch 😄

PR #42473, opened 2022-05-21
[9.x] Added getDirectorySize Method to Filesystem.php

hmennen90:
This Pullrequest adds the getDirectorySize Method to Filesystem.php Facade.

The Size of the given Directory will be returned in Bytes using the allFiles and size Methods.


ankurk91:
Does it works with S3 (Cloud) driver ?

hmennen90:
@ankurk91 don't know yet, but it should: I think the allFiles and size Methods are working on S3.

taylorotwell:

PR #42474, opened 2022-05-22
[9.x] Request retrieve date from JSON format

DarkGhostHunter:
# What?

Retrieves a date from a JSON encoded `Date` Javascript object. Includes constant to easily validate the format.

```php
use Illuminate\Http\Request;
use Illuminate\Support\DateFactory;

// {
//     "kickoff": "2020-01-01T16:00:00.000+05:00"
// }
public function kickoff(Request $request)
{
    $request->validate([
        'kickoff' => 'date_format:' . DateFactory::JSON_DATE
    ]);
    
    $request->jsonDate('kickoff');
    // "2020-01-01 16:00:00.000 (+05:00)"
}
```

Current browsers encode the `Date` object in JSON with milliseconds, but Laravel encodes them with microseconds. This format accepts both sub-second precisions (3 or 6 digits), so the same data can be reused (like for inter-Laravel-apps-communication).

Also, this optionally shift the date to UTC using `jsonDate('kickoff', true)`. This is just convenience as the format already includes the timezone, making it ready to be persisted in the database.

```php
$request->jsonDate('kickoff', true);
// "2020-01-01 11:00:00.000 UTC (+00:00)"
```

The constant is added to the `DateFactory` mostly to avoid repeating `Y-m-d\TH:i:s.uP` every time the user want to validate a date. Also used by the helper.

> This separate PR #42475 allows `date_format:json` to avoid shoving the constant or using the full format string.

## BC?

None, only adding a method.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -350,7 +350,7 @@ trait InteractsWithInput
      * @param  string  $key
      * @return \Illuminate\Support\Carbon|null
      */
-    public function jsDate($key)
+    public function jsonDate($key)
     {
         return $this->date($key, DateFactory::JSON_DATE)?->utc();
     }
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #42475, opened 2022-05-22
[9.x] Adds JSON format helper to validate ISO-8601/RFC-3339 dates

DarkGhostHunter:
## What?

Aliases ISO-8601/RFC-3339 date format to `json` when using the `date_format` rule:

```php
validator([
    'kickoff' => '2020-01-01T21:30:25.123456+05:00',
], [
    // Before
    'kickoff' => 'date_format:Y-m-d\TH:i:s.uP',
    // After
    'kickoff' => 'date_format:json',
]);
```

This is a separate PR for #42474, where the validation rule must use a const or use that format.

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -13,6 +13,7 @@ use Egulias\EmailValidator\Validation\RFCValidation;
 use Exception;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
+use Illuminate\Support\DateFactory;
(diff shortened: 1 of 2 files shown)


johanrosenson:
Just curious, why is the alias `json` and not `iso-8601` or `rfc-3339`?

Is there some spec that says these date formats should be used in JSON or are they just extra commonly used?

DarkGhostHunter:
There is no spec for JSON dates, but JSON comes from JavaScript, and Date objects are encoded using that format in particular. 

I thought that using the spec name would make more sense too, but using `json` is more self-explanatory about the context.

Another alternative would be to create a separate rule called `date_json`, since the date in a JSON string follows that spec only (the `date_format` supports multiple formats).

PR #42490, opened 2022-05-24
[9.x] Added withoutCount method in Eloquent Builder

4msar:
Added `withoutCount` method in Eloquent Builder

Sometime we have to ignore the withCount property in runtime that's why I've added a method in eloquent builder which is `withoutCount`

### Usage

```php

$categories = Category::latest()->withoutCount('articles')->get();
// or 
$categories = Category::latest()->withoutCount('articles', 'votes')->get();
// or
$categories = Category::latest()->withoutCount(['articles', 'votes'])->get();

```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -185,6 +185,19 @@ class Builder implements BuilderContract
 
         return $this;
     }
+    
+    /**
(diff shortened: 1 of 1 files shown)


4msar:
Need to update this PR more, so I just draft this.

bernardwiesner:
Is there any other alternative to run a Model query without the `withCount` ? For the `with` we can use `query()->setEagerLoads([])`.

PR #42491, opened 2022-05-24
[9.x] Model::whereRelation add callback function

adideas:
Example
```php
// I want to optimize the query. In callback function.
$userSQL = User::whereRelation(
    "posts",
    function($query) {
        $query->limit(1)->where("public", true);
    }
)->toSql();
```
Before this PR
```sql
select * 
from "users" 
where exists (
   select *
   from "posts" 
   where "users"."id" = "posts"."user_id" and ("public" = ?)
)
```

After this PR
```sql
select * 
from "users" 
where exists (
   select *
   from "posts" 
   where "users"."id" = "posts"."user_id" and "public" = ? limit 1
)
```

I added tests for checks.
This is not in the documentation, but it is logical.
It would be very convenient.

I think we're talking about a supplement. I'm not sure if this is a mistake. But many expect a different result using this method.


adideas:
I can change my approach. 
Please write your idea, I'll change it.

I followed the Laravel approach.
This is not a critical addition.

PR #42498, opened 2022-05-24
[10.x] Cleaning up unused function in RedirectResponse

vanliederen:
Removal unused function `getOriginalContent` in redirectResponse in Laravel 10.x


rodrigopedra:
This is overriding the method with the same name defined in `ResponseTrait` trait. 

I guess the idea is to prevent trying to process any response output, as a redirect response has no actual content besides its headers.

PR #42500, opened 2022-05-24
[9.x] Add Conditionable Trait to Illuminate\Support\Carbon

ralphjsmit:
Today I had a situation where I needed to set the week-number of a Carbon date, but only when a user selected a filter. Otherwise the Carbon instance should default to the current time. 

I think that many developers as well have had the situation where they wanted to conditionably apply a method to a Carbon object. At least, I've had this feeling multiple times previously.

I could've done this with an `if`-condition, but I really missed the elegance of using the `->when()` method from the Conditionable, so this looked like the perfect situation for adding the Conditionable trait. 

### Simple example

```php
// Could be `null` or `[$year, $week]`
$filter = /** */

// Old:
$carbon = $filter ? Carbon::now()->setIsoDate(...$filter) : Carbon::now();

// New:
$carbon = Carbon::now()->when($filter, fn(Carbon $carbon) => $carbon->setIsoDate(...$filter));
```

I hope this will be useful to many developers!

### DIFF
--- a/src/Illuminate/Support/Carbon.php
+++ b/src/Illuminate/Support/Carbon.php
@@ -4,9 +4,12 @@ namespace Illuminate\Support;
 
 use Carbon\Carbon as BaseCarbon;
 use Carbon\CarbonImmutable as BaseCarbonImmutable;
+use Illuminate\Support\Traits\Conditionable;
 
 class Carbon extends BaseCarbon
 {
+    use Conditionable;
+
     /**
      * {@inheritdoc}
      */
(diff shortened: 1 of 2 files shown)


DarkGhostHunter:
I was just gonna make this PR...

driesvints:
I actually think the old way is so much cleaner here 😅

ralphjsmit:
> I actually think the old way is so much cleaner here 😅

😅 There might be better examples, where it's more of an improvement. This is just rather simplified, as everyone knows how the Conditionable works :)

How about this?

```php
        $carbon = Carbon::now()
            ->when($filter, fn(Carbon $carbon) => $carbon->setIsoDate(...$filter))
            ->startOfWeek();
        
        // Versus
        $carbon = $filter 
            ? Carbon::now()->setIsoDate(...$filter)->startOfWeek()
            : Carbon::now()->startOfWeek();
        
        // Or with proxies...
        $carbon = Carbon::now()
            ->when($filter)->setIsoDate(...$filter)
            ->startOfWeek();
```

It's just that I prefer closures over proxies, because it's annoying to lose autocompletion.

PR #42507, opened 2022-05-25
[9.x] add enum support to collection

agus24:
REF #42043 #42031

when doing `groupBy` or `keyBy` with enum field that casted from a model sometimes it returns an enum instance. it will cause an error `array_key_exists(): Argument #1 ($key) must be a valid array offset type`.
for example like below

```php
enum CustomerStatus: string
{
    case Bussines = 'bussines';
    case Private = 'private';
}
```

```php
class Customer  extends Model {
    protected $casts = [
    ];
}
```

then when we do 
```php
Customer::all()->groupBy('status')
``` 

it will cause an error 
```
array_key_exists(): Argument #1 ($key) must be a valid array offset type
```

we need to workaround with enum for example `Customer::all()->groupBy('status.value')`
this PR will make accessing value of an enum directly like `Customer::all()->groupBy('status')`

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -5,6 +5,7 @@ namespace Illuminate\Support\Traits;
 use CachingIterator;
 use Closure;
 use Exception;
+use UnitEnum;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Contracts\Support\Jsonable;
 use Illuminate\Support\Arr;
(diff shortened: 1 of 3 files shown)


Peshino:
When can we expect this issue to be fixed please? Maybe I'm missing something, but I'm still getting an error `array_key_exists(): Argument #1 ($key) must be a valid array offset type`. I'm not sure if modifing only the **groupBy** method is the solution because for example working with **mapToGroups** method is also not working (`Illegal offset type`) with enum field that casted from a model.
Thanks in advance for a reply.

PR #42509, opened 2022-05-25
Added ucwords to String helper

bretto36:
I needed to take an uppercase string, lower it then upper case the first letter of each word. 

Currently i have to do this: `ucwords((string)Str::of($value)->lower())`,

Now i can do this: `(string)Str::of($value)->lower()->ucwords()`,

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -739,6 +739,18 @@ class Str
     }
 
     /**
+     * Uppercase the first character of each word in a string.
+     *
+     * @param  string  $value
+     * @param  string  $separators
+     * @return string
+     */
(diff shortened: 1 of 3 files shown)


derekmd:
Would the `title()` method already cover this?

```php
Str::of('MIXED case WoRdS')->title()->value()
// "Mixed Case Words"
```

bretto36:
Ah probably. I thought it must have been in there I just couldn’t find it. Thanks

On 25 May 2022, at 4:59 pm, Derek MacDonald ***@***.***> wrote:

﻿

Would the title() method already cover this?

Str::of('MIXED case WoRdS')->title()->value()
// "Mixed Case Words"

PR #42512, opened 2022-05-25
[9.x] Add iterable type hinting to Eloquent concerns

markwalet:
Changes the `array` type hinting to `string[]`.

This was already partially done in 9c047344505da509c2b893eed7aac369a73e07eb and this PR applies the same principle to the other properties in the traits.

I could add this extra typehinting to the different getters/setters in the trait if that is desirable. But the original PR didn't include them.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/GuardsAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/GuardsAttributes.php
@@ -28,7 +28,7 @@ trait GuardsAttributes
     /**
      * The actual columns that exist on the database and can be guarded.
      *
-     * @var array
+     * @var string[]
      */
     protected static $guardableColumns = [];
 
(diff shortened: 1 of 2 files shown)


rodrigopedra:
They moved to another syntax, if I am not mistaken to fit better static analyzers

https://github.com/laravel/framework/blob/21959000a2d64e1aae8d8c12d256c0a0234715f2/src/Illuminate/Collections/Traits/EnumeratesValues.php#L63-L68

markwalet:
That's fine by me. I'll change them later this day.

rodrigopedra:
It should be something like this:

~~~php
* @var array<int, string> 
~~~

The first parameter inside the generic specification is the key type, and the second is the value type. It reads like this: "An array with integer keys and string values".

Sorry if I look picky, I know it is verbose, I just want to help get it merged =)

markwalet:
When an iterable doesn't have an explicit key type (in some cases only the value is used), you can use `array<string>` instead. Laravel doesn't care if you use string or int keys for those arrays so that's why I kept it open. 

I used this as an example: https://phpstan.org/writing-php-code/phpdoc-types#general-arrays

markwalet:
You're not picky by the way, I like the strict reviewing haha. But I'm not sure if the iterable really needs a key type in this case :)

PR #42518, opened 2022-05-25
[9.x] New methods of eloquent - createAnd and updateAnd

m-mohammadi1:
Most of the time we need to do this
```php
$model = User::create[...]);

if ($model) {
    // do something ...
}
```
for this one we could just use this shortcut
```php
$model = User::createAnd([...], function ($created_model) {
      // do something ...
  });
```

And for the same reason we could use updateAnd on Model and Eloquent Builder.
For Model
```php
$model = Model::find(1);
$model->updateAnd([...], function ($updated_model) {
    // more code ...
});
```
on Builder we can use this
```php
$models = Model::query();

$models->updateAnd([...], function ($updated_models) {
    // code more ...
});
```

I think it is good practice we do every thing related to an stuff in a area that relates to it.
and these are like an inline model observers that we can use in our logic.


src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -608,6 +608,44 @@ public function firstOr($columns = ['*'], Closure $callback = null)
         return $callback();
     }
 
+    /**
+     * Create record in database and call user given callback with created model.
+     *
+     * @param  array  $attributes
+     * @param  Closure  $callback
+     * @return mixed|null
```
GrahamCampbell, line 616: `mixed` cannot exist next to any other type in a union. `null` is already contained in `mixed`.

GrahamCampbell (changes requested):

rodrigopedra:
Why not using `tap()`?

~~~php
\tap(new Model, function ($model) {
 // do something ...
});
~~~

reference: https://laravel.com/docs/9.x/helpers#method-tap

src/Illuminate/Database/Eloquent/Builder.php
m-mohammadi1: > Why not just use `tap()`?
> 
> ```
> \tap(new Model, function ($model) {
>  // do something ...
> });
> ```
> 
> reference: https://laravel.com/docs/9.x/helpers#method-tap
I think this is more user friendly and can help new users and actually it is obvious what it does.

m-mohammadi1: > `mixed` cannot exist next to any other type in a union. `null` is already contained in `mixed`.

yes I should fix it.

m-mohammadi1:
> Why not just use `tap()`?
> 
> ```
> \tap(new Model, function ($model) {
>  // do something ...
> });
> ```
> 
> reference: https://laravel.com/docs/9.x/helpers#method-tap

I think this is more user friendly and can help new users and actually it is obvious what it does.

taylorotwell:

Issue #40264, opened 2022-01-05
Rules digits and digits_between don't work with fractional numbers


sera527:
- Laravel Version: 5.2 and later

### Description:
I believe that the fix on this [issue](https://github.com/laravel/framework/issues/14647) should not have been approved. The rules now only supports integer values. And this does not match the description in the documentation: "The field under validation **must be numeric** and must have an exact length of value."

### Steps To Reproduce:

```php
<?php

$validator = Validator::make(
    ['fractional_value' => 1.2],
    ['fractional_value' => 'digits_between:1,10']
);
$validator->errors()->first(); // The fractional value must be between 1 and 10 digits.

driesvints:
We've merged this and it'll be released next Tuesday. Thanks for reporting.

PR #42522, opened 2022-05-25
[9.x] New validation rules `float` and `float_between`

taai:
Validation rules ```digits``` and ```digits_between``` validated, well, digits. They worked perfectly fine since [August 6, 2016](https://github.com/laravel/framework/pull/14650), for more than 5 years… On January 5, 2022, a **single** person suddenly woke up and [demanded](https://github.com/laravel/framework/issues/40264) that the rules should actually accept floats (to allow a dot between digits) and his wish was [fulfilled](https://github.com/laravel/framework/pull/40278) the next day.

Since nobody cares that [floats are not actually digits](https://github.com/laravel/framework/issues/42326) and [does not want to compromise](https://github.com/laravel/framework/pull/42339), here are 2 **"new"** validation rules that validate floats (digits, but nobody cares about the difference, right?).

Currently, [there are no other validation rules](https://github.com/laravel/framework/pull/42339#issuecomment-1122779620) (or rule combinations) that can guarantee that the input contains only digits and that the input contains only a specific number of digits.

For example, with the **"new"** validation rules ```float``` and ```float_between``` you can validate:
* PIN codes (```float:4``` passes ```0123``` and ```float_between:4,8``` passes ```01234567```)
* US Zip codes (```float:5``` passes ```02421```)
* Phone numbers
* Digits!

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1031,6 +1031,54 @@ trait ValidatesAttributes
     }
 
     /**
+     * Validate that an attribute has a given number of digits.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @param  array<int, int|string>  $parameters
+     * @return bool
+     */
+    public function validateFloat($attribute, $value, $parameters)
+    {
+        $this->requireParameterCount(1, $parameters, 'float');
+
+        if (!is_numeric($value)) {
+            return false;
+        }
+
+        if (!is_string($value)) {
+            $value = strval($value);
+        }
+
+        return strlen($value) == $parameters[0] && ctype_digit($value);
+    }
+
+    /**
+     * Validate that an attribute is between a given number of digits.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @param  array<int, int|string>  $parameters
+     * @return bool
+     */
+    public function validateFloatBetween($attribute, $value, $parameters)
+    {
+        $this->requireParameterCount(2, $parameters, 'float_between');
+
+        if (!is_numeric($value)) {
+            return false;
+        }
+
+        if (!is_string($value)) {
+            $value = strval($value);
+        }
+
+        $length = strlen($value);
+
+        return $length >= $parameters[0] && $length <= $parameters[1] && ctype_digit($value);
+    }
+
+    /**
      * Validate that an attribute is greater than another attribute.
      *
      * @param  string  $attribute
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1042,11 +1042,11 @@ trait ValidatesAttributes
     {
         $this->requireParameterCount(1, $parameters, 'float');
 
-        if (!is_numeric($value)) {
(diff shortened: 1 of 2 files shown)


driesvints:
What about `number` and `number_between`?

taai:
> What about `number` and `number_between`?

I would be **very** happy to swap the names of ```float``` and ```float_between``` with ```digits``` and ```digits_between```.


Let's talk about floats (really, floats)!

Name ```float_between``` or ```number_between``` suggests that it does the same as ```min``` and ```max``` (compares the numeric value, not the number of characters) - that's why I suggest **not** to use names ```number```, ```number_between```, ```float```, ```float_between```.

The word ```number``` is ambiguous - there are many types of numbers (digits, integers, floats, hexadecimals, binary, scientific). If there is a need for validating any type of numbers, use the already existing rule ```numeric```!
Float is unambiguous - every programmer must know [what float is](https://en.wikipedia.org/wiki/Floating-point_arithmetic) and that it may contain a dot (```.```). In my opinion, ```decimal``` could be **even better name** than ```float``` - that makes more sense, because floats are prone to rounding errors, but we are talking about strings and precise length.

Floats/decimals are more complicated than digits. If there would be a rule ```float``` (or ```decimal```), it should validate that the input **is** a valid decimal number (that is either zero or does not star with zero like ```01``` or ```00.123```) at all. And I would not be interested in the complete length of the whole input, but I would be interested in precision - how many digits are before the dot and how many digits are after the dot. For example, when I'm validating product price, I'm talking about numbers like ```1.99``` (I suggest ```decimal:12,2```), but when I'm validating GPS coordinates (using the same precision what *Google Maps* use), I'm talking about numbers like ```60.8656662``` (I suggest ```decimal:9,7```) and ```159.6511265``` (I suggest ```decimal:10,7```). And probably there is no need for ```decimal_between``` rule for floats/decimals when we are talking about precision - usually the minimum number of digits does not matter, only the maximum. Of course, floats can be also negative numbers.

And if we are talking about ```decimal_between``` rule - it should not compare precision of digits, but it should compare the float/decimal value itself by using PHP function [bccomp](https://www.php.net/manual/en/function.bccomp.php). For example, if I would want the user pay at least one cent, I would use validation rule ```decimal_between:0.01,1000000,2``` (the third parameter would check the precision - limit the number of digits after the dot).

In conclusion...
* Do you agree to change back the previous behavior of rules ```digits``` and ```digits_between```?
* Shall I make a validation rule ```decimal``` and ```decimal_between``` as I described them?

I'm willing to program all of this, of course, just give me a green light! :wink:

taylorotwell:
Honestly I wonder if that dot change should just be reverted entirely.

taylorotwell:
Do you actually need a new validation rule if you just combine the `integer` and `digits` rules?

taai:
> Do you actually need a new validation rule if you just combine the `integer` and `digits` rules?
> 
> `'pin' => 'required|integer|digits:4'`

@taylorotwell As [I have already answered you](https://github.com/laravel/framework/pull/42339#issuecomment-1122779620), the validation rule ```integer``` will fail on pin number ```0123```.

```php
$value = '0123';
$isValid = filter_var($value, FILTER_VALIDATE_INT) !== false; // false
```



> Honestly I wonder if that dot change should just be reverted entirely.

Yes, please!

taylorotwell:
Yeah, agree. Reverting it now.

Issue #41313, opened 2022-03-03
Factories with relationship fields are calling create() when calling make()


robertoarruda:
### Discussed in https://github.com/laravel/framework/discussions/36169

<div type='discussions-op-text'>

<sup>Originally posted by **robertoarruda** February  6, 2021</sup>
- Laravel Version: 8.22.0
- PHP Version: 7.4.13

### Description:
I use factories in my unit tests. My tests never touch the database, so I don't use connections or migrations on it.
In my unit tests, factories use the make method only to provide fake data. However, even using the make method, apparently the related factories are using the create method.
My code has several relationships in the factories, but below is a simple example of the problem.

### Steps To Reproduce:
PositionFactory:
```php
<?php

namespace Database\Factories;

use Illuminate\Database\Eloquent\Factories\Factory;
use App\Models\Firm;
use App\Models\Position;

class PositionFactory extends Factory
{
    /**
     * @var string
     */
    protected $model = Position::class;

    /**
     * @return array
     */
    public function definition(): array
    {
        return [
            'name' => $this->faker->name,
            'description' => $this->faker->text,
            'firm_id' => Firm::factory(),
        ];
    }
}

``` 
FirmFactory:
```php
<?php

namespace Database\Factories;

use Illuminate\Database\Eloquent\Factories\Factory;
use App\Models\Firm;

class FirmFactory extends Factory
{
    /**
     * @var string
     */
    protected $model = Firm::class;

    /**
     * @return array
     */
    public function definition(): array
    {
        return [
            'name' => $this->faker->company,
            'description' => $this->faker->paragraph,
            'website' => $this->faker->url,
            'discount' => $this->faker->randomFloat(),
            'supplier' => $this->faker->word,
            'status' => $this->faker->randomElement(['active', 'disabled']),
        ];
    }
}

``` 

The error presented is when running the tests, it is the attempt to create a user. I imagine it’s because most relationships are tied to that table
```shell
  • Tests\UT\Repositories\TemplateFieldPropertyValueRepositoryTest > attach dependency existing dependency
   Illuminate\Database\QueryException 

[...]

PR #42524, opened 2022-05-25
Avoid creating data when make (#41313)

robertoarruda:
This update prevents a [model factory with a defined relationship field](https://laravel.com/docs/9.x/database-testing#defining-relationships-within-factories) from creating data in the database when using `make`.

For that we create the status field `isCreating`, which is set to true only when a model factory is being created, and then we check this new field to see if it can create the related model factory or not.

This update may require user adjustments to existing tests.
Example:
```php
class AddressFactory extends Factory
{
    public function definition(): array
    {
        return [
            'user_id' => User::factory(),
        ];
    }
}
```

Before:
```php
Address::factory()->make()->user_id;
```

After:
```php
Address::factory()->make(['user_id' => 1])->user_id;
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/Factory.php
+++ b/src/Illuminate/Database/Eloquent/Factories/Factory.php
@@ -90,6 +90,13 @@ abstract class Factory
     protected $faker;
 
     /**
+     * The "is creating" is true when a factory is created.
+     *
+     * @var bool
+     */
+    protected $isCreating = false;
+
+    /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
This would be a breaking change.

robertoarruda:
> This would be a breaking change.

But without this adjustment, in the example below, a new data (user) will be created in the database without the consent of the developer. In my opinion this is a serious mistake.

```php
Address::factory()->make();
```

PR #42525, opened 2022-05-25
[9.x] Add `pipe` helper

michael-rubel:
### Inspiration
This helper was inspired by the combination of `tap()` and `collect()->pipe()`. Sometimes it's handy if you can pipe something without creating a collection, the same manner we can go with `tap()`.

### Example
```php
return pipe(static::$tooLongObjectName, function ($shortened) {
    if ($shortened->someMethodIsTrue()) {
        return false;
    }

    if (! $shortened->anotherMethodIsFalse()) {
        return false;
    }

    return true;
});
```

### DIFF
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -305,6 +305,20 @@ if (! function_exists('tap')) {
     }
 }
 
+if (! function_exists('pipe')) {
+    /**
+     * Pass the value to the given callback and return the result.
+     *
+     * @param  mixed  $value
+     * @param  callable  $callback
+     * @return mixed
+     */
+    function pipe($value, $callback)
+    {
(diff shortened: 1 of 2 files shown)


crynobone:
You can already use `with()` or `transform()` at the moment.

DarkGhostHunter:
```php
return with(static::$tooLongObjectName, function ($shortened) {
    if ($shortened->someMethodIsTrue()) {
        return false;
    }

    if (! $shortened->anotherMethodIsFalse()) {
        return false;
    }

    return true;
});
```

michael-rubel:
@DarkGhostHunter Thanks, but it's more about consistency with the method available in the collection. I think it's not only me who may look for this method after using it in collections. Let's see what @taylorotwell says.

taylorotwell:

PR #38486, opened 2021-08-21
[8.x] Avoid notifications locale being null

milewski:
This fixes something that I few is very annoying when working with localized notifications, currently, whenever we use the `HasLocalePreference` on any notifiable class, we can only retrieve the locale via `app()->getLocale()` instead of `$this->locale`, for example:

https://laravel.com/docs/8.x/notifications#user-preferred-locales

```php
use Illuminate\Contracts\Translation\HasLocalePreference;

class User extends Model implements HasLocalePreference
{
    public function preferredLocale()
    {
        return $this->locale;
    }
}
```

```php
class SampleNotification extends Notification
{
    public function toArray(): array
    {
        return [
            'example1' => trans('notifications.example1', locale: app()->getLocale()),
            'example2' => trans('notifications.example2', locale: $this->locale) // this will be null,
        ];
    }
}
```

But when setting the locale manually on the notification, the following works as expected:

```php
$notifiable->notify((new SampleNotification())->locale('hue_BR'));
```

```php
class SampleNotification extends Notification
{
    public function toArray(): array
    {
        return [
            'example1' => trans('notifications.example1', locale: app()->getLocale()), // works as expected
            'example2' => trans('notifications.example2', locale: $this->locale) // works as expected,
        ];
    }
}
```

So what this PR is doing is `ALWAYS` setting the property `$locale` on the notifications in order to make the `$this->locale` more reliable 

This is not a breaking change.

PR #30429, opened 2019-10-25
[6.x] Queued notifications respect user's preferred locale

thijskok:
We use both queued and non-queued notifications and rely on the **preferredLocale** method for determining the notification's locale. While the non-queued one's are translated using the correct locale, the queued one's still used the default app locale.

It turns out queued notifications can only be configured using the **locale** fluent method. The preferredLocale method isn't referenced or available in this case. The [documentation](https://laravel.com/docs/6.x/notifications#localizing-notifications) doesn't mention this exception.

This allows you use the preferredLocale method for notifications, whether you use queuing or not.

PR #25752, opened 2018-09-24
[5.7] Mail recipient and notifiable preferred locale

derekmd:
This wraps up localized notifications introduced by https://github.com/laravel/framework/pull/23178, https://github.com/laravel/framework/pull/24451, and https://github.com/laravel/framework/pull/24919.

Users can have a preferred locale (likely stored in the `users` database table) which the `Mail` and `Notification` services will use when queuing jobs.

## Why store a user locale preference?

1. To allow admin-triggered or automated scheduled notifications in the expected language.
2. Clean up excessive need to call `Mail::locale()` and `Notification::locale()`. A typical localized scenario:
   * Default app locale: `'en'`
   * http://mydomain.com/de/ requests have middleware call `App::setLocale('de')` for German translations
   * When queued, this mailable is actually delivered using the app's default `'en'` instead of '`de`'
     ```php
     User::create(request()->all());
     Mail::to($user)->queue(new OrderConfirmation($order));
     ```
   * This call will translate the mailable as expected:
     ```php
     Mail::to($user)->locale(app()->getLocale())->queue(new OrderConfirmation($order))
     ```
   * However this means **every translated mailable sent in a codebase needs to call** `Mail::to()->locale()`
     * likewise for `Notification::locale()`

## Example `User` model with stored locale preference

```php
use Illuminate\Contracts\Translation\HasLocalePreference;

class User extends Model implements HasLocalePreference
{
    public function preferredLocale()
    {
        return $this->locale;
    }
}
```

```php
$user = User::create([
    'email' => 'taylor@laravel.com',
    'locale' => 'de',
]);
```

These four calls will send the notification in German:

* `$user->notify(new OrderConfirmation($order))`
* `Notification::send($user, new OrderConfirmation($order))`
* `Mail::to($user)->queue(new OrderConfirmation($order)`
* `Mail::to($user)->send(new OrderConfirmation($order)`

## Notifying multiple recipients with mixed locale preferences

```php
$users = collect([
    User::create([
        'email' => 'taylor@laravel.com',
        'locale' => 'en',
    ]),
    User::create([
        'email' => 'mohamed@laravel.com',
        'locale' => 'ar',
    ]),
]);
```

This call will notify Taylor in English and Mohamed in Arabic:

* `Notification::send($users, new OrderConfirmation($order))`

This call will send the email _in the application's default locale_:

* `Mail::to($users)->queue(new OrderConfirmation($order))`

[...]

PR #42527, opened 2022-05-25
[9.x] Use `preferredLocale` when notification is queued

welderlourenco-candc:
Description
--

[Laravel's documentation](https://laravel.com/docs/8.x/mail#localizing-mailables) states
> [...] Laravel allows you to send mailables in a locale other than the request's current locale, and will even remember this locale if the mail is queued.

> [...] Once you have implemented the interface, Laravel will automatically use the preferred locale when sending mailables and notifications to the model.

however, this is only functioning for synced notifications, this PR fixes this by adding the functionality of sending a notification in the user's preferred locale when a notification is queued.

Example
--

```php
// config/app.php
    'locale' => 'en',
    'fallback_locale' => 'en',

// en/notifications/example.php
return [
    'message' => 'English message'
];

// pt-br/notifications/example.php
return [
    'message' => 'Mensagem em português'
];

```

```php
class User extends Authenticatable implements HasLocalePreference
{
     // ...

     public function preferredLocale(): string
     {
         return 'pt-br';
     }

     // ...
}
```

```php
class ExampleNotification extends BaseNotification implements ShouldQueue
{
    use Queueable;
 
    /**
     * Method called by base notification, running inside worker, that fetches/formats a message.
     */
    protected function getMessage()
    {
        return __('notifications/example/message');
    }
}
```

```php
$user->notify(new ExampleNotification());

// expected message: 'Mensagem em português'
// rendered message: 'English message'
```

### DIFF
--- a/src/Illuminate/Notifications/NotificationSender.php
+++ b/src/Illuminate/Notifications/NotificationSender.php
@@ -193,6 +193,8 @@ class NotificationSender
 
                 $notification->id = $notificationId;
 
+                $notification->locale = $this->preferredLocale($notifiable, $notification);
+
                 if (! is_null($this->locale)) {
                     $notification->locale = $this->locale;
                 }


derekmd:
> however, this is only functioning for synced notifications

Queued notifications do support the `HasLocalePreference` contract: https://github.com/laravel/framework/pull/25752

Where is the `ExampleNotification::getMessage()` method being called? The details of the example's `BaseNotification` class need to be known to give enough context about what this is attempting to fix.

Related previous reports: https://github.com/laravel/framework/pull/30429 https://github.com/laravel/framework/pull/38486

welderlourenco-candc:
Thanks for replying, you're absolutely right. It's gotta be some heavy caching with out current queue setup, tested on a fresh laravel install and it works fine. Please disregard.

PR #42528, opened 2022-05-26
Revert "Revert "Update Encrypter.php""

diadal:
Reverts laravel/framework#42372

### DIFF
--- a/src/Illuminate/Encryption/Encrypter.php
+++ b/src/Illuminate/Encryption/Encrypter.php
@@ -148,7 +148,7 @@ class Encrypter implements EncrypterContract, StringEncrypter
      */
     public function decrypt($payload, $unserialize = true)
     {
-        $payload = $this->getJsonPayload($payload);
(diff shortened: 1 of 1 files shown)


derekmd:
Requesting to revert a revert by the framework's author is certainly a choice.

Stop allowing `null` to be passed into a parameter documented as `@param string $payload`.

diadal:
> Requesting to revert a revert by the framework's author is certainly a choice.
> 
> Stop allowing `null` to be passed into a parameter documented as `@param string $payload`.

this is not an issue on my side I just felt it best.

like you said it is a choice, I know about `@param string $payload` I also know how easy to bypass `@param` when it comes to security, files like `Encrypter.php` I will strongly advise all the `param` should be converted to expected variables and not rely on `@param` or make sure it crashes. anyway, I leave this to the Laravel team to decide.

you can be sure of `strval($str)` or `(stirng) $str`

PR #42533, opened 2022-05-26
[9.x] Make HasTimestamps::updateTimestamps chainable

leandro-hermes:
This PR makes the method `HasTimestamps::updateTimestamps` chainable, so we can do:

```php
$model->updateTimestamps()->save();
```


taylorotwell:
Don't mind merging this, but for your example you can just do `$model->touch()`

PR #42535, opened 2022-05-26
[9.x] Add html debug comments around components.

haringsrob:
This proposed change adds a small comment around blade components.

The purpose is to allow for easier debugging while inspecting the code using a browser inspector so that you can navigate faster or have more information about the context of what you are looking at.

In this implementation, I added this using `app.debug` but we can also introduce a new config variable.

Example:

Screenshot 2022-05-26 at 22 29 39


haringsrob:
This does not work with livewire as livewire expects the first element to be a div. Closing this for now until I have new ideas for better implementation. Maybe a data-component-info="component-name" attribute that we inject into the first element.

PR #42536, opened 2022-05-26
[9.x] Support kebab case for slot name shortcut

Waghabond:
The inline slot name shortcut was added in laravel 9 but it does not comply with the blade component kebab case naming convention. As a result I had to do some digging into why the slot names weren't working how I would expect them to. This PR will make the feature fit intuitively with the blade component naming conventions. 

Note: The slot's contents will be available as a variable with the slot name converted to snake_case.
e.g. 
```xml
<!-- /resources/views/components/my-layout.blade.php -->
<html lang="en">
    <body>
        {{ $my_slot }}
    </body>
</html>

<!-- /resources/views/components/child.blade.php -->
<x-my-layout>
    <x-slot:my-slot>
        <div>My super cool content</div>
    </x-slot:my-slot>
</x-my-layout>
```
Previously this would have to be done like this:  
```xml
<!-- /resources/views/components/child.blade.php -->
<x-my-layout>
    <x-slot:my_slot>
        <div>My super cool content</div>
    </x-slot:my_slot>
</x-my-layout>
```
Not bad but it doesnt fit conventions and a multiword slot-name example isn't provided in the documentation. Just a small oversight when the feature was added I guess.

Please let me know if these proposed changes will cause any breaks! They seem fine based on the testing I've done personally =)

### DIFF
--- a/src/Illuminate/View/Compilers/ComponentTagCompiler.php
+++ b/src/Illuminate/View/Compilers/ComponentTagCompiler.php
@@ -426,7 +426,7 @@ class ComponentTagCompiler
             <
                 \s*
                 x[\-\:]slot
-                (?:\:(?<inlineName>\w+))?
+                (?:\:(?<inlineName>\w+(-\w+)*))?
                 (?:\s+(:?)name=(?<name>(\"[^\"]+\"|\\\'[^\\\']+\\\'|[^\s>]+)))?
                 (?<attributes>
                     (?:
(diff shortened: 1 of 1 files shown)


Waghabond:
Excuse the repetitive reopening and closing of the pull request. I wasnt sure how to use github correctly

taylorotwell:
Personally I would prefer it to convert to camelCase if kebab case is used on the slot name. Are you able to make a PR for that instead?

Waghabond:
@taylorotwell sure can! gimme a sec should i just add a commit then reopen this or create a whole new PR?

PR #42538, opened 2022-05-26
[9.x] PendingMailFake matching PendingMail

CornerMonkey:
The `PendingMail` send function was altered to return the result of the mailer send call (see commit https://github.com/laravel/framework/commit/68971ff65d737a8caba125c72e4a4b8b25c1cec6).  However, this change was not reflected in the `PendingMailFake` class.  So, when testing the result of the send call will always return nothing when using `Mail::fake()`.  This makes it difficult to properly unit test functionality without these changes.


taylorotwell:
So what will it return when faked? Does it actually return a SentMessage instance?

driesvints:
We didn't change this because as you can see in the MailFake class, nothing is returned anyway. That's the whole thing about the faking of mails. There isn't anything sent.

https://github.com/laravel/framework/pull/42048#issuecomment-1105876702

PR #42545, opened 2022-05-27
[9.x] Adds `Str::isJson`

ecrmnn:
Adds `isJson()` to `Str` and `Stringable`.

```php
Str::isJson($data); 
//=> boolean

Str::of($data)->isJson()
//=> boolean

str($data)->isJson();
//=> boolean
```

Uses `JSON_THROW_ON_ERROR` flag introduced in PHP 7.3 under the hood.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -397,7 +397,7 @@ class Str
         if (! is_string($value)) {
             return false;
         }
-        
+
         try {
             json_decode($value, true, 512, JSON_THROW_ON_ERROR);
         } catch (JsonException) {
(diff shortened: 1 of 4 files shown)


Florisbosch:
Why is an empty and a null value expected to be false? 
JSON can be empty. If I persist a payload to the db with json and I want to make it empty or update with a new json payload it should test as true if it's empty. No need to make extra if statement to manually update null to a payload field in the db

ecrmnn:
@Florisbosch 

By it self `''` isn't valid JSON. 
```php
json_decode('', false, 512, JSON_THROW_ON_ERROR);

// [THROWS] JsonException with message 'Syntax error'
```

`null` isn't valid because this is a string helper, therefore `null` will never be valid because it's not a `string`

PR #42547, opened 2022-05-27
Added whereCreatedBefore/After, whereUpdatedBefore/After Eloquent methods

nickakitch:
This PR introduces four new Eloquent builder methods `whereCreatedBefore`, `whereCreatedAfter`, `whereUpdatedBefore`, and `whereUpdatedAfter`. Which provide nice syntax for querying database record creation/updates.

Example usage:
```php
$reports = Report::query()
    ->whereCreatedBefore('2022-01-01 16:00:00')
    ->orWhereUpdatedAfter('2022-02-05 10:00:00')
    ->get();
```
is equivalent to
```php
$reports = Report::query()
    ->where('created_at', '<', '2022-01-01 16:00:00')
    ->orWhere('updated_at', '>', '2022-02-05 10:00:00')
    ->get();
```

I image this will be particularly helpful for more complex reporting, or filtering records when working with large data sets. Especially when additional functions or methods are used to create the date-time portion.

```php
$reports = Report::query()
    ->whereCreatedBefore(Carbon::yesterday())
    ->get();
```

There is also the option to specify the column name as the second parameter:
```php
$reports = Report::query()
    ->whereCreatedBefore('2022-01-01 16:00:00', 'created')
    ->get();
```

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1632,8 +1632,8 @@ class Builder implements BuilderContract
     /**
      * Add a where updated before statement to the query.
      *
-     * @param  mixed $value
-     * @param  string $column
+     * @param  mixed  $value
+     * @param  string  $column
      * @return $this
      */
     public function andWhereUpdatedBefore($value, $column = 'updated_at')
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1536,8 +1536,8 @@ class Builder implements BuilderContract
     /**
      * Add a where created before statement to the query.
      *
-     * @param mixed  $value
-     * @param string  $column
+     * @param  mixed  $value
+     * @param  string  $column
      * @return $this
      */
     public function whereCreatedBefore($value, $column = 'created_at')
@@ -1548,8 +1548,8 @@ class Builder implements BuilderContract
     /**
      * Add a where created before statement to the query.
      *
-     * @param mixed  $value
-     * @param string  $column
+     * @param  mixed  $value
+     * @param  string  $column
      * @return $this
      */
     public function orWhereCreatedBefore($value, $column = 'created_at')
@@ -1560,8 +1560,8 @@ class Builder implements BuilderContract
     /**
      * Add a where created before statement to the query.
      *
-     * @param mixed  $value
-     * @param string  $column
+     * @param  mixed  $value
+     * @param  string  $column
      * @return $this
      */
     public function andWhereCreatedBefore($value, $column = 'created_at')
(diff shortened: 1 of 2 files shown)


ankurk91:
No more methods in eloquent please , just to save some keystrokes.

Similar PR was closed in past

nickakitch:
@ankurk91 why no more methods? Are you anti-improvement or something? Lol

DarkGhostHunter:
I barely have an use case for that.

johanrosenson:
The optional column name argument seems weird. It will accept any date column, not only updated/created. 

A more fitting name then would be `whereDateBefore` (already exists i believe?). 

Or just remove that argument since a method with this name should be assumed to only check the column in the method name, and can be assumed to be able to figure that out by itself. 
 
The whole method loses usefulness the second I have to tell it the column name, then I can just use any other eloquent date method.

johanrosenson:
Your methods doesn't support when the column names have been changed on the model, is it possible for this method to figure it out by itself?

nickakitch:
The optional column name parameter is there for cases when the database doesn't have the standard `created_at` or `updated_at` columns. I'm sure we've all worked with legacy database that use things like `LASTUPDATEDTIME` etc.

I'm thinking if there's an appetite for these methods then I can add in methods such as `->createdBetween(startDate, endDate)`. Again these would be overkill for simple cases, but I think provide clarity for more complex queries.

johanrosenson:
> The optional column name parameter is there for cases when the database doesn't have the standard `created_at` or `updated_at` columns. I'm sure we've all worked with legacy database that use things like `LASTUPDATEDTIME` etc.
> 
> I'm thinking if there's an appetite for these methods then I can add in methods such as `->createdBetween(startDate, endDate)`. Again these would be overkill for simple cases, but I think provide clarity for more complex queries.

Yes, but the Laravel Model already have methods to retrieve the column name if it's been changed to anything other than the standard.

https://laravel.com/docs/9.x/eloquent#timestamps

Constants on the model: https://github.com/laravel/framework/blob/acddea7ab355f218f193bea071d6d63257a68c46/src/Illuminate/Database/Eloquent/Model.php#L189

methods to retrieve said constants: https://github.com/laravel/framework/blob/acddea7ab355f218f193bea071d6d63257a68c46/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php#L124

My opinion is that if your methods can't figure this out by itself, then the methods should be generic and named `whereDateBefore` / `whereDateAfter` instead.

The power of these constants is that you define them in one place and the rest of the system recognizes this, I just believe that methods such as yours should take advantage of this.

The model is set when creating a new Eloquent query builder, so you should be able to retrieve these values during runtime in the Eloquent query builder.

https://github.com/laravel/framework/blob/acddea7ab355f218f193bea071d6d63257a68c46/src/Illuminate/Database/Eloquent/Model.php#L1375

These are only my personal opinions on how your methods should be structured, I'm not convinced these methods are a good idea anyway, but if they are I think they should be changed, but even better would be more generic methods.

But as @ankurk91 says; similar PRs have been closed recently:
* https://github.com/laravel/framework/pull/39332
* https://github.com/laravel/framework/pull/42081

These where generic methods, yours only handles 2 columns (but adds 12 methods to the query bilder), I'm not sure how much chance it stands.

nickakitch:
Yep, you make good points - I'll add in support for the constants.

taylorotwell:

PR #42548, opened 2022-05-27
[9.x] Adds `glue()` method to Arr (better implode)

DarkGhostHunter:
## What?

Adds the `glue()` method to implode multiple subjects with a string that doesn't repeat between each of them, because `implode` doesn't care.

```php
$subjects = ['category/', '/post' , '/comment/'];

implode($subjects, '/');
// 🤦‍♂️ "category///post//comment/"

Arr::glue($subjects, '/');
// 😍 "category/post/comment"
```

I believe this helper could be used for handling paths or small lists with commas.

It also goes the extra mile to ensure the glue... well, glues a subject equal to the glue, because is expected that `glue` to work almost like `implode()`.

```php
Arr::glue(['foo', 'bar', 'baz'], 'foo');
// foofoobarfoobaz
```

It also supports final glue like `join()`.

```php
Arr::join(['a, ', 'b, ', ' and c'], ', ', ' and '));
// 🤦‍♂️ a, , b,  and  and c

Arr::glue(['a, ', 'b, ', ' and c'], ', ', ' and '));
// 😍 a, b and c
```

> This method **does not trim** the subjects, otherwise _hilarity ensues_.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -455,6 +455,51 @@ class Arr
     }
 
     /**
+     * Implode multiple subjects with a string that doesn't repeat between each subject.
+     *
+     * @param  array<int|string, string>  $subjects
+     * @param  string  $glue
+     * @param  string  $finalGlue
+     * @return string
+     */
+    public static function glue($subjects, $glue, $finalGlue = '')
+    {
+        if (count($subjects) === 0) {
+            return '';
+        }
+
+        if (count ($subjects) === 1) {
+            return end($subjects);
+        }
+
+        if ($finalGlue === '') {
+            $finalGlue = $glue;
+        } elseif (count($subjects) === 2) {
+            $glue = $finalGlue;
+        }
+
+        $first = array_shift($subjects);
+        $last = array_pop($subjects);
+
+        $first = Str::beforeLast($first, $glue) ?: $first;
+        $last = Str::after($last, $finalGlue) ?: $last;
+
+        if (count($subjects) === 0) {
+            return $first.$finalGlue.$last;
+        }
+
+        foreach ($subjects as $key => $subject) {
+            if ($subject !== $glue) {
+                $subjects[$key] = Str::between($subject, $glue, $glue);
+            }
+        }
+
+        array_unshift($subjects, $first);
+
+        return implode($glue, $subjects).$finalGlue.$last;
+    }
+
+    /**
      * Key an associative array by a field or using a callback.
      *
      * @param  array  $array
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -490,7 +490,7 @@ class Arr
 
         foreach ($subjects as $key => $subject) {
             if ($subject !== $glue) {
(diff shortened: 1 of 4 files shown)


johanrosenson:
> Arr::join(['a ', ' b ', ' c'], ' ', ' and ');
> // 🤦‍♂️ a ,  b  and  c
> 
> Arr::glue(['a ', ' b ', ' c'], ' ', ' and ');
> // 😍 a, b and c

Your examples are not correct, right?

Shouldn't it be:

```
Arr::join(['a ', ' b ', ' c'], ', ', ' and ');
// 🤦‍♂️ a ,  b  and  c

Arr::glue(['a ', ' b ', ' c'], ', ', ' and ');
// 😍 a, b and c
```

DarkGhostHunter:
> > Arr::join(['a ', ' b ', ' c'], ' ', ' and ');
> > // 🤦‍♂️ a ,  b  and  c
> > Arr::glue(['a ', ' b ', ' c'], ' ', ' and ');
> > // 😍 a, b and c
> 
> Your examples are not correct, right?
> 
> Shouldn't it be:
> 
> ```
> Arr::join(['a ', ' b ', ' c'], ', ', ' and ');
> // 🤦‍♂️ a ,  b  and  c
> 
> Arr::glue(['a ', ' b ', ' c'], ', ', ' and ');
> // 😍 a, b and c
> ```

Yep, fixed.

johanrosenson:
> ```
> Arr::join(['a ', ' b ', ' c'], ', ', ' and ');
> // 🤦‍♂️ a ,  b  and  c
> 
> Arr::glue(['a ', ' b ', ' c'], ', ', ' and ');
> // 😍 a, b and c
> ```

Can you add these examples to your tests? I don't think they pass.

I test ran your method in Tinkerwell and i don't get the same output as your examples.

DarkGhostHunter:
> > ```
> > Arr::join(['a ', ' b ', ' c'], ', ', ' and ');
> > // 🤦‍♂️ a ,  b  and  c
> > 
> > Arr::glue(['a ', ' b ', ' c'], ', ', ' and ');
> > // 😍 a, b and c
> > ```
> 
> Can you add these examples to your tests? I don't think they pass.
> 
> I test ran your method in Tinkerwell and i don't get the same output as your examples.

You're right, `glue()` won't trim the strings, it will respect their spaces in between. I fixed it now.

johanrosenson:
Great great, I have another edge for you, what would be the expected output of this example?

```
Arr::glue(['a,', 'b'], ', '), // "a, b" or "a,, b" ?
```

The reason I'm asking is that trying to use multi length glue in methods like these always produces a lot of edge cases when the glue only partially matches the start/end of the segments.
I've have done some experiments in the same area previously and ran in to many edge cases.

DarkGhostHunter:
> ```
> Arr::glue(['a,', 'b'], ', '), // should it be "a, b" or "a,, b" ?
> ```

No prob. Guys had it tough before the internetz tho.

The glue should match exactly the start or the end. In that case, it should be `a,, b`. 

Most of edge cases I can think of can be fixed by pushing the subjects after a `trim`. `Arr::trim()` could be an excellent addition to trim multiple subjects.

taylorotwell:

PR #42550, opened 2022-05-28
[9.x] Allows BelongsToMany sync method to bulk insert the new relations

pmcpimentel:
A simple refactor that allows the `$entity->relation()->sync()` method used by the `InteractsWithPivotTable` trait method to bulk insert any new relations sent to the method, instead of inserting them one by one.

**Before:**
```
$user->roles()->sync([1, 2, 3, 4, 5]);

// Results in:
// INSERT INTO user_role (user_id, role_id) VALUES (1, 1)
// INSERT INTO user_role (user_id, role_id) VALUES (1, 2)
// INSERT INTO user_role (user_id, role_id) VALUES (1, 3)
// INSERT INTO user_role (user_id, role_id) VALUES (1, 4)
// INSERT INTO user_role (user_id, role_id) VALUES (1, 5)

$user->roles()->sync([4, 5, 6, 7, 8, 9, 10]);

// Results in:
// DELETE FROM user_role WHERE user_id = 1 AND role_id in (1, 2, 3)
// INSERT INTO user_role (user_id, role_id) VALUES (1, 6)
// INSERT INTO user_role (user_id, role_id) VALUES (1, 7)
// INSERT INTO user_role (user_id, role_id) VALUES (1, 8)
// INSERT INTO user_role (user_id, role_id) VALUES (1, 9)
// INSERT INTO user_role (user_id, role_id) VALUES (1, 10)
```

**After:**
```
$user->roles()->sync([1, 2, 3, 4, 5]);

// Results in:
// INSERT INTO user_role (user_id, role_id) VALUES (1, 1), (1, 2), (1, 3), (1, 4), (1, 5)

$user->roles()->sync([4, 5, 6, 7, 8, 9, 10]);

// Results in:
// DELETE FROM user_role WHERE user_id = 1 AND role_id in (1, 2, 3)
// INSERT INTO user_role (user_id, role_id) VALUES (1, 6), (1, 7), (1, 8), (1, 9), (1, 10)
```

The `attach()` method used by this method to insert new relations already accepts an array with attributes, so this refactor is possible without breaking the framework code in any way.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
@@ -157,7 +157,7 @@ trait InteractsWithPivotTable
     }
 
     /**
-     * Attach all of the records that aren't in the given current records.
+     * Attach all the records that aren't in the given current records.
      *
      * @param  array  $records
      * @param  array  $current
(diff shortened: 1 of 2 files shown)


ankurk91:
This may break some existing 3rd Party packagees which are using model events

sneycampos:
> 

Since pivot tables has not model associated to it, how can model events exists?

ankurk91:
https://github.com/GeneaLabs/laravel-pivot-events

pmcpimentel:
> https://github.com/GeneaLabs/laravel-pivot-events

Since the sync result is the same for the original code & the refactored code, the current sync events for that package remain the same.

taylorotwell:
Rather than risk breaking existing applications, I suggest just doing the queries yourself manually in a database transaction if you encounter a performance bottleneck because of this.

PR #42551, opened 2022-05-28
[9.x] Add counts to route:list command

ash-jc-allen:
Hi again! This is just another small PR to the `route:list` command that I think could come in handy. It adds 2 counts to the CLI output showing the amount of routes that are being displayed. One count is placed at the bottom and one is placed at the top.

Here's an example of how it could look if you don't have a lot of routes and can see both counts:

Screenshot 2022-05-28 at 02 01 30

Here's an example of how it could look if you have a lot of routes and can only see the bottom count:

Screenshot 2022-05-28 at 02 01 15

I think could just be another small UI nicety to make the route list even easier (and quicker) to read and understand.

I couldn't seem to get the tests fully working for this. I have a feeling that it might be something to do with me outputting the same line twice, but that's just a guess. But I have updated the tests to pass with what I've done so far.

If you think this is something that might be worth merging, please let me know if you'd like any changes making 😄

### DIFF
--- a/src/Illuminate/Foundation/Console/RouteListCommand.php
+++ b/src/Illuminate/Foundation/Console/RouteListCommand.php
@@ -405,7 +405,7 @@ class RouteListCommand extends Command
         })
             ->flatten()
             ->filter()
-            ->prepend('')->prepend($routeCount)->prepend('')
+            ->prepend('')
             ->push('')->push($routeCount)->push('')
             ->toArray();
     }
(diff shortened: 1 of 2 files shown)


DarkGhostHunter:
Could this also add the total?

    Showing [4] of 88 routes.

If the total is the same, then this should say:

    Showing all [88] routes.

ash-jc-allen:
I did consider adding the total but couldn't decide whether to include it. If this PR is something that Taylor might want to merge, I'm happy to add it if it's something that would be useful 😄

taylorotwell:
I would personally just show it at the bottom. Mark as ready for review when tests are passing. Thanks!

ash-jc-allen:
I've removed the count from the top now, so it's only at the bottom. I've also updated the tests so they should be passing now too :)

PR #42552, opened 2022-05-28
[9.x] Add `charAt` convenience method to Str and Stringable

boserup:
Sometimes it is necessary to access a specific character in a string - i.e. for parsing data. This PR adds a convenient way of retrieving a single character from a string, which arguably is cleaner than just using `substr`, when the string length is predetermined.

It uses `Str::substr` under the hood.

Usage:
```php
Str::charAt('Tay', 1)
Str::of('Tay')->charAt(1)
str('Tay')->charAt(1)
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -202,6 +202,17 @@ class Str
     }
 
     /**
+     * Retrieve character at specified position in string.
+     *
+     * @param string  $value
+     * @param int  $index
+     * @return string
+     */
+    public static function charAt($value, $index) {
+        return static::substr($value, $index, 1);
+    }
+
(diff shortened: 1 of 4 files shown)


nielsdos:
Why not use ```$string[$index]``` instead of substr in the implementation? :)

rodrigopedra:
@nielsdos under the hood it uses `mb_substr()` which account for unicode characters. 

Offset get operator on strings does not account for unicode characters.

~~~php
$word = 'façade';

assert($word[2] === mb_substr($word, 2, 1));
~~~

nielsdos:
Ah I didn't realise the string methods of Laravel were multibyte safe. My bad, thanks for the comment.

PR #42553, opened 2022-05-28
[9.x] Adds 'finished' method to jobs

yoeriboven:
This PR adds a `finished` method to jobs.

```php
class ImportContactsJob implements ShouldQueue
{
    use Batchable, Dispatchable, InteractsWithQueue, Queueable, SerializesModels;

    public function __construct(protected int $userId)
    {
        //
    }

    public function handle()
    {
        //
    }

    // This method is called when the job has finished
    public function finished()
    {
        ImportContactsBatchUpdated::dispatch($this->userId, $this->batch()->jsonSerialize());
    }
}
```

I'm trying to broadcast the progress of a batch of jobs to my user.
When dispatching the event at the end of the `handle` method, the batch isn't updated yet.

The last job would say:

```
totalJobs : 10
pendingJobs: 1
processedJobs: 9
```

This is because the job was not marked as finished yet.

### Possible solutions
I tried to use this:

```php
Queue::after(function (JobProcessed $event) {
    // $event->connectionName
    // $event->job
    // $event->job->payload()
});
```
But `$event->job` is not an instance of `App\Jobs\ImportContactsJob` but of `Illuminate\Queue\Jobs\Job`.

What you could also do is sending the event in the `then` or `finally` method of `Bus::batch`.

```php
$batch = Bus::batch([
    new ImportContactsJob(auth()->id()),
])->then(function (Batch $batch) {
    ImportContactsBatchUpdated::dispatch(auth()->id(), $batch()->jsonSerialize());
})->name('Import Contacts')->dispatch()
```

I don't like this because you then have to dispatch the event in two places (the job for progress and the `then` method for 100% progress)

### DIFF
--- a/src/Illuminate/Queue/CallQueuedHandler.php
+++ b/src/Illuminate/Queue/CallQueuedHandler.php
@@ -77,6 +77,10 @@ class CallQueuedHandler
             $this->ensureSuccessfulBatchJobIsRecorded($command);
         }
 
+        if (method_exists($command, 'finished')) {
+            $command->finished();
+        }
+
         if (! $job->isDeletedOrReleased()) {
             $job->delete();
         }
--- a/tests/Integration/Queue/CallQueuedHandlerTest.php
+++ b/tests/Integration/Queue/CallQueuedHandlerTest.php
@@ -216,7 +216,8 @@ class TestJobMiddleware
     }
 }
 
-class CallQueuedHandlerWithFinishedMethod {
+class CallQueuedHandlerWithFinishedMethod
+{
     use InteractsWithQueue;
 
     public static $finished = false;
--- a/tests/Integration/Queue/CallQueuedHandlerTest.php
+++ b/tests/Integration/Queue/CallQueuedHandlerTest.php
@@ -42,6 +42,26 @@ class CallQueuedHandlerTest extends TestCase
         $this->assertTrue(CallQueuedHandlerTestJob::$handled);
     }
 
+    public function testJobCanBeFinished()
+    {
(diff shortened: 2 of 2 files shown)


lucasmichot:
What's the difference between `finished()` and `__destruct()`?

yoeriboven:
@lucasmichot I tested this but it seems that `__destruct()` is called multiple times per job. `finished()` will only run once it has been completed.

lucasmichot:
Strange, if you rename `finished()` by `__destruct()` in `CallQueuedHandlerWithFinishedMethod`, `testJobCanBeFinished()` still passes.

yoeriboven:
In that test only a part of the queue system is tested. Maybe in another part the job is also created and destroyed.

yoeriboven:
I just realized it doesn't run `finished` if an exception is thrown in the job. Not sure what we want the expected behavior to be.

Will wait for a reply from the team.

rodrigopedra:
Wouldn't listen for the `JobProcessed` event work? You can also register a closure to dispatched with the Queue façade  `Queue::after()` method.

https://laravel.com/docs/9.x/queues#job-events

Or maybe using a job middleware?

https://laravel.com/docs/9.x/queues#job-middleware

You could see how the `WithoutOverlapping` is implemented which accounts for exceptions.

rodrigopedra:
You can register the listener on the job constructor if you prefer so.

Also I might consider using middleware for this, as they were added to solve the very same problem you describe. 

An eventual middleware can receive parameters through its constructor, and an eventual listener registered on the job's constructor, can also be bound to the job instance's variable.

PR #42554, opened 2022-05-28
[9.x] Add insert helper to add new item in array on any position

mehdirajabi59:
Add insert helper to add new item in array on any position

```
$alphabet = ['a', 'b', 'd'];
$alphabet = Arr::insert($alphabet, 2, 'c');

//output => [a, b, c, d]

 $alphabet = Arr::insert($alphabet, 4, 'e');

//output => [a, b, c, d, e]

```

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -828,10 +828,8 @@ class Arr
      */
     public static function insert(array &$array, int $pos, $value)
     {
-        return array_merge(
-            array_slice($array, 0, $pos),
-            static::wrap($value),
-            array_slice($array, $pos, null, true)
-        );
+        array_splice($array, $pos, 0, static::wrap($value));
+
+        return $array;
     }
 }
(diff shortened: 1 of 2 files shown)


src/Illuminate/Collections/Arr.php
```diff
@@ -817,4 +817,21 @@ public static function wrap($value)
...
+     * @param $value
+     * @return array
+     */
+    public static function insert(array &$array, int $pos, $value)
+    {
+        return array_merge(
+            array_slice($array, 0, $pos),
+            static::wrap($value),
+            array_slice($array, $pos, null, true)
+        );
```
rodrigopedra, line 835: Wouldn't `array_splice()` work here?

~~~php
public static function insert(array &$array, int $pos, $value)
{
    array_splice($array, $pos, 0, static::wrap($value));

    return $array;
}
~~~

https://www.php.net/manual/en/function.array-splice.php

mehdirajabi59: apply it

PR #42555, opened 2022-05-29
[9.x] Conditionable should return HigherOrderWhenProxy only when the args number is exactly 1

leo108:
```
$extraFilter = $condition ? null : function ($query) { $query->where('field', 'value'); };
$query = User::query()->when($extraFilter !== null, $extraFilter);

// When $condition is false, which means $extraFilter is null
// Before: Illuminate\Support\HigherOrderWhenProxy
// After: Illuminate\Database\Eloquent\Builder
dump(get_class($query));
```


BrandonSurowiec:
A simple test should be added.

leo108:
@BrandonSurowiec Done

PR #42556, opened 2022-05-29
[9.x] Adds throwUnless function

mahmoudmohamedramadan:
This PR adds throwUnless function


DarkGhostHunter:
I think it should be better to add `value()` to the condition _before_ adding `throwUnless()`.

PR #42559, opened 2022-05-29
[9.x] Adds `isDirectoryEmpty()` and `isDirectoryNotEmpty()` to Filesystem

DarkGhostHunter:
## What?

Allows to check if a directory has files/directories or not. Even hidden ones.

```php
use Illuminate\Support\Facades\File;

if (File::isDirectoryEmpty('john/photos')) {
    return 'You do not have any photos';
}

if (File::isDirectoryNotEmpty('john/videos', true)) {
    return 'You may have one or more videos, even hidden ones.';
}
```

## BC?

Nope, just adding methods.


johanrosenson:
Your examples are confusing, your returned strings doesn't match what you expect should happen.

src/Illuminate/Filesystem/Filesystem.php
```diff
@@ -500,7 +500,7 @@ public function isDirectoryEmpty($directory, $hidden = false)
     public function isDirectoryNotEmpty($directory, $hidden = false)
     {
         return $this->isDirectory($directory)
```
johanrosenson, line 502: could this not be written as

```
return ! $this->isDirectoryEmpty();
```

DarkGhostHunter: Now that I can think about it more clearly I think it should return an exception if it's not a directory.

johanrosenson: What i meant was that the `isDirectoryNotEmpty` method doesn't need to repeat the logic, it can just use the `isDirectoryEmpty` method and negate the result.

Then you don't need to have almost the same logic in two different methods.

PR #42560, opened 2022-05-30
[9.x] Add ability to resolve `Redirect::home()` via a callback

hailwood:
Right now we have `return Redirect::home()` which returns a redirect to the route registered as 'home'.
It's useful but rather rigid.

This PR adds the ability to register a home resolver to allow end users to resolve where `home` is however they wish.
E.g.

**RouteServiceProvider.php**
```php
use Illuminate\Routing\UrlGenerator;
use Illuminate\Support\Facades\Auth;
use Illuminate\Support\Facades\Redirect;

//...
Redirect::resolveHomeUsing(fn (UrlGenerator $url) => match(Auth::user()?->role) {
    'admin' => $url->route('users.index'),
    'owner' => $url->route('organization.dashboard'),
    default => $url->route('user.dashboard')
});
```

The default resolver is set to the current resolution logic meaning there should be no BC breaks.

### DIFF
--- a/src/Illuminate/Routing/Redirector.php
+++ b/src/Illuminate/Routing/Redirector.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Routing;
 
+use Closure;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Session\Store as SessionStore;
 use Illuminate\Support\Traits\Macroable;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Not sure we even document Redirect::home()... I would just redirect to a specific named route.

hailwood:
@taylorotwell do you mean redirect to the named route and then have the route do a further redirect?

PR #42563, opened 2022-05-30
[9.x] Improve file attachment for mail and notifications

timacdonald:
## Purpose

This PR aims to bring the concept of an "Attachable" for re-use of attaching files to a mail or other notification-like resources. It also standardises the attachable features across:

- `Illuminate/Mail/Mailable`
- `Illuminate/Mail/Message`
- `Illuminate/Notifications/Messages/MailMessage`

which is currently inconsistent, with `Mailable` supporting attaching from storage, while the others do not.

## Why this feature is useful

Often in applications you have a model or POPO that represents a "file" that you attach to a notification. The best of example of this is the ever classic example of an `Invoice`. You may have in invoice model in your system that has an associated PDF that you attach to notifications.

There may be several emails that you want the invoice attached to, and what's more is you might have some extra logic involved in retrieving the invoice before attaching to an email.

Currently we can attach files to an email in the following ways...

```php
class InvoiceCreated extends Mailable
{
    public function __construct(private Invoice $invoice)
    {
        //
    }

    public function build()
    {
        // From a file on the local filesystem...

        $this->attach($this->invoice->path);

        // As above, but with a nice filename and specifying the mime type...

        $this->attach($this->invoice->path, [
            'as' => "Acme Invoice {$this->invoice->number}.pdf",
            'mime' => 'application/pdf',
        ]);

        // From in-memory data...

        $this->attachData(
            $this->invoice->content,
            "Acme Invoice {$this->invoice->number}.pdf",
            ['mime' => 'application/pdf']
        );

        // Getting the contents on the fly...

        $data = Http::withHeaders(['Accept' => 'application/pdf'])
            ->get("https://api.xero.com/api.xro/2.0/Invoices/{$this->invoice->xero_id}")
            ->throw()
            ->body();

        $this->attachData(
            $data,
            "Acme Invoice {$this->invoice->number}.pdf",
            ['mime' => 'application/pdf']
        );


        /*
         * The following are only available on the `Mailable` class, making the API inconsistent...
         */

        // From the default storage disk...

        $this->attachFromStorage(
            $this->invoice->path,
            "Acme Invoice {$this->invoice->number}.pdf",
            ['mime' => 'application/pdf']
        );

        // From a specific storage disk...

        $this->attachFromStorageDisk(
            'backblaze',
            $this->invoice->file_path,
            "Acme Invoice {$this->invoice->number}.pdf",
            ['mime' => 'application/pdf']
        );
    }
}

```

When you then implement another email that sends this same file, you may then find yourself duplicating this code or reaching for the docs to remember how the APIs work, etc.

This PR aims to improve experience of attaching files to an email or other notification system.

## Usage

There is now an new contract called `Illumiante\Contracts\Mail\Attachable` that is meant for your model and a new `Illuminate\Mail\Attachment` class that you return from the contract method.

You can specify how you want the attachment to be attached to you mail / notification...

```php
class Invoice extends Model implements Attachable
{
    public function toMailAttachment()
    {
        // From a file on the local filesystem...

        return Attachment::fromPath($this->path);

        // As above, but with a nice filename and specifying the mime type...

        return Attachment::fromPath($this->path)
            ->as("Acme Invoice {$this->number}.pdf")
            ->withMime('application/pdf');

        // From in-memory data...

        return Attachment::fromData(fn () => $this->content)
            ->as("Acme Invoice {$this->number}.pdf")
            ->withMime('application/pdf');

        // Getting the contents on the fly...

        return Attachment::fromData(fn () => Http::withHeaders(['Accept' => 'application/pdf'])
            ->get("https://api.xero.com/api.xro/2.0/Invoices/{$this->xero_id}")
            ->throw()
            ->body()
        )->as("Acme Invoice {$this->invoice->number}.pdf")->withMime('application/pdf');
 

        /*
         * The following are now available on the `Mailable` and mail messages, both mail and notifications,
         * making the API consistent 🎉
         */


        // From the default storage disk...

        return Attachment::fromStorage($this->path)
            ->as("Acme Invoice {$this->number}.pdf")
            ->withMime('application/pdf');

        // From a specific storage disk...

        return Attachment::fromStorageDisk('backblaze', $this->path)
            ->as("Acme Invoice {$this->number}.pdf")
            ->withMime('application/pdf');
    }
}
```

Then to attach the invoice to a built in mailable resource the following API is used...

```php
class InvoiceCreated extends Mailable
{
    public function __construct(private Invoice $invoice)
    {
        //
    }

    public function build()
    {
        $this->attach($this->invoice);
    }
}
```

It is also possible to set the filename or mine from within the mailable itself, if that is wanted...

```php
class InvoiceCreated extends Mailable
{
    public function __construct(private Invoice $invoice)
    {
        //
    }

    public function build()
    {
        $this->attach(
            $this->invoice->toMailAttachment()->as('Pay Me.pdf')
        );
    }
}
```

## Mutiple attachments

Sometimes your model may several attachments. No troubles...

```php
class Invoice extends Model implements Attachable
{
    public function overdueAttachment()
    {
        return Attachment::fromData(/* ... */);
    }

    public function renewalAttachment()
    {
        return Attachment::fromData(/* ... */);
    }

    public function toMailAttachment()
    {
        return Attachment::fromData(/* ... */);
    }
}


// in the build function of your mailable....


$this->attach($this->invoice);

$this->attach($this->invoice->renewalAttachment());

$this->attach($this->invoice->overdueAttachment());

```

it is also possible to just use the attachment class right inline while attaching to a `Mail\Mailable`, `Mail\Message`, and `Notification\Messages\MailMessage`, with the latter two now being able to utilise the storage machanism.

```php
// in the build function…

$this->attach(Attachment::fromStorageDisk('s3', 'logo.png'));
```

## General API for 3rd parties

This new class has two ways of attaching data. By referencing a path or by supplying a Cloure that resolves a string of data. All APIs are just wrappers around these two features.

Third party packages that want to integrate with this class may specify how they want to handle an attachment. They must specify both a handler for a path based resource and a data based resource.

```php
class NotificationFrom3rdParty
{
    public function attach(Attachment $attachment): void
    {
        $attachment->attachWith(
            fn (string $path) => $this->attachPath($path, $attachment->as, $attachment->mime),
            fn (Closure $data) => $this->attachData($data, $attachment->as, $attachment->mime),
        );
    }

    public function attachData(Closure $data, string $filename, string $mimeType): void
    {
        //
    }

    public function attachPath(string $path, string $filename, string $mimeType): void
    {
        //
    }
}

$notificationFrom3rdPartyInstance->attach($invoiceModel);

```

You can see that the `$data` is not eagerly resolved, allowing the integration to decide _when_ they want to resolve the data - which is useful for delaying the resolution from storage etc, which is how the current mail system works.

### Extending the dev facing API

The attachment class is also Macroable, so it is possible to create nice named constructors from 3rd party libraries...

```php
// Service provider...

Attachment::macro('fromXeroInvoice', function ($id) {
    return Attachment::fromData(fn () => Http::withHeaders(['Accept' => 'application/pdf'])
        ->get("https://api.xero.com/api.xro/2.0/Invoices/{$id}")
        ->throw()
        ->body());
});

// Attachable...

public function toMailAttachment()
{
    return Attachment::fromXeroInvoice($this->xero_id)
        ->as("Acme Invoice {$this->number}.pdf")
        ->withMime('application/pdf');;
}
```

Or even just a more generic way of retrieving from a public API...

```php
Attachment::macro('fromUrl', function ($url) {
    return Attachment::fromData(fn () => Http::get($url)->throw()->body();
});
```

## Notes

- I've put the new contract and the new Attachment class in the `Mail` namespace, but as shown above it is kinda just generic "Notification". I'm happy to move it to the `Notifications` namespace if we feel that is a better fit.
- This new implementation does not dedupe "fromStorage" attachments in the same way that the current Mailable implementation does, as it has to resolve the Data the know how to dedupe.
- The current mail implementation does eagerly load the data during the "build" function call.
- It would be possible to consolidate both the `fromPath` and `fromData` APIs so that everything at the end of the day results in a `fromData`. This would simplify the integration as well, as you no longer have to handle both the "Path" and "Data" based attachments. The reason I didn't opt for a simple `file_get_contents` is that the Symfony mailer is doing a heap under the hood that I didn't want to have to replicate, so we continue to have two APIs - just like we currently do.
- This PR does not support inline attachments, as I wanted to make sure it was wanted before doing the work. Will do a follow up PR to ensure they work with inline attachments if this is merged / wanted. I'll also bring the embed feature to `Mailable` and `Notifications/Messages/MailMessage` which do not currently support inline embeds as a first class feature.

### DIFF
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -4,6 +4,7 @@ namespace Illuminate\Mail;
 
 use Illuminate\Container\Container;
 use Illuminate\Contracts\Filesystem\Factory as FilesystemFactory;
+use Illuminate\Contracts\Mail\Attachable;
 use Illuminate\Contracts\Mail\Factory as MailFactory;
 use Illuminate\Contracts\Mail\Mailable as MailableContract;
 use Illuminate\Contracts\Queue\Factory as Queue;
@@ -867,12 +868,20 @@ class Mailable implements MailableContract, Renderable
     /**
      * Attach a file to the message.
      *
-     * @param  string  $file
+     * @param  string|\Illuminate\Contracts\Mail\Attachable|\Illuminate\Mail\Attachment  $file
      * @param  array  $options
      * @return $this
      */
     public function attach($file, array $options = [])
     {
+        if ($file instanceof Attachable) {
+            $file = $file->toMailAttachment();
+        }
+
+        if ($file instanceof Attachment) {
+            return $file->attachTo($this);
+        }
+
         $this->attachments = collect($this->attachments)
                     ->push(compact('file', 'options'))
                     ->unique('file')
--- a/src/Illuminate/Mail/Message.php
+++ b/src/Illuminate/Mail/Message.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Mail;
 
+use Illuminate\Contracts\Mail\Attachable;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\ForwardsCalls;
 use Symfony\Component\Mime\Address;
@@ -23,6 +24,8 @@ class Message
 
     /**
      * CIDs of files embedded in the message.
+     *
+     * @deprecated Will be removed in a future Laravel version.
      *
      * @var array
      */
@@ -290,12 +293,20 @@ class Message
     /**
      * Attach a file to the message.
      *
-     * @param  string  $file
+     * @param  string|\Illuminate\Contracts\Mail\Attachable|\Illuminate\Mail\Attachment  $file
      * @param  array  $options
      * @return $this
      */
     public function attach($file, array $options = [])
     {
+        if ($file instanceof Attachable) {
+            $file = $file->toMailAttachment();
+        }
+
+        if ($file instanceof Attachment) {
+            return $file->attachTo($this);
+        }
+
         $this->message->attachFromPath($file, $options['as'] ?? null, $options['mime'] ?? null);
 
         return $this;
@@ -319,11 +330,32 @@ class Message
     /**
      * Embed a file in the message and get the CID.
      *
-     * @param  string  $file
+     * @param  string|\Illuminate\Contracts\Mail\Attachable|\Illuminate\Mail\Attachment  $file
      * @return string
      */
     public function embed($file)
     {
+        if ($file instanceof Attachable) {
+            $file = $file->toMailAttachment();
+        }
+
+        if ($file instanceof Attachment) {
+            return $file->attachWith(
+                function ($path) use ($file) {
+                    $cid = $file->as ?? Str::random();
+
+                    $this->message->embedFromPath($path, $cid, $file->mime);
+
+                    return "cid:{$cid}";
+                },
+                function ($data) use ($file) {
+                    $this->message->embed($data(), $file->as, $file->mime);
+
+                    return "cid:{$file->as}";
+                }
+            );
+        }
+
         $cid = Str::random(10);
 
         $this->message->embedFromPath($file, $cid);
--- a/src/Illuminate/Contracts/Mail/Attachable.php
+++ b/src/Illuminate/Contracts/Mail/Attachable.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Illuminate\Contracts\Mail;
+
+interface Attachable
+{
+    /**
+     * Get an attachment instance for this entity.
+     *
+     * @return \Illuminate\Mail\Attachment
+     */
+    public function toMailAttachment();
+}
--- a/src/Illuminate/Mail/Attachment.php
+++ b/src/Illuminate/Mail/Attachment.php
@@ -5,7 +5,6 @@ namespace Illuminate\Mail;
 use Closure;
 use Illuminate\Container\Container;
 use Illuminate\Contracts\Filesystem\Factory as FilesystemFactory;
-use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
 
 class Attachment
--- a/src/Illuminate/Mail/Attachment.php
+++ b/src/Illuminate/Mail/Attachment.php
@@ -1,0 +1,156 @@
+<?php
+
+namespace Illuminate\Mail;
+
+use Closure;
+use Illuminate\Container\Container;
+use Illuminate\Contracts\Filesystem\Factory as FilesystemFactory;
+use Illuminate\Support\Str;
+use Illuminate\Support\Traits\Macroable;
+
+class Attachment
+{
+    use Macroable;
+
+    /**
+     * The attached file's filename.
+     *
+     * @var string|null
+     */
+    public $as;
+
+    /**
+     * The attached file's mime type.
+     *
+     * @var string|null
+     */
+    public $mime;
+
+    /**
+     * A callback that attaches the attachment to the mail message.
+     *
+     * @var \Closure
+     */
+    protected $resolver;
+
+    /**
+     * Create a mail attachment.
+     *
+     * @param  \Closure  $resolver
+     * @return void
+     */
+    private function __construct(Closure $resolver)
+    {
+        $this->resolver = $resolver;
+    }
+
+    /**
+     * Create a mail attachment from a path.
+     *
+     * @param  string  $path
+     * @return static
+     */
+    public static function fromPath($path)
+    {
+        return new static(fn ($attachment, $pathStragegy) => $pathStragegy($path, $attachment));
+    }
+
+    /**
+     * Create a mail attachment from in-memory data.
+     *
+     * @param  \Closure  $data
+     * @param  string  $name
+     * @return static
+     */
+    public static function fromData(Closure $data, $name)
+    {
+        return (new static(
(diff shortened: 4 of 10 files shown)


timacdonald:
Looks like there are some tests failing on `prefer-lowest` as it isn't detecting the fallback mime type  in the same way. Marking as draft until I can look at that tomorrow.

driesvints (approved):
Looks like a great addition to the mail component 👍 

I think the Attachment class can definitely live in the Mail namespace.

timacdonald:
Naturally in the middle of the night my brain decided there was more improvements to be made to allow use by 3rd parties. Will send through an update tomorrow to improve the way third party packages can use their existing APIs, and not have to implement new ones, with this class

timacdonald:
This one is now ready to roll. I've tested this against a real mail server as well to ensure this all actually works IRL as expected.

src/Illuminate/Mail/Message.php
```diff
@@ -24,6 +25,8 @@ class Message
     /**
      * CIDs of files embedded in the message.
      *
+     * @deprecated Will be removed in a future Laravel version.
+     *
      * @var array
      */
     protected $embeddedFiles = [];
```
timacdonald, line 32: This is not used anywhere, so I deprecated it. Happy for it to be reverted if it has a purpose.

timacdonald: // cc @driesvints

driesvints: Sounds good

taylorotwell:
@timacdonald how does this embed stuff actually work?

Normally I think of embeds as being in the message itself?

CleanShot 2022-06-02 at 10 21 37@2x

src/Illuminate/Mail/Mailable.php
```diff
@@ -438,6 +453,20 @@ protected function buildAttachments($message)
             );
         }
 
+        foreach ($this->embeds as $embed) {
+            $message->embed(Attachment::fromPath($embed['file'])
+                ->as($embed['options']['as'])
+                ->withMime($embed['options']['mime'] ?? null));
+        }
```
timacdonald, line 460: You will notice this re-creates the `Attachment` class. This is to work around the inconsistent `Mail\Message` embed API.

src/Illuminate/Mail/Mailable.php
```diff
@@ -936,6 +973,60 @@ public function attachData($data, $name, array $options = [])
...
+                'data' => $data,
+                'name' => $name,
+                'options' => $options,
+            ])
+            ->unique(fn ($file) => $file['name'].$file['data'])
+            ->all();
+
+        return "cid:{$name}";
+    }
+
```
timacdonald, line 1029: I've given these functions the same API as the `attach` function for consistency, however `$name` is option in the data function.

This does mean that it is inconsistent with the `Mail\Message` embed API, but I think that is okay as that class is kind mostly internal and I think it should be made consistent with these APIs in a future major version.

src/Illuminate/Notifications/Channels/MailChannel.php
```diff
@@ -246,6 +247,20 @@ protected function addAttachments($mailMessage, $message)
...
+                ->withMime($embed['options']['mime'] ?? null));
+        }
+
+        foreach ($message->rawEmbeds as $embed) {
+            $mailMessage->embedData(
+                $embed['data'],
+                $embed['name'],
+                $embed['options']['mime'] ?? null
+            );
+        }
```
timacdonald, line 263: See: https://github.com/laravel/framework/pull/42563/files?show-viewed-files=true&file-filters%5B%5D=#r888534732

src/Illuminate/Mail/Message.php
```diff
@@ -319,11 +330,32 @@ public function attachData($data, $name, array $options = [])
...
     public function embed($file)
     {
+        if ($file instanceof Attachable) {
+            $file = $file->toMailAttachment();
+        }
+
+        if ($file instanceof Attachment) {
+            return $file->attachWith(
+                function ($path) use ($file) {
+                    $cid = $file->as ?? Str::random();
```
driesvints, line 345: Maybe use `random(10)` like the below.

```suggestion
                    $cid = $file->as ?? Str::random(10);
```

Is there a huge reason why you prefer `$file->as` and not just always pick `Str::random(10)`? Could simplify things a bit.

timacdonald: This is the name that is given to the file / image / document in the email. If you drag the image / document from the email onto your own system you get the nicer name specified by the developer, rather than an extension-less file...

Felt it was nicer and more consistent with attachments as well. I think we should allow the `embed` function to accept a name - but as that would be breaking I didn't want to make that change, of course.

Screen Shot 2022-06-03 at 5 26 00 pm

timacdonald: but not a hill I'm gonna die on. Happy for it to just always give a random name if we prefer 👍

timacdonald: oh, and for "web based" mail clients, they often show a list of attachments - inline or not, and it will show the name there as well.

So mostly just that it isn't completely hidden and the name is sometimes exposed to people.

driesvints: Definitely makes sense to use the file name I think. All good 👍

Issue #42493, opened 2022-05-24
BelongsTo relation with nullable relations


pionl:
- Laravel Version: 9.13.0
- PHP Version: 8.1.2

### Description:

- While using belongsTo relations (`with/load`) there can be situations when  there are no relations (nullable relations) to be loaded and query is "invalid". -> ```select `id`, `name`, `public_name` from `options` where 0 = 1 ```. 
- This similar issue can be potentially found on other relations
- This bug is not noticeable if the results contains at least 1 value in the relation (if you have pagination, some pages can contain this error).

### Steps To Reproduce:

1. Create a table with optional belongsTo relations
2. Build a query to get the results (all results should contain null)
3. Here a screenshot of a real queries 
<img width="2025" alt="Snímek obrazovky 2022-05-24 v 14 22 56" src="https://user-images.githubusercontent.com/1878831/170034097-9867c867-bf9c-4100-bff0-6ecee1426097.png">

#### Fix

1. In `Illuminate\Database\Eloquent\Relations\Relation` add property `protected bool $hasValidConstrains = true;`
2. Change the implementation of `Illuminate\Database\Eloquent\Relations\Relation::getEager `
```php
 public function getEager()
    {
        if ($this->hasValidConstrains) {
            return $this->get();
        }
        return new Collection();
    }
```
3. Change the implementation of `Illuminate\Database\Eloquent\Relations\BelongsTo::addEagerConstraints`
```php
public function addEagerConstraints(array $models)
    {
        // We'll grab the primary key name of the related models since it could be set to
        // a non-standard name and not "id". We will then construct the constraint for
        // our eagerly loading query so it returns the proper models from execution.
        $key = $this->related->getTable().'.'.$this->ownerKey;

        $whereIn = $this->whereInMethod($this->related, $this->ownerKey);

        $eagerModelKeys = $this->getEagerModelKeys($models);

        if ($eagerModelKeys === []) {
            $this->hasValidConstrains = false;
            return;
        }

        $this->query->{$whereIn}($key, $eagerModelKeys);
    }
```

**Result**

<img width="1650" alt="Snímek obrazovky 2022-05-24 v 14 26 39" src="https://user-images.githubusercontent.com/1878831/170034353-efb760fa-95d9-47a6-873f-f2735f05a72c.png">

I can make PR if you consider this a bug (which I believe it is) and don't have the time to make the fix.

Thank you for your time.
Martin.

driesvints:
[...]

PR #42564, opened 2022-05-30
[9.x] Prevent triggering belongs to query if there are no relations to load

pionl:
References issue #42493

- At this moment add the protection only for BelongsTo relation (can be added to `Relation` if more use cases are found)
- Prevents a query with empty whereIn value which reduces queries to DB.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/BelongsTo.php
+++ b/src/Illuminate/Database/Eloquent/Relations/BelongsTo.php
@@ -15,6 +15,8 @@ class BelongsTo extends Relation
         InteractsWithDictionary,
         SupportsDefaultModels;
 
+    protected bool $hasValidConstrains = true;
+
     /**
      * The child model instance of the relation.
      *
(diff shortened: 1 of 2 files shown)


taylorotwell:

pionl:
Hi @taylorotwell, I truly believe this is a bug that bothers me from L5.4. Why would framework want to trigger a query on database that will return "nothing". This is a quite simple fix.

Of course I can make my "own" belongs to relation and use it, but I believe that hurts the quality of the framework that all the people like.

Here is a copy of explanation from issue. Thank you for your time.




- Laravel Version: 9.13.0
- PHP Version: 8.1.2

### Description:

- While using belongsTo relations (`with/load`) there can be situations when  there are no relations (nullable relations) to be loaded and query is "invalid". -> ```select `id`, `name`, `public_name` from `options` where 0 = 1 ```. 
- This similar issue can be potentially found on other relations
- This bug is not noticeable if the results contains at least 1 value in the relation (if you have pagination, some pages can contain this error).

### Steps To Reproduce:

1. Create a table with optional belongsTo relations
2. Build a query to get the results (all results should contain null)
3. Here a screenshot of a real queries 
Snímek obrazovky 2022-05-24 v 14 22 56

PR #42571, opened 2022-05-30
[9.x] Map integer parameter to parameter name when resolving binding field

### LINKED ISSUE #42541: Scoped route binding in apiResource routes no longer working
- Laravel Version: 9.14.1
- PHP Version: 8.1.2
- Database Driver & Version:

### Description:

I've recently updated Laravel from 9.13.0 to 9.14.1 and I have a series of nested apiResource routes defined like so:
`$router->apiResource('/product/{product}/image', ImageController::class)->scoped();`

I've found that when I request an image that doesn't belong to the product id specified in the URL, I get a 404 like I'd expect in v9.13.0, but in v9.14.1 the request falls through to the controller incorrectly... I may be doing something wrong but it has been working until today.

### Steps To Reproduce:

Define a scoped API resource like the one above.
Request a URL where the child model is not owned by the parent model.

Expected: The app throws a 404
Actual: The app calls the controller

I've made a new laravel project to demonstrate: https://github.com/joe-pritchard/scoped-resource-bug

There's a test in there at `Tests\Feature\Models\ChildTest` which shows the issue. The test fails, but it will pass if you downgrade the `laravel/framework` dependency to 9.13.0

ksassnowski:
Fixes #42541

First of all, I'm really sorry about all the issues the initial PR has caused 😞

## Summary

This PR reverts most of the changes of the original PR (#42425) while still fixing the bug described in the PR. The only method that was changed was the `bindingFieldFor` method of the `Route` class. All changes to `ImplicitRouteBinding` and `RouteUri` were reverted.

To fix the bug described in the original PR, we now resolve the parameter index to the parameter name by looking it up in the route's `parameterNames` array. 

```php
public function bindingFieldFor($parameter)
{
    if (is_int($parameter))  {
        $parameter = $this->parameterNames()[$parameter];
    }

    return $this->bindingFields[$parameter] ?? null;
}
```

This means that we no longer have to change the structure of the `bindingFields` array itself like in the original PR. The bug described in the related issue (#42541) was due to the fact that the `ResourceRouteRegistrar` _also_ filled the route's binding fields with `null` values, just like my change in the original PR. The problem with that is that both of these implementations used `null` to signal the exact opposite of each other. My PR used `null` to mean "don't scope this parameter", whereas the `ResourceRouteRegistrar` used it to mean "scope this parameter but default to the model's route key name".

I left in all the additional tests I added since all the problems that got reported due to my original PR were not covered by any existing tests.

### DIFF
--- a/src/Illuminate/Routing/ImplicitRouteBinding.php
+++ b/src/Illuminate/Routing/ImplicitRouteBinding.php
@@ -46,7 +46,7 @@ class ImplicitRouteBinding
                         ? 'resolveSoftDeletableRouteBinding'
                         : 'resolveRouteBinding';
 
-            if ($parent instanceof UrlRoutable && ($route->enforcesScopedBindings() || $route->bindingFieldFor($parameterName) !== null)) {
+            if ($parent instanceof UrlRoutable && ($route->enforcesScopedBindings() || array_key_exists($parameterName, $route->bindingFields()))) {
                 $childRouteBindingMethod = $route->allowsTrashedBindings() && in_array(SoftDeletes::class, class_uses_recursive($instance))
                             ? 'resolveSoftDeletableChildRouteBinding'
                             : 'resolveChildRouteBinding';
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -535,9 +535,11 @@ class Route
      */
     public function bindingFieldFor($parameter)
     {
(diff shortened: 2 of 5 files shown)


mortenscheel:
FYI I think this PR introduced a breaking change in a rather exotic edge case. I posted a bug report (https://github.com/laravel/framework/issues/42605), but I closed it myself when I realized that it only happens if you ask the route() helper to turn models into query params. Which is obviously a mistake, but up until v9.14 it produced valid URL's. Now it produces 500 errors.

PR #42577, opened 2022-05-30
[9.x] Predis v2.0

driesvints:
Predis v2 is released now: https://github.com/predis/predis/releases/tag/v2.0.0

It's only a maintenance release so nothing should break.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -92,7 +92,7 @@
         "pda/pheanstalk": "^4.0",
         "phpstan/phpstan": "^1.4.7",
         "phpunit/phpunit": "^9.5.8",
-        "predis/predis": "^1.1.9",
+        "predis/predis": "^1.1.9|^2.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -92,7 +92,7 @@
         "pda/pheanstalk": "^4.0",
         "phpstan/phpstan": "^1.4.7",
         "phpunit/phpunit": "^9.5.8",
-        "predis/predis": "^1.1.9",
+        "predis/predis": "^2.0",
```
GrahamCampbell, line 95: `^1.1.9|^2.0`, so prefer lowest gets 1.x?

GrahamCampbell (changes requested):

composer.json
driesvints: I'm just experimenting with this PR to force 2.0. I'll revert that.

PR #42578, opened 2022-05-30
[9.x] Allow underline in numeric validation values

imanghafoori1:
PHP 7.4 allows putting the underline character in integers for more readability, like:
"1_500_000" instead of: "1500000"
This PR allows putting underline in numbers of validation rules: ```Numeric|Between:500_000,1_000_000```

- It also adds some important missing tests.

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -407,7 +407,7 @@ trait ValidatesAttributes
 
         $size = $this->getSize($attribute, $value);
 
-        return $size >= $parameters[0] && $size <= $parameters[1];
+        return $size >= str_replace('_', '', (string) $parameters[0]) && $size <= str_replace('_', '', (string) $parameters[1]);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Illuminate/Validation/Concerns/ValidatesAttributes.php
```diff
@@ -407,7 +407,7 @@ public function validateBetween($attribute, $value, $parameters)
 
         $size = $this->getSize($attribute, $value);
 
-        return $size >= $parameters[0] && $size <= $parameters[1];
+        return $size >= (int) str_replace('_', '', (string) $parameters[0]) && $size <= (int) str_replace('_', '', (string) $parameters[1]);
```
lucasmichot, line 410: They should be casted to float, as the boundaries might not be integers.

imanghafoori1: Is it better to remove the (int) cast and leave it to PHP?

lucasmichot: I would add extra tests to cover this case and adjust your code accordingly

imanghafoori1: I force pushed some more tests to cover the floats.

PR #42581, opened 2022-05-31
[9.x] Fix queueable notification's ID overwritten

jay-youngn:
When the notification ID has been assigned, It should not be overwritten. Even in the queue.

The case for database notification:
I've defined a special unique ID, but it's overwritten here.

If this option is not desirable, the `Str::uuid()` should also be replace to `Str::orderedUuid()` for keep order of database index.

### DIFF
--- a/src/Illuminate/Notifications/NotificationSender.php
+++ b/src/Illuminate/Notifications/NotificationSender.php
@@ -191,7 +191,9 @@ class NotificationSender
             foreach ((array) $original->via($notifiable) as $channel) {
                 $notification = clone $original;
 
-                $notification->id = $notificationId;
+                if (! $notification->id) {
+                    $notification->id = $notificationId;
+                }
 
                 if (! is_null($this->locale)) {
                     $notification->locale = $this->locale;


ginnerpeace:
The case is via database notification

taylorotwell:
How are you defining a special unique ID that is being overwritten?

ginnerpeace:
> How are you defining a special unique ID that is being overwritten?

@taylorotwell  Thanks for your attention

My cases like this:

Part 1:
I have a function to generate unique IDs in ascending order.
```php
<?php

namespace App\Supports;

use Str;

class UniqueID
{
    /**
     * Generates a string UID.
     */
    public static function generateString(): string
    {
        return str_replace('-', '', (string) Str::orderedUuid());
    }
}

```

Part 2:
And I assigned notification ID when an instance created.
(Because I want the primary key to be ordered in table `notifications`)
```php
<?php

namespace App\Notifications;

use App\Enums\EntityAlias;
use App\Enums\InteractType;
use App\Models\User;
use App\Supports\UniqueID;
use Illuminate\Bus\Queueable;
use Illuminate\Notifications\Messages\MailMessage;
use Illuminate\Notifications\Notification;

class UserInteractsNotification extends Notification
{
    use Queueable;

    private EntityAlias $topic = EntityAlias::User;

    /**
     * Create a new notification instance.
     *
     * @return void
     */
    public function __construct(
        private InteractType $type,
        private User $fromUser,
    ) {
        $this->id = UniqueID::generateString();
    }

    /**
     * Get the notification's delivery channels.
     *
     * @param  mixed  $notifiable
     * @return array
     */
    {
        return ['database'];
    }

    // ...
}

```

Part 3:
If the `UserInteractsNotification` implements `ShouldQueue`, the earlier code in `Illuminate\Notification\NotificationSender` will always overwrite notification ID.

The important thing is... When queue driver is `sync`, overwrite will not happen, this is because there is such a piece of code in method `NotificationSender::sendToNotifiable()`:
https://github.com/laravel/framework/blob/3b8d805319f3d41c870241edd76865a8bf7b4231/src/Illuminate/Notifications/NotificationSender.php#L138-L146
When notification should queue, the method `NotificationSender::queueNotification()` will be called, and it does not have a determination of whether the ID exists or not.

So I added same code from L194 to L196
https://github.com/laravel/framework/blob/3b8d805319f3d41c870241edd76865a8bf7b4231/src/Illuminate/Notifications/NotificationSender.php#L188-L197

PR #42583, opened 2022-05-31
[9.x] Adds `addArrayOfWheres` function

mahmoudmohamedramadan:
This function allows us to pass an array of keys and values where the key will be the column name and the value will be the value of this column.

**Before adding this function**

```PHP
Rule::unique('table_name')
    ->where('column_name', 'the_value')
    ->where('another_column_name', 'the_value')
```

**After adding this function**

```PHP
Rule::unique('table_name')
    ->where([
        ['column_name', 'the_value'],
        ['another_column_name', 'the_value']
    ])
```

### DIFF
--- a/src/Illuminate/Validation/Rules/DatabaseRule.php
+++ b/src/Illuminate/Validation/Rules/DatabaseRule.php
@@ -76,16 +76,35 @@ trait DatabaseRule
 
         return $table;
     }
+    
+    /**
+     * Add an array of where clauses to the query.
+     *
+     * @param  array  $column
+     * @return $this
+     */
+    protected function addArrayOfWheres($column)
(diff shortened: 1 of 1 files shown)


johanrosenson:
You should probably make it so it behaves like the `where` on the `QueryBuilder`, there you it an associative array of column and values.

usage example:

```
->where([
    'column_name' => 'the_value',
    'another_column_name' => 'the_value',
])
```

it also supports your syntax but with two or three arguments where the second argument is an optional operator, if left out it defaults to `=`.

```
->where([
    ['column_name', '>', 'the_value'],
    ['another_column_name', 'the_value']
])
```

taylorotwell:

PR #42587, opened 2022-05-31
[10.x] Remove deprecated dates property

driesvints:
[It's been over two years since the dates property became deprecated](https://github.com/laravel/framework/commit/c0e0bb52a8e98169b4f5d2415890c7e80fdb04e5) and even longer since the introduction of casts. I feel it's time to remove this property so people can solely rely on the model casts for any datetime attributes.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -106,15 +106,6 @@ trait HasAttributes
     ];
 
     /**
-     * The attributes that should be mutated to dates.
-     *
-     * @deprecated Use the "casts" property
-     *
-     * @var array
-     */
-    protected $dates = [];
-
-    /**
      * The storage format of the model's date columns.
      *
      * @var string
@@ -1366,16 +1357,10 @@ trait HasAttributes
      */
     public function getDates()
     {
-        if (! $this->usesTimestamps()) {
-            return $this->dates;
-        }
-
-        $defaults = [
+        return $this->usesTimestamps() ? [
             $this->getCreatedAtColumn(),
             $this->getUpdatedAtColumn(),
-        ];
-
-        return array_unique(array_merge($this->dates, $defaults));
+        ] : [];
     }
 
     /**


ghoshriju33:
I know this is a good thing. Sorry if I posted in the wrong place, but the dates property is very useful. Without the `$dates` property, I have to use `$casts` and also define an accessor to convert the value to a carbon instance. While `$casts` is the recommended way, it only works with serialized models. To be honest, the `$dates` property is really a timesaver.

driesvints:
@ghoshriju33 what do you mean? why do you need to define an accessor?

nanaya:
Sorry for posting on an old PR, but was this deprecation announced anywhere? I checked upgrade guide and release notes for 7.x and later and none seem to mention it was being deprecated - only it being suddenly removed in 10.x.

driesvints:
@nanaya yes here: https://laravel.com/docs/10.x/upgrade#model-dates-property

driesvints:
The `@deprecation` annotation has been in place for some time now.

nanaya:
Annotation that isn't visible anywhere isn't really useful. It also doesn't help the update guide is only given on removal instead of deprecation (although for this one it's pretty simple).

decadence:
Agreed with @nanaya. Deprecations should be noted in the next Upgrade Guides: X-thing is deprecated and will be deleted in future releases, use Y-thing instead. Not everybody monitors GitHub Pull Requests and laravel/framework source code. Unless it's hard to maintain.

cwilby:
With `$dates` gone, is the cleanest approach to JsonResource serialization to now add `->format('Y-m-d')` to the end of each field?

Before:
```php
<?php

class Matter extends Model
{
    protected $dates = [
        'bar_date' => 'Y-m-d'
    ];
}

class MatterResource extends JsonResource
{
    {
        dd((string) $this->bar_date); // 2023-02-26

        return [
            'id' => $this->id,
            'bar_date' => $this->bar_date
        ];
    }
}
```

Now

```php
<?php

class Matter extends Model
{
    protected $casts = [
        'bar_date' => 'date:Y-m-d'
    ];
}

class MatterResource extends JsonResource
{
    {
        dd((string) $this->bar_date); // 2023-02-26T00:00:00Z

        return [
            'id' => $this->id,
            'bar_date' => $this->bar_date->format('Y-m-d') // 👈 must be added everywhere `$dates` was used.
        ];
    }
}
```

I thought about writing a resource that uses `->toArray()` to leverage `$casts`, but that kind of made me question the value of JsonResources at that point.

```
class MatterResource extends JsonResource
{
    {
        return $this->resource->toArray();
    }
}
```

Anyway - it's not clear, and I thought it may be helpful to show why.

tentus:
This change makes the annotation on `getDates()` a bit misleading: my code depended on that method to get a list of attributes to then apply a user-specified format to, but now that code skips over everything except the created & updated timestamps.

For anyone else looking for a quick-and-ugly hack to get back to the previous behavior, this worked for me as a short term solution until i can remove references to getDates entirely:
```
    public function getDates()
    {
        return array_keys(
            array_flip(array_filter(parent::getDates())) + array_filter($this->getCasts(), function ($v) {
                return $v == 'datetime';
            })
        );
    }
```

PR #42591, opened 2022-05-31
[10.x] Remove deprecated dispatchNow functionality

driesvints:
This will remove the deprecated functions and methods around dispatchNow. The internals will remain to not make the upgrade path too hard but in general we want to push people more to use `dispatchSync` which is the only supported way to dispatching immediately.


amitshah-sgcm:
I just realized, unlike dispatchNow we can not mention the onQueue in dispatchSync. this is mandatory in my case to pass dynamic queue name. so dispatchNow is an essential function for me. Please update dispatchSync to accept queue name.

PR #42597, opened 2022-05-31
[9.x] Added Eloquent withWhereHas method

eusonlito:
There are a lot of times that we need to load a relationship using `with` with the same conditions that we use for `whereHas`.

This method simplifies this selection by avoiding repeating the code used both to filter with `whereHas` and then to select that same record using `with`. There is less code, it is more readable and easier to maintain.

## Before

```php
CollectionModel::whereHas('products', function ($query) {
    $query->where('enabled', true)->where('sale', true);
})->with(['products' => function ($query) {
    $query->where('enabled', true)->where('sale', true);
});
```
## After

```php
CollectionModel::withWhereHas('products', fn ($query) => $query->where('enabled', true)->where('sale', true));
```

If it's ok for you, I will add tests :)

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -153,6 +153,23 @@ trait QueriesRelationships
     }
 
     /**
+     * Add a relationship count / exists condition to the query with where clauses.
+     * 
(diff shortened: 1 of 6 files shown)


taylorotwell:
I think it seems pretty decent if you want to add tests. Mark as ready for review when done.

lovemintblue:
nice！

PR #42606, opened 2022-06-01
[9.x] Handle undefined array key error

### LINKED ISSUE #42605: v9.15 has breaking change in route() helper: Unidentified array key 0
- Laravel Version: 9.15.0
- PHP Version: 8.0.18

### Description:
In v9.15.0 calling the route() helper with model parameters in a non-associative array will result in an ErrorException "Unidentified array key 0" being thrown in Route.php line 539.
I realize that the use of non-associative arrays is no longer documented, but as long as the order of the parameters was correct, it has worked until v9.14.0. 
If there have been any warnings in the upgrade guides that this would happen, I must have missed it. And it's especially unexpected in a minor release.

### Steps To Reproduce:
```php
route('route.name', [$someModel]); // Throws ErrorException
route('route.name', $someModel); // Also worked up until v9.14.0, now it throws ErrorException
```

ksassnowski:
Fixes #42605 

Oh boy, the gift that keeps on giving. Looks like there's yet another edge case.

## Summary

This PR fixes a regression when trying passing a model to the `route` helper for a route without route parameters.

```php
$model = new User(['id' => 1]);

Route::get('/foo')->name('foo');

route('foo', $model); // Used to return `/foo?1`, now throws an exception
```

While I think this behavior is pretty weird, I guess it shouldn't break between minor versions.

### DIFF
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -536,7 +536,13 @@ class Route
     public function bindingFieldFor($parameter)
     {
         if (is_int($parameter)) {
-            $parameter = $this->parameterNames()[$parameter];
+            $parameters = $this->parameterNames();
+
+            if (!isset($parameters[$parameter])) {
+                return null;
+            }
+
+            $parameter = $parameters[$parameter];
         }
 
         return $this->bindingFields[$parameter] ?? null;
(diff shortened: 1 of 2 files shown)


driesvints:
At least we're adding tests for all of this 👍

PR #42609, opened 2022-06-01
[9.x] Utilize `Str::isJson` in JSON validation rule

ecrmnn:
This PR replaces the code used to validate JSON strings with `Str::isJson()`


taylorotwell:
If the current code is working not going to change it in case there are subtle behavior changes.

PR #42611, opened 2022-06-01
[9.x] Adds `Arr::filter`

ecrmnn:
I've been using `Arr` heavily and felt that it was missing a `filter` method.
Turns out it has been there all along, but is named `Arr::where`.

Almost every programming language as the concept of `map` and `filter`, 
so I think it would be more intuitive to call it `Arr::filter`.

To avoid a breaking change I've moved the logic to the new `filter` method, but kept `where` as a simple alias.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -177,6 +177,18 @@ class Arr
     }
 
     /**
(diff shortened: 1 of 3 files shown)


taylorotwell:
Think I'll just leave things as is.

PR #42613, opened 2022-06-01
[10.x] Adds int-backed `Enum` as an option for implicit route binding

przemyslaw-przylucki:
This pull request introduces implicit int-backed `Enum` route binding.

The original pull request which added enum route binding (https://github.com/laravel/framework/pull/40281) rejected them as an option, however, it didn't include the reason behind it.

If this PR gets rejected, we should still consider specifying in the docs that only string-backed enums are a viable option for the routing


src/Illuminate/Support/Reflector.php
```diff
@@ -141,20 +141,19 @@ public static function isParameterSubclassOf($parameter, $className)
...
 
     /**
-     * Determine if the parameter's type is a Backed Enum with a string backing type.
+     * Determine if the parameter's type is a Backed Enum.
      *
      * @param  \ReflectionParameter  $parameter
      * @return bool
      */
-    public static function isParameterBackedEnumWithStringBackingType($parameter)
+    public static function isParameterBackedEnum($parameter)
```
driesvints, line 149: Changing this method name is a breaking change unfortunately.

PR #42619, opened 2022-06-02
[9.x] Add additional uuid testing helpers

timacdonald:
This PR adds some testing helpers to the generation of UUIDs via the `Str` helper.

Currently in order to control UUID creation, you need to do the following...

```php
$uuid = Str::uuid();
Str::createUuidsUsing(fn () => $uuid);

// do stuff...

Str::uuid() === Str::uuid() === $uuid;

// cleanup...

Str::createUuidsNormally();
```

This locks the helper so that it only returns the one UUID throughout the whole process.

I find myself always having to re-learn how this works and that I have to capture a UUID first or needing to fake a specific UUID generation.

With this PR the following is now possible...

```php
Str::freezeUuids();

// do stuff...

Str::uuid() === Str::uuid() === $uuid;

// cleanup...

Str::createUuidsNormally();
```

You may also pass a closure to only freeze creation for the duration of the closure...

```php
Str::freezeUuids(function ($uuid) {
    // do stuff...
    Str::uuid() === Str::uuid() === $uuid;
});
```

Finally, it is also possible to provide a sequence of UUIDs to return. 
```php
Str::createUuidsUsingSequence([
    $zeroth = Str::uuid(),
    $first = Str::uuid(),
]);

Str::uuid() === $zeroth;
Str::uuid() === $first;
Str::uuid(); // back to random UUIDs
```

You may specify the keys if there are only certain instances you wish to sequence....

```php
Str::createUuidsUsingSequence([
    0 => ($zeroth = Str::uuid()),
    3 => ($third = Str::uuid()),
]);

Str::uuid() === $zeroth;
Str::uuid(); // random
Str::uuid(); // random
Str::uuid() === $third;
```

If you would like to throw an exception when there is no more UUIDs in the array, you pass a closure to control what happens when finding an index that has no specified UUID...

```php
Str::createUuidsUsingSequence([
    $zeroth = Str::uuid(),
    $first = Str::uuid(),
], fn () => throw new Exception('Sequence ended'));

Str::uuid() === $zeroth;
Str::uuid() === $first;
Str::uuid(); // THROWS
```

This closure is also called when there is a missing key if you specify specific keys in the sequence...

```php
Str::createUuidsUsingSequence([
    0 => Str::uuid(),
    3 => Str::uuid(),
], fn () => throw new Exception('Sequence ended'));

Str::uuid() === $zeroth;
Str::uuid(); // THROWS
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -40,7 +40,7 @@ class Str
     /**
      * The callback that should be used to generate UUIDs.
      *
-     * @var callable
+     * @var callable|null
      */
     protected static $uuidFactory;
 
@@ -1113,7 +1113,14 @@ class Str
         static::$uuidFactory = $factory;
     }
 
-    public static function createUuidsUsingSequence(array $sequence, Closure $whenMissing = null)
+    /**
+     * Set the sequence that will be used to generate UUIDs.
+     *
+     * @param  array  $sequence
+     * @param  callable|null  $whenMissing
+     * @return void
+     */
+    public static function createUuidsUsingSequence(array $sequence, $whenMissing = null)
     {
         $next = 0;
 
@@ -1140,6 +1147,12 @@ class Str
         });
     }
 
+    /**
+     * Always return the same UUID when generating new UUIDs.
+     *
+     * @param  \Closure|null  $callback
+     * @return \Ramsey\Uuid\UuidInterface
+     */
     public static function freezeUuids(Closure $callback = null)
     {
         $uuid = Str::uuid();
@@ -1154,7 +1167,6 @@ class Str
 
         return $uuid;
     }
-
 
     /**
      * Indicate that UUIDs should be created normally and not using a custom factory.
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Support;
 
+use Closure;
 use Illuminate\Support\Traits\Macroable;
 use JsonException;
 use League\CommonMark\GithubFlavoredMarkdownConverter;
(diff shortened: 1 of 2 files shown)


timacdonald:
Please feel free to suggest / improve the naming. Went down the rabbit hole of "recycle" like words, but this felt the best at the end of the day for me.

src/Illuminate/Support/Str.php
```diff
@@ -1112,6 +1113,61 @@ public static function createUuidsUsing(callable $factory = null)
         static::$uuidFactory = $factory;
     }
 
+    /**
+     * Set the sequence that will be used to generate UUIDs.
+     *
+     * @param  array  $sequence
+     * @param  callable|null  $whenMissing
```
ksassnowski, line 1120: Any chance to change the `callable` type hint to something static analysis better understands? The recent updates to the Collections doc blocks were a really nice QoL improvement and I'd love for more code to be annotated like that :)

```php
/**
 * @param null|callable(): UuidInterface $whenMissing
 */
```

Same for all the other methods that take a callback as a parameter.

tests/Support/SupportStrTest.php
```diff
@@ -891,6 +891,95 @@ public function testTransliterateStrict(string $value, string $expected): void
...
+        $this->assertSame((string) $uuids[0], (string) $uuids[1]);
+        $this->assertSame($uuids[1], $uuids[2]);
+        $this->assertSame((string) $uuids[1], (string) $uuids[2]);
+        $this->assertNotSame(Str::uuid(), Str::uuid());
+        $this->assertNotSame((string) Str::uuid(), (string) Str::uuid());
+
+        Str::createUuidsNormally();
+    }
+
+    public function testItCanSpecifyASquenceOfUuidsToUtilise()
```
ksassnowski, line 936: Typo in `Squence`

tests/Support/SupportStrTest.php
```diff
@@ -891,6 +891,95 @@ public function testTransliterateStrict(string $value, string $expected): void
...
+            0 => ($zeroth = Str::uuid()),
+            1 => ($first = Str::uuid()),
+            // just generate a random one here...
+            3 => ($third = Str::uuid()),
+            // continue to generate random uuids...
+        ]);
+
+        $retrieved = Str::uuid();
+        $this->assertSame($zeroth, $retrieved);
+        $this->assertSame((string) $zeroth, (string) $retrieved);
```
ksassnowski, line 948: Is the second assertion necessary? We're already checking that they're the same _instance_.

PR #42624, opened 2022-06-02
[9.x] Add `firstOrTap()` Eloquent methods

stevebauman:
### Description

This PR adds a `firstOrTap($closure, $attributes = [])` method on the Eloquent `Builder`, `BelongsToMany` relation, `HasManyThrough` relation, and the `HasOneOrMany` relation.

When no existing model is found, a new instance is provided to the callback.

An optional array of attributes can be supplied in the second parameter as where clauses, which will pre-populate the new model instance's attributes if no model is found.

### Purpose

These methods provide convenient encapsulation of logic that may obscure some of the outer logic.

### Example

```php
$user = User::where('...')->firstOrTap(function (User $User) {
    $user->fill(['...'])->save();
});
```

```php
$meta = $book->metadata()->where('key', 'author')->firstOrTap(function (Metadata $meta) {
    $meta->fill(['key' => 'author', 'name' => 'John Doe'])->save();
});
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -522,6 +522,20 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Get the first matching record or tap a new instance.
+     *
+     * @param Closure $callback
+     * @param array   $attributes
+     * @return \Illuminate\Database\Eloquent\Model|static
+     */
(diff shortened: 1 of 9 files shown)


eusonlito:
Is so easy as, no?

```php
$user = User::where('...')->firstOr(fn () => User::create($data));
```

Or using the `firstOrCreate` method.

PR #42626, opened 2022-06-02
[9.x] Adds `plain_text` validation rule

ecrmnn:
### Why?
When validating user input (especially non-English) rules like `alpha`, `alpha_dash` and `alpha_num` doesn't work well, while `string` is too inclusive.

We want to allow as much as possible _(characters, quotes, dashes and emojis)_ -- while making sure there are no code injection / executable code.

| Test | `plain_text` | `alpha` | `alpha_dash` | `alpha_num` | `string` |
|-|-|-|-|-|-|
| Joe O'Brian | ✅ | ❌ | ❌ | ❌ | ✅ |
| Mary-Kate | ✅ | ❌ | ✅ | ❌ | ✅ |
| X Æ A-Xii | ✅ | ❌ | ❌ | ❌ | ✅ |
| Mrs. Jackson | ✅ | ❌ | ❌ | ❌ | ✅ |
| Can contain "double quotes" | ✅ | ❌ | ❌ | ❌ | ✅ |
| Blåbærsyltetøy | ✅ | ✅ | ✅ | ✅ | ✅ |
| ラーメン | ✅ | ✅ | ✅ | ✅ | ✅ |
| 🍑 | ✅ | ❌ | ❌ | ❌ | ✅ |
| https://url.com | ✅ | ❌ | ❌ | ❌ | ✅ |
| !"#$%&/()=_-@*^¨1234567890 | ✅ | ❌ | ❌ | ❌ | ✅ |
|     John Doe | ✅ | ❌ | ❌ | ❌ | ✅ |
| John\nDoe | ❌ | ❌ | ❌ | ❌ | ✅ |
| `<p>HTML tags is not OK</p>` | ❌ | ❌ | ❌ | ❌ | ✅ |
| \t\n\r\0\v is not OK | ❌ | ❌ | ❌ | ❌ | ✅ |
| <script>console.log('hello');</script> | ❌ | ❌ | ❌ | ❌ | ✅ |

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1463,7 +1463,6 @@ trait ValidatesAttributes
      */
     public function validatePlainText($attribute, $value)
     {
-        dump(trim(strip_tags($value)), $value);
         return trim(strip_tags($value)) === $value;
     }
 
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1455,6 +1455,19 @@ trait ValidatesAttributes
     }
 
     /**
+     * Validate that an attribute is plain text.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @return bool
+     */
+    public function validatePlainText($attribute, $value)
+    {
+        dump(trim(strip_tags($value)), $value);
+        return trim(strip_tags($value)) === $value;
+    }
+
+    /**
      * Validate that an attribute exists even if not filled.
      *
      * @param  string  $attribute
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1463,7 +1463,6 @@ trait ValidatesAttributes
      */
     public function validatePlainText($attribute, $value)
     {
-        dump(str_replace(["\n","\r","\t","\v","\0","\x00"], '', strip_tags($value)), $value);
         return str_replace(["\n","\r","\t","\v","\0","\x00"], '', strip_tags($value)) === $value;
     }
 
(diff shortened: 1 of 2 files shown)


dennisprudlo:
I like the idea, however with `trim(strip_tags($value)) === $value` the `John\nDoe` example will pass (there is no tag and nothing to trim).

Also, IMO leading and trailing spaces shouldn't be a reason to fail. If a space between words is considered "plain text" why isn't a space (or more) after the full text.

ecrmnn:
@dennisprudlo Thanks for the heads up. Also, I agree that it probably shouldn't fail on leading/trailing spaces. I've updated the code to fix this.

ecrmnn:
@johanrosenson Hmm, while I do agree that newlines are part of plain text, I generally don't want them. At least that's one of the things I want this rule to filter out.

Should I change the name to something different? Any suggestions?

Maybe there should be two rules?
- `plain_text` for simply stripping tags
- `no_white_space` for removing white space like [trim()](https://www.php.net/manual/en/function.trim.php) does

|Test|plain_text|no_white_space|plain_text + no_whitespace
|-|-|-|-|
|Joe\nDoe|✅|❌|❌|
|`<p>HTML tags is not OK</p>`|❌|✅|❌|

judgej:
Essentially you are pulling in a collection of simpler rules, and giving it a name. I'm sure no matter what rules you use, it will not be suitable for someone's specific use-case.

Would a rule macro or meta rule approach, with some default examples, be a better approach? It would have a very wide range of use-cases.

ecrmnn:
@judgej Yup, but you can say the same for every validation rule though. 

What I want is something that solves 90% of all user input. Stuff like, first/middle/last name, address, city, title, description, excerpt. Just to make sure there are no code injection or weird formatting.

Existing rules falls short.

Maybe the name is misleading and it should be named something else to make it clear, e.g.;
- `plain_string`
- `without_tags`
- `trimmed`

taylorotwell:
I feel like this will have a lot of opinions around it. Feel free to release as a package or just as a simple custom rule people can copy and paste.

judgej:
That was my concern. Whatever the rules in it, it would be wrong for too many people.

Creating a custom rule and using that to wrap a bunch of more primitive validation rules, is incredibly easy now, as is providing new validation rules through a package.

PR #42629, opened 2022-06-02
[10.x] Introduce int backed enum for implicit route binding

przemyslaw-przylucki:
This pull request introduces implicit int-backed Enum route binding.

The original pull request which added enum route binding (https://github.com/laravel/framework/pull/40281) rejected them as an option, however, it didn't include the reason behind it.

If this PR gets rejected, we should still consider specifying in the docs that only string-backed enums are a viable option for the routing


taylorotwell:
Is there a particular reason you sent this to 10.x branch instead of 9.x?

przemyslaw-przylucki:
@taylorotwell In my initial PR, I've created this PR for `9.x`, but @driesvints told me this would be a breaking change, so I changed the target to master, and then I messed up the branch so I wanted to open a new PR without all the junk (I suck at git, sorry)
 
I've created a new PR which solves the issue you've mentioned above

PR #42634, opened 2022-06-02
[9.x] Add onFailure method to Pipeline

reziamini:
Now, we are not able to do some stuff with the pipeline if there is a failed pipe.

With this new method, we will be able to throw a custom exception or our custom functionality (like logging, ...)

### DIFF
--- a/src/Illuminate/Pipeline/Pipeline.php
+++ b/src/Illuminate/Pipeline/Pipeline.php
@@ -25,6 +25,13 @@ class Pipeline implements PipelineContract
     protected $passable;
 
     /**
+     * The callback to be executed on failure pipeline.
+     *
+     * @var Closure
+     */
+    protected $onFailure;
+
+    /**
      * The array of class pipes.
      *
      * @var array
@@ -182,6 +189,10 @@ class Pipeline implements PipelineContract
 
                     return $this->handleCarry($carry);
                 } catch (Throwable $e) {
+                    if ($this->onFailure) {
+                        return call_user_func($this->onFailure, $pipe);
+                    }
+
                     return $this->handleException($passable, $e);
                 }
             };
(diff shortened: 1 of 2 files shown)


src/Illuminate/Pipeline/Pipeline.php
```diff
@@ -182,6 +189,10 @@ protected function carry()
 
                     return $this->handleCarry($carry);
                 } catch (Throwable $e) {
+                    if ($this->onFailure) {
+                        return call_user_func($this->onFailure, $pipe);
```
rodrigopedra, line 193: It would be nice to also send the exception to the `onFailure` method, for logging purposed, for example.

I'd also suggest to send it as the first parameter.

rezaamini-ir: I don't have any opinion on that, let's see what will Taylor say.

rezaamini-ir: @michael-rubel Didn't get it, what's wrong with `call_user_func`?
Could you please be more specific?

taylorotwell:
But you can do some stuff is there is a failed pipe using normal PHP try / catch syntax?

```php
    try {
        $pipeline = (new Pipeline)->send('data')->through([
            fn () => throw new Exception('foo'),
        ])->then(fn ($piped) => $piped);
    } catch (Exception $e) {
        dd('here');
    }
```

michael-rubel:
Hey @rezaamini-ir 

Your PR inspired me to create a modified version of Pipeline with this method and a DB transaction as a separate package:
https://github.com/michael-rubel/laravel-enhanced-pipeline

Try it out if needed and let me know if you want to add something else. 🙂

src/Illuminate/Pipeline/Pipeline.php
michael-rubel: @rezaamini-ir I meant using of more modern approach `($this->onFailure)($pipe)` instead of `call_user_func`.
Look at this line: https://github.com/michael-rubel/laravel-enhanced-pipeline/blob/main/src/Pipeline.php#L236

PR #42635, opened 2022-06-02
[9.x] Possibility to force safe retrieval in Form Request

michael-rubel:
### The case
FormRequest values are usually accessed by using `$request->value`, i.e. by a magic method. If `$request->value` is accessed, it means that all defined values have been validated, however, it does not mean that the value itself has been validated and wanted by the application. It could be resolved by using safe or validated methods, but I cannot find an option to set this behavior by default when requesting values from a form request.

This PR adds the possibility to force a safe retrieval from the `validated` method when accessing values in Form Requests.
You can turn it on by overriding class property:
```php
protected $safeInputElements = true;
```

This saves us from extracting validated input each time:
```php
public function controllerMethod(FormRequest $request): mixed
{
    $request = $request->safe();
}
```

### Breaking?
It's fully backward-compatible since this feature is optional and turned off by default. When disabled, the new `__get` just forwards the call to the parent.

### Any thoughts?
What do you think about this? The PR will stay as a draft until I see the green light from the community, or close it if people don't want this feature. 🔨

### DIFF
--- a/src/Illuminate/Foundation/Http/FormRequest.php
+++ b/src/Illuminate/Foundation/Http/FormRequest.php
@@ -292,6 +292,6 @@ class FormRequest extends Request implements ValidatesWhenResolved
     {
         return ! $this->safeInputElements
             ? parent::__get($key)
-            : Arr::get($this->safe(), $key, fn () => $this->route($key));
+            : Arr::get($this->validated(), $key, fn () => $this->route($key));
     }
 }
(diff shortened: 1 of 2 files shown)


michael-rubel:
I'm going to risk opening the idea for a review, as it seems to be liked by some people and doesn't have any downvotes at the moment.  🙂

taylorotwell:

michael-rubel:
Released this as a package if someone's interested in this feature:
https://github.com/michael-rubel/laravel-safe-request

PR #42640, opened 2022-06-02
Add php array_change_key_case() support to Collection

meyer59:
This PR add provide the php `array_change_key_case()`  to Collection and LazyCollection.
**Benefit**:
When you retrieve data from onlines API's, sometimes they are not following any name convention. For the sake of code readability it's more convenient to work with standardize array keys. 
Also, when the api result have 100 array keys, it's hard to remember how they named/cased everything.
```
$apiData = collect([
     "ID"       => 1,
     "NAME"     => "some name",
     "Lastname" => "some lastname",
 ]);
 //before this PR
 $data = array_change_key_case($apiData->all());
 collect($data)->dd();
 //after this PR
 $apiData->changeKeyCase()->dd();
```

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -1708,12 +1708,13 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
 
     /**
      * apply the same case to all the collection key
+     *
      * @param int $case
      *
      * @return $this
      */
     public function changeKeyCase(int $case = CASE_LOWER): self
     {
-        return new static(array_change_key_case($this->items,$case));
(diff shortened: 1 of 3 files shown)


ecrmnn:
This already super easy with `mapWithKeys` tho.

```php
collect(['FiRst' => 'Hello', 'Second' => 'hello'])
    ->mapWithKeys(function ($v,$k) {
        return [mb_strtoupper($k) => $v];
    });
```

meyer59:
- A lot of collection method are using map under the hood and they still exists even if we can leverage `map()`. 
- Obviously, whenever i can, i would rather go for a native php function to do the job instead of me looping and using `mb_strtoupper()`

taylorotwell:

PR #41950, opened 2022-04-13
[9.x] Add withoutEagerLoads() method to Builder

iamgergo:
This PR just adds the `withoutEagerLoads` method to the `Eloquent\Builder` class.

```php
// Before
$query->setEagerLoads([]);

// After
$query->withoutEagerLoads();
```

PR #42641, opened 2022-06-02
[9.x] Adds `withoutEagerLoad` function

mahmoudmohamedramadan:
According to this [PR](https://github.com/laravel/framework/pull/41950) I have added a `withoutEagerLoad` to prevent the given relationships from being eagerly loaded.

```PHP
$user = \App\Models\User::with(['post', 'attachment'])->first();

// This line will return the first user with the post relationship
return $user->withoutEagerLoad(['attachment'])->first();
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1575,12 +1575,12 @@ class Builder implements BuilderContract
     }
     
     /**
-     * Flush the given relationships being eagerly loaded.
+     * Indicate that the given relationships should not be eagerly loaded.
      *
      * @param  array  $relations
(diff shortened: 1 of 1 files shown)


src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -1573,6 +1573,19 @@ public function setEagerLoads(array $eagerLoad)
 
         return $this;
     }
+    
+    /**
+     * Flush the given relationships being eagerly loaded.
```
rodrigopedra, line 1578: I guess this comment needs to be changed, it is the same as the method below (`withoutEagerLoads`) and is not describing intended feature.

PR #42649, opened 2022-06-03
[9.x] Adds the Dispatchable trait

imanghafoori1:
Laravel event classes can use the `Dispatchable` trait and be dispatched like this: 
### Usage:
```php
OrderShipped::dispatch($order);
```

OrderShipped:
```php
class OrderShipped
{
    use Dispatchable;

    public function __construct(public Order $order)
    {
          //
    }
}

```


siarheipashkevich:
It's already done, please see https://laravel.com/docs/9.x/events#dispatching-events

PR #42656, opened 2022-06-03
now we can run migrations when we start server by passing m flag

manadinho:
With this we can run migrations at the time when we start our server by passing m flag with serve command.
Like:
**php artisan serve -m**
It would be helpful when more than one developers are working on a project, So they don't need to tell to run migrations in his branch.


taylorotwell:

PR #42657, opened 2022-06-03
[9.x] Adds rename key to Arr

imanghafoori1:
Adds the ability to rename some keys of an associative array without changing the order of the values.

This is usually needed when we get some data from a source and want to expose it in a public API, but the keys are not very readabale.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -817,4 +817,22 @@ class Arr
 
         return is_array($value) ? $value : [$value];
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
mapWithKeys

imanghafoori1:
@taylorotwell

1.  `mapWithKeys` is only available in `Collection`
2. This does not replace nested keys and is not very easy to write.
```php
Collection::make($array)->mapWithKeys(
            fn ($item, $key) => [($replace[$key] ?? $key) => $item]
        )->all();
```

PR #42661, opened 2022-06-03
[9.x] Adds renameKeys to the Collection

imanghafoori1:
Adds the ability to rename some keys of an associative array collection without changing their order.

Following the previous PR #42657, this implementation is based on `mapWithKeys` with more tests covering edge cases.

```php
$array = [
  'key1' => 'val_1',
  'key2' => 'val_2',
  'nested' => [ 'key1' => 'val_1', 'key2' => 'val_2'],
];

// we want to rename "key2" to "new_key_2"
$replace = [
  'key2' => 'new_key_2',
];


$result = Collection::make($array)->renameKeys($replace, recursive: true)->all();


$result = [
  'key1' => 'val_1',
  'new_key_2' => 'val_2',
  'nested' => [ 'key1' => 'val_1', 'new_key_2' => 'val_4'],
];
```
- **Use cases**:
This is usually needed when we get some data from a source and want to expose it in a public API, but the keys are not very readable.
Another typical use case is when we receive a form request which represents data to be inserted in a table but the key names in the request does not exactly map to the name of the database columns.

```php
$data = request()->all();

$data = Arr::renameKeys($data, ....);

MyModel::create($data);

```

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -817,4 +817,16 @@ class Arr
 
         return is_array($value) ? $value : [$value];
     }
+
+    /**
+     * Renames the given array keys.
+     *
+     * @param  array<string, mixed>  $array
+     * @param  array<string, string>  $replace
+     * @return array
+     */
+    public static function renameKeys(array $array, array $replace, $recursive = true)
+    {
+        return Collection::make($array)->renameKeys($replace, $recursive)->all();
+    }
 }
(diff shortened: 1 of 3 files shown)


johanrosenson:
Would not `keyBy` work for this?

```
$collection->keyBy(function ($value, $key) {
    $newKey = 'something';

    return $newKey;
});
```

It accepts a callback which receives the value and the current key.

imanghafoori1:
@johanrosenson No that is totally a different thing. Check the tests at `SupportCollectionTest@testKeyBy`.

johanrosenson:
@imanghafoori1 

ok, could you include an usage example in your original post so it's easier to quickly understand how this is supposed to be used?

taylorotwell:

PR #42664, opened 2022-06-04
[9.x] Fixed grouping for user authorization

rennokki:
See: https://github.com/laravel/framework/pull/42531#issuecomment-1146336499

### DIFF
--- a/src/Illuminate/Broadcasting/BroadcastManager.php
+++ b/src/Illuminate/Broadcasting/BroadcastManager.php
@@ -58,32 +58,56 @@ class BroadcastManager implements FactoryContract
     /**
      * Register the routes for handling broadcast authentication and sockets.
      *
-     * @param  array|null  $channelAuthorizationAttributes
-     * @param  array|null  $userAuthorizationAttributes
+     * @param  array|null  $attributes
      * @return void
      */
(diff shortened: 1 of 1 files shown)


src/Illuminate/Broadcasting/BroadcastManager.php
```diff
@@ -74,14 +74,42 @@ public function routes(array $attributes = null)
...
+        });
+    }
+
+    /**
+     * Register the routes for handling broadcast authentication.
+     *
+     * @param  array|null  $attributes
+     * @return void
+     */
+    public function userAuthenticationRoutes(array $attributes = null)
```
driesvints, line 86: Probably best to just make this an empty array fallback rather than use two different types.

src/Illuminate/Broadcasting/BroadcastManager.php
```diff
@@ -74,14 +74,42 @@ public function routes(array $attributes = null)
...
         });
     }
 
+    /**
+     * Alias for routes().
+     *
+     * @param  array|null  $attributes
+     * @return void
+     */
+    public function channelAuthorizationRoutes(array $attributes = null)
```
driesvints, line 108: Probably best to just make this an empty array fallback rather than use two different types.

PR #42665, opened 2022-06-04
[9.x] Dispatch jobs with the ShouldBeUnique interface only once after response

redelschaap:
When using the `dispatchAfterResponse` function on a job that implements the `ShouldBeUnique` interface, I would expect it to be dispatched only once after response. Instead, these jobs are dispatched as many times as you call the function.

Dispatching them only once is very useful when you have more than one reason to dispatch a job, but you need it to be actually running one time.

This PR offers a solution, just not sure if this would be considered a breaking change.

### DIFF
--- a/src/Illuminate/Bus/Dispatcher.php
+++ b/src/Illuminate/Bus/Dispatcher.php
@@ -4,6 +4,7 @@ namespace Illuminate\Bus;
 
 use Closure;
 use Illuminate\Contracts\Bus\QueueingDispatcher;
+use Illuminate\Contracts\Cache\Repository as Cache;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Contracts\Queue\Queue;
 use Illuminate\Contracts\Queue\ShouldBeUnique;
@@ -51,11 +52,6 @@ class Dispatcher implements QueueingDispatcher
      * @var \Closure|null
      */
     protected $queueResolver;
-
-    /**
-     * @var array
-     */
-    private array $dispatchingJobsAfterResponse = [];
 
     /**
      * Create a new command dispatcher instance.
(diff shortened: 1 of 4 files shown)


ankurk91:
It should rather check the lock key.

How does it work with Octane?

redelschaap:
Let me see if I can make this work using a lock key.

Never worked with Octane myself. Let me also see if I can try that out.

redelschaap:
Lock key is now implemented.

I'm afraid I don't have the ability to set up and test with Octane now. Is there "an easy way" to test this or could someone else test this on Octane?

ankurk91:
Since you are no longer storing the keys within class, there is no concerns with Octane now

redelschaap:
Good to know!

taylorotwell:
Please provide examples for how to recreate the broken situation now so we can see what you are referring to.

redelschaap:
In the following situation, the job will be executed thrice:

```php
JobImplementingTheShouldBeUniqueInterface::dispatchAfterResponse();
JobImplementingTheShouldBeUniqueInterface::dispatchAfterResponse();
JobImplementingTheShouldBeUniqueInterface::dispatchAfterResponse();
```

For example, I use the `dispatchAfterResponse()` method on jobs that needs to run for different types of a specific model, while looping through them. Let's say I have 10 records with in total 3 different model types, I need this job to be dispatched 3 times. Instead, the job will be dispatched and executed 10 times.

However, because they implement the `ShouldBeUnique` interface, I would expect the job only to be executed 3 times (assuming the uniqueId is set up correctly). Especially because all (unique) jobs will do the same work and thus are being run unnecessarily.

taylorotwell:
I don't find the behavior that unexpected. Or, at least, it seems undefined to me. Telling the job to dispatch after the response doesn't seem compatible with being "unique". You are telling the job to run synchronously within the current process - not on the queue.

redelschaap:
@taylorotwell Well maybe the thing is that it is undefined. But I guess the job is synchronously, so it should run multiple times. On the other hand, what's the value of running the same job multiple times at the end of a process?

Another thought... Should a job that implements the `ShouldQueue` interface be actually queued when using the `afterResponse` methods (instead of being executed)? That would make more sense to me.

If not, would a `dispatchOnceAfterResponse` method be a good addition to the framework?

Queueing the job is just not enough in some situations, because a worker can already executed it before the process that wants to queue it is fully finished.

PR #42668, opened 2022-06-04
[8.x] Pass the full custom pivot model to the `deleting` & `deleted` events.

r-kujawa:
The changes made in this PR will provide the actual pivot model to be passed into the `deleting` & `deleted` events, instead of solely the foreign pivot key & related pivot key.

### Real life example:
I have a `Company::class` model and a `PaymentMethod::class` model along with a `CompanyPaymentMethod::class` pivot model . The pivot table has an additional boolean column called `is_primary`. In this event I would like to apply some business logic and do certain things when a primary payment method relationship is being destroyed, but I do not have access to the "full" `CompanyPaymentMethod::class` pivot model's attributes as it only comes in with the foreign & primary keys.

The workaround I currently provided in in project was to change my event from `deleted` to `deleting` in order to be able to query the table myself, it is not a good solution as it will cause a `n + 1`, in case a `sync()` or multiple `detach()` is being made.
```
public function deleting(CompanyPaymentMethod $pivot)
{
    $pivot = CompanyPaymentMethod::where([
        'company_id' => $pivot->company_id,
        'payment_method_id' => $pivot->payment_method_id
    ])->first();

    if ($pivot->is_primary) {
        // do something
    }
}
```

🔴  A disadvantage I would see vs the current solution is the fact that this will be making an additional query to retrieve the pivot models that will be deleted. But it could prevent hundreds of queries if there are people applying a workaround like the one I mention above and could also save countless hours in debugging this as all other pivot model events pass in the full pivot model as it is expected.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
@@ -475,7 +475,7 @@ trait InteractsWithPivotTable
 
         $pivots = $this->using::where($this->foreignPivotKey, $this->parent->{$this->parentKey})->whereIn($this->relatedPivotKey, $this->parseIds($ids))->get();
 
-        foreach($pivots as $pivot) {
+        foreach ($pivots as $pivot) {
             $results += $pivot->setPivotKeys($this->foreignPivotKey, $this->relatedPivotKey)->delete();
         }
 
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
@@ -473,7 +473,10 @@ trait InteractsWithPivotTable
     {
         $results = 0;
 
-        $pivots = $this->using::where($this->foreignPivotKey, $this->parent->{$this->parentKey})->whereIn($this->relatedPivotKey, $this->parseIds($ids))->get();
+        $pivots = $this->using::withoutGlobalScopes()
+            ->where($this->foreignPivotKey, $this->parent->{$this->parentKey})
+            ->whereIn($this->relatedPivotKey, $this->parseIds($ids))
+            ->get();
 
         foreach ($pivots as $pivot) {
             $results += $pivot->setPivotKeys($this->foreignPivotKey, $this->relatedPivotKey)->delete();
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
@@ -473,11 +473,10 @@ trait InteractsWithPivotTable
     {
         $results = 0;
 
-        foreach ($this->parseIds($ids) as $id) {
-            $results += $this->newPivot([
-                $this->foreignPivotKey => $this->parent->{$this->parentKey},
-                $this->relatedPivotKey => $id,
-            ], true)->delete();
+        $pivots = $this->using::where($this->foreignPivotKey, $this->parent->{$this->parentKey})->whereIn($this->relatedPivotKey, $this->parseIds($ids))->get();
+
+        foreach($pivots as $pivot) {
+            $results += $pivot->delete();
         }
 
         return $results;
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
@@ -476,7 +476,7 @@ trait InteractsWithPivotTable
         $pivots = $this->using::where($this->foreignPivotKey, $this->parent->{$this->parentKey})->whereIn($this->relatedPivotKey, $this->parseIds($ids))->get();
 
         foreach($pivots as $pivot) {
-            $results += $pivot->delete();
+            $results += $pivot->setPivotKeys($this->foreignPivotKey, $this->relatedPivotKey)->delete();
         }
 
         return $results;


rodrigopedra:
In the case you described, you are moving the extra calls from your listeners to the core. I see by using `WHERE IN` the aim is to make less calls, but not all applications has more than one record in a pivot table for a given related pair, so for most applications the number of queries will increase without need.

So instead of extra queries being needed just for particular use cases like yours, they will end up being made for all applications regardless they use extra fields, or have more than one record for a particular foreign key pair as it seems to be your case.

As one workaround -- in case this PR doesn't get merged -- you could override this method on your app's Pivot models.

A second workaround -- not sure this will work, I didn't test it -- is to add the pivot's company relation to the pivot's `$touches` attribute, and listed for the `saved` event on the `Company` to perform any logic when its primary payment method is deleted.

reference: https://laravel.com/docs/9.x/eloquent-relationships#touching-parent-timestamps

r-kujawa:
@rodrigopedra I don't think your understanding the point of this PR, I am aware of the additional query that is being made every time a `detach()` or `sync()` function is called on a many to many relationship, that's why I try to make it very clear of this disadvantage in the description, and I am aware this PR might not get merged solely for this reason. But I am trying to make a point on all the advantages it may bring to the table, hopefully these will overrule the disadvantage.

The way I see it is we will be making one extra query (only for pivot models that were explicitly made pivot models by the developer) to gain the following:
- We will be able to make use of any additional column we have on our pivot table in the `deleting` & `deleted` events, because let's face it, if you're explicitly `->using()` a pivot model on your `->belongsToMany()` relationship it is very probable you are doing it because you have additional columns in that pivot and/or you are taking advantage of the model events.
- You will standardize the code between all model events (they will now send the full model instance to the `deleting` & `deleted` events just like in the `creating`, `created`, `updating`, `updated`, `saving` & `saved` events).
- If a developer ever needs the additional columns in the `deleted` event he will not be able to retrieve them since they no longer exist, with this the bug is fixed.

**Side Note**: Just like normal models this additional query will not be made and no `deleting` or `deleted` event will be fired if this is considers a mass delete of the pivot models (e.g. `$company->paymentMethods()->detach()`), this is the current behaviour.

Another thing to consider would be that, it is probable that many applications are having this issue and don't even know about it, since I could argue that most developers would assume they are receiving their "full" pivot model columns to the event, and they are probably running some sort of `if ()` statement on another column which would be `null` (falsy) by nature.

Nowadays one "simple" query to the database shouldn't impact it's overall performance, if it does, the source of the issue is probably being originated elsewhere.

The first workaround would work for me, my only concern would be for future developers when upgrading Laravel versions. But the second one would not work since I would have to implement a lot more logic in order to retrieve my pivot model that is being deleted to find out whether or not it is primary.

Also, what would be the point of a pivot table if you will not be using it to store more than one record for a given related pair? Would that make it a one to one (or one to many) relation where the pivot would be useless?

rodrigopedra:
I got your point now, thanks for the clarification. Indeed it is fair to think a developer will reach out for using a custom pivot model when they have custom pivot columns.

Regarding this:

> Also, what would be the point of a pivot table if you will not be using it to store more than one record for a given related pair? 

Well, usually there is just a single related pair of foreign keys. For example let's say there are students A, B, and C, and the three of them are enrolled on classes 1 and 2.

The way I see this pivot is like this

| student | classroom | 
|---------|-----------| 
| A | 1 |
| A | 2 |
| B | 1 |
| B | 2 |
| C | 1 |
| C | 2 |

So there is no repetition for a given pair, let's say `(A, 2)` only shows a single time.

I assumed, and I might be wrong, that in your case as you have a `is_primary` column, that could mean a user could have the following pairs: `(A, 2, true)` (`true` for `is_primary`) and `(A, 2, false)`, in this case the foreign keys pair would be repeated.

Of course on a simple example like this, assuming `is_primary` means, for example, which "classroom" (in my example above)  a would be the preferred for a student, so a repeating foreign key pair wouldn't make sense. 

But, on the other hand, I saw some projects trying to keep soft deleted pivot records in a sort of history tracking, so as a foreign key pair can be soft-deleted, and later re-added, in that case there could some repeating pairs.

But now after you explained your use case, the case I described seems to be too edgy to justify not adding your PR.

Just to be clear, I just answered the last question for clarification as you said you didn't understand, but now I agree that what I described was due to my misunderstanding of the matter.

Good luck =)

axlon:
I think this PR would be fine if it was both opt-in and if the query was configurable. As it stands it adds an additional query (which is likely to be redundant for a lot of people) and it is likely to break applications that use global scopes on their pivots because there is no way to disable them here.

Btw I don't think this PR should go to 8.x, IMO it should go to 10.x if it is not opt-in (and 9.x if it is)

r-kujawa:
@axlon I do believe this is opt-in as the developer opts-in for this when they decide to use a pivot model, these are the reasons why I believe this is an actual bug of the framework:
1. You are only getting half your pivot model to your `deleting` & `deleted` functions (I could compare this to only receiving the 'id' attribute in the same events for a normal model, that wouldn't make sense right?)
2. The name of the function itself (`detachUsingCustomClass()`) is describing what it does and it is not doing it correctly if it will only "simulate" the class with the id's for the sole purpose of generating the delete query.

Good catch with the global scopes, I will check to see if it is viable to apply a `disableGlobalScopes()` here 😄.

r-kujawa:
@rodrigopedra Thanks for the clarification, in my example the `is_primary` column is meant to mark a single `PaymentMethod::class` model relationship to the `Company::class` model as primary, I never actually have 2 relations with the same id attached to a single company. I actually leverage the `saving()` event to check if the saving pivot model `is_primary` to update all the company's payment method pivot records to `['is_primary' => false]` to ensure that a company has only one primary payment method at all times, and although we do our best to not allow the removal of a `is_primary` record, in some cases it is impossible to prevent, so we apply some business logic to that event.

taylorotwell:
I'm not changing this behavior on 8.x. Too risky.

PR #42671, opened 2022-06-05
[9.x] Adds `compileOld` function

mahmoudmohamedramadan:
This PR makes accessing the old input value easier like so,

**Old version**

```HTML
<input type="text" name="name" class="form-control" value="{{ old('name', 'defalut_value') }}">
```

**New version**

```HTMl
<input type="text" name="name" class="form-control" value="@old('name', 'defalut_value')">
```


ankurk91:
Echo() is prone to XSS, the output of old() function must be escaped

taylorotwell:

PR #42677, opened 2022-06-06
[9.x] Add `Arr::nth` and `Arr::trim`

redelschaap:
This PR adds two small functions to the Arr class that I now often solve with ugly code.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -802,7 +802,7 @@ class Arr
      */
     public static function trim(array $array, string $characters = " \n\r\t\v\x00"): array
     {
-        foreach ($array as & $value) {
+        foreach ($array as &$value) {
             $value = is_array($value) ? static::trim($value, $characters) : trim($value, $characters);
         }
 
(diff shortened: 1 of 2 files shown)


ankurk91:
Convert empty strings to null, can also be a nice addition to Arr class

bert-w:
I think I've seen a similar PR before for the `nth` function. The thing is that if you try to access the array:
```
$arr = ['foo' => 'bar', 'k' => 'v']
```
using a numerical index, it becomes ambiguous when the array is defined as:
```
$arr = ['k' => 'v', 'foo' => 'bar']
```

Arguably, those two arrays are the same (internally they have a different order though).

I'm sure there are cases though to access the keys by index but at first sight it looks like an antipattern. If all else fails, you could do it with raw php using:
```
$second = $arr[array_keys($arr)[1]];
```
Or the collection variant:
```
$collection = collect($arr);
$second = $collection[$collection->keys()[1]];
```
Also note that there is an `nth` function in the Collection class which "Creates a new collection consisting of every n-th element.". So there's a bit of a meaning/naming collision there.

PR #42679, opened 2022-06-06
[9.x] Improve tap() documented return type

axlon:
This PR aims to improve the documented return type of `tap()`. It still is not 100% accurate, but it's a fair bit better than `mixed` in my opinion.

(This also makes the global tap helper return type consistent with `Tappable::tap()`)


axlon:
@nunomaduro Understandable 👍

A 100% accurate version is possible with conditional return types if you guys are open to that. As far as I can tell both PHPStan and Psalm support them, don't know about IDEs though.

PR #42683, opened 2022-06-06
[9.x] Adds `not_starts_with` validation

mahmoudmohamedramadan:
This PR adds the ability to check if the string does not start with the given values, and I think this validation is important and makes the validation easier and we will not resort to using something else such as Regex.

```PHP
$request->validate([
    'name' => ['required', 'not_starts_with:mahmoud'],
]);
```

@taylorotwell The PR sent to 9.x branch as you told :)


michaelnabil230:
Change the title to [9.x] Add `not_starts_with` in to validation rules

PR #42686, opened 2022-06-07
[10.x] Add `can` middleware helper to resource routes and route registrar, improve consistency

tontonsb:
I really enjoy the `can` method added in #39464. So much that I often run into situations where I try to use it but it's not available there (yet).

This PR adds the ability to use of `can` in these two constructions:

```php
Route::resource('users', UserController::class)->can('manage-users');

Route::can('view-enums')->group(function () {
    Route::get('statuses', fn() => 'Ok');
});
```

To prevent `can('action')->middleware('mw')` or `middleware('mw')->can('action')` from overriding each other I changed the behaviour of `->middleware()` to merge the middleware stack instead of overriding it:

```php
Route::middleware('a')
    ->middleware('b')
    ->get('users', fn() => 'Ok');

Route::resource('users', UserController::class)
    ->middleware('a')
    ->middleware('b');
```

These previously lead to middleware stack `['b']`, but now lead to middleware stack `['a', 'b']` and that is why this PR is a breaking change. This new behaviour is consistent with how `->middleware('x')` behaves on a route (e.g. when chaining after `Route::get()`).

And one more consistency improvement is that comma-separated arg list can now be used to set multiple middlewares on resources, just like it was already available on `Route::middleware()` and `Route::get()->middleware()`:

```php
// Previously 'b' and 'c' would be discarded
Route::resource('users', UserController::class)->middleware('a', 'b', 'c');
```

### DIFF
--- a/src/Illuminate/Routing/PendingResourceRegistration.php
+++ b/src/Illuminate/Routing/PendingResourceRegistration.php
@@ -149,15 +149,33 @@ class PendingResourceRegistration
      */
     public function middleware($middleware)
     {
-        $middleware = Arr::wrap($middleware);
+        if (! is_array($middleware)) {
+            $middleware = func_get_args();
+        }
 
         foreach ($middleware as $key => $value) {
             $middleware[$key] = (string) $value;
         }
 
-        $this->options['middleware'] = $middleware;
+        $this->options['middleware'] = array_merge(
+            (array) ($this->options['middleware'] ?? []), $middleware
+        );
 
         return $this;
+    }
+
+    /**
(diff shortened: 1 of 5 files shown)


tontonsb:
@taylorotwell If I may continue the discussion a tiny bit, is this API inconsistency really intentional/desired?

```php
// Merges middlewares to produce ['a', 'b']
Route::get('ok', fn() => 'ok')->middleware('a')->middleware('b');

// Overrides middlewares producing ['b']
Route::resource('users')->middleware('a')->middleware('b');
```

Or is the objection only about the merging on registrar? Or just my merge implementation looking fishy? 😂

PR #42687, opened 2022-06-07
[10.x] Pass the full custom pivot model to the `deleting` & `deleted` events.

r-kujawa:
The changes made in this PR will provide the actual pivot model to be passed into the `deleting` & `deleted` events, instead of solely the foreign pivot key & related pivot key.

🔴 Full description & conversation in #42668. (Closed for 8.x)


r-kujawa:
@taylorotwell,

It is a lot simpler than you might think, the modified method is only being called from one function (`detach()`) and has many constraints on it  to make sure it is only being called in the rightful scenario. The only thing that could possibly be affected in the end result of this is that instead of receiving solely the related model's keys it will be populated with the full pivot model's attributes to be used in the `deleting` & `deleted` events. So it only adds data, doesn't affect any other functionality.

Either way I respect your decision, and understand how hard it is to keep up with all of this.

PR #42689, opened 2022-06-07
[9.x] Introduce Invokable validation classes

timacdonald:
This PR aims to introduce a new class based validation implementation that mixes the brevity + simplicity of Closure based rules with the shareable, extendable, and chainable nature of class based rules by introducing an Invokable validation rule.

Some additional notes:

- Brings 1st party translation support to failure messages, without the need to reach for the translation helper.
- You no longer need to manage state around messages and failures when it comes to complex validation where multiple validation errors may occur.
- Within a complex validation rule, you can no longer add a messaged and return true. Adding a messages causes the rule to fail - so no more double handling.
- Having the error messages managed in another layer also creates consistency across validation rules.
- An exception is thrown when it is unable to find a validation key, if `$fail('key')->translate()` is called. I've see this a bit where validation messages are not added. This makes it strict about finding the validation message.

Below is a comparison of the two implementations with a rule that handles a rich object...

```php
class Quantity implements Rule
{
    protected $messages = [];

    {
        if (! is_array($value)) {
            $this->messages[] = trans('validation.quantity.must_be_an_object');

            return false;
        }

        if (! array_key_exists('magnitude', $value)) {
            $this->messages[] = trans('validation.quantity.missing_magnitude');
        }

        if (! array_key_exists('units', $value)) {
            $this->messages[] = trans('validation.quantity.missing_units');
        }

        return $this->messages === [];
    }

    public function message()
    {
        return $this->messages;
    }
}
```

```php
class InvokableQuantity implements InvokableRule
{
    {
        if (! is_array($value)) {
            return $fail('validation.quantity.must_be_an_object')->translate();
        }

        if (! array_key_exists('magnitude', $value)) {
            $fail('validation.quantity.missing_magnitude')->translate();
        }

        if (! array_key_exists('units', $value)) {
            $fail('validation.quantity.missing_units')->translate();
        }
    }
}
```

### DIFF
--- a/src/Illuminate/Contracts/Validation/InvokableRule.php
+++ b/src/Illuminate/Contracts/Validation/InvokableRule.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Illuminate\Contracts\Validation;
+
+interface InvokableRule
+{
+    /**
+     * Run the validation rule.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @param  \Closure(string): \Illuminate\Translation\PotentiallyTranslatedString  $fail
+     * @return void
+     */
+    public function __invoke($attribute, $value, $fail);
+}
--- a/src/Illuminate/Translation/PotentiallyTranslatedString.php
+++ b/src/Illuminate/Translation/PotentiallyTranslatedString.php
@@ -32,7 +32,7 @@ class PotentiallyTranslatedString implements Stringable
      * Create a new potentially translated string.
      *
      * @param  string  $string
-     * @param \Illuminate\Contracts\Translation\Translator  $translator
+     * @param  \Illuminate\Contracts\Translation\Translator  $translator
      */
     public function __construct($string, $translator)
     {
--- a/src/Illuminate/Translation/PotentiallyTranslatedString.php
+++ b/src/Illuminate/Translation/PotentiallyTranslatedString.php
@@ -1,0 +1,89 @@
+<?php
+
+namespace Illuminate\Translation;
+
+use RuntimeException;
+use Stringable;
+
+class PotentiallyTranslatedString implements Stringable
+{
+    /**
+     * The string that may be translated.
+     *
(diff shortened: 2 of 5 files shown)


src/Illuminate/Translation/PotentiallyTranslatedString.php
```diff
@@ -0,0 +1,89 @@
...
+     */
+    protected $translator;
+
+    /**
+     * Create a new potentially translated string.
+     *
+     * @param  string  $string
+     * @param  \Illuminate\Contracts\Translation\Translator  $translator
+     */
+    public function __construct($string, $translator)
```
driesvints, line 37: I'd type-hint the contract here.

src/Illuminate/Contracts/Validation/InvokableRule.php
```diff
@@ -0,0 +1,16 @@
...
+{
+    /**
+     * Run the validation rule.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @param  \Closure(string): \Illuminate\Translation\PotentiallyTranslatedString  $fail
+     * @return void
+     */
+    public function __invoke($attribute, $value, $fail);
```
driesvints, line 15: Maybe best to type-hint the `Closure` here?

src/Illuminate/Validation/InvokableValidationRule.php
```diff
@@ -0,0 +1,183 @@
...
+     */
+    protected $data = [];
+
+    /**
+     * Create a new implicit or explicit Invokable validation rule.
+     *
+     * @param  \Illuminate\Contracts\Validation\InvokableRule  $invokable
+     * @return \Illuminate\Contracts\Validation\ImplicitRule
+     */
+    public static function make($invokable)
```
driesvints, line 54: I'd type-hint the contract here.

driesvints (changes requested):

PR #42690, opened 2022-06-07
[9.x] Add ability to define "with" relations as a nested array

timacdonald:
## Purpose

This PR aims to add the ability to specify relationship "withs" as a PHP array, rather than dot notation.

## Usage

```php
// before...

User::with([
    'avatar',
    'posts.tags',
    'posts.author',
    'posts.featureImage',
    'posts.comments.tags' => fn ($q) => $q->latest(),
])->get();

// after...

User::with([
    'avatar',
    'posts' => [
        'tags',
        'author',
        'featureImage',
        'comments' => [
            'tags' => fn ($q) => $q->latest(),
        ],
    ],
])->get();

```

### Advanced

```php
// supports chaining...

User::with([
    'posts' => fn ($q) => $q->withCount('comments')->with([
        'comments' => [
            'tags',
        ],
    ]),
])->get();

// supports specifying attributes inline, although I think with this notation you
// should probably be specifying selects via a constraint.

User::with([
    'posts:id,title,user_id' => [
        'comments:id,content,post_id' => [
            'tags',
        ],
    ],
])->get();

// supports mixing notation. Not saying you should, just saying you can 🙈

User::with([
    'posts' => [
        'comments',
    ],
    'posts.image',
])->get();

// merges constraints from different notation - again, not saying you should. just saying you can.

User::with([
    'posts.comments' => fn ($q) => $q->with('tags'),
    'posts:id,title,user_id' => [
        'comments' => fn ($q) => $q->withCount('tags'),
    ],
])->get();


```

## What isn't supported

I should note that merging of selects from different notation is not supported. I don't see this as a problem, as the current dot notation suffers from the same constraint...

```php
// does not select "id" and "name"
User::query()
    ->with(['users:id,name'])
    ->with(['users:email'])
    ->get();

// nor does this...

User::query()
    ->with(['users' => fn ($q) => $q->select(['id', 'name'])])
    ->with(['users:email'])
    ->get();
```

and so the following is explicitly not supported...

```php
User::with([
    'posts.comments' => fn ($q) => $q->addSelect(['id']),
    'posts' => [
        'comments' => fn ($q) => $q->addSelect(['post_id'])
    ],
])->get();

User::with([
    'posts' => [
        'comments:post_id,title'
    ],
    'posts.comments:id',
])->get();
```

or any other mixture of notation with duplicate selects.

## Notes

I'd love to get some eyes on this from a range of people who are deep in Eloquent land.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1430,22 +1430,13 @@ class Builder implements BuilderContract
      */
     protected function parseWithRelations(array $relations)
     {
+        if ($relations === []) {
+            return [];
+        }
+
         $results = [];
 
-        foreach ($relations as $name => $constraints) {
-            // If the "name" value is a numeric key, we can assume that no constraints
-            // have been specified. We will just put an empty Closure there so that
-            // we can treat these all the same while we are looping through them.
-            if (is_numeric($name)) {
-                $name = $constraints;
-
-                [$name, $constraints] = str_contains($name, ':')
-                            ? $this->createSelectWithConstraint($name)
-                            : [$name, static function () {
-                                //
-                            }];
-            }
-
+        foreach ($this->prepareNestedWithRelationships($relations) as $name => $constraints) {
             // We need to separate out any nested includes, which allows the developers
             // to load deep relationships using "dots" without stating each level of
             // the relationship with its own key in the array of eager-load names.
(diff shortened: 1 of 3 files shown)


src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -1457,6 +1448,91 @@ protected function parseWithRelations(array $relations)
...
+        return $preparedRelationships;
+    }
+
+    /**
+     * Combine an array of constraints into a single constraint.
+     *
+     * @param  array  $constriants
+     * @return \Closure
+     */
+    protected function combineContraints(array $constraints)
```
rodrigopedra, line 1510: Looks like a typo on the method's name

mateusjunges: Also in the php doc, it should be `$constraints`

timacdonald: Thanks peeps! Fixed.

OzanKurt:
Does this also work on `->load()`?

PR #42693, opened 2022-06-07
[9.x] Add trashed() method phpDoc for Eloquent factory

adelf:
This will improve IDE completion.


jasonmccreary:
Worth noting not every factory has the `trashed` method.

adelf:
@jasonmccreary agree, but I think it's better to have it in completion...

jasonmccreary:
Sure. Just putting something out on the webs for the poor developer who might try to use the `trashed` state (learned via autocompletion) for a model that isn't actually using the `SoftDeletes` trait. 😅

PR #42699, opened 2022-06-07
[9.x] Add `contains_with`, `doesnt_contains_with` in validation rules

michaelnabil230:
This PR adds the ability to check if the string contains the given values, and I think this validation is important and makes the validation easier and we will not resort to using something else such as Regex.

```php
$request->validate([
    'comment' => ['required', 'contains_with:f***,s***'],
]);
```

```php
$request->validate([
    'comment' => ['required', 'doesnt_contains_with:other words'],
]);
```


michaelnabil230:
Can be added mask string in to message for hidden the words contains in input. Can making that if the idea is good

taylorotwell:

PR #42700, opened 2022-06-07
[9.x] Add rememberUntil method to cache

mehdirajabi59:
cached data may also be invalidated according to some dependency changes. For example, if we are caching the content of some table or file and the table is changed, we should invalidate the cached copy and read the latest content from the table instead of the cache.

for example we want get users that have greater than two like and cache result it then when users update so cache will be expired and restore cache
```
$users = cache()->rememberUntil(
        'user_like',
        fn() => User::select('name', 'email')->where('like','>', '2')->get(),
        fn() => cache('user_like')->max('updated_at') == User::where('like', '>', 2 )->max('updated_at')
    );
```

### DIFF
--- a/src/Illuminate/Cache/Repository.php
+++ b/src/Illuminate/Cache/Repository.php
@@ -505,6 +505,35 @@ class Repository implements ArrayAccess, CacheContract
     }
 
     /**
+     * Get an item from the cache, or execute the given Closure and store the result until dependency is false.
+     *
+     * @param $key
+     * @param $value
+     * @param $dependency
+     * @return mixed
+     */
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can already do this by incorporating the last update timestamp of the model / file into the cache key itself... "user_like_{timestamp}"

mehdirajabi59:
> You can already do this by incorporating the last update timestamp of the model / file into the cache key itself... "user_like_{timestamp}"

@taylorotwell We already can do this with random key but have limit on duration and key of cache now we can cache forever until closure returns false with static key of cache.


model cache is example we might want cache that needs cache forever and expired with closure in any problems

PR #42701, opened 2022-06-07
[9.x] Unless validation rule

msmahon:
https://github.com/laravel/framework/discussions/42684

The `Conditionable` trait includes a `when()` method for applying a callback if the given value is (or resolves to) truthy. It also provides an `unless()` method that acts in an opposite fashion. This change applies that same paradigm to the Rule class for use in request validation.

## Context

Existing "When" method:

```php
request()->validate([
    'credential' => [
        'min:5',
        Rule::when($isNotAdmin, ['required'])
    ],
]);
```

Proposed "Unless" method:
```php
request()->validate([
    'credential' => [
        'min:5',
        Rule::unless($isAdmin, ['required'])
    ],
]);
```

I think this rule's existence is intuitive (as it was for me).

### DIFF
--- a/src/Illuminate/Validation/ConditionalRules.php
+++ b/src/Illuminate/Validation/ConditionalRules.php
@@ -40,7 +40,7 @@ class ConditionalRules
      * @param  callable|bool  $condition
      * @param  array|string|\Closure  $rules
      * @param  array|string|\Closure  $defaultRules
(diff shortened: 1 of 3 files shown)


msmahon:
@taylorotwell I don't know what you mean but ok!

PR #42702, opened 2022-06-07
[9.x] Define default Model Validation Rules directly in Model

hmennen90:
This PR adds a property to GuardsAttributes.php => rules. This can be retrieved using getRules() Method.

Additionally Rules.php got a Model Method expecting a string or instance of Illuminate\Database\Eloquent\Model.

Within the rules Property in Eloquent Model, default Validation Rules can be set which are retrieved by Rule::model Method.

For Example, instead of:

```
<?php 

class StoreRequest extends Request
{
public function rules() 
{
return [
'name' => ['required','string','max:255']
];
}
}

class UpdateRequest extends Request
{
public function rules() 
{
return [
'name' => ['required','string','max:255']
];
}
}

```

You can now easily do this:

```
<?php

/* namespace App\Models; */

class MyModel extends Model
{
protected $rules = [
'name' => ['required', 'string', 'max:255'
];
}


class StoreRequest extends Request
{
public function rules() 
{
return Rule::model(MyModel::class);
}
}

class UpdateRequest extends Request
{
public function rules() 
{
return Rule::model(MyModel::class);
}
}
```

This is covered by two Checks: 

- If a new Model Instance is used as Argument, it directly returns the result of getRules().
- If the class exists, which is submitted as string and it extends the Illuminate\Database\Eloquent\Models it instantiate the Model and returns the result of getRules().

If a Modelclass was not found, a ValidationException is thrown with declarative Message, containing the Model Name which cannot be resolved for Validation Rules.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/GuardsAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/GuardsAttributes.php
@@ -12,6 +12,13 @@ trait GuardsAttributes
     protected $fillable = [];
 
     /**
+     * The attributes default Validation Rules
+     *
+     * @var array<string>
+     */
+    protected $rules = [];
+
+    /**
      * The attributes that aren't mass assignable.
      *
      * @var array<string>|bool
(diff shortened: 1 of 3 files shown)


hmennen90:
@browner12 yes - but some fields can always be equal. Also you may use these Rules in different Controllers / Requests.

PR #42711, opened 2022-06-08
[9.x] Allow the "order by field()" use, in MySQL

fgaroby:
This PR adds the "order by FIELD()"  sorting, in MySQL.

Currently, if we want to order a query with the `FIELD()` method, we have to create a raw query, but it's not portable (if we decide to migrate from MySQL to another database).
With this PR, we have a new method `orderByField()`, which works with MySQL.
Use with MySQL:
```php
$builder->select('*')->from('users')->orderByField('age', [5, 7, 9]);
// select * from `users` order by field(`age`, "5", "7", "9")
```

Of course, if we use another database than MySQL, this method works too!
The same query, but with another database:
```php
$builder->select('*')->from('users')->orderByField('age', [5, 7, 9]);
// select * from "users" order by case when "age"="5" then 1 when "age"="7" then 2 when "age"="9" then 3 else 0'
```

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -2243,33 +2243,6 @@ class Builder implements BuilderContract
     }
 
     /**
-     * Add an "order by field()" clause to the query.
-     *
-     * @param  \Closure|\Illuminate\Database\Query\Builder|\Illuminate\Database\Query\Expression|string  $column
-     * @param  array  $field
-     * @return $this
-     *
-     * @throws \InvalidArgumentException
-     */
-    public function orderByField($column, $field)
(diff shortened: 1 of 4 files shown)


PaolaRuby:
Question
Default value on `else`, shouldn't it be zero?
```
// select * from "users" order by case when "age"="5" then 1 when "age"="7" then 2 when "age"="9" then 3 else 0
```

fgaroby:
Yes, you're right.
I'm updating my code, to have 0 as `else` value, and other stuff.

PaolaRuby:
`order by field()` also supports direction, (`DESC`, `ASC`)

taylorotwell:

PR #42712, opened 2022-06-08
[9.x] Allow Carbon Period for HavingBetween QueryBuilder

Joorren:
Allow a Carbon Period as the second parameter in the HavingBetween Query Builder method, implemented the same as for the WhereBetween method.

Preview of the [whereBetween](https://github.com/Joorren/framework/blob/48ef2a24eef2931798453d11d50541b0a489a525/src/Illuminate/Database/Query/Builder.php#L1184) method:

```php
    /**
     * Add a where between statement to the query.
     *
     * @param  string|\Illuminate\Database\Query\Expression  $column
     * @param  iterable  $values
     * @param  string  $boolean
     * @param  bool  $not
     * @return $this
     */
    public function whereBetween($column, iterable $values, $boolean = 'and', $not = false)
    {
        $type = 'between';

        if ($values instanceof CarbonPeriod) {
            $values = $values->toArray();
        }

        $this->wheres[] = compact('type', 'column', 'values', 'boolean', 'not');

        $this->addBinding(array_slice($this->cleanBindings(Arr::flatten($values)), 0, 2), 'where');

        return $this;
    }
```

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -2151,14 +2151,18 @@ class Builder implements BuilderContract
      * Add a "having between " clause to the query.
      *
      * @param  string  $column
-     * @param  array  $values
+     * @param  iterable  $values
      * @param  string  $boolean
      * @param  bool  $not
      * @return $this
      */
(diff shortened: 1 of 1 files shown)


driesvints:
This is a breaking change to the method signature. Please send this to master.

Issue #42691, opened 2022-06-07
GetChanges() on model fails when data stored in MySQL json type


sfilzek:
- Laravel Version: 9.16.0
- PHP Version: PHP 8.1.3
- Database Driver & Version: MySQL 5.7.24

### Description:
When updating a model which uses a mysql json column, the getChanges() call on the model always returns the json data as having changed even when it has not.

I believe this is because mysql re-orders the json data so that keys are sorted from shortest to longest and then alphabetically, and this causes the comparison to fail in Illuminate\Database\Eloquent\Concerns\HasAttributes where originalIsEquivalent() using castAttribute() performs an === equivalence check to see if the two arrays match. Even if the arrays contain the same data, if the keys have a different order then the test fails.

I'm not sure if this is just a mysql thing or if other databases also mess with the order of the stored json data.

## Model
```
namespace App\Models;

use Illuminate\Database\Eloquent\Model;

class JsonTest extends Model
{
    public $timestamps = false;
    protected $casts = [
        'sample_data' => 'array',
    ];    
}
```
## Migration
```
use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class extends Migration
{
    public function up()
    {
        Schema::create('json_tests', function (Blueprint $table) {
            $table->increments('id');
            $table->json('sample_data');
        });
    }
};
```
## Test
```
$sampleData = [
    'aa' => 1, // Note the ordering.
    'b' => 2   // Mysql will reorder this element to be first
];

$jsonTest = new App\Models\JsonTest;
$jsonTest->sample_data = $sampleData;
$jsonTest->save();

$jsonTest2 = App\Models\JsonTest::find(1);

$jsonTest2->sample_data = $sampleData;
$jsonTest2->save();
dump($jsonTest2->getChanges());
```

This will output:
```
[
    "sample_data" => "{"aa":1,"b":2}",
]
```

eusonlito:
Same with PostgreSQL, but PostgreSQL store data as binary, then you can not know how a JSON will be returned. Only array order is preserved on store.

hmennen90:
@sfilzek @eusonlito I've opened a PR after your Steps to Reproduce - the getChanges() Array is empty as expected. (SQLite)

Additionally I will setup tests using PostgreSQL and MYSQL too.

hmennen90:
Testing at MySQL, i can confirm this.

[...]

PR #42715, opened 2022-06-08
[9.x] Updating JSON Column with identical Data affecting changed Data #42691

hmennen90:
As mentioned in https://github.com/laravel/framework/issues/42691 this PR adds a Test on SQLite, MySQL and PGSQL to be sure, that Updating a JSON Column with identical Data does not appear as a change.


taylorotwell:
Why did you make the change from === to ==?

hmennen90:
@taylorotwell I found this:

https://stackoverflow.com/questions/59045471/how-to-prevent-the-json-data-output-being-alphabetically-sorted-and-stored-in-my

My Solution is pushed

taylorotwell:
There's now way I'm changing other methods like that - things will break.

PR #42717, opened 2022-06-08
[9.x] Adds `compileReadonly` function

mahmoudmohamedramadan:
This PR adds `compileReadonly` function

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php
@@ -337,4 +337,15 @@ trait CompilesConditionals
     {
         return "<?php if{$condition}: echo 'disabled'; endif; ?>";
     }
+
+    /**
+     * Compile a readonly block into valid PHP.
+     *
+     * @param  string  $condition
+     * @return string
+     */
+    protected function compileReadonly($condition)
+    {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Hmm - I wonder how many more HTML attributes we would possibly add directives for 😨  I'm worried we will get too many.

DarkGhostHunter:
`@if($condition) readonly @endif`.

If there is minimal keystrokes saved without tackling verbosity, I reject.

At this point, I think is just better to create a package to handle forms separately.

mahmoudmohamedramadan:
@taylorotwell maybe you are right but actually, not all HTML attributes will be rendered to directives because the attributes that must be rendered must have a common use like `@readonly` which is similar to `@disabled` it's common to check about something to apply these attributes So, it'll be nice to render these attributes to directives in Laravel because Laravel tries to make everything awesome and easier.

mahmoudmohamedramadan:
@DarkGhostHunter the same thing can be applied to `@disabled` and `@selected` because we can use `@if($condition) disabled @endif` also the same thing with `@selected` but I see that it'll be nice if we use minimal keystrokes to handle these scenarios.

taylorotwell:
I looked over the attributes - looks like there isn't much of anything remaining after this.

PR #42723, opened 2022-06-09
[9.x] Add new method `lockInProduction` in Route

michaelnabil230:
This PR adds a new method in routes that makes sure that the application is in development mode or not and will help many developers because they are doing routes for experimentation and raising the project to the public without the be prohibited. and will also help developers not to make mistakes in any way.

Before
```php
Route::get('/run-artisan-command', function () {
    Artisan::call('migrate:fresh');
    return 'Done';
});

Route::get('/run-artisan-command', function () {
    if (app()->environment('local')) {
        Artisan::call('migrate:fresh');
        return 'Done';
    }
});
```

After
```php

Route::prefix('dev')->lockInProduction()->group(base_path('routes/dev.php'));

Route::get('/run-artisan-command', function () {
    Artisan::call('migrate:fresh');
    return 'Done';
})->lockInProduction();
```

Note:
The code needs a lot of development to be good, but it's only the beginning.

### DIFF
--- a/src/Illuminate/Routing/Matching/IsLockInProductionValidator.php
+++ b/src/Illuminate/Routing/Matching/IsLockInProductionValidator.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Illuminate\Routing\Matching;
+
+use Illuminate\Http\Request;
+use Illuminate\Routing\Route;
+
+class IsLockInProductionValidator implements ValidatorInterface
+{
+    /**
+     * Validate a given rule against a route and request.
+     *
+     * @param  \Illuminate\Routing\Route  $route
+     * @param  \Illuminate\Http\Request  $request
+     * @return bool
+     */
+    public function matches(Route $route, Request $request)
+    {
+        return ! $route->getLockInProduction();
+    }
+}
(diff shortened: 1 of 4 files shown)


mateusjunges:
Why not just create a new route file and egister it in local environment only?

```php
if ( app()->environment('local')) {
    Route::middleware('api')
        ->prefix('dev')
        ->group(base_path('routes/dev.php'));
}

**Comment by michaelnabil230** [comment]
> Why not just create a new route file and egister it in local environment only?
> 
> ```
> if (app()->environment('local')) {
>     Route::prefix('dev')->group(base_path('routes/dev.php'));
> }
> ```

This will be a more conceptual abbreviation. And also some developers are lenient in writing the code. 
And also can be written the same function thus:
```php
Route::prefix('dev')->lockInProduction()->group(base_path('routes/dev.php'));
```

driesvints:
I don't think we should add an extra safe guard here. Just remove the routes again which you don't need. We can't protect people from themselves.

dennisprudlo:
You could also use the check outside the route definition so the route does not even exist in production. This is how I add my "playground" routes:

```php
// routes/web.php

// application routes...

if (app()->isLocal()) {
    Route::get('/mailable', fn () => new \App\Mail\Users\SignedUp(App\Models\User::find(1)));
    Route::view('playground', 'playground');
    Route::get('/abort/{status}', fn (string $status) abort($status));
}
```

michaelnabil230:
> I don't think we should add an extra safe guard here. Just remove the routes again which you don't need. We can't protect people from themselves.

Do you think it is useless?

PR #42726, opened 2022-06-09
Eloquent Expose morphType in MorphToMany

vFouailly:
When working with legacy db, one sometimes need to adapt to existing column names, some are already changeable in eloquent, but morphType was not.
This PR propose to expose the type custom column name just like other fields (namely foreignPivotKey, relatedPivotKey, parentKey and relatedKey) when doing $this->morphedByMany()
Existing use isn't broken because the parameter is optional in each context and stands at the end of the function/class signature
It makes Eloquent more flexible when subject DB cannot be modified easily (think legacy db with production project running on it) and allows for more progressive integration of Laravel into older projects


driesvints:
There's unfortunately, breaking changes to method signatures in this PR. Please send to master.

PR #42731, opened 2022-06-09
[9.x]  Use Arrow Function

mehdirajabi59:
Use Arrow Function in CacheServiceProvider.php

### DIFF
--- a/src/Illuminate/Cache/CacheServiceProvider.php
+++ b/src/Illuminate/Cache/CacheServiceProvider.php
@@ -15,27 +15,19 @@ class CacheServiceProvider extends ServiceProvider implements DeferrableProvider
      */
     public function register()
     {
-        $this->app->singleton('cache', function ($app) {
-            return new CacheManager($app);
-        });
+        $this->app->singleton('cache', fn ($app) => new CacheManager($app));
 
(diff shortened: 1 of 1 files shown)


src/Illuminate/Cache/CacheServiceProvider.php
```diff
@@ -15,27 +15,18 @@ class CacheServiceProvider extends ServiceProvider implements DeferrableProvider
...
 
-        $this->app->singleton('cache.psr6', function ($app) {
-            return new Psr16Adapter($app['cache.store']);
-        });
+        $this->app->singleton('cache.psr6', fn ($app) => new Psr16Adapter($app['cache.store']));
 
-        $this->app->singleton('memcached.connector', function () {
-            return new MemcachedConnector;
-        });
+        $this->app->singleton('memcached.connector', new MemcachedConnector);
```
driesvints, line 24: You're changing the behavior here from deferring the creation of the connector to immediate instantiation. Please re-add the closure (or short closure)

driesvints (changes requested):

taylorotwell:
I think it's fine as is if it's working.

PR #42733, opened 2022-06-09
[9.x] Add authorize function to Authorizable trait

zackrowe:
This PR adds an `authorize` function to the `Authorizable` trait that mirrors the `authorize` function found on the `Gate` class. This provides the convenience of being able to throw an authorization exception from an authorizable entity without using the Gate facade.

Before:
```
Gate::forUser($user)->authorize('update-post', $post);
```
After:
```
$user->authorize('update-post', $post);
```

This update doesn't change or remove any existing functionality just adds another convenient way to perform an authorization check.


taylorotwell:
So - I like it! But... I'm worried if people have already added this method to their user model this could be a breaking change. 🫠

So, for now, I think I'll have to leave it to you to define this on your own User model.

PR #42737, opened 2022-06-09
[9.x] Add ability to set backoff in broadcast events

jpbalda:
This PR allows to set a backoff to broadcast events. This is useful when the websockets server is unstable and you want to give the server some time to recover.

### DIFF
--- a/src/Illuminate/Broadcasting/BroadcastEvent.php
+++ b/src/Illuminate/Broadcasting/BroadcastEvent.php
@@ -36,6 +36,13 @@ class BroadcastEvent implements ShouldQueue
     public $timeout;
 
     /**
+     * The number of seconds to wait before retrying a job that encountered an uncaught exception.
+     *
+     * @var int
+     */
+    public $backoff;
+
+    /**
      * Create a new job handler instance.
      *
      * @param  mixed  $event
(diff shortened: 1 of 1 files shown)


taylorotwell:
How would this be customized by the end user?

jpbalda:
You can customize it in the broadcast event class as a public property. Currently, you can only set `$tries` and `$timeout` properties, so the PR is adding another property to customize the job behavior.

Using the docs example, it'd look something like this:

```php
<?php
 
namespace App\Events;
 
use App\Models\User;
use Illuminate\Broadcasting\Channel;
use Illuminate\Broadcasting\InteractsWithSockets;
use Illuminate\Broadcasting\PresenceChannel;
use Illuminate\Broadcasting\PrivateChannel;
use Illuminate\Contracts\Broadcasting\ShouldBroadcast;
 
class ServerCreated implements ShouldBroadcast
{
    use SerializesModels;
 

    public $tries = 3;

    public $backoff = 15; // <--- new property


    /**
     * The user that created the server.
     *
     * @var \App\Models\User
     */
    public $user;
 
    /**
     * Create a new event instance.
     *
     * @param  \App\Models\User  $user
     * @return void
     */
    public function __construct(User $user)
    {
        $this->user = $user;
    }
 
    /**
     * Get the channels the event should broadcast on.
     *
     * @return Channel|array
     */
    public function broadcastOn()
    {
        return new PrivateChannel('user.'.$this->user->id);
    }
}
```

jpbalda:
Thanks for merging my first PR 😃

PR #42741, opened 2022-06-09
[9.x] Add Password Generator that Adheres to Validation Rules

aidan-casey:
Hey there!

This pull request adds the static `make` and non-static `create` methods to `Illuminate\Validation\Rules\Password`. These generate a password that adheres to the validation rules specified (in the case of `make`, the default validation rules are used).

Two primary use cases come to mind for this functionality: 1) The ability to generate adherent passwords for tests and 2) the ability to generate adherent passwords on behalf of the user (e.g., suggest a password when changing it).


johanrosenson:
Neat, but it feels very wrong that a validation rule also serves as a password generator, IMO one class should not have both these responsibilities.

aidan-casey:
> Neat, but it feels very wrong that a validation rule also serves as a password generator, IMO one class should not have both these responsibilities.

@johanrosenson - I definitely don't disagree. I'm open to alternative suggestions, though it may be determined that it is better housed outside of the framework at that point.

src/Illuminate/Validation/Rules/Password.php
```diff
@@ -279,6 +290,50 @@ public function rules($rules)
...
+     * @param  int  $length
+     * @return string
+     */
+    public function create(int $length = 8): string
+    {
+        $length = ($length >= $this->min) ? $length : random_int($this->min, $this->min + 10);
+
+        $password = [];
+
+        $upper = ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'X', 'Y', 'Z'];
```
ankurk91, line 305: You can generate uppercase letters like

```php
$upper = array_reduce(
            range('A', 'Z'),
            fn ($p, $n) => $p.$n
        );
```

aidan-casey: @ankurk91 - Fair enough. I'm not sure it really matters in the grand scheme of things, but I believe you just need `range('A', 'Z')` to generate the array.

BrandonSurowiec: Works for lowercase and numbers as well.

```
range('a', 'z');
range(0, 9);
```

taylorotwell:
If this should exist at all, it feels like it should maybe be on the `Str` class? But, in general, I'm worried it would encounter PR bloat with lots of PRs for various options and customizations.

aidan-casey:
@taylorotwell - That makes total sense.

Would you be open to a PR that implements this as a method on the `Str` class and by default uses the most secure option (mixed case, letters, and symbols) but allows for a factory to be defined like `createRandomStringsUsing` does? Or is this just something that poses no real value?

PR #42742, opened 2022-06-09
[9.x] Add invokable option to make rule command

snarcraft:
Following the PR https://github.com/laravel/framework/pull/42689, this new PR adds an option to make an invokable rule to the existing make rule command.

This minor feature comes with no breaking change, the previous command will still work as usual.
For the end users, this additional option can be useful to generate an invokable rule.

### DIFF
--- a/src/Illuminate/Foundation/Console/stubs/rule.invokable.implicit.stub
+++ b/src/Illuminate/Foundation/Console/stubs/rule.invokable.implicit.stub
@@ -9,7 +9,7 @@ class {{ class }} implements InvokableRule
     /**
      * Indicates whether the rule should be implicit.
      *
-     * @var bool|null
+     * @var bool
      */
     public $implicit = true;
 
--- a/src/Illuminate/Foundation/Console/stubs/rule.invokable.implicit.stub
+++ b/src/Illuminate/Foundation/Console/stubs/rule.invokable.implicit.stub
@@ -1,0 +1,28 @@
+<?php
+
+namespace {{ namespace }};
+
+use Illuminate\Contracts\Validation\InvokableRule;
+
+class {{ class }} implements InvokableRule
+{
+    /**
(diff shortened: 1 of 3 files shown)


timacdonald:
@taylorotwell I do feel that making these the "default" is the way to go. It creates consistency across the rule APIs (Invokable class / Closure based), they have better DX for for more complicated rules (as shown in the PR), and is one less decision for developers to make when creating rules e.g. "which rule type do we use".

taylorotwell:
Should we be handling the combination of `--invokable` and `--implicit`?

timacdonald:
Yea, for 9.x I think we should handle both of these. @snarcraft, if you have the space, could you add support for the following....

```
artisan make:rule --invokable
artisan make:rule --invokable --implicit
```

If you don't have space right now, just let me know and I can jump in for ya.

Will convert this one to a draft until we have it ready.

src/Illuminate/Foundation/Console/stubs/rule.invokable.implicit.stub
```diff
@@ -0,0 +1,28 @@
...
+namespace {{ namespace }};
+
+use Illuminate\Contracts\Validation\InvokableRule;
+
+class {{ class }} implements InvokableRule
+{
+    /**
+     * Indicates whether the rule should be implicit.
+     *
+     * @var bool|null
```
timacdonald, line 12: Although I understand why you've put `null` in the type, if this property exists, I don't think it makes sense to indicate that it is `null`. It should only be `true` or `false` if the property exists.

Yes, `null` is supported, but it isn't intended to be set to `null`.

```suggestion
     * @var bool
```

snarcraft: I agree with you
Actually what I had in mind is that : "this property is not required", I was trying to express that :)

PR #42746, opened 2022-06-10
[9.x] Fix incorrect PHPDoc syntax in `Collection`

devfrey:
Adds missing `>` to the PHPDocs for `Collection::sortBy()`, `sortByMany()`, and `sortByDesc()`.


driesvints:
It would seem the docblock here is correct after all judging from the failing types test.

devfrey:
@driesvints The PHPDoc syntax is incorrect and therefore ignored by PHPStan. The failing tests now only show up because the PHPDoc was never correct.

driesvints:
Good remark 👍 

Can you fix the tests?

PR #42752, opened 2022-06-10
Uses `Conditionable` trait

mahmoudmohamedramadan:
I think it's important to use the `Conditionable` trait because we may need to check some values in the request to perform some actions.

**Old version**

```PHP
if ($request->get('key') == 'some_value') {
    // perform actions here...
}
```

**New version**

```PHP
$request->when($request->get('key') == 'some_value', function () {
    // perform some actions here...
});
```

### DIFF
--- a/src/Illuminate/Http/Request.php
+++ b/src/Illuminate/Http/Request.php
@@ -8,6 +8,7 @@ use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Session\SymfonySessionDecorator;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
+use Illuminate\Support\Traits\Conditionable;
 use Illuminate\Support\Traits\Macroable;
 use RuntimeException;
 use Symfony\Component\HttpFoundation\Exception\SessionNotFoundException;
@@ -25,7 +26,8 @@ class Request extends SymfonyRequest implements Arrayable, ArrayAccess
     use Concerns\InteractsWithContentTypes,
         Concerns\InteractsWithFlashData,
         Concerns\InteractsWithInput,
-        Macroable;
+        Macroable,
+        Conditionable;
 
     /**
      * The decoded JSON content for the request.


antonkomarev:
What's wrong with if?

antonkomarev:
With new syntax we need to pass variables inside of the function with `use`

mahmoudmohamedramadan:
I think it's no problem to pass the variable to the closure and the `when` function is readable more than the `if` condition for illuminate see the next example

**Using `if` condition**

```PHP
if ($request->get('key') == 'some_value') {
    collect($array)->filter(fn ($el) => dd($el));
} elseif ($request->get('key') == 'other_value') {
    collect($array)->filter(fn ($el) => dd($el));
}
```

**Using `when` function**

```PHP
$request
    ->when($request->get('key') == 'some_value', function () use ($array) {
        collect($array)->filter(fn ($el) => dd($el));
    })
    ->when($request->get('key') == 'other_value', function () use ($array) {
        collect($array)->filter(fn ($el) => dd($el));
    });
```

for me, I see that the `when` function is better than `if` condition as in the previous scenario.

PaolaRuby:
Maybe yo have to add some tests

taylorotwell:

antonkomarev:
@mahmoudmohamedramadan you can make request class in your app with this trait and inherit it

PaolaRuby:
Also you can make a macro

mahmoudmohamedramadan:
@PaolaRuby Hmm- I think If I create a class that extends the `src/Illuminate/Http/Request.php` then I use the `Conditionable ` trait is easier, like so

```PHP
<?php

namespace App\Classes;

use Illuminate\Http\Request;
use Illuminate\Support\Traits\Conditionable;

class MyCustomRequest extends Request
{
    use Conditionable;
}
```

then I can type-hint my `$request` variable with this class, like so

```PHP
/**
 * Store a newly created resource in storage.
 *
 * @param \App\Classes\MyCustomRequest $request
 */
public function store(MyCustomRequest $request)
{
    $request->when(true == true, function () {
        // perform something here...
    });

    return redirect()->back();
}
```

PR #42755, opened 2022-06-10
Add `WithPolicies, WithEvents, WithRoutes` service provider traits

HanakJakub:
Suggestion about moving code from: 
- `AuthServiceProvider` to `WithPolicies`
- `EventServiceProvider` to `WithEvents`
- `RouteServiceProvider` to `WithRoutes`

This could help to build providers more easily especially providers like module, domain, support service providers. 

Currently to make those kind of providers you need to create {Module}EventServiceProvider, {Module}RouteServiceProvider etc. and the only way is by extending main `RouteServiceProvider`, `EventServiceProvider`, `AuthServiceProvider`. 

Also currently if provider doesn't extends `EventServiceProvider` it won't be discovered (unless automatic event discovery is enabled)

With provider traits we could be making service providers like this:
```php

// Have main provider with the logic we need
class ModuleServiceProvider extends \Illuminate\Support\ServiceProvider
{
   use WithEvents;
   use WithRoutes;
   use WithPolicies;
}

// Have e.g. domain level service provider that only need to register events
class DomainServiceProvider extends \Illuminate\Support\ServiceProvider
{
   use WithEvents;
}

```


taylorotwell:

PR #42759, opened 2022-06-11
[8.x] Do not resolve already set headers

### LINKED ISSUE #42758: Custom file formats not supported
Re-opening this because it's an issue for every application that use non-standard filetypes.
https://github.com/laravel/framework/issues/41960#issuecomment-1152703807_

- Laravel Version: 9.8.1
- PHP Version: 8.1.3
- league/flysystem 3.0.16

### Description:
```
$headers["Content-Type"] = "text/plain";
$headers["Content-Length"] = $file_size;
$headers["Accept-Ranges"] = "bytes";
$headers["Content-Disposition"] = "attachment; filename={$id}.left";
$headers["Content-Range"] = "bytes " . 0 . "-" . ($file_size) . "/" . $file_size;
Storage::disk("localdisk")->download($id.".found", $headers);
```
Results in the following error:
```
local.ERROR: Unable to retrieve the mime_type for file at location: 737.found.  {"exception":"[object] (League\\Flysystem\\UnableToRetrieveMetadata(code: 0): Unable to retrieve the mime_type for file at location: 737.found.  at /var/www/hostname/vendor/league/flysystem/src/UnableToRetrieveMetadata.php:49)
```
I believe this is caused by me having a non-standard extension, which is fine since I provided the content-type in the headers explicitly. But instead I'm getting errors now with my recent upgrade from laravel 8 to laravel 9. Any way to make this work?

rodrigopedra:
Currently `Illuminate\Filesystem\FilesystemAdapter@response`, will implicitly calculate the `Content-Type`, `Content-Length`, and `Content-Disposition` default headers values, even when the user already provided those headers.

Also, the calculated default values end up not being used, as the default headers are merged to the user provided headers using the `+` operator, which preserve keys already present.

~~~php
$response->headers->replace($headers + [
    'Content-Type' => $this->mimeType($path),
    'Content-Length' => $this->size($path),
    'Content-Disposition' => $disposition,
]);
~~~

As outlined in issue #42758 , this can create an issue, for example, when a developer is trying to send a custom file extension, even when they provide a custom `Content-Type` header, as the `FilesystemAdapter@mimeType` method would be called anyways.

This PR:

- Checks if each of the  `Content-Type`, `Content-Length`, and `Content-Disposition` headers are not present before calculating their corresponding default
- Adds test cases to ensure the methods are not called when those headers are provided

Note: Per support policy this PR was sent to the 8.x branch, as the issue's reporter reports the issue as a bug.

If accepted, when merging this PR to the 9.x branch, the partial mock creation on the added test cases will need to be changed, as  the `FilesystemAdapter` constructor expects an extra adapter argument on the 9.x branch.

Short, when merging to 9.x, mock definitions like this...

~~~php
$files = m::mock(FilesystemAdapter::class, [$this->filesystem])->makePartial();
~~~

... will need to be changed to this:

~~~php
$files = m::mock(FilesystemAdapter::class, [$this->filesystem, $this->adapter])->makePartial();
~~~

No further changes will be needed on the test cases.

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemAdapter.php
+++ b/src/Illuminate/Filesystem/FilesystemAdapter.php
@@ -177,7 +177,7 @@ class FilesystemAdapter implements CloudFilesystemContract
      *
      * @param  string  $path
      * @param  string|null  $name
-     * @param  array|null  $headers
+     * @param  array  $headers
      * @param  string|null  $disposition
      * @return \Symfony\Component\HttpFoundation\StreamedResponse
      */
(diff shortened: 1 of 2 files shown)


rodrigopedra:
Closed as in 8.x the underlying Flysystem implementation on 8.x does not throw an Exception with a MimeType it can't guess.

PR #42763, opened 2022-06-11
[9.x] Removed unused parameter

bvtterfly:
This small PR removes the unused `$request` parameter from the `AuthorizesRequests` trait's `authorizeResource` method.


driesvints:
This is a breaking change unfortunately.

PR #42769, opened 2022-06-11
[9.x] Deprecated message fix

MrKampf:
Deprecated message on Str::length. Check is null than return 0;

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -438,6 +438,10 @@ class Str
         if ($encoding) {
             return mb_strlen($value, $encoding);
         }
+        
+        if (is_null($value)) {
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
👎 here. You are not allowed to call Str::length with `null` value. You MUST pass through only strings.

MrKampf:
Yeah but currently is allowed.

Screenshot_1730

MrKampf:
Now i changed to allowed only string type

ankurk91:
You should cast the $value to string,

MrKampf:
Yeah this one solution. But is it not better when people get a execption? With the "new" php verions, is it allowed to set fixed types in the function parameters, for situations like this.

Issue #42765, opened 2022-06-11
there is a difference between $request and request()


Ham3D:
- Laravel Version: 9.x
- PHP Version: 8.1
- Database Driver & Version:

### Description:
when you attach something to **$request** , you dont have access to it when you call **request()->get('something');**

### Steps To Reproduce:

There is a wired bug in the request.
I fetch something inside a policy class (which is loaded inside a FormRequest class)
imagine i have to:

    $chat = Chat:first();

and then inside my services i need to again get this `$chat`,  but because i have it in the first place, i do this inside policy [here we dont have access to **$request**, so i used **request()** helper]:

    request()->request->add(['chat' => $chat]);
or

    request()->merge(['chat' => $chat]);

the final result is the same (both buggy)

and inside my services, or in controller, etc, i do this:

    $chat = $request()->get('chat');    // this will be null
    $chat = request()->get('chat') // this one has the actual chat model

nielsdos:
I believe the issue is that you're using `$request()` as a function call instead of `$request` as a variable, in the first (buggy) example.

Ham3D:
> I believe the issue is that you're using `$request()` as a function call instead of `$request` as a variable, in the first (buggy) example.

no, the bug is there, give it a try!

i fixed the typo.

igorgaming:
+1.  I also have this bug. A workaround its a passing of the current $request to the service.

rodrigopedra:
Can you please provide a GitHub repository to demonstrate your issue.

You can use this command, if you have the Laravel Installer installed locally:

~~~bash
$ laravel new bug-report --github="--public"
~~~


Right now, I could not reproduce. Steps I did:

~~~bash
$ laravel new issue-42765
$ cd issue-42765
$ php artisan make:policy UserPolicy --model=User
$ php artisan serve
~~~

Then edit `./routes/web.php` to:

~~~php
<?php

use App\Models\User;
use Illuminate\Http\Request;
use Illuminate\Support\Facades\Auth;
use Illuminate\Support\Facades\Route;

// avoid setting up a database
Auth::login(User::factory()->makeOne()->forceFill(['id' => 1]));

Route::get('/', function (Request $request) {
    return [
        $request->get('chat'),
        request()->get('chat'),
        request('chat'),
    ];
})->can('viewAny', User::class);
~~~

And change `UserPolicy@viewAny` method to:

~~~php
public function viewAny(User $user)
{
    request()->merge(['chat' => now()->timestamp]);

    return true;
[...]

PR #42770, opened 2022-06-11
[8.x] Replace resolved form request in container

rodrigopedra:
Currently when a Form Request is resolved from the container the container's current `Request` instance is not replaced, thus if a developer typehints a FormRequest in a controller action and then from a service class they use the `request()` helper to access the request instead, they will get different request instances.

This is problematic, when a developer wants to merge new values, for example in the `FormRequest@passedValidation` and later wants to access this value from the request helper.

For example consider this form request:

~~~php
<?php

namespace App\Http\Requests;

class SampleRequest extends FormRequest
{
    {
        return [];
    }

    protected function passedValidation()
    {
        $this->merge(['foo' => 'bar']);
    }
}
~~~

And this route:

~~~php
Route::get('/', fn (SampleRequest $request) => request('foo'));
~~~

Currently this route would return `null` as the `request()` helper would be trying to get the field from the original request instance, not from the resolved form request.

This behavior was reported in issue #42765 

This PR:

- Replaces the `'request'` instance in the container after creating a new Form Request
- No tests were added as this is done inside the `FormRequestServiceProvider` and all `FormRequests` tests manually creates a new instance, so I could not figure out how to test it out in a simple manner.
-

### DIFF
--- a/src/Illuminate/Foundation/Providers/FormRequestServiceProvider.php
+++ b/src/Illuminate/Foundation/Providers/FormRequestServiceProvider.php
@@ -32,6 +32,7 @@ class FormRequestServiceProvider extends ServiceProvider
 
         $this->app->resolving(FormRequest::class, function ($request, $app) {
             $request = FormRequest::createFrom($app['request'], $request);
+            $app->instance('request', $request);
 
             $request->setContainer($app)->setRedirector($app->make(Redirector::class));
         });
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure this should be changed on a patch release. All sorts of container rebinding action happens when "rebinding" a request in the container. I'm also not sure I expect the request() helper to resolve form requests. To me the base HTTP request and a form request are two separate objects.

PR #38353, opened 2021-08-12
[8.x] Update return type for tap

driesvints:
`when` already has `$this|mixed` so this should be reflected in `tap` on `BuildsQueries`. If you look at the logic inside `when` you'll see that this is the correct return type as it's either the return value of the given callback or `$this`. A previous attempt was made here: https://github.com/laravel/framework/pull/32717

See a more thorough explanation here: https://github.com/laravel/framework/issues/38343

Issue #38343, opened 2021-08-11
[8.x] BuildsQueries::tap() doesn't always return $this despite being documented as such


axlon:
- Laravel Version: 8.53.0 (I first spotted this in 7.x, but it may have been around for longer)
- PHP Version: 8.0.8
- Database Driver & Version: n/a

### Description:

The `BuildsQueries::tap()` method is documented to return `$this`, but in reality does not always return `$this` or even a builder instance at all. This method calls (and returns) `Conditionable::when()` which returns the return value of the given callback if it is truthy, if the return value is not truthy it will return `$this`. 

This leads to situations where a query chain may break in a type error.

By contrast, this is the contents of `EnumeratesValues::tap()` (which is used by collections):

```php
    public function tap(callable $callback)
    {
        $callback(clone $this);

        return $this;
    }
```

Collection's implementation always return `$this` regardless of the callback's return value.

Another example is the global `tap()` helper:

```php
    function tap($value, $callback = null)
    {
        if (is_null($callback)) {
            return new HigherOrderTapProxy($value);
        }

        $callback($value);

        return $value;
    }
```

This function can also return a proxy when there is no callback, but when there _is_ a callback it's return value is always ignored and `$value` is returned.

(Previous PR about this issue #32717 on 7.x)

### Steps To Reproduce:

1. Create a new Laravel project and run the migrations
2. Paste the following snippet in the default route and open the project in your browser:

```php
// The first time you run this it should work without much issue,
// after confirming the code does not break uncomment the line below:
//
//    \Database\Factories\UserFactory::new()->createOne();

    \App\Models\User::query()
        ->tap(fn ($query) => $query->update(['name' => 'Taylor Otwell']))
        ->count();

    return 'It works!';
```

### How to fix:

I see two ways to fix this:
- Accept that this is how this method works, in which case I believe it's return type should be adjusted to include `mixed` (`$this` is part of mixed, but making it mixed only would break auto completion for a lot of people)
- Fix this method to be consistent with other versions of tap, while it "technically" would not be a breaking change as this function should only return `$this` as is, it would probably be best to make this change in 9.x to avoid breaking any projects relying on current behavior

Personally I think the second fix would be the way to go, but I'm willing to provide a PR for either change if accepted.

driesvints:
Thanks for the thorough issue. I've sent in a PR with the first suggestion: https://github.com/laravel/framework/pull/38353

PR #36217, opened 2021-02-10
[8.x] Compatibility between EnumeratesValues and BuildsQueries on when() method

ImJustToNy:
This simple PR aims to ensure compatibility between `src/Illuminate/Collections/Traits/EnumeratesValues.php` and `src/Illuminate/Database/Concerns/BuildsQueries.php` on `when()` method. 

Before, using `when()` method on collection **without** returning value, lead to collection being null (user must always return the same collection to chain further methods). This is understandable, but when we're doing the same thing inside query builder, we may return nothing (null) and Laravel corrects us returning builder again:
https://github.com/laravel/framework/blob/a6680d98f9dadaa363aa7d5218517a08706cee64/src/Illuminate/Database/Concerns/BuildsQueries.php#L208-L210

This leads to confusion in my opinion.
Thank you for reading.

PR #42775, opened 2022-06-12
[8.x] Make BuildsQueries::tap consistent with other versions of tap

HazzazBinFaiz:
As previously mentioned on https://github.com/laravel/framework/issues/38343 and https://github.com/laravel/framework/pull/38353.

This PR makes `BuildsQueries::tap()` consistent with `EnumeratesValues::tap()` and `tap()`, meaning it will always return `$this` after executing the given callback. This change should not be breaking unless the developer was relying on undocumented behaviour (prior to https://github.com/laravel/framework/pull/38353).

This was merged in 9.x with https://github.com/laravel/framework/pull/38359 PR but still remains in 8.x.

### DIFF
--- a/src/Illuminate/Database/Concerns/BuildsQueries.php
+++ b/src/Illuminate/Database/Concerns/BuildsQueries.php
@@ -487,10 +487,12 @@ trait BuildsQueries
      * Pass the query to a given callback.
      *
      * @param  callable  $callback
-     * @return $this|mixed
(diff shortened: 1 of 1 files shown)


derekmd:
The previous PR targeted 9.x (unreleased at the time) because it was considered a breaking change for 8.x: https://github.com/laravel/framework/pull/36217

PR #42778, opened 2022-06-12
[9.x] Adding PHP DocBlock @mixin.

adideas:
The mixin DocBlock allows you to signal that all properties and methods of another class are available on the class where the mixin is applied upon.
### I think this will help a lot in learning or using Laravel!

```php
namespace App\Models;

class Thread extends Model
{

}
```

Before: 
alt text

After:
alt text
alt text
alt text

### I think this will help a lot in learning or using Laravel!


adideas:
I tested in PhpStorm.

ankurk91:
https://github.com/barryvdh/laravel-ide-helper

adideas:
What's the point of adding an additional dependency?  in a large laravel project this doesn't work.  this pr will tell the editor a bit about what methods exist.

adideas:
on click I will get to the ide helper and not to the vendor

src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -23,6 +23,9 @@
 use JsonSerializable;
 use LogicException;
 
+/**
+ * @mixin Builder
```
rodrigopedra, line 27: If you search for other `@mixin` usage throughout Laravel codebase, you will see every time a Full Qualified Class Name is used.

So I would change this to:

~~~php
/**
 * @mixin \Illuminate\Database\Eloquent\Builder
~~~

PR #42780, opened 2022-06-12
[9.x] Improve return value in Container

michael-rubel:
The method says it should return `\Closure|string|array|null`, but it returns nothing instead of `null` after checking aliases.
This PR replaces nothing with `null` in `getContextualConcrete` method.

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -820,7 +820,7 @@ class Container implements ArrayAccess, ContainerContract
         // given abstract type. So, we will need to check if any aliases exist with this
         // type and then spin through them and check for contextual bindings on these.
         if (empty($this->abstractAliases[$abstract])) {
-            return;
+            return null;
         }
 
         foreach ($this->abstractAliases[$abstract] as $alias) {


stevebauman:
Won't the return value still be `null` even with this explicit return?

There is currently no instance in the framework that I could find where there is an explicit `return null`, so I don't see this being merged -- as there is no precedent for this.

rodrigopedra:
`return;` actually returns `null`. 

Also not even returning anything, returns `null`, in case someone uses the return value.

Take a look at the output of the following sample code: (available at https://3v4l.org/aBeLC )

~~~php
<?php

function returnNull() { return null; }
function returnNothing() { return; }
function noExplicitReturn() { ; }

var_dump(array(
    is_null(returnNull()),
    is_null(returnNothing()),
    is_null(noExplicitReturn()),
));
~~~

All three comparisons return `true`.


As @stevebauman already said, this is the code style throughout Laravel's code base.

michael-rubel:
It returns `null` for nothing due to an undefined return type. If Laravel someday adds types to the functions as it is described in docblocks, it will fail if you return nothing.

rodrigopedra:
If someday they decide to add types, I am sure they will review every missing return, and also make all the tests pass.

But right now it works and is not their code style. And if it is not broken, why fix it?

PR #42785, opened 2022-06-13
[9.x] Vite

jessarcher:
:zap: Vite is a modern frontend build tool that provides an extremely fast development environment and bundles your code for production.

This PR introduces a `@vite` Blade directive to the framework that will be used in combination with the official [Laravel Vite Plugin](https://github.com/laravel/vite-plugin). This PR does not remove or deprecate the existing `mix()` helper. There are also PRs to [laravel/laravel](https://github.com/laravel/laravel/pull/5904) as well as [laravel/breeze](https://github.com/laravel/breeze/pull/158) and [laravel/jetstream](https://github.com/laravel/jetstream/pull/1068) that will make this the default for new Laravel applications. There is also, of course, [documentation](https://github.com/laravel/docs/pull/7984).

The Laravel side of the implementation is similar to Mix and should feel quite familiar.

The main difference is that instead of using the `mix()` helper, you would now use the `@vite` Blade directive. The reason for the directive is that Laravel needs to output complete `<script>` and `<link rel="stylesheet">` tags, rather than just a path as with Mix.

For the best experience with a Blade app, you will want to specify both CSS and JavaScript entry points, which would look something like this in your Blade layout template:

```diff
- <link rel="stylesheet" href="{{ mix('css/app.css') }}">
- <script src="{{ mix('js/app.js') }}" defer></script>
+ @vite(['resources/css/app.css', 'resources/js/app.js'])
```

In development mode, this will also load the Vite client, which will automatically reload your JavaScript and CSS when you make changes. In production mode, it will automatically load the versioned assets for you.

For the best experience with a SPA (e.g. Inertia), you will want to import your CSS via your JavaScript entry point(s) and then only specify the JavaScript entry point to the `@vite` directive, which would look something like this: 

```diff
  import './bootstrap.js';
+ import '../css/app.css';
```

```diff
- <link rel="stylesheet" href="{{ mix('css/app.css') }}">
- <script src="{{ mix('js/app.js') }}" defer></script>
+ @vite('resources/js/app.js')
```

When importing your CSS via JavaScript, Vite will automatically inject a `<style>` tag in development mode that it will keep up to date. When building for production, a compiled `app.{version}.css` file will be generated that the `@vite` directive will automatically load via a `<link rel="stylesheet">` tag.

When using React, you will also need to place a `@viteReactRefresh` directive above your `@vite` directive that will inject Vite's React Refresh Runtime.

Everything will be configured for you with Laravel's updated starter kits.

As with Mix, The `@vite` directive will automatically detect when the Vite development server is running, otherwise, it will use the compiled and versioned assets from the manifest.

The Laravel plugin also handles Vite SSR builds and works well with tools like Sail and Vapor.

The documentation will include full instructions, examples, and advanced configuration.

**This implementation requires a specific Vite configuration that the official Laravel Vite Plugin handles for you. For existing applications, there are also some required changes documented in the [upgrade guide](https://github.com/laravel-labs/vite-plugin-laravel/blob/main/UPGRADE.md).**

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
@@ -46,4 +46,15 @@ trait CompilesHelpers
     {
         return "<?php echo method_field{$method}; ?>";
     }
+
+    /**
+     * Compile the "vite" statements into valid PHP.
+     *
+     * @param  string  $arguments
+     * @return string
+     */
+    protected function compileVite($arguments)
+    {
+        return "<?php echo vite{$arguments}; ?>";
+    }
 }
--- a/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
@@ -2,6 +2,8 @@
 
 namespace Illuminate\View\Compilers\Concerns;
 
+use Illuminate\Foundation\Vite;
+
 trait CompilesHelpers
 {
     /**
@@ -55,6 +57,8 @@ trait CompilesHelpers
      */
     protected function compileVite($arguments)
     {
-        return "<?php echo vite{$arguments}; ?>";
+        $class = Vite::class;
+
+        return "<?php echo app('$class'){$arguments}; ?>";
     }
 }
--- a/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
@@ -61,4 +61,16 @@ trait CompilesHelpers
 
         return "<?php echo app('$class'){$arguments}; ?>";
     }
+
+    /**
+     * Compile the "viteReactRefresh" statements into valid PHP.
+     *
+     * @return string
+     */
+    protected function compileViteReactRefresh()
+    {
+        $class = Vite::class;
+
+        return "<?php echo app('$class')->reactRefresh(); ?>";
+    }
 }
--- a/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
@@ -52,7 +52,7 @@ trait CompilesHelpers
     /**
      * Compile the "vite" statements into valid PHP.
      *
-     * @param  string  $arguments
+     * @param  ?string  $arguments
      * @return string
      */
     protected function compileVite($arguments)
--- a/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
@@ -57,6 +57,8 @@ trait CompilesHelpers
      */
     protected function compileVite($arguments)
     {
+        $arguments ??= '()';
+
         $class = Vite::class;
 
         return "<?php echo app('$class'){$arguments}; ?>";
--- a/tests/Foundation/FoundationViteTest.php
+++ b/tests/Foundation/FoundationViteTest.php
@@ -1,0 +1,116 @@
+<?php
+
+namespace Illuminate\Foundation;
+
+use PHPUnit\Framework\TestCase;
+
+class FoundationViteTest extends TestCase
+{
+    public function testViteWithoutCss()
+    {
+        $this->makeViteManifest();
+
+        $result = (new Vite)(['resources/js/app-without-css.js']);
+
+        $this->cleanViteManifest();
+
+        $this->assertSame('<script type="module" src="/build/assets/app-without-css.versioned.js"></script>', $result->toHtml());
+    }
+
+    public function testViteWithCss()
+    {
+        $this->makeViteManifest();
+
+        $result = (new Vite)(['resources/js/app-with-css.js']);
+
+        $this->cleanViteManifest();
+
+        $this->assertSame(
+            '<link rel="stylesheet" href="/build/assets/app.versioned.css" />'
+            . '<script type="module" src="/build/assets/app-with-css.versioned.js"></script>',
+            $result->toHtml()
+        );
+    }
+
+    public function testViteWithSharedCss()
+    {
+        $this->makeViteManifest();
+
+        $result = (new Vite)(['resources/js/app-with-shared-css.js']);
+
+        $this->cleanViteManifest();
+
+        $this->assertSame(
+            '<link rel="stylesheet" href="/build/assets/app.versioned.css" />'
+            . '<script type="module" src="/build/assets/app-with-shared-css.versioned.js"></script>',
+            $result->toHtml()
+        );
+    }
+
+    public function testViteHotModuleReplacement()
+    {
+        $this->makeViteHotFile();
+
+        $result = (new Vite)(['resources/js/app-with-css.js']);
+
+        $this->cleanViteHotFile();
+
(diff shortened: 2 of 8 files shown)


bram-pkg:
Will the Vite manifest SRI plugin be supported by this? That would be nice.

jessarcher:
> Will the Vite manifest SRI plugin be supported by this? That would be nice.

It's on my list of things to look at post-launch and should be easy to implement :+1:

src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
```diff
@@ -46,4 +48,31 @@ protected function compileMethod($method)
...
+     * @param  ?string  $arguments
+     * @return string
+     */
+    protected function compileVite($arguments)
+    {
+        $arguments ??= '()';
+
+        $class = Vite::class;
+
+        return "<?php echo app('$class'){$arguments}; ?>";
```
hailwood, line 64: Seems like this needs brackets around arguments?

```php
return "<?php echo app('$class')({$arguments}); ?>";
```

bjhijmans:
```diff
- <link rel="stylesheet" href="{{ mix('css/app.css') }}">
- <script src="{{ mix('js/app.js') }}" defer></script>
+ @vite(['resources/css/app.css', 'resources/js/app.js'])
```
How would you add nonces to this vite call?

tests/Foundation/FoundationViteTest.php
```diff
@@ -0,0 +1,164 @@
+<?php
+
+namespace Illuminate\Foundation;
```
lucasmichot, line 3:
```suggestion
namespace Illuminate\Tests\Foundation;
```

src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
hailwood: Ah righto, that makes sense (I was testing vite by adding a custom directive)

tabuna:
Hi @jessarcher 

> The main difference is that instead of using the mix() helper, you would now use the @vite Blade directive. The reason for the directive is that Laravel needs to output complete <script> and <link rel="stylesheet"> tags, rather than just a path as with Mix.

Is there a way to output the path like it was with `mix()` because it is often required to specify additional things like for `link` tags:

```blade
<link rel="stylesheet" type="text/css" href="{{  mix('/css/light.css') }}"
 media="(prefers-color-scheme: light)">
```

For `script`:
```blade
<script src="{{  mix('/app/part/landing.js') }}" 
data-turbo-eval="false" data-turbolinks-eval="false"></script>
```

timacdonald:
@tabuna there isn't a way to do this currently with Vite, however we are considering enabling the specification of arbitrary attributes on entry point tags.

mikemand:
> @tabuna there isn't a way to do this currently with Vite, however we are considering enabling the specification of arbitrary attributes on entry point tags.

I was thinking about this also, about how I could add a CSP nonce to the script tag. What about a blade template that can be overridden? That way, instead of calling `makeTag`/`makeScriptTag`/`makeStylesheetTag`, the URLs are pushed to the collection instead and then passed to a blade template.

nlvedwin:
after installing a fresh app 
image
image

timacdonald:
@nlvedwin it looks like you are visiting the Vite dev server. You still need to utilise `php artisan serve`, Valet or similar.

For example if you are using Valet on your system, you should visit `laravel-vite.test` instead.

bram-pkg:
No, you load the normal artisan serve URL, ***@***.***` will take of making it
able to hot reload from the development server.

nlvedwin:
> @nlvedwin it looks like you are visiting the Vite dev server. You still need to utilise `php artisan serve`, Valet or similar.
> 
> For example if you are using Valet on your system, you should visit `laravel-vite.test` instead.



> @nlvedwin it looks like you are visiting the Vite dev server. You still need to utilise `php artisan serve`, Valet or similar.
> 
> For example if you are using Valet on your system, you should visit `laravel-vite.test` instead.

got it now, it feels like magic 🧙

mikemand:
For others who are having a 404 when trying to check the assets (http://localhost:3000/build/css/app.css), run `vite build` first to see if there are any errors compiling. I moved from SASS to PostCSS at the same time as upgrading to Vite and had a few errors that PostCSS couldn't deal with (commented out `@apply`, for one).

chrispage1:
Hi all, referred here by @tabuna. For those of you who are looking for a direct replacement of the mix() helper, please see my Vite helper package on Packagist - https://packagist.org/packages/motomedialab/laravel-vite-helper

timacdonald:
Hey folks,

A lot of questions in this thread about adding a CSP nonce, Subresource integrity, or arbitrary attributes (turoblinks, media query stuff, etc) to the Vite generated tags.

I've just opened a PR to support all of those use-cases.

You can check it out here: https://github.com/laravel/framework/pull/43442

Any feedback would be appreciated.

bram-pkg:
Awesome stuff @timacdonald, thank you so much!

kitro:
Great work @jessarcher  :+1: ,
I think it will be nice if we can also use `@vitejs/plugin-legacy` to inject nomdule script tag. In some cases like using the iPhone with iOS 13 that not support modular bundles `<script type="module">`.
Thank you,

kitro:
> I think it will be nice if we can also use `@vitejs/plugin-legacy` to inject nomdule script tag. In some cases like using the iPhone with iOS 13 that not support modular bundles `<script type="module">`.

After https://github.com/laravel/framework/pull/43442 I get a workaround to inject legacy generated files 
```php
        Vite::useScriptTagAttributes(function (string $src, string $url, array|null $chunk, array|null $manifest) {
            if (in_array(
                $src,
                [
                    'vite/legacy-polyfills',
                    'app/resources/js/app/app-legacy.js',
                    'app/resources/sass/app/app-legacy.scss',
                ]
            )) {
                return [
                    'type' => 'text/javascript',
                    'nomodule',
                ];
            }

            return [];
        });
```

FreekVR:
Late comment, but while the above works (thanks!) it is still loading the legacy script via preloads for modern browsers

eg.
`<link rel="modulepreload" href="/build/assets/app-legacy-d2b2bf94.js" />`

So far I've been unable to find a workaround for this

FreekVR:
I was able to find an ondocumented callback for the preloads, in which returning false causes no preloads to be included.
By default, a normal or module preload will cause the legacy bundle(s) to be downloaded even in modern browsers.

The following worked for us in a ViteServiceProvider we added to the project.
```
<?php

namespace App\Providers;

class ViteServiceProvider extends ServiceProvider
{
    protected const NONMODULE_FILES = [
        'vite/legacy-polyfills-legacy',
        'resources/frontend/app/app-legacy.js',
    ];

    /**
     * Register services.
     *
     * @return void
     */
    public function register()
    {
        //
    }

    /**
     * Bootstrap services.
     *
     * @return void
     */
    public function boot()
    {
        Vite::usePreloadTagAttributes(function ($src, $url, $chunk, $manifest) {
            // Don't preload non-module files at all
            if (in_array($src, self::NONMODULE_FILES)) {
                return false;
            }
            // Use the default attributes for the rest
            return [];
        });
        Vite::useScriptTagAttributes(function (string $src, string $url, array|null $chunk, array|null $manifest) {
            if (in_array($src, self::NONMODULE_FILES)) {
                return [
                    'type' => 'text/javascript',
                    'nomodule',
                ];
            }
            return [];
        });
    }
}
```

PR #42787, opened 2022-06-13
[9.x] Fix boolean values used as a date/datetime value when casting

Tarpsvo:
When a boolean value of `true` reaches the HasAttributes trait's addCastAttributesToArray function, it will be casted into a `date` as the `in_array` function isn't set to strict comparison mode. Please see the following screenshot:

image


GrahamCampbell:
Under what scenario could that happen? Could you add a test that failed before, but passes now?

Tarpsvo:
We have a project in which this reproduces fairly easily. I will look into adding the test and get back to you. Thanks!

Tarpsvo:
Please forget you've ever seen me.

image

PR #42793, opened 2022-06-13
[8.x] Loose comparison causes the value not to be saved

vsg24:
**The PR back-ports  https://github.com/laravel/framework/issues/41337 to 8.x branch.**

The description below is copied from the original PR.
---
Given an attribute with the cast object or collection, the $model->isDirty() method returns the wrong value due to the use of loose comparison.

Example from issue https://github.com/laravel/framework/issues/37961:

```php

class Test extends Model
{
    protected $casts = [
        'collection' => 'collection',
    ];
    protected $fillable = [
        'collection',
    ];
}

$model = new Test(['collection' => ['key' => true]]);
$model->syncOriginal();
$model->fill(['collection' => ['key' => 'value']]);

$model->isDirty(); // false but should be true

// or
$model = new Test(['collection' => ['key' => null]]);
$model->syncOriginal();
$model->fill(['collection' => ['key' => false]]);

$model->isDirty(); // false but should be true
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1956,8 +1956,8 @@ trait HasAttributes
             return $this->fromDateTime($attribute) ===
                 $this->fromDateTime($original);
         } elseif ($this->hasCast($key, ['object', 'collection'])) {
-            return $this->castAttribute($key, $attribute) ==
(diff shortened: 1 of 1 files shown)


MasterRO94 (commented):
This PR breaks BC. If original $value is `null` then `PHP Deprecated:  json_decode(): Passing null to parameter #1 ($json) of type string is deprecated` is thrown

Issue #42794, opened 2022-06-13
Incorrect Query Params by Arr::query()


comes:
- Laravel Version: 8.83.16 & 9.17.0
- PHP Version: 8.0.19 & 8.1.6
- Database Driver & Version: unneeded

### Description:

Query params built by `Arr::query()` didn't match the expected result.

### Steps To Reproduce:

```php
$query = ['foo' => 1, 'bar' => 'hello world', 'nullable' => null, 'empty' => ''];

dump(http_build_query($query, '', '&', PHP_QUERY_RFC3986)); // foo=1&bar=hello%20world&empty=

dump(\Arr::query($query)); // foo=1&bar=hello%20world&empty=
```

the expected output should be `foo=1&bar=hello%20world&empty=&nullable=`

`Arr::query()` use an internal php function which did not produce the expected results.

```
Route::any('test', function (Request $request) {
    $laravelQuery = $request->all();
    $vanillaQuery = $_GET;

    return print_r([$vanillaQuery,
        $laravelQuery,
        $vanillaQuery === $laravelQuery,
        \Arr::query($vanillaQuery),
        \Arr::query($laravelQuery)
    ]);
});
```

output:
```

Array
(
    [0] => Array
        (
            [foo] => bar
            [empty] => 
        )

    [1] => Array
        (
            [foo] => bar
            [empty] => 
        )

    [2] => 
    [3] => foo=bar&empty=
    [4] => foo=bar
)

```

The reason for the null-value is the default middleware `\Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,` in https://github.com/laravel/laravel/blob/1f9eee06ddb38abf86e2f7df37a20281cd774e3d/app/Http/Kernel.php#L23

but the reason for the wrong query parameters is in `http_build_query` see https://www.php.net/manual/en/function.http-build-query.php#60523

The follow up error here is, if you create a pagination `withQueryParams()`, the resulting query parameter is something different, than the original one.

rodrigopedra:
Sent PR #42795

rodrigopedra:
Hey @comes , I tried PR #42795 targeting 8.x, as it seemed a bug fix, and PR #42806 targeting master as the first PR was rejected due to being a breaking change on expected behavior. 

[...]

PR #42795, opened 2022-06-13
[8.x] Do not remove null query parameters

rodrigopedra:
As outlined in issue #42794 , `Arr:query()` removes parameters set to null as a side effect of using `http_build_query` internally.

For example:

~~~php
$query = ['foo' => 1, 'bar' => 'hello world', 'nullable' => null, 'empty' => ''];
~~~

Removes the `nullable => null` parameter, while keeping `empty => ''`, which might be confusing.

One particular scenario where this issue can occur is with GET routes which expect query parameters to get back to a filtered state.

With the current implementation, if this particular route expects a parameter just to be present, for example as a boolean parameter, the `UrlGenerator` would remove these null parameters when, for example generating redirects, or links in a mailable, among other scenarios.

This parameters could come from a view with a link "share report by email", as `ConvertEmptyStringsToNull` middleware convert all empty strings to null, the request would generate unexpected links in the sent mail.

This PR:

- Converts all empty string to `null` before calling `http_build_query` in `Arr@query()`
- Fixes test cases, and test helpers to account for nullable query parameters.

~~~

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -531,6 +531,12 @@ class Arr
      */
     public static function query($array)
     {
+        array_walk_recursive($array, function (&$value) {
+            if ($value === null) {
+                $value = '';
+            }
+        });
+
         return http_build_query($array, '', '&', PHP_QUERY_RFC3986);
     }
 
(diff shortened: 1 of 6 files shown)


PaolaRuby:
Maybe it is a breaking change

GrahamCampbell:
This is a huge breaking change, for sure. This should target 10.x.

taylorotwell:
Sorry, as noted - this could be a breaking change?

rodrigopedra:
Sure, it looks like so. I will send to 10.x then.

PR #42798, opened 2022-06-14
[9.x] Add `toResourceCollection` and `toAnonymousResourceCollection` to Collections

bennett-treptow:
This PR aims to add two new collection methods: `toResourceCollection` and `toAnonymousResourceCollection`.

`toResourceCollection` allows a user to specify a `ResourceCollection` that the current collection can be cast into instead of having to wrap the collection in a `ResourceCollection::make()`.

Same with `toAnonymousResourceCollection` -- this allows a user to specify a `JsonResource` that the current collection would be put into with the existing `JsonResource::collection()` method.

```php
class UserResource extends JsonResource {
}

class UserResourceCollection extends ResourceCollection {
    public $collects = UserResource::class;
}

// UserController.php
// using toResourceCollection
class UserController extends Controller {
    public function oldIndex(Request $request){
        return UserResourceCollection::make(
            User::whereIn('name', ['Bennett', 'Taylor', 'et all'])->get()
        );
    }

    public function newIndex(Request $request){
        return User::whereIn('name', ['Bennett', 'Taylor', 'et all'])
            ->get()
            ->toResourceCollection(UserResourceCollection::class);
    }
}

// UserController.php
// using toAnonymousResourceCollection()
class UserController extends Controller {
    public function oldIndex(Request $request){
        return UserResource::collection(
            User::whereIn('name', ['Bennett', 'Taylor', 'et all'])->get()
        );
    }

    public function newIndex(Request $request){
        return User::whereIn('name', ['Bennett', 'Taylor', 'et all'])
            ->get()
            ->toAnonymousResourceCollection(UserResource::class);
    }
}
```

The benefit here is that this method is just end of the chain of the collection pipeline and doesn't need to be wrapped in the resource methods to achieve the same effect.
This is definitely just a DX improvement, and does not introduce any new functionality so it would not break any existing features related to collections and resources.

**Note:** I am not 100% on the PHPStan type assertions -- I have them passing now but I don't think they're as specific as they can get, but I've tried all the combinations I could think of to get the classes to pass around correctly, but no dice. 
Also, not hard-stuck on the names `toResourceCollection`, `toAnonymousResourceCollection`.

**EDIT:** I am now aware there is a `pipe` method that would potentially allow userland to implement this exact function -- however, since resources are a first-class Laravel object, I think it would be fitting to have a collection method to reflect that. I understand this is just syntactic sugar in that case, but would be curious to see if that puts this at a tipping point for acceptance.

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -925,6 +925,7 @@ trait EnumeratesValues
 
     /**
      * Get the collection of items as a resource collection.
+     *
      * @template TResourceCollectionClass of \Illuminate\Http\Resources\Json\ResourceCollection
      *
      * @param  class-string<TResourceCollectionClass>  $resourceCollectionClass
@@ -937,6 +938,7 @@ trait EnumeratesValues
 
     /**
      * Get the collection of items as an anonymous resource collection.
+     *
      * @template TResourceClass of \Illuminate\Http\Resources\Json\JsonResource
      *
      * @param  class-string<TResourceClass>  $resourceClass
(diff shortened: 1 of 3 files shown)


bennett-treptow:
TIL there is a `pipe` method that does this very thing. I just was looking through methods again to check if there was something that does this, and found the `pipe` method for the first time (and I've used just about every collection method at least once before -- just not ever pipe!)
I'm going to leave this open though to see if the `toResourceCollection` or `toAnonymousResourceCollection` has any interest -- feel free to close otherwise.

taylorotwell:

PR #42799, opened 2022-06-14
[9.x] Add nonconsecutive option to `Str::ucsplt`

Riley19280:
This PR adds an option to `Str::ucsplt` to keep groups of uppercase characters together. 
Many times ucsplit is used to split text that contain acronyms such as 'PHP', which results in these acronyms being broken up, for example:
```php
Str::ucsplit('PHPIsCool'); // ['P', 'H', 'P', 'Is,', 'Cool']
``` 
Passing `$consecutive = false` results in
```php
Str::ucsplit('PHPIsCool'); // ['PHP', 'Is,', 'Cool']
``` 

Defaulting the `$consecutive` argument to true allows for full backwards compatibility with existing code.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -1115,7 +1115,7 @@ class Str
      */
     public static function ucsplit($string, $consecutive = true)
     {
-        if($consecutive) {
+        if ($consecutive) {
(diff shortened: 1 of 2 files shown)


rodrigopedra:
Adding a new parameter to a public function is a breaking change, regardless if it has a default value.

It will break any class that extends `Illuminate\Support\Str`, which people might be doing in their projects to add new methods to it, instead of using macros

taylorotwell:

PR #42800, opened 2022-06-14
[10.x] Bring translatable error messages to closure based rules

timacdonald:
This PR brings the translatable validation error messages [from the Invokable validation rule](https://github.com/laravel/framework/pull/42689) to Closure based rules for the next version of Laravel.

```php
public function rules()
{
    'foo' => [
            $fail('validation.translation.key')->translate();
        },
    ],
}
```

Targeting master as I feel there are two (minor) breaking changes here...

1. Calling `$fail` more than once now appends to an array, so if you were calling it twice in a closure (which is probably a bug where you forgot to return), the behaviour would change.
2. $fail now returns an object , rather than void / null, so if you have typehinted your closure and are returning the result of the `$fail` call, it would break.

### DIFF
--- a/src/Illuminate/Validation/ClosureValidationRule.php
+++ b/src/Illuminate/Validation/ClosureValidationRule.php
@@ -3,8 +3,10 @@
 namespace Illuminate\Validation;
 
 use Illuminate\Contracts\Validation\Rule as RuleContract;
+use Illuminate\Contracts\Validation\ValidatorAwareRule;
+use Illuminate\Translation\PotentiallyTranslatedString;
 
-class ClosureValidationRule implements RuleContract
+class ClosureValidationRule implements RuleContract, ValidatorAwareRule
 {
     /**
      * The callback that validates the attribute.
@@ -21,11 +23,18 @@ class ClosureValidationRule implements RuleContract
     public $failed = false;
 
     /**
-     * The validation error message.
+     * The validation error messages.
      *
-     * @var string|null
+     * @var array
      */
-    public $message;
+    public $messages = [];
+
+    /**
+     * The current validator.
+     *
+     * @var \Illuminate\Validation\Validator
+     */
+    protected $validator;
 
     /**
      * Create a new Closure based validation rule.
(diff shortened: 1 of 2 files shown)


dennisprudlo:
What's the benefit of passing a translation key as a parameter and then calling the "translate" method on the result over just `$fail(trans('validation.translation.key'))`?

taylorotwell:
@timacdonald I'm also a bit curious about the question above

timacdonald:
Hey folks, sorry I didn't reply sooner. Got side tracked. So this was my reasoning as to why I wanted to introduce this functionality.


1. Translation is a first party concern for validation errors, yet there is no affordance given for rules to perform such translation. The only option we have is reach out into the global world with helpers, facades, etc. This is, as we know, functionally the same, but it _forces_ you to do this. I love that Facades and global helpers exist (love 'em in my views), but I also love the _"you can use them if you want to"_ approach that Laravel provides. So this functionality is about promoting a first-party solution for translating validation error messages. You can still use helpers if you want, or you can use this solution.
2. It is not a 1:1 mapping to `$fail(trans('validation.required'))`, as the `trans` helper is lax. The new functionality is the opposite. It is strict, ensuring that the user does not actually see the validation error "validation.required", when it should have seen the translated "Is Required". The `$fail()->translate()` functionality is strict and will throw an exception when there is no translation present. I've seen this plenty of times and would love to not silently fail so it can be addressed.
3. The way I see it, Validation objects are value objects...

```php
return [
    'options' => new InRule([5, 6, 7])
];
```

Having value objects rely on global state or be half value object / half service object by injecting the translater itself is clunky as heck - yes, this is possible as the `rules()` method does support dependency injection, but it feels hacky and weird to me.

```php
public function rules(Translator $translator)
{
    return [
        'options' => new InRule([5, 6, 7], $translator)
    ];
}
```

Hopefully you can see how this is, IMO, a valuable addition to the framework.

```php
$fail('rant.over')->translate();
```

dennisprudlo:
@timacdonald @taylorotwell might it be worth a thought to think about how this could include parameters and pluralization? As far as I can see this approach only supports single-fixed translation strings. So one would still have to use both (translate method on the result and global helpers when using parameters and pluralization)

timacdonald:
@dennisprudlo see https://github.com/laravel/framework/pull/42689

PR #42801, opened 2022-06-14
[9.x] Allow invokable rules to push messages to nested (or other) attributes

timacdonald:
This allows invokable validation rule classes to push errors to nested or other attributes. This allows greater integration with the validator and allows userland rules to validate rich object structures, but apply errors to the correct keys.

```php
class UserRule implements InvokableRule
{
    {
        if (! is_array($attribute) || array_is_list($attribute)) {
            return $fail('Must be an object.'); // apply to $attribute
        }

        if (! array_key_exists('name', $attribute)) {
            return $fail("{$attribute}.name", 'Is required.'); // apply to nested attribute
        }

        /* ... */
    }
}
```

Potential error messages...

```php
Validator::make(
    [
        'user_1' => ['xxxx'],
        'user_2' => ['age' => 23],
    ],
    [
        'user_1' => [new UserRule()],
        'user_2' => [new UserRule()],
    ]
);

// errors...

[
    'user_1' => ['Must be an object.'],
    'user_2.name' => ['Is required.'],
]

```

This feature allows developers to create validation rules that can handle rich objects and validate not only the top level value, but also nested values.

In combination with the `DataAware` interface, a developer may also push validation errors to other keys. For example a developer could re-create the prohibits validation rule, but put the error message on the attribute that should not exist, rather than the attribute that specified the validation rule.

### DIFF
--- a/src/Illuminate/Validation/InvokableValidationRule.php
+++ b/src/Illuminate/Validation/InvokableValidationRule.php
@@ -61,7 +61,7 @@ class InvokableValidationRule implements Rule, ValidatorAwareRule
      * Create a new implicit or explicit Invokable validation rule.
      *
      * @param  \Illuminate\Contracts\Validation\InvokableRule  $invokable
-     * @return \Illuminate\Contracts\Validation\ImplicitRule
+     * @return \Illuminate\Contracts\Validation\Rule
      */
     public static function make($invokable)
     {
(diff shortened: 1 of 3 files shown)


ankurk91:
Tim, you are on fire 🔥

timacdonald:
❤️

PR #42805, opened 2022-06-14
Added new support method to the existing Str class helper that performs a preg replace

chrisjk123:
I saw an opportunity to extend the `Str` helper class and felt like we needed a regular `preg` method here.


fgaroby:
I think `preg()` is not a clear method name : is it a `preg_replace`? A `preg_match`?
Isn't it more clear to update the `replace`method, instead?

taylorotwell:

PR #42806, opened 2022-06-14
[10.x] Do not remove null query parameters

rodrigopedra:
**This is resend of #42795 to master**

As outlined in issue #42794 , `Arr:query()` removes parameters set to null as a side effect of using `http_build_query` internally.

For example:

~~~php
$query = ['foo' => 1, 'bar' => 'hello world', 'nullable' => null, 'empty' => ''];
~~~

Removes the `nullable => null` parameter, while keeping `empty => ''`, which might be confusing.

One particular scenario where this issue can occur is with GET routes which expect query parameters to get back to a filtered state.

With the current implementation, if this particular route expects a parameter just to be present, for example as a boolean parameter, the `UrlGenerator` would remove these null parameters when, for example generating redirects, or links in a mailable, among other scenarios.

This parameters could come from a view with a link "share report by email", as `ConvertEmptyStringsToNull` middleware convert all empty strings to null, the request would generate unexpected links in the sent mail.

This PR:

- Converts all `null`s to empty strings before calling `http_build_query` in `Arr@query()`
- Fixes test cases, and test helpers to account for nullable query parameters.

**Edit**: fixed wording, previously it said it would convert empty strings to nulls, but it is actually the other way around,

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -602,6 +602,12 @@ class Arr
      */
     public static function query($array)
     {
+        array_walk_recursive($array, function (&$value) {
+            if ($value === null) {
+                $value = '';
+            }
+        });
+
         return http_build_query($array, '', '&', PHP_QUERY_RFC3986);
     }
 
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -332,9 +332,9 @@ EOF;
         );
 
         if (! is_null($name)) {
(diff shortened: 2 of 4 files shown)


taylorotwell:
I don't plan to break this behavior.

rodrigopedra:
@taylorotwell would you consider a different method?

Otherwise I can tell #42794 OP to use a macro instead

comes:
Hej all, just want to point to a more advanced example https://github.com/laravel/framework/issues/42794#issuecomment-1156454217

PR #42809, opened 2022-06-14
Adds maskWords() for masking each words of the string separated by a separator

yusufonur:
Not every word can be masked in the text that is masked with mask. It takes extra effort to do this.

Each word can be masked according to the separator specified with maskWords.

Example:
```
use Illuminate\Support\Str;

$str = 'Yusuf Onur SARI';

return [
    Str::mask($str, "*", 1),
    Str::maskWords($str),
];

/*
[
    "Y**************",
    "Y**** O*** S***",
]
*/
```


taylorotwell:

PR #42812, opened 2022-06-14
[9.x] Fix callOnce seeder method to properly process arrays

wreality:
Update the callOnce method in Seeder.php to correctly handle an array of seeder classes being supplied.

@sebastiandedeyne I'm assuming this was the intent of the original based on the doc block for the $class parameter (`string|array`) and your examples in your [post ](https://sebastiandedeyne.com/composable-laravel-seeders-with-callonce/)about callOnce being added.

### DIFF
--- a/src/Illuminate/Database/Seeder.php
+++ b/src/Illuminate/Database/Seeder.php
@@ -105,7 +105,7 @@ abstract class Seeder
 
         foreach($classes as $class) {
             if (in_array($class, static::$called)) {
-                return;
+                continue;
             }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:

Thanks!

wreality:
Fair enough.  I assumed the change was straight forward enough to not need a lot of explanation.  I'll come back to this when I have time and see if I can flesh something out.  The original inclusion of callOnce didn't include tests, which probably would have caught this before it was included.

Appreciate the time in any case.

PR #42825, opened 2022-06-15
[9.x] Add array support to @error blade directive

npostman:
This update allows for the @error blade directive to accept an array of keys to check for in the MessageBag.

When given an array, it will check if the error bag has an error in any of the keys, if so, the first error in the first key that has errors will be available in the `$message` variable.

Use case:
```php
<input id="accessCode" type="text" class="form-control @error(['code', 'throttled']) is-invalid @enderror"
placeholder="Code" name="code" value="{{ old('code') }}" aria-describedby="accessCodeFeedback">
<label>{{ trans('labels.access_code') }}</label>
@error(['code', 'throttled'])
<div id="accessCodeFeedback" class="invalid-feedback">{{ $message }}</div>
@enderror
```

I'm not sure on how to add testing coverage to this. Some help/guidance is welcome.

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesErrors.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesErrors.php
@@ -16,9 +16,15 @@ trait CompilesErrors
 
         return '<?php $__errorArgs = ['.$expression.'];
 $__bag = $errors->getBag($__errorArgs[1] ?? \'default\');
-if ($__bag->has($__errorArgs[0])) :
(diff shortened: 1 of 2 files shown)


taylorotwell:

npostman:
I've put it in a Gist as Blade extensions registered in the AppServiceProvider. Felt a bit like overkill to create a package for just this simple function.

https://gist.github.com/npostman/273b9a8d165b5b1872a62e90cdb33098

PR #42828, opened 2022-06-15
[9.x] Avoid Storing Turbo Frame Visits as Previous URL

tonysm:
There was [an issue](https://github.com/tonysm/turbo-laravel/issues/60#issuecomment-1123142591) in the Turbo Laravel package where someone noticed the `url()->previous()` was behaving weirdly. I traced it down to how Laravel stores the previously visited URL in the session.

I think it's safe to ignore visits from Turbo Frames to get stored as previously visited URL, since that's generally not a URL you would want to redirect users to (since it's mostly for UI fragments, not always entire pages). The only way to override this behavior was by creating a new middleware that extends the StartSession and override that method.

I think it wouldn't hurt to add this to the core or at least provide a way where I could customize this behavior in the package.

Couldn't find tests for this behavior, so I also added a test case just for that behavior.

### DIFF
--- a/src/Illuminate/Session/Middleware/StartSession.php
+++ b/src/Illuminate/Session/Middleware/StartSession.php
@@ -202,7 +202,8 @@ class StartSession
         if ($request->isMethod('GET') &&
             $request->route() instanceof Route &&
             ! $request->ajax() &&
-            ! $request->prefetch()) {
+            ! $request->prefetch() &&
+            ! $request->headers->has('Turbo-Frame')) {
             $session->setPreviousUrl($request->fullUrl());
         }
     }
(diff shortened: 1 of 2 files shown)


johanrosenson:
This feels like opening pandoras box; this is very specific to a single framework, IMO this will only end with more checks for other similar solutions in other frameworks.

I think this belongs in a custom middleware in the projects that require this.

tonysm:
Tbh, I agree. I mainly wanted to see if there was a way to make this more customizable at app or package levels without requiring an entirely new middleware. I'll try to find another way and if I'm not able to, I guess I can only document the need to override this.

PR #42847, opened 2022-06-17
[9.x] form requests: populate input to public properties

MohannadNaj:
This PR introduces a suggested feature to form requests: populate input to the typed public properties.

If a `FormRequest` instance had a `$name` public property and the request had `name` input, the input value will be populated to the property.

``` php

class CreateUserRequest extends FormRequest
{
    public string $name;

    public function rules()
    {
        return [
            'name' => ['required', 'string',]
        ];
    }
}
```
``` php
        $this->post('user', [
            'name' => 'abc',
        ]);
```
``` php
    public function store(CreateUserRequest $request)
    {
        $request->name; // 'abc'
    }
```

> In addition, please describe the benefit to end users; the reasons it does not break any existing features; how it makes building web applications easier, etc.

Mainly, IDE support and better development experience.

### DIFF
--- a/src/Illuminate/Validation/ValidatesWhenResolvedTrait.php
+++ b/src/Illuminate/Validation/ValidatesWhenResolvedTrait.php
@@ -2,6 +2,9 @@
 
 namespace Illuminate\Validation;
 
+use ReflectionClass;
+use ReflectionProperty;
+
 /**
  * Provides default implementation of ValidatesWhenResolved contract.
  */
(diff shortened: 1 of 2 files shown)


MohannadNaj:
Caveats related to this PR:

### Behaviour after `$request->merge` & `$request->replace`:

The current submitted implementation doesn't refresh the properties. However, it exposes a public `refresh` function so it's up to the userland code to call it if needed.

### Possible misuse: Setting a property and expecting it in `->all()`

``` php

    public function store(CreateUserRequest $request)
    {
        $request->name = 'xxx';

        $request->all(); // [] <-- setting the name ignored
    }
```

taylorotwell:
Not sure I can add this on a patch release. Maybe you can release a package that does this?

PR #42857, opened 2022-06-17
[9.x] Add contains rule for validation

reziamini:
With these new changes, we will be able to use the `contains` rule for validations.
For example:
```PHP
return [
    'foo' => 'required|contains:Foo,Bar'
];
```
If the value of `foo` contains `Foo` or `Bar` it will be passed.

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1951,6 +1951,19 @@ trait ValidatesAttributes
     }
 
     /**
+     * Validate the attribute contains a given substring.
+     *
+     * @param  string  $attribute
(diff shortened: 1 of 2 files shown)


willrowe:
Is this different from [in](https://laravel.com/docs/9.x/validation#rule-in)?

rezaamini-ir:
@willrowe Of course, in will check the given value is exactly equal to the given attribute in the rules.
But it will check does the given value contains one of the attributes in the rules. 
Imagine the value is `Will Rowe`:
```php
'contains:Will', // ✅
'in:Will', // ❌
```

rezaamini-ir:
Yeah, in behind scene it uses ‘Str::contains()’ from Laravel.

taylorotwell:
I would just use a custom closure rule.

PR #42859, opened 2022-06-17
[9.x] Development in function `logoutOtherDevices`

michaelnabil230:
This PR is developing a `rehashUserPassword` function Some developers use `Attribute` on the user’s password for example so they can put hash every time they update the password or register or forget the password the user’s  

```php
{
    use HasApiTokens;
    use Notifiable;

    /**
     *
     * @var string[]
     */
    protected $fillable = [
        'name',
        'email',
        'password',
    ];

    /**
     * Set the user's Hash password.
     *
     * @return Attribute
     */
    {
        return Attribute::set(fn ($value) => Hash::make($value));
    }
}
```


DarkGhostHunter:
Let me translate this PR:

This PR introduces the `rehashUserPassword` parameter to the `logoutOtherDevices` function. It accepts a closure to allow manually rehashing the password.

I don't see the use case for this.

PR #42861, opened 2022-06-17
[9.x] Add localization for server error messages.

NanoNulla:
It adds localization support to exception hander's error messages.


taylorotwell:
These messages are not really intended to be displayed to the end user and thus do not need to be localized.

hamidafghan:
Thanks @taylorotwell

PR #42862, opened 2022-06-17
Create resource file for a model when use make:model command

AbdlrahmanSaberAbdo:
**What**
-----------------------

This PR is for creating `resource` file with `make:model` command when using `-a` or `--resource/-R` options also support creating resource collection by adding `--collection` to the command. for example:

```
php artisan make:model post -a
php artisan make:model post -R
```

The above command will create all old files + resource file with the name `PostResource`.

if you'd like to create a collection resource file you can add `--collection` option. for example

```
php artisan make:model post -a --collection /
php artisan make:model post -R --collection
```

**Why**
---------------------------
I always wondered why `resource` file didn't create with other files when using  `make:model` with `-a` or `--resource/-R` option, is that for a specific reason? maybe there's a reason I don't know it, also I think it'll save time because when we use `-a` option with `make:model` command we use it to create all other files to save time so I think it'll be good if create resource file to avoid writing another command to create it, and when use `--resource/-R` it'll be good to create a resource file also.

I tried to search why it doesn't create a `resource` file but I didn't find an answer, I found a question on Stackoverflow that asked the same question from 2 years ago for Larave 6,  you can check it from [here](https://stackoverflow.com/questions/57944093/why-make-model-a-command-doesnt-create-resource-file).

### DIFF
--- a/src/Illuminate/Foundation/Console/ModelMakeCommand.php
+++ b/src/Illuminate/Foundation/Console/ModelMakeCommand.php
@@ -84,6 +84,10 @@ class ModelMakeCommand extends GeneratorCommand
         if ($this->option('policy')) {
             $this->createPolicy();
         }
+
+        if ($this->option('resource')) {
+            $this->createResource();
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think it's probably because many Laravel applications don't have external APIs, which are what resources are for. Think I will let things be for now.

AbdlrahmanSaberAbdo:
> I think it's probably because many Laravel applications don't have external APIs, which are what resources are for. Think I will let things be for now.

You're right, I missed it, I should create it when the user adds `--api` option, what do you think? if this is acceptable shall I update this pull request or create a new one? @taylorotwell

PR #42864, opened 2022-06-17
[9.x] Adds `silent` function to silence a callable thrown exception

DarkGhostHunter:
## What?

The `silent` function method will catch any `Exception` instance and return `false` instead of the callback result, while reporting the exception to the application. This removes the necessity of going for a try-catch-all block and reporting manually.

```php
$result = silent(Something::mayBreak(...));

if (! $result) {
   return 'Something broke!');
}
```

This can be very useful when using with `retry()`. A _silent retry_:

```php
$result = silent(retry(...), Something::mayBreak(...));
```

Of course, it won't silence `Errors` like invalid arguments or return types.

### DIFF
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -257,6 +257,24 @@ if (! function_exists('retry')) {
     }
 }
 
+if (! function_exists('silent')) {
+    /**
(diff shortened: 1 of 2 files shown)


DarkGhostHunter:
Looooooool 😂 

Italo Baeza C.

PR #42868, opened 2022-06-18
[9.x] Create resource file when writes make:model with `-a` and `--api` options

AbdlrahmanSaberAbdo:
**What**
-----------------------

This PR is for creating `resource` file with `make:model` command when using `-a/--all` with `--api` options also supports creating resource collection by adding `--collection` to the command. for example:

```
php artisan make:model post -a --api
```

The above command will create all old files + resource file with the name `PostResource`.

if you'd like to create a collection resource file you can add `--collection` option. for example

```
php artisan make:model post -a --api --collection
```

**Why**
---------------------------
I made my first pull request here about this https://github.com/laravel/framework/pull/42862, after seeing @taylorotwell response he show me some unclear points to me so that's why I create this pull request, let me explain what's the changes that I made 

I think we can create a resource file when the user adds `-a/--all` option with `--api` option because like @taylorotwell said 

> many Laravel applications don't have external APIs, which are what resources are for.

So the whole command will be like this: `php artisan make:model post -a --api`

currently when the user writes `-a` with `--api` it creates a different controller with all API functions, if the user writes `-a` without `--api` option it'll create a controller with all resource functions so I think it'll be more useful when creating a resource file when a user writes `--api` option with `-a` option.

### DIFF
--- a/src/Illuminate/Foundation/Console/ModelMakeCommand.php
+++ b/src/Illuminate/Foundation/Console/ModelMakeCommand.php
@@ -84,6 +84,10 @@ class ModelMakeCommand extends GeneratorCommand
         if ($this->option('policy')) {
             $this->createPolicy();
         }
+
+        if ($this->option('resource')) {
+            $this->createResource();
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


taylorotwell:

AbdlrahmanSaberAbdo:
> 
> If you feel absolutely certain that this code corrects a bug in the framework, please "**@**" mention me in a follow-up comment with a further explanation so that GitHub will send me a notification of your response.

Hey @taylorotwell, no it's not correct a bug, it's to save time, the purpose of using `-a` command is to create all needed files and, when using it with `--api` my opinion is that should create also all files and here's come this pull request if you use `-a` with `--api` that means that you build APIs for that model so instead of writing another command to create a resource we can create it automatically when user write `-a` with `--api`.

I think it's not possible to be a package because the changes are too small so that's why I did it here.

Anyway, I understand that you want to be very careful when adding new code.

PR #42871, opened 2022-06-19
Use arrow function

afsharone:


GrahamCampbell:
What value does this add? In some of these instances, it makes readability worse IMO.

PR #42884, opened 2022-06-20
[9.x] Fix 0 division working with variable length pages

Aethedigm:
Divide by zero fix for length aware paginator. Fixes a few issues with argument based pagination.

### DIFF
--- a/src/Illuminate/Pagination/LengthAwarePaginator.php
+++ b/src/Illuminate/Pagination/LengthAwarePaginator.php
@@ -46,7 +46,7 @@ class LengthAwarePaginator extends AbstractPaginator implements Arrayable, Array
         }
 
         $this->total = $total;
-        $this->perPage = $perPage;
+        $this->perPage = (int) $perPage;
         $perPage = $perPage > 0 ? $perPage : 1;
         $this->lastPage = max((int) ceil($this->total / $perPage), 1);
         $this->path = $this->path !== '/' ? rtrim($this->path, '/') : $this->path;
--- a/src/Illuminate/Pagination/LengthAwarePaginator.php
+++ b/src/Illuminate/Pagination/LengthAwarePaginator.php
@@ -46,8 +46,9 @@ class LengthAwarePaginator extends AbstractPaginator implements Arrayable, Array
         }
 
         $this->total = $total;
-        $this->perPage = $perPage > 0 ? (int) $perPage : 1;
-        $this->lastPage = max((int) ceil($this->total / $this->perPage), 1);
+        $this->perPage = $perPage;
+        $perPage = $perPage > 0 ? $perPage : 1;
+        $this->lastPage = max((int) ceil($this->total / $perPage), 1);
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Thanks for the PR. Are you able to add a test covering this, please?

Aethedigm:
Might need to look further into the unit test, as it passes in my local.

tests/Pagination/LengthAwarePaginatorTest.php
```diff
@@ -124,4 +124,11 @@ public function testItRetrievesThePaginatorOptions()
     {
         $this->assertSame($this->options, $this->p->getOptions());
     }
+
+    public function testLengthAwarePaginatorShouldAllowForZeroPerPage()
+    {
+        $p = new LengthAwarePaginator($array = ['item1', 'item2', 'item3', 'item4'], 4, 0, 2, $this->options);
+
+        $this->assertSame(0, $p->toArray()['per_Page']);
```
BrandonSurowiec, line 132: Why not use the methods that are readily available?

```
$this->assertSame(0, $p->perPage());
// $this->assertSame(?, $p->lastPage());
// etc.
```

It would also help if you showed an example in your PR with how you are using this code.

Aethedigm: Locally we are using query builder and retrieving results directly back from our database. As far as operating directly against last page, my goal was to stop the application from dying on a 0.

I am currently learning my way around unit tests however, Is there a particular reason I should assert on lastPage over perPage?

Aethedigm: Made the suggested change, also realized that the member perPage not keeping value might be an unwelcome change for users. Keeps 0 value, avoids 0 division now.

src/Illuminate/Pagination/LengthAwarePaginator.php
```diff
@@ -46,8 +46,9 @@ public function __construct($items, $total, $perPage, $currentPage = null, array
         }
 
         $this->total = $total;
-        $this->perPage = (int) $perPage;
```
BrandonSurowiec, line 49: The `(int)` cast should not be removed.

Aethedigm: Accidental oversight. Fixing now.

taylorotwell:
I would honestly rather just throw an exception on 0.

dennisprudlo:
I also think an exception is better. It makes no sense to pass 0 as the `$perPage` argument. A paginator will always need at least 1 item per page, even if the collection of items to paginate itself is empty. So instead of allowing 0 i would suggest to do this in userland:

```php
$items = [ ... ];
$paginator = new LengthAwarePaginator($items, count($items), max(1, count($items))
```

Aethedigm:
> I would honestly rather just throw an exception on 0.

Just to pick your brain a bit if you have the time, do you prefer throwing an exception because it's not the intended behavior for the class?


> _In this example the paginator is useless, I just mean we can prevent a 0 `$perPage` value when creating an instance_

This will probably be the way we move forward

PR #42887, opened 2022-06-20
[9.x] Don't ignore nested chained jobs while dispatching next job in chain

zlodes:
It sounds kinda crazy, but...
`Queueable::dispatchNextJobInChain` replaces chained job `chained` property with the rest of current job's chained jobs.

But chained job also may have it's chained jobs... 

## Code example:

```php
$firstLevelJob = (new JobLevelFirst(1))->chain([
    (new JobLevelSecond(1))->chain([
        new JobLevelThird(1),
        new JobLevelThird(2),
        new JobLevelThird(3),
    ]),
    (new JobLevelSecond(2))->chain([
        new JobLevelThird(3),
        new JobLevelThird(4),
        new JobLevelThird(5),
    ]),
]);

$dispatcher->dispatch($firstLevelJob);
```

## Expected result: 
All the jobs were dispatched.
## Actual result: 
Only JobLevelFirst and JobLevelSecond jobs were dispatched.

## What is happening there?

There are three commits:
* First contains only test. 
* The second one fixes the test.
* The third one changes `array_merge` with `array_push` because it's faster

### DIFF
--- a/src/Illuminate/Bus/Queueable.php
+++ b/src/Illuminate/Bus/Queueable.php
@@ -249,7 +249,7 @@ trait Queueable
     {
         if (! empty($this->chained)) {
             dispatch(tap(unserialize(array_shift($this->chained)), function ($next) {
-                $next->chained = $this->chained;
+                $next->chained = array_merge($next->chained, $this->chained);
 
                 $next->onConnection($next->connection ?: $this->chainConnection);
                 $next->onQueue($next->queue ?: $this->chainQueue);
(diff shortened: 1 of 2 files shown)


driesvints:
I believe this is by design. I don't think it's meant to nest chained jobs like that as we don't document that: https://laravel.com/docs/9.x/queues#job-chaining

zlodes:
@driesvints maybe, but it's very handy ability to build jobs tree.
In the current job implementation we should dispatch job from another job to reach the same result. 
It isn't a good idea to ignore nested chained jobs — it's really a bug. Also this PR won't break anything. :-)

Also I can make a PR to the docs repo with this functionality.

taylorotwell:
I'm a little hesitant to merge this on a patch release, especially since there would be no behavioral difference between this and just passing one array containing the entire chain which you can already do.

zlodes:
@taylorotwell there is behavioral difference in case of non-sync conneciton.

PR #38869, opened 2021-09-19
[8.x] Fix getDirty method when using `AsArrayObject` / `AsCollection`

gdebrauwer:
A less impactful version of #38774.

I updated the `originalIsEquivalent` method to make a custom comparison when the attribute uses `AsArrayObject` or `AsCollection` cast. This way, Eloquent will not think there is a difference if there are just some extra spaces in the original json string.

```php
// If you try to save this in a MySQL json column:
// {"foo":"bar"}
// Then MySQL will save it like this:
// {"foo": "bar"}

// This means that Eloquent currently thinks that a json value is changed
// because there is an extra space in the original value.
'{"foo": "bar"}' !== '{"foo":"bar"}'
```

PR #42888, opened 2022-06-20
[9.x] Add support to detect dirty encrypted model attributes

PhiloNL:
Currently, a model always determines that an encrypted attribute is modified when you fill a model, even if the value that is being encrypted is the same. This is because it compares the encrypted string, which changes every time you encrypt something regardless of whether the encrypted value is the same.

This PR will add support to detect dirty encrypted model attributes.
```
$model = Model::create(['billing_address' => ['street' => 'foo']);

$model->fill(['billing_address' => ['street' => 'foo']);
$model->originalIsEquivalent('billing_address'); // Returns false while it must return true
```

I wasn't sure if the tests should be part of `EloquentModelEncryptedCastingTest` or `DatabaseEloquentModelTest`. If they need to be moved, let me know 🙌🏻

This follows up on https://github.com/laravel/framework/pull/38869 and the question from Taylor https://github.com/laravel/framework/pull/38869#issuecomment-925117393.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -11,6 +11,8 @@ use Illuminate\Contracts\Database\Eloquent\CastsInboundAttributes;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Database\Eloquent\Casts\AsArrayObject;
 use Illuminate\Database\Eloquent\Casts\AsCollection;
+use Illuminate\Database\Eloquent\Casts\AsEncryptedArrayObject;
+use Illuminate\Database\Eloquent\Casts\AsEncryptedCollection;
 use Illuminate\Database\Eloquent\Casts\Attribute;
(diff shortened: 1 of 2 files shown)


PhiloNL:
Point of discussion: https://github.com/laravel/framework/blob/b9203fca96960ef9cd8860cb4ec99d1279353a8d/src/Illuminate/Database/Eloquent/Casts/AsEncryptedCollection.php#L31-L38

All though this PR fixes the `$model->originalIsEquivalent()`, `$model->isDirty()` behavior calling `$model->update()` will still update the database with a new encrypted key even though the contents is the same. Preferrably this update is skipped because it's unnecessary. 

A possible solution is to add a check to the`set()` method on the `caster` and only encrypt if the content is different:

```php
{
    if (! is_null($value)) {
        if(json_encode($model->getOriginal($key)) === json_encode($value)) {
            return [$key => $model->getRawOriginal($key)];
        }

        return [$key => Crypt::encryptString(json_encode($value))];
    }

    return null;
}
```

PR #42890, opened 2022-06-21
dynamic database connection get from the new configuration file

manojkumar-net:
Users will be able to split projects more dynamically, including config files.
Don't have to write all database configurations in one file(config/database.php).


manojkumarlinux:
```
Config::set('database.connections.mysql', Config::get('modules.database.connection'));
  
DB::purge('mysql');
```
Now I understand the purge method. 

 https://stackoverflow.com/questions/46775265/change-the-database-connection-dynamically-in-laravel

PR #42909, opened 2022-06-22
[9.x] Add "orWhere" to Collection

imanghafoori1:
This PR adds the missing `orWhere` method to collection and LazyCollection.
The `orWhere` method is designed to only be called immediately after a `where` method call to add to it.

```php
  $results = $c->where('v', '<', 2)->orWhere('v', '>', 5) ->values()->all();
```
- Tests are included.

### DIFF
--- a/tests/Support/SupportCollectionTest.php
+++ b/tests/Support/SupportCollectionTest.php
@@ -928,6 +928,34 @@ class SupportCollectionTest extends TestCase
     /**
      * @dataProvider collectionClassProvider
      */
+    public function testOrWhere($collection)
+    {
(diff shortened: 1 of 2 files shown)


tests/Support/SupportCollectionTest.php
```diff
@@ -5130,6 +5158,129 @@ public function collectionClassProvider()
             [LazyCollection::class],
         ];
     }
+
+    public function testRenameKeys()
```
fgaroby, line 5162: Why is there this test? It doesn't test your new method `orWhere`...

imanghafoori1: hmm... it is committed by mistake. Thanks for mentioning it.

PR #42912, opened 2022-06-22
[9.x] Add "orWhere" to collection

imanghafoori1:
This PR adds the missing `orWhere` method to collection and LazyCollection.
The `orWhere` method is designed to only be called immediately after a `where` method call to add to it.

```php
  $results = $c->where('v', '<', 2)->orWhere('v', '>', 5) ->values()->all();
```
- Tests are included.
- This is not included in the `LazyCollection` since it requires a more elaborate solution. Simply I was not able to workout a solution for the LazyCollections. Maybe @JosephSilber can.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -32,6 +32,13 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     protected $items = [];
 
     /**
+     * The items contained in the collection before applied where clause.
+     *
+     * @var array<TKey, TValue>
+     */
+    protected $itemsBeforeWhere = [];
+
+    /**
      * Create a new collection.
      *
      * @param  \Illuminate\Contracts\Support\Arrayable<TKey, TValue>|iterable<TKey, TValue>|null  $items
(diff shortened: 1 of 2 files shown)


JosephSilber:
Do we really want to carry around all of the original memory whenever someone calls `where`? That's really unexpected.

```php
function get_stuff() {
    return fetch_something_huge()->where('name', 'Iman');
}
```

With the current change, the memory of that original huge collection cannot be deallocated. This is quite unexpected!

JosephSilber:
It might make more sense to have it as a separate method. I don't have a good name for it ATM, but something like this:

```php
$items->whereChain(fn ($items) => $items->where(...)->orWhere(...));
```

imanghafoori1:
@JosephSilber after calling the next method the memory gets free-up and the property will get deleted with the object.
I do not think that will have a huge impact on memory.

In this example when we call `contains` the items get deleted.
and they are loaded into memory before we call `where` anyways.
```php
function get_stuff() {
    return fetch_something_huge()->where('name', 'Iman')->contains()->value()->all();
}
```

johanrosenson:
What about naming it `whereAny` instead? Which takes a list of criterias where only one of them have to be true.

```
$collection->orWhere([
    'key' => 'value',
    'another_key' => 'another_value',
    ['key', '<', 100],
    ...
]);
```

That would solve the memory issue, allow more than one chained `orWhere` (your solution doesn't allow `->where()->orWhere()->orWhere()` right?) and also not introduce a new `orWhere` that works differently than the `orWhere` on the _query builder_ (to avoid confusion).

taylorotwell:

PR #42918, opened 2022-06-23
[10.x] Restrict `__()` helper to return `string|null`

crynobone:
Target branch: 10.x due to breaking change 
Related issues:
- https://github.com/laravel/nova-issues/issues/4467

-------

`__()` is commonly used to translate JSON or short key translation value,
however, since it replicates the functionality of `trans()` method it is
also possible to return the whole `array` if the given key matches
a translation file which causes the following limitation:

image

The changes would restrict those use-case and instead would return translated
`$key`. Developer who wishes to still return `array` should use
`trans()` method instead.

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -880,7 +880,7 @@ if (! function_exists('__')) {
      * @param  string|null  $key
      * @param  array  $replace
      * @param  string|null  $locale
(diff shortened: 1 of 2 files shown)


driesvints (approved):
Good idea 👍

taylorotwell:
Not sure I'm going to make this breaking change since we fixed it on the Nova side.

PR #42922, opened 2022-06-23
Escape values before checking for their presence in HTML part of Mailables

turanjanin:
Laravel offers the ability to assert presence of a value inside the HTML for both views and mailables. However, the implementation of these assertions differ in a way that may cause confusion and occasional failing tests:

`TestResponse::assertSee()` escapes the HTML by default,
`Mailable::assertSeeInHtml()` does not escape the HTML.

This means that, for the same HTML, `$response->assertSee("O'Connor");` will pass, while `$mail->assertSeeInHtml("O'Connor");` will fail.

The goal of this PR is to align the behaviour of Mailable assertion methods with the rest of the similar assertion methods.

### DIFF
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -976,10 +976,13 @@ class Mailable implements MailableContract, Renderable
      * Assert that the given text is present in the HTML email body.
      *
      * @param  string  $string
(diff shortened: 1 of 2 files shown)


driesvints:
This PR unfortunately contains breaking changes to method signatures. Please send this to master.

PR #42928, opened 2022-06-23
[9.x]Facade mock should not be created from a final class

leo108:
It will show the following error message if try to create a Facade mock with final class before this PR

> Mockery\Exception: The class \Illuminate\Tests\Support\FinalClass is marked final and its methods cannot be replaced. Classes marked final can be passed in to \Mockery::mock() as instantiated objects to create a partial mock, but only if the mock is not subject to type hinting checks.


taylorotwell:

leo108:
@taylorotwell Yes, this should be a bugfix instead of a new syntactic sugar

PR #42931, opened 2022-06-23
Add collect method to config

Kyzegs:
I've added the collect method to the config repository. I see this being useful for quite a few users:

Users might have arrays within their configuration that they want to turn into a collection. What the user needs to do now to achieve this:

```php
collect(config('application.commands'));
```

The new approach will allow for the following:

```php
config()->collect('application.commands');
```

`\Illuminate\Http\Request` already has this method and is widely adopted. Therefore I reckon this might be useful for quite a lot of users.


taylorotwell:

Issue #42932, opened 2022-06-24
undefined constant error when try to define dynamic slot names


monurakkaya:
- Laravel Version: 9.18.0
- PHP Version: 8.1.2
- Database Driver & Version:

### Description:

Since v9.15, After the merge of this  pull request: #42574 
```
<x-slot name="$language->code"> 
```
throws error as undefined constant "code". Because compiled view is looking like this:

`<?php $__env->slot($language>code null, []); ?> <?php echo e($language->name); ?> <?php $__env->endSlot(); ?>`
instead
`<?php $__env->slot($language->code null, []); ?> <?php echo e($language->name); ?> <?php $__env->endSlot(); ?>`
Because it's trying to make camel case `$language->code`

### Steps To Reproduce:

create a blade component: 

`php artisan make:component Test`

In any blade file: 

(It does not matter content of the `x-test` component below)

```
@php
    $languages = [
        [
            'name' => 'Turkish',
            'code' => 'tr'
        ],
        [
            'name' => 'English',
            'code' => 'en'
        ]
    ];

    $languagesAsArrayOfObjects = [
        (object)[
            'name' => 'Turkish',
            'code' => 'tr'
        ],
        (object)[
            'name' => 'English',
            'code' => 'en'
        ]
    ];
@endphp


##Works
<x-test>
    @foreach($languages as $language)
        <x-slot :name="$language['code']">{{ $language['name'] }}</x-slot>
    @endforeach
</x-test>

<br><br>

##Works
<x-test>
    <x-slot name="tr">Turkish</x-slot>
    <x-slot name="en">English</x-slot>
</x-test>


<br><br>

###Does not work (undefined constant code)
<x-test>
    @foreach($languagesAsArrayOfObjects as $language)
        <x-slot :name="$language->code">{{ $language->name }}</x-slot>
    @endforeach
</x-test>

<br><br>

##Works

<x-test>
    @foreach($languagesAsArrayOfObjects as $language)
        @php($code = $language->code)
        <x-slot :name="$code">{{ $language->name }}</x-slot>
    @endforeach
</x-test>

```

driesvints:
Closing this as a PR was opened. Thanks

PR #42933, opened 2022-06-24
fix undefined constant error on dynamic slot names as key of object

monurakkaya:
It's solution for #42932

I've described problem in the issue. 

With 9.15 and after we cannot use `<x-slot :name="$model->name">` because it is rendering as  
`<?php $__env->slot($model>name, null, []); ?>` 
instead 
`<?php $__env->slot($model->name, null, []); ?>`


dennisprudlo:
```php
<x-slot :name="$model->name">
```
That should work

driesvints:
@dennisprudlo I guess they want to pass it as a literal string.

monurakkaya:
@dennisprudlo I updated the example. Forgot to put `:`.

taylorotwell:
Please resubmit with a test.

PR #42934, opened 2022-06-24
[10.x] Allow closure based rules to push messages to nested (or other…

nshiro:
This PR allows closure based rules to push messages to nested (or other) attributes.
This is almost copy and paste from Tim's PR. #42801

I also removed a test method which asserts `expectException` because the `translate` method of PotentiallyTranslatedString class does not throw RuntimeException any more when there's no translation key. #42873


timacdonald:
Thanks @nshiro. This was on my todo list and you beat me to it ❤️

nshiro:
Thanks always. @timacdonald
You're on fire.❤️🔥

PR #42935, opened 2022-06-24
[9.x] Vite assets have deduplicated CSS <link>s

Magnesium38:
If two entrypoints are used that end up importing the same CSS chunk, multiple identical `<link>`s are output. This just makes the `$stylesheets` collection unique first. One caveat on this PR is a situation in which the `<link>`s are not consecutive.

```html
<!-- Current Output -->
<link rel="stylesheet" type="text/css" href="first.css">
<link rel="stylesheet" type="text/css" href="second.css">
<link rel="stylesheet" type="text/css" href="first.css">

<!-- Updated Output -->
<link rel="stylesheet" type="text/css" href="first.css">
<link rel="stylesheet" type="text/css" href="second.css">
```

Because CSS rules of equal specificity takes the later rule, the duplicated `first.css` stylesheet ends up overriding any overrides that came from the `second.css` stylesheet.

A trivial fix would be to chain more collection methods to ensure that the last duplicate of a file is the one that's retained. Something like `$stylesheets->reverse()->unique()->reverse()->join('')` would do it. This fix (or a more performant version) isn't included in this PR because I felt that the resulting order of the stylesheets shouldn't be a concern of this code. However, I know that my opinion comes from a place of bias. Nearly every project I work on has a singular stylesheet consisting of utility classes. If desired, I would be able to update this PR to account for it.

### DIFF
--- a/src/Illuminate/Foundation/Vite.php
+++ b/src/Illuminate/Foundation/Vite.php
@@ -75,7 +75,7 @@ class Vite
 
         [$stylesheets, $scripts] = $tags->partition(fn ($tag) => str_starts_with($tag, '<link'));
 
-        return new HtmlString($stylesheets->join('').$scripts->join(''));
+        return new HtmlString($stylesheets->unique()->join('').$scripts->join(''));
     }
 
     /**
--- a/tests/Foundation/FoundationViteTest.php
+++ b/tests/Foundation/FoundationViteTest.php
@@ -74,6 +74,19 @@ class FoundationViteTest extends TestCase
         );
     }
 
+    public function testViteWithDuplicateCssImport()
+    {
+        $this->makeViteManifest();
+
+        $result = (new Vite)(['resources/js/app-with-duplicate-css.js']);
+
+        $this->assertSame(
+            '<link rel="stylesheet" href="https://example.com/build/assets/common.versioned.css" />'
+            .'<script type="module" src="https://example.com/build/assets/app-with-duplicate-css.versioned.js"></script>',
+            $result->toHtml()
+        );
+    }
+
     public function testViteHotModuleReplacementWithJsOnly()
     {
         $this->makeViteHotFile();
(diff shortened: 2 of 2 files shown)


jessarcher:
Hi @Magnesium38,

Are you loading both entry points on the same page? Or do the styles get duplicates even if you're only loading one of the entry points?

taylorotwell:
Marking as draft until @jessarcher marks as ready for review.

Magnesium38:
I've been unsuccessful in attempting to generate a manifest that results in duplicate CSS when given a singular entrypoint. I've been able to generate plenty of manifests that would do it given multiple entrypoints though for what it's worth. However, with my tinkering I found another potential issue. Again, I haven't been able to actually produce a manifest that contains the issue, however writing one manually with it is trivial.

Currently, the directive takes some entrypoints, iterates over them and their imports to retrieve a list of CSS files to convert into `<link>`s. It is possible for an import to have imports of it's own, that may again have CSS files. This could be recursive for an arbitrary amount of layers. I briefly investigated this and couldn't product an issue definitively. The nested imports were always duplicated at the parent level. I took a look at the [manifest plugin's source](https://github.com/vitejs/vite/blob/main/packages/vite/src/node/plugins/manifest.ts) to see if that was something we could rely on, but I'm familiar enough with Vite plugins to be certain.

I, personally, would err on the side of not trusting it and would instead base the implementation on the potential file structure. It may be possible that another plugin could interfere with that guarantee that may or may not actually exist. I haven't taken the time to create a PR with a recursive solution instead yet, but I would be able to get one made later today/tomorrow if there's interest.

timacdonald:
@Magnesium38 Thanks for your PR. I'm gonna close this one until we can create an actual reproduction of the issue we are trying to fix, without manually editing the manifest.

If you do find the issue and can reproduce it, please let us know - and thanks again!

PR #42936, opened 2022-06-24
[9.x] Add modulepreload links for Vite assets

Magnesium38:
It is currently possible to end up with a waterfall problem where the entrypoints end up importing a chunk that was externalized to be shared between various entrypoints. Depending on the how the dependencies are determined, it could turn into a rather large waterfall of requests. This PR adds a step of converting the entrypoint's imports into `<link rel="modulepreload">`s.

`modulepreload` isn't terribly [well supported](https://caniuse.com/link-rel-modulepreload) right now. However, the [Vite documentation](https://vitejs.dev/config/#build-polyfillmodulepreload) says to import a polyfill to account for it. I think that how Laravel handles the inclusion of that polyfill as beyond the scope of this PR. As far as I'm aware, browsers that don't understand `modulepreload` will just ignore them. I believe this PR could be merged without any negative impact and then followed up with another one to resolve the handling of the polyfill once a decision is made.

### DIFF
--- a/src/Illuminate/Foundation/Vite.php
+++ b/src/Illuminate/Foundation/Vite.php
@@ -64,6 +64,8 @@ class Vite
 
             if (isset($manifest[$entrypoint]['imports'])) {
                 foreach ($manifest[$entrypoint]['imports'] as $import) {
+                    $tags->push($this->makeModulePreloadTag(asset("{$buildDirectory}/{$manifest[$import]['file']}")));
+
                     if (isset($manifest[$import]['css'])) {
                         foreach ($manifest[$import]['css'] as $css) {
                             $tags->push($this->makeStylesheetTag(asset("{$buildDirectory}/{$css}")));
@@ -131,6 +133,17 @@ class Vite
     protected function makeScriptTag($url)
     {
         return sprintf('<script type="module" src="%s"></script>', $url);
+    }
+
+    /**
+     * Generate a modulepreload tag for the given URL.
+     *
+     * @param  string  $url
+     * @return string
+     */
+    protected function makeModulePreloadTag($url)
+    {
+        return sprintf('<link rel="modulepreload" href="%s" />', $url);
     }
 
     /**
--- a/tests/Foundation/FoundationViteTest.php
+++ b/tests/Foundation/FoundationViteTest.php
@@ -35,6 +35,19 @@ class FoundationViteTest extends TestCase
         $this->assertSame('<script type="module" src="https://example.com/build/assets/app.versioned.js"></script>', $result->toHtml());
     }
 
+    public function testViteWithNestedJs()
+    {
+        $this->makeViteManifest();
+
+        $result = (new Vite)('resources/js/parent.js');
+
+        $this->assertSame(
+            '<link rel="modulepreload" href="https://example.com/build/assets/child.versioned.js" />'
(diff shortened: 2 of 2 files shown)


jessarcher:
Hi @Magnesium38,

Thanks for this. With regards to the polyfill, I think the easiest solution would be to add the import to the `app.js` file in the skeleton, Jetstream, and Breeze. Unless you can think of a better solution?

Magnesium38:
Hey @jessarcher,

Yeah, I think including it in the skeletons and such would be a reasonable option, although personally I can say that I always immediately blow away the skeletons and it should be called out in the docs somewhere in the Vite section if that's the route that's chosen. It's definitely the simplest solution because it means that it isn't something that Laravel has to maintain in any way.

I had the thought that if Laravel were to generate a `modulepreload` link that it could also create an additional script that includes the polyfill. It's fairly lightweight so I don't think it would be a terrible solution. I think this falls apart once we take other Vite plugins into account. To be fair, they should have the chance to do work on the polyfill if they wanted to.

That could maybe be alleviated by always creating an entrypoint that includes only the polyfill that would also get added transparently. As in the laravel-vite-plugin would inject an additional entrypoint that has a known name, then the `@vite` directive would additionally output that chunk if it necessary.

On the plugin side, I haven't looked at all if it would be possible, but I think the most reasonable solution would be if the laravel-vite-plugin would automatically inject the polyfill into the entrypoints based on the same Vite config property (`build.polyfillModulePreload`). That way it's transparent on the framework side and is solely the responsibility of the plugin. I feel like this is the most appropriate solution and the closest to how Vite would work on it's own.

timacdonald:
@Magnesium38 thanks again for this PR. We have been experimenting on our end with this.

We have been unable to create a situation where we found that `modulepreload` had a noticeable positive impact on load times - even in Chrome where the feature is natively supported.

We also found that the polyfill made things _slower_ in other browsers as it would attempt to fetch the same resource multiple times, often fetching the same resource twice _at the same time_.

With back-end implementations, because the polyfill is in the entrypoint, it means that the polyfill is not even activated until after the entrypoint has been loaded and executed. We have considered injecting the polyfill for developers before the script tags instead of having it in the entrypoints, but we still were not able to find a positive impact.

Although I understand in theory this feature makes sense and it is available in other back-end Vite implementations, we are hesitant to merge it until we can see a reproduction of the feature having a noticeable positive impact on performance. It could just be our contrived testing that is the problem.

If we can see this feature in action and having an improvement, we would re-consider it for sure.

Thanks again.

Magnesium38:
@timacdonald Thanks for the detailed response on this. I can agree that waiting on this is sensible choice. I knew going into creating these PRs that I likely wouldn't be able to prove what was broken. It's always difficult to simulate real world usage.

I made this PR and #42935 because I have a project that I'm wanting to immediately pivot into using Vite and they both address potential issues that I discovered whilst playing around with the prerelease. The project scale is small enough to likely not be affected by the potential waterfall issue and we use Tailwind for the CSS so duplicate files is pretty unlikely too. However, it'll be a better playground to test performance against at the least. Once the project is decently established I'll do some performance testing and follow up with some more tangible data if I can see any notable results.

Again, thanks for the detailed response. It's much appreciated!

timacdonald:
Would love you to follow up with anything you find! Thank you ❤️

PR #42938, opened 2022-06-24
[9.x] Adds `Json` utility class

DarkGhostHunter:
## What?

A utility class to manipulate JSON trees. 

```php
use Illuminate\Support\Json;

Json::make(['foo', 'bar']);
```

It's _memory friendly_. It uses `data_get()` and `data_set()` behind the scenes.

> Memory friendly as it never transforms to array objects until serialization. It also decodes JSON into objects instead of arrays.

## Why?

Because sometimes JSON is not a list of values, but rather a data tree (like the app config), that can be very cumbersome to parse and manipulate.

Usually, there are workarounds to manipulate JSON:

- If it's part of the request, you would use `json()` to retrieve all or part of the contents.
- If it's part of a response, you would use `data_get()` and `data_set()` on the `getContents()`.
- You would extend the Config `Repository` class.
- You would use Collections, and travel through a given key.
- Manipulate an array using `Arr`
- Use `data_get()` and `data_set()` indiscriminately.

Instead, the point of having a `Json` object is to centralize these operations into one object that can be shared across the application without the risk of memory duplication (much like Collections).

```php
use Illuminate\Support\Json;

$json = Json::make(['foo' => ['bar', 'baz']]);

$json->get('foo.bar'); // "baz"
```

It supports a lot of methods to make it convenient, all using dot notation like you would with the application config instance: get, set, forget, has, missing, dynamic access and array access. It also can be totally or partially transformed to a Collection using `collect()`, and is _iterable_.

```php
$collection = $json->collect()->map(fn ($item) => "$item is cool");

foreach ($json as $key => $item) {
    // ...
}
```

Of course, it also transforms to JSON when casting it to a string.

```php
use Illuminate\Support\Json;

$json = Json::make(['foo' => ['bar', 'baz']]);

(string) $json; // {"foo":{"bar":"baz"}}
```

It will also play nice with a Collections, as these enter as objects. Since objects are never transformed before JSON serialization, values can be edited without needing to use the `Json` instance itself.

```php
use Illuminate\Support\Json;

$collection = collect(['bar' => 'baz']);

$json = Json::make();

$json->set('foo', $collection);

$collection->put('bar', 'qux');

echo $json->get('foo.bar'); // "qux"
```

### DIFF
--- a/src/Illuminate/Collections/Json.php
+++ b/src/Illuminate/Collections/Json.php
@@ -24,7 +24,7 @@ class Json implements Stringable, ArrayAccess, JsonSerializable, IteratorAggrega
     protected $data;
 
     /**
-     * Create a new JSON instance
+     * Create a new JSON instance.
      *
      * @param  mixed  $data
      */
@@ -101,7 +101,7 @@ class Json implements Stringable, ArrayAccess, JsonSerializable, IteratorAggrega
      */
     public function missing($key)
     {
-        return !$this->has($key);
+        return ! $this->has($key);
     }
 
     /**
@@ -259,7 +259,7 @@ class Json implements Stringable, ArrayAccess, JsonSerializable, IteratorAggrega
     }
 
     /**
-     * Retrieve an external iterator
+     * Retrieve an external iterator.
      *
      * @return \ArrayIterator
      */
--- a/src/Illuminate/Collections/Json.php
+++ b/src/Illuminate/Collections/Json.php
@@ -110,7 +110,7 @@ class Json implements Stringable, ArrayAccess, JsonSerializable, IteratorAggrega
      * @param  string|int  $key
      * @return void
      */
-    public function unset($key)
+    public function forget($key)
     {
         $segment = $this->data;
 
@@ -190,7 +190,7 @@ class Json implements Stringable, ArrayAccess, JsonSerializable, IteratorAggrega
      */
     public function __unset($name): void
     {
-        $this->unset($name);
+        $this->forget($name);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


BrandonSurowiec:
Seems like there is little to gain versus using a collection directly:

```php
collect(['a' => 3, 'b'])->toJson(); // "{"a":3,"0":"b"}"
```

The only thing a collection can't do is a `get()` with "array dot" syntax. 

```php
collect(['a' => 1, 'b' => ['c' => ['a', 'y']]])->toJson(); // "{"a":1,"b":{"c":["a","y"]}}"
collect(['a' => 1, 'b' => ['c' => ['a', 'y']]])->get('b.c'); // null versus expected ['a', 'y']
```

So `$collect->get('*.*')` will always return `null` on a collection.

dennisprudlo:
It's not fluent but you can read from a collection with dot notation using `data_get`.
```php
$collection = collect(['a' => 1, 'b' => ['c' => ['a', 'y']]]);

data_get($collection, 'b.c'); // [ 'a', 'y' ]
```

So maybe a fluent method on the collection to retrieve data using dot notation would suffice. I think I saw a PR every now and then but they were rejected. Don't know why, though.

DarkGhostHunter:
I see these points, but I seems you haven't worked with this kind of JSON:

```json
{
  "glossary":{
    "title":"example glossary",
    "GlossDiv":{
      "title":"S",
      "GlossList":{
        "GlossEntry":{
          "ID":"SGML",
          "SortAs":"SGML",
          "GlossTerm":"Standard Generalized Markup Language",
          "Acronym":"SGML",
          "Abbrev":"ISO 8879:1986",
          "GlossDef":{
            "para":"A meta-markup language, used to create markup languages such as DocBook.",
            "GlossSeeAlso":[
              "GML",
              "XML"
            ]
          },
          "GlossSee":"markup"
        }
      }
    }
  }
}
```

taylorotwell:

PR #42942, opened 2022-06-24
[9.x] Fix default parameter bug

### LINKED ISSUE #42707: Route generation broken
- Laravel Version: 9.17.0
- PHP Version: 8.1.0

### Description:
When updating to Laravel > 9.13 route generation breaks.
I tested with 9.14, 9.15, 9.16 and 9.17.
I think the bug got introduced by #42425 

### Steps To Reproduce:

Route in web.php:
```PHP
Route::group([
    'prefix' => 'extranet',
    'as' => 'extranet.',
    'middleware' => 'auth',
], function () {
    Route::get('gallery/{folder:slug}/albums', [GalleryFolderAlbumsController::class, 'index'])->name('gallery-folders.albums.index');
});
```

Laravel v9.13 (correct behaviour):
```PHP
dd(route('extranet.gallery-folders.albums.index', $folder)); // -> http://localhost/de/extranet/gallery/aut-neque-dicta-qui-possimus-porro-sit-vel-iusto-voluptatem-quia-ut-excepturi-quaerat-labore-nobis-nesciunt-id-alias-voluptas-voluptatem/albums
```

Laravel v9.17 (wrong behaviour):
```PHP
dd(route('extranet.gallery-folders.albums.index', $folder)); // -> http://localhost/de/extranet/gallery/1/albums
```

driesvints:
This took me the better part of the day to figure out but I finally found a solution that resolves https://github.com/laravel/framework/issues/42707. After the introduction of https://github.com/laravel/framework/pull/42425 several bugs got introduced as well. This is the last remaining one (we know of so far).

Let me try to explain the reasoning here. In the `compileParameterNames` of the `Route` object we return all parameter names of a route. These are mapped in an index array from 0 to up. Because this method has no knowledge of what a default parameter is it incorrectly matches a default param instead of the correct regular param. What we need to do is filter out the default params here before we try to map the regular passed params that we feed to the `route` method. The `Route` object does not have any knowledge about these so we need to pass in to the route so it can filter the params out. Afterwards we can match our own params properly with the non-default ones.

This fix is a bit of a crutch and I'm not 100% sure it's the correct one but without more examples I cannot make sure it's going to cover all use cases. At least all tests pass now.

Fixes https://github.com/laravel/framework/issues/42707

### DIFF
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -94,6 +94,13 @@ class Route
     protected $originalParameters;
 
     /**
+     * The named parameter defaults.
+     *
+     * @var array
+     */
+    public $defaultParameters = [];
+
+    /**
      * Indicates "trashed" models can be retrieved when resolving implicit model bindings for this route.
      *
      * @var bool
@@ -507,9 +514,11 @@ class Route
     {
         preg_match_all('/\{(.*?)\}/', $this->getDomain().$this->uri, $matches);
 
-        return array_map(function ($m) {
+        return array_values(array_filter(array_map(function ($m) {
             return trim($m, '?');
-        }, $matches[1]);
+        }, $matches[1]), function ($parameterName) {
+            return ! array_key_exists($parameterName, $this->defaultParameters);
+        }));
     }
 
     /**
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -94,11 +94,11 @@ class Route
     protected $originalParameters;
 
     /**
-     * The named parameter defaults.
+     * The parameters to exlude when determining the route's parameter names.
      *
      * @var array
      */
-    public $defaultParameters = [];
+    public $excludedParameters = [];
 
     /**
      * Indicates "trashed" models can be retrieved when resolving implicit model bindings for this route.
@@ -517,7 +517,7 @@ class Route
         return array_values(array_filter(array_map(function ($m) {
             return trim($m, '?');
         }, $matches[1]), function ($parameterName) {
-            return ! array_key_exists($parameterName, $this->defaultParameters);
+            return ! array_key_exists($parameterName, $this->excludedParameters);
         }));
     }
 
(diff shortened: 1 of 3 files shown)


mateusjunges:
Just tested it locally and the route is generated as expected.

taylorotwell:
Feels a bit like we're entering into kinda rough territory here setting public properties before compiling, etc. I guess I will go ahead and merge this but if we run into anymore issues at all I think we should revert #42425

cole-flexsim:
@driesvints @taylorotwell This causes issues with Octane and Url Defaults as described in this post I made on the laravel discord server:  

> I have routes like so:
> 
> ```
> /teams/{team}/products
> /teams/{team}/settings
> ```
> 
> I have a middleware (prioritized before substitute bindings like in the docs) that sets the default value for a $team route parameter to be the value of the current route team: 
> 
> ```
> URL::defaults(['team' => $request->team])
> ```
> 
> This works when I first navigate to `/teams/myteam/products` and anytime I refresh or navigate to that page after that. However if I try to go to `/teams/myteam/settings` (or any other route with `{team}` in it) then the $team parameter is now null
> 
> this only happens on octane (I'm using the latest version of laravel 9 and octane). I don't remember having this issue on laravel 8 so maybe something in the framework changed

sebastiaanluca:
We upgraded Laravel, but one of our tests started failing. Running tests locally, without Octane enabled. Same issue as Cole.

In our test we're making a request, then follow the redirect returned from its response:
```php
$this
    ->actingAs($user)
    ->followingRedirects()
    ->post(route('checkout_cart', $store))
    ->assertViewIs('summary');
```

The route is defined as:
```php
Route::group(['domain' => config('app.domain'), 'prefix' => 'stores/{storeSlug}', 'middleware' => $middleware], function (): void {
    Route::get('cart/checkout/summary', ShowCartCheckoutSummary::class)->name('checkout_cart');
    });
```

We're also defining a default in a middleware:
```php
URL::defaults(['storeSlug' => $defaultStore->slug])
```

In in the initial request, the store (slug) is set (from the URL). In the second request, coming from the `response()->redirect()` returned from the checkout controller, `storeSlug` is `null`.

driesvints:
@sebastiaanluca we already reverted this pr and will be reverting all code related to the original pr as well soon. Tuesday's release will contain the fix.

PR #42943, opened 2022-06-24
[9.x] Fix undefined constant error when use slot name as key of object

monurakkaya:
It's solution for #42932

I've described problem in the issue. 

With 9.15 and after we cannot use `<x-slot :name="$model->name">` because it is rendering as  

`<?php $__env->slot($model>name, null, []); ?>` 

instead 

`<?php $__env->slot($model->name, null, []); ?>`


monurakkaya:
@taylorotwell I’ve added the test which will fail on 9.15 - 9.18 tags.

PR #42944, opened 2022-06-24
[9.x] Add `grep()` and `grepFilled()` to request object

ghost:
This PR introduces two new methods to the `Request` object.

Currently, you might check for certain key existence first then fill it into a variable like so:

```php
if ($request->has('name')) {
    $name = $request->input('name');
    /* ... */
}
```

With this PR now, you can do the following:

```php
if ($request->grep('name', $name)) {
    /* ... */
}
```

You can also need to check if the `input` is not empty like so:

```php
if ($request->filled('name')) {
    $name = $request->input('name');
    /* ... */
}
```

But now, you can do the following:

```php
if ($request->grepFilled('name', $name)) {
    /* ... */
}
```

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -566,4 +566,36 @@ trait InteractsWithInput
 
         return $this;
     }
+
+    /**
+     * Determine if the request contains a given input item key,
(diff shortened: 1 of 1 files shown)


ankurk91:
Rather than solving real world issues related to framework , developers are saving some keystrokes by introducing new methods to request class.

👎

taylorotwell:

PR #38124, opened 2021-07-24
[8.x] Add ignoreTrashed to Unique validation rule

PhiloNL:
Hi all!

I find myself adding `whereNull('deleted_at')` to the `Rule::unique()` quite often when working with models which can be soft deleted. This PR will add a simple helper to the unique validation rule. 

Have a great day! 😄 

```php
// Before
[
  'email'=> [
    Rule::unique('users')->whereNull('deleted_at'),
  ],
];

// After
[
  'email'=> [
    Rule::unique('users')->ignoreTrashed(),
  ],
];
````

PR #42948, opened 2022-06-25
[9.x] uniqueIgnoreTrashed index

morrislaptop:
It's currently difficult to add a unique index to a table with SoftDeletes, as the unique index will prevent duplicate rows of a single column, even if records have been deleted.  

Creating a composite unique index doesn't work as most databases ignore null columns for unique comparison. Further reading at https://www.dbi-services.com/blog/nulls-in-composite-keys/

This PR adds a shortcut method to add this functionality, via stored generated columns and a unique index. The generated column is set to NULL if deleted_at is set (allowing multiple NULLs) and to TRUE if deleted_at is not set (only allowing a single TRUE)

```php
$table->string('username')->uniqueIgnoreTrashed();
```

or

```php
$table->string('username');
$table->uniqueIgnoreTrashed('username');
```

This approach is supported by all databases, [however](https://www.enterprisedb.com/postgres-tutorials/postgresql-unique-constraint-null-allowing-only-one-null) [SQL Server will prevent multiple null values](https://dba.stackexchange.com/questions/80514/why-does-a-unique-constraint-allow-only-one-null)

Naming inspired from #38124

### DIFF
--- a/src/Illuminate/Database/Schema/Grammars/MySqlGrammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/MySqlGrammar.php
@@ -229,6 +229,24 @@ class MySqlGrammar extends Grammar
         return $this->compileKey($blueprint, $command, 'unique');
     }
 
+/**
+     * Compile a unique index ignoring a null column command.
+     *
+     * @param  \Illuminate\Database\Schema\Blueprint  $blueprint
+     * @param  \Illuminate\Support\Fluent  $command
+     * @return string
+     */
+    public function compileUniqueIgnoreNull(Blueprint $blueprint, Fluent $command)
+    {
+        return sprintf('create unique index %s on %s%s (%s, %s)',
+            $this->wrap($command->index),
+            $this->wrapTable($blueprint),
+            $command->algorithm ? ' using '.$command->algorithm : '',
+            $this->columnize($command->columns),
+            "(CASE WHEN {$this->wrap($command->nullColumn)} IS NULL THEN 0 ELSE {$this->wrap($command->nullColumn)} END)",
+        );
+    }
+
     /**
      * Compile a plain index key command.
      *
--- a/src/Illuminate/Database/Schema/Grammars/MySqlGrammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/MySqlGrammar.php
@@ -229,7 +229,7 @@ class MySqlGrammar extends Grammar
         return $this->compileKey($blueprint, $command, 'unique');
     }
 
-/**
+    /**
      * Compile a unique index ignoring a null column command.
      *
      * @param  \Illuminate\Database\Schema\Blueprint  $blueprint
--- a/src/Illuminate/Database/Schema/Grammars/MySqlGrammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/MySqlGrammar.php
@@ -230,24 +230,6 @@ class MySqlGrammar extends Grammar
     }
 
     /**
-     * Compile a unique index ignoring a null column command.
-     *
-     * @param  \Illuminate\Database\Schema\Blueprint  $blueprint
-     * @param  \Illuminate\Support\Fluent  $command
-     * @return string
-     */
-    public function compileUniqueIgnoreNull(Blueprint $blueprint, Fluent $command)
-    {
-        return sprintf('create unique index %s on %s%s (%s, %s)',
-            $this->wrap($command->index),
-            $this->wrapTable($blueprint),
-            $command->algorithm ? ' using '.$command->algorithm : '',
-            $this->columnize($command->columns),
-            "(CASE WHEN {$this->wrap($command->nullColumn)} IS NULL THEN 0 ELSE {$this->wrap($command->nullColumn)} END)",
-        );
-    }
-
-    /**
      * Compile a plain index key command.
      *
      * @param  \Illuminate\Database\Schema\Blueprint  $blueprint
--- a/src/Illuminate/Database/Schema/Grammars/SqlServerGrammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/SqlServerGrammar.php
@@ -140,6 +140,25 @@ class SqlServerGrammar extends Grammar
     }
 
     /**
+     * Compile a unique index ignoring a null column command.
+     *
+     * @param  \Illuminate\Database\Schema\Blueprint  $blueprint
+     * @param  \Illuminate\Support\Fluent  $command
+     * @return string
+     */
+    public function compileUniqueIgnoreNull(Blueprint $blueprint, Fluent $command)
+    {
+        $columns = $command->columns;
+        $columns[] = $command->nullColumn;
+
+        return sprintf('create unique index %s on %s (%s)',
+            $this->wrap($command->index),
+            $this->wrapTable($blueprint),
+            $this->columnize($columns)
+        );
+    }
+
+    /**
      * Compile a plain index key command.
      *
      * @param  \Illuminate\Database\Schema\Blueprint  $blueprint
--- a/src/Illuminate/Database/Schema/Grammars/SqlServerGrammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/SqlServerGrammar.php
@@ -148,13 +148,12 @@ class SqlServerGrammar extends Grammar
      */
     public function compileUniqueIgnoreNull(Blueprint $blueprint, Fluent $command)
     {
-        $columns = $command->columns;
-        $columns[] = $command->nullColumn;
-
-        return sprintf('create unique index %s on %s (%s)',
+        return sprintf('create unique index %s on %s%s (%s) where %s is null',
             $this->wrap($command->index),
             $this->wrapTable($blueprint),
-            $this->columnize($columns)
+            $command->algorithm ? ' using '.$command->algorithm : '',
+            $this->columnize($command->columns),
+            $this->wrap($command->nullColumn),
         );
     }
 
--- a/src/Illuminate/Database/Schema/Grammars/SqlServerGrammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/SqlServerGrammar.php
@@ -140,24 +140,6 @@ class SqlServerGrammar extends Grammar
     }
 
     /**
-     * Compile a unique index ignoring a null column command.
-     *
-     * @param  \Illuminate\Database\Schema\Blueprint  $blueprint
-     * @param  \Illuminate\Support\Fluent  $command
-     * @return string
-     */
-    public function compileUniqueIgnoreNull(Blueprint $blueprint, Fluent $command)
-    {
-        return sprintf('create unique index %s on %s%s (%s) where %s is null',
-            $this->wrap($command->index),
-            $this->wrapTable($blueprint),
-            $command->algorithm ? ' using '.$command->algorithm : '',
-            $this->columnize($command->columns),
-            $this->wrap($command->nullColumn),
-        );
-    }
-
-    /**
      * Compile a plain index key command.
      *
      * @param  \Illuminate\Database\Schema\Blueprint  $blueprint
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -561,15 +561,15 @@ class Blueprint
     }
 
     /**
-     * Specify a unique index for the table, ignoring
-     * soft deleted rows.
+     * Specify a unique index for the table. Ignore soft
+     * deleted models during the unique check.
      *
      * @param  string|array  $columns
      * @param  string|null  $name
      * @param  string|null  $algorithm
      * @return \Illuminate\Database\Schema\IndexDefinition
      */
-    public function uniqueNotTrashed($columns, $name = null, $algorithm = null)
+    public function uniqueIgnoreTrashed($columns, $deletedAtColumn = 'deleted_at', $name = null, $algorithm = null)
     {
         return $this->indexCommand('unique', $columns, $name, $algorithm);
     }
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -5,9 +5,11 @@ namespace Illuminate\Database\Schema;
 use BadMethodCallException;
 use Closure;
 use Illuminate\Database\Connection;
+use Illuminate\Database\Query\Builder;
 use Illuminate\Database\Query\Expression;
 use Illuminate\Database\Schema\Grammars\Grammar;
 use Illuminate\Database\SQLiteConnection;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Fluent;
 use Illuminate\Support\Traits\Macroable;
 
@@ -571,7 +573,7 @@ class Blueprint
      */
     public function uniqueIgnoreTrashed($columns, $deletedAtColumn = 'deleted_at', $name = null, $algorithm = null)
     {
-        return $this->indexCommand('unique', $columns, $name, $algorithm);
+        return $this->indexIgnoreNullCommand('unique', $columns, $deletedAtColumn, $name, $algorithm);
     }
 
     /**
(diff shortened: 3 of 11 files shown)


driesvints:
Please note that draft PR's aren't reviewed.

morrislaptop:
@driesvints yep, ok to leave here and I'll mark ready when I've finished the todo's (happy to close and re-open if better for you)

morrislaptop:
@driesvints I've got the core DB engines supported now - except there are variations between versions of MySQL / Maria for this PR. Can I ask what's the preferred option here (answer will not guarantee merge 🤓):

1.  Find the lowest common solution for all engines (+ less complexity, - not as performant)
2. Check engine / version numbers and use best solution (+ performant, - more complexity)

src/Illuminate/Database/Schema/Grammars/MySqlGrammar.php
```diff
@@ -229,6 +229,24 @@ public function compileUnique(Blueprint $blueprint, Fluent $command)
...
+     * @return string
+     */
+    public function compileUniqueIgnoreNull(Blueprint $blueprint, Fluent $command)
+    {
+        return sprintf('create unique index %s on %s%s (%s, %s)',
+            $this->wrap($command->index),
+            $this->wrapTable($blueprint),
+            $command->algorithm ? ' using '.$command->algorithm : '',
+            $this->columnize($command->columns),
+            "(CASE WHEN {$this->wrap($command->nullColumn)} IS NULL THEN 0 ELSE {$this->wrap($command->nullColumn)} END)",
```
morrislaptop, line 246: The `0` is any value for the column that will not be treated as unique when doing comparisons. 0 is good for ints / timestamps etc.. but might not be suitable for all scenarios

src/Illuminate/Database/Schema/Grammars/SqlServerGrammar.php
```diff
@@ -139,6 +139,25 @@ public function compileUnique(Blueprint $blueprint, Fluent $command)
...
+     * Compile a unique index ignoring a null column command.
+     *
+     * @param  \Illuminate\Database\Schema\Blueprint  $blueprint
+     * @param  \Illuminate\Support\Fluent  $command
+     * @return string
+     */
+    public function compileUniqueIgnoreNull(Blueprint $blueprint, Fluent $command)
+    {
+        $columns = $command->columns;
+        $columns[] = $command->nullColumn;
```
morrislaptop, line 152: Just adding the column to the index works as SQL Server treats NULL values as the same. This is slightly less performant as a partial index as the null column is included in the index taking up storage / memory.

Sophist-UK:
### Description
I think what is missing from the description of this PR is the definition of the unique index on a table with softDeletes defined.

AFAIK, `$table->unique()` does not automatically add the `deleted_at` field at the end of the unique index definition - please correct me if I have this wrong.

If this is the case, then we have two distinct use cases:

1. `$table->unique('username')`
2. `$table->unique(['username', 'deleted_at'])`

Then we have at least the following 2 use cases:
1. Insert into this table when there are one or more existing records with the same `username` either all soft-deleted or one non soft-deleted. In the example link provided this seems to work consistently. It is inserting into a child table that varies between RDBMS'.
2. What happens when we try to insert a record into a separate table with a foreign key pointing to `username` when `username` is or is not defined in the first table and when it is or is not soft-deleted. TBH this is where the whole thing gets a bit hairy - probably the foreign-key should be `[table2.username, null]->[table1.username, table1.deleted_at]` so you can only insert the child record if the primary record is not soft-deleted. Then you have to deal with what happens to this relationship when the child is inserted and the parent record is subsequently soft-deleted.

I think that the real answer is to use an autoincrement column in the parent table as the foreign key, and to get the value of that primary key for matching username and deleted_at==null. 

And once this is understood, then we can test against the various RDBMS' that Laravel supports.

What I would hope is that Laravel would hide the implementation differences between RDBMS' so that the developer can write code without worrying about such things and the code will work the same regardless.

### So what is the problem?
1. Should Laravel automatically add `deleted_at` to unique indexes / primary keys because that would be an equivalent uniqueness to real deletes?
2. Should Laravel allow child inserts for a parent that is soft deleted?
3. What special handling code is needed inside Laravel to create consistent functionality regardless of RDBMS?
4. What documentation is needed to warn users of potential pitfalls?

morrislaptop:
@taylorotwell it's currently tricky to have a unique index on a column for a model with soft deletes and you want to ignore the deleted columns from the unique index. 

Taking an excerpt from [this article](https://dev.to/mmollick/using-unique-columns-and-soft-deletes-in-laravel-470p) 

This can be a common scenario if you use soft-deletes on your users table, where you have a unique constraint on the email field. With soft deletes this means if a user deletes their account they'll be unable to recreate the account at a later date because the email still exists in the table. Unless we intend to write user-hostile software, we probably want the user to be able to re-register with the same email.

To start off let's cover some basic assumptions about our application. We have a base Laravel user table migration, with the soft deletes attribute added.

```php
    $table->id();
    $table->string('email')->unique();
    $table->softDeletes();
});
```

We also have the base Laravel User model, with the SoftDeletes attribute added.

```php
{
    use Notifiable, SoftDeletes;
    ...
```

With this setup we can create uses and delete users, but creating a new user with the same email results in the "Integrity constraint violation" error.

1. Create User

```php
User::create([
    'email' => 'sam@example.com',
]);
```

Users Table:

| id | email | deleted_at |
| -- | ----- | ----------- |
| 1 | sam@example.com |

3. Delete User
```php
User::find(1)->delete();
```

Users Table:

| id | email | deleted_at |
| -- | ----- | ----------- |
| 1 | sam@example.com | 2020-07-29 16:40:56 |

5. Recreating user produces an error

```php
User::create([
    'email' => 'sam@example.com',
]);
```

> **Warning**: Illuminate/Database/QueryException with message 'SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'sam@example.com' for key 'users.users_email_unique' (SQL: insert into `users` (`email`) values (sam@example.com))'

The most obvious solution to try is to make an index with `$table->unique(['email', 'deleted_at'])` however this doesn't people would expect due [to this](https://www.dbi-services.com/blog/nulls-in-composite-keys/) and would allow a table to look like this.

| id | email | deleted_at |
| -- | ----- | ----------- |
| 1 | sam@example.com |
| 2 | sam@example.com |

This PR creates a unique index that works as expected.

Sophist-UK:
@morrislaptop Craig

1. We need tests on the existing code that shows that this problem can actually occur as you just described. (I don't disbelieve you, but the purpose of tests is to show that there is a problem and then to show that your solution actually fixes it.) So what we probably need are 2 tests:
    1. A table **without** soft deletes and a unique index. Add a record, attempt to add the same record again and get an exception, delete the record and add the record again successfully, then attempt to add it a second time and get an exception.
    2. The same but **with** soft deletes.
2. For softDelete to work as advertised i.e. it is equivalent to a delete unless you explicitly request that soft-deleted records are included, IMO your code needs to be added to `unique` rather than creating a whole new method. However, @taylorotwell should probably comment on this as we should try to avoid backwards incompatibilities.
3. If we are going to have a separate method, how about calling the new method `uniqueSoftDelete` so it is clear that this should be used with soft-delete tables.

morrislaptop:
> We need tests on the existing code that shows that this problem can actually occur as you just described. (I don't disbelieve you, but the purpose of tests is to show that there is a problem and then to show that your solution actually fixes it.) So what we probably need are 2 tests:
> A table without soft deletes and a unique index. Add a record, attempt to add the same record again and get an exception, delete the record and add the record again successfully, then attempt to add it a second time and get an exception.
> The same but with soft deletes.

Cool happy to show this, this is just showing how the database index works? Would you be happy with SQL Fiddle links?

> For softDelete to work as advertised i.e. it is equivalent to a delete unless you explicitly request that soft-deleted records are included, IMO your code needs to be added to unique rather than creating a whole new method. However, @taylorotwell should probably comment on this as we should try to avoid backwards incompatibilities.

I initially explored `$table->string('email')->unique()->ignoreTrashed()`, but I couldn't quite crack the implementation. Happy to take direction here. 

> If we are going to have a separate method, how about calling the new method uniqueSoftDelete so it is clear that this should be used with soft-delete tables.

Naming taken from https://github.com/laravel/framework/pull/38124

Sophist-UK:
> Would you be happy with SQL Fiddle links?

No - these should be phpunit tests.

> Naming taken from #38124

Well, perhaps with hindsight `ignoreTrashed()` should have been called `ignoreSoftDeleted()`.

IMO consistency of conceptual naming is important - and this is a Schema method and Schema uses `softDelete`.

But as previously stated, IMO this should simply be part of the existing `unique` method.

morrislaptop:
Thanks for your working on this with me @Sophist-UK !

> No - these should be phpunit tests.

I've added those tests. 

> Well, perhaps with hindsight ignoreTrashed() should have been called ignoreSoftDeleted().
> IMO consistency of conceptual naming is important - and this is a Schema method and Schema uses softDelete.

I'd prefer to keep it `ignoreTrashed()` so it's consistent with the rest of the framework (there is `$model->trashed()` as well). Leaving this to @taylorotwell 

> But as previously stated, IMO this should simply be part of the existing unique method.

I wouldn't assume that soft deleted records in a unique index should be ignored by default - some applications may disallow "used" values even if those records have been deleted. Better to be explicit on the behaviour with a new method. Leaving this to @taylorotwell

dammy001:
@morrislaptop faced this issue recently while performing validation `'email' => ['email', Rule::unique('users', 'email')->withoutTrashed()]`. This validation passed through but throw database error due to the unique index. Well done.

Sophist-UK:
I think the idea of softDelete is that when you soft delete records, all further changes should behave exactly as you would expect if the record had been hard deleted - except if you explicitly ask that softDeleted records are included.

This is why IMO `deleted_at` (or your calculated column) should be added to unique indexes automatically for tables with softDelete defined.

P.S. On reflection, I think that your calculated column needs to be changed too, because you might have 3 records:

|Value|deleted_at|Calculated_field|
|Name|null|true|
|Name|2022-07-04|false|
|Name|2022-07-03|false|

Would your calculated field allow there to be two deleted versions in the unique index?

morrislaptop:
> I think the idea of softDelete is that when you soft delete records, all further changes should behave exactly as you would expect if the record had been hard deleted - except if you explicitly ask that softDeleted records are included.
> 
> This is why IMO `deleted_at` (or your calculated column) should be added to unique indexes automatically for tables with softDelete defined.

This isn't the current behaviour of the framework - see Unique and Exists validation rules. 

> 
> P.S. On reflection, I think that your calculated column needs to be changed too, because you might have 3 records:
> 
> ```
> |Value |deleted_at|calc |
> -------------------------
> |Name  |null      |true |
> |Name  |2022-07-04|false|
> |Name  |2022-07-03|false|
> ```
> 
> Would your calculated field allow there to be two deleted versions in the unique index?

Yes, there is no reason to prevent this from happening. The application can enforce stricter rules if needed.

Sophist-UK:
> This isn't the current behaviour of the framework - see Unique and Exists validation rules.

But IMO this is how it should behave. And now that you mention it, the unique and exists validations should also operate on softDeleted records as if they had been hard deleted i.e. deleted_at should be included automatically in detect or exists validations.

> > P.S. On reflection, I think that your calculated column needs to be changed too, because you might have 3 records:
> > ```
> > |Value |deleted_at|calc |
> > -------------------------
> > |Name  |null      |true |
> > |Name  |2022-07-04|false|
> > |Name  |2022-07-03|false|
> > ```
> > Would your calculated field allow there to be two deleted versions in the unique index?
> 
> Yes, there is no reason to prevent this from happening. The application can enforce stricter rules if needed.

So your calculated field would produce false whenever deleted_at was not null, and you would need to include both deleted_at and your calculated field in the index to check for uniqueness?

morrislaptop:
> But IMO this is how it should behave. And now that you mention it, the unique and exists validations should also operate on softDeleted records as if they had been hard deleted i.e. deleted_at should be included automatically in detect or exists validations.

This is a little outside the scope of this PR - perhaps [discussions](https://github.com/laravel/framework/discussions) would be a good place for this. 

> So your calculated field would produce false whenever deleted_at was not null, and you would need to include both deleted_at and your calculated field in the index to check for uniqueness?

the calculated field produces NULL whenever deleted_at was not null and since null is not tested in a unique index, it works as expected. I've expanded the test to cover this, thanks for flagging.

Sophist-UK:
> This is a little outside the scope of this PR - perhaps [discussions](https://github.com/laravel/framework/discussions) would be a good place for this.

Actually, this is very much in scope for this PR. This PR is attempting to fix issues where softDeleted records do **not** operate as intended because a softDeleted record should not impact any other operations. The whole point of Eloquent / Schema is that it hides these sorts of complexities (and differences in how different RDBMs behave) from the user and gives a consistent developer experience.

> the calculated field produces NULL whenever deleted_at was not null and since null is not tested in a unique index, it works as expected. I've expanded the test to cover this, thanks for flagging.

In other words, you are relying on an eccentricity of a database to fix and issue caused by the same eccentricity. IMO this is NOT a recipe for a reliable long-term solution which will work identically on every supported RDBM. An alternative fix might be to create a calculated field which copies the deleted_at timestamp but if it is null then we set it to the earliest allowed timestamp - and this would result in a non-null field that will work regardless of the idiosyncrasies of any specific RDBM.

morrislaptop:
@Sophist-UK this PR is not fixing any issues, both soft deleted and the unique index work as intended. This PR adds a feature allowing unique values in a table which ignores soft deleted records whilst hiding the complexities to make this possible. This is not eccentricity, it is the [SQL standard](https://www.dbi-services.com/blog/nulls-in-composite-keys/). 

I think this PR has some had healthy discussion which has contributed to the design and shape. At this point I will leave it to the maintainers to merge or close as is. Thank you for feedback.

Sophist-UK:
@morrislaptop Craig - Sorry, but it is my belief that softDelete is supposed to mimic a hard delete but be recoverable, and therefore my own view is that it is not currently working as intended.

The SQL standard AFAIK is NOT to ignore null fields when checking a unique index for uniqueness. The definition in the SQL standard is apparently a bit open to interpretation, and different RDBMs implement it differently. So it is indeed a quirk.

But even if it is not a quirk, my point about fixing it is still relevant - implementing a fix in a way that is completely unambiguous is IMO a good thing as it makes it less likely that some quirk will create problems with it at a later date.

(I should add that I do not see my comments as a criticism of your views or the work you have done - anyone who takes the time and effort to create a PR deserves recognition for their contribution. I just have a different perspective. But the purpose of these reviews is to try to ensure that changes are done right, so that they don't require later fixes and don't create backward compatibility issues, and on that basis I believe that these sorts of debates are exactly what should be happening in the interests of quality.)

I guess we need @taylorotwell to adjudicate on this one. I have made my case, and if Taylor disagrees with me and agrees with you, that is still a good result either way because the debate has helped ensure that the right decision is made.

morrislaptop:
Thanks for that @Sophist-UK, a rising tide lifts all ⛵ 🛥️

taylorotwell:
No plans to modify or add any behavior around this at the moment.

Sophist-UK:
@morrislaptop Craig

As per my comment in #41007, and as stated above, I believe that this behaviour is actually a bug.

I am wondering whether you might get more traction creating a PR for [10.x] along the lines of #41007?

The idea here is not to do an extra field, but instead to use a non-null value for `deleted_at` when the record is **not** soft-deleted. To avoid backwards compatibility if existing code checks for `deleted_at == null`, we might want the Model class to continue to deliver the value as null even though it is stored in the database differently. The Schema code to make the field `nullable()` would also need to be replaced with a `default('1970-01-01 00:00:01'). Field resets and record restores would need to be handled. 

A other considerations:

1. Migration from v9 to v10 would require some SQL statements to be run:
```
UPDATE table_name SET `deleted_at` = 1970-01-01 00:00:01 WHERE `deleted_at` IS NULL;
ALTER TABLE table_name MODIFY `deleted_at` DATA_TYPE NOT NULL;
```
    Obviously it gets more complicated if the `deleted_at` field is part of a primary or foreign key.

2. It might be sensible to include code to detect if the `deleted_at` field is NULL and raise an exception or warn the user or run the above statements automatically (yes a performance hit - but it would be a one off and it would make the thing self-repairing).

3. If this is not considered a bug, then it would be likely that such a PR would be rejected, and then the alternative is to create a small "Fix softDelete" package that would modify the behaviour as above.

morrislaptop:
Hi @Sophist-UK , I'm just doing something application / database specific now!

The example from https://github.com/tpetry/laravel-postgresql-enhanced#partial-indexes works perfectly for this use case. Hope this helps other people arriving at this thread.

morrislaptop:
Timely news for anyone interested https://blog.rustprooflabs.com/2022/07/postgres-15-unique-improvement-with-null

Sophist-UK:
@morrislaptop Craig - IMO the spec was wrong or it was wrongly interpreted. null should never have been "considered non-equal". I think most of us would expect `null == null` to be true, and we would expect a Unique index to ensure that items are actually unique, and so treating nulls as non-equal is anti-intuitive and TBH I struggle to envisage a real-life use case.

However IMO the key point is that one of the key selling points of frameworks like Laravel and ORMs like Eloquent is that they should hide the inconsistencies from users. So whilst we should welcome this Postgres initiative, until it becomes a de-facto standard across all SQL RDBMs it is IMO Eloquent's job to hide these inconsistencies from users and make Laravel behave consistently across DBs.

PR #42954, opened 2022-06-26
[9.x] Add `CREATE LIKE` grammar for MySQL and Postgres

lupinitylabs:
This PR adds support for the `CREATE LIKE` syntax for both MySQL and PostgreSQL.

The CREATE LIKE syntax in MySQL and PostgreSQL creates a copy of an existing table schema.
This PR allows to create a copy of the users table into users2 by executing:

```php
Schema::createLike(table: 'users2', fromTable: 'users')
```

Currently, raw statements have to be issued to achieve the same, e.g.
```php
DB::statement('CREATE TABLE `users2` LIKE `users`')
```

### Partial PostgreSQL Support

PostgreSQL supports options allowing for the inclusion or exclusion of copied features like comments, constraints etc. By default, none of the features will be copied to the new table. I decided to keep the implementation minimal for now, but it looks like it would be a good idea to include these options for PostgreSQL because I suppose it is desirable to keep some of the features on schema copies.

I'm open for opinions. @tpetry ?

### SQL Server Support

While it seems to be possible to achieve a similar effect with `SELECT top 0 * INTO users2 FROM users` in SQL Server, I decided not to implement this because it does not have native support and I could not test the implementation.

### SQLite Support

SQLite does not have native support for copying table schemas either. `CREATE TABLE users2 AS SELECT * FROM users WHERE 0` can be used to create a copy, but only basic type information is retained.

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -280,6 +280,16 @@ class Blueprint
     }
 
     /**
+     * Indicate that the table needs to be created from an existing table.
+     *
+     * @return \Illuminate\Support\Fluent
+     */
+    public function createLike($fromTable)
+    {
+        return $this->addCommand('createLike', compact('fromTable'));
+    }
+
+    /**
      * Indicate that the table needs to be temporary.
      *
      * @return void
--- a/src/Illuminate/Database/Schema/Builder.php
+++ b/src/Illuminate/Database/Schema/Builder.php
@@ -258,6 +258,21 @@ class Builder
     }
 
     /**
+     * Create a new table on the schema after an existing table.
+     *
+     * @param  string  $table
+     * @param  string  $fromTable
+     * @return void
+     */
+    public function createLike($table, $fromTable)
+    {
+        $this->build(tap(
+            $this->createBlueprint($table),
+            fn ($blueprint) => $blueprint->createLike($fromTable))
+        );
+    }
+
+    /**
      * Drop a table from the schema.
      *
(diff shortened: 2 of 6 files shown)


tpetry:
Supporting all the include/exclude options would be really nice. You can do a lot of cool stuff by selectively including/excluding some data. I would propose adding a generic `array $options = []` as last parameter to implement engine specific functionality. It could look like this for PostgreSQL:
```php
// The parameter order should be changes so it can be read like "copy table like {old} to {new}"
Schema::createLike('users', 'users2', [
  'excluding' => ['CONSTRAINTS', 'INDEXES'],
])
```

Do you see any possibility of a different syntax? The following would be awesome and feel much more naturally:
```php
Schema::create('users2', function(Blueprint $table) {
  $table->like('users', [
    'excluding' => ['CONSTRAINTS', 'INDEXES'],
  ]);

  $table->text('newcolumn');
  $table->index(['id', 'newcolumn']);
});
```

That's how I wanted to implement it when I find time for it.

lupinitylabs:
My intention was to stick as close to the existing generic create table syntax as possible.
If anyone from the core can decide whether such a deviation would be ok, I'm happy to change it accordingly.

From how I understand the PostgreSQL docs,

```php
Schema::createLike('users', 'users2', [
  'excluding' => ['CONSTRAINTS', 'INDEXES'],
])
```
would have to generate a `INCLUDING ALL EXCLUDING CONSTRAINTS EXCLUDING INDEXES` syntax (because everything is excluded by default)?

taylorotwell:

PR #18989, opened 2017-04-28
Add str_ordinal helper

shawnlindstrom:
Adds a helper function to append an ordinal indicator to a given number and optionally wrap the indicator in superscript tags.

    $value = str_ordinal('1');

    // 1st

    $value = str_ordinal('2', true);

    // 2<sup>nd</sup>

PR #42955, opened 2022-06-26
[9.x] Adds a new `ordinal` method to Str and Stringable classes

mateusjunges:
This PR adds a new `ordinal` method to the Str and Stringable classes.

I'm building an application where I need to display a certain user's position in a ranking and I found that there is no built-in method to do this. I could easily create a function in my own app to do this, but I think it would be useful to have this available in the Str/Stringable classes.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -642,9 +642,10 @@ class Str
      * Generate an ordinal suffix for the given value.
      *
      * @param  int  $value
+     * @param  bool  $superscript
      * @return string
      */
-    public static function ordinal($value)
+    public static function ordinal($value, $superscript = false)
     {
         if (! is_numeric($value)) {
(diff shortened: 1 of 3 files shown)


ankurk91:
This method should go to HtmlString class instead?

bert-w:
> This method should go to HtmlString class instead?

Because it can output HTML? I dont know if there is a class that does that specifically. I use the `HtmlString` class only as a simple wrapper for HTML data (so that `{{ my_html_func() }}` can still output correctly even though you use `{{ }}`).

Also this function might be a bit restrictive, since you can't send params to `number_format()` now and the `<sup>` tag cannot be replaced by another tag (imagine having a css rule for `<span>`). But overall this is fairly nice (I never considered the logic behind st/th/nd suffixes).

bonzai:
#18989

PR #42961, opened 2022-06-26
[9.x] Collection Helper data_get Not Using Default Value

tehomen:
Currently the function will only return default value for the parameter $default. This fix corrects the issue on * segments and will now return the value specified in $default.

### DIFF
--- a/src/Illuminate/Collections/helpers.php
+++ b/src/Illuminate/Collections/helpers.php
@@ -68,7 +68,7 @@ if (! function_exists('data_get')) {
                 $result = [];
 
                 foreach ($target as $item) {
-                    $result[] = data_get($item, $key);
+                    $result[] = data_get($item, $key, $default);
                 }
 
                 return in_array('*', $key) ? Arr::collapse($result) : $result;


selcukcukur:
@tehomen There is already a parsing before the process gets there, so if it needs to get the default value, it has already taken it, no need for it to be there. It is not included because it will return null under all conditions.

tehomen:
> @tehomen There is already a parsing before the process gets there, so if it needs to get the default value, it has already taken it, no need for it to be there. It is not included because it will return null under all conditions.

When I call the function like so:
```php 
data_get([ 'items' => [ [ 'name' => 'foo', 'id' => 12 ], [ 'id' => 13 ] ] ], 'items.*.name', 'bob');
```
I would get back an array with [ 'foo', null ] even though I specified to return 'bob' by default.

selcukcukur:
https://github.com/laravel/framework/blob/4d46bda6f19d490080512fa053065081f737bf2f/tests/Support/SupportHelpersTest.php#L102
https://github.com/laravel/framework/blob/4d46bda6f19d490080512fa053065081f737bf2f/tests/Support/SupportHelpersTest.php#L105
https://github.com/laravel/framework/blob/4d46bda6f19d490080512fa053065081f737bf2f/tests/Support/SupportHelpersTest.php#L117

If you change the `null` parts to `irrelevant`, the errors you get will be gone.

driesvints:
This seems to break the test suite. Please resubmit this with passing tests.

PR #42973, opened 2022-06-27
allow dot symbol (.) in database connection when using Unique/Exists …

thyseus:
Scenario: a database connection with a dot (.) symbol like: `mysql_tunnel_to_192.168.10.10`.

Assuming the following rule: `'email' => 'required|email|unique:mysql_tunnel_to_192.168.10.10.users'`

Before this fix, Unique/Exists Validators will extract:

connection: `mysql_tunnel_to_192.`
table: `168.10.10.users`

which is wrong. With this fix,

connection: `mysql_tunnel_to_192.168.10.10`
table: `users`

is being extracted which is correct.
If this is too hacky or you have an better idea, i suggest to throw an exception when a developer attempts to create a database connection with a dot in it. It can lead to unexpected behavior.

Thanks a lot !

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased](https://github.com/laravel/framework/compare/v9.18.0...9.x)
 
+### Fixed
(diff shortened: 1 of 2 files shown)


taylorotwell:
No plans to merge this atm. Feel free to use a custom rule if needed.

PR #42980, opened 2022-06-27
[9.x] Passing null to parameter #2 ($string) of type string is deprecated

buismaarten:


kayw-geek:
@taylorotwell  Excuse me, Could you tell me why these PRs would  be unmerciful close. #43510 #43507 #43508

kayw-geek:
> @taylorotwell Excuse me, Could you tell me why these PRs would be unmerciful close. #43510 #43507 #43508

I don't know if me offend him?

Issue #42952, opened 2022-06-25
Unexpected behaviour when registering multiple Model::creating() event callbacks


JackWH:
- Laravel Version: 9.18.0
- PHP Version: 8.1.6
- Database Driver & Version: MySQL 8

### Description:

Applications may register several Eloquent model callbacks for the same event. In my app, I'm automatically assigning certain values to models throughout their lifecycle. When registering multiple `creating` callbacks, they cause an unexpected early `INSERT` if any of the callbacks return a "truthy" value.

This might be a bug, or it might just be inconsistent behaviour, I'm not sure. Either way it doesn't seem to be documented anywhere, and [works the opposite way](https://laravel.com/docs/9.x/events#stopping-the-propagation-of-an-event) to how we'd normally stop event propagation. If I register several `static::creating()` handlers on the same model class, I'd expect them to run sequentially, and for them *all* to complete, before the model can be considered `created`.

I was caught out by this because of using PHP's short function syntax to define my callback, as shown in the example below. 

### Steps To Reproduce:

1. Create the database schema:

```php
<?php

use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class extends Migration {
    public function up(): void
    {
        Schema::create('my_models', function(Blueprint $table) {
            $table->increments('id');
            $table->string('optional_column')->nullable();
            $table->string('required_column');
            $table->timestamps();
        });
    }
}
```

2. Create the model:

```php
<?php

namespace App\Models;

use Illuminate\Database\Eloquent\Model;
use Illuminate\Support\Facades\Log;

class MyModel extends Model
{
    protected $fillable = [
        'optional_column',
        'required_column',
    ];

    protected static function booted(): void
    {
        // 1: Log a message with the Model's initial (empty) attributes when it's being created
        // (This behaves "normally", it logs and continues execution to the next callback as we'd expect.)
        static::creating(static fn(Model $model) => Log::info('Log 1', $model->getAttributes()));

        // 2A: Assign an optional value to the Model, if not already set, using arrow syntax.
        // (🧨 This causes an SQL INSERT error. Returning the model in the callback caused an immediate insert.)
        static::creating(static fn(Model $model) => $model->optional_column === null
            ? $model->optional_column = 'hello world'
            : $model
        );

        // 2B: Assign an optional value to the Model, if not already set, WITHOUT using arrow syntax.
[...]

PR #42981, opened 2022-06-27
[9.x] Proposal to change the  behavior of all "before" model events

mateusjunges:
As discussed in #42952, currently, all model events ending in `-ing`, are dispatched using `Dispatcher@until`, which will stop in the first non-null response from a listener, so we can't define any listeners that return any truthly value.

For example, in the code snipeed below, the framework will try to insert the model after executing the second `creating` callback, and this insertion will fail because the `required_column` is missing until the fourth callback executes.
In this example, I could easily perform all actions in a single `creating` callback, but in a real world project, there might be a third party package defining their own `creating` listeners, and if one of them returns a non-null value, then the following callbacks won't be executed.

```php
// Migration
return new class extends Migration {
    public function up(): void
    {
        Schema::create('my_models', function(Blueprint $table) {
            $table->increments('id');
            $table->string('optional_column')->nullable();
            $table->string('required_column');
            $table->timestamps();
        });
    }
}

//Model
class MyModel extends Model
{
    protected $fillable = [
        'optional_column',
        'required_column',
    ];

    protected static function booted(): void
    {
        // 1: Log a message with the Model's initial (empty) attributes when it's being created
        // (This behaves "normally", it logs and continues execution to the next callback as we'd expect.)
        static::creating(static fn(Model $model) => Log::info('Log 1', $model->getAttributes()));

        // 2A: Assign an optional value to the Model, if not already set, using arrow syntax.
        // (🧨 This causes an SQL INSERT error. Returning the model in the callback caused an immediate insert.)
        static::creating(static fn(Model $model) => $model->optional_column === null
            ? $model->optional_column = 'hello world'
            : $model
        );

        // 2B: Assign an optional value to the Model, if not already set, WITHOUT using arrow syntax.
        // (✅ This works as expected. The model attribute is updated and the next callback is handled.)
        static::creating(static function(Model $model) {
            $model->optional_column === null
                ? $model->optional_column = 'hello world'
                : $model;
        });
        
        // 4: Assign a required value to the Model.
        // After this callback executes, the model will be saved successfully.
        static::creating(static function(Model $model) {
            $model->required_column = 'lorem ipsum';
        });
    }
}
```

This PR changes all before events to be dispatched without using event halting, which allows to run all the defined callbacks before returning the response.

We can't just pass `false` as the second param for the `fireModelEvent` method where `before` events  are dispatched, because we still want to stop the execution if some of the listeners return false.

I can add some tests if Taylor and the laravel team consider to merge this PR. Also, I'd appreciate a review from @rodrigopedra and @JackWH here.

If this should go to 10.x please let me know.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php
@@ -202,7 +202,7 @@ trait HasEvents
             return null;
         }
 
-        $eventResponses = array_map(function($response) {
+        $eventResponses = array_map(function ($response) {
             return ! ($response === false);
         }, Arr::wrap($eventResponses));
 
--- a/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php
@@ -194,20 +194,10 @@ trait HasEvents
     {
         $eventResponses = $this->fireModelEvent($event, false);
 
-        if ($eventResponses === false) {
-            return false;
-        }
-
-        if ($eventResponses === null) {
-            return null;
-        }
-
-        $eventResponses = array_map(function ($response) {
-            return ! ($response === false);
-        }, Arr::wrap($eventResponses));
-
-        if (in_array(false, $eventResponses)) {
-            return false;
+        foreach (Arr::wrap($eventResponses) as $response) {
+            if ($response === false) {
+                return false;
+            }
         }
 
         return true;
--- a/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php
@@ -190,6 +190,29 @@ trait HasEvents
         );
     }
 
+    protected function fireModelBeforeEvent($event)
+    {
+        $eventResponses = $this->fireModelEvent($event, false);
+
+        if ($eventResponses === false) {
+            return false;
+        }
+
+        if ($eventResponses === null) {
+            return null;
+        }
+
+        $eventResponses = array_map(function($response) {
+            return ! ($response === false);
+        }, Arr::wrap($eventResponses));
+
+        if (in_array(false, $eventResponses)) {
+            return false;
+        }
+
+        return true;
+    }
+
     /**
      * Fire a custom model event for the given event.
      *
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -222,7 +222,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
         if (! isset(static::$booted[static::class])) {
             static::$booted[static::class] = true;
 
-            $this->fireModelEvent('booting', false);
+            $this->fireModelBeforeEvent('booting');
 
             static::booting();
             static::boot();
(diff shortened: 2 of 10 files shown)


mateusjunges:
We could also change the `Dispatcher@dispatch` method to always return the first `false` response from a listener, so if my first listener returns `null` and the second one returns `false`, instead of returning a array with `[null, false]`, it will return only false, and we wont try to insert the model into the database, but I'll let this decision to taylor.

With this approach, I think we can change the second parameter to `false` when dispatching "before" events, and the `Dispatch@dispatcher` method will look like this:

```php
public function dispatch($event, $payload = [], $halt = false)
{
    // When the given "event" is actually an object we will assume it is an event
    // object and use the class as the event name and this event itself as the
    // payload to the handler, which makes object based events quite simple.
    [$event, $payload] = $this->parseEventAndPayload(
        $event, $payload
    );

    if ($this->shouldBroadcast($payload)) {
        $this->broadcastEvent($payload[0]);
    }

    $responses = [];

    foreach ($this->getListeners($event) as $listener) {
        $response = $listener($event, $payload);

        // If a response is returned from the listener and event halting is enabled
        // we will just return this response, and not call the rest of the event
        // listeners. Otherwise we will add the response on the response list.
        if ($halt && ! is_null($response)) {
            return $response;
        }

        // If a boolean false is returned from a listener, we will stop propagating
        // the event to any further listeners down in the chain, else we keep on
        // looping through the listeners and firing every one in our sequence.
        if ($response === false) {
            return false;
        }

        $responses[] = $response;
    }

    return $halt ? null : $responses;
}
```

rodrigopedra:
ah ok, got it... tricky. Thanks @mateusjunges =)

src/Illuminate/Database/Eloquent/Concerns/HasEvents.php
```diff
@@ -190,6 +190,29 @@ protected function fireModelEvent($event, $halt = true)
...
+
+        if ($eventResponses === false) {
+            return false;
+        }
+
+        if ($eventResponses === null) {
+            return null;
+        }
+
+        $eventResponses = array_map(function ($response) {
```
rodrigopedra, line 205: Wouldn't this be simpler to understand? Some other parts of the framework use this approach:

~~~php
foreach (Arr::wrap($eventResponses) as $response) {
    if ($response === false) {
        return false;
    }
}

return true;
~~~

Some examples:

https://github.com/laravel/framework/blob/49368f4f50e63c543511bad983e7bcef038e9794/src/Illuminate/Auth/SessionGuard.php#L442-L451

https://github.com/laravel/framework/blob/49368f4f50e63c543511bad983e7bcef038e9794/src/Illuminate/Collections/Collection.php#L545-L556

rodrigopedra, line 205: Wouldn't this be simpler to understand? Some other parts of the framework use this approach:

~~~php
foreach (Arr::wrap($eventResponses) as $response) {
    if ($response === false) {
        return false;
    }
}

return true;
~~~

Some examples:

https://github.com/laravel/framework/blob/49368f4f50e63c543511bad983e7bcef038e9794/src/Illuminate/Auth/SessionGuard.php#L442-L451

https://github.com/laravel/framework/blob/49368f4f50e63c543511bad983e7bcef038e9794/src/Illuminate/Collections/Collection.php#L545-L556

mateusjunges: Sure! Just sent a new commit with your suggestion

taylorotwell:
We could never change this behavior on a patch release.

mateusjunges:
@taylorotwell shoul I send this to 10.x then?

driesvints:
@mateusjunges yes

PR #42986, opened 2022-06-28
[10.x] Proposal to change the behavior of all "before" model events

### LINKED ISSUE #42952: Unexpected behaviour when registering multiple Model::creating() event callbacks

mateusjunges:
This is a follow up on #42981 and closes #42952.

Basicallly, all model events ending in `-ing`, are dispatched using `Dispatcher@until`, which will stop in the first non-null response from a listener, so we can't define any listeners that return any truthly value.

This PR changes this behavior by changing all the "before" model events to be dispatched without using event halting, which allows to run all event listeners until we get a `false` response.

[This comment ](https://github.com/laravel/framework/pull/42981#issuecomment-1168002195) has another approach I've considered.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasEvents.php
@@ -191,6 +191,25 @@ trait HasEvents
     }
 
     /**
+     * Fire the given "before" event for the model.
+     *
+     * @param  string  $event
+     * @return bool
+     */
+    protected function fireModelBeforeEvent($event)
+    {
+        $eventResponses = $this->fireModelEvent($event, false);
+
+        foreach (Arr::wrap($eventResponses) as $response) {
+            if ($response === false) {
+                return false;
+            }
+        }
+
+        return true;
+    }
+
+    /**
      * Fire a custom model event for the given event.
      *
      * @param  string  $event
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -222,7 +222,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
         if (! isset(static::$booted[static::class])) {
             static::$booted[static::class] = true;
 
-            $this->fireModelEvent('booting', false);
+            $this->fireModelBeforeEvent('booting', false);
 
             static::booting();
             static::boot();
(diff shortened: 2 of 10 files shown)


taylorotwell:
Honestly after so many years of this behavior I don't really feel compelled to change it and potentially break existing applications off of a single report, etc.

PR #42987, opened 2022-06-28
[9.x] Fix BC break for `Log` feature tests

michael-rubel:
## About
PR #41925 introduced a breaking change for `Log` feature tests. This PR reverts `Handler` class behavior to call log levels on logger instead of always relying on `log` method. If the original log level doesn't exist in the Logger class as a method, we're directing it to the `log` method.

### Steps to reproduce

Use `Log` facade to access logger:
```php
Log::debug('Message');
```

Then we expect `debug` to be called in feature tests:
```php
Log::shouldReceive('debug')
    ->withArgs(fn ($message) => $message === 'Message')
    ->once();
```

But the test fails, because:
```shell
Received Mockery_0_Illuminate_Log_LogManager::log(), but no expectations were specified
```

### DIFF
--- a/src/Illuminate/Foundation/Exceptions/Handler.php
+++ b/src/Illuminate/Foundation/Exceptions/Handler.php
@@ -265,15 +265,15 @@ class Handler implements ExceptionHandlerContract
             $this->levels, fn ($level, $type) => $e instanceof $type, LogLevel::ERROR
         );
 
-        $logger->log(
-            $level,
-            $e->getMessage(),
-            array_merge(
-                $this->exceptionContext($e),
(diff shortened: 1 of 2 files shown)


ankurk91:
This breaking change was introduced in April, and we updated our tests accordingly.

But after this PR, we need to revert our tests. :(

This PR has created more work for us (again) , I think it is too late to revert this PR.

driesvints:
@ankurk91 why didn't you report the breaking change?

ankurk91:
@olsza 
I agree with you, i thought It was just me. Will take care next time.

PR #43004, opened 2022-06-29
[9.x] Collection `get()` allows to get nested data using dot notation

DarkGhostHunter:
## What?

Since my last PR #42938 was shot down (probably too big), this brings the ability to use `get()` con Collections to retrieve deeply nested data using "dot" notation.

```php
$collection = collect([
    'fred' => ['id' => 99, ...]
    ...
]);

// Before
data_get($collection, 'fred.id');

// Now
$collection->get('fred.id');
```

> `data_get()` also works with numeric indexes, like `0.id`.

## BC?

None. Users that expects array keys using dots are on a world of pain anyway.


DarkGhostHunter:
I think I'm gonna rework this using a dumber-down version for building JSON.

stephan-v:
@DarkGhostHunter Was there a specific breaking reason why this never got merged?

DarkGhostHunter:
> @DarkGhostHunter Was there a specific breaking reason why this never got merged?

`get()` is used to retrieve an item by their key. Seems that if a key has dots then is a BC. 

This could be refactored by `dataGet()` instead.

PR #43006, opened 2022-06-29
[9.x] Adds simpler JSON builder class

DarkGhostHunter:
## What?

Retry of #42938. Simpler version of the JSON builder object, which allows to build complex JSON trees easily across the application.

```php
use Illuminate\Support\Json;
use Illuminate\Support\Arr;

// Before
$json = ['this' => ['is' => 'cool']]);

Arr::set($json, 'this.is', 'cool');

// Now
$json = Json::make()->set('this.is', 'cool');
```

Features:
- Supports get, set and unset from nested arrays or objects.
- Serializes into JSON automatically.
- Supports dynamic access, array access.
- The `make()` method supports creating from an iterable or a JSON-encoded string.

There is potential to adopt this class in multiple JSON-responsable services, like translation, requests, responses and Model JSON column casting (sign me up for that!).

## Why?

Because Collection and Fluent are focused on single-dimensional lists. Playing with multidimensional arrays quickly becomes spaghettis code.

## BC?

Only adds a class and its tests.

### DIFF
--- a/src/Illuminate/Collections/Json.php
+++ b/src/Illuminate/Collections/Json.php
@@ -1,0 +1,255 @@
+<?php
+
+namespace Illuminate\Support;
+
+use ArrayAccess;
+use Illuminate\Contracts\Support\Arrayable;
+use Illuminate\Contracts\Support\Jsonable;
+use Illuminate\Support\Traits\Conditionable;
+use Illuminate\Support\Traits\Tappable;
+use JsonSerializable;
+use Stringable;
+
+class Json implements ArrayAccess, Arrayable, JsonSerializable, Jsonable, Stringable
+{
+    use Conditionable;
+    use Tappable;
+
+    /**
+     * The underlying JSON data.
+     *
+     * @var array
+     */
+    protected $data;
+
+    /**
+     * Create a new Json instance.
+     *
+     * @param  iterable  $data
+     */
+    public function __construct($data = [])
+    {
+        foreach ($data as $key => $value) {
+            $this->data[$key] = $value;
+        }
+    }
+
+    /**
+     * Returns a JSON value in "dot" notation.
+     *
+     * @param  array|string|int|null  $key
+     * @param  mixed|null  $default
+     * @return mixed
+     */
+    public function get($key, $default = null)
+    {
+        return data_get($this->data, $key, $default);
+    }
+
+    /**
+     * Sets a JSON value in "dot" notation.
+     *
+     * @param  array|string|int  $key
+     * @param  mixed  $value
+     * @param  bool  $overwrite
+     * @return $this
+     */
+    public function set($key, $value, $overwrite = true)
+    {
+        data_set($this->data, $key, $value, $overwrite);
+
+        return $this;
+    }
+
+    /**
+     * Determine a given key in "dot" notation exists in the JSON data.
+     *
+     * @param  string|int  $key
+     * @return bool
+     */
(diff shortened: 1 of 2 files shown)


src/Illuminate/Collections/Json.php
```diff
@@ -0,0 +1,255 @@
...
+    /**
+     * Create a new Json instance.
+     *
+     * @param  iterable  $data
+     */
+    public function __construct($data = [])
+    {
+        foreach ($data as $key => $value) {
+            $this->data[$key] = $value;
+        }
```
mateusjunges, line 34: Do you need to map through the entire array here? 
The only thing I can think is that if the array has a duplicated key, then you will keep the last one, but it already happens even if you just assign the value using `$this->data = $data`.

Check this example. Using foreach:
```php
class Json {
    public array $data;
  
    public function __construct($data = []) {
        foreach($data as $key => $value) {
            $this->data[$key] = $value;
        }
  }
}

$data = ['test' => 'a', 'b' => 'c', 'test' => 1, 'test' => 2]; 

new Json($data)

/*
This code outputs
Json {#2885
     +data: [
       "test" => 2,
       "b" => "c",
     ],
   }
*/
``` 

Using `$this->data = $data`:
```php
class Json {
    public array $data;
  
    public function __construct($data = []) {
        $this->data = $data;
   }
}

$data = ['test' => 'a', 'b' => 'c', 'test' => 1, 'test' => 2]; 

new Json($data)

/*
This code outputs

Json {#2885
     +data: [
       "test" => 2,
       "b" => "c",
     ],
   }
*/
```

DarkGhostHunter: It's the same approach the `Fluent` class. 

Since you can issue an `iterable` object (like something `IteratorAggregate`), this ensures all the items are rooted into an `array`. Otherwise, I could copy how the `Collection` does it, which does not map the entire array but rather checks which is:

```php
if (is_array($items)) {
    return $items;
} elseif ($items instanceof Enumerable) {
    return $items->all();
} elseif ($items instanceof Arrayable) {
    return $items->toArray();
} elseif ($items instanceof Jsonable) {
    return json_decode($items->toJson(), true);
} elseif ($items instanceof JsonSerializable) {
    return (array) $items->jsonSerialize();
} elseif ($items instanceof Traversable) {
    return iterator_to_array($items);
}

return (array) $items;
```

mateusjunges: Got it. Thanks!

DarkGhostHunter: I said "I could", not that I will add it to the PR because, sincerely, I think it adds more boilerplate.

taylorotwell:
It feels like this could be a generic PHP package.

mateusjunges:
Hey @DarkGhostHunter if you don't mind I'd like to extract this into a package. Actually I've some use cases for this in some personal projects.

DarkGhostHunter:
> Hey @DarkGhostHunter if you don't mind I'd like to extract this into a package. Actually I've some use cases for this in some personal projects.

I've already made this as a package, releasing next week. It's macroed too and comes with a model cast.

```php
request()->asJson('foo');

$casts = [
    'config' => Json::class,
    'secureThing' => JsonEncrypted::class,
];
```

DarkGhostHunter:
@mateusjunges [There you go](https://github.com/Laragear/Json)

PR #43010, opened 2022-06-30
Add Gate helper

sawirricardo:
I think it's nice to simply gate('some-ability') instead of Gate::allows(), especially when you're dealing with inspect().

Hopefully can reduce boilerplate code.

```
gate('can-do-this'); // true
gate('can-do-that', true); // Response::deny();
gate(); // Gate instance
```


driesvints:
Seems like tests are broken here.

sawirricardo:
Hi @driesvints, just wondering if there's any clue on how to setup the Gate's interface (which is the error stated in the test) ? Would love to know how this should play out in setting up the test

PR #43014, opened 2022-06-30
[9.x] Fix issue with AssertableJson 'where' function and floating point numbers

Tarpsvo:
We encountered an issue in one of our projects where we try to assert JSON with a float value that ends with .0.

During JSON (de)serialization, the double value of "1234.0" gets converted into an integer and comparing the same value in an `AssertableJson ... ->where('value', 1234.0)` fails as the `value` has become an integer.

I have added a third property to the AssertableJson's ->where() method called `$strict` which allows the user to choose between `assertEquals` and `assertSame` functions. This allows the user to skip the type check in cases such as the one described.

I have also added two tests with this PR that confirms the functionality works.

The first test `testAssertWhereFailsWithFloatInStrictMode` confirms that the original `->where()` syntax continues working exactly as it was and also provides a repro case for the issue I encountered.

The second test `testAssertWhereMatchesWithFloatInNonStrictMode` confirms the newly added functionality works and the floating point number passes the assertion non-strict mode.

### DIFF
--- a/tests/Testing/Fluent/AssertTest.php
+++ b/tests/Testing/Fluent/AssertTest.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Tests\Testing\Fluent;
 
 use Illuminate\Support\Collection;
+use Illuminate\Testing\AssertableJsonString;
 use Illuminate\Testing\Fluent\AssertableJson;
 use Illuminate\Tests\Testing\Stubs\ArrayableStubObject;
 use PHPUnit\Framework\AssertionFailedError;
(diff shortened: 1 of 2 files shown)


tests/Testing/Fluent/AssertTest.php
```diff
@@ -271,6 +272,38 @@ public function testAssertWhereMatchesValue()
...
 
+    public function testAssertWhereFailsWithFloatInStrictMode()
+    {
+        $doubleValue = (float) 1234.0;
+
+        $stringifiedData = json_encode([
+            'bar' => $doubleValue,
+        ]);
+
+        $assertableJsonString = new AssertableJsonString($stringifiedData);
```
Tarpsvo, line 283: If using AssertableJsonString in AssertTest is wrong, please let me know and I will refactor it.

Tarpsvo:
@taylorotwell Thanks for your time. Can you suggest a different solution that would fix the case described in the test case `testAssertWhereFailsWithFloatInStrictMode` in this PR? Or would you like to guide me to a more acceptable acceptable solution with which I could submit a new PR? Ie a function `->whereSame()`?

Currently we used a macro to add the functionality.

PR #43021, opened 2022-06-30
Add Gate helper

sawirricardo:
I think it's nice to simply gate('some-ability') instead of Gate::allows(), especially when you're dealing with inspect().

Hopefully can reduce boilerplate code.

gate('can-do-this'); // true
gate('can-do-that', true); // Response::deny();
gate(); // Gate instance

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -953,3 +953,27 @@ if (! function_exists('view')) {
         return $factory->make($view, $data, $mergeData);
     }
 }
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
Feel free to define this in your own project if you wish!

sawirricardo:
Hi @taylorotwell, Thanks for reviewing this. Just want to know if you have any reason why this is not cut out to be in the framework?

Just want to understand better when I can contribute again, hopefully I know what to contribute next.

Thank you!

PR #43023, opened 2022-06-30
[9.x] Fix missing return

wfeller:


GrahamCampbell:
Thanks for the PR. What exactly does this fix, please?

wfeller:
```php
use \Illuminate\Mail\Attachment;
use \Illuminate\Notifications\Messages\MailMessage;

$message = new MailMessage;

// this should return the message but returns null, which prevents from chaining other message methods
$message->attach(
    Attachment::fromStorageDisk('local', 'path.pdf')
);
```

timacdonald:
I've verified this is indeed a fix and added a test for good measure that fails without this change.

Thanks for the PR @wfeller.

timacdonald (approved):

PR #43024, opened 2022-06-30
[9.x] allow to chain jobs when condition is true

fgaroby:
This PR adds a new `when`-like method, which allows to chain an array of jobs when a condition is `true`.

For example:
```php
JobChainingTestFirstJob::dispatch()->chainWhen(false, [
            new JobChainingTestSecondJob // won't be chained because the condition is false
        ])
        ->chainWhen(true, [
            new JobChainingTestThirdJob // will be chained because the condition is true
        ]);
```

Of course, it is possible to chain a default array of jobs, if the condition is `false` and if the third parameter is not `null`.
```php
JobChainingTestFirstJob::dispatch()->chainWhen(false, [
            new JobChainingTestSecondJob // Won't be chained because the condition is false
        ], [
            new JobChainingTestThirdJob // Will be chained because the condition is false
        ]);
```

### DIFF
--- a/src/Illuminate/Foundation/Bus/PendingDispatch.php
+++ b/src/Illuminate/Foundation/Bus/PendingDispatch.php
@@ -147,7 +147,7 @@ class PendingDispatch
      */
     public function chainWhen($condition, $chain, $defaultChain = null)
     {
-        if($condition) {
+        if ($condition) {
             $this->job->chain($chain);
         } elseif ($defaultChain) {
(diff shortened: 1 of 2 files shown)


fgaroby:
If this PR is OK, I will add some other stuff like `chainUnless` (in another PR or in this one ? As you wish), and I will write the Laravel documentation.

taylorotwell:

PR #43025, opened 2022-06-30
Added type hinting to $options parameter in toJson

gsinti:


taylorotwell:
I don't think we can add type hints on a patch release without breaking changes?

PR #43026, opened 2022-06-30
[9.x] Added missing method to CursorPaginator interface

buismaarten:
Autocompletion in IDEs doesn't recognise the `withQueryString` method because `paginate`, `simplePaginate` and `cursorPaginate` returns the CursorPaginator interface that doesn't contain the method `withQueryString`.

The interface method can be safely added because the class that implements `CursorPaginator`, has a base class with the `withQueryString` implemented.


GrahamCampbell:
Indeed this is a major breaking change, and cannot be added to 9.x.

buismaarten:
Sounds reasonable, tomorrow I will propose the changes to the master branche.

PR #43029, opened 2022-07-01
[9.x] wasChanged and isDirty phpdoc fix

daniilskr:
Right now methods **isDirty()** and **wasChanged()** have the same phpdoc comment and since their functionality is quite similar it is very confusing sometimes and may lead to bugs.

This PR proposes a fix to PHPDoc comments of those methods to make the difference and their purpose clearer.

The comment for **wasChanged()** is taken from Laravel docs
> method determines if any attributes were changed when the model **was last saved** within the current request cycle

The comment for **isDirty()** is not something I am sure of. The Laravel documentation might be wrong since it says
> The `isDirty` method determines if any of the model's attributes have been changed since the model was **retrieved**

When actually it checks if attribute(s) were changed since last sync (which happens on save)

Better suggestions for phpdoc comments are welcome

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1834,7 +1834,7 @@ trait HasAttributes
     }
 
     /**
-     * Determine if the model or any of the given attribute(s) have been modified.
+     * Determine if the model or any of the given attribute(s) have been modified since last save.
      *
      * @param  array|string|null  $attributes
      * @return bool
@@ -1858,7 +1858,7 @@ trait HasAttributes
     }
 
     /**
-     * Determine if the model or any of the given attribute(s) have been modified.
+     * Determine if the model or any of the given attribute(s) were changed on save.
      *
      * @param  array|string|null  $attributes
      * @return bool


IgorLockyh:
I forgot to create a new branch for commit.
Though this is more of a proposal, I will re-submit this if this PHPDoc is ok

src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
```diff
@@ -1858,7 +1858,7 @@ public function isClean($attributes = null)
     }
 
     /**
-     * Determine if the model or any of the given attribute(s) have been modified.
+     * Determine if the model or any of the given attribute(s) were changed on save.
```
taylorotwell, line 1861: Little unclear what this means. Do you mean these attributes were actually changed in the database on the last save?

IgorLockyh: Thanks for reviewing, Taylor! I guess this is what the method does. My English is far from perfect so I will be happy if someone will do a better job. My goal is just to have PHPDoc comments which could minimize confusion between the methods

taylorotwell:
I'll make a note to look into this this week.

taylorotwell:
Fixed.

PR #43035, opened 2022-07-01
feat: add singularity and plurality check methods for strings 🔬🧐

KhomerikiK:
This Pull Request adds new methods to the `Support Str` class to easily check strings plurality and singularity.

Added methods:
- `Str::isSingular('cat')//true`
- `Str::isPlural('cats')//true`

## Check if the given string is in a singular form:

> Before
```php
Str::singular($value) === $value;
```

> After
```php
Str::isSingular($value);
```

## Check if the given string is in a plural form:

> Before
```php
Str::singular($value) !== $value || $value === Str::plural($value);
```

> After
```php
Str::isPlural($value);
```

## Note

In some cases, if `isSingular` method returns `true`, it doesn't mean that `isPlural` method will return `false` for the same word
```php
Str::isSingular('sheep');//true
Str::isPlural('sheep');//true
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -921,6 +921,26 @@ class Str
     }
 
     /**
+     * check if the given value is in a singular form.
+     *
+     * @param  string  $value
+     * @return bool
+     */
+    public static function isSingular(string $value){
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #43042, opened 2022-07-02
[9.x] Add pseudo namespace for tables for `make:model` command 

Sophist-UK:
Laravel makes extensive use of php namespaces to allow packages to co-exist, however the same is not true of SQL tables, with the consequence that the package that creates tables cannot easily be identified and increasing the chances of clashes on table names between packages.

This pull request adds pseudo namespace support for tables to the `make:model` command by pre-pending the path specified on the command line to the table name in the model and migration files. So currently...
* `php artisan make:model Test` uses `test` as the table name
* `php artisan make:model Project\Test` also uses `test` as the table name

This pull request prepends the path to the table name in the migration and model files as follows:
* `php artisan make:model Test` still uses `test` as the table name
* `php artisan make:model Project\SubProj\Test` now uses `project_subproj_test` as the table name

Obviously the `make:model` command only creates stub files, and the user then has the option to change the table names, so this PR should not introduce any real backward compatibility issues, but hopefully this change will encourage package providers to group their tables.

Note: This is my first PR on Laravel so please excuse any mistakes in the coding or submission.

### DIFF
--- a/src/Illuminate/Console/GeneratorCommand.php
+++ b/src/Illuminate/Console/GeneratorCommand.php
@@ -289,7 +289,9 @@ abstract class GeneratorCommand extends Command
     {
         $stub = $this->files->get($this->getStub());
 
-        return $this->replaceNamespace($stub, $name)->replaceClass($stub, $name);
+        return $this->replaceNamespace($stub, $name)
+                     ->replaceTable($stub, $name)
+                     ->replaceClass($stub, $name);
     }
 
     /**
@@ -327,6 +329,22 @@ abstract class GeneratorCommand extends Command
     protected function getNamespace($name)
     {
         return trim(implode('\\', array_slice(explode('\\', $name), 0, -1)), '\\');
+    }
+
+    /**
+     * Replace the table name for the given stub.
+     *
+     * @param  string  $stub
+     * @param  string  $name
+     * @return $this
+     */
+    protected function replaceTable(&$stub, $name)
+    {
+        $table = Str::snake(Str::pluralStudly(Str::studly(Str::replace('\\', '_', $this->getNameInput()))));
+
+        $stub = str_replace(['DummyTable', '{{ table }}', '{{table}}'], $table, $stub);
+
+        return $this;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


taylorotwell:

Sophist-UK:
@taylorotwell Taylor

This is the second PR which has been closed this way. This one is not a bug either but IMO this fixes a poor design whereby component developers are **NOT** encouraged to namespace their tables.

Whilst I understand the point you are making, can you point me to some documentation which explains this policy?

Does the Laravel project accept any enhancement PRs? If so, what are the criteria, how do I ascertain whether an idea will be accepted if I develop a PR for it before I waste the effort?

dennisprudlo:
This is a template message for closing PRs. In this case the first paragraph might be the most important one. There's not really a checklist to see what to do so a PR gets merged. You can always start a discussion first about a feature you plan to implement and see what other developers/contributors think about what you plan to PR. In the end Taylor decides what gets merged into his project, but if a PR receives positive feedback from the community it's more likely to get merged.

Me personally, I don't like this part where the `$table` property is always added when creating a model. The laravel way is to let the framework assume what the table name might be. And only if necessary allow users to add this property to define the table name.

BrandonSurowiec:
Laravel has the ability to export the stubs so you can customize them for your own projects: https://laravel.com/docs/8.x/artisan#stub-customization

Granted, in this case you wouldn't be able to fill in the `$table` property, but you could update your stub to append it. 

However, like @dennisprudlo mentioned, Laravel has conventions to make things easier / require less customization. The more you can follow these conventions, the easier your life as a developer will be: https://laravel.com/docs/8.x/eloquent#table-names

So in short, this PR strays from the Laravel conventions, *and you can customize the stubs for your project) which is why it was closed. The "canned" response isn't always clear, I agree. It would be nice to have it mention a specific reason, but it saves Taylor a bunch of time. It is what it is.

Sophist-UK:
> Me personally, I don't like this part where the `$table` property is always added when creating a model. The laravel way is to let the framework assume what the table name might be. And only if necessary allow users to add this property to define the table name.

Yes - I agree - but stubs don't have conditional processing so you can't omit the `$table_name` block if the model is a root object.

PR #43046, opened 2022-07-03
Add totalCount to cursor pagination

leonardocustodio:
This adds total count to cursor pagination: https://github.com/laravel/framework/discussions/43045

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,25 @@
 # Release Notes for 9.x
 
-## [Unreleased](https://github.com/laravel/framework/compare/v9.18.0...9.x)
+## [Unreleased](https://github.com/laravel/framework/compare/v9.19.0...9.x)
+
+
+## [v9.19.0](https://github.com/laravel/framework/compare/v9.18.0...v9.19.0) - 2022-06-28
+
+### Added
+- Add new allowMaxRedirects method to PendingRequest ([#42902](https://github.com/laravel/framework/pull/42902))
+- Add support to detect dirty encrypted model attributes ([#42888](https://github.com/laravel/framework/pull/42888))
+- Added Vite ([#42785](https://github.com/laravel/framework/pull/42785))
+
+### Fixed
+- Fixed bug on forceCreate on a MorphMay relationship not including morph type ([#42929](https://github.com/laravel/framework/pull/42929))
+- Fix default parameter bug in routes ([#42942](https://github.com/laravel/framework/pull/42942))
+- Handle cursor paginator when no items are found ([#42963](https://github.com/laravel/framework/pull/42963))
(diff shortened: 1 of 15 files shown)


bonzai:
This will run an additional query even for people who don't need it (`count()` can be slow for big tables, so cursor pagination will no longer have benefits).

Also, I think you should send it to master branch since it's a breaking change (some method signatures were changed).

leonardocustodio:
Ok, maybe we could make this an optional feature then. I will move it to draft and wait for @taylorotwell feedback on this.

leonardocustodio:
Ah no, no need to check the PR, wait the discussion on https://github.com/laravel/framework/discussions/43045 then if it is something that we should consider I can finish up this ;D
Thanks a lot for the suggestions :)

GrahamCampbell:
Adding total count to cursor pagination completely defeats the point of cursor pagination in that you are allowing the DB to only look quickly at a local region of the data.

leonardocustodio:
I agree with you but that's the GraphQL spec for cursor-based pagination it has a formal spec on this and also a "best-practice" section for cursor pagination. And they use the totalCount as mandatory field. I think since graphQL is growing a lot in popularity and a lot of APIs are being done with GraphQL. I think we should at least consider it. Maybe as you said not as a default behavior.

Resources:
https://graphql.org/learn/pagination/
https://relay.dev/graphql/connections.htm

GrahamCampbell:
Where in the spec does it say that it is mandatory? All I can see is an example where it it is used.

GrahamCampbell:
**Not** including the total count is what I'd consider best practice for cursor paginated APIs that need to scale.

leonardocustodio:
I was almost sure I saw that totalCount and pageInfo were non-null. But now that you said I took a look again, and you are right 😆 sorry the confusion, closing this. Thanks for the guidance.

cwilby:
Just putting a note in here that cursor pagination is not always used "at scale" - and there could be a benefit to having the count be _available_, if not the default.  We're using cursor pagination instead of offset pagination to implement a "realtime grid" with web sockets.

Example - when a new item is created, each client receives the new item, and decides whether that new item falls within the currently loaded set.  If it does, it is added to the grid.  If it does not, it will be correctly loaded on subsequent pages.

Also, when an item is removed from the grid, each client receives a unique key, and if that new item is visible in the grid it is removed, if it is not, then subsequent paginations will work correctly.

Updates are implemented in a similar fashion.

We also show the total count of items in the grid, for which we do have to perform an extra count query.

If we were to use offset-based pagination, then subsequent pagination events would load incorrect data.  Thankfully we are able to use cursor-based pagination to solve those issues, with the tradeoff of not being able to jump to a specific page.

I'm not campaigning for this feature to be added, just highlighting a potential use case that might have been glanced over.

PR #43047, opened 2022-07-04
[9.x] Add `BadRequest` assertion to `TestResponse`

michael-rubel:
### About
We have conventional methods for `40x` HTTP responses, like `assertNotFound`, `assertForbidden`, `assertUnauthorized`. But the main `400` (Bad Request) is missing. This PR adds `assertBadRequest` method to the pool of available assertions.

```php
$response->assertBadRequest();
```


taylorotwell:
I recommend just asserting against the actual status code.

PR #43048, opened 2022-07-04
[9.x] Ability to attach an array of files in MailMessage

Illusionist3886:
From Notifications ```toMail()``` method:

```
$attachments = [
            public_path('file1.xlsx'),
            public_path('file2.mp4')
        ];
```
```
return (new MailMessage)
            ->view('mails.mail-template')
            ->attachFiles($attachments);
```


dennisprudlo:
I think this should be implemented in the `attach` method so the `$files` parameter can also be an array. With this I will still need to chain the `attach`-method if I have multiple `Attachable` objects I want to add.

taylorotwell:
See above.

PR #43061, opened 2022-07-04
Fix pushing notification to notifications when notifiable channels is…

faisalabuzaid:
… empty


taylorotwell:
Does not detail the bug or what happens.

PR #43065, opened 2022-07-05
[9.x] Introducing a fresh new look for Artisan

nunomaduro:
Hi, community. 👋🏻

For the past two weeks, I've been working on this pull request that introduces a **fresh new look for Artisan**. Before looking at the technical details, let's see some before/after comparisons:

---

## Technical details:

A common change you will see in this pull request is all the previous command calls to `$this->info`, `$this->warn`, etc, have been replaced by `$this->components->info`, etc. This was made intentionally, to ensure only framework internal commands are affected by this new output.

Now, besides the regular `$this->info`, `$this->warn`, etc methods, we also introduce - in this pull request - new components that are used across the pull request. This list of components is: `Alert`, `BulletList`, `Choice`, `Component`, `Confirm`, `Error`, `Factory`, `Info`, `Line`, `Task`, `TwoColumnDetail`, `Warn`.

As an example, the `$this->task` is used on the "queue:work" command to display the queue job description, the result of the queue job, and its duration.

Screenshot 2022-07-11 at 19 08 39

Now, keep in mind, that the `$this->components` property is marked as **internal**. Meaning that this pull request was not meant to introduce a public API for those components, but instead, simply to improve the output of Artisan itself.

Also, these new components share something in common: they mutate the given information. By "mutate", I mean: we ensure punctuation, relative paths instead of absolute paths, and more. This is necessary to ensure things *always* look pretty.

This pull request effects more than 100 commands. Yet, it does not affect the following commands: `completion`, `help`, `List`, `serve`, `tinker`, and commands of Laravel packages such as Horizon or Telescope. In addition, it does not affect errors command from Symfony, such as invalid arguments, etc. All of this, can be done later, so we deliver this pull request first.

Finally, would be very appreciated if the community could test this pull request in their L9 projects, as we plan to ship these changes in a minor release. To test this pull request locally, please follow this instructions: 

```diff
-        "laravel/framework": "^9.19.0",
+        "laravel/framework": "dev-feat/console-ui-improvements as 9.19.0",
```

```bash
composer update laravel/framework -W
```


src/Illuminate/Console/Contracts/NewLineAware.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+namespace Illuminate\Console\Contracts;
+
+interface NewLineAware
```
bastien-phi, line 5: Is this interface meant to be empty ?

Lloydinator:
Much cleaner. I like how it looks, and I like how you're separating feedback into Info and Error.

denniseilander:
First of all, I really like the new interface! Great job on this one.

The only thing I'm not sure about is the choice component.
It feels like it's harder to read compared to the old version (left) when running `php artisan vendor:publish`

The "Provider" and "Tag" labels are not clearly distinguishable, maybe because they don't have a highlight anymore.

Screenshot 2022-07-11 at 23 19 55

The same goes for some other commands with the choice output:
On the left the old interface, on the right the new interface.

Screenshot 2022-07-11 at 23 25 46

What do you think?

nunomaduro:
@denniseilander Done.

src/Illuminate/Console/QuestionHelper.php
```diff
@@ -0,0 +1,77 @@
...
+        $text = OutputFormatter::escapeTrailingBackslash($question->getQuestion());
+
+        $text = $this->ensureEndsWithPunctuation($text);
+
+        $text = "  <fg=default;options=bold>$text</></>";
+
+        $default = $question->getDefault();
+
+        if ($question->isMultiline()) {
+            $text .= sprintf(' (press %s to continue)', 'Windows' == PHP_OS_FAMILY
```
ozanakman, line 29:
```suggestion
            $text .= sprintf(' (press %s to continue)', 'Windows' === PHP_OS_FAMILY
```

src/Illuminate/Foundation/Console/EnvironmentCommand.php
```diff
@@ -40,6 +40,9 @@ class EnvironmentCommand extends Command
      */
     public function handle()
     {
-        $this->line('<info>Current application environment:</info> <comment>'.$this->laravel['env'].'</comment>');
+        $this->components->info(sprintf(
+            'The application is in the [%s] environment.',
```
PHLAK, line 44: The application isn't necessarily _in_ a specific environment, rather the environment is set to a specific value. For example, I could set `APP_ENV` to `production` on my local system.

```suggestion
            'The application environment is [%s].',
```

PHLAK:
In the `vendor:publish` command screenshot pre-existing files have trailing ellipses with no following word (i.e. `DONE`). Instead they should show `SKIPPED` at the end of the ellipses to indicate what action was taken (i.e. they were skipped). For example.

```
Copying directory [...] to [...] .................... DONE
File [...] already exists ........................ SKIPPED
```

src/Illuminate/Foundation/Console/EnvironmentCommand.php
tomasbreffitt: I also think putting the value in the middle of a sentence is harder to mentally parse, I like this change since it returns it to the end of the line

hmazter:
Great work!
I don't know if it's an intentional change, but the durations for each migration file are very sporadic with this change in the `migrate` command.

**The old output:**
Duration timings on each migration
Screenshot 2022-07-12 at 08 13 52

**The new output:**
Duration only appears on 1 of these 5
Screenshot 2022-07-12 at 08 14 59

src/Illuminate/Foundation/Console/EnvironmentCommand.php
nunomaduro: [Done.](https://github.com/laravel/framework/pull/43065/commits/62b2dce73285595105a80879e45dcb559ca3587b)

nunomaduro:
@PHLAK done.

nunomaduro:
@hmazter Done. The duration is always displayed now.

Krisell:
Awesome work!

The `Seeding:` and `Seeded:` output looks a bit misalign, but no big deal and perhaps they are intentionally not changed:
image

shaedrich:
I don't know, how other people prefer that, but what about removing the underscores from the migration file names just like PHPUnit does with the test names, so it looks a bit nicer? Perhaps, when doing that, the date could be formatted to be ISO 8601 or the like.

shaedrich:
Since you go for relative paths, you could go even further by doing it [zsh-style](https://ohmyz.sh/) and extract the "base path" for your command and display it separately to keep the rest of the output dry (e.g. `App\Jobs` for `php artisan queue:work`)

shaedrich:
grafik
grafik
While when running `php artisan make:job` already existing jobs result in an error, publishing existing files, doesn't even result in an info. Isn't that a bit inconsistent?

Besides that, I liked the commands where variable stuff inside `[` and `]` was highlighted. Was that intentionally omitted?

nunomaduro:
@shaedrich All your feedback can be addressed later.

Wirone:
I've just installed it and on full-screen iTerm2 on M1 it looks weird:

image

IMHO that full-width stretch is unnecessary, it's almost impossible to see everything at once and it's hard to match left side's data with right side's. Also it does not work well with cmd+D split:

image

Regular, left-aligned output would fit even after resizing.

adrianobnu (commented):
Excellent implementation @nunomaduro! Really, on a day-to-day basis working with Laravel will make the experience more pleasant.

nunomaduro:
@Wirone This is fixed. We now ensure a max of 150 chars per line:

Screenshot 2022-07-13 at 12 06 29

huangdijia:
Can add phpdoc for Factory ?

Saifallak:
it's awesome, but for `queue:work` 



i would prefer to see when the job is pushed to process not just after process completed

shaedrich:
@Saifallak Maybe there should be a modified progress bar 🤔

nunomaduro:
@Saifallak @shaedrich Not sure where you got that idea.

But, when a job starts goes into "processing" you see the time and the job name. The "DONE" on the right only appears when is actually done.

Saifallak:
> @Saifallak @shaedrich Not sure where you got that idea.
> 
> But, when a job starts goes into "processing" you see the time and the job name. The "DONE" on the right only appears when is actually done.

I got that idea from screenshots,.
You made examples for everything.
Glad to know that,..
Will try it out

L3o-pold:
This removed the id of a job when processing 😞  https://github.com/laravel/framework/commit/a105ae5d14a86c75706cf2e5fa802a215ae60666#diff-68b410043fae33c9cecda76c3cf7c54a9073dd08eca46d44cb8d4424fae8dfd8L210

Eldair:
@nunomaduro was this tested on windows? This is MS terminal with WSL

image

streamingsystems:
Hi All,

I opened an issue a few minutes ago: 

https://github.com/laravel/framework/issues/43325

I am trying to hunt down why my migrations don't work like they have since I upgraded to 9.21.0. I am wondering if at a glance what I am seeing might ring a bell. When I go back to 9.20.0 it works great, when I upgrade to 9.21.0 it's broken. Maybe the syntax or something changed, but that would seem odd. Anyhow, just trying to hunt it down.

jensolafkoch:
> @Wirone This is fixed. We now ensure a max of 150 chars per line:
> 
> Screenshot 2022-07-13 at 12 06 29

jensolafkoch:
> @Wirone This is fixed. We now ensure a max of 150 chars per line:
> 
> Screenshot 2022-07-13 at 12 06 29

@nunomaduro 

Wouldn't it be nice to have a config option for the character count (width)? Just in case that the default seems odd in some device/environment? 

And I totally love the new design! :-)

roberttolton:
Looks great @nunomaduro but I think something might be wrong with the terminal / window width detection, or something?

Running the latest Laravel 9.26.

Screenshot 2022-08-23 at 17 32 01

PR #43072, opened 2022-07-06
Add test for functional values (wrapped in parens) in wrap()

cheesegrits:
As of MySQL 8.0.13 (or thereabouts), indexes can be specified with [functional key parts](https://dev.mysql.com/doc/refman/8.0/en/create-index.html), a very powerful feature which allows us to to customize performance and functionality of indexes.  But in Laravel migrations, this would require something like this ...

``            $table->unique(['name', '(COALESCE(`scan_code`, ""))', '(COALESCE(`global_scan_type_id`, 0))'], 'global_manufacturer_aliases_unique_index');``

This currently won't work, because the Grammar wrap() method simply wraps the entire parenthesized expression in backticks.  My inclination was to modify wrapValue() in MySqlGrammar and test to see if the value already contained a backtick, and if so leave it untouched (on the "let's assume they know what they are doing" basis).  But for reasons not clear to me, that function already converts single backticks to double backticks.  Hence a modification to wrap() in the main Garmmar.

Thank you for your consideration.

### DIFF
--- a/src/Illuminate/Database/Grammar.php
+++ b/src/Illuminate/Database/Grammar.php
@@ -70,6 +70,14 @@ abstract class Grammar
             return $this->wrapJsonSelector($value);
         }
 
+        // If the given value is functional (wrapped in parentheses), we cannot make any
+        // assumptions, and should just return the value as is.  For example, as of MySQL
+        // 8.0.13, indexes can include functional key parts wrapped in parens, like
+        // (COALESCE(`scan_code`, "")), as well as simple column names.
+        if ($this->isFunctional($value)) {
+            return $value;
+        }
+
         return $this->wrapSegments(explode('.', $value));
     }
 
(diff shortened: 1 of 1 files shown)


cheesegrits:
I noticed it doesn't like my shortcut for testing the string, I can convert that to use strpos if the PR would be accepted.

BrandonSurowiec:
Would this work "as-is" without the PR?

```php

// ...

$table->unique([
    'name', 
    DB::raw('COALESCE(`scan_code`, "")', 
    DB::raw('COALESCE(`global_scan_type_id`, 0)',
], 'global_manufacturer_aliases_unique_index');
```

taylorotwell:

cheesegrits:
> ```
> DB::raw('COALESCE(`global_scan_type_id`, 0)'),
> ```

Yes, that does work, thank you!

I was working round it by just doing the whole index creation with a DB:statement(), but your solution is better.

Many thanks.

   -- hugh

PR #43076, opened 2022-07-06
[9.x] Add serialized cast

iamgergo:
This PR adds the `serialized` model attribute cast.

Recently, I found myself serializing/unserializing model attributes with accessors and mutators.
However I think, it would be better to have a native support for serialized value casting.

This is only additive, should not contain any BC.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -103,6 +103,7 @@ trait HasAttributes
         'json',
         'object',
         'real',
+        'serialized',
         'string',
         'timestamp',
     ];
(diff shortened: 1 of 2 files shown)


dennisprudlo:
@bert-w If it's good practice or not is another question but just for the sake of the question the `json` cast would not be handling classes correctly.

```php
$user = User::first();
$data = [
    'some' => 'data',
    'user' => $user
];

// will return the $data array with a full eloquent user model.
unserialize(serialize($user));

// will return the $data array with a stdClass having all columns as properties.
json_decode(json_encode($user));
```

iamgergo:
Well, most of the cases I use JSON as well. Simple and portable. 

But in some scenarios it's possible the developer want to store and keep the native PHP types of the data structure.

Also serialization has some other advantages like using `__wakeup()`, but I think for a model attribute it would be an overkill.

Overall, I find it's a nice to have.

taylorotwell:

PR #41861, opened 2022-04-07
[9.x] Allow Schema::enum to accept an Enum class

geidelguerra:
This allows a cleaner syntax to define enum column in migrations.

Example
```php
$table->enum('status', Status::class);
```

PR #43077, opened 2022-07-06
[feature] automatically detect allowed values

Com-Digital:
This mr makes it possible to pass a enum class name as 2nd argument.
Depending on enum type possible values will be added automatically.
The exception protects the user from unexpected errors.


dennisprudlo:
This kind of PR has been rejected in the past (#41861, https://github.com/laravel/framework/pull/40662 https://github.com/laravel/framework/pull/40913 https://github.com/laravel/framework/pull/41026)

When using an enum in your migration and then add or remove a case to/from the enum in the future, your migrations will be out of sync.

PR #43080, opened 2022-07-06
[9.x] Ability to attach an array of files in MailMessage

Illusionist3886:
Passing a filepath or an array of filepaths will now attach one or multiple files in mail.

```
$files = ['file.jpeg', 'file2.jpeg'];
```
Then in ***build*** method
```
return $this->view('welcome')
        ->attach($files)
        ->with('order', $this->order);
```

### DIFF
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -874,13 +874,11 @@ class Mailable implements MailableContract, Renderable
      */
     public function attach($files, array $options = [])
     {
-        if(!is_iterable($files))
-        {
+        if(!is_iterable($files)) {
             $files = explode(',', $files);
         };
         
-        foreach ($files as $file)
-        {
+        foreach ($files as $file) {
             if ($file instanceof Attachable) {
                 $file = $file->toMailAttachment();
             }
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -872,7 +872,7 @@ class Mailable implements MailableContract, Renderable
      * @param  array  $options
      * @return $this
      */
-    public function attach($files, array $options = [])
+    public function attach($file, array $options = [])
     {
         if ($file instanceof Attachable) {
             $file = $file->toMailAttachment();
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -877,19 +877,19 @@ class Mailable implements MailableContract, Renderable
         if ($file instanceof Attachable) {
             $file = $file->toMailAttachment();
         }
-    
+
         if ($file instanceof Attachment) {
             return $file->attachTo($this);
         }
-    
+
         $this->attachments = collect($this->attachments)
-            ->push(compact('file', 'options'))
-            ->unique('file')
-            ->all();
+                    ->push(compact('file', 'options'))
+                    ->unique('file')
+                    ->all();
 
         return $this;
     }
-    
+
     /**
      * Attach multiple files to the message.
      *
@@ -908,7 +908,7 @@ class Mailable implements MailableContract, Renderable
         } else {
             $this->attach($files, $options = []);
         }
-        
+
         return $this;
     }
 
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -874,25 +874,41 @@ class Mailable implements MailableContract, Renderable
      */
     public function attach($files, array $options = [])
     {
-        if(!is_iterable($files)) {
-            $files = explode(',', $files);
-        };
-        
-        foreach ($files as $file) {
-            if ($file instanceof Attachable) {
-                $file = $file->toMailAttachment();
-            }
-    
-            if ($file instanceof Attachment) {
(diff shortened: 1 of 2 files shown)


src/Illuminate/Mail/Mailable.php
```diff
@@ -872,21 +872,27 @@ public function with($key, $value = null)
...
+    public function attach($files, array $options = [])
     {
-        if ($file instanceof Attachable) {
-            $file = $file->toMailAttachment();
-        }
-
-        if ($file instanceof Attachment) {
-            return $file->attachTo($this);
+        if(!is_iterable($files)) {
+            $files = explode(',', $files);
```
BrandonSurowiec, line 878: This doesn't work when `$files` is an attachable. (Tests are failing.) You'll also need to add some tests for the new behavior.

timacdonald:
Although I can absolutely get behind the intended functionality of this PR and think it is a nice addition to the framework, I don't feel this is the best way to handle it.

Take the following example...

```php
$mailable->attach(['forge.png', 'vapor.jpg'], [
    'as' => 'logo',
    'mime' => 'image/jpg',
]);
```

This means that you cannot assign intended names / mime types to the individual elements in the array. Instead of trying to pipe this all through the existing method, I would personally prefer to see an `attachMany` method, potentially with the following API...

```php
$mail->attachMany([
    'forge.png' => ['as' => 'forge-logo', 'mime' => 'image/png'],
    'vapor.jpg' => ['as' => 'forge-logo', 'mime' => 'image/png'],
    $invoice, // instance of attachable 
    $anotherOne, // instance of attachable 
]);
```

If this is wanted for the framework, I would be happy to assist @Illusionist3886 getting this one in shape and tested.

Illusionist3886:
@tmacdonald
I also think your approach will be much cleaner. And give the developer more power to manage attachables

Illusionist3886:
@tmacdonald
I also think your approach will be much cleaner. And give the developer more power to manage attachables

Illusionist3886:
@tmacdonald
I also think your approach will be much cleaner. And give the developer more power to manage attachables

timacdonald:
@Illusionist3886 would you like me to jump into this and make the necessary changes? Or did you want to have a go at it first?

Illusionist3886:
Let me try a few days more. I can surely use your help about writing tests.

timacdonald:
No worries at all. Ping me and I can jump in, otherwise I'll check back with you in a few days.

Illusionist3886:
@timacdonald 
I've implemented the new method **attachMany**. 
Thanks for your patience, support and ideas.
We can pass **filename**, **filenames array**,  **attachable object** through this method.

```
$items = ['index.php', 'robots.txt'];
```
```
$item = 'forge.svg';
```
```
$item_details = [
                 'dark-gold.jpeg' => ['as' => 'Gold Dust.jpeg', 'mime' => 'image/jpeg'],
                 'forge.svg', // just a filename
                 'vapor.svg', // just a filename
                ];
```

We can pass any of this and also the ***attachable object*** too.

Can you confirm and write some tests for this, please?

src/Illuminate/Mail/Mailable.php
```diff
@@ -877,18 +877,40 @@ public function attach($file, array $options = [])
...
+    public function attachMany($files, array $options = [])
+    {
+        if(is_array($files))
+        {
+            foreach ($files as $file => $options)
+            {
+                $this->attach((is_int($file) ? $options : $file), (is_int($file) ? [] : $options));
+            }
+        } else {
+            $this->attach($files, $options = []);
```
BrandonSurowiec, line 909: You're forcing `$options` to always pass through an empty array here. But his method should probably force you to include an array or iterable object to not even have this "else" path, So if `$files` is null or empty an early return would probably be best.

timacdonald:
Awesome stuff @Illusionist3886 Gonna take a look at this today and then will get some tests written up and sent to your branch.

src/Illuminate/Mail/Mailable.php
timacdonald: This should be addressed with the current changes.

Illusionist3886:
Thanks for the support ☺️. 
Do you think we should add a check if the passed items are iterable or not? 
May act as a failsafe. Other than that, I think we are ready for review.

timacdonald:
No troubles at all. Always excited to see new contributors making things happen in the framework!

> Do you think we should add a check if the passed items are iterable or not?

I don't believe we should add an additional type check. We have specified the type in the docblock, which is the Laravel convention.

taylorotwell:
Sorry - I don't follow. You can't just call `attach` multiple times with notification mail messages? /cc @timacdonald

Illusionist3886:
Yes, we can definitely `attach` multiple times.
```
return $this->view('welcome')
        ->attach('goat.png')
        ->attach('dark-gold.jpeg')
        ->attach('forge.svg')
        ->with('order', $this->order);
```

But if we have **unknown** amount of items to be attached we need to do a `foreach` loop like below.
```
foreach ($item_details as $item) {
            $this->attach($item);
        }

return $this->view('welcome')
            ->with('order', $this->order);
```

So, I thought why not create a `attachMany` method which can help to write more cleaner code. Just my opinion.

PR #43083, opened 2022-07-06
[9.x] Adds Str::space() helper

aidan-casey:
While this is a pretty nominal feature, I have reached for the functionality in many projects, so I thought I would see about getting it merged into the framework. This pull request adds a `Str::space()` helper that will join the passed words with a space and ignore nullable values. In particular, this is useful when joining model attributes that could be `null`. For example:

```php
// Could be: "John", "John Doe", or "John James Doe"

Str::space($contact->first_name, $contact->middle_name, $contact->last_name);
```

The filter callback itself specifically checks for a nullable value or an empty string so as to allow for things like the digit `0` in the scenario that someone might desire to pass that.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -950,6 +950,19 @@ class Str
     }
 
     /**
+     * Join strings with a single space, ignoring nullable values.
+     *
+     * @param  mixed  ...$words
+     * @return string
+     */
+    public static function space(mixed ...$words): string
+    {
+        return join(' ', array_filter(
+            $words, fn ($value) => ! is_null($value)
+        ));
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


src/Illuminate/Support/Str.php
```diff
@@ -949,6 +949,19 @@ public static function slug($title, $separator = '-', $language = 'en')
         return trim($title, $separator);
     }
 
+    /**
+     * Join strings with a single space, ignoring nullable values.
+     *
+     * @param  mixed  ...$words
+     * @return string
+     */
+    public static function space(mixed ...$words): string
```
mateusjunges, line 958: Isn't it better to call this method `join`  and allow the user to pass the glue as a parameter?

taylorotwell:

PR #43088, opened 2022-07-07
[9.x] Add whereBoolean

milwad-dev:


src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -1914,4 +1914,12 @@ public function __clone()
     {
         $this->query = clone $this->query;
     }
+
+    /**
+     * Check boolean column.
+     */
```
GrahamCampbell, line 1920: You need to provide phpdoc.

GrahamCampbell (changes requested):

taylorotwell:

GrahamCampbell:
Why do we need this. Why can't someone do `->where('foo', false)`?

PR #43101, opened 2022-07-08
[9.x] Add `whenCounted` to JsonResource

stevebauman:
## Description

This PR adds a `whenCounted` method to the JSON resource class giving us the ability to conditionally include attributes in a response when a relation count is set on the model via `Post::withCount('comments')`.

This method acts nearly identically to `whenLoaded`, with the exception that the retrieved count attribute value will be passed into the `$value` callback (if one is provided).

I've also added a `hasAttribute` method to models to easily determine if an attribute exists inside of the models raw `$attributes` array, so I didn't have to fetch all of the models attributes to determine if it exists:

```php
if (! isset($this->resource->getAttributes()[$attribute])) {
    // ...
}
```

## Example

```php
class PostController
{
    public function show(Post $post)
    {
        return new PostResource($post->loadCount('comments'));
    }
}
```

```php
class PostResource extends PostResource
{
    {
        return [
            'id' => $this->id,
            'comments_count' => $this->whenCounted('comments'),
        ];
    }
}
```

## Important

If you'd prefer not having the `hasAttribute` method added onto Eloquent models, let me know and I can remove it!

Thanks for your time! ❤️

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -559,6 +559,17 @@ trait HasAttributes
     }
 
     /**
+     * Determine if an attribute exists.
+     *
+     * @param  string  $key
+     * @return bool
+     */
+    public function hasAttribute($key)
+    {
+        return isset($this->attributes[$key]);
+    }
+
+    /**
      * Determine if a get mutator exists for an attribute.
      *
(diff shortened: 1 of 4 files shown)


taylorotwell:
@stevebauman yeah, you can remove that `hasAttribute` method from Eloquent.

taylorotwell:
Can you PR docs for this @stevebauman ?

PR #43104, opened 2022-07-08
Using model's $table property (if available) as table name in foreignIdFor methods

bobanum:
Hello,

I'm new at "pull requesting". I hope I'm doing things right.

Currently, the $table->foreignIdFor always uses the plural version of table part of the foreign key to determines the foreign table name. So if the name is changed in the model, the name will be wrong.

I made small changes to account for the $table property. Hope that will be helpful (it is for me).

P.S.: Since getTable() returns the plural version of the model and getForeignKey() is also based on the model, I think we could get rid of the Str::plural(...) part, but I didn't dare to.

Best regards.
Martin Boudreau

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -1259,7 +1259,7 @@ class Blueprint
      * Create a new UUID column on the table with a foreign key constraint.
      *
      * @param  string  $column
-     * @param  string|null $table
+     * @param  string|null  $table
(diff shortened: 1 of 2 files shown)


taylorotwell:

bobanum:
@taylorotwell Hello, As a Laravel user, I would really need the method `foreignIdFor` to use the `$table` property. Please, help me make the right things to help you fix this bug. The changes are not intrusive. I'm new to submitting pull request. Sorry if I do anything wrong. Thnak you very much.

PR #43107, opened 2022-07-09
add getters methods to view finder interface

ph7jack:
Hello all, I think that adding some getters methods to the `ViewFinderInterface` can be helpful when creating packages for laravel.
I know this interface can be used outside laravel, if so this PR will break the compatibility in other projects, maybe the next laravel version can have more interface types?


BrandonSurowiec:
Like you mentioned this is a BC break, so you'll need to target 10.x (master).

PR #43108, opened 2022-07-09
Added initial option to key generate command

faustoFF:
Sometimes `key:generate` command is used in deploy scripts, which runs more than one time (each merge, for example). On production environment you dont need to set/overwrite application key each deploy.

Added option `--initial` to `key:generate` command allows to not set/overwrite application key if it already set.


dennisprudlo:
I think the "key:generate" command shouldn't be in a deploy script that runs on every merge.

It's like having the "db:seed" command in the deploy script but it should only execute on the first deployment.

It's probably better to have a provision script that you execute when setting up a project.

PR #43112, opened 2022-07-09
add the beforeCompilers feature

ph7jack:
At the moment we can't add a blade compiler to be run before the default blade compiler, this PR will add the ability to process the blade as we want before laravel process the blade.
I'm building a icons components [package](https://github.com/wireui/heroicons), I don't want to register all icons components in a loop or to use the namespace syntax

It works if register a components namespace (but it now includes the Icon class, just the blade file)
```
<x-icons::user />
<x-icons::home /> 
```

But it is better.
```
<x-icons.user />
<x-icons.home /> 
```
All icons use the same Icon component class but different blade views.

Thanks, Laravel team

### DIFF
--- a/src/Illuminate/Support/Facades/Blade.php
+++ b/src/Illuminate/Support/Facades/Blade.php
@@ -23,7 +23,7 @@ namespace Illuminate\Support\Facades;
  * @method static void extend(callable $compiler)
  * @method static void if(string $name, callable $callback)
  * @method static void include(string $path, string|null $alias = null)
- * @method static void beforeCompiler(callable $beforeCompiler)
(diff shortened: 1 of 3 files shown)


taylorotwell:

PH7-Jack:
@taylorotwell Why close it? It doesn't have any break changes and will be very useful for blade development. The laravel has the precompilers but it's called after compiling the blade. Please, reconsider reviewing this PR, I really need a compiler before compiling the blade.

PH7-Jack:
@taylorotwell This feature can't be a custom package, I can't touch the laravel core code from a package. If it's possible to do from a package let me know

PR #43114, opened 2022-07-09
Add missing translation prefix to pagination file

komandar:
Currently some translations of the pagination templates are not translated. Since the prefix for the strings is missing, a translation is also not easily possible without adapting the template.

The strings have now been translated in the language file and the prefixes added.

This ticket belongs to this ticket:
https://github.com/laravel/laravel/pull/5937

### DIFF
--- a/src/Illuminate/Pagination/resources/views/bootstrap-5.blade.php
+++ b/src/Illuminate/Pagination/resources/views/bootstrap-5.blade.php
@@ -29,13 +29,13 @@
         <div class="d-none flex-sm-fill d-sm-flex align-items-sm-center justify-content-sm-between">
             <div>
                 <p class="small text-muted">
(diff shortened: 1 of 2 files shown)


derekmd:
This would have to target the `master` branch since it would break 9.x installs bumping to the next patch release. The laravel/laravel PR would only apply to new Laravel app installs so this change would cause existing apps to show strings `pagination.showing`, `pagination.to`, etc. in HTML output because the translation keys are missing. For ongoing support, the new translation keys would need to be covered on https://laravel.com/docs/master/upgrade for the next 10.x major release.

PR attempts on the framework's internal pagination views are usually closed. Instead devs are given a suggestion to run `php artisan vendor:publish` and customize the views as needed in your app namespace.

PR #43118, opened 2022-07-10
Arr::except add support for asterisks as wildcard

Com-Digital:
This MR makes it possible to use asterisks as wildcard for Arr::except.
Imagine having an array like this:
```
use Illuminate\Support\Arr;

$array = [
    "users" => [
        [
            "name" => "Jakob",
            "secret" => "Jakob's secret", // should be removed
            "phone" => "123456",
        ],
        [
            "name" => "Tylor",
            "secret" => "Tylor's secret", // should be removed
            "phone" => "123456",
        ],
    ],
];
```
and you want to remove all "secret" keys.
You can now simply use this syntax:
```php
$filtered = Arr::except($array, ["users.*.secret"])
```
Before you had to write this:
```php
$filtered = Arr::except($array, ["users.0.secret", ""users.1.secret"])
```
The solution works, but it cannot handle a dynamic number of items.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -286,7 +286,13 @@ class Arr
             while (count($parts) > 1) {
                 $part = array_shift($parts);
 
-                if (isset($array[$part]) && static::accessible($array[$part])) {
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #43120, opened 2022-07-10
Factory fixes

eloquentcoder:
The factoryForModel method has a param type of 
````
class-string<\Illuminate\Database\Eloquent\Model>
````
 for the $modelName attribute in its phpdoc which makes the editor throw a red underline error. The attribute itself accepts $modelName as string type.
Screenshot 2022-07-10 at 15 15 13. 
Using a parameter type of  
````
@param  string|<\Illuminate\Database\Eloquent\Model>  $modelName 
````
removes the error and properly defines the modelName parameter type.

Similarly the modelName function has a return type of 
````
   @return class-string<\Illuminate\Database\Eloquent\Model|TModel>
 ````

which throws an error in the has function and updating the return type to 
````
     * @return string|<\Illuminate\Database\Eloquent\Model|TModel>
 ````

removes the red underline


timacdonald:
`class-string` is the correct type here. To understand it better and learn how to address the issue in your project, check out the PHPStan docs on `class-string` 

See: https://phpstan.org/writing-php-code/phpdoc-types#class-string

eloquentcoder:
> `class-string` is the correct type. To understand it better and learn how to address the issue in your project, check out the PHPStan docs on `class-string`
> 
> See: https://phpstan.org/writing-php-code/phpdoc-types#class-string

Alright already seen that. Thanks for the correction

PR #43123, opened 2022-07-11
[9.x] Add colon to database cache driver prefix key

shatterproof:
This PR makes the database cache driver prefix key consistent with the other drivers by adding a colon to the end of it, separating the prefix from the key name like so: `laravel_cache:key`.

This solves the issue presented by laravel/laravel#5817 in a more efficient manner by targeting the fix only on the database driver rather than adding an excessive underscore to the prefix keys across all of the cache drivers.

Also see laravel/laravel#5938 for the PR that ties into this one.


taylorotwell:
I don't want to make a breaking change here.

PR #43125, opened 2022-07-11
[9.x] Avoid deprecation-warnings in Str::contains with PHP 8.1

hersche:
Hi

Not sure if this should be added in front, as it would be true, but the method is only for strings :shrug: 

        if ($haystack === $needles) {
            return true;
        }

The warning we get:

> str_contains(): Passing null to parameter #1 ($haystack) of type string is deprecated in /var/www/redacted/vendor/illuminate/support/Str.php on line 229

We do not use Str::contains in our custom code so i assume it's a missing check in a dependency. It would still make sense in my opinion to have this fixed at this central place :) 

This happens in Lumen 9 with illuminate/support 9.19 running with PHP 8.1.7

If somethings missed, please let me know.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -223,6 +223,7 @@ class Str
         if (is_null($haystack) || is_null($needles)) {
             return false;
         }
+
         if ($ignoreCase) {
             $haystack = mb_strtolower($haystack);
             $needles = array_map('mb_strtolower', (array) $needles);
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
👎 You are not allowed to pass null into this function, as per the phpdoc.

hersche:
@GrahamCampbell I agree with you, but i checked our own code and we do not use this function. That's why i suspect this in the dependencies.

 @taylorotwell Thank you!

driesvints:
We reverted this. Please type-cast to the proper type as indicated in the docblock.

hersche:
As said, it's not in our custom code. So i either need to check where the issue in Lumen, Laravel or other third party dependencies is. Thanks.

PR #43127, opened 2022-07-11
Added Feature `Str::acronym()` method to generate acronym from given string.

Jagdish-J-P:
This method will be helpful to end users when they needs to generate acronym for the longer string. Many users doesn't comfortable with regular expression or they have to iterate string by exploding it in parts and then concatenating first character. But this method will help users to make it quick and easier.

Tests are also created for the same.


taylorotwell:

PR #43131, opened 2022-07-11
[9.x] Add `modelKeys()` Eloquent Builder method

stevebauman:
This PR adds a method to easily fetch a model's primary keys via the `Model::modelKeys()` method. This provides the convenience of not having to specify the primary key yourself, and having to update the call later in case you switch primary key names (ex. UUID's).

```php
// Before...
$primaryKeys = Post::pluck('id');

// After...
$primaryKeys = Post::modelKeys();
```

This method is synonymous with the Eloquent Collection method `modelKeys()`.


taylorotwell:

PR #43133, opened 2022-07-11
[9.x] Str::contains(All) - parameter-datatypes to avoid warnings

hersche:
Hi,

First, thanks again for merging my last PR that fast, @taylorotwell  i felt honored!

The comment from @GrahamCampbell  made be a bit unhappy, since i see it isn't the cleanest code. So i wanted to suggest this as well as a clean, but really strict solution that would crash if you try to pass null. It's in my perspective the better approach but obviusly causes more work.
What's the solution you people want to do should match the mindset of the Laravel-team, not mine (in case it's too strict, pls just close it). But i'm really glad if we can stop the flood of those warnings in one or the other way :)

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -222,12 +222,8 @@ class Str
      * @param  bool  $ignoreCase
      * @return bool
      */
-    public static function contains($haystack, $needles, $ignoreCase = false)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I think what you have is fine. Changing these signatures on a patch release could be a breaking change. 👍

PR #43140, opened 2022-07-12
[9.x] Allow `@class()` for component tags

### LINKED ISSUE #42977: Can't use @class on nested anonymous component
- Laravel Version: 9.18.0
- PHP Version: 8.1.7
- Database Driver & Version: MySQL (not relevant)

### Description:

When using `@class` in Blade, on an anonymous, nested component, it does not pick up the component and renders `<x-icon></x-icon>`, without any processing.

### Steps To Reproduce:

https://github.com/bram-pkg/class-directive-bug

Use `@class` on a component, results in the component not rendering as the contents of that component, but just the raw tag. In this case `<x-icon/>`

aguingand:
Closes https://github.com/laravel/framework/issues/42977. Example of use:
```blade
<x-icon @class(['lg' => $large]) />
```

### Description
My attempt to allow `@class()` directive alongside component tag attributes. I've adapted the regex used for statements in the blade compiler.

I have made 2 choices in this PR:
- The regex only match the statement `class` with an expression. Because it's the only one allowed here, it's seems more comprehensible instead of compiling the component and ignoring not handled directives.
- Double quotes are naively converted to single quotes inside the directive expression. Since we compile to `:class="..."`, double quotes would break the compiled code.

### DIFF
--- a/src/Illuminate/View/Compilers/ComponentTagCompiler.php
+++ b/src/Illuminate/View/Compilers/ComponentTagCompiler.php
@@ -112,6 +112,10 @@ class ComponentTagCompiler
                         \s+
                         (?:
                             (?:
+                                @(?:class)(\( (?: (?>[^()]+) | (?-1) )* \))
+                            )
+                            |
+                            (?:
                                 \{\{\s*\\\$attributes(?:[^}]+?)?\s*\}\}
                             )
                             |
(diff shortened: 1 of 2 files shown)


driesvints (approved):

taylorotwell:
Do we convert double quotes to single quotes in other places in this way?

aguingand:
@taylorotwell I don't think so. The same problem exists for `{{ $attributes }}` parsing, e.g.
```blade
<x-icon {{ $attributes->class("test") }} /> {{-- double quotes breaks compiled code --}}
```
https://github.com/laravel/framework/blob/ca6a9286ca0f0e99942726daa4fcda3615c0f312/src/Illuminate/View/Compilers/ComponentTagCompiler.php#L554

I have no opinion here, it looks like a bug but maybe a warning in the docs is enough.

taylorotwell:
Thanks!

PR #43147, opened 2022-07-12
[9.x] Artisan `about` Command

jbrooksuk:
This PR introduces a new `about` command that displays output about the system environment and configuration of the user's application. The command is inspired by Rails' own [`about`](https://guides.rubyonrails.org/v2.3/command_line.html#about) command, though it's beautiful and has some additional powers 🪄 

CleanShot 2022-07-13 at 12 08 46@2x

The command includes its own API so that applications and packages may push data into existing sections or create a new section with its own data:

```php
use Illuminate\Foundation\Console\AboutCommand;

AboutCommand::add('Laravel Vapor', 'Vapor Runtime', '1.0.0');
```

You can also add multiple items:

```php
AboutCommand::add('Environment', [
    'Laravel Cashier' => 'Stripe',
    'Laravel Nova' => '4.0.0'
]);
```

It's expected that packages will start to add their own sections of information, which could make this command output a lot of information. You can choose to filter down certain sections with `--only` and pass a comma-separated list of sections in `snake_case`.

CleanShot 2022-07-13 at 12 10 13@2x

Finally, you can choose to output this information as JSON by using the `--json` option (which can also be combined with `--only`):

CleanShot 2022-07-13 at 12 11 40@2x

### DIFF
--- a/src/Illuminate/Foundation/Console/AboutCommand.php
+++ b/src/Illuminate/Foundation/Console/AboutCommand.php
@@ -1,0 +1,217 @@
+<?php
+
+namespace Illuminate\Foundation\Console;
+
+use Illuminate\Console\Command;
+use Illuminate\Support\Composer;
+use Illuminate\Support\Str;
+use Symfony\Component\Console\Attribute\AsCommand;
+
+#[AsCommand(name: 'about')]
+class AboutCommand extends Command
+{
+    /**
+     * The console command signature.
+     *
+     * @var string
+     */
+    protected $signature = 'about {--only= : The section to display}
+                {--json : Output the information as JSON}';
+
+    /**
+     * The name of the console command.
+     *
+     * This name is used to identify the command during lazy loading.
+     *
+     * @var string|null
+     *
+     * @deprecated
+     */
+    protected static $defaultName = 'about';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Display basic information about your application';
+
+    /**
+     * The Composer instance.
+     *
+     * @var \Illuminate\Support\Composer
+     */
+    protected $composer;
+
+    /**
+     * The data to display.
+     *
+     * @var array
+     */
+    protected static $data = [];
+
+    /**
+     * Create a new command instance.
+     *
+     * @param  \Illuminate\Support\Composer  $composer
+     * @return void
+     */
+    public function __construct(Composer $composer)
+    {
+        parent::__construct();
+
+        $this->composer = $composer;
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return int
+     */
+    public function handle()
+    {
+        $this->gatherApplicationInformation();
+
+        collect(static::$data)->sortBy(function ($data, $key) {
+            $index = array_search($key, ['Environment', 'Cache', 'Drivers']);
+
+            if ($index === false) {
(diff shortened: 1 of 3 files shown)


src/Illuminate/Foundation/Console/AboutCommand.php
```diff
@@ -0,0 +1,154 @@
...
+     * @deprecated
+     */
+    protected static $defaultName = 'about';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'About your application\'s environment';
```
nunomaduro, line 35: It seems a little bit more than the Application's environment at this point?

src/Illuminate/Support/Composer.php
```diff
@@ -107,4 +107,21 @@ public function setWorkingPath($path)
...
     }
+
+    /**
+     * Get the version of Composer.
+     *
+     * @return string|null
+     */
+    public function getVersion()
+    {
+        $command = array_merge($this->findComposer(), ['-V']);
```
nunomaduro, line 118: @jbrooksuk Did you manage to test this with different composer versions?

danharrin:
This is great, for package maintainers, it will allow us to more easily get debugging information from our users, which will help us fix issues. Not just version numbers, but checking if they have fallen into common traps like not linking their public storage directory.

Very nice!

src/Illuminate/Foundation/Console/AboutCommand.php
```diff
@@ -0,0 +1,154 @@
...
+
+    /**
+     * Gather data about the application's environment.
+     *
+     * @return array
+     */
+    protected function gatherEnvironmentData()
+    {
+        static::add('Environment', [
+            'Laravel Version' => app()->version(),
```
sebdesign, line 107: Why not `$this->laravel->version()`?

jbrooksuk: I've switched to using this `$this->laravel` for consistency.

pxlrbt:
For sites with many packages, I think it might be useful to filter results. For example `php artisan about --only=environment,laravel_nova`

voidgraphics:
Great PR, it's a little hard to read though with the values being so far from the keys. Especially in full screen.

fgaroby:
@pxlrbt: wouldn't it be better using the `vendor/package_name` syntax ?

jbrooksuk:
Okay, I've added `--only` which accepts a command-separated list of section names in snake case, i.e. `Laravel Nova => laravel_nova`.

I've also added an `about` method to `ServiceProvider` as a shortcut to adding sections.

jbrooksuk:
> Great PR, it's a little hard to read though with the values being so far from the keys. Especially in full screen.

The about is dynamic based on the width of your terminal:

CleanShot 2022-07-13 at 10 56 40@2x

jbrooksuk:
We'll address this in the Console UI Improvements PR, on which this PR is based on.

dennisprudlo:
@jbrooksuk Just a thought: How about getting rid of the `--only` option and use an optional argument list instead?
So `about {sections?*} {--json}`. This way the command reads better:
```shell
php artisan about environment
php artisan about environment laravel_nova some_other_section --json
```

omerbaflah:
Great work!  @jbrooksuk  👏🏻

PR #43152, opened 2022-07-12
[9.x]: add `driver` docblock in `cache` facade

dammy001:


taylorotwell:
I think you should just use `store`.

PR #43153, opened 2022-07-13
[9.x] Add support for a secondary database connection and custom migration path to the Testing/RefreshDatabase trait

berrugo:
# Type: Improvement

This pull requests allows users to specify a custom `$database` and/or  migration `$path` when using the RefreshDatabase trait.

## Use case
My application uses the default database and a secondary database with its own custom migration files path. My FeatureTests require the secondary database and I wasn't able to use the RefreshDatabase trait because it only supports the default database and the default migration files path.
The purpose of this PR is to make devs life easier and be able to use RefreshDatabase out of the box when your tests require a database that's not the default one, without the need to call custom artisan commands on setup() 

## How it works
This works pretty much like the `$seeder` [property](https://laravel.com/docs/9.x/database-testing#running-seeders). Simply specify in your test case a `$database` and/or a `$path` and the RefreshDatabase trait will add these as parameters to the `migrate:fresh` artisan command.

```
class BeccaTest extends TestCase
{
    use RefreshDatabase;

    /**
     * Use a custom database connection
     *
     * @var string
     */
    protected $database = 'becca';

    /**
     * Use a custom migration files path
     *
     * @var string
     */
    protected $path = 'database/migrations/becca';
}
```
After setting these 2 properties the migrate:fresh command will be executed with these params:

```
->with('migrate:fresh', [
                '--drop-views' => false,
                '--drop-types' => false,
                '--seed' => false,
		'--database' => 'becca',
		'--path' => 'database/migrations/becca'
```

The Feature tests are passing and I also added few more to test these new properties.

### DIFF
--- a/src/Illuminate/Foundation/Testing/RefreshDatabase.php
+++ b/src/Illuminate/Foundation/Testing/RefreshDatabase.php
@@ -58,7 +58,7 @@ trait RefreshDatabase
     {
         return array_merge([
             '--seed' => $this->shouldSeed(),
-            '--seeder' => $this->seeder()],
+            '--seeder' => $this->seeder(), ],
             $this->database() ? ['--database' => $this->database()] : [],
             $this->path() ? ['--path' => $this->path()] : []
         );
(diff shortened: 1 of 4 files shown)


taylorotwell:

PR #43156, opened 2022-07-13
[9.x] Artisan `model:show` command

jessarcher:
This PR introduces a new `model:show` command that displays information for a model. It combines information from the database alongside information from Eloquent to give you a complete overview of your model.

This command is especially useful as your application grows and you have multiple migrations for a single table, with all sorts of casts, accessors, etc.

image

Not only does it show columns from the database, but it also displays "virtual" attributes - I.e. accessors/attributes that don't have a matching column in the database (See `first_name` and `last_name` in the screenshot).

_Note that the column type information depends on the database. For example, PostgreSQL doesn't have unsigned information and SQLite doesn't have bigints or column lengths._

It also conveniently displays model relationships, including those defined in traits.

There is a `--verbose` (or `-v`) flag that will also show any default values. It first checks for any defaults specified in the `$attributes` key of your model before falling back to any default specified in the database.

image

There is also a handy `--json` flag that will return everything as JSON, which I could imagine useful for tools like https://github.com/barryvdh/laravel-ide-helper, https://github.com/nunomaduro/larastan, and https://laravelshift.com.

<details>
<summary>JSON Example</summary>

```json
{
  "class": "App\\Models\\User",
  "database": "mysql",
  "table": "users",
  "attributes": [
    {
      "name": "id",
      "type": "bigint unsigned",
      "increments": true,
      "nullable": false,
      "default": null,
      "unique": true,
      "fillable": false,
      "hidden": false,
      "appended": null,
      "cast": "int"
    },
    {
      "name": "name",
      "type": "string(255)",
      "increments": false,
      "nullable": false,
      "default": "Jess",
      "unique": false,
      "fillable": false,
      "hidden": false,
      "appended": null,
      "cast": null
    },
    {
      "name": "email",
      "type": "string(255)",
      "increments": false,
      "nullable": false,
      "default": null,
      "unique": true,
      "fillable": true,
      "hidden": false,
      "appended": null,
      "cast": null
    },
    {
      "name": "email_verified_at",
      "type": "datetime",
      "increments": false,
      "nullable": true,
      "default": null,
      "unique": false,
      "fillable": false,
      "hidden": false,
      "appended": null,
      "cast": "datetime"
    },
    {
      "name": "password",
      "type": "string(255)",
      "increments": false,
      "nullable": false,
      "default": null,
      "unique": false,
      "fillable": true,
      "hidden": true,
      "appended": null,
      "cast": null
    },
    {
      "name": "remember_token",
      "type": "string(100)",
      "increments": false,
      "nullable": true,
      "default": null,
      "unique": false,
      "fillable": false,
      "hidden": true,
      "appended": null,
      "cast": null
    },
    {
      "name": "created_at",
      "type": "datetime",
      "increments": false,
      "nullable": true,
      "default": null,
      "unique": false,
      "fillable": false,
      "hidden": false,
      "appended": null,
      "cast": "date"
    },
    {
      "name": "updated_at",
      "type": "datetime",
      "increments": false,
      "nullable": true,
      "default": null,
      "unique": false,
      "fillable": false,
      "hidden": false,
      "appended": null,
      "cast": "datetime"
    },
    {
      "name": "foo",
      "type": "string(255)",
      "increments": false,
      "nullable": false,
      "default": "a default",
      "unique": false,
      "fillable": false,
      "hidden": false,
      "appended": null,
      "cast": "App\\Casts\\CustomCast"
    },
    {
      "name": "bar",
      "type": "string(255)",
      "increments": false,
      "nullable": false,
      "default": null,
      "unique": false,
      "fillable": false,
      "hidden": false,
      "appended": null,
      "cast": "accessor"
    },
    {
      "name": "baz",
      "type": "string(255)",
      "increments": false,
      "nullable": false,
      "default": null,
      "unique": false,
      "fillable": false,
      "hidden": false,
      "appended": null,
      "cast": "attribute"
    },
    {
      "name": "team_id",
      "type": "bigint unsigned",
      "increments": false,
      "nullable": false,
      "default": null,
      "unique": false,
      "fillable": false,
      "hidden": false,
      "appended": null,
      "cast": null
    },
    {
      "name": "first_name",
      "type": null,
      "increments": false,
      "nullable": null,
      "default": null,
      "unique": null,
      "fillable": false,
      "hidden": false,
      "appended": true,
      "cast": "accessor"
    },
    {
      "name": "last_name",
      "type": null,
      "increments": false,
      "nullable": null,
      "default": null,
      "unique": null,
      "fillable": false,
      "hidden": false,
      "appended": false,
      "cast": "attribute"
    }
  ],
  "relations": [
    {
      "name": "tokens",
      "type": "MorphMany",
      "related": "Laravel\\Sanctum\\PersonalAccessToken"
    },
    {
      "name": "notifications",
      "type": "MorphMany",
      "related": "Illuminate\\Notifications\\DatabaseNotification"
    }
  ]
}
```
</details>

The results do depend on the current state of the database, so any changes made outside of a migration _will_ be reflected here, but any changes from not-run migrations will not.

I'd also like to acknowledge https://github.com/barryvdh/laravel-ide-helper for some of the logic around finding attributes and relationships, and @nunomaduro for his help with the UI :nail_care:


lucasgiovanny:
That's awesome to have! I see myself all the time running to the database to check some model information, especially when you jump from project to project. Great job! 😃

src/Illuminate/Foundation/Console/ShowModelCommand.php
```diff
@@ -0,0 +1,433 @@
...
+     * Qualify the given model class base name.
+     *
+     * @param  string  $model
+     * @return string
+     *
+     * @see \Illuminate\Console\GeneratorCommand
+     */
+    protected function qualifyModel(string $model)
+    {
+        $model = ltrim($model, '\\/');
```
innocenzi, line 419:
```suggestion
        if (class_exists($model)) {
            return $model;
        }
        
        $model = ltrim($model, '\\/');
```

Currently, it's not possible to use the command when not following Laravel's default architecture - using a domain-driven approach, models are not necessarily in `App/Models`. 

With this condition, I can do `php artisan model:show Domain\\Reporting\\Models\\Duty`

SjorsO:
Looks great!

It would be useful if it also shows indexes that each column has (unique, primary, foreign key). I'm working on a big project where some columns are missing important indexes (sometimes even primary keys). This command could make it easier to find those missing indexes.

Maybe the command could even display a yellow warning banner: 
> **Warning**
> This table has no primary key index

imanghafoori1:
I suggest `php artisan model:list User` and `php artisan model:list` (with no args) will loop through all the psr-4 paths of the composer.json and show a list of all the `Model` classes with some basic information.

hamidafghan:
@jessarcher @taylorotwell 

Yes, It is a great feature to have, but it will be very rare to use the command. I am afraid you push some feature which is not used widely. Instead, if you are able to ship it as an official Laravel package, it would be super-efficient to expand it later as well.

nunomaduro:
@jessarcher I've put this pull request on draft, so you can review my suggestion to the layout: https://github.com/laravel/framework/pull/43170.

src/Illuminate/Foundation/Console/ShowModelCommand.php
jessarcher: Makes sense!

This method was taken pretty much verbatim from the `make:*` commands so I'm assuming they're similarly affected?

https://github.com/laravel/framework/blob/77560968f6a308d2abddb832ad733bee2b15392e/src/Illuminate/Console/GeneratorCommand.php#L207-L228

innocenzi: They are, unfortunately :(

With a DDD architecture, `make:*` commands are not really usable - we have to move files and update namespaces manually

freekmurze:
Very nice work! 👍

I think it could be worthwile to refactor the column/relation finding logic to it’s own class that the command would use. This way, others can also make use of the gathered information.

It could also be cool to somehow let a model append custom information to the output.

M6268:
A suggestion , add this info as doc inside each model (like laravel-ide-helper does).
whenever migration occurs or manually using the command (model:show) update it in the models.
So we can reduce checking each time by manually querying a command.

omerbaflah:
Great work! @jessarcher  👏🏻

spekulatius:
This looks awesome, thank you @jessarcher!

PR #43157, opened 2022-07-13
Added Eloquent withCountWhereHas method

tnt-freskim-veliu:
There are a lot of times that we need to count a relationship using `withCount` with the same conditions that we use for `whereHas`.

This method simplifies this selection by avoiding repeating the code used both to filter with `whereHas` and then to count the same records using `withCount`. 

### Before
```
CollectionModel::query()->whereHas('products', function ($query) {
    $query->where('is_enabled', true)->where('is_for_sale', true);
})->withCount(['products' => function ($query) {
    $query->where('is_enabled', true)->where('is_for_sale', true);
});
```
### After
```
CollectionModel::query()->withCountWhereHas('products', function ($query) {
    $query->where('is_enabled', true)->where('is_for_sale', true);
});
```

I will add tests later 😄 

PS. The idea came from this PR: https://github.com/laravel/framework/pull/42597

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -170,6 +170,23 @@ trait QueriesRelationships
     }
 
     /**
+     * Add a relationship count / exists condition to the query with where clauses.
+     *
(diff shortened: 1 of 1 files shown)


taylorotwell:

Thanks!

Issue #43158, opened 2022-07-13
PendingMail send function missing a @throws TransportExceptionInterface annotation


Vrolyx:
- Laravel Version: 9.19
- PHP Version: 8.1.6

### Description:
PendingMail send function is missing a @throws TransportExceptionInterface annotation. This missing annotation will result in a false positive with phpstan. The MailerInterface throws this exception within the Symfony mailer component.

### Steps To Reproduce:

```
try {
            Mail::to(
                'user@example.com'
            )->send($mail);
        } catch (\Exception $e) {
            // A TransportException is thrown on failure, but phpstan gives a false positive due to the missing @throws TransportExceptionInterface annotation
        }
```

driesvints:
Heya, thanks but these methods don't directly throw this exception. We only add throw annotations to methods that directly throw an exception.

PR #43159, opened 2022-07-13
Add @throws annotation to send function in PendingMail and Mailer class

Vrolyx:
Fixes issue #43158.


driesvints:
Heya, thanks but these methods don't directly throw this exception. We only add throw annotations to methods that directly throw an exception.

PR #43138, opened 2022-07-12
[file]{SessionGuard.php}=> add `getGuardName()` method

nerdy-panda:
some times for each guard , we need receive data from config or database ,for this should get guard name 
for solve this problem i add ``getGuardName() `` to ``SessionGuard.php``

PR #43163, opened 2022-07-13
[10.x] Set `SessionGuard::$name` as readonly

crynobone:
Alternative to #43138 and make it easier to do the following:

image


nerdpanda-ir:
good idea thanks

PR #43170, opened 2022-07-13
[9.x] Suggest a different layout for the `model:show` command

nunomaduro:
This pull request suggest a different layout for the `model:show` command, that handlers better small terminal sizes, respects padding and margins, and displays the information in a prettier way overall. 

Before:
before

After:
after


taylorotwell:
Will let @jessarcher review this

jessarcher:
I've cherry-picked this into my branch, thanks heaps Nuno!

PR #43172, opened 2022-07-13
create getGuardName() method

nerdy-panda:

### DIFF
--- a/src/Illuminate/Auth/AuthManager.php
+++ b/src/Illuminate/Auth/AuthManager.php
@@ -54,6 +54,16 @@ class AuthManager implements FactoryContract
     }
 
     /**
+     * get The array of created "drivers".
+     *
+     * @return array
+     */
+    public function getGuards()
+    {
+        return $this->guards;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


driesvints:
Please resubmit this with a thorough explanation why you need this. Thanks

nerdpanda-ir:
hey @driesvints 
i have two guard in system (admin,user(web))
per admin guard  and  web (user) guard i have segregate login page and panel page !!!
in guest middleware i should redirect to panel via guard name 
for example if guard is ``null`` or ``web`` i should redirect to ``user-panel``
but if guard is ``admin`` i should redirect to ``admin-panel``
in first method i add if statements per available guards 
like that 
firstMethod
but this method is very bad because is not dynamic and 
if tomorrow  i add more guards to system i should go to middlewares and add if statements 
this method failed open-close principe 

I thought about this problem and find way for that :
i went in ``config/auth.php`` and  defined panel route per each guards like this : 
auth php
and then modify guest middleware : 
RedirectIfAuthenticated php2
my code is more beautiful  if i have ``getGuardName()`` method  : 
index

PR #43173, opened 2022-07-13
[9.x] Added env_as_array helper.

bokoch:
This PR adds `env_as_array` helper.
Sometimes it's useful to be able to set array values in an env variable. Here is an example of usage in a config file:
`'foo' => env_as_array('FOO'),`
or with a defined values separator and a default value
`'foo' => env_as_array('FOO', '|', ['bar', 'baz']),`
And in `.env` file it looks like that:
`FOO=bar,baz`

### DIFF
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -136,11 +136,11 @@ if (! function_exists('env')) {
 
 if (! function_exists('env_as_array')) {
     /**
-     * Gets the value of an environment variable as array
+     * Gets the value of an environment variable as array.
      *
-     * @param string $key
-     * @param string $separator
-     * @param array $default
+     * @param  string  $key
+     * @param  string  $separator
+     * @param  array  $default
      * @return array
      */
     function env_as_array(string $key, string $separator = ',', array $default = []): array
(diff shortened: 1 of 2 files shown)


antonkomarev:
After that, I can foresee someone will want to return collections :)

antonkomarev:
```
$array = explode(',', env('KEY', ''));
```

antonkomarev:
What if `trim` of each array value is required?

src/Illuminate/Support/helpers.php
```diff
@@ -134,6 +134,26 @@ function env($key, $default = null)
...
+     * @return array
+     */
+    function env_as_array(string $key, string $separator = ',', array $default = []): array
+    {
+        $value = env($key);
+        if (null === $value) {
+            return $default;
+        }
+
+        return explode($separator, $value);
```
antonkomarev, line 153: This may fail if value is null in future PHP versions. Explode should accept only strings

taylorotwell:

GrahamCampbell:
I find it fine to just put `explode(',', env('FOO') ?? '')` in my config files.

PR #43176, opened 2022-07-13
[8.x] Fix transaction attempts counter for sqlsrv

krasucki:
This brings back transactions attempts for sqlsrv driver

### DIFF
--- a/src/Illuminate/Database/SqlServerConnection.php
+++ b/src/Illuminate/Database/SqlServerConnection.php
@@ -29,7 +29,7 @@ class SqlServerConnection extends Connection
     {
         for ($a = 1; $a <= $attempts; $a++) {
             if ($this->getDriverName() === 'sqlsrv') {
-                return parent::transaction($callback);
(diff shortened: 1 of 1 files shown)


taylorotwell:
When was it removed and why?

krasucki:
First introduced in #1883 (well not precisely, because at that time there wasn't $attempts implemented yet, but that was the initial version of this sqlsrv condition) , then during some bug fixing in f4588f8851aab1129f77d87b7dc1097c842390db, for some reason the $attempts was omitted. For me it look like mistake, because there is no reason not to re-run transactions for (pdo) sqlsrv driver.

This is interesting because during all that time nobody noticed this (I must admit that driver is not as popular, but still..) 
 
maybe  @GrahamCampbell can help here?

GrahamCampbell:
The commit you reference does not remove the parameter, but it's possible that not adding it at that point was a mistake.

GrahamCampbell:
I think we could go-ahead and merge this, and see if anyone complains.

PR #43179, opened 2022-07-14
[9.x] - Convert locale to string to avoid null

hersche:
Context: #43125 

This is causing the warning, at least, maybe there are more places. I know it's needed to pass a string here, so as a first action, i added a strval to the calls, which didn't helped, so i added it in the method itself. As this is solved with casting (converting in this case, to be exact i think) i wouldn't expect issues and hope you people are fine with it.

### DIFF
--- a/src/Illuminate/Translation/Translator.php
+++ b/src/Illuminate/Translation/Translator.php
@@ -413,7 +413,7 @@ class Translator extends NamespacedItemResolver implements TranslatorContract
      */
     public function setLocale($locale)
     {
-        if (Str::contains($locale, ['/', '\\'])) {
+        if (Str::contains(strval($locale), ['/', '\\'])) {
             throw new InvalidArgumentException('Invalid characters present in locale.');
         }
 


hersche:
Investigated a bit more, issue was: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Translation/TranslationServiceProvider.php#L25

This config-value was not set (it doesn't make too much sense for our use). Not sure this is because of Lumen or because we missed it in a upgrade-migration from v4 -> v9. Anyway, it solves our issue.

I guess a cast to a empty string in the TranslationServiceProvider would make even more sense than in the suggested Translator - what do you think?

If you think, no change is needed you can also close it as it's solveable.

driesvints:
> This config-value was not set (it doesn't make too much sense for our use).

Heya, this config value should always be set. Please do not remove it.

hersche:
> Heya, this config value should always be set. Please do not remove it.

I need to add it to the .env and to config/app.php as it was not there. The suggestion wouldn't be to remove it from TranslationServiceProvider of course but to put the config-value into a strval, so if someone else have not set it, it would not be null.

But i guess that will not happen.

hersche:
Well, i did that (in our case with the DEFAULT_LOCALE we had already), and for me the issue is solved.

I just thinked if there can be other people which are flooded with this warning, the more-correct behaviour in my opinion on this line: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Translation/TranslationServiceProvider.php#L25

would be like
> $locale = strval($app['config']['app.locale']);

So if it's not set, it would be "" at least. Or, if not set, change it to en (as this is the default value in config/app.php if you would do a new setup).

I simply dislike the first answer as we never removed this config nor i ment to remove it from the above line. Either i wasn't clear with my description or i misunderstood the answer.

dennisprudlo:
Maybe `""` won't throw a warning, however it is as invalid for a locale as `null`. I think @driesvints meant, that the `app.locale` key is a default setting that comes with laravel and one should not remove it. Just like `view.paths` is required to be set for laravel to work.

Some other developer must have removed that configuration then, because it's expected to be there.

driesvints:
> the app.locale key is a default setting that comes with laravel and one should not remove it

This is what I meant indeed.

hersche:
Yeah, i guess it only affects projects which upgraded over major-versions.

I don't really agree it's invalid on a same level, as null is the wrong datatype by the (mentioned by driesvints in the other ticket) docs.
As the warnings are flooding the logs, people might will miss important stuff by that.

However, i don't want to make this endless. There's a decision against it, so it i can respect this. It's closed, so should be fine like that.

PR #43188, opened 2022-07-14
[9.x] Add isJsPath() to Vite

hosseinhezami:
This PR for determine whether the given path is a JavaScript file.
It can be used in makeTag() method.

```
$this->isJsPath($url)
// true | false
```

[https://vitejs.dev/config/shared-options.html#resolve-extensions](https://vitejs.dev/config/shared-options.html#resolve-extensions)


hosseinhezami:
@taylorotwell

PR #43191, opened 2022-07-15
Add a Terminate method to the Job Middleware so it operated the same as Request Middleware

bretto36:
Hello there. 

I've recently needed to incorporate a Job Middleware into my project. I want to be able to run something at the end of the job without having to mess with the `handle` method of the job. 

The job middleware does a great job of handling something BEFORE the job runs. But i'd love to trigger something at the end. 

This is my attempt at demonstrating what i mean. I'm not 100% sure on the position of the method within the Queue Handler or whether i've completely misplaced it. I just found the `dispatchThroughMiddleware` method and based it on that. 

I've added some tests as well and am happy to receive feedback on how to make those a bit nicer. I noticed there wasn't a test for proving the Middleware's handle method was called so i added that as well. 

Thank you!

### DIFF
--- a/src/Illuminate/Queue/CallQueuedHandler.php
+++ b/src/Illuminate/Queue/CallQueuedHandler.php
@@ -72,6 +72,8 @@ class CallQueuedHandler
             $this->ensureUniqueJobLockIsReleased($command);
         }
 
+        $this->terminateThroughMiddleware($job, $command);
+
         if (! $job->hasFailed() && ! $job->isReleased()) {
             $this->ensureNextJobInChainIsDispatched($command);
             $this->ensureSuccessfulBatchJobIsRecorded($command);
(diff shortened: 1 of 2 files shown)


src/Illuminate/Queue/CallQueuedHandler.php
```diff
@@ -125,6 +127,28 @@ protected function dispatchThroughMiddleware(Job $job, $command)
...
+     * @return mixed
+     */
+    protected function terminateThroughMiddleware(Job $job, $command)
+    {
+        if ($command instanceof \__PHP_Incomplete_Class) {
+            throw new Exception('Job is incomplete class: '.json_encode($command));
+        }
+
+        return (new Pipeline($this->container))->send($command)
+                                               ->through($command->middleware ?? [])
```
bretto36, line 144: I don't know if i need to recalculate the middleware array, i assume that it's already using the array merge from the `dispatchThroughMiddleware` method

bretto36:
I will fix the other failing tests once i have some feedback.

uintaam:
@taylorotwell This would be a really useful feature as we have things like data pipelines that batch data and push on the terminate call in a middleware. Without this in jobs we end up with missing data or not batching.

PR #43193, opened 2022-07-15
[9.x] Change checking null condition for $key parameter in Arr::prepend()

hosseinhezami:
Change checking null condition for $key parameter in Arr::prepend()

`if (func_num_args() == 2)`
to
`if (is_null($key))`


driesvints:
As you can see in the failing tests, this is a breaking change.

PR #43196, opened 2022-07-15
[8.x] Fix broken regex rule in single string-based rule case

AbdlrahmanSaberAbdo:
The current behavior in Laravel 8.x when using a single string-based rule, ex 
```
  $data = ['users' => [['name' => 'john']]]
  
  // Fails. preg_match throws malformed regex exception.
  $rules = ['users.*' => 'regex:/^(john|doe)$/i'];

  // Fails. preg_match throws malformed regex exception.
 $rules = ['users.*' => 'not_regex:/^(john|doe)$/i'];
```

after debugging I figure out that when pass `regex` contains `|` it broke the `regex` string to different arrays.

Now you can supply a regex validation rule inside of a single string-based rule, ex:

```
  $data = ['users' => [['name' => 'john']]]
  
  // passes
  $rules = ['users.*' => 'regex:/^(john|doe)$/i'];

  // passes
  $rules = ['users.*' => 'not_regex:/^(john|doe)$/i'];
```

PR #43198, opened 2022-07-15
[8.x] Fix breaks in tests according to psr/container changes

AbdlrahmanSaberAbdo:
This PR fix two breaks happen when run `tests`

# First
According to `psr/container` changes should put data type to return value in `has` method, check this changes https://github.com/php-fig/container/commit/6c2bc7fc1619d0f3f3442be69f177e6103d27dc7

# Second
seems that the `ninja` password appears in data leaks so it was returning true instead of returning false.

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -191,7 +191,7 @@ class Container implements ArrayAccess, ContainerContract
      *
      * @return bool
      */
-    public function has($id)
+    public function has($id):bool
     {
         return $this->bound($id);
     }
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
Laravel 8 does not support the psr container v2. https://github.com/laravel/framework/blob/8.x/composer.json#L31

GrahamCampbell:
Moreover, adding that type would be a major breakng change.

AbdlrahmanSaberAbdo:
You're right, I made another pull request before that but it falls in test cases not related to my changes that's why I tried to fix these breaks, please check it https://github.com/laravel/framework/pull/43196 @GrahamCampbell

PR #43199, opened 2022-07-15
[8.x] Fix broken regex rule in single string-based rule case

AbdlrahmanSaberAbdo:
The current behavior in Laravel 8.x when using a single string-based rule, ex 
```
  $data = ['users' => [['name' => 'john']]]
  
  // Fails. preg_match throws malformed regex exception.
  $rules = ['users.*' => 'regex:/^(john|doe)$/i'];

  // Fails. preg_match throws malformed regex exception.
 $rules = ['users.*' => 'not_regex:/^(john|doe)$/i'];
```

after debugging I figure out that when pass `regex` contains `|` it broke the `regex` string to different arrays.

Now you can supply a regex validation rule inside of a single string-based rule, ex:

```
  $data = ['users' => [['name' => 'john']]]
  
  // passes
  $rules = ['users.*' => 'regex:/^(john|doe)$/i'];

  // passes
  $rules = ['users.*' => 'not_regex:/^(john|doe)$/i'];
```

### DIFF
--- a/src/Illuminate/Validation/ValidationRuleParser.php
+++ b/src/Illuminate/Validation/ValidationRuleParser.php
@@ -306,7 +306,7 @@ class ValidationRuleParser
     /**
      * Determine if the rule is a regular expression rule.
      *
-     * @param string $rule
+     * @param  string  $rule
      * @return bool
      */
     protected static function ruleIsRegex(string $rule): bool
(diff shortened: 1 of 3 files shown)


taylorotwell:
This is already documented I believe.

taylorotwell:
CleanShot 2022-07-15 at 08 34 19@2x

AbdlrahmanSaberAbdo:
@taylorotwell

PR #43207, opened 2022-07-15
[9.x] Add autocomplete directive

intrepidws:
I know that there have been a flurry of similar directives added recently, but I believe that this adds value and allows for a smooth developer experience.

This PR allows use of the `autocomplete` HTML attribute through a shorthand syntax via a Blade directive. While my use case for this directive is to disable autocomplete, this attribute has many accepted values ('on', 'name', 'email', etc). Since this attribute requires a value, my implementation of this directive works a little differently from the `@required()`, `@readonly()` and `@disabled()` directives:

```
<input type="text" @autocomplete($conditional, 'off') />
```

If a second argument is not provided, the value will default to `on`.

If merged, I will PR to the docs.

Note that I have also added a missing test for the recently added `@readonly()` directive.

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesConditionals.php
@@ -361,6 +361,20 @@ trait CompilesConditionals
     }
 
     /**
+     * Compile an autocomplete block into valid PHP.
+     *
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't think I'm going to merge anymore of these right now. People should be able to create custom directives for them.

PR #43209, opened 2022-07-15
fix: Fix empty arrays on request bodies

PAG-Erik:
This PR fixes a problem I've encountered some times. The problem is that on some APIs, an empty array as the request's body may lead to false errors regarding parsing the payload. While it may seem that sending nothing on an Http::post($url) method may suffice it, it still sends an empty "[]" array.

**TEST:**
proof

Making the request using solely _curl_ worked as expected.


dennisprudlo:
Isn't that breaking, when some API actually expects an empty array? This will convert all empty arrays to null.

driesvints:
No plans to change this, sorry.

PR #43211, opened 2022-07-15
`valueOr` method added to Eloquent query builder

chinmaypurav:
In addition to the `value` and `valueOrFail` methods, this method will serve as a way to provide a fallback default value if the `value === null`.

This feature is inspired from the `firstOr` method.

**Note:** Blank string will be treated as a valid value and will not return the fallback value.

Relevant tests have also been written.


taylorotwell:

PR #43215, opened 2022-07-15
[8.x] clear Facade resolvedInstances in queue worker resetScope callback

### LINKED ISSUE #43151: Scoped container bindings accessed via Facade are not reset
- Laravel Version: 8.x
- PHP Version: 8.0.26
- Database Driver & Version: Mysql 8

### Description:
Accessing a scoped binding via a Facade will cache the resolved instance statically on the Facade meaning its not reset when the `scopedInstances` are cleared on the container.

I don't belive this is intended/desired or intuitive behavior and it certainly wasn't immediately obvious why my service was not being reinitialised within my workers.

I would be happy to attempt a pr to resolve if it is agreed that this this should be fixed.

### Steps To Reproduce:
- Create a scoped singleton binding in the container.
- Resolve this binding via a facade within a worker or octane server 
- resolved instance accessed via Facade is not cleared/reset between request/jobs.

ankorstore-haddowg:
Fixes #43151

Ensure scoped bindings are cleared from resolved facade instances when a worker resets its scope, as per https://github.com/laravel/octane/blob/1.x/src/CurrentApplication.php#L24

### DIFF
--- a/src/Illuminate/Queue/QueueServiceProvider.php
+++ b/src/Illuminate/Queue/QueueServiceProvider.php
@@ -16,6 +16,7 @@ use Illuminate\Queue\Failed\DatabaseUuidFailedJobProvider;
 use Illuminate\Queue\Failed\DynamoDbFailedJobProvider;
 use Illuminate\Queue\Failed\NullFailedJobProvider;
 use Illuminate\Support\Arr;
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks @ankorstore-haddowg

PR #43219, opened 2022-07-16
[9.x] Collection->random() add preserve keys option

NathanaelGT:
This PR adds preserveKeys parameter to `Collection->random()`

before:
```php
Arr::random($collection->all(), 2, true);
```

after:
```php
$collection->random(2, true);
```


driesvints:
Changing the method signature here is a breaking change unfortunately. Please send this in to master.

PR #43220, opened 2022-07-16
[9.x] Make incrementQuietly and decrementQuietly public to enforce consistency with similar features

eduance:
This PR enforces the consistency for the incrementQuietly and decrementQuietly to make the methods public. The reasoning behind this is that the usage for this method should be accessible from any model instead a developing having to worry about implementing the identical code himself.

**Is this backwards compatible?**

Yes. A change from protected to public should not affect any use cases. This PR has been chosen after carefully reading the contribution guide. :-) If you think this could be a breaking change or a too dangerous adjustment, let me know.

**Why shouldn't you just do this at the subclass level?**

In my opinion, 

- to keep the code consistent with similar features like saveQuietly and updateQuietly which both also have a public access modifier. 
- a user model shouldn't have to implement the incrementQuietly() code on its own to call the parent method, this would violate the single responsibility principle.
- why repeat yourself if Laravel provides you with this identical logic? :-)

Thanks,

Eduance

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -929,7 +929,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
      * @param  array  $extra
      * @return int
      */
-    protected function incrementQuietly($column, $amount = 1, array $extra = [])
+    public function incrementQuietly($column, $amount = 1, array $extra = [])
     {
         return static::withoutEvents(function () use ($column, $amount, $extra) {
             return $this->incrementOrDecrement($column, $amount, $extra, 'increment');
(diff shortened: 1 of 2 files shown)


Sophist-UK:
See #43210 for discussion on this.

driesvints:
Changing a method visibility like this can be a breaking change: https://3v4l.org/dT64C

Please send this to master.

Sophist-UK:
@eduance Brandon

You should be able to rebase your own branch against master and resubmit this PR. However I am not sanguine that it will get merged for v10 for the same backward-compatibility reasons as above.

Not: On reflection I think that the need for `publicIncrementQuietly()` and `publicDecrementQuietly()` in the unit tests is actually an indicator that these should be public - so maybe this is a v10 change which will be accepted.

eduance:
@Sophist-UK Appreciate it. Will indeed give it a try. :-)

PR #43222, opened 2022-07-16
Create a UUID based migration

TechyMS:


GrahamCampbell:
I find that the lack of ordering makes uuids pretty unsuitable as a primary key in practice. Consider using something like ksuids instead.

PR #43224, opened 2022-07-16
[9.x] Setup many-to-many relation via pivot model

l3aro:
This add ability to setup a `belongsToMany` relationship via `Pivot` model

## Problem

Currently, in order to define a `belongsToMany` relationship with custom table name and pivot keys, we need to delcare with primitive value with specific (and a bit confuse) order. For example:

```php
<?php

namespace Modules\Work\Models;

class Project extends Model
{
    protected $table = 'work__projects';

    protected $fillable = [];

    public function categories()
    {
        return $this->belongsToMany(ProjectCategory::class, 'work__project_category', 'project_id', 'category_id');
    }
}
```

```php
<?php

namespace Modules\Work\Models;

class ProjectCategory extends Model
{
    protected $table = 'work__project_categories';

    protected $fillable = [];

    public function projects()
    {
        return $this->belongsToMany(Project::class, 'work__project_category', 'category_id', 'project_id');
    }
}
```

I also notice that we have [Custom Intermediate model](https://laravel.com/docs/9.x/eloquent-relationships#defining-custom-intermediate-table-models) but since the setup relation process has no use again this `using` so we have no benefit from declare table name here!

## Concept

I would like to have nice setup with the name and the pivot keys are declared inside the pivot model. So we can setup the relation like below

```php
<?php

namespace Modules\Work\Models;

class Project extends Model
{
    protected $table = 'work__projects';

    protected $fillable = [];

    public function categories()
    {
        return $this->belongsToMany(ProjectCategory::class, ProjectCategoryPivot::class);
    }
}
```

```php
<?php

namespace Modules\Work\Models;

class ProjectCategory extends Model
{
    protected $table = 'work__project_categories';

    protected $fillable = [];

    public function projects()
    {
        return $this->belongsToMany(Project::class, ProjectCategoryPivot::class);
    }
}
```

```php
<?php

namespace Modules\Work\Models;

class ProjectCategoryPivot extends Pivot
{
    protected $table = 'work__project_category';

    protected $pivotKeys = [
        Project::class => 'project_id',
        ProjectCategory::class => 'category_id',
    ];
}
```

## Solution

My solution is in the setup relationship process `belongsToMany`@`Illuminate/Database/Eloquent/Concerns/HasRelationships.php`, when table name is instance of `Pivot` model, we will extract data defined and pass them to belongsToMany's constructor

## Issues

I'm not so good at english, so please look carefully at comment description and have them fixed if its not that clear

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
@@ -500,11 +500,38 @@ trait HasRelationships
             $table = $this->joiningTable($related, $instance);
         }
 
-        return $this->newBelongsToMany(
+        $pivotTable = $this->resolvePivotTable($table);
+        if ($pivotTable instanceof Pivot) {
+            $foreignPivotKey = $pivotTable->getPivotKeyFor(get_class($this)) ?: $foreignPivotKey;
+            $relatedPivotKey = $pivotTable->getPivotKeyFor($related) ?: $relatedPivotKey;
+            $table = $pivotTable->getTable();
+        }
+
+        $relationInstance = $this->newBelongsToMany(
             $instance->newQuery(), $this, $table, $foreignPivotKey,
             $relatedPivotKey, $parentKey ?: $this->getKeyName(),
             $relatedKey ?: $instance->getKeyName(), $relation
         );
+
+        if ($pivotTable instanceof Pivot) {
+            $relationInstance->using($pivotTable);
+        }
+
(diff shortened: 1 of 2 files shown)


driesvints:
I suggest to just create an actual model for the Pivot model.

l3aro:
@driesvints You mean using an actual model instead of pivot one, right? If so, then I dont see why it's better than have a pivot model with only extra property `pivotKeys`, then have benefit of `Custom intermediate model`

taylorotwell:

PR #43233, opened 2022-07-17
use arrow function

moharami:
In some cases where there is a variable by the "use" keyword, it would be replaced with fn.


GrahamCampbell:
This change is not necessary, and I don't feel like we should merge this class of PRs.x

PR #43238, opened 2022-07-17
[9.x] add whereLike

milwad-dev:


ankurk91:
How does it work in Postgres?

What about `whereiLike() `

milwad-dev:
@ankurk91 
I think they are the same

ankurk91:
@milwad-dev 

`like` and `ilike` are not the same for postgresql

milwad-dev:
@ankurk91 
Sorry, I had another thought.
We can make new method for ``ilike`` or make property for check like & ilike.

GrahamCampbell (changes requested):

taylorotwell:

PR #43169, opened 2022-07-13
[10.x] Adds enum casting to `Request`

emargareten:
This PR allows to conveniently retrieve an input from the Request as an enum.

```php
// Before
public function post(Request $request)
{
    $status = StatusEnum::tryFrom($request->input('status'));

    // do stuff with status enum...
}

// After
public function post(Request $request)
{
    $status = $request->enum('status', StatusEnum::class);

    // do stuff with status enum...
}
```

PR #43239, opened 2022-07-17
[9.x] Adds enum casting to Request

emargareten:
(same as https://github.com/laravel/framework/pull/43169 for 9.x) 

This PR allows to conveniently retrieve an input from the Request as an enum.

```php
// Before
public function post(Request $request)
{
    $status = StatusEnum::tryFrom($request->input('status'));

    // do stuff with status enum...
}

// After
public function post(Request $request)
{
    $status = $request->enum('status', StatusEnum::class);

    // do stuff with status enum...
}
```


supun-io:
I just [added](https://github.com/laravel/framework/pull/44295) a PR to improve type definitions of this.

By the way, I was thinking, shouldn't there be a way to get non-nullable Enum?

For example, if I validate the request input first as an enum, I just want an enum later.

```php
$request->validate(['type' => ['required', new Enum(MyEnum::class)]);

$type = $request->enum('type', MyEnum::class);
// currently this is MyEnum | null
```

Currently, `$type` can be null because `tryFrom` is used. But, as we have the validation earlier, we may just need the exact enum. It is annoying to deal with `null`.

I had two ideas.

1. Add a third-parameter to the `enum` method.

```
public function enum($key, $enumClass, $nullable = true) {}
```

Then, call it as `$request->enum($key, $enumClass, false)`. But, this is not very fluent.

2. Add a different method

As this is already deployed to public 9.0, we will need to reserve the `enum` name for nullable method. I am not sure about a name for the other method. `enumNotNullable` is too long.

Any ideas? @emargareten

PR #43251, opened 2022-07-18
[10.x] Make incrementQuietly and decrementQuietly public to enforce consistency with similar features #43220

eduance:
This PR enforces the consistency for the incrementQuietly and decrementQuietly to make the methods public. The reasoning behind this is that the usage for this method should be accessible from any model instead a developing having to worry about implementing the identical code himself.

**This could be a breaking change! But its a worthy one.**

**Why shouldn't you just do this at the subclass level?**

In my opinion, 

- to keep the code consistent with similar features like saveQuietly and updateQuietly which both also have a public access modifier. 
- a user model shouldn't have to implement the incrementQuietly() code on its own to call the parent method, this would violate the single responsibility principle.
- why repeat yourself if Laravel provides you with this identical logic? :-)

Thanks,

Eduance

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -929,7 +929,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
      * @param  array  $extra
      * @return int
      */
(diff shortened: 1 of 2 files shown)


eduance:
@driesvints We also have the `increment()` method, which is public on the Builder, but protected on the Model. If this is wished, we could process this change into the merge request as well. Let me know your take on this.

PR #43273, opened 2022-07-19
[9.x] Added support for custom proxy urls

gsguglielmo:
Implementation of this feature #43262 

This pull request adds the option to manually specify the base url of the vite development server, so it can be used behind a reverse proxy. 
If the config option `app.vite_hot_proxy_url` is present, all tags generated with the `@vite` directive, will have the specified value as it's base URL. 
Mix has a similar feature with `app.mix_hot_proxy_url`.

### DIFF
--- a/src/Illuminate/Foundation/Vite.php
+++ b/src/Illuminate/Foundation/Vite.php
@@ -27,6 +27,12 @@ class Vite
         if (is_file(public_path('/hot'))) {
             $url = rtrim(file_get_contents(public_path('/hot')));
 
+            $customUrl = app('config')->get('app.vite_hot_proxy_url');
+
+            if (! empty($customUrl)) {
+                $url = $customUrl;
+            }
+
             return new HtmlString(
                 $entrypoints
                     ->map(fn ($entrypoint) => $this->makeTag("{$url}/{$entrypoint}"))
@@ -90,6 +96,12 @@ class Vite
         }
 
         $url = rtrim(file_get_contents(public_path('/hot')));
+
+        $customUrl = app('config')->get('app.vite_hot_proxy_url');
+
+        if (! empty($customUrl)) {
+            $url = $customUrl;
+        }
 
         return new HtmlString(
             sprintf(


taylorotwell:
Will let @jessarcher review this - @jessarcher please mark as ready for review when you're ready for me to take a look

jessarcher:
Hi @gsguglielmo, thanks for the PR.

I think before we look at adding this customisation, it would be good to see if we can get your custom host into the hot file itself.

Vite allows you to configure where its client should look for the HMR server, and the Laravel Vite plugin also respects this since v0.3.

Try adding the following to your Vite config:

```js
export default defineConfig({
  // ...
  server: {
    hmr: {
      // protocol: 'wss', // Only if you need a https URL
      host: 'your-custom-host',
    },
  },
});
```

If you need to customise the port, you may specify the `VITE_PORT` environment variable.

gsguglielmo:
The host part does seem to solve part of the issue, the hostname part od the `hot` file is ok, but the port is not.

I've tried the `VITE_PORT` environment variable but does not seem to have an effect on the tag that is being rendered by the blade directive.
image 
I've tried to use the `server.port` and `server.hmr.port` but in both cases vite will try to bind the server to that port which if the dev server port and the final forwarded port don't match, it won't work. 

If you set `server.hmr.port` in the config Vite will also resolve the FQDN and will try to bind the server to that address too, in my case the server is running inside a docker container, so the ip associated to the FQDN doesn't match any ip that the sev server can bind to. 
image

For the port part of the `hot` file it will always be the port of the http server:
https://github.com/laravel/vite-plugin/blob/a766dd50825037b7914b66b1f0b6f29199a219ff/src/index.ts#L152-L158
https://github.com/laravel/vite-plugin/blob/a766dd50825037b7914b66b1f0b6f29199a219ff/src/index.ts#L340-L351

jessarcher:
Hi @gsguglielmo, I'd be open to updating `laravel-vite-plugin` to respect the port configured by the user.

There is already a PR at https://github.com/laravel/vite-plugin/pull/63 that I will look at merging.

timacdonald:
@gsguglielmo because the PR @jessarcher mentioned has been merged, does that solve your issue and should we close this one?

jessarcher:
Going to close this as https://github.com/laravel/vite-plugin/pull/63 is merged and should hopefully address the issue.

PR #43277, opened 2022-07-19
[9.x] Add increment() and decrement() Arr helpers

gbradley:
This pull request makes it easy to increment or decrement a value inside a nested array. While not hard to write manually, this PR makes it a one-liner.

The methods both accept the array, a key in dot notation, and optionally the amount to increment/decrement by (the default is 1). If the key does not exist, it will be created and set to the specified value.

    $array = [
        'nested' => [
            'key' => [1, 2, 3]
        ]
    ];

    Arr::increment($array, 'nested.key.1', 10) // 12
    Arr::decrement($array, 'missing', 10) // -10


taylorotwell:

Issue #43274, opened 2022-07-19
EloquentUserProvider RetrieveById & RetrieveByToken - Ambiguous ID


BenWalters:
- Laravel Version: 9.11.0
- PHP Version: 8.1.8
- Database Driver & Version: MySQL - 8.0.26

### Description:
SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous
Caused by Illuminate\Auth\EloquentUserProvider:52 (retrieveById)

When applying a global scope to our Users model that also has an ID field, the above mentioned function is not qualifying the field name (adding the table prefix), thus causing the "'id' in where clause is ambiguous" error.

Seemingly the same issue was raised here - https://github.com/laravel/ideas/issues/1956
I believe that wrapping `$model->getAuthIdentifierName()` in `$model->qualifyColumn()` would resolve the issue and should be implemented in both the `retrieveById` and `retrieveByToken` functions of EloquentUserProvider.

driesvints:
Let's see how your PR goes 👍

PR #43278, opened 2022-07-19
[8.x] Protect against ambiguous columns

BenWalters:
Resolving https://github.com/laravel/framework/issues/43274

When applying a global scope to a User model that joins a secondary table, which also has an ID field, the `retrieveById` and `retrieveByToken` functions of Illuminate\Auth\EloquentUserProvider do not qualify the identifier field name in the where clause which. This subsequently causes a "Integrity constraint violation: 1052 Column 'id' in where clause is ambiguous" error.

Wrapping `$model->getAuthIdentifierName()` in `$model->qualifyColumn()` on both methods protects against this. 
This is a non-breaking change and is in-line with other abstract queries/builders. Tests updated accordingly.

Example global scope:
```
class ActiveUsersScope extends AbstractGlobalScope implements Scope
{
    /**
     * Apply the scope to a given Eloquent query builder.
     *
     * @param Builder $builder
     * @param Model $model
     * @return void
     */
    public function apply(Builder $builder, Model $model): void
    {
        $userOrganizationModel = new UserOrganization();

        $builder
            ->join(
                $userOrganizationModel->getTable(),
                $userOrganizationModel->qualifyColumn('user_id'),
                '=',
                $model->qualifyColumn('id')
            )
            ->where($userOrganizationModel->qualifyColumn('status'), UserStatus::ACTIVE);
    }
}
```

### DIFF
--- a/src/Illuminate/Auth/EloquentUserProvider.php
+++ b/src/Illuminate/Auth/EloquentUserProvider.php
@@ -49,7 +49,7 @@ class EloquentUserProvider implements UserProvider
         $model = $this->createModel();
 
         return $this->newModelQuery($model)
-                    ->where($model->getAuthIdentifierName(), $identifier)
+                    ->where($model->qualifyColumn($model->getAuthIdentifierName()), $identifier)
                     ->first();
     }
 
@@ -65,7 +65,7 @@ class EloquentUserProvider implements UserProvider
         $model = $this->createModel();
 
         $retrievedModel = $this->newModelQuery($model)->where(
-            $model->getAuthIdentifierName(), $identifier
+            $model->qualifyColumn($model->getAuthIdentifierName()), $identifier
         )->first();
 
         if (! $retrievedModel) {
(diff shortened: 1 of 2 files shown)


driesvints:
@BenWalters please add a thorough explanation to your main PR description (see the PR template) so Taylor knows what this is about. Thanks.

BenWalters:
@driesvints updated! Let me know if anything else is needed.

driesvints:
@BenWalters maybe also share an example of the global scope

BenWalters:
@driesvints it's a WIP but I've added an example.

driesvints:
This looks like a breaking change so we're reverting this: https://github.com/laravel/framework/issues/43353

BenWalters:
@driesvints  no problem, I would however say that this is 'breaking' because of a 3rd party packages bad implementation. The Mongo DB package does not prefix fields with table names. They have only overridden some of the qualify column methods that are implemented on the abstract Model class that Laravel Framework provides.

BenWalters:
@driesvints  sorry, I'd not seen the octane issue also. I cannot comment on that one.

BenWalters:
Having just looked at it though it's also because MongoDB is being used.

PR #43291, opened 2022-07-19
[9.x] Add syntactic sugar travelIntoThePast() method

gjinali:
This adds syntactic sugar to `InteractsWithTime` trait. Basically it's very similar to `travel()` method but instead, it will pass the `$value` argument as a negative integer to the `Wormhole:class`, thus it'll travel to the past.

Before:
```
$this->travel(-2)->days();
```

Now:
```
$this->travelIntoThePast(2)->days();
```


taylorotwell:

PR #43297, opened 2022-07-19
better style event:list

milwad-dev:
This help to better style event:list command


jbrooksuk:
Please can you include a screenshot of the difference here.

milwad-dev:
if instead of INFO can write Event or Listener i think is very nice :) 

Before


After

PR #43305, opened 2022-07-20
[9.x] Restore job id in worker command logs

L3o-pold:
This PR restore the job id in worker command log that was removed in https://github.com/laravel/framework/pull/43065.
I think it's important to display the job id for better debugging. 

Previous logs looks like:

```
2022-07-20 07:40:51 App\Jobs\Mobile\Posted .................... 22.32ms DONE
```

New log looks like:

```
2022-07-20 08:12:57 ea280c14-90ee-4ec0-ba1f-b7237291371f App\Jobs\SetUserAway  25.57ms DONE
```

### DIFF
--- a/src/Illuminate/Queue/Console/WorkCommand.php
+++ b/src/Illuminate/Queue/Console/WorkCommand.php
@@ -214,17 +214,17 @@ class WorkCommand extends Command
 
             $formattedStartedAt = Carbon::now()->format('Y-m-d H:i:s');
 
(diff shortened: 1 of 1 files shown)


timrspratt:
Maybe this extra info should only be outputted if the verbose level was passed?

Alternative might be to do what GitHub do with commit hashes, and only show the first part - the whole thing is long and does someone ruin the cleanliness of the new layout imo.

taylorotwell:
Yeah this feels like something for verbose mode? Will mark as draft for now.

PR #43309, opened 2022-07-20
for php 8.0.x support

EgorGruzdev:
fix "Cannot unpack Traversable with string keys" on PHP 8.0.x


driesvints:
There's already a PR open. Thanks

PR #43317, opened 2022-07-20
[9.x] Add supportsTemporaryUrl to FilesystemAdapter

Jasonej:
This PR adds a `supportsTemporaryUrl()` method to `FilesystemAdapter` that checks if the adapter has `getTemporaryUrl` or a `temporaryUrlCallback`.

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemAdapter.php
+++ b/src/Illuminate/Filesystem/FilesystemAdapter.php
@@ -671,6 +671,17 @@ class FilesystemAdapter implements CloudFilesystemContract
     }
 
     /**
+     * Check if temporary urls are supported.
+     *
+     * @return bool
+     */
+    public function supportsTemporaryUrl(): bool
+    {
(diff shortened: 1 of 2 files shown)


taylorotwell:
What is this fixing or doing?

Jasonej:
Hi, it's seeking to allow the following in situations where temp urls are preferred when supported, but are not required.
```php
$filesystem = Storage::disk($disk);

$url = $filesystem->supportsTemporaryUrl()
    ? $filesystem->temporaryUrl($path, $expiration)
    : $filesystem->url($path);
```

Currently, the temporaryUrlCallback is not exposed as it is protected, so to accomplish this behavior the options are to assume that no callback exists and rely on a `method_exists()` check or do a `try/catch` that checks for the RuntimeException that is thrown by `temporaryUrl()`.

driesvints:
Renamed to `ProvidesTemporaryUrls` btw.

PR #43321, opened 2022-07-20
[9.x] Pretty output for missing command with suggestions

slavarazum:
Continuing to improve the Artisan output style 💅

Before:
Before

After:
After


nunomaduro:
What do you think about this one? https://github.com/laravel/framework/pull/43323. I've added you as co-author.

slavarazum:
Agreed, I considered this option, looks much cleaner. Also found a better command handling without alternatives.
Thanks for the mention 🤜🤛

PR #43329, opened 2022-07-21
[9.x] Allow section payload to be lazy in the "about" command

timacdonald:
With the [introduction of the new fancy (💅) artisan about command](https://github.com/laravel/framework/pull/43147), all data registered was done able to be done in an eager manner:

```php
AboutCommand::add('Settings', [
    'value' => SettingModel::first()->value,
    'App Balance' => BalanceService::fetch(),
    'Computed Metric' => 6 ÷ 2,
]);
```

A [follow up PR](https://github.com/laravel/framework/pull/43225) was submitted to allow individual items to be lazily evaluated:

```php
AboutCommand::add('Settings', [
    'value' => fn () => SettingModel::first()->value,
    'App Balance' => fn () => BalanceService::fetch(),
    'Computed Metric' => fn () => 6 ÷ 2,
]);
```

This PR pushes this lazy feature further by allowing the entire data payload lazy, rather than individual entries:

```php
AboutCommand::add('Settings', fn () => [
    'value' => SettingModel::first()->value,
    'App Balance' => BalanceService::fetch(),
    'Computed Metric' => 6 ÷ 2,
    'version' => app(Package::class)->version(),
]);
```

This will improve DX (not having to define a new closure on each line) and also improve boot performance / memory usage.

Not only does it allow a Closure to be passed, as seen above, but it will also allow invokable classes. This allows packages and apps to specify their "about" data in a dedicated class, which also has the benefits of constructor and method injection (which is possible in the service provider, but feels more noisy).

```php

namespace Acme;

class About
{
    public function __invoke(Package $package)
    {
        return [
            'value' => SettingModel::first()->value,
            'App Balance' => BalanceService::fetch(),
            'Computed Metric' => 6 ÷ 2,
            'version' => $package->version(),
        ];
    }
}

// App service provider...

AboutCommand::add('Acme', About::class);
```

Although it has not actual benefit to the framework, I've modified the built in data registration to demonstrate lazy loading. When looking on how to do things, a lot of developers look at how Laravel does things under the hood. As I believe this should be the Laravel prescribed way of registering data to guide developers in the best direction for performant application booting - I thought it also made sense to make our data registration model this way of doing things.

I also believe the lazy evaluation should be the way we document things in the Laravel docs by default to promote this practice. If we don't want to do this, we should probably recommend specifying things in the boot function rather than the register function, otherwise if people try to pull things out of the cache or something they might run into trouble if things haven't finished registering into the container.

## Example...

```php
<?php

namespace App\Providers;

use Illuminate\Foundation\Application;
use Illuminate\Foundation\Console\AboutCommand;

{
    public function register()
    {
        //
    }

    public function boot()
    {
        AboutCommand::add('Eager', ['eager' => 'load']);
        AboutCommand::add('Lazy', fn () => ['lazy' => 'load']);
        AboutCommand::add('lazy with class', About::class);
    }
}

class About
{
    public function __invoke(Application $app)
    {
        return [
            'version' => $app->version(),
        ];
    }
}
```

Screen Shot 2022-07-21 at 11 51 21 am

### DIFF
--- a/src/Illuminate/Foundation/Console/AboutCommand.php
+++ b/src/Illuminate/Foundation/Console/AboutCommand.php
@@ -87,7 +87,7 @@ class AboutCommand extends Command
                         ->map(fn ($value, $key) => [$key, $value])
                         ->values()
                         ->all();
-                })->flatten(1)->all()
+                })->flatten(1)
             )
             ->sortBy(function ($data, $key) {
                 $index = array_search($key, ['Environment', 'Cache', 'Drivers']);
@@ -132,13 +132,11 @@ class AboutCommand extends Command
 
             $this->components->twoColumnDetail('  <fg=green;options=bold>'.$section.'</>');
 
-            sort($data);
-
-            foreach ($data as $detail) {
+            $data->sort()->each(function ($detail) {
                 [$label, $value] = $detail;
 
                 $this->components->twoColumnDetail($label, value($value));
-            }
+            });
         });
     }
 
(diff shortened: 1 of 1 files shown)


driesvints:
I fixed the broken build on 9.x. Please rebase your branch and mark as ready when done, thanks!

timacdonald:
Thanks for fix that @driesvints. Rebased.

jbrooksuk (approved):

PR #43330, opened 2022-07-21
Using parent::getActualClassNameForMorph() instead of Model::getActualClassNameForMorph()

huangdijia:
```php
class User extends Model
{
    public function foo()
    {
        return $this->morphOne(Foo::class);
    }

    public function getMorphClass()
    {
        return 'user';
    }
}

class Post extends Model
{
    public function foo()
    {
        return $this->morphOne(Foo::class);
    }

    public function getMorphClass()
    {
        return 'post';
    }
}

class Foo extends Model
{
    public function morphable()
    {
        return $this->morphTo();
    }

    public static function getActualClassNameForMorph($class)
    {
        return [
            'user' => User::class,
            'post' => Post::class,
        ][$class]??$class;
    }
}
```

Fixed the error when using like:

```php
Foo::with('morphable')->get();
Foo::whereHasMorph('morphable', ['*'])->get();
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -231,7 +231,7 @@ trait QueriesRelationships
         }
 
         foreach ($types as &$type) {
-            $type = Relation::getMorphedModel($type) ?? $type;
+            $type = $this->model::getActualClassNameForMorph($type) ?? Relation::getMorphedModel($type) ?? $type;
         }
 
         return $this->where(function ($query) use ($relation, $callback, $operator, $count, $types) {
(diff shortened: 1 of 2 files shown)


driesvints:
https://github.com/laravel/framework/commit/905ae8aa0d53f367257353e89e9f26c777acdf06

taylorotwell:
This will need tests - just glancing at it looks like it will break applications because getActualClassNameForMorph **always** returns a value. The other conditions of your ternary can never be hit.

huangdijia:
```php

public static function getActualClassNameForMorph($class)
    {
        return [
            'user' => User::class,
            'post' => Post::class,
        ][$class]??$class;
    }

```

This is just a demo, just for overloading `Model:: getActualClassNameForMorph()`.

Because it strongly binds the global Relation.

huangdijia:
```php
    /**
     * Get the model associated with a custom polymorphic type.
     *
     * @param  string  $alias
     * @return string|null
     */
    public static function getMorphedModel($alias)
    {
        return static::$morphMap[$alias] ?? null;
    }
```

It's same as:

```php
    /**
     * Retrieve the actual class name for a given morph class.
     *
     * @param  string  $class
     * @return string
     */
    public static function getActualClassNameForMorph($class)
    {
        return Arr::get(Relation::morphMap() ?: [], $class, $class);
    }
```

PR #43331, opened 2022-07-21
[9.x] Improve test for intersectByKeys in collection

moharami:
if there is duplicate key in the given array is present, the item in the original collection will be return


driesvints:
I fixed the broken build on 9.x. Please rebase your branch and mark as ready when done, thanks!

PR #43335, opened 2022-07-21
[9.x] Fixed ```Illuminate\Support\Facades\Http``` phpdoc

bvtterfly:
Add  missing ```resource``` type for ```attach``` method on ```$contents``` parameter  in the phpdoc.


driesvints:
Thanks!

PR #43340, opened 2022-07-21
[8.x] Preventing ambiguous columns when using relational pivot tables.

BenWalters:
When setting Pivot keys, they are currently set without qualifying the field names. In turn, this means there is an opportunity to cause an ambiguous field error. 
This change is making use of pre-existing functions to qualify the pivot keys and is non-breaking.


driesvints:
This seems to break the test suite.

BenWalters:
@driesvints Sorry the only reason I put the PR in is because I'm unable to successfully run all tests locally (without any changes). Are there any docs on how to properly configure your dev environment to run the tests?

driesvints:
@BenWalters you can use GitHub Actions on your fork to verify if the tests are passing.

PR #43343, opened 2022-07-21
Using parent::getActualClassNameForMorph() instead of Model::getActualClassNameForMorph() 

huangdijia:
#43330 43330

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -231,7 +231,7 @@ trait QueriesRelationships
         }
 
         foreach ($types as &$type) {
-            $type = $this->model::getActualClassNameForMorph($type) ?? Relation::getMorphedModel($type) ?? $type;
+            $type = $this->model::getActualClassNameForMorph($type) ?? $type;
         }
 
         return $this->where(function ($query) use ($relation, $callback, $operator, $count, $types) {
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -231,7 +231,7 @@ trait QueriesRelationships
         }
 
         foreach ($types as &$type) {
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
```diff
@@ -231,7 +231,7 @@ public function hasMorph($relation, $types, $operator = '>=', $count = 1, $boole
         }
 
         foreach ($types as &$type) {
-            $type = Relation::getMorphedModel($type) ?? $type;
+            $type = $this->model::getActualClassNameForMorph($type) ?? $type;
```
huangdijia, line 234: If the subclass does not define `getActualClassNameForMorph()`, it will still call `Model::getActualClassNameForMorph()`, Doing this allows subclasses to override `getActualClassNameForMorph()`, avoiding the dependency on `Relation::$morphMap`

huangdijia (commented):
`Relation::$morphMap` is global, Definitions can be made private by overloading `getActualClassNameForMorph() `, won't break the application.

src/Illuminate/Database/Eloquent/Relations/MorphTo.php
```diff
@@ -181,7 +181,7 @@ protected function gatherKeysByType($type, $keyType)
      */
     public function createModelByType($type)
     {
-        $class = Model::getActualClassNameForMorph($type);
+        $class = $this->parent::getActualClassNameForMorph($type);
```
huangdijia, line 184: If the subclass does not define `getActualClassNameForMorph()`, it will still call `Model::getActualClassNameForMorph()`, Doing this allows subclasses to override `getActualClassNameForMorph()`, avoiding the dependency on `Relation::$morphMap`

taylorotwell:
The code for `getActualClassNameForMorph`

CleanShot 2022-07-21 at 09 42 42@2x

How will your code ever hit the right side of your `??`?

taylorotwell:
You're not testing your changes.

huangdijia:
`$type = Relation::getMorphedModel($type) ?? $type;`  to  `$type = $this->model::getActualClassNameForMorph($type);`

PR #43344, opened 2022-07-21
[8.x] Preventing ambiguous columns when generating delete queries.

BenWalters:
When attempting to detach a record from a model which has a global scope joining another table, the delete query fails to qualify the field names in the where clause. In turn, this means there is an opportunity to cause an ambiguous field error.
This change is making use of pre-existing functions to qualify the pivot keys and is non-breaking, test updated accordingly.

### Example Scenario:
**User**
```php
public function entities(): MorphToMany
    {
        return $this
            ->morphedByMany(Entity::class, 'group', 'user_groups')
            ->using(UserGroup::class)
            ->withTimestamps();
    }
```

**UserGroupScope**
```php
    public function apply(Builder $builder, Model $model): void
    {
        $userEntityModel = new UserEntity();

        $builder
            ->join(
                $userEntityModel->getTable(),
                fn(JoinClause $join) => $join
                    ->on(
                        $userEntityModel->qualifyColumn('user_id'),
                        '=',
                        $model->qualifyColumn('user_id')
                    )
                    ->on(
                        $userEntityModel->qualifyColumn('entity_id'),
                        '=',
                        $model->qualifyColumn('entity_id')
                    )
            )
            ->where($userEntityModel->qualifyColumn('status'), UserStatus::ACTIVE);
    }
```

**Function Call**
```php
$user->entities()->detach({{ID}})
```

This would result in an error similar to the following:
`SQLSTATE[23000]: Integrity constraint violation: 1052 Column 'user_id' in where clause is ambiguous`
```sql
delete `user_groups` 
from `user_groups` 
inner join `user_entity` 
on `user_entity`.`user_id` = `user_groups`.`user_id` 
and `user_entity`.`entity_id` = `user_groups`.`entity_id` 
where (`user_id` = 2659 and `group_id` = 2) 
and `group_type` = ENTITY 
and `user_entity`.`status` = 1
```

This change will change the query to:
```sql
delete `user_groups` 
from `user_groups` 
inner join `user_entity` 
on `user_entity`.`user_id` = `user_groups`.`user_id` 
and `user_entity`.`entity_id` = `user_groups`.`entity_id` 
where (`user_groups`.`user_id` = 2659 and `user_groups`.`group_id` = 2) 
and `group_type` = ENTITY
and `user_entity`.`status` = 1
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php
@@ -146,8 +146,14 @@ trait AsPivot
     protected function getDeleteQuery()
     {
         return $this->newQueryWithoutRelationships()->where([
-            $this->foreignKey => $this->getOriginal($this->foreignKey, $this->getAttribute($this->foreignKey)),
-            $this->relatedKey => $this->getOriginal($this->relatedKey, $this->getAttribute($this->relatedKey)),
+            $this->qualifyColumn($this->foreignKey) => $this->getOriginal(
+                $this->foreignKey,
+                $this->getAttribute($this->foreignKey)
+            ),
+            $this->qualifyColumn($this->relatedKey) => $this->getOriginal(
+                $this->relatedKey,
+                $this->getAttribute($this->relatedKey)
+            ),
         ]);
     }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Honestly I think creating a global scope that joins another table is just a bad idea and I don't recommend it. It's going to lead to all sorts of problems like this.

PR #43347, opened 2022-07-21
Add `table` option to `make:model` command

azizramdan:
Sometimes we create models that do not meet the naming conventions in laravel, this is often encountered when migrating projects to laravel and using existing databases, or on projects that are not based on English.

Currently we have to create a model and then add the `$table` property manually. By adding the `table` option then we can use the following command `artisan make:model Foo --table=bar`, this will create a model with the name Foo associated with the table `bar`. It will also create a table with the name `bar` if we use the `migration` option.


taylorotwell:

PR #43098, opened 2022-07-08
Created vite() helper to replicate functionality of mix() helper

chrispage1:
For my project, I was using the `mix()` helper function in various areas to load the absolute path to compiled mix assets.

With the addition of Vite (which is awesome!), there is no such helper to retrieve absolute file paths, only strings bundled with style & script tags.

This helper allows a user to retrieve an absolute path to a Vite asset, either in hot reload or built asset mode.

Example:

`@vite('resources/css/app.scss')` results in a compiled HTML tag - 

```
<link rel="stylesheet" href="http://localhost/build/assets/app.3c62ae02.css" />
```

whereas `{{ vite('resources/css/app.scss') }}` results in an absolute file path - 

```
http://localhost/build/assets/app.3c62ae02.css
```

I hope you'll see this as a useful contribution to Laravel! I'd be happy to write up some documentation on this addition.

PR #43348, opened 2022-07-21
[9.x] Add missing vite() helper

Okipa:
Hello there,

Here is my proposal to add a `vite()` helper that mimic the `mix()` helper behavior.

We sometimes need to get a versioned file path outside of a Blade view and it feels that it missing from the framework right now, as we currently only have a Blade directive to use it.

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -15,6 +15,7 @@ use Illuminate\Contracts\View\Factory as ViewFactory;
 use Illuminate\Foundation\Bus\PendingClosureDispatch;
 use Illuminate\Foundation\Bus\PendingDispatch;
 use Illuminate\Foundation\Mix;
+use Illuminate\Foundation\Vite;
 use Illuminate\Http\Exceptions\HttpResponseException;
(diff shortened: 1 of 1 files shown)


jessarcher:
Hi @Okipa, I don't think this will do what you're expecting.

The `Vite` class returns full `<script>` and `<link rel="stylesheet">` HTML tags, unlike the `Mix` class which just returned a path. This is because a single JS entry point may also contain CSS files that need to be loaded, in addition to the Vite HMR client itself.

You would need to add a new method to the `Vite` class (perhaps called `asset`) that only returned the resolved path for the given asset, which I'd be open to, especially if the other code in the class was able to be refactored to also use that method.

Another issue is that this function would create a conflict with this popular third-party package https://laravel-vite.dev/guide/features/helpers.html#vite by @innocenzi. Would the `Vite` class method be enough without the helper?

Marking this as draft, but feel free to close if you don't wish to pursue it.

ankurk91:
https://github.com/laravel/framework/pull/43098
https://github.com/motomedialab/laravel-vite-helper

driesvints:
Going to close this already. See the above.

PR #43352, opened 2022-07-21
add is_nulls helper

milwad-dev:
This helper help to check multiple column is null.
For example : 
Imagine you have some data and you are going to check if each data is null or not null. Your code like this:

carbon

But with this helper your code like this:

carbon (1)


taylorotwell:

dennisprudlo:
@milwad-dev You can use multiple parameters in the `isset` function.
```php
if (! isset($item1, $item2, $item3, $item4, $item5)) {
    echo 'test';
}
```

PR #43360, opened 2022-07-22
[9.x] Fixes default attribute value when using enums on `model:show`

josecl:
## Problem

Having an `UnitEnum` or `BackedEnum` as default attribute value in a Model, like this:

```php
    protected $casts = [
        'status' => InvoiceStatus::class,
    ];

    protected $attributes = [
        'status' => InvoiceStatus::I_LOVE_LARAVEL,
    ];
```

Triggers an error on `model:show` command:

`ERROR: Object of class App\Enums\InvoiceStatus could not be converted to string`

## Fix

This PR fixes it displaying the value of enums `name` attribute:

```bash
 status fillable ................................... integer / App\Enums\InvoiceStatus
  ⇂ default: I_LOVE_LARAVEL
```

### DIFF
--- a/src/Illuminate/Foundation/Console/ShowModelCommand.php
+++ b/src/Illuminate/Foundation/Console/ShowModelCommand.php
@@ -330,7 +330,7 @@ class ShowModelCommand extends Command
 
             if ($attribute['default'] !== null) {
                 $this->components->bulletList(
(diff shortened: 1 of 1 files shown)


jessarcher:
Hey @josecl,

Thanks for this! I think the error will still occur with the `--json` flag, so it might be better to move this change to the `getColumnDefault()` method to fix it everywhere. I'll work up a PR now.

PR #43365, opened 2022-07-22
[8.x] Fix DynamoDB locks with 0 seconds duration

### LINKED ISSUE #43280: ShouldBeUnique with DynamoDB provider doesn't work as expected with default configuration
- Laravel Version: v9.19.0
- PHP Version: 8.1.6

### Description:

According to [documentation](https://laravel.com/docs/9.x/queues#unique-jobs), implementing ShouldBeUnique on job class is enough to ensure only one instance of given job is present in the queue. However, using `dynamodb` cache driver without setting `$uniqueFor` explicitly makes the lock item expire after 1 second, because `expires_at` is set to current timestamp when `$uniqueFor` is 0 (default value).

This is the change event in DynamoDB table after `putItem` request when trying to acquire job lock:

```
{
  "awsRegion": "eu-central-1",
  "eventID": "4cb2d5b3-4954-451e-8332-01671dd2bf3e",
  "eventName": "INSERT",
  "userIdentity": null,
  "recordFormat": "application/json",
  "tableName": "vapor_cache",
  "dynamodb": {
    "ApproximateCreationDateTime": 1658222276198,
    "Keys": {
      "key": {
        "S": "cloud-api-production:cloud-api-production:laravel_unique_job:App\\Jobs\\<job-name>"
      }
    },
    "NewImage": {
      "value": {
        "S": "s:16:\"D7X1915ojxbQW6sU\";"
      },
      "key": {
        "S": "cloud-api-production:cloud-api-production:laravel_unique_job:App\\Jobs\\<job-name>"
      },
      "expires_at": {
        "N": "1658222276"
      }
    },
    "SizeBytes": 237
  },
  "eventSource": "aws:dynamodb"
}
```

`expires_at` is equal to current timestamp (see `ApproximateCreationDateTime`), so trying to acquire the lock in the following second will work because of used condition:

https://github.com/laravel/framework/blob/d425952e8cc664b01de8bd654336c1be8dcf7444/src/Illuminate/Cache/DynamoDbStore.php#L273

It seems like the behavior is not consistent across other cache providers, e.g. `database` cache provider uses lock expiration time of 1 day if no time was provided:

https://github.com/laravel/framework/blob/d425952e8cc664b01de8bd654336c1be8dcf7444/src/Illuminate/Cache/DatabaseLock.php#L92-L95

`redis` provider creates non-expiring item:

https://github.com/laravel/framework/blob/d425952e8cc664b01de8bd654336c1be8dcf7444/src/Illuminate/Cache/RedisLock.php#L35-L42

`memcached` items doesn't expire with expiration time set to 0.

While `dynamodb` items expire in the next second after the lock is acquired.

Possible solution would be to use 1 day as default expiration value, as in `database` cache provider case.

### Steps To Reproduce:

1. Use `dynamodb` cache provider
2. Create a job implementing `ShouldBeUnique` that takes longer than a few seconds to complete
3. Try to dispatch the job two times, more than 1 second apart
4. The job will be executed concurrently despite implementing `ShouldBeUnique`

Repository demonstrating the issue: https://github.com/Wowu/laravel-dynamodb-bug-report

lukaszmtw:
I've added a default lock duration of 1 day for DynamoDB locks.

This makes the behavior consistent with `DatabaseLock`:
https://github.com/laravel/framework/blob/d425952e8cc664b01de8bd654336c1be8dcf7444/src/Illuminate/Cache/DatabaseLock.php#L92-L95

Fix #43280

### DIFF
--- a/src/Illuminate/Cache/DynamoDbLock.php
+++ b/src/Illuminate/Cache/DynamoDbLock.php
@@ -34,9 +34,12 @@ class DynamoDbLock extends Lock
      */
     public function acquire()
     {
-        return $this->dynamo->add(
-            $this->name, $this->owner, $this->seconds
-        );
+        if ($this->seconds > 0) {
+            return $this->dynamo->add($this->name, $this->owner, $this->seconds);
+        } else {
+            // default to 1 day
+            return $this->dynamo->add($this->name, $this->owner, 86400);
+        }
     }
 
     /**
--- a/src/Illuminate/Cache/DynamoDbLock.php
+++ b/src/Illuminate/Cache/DynamoDbLock.php
@@ -37,7 +37,6 @@ class DynamoDbLock extends Lock
         if ($this->seconds > 0) {
             return $this->dynamo->add($this->name, $this->owner, $this->seconds);
         } else {
-            // default to 1 day
             return $this->dynamo->add($this->name, $this->owner, 86400);
         }
     }


driesvints:
Thanks @lukaszmtw

PR #43367, opened 2022-07-22
[9.x] New `db:show`, `db:table` and `db:monitor` commands

jbrooksuk:
This PR adds three new Artisan commands which provide information about the database directly from a Laravel project.

In summary, it exposes information such as:

- A list of all tables with row count and table size
- Overall size of the database
- Current number of database connections
- Column types
- List of indexes
- List of foreign keys and table references
- Ability to monitor the database and listen for events when the database is under load

---

### `db:show`

This command gives you an overview of the database as a whole. Out of the box, it will utilise your default connection, but you may optionally choose the connection you wish to inspect.

Upon running the command, you will see a summary of the database including database type, connection details, number of connections, total rows and overall size.

You will also see a breakdown of tables including table size and optionally (using the `--with-counts` option), the number of rows. This is an option as it can be an expensive operation on larger databases.

Additionally, using the `--with-views` option, you may also see a summary of the database views.

The output for this command is also available in JSON using the `--json` option.

<details>
  <summary><code>Example JSON</code></summary>

  ```json
  {
    "platform": {
      "config": {
        "driver": "mysql",
        "url": null,
        "host": "127.0.0.1",
        "port": "3306",
        "database": "vapor_app",
        "username": "root",
        "unix_socket": "",
        "charset": "utf8mb4",
        "collation": "utf8mb4_unicode_ci",
        "prefix": "",
        "prefix_indexes": true,
        "strict": true,
        "engine": null,
        "options": []
      },
      "name": "MySQL 8",
      "open_connections": 2
    },
    "tables": [
      {
        "table": "failed_jobs",
        "size": 16384,
        "rows": null,
        "engine": "InnoDB",
        "comment": ""
      },
      {
        "table": "migrations",
        "size": 16384,
        "rows": null,
        "engine": "InnoDB",
        "comment": ""
      },
      {
        "table": "password_resets",
        "size": 16384,
        "rows": null,
        "engine": "InnoDB",
        "comment": ""
      },
      {
        "table": "personal_access_tokens",
        "size": 16384,
        "rows": null,
        "engine": "InnoDB",
        "comment": ""
      },
      {
        "table": "posts",
        "size": 32768,
        "rows": null,
        "engine": "InnoDB",
        "comment": ""
      },
      {
        "table": "users",
        "size": 16384,
        "rows": null,
        "engine": "InnoDB",
        "comment": ""
      }
    ]
  }
  ```
</details>

This command is really useful to get a better understanding of the health of your database.

Example db:show command

---

### `db:table`

Where `db:show` allows you to see the state of a database as a whole, `db:table` allows you to see the state of an individual table by passing the table name as an argument or selecting the table from the list of tables provided.

As with `db:show`, the command will use your default database connection, but you may pass the connection you wish to use.

This command provides a summary of the table including the size and number of rows as well as a breakdown of every column along with its attributes and data type.

Additionally, it presents an overview of all indexes and foreign keys including the columns and tables referenced. It’s a great way to get a better understanding of a given table and how it relates to others.

The output for this command is also available in JSON using the `--json` option.

<details>
  <summary><code>Example JSON</code></summary>

  ```json
  {
    "table": {
      "name": "posts",
      "columns": 4,
      "size": 32768
    },
    "columns": {
      "id": {
        "column": "id",
        "attributes": {
          "0": "autoincrement",
          "type": "bigint",
          "1": "unsigned"
        },
        "default": null,
        "type": "bigint"
      },
      "user_id": {
        "column": "user_id",
        "attributes": {
          "type": "bigint",
          "1": "unsigned"
        },
        "default": null,
        "type": "bigint"
      },
      "created_at": {
        "column": "created_at",
        "attributes": {
          "type": "datetime",
          "2": "nullable"
        },
        "default": null,
        "type": "datetime"
      },
      "updated_at": {
        "column": "updated_at",
        "attributes": {
          "type": "datetime",
          "2": "nullable"
        },
        "default": null,
        "type": "datetime"
      }
    },
    "indexes": {
      "posts_user_id_foreign": {
        "name": "posts_user_id_foreign",
        "columns": [
          "user_id"
        ],
        "attributes": []
      },
      "primary": {
        "name": "PRIMARY",
        "columns": [
          "id"
        ],
        "attributes": [
          "unique",
          "primary"
        ]
      }
    },
    "foreign_keys": {
      "posts_user_id_foreign": {
        "name": "posts_user_id_foreign",
        "local_table": "posts",
        "local_columns": [
          "user_id"
        ],
        "foreign_table": "users",
        "foreign_columns": [
          "id"
        ],
        "on_update": "no action",
        "on_delete": "cascade"
      }
    }
  }
  ```
</details>

Example db:table command

---

### `db:monitor` 

Much like the `queue:monitor` command, `db:monitor` allows you to quickly see the number of connections your database is handling, which is a great indicator of the load it is currently under.

In addition, you may pass the `--max` option which will dispatch a `DatabaseBusy` event when the number of connections is ≥ the max provided. If no `--max` value is provided, no events will be dispatched. Your application may listen for this event in order to, for instance, send an email to the operations team notifying them the database is under strain.

Typically, this would be run as a scheduled command as often as you would like the check to be carried out.

Example db:monitor command

### DIFF
--- a/src/Illuminate/Database/Console/AbstractDatabaseCommand.php
+++ b/src/Illuminate/Database/Console/AbstractDatabaseCommand.php
@@ -9,12 +9,33 @@ use Illuminate\Database\MySqlConnection;
 use Illuminate\Database\PostgresConnection;
 use Illuminate\Database\SQLiteConnection;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Composer;
 use Symfony\Component\Process\Exception\ProcessSignaledException;
 use Symfony\Component\Process\Exception\RuntimeException;
 use Symfony\Component\Process\Process;
 
 abstract class AbstractDatabaseCommand extends Command
 {
+    /**
+     * The Composer instance.
+     *
+     * @var \Illuminate\Support\Composer
+     */
+    protected $composer;
+
+    /**
+     * Create a new command instance.
+     *
+     * @param  \Illuminate\Support\Composer  $composer
+     * @return void
+     */
+    public function __construct(Composer $composer)
+    {
+        parent::__construct();
+
+        $this->composer = $composer;
+    }
+
     /**
      * Get a human-readable platform name.
      *
--- a/src/Illuminate/Database/Console/AbstractDatabaseCommand.php
+++ b/src/Illuminate/Database/Console/AbstractDatabaseCommand.php
@@ -110,33 +110,33 @@ abstract class AbstractDatabaseCommand extends Command
      * Get the number of open connections for a Postgres database.
      *
      * @param  \Illuminate\Database\ConnectionInterface  $connection
-     * @return mixed
+     * @return int
      */
     protected function getPgsqlConnectionCount(ConnectionInterface $connection)
     {
-        return $connection->selectOne('select count(*) as connections from pg_stat_activity')->connections;
+        return (int) $connection->selectOne('select count(*) as connections from pg_stat_activity')->connections;
     }
 
     /**
      * Get the number of open connections for a MySQL database.
      *
      * @param  \Illuminate\Database\ConnectionInterface  $connection
-     * @return mixed
+     * @return int
      */
     protected function getMySQLConnectionCount(ConnectionInterface $connection)
     {
-        return $connection->selectOne($connection->raw('show status where variable_name = "threads_connected"'))->Value;
+        return (int) $connection->selectOne($connection->raw('show status where variable_name = "threads_connected"'))->Value;
     }
 
     /**
      * Get the number of open connections for an SQL Server database.
      *
      * @param  \Illuminate\Database\ConnectionInterface  $connection
-     * @return mixed
+     * @return int
      */
     protected function getSqlServerConnectionCount(ConnectionInterface $connection)
     {
-        return $connection->selectOne('SELECT COUNT(*) connections FROM sys.dm_exec_sessions WHERE status = ?', ['running'])->connections;
+        return (int) $connection->selectOne('SELECT COUNT(*) connections FROM sys.dm_exec_sessions WHERE status = ?', ['running'])->connections;
     }
 
     /**
--- a/src/Illuminate/Database/Console/AbstractDatabaseCommand.php
+++ b/src/Illuminate/Database/Console/AbstractDatabaseCommand.php
@@ -41,7 +41,7 @@ abstract class AbstractDatabaseCommand extends Command
      *
      * @param  \Illuminate\Database\ConnectionInterface  $connection
      * @param  string  $table
-     * @return null
+     * @return int|null
      */
     protected function getTableSize(ConnectionInterface $connection, string $table)
     {
--- a/src/Illuminate/Database/Console/AbstractDatabaseCommand.php
+++ b/src/Illuminate/Database/Console/AbstractDatabaseCommand.php
@@ -5,6 +5,9 @@ namespace Illuminate\Database\Console;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
 use Illuminate\Console\Command;
 use Illuminate\Database\ConnectionInterface;
+use Illuminate\Database\MySqlConnection;
+use Illuminate\Database\PostgresConnection;
+use Illuminate\Database\SQLiteConnection;
 use Illuminate\Support\Arr;
 use Symfony\Component\Process\Exception\ProcessSignaledException;
 use Symfony\Component\Process\Exception\RuntimeException;
@@ -42,10 +45,10 @@ abstract class AbstractDatabaseCommand extends Command
      */
     protected function getTableSize(ConnectionInterface $connection, string $table)
     {
-        return match (class_basename($connection)) {
-            'MySqlConnection' => $this->getMySQLTableSize($connection, $table),
-            'PostgresConnection' => $this->getPgsqlTableSize($connection, $table),
-            'SqliteConnection' => $this->getSqliteTableSize($connection, $table),
+        return match (true) {
+            $connection instanceof MySqlConnection => $this->getMySQLTableSize($connection, $table),
+            $connection instanceof PostgresConnection => $this->getPgsqlTableSize($connection, $table),
+            $connection instanceof SQLiteConnection => $this->getSqliteTableSize($connection, $table),
             default => null,
         };
     }
@@ -158,7 +161,7 @@ abstract class AbstractDatabaseCommand extends Command
     /**
      * Ensure the dependencies for the database commands are available.
      *
-     * @return int
+     * @return int|null
      */
     protected function ensureDependenciesExist()
     {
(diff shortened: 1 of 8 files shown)


src/Illuminate/Database/Console/MonitorCommand.php
```diff
@@ -0,0 +1,145 @@
...
+#[AsCommand(name: 'db:monitor')]
+class MonitorCommand extends AbstractDatabaseCommand
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'db:monitor
+                {--databases= : The names of the databases to monitor}
```
nunomaduro, line 19: I think this should an argument (and not an option), that should inspect the `default` database my default.

src/Illuminate/Database/Console/AbstractDatabaseCommand.php
```diff
@@ -0,0 +1,205 @@
...
+    /**
+     * Get the size of a table in bytes.
+     *
+     * @param  \Illuminate\Database\ConnectionInterface  $connection
+     * @param  string  $table
+     * @return null
+     */
+    protected function getTableSize(ConnectionInterface $connection, string $table)
+    {
+        return match (class_basename($connection)) {
```
nunomaduro, line 45: I would do `match(true) { $connection instanceof ConnectionClass => ...` to avoid issues with custom connections.

src/Illuminate/Database/Console/AbstractDatabaseCommand.php
```diff
@@ -0,0 +1,205 @@
...
+            default => $database,
+        };
+    }
+
+    /**
+     * Get the size of a table in bytes.
+     *
+     * @param  \Illuminate\Database\ConnectionInterface  $connection
+     * @param  string  $table
+     * @return null
```
nunomaduro, line 41: This should be @return `null and something else`.

src/Illuminate/Database/Console/AbstractDatabaseCommand.php
```diff
@@ -0,0 +1,205 @@
...
+    {
+        $database ??= config('database.default');
+
+        return Arr::except(config('database.connections.'.$database), ['password']);
+    }
+
+    /**
+     * Ensure the dependencies for the database commands are available.
+     *
+     * @return int
```
nunomaduro, line 161: `@return int|null`.

src/Illuminate/Database/Console/ShowCommand.php
```diff
@@ -0,0 +1,183 @@
...
+
+#[AsCommand(name: 'db:show')]
+class ShowCommand extends AbstractDatabaseCommand
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'db:show {--database= : The database connection to use}
```
nunomaduro, line 21: I think this should be an argument, with a default value, and not an option.

src/Illuminate/Database/Events/DatabaseBusy.php
```diff
@@ -0,0 +1,32 @@
...
+     * @var string
+     */
+    public $database;
+
+    /**
+     * The number of open connections.
+     *
+     * @var int
+     */
+    public $connections;
```
nunomaduro, line 19: I would rename this to something more explicit.

SuperDJ:
Would this also show table check constraints?

src/Illuminate/Database/Console/MonitorCommand.php
joedixon: For this, we took the lead from existing commands such as `migrate`, `show:model` and `schema:dump` so I think it probably makes sense to leave it as an option.

jbrooksuk:
> Would this also show table [check constraints](https://dev.mysql.com/blog-archive/mysql-8-0-16-introducing-check-constraint/)?

Not currently it doesn't.

src/Illuminate/Database/Console/MonitorCommand.php
jbrooksuk: Changed for this command so that it matches `queue:monitor`.

driesvints:
I want to suggest to keep `database` as an option as that's by far the most frequently used way of doing that across our commands.

src/Illuminate/Database/Console/MonitorCommand.php
```diff
@@ -0,0 +1,151 @@
...
+     * @var \Illuminate\Contracts\Events\Dispatcher
+     */
+    protected $events;
+
+    /**
+     * Create a new db monitor command.
+     *
+     * @param  \Illuminate\Database\ConnectionResolverInterface  $connection
+     * @param  \Illuminate\Contracts\Events\Dispatcher  $events
+     *                                                           * @param  \Illuminate\Support\Composer  $composer
```
bastien-phi, line 60:
```suggestion
      * @param  \Illuminate\Support\Composer  $composer
```

timacdonald:
I feel like the `db:table` command data may potentially be better off incorporated into the `model:show` command. I know at one point @jessarcher had the command showing the index info behind a verbose flag, at which point they feel kinda identical - but the db version doesn't show model related stuff as well.

I get that there might be tables that are not backed by a model, but just wanted to make that suggestion.

src/Illuminate/Database/Console/MonitorCommand.php
driesvints: Thanks!

jbrooksuk:
@timacdonald the way @joedixon and I see this command is that `db:` is more for production use. You're probably not _too_ interested in what a model looks like when you're on a production server, but you do want to know what's going on with the tables - it's like TablePlus in your Terminal.

imManish:
@jbrooksuk - What if migrations are not present, and framework directly connect with remote database on production. `db:show` able to show the `schema` of the database. hope this works with remotely connected database though.

driesvints:
@timacdonald there's still Laravel apps that don't use Eloquent and just use the database query builder to perform database operations. In that regard it makes sense to me to keep `db:table` separately so you have the least amount of abstraction.

jbrooksuk:
> @jbrooksuk - What if migrations are not present, and framework directly connect with remote database on production. `db:show` able to show the `schema` of the database. hope this works with remotely connected database though.

This isn't using migrations to generate the output.

imManish:
> information

@jbrooksuk - I understand that! but the concern is if the app connected with multiple database such as DB1 and DB2 will is there any option that i have to parse along with command `db:show --name=db1`,`db:monitor --name=db1` ,`db:table --name=db1` i guess it should have work like this.

could you let us know show it will work with multiple database connection.  can't we have it within model which would have connection property which will display which model connected with which database and will display accordingly.

joedixon:
> @jbrooksuk - I understand that! but the concern is if the app connected with multiple database such as DB1 and DB2 will is there any option that i have to parse along with command `db:show --name=db1`,`db:monitor --name=db1` ,`db:table --name=db1` i guess it should have work like this.
> 
> could you let us know show it will work with multiple database connection. can't we have it within model which would have connection property which will display which model connected with which database and will display accordingly.

Hi @imManish - This set of commands doesn't have any relation to models. It specifically allows you to get a snapshot of a single database/table. The commands will use your default connection, however you are free to choose which database you which to inspect by using the `--database` option.

PR #43377, opened 2022-07-22
[9.x] Filter out disabled default timestamps

alisaleem27:
I can't seem to find it in the documentation anymore, but setting `public const CREATED_AT = null;` (or similarly for `UPDATED_AT`) in a model would disable that timestamp attribute, while keeping the functionality for the other intact. This is for when you only want to disable either `created_at` or `updated_at`, not both.

If one of these default timestamp attributes is disabled, the `getDates()` function would return a `null` for that attribute's name rather than omitting it.
```
[null, 'updated_at']
```
whereas the correct return value should be
```
['updated_at']
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1376,10 +1376,10 @@ trait HasAttributes
             return $this->dates;
         }
 
-        $defaults = [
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't really plan on changing this on a patch release.

alisaleem27:
@taylorotwell should I create this as a PR to the 10.x branch?

PR #43378, opened 2022-07-22
[10.x] Add support for multiple hash algorithms

medeirosinacio:
Hello.

Is this modification valid?
We often need the `sha256` of the file, it would be interesting for `Filesystem` to get it.
If it is already valid, I can include it in the tests.

I was always confused by the original implementation, as "hash" refers me to all sorts of algorithms. That way we don't break what already exists (md5 default) and enable hashing like sha256 or sha512.

https://www.php.net/manual/en/function.hash-algos.php

```php

File::hash($path)
// 196b0f14eba66e10fba74dbf9e99c22f => default md5

File::hash($path, 'sha256')
// 19c451aedfb24c338a9a2a5c31d553ed77e7cdefc655035f390176ac24066051

```

### DIFF
--- a/src/Illuminate/Filesystem/Filesystem.php
+++ b/src/Illuminate/Filesystem/Filesystem.php
@@ -167,6 +167,7 @@ class Filesystem
      * Get the hash of the file at the given path.
      *
      * @param  string  $path
+     * @param  string  $algorithm
      * @return string
      */
     public function hash($path, $algorithm = 'md5')
--- a/src/Illuminate/Filesystem/Filesystem.php
+++ b/src/Illuminate/Filesystem/Filesystem.php
@@ -164,14 +164,14 @@ class Filesystem
     }
 
     /**
-     * Get the MD5 hash of the file at the given path.
+     * Get the hash of the file at the given path.
      *
      * @param  string  $path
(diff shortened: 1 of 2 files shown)


src/Illuminate/Filesystem/Filesystem.php
```diff
@@ -164,14 +164,14 @@ public function lines($path)
...
 
     /**
-     * Get the MD5 hash of the file at the given path.
+     * Get the hash of the file at the given path.
      *
      * @param  string  $path
      * @return string
      */
-    public function hash($path)
+    public function hash($path, $algorithm = 'md5')
```
GrahamCampbell, line 172: you need to add param doc for this

GrahamCampbell (changes requested):

src/Illuminate/Filesystem/Filesystem.php
medeirosinacio: right

dinhquochan:
It's compatible for 9.x

medeirosinacio:
Should I do PR for laravel 9?

taylorotwell:
Sure

PR #43381, opened 2022-07-23
Added a new array helper function

Hetarth02:
Added a new Array helper function name `depth` which determines how nested an array is or in short the depth of an array.

> The function will return 0 for an empty array and data types other than array including `null`. This behaviour is taken from this [Flat array depth](https://aplwiki.com/wiki/Depth).

- It will be helpful as a breaking condition in cases where one has to iterate over an extremely nested array or unknown levels of array.
- It mostly implements the logic using core php functions except for the Array `dot` method which in itself isn't breaking any existing features. So, the function would be stable.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -839,7 +839,7 @@ class Arr
                 },
                 $depth_array
             );
-            
+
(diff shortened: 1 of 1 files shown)


taylorotwell:

PR #43383, opened 2022-07-23
[9.x] Add whereNot method to Fluent JSON testing matchers

mateusjatenee:
I have a use case where I want to test that a given item *does* not exist on a JSON collection. 
I found it a bit difficult to do this with the existing API, but it is quite possible (and likely) that I just missed something.

Here's an example of my use case: https://twitter.com/mateusjatenee/status/1550574619995815936

The tests are a bit messy because I started out scoping then and then implementing (tried out a couple different things as well), but if the PR looks good I can make them a little better.

### DIFF
--- a/src/Illuminate/Testing/Fluent/Concerns/Matching.php
+++ b/src/Illuminate/Testing/Fluent/Concerns/Matching.php
@@ -48,6 +48,49 @@ trait Matching
     }
 
     /**
+     * Asserts that the property does not match the expected value.
+     *
+     * @param  string  $key
+     * @param  mixed|\Closure  $expected
+     * @return $this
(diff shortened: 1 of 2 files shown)


mateusjatenee:
Whoopsie. Will check tests when I get back home.

morloderex:
@mateusjatenee How is this different than the `missing` method with a potentially new callback parameter added to it?

mateusjatenee:
@morloderex if we could figure out a nice way to handle the callback I don't think it'd be much different. The goal is to just make sure that a certain item does not exist (by filtering) — right now `missing` just makes sure a property doesn't exist.

PR #43385, opened 2022-07-23
[8.x] Bump voku/portable-ascii to v2

rgson:
Backport of #40618

Pulls in newer dependencies, including bug fixes, and makes the transition to 9.x smoother.


GrahamCampbell:
This is a major breaking change, and cannot be done on 8.x. If you want v2, you need to upgrade to Laravel 9.

PR #43387, opened 2022-07-23
[9.x] Added conditional lines to MailMessage

shino47:
Currently, If we need to add a line to a MailMessage based on a condition, we have to do something like this:

```php
{
    $message = new MailMessage();
    $message->greeting('Hello there!');
    $message->line('Your order has been canceled');
    if ($this->amount > 0) {
        $message->line("The refunded amount is {$this->amount}");
    }
    $message->line('...');
    return $message;
}
```

With these methods, we can do:

```php
{
    return (new MailMessage)
        ->greeting('Hello there!')
        ->line('Your order has been canceled')
        ->lineIf($this->amount > 0, "The refunded amount is {$this->amount}")
        ->line('...');
}
```

Which looks similar to the docs.

### DIFF
--- a/src/Illuminate/Notifications/Messages/SimpleMessage.php
+++ b/src/Illuminate/Notifications/Messages/SimpleMessage.php
@@ -158,6 +158,22 @@ class SimpleMessage
     }
 
     /**
+     * Add a line of text to the notification if the given condition is true.
+     *
+     * @param  bool  $boolean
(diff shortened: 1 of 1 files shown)


mateusjatenee:
I wonder if using the `Conditionable` trait isn't a better approach here — that'd allow us to call any methods based on a condition.

PR #43388, opened 2022-07-23
[9.x] Improve pop method in Collection

moharami:
when we use pop with count zero or negative, doesn't remove items correctly. I fix that and add some tests for this purpose.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -878,6 +878,10 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
      */
     public function pop($count = 1)
     {
+        if ($count <= 0) {
+           return;
+        }
+
         if ($count === 1) {
             return array_pop($this->items);
         }
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -879,7 +879,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     public function pop($count = 1)
     {
         if ($count <= 0) {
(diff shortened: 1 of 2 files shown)


taylorotwell:
What was the previous behavior?

ankurk91:
This statement should already take care of negative count

`min($count, $collectionCount)`

moharami:
**Previous Behavior**
    
    $c = collect(['foo', ' bar', 'baz']);
    $c->pop(0); //["baz", " bar"]
    $c->all(); //["foo"]


    $c = collect(['foo', ' bar', 'baz']);
    $c->pop(-1); //["baz", " bar", "foo"]
    $c->all(); //[]


    $c = collect(['foo', ' bar', 'baz']);
    $c->pop(-5); //["baz", " bar", "foo", null, null, null, null]
    $c->all(); //[]



**Explain Behavior Case 1:**

  for example, when we put -5 in count in pop methods, it goes here

     $collectionCount = $this->count();

        foreach (range(1, min($count, $collectionCount)) as $item) {
            array_push($results, array_pop($this->items));
        }

        return new static($results);

and it behaviour like this

        min($count, $collectionCount) // min(-5 , 3)  = -5
        range(1, min($count, $collectionCount) // range(1, -5) = [1, 0,-1,-2,-3,-4,-5]

and array_pop iterate 7 time, and it removes 7 item from collection(that has just 3 item)

**Explain Behavior Case 2:**
when we put zero

    min($count, $collectionCount) // min(0 , 3)  = 0
    range(1, min($count, $collectionCount) // range(1, 0) = [1, 0]

and array_pop remove 2 item from collection

taylorotwell:
I don't know if I really want to mess with this on a patch release.

Issue #42967, opened 2022-06-27
Unexpected behavior in firstOrNew method in belongs to many relationships


ashkanRimer:
- Laravel Version: v9.17.0
- PHP Version: 8.1.7
- Database Driver & Version: MySQL 8.0.29

### Description:
There is an unexpected behavior in `firstOrNew` method in `belongsToMany` relationships. When I try to `firstOrNew` a related model which it doesn't exist, I get the first record of database which I shouldn't.

### Steps To Reproduce:
I have a `ContentPost` model which has a `belongsToMany` relationship with `UploaderFile` and the pivot table is called `uploader_attachments`. 
This is the relation method in `ContentPost`:
```php
public funciton files()
{
   return $this->belongsToMany(UploaderFile::class, "uploader_attachments", "model_id", "file_id")
                    ->withPivot("group", "model_name")
                    ->withTimestamps()
                    ->where("model_name", "ContentPost")
            ;
}
```

When I try to find a `UploaderFile` with a `original_name` which doesn't exist, It returns a new empty instance as below:
```php
# executed query on uploader-file model:
(new UploaderFile)->where("original_name","something not exists")->firstOrNew();

# results:
App\Models\UploaderFile {
     id: 0,
 }
```
But when I execute this query on the ContentPost's files relation method. It gives me the first record.
```php
# executed query on content-post files relation method:
(new ContentPost)->files()->where("original_name","something not exists")->firstOrNew();

# results:
App\Models\UploaderFile {
     id: 1,
     original_name: "24432660_0.jpg",
     extension: "jpg",
     mime_type: "image/jpeg",
     given_name: "1578837071_SW4zL94LHGn1ZYb5VrHHkFljJ8PCHjH6BCYfMGAx",
     private: 0,
     created_at: "2020-01-12 17:21:11",
     updated_at: "2020-01-12 17:21:11",
     deleted_at: null,
     created_by: 2,
     updated_by: 0,
     deleted_by: 0,
     meta: "{"versions":["panel_thumb"]}",
   }
```

vaites:
Same problem here, I found it with something like this, where there's no addresses associated to the person:

```php
$person->addresses()->where('postal_code', 12345)->first();      // null
$person->addresses()->where('postal_code', 12345)->firstOrNew(); // address with ID 1 (not related with person)
```

The `firstOrNew()` method must respect all previous conditions.

Please let me know if I can help by providing an example.

driesvints:
[...]

PR #43390, opened 2022-07-24
[9.x] Fix first or new method in belongs to many relations

ashkan-jafarzadeh:
This pr is created to fix the problem that I have described in this [issue](https://github.com/laravel/framework/issues/42967).

`firstOrNew` method's behavior in `belongsToMany` relations should be the same as `first` method  when the model exists, but it is not. It should consider the chained builder conditions. 

Currently, the `firstOrNew` method in `belongsToMany` relations always returns the model of the first record of the database, no matter the specified model already has it own relations ([described in example](https://github.com/laravel/framework/issues/42967#issuecomment-1193129527)).

Please consider checking the [issue](https://github.com/laravel/framework/issues/42967) which has more details and examples.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php
@@ -600,7 +600,7 @@ class BelongsToMany extends Relation
      */
     public function firstOrNew(array $attributes = [], array $values = [])
     {
-        if (is_null($instance = $this->related->where($attributes)->first())) {
+        if (is_null($instance = $this->where($attributes)->first())) {
             $instance = $this->related->newInstance(array_merge($attributes, $values));
         }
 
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
I think this needs unit tests providing, before it could be considered for merge.

taylorotwell:
I really don't think we would change this on a patch release.

ashkanRimer:
@taylorotwell Thank you for your attention,

Since this unexpected behavior exists, should I create this pr on master branch?

driesvints:
@ashkanRimer I don't think Taylor wants to make any changes here right now.

vaites:
As [discussed here](https://github.com/laravel/framework/issues/42967#issuecomment-1193129527), this must be seen as a bug and it must be fixed ASAP. Is not an _unexpected behaviour_, is a serious bug when used on relationships...

Snowmanko:
@ashkan-jafarzadeh  @driesvints @vaites did you managed to solve this, we have same problem.

vaites:
Just put the conditions in `firstOrNew()` method instead of using wheres. Take a look [here](https://github.com/laravel/framework/issues/42967#issuecomment-1189301036).

I still think that this must be fixed.

NielsJanssen:
@taylorotwell could you please take another look at this issue. This is really dangerous behaviour, and the PR seems to fix it.

PR #43393, opened 2022-07-24
Removed redundant `elseif`

AbdelrhmanSaid:
There's no need to use `elseif` when using `if -> return`, as well as create unnecessary extra indentation for the main code of the function.


taylorotwell:
It's OK for now if it's not broken.

PR #43397, opened 2022-07-24
[9.x] Break out quotes method

dwightwatson:
This is a little odd so I'm happy if this considered not worth fixing, but I've been using `Illuminate\Foundation\Inspiring::quote()` in my production app - showing a nice quote to our internal teams when they've completed the list of pending tasks.

Despite the lovely new changes to the console, it appears the quote is formatted for the console regardless of the context:

```
<options=bold>“ Happiness is not something readymade. It comes from your own actions. ”</> <fg=gray>— Dalai Lama</>
```

This breaks out a `quotes` method so the quotes can be accessed separately. It'll mean I can just do `Inspiring::quotes::random()` in my own code and not impact the existing use-cases of `quote()`.

### DIFF
--- a/src/Illuminate/Foundation/Inspiring.php
+++ b/src/Illuminate/Foundation/Inspiring.php
@@ -55,6 +55,18 @@ class Inspiring
      */
     public static function quote()
     {
+        return static::quotes()
+            ->map(fn ($quote) => static::formatForConsole($quote))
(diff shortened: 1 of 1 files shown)


ahinkle:
Breaking change and one I would find to not be commonly used. You can overwrite the function in your own app.

PR #43400, opened 2022-07-25
[9.x] Round milliseconds in console output runtime

SjorsO:
Commands always show how long they took to run in milliseconds with 2 decimal precision. This PR rounds the milliseconds, because having 2 decimal precision on milliseconds is unnecessary.

Before:
image

After:
image

### DIFF
--- a/src/Illuminate/Console/View/Components/Task.php
+++ b/src/Illuminate/Console/View/Components/Task.php
@@ -38,7 +38,7 @@ class Task extends Component
             throw $e;
         } finally {
             $runTime = $task
-                ? (' '.number_format((microtime(true) - $startTime) * 1000, 2).'ms')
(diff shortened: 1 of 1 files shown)


driesvints (approved):

src/Illuminate/Console/View/Components/Task.php
```diff
@@ -38,7 +38,7 @@ public function render($description, $task = null, $verbosity = OutputInterface:
             throw $e;
         } finally {
             $runTime = $task
-                ? (' '.number_format((microtime(true) - $startTime) * 1000, 2).'ms')
+                ? (' '.number_format((microtime(true) - $startTime) * 1000).'ms')
```
rcerljenko, line 41: you could also use `round()` here...

PR #43402, opened 2022-07-25
[9.x] Allow customization of script and stylesheet tag generation with Vite

axlon:
When using the new Vite plugin to compile assets its currently not possible to define how style and script tags are generated, this PR aims to change that.

This will allow developers to change the tags for their assets to their specific needs (my own use-case being a CSP nonce).

Usage would be:

```php
use Illuminate\Foundation\Vite;

Vite::makeScriptTagUsing(function (string $url) {
    return sprintf('<script type="module" src="%s" defer></script>', $url);
});

Vite::makeStylesheetTagUsing(function (string $url) {
    return sprintf('<link rel="stylesheet" href="%s" crossorigin />', $url);
});
```

### DIFF
--- a/src/Illuminate/Foundation/Vite.php
+++ b/src/Illuminate/Foundation/Vite.php
@@ -9,6 +9,20 @@ use Illuminate\Support\Str;
 class Vite
 {
     /**
+     * The callback to be used when making script tags.
+     *
+     * @var \Closure|null
+     */
+    protected static $scriptTagCallback;
+
+    /**
+     * The callback to be used when making stylesheet tags.
+     *
+     * @var \Closure|null
+     */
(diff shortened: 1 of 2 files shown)


timacdonald:
Hey @axlon, thank you for the PR.

We dig idea, however as it stands the PR will only allow you to add generic attributes - but we want to see if we can come up with a solution that will allow us to handle CSP, SRI, and generic attributes.

We are actively working on a holistic solution to all of these, as they all share a similar problem space, and will send up an alternative PR shortly and get your thoughts and feedback on it.

axlon:
Sounds good @timacdonald !

axlon:
Closing this in favor of #43442

PR #43403, opened 2022-07-25
[9.x] Fix parameter types

halloei:
In #37759 I made a change without adjusting the method's parameter types. This fixes it to make IDEs and static code analysis happy.


driesvints:
Thanks!

PR #43404, opened 2022-07-25
Typehints the real model in Auth()->user

deployHuman:
Only way to get intellisense typehinting in the code editor is to have it point to the original app\models\user and in there one can use @property int $id etc for each param.
This way there is no need to create a wrapper for Auth or extend it, or even edit vendor files.


driesvints:
The purpose of this method is to return an authenticable. That could be anything else besides a User model. Making this concrete is not the right way forward.

PR #43405, opened 2022-07-25
[9.x] PHPDoc fix: Add missing `null` return type for `min` method

chescos:
The `min` method can also return `null`, which is currently not reflected in the PHPDoc of the method. This PR adds `null` as a possible return type, which improves IDE experience and reduces unnecessary static analysis errors.


driesvints:
This seems to break the tests.

PR #43410, opened 2022-07-25
[9.x] Improve test for reverse method in Collection

moharami:


taylorotwell:
I think I'm OK with our tests now.

PR #43414, opened 2022-07-25
[9.x] Added `notFound` on Http/Client/Response

foremtehan:
The 404 status is common after 200


taylorotwell:
You can check the status manually.

PR #43432, opened 2022-07-26
[9.x]: drop column if exists

maloun96:
Unfortunately, when we use `->dropColumn` it triggers errors if the column doesn't exist. This PR adds a check if the table has the column and deletes it.

### DIFF
--- a/src/Illuminate/Database/Schema/Builder.php
+++ b/src/Illuminate/Database/Schema/Builder.php
@@ -292,9 +292,13 @@ class Builder
      */
     public function dropColumns($table, $columns)
     {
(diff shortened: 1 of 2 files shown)


taylorotwell:

maloun96:
@taylorotwell Can you explain to me please why this PR is not good? Because we have a big amount of code in the core app? Then why this feature is released? I mean, for me this feature release is for lazy people, but really really good features are ignored. Waiting for your response.
Thanks.
image

maloun96:
here is the second version of this feature 
https://github.com/laravel/framework/pull/43422

PR #43434, opened 2022-07-26
Change successful status hardcoded number to constant in console.stub

mhabedini:
In console.stub file when returning the handle method result, it's beneficial to use Console constants as result, so that developer know what is exactly the zero in here and what are the alternate values to use.


driesvints:
See many previous attempts.

PR #43440, opened 2022-07-26
[9.x] Add test for add method in Collection

moharami:


ankurk91:
Collection has so many methods that You can send one PR daily :laughing:

PR #43441, opened 2022-07-27
[9.x] Add withoutUpdatedTimestamp()

patrickomeara:
Many times I've needed to run a command that updates a model without updating the `updated_at` column. This might be migrating data or a client wants to make a batch change to their data without losing the existing `updated_at`.

Usually I reach for one of the two options:

```php
// iterate over and set timestamps to false - slow
$users->each(function ($user) {
    $user->timestamps = false;
    $user->update(['something' => 'here']);
});

// extend the model in the command and set timestamps to false - dirty
UserWithoutTimestamps::query()->update(['something' => 'here']);

// ...
class UserWithoutTimestamps extends User {
    public $timestamps = false;
}
```

Recently I've come across the `withoutTouching` static method, which had me thinking "why don't we have this for `updated_at`?" 🤔 

So here it is.

You can now ignore all models `updated_at` timestamps via `Model::withoutUpdatedTimestamp()`

```php
Model::withoutUpdatedTimestamp(function () use ($user, $post, $comment) {
    // no change to updated_at
    $user->update(['something' => 'here']);
    $post->update(['something' => 'here']);
    $comment->update(['something' => 'here']);
});
```

...or ignore mulitple models `updated_at` timestamp via `Model::withoutUpdatedTimestampOn()`

```php
Model::withoutUpdatedTimestampOn([User::class, Post::class], function () use ($user, $post, $comment) {
    // no change to updated_at
    $user->update(['something' => 'here']);
    $post->update(['something' => 'here']);
    // updated_at will be set
    $comment->update(['something' => 'here']);
});
```

...or ignore a single model `updated_at` timestamp via `User::withoutUpdatedTimestamp()`

```php
User::withoutUpdatedTimestamp(function () use ($user, $post, $comment) {
    // no change to updated_at
    $user->update(['something' => 'here']);
    // updated_at will be set
    $post->update(['something' => 'here']);
    $comment->update(['something' => 'here']);
});
```

Laravel sets both `updated_at` and `created_at` when a model is created. This functionality has stayed the same as it's a data integrity issue. So creating models in the callback will still have `updated_at` set.

```php
User::withoutUpdatedTimestamp(function() use ($user) {
    // no change to updated_at
    $user->update(['something' => 'here']);

    // updated_at will be set as usual
    $player = User::create(['name' => 'Patrick']);
});
```

This also works for the query builder

```php
User::withoutUpdatedTimestamp(function() use ($user) {
    // no change to updated_at
    User::query()
        ->where('something', 'there')
        ->update(['something' => 'here']);
});
```

and upserting

```php

User::create(['email' => 'taylor@laravel.com']);

User::withoutUpdatedTimestamp(function() use ($user) {
    User::upsert(
        [
            // no change to updated_at
            ['email' => 'taylor@laravel.com', 'name' => 'Taylor'],
            // updated_at will be set with created_at as the model is new
            ['email' => 'pat@domain.com', 'name' => 'Pat'],
        ], 
        ['email'], 
        ['name'],
    );
});
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1067,7 +1067,8 @@ class Builder implements BuilderContract
     protected function addUpdatedAtColumn(array $values)
     {
         if (! $this->model->usesTimestamps() ||
-            is_null($this->model->getUpdatedAtColumn())) {
+            is_null($this->model->getUpdatedAtColumn()) ||
+            $this->model::isIgnoringUpdatedTimestamp()) {
             return $values;
         }
 
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1126,7 +1126,7 @@ class Builder implements BuilderContract
      */
     protected function addUpdatedAtToUpsertColumns(array $update)
     {
-        if (! $this->model->usesTimestamps()) {
+        if (! $this->model->usesTimestamps() || $this->model::isIgnoringUpdatedTimestamp()) {
             return $update;
         }
 
--- a/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
@@ -14,6 +14,13 @@ trait HasTimestamps
     public $timestamps = true;
 
     /**
+     * The list of models classes that should not have timestamps updated on update.
+     *
+     * @var array
+     */
+    protected static $ignoreUpdatedTimestamp = [];
+
+    /**
      * Update the model's update timestamp.
      *
      * @param  string|null  $attribute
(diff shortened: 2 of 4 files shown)


henzeb:
I think you found a bug... ignoreTouching is supposed to do that, but it's currently only doing that for the touch method. all it has to do is using the ignoringTouch method where they now simply check use timestamps and no need for another method. Unless I am missing something that might cause breaking changes?

patrickomeara:
Hi @henzeb thanks for looking through this PR.

As far as I am aware, `withoutTouching()` only prevents touching parent relationship's timestamps not the model itself.

### Docs

https://laravel.com/docs/master/eloquent-relationships#touching-parent-timestamps

### Existing Test

Testing asserting that the model is touched but parent isn't

https://github.com/laravel/framework/blob/ae76bf60795117e28e7beb64346bc2011e195c23/tests/Database/DatabaseEloquentIntegrationTest.php#L1858-L1883

henzeb:
I am not core team. But you're right. It does only work on parents. But it makes no sense to me if you are encapsulating it in a method that puts itself in the Model::$ignoreOnTouch and then it does not ignore the timestamps.

taylorotwell:

PR #43445, opened 2022-07-27
Update Repository.php

Mohammadreza-73:
Implement pull method with arrow function.


GrahamCampbell:
👎 This change is incorrect. Arrow functions cannot be used with void return types.

PR #43449, opened 2022-07-27
[9.x] Support conditionables that get condition from target object

ralphjsmit:
This PR allows people to use the `->when()` and `->unless()` methods without passing in any parameters. 

Currently, people are required to pass in the condition as the first parameter:

```php
Storage::disk('orders')
    ->when($condition = fn(Filesystem $disk) => $disk->missing('someDirectory'))
    ->makeDirectory('someDirectory');
```

The `Conditionable` will execute the closure and store that as the condition. The next method call ("makeDirectory") will be proxied to the original object based on that condition.

This PR allows to create a proxy without an initial condition. The first method call to the proxy will become the condition, and the second call will be proxied to the original object based on that condition.

With this change, the above example would become:

```php
Storage::disk('orders')
    ->when()->missing('someDirectory')
    ->makeDirectory('someDirectory');
```

(In this example I'll assume that the #43450 PR gets merged first, but you certainly can get the general idea.)

Thanks!

### DIFF
--- a/src/Illuminate/Conditionable/HigherOrderWhenProxy.php
+++ b/src/Illuminate/Conditionable/HigherOrderWhenProxy.php
@@ -30,7 +30,7 @@ class HigherOrderWhenProxy
      *
      * @var bool
      */
-    protected $negateCondition;
+    protected $negateConditionOnCapture;
 
     /**
      * Create a new proxy instance.
@@ -62,9 +62,9 @@ class HigherOrderWhenProxy
      *
      * @return $this
      */
-    public function negateCondition()
+    public function negateConditionOnCapture()
     {
-        $this->negateCondition = true;
+        $this->negateConditionOnCapture = true;
 
         return $this;
     }
@@ -94,7 +94,7 @@ class HigherOrderWhenProxy
         if (! $this->hasCondition) {
             $condition = $this->target->{$method}(...$parameters);
 
-            return $this->condition($this->negateCondition ? ! $condition : $condition);
+            return $this->condition($this->negateConditionOnCapture ? ! $condition : $condition);
         }
 
         return $this->condition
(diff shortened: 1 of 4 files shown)


src/Illuminate/Conditionable/HigherOrderWhenProxy.php
```diff
@@ -18,17 +18,55 @@ class HigherOrderWhenProxy
...
+    protected $negateCondition;
+
     /**
      * Create a new proxy instance.
      *
      * @param  mixed  $target
-     * @param  bool  $condition
      * @return void
      */
-    public function __construct($target, $condition)
```
driesvints, line 28: Removing this is a breaking change.

src/Illuminate/Conditionable/Traits/Conditionable.php
```diff
@@ -13,17 +13,21 @@ trait Conditionable
...
      * @template TWhenReturnType
      *
-     * @param  (\Closure($this): TWhenParameter)|TWhenParameter  $value
+     * @param  (\Closure($this): TWhenParameter)|TWhenParameter|null $value
      * @param  (callable($this, TWhenParameter): TWhenReturnType)|null  $callback
      * @param  (callable($this, TWhenParameter): TWhenReturnType)|null  $default
      * @return $this|TWhenReturnType
      */
-    public function when($value, callable $callback = null, callable $default = null)
+    public function when($value = null, callable $callback = null, callable $default = null)
```
driesvints, line 21: Changing the method signature here is a breaking change.

src/Illuminate/Conditionable/HigherOrderWhenProxy.php
ralphjsmit: You are right that it is a breaking change, though isn't it considered part of the internal API?

chu121su12:
@ralphjsmit What's the rationale to not support proxy call similar to map/each? So that the usage would be without parenthesis `$x->when->has('init')->call()`

PR #43451, opened 2022-07-27
[9.x] Add queued command function to scheduler

dmason30:
**Description**
```php
$schedule->queuedCommand(FooCommand::class)->everyMinute();
$schedule->queuedCommand(FooCommand::class, ['some-argument' => 'foo']);
$schedule->queuedCommand(FooCommand::class, ['some-argument' => 'foo'], 'queue');
$schedule->queuedCommand(FooCommand::class, ['some-argument' => 'foo'], 'queue', 'connection');
```
Adds `queuedCommand` function to the scheduler that executes the equivalent of `Artisan::queue(...)` on the given schedule. It also provides the ability to optionally customise the queue and connection if needed.

**Motivation**
When running Laravel on serverless (e.g. Laravel Vapor) applications we cannot reliably use the `->runInBackground()` as the Lambda instance could be stopped before the background processes are executed. 

When a package wants you to stick something in the scheduler they usually provide this as an artisan command, for example, the `PruneBatchesCommand`, `PruneFailedJobsCommand` and other `prune:*` commands in Telescope/Horizon etc. It would be great if we could have a function like this so we can choose to queue these commands if we want to.

### DIFF
--- a/src/Illuminate/Console/Scheduling/Schedule.php
+++ b/src/Illuminate/Console/Scheduling/Schedule.php
@@ -9,6 +9,7 @@ use Illuminate\Console\Application;
 use Illuminate\Container\Container;
 use Illuminate\Contracts\Bus\Dispatcher;
 use Illuminate\Contracts\Cache\Repository as Cache;
+use Illuminate\Contracts\Console\Kernel;
 use Illuminate\Contracts\Container\BindingResolutionException;
 use Illuminate\Contracts\Queue\ShouldBeUnique;
 use Illuminate\Contracts\Queue\ShouldQueue;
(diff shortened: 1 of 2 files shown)


driesvints:
Maybe `queueCommand`?

driesvints:
Let's see what @taylorotwell says.

taylorotwell:
Honestly I would just use what we already have available:

```php
$schedule->call(fn () => dispatch(fn() => Artisan::call('whatever-command-you-want')));
```

dmason30:
For anyone in the future searching this, I have extracted the change to a package here:
https://github.com/fidum/laravel-schedule-queue-command

Feel free to install or fork as needed. ❤️

PR #43458, opened 2022-07-28
[9.x] Get the total items with simplePagination

mtawil:
Since we already did counting for all items being paginated with `simplePaginator`, I suggest setting the items count before slicing and adding the ability to get it using the `total` method, the same as on the `LengthAwarePaginator` class.


taylorotwell:
That is not the complete total.

PR #43462, opened 2022-07-28
[9.x] Get filesystem config options using dot array syntax

ziadoz:
This makes it easier to get configuration options from a filesystem, and mirrors the same method on database connections:

```
$filesystem->getConfig();      // ['foo' => 'bar']
$filesystem->getConfig('foo'); // 'bar'
```

In my use case I'm storing additional options in the config used for logging, and this makes it easier to get at.


taylorotwell:

PR #43463, opened 2022-07-28
[9.x] Fixes the issue of overriding `$wrap` property

mahmoudmohamedramadan:
So, let me explain the issue in more detail:

Laravel [says](https://laravel.com/docs/9.x/eloquent-resources#data-wrapping) that I can change `data` keyword via defining a `$wrap` attribute on the resource class like so,

```PHP
/**
 * The "data" wrapper that should be applied.
 *
 * @var string|null
 */
 public static $wrap = 'user';
```

But when I tried this way nothing changed as we can see in the next figure:

before-this-PR

- - -

**If this PR gets merged**

As we can see in the next figure that this issue has been solved 🎉

after-PR-was-accepted

### DIFF
--- a/src/Illuminate/Http/Resources/Json/JsonResource.php
+++ b/src/Illuminate/Http/Resources/Json/JsonResource.php
@@ -80,6 +80,10 @@ class JsonResource implements ArrayAccess, JsonSerializable, Responsable, UrlRou
             if (property_exists(static::class, 'preserveKeys')) {
                 $collection->preserveKeys = (new static([]))->preserveKeys === true;
             }
+            
+            if (property_exists(static::class, 'wrap')) {
+                $collection::$wrap = static::$wrap;
+            }
         });
     }
 


rodrigopedra:
This is a major breaking change. `$wrap` works fine for individual resources. 

An application using `$wrap` for individual resources (the default behavior) is already expecting unwrapped resource collections when calling its `::collection(...)` method.

I am not opposing to you to propose this change, but this should be sent to master instead.

In the meanwhile you can use the docs suggestions on how to wrap resource collections:

> If you would like all resource collections to be wrapped in a `data` key, regardless of their nesting, you should **define a resource collection class for each resource** and return the collection within a `data` key.
> 
> ~~~php
> <?php
>  
> namespace App\Http\Resources;
>  
> use Illuminate\Http\Resources\Json\ResourceCollection;
>  
> class CommentsCollection extends ResourceCollection
> {
>     /**
>      * Transform the resource collection into an array.
>      *
>      * @param  \Illuminate\Http\Request  $request
>      * @return array
>      */
>     public function toArray($request)
>     {
>         return ['data' => $this->collection];
>     }
> }
> ~~~

reference: https://laravel.com/docs/9.x/eloquent-resources#wrapping-nested-resources

PR #43467, opened 2022-07-28
SoftDeletes trait new runRestore method

medilies:
While developing [Eloquent Logs package](https://github.com/elaborate-code/laravel-eloquent-logs) I ran into the following issues when a model has the `SoftDeletes` trait:
1. **Soft deleting** emits both `delete` and `soft delete` events.
2. **Force deleting** emits both `delete` and `force delete` events.
3. **Restoring emits** both `update` and `restore` events.

Here is how I register my event handlers:

```php
trait HasLogs
{
    public static function bootHasLogs(): void
    {
        self::created(callback: fn ($model) => self::log($model, 'created'));
        self::updated(callback: fn ($model) => self::log($model, 'updated'));

        self::deleted(callback: function ($model) {
            in_array('Illuminate\Database\Eloquent\SoftDeletes', (class_uses(self::class))) ?
                "This is a softDeleted or a forceDeleted so don't log" :
                self::log($model, 'deleted');
        });

        if (in_array('Illuminate\Database\Eloquent\SoftDeletes', (class_uses(self::class)))) {
            self::softDeleted(callback: fn ($model) => self::log($model, 'soft deleted'));
            self::restored(callback: fn ($model) => self::log($model, 'restored'));
            self::forceDeleted(callback: fn ($model) => self::log($model, 'force deleted'));
        }
    }
    // ...
}
```

As you can see I tackled the issue of deleting by checking for the `SoftDeletes` Trait. But couldn't find a similar solution for restoring.

Then I decided to check why `runSoftDelete()` does not emit the **update** event while it does the same thing as `restore()`, setting `updated_at` to `CURRENT` and `deleted_at` to `CURRENT` or `null`, and found out that `runSoftDelet()` uses directly the `Builder::update()`, on the counterpart `restore` goes through this chain of action:

1. `Model::restore()` on the `SoftDeletes` trait.
2. ---- `restoring` is fired.
3. ---- `Model::save()`
4. -------- `saving` is fired.
5. -------- `Model::performUpdate()`
6. ------------ `updating` is fired.
7. ------------ `Builder::update()`
8. ------------ **`updated` is fired**.
9. -------- `Model::finishSave()`
10. ------------ `saved` is fired.
11. ---- `restored` is fired.

So my edit tries to clone the behavior of  `runSoftDelete()` into `runRestore`.

And I'm sure that many devs are adding too much hacky code when listening to events to execute an accurate response.

### DIFF
--- a/src/Illuminate/Database/Eloquent/SoftDeletes.php
+++ b/src/Illuminate/Database/Eloquent/SoftDeletes.php
@@ -114,18 +114,40 @@ trait SoftDeletes
             return false;
         }
 
+        $this->runRestore();
+
+        return true;
+    }
+
+    /**
+     * Perform the actual restore query on this model instance.
+     *
+     * @return void
+     */
+    public function runRestore()
+    {
+        $query = $this->setKeysForSaveQuery($this->newModelQuery());
+
+        $columns = [$this->getDeletedAtColumn() => null];
+
         $this->{$this->getDeletedAtColumn()} = null;
+
+        if ($this->timestamps && !is_null($this->getUpdatedAtColumn())) {
+            $time = $this->freshTimestamp();
+
+            $this->{$this->getUpdatedAtColumn()} = $time;
+
+            $columns[$this->getUpdatedAtColumn()] = $this->fromDateTime($time);
+        }
 
         // Once we have saved the model, we will fire the "restored" event so this
         // developer will do anything they need to after a restore operation is
         // totally finished. Then we will return the result of the save call.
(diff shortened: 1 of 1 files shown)


driesvints:
Tests are failing here

medilies:
@driesvints how can I run the tests locally to see how to fix that? and did you consider the idea of the PR?

driesvints:
Run it through Docker with `./bin/test.sh`. I don't consider PR's, only Taylor does. Please resend this with passing tests.

PR #43477, opened 2022-07-29
[9.x] Fix Missing exception variable in `Arr.php`

karam-mustafa:
Fix missing exception variable


fragkp:
Catch variables are optional since PHP 8.0 - so no need here for this change.

karam-mustafa:
@fragkp , yes this was checked recently, I was installing Laravel on php 7, so if it's not necessary, the Laravel team will close it

PR #43481, opened 2022-07-30
[9.x] Generate table's columns as phpDoc properties for `make:model` command

sethsandaru:
Hi all,

This PR introduces a new option `--phpdoc` for the `php artisan make:model` command. 

It will generate the Model file **together** with the phpDoc's properties - the table's columns.

It would be a great helper for IDEs and for us to know the Models' columns. Makes the development process around Eloquent more enjoyable. Quite helpful for some tables that had a lot of columns since the beginning, we don't have to add the `@property` manually, takes time.

Usage:

```bash
php artisan make:model User --phpdoc
```

Without `--phpdoc` - same as today:

image

With `--phpdoc`:

image

Note: this feature requires `doctrine/dbal` package.

Cheers!

### Checklists
- [x] Implement feature
- [x] Unit testing
- [x] Test


medilies:
Will it support refreshing the docs after changing the migration ?

sethsandaru:
@medilies that's what I'm about to do next 😄

foremtehan:
Wish this happen for Facades @method docblocks.

taylorotwell:

sethsandaru:
Thanks for the information @taylorotwell . 

But I believe it is a small helper/nice-to-have feature and it wouldn't add a big burden for the `make:model` command, we also have unit testing, can we process? Thanks 😄

PR #43483, opened 2022-07-30
Add support for BackedEnums in Blueprint::enum() method

taylormhicks90:
This allows you to pass the class name of a Backed Enum to the Blueprint::enum() method as well as an array. It now supports both cases below when creating an enum column.
```
Schema::table('table_name', function (Blueprint $table) {
    $table->enum('my_enum',MyEnum::class);
    $table->enum('another_enum',['foo','bar']);
});
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "^8.0.2",
+        "php": "^8.1",
         "ext-mbstring": "*",
         "ext-openssl": "*",
         "doctrine/inflector": "^2.0",
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "^8.1",
+        "php": "^8.0.2",
         "ext-mbstring": "*",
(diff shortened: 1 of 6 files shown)


taylormhicks90:
This requires php 8.1 for enum support. Should I instead submit a PR for the master branch for a future release?

BrandonSurowiec:
Wouldn't submit another PR, it will be closed. You can read why below.

> Database migrations referencing PHP enums has been declined many times: https://github.com/laravel/framework/pull/40662 https://github.com/laravel/framework/pull/40913 https://github.com/laravel/framework/pull/41026
> 
> Refer to this comment why that's the case: https://github.com/laravel/framework/pull/40662#issuecomment-1023048467
> 
> _Originally posted by @derekmd in https://github.com/laravel/framework/issues/41861#issuecomment-1091039790_

taylormhicks90:
@BrandonSurowiec thanks for the reply. #40662 makes complete sense and I had the same concern with using the enum but was tired of hard coding the migration after writing the enum. Guess hard coding it is the best way though.

PR #43484, opened 2022-07-30
[9.x] Create config files from artisan CLI

aknEvrnky:
Hi everybody,
this PR targets creating config files from artisan CLI. Since we could create almost every files from artisan, creating the config files were missing. We can create config files now.

## usage:

```shell
php artisan make:config theme
```

## result
In base directory, config/theme.php file is created. Now we can fill inside how we want.
```php
<?php

return [
    'default' => env('ADMIN_THEME', 'main'),

    'themes' => [
        'main' => [
            'admin_color_1' => '#188ea4',
            'admin_color_2' => '#312cb8',
            'admin_color_3' => '#1325a3',
            'admin_login_image' => '/admin/bgs/0.jpg',
            // ...
        ],
    ]
];

```

Tests are run, nothing is broken. Codes are generated like other <code>php artisan make</code> commands.

Main purpose of code is that, we should be able to create configs from CLI.


src/Illuminate/Foundation/Console/stubs/config.stub
```diff
@@ -0,0 +1,14 @@
+<?php
+
+return [
+
+    /*
+    |--------------------------------------------------------------------------
+    | {{ class }} Config File
+    |--------------------------------------------------------------------------
+    |
+    | This file is for storing the credentials of the {{ class }} application
```
GrahamCampbell, line 10: Typically one would put these in the services.php file, so this doesn't make for a good example. Moreover, I think it's rare app developers would want to make another config file, since the main use case for config is exactly the one you identified - and that is already serviced by the services.php config file (pun intended :P).

GrahamCampbell (changes requested):

src/Illuminate/Foundation/Console/stubs/config.stub
aknEvrnky: Yes, it's true but of course the moment that developer want to create config file will be occured. Since laravel ecosystem includes lots of make command, the config file should be included. That was the main purpose.

ankurk91:
Why did you clear cache in the command?

`{{class}} ` is not being populated?

aknEvrnky:
> Why did you clear cache in the command?
> 
> `{{class}} ` is not being populated?

My thought is that if you are creating a config file, your cached config will not include your new config. Maybe clearing the cache can be optional.

taylorotwell:

PR #43485, opened 2022-07-30
[9.x] improve finding model namespace in Model:show command

alissn:
### hi
in this PR update method `qualifyModel` of `model:show` command to improvement finding the models.

in this method first find with `glob` function in models folder, After finding the matched files, return **exact namespace** of the file.

actually if model namespace failure to follow PSR-4 rule, old version can't show model detail because **namespace generated manually** ,  but in new method if matched file name with argument, return namespace of model to continue command processing.

thanks.

### DIFF
--- a/src/Illuminate/Foundation/Console/ShowModelCommand.php
+++ b/src/Illuminate/Foundation/Console/ShowModelCommand.php
@@ -481,13 +481,23 @@ class ShowModelCommand extends Command
 
         $rootNamespace = $this->laravel->getNamespace();
 
-        if (Str::startsWith($model, $rootNamespace)) {
+        $pattern = is_dir(app_path('Models'))
+            ? app_path('Models').DIRECTORY_SEPARATOR.$model.'.php'
+            : app_path().DIRECTORY_SEPARATOR.$model.'.php';
+
+        $modelPath = glob($pattern);
+
+        if (! count($modelPath)) {
             return $model;
         }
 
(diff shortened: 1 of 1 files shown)


jessarcher:
Hi @alissn, could you please provide reproduction steps for what does not currently work?

The current implementation is based on the same logic that is used by the [generator commands](https://github.com/laravel/framework/blob/55826147af150a51ab5ca19c1be689cc222ac387/src/Illuminate/Console/GeneratorCommand.php#L217-L232) such as `make:model`, with a small tweak to support passing a full model namespace.

alissn:
Sometimes it happens the namespace of model failure to follow PSR-4 rule like bellow : 
the path of model is :` app/Models/User.php`
```
<?php

namespace App;

class User extends \Modules\User\Entities\User
{
    //
}
```

if you use `php artisan model:show User` 
the previous method return : `App\Models\User` 
but new method return : `App\User` (this correct namespace of user model )

 Sometimes you can't change namespace of model... this PR handle this issue for searching model name based on File name.

rodrigopedra:
> Sometimes it happens the namespace of model failure **to follow PSR-4**

I don't think that should be supported. Also how are you loading this model's class in other files? By requiring/including it directly? Or by using a custom autoloader?

Loading classes without composer autoload seems to be a very edgy case, it would make generator commands very hard to maintain if they should cover every unusual use case.

taylorotwell:

PR #43487, opened 2022-07-31
[9.x] Improve getter for morphed model alias

sawirricardo:
Background, I uses polymorphic models heavily in apps, which at times requires to do something like
```
Relation::enforceMorphMap(['user' => 'App\Models\User']);

Model::query()->whereModelType('user');
```

now, what if we updated the morph map from 'user' to 'other_user'?

Of couse we can imagine how many code we need to change.

```
// current way
User::make()->getMorphClass(); // we can do this better
User::morphAlias(); // 😍 returns 'other_user'
Relation::getMorphedModelAlias('App\Models\User') // 👍 also returns 'other_user'
```

Of course, suggestions are welcomed

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
@@ -749,7 +749,7 @@ trait HasRelationships
      *
      * @return int|string|false
      */
-    public static function morphMapAlias()
+    public static function morphAlias()
     {
         return Relation::getMorphedModelAlias(static::class);
     }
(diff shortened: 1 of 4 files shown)


taylorotwell:

JamesHemery:
Hi @taylorotwell,

What do you think about just adding this method on `HasRelationships` trait ? 

```php
    public static function morphAlias()
    {
        return (new static)->getMorphClass();
    }
```

It's a small DX improvement.

_I also thought of changing the visibility of getMorphClass as protected, so that it could be called by __call and __callStatic from Model, BUT the method is mocked in several places and this would be a major change._

PR #43495, opened 2022-08-01
[9.x] PHP 8.2 Support

Jubeki:
I don't think there is much to explain.

I extended the matrix to run tests with PHP 8.2 on linux and windows.
These tests are marked as `continue-on-error` so that there should be no problems with normal pull requests.

Additional Note:
PHP 8.2 will be released ...
- ... after Laravel 6 supports ends. Should the tests still be run there?
- ... before Laravel 8 support ends. Tests probably should also be run there. Should I send the PR for Laravel 8 instead?

You can simply close this PR if you think it is too early to start testing against PHP 8.2

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,8 +3,8 @@ name: tests
 on:
   push:
   pull_request:
-  # schedule:
-  #   - cron: '0 0 * * *'
+  schedule:
+    - cron: '0 0 * * *'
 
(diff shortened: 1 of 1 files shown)


driesvints:
Thank you but we'll start testing once the first release candidate is out.

driesvints:
Decided to start fresh here: https://github.com/laravel/framework/pull/43910

PR #43496, opened 2022-08-01
Fix typo in BelongsToMany

jnoordsij:
Note: I figured this should be `at` instead of `to`, because the method is also present in [`MorphToMany`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Relations/MorphToMany.php#L167).


driesvints:
Thanks!

Issue #23298, opened 2018-02-26
Laravel can't access pivot additional column from pivot model itself


shabaz-ejaz:
- Laravel Version: 5.5.35
- PHP Version: 7.0.10
- Database Driver & Version: MySQL 5.7.14

### Description:
I have a pivot model called UserTask in which I have an accessor function:

```
class UserTask extends Pivot implements HasMedia
{
    use HasMediaTrait;

    public function getCompletedAttribute()
    {
        return $this->getMedia()->isEmpty() && $this->completed;
    }



    public function task()
    {
        return $this->belongsTo(Task::class);
    }

}
```

I specify the relation in my Task model like this:

```
class Task extends Model
{
    public function users()
    {
        return $this->belongsToMany(User::class, 'user_task')->using('App\Models\UserTask')->withPivot('completed');
    }
}
```

I get the following error:

> "message": "Undefined property: App\Models\UserTask::$completed",

Anyone know why this is happening?

kevdotbadger:
Don't you need to use `->withPivot('completed');` to ask Eloquent to get this attribute?

staudenmeir:
Do you still have that problem?

CyberiaResurrection:
@staudenmeir , I've also tripped over this problem - to both cast the pivot values as a model and to also be able to access pivot model columns directly (without having to do an additional $PIVOT_MODEL::find($pivot_id) ), I have to use both ->using(... and withPivot(.... , like so:
```
return $this->belongsToMany(ForeignModel::class, 'pivot_table)->using(PivotModel::class)->withPivot(['firstCol', 'secondCol', .... ])

```

staudenmeir:
@CyberiaResurrection Can you please provide more details? What exactly is the issue?

CyberiaResurrection:
If I define the pivot table with additional columns but don't cast the pivot values to a model, then I get a bunch of values in $foo->manyToManyRelation()->first()->pivot and not a model.  This is problematic when said pivot models also have their own, independent relations to other models (not just the implied belongsTo relations to the parent models on each side of the original M:N relation).

If I go the other way, and cast the pivot values to a model but don't specify the extra columns, that usually breaks the extra relations where they are defined on the pivot models, as the requisite key value is usually missing.

As (Morph)Pivot extends the Model class, I don't see a reason why, if my application setup needs it, to refrain from adding those extra relations directly onto pivot models.

[...]

PR #43497, opened 2022-08-01
[9.x] Add pivot default columns

jnoordsij:
Currently, loading columns for a `Pivot` class in a many-to-many relation requires manually adding all column names on the relation method. This is [slightly confusing sometimes](https://github.com/laravel/framework/issues/23298), and moreover leads to some code duplication if you want to query the same pivot columns from both sides of the relation.

This PR adds a `withColumn` property to the `Pivot` class, where one can mark columns that should be loaded on all relations referencing the pivot.

Alternative approaches: 
- use the `attributes` property, though this would require adding a default value for all columns
- use the `casts` property, though this would require adding a cast-type for all columns

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php
@@ -334,6 +334,10 @@ class BelongsToMany extends Relation
     {
         $this->using = $class;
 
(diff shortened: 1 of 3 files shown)


driesvints:
Seems tests fail here. Please resend this with passing tests. Thanks

PR #43498, opened 2022-08-01
[9.x] Updates `dd|dump` color palette

nunomaduro:
This pull request updates the `dump` global function CLI color palette output (that's also used by `dd`). It's inspired by the Material Community theme, and the only difference is the `false` and `true` having different colors.

Screenshot 2022-08-01 at 16 21 03

### DIFF
--- a/src/Illuminate/Foundation/Console/CliDumper.php
+++ b/src/Illuminate/Foundation/Console/CliDumper.php
@@ -45,7 +45,7 @@ class CliDumper extends BaseCliDumper
 
         switch ($type) {
             case 'boolean':
-                $style = sprintf('boolean-%s',  $value = $value ? 'true' : 'false');
+                $style = sprintf('boolean-%s', $value = $value ? 'true' : 'false');
                 break;
         }
 
--- a/src/Illuminate/Foundation/Console/CliDumper.php
+++ b/src/Illuminate/Foundation/Console/CliDumper.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace Illuminate\Foundation\Console;
+
+use Symfony\Component\VarDumper\Cloner\Cursor;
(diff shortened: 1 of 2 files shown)


jeremynikolic:
Hi @nunomaduro 

f this may help, I've given this a try on Windows (php 8.0) both standard cmd line and phpstorm terminal.
From what I can tell, it looks great.

Cmd line:
image

PhpStorm terminal: 
image

Happy to try some specific scenarios, let me know 👍

lupinitylabs:
The numbers seem to have the same color like false and null now. Is this intentional?

AdamGaskins:
> I want to share some considerations here. I've heard multiple times already that people prefer the old color palette instead of the new one. The old colors are brighter, easier to read and (imo) make a more coherent color palette.
> 
> Do with it what you want, but the new colors isn't _necessarily_ for all people an improvement.


I see your point that color schemes are inherently subjective, but I think this new theme is much easier on the eyes. In the old theme every character had a bright color making everything visually the same weight, but in the new theme structure characters like `"`, `{`, `[`, `=>` etc. are a subtle gray, greatly increasing readability.

ralphjsmit:
> I see your point that color schemes are inherently subjective, but I think this new theme is much easier on the eyes. In the old theme every character had a bright color making everything visually the same weight, but in the new theme structure characters like `"`, `{`, `[`, `=>` etc. are a subtle gray, greatly increasing readability.

Sure, they are subjective. I can understand your argument for the 'subtle gray, which increases readability'. That's indeed one side of the coin. The other side is that for example the highlighted string quotes `"` underscore the _type_ of the value and not the value itself. I know that it's hard to please everyone, since it's subjective, but because of that I probably makes sense to give people a choice and not force them over.

taylorotwell:
I may table this for now since early reactions are pretty mixed. Will revisit later.

PR #43499, opened 2022-08-01
Added property for email verified at name

AntonLindell:
Allows for overriding the email verified at column name from the user model.


taylorotwell:

AntonLindell:
@taylorotwell Hey Taylor, thank you for the quick review. 

As the framework allows for not using snake_case in Eloquent models, and also columns like remember_token on the User model allows for overriding the rememberTokenName, I figured this addition to the MustVerifyEmail-class would make sense as it is the only snake_case attribute that is forced upon the user.

PR #43504, opened 2022-08-01
[9.x] Added createMany method to HasManyThrough

lee-to:
Maybe I didn't understand the concept of HasManyThrough relationship and therefore it seemed strange to me that there was no "createMany" method! After all, it fits perfectly, works and has practical importance (in my opinion)! If not, then I will be glad to know why this method is not available in HasManyThrough! It would be useful to me in the project, if it is still needed, then I can modify it and cover it with tests from above


taylorotwell:
I think because it doesn't really buy you much convenience. You still have to provide the foreign key.

PR #43505, opened 2022-08-02
[9.x] Add support for CarbonImmutable

mateusjatenee:
This is more of an idea...   

@michaeldyrynda has a [great blog post](https://dyrynda.com.au/blog/laravel-immutable-dates/) on how to use `CarbonImmutable` as the default Carbon class in Laravel. The problem is that swapping that class means the `Date` facade docblocks become invalid, and things such as the following start to be picked up by IDE and static analysis:

```php    
someMethod(now()); // wrong type notice
someMethod($request->date('date')); // wrong type notice. Expects DateTimeImmutable but was given Carbon

function someMethod(CarbonImmutable $date) 
{
  //
 }
```   

I wonder if we could add some sort of first-party support to immutable carbon. Let me know what you think. Thanks in advance!


taylorotwell:
Let me think on this a bit more.

PR #43509, opened 2022-08-02
Update RouteRegistrar.php

Ken-vdE:
Allow a route groups to also call the dynamic where methods described in https://laravel.com/docs/9.x/routing#parameters-regular-expression-constraints

E.g.
```php
Route::prefix('/{something}')->whereIn('something', ['foo', 'bar'])->group(function() {});
```


Ken-vdE:
@taylorotwell This does not work and should be reverted. It seems the RouteRegistrar::__call() method does not pass the parameters correctly. The passed array is omitted there. I don't see a clear solution other then adding another `if` exception for the 'whereIn' just like the `if ($method === 'middleware') {` statement in that __call method.
Do you think that would be an acceptable solution?

ankurk91:
We should not accept PRs without tests.

PR #43516, opened 2022-08-02
[9.x] Fix unique locking on broadcast events

DougSisk:
This is a fix for #43416. The lock was not being acquired because the broadcast manager was pushing straight to the queue and not using the dispatcher. Acquiring the lock within the `queue` method seems like the best option, as rewriting to use a dispatcher seems like it could cause a ton of headaches.


taylorotwell:
Now `uniqueVia` is not respected right?

DougSisk:
I think we can eliminate the `UniqueBroadcastEvent`, as it's not needed since we can generate the lock key from the event. I think there's also a potential issue with having `uniqueVia` be a property, as I believe the cache repository can't be serialized.

PR #43517, opened 2022-08-02
Make 'APP_KEY=' on the fly if it doesn't exist

salmanarshad321:


taylorotwell:
I'm not longer making changes to this command.

salmanarshad321:
> I'm not longer making changes to this command.

In the absence of APP_KEY, artisan incorrectly states that the key was set successfully. I think either it should be made clear the key wasn't set, or APP_KEY should be generated on the fly.

PR #43519, opened 2022-08-02
[9.x] Add Pint Command to Artisan Commands

alexanderkroneis:
Hi 👋

I was wondering since Pint was implemented into Laravel why it's not part of Artisans commands. That's why I decided to implement this command. It's not adding any functionality to be honest just allow developers to use `php artisan pint` instead of `./vendor/bin/pint`.


nunomaduro:
Sorry, no plans to add Pint as "artisan" command for now.

PR #43520, opened 2022-08-02
[9.x] Add `Config::collect()` method

PHLAK:
When retrieving configuration values sometimes the value needs to be converted to a collection for further processing. This is trivial to do by wrapping the `config()` call with `Collection::make()` or `collect()`, however this feels unnecessarily verbose.

This PR introduces the `Config::collect()` method to combine these actions and allow retrieving a configuration value as a collection in one call. Also, since `Config::collect()` returns an instance of `Collection` this allows fluent chaining for cleaner, more enjoyable code.

#### Implementation Details

  - `Config::collect()` will return an empty collection by default if the configuration value is not found by it's key
  - The `$default` parameter accepts any [`iterable`](https://www.php.net/manual/en/language.types.iterable.php) value including an array, a `Collection` object or an object that implements the `Traversable` interface
  - When retrieving a non-array configuration value it will be wrapped and returned as a collection (same behavior as `Collection::make()`)

#### Given a confiuration file of `config/test.php`

```php
return [
    'array' => ['foo', 'bar', 'baz'],
    'associative' => [
        'foo' => 123,
        'bar' => 456,
    ],
    'text' => 'foobar',
    'boolean' => true,
];
```

```php
// Retrieve a configuration value as a collection
$words = Config::collect('test.array'); // Collection<'foo', 'bar', 'baz'>

// Retrieve a value and fluently chain collection methods
Config::collect('test.array')->map(fn ($word) => Str::title($word))->toArray(); // ['Foo', 'Bar', 'Baz']

// Wrap a non-array value in a collection on retrieval
Config::collect('test.boolean'); // Collection<true>

// Retrieve with default values for non-existent items
Config::collect('non.existent.item', ['apple', 'orange']); // Collection<'apple', 'orange'>
```

---

Comments and criticisms welcome. Let me know if there's any way I can improve this PR.

### DIFF
--- a/src/Illuminate/Config/Repository.php
+++ b/src/Illuminate/Config/Repository.php
@@ -5,6 +5,7 @@ namespace Illuminate\Config;
 use ArrayAccess;
 use Illuminate\Contracts\Config\Repository as ConfigContract;
 use Illuminate\Support\Arr;
+use Illuminate\Support\Collection;
 
 class Repository implements ArrayAccess, ConfigContract
 {
@@ -72,6 +73,18 @@ class Repository implements ArrayAccess, ConfigContract
         }
 
         return $config;
+    }
+
+    /**
+     * Get the specified configuration value as a Collection.
+     *
+     * @param  string  $key
+     * @param  iterable  $default
+     * @return Collection
+     */
+    public function collect($key, $default = [])
+    {
+        return Collection::make(Arr::get($this->items, $key, $default));
     }
 
     /**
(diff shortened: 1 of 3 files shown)


taylorotwell:

PHLAK:
Hey @taylorotwell, as a package maintainer myself I totally understand the sentiment of maintenance overhead and don't really expect to change your mind, however, the fact that the `Config` facade is _not_ macroable makes implementing this by an end-user not really possible. Also, for what it's worth, I don't believe this is a substantial change that would require much (if any) maintenance overhead. Anyway, I'd really appreciate it if this were re-considered but, again, totally understand. No hard feelings either way.

PHLAK:
Actually, a quick tests shows that adding the `Macroable` trait to the `Illuminate/Config/Repository` class is all that's needed to make it macroable. Would you be willing to accept a PR to make `Config` macroable?

PR #43523, opened 2022-08-03
Revert "Update RouteRegistrar.php"

Ken-vdE:
Reverts laravel/framework#43509

@taylorotwell This does not work and should be reverted. It seems the RouteRegistrar::__call() method does not pass the parameters correctly. The passed array is omitted there. I don't see a clear solution other then adding another if exception for the 'whereIn' just like the `if ($method === 'middleware') {` statement in that __call method.
Do you think that would be an acceptable solution?


taylorotwell:
If there is a better solution feel free to PR.

ollieread:
@Ken-vdE did you ever figure out a solution for this? I'm happy to take a look.

Ken-vdE:
@ollieread I haven't gotten around to it yet. Have at it ;)

ollieread:
> @ollieread I haven't gotten around to it yet. Have at it ;)

Have added a new PR addressing this @Ken-vdE #43731 43731

PR #43527, opened 2022-08-03
[9.x] Add failing test for unregistered blade directive output

JayBizzle:
We have quite a few `blade` templates for emails that include the css `@media` query such as...

```css
@media only screen and (min-width:480px) {
```

When these get rendered, blade outputs the following...

```diff
- @media only screen and (min-width:480px) {
+ @media  only screen and (min-width:480px) {
```

Notice the extra space between `@media` and `only`

I know we can (and currently do) escape the `@media` string like so - `@@media` but we don't always remember, and it seems like escaping should only be required for directives that _are actually_ directives.

I had a look at a potential fix for this, but didn't get very far, so thought I would create this PR with a failing test to...
1) See if it should be fixed
2) See if anyone more clued up than me had any suggested fixes

Thanks

### DIFF
--- a/src/Illuminate/View/Compilers/BladeCompiler.php
+++ b/src/Illuminate/View/Compilers/BladeCompiler.php
@@ -512,14 +512,16 @@ class BladeCompiler extends Compiler implements CompilerInterface
     protected function compileStatement($match)
     {
         if (str_contains($match[1], '@')) {
-            $match[0] = isset($match[3]) ? $match[1].$match[3] : $match[1].$match[2];
+            $match[0] = isset($match[3]) ? $match[1].$match[3] : $match[1];
         } elseif (isset($this->customDirectives[$match[1]])) {
             $match[0] = $this->callCustomDirective($match[1], Arr::get($match, 3));
         } elseif (method_exists($this, $method = 'compile'.ucfirst($match[1]))) {
             $match[0] = $this->$method(Arr::get($match, 3));
+        } else {
+            return $match[0];
         }
 
-        return $match[0];
+        return isset($match[3]) ? $match[0] : $match[0].$match[2];
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Illuminate/View/Compilers/BladeCompiler.php
```diff
@@ -512,14 +512,14 @@ protected function compileStatements($value)
...
         if (str_contains($match[1], '@')) {
-            $match[0] = isset($match[3]) ? $match[1].$match[3] : $match[1];
+            $match[0] = isset($match[3]) ? $match[1].$match[3] : $match[1].$match[2];
         } elseif (isset($this->customDirectives[$match[1]])) {
             $match[0] = $this->callCustomDirective($match[1], Arr::get($match, 3));
         } elseif (method_exists($this, $method = 'compile'.ucfirst($match[1]))) {
             $match[0] = $this->$method(Arr::get($match, 3));
         }
 
-        return isset($match[3]) ? $match[0] : $match[0].$match[2];
```
taylorotwell, line 522: Can you full explain this change?

JayBizzle: I'll try...

For the string `@media only screen and (min-width:480px) {`

The regex in the `compileStatements()` function returns the following matches...

```
^ array:3 [
  0 => "@media "
  1 => "media"
  2 => " "
]
```

`matches[0]` contains the matched directive including a trailing space. This was originally returned along with the space that was matched in `$matches[2]`. Infact, all we actually need to do in this instance is return `$matches[0]`. None of the `if` statements in the `compileStatement` function match, so the directive is just returned as-is, which is exactly what we want.

I `dd()`'d all the `$matches` across the entire test suite to see in which scenarios we actually need to return the extra trailing space.

It seems that the only time the trailing space in `matches[2]` needs to be returned is when a directive has been escaped...

The following string `@@continue` outputs the following matches...

```
^ array:3 [
  0 => "@@continue "
  1 => "@continue"
  2 => " "
]
```

Here, the string we actually want to output is `$matches[1]`, but that does not contain the trailing space, so we need to add that from `$matches[2]`, The first `if` statement in the `compileStatement` function now handles this.

If you need any further explanation, please let me know 👍

taylorotwell:
@JayBizzle what do you think of the change I made? All tests still seem to pass.

JayBizzle:
Yes, that change you made has the same effect, so all looks good to me 👍

PR #43528, opened 2022-08-03
[9.x] Add specific queue parameter when pending mail

TyperEJ:
it's easier to use Mail facade to dispatch different queue.

`Mail::to($users)->queue(new Mailable(), 'mail');`


taylorotwell:
https://laravel.com/docs/9.x/mail#pushing-to-specific-queues

PR #43529, opened 2022-08-03
[9.x] Make more errors in validation error response translatable

djlimix:
Right now, when validation fails and there is more than 1 errors, it returns a message like this:
```
Name is required. (and 1 more error)
```

However, let's say that you've translated validation messages. Then, the message could be something like this:
```
Meno je požadované. (and 1 more error)
```

The issue is that there's no way to translate the ```(and 1 more error)``` string and so you end up with this half-translated error message, which is not ideal.

This pull request is a response to #41230.


driesvints:
See many many previous attempts.

PR #43533, opened 2022-08-03
fix: select maximum supported timestamp precision

### LINKED ISSUE #41154: Cannot rename a timestamp column "Too-big precision 10 specified"
- Laravel Version: 9.1.0
- PHP Version: 8.0.5
- Database Driver & Version: Mysql 8.0.19

### Description:
Attempting to use `$table->renameColumn('service_starts_at', 'sort_date');` on a timestamp column (created using `$table->timestamp('service_starts_at')->nullable();`) fails with the following error

> SQLSTATE[42000]: Syntax error or access violation: 1426 Too-big precision 10 specified for 'sort_date'. Maximum is 6. (SQL: ALTER TABLE funerals CHANGE service_starts_at sort_date TIMESTAMP(10) NULL DEFAULT NULL)

### Steps To Reproduce:

1. Add a new `timestamp()` column to a table via a migration.
2. Attempt to rename the created table in a later `Schema::table`

nandi95:
When renaming a timestamp column it does not correctly gets the precision using the `TIMESTAMP()` function in the alter statement.
While I did not investigate why we're getting too big precision to beging with, I made sure that we're not passing in higer precision than supported by the platforms.
There are no tests as test currently are using sqlite I think. It has however fixed the issue I had with renaming.
Can you please advise how this is tested with mysql? (I can see workflows referencing it)

The test would be something akin to
```php
public function testItDoesNotSetPrecisionHigherThanSupportedWhenRenamingTimestamps()
{
    $this->db->connection()->getSchemaBuilder()->table('users', function (Blueprint $table) {
        $table->timestamp('created_at');
    });
    $this->db->connection()->getSchemaBuilder()->table('users', function (Blueprint $table) {
        $table->renameColumn('created_at', 'new_created_at');
    });
    // Illuminate\Database\QueryException::class was not thrown
}
```

Rest of the changes are updates advised by deprecations.
Please see linked issue for context.

closes #41154

### DIFF
--- a/src/Illuminate/Database/DBAL/TimestampType.php
+++ b/src/Illuminate/Database/DBAL/TimestampType.php
@@ -4,6 +4,14 @@ namespace Illuminate\Database\DBAL;
 
 use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
+use Doctrine\DBAL\Platforms\MariaDb1027Platform;
+use Doctrine\DBAL\Platforms\MySQL57Platform;
+use Doctrine\DBAL\Platforms\MySQL80Platform;
+use Doctrine\DBAL\Platforms\MySQLPlatform;
+use Doctrine\DBAL\Platforms\PostgreSQL100Platform;
+use Doctrine\DBAL\Platforms\PostgreSQL94Platform;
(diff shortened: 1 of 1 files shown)


taylorotwell:

Thanks!

PR #43535, opened 2022-08-03
[9.x] Fix: select maximum supported timestamp precision

nandi95:
On the latest version (9.23.0) when renaming a timestamp column, it does not correctly gets the precision when using the `TIMESTAMP()` function in the generated alter statement. `$table->renameColumn('created_at', 'old_created_at');` would result something similar to:
```
Illuminate\Database\QueryException
SQLSTATE[42000]: Syntax error or access violation: 1426 Too big precision 10 specified for 'old_created_at'. Maximum is 6....
```

To replicate, rename a timestamp while using sql server, postgres or mysql.

**NOTE:** In testing, I cannot replicate the issue as the test suit seems to be ran with sqlite which ignores the precision and uses `DATETIME`. If the testing workflow also run the tests using the above platforms, then they won't fail due to my changes and this was not covered before. In any case my migration now succesfully runs after this change.

This PR ensures when the inferred precision value is higher than what the platform supports, it defaults back to the highest supported value.

Rest of the changes are updates advised by deprecations (using classnames as oppoed to deprecated `->getName()` and match the abstract function's argument name).

Issue where this was previously reported: https://github.com/laravel/framework/issues/41154

### DIFF
--- a/src/Illuminate/Database/DBAL/TimestampType.php
+++ b/src/Illuminate/Database/DBAL/TimestampType.php
@@ -4,14 +4,12 @@ namespace Illuminate\Database\DBAL;
 
 use Doctrine\DBAL\Exception as DBALException;
 use Doctrine\DBAL\Platforms\AbstractPlatform;
-use Doctrine\DBAL\Platforms\MariaDb1027Platform;
-use Doctrine\DBAL\Platforms\MySQL57Platform;
+use Doctrine\DBAL\Platforms\MariaDBPlatform;
 use Doctrine\DBAL\Platforms\MySQL80Platform;
 use Doctrine\DBAL\Platforms\MySQLPlatform;
-use Doctrine\DBAL\Platforms\PostgreSQL100Platform;
-use Doctrine\DBAL\Platforms\PostgreSQL94Platform;
+use Doctrine\DBAL\Platforms\PostgreSQLPlatform;
 use Doctrine\DBAL\Platforms\SqlitePlatform;
-use Doctrine\DBAL\Platforms\SQLServer2012Platform;
+use Doctrine\DBAL\Platforms\SQLServerPlatform;
 use Doctrine\DBAL\Types\PhpDateTimeMappingType;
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't think I'm going to mess with this on a patch release.

nandi95:
> patch release

Should I target master then? Or what should be done to get this bug fixed?

In the meantime for people who are trying to rename timestamp columns (or order them for example):

```php
// when timestamps already exist

// create temporary timestamps
Schema::table('my_table', static function (Blueprint $table) {
    $table->timestamp('temp_created_at')->nullable();
    $table->timestamp('temp_updated_at')->nullable();
});
// copy data to temporary stamps
DB::table('my_table')->update([
    'temp_created_at' => DB::raw('created_at'),
    'temp_updated_at' => DB::raw('updated_at'),
]);
// drop original timestamps
Schema::table('my_table', static function (Blueprint $table) {
    $table->dropColumn(['created_at', 'updated_at']);
});
// create new timestamps (at the right order/with the right name)
Schema::table('my_table', static function (Blueprint $table) {
    $table->timestamp('created_at')->after('column_i_want_this_after')->nullable();
    $table->timestamp('updated_at')->after('created_at')->nullable();
});
// update new timestamps
DB::table('my_table')->update([
    'created_at' => DB::raw('temp_created_at'),
    'updated_at' => DB::raw('temp_updated_at'),
]);
// remove temporary timestamps
Schema::table('my_table', static function (Blueprint $table) {
    $table->dropColumn(['temp_created_at', 'temp_updated_at']);
});
```

satoved:
Also a simpler fix:
```
\Illuminate\Support\Facades\DB::unprepared('alter table my_table change old_field_name new_field_name timestamp null;');
```

Or for Mysql 8.0:
```
\Illuminate\Support\Facades\DB::unprepared('ALTER TABLE my_table RENAME COLUMN old_field_name TO new_field_name;');
```

nandi95:
@taylorotwell Should I re-open this targeting master or forget about it?

PR #43544, opened 2022-08-04
change switch statement with match expression

Mehran-tr:
Change switch statement with PHP 8 match expression.


Mehran-tr:
test fails

PR #43546, opened 2022-08-04
[9.x] Add configurable 'hot' file path to Vite

iamgergo:
I'm using Vite for one of my packages as well, instead of mix. However the `hot` file path is not configurable, and it make not possible to run the Vite dev server correctly from a package.

-----
My package comes with pre-bundled assets, and they are being published to `public/vendor/package-name` when running composer update. For development I just make a symlink to fake the published status of the assets.

However with Vite I can't run the dev server properly from the package itself:

```sh
cd packages/conedevelopment/package-name

# works well
yarn build

# runs, but does not connect to the dev server
yarn dev
```

In my package base layout I use the following `@vite` blade directive:

```blade
@vite('resources/js/app.js', '/vendor/package-name/build')
```

By customizing the `build` directory path, it allows to have access to the published or symlinked assets for production and development as well.

However right now it's not possible to configure the path to the `hot` file. When running the `yarn dev` command from the package, the hot file is placed in the `packages/conedevelopment/package-name/public/hot` that is linked to the `public/vendor/package-name/hot` path.

-------

This PR attempts to make configurable the path to the `hot` file.

So we could do the following:

```blade
@vite('resources/js/app.js', '/vendor/package-name/build', '/vendor/package-name/hot')
```

------

This PR is only additive, should not hold any BC.

Also, I was not sure how to add tests for this, but if you are willing to merge this, I can add some tests as well.

### DIFF
--- a/src/Illuminate/Foundation/Vite.php
+++ b/src/Illuminate/Foundation/Vite.php
@@ -110,19 +110,21 @@ class Vite
      *
      * @param  string|string[]  $entrypoints
      * @param  string  $buildDirectory
+     * @param  string  $hotPath
      * @return \Illuminate\Support\HtmlString
      *
      * @throws \Exception
      */
-    public function __invoke($entrypoints, $buildDirectory = 'build')
+    public function __invoke($entrypoints, $buildDirectory = 'build', $hotPath = 'hot')
     {
         static $manifests = [];
 
         $entrypoints = collect($entrypoints);
         $buildDirectory = Str::start($buildDirectory, '/');
+        $hotPath = Str::start($hotPath, '/');
 
-        if (is_file(public_path('/hot'))) {
-            $url = rtrim(file_get_contents(public_path('/hot')));
+        if (is_file(public_path($hotPath))) {
+            $url = rtrim(file_get_contents(public_path($hotPath)));
 
             return new HtmlString(
                 $entrypoints


iamgergo:
A test failed, but I don't think it has anything to do with this PR.

driesvints:
@iamgergo re-ran that for you 👍

taylorotwell:
@jessarcher please mark as ready for review once you have taken a look 👍

jessarcher:
My only concern with adding this as the third parameter is that some folks also want to configure the manifest file name as well. If we were to add that, I think it should be the third parameter because it's related to the second parameter (`$buildDirectory`).

Any thoughts @timacdonald?

iamgergo:
@jessarcher @timacdonald @taylorotwell 

Just an idea then: What if the second parameter is a string by default, which determines the path to the `build` directory. Alternatively, we allow the developer to pass an array, with the paths.

```blade
{{-- Default, No BC, will cast it into an array in the Vite class --}}
@vite('resources/js/app.js', '/vendor/package-name/build')

{{-- Passing an array --}}
@vite('resources/js/app.js', [
    'buildDirectory' => '/vendor/package-name/build',
    'hotPath' => '/vendor/package-name/hot',
    'manifestpath' => '/vendor/package-name/...',
])
```

In the `Vite` class there are several ways to manage this, but with this approach we can stay with two parameters instead of 3 or 4 (or more?).

If you aggree with the concept I can update the PR, will see if you like it.

Ps.: I find named arguments a bit overkill here, personally I prefer passing an array. But it might be simpler in some cases thats for sure.

timacdonald:
My main concern with this is that I believe the signature change is a breaking change.

I believe that his is tied together with the manifest file naming, as @jessarcher mentioned, and we need to come up with a holistic approach for everything here if we wanna tackle it.

jessarcher:
I like that @timacdonald.

My initial concern with doing this was that people wouldn't be able to specify different hot file locations on a per-template basis for the scenario where they have multiple builds, but your solution looks to solve this without adding more parameters to the `@vite` directive :clap:

iamgergo:
Then would it be a good idea using an inline blade component that wraps the functionality that @timacdonald suggests?

So we could have a fresh new 'Vite' instance in its constructor and passthru the component props to the Vite methods.

This would be more elegant for me, but it feels a bit weird to have a directive and a component for the same thing.

timacdonald:
I don't think we should have a component. This is really just an escape hatch, so I don't want to build infrastructure around it that might promote it further than just that.

I would personally prefer that we just enable these affordances on the Vite class (and docblock on the facade) and allow the inline building approach.

timacdonald:
I do wanna brainstorm what we call the named method that aliases `__invoke`.

- `toHtmlTags($entrypoints)`
- `toEntryPoints($entrypoints)`
- Other ideas?

We can't use `toHtml` as that is confusing with the `Htmlable` contract, and we need to be able to pass through the entrypoints to the function (I don't feel we should have a named method for entry points).

```blade
{{ Vite::useHotFile('path/to/hot')->toHtmlTags('resources/js/app.js') }}

{{ Vite::useHotFile('path/to/hot')->toEntrypoints('resources/js/app.js') }}
```

iamgergo:
I think `toEntryPoints($entrypoints)` is nice and expressive. Some other alternatives that came in my mind:

- `toAssets($entrypoints)`
- `toBuild($entrypoints)`
- or just simply `build($entrypoints)`
- `render` but it has the same issue as `toHtml`

timacdonald:
Lets roll with `toEntryPoints` for now, but also pinging @jessarcher for naming input.

@iamgergo are you happy to make the suggested changes in this PR or would you like me to send up some commits to your branch?

We will need a PR for the Vite plugin as well to allow configuring the hot path.

timacdonald:
@iamgergo After chatting with Jess, we have decided the following is the way forward.


1. New func `withEntryPoints` that accepts entry points and is merged into an array on the class.
2. New func `useHotFile` that accepts a path.
3. Make the Vite class `Htmlable`, with `toHtml` invoking the class.

```blade
{{ Vite::withEntryPoints(['app.js'])->useHotFile('path/to/hot') }}
```

These functions can then be called in any order. Let me know if you would like any assistance. Happy to help out and get this across the line.

iamgergo:
@timacdonald okay, sure I'm happy to make a PR.

I'll submit it soon, then we'll see what to change/modify.

So I'm closing this for now.

Thank you!

PR #43553, opened 2022-08-04
[9.x] Fix bullet output transparency in Artisan commands

reziamini:
The transparency of the bullet list was not as good as it should be, especially in IDEs' terminal with dark mode and the readability is 

Before:
before

After:
Screenshot 2022-08-04 230803


taylorotwell:
Marking as draft until @nunomaduro reviews.

nunomaduro:
Not sure what terminal schema are you using. But we have multiple output ( event:list, and route:list, etc ) using gray, and so far, we have no faced issues here.

So, it may be something specific to your theme. Do you mind of considering updating your terminal schema so we don't have to update this for everyone?

Screenshot 2022-08-05 at 09 46 37

PR #43556, opened 2022-08-04
[10.x] Remove trailing slash from ASSET_URL

hansnn:
Update: The description below is outdated (and doesn't match the title of the PR). See the discussion in the comments below.

Setting `ASSET_URL` to an empty string in your env file currently yield asset urls that are prefixed your `APP_URL` (e.g. `<APP_URL>/path/to/asset`).

This PR changes this behaviour so that setting `ASSET_URL` to an empty string yield asset urls that are prefixed with a single slash (i.e. `/path/to/asset`).

The need for this change came up when moving from Mix to Vite.
Mix always output the assets with a single slash in front (i.e. `/path/to/asset`). However, Vite uses the `asset()` helper, which by default prefixes the asset urls with the `APP_URL` env variable (i.e. `<APP_URL>/path/to/asset`). This becomes a problem when trying to access the site using different hostnames (usually locally during testing and such). A solution would be to remove the `<APP_URL>` part of the asset urls, but as far as I can tell that's not possible without the changes in this PR.

This will be a breaking change for anyone that has currently set `ASSET_URL` to an empty string.

### DIFF
--- a/src/Illuminate/Routing/UrlGenerator.php
+++ b/src/Illuminate/Routing/UrlGenerator.php
@@ -253,7 +253,7 @@ class UrlGenerator implements UrlGeneratorContract
         // Once we get the root URL, we will check to see if it contains an index.php
         // file in the paths. If it does, we will remove it since it is not needed
         // for asset paths, but only for routes to endpoints in the application.
-        $root = $this->assetRoot ?: $this->formatRoot($this->formatScheme($secure));
+        $root = $this->assetRoot ?? $this->formatRoot($this->formatScheme($secure));
 
         return $this->removeIndex($root).'/'.trim($path, '/');
     }
--- a/src/Illuminate/Routing/UrlGenerator.php
+++ b/src/Illuminate/Routing/UrlGenerator.php
@@ -253,7 +253,9 @@ class UrlGenerator implements UrlGeneratorContract
         // Once we get the root URL, we will check to see if it contains an index.php
         // file in the paths. If it does, we will remove it since it is not needed
         // for asset paths, but only for routes to endpoints in the application.
-        $root = $this->assetRoot ?? $this->formatRoot($this->formatScheme($secure));
+        $root = $this->assetRoot
+            ? Str::of($this->assetRoot)->rtrim('/')
+            : $this->formatRoot($this->formatScheme($secure));
 
         return $this->removeIndex($root).'/'.trim($path, '/');
     }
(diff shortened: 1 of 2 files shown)


jessarcher:
At the moment if you set `ASSET_URL` to `/` then you get URLs like `//path/to/asset`, which I don't think is useful and feels unexpected. I wonder whether it would be better to make it so that setting an `ASSET_URL` of `/` would give you URL's like `/path/to/asset`.

Is there a reason you can't set `APP_URL` to something appropriate for your environment?

hansnn:
> At the moment if you set ASSET_URL to / then you get URLs like //path/to/asset, which I don't think is useful and feels unexpected. I wonder whether it would be better to make it so that setting an ASSET_URL of / would give you URL's like /path/to/asset.

I wonder the same. I actually already created [a PR with that approach](https://github.com/laravel/framework/pull/43555), but closed it in favour of this one. One of the downsides of that PR is that it will be impossible to get the double slash asset urls (e.g. `//path/to/asset`). You will either get a single slash by setting `ASSET_URL` to `/` or three slashes by setting it to `//`. Not that I think the double slash is useful, but for all I know someone needs it 🤷🏻 

> Is there a reason you can't set APP_URL to something appropriate for your environment?

My `APP_URL` is simply `localhost` when working locally, which works fine most of the time. There are a couple of cases where it doesn't work, though:

1. I use docker for everything. Including Cypress testing. My Cypress container needs to access my laravel/nginx container, and does so over the "docker network" by going to `http://my-nginx-container-name`. This is of course not the same as `APP_URL` and hence the asset urls won't work in that context.
2. I sometimes access my local site using my mobile device. In that case the url I access is `http://my-machines-ip-address`, this is also of course not the same as `APP_URL`.

jessarcher:
What if it was made so that an `ASSET_URL` of `//` output `//path/to/asset`?

Basically, if `ASSET_URL` ends with a `/`, then you'd remove any leading slashes from the path before joining them together?

hansnn:
Great suggestion! I implemented it in the latest commit.

hansnn:
I realized that my reasoning behind this change was a bit flawed. My use case is quite specific in that I don't need the `APP_URL` to match the url I'm using to access the application, but the "correct" solution would probably anyway be to make sure that I access the application using the `APP_URL`. Which I also eventually found a way to do (using the "network_mode: host" option in Docker).

While I think this change could makes sense (especially after incorporating @jessarcher's input), I'll close the PR as I'm not convinced that it's worth anyone else spending their time on it. I'l happily reopen it if anyone thinks it makes sense.

PR #43560, opened 2022-08-05
[9.x] Add possibility to run code before and after a command

alexanderkroneis:
This PR introduces a possibility to run code before and after a command.

An example of using this would be to create a Laravel package which provides a set of commands and there is some specific logic to execute before and/or after every command of this package. In some of these cases `$this->error` and `$this->components->error` are required, which are not available in the constructor.

**MakeDataObjectCommand.php**
```
class MakeDataObjectCommand extends Command
{
    protected $signature = 'package:make:data {name}';

    protected $description = 'Make a new data object';
    
    protected $requiredPackages = [
        'spatie/laravel-data',
    ];

    public function before(): void
    {
        $composer = Composer::getInstance();

        if ($composer->packagesAreInstalled($this->requiredPackages) === false) {
            $this->components->error('There are some missing dependencies.');
        }
    }

    {
        // ...
    }
    
    public function after(): void
    {
        // ...
    }
}
```

### DIFF
--- a/src/Illuminate/Console/Command.php
+++ b/src/Illuminate/Console/Command.php
@@ -134,9 +134,19 @@ class Command extends SymfonyCommand
      */
     protected function execute(InputInterface $input, OutputInterface $output)
     {
+        if (method_exists($this, 'before')) {
+            $this->laravel->call([$this, 'before']);
+        }
+
         $method = method_exists($this, 'handle') ? 'handle' : '__invoke';
 
-        return (int) $this->laravel->call([$this, $method]);
+        $code = (int) $this->laravel->call([$this, $method]);
+
+        if (method_exists($this, 'after')) {
+            $this->laravel->call([$this, 'after']);
+        }
+
+        return $code;
     }
 
     /**


jmarcher:
What is the gain of doing the same for the specific command like:
```php
class MakeDataObjectCommand extends Command
{
    protected $signature = 'package:make:data {name}';

    protected $description = 'Make a new data object';
    
    protected $requiredPackages = [
        'spatie/laravel-data',
    ];

    public function before(): void
    {
        $composer = Composer::getInstance();

        if ($composer->packagesAreInstalled($this->requiredPackages) === false) {
            $this->components->error('There are some missing dependencies.');
        }
    }

    {
        $this->before();
        // ... Your code
        $this->after();

        return $code;
    }
    
    public function after(): void
    {
        // ...
    }
}
```

alexgaal:
We are developing a package with e.g. 20 commands, which are executing the same before logic every time. If I add this to `handle` it would hurt DRY principle.

**Command.php**
```
class Command extends BaseCommand {
    public function before()
    {
        $this->check();
    }
    
    public function check() {
        // ...
    }
}

class MakeDataObjectCommand extends Command
{
    public function handle() {
        // ...
    }
}
```

It would decrease code repetition in this case and also important: It increases DX with a low amount of code.

rodrigopedra:
Wouldn't adding listeners for `Illuminate\Console\Events\CommandStarting` and `Illuminate\Console\Events\CommandFinished` suffice?

rodrigopedra:
Also the base Symfony Command class already provides an empty `Command@initialize` method meant to be overridden.

rodrigopedra:
Also, as it seems your package will use a base command to avoid DRY, if you prefer this approach you can override the base command's `execute` method on your base command.

alexgaal:
I am pretty sure it would also work with overwriting `execute` or implement these listener. It's just a different approach and I think tbh it's the easiest one without breaking anything.

taylorotwell:

PR #43563, opened 2022-08-05
Factory `for` method support for dynamic relationship name with existing model instance 

hassanhe:
Since `for` method already support dynamic relationship name for state, we can support passing model instance directly and infer the relationship name from the call:
```PHP
// Before
$postFactory->for($model, 'author');

// After
$postFactory->forAuthor($model);
```


taylorotwell:

PR #43564, opened 2022-08-05
[9.x] Fix type for `instance` container method

olivernybroe:
A small type fix for `instance` method on Container class


driesvints:
I think it was said that right now we're not pursuing these kind of generics other than in collections. /cc @nunomaduro

nunomaduro:
Dries is correct. Plus, it's complicated make container resolvers generic, as in theory, you can have stuff like `app('request')` resolving an instance of `Illuminate\Http\Request`.

olivernybroe:
Ah alright, no worries 👍

PR #43565, opened 2022-08-05
New method Str::end added. Should do the opposite of Str::start.

DerFichtl:
Surprised me that Str::end method is missing, because there is a method Str::start and also Str::startsWith and Str::endsWith both versions are implemented. On purpose missing?

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -892,6 +892,20 @@ class Str
     }
 
     /**
+     * End a string with a single instance of a given value.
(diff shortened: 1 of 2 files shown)


adevade:
The method is called `finish()`, instead of `end()`.

https://github.com/laravel/framework/blob/5a25794264b0947df277abb7b33ec9c490e4dc14/src/Illuminate/Support/Str.php#L319-L331

DerFichtl:
Nice. Then this can be closed.

PR #43568, opened 2022-08-05
adding an attribute to HasAttributes Class to hide elements from obse

omer-hanaraey:
…rver

the purpose of that is when you observe a model and you don't want to get the changes of some columns like "updated_at" or some other fields. you can specify it from its Model and it will not show when you call $mode->getDirty().


omer-hanaraey:
@taylorotwell Unfortunately it is not correcting any bug from the framework, but it gives a better ability to the framwork only by adding a small condition.

I already faced this problem and in every observer class, I should remove all unnecessary fields after calling `getDirty` function. this pull request makes it simple. only by adding a protected field for the Model and adding those column names that you don't want to be observed when it is changed.

PR #43569, opened 2022-08-05
[9.x] Eloquent Accessing Nested Array Value

dbpolito:
Fix inconsistence for dealing with nested array on Eloquent.

Currently you can set nested array values, but not access them:

```php
$model = new Model;
$model->{'a->b'} = 'something'; // this works

dump($model->{'a->b'}) // this will return null
```

This PR fixes this behavior.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -447,6 +447,13 @@ trait HasAttributes
             return;
         }
 
+        // If this attribute contains a JSON ->, we'll get the proper value in the
+        // attribute's underlying array. This takes care of properly nesting an
+        // attribute in the array's value in the case of deeply nested items.
+        if (str_contains($key, '->')) {
+            return $this->getJsonAttribute($key);
+        }
+
         return $this->getRelationValue($key);
     }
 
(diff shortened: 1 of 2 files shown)


morloderex:
@dbpolito I feel like you are missing a json cast or an ArrayObject cast for this attribute. Am i missing something?

dbpolito:
You mean here? https://github.com/laravel/framework/blob/9.x/tests/Database/DatabaseEloquentModelTest.php#L2451 add `meta` as an `array` cast?

We can definitely do that, but it won't impact this test anyway... i did notice that and i didn't add because there are other tests already related to json field, like: https://github.com/laravel/framework/blob/9.x/tests/Database/DatabaseEloquentModelTest.php#L1228

But i can change to use this other model instead: https://github.com/laravel/framework/blob/9.x/tests/Database/DatabaseEloquentModelTest.php#L2792

dbpolito:
Just updated my tests, let me know if you also want to update this one https://github.com/laravel/framework/blob/9.x/tests/Database/DatabaseEloquentModelTest.php#L1228-L1245 to use the proper model with cast too...

timacdonald:
Can't you just use `$model->a->b`?

taylorotwell:

dbpolito:
Well this isn't a bug per say but it's an inconsistency which has bitten me...

We can access by `$model->a->b` but let's say you're using a dynamic variable:

```php
$path = 'a->b';

$model->{$path} = 'something'; // works

$model->{$path}; // doesn't work
```

Setting an attribute this way works, but accessing it doesn't work... which creates an inconsistency...

In my case i used this to set stuff several times, and got the assumption that accessing would also work, which didn't.

Maybe we should remove the support to setting attribute this way.

@taylorotwell 👆

PR #43588, opened 2022-08-08
Update helpers.php

bilogic:
It takes quite a bit of effort to track down `abort(404)` or other aborts when developing more complex packages. This help track down the source quickly if we are in debug mode.


bilogic:
Ok, it is more complicated than I thought since I manipulated the original `$code`

PR #43589, opened 2022-08-08
[9.x] Allow using static closures as callable for macros

paulbalandan:
This PR allows using static closures as the callable when registering a macro.
```php
TestMacroable::macro('speed', static fn () => 'I am speed.');
echo (new TestMacroable)->speed(); // 'I am speed.'
```

I am using PHP-CS-Fixer's [`static_lambda`](https://github.com/FriendsOfPHP/PHP-CS-Fixer/blob/master/doc/rules/function_notation/static_lambda.rst) rule. So, whenever a closure is not referencing `$this` in the closure body, it will be converted to static. This conversion, however, currently breaks macroable binding to the calling class since we'll get a `Cannot bind an instance to a static closure` error when the macro is called through `__call`.

### DIFF
--- a/src/Illuminate/Macroable/Traits/Macroable.php
+++ b/src/Illuminate/Macroable/Traits/Macroable.php
@@ -5,6 +5,7 @@ namespace Illuminate\Support\Traits;
 use BadMethodCallException;
 use Closure;
 use ReflectionClass;
+use ReflectionFunction;
 use ReflectionMethod;
 
 trait Macroable
@@ -118,7 +119,14 @@ trait Macroable
         $macro = static::$macros[$method];
 
         if ($macro instanceof Closure) {
-            $macro = $macro->bindTo($this, static::class);
+            $fn = new ReflectionFunction($macro);
+            $bindable = $fn->getClosureScopeClass() === null || $fn->getClosureThis() !== null;
+
+            if ($bindable) {
+                $macro = $macro->bindTo($this, static::class);
+            } else {
+                $macro = $macro->bindTo(null, static::class);
+            }
         }
 
         return $macro(...$parameters);
--- a/src/Illuminate/Macroable/Traits/Macroable.php
+++ b/src/Illuminate/Macroable/Traits/Macroable.php
@@ -119,8 +119,9 @@ trait Macroable
         $macro = static::$macros[$method];
 
         if ($macro instanceof Closure) {
-            $fn = new ReflectionFunction($macro);
-            $bindable = $fn->getClosureScopeClass() === null || $fn->getClosureThis() !== null;
+            $reflection = new ReflectionFunction($macro);
+
+            $bindable = $reflection->getClosureScopeClass() === null || $reflection->getClosureThis() !== null;
 
             if ($bindable) {
                 $macro = $macro->bindTo($this, static::class);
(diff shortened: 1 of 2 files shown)


serpentblade:
Reflection is an expensive operation to be added into a potentially critical code path.  This solution seems easy enough to support the need for static macros, but will add significant overhead to all macro calls, where we can assume most of them are not static closures as they weren't previously supported.

The better solution should be to determine at the time of registering the macro if it is static or not and modify the behavior of the call function based on that predetermined assessment.

driesvints:
We reverted this.

Tofandel:
ReflectionFunction is an inexpensive operation taking 0.3 microseconds that is cached on subsequent calls and taking 0.01 microsecond each other call after that (even in a different process)

`php -r '$start = microtime(true);new ReflectionFunction(fn () => true);microtime(true);echo ($end-$start)*1000;`

ollieread:
> Reflection is an expensive operation to be added into a potentially critical code path. This solution seems easy enough to support the need for static macros, but will add significant overhead to all macro calls, where we can assume most of them are not static closures as they weren't previously supported.
> 
> The better solution should be to determine at the time of registering the macro if it is static or not and modify the behavior of the call function based on that predetermined assessment.

I'd just like to add that reflection in PHP is not expensive at all. To run code, the engine will parse a class, function, closure, etc., into a Zend object, which is how it knows which methods, properties, etc. exist. All reflection does is instantiate an instance of the appropriate `Reflector` implementation and populate it. 

Source: I've made a few PRs to php-src for reflection and have spoken to the core devs about it; it's really not that bad, it's just presenting information already present in memory.

serpentblade:
There's still an overhead to creating that Reflector instance, making the calls to evaluate if the function is bindable, and then branching based on that evaluation.  My argument is that 100% of existing code doesn't need to evaluate for that at all.

Additionally, the performance test provided is not an accurate evaluation either. You should be timing the entirety of the code added, and not a single line. So that should include the creation of `ReflectionFunction`, the evaluation of `$bindable` and then the branching based on the value of `$bindable` to get an accurate read.  Additionally, `hrtime` should be used when measuring time variations instead of `microtime`.

But really this entire argument is moot anyway as `Macroable` already supports static methods as there's a __callStatic method implemented in the trait.

So the original code in this PR:
```php
TestMacroable::macro('speed', static fn () => 'I am speed.');
echo (new TestMacroable)->speed(); // 'I am speed.'
```

Can actually be called use when called like a static method:
```php
TestMacroable::macro('speed', static fn () => 'I am speed.');
echo TestMacroable::speed(); // 'I am speed.'
```

The nice takeaway here is that Laravel already supports what you're trying to do, no PR necessary.

paulbalandan:
I will not try to argue on the reflection. It seems you are more experienced on that field than I am.

My takeaway is that the requirement is to call the registered macro on a non-static context, but presented with the problem of binding a static closure. Yes, I am aware `Macroable` has a `__callStatic` method. That is a nice alternative since PHP supports calling a static method on an instance. But that is not the requirement wants. (If you're curious, the requirement was to register the macro on `Illuminate\Http\Request` class).

This argument is moot and academic anyway as this PR is already reverted. I just hacked my way to achieve the requirement.

PR #43590, opened 2022-08-08
Update 404.blade.php

bilogic:
Allow $message to show since `NotFoundHttpException(...);` accepts a message


driesvints:
You can overwrite this view if you want.

PR #43591, opened 2022-08-08
[9.x] Fix type for `flatmap` collection method

olivernybroe:
This adds types for when using flatmap in collections via phpstan. 

This makes use of a special phpstan feature called `value-of`. 

The reason why it still returns `mixed` if not iterable is because that is actually how `flatMap` works compared to `flatten` (might be a bug?) 

if you flatten vs. flatmap the following you get different results
image
however instead of looking into if this is a bug, I simply resolve the types correctly.

The reason why this pr does not contain types for flatten, is because I have not figured out how to solve the depth parameter yet.

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -386,8 +386,9 @@ trait EnumeratesValues
     /**
      * Map a collection and flatten the result by a single level.
      *
-     * @param  callable(TValue, TKey): mixed  $callback
-     * @return static<int, mixed>
+     * @template T
+     * @param  callable(TValue, TKey): T  $callback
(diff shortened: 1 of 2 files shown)


olivernybroe:
ping @nunomaduro

nunomaduro:
Because PHPStorm does not support conditional return types, we should not merge this. Otherwise, we would broke auto-completion for the sake of having better phpstan support.

Plytas:
Looks like conditional return types are now supported in [PhpStorm 2022.3.1](https://blog.jetbrains.com/phpstorm/2022/12/phpstorm-2022-3-1-is-released/)
image
@nunomaduro @olivernybroe

olivernybroe:
Interesting. 
@nunomaduro could this be taken into consideration again now that phpstorm supports it?

PR #43592, opened 2022-08-08
Add Arr::empty() helper

schulerj89:
This adds the `Arr::empty()` helper. This could provide a simpler way to check if an array has a empty key/value pair. It also supports dot notation.

Before
```php
$example = [
    'product' => [
        'name' => 'Product Name',
        'label' => 'Product Label',
        'slug' => '/product'
    ]
];

if (!empty($example['product']['label']) && !empty($example['product']['slug'])) {
    // do stuff
}
```

After
```php
use Illuminate\Support\Arr;

$example = [
    'product' => [
        'name' => 'Product Name',
        'label' => 'Product Label',
        'slug' => '/product'
    ]
];

if (!Arr::empty($example, ['product.name', 'product.slug'])) {
    // do stuff
}
```


taylorotwell:

PR #43594, opened 2022-08-08
[9.x] Improved check on long Blade strings

madbob:
Further improvement of #41956, as commented in #32254.

The full concatenation of a Component's Blade string AND the Views' folder path must be considered to avoid that a string longer than `PHP_MAXPATHLEN` characters arrives to `Filesystem::exists()`, triggering an error like `File name is longer than the maximum allowed path length on this platform`.


taylorotwell:
No plans to make further changes here.

PR #43598, opened 2022-08-08
[9.x] Make `Config` repository macroable

PHLAK:
Make the `Config` repository macroable so users can extend it in their own apps.

**Reasoning:** Since #43520 was closed without merging I was hoping to implement `Config::collect()` via a macro, however the `Config` repository is _not_ currently macroable.


driesvints:
Renamed this PR title to what it actually does.

PR #43599, opened 2022-08-08
[9.x] Add "trashing" event for soft deletes

imanghafoori1:
There is no event fired before trying to soft-delete a model.
This PR makes soft deletes consistent other operations on models which include two events before and after each operation.

- This event is only fired when the model is being "**soft-deleted**" and it is `not fired` when the model is being "**force-deleted**".

- Currently, the `deleteing` and `deleted` events are fired when soft-deleting models, but they are not exclusive to soft-deletes and are also fired when the model is being `force-deleted`.


imanghafoori1:
This PR was closed because the implementation is not fully correct.

PR #33368, opened 2020-06-28
[7.x] Include class name in output of GeneratorCommand

shalvah:
It's often useful (especially for Laravel newbies) to know exactly what files were created after running a command. This PR adds the name of the created class for `make:` and other generator commands, making the output more useful.

This shouldn't be a breaking change unless someone is asserting on the output of these commands, which would be a pretty flaky test (you should be checking if the expected files were created).

PR #43605, opened 2022-08-09
[9.x] add filename in output that is generated by generator commands

hAz5:
add filename in input that is generated by generator commands

before

before

after
after

### DIFF
--- a/src/Illuminate/Console/GeneratorCommand.php
+++ b/src/Illuminate/Console/GeneratorCommand.php
@@ -183,6 +183,8 @@ abstract class GeneratorCommand extends Command
         }
 
         $this->components->info($info.' created successfully.');
+
+        $this->components->warn(sprintf('Path: [%s]', $path));
     }
 
     /**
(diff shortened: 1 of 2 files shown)


timacdonald:
If we consider this PR, please see this comment (https://github.com/laravel/framework/pull/33368#issuecomment-656995736) for reasoning and why it might be good to actually output the entire path, not just the filename.

hAz5:
@timacdonald thank you for making me aware

this is the result of my last changes:

after2

@nunomaduro @timacdonald  Please If there is an issue, please let me know.

driesvints:
I don't think we should add a warning there... It also feels weird the migration doesn't have a full path but the other ones do.

hAz5:
> 

the migration command has --fullpath option but others don't have

can you tell me which tag I can use?

nunomaduro:
Feel free to re-submit this pull request. Yet, the output should be consistent with other commands, and affect all the `make:*` commands. As such, would be cool something like the following:

If single file created:
```

[ INFO ] Model [app/Models/Foo.php] created successfully.

```

If multiple files created:

```

[ INFO ] Model and Factory created successfully.

    ⇂ app/Models/Foo.php
    ⇂ database/Factories/FooFactory.php

```

PR #43606, opened 2022-08-09
Optimize when using static closures as callable for macros

huangdijia:
This PR optimize for  #43589


taylorotwell:
Going to revert the original PR for now.

Tofandel:
And otherwise see https://gist.github.com/mindplay-dk/3359812

There is no benefit to caching reflection, php already does it, what's expensive is the first reflection call only

PR #43610, opened 2022-08-09
[9.x] Support for BackedEnum in `value` helper

foremtehan:
I think it make sense when we pass BackedEnum to `value` function it should return 'value' of it not the enum itself.


taylorotwell:
Breaking change.

PR #43611, opened 2022-08-09
[9.x] Override notification display name

mixdf:
Currently, it's impossible to override the notification display name. The `\Illuminate\Notifications\SendQueuedNotifications::displayName` method simply returns the notification class name, which is not favorable in some cases. 

The expected behavior is to allow notifications to override display name, while maintaining backward compatibility.

This PR adds a simple check to allow notifications to override the displayed name, while reverting to the default class name for backward compatibility.


taylorotwell:
Why is it not favorable in some cases? That's how mailables work as well.

PR #43612, opened 2022-08-09
added the ability to assert using a json resource instead of fully wr…

Ken-vdE:
added the ability to assert a json response equals a used JsonResource instead of having to fully write-out an array.
Based on https://stackoverflow.com/a/61960799/3017716

E.g.
```php

        $model = SomeModel::factory()->create();
        $route = route('api.some-model.show', $model);
        $response = $this->getJson($route);
        $response->assertOk();

        $resource = new TestModelJsonResource($model);
        $response->assertJsonResource($resource);
```
```php
        $model = SomeModel::factory(2)->create();
        $route = route('api.some-model.index');
        $response = $this->getJson($route);
        $response->assertOk();

        $paginator = new LengthAwarePaginator($models, 2, 15, 1);
        $resource = TestModelJsonResource::collection($paginator);
        $response->assertJsonResource($resource);
```

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -10,6 +10,7 @@ use Illuminate\Database\Eloquent\Collection as EloquentCollection;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Http\RedirectResponse;
 use Illuminate\Http\Request;
+use Illuminate\Http\Resources\Json\JsonResource;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Collection;
@@ -40,6 +41,13 @@ class TestResponse implements ArrayAccess
     public $baseResponse;
 
     /**
+     * The request that was used to make the application return our response.
+     *
+     * @var \Illuminate\Http\Request
+     */
+    public $request;
+
+    /**
      * The collection of logged exceptions for the request.
      *
      * @var \Illuminate\Support\Collection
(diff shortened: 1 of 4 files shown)


src/Illuminate/Testing/TestResponse.php
```diff
@@ -57,23 +65,26 @@ class TestResponse implements ArrayAccess
      * Create a new test response instance.
      *
      * @param  \Illuminate\Http\Response  $response
+     * @param  \Illuminate\Http\Request  $request
      * @return void
      */
-    public function __construct($response)
+    public function __construct($response, $request)
```
driesvints, line 71: This is a breaking change.

src/Illuminate/Testing/TestResponse.php
```diff
@@ -57,23 +65,26 @@ class TestResponse implements ArrayAccess
...
 
     /**
      * Create a new TestResponse from another response.
      *
      * @param  \Illuminate\Http\Response  $response
+     * @param  \Illuminate\Http\Request  $request
      * @return static
      */
-    public static function fromBaseResponse($response)
+    public static function fromBaseResponse($response, $request)
```
driesvints, line 85: This is a breaking change.

driesvints:
Seems like there's several breaking changes in this PR. Please resend this without those or send this to master. Thanks

src/Illuminate/Testing/TestResponse.php
Ken-vdE: Yes, but it's not used anywhere else in the framework.

src/Illuminate/Testing/TestResponse.php
Ken-vdE: Yes, but it's not used anywhere else in the framework.

PR #43617, opened 2022-08-09
[9.x] foreignIdFor() with custom column name: use correct table, fixes #42519

### LINKED ISSUE #42519: Migration: foreignIdFor: $column argument doesn't make sense currently
- Laravel Version: 9.13.0
- PHP Version: 8.1.6
- Database Driver & Version: MariaDB 10.7.3 / PDO

### Description:

The `foreignIdFor()` method in migrations has a second argument `$column` that doesn't make sense in how it (not) works currently.

Imagine I have a `User` model and a `Post` model and a post has both a `author_user_id` and an `reviewer_user_id`. Intuitively, the migration would look like

```php
$table->foreignIdFor(App\Models\User::class, 'author_user_id')->constrained();
```

Which would be the short form for:

```php
$table->unsignedBigInteger('author_user_id');
$table->foreign('author_user_id')->references('id')->on('users');
```

Instead, it behaves as if the code had an `->on('author_users')`.

In my understanding, either the `$column` argument of the `foreignIdFor()` method is useless (because if the model is ignored, it's just the same as the `foreignId()` method - why make it available as another method as well?), or, rather, `foreignIdFor($model, $class)` should get the table name (here: `'users'`) from the first argument, from the model.

levu42:
Currently, `foreignIdFor(App\Models\User::class, 'author_user_id')->constrained()` calls create SQL that wrongly tries to guess the table name from the custom column name (`'author_user_id'` => `author_users`) instead of using the table name from the model (to get the correct `users` table in this example).

I explained the problem in a bit more detail in #42519.

This PR adds `ForeignIdColumnDefinition::$table` that can be set via `ForeignIdColumnDefinition::withTabl('users')`. This table name is used instead of guessing the table name when present. Also `Blueprint::foreignIdFor()` was changed to set the table name from the model.

There is one test for each of the 4 SQL grammars included for this usage of `foreignIdFor()` with a custom column name & subsequent `constrained()` call.

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -934,7 +934,7 @@ class Blueprint
         $foreignIdDefinition = $model->getKeyType() === 'int' && $model->getIncrementing()
                     ? $this->foreignId($column ?: $model->getForeignKey())
                     : $this->foreignUuid($column ?: $model->getForeignKey());
-        
+
         return $foreignIdDefinition->withTable($model->getTable());
     }
 
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -931,9 +931,11 @@ class Blueprint
             $model = new $model;
         }
 
-        return $model->getKeyType() === 'int' && $model->getIncrementing()
+        $foreignIdDefinition = $model->getKeyType() === 'int' && $model->getIncrementing()
                     ? $this->foreignId($column ?: $model->getForeignKey())
                     : $this->foreignUuid($column ?: $model->getForeignKey());
+        
+        return $foreignIdDefinition->withTable($model->getTable());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


taylorotwell:
If this convenience method doesn't work for you then you should use the `foreign` method to create the foreign key definition - it provides more flexibility.

levu42:
@taylorotwell It does work for me, but the method arguments currently don't make sense.

If I say `foreignIdFor(User::class, 'author_user_id')->constrained()` it generates a foreign ID definition that's linking to a non-existing table when it has all the info to link to the correct table.

What is the reason to provide a `foreignIdFor()` method that takes the class name and a custom column name when I can either have it not-constrained (then I don't need the class name at all) or I have it constrained but need to repeat the table name, even though it's possible to get it from the model?

If you don't want this flexibility in `foreignIdFor()`, because it's just a convenience method, then what's the point of having the custom column name argument there, if it generates wrong SQL when used with `constrained()`?

PR #43629, opened 2022-08-09
[9.x] Adding dot notation support to Arr::only helper

godismyjudge95:
This pull request adds support for dot notation which is supported elsewhere in the framework but currently is not supported via the `Arr::only()` helper.

Example:
```php
$array = ['taxonomies' => ['colors' => ['red', 'blue'], 'positions' => ['top', 'bottom']]];

$newArray = Arr::only($array, 'taxonomies.colors');
```

`$newArray` would then be equal to:
```php
['taxonomies' => ['positions' => ['top', 'bottom']]]
```

This also works with an array of dot notion keys:
```php
$newArray = Arr::only($array, ['taxonomies.colors', 'taxonomies.positions'])
```

---

A few notes:
- I have added new test assertions that should cover all new cases of using `Arr::only()`
- There are a few random changes that had to be made to the test suite where the tests were depending on the order of the array output of `Arr::only()`.  As the array order does not actually matter in any of these cases, I changed the tests to match the new order so they would pass.
- This should not cause any breaking changes as the output stays the same regardless of what is passed in for the `$keys` parameter (ie. a string, dot notation string, array of normal keys, or array of dot notation keys)

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -498,6 +498,7 @@ class Arr
                 static::set($newArray, $key, static::get($array, $key));
             }
         }
+
         return $newArray;
     }
 
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -490,11 +490,11 @@ class Arr
      */
     public static function only($array, $keys)
     {
-    	$keys = static::wrap($keys);
+        $keys = static::wrap($keys);
     	
-    	$newArray = array_intersect_key($array, array_flip($keys));
+        $newArray = array_intersect_key($array, array_flip($keys));
         foreach ($keys as $key) {
-            if (!static::exists($newArray, $key) && static::has($array, $key)) {
+            if (! static::exists($newArray, $key) && static::has($array, $key)) {
                 static::set($newArray, $key, static::get($array, $key));
             }
         }
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -491,7 +491,7 @@ class Arr
     public static function only($array, $keys)
     {
         $keys = static::wrap($keys);
-    	
+
         $newArray = array_intersect_key($array, array_flip($keys));
         foreach ($keys as $key) {
             if (! static::exists($newArray, $key) && static::has($array, $key)) {
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -490,7 +490,7 @@ class Arr
      */
     public static function only($array, $keys)
     {
-        if (!is_array($keys)) {
+        if (! is_array($keys)) {
             $keys = [$keys];
         }
 
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -490,24 +490,15 @@ class Arr
      */
     public static function only($array, $keys)
     {
-        if (! is_array($keys)) {
-            $keys = [$keys];
-        }
-
-        $newArray = [];
+    	$keys = static::wrap($keys);
+    	
+    	$newArray = array_intersect_key($array, array_flip($keys));
         foreach ($keys as $key) {
-            if (Arr::exists($array, $key)) {
-                $newArray[$key] = Arr::get($array, $key);
-            } elseif (Arr::has($array, $key)) {
-                Arr::set($newArray, $key, Arr::get($array, $key));
+            if (!static::exists($newArray, $key) && static::has($array, $key)) {
+                static::set($newArray, $key, static::get($array, $key));
             }
         }
-
-        $oldKeys = array_fill_keys(array_keys($array), null);
-        $orderedNewArray = array_replace($oldKeys, $newArray);
-        $filteredOrderedNewArray = array_intersect_key($orderedNewArray, $newArray);
-
-        return $filteredOrderedNewArray;
+        return $newArray;
     }
 
     /**
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -490,7 +490,24 @@ class Arr
      */
     public static function only($array, $keys)
     {
-        return array_intersect_key($array, array_flip((array) $keys));
+        if (! is_array($keys)) {
+            $keys = [$keys];
+        }
+    
+        $newArray = [];
+        foreach ($keys as $key) {
+            if (Arr::exists($array, $key)) {
+              $newArray[$key] = Arr::get($array, $key);
+            } else if (Arr::has($array, $key)) {
+                Arr::set($newArray, $key, Arr::get($array, $key));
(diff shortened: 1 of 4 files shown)


Tofandel:
@godismyjudge95 I really'd like to have dot notation for Arr::only but your assumption that the order doesn't matter or that it's not a breaking change is wrong 

If you had to change existing tests it definitely is a breaking change because it will also cause failing test for people relying on Arr::only in their tests

Order change could have some unexpected issues as well if you were for example using the values in their respecitve order but I do agree it's rare to have order matter when using keyed arrays, if you want to have it in 9.X you need to figure out a way to preserve the order

```php
echo join(' ', Arr::only(['I' => 'He', 'like' => 'loves', 'hates' => 'dislikes', 'mars' => 'twix'], ['mars', 'I', 'like']));
// Currently: He loves twix
// With your version: twix He loves
```
Yoda

godismyjudge95:
@Tofandel ah ok.  I didn't realize DSN was order sensitive.  I will look into how difficult/complicated it is to keep the order.  Was trying to keep the code as simple as possible but like you pointed out there are a few edge cases where array order matters.  I have just never assumed a string keyed array would be in the order I expect so I thought this would be fine.

Thanks for the review.

Tofandel:
No problem

It will be quite hard to keep the order, your other bet is to send it to the master branch so that it is included in 10.X as a breaking change

Here is my best demonstration


```php
$url = join('', Arr::only(['protocol' => 'https://', 'domain' => 'example.com', 'port' => '8000', 'path' => '/foo'], ['domain', 'protocol']);
// https://example.com
// vs
// example.comhttps://
```

Obviously we could even use the fact that now it's ordered based on the keys we passed, and it makes it more predictable and usefull, and I prefer that, but yes sadly a breaking change

$url = join('', Arr::only(['domain' => 'example.com', 'protocol' => 'https://', 'port' => '8000', 'path' => '/foo'], ['protocol', 'domain']);
// example.comhttps://
// vs
// https://example.com

godismyjudge95:
@Tofandel Good news!  I figured out a way to keep the order, have dot notation, not break any existing tests/code, and along the way found/fixed a bug where the key had a dot in it.

The pull request now only has changes to the Arr::only method and the related tests in SupportArrTest

Tofandel:
Nice, I'm reviewing it, and I noticed that yes you don't need to preserve the order of nested keys because that would be part of the new feature anyways that makes it easier to implement

src/Illuminate/Collections/Arr.php
```diff
@@ -490,7 +490,24 @@ public static function prependKeysWith($array, $prependWith)
...
+            } elseif (Arr::has($array, $key)) {
+                Arr::set($newArray, $key, Arr::get($array, $key));
+            }
+        }
+
+        $oldKeys = array_fill_keys(array_keys($array), null);
+        $orderedNewArray = array_replace($oldKeys, $newArray);
+        $filteredOrderedNewArray = array_intersect_key($orderedNewArray, $newArray);
+
+        return $filteredOrderedNewArray;
```
Tofandel, line 510:
```suggestion
    	$keys = static::wrap($keys);
    	
    	$newArray = array_intersect_key($array, array_flip($keys));
        foreach ($keys as $key) {
            if (!static::exists($newArray, $key) && static::has($array, $key)) {
                static::set($newArray, $key, static::get($array, $key));
            }
        }
        return $newArray;
```

tests/Support/SupportArrTest.php
```diff
@@ -492,9 +492,16 @@ public function testIsList()
...
         $array = ['name' => 'Desk', 'price' => 100, 'orders' => 10];
-        $array = Arr::only($array, ['name', 'price']);
-        $this->assertEquals(['name' => 'Desk', 'price' => 100], $array);
+        $this->assertEquals(['name' => 'Desk', 'price' => 100], Arr::only($array, ['price', 'name']));
         $this->assertEmpty(Arr::only($array, ['nonExistingKey']));
+
+        $array = ['taxonomies' => ['colors' => ['red', 'blue'], 'sizes' => ['small', 'medium', 'large']]];
+        $this->assertEquals(['taxonomies' => ['colors' => ['red', 'blue']]], Arr::only($array, 'taxonomies.colors', true));
+        $this->assertEquals(['taxonomies' => ['sizes' => ['small', 'medium', 'large']]], Arr::only($array, 'taxonomies.sizes', true));
+        $this->assertEquals($array, Arr::only($array, ['taxonomies.sizes', 'taxonomies.colors'], true));
```
Tofandel, line 501:
```suggestion
        $this->assertEquals(['taxonomies' => ['colors' => ['red', 'blue']]], Arr::only($array, 'taxonomies.colors'));
        $this->assertEquals(['taxonomies' => ['sizes' => ['small', 'medium', 'large']]], Arr::only($array, 'taxonomies.sizes'));
        $this->assertEquals($array, Arr::only($array, ['taxonomies.sizes', 'taxonomies.colors']));
```

Tofandel (commented):
I had good look at it and saw there was a way to make a more memory safe and performant implementation, by basically applying the old behavior and then just doing the dot notation on top

`array_fill_keys(array_keys) + array_replace` would have killed the memory for a large array by keeping 3 copies of the array in memory at the same time, it goes from a O(2n+2y) implementation to a O(2y) as well, so we don't depend on the size of the original array anymore and only of the keys

You should also use static and not Arr like in the rest of the methods

Good job on spotting up that edge case 🎊

tests/Support/SupportArrTest.php
Tofandel: Little unexisting param addition there

godismyjudge95:
> array_fill_keys(array_keys) + array_replace would have killed the memory for a large array by keeping 3 copies of the array in memory at the same time, it goes from a O(2n+2y) implementation to a O(2y) as well, so we don't depend on the size of the original array anymore and only of the keys

Wow, yeah didn't think about time complexity and memory usage on this - probably should get in the habit of doing so 😄 

> You should also use static and not Arr like in the rest of the methods

I know... haha this was left over from me experimenting on the laravel playground to get it to work.

Thanks for reviewing this btw, it really helps me learn a lot.

godismyjudge95:
Also apologies for all the commits (if you get notifications for them), not familiar with the Laravel formatting styles.

Tofandel:
Ha no worries, I always have the same issue with the styling, will be squashed anyways. Would be good to have an autocorrecting github action like I have seen on some spatie repos

Looks all good to me

taylorotwell:

godismyjudge95:
@taylorotwell Almost all the other Arr helper methods support dot notation so it is very confusing when Arr::only does not support it.  This request is more about being consistent in what the Arr helper methods support than adding a new feature.

Tofandel:
I also agree with this, almost all array helpers support the dot notation, it is a bit confusing when Arr::only doesn't

This is a good PR 🙏

hafezdivandari:
This also can be used on `Arr:pull` I opened a discussion for this: #43763

PR #43633, opened 2022-08-10
Return type declarations added

developersharif:


driesvints:
No plans to add these

PR #43637, opened 2022-08-10
[9.x] Mistake in console choice phpdoc

mro95:
Writing a console command, there is a function `$this->choice()`. The `$default` property should be the default key as described in the [Laravel documentation](https://laravel.com/docs/9.x/artisan#multiple-choice-questions), which is `int` and not `string`.


timacdonald:
When choice is passed an associated array, a string is a valid value here.

```php
$day = $this->choice(
    'What day is it?',
    [
       'mon' => 'Monday',
       'tues' => 'Tuesday'
    ],
    'mon'
);
```

Likely, it should be `null|string|int`

driesvints:
Thanks!

PR #43638, opened 2022-08-10
[9.x] Updated hydrate method to calculate array length once

egesu:
The method was counting the `$items` on each loop. I've added a variable to keep length of `$items` and pass it to the callback function instead of passing the full array. Just a little performance update.


Tofandel:
This is an unecessary optimisation, in php count is O(1) the size of the array is stored in the array and it takes the same time to retrieve it than to retrieve a variable

egesu:
Ah I didn't know that. Closing the MR then.

PR #43635, opened 2022-08-10
[9.x] Prevent error in db/model commands when using unsupported columns

joedixon:
The PR resolves #43630 #43622 

The `doctrine/dbal` package does not support enum types when listing database columns. This PR adds support for mapping `enum` and `sysname` (issue with some distributions of sqlsrv) to `string`. The Artisan `db:show`, `db:table` and `model:show` commands have all been updated to use it.

Additonally, `db:show` uses the `rescue` helper to prevent an error if the engine type is not available. This PR also prevents the error from being reported.

PR #43645, opened 2022-08-10
[9.x] fix mySQL enum support for db:show command

lupinitylabs:
### The Issue

Currently, the new `db:show` console command throws a `Doctrine\DBAL\Exception` if any table in the application uses `ENUM` columns on a MySQL database:

grafik

### The Fix

[The Doctrine Project addresses the issue with MySQL `ENUM`s in their documentation](https://www.doctrine-project.org/projects/doctrine-orm/en/2.13/cookbook/mysql-enums.html) and as a global workaround suggests mapping the type to `VARCHAR`s to mitigate the issue.

This PR registers such a type mapping dynamically within the command for all MySQL platforms.

### Further unsupported types

The same problem exists for `POINT` types, but I first wanted to see if there is an intent to address these issues.
If the fix is appreciated, I'm happy to add one for `POINT`s too.

### DIFF
--- a/src/Illuminate/Database/Console/ShowCommand.php
+++ b/src/Illuminate/Database/Console/ShowCommand.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Database\Console;
 
+use Doctrine\DBAL\Platforms\MySQLPlatform;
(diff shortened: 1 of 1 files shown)


lupinitylabs:
I am closing this for now in favor of #43635.

Issue #43113, opened 2022-07-09
Model `increment()` does not update timestamps immediately


bert-w:
- Laravel Version: 9.x
- PHP Version: 8.x
- Database Driver & Version: N/A?

### Description:
Calling `increment()` (or `decrement()` too probably) on an Eloquent model with timestamps enabled doesn't update the `updated_at` column immediately.

### Steps To Reproduce:
```php
// Get a random user object.
$user = User::find(1);
// Dump current `updated_at`.
dump($user->updated_at);
// Increment some numerical column.
$user->increment('credits', 10);
// Dump the new `updated_at`.
dump($user->updated_at);
```
Result: the second `updated_at` dump shows the same value.
Expected result: show the new `updated_at` value.

When calling `$user->refresh()` after the last dump (or checking the database), the `updated_at` column DOES appear to have been updated though.

I went through the code looking for a spot to fix it but I got lost at the syncing logic that writes results back to the model. I'm not sure how different the `increment()` function is to the `update()` one and if this bug is feasible to be fixed.

bert-w:
Line from `protected function incrementOrDecrement(...)`:
https://github.com/laravel/framework/blob/6dd688461ff7a402efc7f939267087ccafebcc96/src/Illuminate/Database/Eloquent/Model.php#L861

Just wanted to add that apparently any properties that you change during the `updating` event are not actually changed in the model. It is simply disregarded. There appears to be an `$extra` parameter for the `increment()` function that accomplishes that using a different route.

bert-w:
https://github.com/laravel/framework/blob/6dd688461ff7a402efc7f939267087ccafebcc96/src/Illuminate/Database/Eloquent/Model.php#L865

Does anything prevent us to use `$this->getDirty()` here for the `$extra` param (or merge them), similar to how it happens in the `update()` function?

milwad-dev:
No bug, I test increment & it's true & update (updated_at)
```
    public function increment($column, $amount = 1, array $extra = [])
    {
        return $this->toBase()->increment(
            $column, $amount, $this->addUpdatedAtColumn($extra)
        );
    }
```

driesvints:
I don't think there's a right way to solve this one. Here's an issue with more details on this: https://github.com/laravel/framework/issues/32567.

There's no way to know if the dirty values at the time of incrementing are going to be the values that need to be saved.

Issue #32567, opened 2020-04-28
Increment method doesn't trigger the updated


mcpraps:
- Laravel Version: 7.2.1
- PHP Version: 7.2.3
- Database Driver & Version: MySQL 5.6.47

### Description:
I'm not sure exactly if it'is a bug, but I think it is. If I create an observer and I want to catch the updated event, it doesn't work if I use the eloquent increase method on the specified model. So the increase method doesn't trigger the updated event. It's so annoying because it updating the updated_at field also by default. That's why I'm think it's a real update event, why it should to trigger the updated event.

### Steps To Reproduce:
Create a model with a counter field, increase this field and check the model's updated event in the the model's observer.

driesvints:
These methods indeed do not trigger these events. Please see other issues on this tracker.

mcpraps:
Thanks but the "bug" wasn't be "fixed". What kind of issues are you thinking of?

driesvints:
This isn't a bug. It's the current expected behavior.

driesvints:
Taylor let me know that we should look into fixing the behavior for increments/decrements on single models. This will need to be targeted to master as it's a breaking change. We'll need to keep the behavior for relationships as we can't emit events for mass updates.

MattStrauss:
Maybe I am looking at this naively, but can't we just add two lines to the `incrementOrDecrementAttributeValue` method ([here](https://github.com/laravel/framework/blob/f27c2192d05e941268522df2ebe30a7764027212/src/Illuminate/Database/Eloquent/Model.php#L611)):

```
  protected function incrementOrDecrementAttributeValue($column, $amount, $extra, $method)
    {
        $this->{$column} = $this->{$column} + ($method === 'increment' ? $amount : $amount * -1);

        $this->forceFill($extra);

        // add this line to fire the 'updating' event and give the dev a chance to stop the model update
        if ($this->fireModelEvent('updating') === false) {
            return false;
        };

        $this->syncOriginalAttribute($column);

        // add this line to fire the 'updated' event on model
        $this->fireModelEvent('updated', false);
    }
```
This will fire the updating and updated events when `increment` or `decrement` is called on a given model. If the dev increments a field on all models, the events should not be fired as the below line in the the `incrementOrDecrement` method ([here](https://github.com/laravel/framework/blob/f27c2192d05e941268522df2ebe30a7764027212/src/Illuminate/Database/Eloquent/Model.php#L587)) will catch that scenario:

```
if (! $this->exists) {

[...]

PR #43646, opened 2022-08-10
[10.x] Rework increment/decrement functions to use default `save()` routine

bert-w:
Should fix https://github.com/laravel/framework/issues/43113, https://github.com/laravel/framework/issues/32567

### Explanation
The `increment()` and `decrement()` functions of `Eloquent\Model` were using some verbose logic to make `updating`/`updated` events work. These functions use a special syntax to update a column table, which results in atomic queries which make sure that concurrent queries to the same column keep working consistently, e.g.:
```
UPDATE `users` SET `coins` = `coins` + 10 WHERE `id` = 5
````
These functions will now include proper handling of `updating`/`updated`/`saving`/`saved` events (the first 2 were formerly supported but somewhat hacked-in).

### `getDirty()` + `getDirtyExpectables()`

The main idea used is this: keep track of model attributes that are about to be updated (comparable to `$model->getDirty()`), but also keep a list of updates that contain calculations (like the statement `"coins" + 10"`). This uses a new class `ExpectableExpression` which extends from `Illuminate\Database\Query\Expression` (aka `DB::raw()`).

These statements are kept inside the `$model->getDirtyExpectables()` and read and flushed inside the `$model->performUpdate()` function, which sends them to the query builder which (already) automatically copies these expressions into the expected query format.

The `ExpectableExpression` uses an additional property called the `$expectedOutcome`, which is the best guess the application has about the outcome of the atomic expression. This expected value is then set as the model property before the update takes place (**this behaviour has not changed**). This workaround is required since the actual outcome of the increment/decrement functions is not known *until* you re-query the database and get the new updated value.

### [10.x]
I targeted the master branch because there are some breaking changes

### CODE REVIEW

**Comment by bert-w** [comment]
It just crossed my mind that this has an added bonus feature:
```php
// $model = ['internet_points' => 10, 'coins' => 5]
$model->update([
    'internet_points' => new AtomicExpression("internet_points + 10", 20),
    'coins' => new AtomicExpression("coins - 5", 0),
]);
```
To update 2 properties atomically in 1 query.  Tthe second parameter is the `$expectedOutcome` as described above.
Obviously this syntax could be greatly improved/simplified to something like:
```php
$model->update([
    'internet_points' => $model::atomic('+', 10),
    'coins' => $model::atomic('-', 5),
]);
```
You can even use multiply and divide operators if you have a use-case for that :)

src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -1053,11 +1055,23 @@ public function increment($column, $amount = 1, array $extra = [])
      */
     public function decrement($column, $amount = 1, array $extra = [])
     {
-        return $this->toBase()->decrement(
-            $column, $amount, $this->addUpdatedAtColumn($extra)
+        return $this->update(
+            $this->toBase()->createDecrementStatement(
```
Jubeki, line 1059:
```suggestion
            $this->createDecrementStatement(
```
You already have the `toBase()` call inside of `createDecrementStatement()`

src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -1038,8 +1038,10 @@ public function upsert(array $values, $uniqueBy, $update = null)
      */
     public function increment($column, $amount = 1, array $extra = [])
     {
-        return $this->toBase()->increment(
-            $column, $amount, $this->addUpdatedAtColumn($extra)
+        return $this->update(
+            $this->toBase()->createIncrementStatement(
```
Jubeki, line 1042:
```suggestion
            $this->createIncrementStatement(
```
You already have the `toBase()` call inside of `createIncrementStatement()`

bert-w: i've moved 2 functions to the `$passthru` variable so they can be called directly. However I've kept the `toBase()` call since I don't see it done for any of the other passthru functions.

tests/Database/DatabaseEloquentModelTest.php
```diff
@@ -1924,7 +1943,7 @@ public function testDecrementQuietlyOnExistingModelCallsQueryAndSetsAttributeAnd
         $model->publicDecrementQuietly('foo', 1, ['category' => 1]);
         $this->assertEquals(2, $model->foo);
         $this->assertEquals(1, $model->category);
-        $this->assertTrue($model->isDirty('category'));
+        $this->assertFalse($model->isDirty('category'));
```
bert-w, line 1946: I changed this behaviour since it's not expected that the property is still dirty after update

tests/Database/DatabaseEloquentModelTest.php
```diff
@@ -1860,17 +1881,15 @@ public function testIncrementOnExistingModelCallsQueryAndSetsAttribute()
...
+        $model->shouldReceive('newQueryWithoutRelationships')->andReturn($this->mockBuilderForModel($model));
 
         $model->publicIncrement('foo', 1);
         $this->assertFalse($model->isDirty());
 
         $model->publicIncrement('foo', 1, ['category' => 1]);
         $this->assertEquals(4, $model->foo);
         $this->assertEquals(1, $model->category);
-        $this->assertTrue($model->isDirty('category'));
+        $this->assertFalse($model->isDirty('category'));
```
bert-w, line 1892: I changed this behaviour since it's not expected that the property is still dirty after update

taylorotwell:

PR #43647, opened 2022-08-10
[9.x] Add joinRelation to Eloquent query builder

leandro-hermes:
When we want to add a join clause to an Eloquent query, we need to either give the raw table name, or instantiate the related model in order to get the table name dynamically.

This PR adds a way to add this join clause taking advantage of the relationship, by doing this:
```php
// using the relation name
$model->joinRelation('relationName');

// using the relation instance
$model->joinRelation($model->relation());
```

This was implemented for the following relation types: `BelongsTo`, `HasOne`, `HasMany`, `MorphOne`, `MorphMany`, `BelongsToMany`, `HasManyThrough` and `HasOneThrough`. Of course, for the last 3 type, two join clauses will be added.

If anyone thinks that there's a better way to do this, or there's anything wrong or missing, please let me know.

This PR doesn't add any breaking changes, as it only introduces new functionality.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -593,7 +593,6 @@ trait QueriesRelationships
      * @param  \Illuminate\Database\Eloquent\Relations\Relation|string  $relation
      * @param  string  $type
      * @param  bool  $where
-     *
      * @return $this
      *
      * @throws \Illuminate\Database\Eloquent\RelationNotFoundException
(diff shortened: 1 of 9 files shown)


taylorotwell:

leandro-hermes:
@taylorotwell Don't you think that this would add a nice functionality to the framework by reducing the need for hardcoding the names of the tables and columns while building queries?

leandro-hermes:
Or, do you think this could be done in a different way?

joostdebruijn:
Hi @leandro-hermes. Personally, I would like such a functionallity in the framework itself. However, there is a good alternative solution, you might want to have a look at this: https://github.com/kirschbaum-development/eloquent-power-joins.

leandro-hermes:
Great @joostdebruijn, thank you. I've never seen this package before, and I can see that it's certainly more powerful than my implementation.

PR #43651, opened 2022-08-11
[9.x] Disallow routes with duplicate URIs

mateusjatenee:
I recently had to deal with an annoying issue on an existing application where routes were being registered twice by mistake. 

This feature would be opt-in. 
I couldn't figure out how to deal with different domains (the failing test), but if you think this isn't a shit idea I can work on it and figure it out.

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -130,7 +130,7 @@ class Router implements BindingRegistrar, RegistrarContract
      *
      * @var bool
      */
-    protected static $enforceUniqueRoutes = false;
+    protected $enforceUniqueRoutes = false;
 
     /**
      * Create a new Router instance.
@@ -480,7 +480,7 @@ class Router implements BindingRegistrar, RegistrarContract
      */
     protected function validateDuplicatesIfNeeded(Route $newRoute)
     {
-        if (! static::$enforceUniqueRoutes) {
+        if (! $this->enforceUniqueRoutes) {
             return;
         }
 
@@ -1392,9 +1392,9 @@ class Router implements BindingRegistrar, RegistrarContract
      * @param bool $value
      * @return void
      */
-    public static function enforceUniqueRoutes($value = true)
+    public function enforceUniqueRoutes($value = true)
     {
-        static::$enforceUniqueRoutes = $value;
+        $this->enforceUniqueRoutes = $value;
     }
 
     /**
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -466,29 +466,30 @@ class Router implements BindingRegistrar, RegistrarContract
      */
     public function addRoute($methods, $uri, $action)
     {
-        $this->validateDuplicatesIfNeeded($methods, $uri);
+        $route = tap($this->createRoute($methods, $uri, $action), fn (Route $route) => $this->validateDuplicatesIfNeeded($route));
 
-        return $this->routes->add($this->createRoute($methods, $uri, $action));
+        return $this->routes->add($route);
     }
 
     /**
      * Throw an exception if a route with the same path already exists.
      *
-     * @param  array|string  $methods
-     * @param  string  $uri
+     * @param  \Illuminate\Routing\Route $newRoute
      * @return void
      * @throws \Illuminate\Routing\Exceptions\RouteAlreadyRegisteredException
      */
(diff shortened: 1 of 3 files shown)


driesvints:
Just want to say that it's a general practice, for example, to overwrite existing routes from packages in your own routes file. So definitely keep this opt in.

driesvints:
Seems like one of your own tests fails. Please mark as ready once it passes.

src/Illuminate/Routing/Router.php
```diff
@@ -124,6 +125,13 @@ class Router implements BindingRegistrar, RegistrarContract
      */
     public static $verbs = ['GET', 'HEAD', 'POST', 'PUT', 'PATCH', 'DELETE', 'OPTIONS'];
 
+    /**
+     * Whether duplicate routes should be disallowed.
+     *
+     * @var bool
+     */
+    protected static $enforceUniqueRoutes = false;
```
Tofandel, line 133: Don't think this should be static

src/Illuminate/Routing/Router.php
```diff
@@ -1350,6 +1385,17 @@ public function setContainer(Container $container)
         return $this;
     }
 
+    /**
+     * Instruct the application to throw an exception if a duplicate route is registered.
+     *
+     * @param bool $value
+     * @return void
+     */
+    public static function enforceUniqueRoutes($value = true)
```
Tofandel, line 1394: Likewise

src/Illuminate/Routing/Router.php
```diff
@@ -454,12 +462,39 @@ public function getLastGroupPrefix()
...
+     * @return void
+     * @throws \Illuminate\Routing\Exceptions\RouteAlreadyRegisteredException
+     */
+    protected function validateDuplicatesIfNeeded($methods, $uri)
+    {
+        if (! static::$enforceUniqueRoutes) {
+            return;
+        }
+
+        $cleanUriParameters = fn (string $uri) => preg_replace('/{(.*?)\}/s', '', $uri);
```
Tofandel, line 488: This preg_replace doesn't seem right

Have a test with params like
```php
        $this->router->get('foo/{foo}/baz/{baz}', fn () => true);
        $this->router->get('foo/{bar}/bar/{baz}', fn () => false);
 ```
 
 Pretty sure it will throw
 
 You also need to account for the fact that params can be assigned different non overlapping "wheres" so you should replace them with their respecting where value instead of removing them

File: src/Illuminate/Routing/Router.php [hunk 4]
```diff
@@ -454,12 +462,39 @@ public function getLastGroupPrefix()
...
 
+    /**
+     * Throw an exception if a route with the same path already exists.
+     *
+     * @param  array|string  $methods
+     * @param  string  $uri
+     * @return void
+     * @throws \Illuminate\Routing\Exceptions\RouteAlreadyRegisteredException
+     */
+    protected function validateDuplicatesIfNeeded($methods, $uri)
```
Jubeki, line 482:
```suggestion
    protected function validateDuplicatesIfNeeded($newRoute)
```

How about passing the new Route, which should be checked against the other routes?
You would then have access to check the domain, which should solve your problem with the failing test.

mateusjatenee: You're right... I tried doing that but only after adding the route, so it obviously didn't work. 😅 thank you

src/Illuminate/Routing/Router.php
mateusjatenee: You're correct, it is a singleton. Thanks

src/Illuminate/Routing/Router.php
mateusjatenee: Wheres are definitely tricky...

mateusjatenee: Theoretically, you'll only know if the routes collide when a request comes in, right? So I guess it should not take them into account.

mateusjatenee:
> Just want to say that it's a general practice, for example, to overwrite existing routes from packages in your own routes file. So definitely keep this opt in.

Since there's a method to enable/disable I believe users would also be able to select on which "group" of routes this should apply.   
And sorry for the broke test — I hadn't figured out the domain stuff last night, but @Jubeki helped me out. 👍

taylorotwell:
Don't think I'll take any action on this right now.

medilies:
I think having the `route:list` command show `(duplicate)` would be useful

PR #43654, opened 2022-08-11
Allow adding of more faker providers

ziming:
Sometimes we want to add additional faker providers or do overriding

https://fakerphp.github.io/#faker-internals-understanding-providers

With this PR

in our config/app.php we can then do

```php
'extra_faker_providers' => [
    // let's say we want to generate Chinese names instead but keep everything else en_US
    \Faker\Provider\zh_CN\Person::class,
   // we may also want to add some 3rd party package faker providers to be used in all factories
  // or our own
];
```


taylorotwell:
I would maybe just do this in a setupBeforeClass method of a base test class instead of adding additional configuration to Laravel

ziming:
thanks for the tip @taylorotwell , so i will create my own Factory superclass for all my factory classes to inherit from.

However i'm not quite sure which method should I override. I'm guessing withFaker(), would you say that is the correct one too?

thanks!

Issue #43627, opened 2022-08-09
Atomic locks do not provide reliable mutual exclusion when using the file cache driver


hbgl:
- Laravel Version: 9.23.0
- PHP Version: 8.0.15
- Database Driver & Version: none

### Description:
The locks returned by [FileStore](https://github.com/laravel/framework/blob/053840f579cf01d353d81333802afced79b1c0af/src/Illuminate/Cache/FileStore.php) are instances of the class [CacheLock](https://github.com/laravel/framework/blob/053840f579cf01d353d81333802afced79b1c0af/src/Illuminate/Cache/CacheLock.php). CacheLock's implementation of the acquire method is not atomic when the lock timeout is 0 (the default):

https://github.com/laravel/framework/blob/053840f579cf01d353d81333802afced79b1c0af/src/Illuminate/Cache/CacheLock.php#L30-L50

When the lock timeout is 0, first a `get` and then a `put` is performed. When combined, these operations are not atomic and thus multiple actors can enter the critical section protected by the lock. The possibility for race conditions was already discussed in the PR https://github.com/laravel/framework/pull/35139 that introduced the change.

#### Proposed quick fix
A quick fix would be to always call `$this->store->add` even if  `$this->seconds <= 0` because the store can handle 0 timeouts just fine:

```php
public function acquire() 
{ 
    return $this->store->add($this->name, $this->owner, $this->seconds); 
}
```

There is no need to even check that the `add` method exists because it is the stores responsibility to return a compatible lock instance.

#### Proposed proper fix
[...]

PR #43661, opened 2022-08-11
Fix race condition in locks issued by the file cache driver

hbgl:
This pull request fixes issue https://github.com/laravel/framework/issues/43627.

TLDR:
Atomic locks issued by [FileStore](https://github.com/laravel/framework/blob/09d50b70557cd061ced973efcd06713a0bf803c4/src/Illuminate/Cache/FileStore.php) or other stores that use the [HasCacheLock](https://github.com/laravel/framework/blob/09d50b70557cd061ced973efcd06713a0bf803c4/src/Illuminate/Cache/HasCacheLock.php) trait have a [race condition](https://github.com/laravel/framework/blob/09d50b70557cd061ced973efcd06713a0bf803c4/src/Illuminate/Cache/CacheLock.php#L43-L49) when the lock timeout is zero (the default). This can cause multiple processes to concurrently enter a critical section protected by the lock.

The [old lock implementation](https://github.com/laravel/framework/blob/09d50b70557cd061ced973efcd06713a0bf803c4/src/Illuminate/Cache/CacheLock.php#L30-L50) would do a non-atomic `get` and `put` when the lock timeout is less than or equal to zero and an atomic `add` otherwise. The [new implementation](https://github.com/hbgl/framework/blob/c93e7f13e5a35b18153fec34cce0d6d4b3a5df0b/src/Illuminate/Cache/CacheLock.php#L30-L43) always does an atomic `add`. This change only breaks code that incorrectly implemented their cache store, i.e. their `add` method cannot handle lock timeouts less or equal to zero.

### DIFF
--- a/src/Illuminate/Cache/CacheLock.php
+++ b/src/Illuminate/Cache/CacheLock.php
@@ -34,19 +34,12 @@ class CacheLock extends Lock
      */
     public function acquire()
     {
-        if (method_exists($this->store, 'add') && $this->seconds > 0) {
-            return $this->store->add(
-                $this->name, $this->owner, $this->seconds
-            );
-        }
-
-        if (! is_null($this->store->get($this->name))) {
-            return false;
-        }
-
-        return ($this->seconds > 0)
-                ? $this->store->put($this->name, $this->owner, $this->seconds)
-                : $this->store->forever($this->name, $this->owner);
+        // Cache stores that return this lock type must ensure that they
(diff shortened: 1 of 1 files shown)


taylorotwell:
Not sure I'm comfortable changing this on a patch release.

driesvints:
@hbgl can you try sending to master?

hbgl:
@driesvints Here is the new PR for master https://github.com/laravel/framework/pull/43686

Issue #33604, opened 2020-07-20
Setting a date on Eloquent model with timezone causes wrong time to be stored in UTC 


grEvenX:
- Laravel Version: 7.20.0
- PHP Version: 7.4.8
- Database Driver & Version: PostgreSQL (all versions), but doesn't seem relevant for the problem

### Description:
The time stored internally in the Eloquent instance (and the DB) is wrong when trying to set a date property of an Eloquent model using a Carbon instance that has a different timezone than UTC set.

It seems the problem is related to the `HasAttribute` trait's `fromDateTime` method that gets called when the Eloquent attribute setter method is called.
The method currently has the following implementation

```!php
public function fromDateTime($value)
{
    return empty($value) ? $value : $this->asDateTime($value)->format(
        $this->getDateFormat()
    );
}
```

This should probably respect the `timezone` config property defined in `config/app.php`, or be set to UTC?
If it was supposed to be in UTC, it could be solved by changing `$this->asDateTime($value)->format(...)` to `$this->asDateTime($value)->utc()->format(...)`

### Steps To Reproduce:

The following code can be used as a test:
```!php
<?php

class MyModel extends Model
{
    protected $dates = [ 'started_at'];
}

$aModel = new MyModel();
$aModel->started_at = new Carbon\Carbon('2018-05-09T07:09:32+02:00');

// The date-time to be stored in UTC should be at 05:09:32 instead of 07:09:32
assert('2018-05-09T05:09:32.000000Z' == $aModel->started_at->toJSON()); // should be true
// $aModel->started_at->toJSON() outputs 2018-05-09T07:09:32.000000Z (lost the 2 hours from the UTC+2 offset from the input date)
```

driesvints:
Please see https://github.com/laravel/framework/issues/33592 and other closed issues. You are responsible for converting to the correct timezone before storing.

UncertaintyP:
Hey @driesvints I just stumbled upon this.

According to the linked issue you said Laravel stores the datetime as timestamp. So is it intended behavior that Laravel changes the timestamp? As you can see in the following example Laravel shifts the timestamp by 1 hour :worried: :cry: 


```
>>> $model->start_at  = now('Europe/Berlin')
=> Illuminate\Support\Carbon @1638313200 {#5413
     date: 2021-12-08 19:30:22.622384 Europe/Berlin (+01:00),
   }
>>> $model->start_at
=> Illuminate\Support\Carbon @1638316800 {#5412
[...]

Issue #33592, opened 2020-07-20
Datetime strings containing timezone data (written to a Laravel API) are stored incorrectly in database


miking7:
- Laravel Version: 7.13.0
- PHP Version: 7.4.1
- Database Driver & Version: mysql 5.7.12

### Description:
Laravel incorrectly throws out timezone data passed to an API (ie: create/update call), and stores times in the database that are based on an "assumed" timezone instead of the timezone specified.

The issue is due to an Eloquent inconsistency already highlighed (unsuccessfully) under [issue #19865](https://github.com/laravel/framework/issues/19865).  @DSpeichert was correct - Laravel is unnecessarily throwing away timezone information here, and is exhibiting inconsistent behaviour between retrieving and storing information to Eloquent Models.  

### Steps to Reproduce:
1. Create an API using any model with datetime values.
2. Try to write a value to the API using any (ie: ISO 8601) text format that includes a timezone offset - ie: `2020-07-20T02:30+10:00`.
3. Laravel completely discards the timezone information and writes invalid values to the database - ie: `2020-07-20 02:30` (using the default database timezone) - which stores an invalid datetime value.
### Temporary Fix - Override the faulty function (coming from HasAttributes.php):
Notes: 
* Insert this code into your models to fix Eloquent's inconsistent handling of timezones when storing vs retrieving from a model.  
* This code is a Laravel 7 update based on a @DSpeichert's original suggested fix in [issue #19865]
* Although we're overriding it on a per-model basis here, this function could completely replace the faulty function defined in `Illuminate/Database/Eloquent/Concerns/HasAttributes.php`

```php
    /**
     * Convert a DateTime to a storable string.  (Override Laravel Default - which inconsistently throws away timezone informatio)
     *
     * @param  mixed  $value
     * @return string|null
     */
    public function fromDateTime($value)
    {
        return empty($value) ? $value : $this->asDateTime($value)->setTimezone(config('app.timezone'))->format(
            $this->getDateFormat()
        );
    }
```

### Alternative temporary fix using a mutator
Note: This code is required for every datetime field you need to protect in your models.  (We had developed this code before finding @DSpeichert 's more generic fix.
```php
    /**
     * Set the "Test" field value (without throwing away timezone data)
     *
     * @param  string  $value
     * @return void
     */
    public function setTestAttribute($value)
    {
        $this->attributes['test'] = Date::parse($value)->setTimeZone(config('app.timezone'));
    }
```

[...]

PR #43662, opened 2022-08-11
[10.x] Fix timestamp change after model setter in a different timezone

Tofandel:
Fixes https://github.com/laravel/framework/issues/42993, https://github.com/laravel/framework/issues/33604, https://github.com/laravel/framework/issues/33592

Currently when setting a date on a model the timestamp of the date that we just set on it doesn't match if it was in a different timezone than the application timezone, this issue is only present with the `datetime` cast and not the `datetime:Y-m-d H:i:s` cast

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1291,7 +1291,7 @@ trait HasAttributes
         // Carbon instances from that format. Again, this provides for simple date
         // fields on the database, while still supporting Carbonized conversion.
         if ($this->isStandardDateFormat($value)) {
-            return Date::instance(Carbon::createFromFormat('Y-m-d', $value)->startOfDay());
+            return Date::instance(Carbon::createFromFormat('Y-m-d', $value, 'UTC')->startOfDay());
         }
 
         $format = $this->getDateFormat();
@@ -1300,12 +1300,12 @@ trait HasAttributes
         // the database connection and use that format to create the Carbon object
         // that is returned back out to the developers after we convert it here.
         try {
-            $date = Date::createFromFormat($format, $value);
-        } catch (InvalidArgumentException $e) {
+            $date = Date::createFromFormat($format, $value, 'UTC')->setTimezone(date_default_timezone_get());
+        } catch (InvalidArgumentException) {
             $date = false;
         }
 
-        return $date ?: Date::parse($value);
+        return $date ?: Date::parse($value, 'UTC')->setTimezone(date_default_timezone_get());
     }
 
     /**
@@ -1327,7 +1327,7 @@ trait HasAttributes
      */
     public function fromDateTime($value)
     {
-        return empty($value) ? $value : $this->asDateTime($value)->setTimezone(date_default_timezone_get())->format(
+        return empty($value) ? $value : $this->asDateTime($value)->setTimezone('UTC')->format(
             $this->getDateFormat()
         );
     }
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1327,7 +1327,7 @@ trait HasAttributes
      */
     public function fromDateTime($value)
     {
-        return empty($value) ? $value : $this->asDateTime($value)->format(
+        return empty($value) ? $value : $this->asDateTime($value)->setTimezone(date_default_timezone_get())->format(
             $this->getDateFormat()
         );
     }
--- a/tests/Integration/Database/EloquentModelDateCastingTest.php
+++ b/tests/Integration/Database/EloquentModelDateCastingTest.php
@@ -64,6 +64,9 @@ class EloquentModelDateCastingTest extends DatabaseTestCase
 
         $this->assertEquals((new Carbon($model->getAttributes()['format_datetime']))->timestamp, $date->timestamp);
         $this->assertEquals((new Carbon($model->getAttributes()['basic_datetime']))->timestamp, $date->timestamp);
+
+        // Make sure the date was not mutated
+        $this->assertEquals('+09:00', $date->getOffsetString());
     }
 
     public function testDatesArePreservedInGetter()
--- a/tests/Integration/Database/EloquentModelDateCastingTest.php
+++ b/tests/Integration/Database/EloquentModelDateCastingTest.php
@@ -53,35 +53,20 @@ class EloquentModelDateCastingTest extends DatabaseTestCase
         $this->assertSame(['2019-10-01', '2019-10-01 10:15:20', '2019-10-01', '2019-10-01 10:15'], $bindings);
     }
 
-    public function testDatesArePreservedInSetter()
+    public function testDatesArePreservedInSetterGetter()
     {
         $date = Carbon::make('2022-08-28T12:52:02+09:00');
 
-        $model = new TestModel1();
+        $model = new TestModel1([
+            'format_datetime' => $date,
+            'basic_datetime' => $date,
+        ]);
 
-        $model->setAttribute('basic_datetime', $date);
-        $model->setAttribute('format_datetime', $date);
-
-        $this->assertEquals((new Carbon($model->getAttributes()['format_datetime']))->timestamp, $date->timestamp);
-        $this->assertEquals((new Carbon($model->getAttributes()['basic_datetime']))->timestamp, $date->timestamp);
+        $this->assertEquals($model->format_datetime, $date);
+        $this->assertEquals($model->basic_datetime, $date);
 
         // Make sure the date was not mutated
         $this->assertEquals('+09:00', $date->getOffsetString());
-    }
-
-    public function testDatesArePreservedInGetter()
-    {
-        $date = Carbon::make('2022-08-28T12:52:02+09:00');
-
-        $model = new TestModel1();
-
-        $model->setRawAttributes([
-            'basic_datetime' => $date,
-            'format_datetime' => $date,
-        ]);
-
-        $this->assertEquals((new Carbon($model->getAttribute('format_datetime')))->timestamp, $date->timestamp);
-        $this->assertEquals((new Carbon($model->getAttribute('basic_datetime')))->timestamp, $date->timestamp);
     }
 
     public function testDatesFormattedArrayAndJson()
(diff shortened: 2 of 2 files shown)


driesvints:
This isn't correct and this is a major breaking change. These datetimes should not have their timezone set by the app. This should be done in userland by whatever domain logic the app warrants.

Tofandel:
How is it not correct? Please explain your train of thoughts because the current behaviour is broken and we're a lot to report it

It's a minor breaking change because people affected by this issue already have to manually convert all their dates to app timezone, in this case this change would not change anything but consider it for v10 at least because its a major bug

The problem is that a model doesn't preserve the timezone when setting a time as you can clearly see, so we end up with a different timestamp in a getter than in a setter

And the worst part is that datetime and datetime:Y-m-d H:i:s behave differently, one has the issue and the other doesn't, to tell you how broken that is

A date should never change its timestamp, never, its the universal time and doesn't depend on the timezone

Tofandel:
I changed the test, so maybe now you'll understand better the underlying issue (only line 66 fails now, and it passes after the fix)

After that, I'm out of ways to explain this

driesvints:
@Tofandel I really don't know what's left to say 😅 This is the expected behavior when you use a custom format. We've documented everything here: https://laravel.com/docs/9.x/eloquent-mutators#date-casting

Screenshot 2022-08-12 at 11 02 58

Tofandel:
> UTC ISO-8601 date string (1986-05-28T21:05:54.000000Z), regardless of the timezone specified in your application's timezone configuration option
This sentence is what I'm talking about, the UTC ISO-8601 is a format that allows for timezone (after the Z) the fact that the timezone is being stripped right there but the date kept as is is a violation of that iso

driesvints:
You're reading that wrong. That's for serialization when the datetime is outputted. The timezone is stripped because the value is prepared for being entered in the database. There is no magic mechanism for saving the timezone as well and there never has been. You need to make sure your datetimes are converted to UTC before saving them to the database.

Tofandel:
There is if you did respect the ISO but even that's not true

```php
public function getDateFormat()
    {
        return 'Y-m-d H:i:s';
    }
```

This is not the iso 8601 format

This is
```php
public function getDateFormat()
    {
        return 'c';
    }
```

https://www.php.net/manual/fr/datetime.format.php

driesvints:
That's the format for storing in the database. Not to output when serializing to JSON. You can't store the timezone in the database in a timestamp.

Tofandel:
I know that you can't store the timezone in the database, and that's why to be in line with that, it needs to be changed to the app timezone BEFORE being converted in that format if that format doesn't preserve the timezone, otherwise you're just storing a wrong value

The thing the doc is saying doesn't explain the bug I'm demonstrating, a cast shouldn't change the timestamp value of a date

Why have all these iso conventions if we're just gonna throw them all out the window anyways

Tofandel:
There is absolutely no way that this is not a bug
```php
$date2 = Carbon\Carbon::make('2022-12-08 14:00:00+02:00');
$user->created_at = $date2;
$this->assertEquals($date2, $user->created_at); // False
```

driesvints:
> it needs to be changed to the app timezone BEFORE being converted in that format if that format doesn't preserve the timezone

Now that's what I'm trying to tell you. That's explicitly what we don't want to do. Time stored in the database shouldn't be based on what your app's timezone is but by UTC. If you have a single point of truth when storing time it's easier to convert it later on to whatever timezone you prefer or what your business logic dictates. 

> There is absolutely no way that this is not a bug

It's not a bug, it's the expected behavior.

Tofandel:
You just said two condradicting statements
> Time stored in the database shouldn't be based on what your app's timezone is but by UTC.
They are already stored in the app timezone not in UTC because php makes dates in the app timezone, it's just a good thing that the default app timezone is UTC, but if you change it, it's wrong

> If you have a single point of truth
And that single point of truth is what this PR is about, right now 2 same dates with different time zones result in a different output

If you want the source of truth to be utc then that should be
```php
->setTimezone('UTC')
```

```php
if ($this->isStandardDateFormat($value)) {
    return Date::instance(Carbon::createFromFormat('Y-m-d', $value)->startOfDay(), 'UTC');
}
```
instead of the mess that it currently is

Tofandel:
Latest commit is what a UTC source of truth looks like

Tofandel:
I have another proposal that can keep everyone happy by adding a `datetime_utc` and `date_utc` cast instead of modifying the existing one

taylorotwell:

yurii-github:
@taylorotwell 

hello

I do not know if it makes any difference on your decision, but below I provide code based on `backtrace` that shows you  the problem with the framework in usage of 2 methods - `fromDateTime()` and `asDateTime ()`.

As you may see, this `BaseModel` introduces usage of new option `app.db_timezone`, while fixing existing option `app.timezone`.

I do not provide my test coverage intentionally.


regards 

p.s.
Code target is Laravel 8.x

```
<?php

declare(strict_types=1);

namespace App\Models;

use Carbon\CarbonInterface;
use Illuminate\Support\Facades\Config;
use Illuminate\Support\Facades\Date;

abstract class BaseModel extends Model
{
    /**
     * Convert a DateTime to a storable string.
     *
     * @param mixed $value
     *
     * @return string|null
     */
    public function fromDateTime($value)
    {
        if (empty($value)) {
            return $value;
        }

        $caller = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1]['function'] ?? null;
        $isRead = in_array($caller, [
            'originalIsEquivalent', 
        ]);
        $isWrite = in_array($caller, [
            'runSoftDelete',
            'setAttribute',
            'freshTimestampString',
        ]);

        if (!$isRead && !$isWrite) {
            throw new \DomainException("FIX ASAP! Unknown caller: ${caller}");
        }

        if ($isRead) {
            $v = $this->asDateTime($value)->setTimezone(Config::get('app.timezone'))->format($this->getDateFormat());
        } else {
            $v = $this->asDateTime($value)->setTimezone(Config::get('app.db_timezone'))->format($this->getDateFormat());
        }

        return $v;
    }

    /**
     * Return a timestamp as DateTime object.
     *
     * @param mixed $value
     *
     * @return \Illuminate\Support\Carbon
     */
    protected function asDateTime($value)
    {
        $caller = debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS, 2)[1]['function'] ?? null;
        $isRead = in_array($caller, ['addDateAttributesToArray', 'castAttribute', 'transformModelValue']);
        $isWrite = in_array($caller, ['fromDateTime']);

        if (!$isRead && !$isWrite) {
            throw new \DomainException("FIX ASAP! Unknown caller: ${caller}");
        }

        // If this value is already a Carbon instance, we shall just return it as is.
        // This prevents us having to re-instantiate a Carbon instance when we know
        // it already is one, which wouldn't be fulfilled by the DateTime check.
        if ($value instanceof CarbonInterface) {
            return Date::instance($value);
        }

        // If the value is already a DateTime instance, we will just skip the rest of
        // these checks since they will be a waste of time, and hinder performance
        // when checking the field. We will just return the DateTime right away.
        if ($value instanceof \DateTimeInterface) {
            return Date::parse(
                $value->format('Y-m-d H:i:s.u'),
                $value->getTimezone()
            );
        }

        // If this value is an integer, we will assume it is a UNIX timestamp's value
        // and format a Carbon object from this timestamp. This allows flexibility
        // when defining your date fields as they might be UNIX timestamps here.
        if (is_numeric($value)) {
            return Date::createFromTimestamp($value);
        }

        // If the value is in simply year, month, day format, we will instantiate the
        // Carbon instances from that format. Again, this provides for simple date
        // fields on the database, while still supporting Carbonized conversion.
        if ($this->isStandardDateFormat($value)) {
            return Date::instance(Carbon::createFromFormat('Y-m-d', $value)->startOfDay());
        }

        $format = $this->getDateFormat();

        // Finally, we will just assume this date is in the format used by default on
        // the database connection and use that format to create the Carbon object
        // that is returned back out to the developers after we convert it here.
        try {
            if ($isRead) { // attribute is read
                $date = Date::createFromFormat($format, $value, Config::get('app.db_timezone'))->setTimezone(Config::get('app.timezone'));
            } else { // attribute is written
                $date = Date::createFromFormat($format, $value, Config::get('app.timezone'));
            }
        } catch (\InvalidArgumentException $e) {
            $date = false;
        }

        if ($date) {
            return $date;
        }

        if ($isRead) {
            $date = Date::parse($value, Config::get('app.db_timezone'))->setTimezone(Config::get('app.timezone'));
        } else {
            $date = Date::parse($value);
        }

        return $date;
    }
}

```

mvdstam:
Hey @Tofandel, @driesvints, just chiming in here after this issue caught us off guard.

I'm a bit surprised that this behaviour is considered "expected behaviour" rather than a bug. I'd like to quote @driesvints (https://github.com/laravel/framework/pull/43662#issuecomment-1212925829) with the following snippet:

```php
$date2 = Carbon\Carbon::make('2022-12-08 14:00:00+02:00');
$user->created_at = $date2;
$this->assertEquals($date2, $user->created_at); // False
```

I'd say it can't get more clearer than this.

Basically, the timezone of the incoming `DateTime` is ignored completely and stored with the assumption of it having the same timezone as the current `config('app.timezone')`. The entire point, to my understanding, is that incoming datetimes should have their timezones normalized to the value of `config('app.timezone')`. It's what I absolutely expect.

The current implementation makes the use of a timezone more or less pointless, since this means that the `datetime` cast will only work correctly if the incoming date has a timezone equal to the current `config('app.timezone')`.

Can this PR please get looked at again?

eroteev:
Hi @Tofandel, @driesvints,

I was quite surprised to discover that assigning a Carbon instance with a specific timezone to a model instance results in data loss. Moreover, the documentation is somewhat misleading in this regard. Initially, I believed that the date would be converted to UTC before being set to the model and later saved in the database.

I suggest updating the implementation to cast dates having timezone to UTC, or at least clarifying in the documentation how dates are handled.

Thank you for your attention to this matter.

driesvints:
As said before in numerous past issues we gladly accept PR's to make the docs more clear. We won't change how anything of this works because any change to this is a major breaking change to every single Laravel application out there.

mvdstam:
@driesvints 

Sorry, but this doesn't make any sense. With that logic, any bugs in the framework (like the one currently discussed) can never be fixed as is bug is simply rebranded as "expected behaviour". Where, in reality, the code is doing something it shouldn't and the developers using said code need to implement workarounds to actually make it work like it's supposed to. Storing timezone-specific data in a Eloquent model through the built-in casts now results in **loss of data**. That's the bug. It doesn't make sense to call it anything else but that.

I don't understand why this cannot be fixed as part of a minor release or even a major release if you feel that this really _is._ a big breaking change. This isn't a _documentation_ issue, this is a _code_ issue. **Please** take a look at this again with a open mind.

driesvints:
There is no bug. This is the expected behaviour. I'm really done explaining this over all of these years 😅

PR #43665, opened 2022-08-11
[9.x] Add ability to `Model::query()->touch()` to mass update timestamps

stevebauman:
This PR adds the `touch($column = null)` method to the Eloquent query builder.

This is a convenience method, allowing you to mass update a model's timestamp in the database table with or without query constraints:

**Mass touching the updated at column**:
```php
$touched = User::query()->touch();
```

**Mass touching the updated at column with constraints**:
```php
$touched = User::where('email', 'like', '%@company.com')->touch();
```

**Mass touching a specific column**:
```php
$published = Post::query()->touch('published_at');
```

It behaves identically to `$model->touch()`, and accepts the same method argument.

Let me know your thoughts! Thanks for your time 🙏

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1030,6 +1030,29 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Update the column's update timestamp.
+     *
+     * @param  string|null  $column
+     * @return int|false
+     */
+    public function touch($column = null)
+    {
+        $time = $this->model->freshTimestamp();
+
+        if ($column) {
+            return $this->toBase()->update([$column => $time]);
+        }
+
+        $column = $this->model->getUpdatedAtColumn();
+
+        if (! $this->model->usesTimestamps() || is_null($column)) {
+            return false;
+        }
+
+        return $this->toBase()->update([$column => $time]);
+    }
+
+    /**
      * Increment a column's value by a given amount.
      *
      * @param  string|\Illuminate\Database\Query\Expression  $column
--- a/tests/Database/DatabaseEloquentBuilderTest.php
+++ b/tests/Database/DatabaseEloquentBuilderTest.php
@@ -2079,6 +2079,61 @@ class DatabaseEloquentBuilderTest extends TestCase
         $this->assertEquals(2, $result);
     }
 
+    public function testTouch()
+    {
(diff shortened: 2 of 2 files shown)


src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -1029,6 +1029,29 @@ public function upsert(array $values, $uniqueBy, $update = null)
...
+     *
+     * @param  string|null  $column
+     * @return int|false
+     */
+    public function touch($column = null)
+    {
+        $time = $this->model->freshTimestamp();
+
+        if ($column) {
+            return $this->toBase()->update([$column => $time]);
```
LukeTowers, line 1043: Is it intentional that you can touch a column if the model `usesTimestamps()` returns false?

stevebauman: Yes:

https://github.com/laravel/framework/blob/fdf3ae9760729652673359d3e2cced727ba25b92/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php#L22-L37

src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -1029,6 +1029,29 @@ public function upsert(array $values, $uniqueBy, $update = null)
...
+        }
+
+        $column = $this->model->getUpdatedAtColumn();
+
+        if (! $this->model->usesTimestamps() || is_null($column)) {
+            return false;
+        }
+
+        return $this->toBase()->update([$column => $time]);
```
LukeTowers, line 1052:
```suggestion
        $time = $this->model->freshTimestamp();
        $column = $column ?: ($this->model->usesTimestamps() ? $this->model->getUpdatedAtColumn() : null);

        if (is_null($column) {
            return false;
        }

        return $this->toBase()->update([$column => $time]);
```
How about using this instead then (or a variation of this without the ternaries if you'd prefer)? That way we can avoid the duplicated calls / early returns.

stevebauman: I think it's fine the way it is. I want the structure to be as close to `Model::touch()` as possible.

Appreciate the PR review 👍

LukeTowers: No problem, I didn't check Model::touch() so if it's matching the existing structure it should be fine.

PR #43666, opened 2022-08-11
[9.x] Allow chunkById on Arrays, as well as Models

BennyC:
This PR allows `chunkById` with `PDO_ASSOC` results from the `Builder`.

I believe this is a very niche scenario, and understand a rejection on this PR. On projects I've previously (and currently) worked on, I've had scenarios where we're using `PDO_ASSOC` as the fetch mode.

This change would allow others to `chunkById` in either scenario, handling Models/objects or associative arrays.


TheGlenn88:
+1

PR #43668, opened 2022-08-11
[9.x] Allow for int value parameters to whereMonth() and whereDay()

lupinitylabs:
When trying to pass an integer value to whereMonth() or whereDay(), e.g. 
```php
User::whereMonth('updated_at', '=', 4)->get();
```
PHPstan will report `Parameter #3 $value of method Illuminate\Database\Eloquent\Builder<App\Models\User>::whereMonth() expects DateTimeInterface|string|null, int given.` 

At the same time,
```php
User::whereYear('updated_at', '=', 2022)->get();
```
will accept integer values.

This PR suggests to make the type definitions for `whereYear()`, `whereMonth()` and `whereDay()` more consistent by adding the int type to the @param docblock for `$value`.

I am swapping out the usage of `str_pad()` with `sprintf()` because the latter will accept mixed arguments, whereas for `str_pad()`, we would have to convert int values to strings first (and it's more compact, too).

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1386,7 +1386,7 @@ class Builder implements BuilderContract
      *
      * @param  string  $column
      * @param  string  $operator
-     * @param  \DateTimeInterface|string|null  $value
+     * @param  \DateTimeInterface|string|int|null  $value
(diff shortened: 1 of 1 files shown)


lupinitylabs:
Looks like the failing tests are unrelated...?

PR #43669, opened 2022-08-11
[9.x] `db:table` command properly handle table who doesn't exist

kichetof:
New `db:table` doesn't proper handle table who doesn't exist.

When I first tried it, I run `artisan db:table user` _(forgot the **s**)_ and instead of getting an error message I got an error:

```
   ErrorException 

  Attempt to read property "size" on null

  at vendor/laravel/framework/src/Illuminate/Database/Console/DatabaseInspectionCommand.php:89
     85▕     {
     86▕         return $connection->selectOne('SELECT (data_length + index_length) AS size FROM information_schema.TABLES WHERE table_schema = ? AND table_name = ?', [
     87▕             $connection->getDatabaseName(),
     88▕             $table,
  ➜  89▕         ])->size;
     90▕     }
     91▕ 
     92▕     /**
     93▕      * Get the size of a Postgres table in bytes.

      +16 vendor frames 
  17  artisan:37
      Illuminate\Foundation\Console\Kernel::handle()

```

This PR fix it by checking if table exists before accessing any properties.


src/Illuminate/Database/Console/TableCommand.php
```diff
@@ -52,6 +52,12 @@ public function handle(ConnectionResolverInterface $connections)
             collect($schema->listTables())->flatMap(fn (Table $table) => [$table->getName()])->toArray()
         );
 
+        if (! $schema->tablesExist([$table])) {
+            $this->warn("Table [{$table}] doesn't exist.");
```
jbrooksuk, line 56:
```suggestion
            $this->components->warn("Table [{$table}] doesn't exist.");
```

jbrooksuk (changes requested):

src/Illuminate/Database/Console/TableCommand.php
kichetof: Totally agree with components but `handle` should return `int`. If we return the warning component, it'll be `void`. What do you think?

jbrooksuk: I've just checked the `ShowModelCommand` and it's the same there; https://github.com/laravel/framework/blob/9.x/src/Illuminate/Foundation/Console/ShowModelCommand.php#L81-L107

kichetof: Same message with PhpStorm on `ShowModelCommand`
`'void' method 'error' result used `

And for `TableCommand`: `'void' method 'warn' result used `

But it's only some warning, everything works as expected. As it's fine for you, it's fine for me :)

jbrooksuk (approved):

PR #43671, opened 2022-08-11
[9.x]  Added whenHasValue method to the request object

loremroot:
Sometimes we need to evaluate the value of a input and execute one thing or another, we can achieve this by using a simple if or with the whenHas method.
```php
$request->whenHas('input', $closure)..
```
The thing with this is that even we know that the input is present, if we need to evaluate the value of the input we have to add an extra if inside the closure.

```php
$request->whenHas('input', function($value) {
    if($value == 'Hello') {

    } else {

    }
})..
```

I think we be really cool if we can pass an extra param to evaluate the value of the input, think is cleaner this way.
The method take 2 closures one of the input has the value (if), and the other closure when this input doesn't match with the value we passed as parameter (else).

```php
$request->whenHasValue('input', 'Hello', function($value) {
    // the input has the value of hello
}, function() {
    // the input does not have the value of hello
})..
```

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -134,6 +134,28 @@ trait InteractsWithInput
     }
 
     /**
+     * Apply the callback if the request contains the given input with the given value.
+     *
+     * @param  string  $key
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #43681, opened 2022-08-12
addedd test to ensure validateAlpha rule does not permit Mark Unicode Pointers

doshu:
the current validate alpha rules, allow some special unicode char that are in the "Mark" category.
this is because the regular expression, include the chars with the options "\pM" and it should not.

in the test, the char U+0488 has been used that represent a special symbol like a circle made with other semicircles 
(actually a strange unusual char)


driesvints:
This seems to break our test suite. Please re-send this with passing tests.

PR #43683, opened 2022-08-12
modified alpha validation rule to avoid mark unicode char pass the validation

doshu:
modified the alhpa validation rule and created the test


driesvints:
Seems tests are still failing here. Please make sure that they pass before you send this in.

doshu:
test fail but not cause of the edit i made.
i don't know the real cause is.

i will try to solve and resend

Il ven 12 ago 2022, 16:58 Dries Vints ***@***.***> ha scritto:

PR #43688, opened 2022-08-12
Optimize eager loading when no keys to be loaded

garygreen:
Currently when eager loading relations that don't have any keys to be loaded Laravel will still perform the query:

**Example:**

```php
class User {
   public function picture()
   {
      return $this->belongsTo(Picture::class);
   }
}

User::create(['picture_id' => null]),
User::create(['picture_id' => null])
$users = User::all()->load(['picture']);
```

The eager load will trigger a select on the database, but because there is no model keys to be loaded it replaces the `where in (...)` with [`where 0 = 1`](https://github.com/laravel/framework/blob/master/src/Illuminate/Database/Query/Grammars/Grammar.php#L292)

```sql
select * from `pictures` where 0 = 1
```

This PR checks to see if there aren't any model keys that can be loaded and if so will implicitly return an empty collection, preventing the unnecessary database query.


garygreen:
Will create a new PR with cleaner commit history.

PR #43689, opened 2022-08-12
Optimize eager loading when no keys to be loaded

garygreen:
Currently when eager loading relations that don't have any keys to be loaded Laravel will still execute the query:

**Example:**

```php
class User {
   public function picture()
   {
      return $this->belongsTo(Picture::class);
   }
}

User::create(['picture_id' => null]),
User::create(['picture_id' => null])
$users = User::all()->load(['picture']);
```

The eager load will trigger a select on the database even though there is no model keys to be loaded.

It does this by replacing replacing the empty `where in (...)` with [`where 0 = 1`](https://github.com/laravel/framework/blob/master/src/Illuminate/Database/Query/Grammars/Grammar.php#L292):

```sql
select * from `pictures` where 0 = 1
```

This PR checks to see if there are not any model keys that can be loaded, and if so will implicitly return an empty collection - this prevents the unnecessary database query.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/BelongsTo.php
+++ b/src/Illuminate/Database/Eloquent/Relations/BelongsTo.php
@@ -113,7 +113,7 @@ class BelongsTo extends Relation
 
         $whereIn = $this->whereInMethod($this->related, $this->ownerKey);
 
-        $this->query->{$whereIn}($key, $this->getEagerModelKeys($models));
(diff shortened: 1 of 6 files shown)


taylorotwell:

garygreen:
@taylorotwell Please could you elaborate more on why this wasn't accepted? In our app we have hundreds of `impossible where` calls to mysql due to this. It's an optimization that removes unnecessary calls to the database...

Would you be willing to accept a PR that at least makes it easier to add this as a custom package?

PR #43293, opened 2022-07-19
[9.x] Add requiredUnless, excludeUnless and prohibitedUnless Validation Methods

intrepidws:
Currently, validation methods exist for `Rule::requiredIf()`, `Rule::excludeIf()` and `Rule::prohibitedIf()` which allow complex conditional logic that goes beyond the simple use of the `required_if`, `exclude_if` and `prohibited_if` validation rules.

This PR seeks to bring parity by adding `Rule::requiredUnless()`, `Rule::excludeUnless()` and `Rule::prohibitedUnless()` validation method alternatives of the simple validation rules: `required_unless`, `exclude_unless` and `prohibited_unless`.

If accepted, I will PR these changes to the docs as well.

Some code examples:

```php
use Illuminate\Support\Facades\Validator;
use Illuminate\Validation\Rule;

Validator::make($request->all(), [
    'role_id' => Rule::requiredUnless($request->user()->is_expired),
]);
 
Validator::make($request->all(), [
    'role_id' => Rule::requiredUnless(fn () => $request->user()->is_expired),
]);

Validator::make($request->all(), [
    'role_id' => Rule::prohibitedUnless($request->user()->is_active),
]);
 
Validator::make($request->all(), [
    'role_id' => Rule::prohibitedUnless(fn () => $request->user()->is_active),
]);

Validator::make($request->all(), [
    'role_id' => Rule::excludeUnless($request->user()->is_active),
]);
 
Validator::make($request->all(), [
    'role_id' => Rule::excludeUnless(fn () => $request->user()->is_active),
]);
```

PR #43690, opened 2022-08-13
[9.x] Add `acceptedIf` and `declinedIf` validation methods

lioneaglesolutions:
This PR adds methods `acceptedIf` and `declinedIf` for the current `accepted_if` and `declined_if` validations respectively.

The reason I made this PR is I am used to using the `requiredIf` method and just *assumed* that other methods for `*_if` validations would exist. 

A recent PR adding similar rules for `*_unless` methods was not merged https://github.com/laravel/framework/pull/43293 however I believe as we already have rules for `requiredIf`, `excludeIf` and `prohibitedIf`, it is confusing to only include some `*_if` validation methods, and not others (as opposed to adding a full set of `*_unless` methods). This PR would mean all `*_if` validations would be available as a method.

This PR adds only 2 very simple classes with almost identical tests as the existing `requiredIf` method and would remove confusion for developers wondering if a validation method for an `*_if` rule exists or not.

### DIFF
--- a/src/Illuminate/Validation/Rule.php
+++ b/src/Illuminate/Validation/Rule.php
@@ -4,6 +4,8 @@ namespace Illuminate\Validation;
 
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Support\Traits\Macroable;
+use Illuminate\Validation\Rules\AcceptedIf;
+use Illuminate\Validation\Rules\DeclinedIf;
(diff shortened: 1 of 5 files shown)


taylorotwell:
It's not super necessary... you can do things like:

```php
'field' => ['required', ...(true ? ['accepted'] : [])]
```

PR #43694, opened 2022-08-13
[9.x] Removed class exists check for `make:migration` command

sethsandaru:
Since Laravel 9.x we are using **anonymous classes** for migrations, thus this check can be removed.

And it is also a nice-to-have for other Laravel projects that migrated to Laravel 9.x: 
- We can quickly create a migration file with a simple name (eg `php artisan make:migration UpdateUsersTable`) even the class already exists. 
- We don't have to think for a long long name, just create the migration and start the development quick.

This improvement can be considered as bug fixing.

### DIFF
--- a/src/Illuminate/Database/Migrations/MigrationCreator.php
+++ b/src/Illuminate/Database/Migrations/MigrationCreator.php
@@ -97,10 +97,6 @@ class MigrationCreator
                 $this->files->requireOnce($migrationFile);
             }
         }
-
-        if (class_exists($className = $this->getClassName($name))) {
-            throw new InvalidArgumentException("A {$className} class already exists.");
-        }
     }
 
     /**
--- a/src/Illuminate/Database/Migrations/MigrationCreator.php
+++ b/src/Illuminate/Database/Migrations/MigrationCreator.php
@@ -148,17 +148,6 @@ class MigrationCreator
     }
 
     /**
-     * Get the class name of a migration name.
-     *
-     * @param  string  $name
-     * @return string
-     */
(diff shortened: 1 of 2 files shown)


ankurk91:
There are 3rd party package which are still using class names in migrations.

sethsandaru:
@ankurk91 would be out-of-scope of this improvement PR since I didn't touch anything related to the migration core, only the `make:migration` command 😄

axlon:
How does this affect older projects that were upgraded to Laravel 9 and don't (exclusively) use anonymous migration classes?

sethsandaru:
> How does this affect older projects that were upgraded to Laravel 9 and don't (exclusively) use anonymous migration classes?

Since Laravel 9, make:migration will use anonymous classes (you can check the stub files) 

But if you intend to change the your new migration class back to the old way then I guess it isn't recommended to do to so?

driesvints:
This will break all existing apps that still use non-anonymous classes. We specifically let this in to allow for an easy upgrade path.

sethsandaru:
> This will break all existing apps that still use non-anonymous classes. We specifically let this in to allow for an easy upgrade path.

@driesvints yes but the thing is, `make:migration` will create an anon class so the check is not relevant at the creation time?

I mean after the creation, people can change it back to the old way and if they did, that's another story and it is their fault - `make:migration` here did nothing wrong, did it?

Additional point: I already had the `UpdateUsersTable` class, now I want to hit the `php artisan make:migration UpdateUsersTable` again => class existed => can't create a migration (even anon class migration) so it is like a small blocker/pain here?

PR #43696, opened 2022-08-13
[9.x] Add price directive for blade

milwad-dev:
Convert number format to directive blade


ankurk91:
Lets call it money or currency

milwad-dev:
@ankurk91 
ok

milwad-dev:
{{ number_format(5000000) }}
@money(5000000)

driesvints:
I wouldn't do this. This is a pretty domain specific keyword that could be used in lots of apps where the expected implementation would be different (different library like MoneyPHP, decimal handling, etc). This is best defined in userland.

milwad-dev:
@driesvints 
Sometimes can use directive to use php function!

taylorotwell:

PR #43698, opened 2022-08-14
[9.x] Add middleware condition for routes

yamitsushi:
this commit will enable the can middleware to parse multiple policies using the "|" character

Sample Code

`Route::get('permissions', [FetchController::class, 'permissions'])->can("attach|detach",Permission::class);`

Since this is my first contribution I didn't know if I place it in the proper location, I do hope that this will be implemented.

### DIFF
--- a/src/Illuminate/Auth/Access/Gate.php
+++ b/src/Illuminate/Auth/Access/Gate.php
@@ -392,13 +392,20 @@ class Gate implements GateContract
     public function inspect($ability, $arguments = [])
     {
         try {
-            $result = $this->raw($ability, $arguments);
+            $abilities = explode("|", $ability);
+            foreach($abilities as $ability) {
+                $result = $this->raw($ability, $arguments);
 
-            if ($result instanceof Response) {
-                return $result;
+                if ($result instanceof Response) {
+                    return $result;
+                }
+
+                if($result) {
+                    return Response::allow();
+                }
             }
 
-            return $result ? Response::allow() : Response::deny();
+            return Response::deny();
         } catch (AuthorizationException $e) {
             return $e->toResponse();
         }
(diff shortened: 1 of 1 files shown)


src/Illuminate/Auth/Access/Gate.php
```diff
@@ -392,13 +392,20 @@ public function authorize($ability, $arguments = [])
...
         try {
-            $result = $this->raw($ability, $arguments);
+            $abilities = explode('|', $ability);
+            foreach ($abilities as $ability) {
+                $result = $this->raw($ability, $arguments);
 
-            if ($result instanceof Response) {
-                return $result;
+                if ($result instanceof Response) {
+                    return $result;
```
robert-sykes, line 400: Hi there! I am not really sure why have you expanded this check out? Was it not simpler and readable with the ternary operator? I think this was fine how it was to be honest.

seriquynh:
In my opinion, this is not necessary. I will create a method called "attachOrDetach" in the PermissionPolicy class. Then, the route definition will be.

```php
<?php

Route::get('permissions', [FetchController::class, 'permissions'])->can("attachOrDetach", Permission::class);
```

yamitsushi:
> In my opinion, this is not necessary. I will create a method called "attachOrDetach" in the PermissionPolicy class. Then, the route definition will be.
> 
> ```
> <?php
> 
> Route::get('permissions', [FetchController::class, 'permissions'])->can("attachOrDetach", Permission::class);
> ```

I also think that I can create a new method specifically for this, but that won't do since I will implement similar to this on multiple routes in my project.

Implementing a dedicated Policy will definitely become a headache in my case as I can't keep track of everything I do, heck I even forgot what I actually did here without reading back my changes.

taylorotwell:

Issue #43705, opened 2022-08-15
Null compatibility broken


jc-skurman:
- Laravel Version: 9.24.0
- PHP Version: 8.1.0
- Database Driver & Version:
MySql 8
Description:
When we use a cast of type 'object', 'collection' and we have the original value null, we get an error when trying to decode json.
This change appeared in the [PR](https://github.com/laravel/framework/pull/42793/files#diff-59d24f1a8f0dd1e51ee7dffc8778133711634e928ecef4a91b63fc3851cb1579) and is still relevant even for version 9.
PR
We get the warning

> PHP Deprecated: json_decode(): Passing null to parameter https://github.com/laravel/framework/issues/1 ($json) of type string is deprecated

driesvints:
@tpetry since you originally created https://github.com/laravel/framework/issues/37961 which sparked https://github.com/laravel/framework/pull/41337: do you know of a way to do a fix so both issues are resolved?

Also want to ping you @JBtje since you're the author of https://github.com/laravel/framework/pull/41337.

@jc-skurman this change has been in the framework now for several months. Did you only update Laravel now recently?

jc-skurman:
I just now caught this bug on the old version and after indicating that you no longer support it, I checked it on the latest version

MasterRO94:
@driesvints 

I've checked issues can be reproduced if you have `null` original value for `object` or/and `collection` cast. This can be fixed replacing `fromJson` call to `castAttribute`
 

 ```php
 } elseif ($this->hasCast($key, ['object', 'collection'])) {
            return $this->fromJson($attribute) ===
                $this->fromJson($original);
 ```
change to 
 ```php
 } elseif ($this->hasCast($key, ['object', 'collection'])) {
            return $this->castAttribute($key, $attribute) ===
                $this->castAttribute($key, $original);             
 ```

tpetry:
I replaced the dirty check with my custom implementation because in my projects I found a lot more edge cases, e.g. different order of object keys.

But based on the implementation of the PR the null value has to be handled too. `null != json, null = null`.

JBtje:
First and foremost, `null` is not valid `json`, so the warning is correct. If you seed your empty JSON fields with a valid JSON value (e.g. empty string), the warning is solved.
I have come across this warning a few times, just to realize i did something wrong (i.e. not seeding JSON field properly).

But if you think `null` should be a valid value for a JSON field, then this could easily be solved:
[...]

PR #43706, opened 2022-08-15
[9.x] Null value for auto-cast field caused deprication warning in php 8.1

JBtje:
Problem:
- PHP 8.1 no longer supports null for json_decode, and throws a deprication warning
https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg
Changed in https://github.com/php/php-src/commit/b10416a652d26577a22fe0b183b2258b20c8bb86#diff-a39ce231303cec12a6d347165931a7f2209ee7dce5989f2e06f8c04143ec7032

However, thr RFC says that `null` is a valid JSON value:
https://datatracker.ietf.org/doc/rfc8259/
```
A JSON value MUST be an object, array, number, or string, or one of
the following three literal names:

    false null true

 The literal names MUST be lowercase.  No other literal names are
 allowed.
```
Thus, auto cast should accept `null` as a valid JSON and not result in a deprecation warning..

also see issue #43705

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1200,7 +1200,7 @@ trait HasAttributes
      */
     public function fromJson($value, $asObject = false)
     {
-        return json_decode($value, ! $asObject);
+        return json_decode($value ?? '', ! $asObject);
     }
 
     /**


JBtje:
Note though:

```
var_dump( json_decode( false ) ); --> NULL
var_dump( json_decode( true ) ); --> int(1)
```

while rfc7159 indicates that `JSON false` -> `false` and `JSON true` -> `true`

php does not even have tests for these cases...
https://github.com/php/php-src/blob/master/ext/json/tests/001.phpt
so would Laravel support them?

marcovo:
The reasoning based on the RFC is not entirely solid. It is true that false, true and null are valid JSON literals, but translating to php, this would mean that the following three statements should successfully return the respective values (which they do):
```
json_decode('true')
json_decode('false')
json_decode('null')
```

Here, `'null'` is the null described by the RFC. This is something else than:
```
json_decode(null)
```
i.e., the null literal instead of the string `'null'`.

While the submitted change is probably perfectly valid, this would be primarily to conform to php's new function signature. The RFC argument is not valid here.
Also, missing a test here?

driesvints:
@marcovo did this break something for you? If so how is that reproducable?

marcovo:
It did not break anything for me; merely pointing out that in the opening post, the deprecation warning is a valid point but the RFC argument is not applicable. Ultimately the change seems fine.

driesvints:
@marcovo right, gotcha.

kayw-geek:
@driesvints #43507 So should it be shut down instantly?

driesvints:
@kayw-geek you didn't exactly explain where this issue was coming from as well as the people in https://github.com/laravel/framework/issues/43705 did. Afaik you could be calling the method directly. Next time please provide a more thorough explanation and your PR might have been accepted.

kayw-geek:
> @kayw-geek you didn't exactly explain where this issue was coming from as well as the people in #43705 did. Afaik you could be calling the method directly. Next time please provide a more thorough explanation and your PR might have been accepted.

OK, I'll write it in detail later.

PR #43714, opened 2022-08-15
Ensure correct 'integrity' value is used for css assets

julesjanssen:
Like with a css resource that is a dependency of an import, a directly imported css file should also use its own manifest record to make sure the correct integrity value is applied.
Before this change, the `<link` would use the integrity key of the JS asset.


timacdonald:
Hey @julesjanssen I'm gonna make this target my existing PR branch, as it conflicts - and I'd love to get this and my PR merged into tomorrows release.

timacdonald (approved):
Thanks for fixing this one. Tested and confirmed.

jessarcher (approved):

PR #43720, opened 2022-08-16
[9.x] Request model helper

rcerljenko:
Hi,

This PR aims to do similar thing like `$request->enum()` helper only in this case we are binding an Eloquent model instance.
Eg:
```php
$post = $request->model('post_id', Post::class);
```
Just like `enum` helper, it returns `null` if the field is missing from the payload or if there's an invalid key provided in the input field itself.

If that's something worth having I'll provide tests to cover all the cases.


michaelnabil230:
@rcerljenko hey
In fact, I was thinking of adding the same pr, but I was a little lazy, but good job 😂♥️

taylorotwell:

PR #43723, opened 2022-08-16
adding Arr::match() method

HijenHEK:
#### description

Arr::match() method return an array with values from the given array where their keys match the given set of keys, it keeps **only** the keys that exist in the list and **adds** the missing keys with a null value 

#### usage
formatting arrays with a standard set of keys

#### example 

```PHP
$array = ['name' => 'Desk', 'price' => 100, 'orders' => 10];
$array = Arr::match($array, ['name', 'price', 'color']);
```

#### output
```PHP
[ 
   'name' => 'Desk', 
   'price' => 100, 
   'color' => null
]

```

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -494,6 +494,21 @@ class Arr
     }
 
     /**
+     * Match array keys with given keys set
+     *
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #43724, opened 2022-08-16
[10.x] Optimize eager loading when no keys to be loaded

garygreen:
Sending this again. @taylorotwell please could you provide suggestions on what I can do to get this merged?

Currently when eager loading relations it causes many impossible wheres to be executed, slowing down application load time. The reason for that is mentioned in the [original PR](https://github.com/laravel/framework/pull/43689) - Laravel will add an impossible where `0 = 1` which would never return a result for an eagerly loaded relation. This PR fixes those eager loading scenarios, which are very common with `NULL` column values.

If you could provide alternative suggestions on how to get this merged that would be great.

image

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Relation.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Relation.php
@@ -41,6 +41,13 @@ abstract class Relation implements BuilderContract
     protected $related;
 
     /**
+     * Indicates whether the eagerly loaded relation should implicitly return an empty collection.
+     *
+     * @var bool
+     */
+    protected $eagerImplicitlyEmpty = false;
+
+    /**
      * Indicates if the relation is adding constraints.
      *
      * @var bool
--- a/src/Illuminate/Database/Eloquent/Relations/Relation.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Relation.php
@@ -161,7 +161,7 @@ abstract class Relation implements BuilderContract
      */
     public function getEager()
     {
-        if ($this->eagerImplicitlyEmpty ?? false) {
+        if ($this->eagerImplicitlyEmpty) {
             return $this->query->getModel()->newCollection();
         }
 
--- a/src/Illuminate/Database/Eloquent/Relations/Relation.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Relation.php
@@ -45,7 +45,7 @@ abstract class Relation implements BuilderContract
      *
      * @var bool
      */
-    protected $eagerImplicitlyEmpty = false;
+    protected $eagerKeysWereEmpty = false;
 
     /**
      * Indicates if the relation is adding constraints.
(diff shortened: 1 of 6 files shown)


taylorotwell:
I'm curious if there is a reason this was sent to 10.x instead of 9.x?

garygreen:
The reason I sent it to 10.x is because it potentially poses a breaking change, but only for a very strange edge case scenario when adding constraints to the eager load:

```php
$user->load(['picture' => function($q) {
	$q->orWhere('id', 'some other id')->orWhere('id', 'another id');
}]);
```

This would currently result in this query:

image

Which would could return results, but with this PR it wouldn't return any results because the model didn't have any loadable relation keys. I would say the latter is expected, and you could say this is actually an unintended bug in Laravel at the moment. Personally I would expect a query more like the below where it should restrict loading of model to that particular set of keys being loaded - i.e. the constraints is ADDITIONAL as oppose to a complete override:

```sql
select * from `pictures` where 0 = 1 and (`id` = 'some other id' or `id` = 'another id')
```

So that's the reason for sending it to 10.x. I'm curious to know what you think about that. I'm open to suggestions on how to change this PR if you aren't comfortable with that type of behaviour change..

Tofandel:
What you describe is a bug, the query should definitely not load entities that don't pertain to that model (especially if the relation is marked as null in the first place), it could in fact be a security issue as that could load data from other users unexpectedly

It should be sent to 9.X or 8.X imo

garygreen:
> What you describe is a bug [...] It should be sent to 9.X or 8.X imo

Agreed, I think it's a bug as well although this PR doesn't really intend to fix that. It technically does fix it for eager loading when all relations are `null` but that's just because it doesn't need to hit the database at all, so doesn't even attempt to build a query. For any others it would still allow a query like this:

```php
select * from `pictures` where in (15, 88, 99) or `id` = 'some other id' or `id` = 'another id'
```

...because there are loadable model keys, so it go onto hitting the database constructing the query as above. Solving that should be a separate PR which wraps those additional eager constraints in a parentheses group.

Tofandel:
I see, I thought you fixed it for all cases, then yes there should be a new PR to fix this issue, but then this PR is still not breaking any expected behavior

src/Illuminate/Database/Eloquent/Relations/Relation.php
```diff
@@ -154,6 +154,10 @@ abstract public function getResults();
      */
     public function getEager()
     {
+        if ($this->eagerImplicitlyEmpty ?? false) {
```
taylorotwell, line 157: Where is `eagerImplicitlyEager` defined?

garygreen: Have now added that as a proper parameter along with docblock comment.

src/Illuminate/Database/Eloquent/Relations/Relation.php
```diff
@@ -392,6 +396,24 @@ protected function whereInMethod(Model $model, $key)
...
+     * @param  Builder  $query
+     * @return void
+     */
+    protected function whereInEager(string $whereIn, $key, array $modelKeys, $query = null)
+    {
+        ($query ?? $this->query)->{$whereIn}($key, $modelKeys);
+
+        if ($modelKeys === []) {
+            $this->eagerImplicitlyEmpty = true;
+        }
```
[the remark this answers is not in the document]
garygreen: Well if the eager is loaded again it'll make calls to the relation methods `addEagerConstraints` which subsequently call `whereInEager` which could change the `eagerImplicitlyEmpty` property because it will know if there are any loadable keys. So once the eager constraints are added, it will set that property.

PR #43728, opened 2022-08-16
Adding support for non-backed enums

ollieread:
This PR adds support for non-backed enums when casting eloquent attributes. It works based on the fact that enum cases are glorified constants, so the `constant()` method will work for retrieving.

It also updates the `$value` param type for `Illuminate\Database\Eloquent\Concerns\HasAttributes::setEnumCastableAttribute` from `\BackedEnum` to `\UnitEnum|string|int`. It's worth noting that the original version should have technically been `\BackedEnum|string|int` because there are code paths for handling when it isn't an instance.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -814,7 +814,11 @@ trait HasAttributes
             return $value;
         }
 
-        return $castType::from($value);
+        if (is_subclass_of($castType, \BackedEnum::class)) {
+            return $castType::from($value);
+        }
+
+        return constant($castType . '::' . $value);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


foremtehan:
please add tests if possible

johanmolen:
Should the non-backed enums also be implemented in the `addCastAttributesToArray` method?

ollieread:
> Should the non-backed enums also be implemented in the `addCastAttributesToArray` method?

This is a very good point!

ollieread:
@laravel there's a valid point raised by @johanmolen, will fix this ASAP

johanmolen:
@ollieread awesome, if you need help let me know.

johanmolen:
@ollieread my quick fix for now (line 310):
```
if ($this->isEnumCastable($key) && (! ($attributes[$key] ?? null) instanceof Arrayable)) {
    if (! isset($attributes[$key])) {
        $attributes[$key] = null;
    } else if ($attributes[$key] instanceof BackedEnum) {
        $attributes[$key] = $attributes[$key]->value;
    } else {
        $attributes[$key] = $attributes[$key]->name;
    }
}
```

ollieread:
Added a fix in #43842 @johanmolen

johanmolen:
@ollieread Awesome 💯

PR #43731, opened 2022-08-16
Add support for additional where* methods to route groups

ollieread:
This PR addresses the change originally mentioned in #43509 and reverted in #43523.

Unlike the original PR, it adds support for all of the additional `where*` methods available on a route, to a route group, by adding the `CreatesRegularExpressionRouteConstraints` trait.

It works directly on the router, handled by `Router::__call` with an explicit check for beginning with `where`, that proxies to the method. There is no check to see if the method exists, because if it does not, it will be passed to the `RouteRegistrar::attributes` method automatically, so will achieve the same without the requirement for an additional condition.

```php
Route::whereIn(['foo', 'bar'], ['one', 'two'])->prefix('/{foo}/{bar}')->group(function () {});
```

And on a `RouteRegistrar`, using the trait.
```php
Route::prefix('/{foo}/{bar}')->whereIn(['foo', 'bar'], ['one', 'two'])->group(function () {});
```

This PR also adds 8 tests that test both of the approaches, as they are slightly different. No changes have been made to `RouteRegistrar::allowedAttributes` because the trait turns them into `where` attributes.

### DIFF
--- a/src/Illuminate/Routing/RouteRegistrar.php
+++ b/src/Illuminate/Routing/RouteRegistrar.php
@@ -29,6 +29,8 @@ use InvalidArgumentException;
  */
 class RouteRegistrar
 {
+    use CreatesRegularExpressionRouteConstraints;
+
     /**
      * The router instance.
      *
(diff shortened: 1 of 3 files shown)


morloderex:
@ollieread Have you tested this with route caching enabled?

ollieread:
> @ollieread Have you tested this with route caching enabled?

I have not, but I'm confident that I do not need to. Route groups already support where attributes, and these methods are wrappers for the where attribute. There is nothing new created because of this, it's just easier.

PR #43734, opened 2022-08-17
check for env() in /app when running php artisan cache:config

gmhafiz:
According to https://laravel.com/docs/9.x/configuration#configuration-caching, running the command `php artisan config:cache`
will result in Laravel ignoring any value retrieval using `env()` as they are now cached. 

However, the artisan command currently does not warn the user that values from `env()` will not be read.

This PR grep the files under `/app` and look for occurrence of `env()` and report the file(s) containing it.

It does not break existing functionalities as it only prints the result without stopping config caching. More than that, it helps users to catch potential bug early before going into production.


taylorotwell:
You can use `env` outside of config if you're retrieving system level environment variables.

PR #43737, opened 2022-08-17
[9.x] Always use the write PDO connection to read the just stored pending batch

corneyl:
Setting the `sticky` option to `false` for a database connection with separate read/write database nodes, results in problems with finding a batchable job right after creation. 

This can happen because the read query is executed on a replicated read-only node directly after writing to the write node. If the select query is executed faster than the database replication takes place, then no batch is found. In that case an exception is thrown because jobs are added to a `null` batch.

This PR fixes this problem by always using the write PDO connection for finding the just created batchable job in the `store()` method in `DatabaseBatchRepository.php`.


taylorotwell:
Probably easier to just always use write PDO in this situation. We can't change a contract method signature on a patch release.

PR #43738, opened 2022-08-17
[9.x] Fix unique lock release for broadcast events

daannet:
This fix fixes not triggered lock release of #43416, #43516. It added back the UniqueBroadcastEvent class created by @DougSisk but now the uniqueId will always include the original event class. 

The regular BroadcastEvent does not implement the 'Queue' ShouldBeUnique and wasn't removing the lock. Now it is possible to dispatch the same event and id after the previous has been processed.

### DIFF
--- a/src/Illuminate/Broadcasting/UniqueBroadcastEvent.php
+++ b/src/Illuminate/Broadcasting/UniqueBroadcastEvent.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Broadcasting;
 
+use Illuminate\Contracts\Cache\Repository;
 use Illuminate\Contracts\Queue\ShouldBeUnique;
 
 class UniqueBroadcastEvent extends BroadcastEvent implements ShouldBeUnique
@@ -19,13 +20,6 @@ class UniqueBroadcastEvent extends BroadcastEvent implements ShouldBeUnique
      * @var int
      */
     public $uniqueFor;
-
-    /**
-     * The cache repository implementation that should be used to obtain unique locks.
-     *
-     * @var \Illuminate\Contracts\Cache\Repository
-     */
-    public $uniqueVia;
 
     /**
      * Create a new job handler instance.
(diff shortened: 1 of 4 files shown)


DougSisk:
This actually brings up another issue:
https://github.com/laravel/framework/blob/57c5672e12a9bcb6143134dc5644a2b45fe9d207/src/Illuminate/Queue/CallQueuedHandler.php#L201-L218
The lock release should be moved to within the `Illuminate\Bus\UniqueLock` instead.

src/Illuminate/Broadcasting/UniqueBroadcastEvent.php
```diff
@@ -0,0 +1,60 @@
...
+     * @var int
+     */
+    public $uniqueFor;
+
+    /**
+     * The cache repository implementation that should be used to obtain unique locks.
+     *
+     * @var \Illuminate\Contracts\Cache\Repository
+     */
+    public $uniqueVia;
```
DougSisk, line 28: This can't be a property, as the cache can't be serialized

daannet:
> This actually brings up another issue:
The UniqueBroadcastEvent has the data to release the lock after processing. Now uniqueVia is taken from the inner event or retrieved from container and working :) 

> The lock release for unique queue jobs should be moved to within the `Illuminate\Bus\UniqueLock` instead.
I don't know how desirable it is to change the queue/jobs logic for broadcast events. Where should we draw the line?

taylorotwell:
IMO we should not be modifying queue / job logic to accommodate this feature. Risks breaking more things.

DougSisk:
> IMO we should not be modifying queue / job logic to accommodate this feature. Risks breaking more things.

I've opened #43740 which explains the issue. It does not pertain to making this feature work; just fixing decoupled logic.

taylorotwell:
Can someone explain the actual problem at its root and how this fixes it? Also, calling the foundation `app` helper from component level code is generally not allowed.

daannet:
> Can someone explain the actual problem at its root and how this fixes it? Also, calling the foundation `app` helper from component level code is generally not allowed.

The current BroadcastEvent (job) doesn't have the queue ShouldBeUnique implementation and the required uniqueVia etc data and the unique lock is not released after processing.

DougSisk:
I believe this is only an issue if you don't define a `$uniqueFor`.

PR #43741, opened 2022-08-17
[9.x] Added `flatGet` to Collection

foremtehan:
It is often when you get data as a collection from external api and you want to only get data from a nested key like so:

```php
//Before

$data = Http::get($url)->collect();

$data->only('data')->collapse();
```

```php
//After

$data = Http::get($url)->collect();

$data->flatGet('data');
```

I'm not quite sure about the name feel free to change if you want to merge the PR.


taylorotwell:

PR #43745, opened 2022-08-17
[8.x] Fixes TypeError when $this->request is null when calling dispatchConnectionFailedEvent

### LINKED ISSUE #43743: [8.83] TypeError on HttpClient with custom middleware and ConnectException
- Laravel Version: 8.83.23
- PHP Version: 7.4
- Database Driver & Version:

### Description:

I'm facing a similar issue as described in #37594 , but in a different place. I think the cause is the same as explained in that issue by @lukeraymonddowning.

#### The error:

    Argument 1 passed to Illuminate\Http\Client\Events\ConnectionFailed::__construct() must be an instance of 
    Illuminate\Http\Client\Request, null given, called in
    /var/www/vendor/laravel/framework/src/Illuminate/Http/Client/PendingRequest.php on line 1093

### Steps To Reproduce:

In my project, I've created a macro to wrap the calls to an API, to handle the authentication. The macro sets a custom middleware in the HandlerStack. The error happens when doing a request to the api and having a Request Timeout for example.

Bellow is part of the code used:

    Http::macro('someApi', function (bool $authenticated = true) : PendingRequest {
        $stack = HandlerStack::create();

        $addRequestSignatureHeaders = static function (bool $authenticated = true) {
            return  static function (callable $handler) use ($authenticated) {
                return static function (
                    RequestInterface $request,
                    array $options
                ) use ($handler, $authenticated) {
                    if ($authenticated) {
                        $token = Cache::get('someApi.auth_token', function () {
                            $tokenResponse = Http::someApi(false)
                            ->retry(3, 1000, function ($exception) {
                                return $exception instanceof ConnectionException;
                            })->post('/auth', [
                                'username' => config('services.someApi.username'),
                                'password' => config('services.someApi.password'),
                            ])->object();

                            Cache::put('someApi.auth_token', $tokenResponse->access_token, now()->diffInSeconds($tokenResponse->expires_in));

                            return $tokenResponse->access_token;
                        });

                        $request = $request->withHeader('Authorization', 'Bearer ' . $token);
                    }

                    return $handler($request, $options);
                };
            };
        };

        $stack->push($addRequestSignatureHeaders($authenticated), 'add_request_signature_headers');

        return Http::withOptions([
                'handler' => $stack,
            ])
            ->timeout(3)
            ->withHeaders([
                'Content-Type' => 'application/json',
                'Accept' => 'application/json',
            ])->baseUrl(config('services.someApi.base_url'));
    });

#### Possible solution:

In the `Illuminate\Http\Client\PendingRequest` class, I changed the `dispatchConnectionFailedEvent` method with the following code and the error stopped:

    protected function dispatchConnectionFailedEvent()
    {
        if (!($dispatcher = optional($this->factory)->getDispatcher()) ||
            ! $this->request) {
                return;
        }

        $dispatcher->dispatch(new ConnectionFailed($this->request));
    }

Is there any other solutions for this? Should I open a PR with the suggested fix?

ivanwitzke:
Possible fix for issue #43743.

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -1089,9 +1089,12 @@ class PendingRequest
      */
     protected function dispatchConnectionFailedEvent()
     {
-        if ($dispatcher = optional($this->factory)->getDispatcher()) {
-            $dispatcher->dispatch(new ConnectionFailed($this->request));
+        if (! ($dispatcher = optional($this->factory)->getDispatcher()) ||
+            ! $this->request) {
+                return;
         }
+
+        $dispatcher->dispatch(new ConnectionFailed($this->request));
     }
 
     /**


driesvints:
8.x is closed for bug fixes

PR #43758, opened 2022-08-18
[9.x] Fix empty collection class serialization

### LINKED ISSUE #43732: Serialization fails for typed custom Eloquent Collections that are empty
- Laravel Version: 9.25.1
- PHP Version: 8.1.6
- Database Driver & Version: MySQL 8.0.27

### Description:

When queueing a job that has a **custom** Eloquent Collection as a typed property, and it holds an empty collection, (un)serializing fails with a Type Error, since it tries to assign the base Eloquent Collection to the property:

```php
class SomeJob implements ShouldQueue
{

    public OrderCollection $orders;

    public function __construct(OrderCollection $orders)
    {
        $this->orders = $orders;
    }

    public function handle()
    {
        //
    }
}

$job = new SomeJob(new OrderCollection());

$serialized = serialize($job)

unserialize($serialized)

```

Result:

`TypeError : Cannot assign Illuminate\Database\Eloquent\Collection to property SomeJob::$orders of type App\Eloquent\OrderCollection`

### Steps To Reproduce:

I've added a failing test to prove the bug is there:
https://github.com/hxnk/framework/commit/ac6ed8ba058ae007c73e352c332fd4fb1070c1df

driesvints:
This fixes an issue where restoring a custom Eloquent collection wouldn't be of the same type. We should save the class type in the ModelIdentifier class to achieve this.

Fixes #43732

### DIFF
--- a/src/Illuminate/Queue/SerializesAndRestoresModelIdentifiers.php
+++ b/src/Illuminate/Queue/SerializesAndRestoresModelIdentifiers.php
@@ -24,7 +24,8 @@ trait SerializesAndRestoresModelIdentifiers
                 $value->getQueueableClass(),
                 $value->getQueueableIds(),
                 $value->getQueueableRelations(),
-                $value->getQueueableConnection()
+                $value->getQueueableConnection(),
+                get_class($value)
             );
         }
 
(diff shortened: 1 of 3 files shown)


src/Illuminate/Queue/SerializesAndRestoresModelIdentifiers.php
```diff
@@ -66,7 +69,9 @@ protected function getRestoredPropertyValue($value)
     protected function restoreCollection($value)
     {
         if (! $value->class || count($value->id) === 0) {
-            return new EloquentCollection;
+            return ! is_null($value->collectionClass ?? null)
```
driesvints, line 72: The `?? null` part here is a BC measure to make sure we don't break already serialized model identifiers instances that don't have this property.

PR #43761, opened 2022-08-18
[9.x] Fix PHPDoc return type of `EnumeratesValues::min()` and `EnumeratesValues::max()`

tyabu12:
The PHPDoc return types of `EnumeratesValues::min()` and `EnumeratesValues::max()` are now `TValue`.

But the both function can return other than `TValue` when a string or a callable is passed to `$callback`.

One of the most possible cases is the min/max of column value of a Model class.

```php
/** @var Collection<int, User> $users */

>> $users->min('id');
=> 1

>> $users->min(static fn(User $user): int => $user->id);
=> 1
```

I think the PHPDoc return types of min/max should be `mixed` as well as `sum`.

### DIFF
--- a/src/Illuminate/Collections/Enumerable.php
+++ b/src/Illuminate/Collections/Enumerable.php
@@ -735,7 +735,7 @@ interface Enumerable extends Arrayable, Countable, IteratorAggregate, Jsonable, 
      * Get the min value of a given key.
      *
      * @param  (callable(TValue):mixed)|string|null  $callback
-     * @return TValue
+     * @return mixed
      */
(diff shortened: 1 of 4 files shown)


taylorotwell:
I'll let @nunomaduro take a look.

nunomaduro:
@tyabu12 Can you also update "Illuminate/Collections/Enumerable.php"?

tyabu12:
@nunomaduro 
Thank you for your reviewing.
I also updated "Illuminate/Collections/Enumerable.php".

PR #43766, opened 2022-08-19
[9.x] Add schedules method on ServiceProvider

alkin:
Service Providers are an easy way to group and provide similar functionalities. They are commonly used by external packages and internal modules to - among other things - register new commands and schedule jobs.

Command registration in a Service Provider is easily executed using the `commands`method:

```php
public function boot()
{
    $this->commands([
        InstallBlogPackage::class,
        AnotherCommand::class
    ]);
}
```

Registering schedules, on the other hand, is not so intuitive. Searching on the internet, people may find one of the following solutions:

```php
public function boot()
{
    // First Solution
        $schedule = $this->app->make(Schedule::class);
        $schedule->command('some:command')->everyMinute();
    });

    // Second Solution
    $this->app->afterResolving(Schedule::class, function (Schedule $schedule) {
        $schedule->command(MyAwesomeCommand::class)->everyMinute();
    });
}
```

This PR makes easier to schedule commands and jobs inside a Service Provider by introducing a new fully backwards compatible method `schedules`. It is a simple wrapper around `callAfterResolving` that makes schedule registration more intuitive and consistent with how commands are registered.

```php
public function boot()
{
    // Register commands
    $this->commands([
        InstallBlogPackage::class,
        AnotherCommand::class
    ]);

    // Register schedules
    $this->schedules(function (Schedule $schedule) {
        $schedule->job(new SendCampaigns())->hourly();
        $schedule->job(new SendNewsletters())->weekly();
    });
}
```

**References**
[Scheduling a Command in the Service Provider](https://laravelpackage.com/06-artisan-commands.html#registering-a-command-in-the-service-provider)
[How to Schedule Artisan Commands from Your Laravel Package](https://ryangjchandler.co.uk/posts/how-to-schedule-artisan-commands-from-your-laravel-package)

### DIFF
--- a/src/Illuminate/Support/ServiceProvider.php
+++ b/src/Illuminate/Support/ServiceProvider.php
@@ -409,16 +409,12 @@ abstract class ServiceProvider
     /**
      * Register the package's custom schedules.
      *
-     * @param  array|mixed  $commands
+     * @param  mixed  $callback
      * @return void
      */
     public function schedules($callback)
     {
-        $this->app->booted(function () use ($callback) {
-            $schedule = $this->app->make(Schedule::class);
-
-            $callback($schedule);
-        });
+        $this->callAfterResolving(Schedule::class, $callback);
     }
 
     /**
--- a/src/Illuminate/Support/ServiceProvider.php
+++ b/src/Illuminate/Support/ServiceProvider.php
@@ -4,6 +4,7 @@ namespace Illuminate\Support;
 
 use Closure;
 use Illuminate\Console\Application as Artisan;
+use Illuminate\Console\Scheduling\Schedule;
 use Illuminate\Contracts\Foundation\CachesConfiguration;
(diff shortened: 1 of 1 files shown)


rodrigopedra:
> by introducing a new fully backwards compatible method

Adding a public method to a class that is meant to be extended is never backwards compatible.

Someone already extending this class and already have a method named `schedules`. As PHP does not allow method overloading, if the existing method signature is different, updating to it will break theirs app.

I would send this to master, but I am not a maintainer, so you can wait for their regards.

timacdonald:
I don't _think_ we consider this a breaking change in Laravel (it's a trade-off), but because it is meant to be called internally, it could be marked `protected`

taylorotwell:
I think I'll hold off on this for now for the breaking change reason, but also because I didn't really intend for packages to register scheduled commands. I prefer the developer register them so they have full control over how they are run, etc.

PR #43768, opened 2022-08-19
Check if middleware is really set on the route

JeroenGovers:
If you currently calling withoutMiddleware within your tests they will pass even if the skipped middleware is not on the route. This can lead to security issues because you are assuming that the middleware is set.

This PR will automaticly check if the skipped middleware is on the requested route. I also added the assertMiddleware check to the TestResponse object so you can check if middleware is present even if you have not skipped it.

This PR can be a breaking change to the test of current laravel projects because they fail now if there is middleware that is skipped that is not set. But I think that is a good thing. Because you are assuming it is set.

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
@@ -64,6 +64,13 @@ trait MakesHttpRequests
     protected $withCredentials = false;
 
     /**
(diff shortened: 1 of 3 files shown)


taylorotwell:
Likely a breaking change - but I don't think this is really a security issue.

JeroenGovers:
Don't you think it is a nice feature that the middleware is checked that it is present on the current route?

PR #43777, opened 2022-08-19
Fix middleware not being registered on pooled requests

wfeller:
```php
    $pool->withMiddleware($middleware)->get('http://website.com'),
]);
```

The middleware is never registered on the Guzzle client since the client is created and locked before the middleware is pushed in the middleware stack. The proposed changes fix this.


wfeller:
We could also delete the `requestsReusableClient` and `getReusableClient` methods, and change the `buildClient` to:
```php
 public function buildClient()
 {
     return $this->client ?? this->createClient($this->buildHandlerStack());
}
```

taylorotwell:
I'm not sure this is explained very well and there are no tests.

PR #43782, opened 2022-08-19
[9.x] Resolve query builders out of the container

imanghafoori1:
Resolving these two classes out of the container makes package developers able to swap them at run-time without touching the models like adding a trait on all the models or changing the parent Model class. (For example creating a driver for neo4j or something.)
### Why
My particular need for this came while I was developing the[ eloquent-mockery ](https://github.com/imanghafoori1/eloquent-mockery) package. If developers want to use my package, `they have to put a dev-only trait on all their models`, which is kinda painful since it will be pushed into production for no reason.
I think with this change packages like neo-eloquent will be able to create a configurable driver for eloquent without changing the parent class of all the model classes.

### Compatibility:
- It is not a breaking change unless the `QueryBuilder::class` is already bound to something else on the container which is extremely unlikely.
- Let me note that the `illuminate/container` is a dependency of the `illuminate/database` and we can use it there.

### DIFF
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -8,6 +8,7 @@ use DateTimeInterface;
 use Doctrine\DBAL\Connection as DoctrineConnection;
 use Doctrine\DBAL\Types\Type;
 use Exception;
+use Illuminate\Container\Container;
 use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Database\Events\QueryExecuted;
 use Illuminate\Database\Events\StatementPrepared;
@@ -332,9 +333,11 @@ class Connection implements ConnectionInterface
      */
     public function query()
     {
-        return new QueryBuilder(
-            $this, $this->getQueryGrammar(), $this->getPostProcessor()
-        );
+        return Container::getInstance()->make(QueryBuilder::class, [
+            'connection' => $this,
+            'grammar' => $this->getQueryGrammar(),
+            'processor' => $this->getPostProcessor(),
+        ]);
     }
 
     /**
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Database\Eloquent;
 
 use ArrayAccess;
+use Illuminate\Container\Container;
 use Illuminate\Contracts\Broadcasting\HasBroadcastChannel;
 use Illuminate\Contracts\Queue\QueueableCollection;
 use Illuminate\Contracts\Queue\QueueableEntity;
@@ -1450,7 +1451,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
      */
     public function newEloquentBuilder($query)
     {
-        return new Builder($query);
+        return Container::getInstance()->make(Builder::class, ['query' => $query]);
     }
 
     /**


deleugpn:
Not a fan of this change 😬

imanghafoori1:
@deleugpn This was what I really needed as a package developer (not as a web app builder).
Could you please explain the downside of it?

rodrigopedra:
> Not a fan of this change

Me neither.

>  Could you please explain the downside of it?

- Two third-party packages register different resolvers for the query builder. Wouldn't a user expect both to work? As they won't work as expected would this user report a bug to each package or to Laravel directly?
- This allows third-party packages to extend the original builder and override an existing method to add some feature. If then Laravel changes its behavior or fix a bug, and the package maintainer take some time to update their overridden methods, would the package user report the inconsistent behavior to the package or to Laravel?

I didn't take a deep look into your linked package, but from skimming it, I think it could benefit from composition over inheritance.

Something like this:

~~~php
// first call to the EloquentMockery@fake would register observers to model events
// and return a mock helper object
EloquentMockery::fake(User::class)->addRow([...]);

// subsequent calls would just returned the mock helper
EloquentMockery::fake(User::class)->addRow([...]);

// Or a user could use a temporary variable if they prefer so
$mocker =  EloquentMockery::fake(User::class);

$mocker->addRow([...]);
$mocker->addRow([...]);
$mocker->addRow([...]);

// By leveraging Eloquent model events you can intercept and prevent basically any DB operation
$user = User::create(['username' => 'iman', 'email' => 'iman@gmail.com']);

// ... do assertions
~~~

Again, I did not deep dive into your package's code so I could be missing something very obvious. From what I saw you could even skip mocking the Builder, DB and Connection objects.

If you do need to mock the even more low level objects such as the Connection object, when registering the mocker you could replace the default connection on the config to a custom fake connection.

In my opinion the risks of a developer not knowing from where a strange behavior comes from, or from a user to expect multiple packages to just work when more than one could be replacing the builder resolver in the container, overcome the benefits.

These are just my two cents.

imanghafoori1:
@rodrigopedra Thank you for taking your time in order to share your knowledge and opinion.

imanghafoori1:
@rodrigopedra Ofcource bugs should be first reported to package maintainers and then they decide to forward it to the Laravel or not. This has happened to me before.

taylorotwell:

rodrigopedra:
@imanghafoori1 thanks for your regards =)

One use case I missed at the time was this:

https://timacdonald.me/dedicated-eloquent-model-query-builders

I might have been sleepy due to the local time, as this is something I use in almost all my projects.

If a developer uses the approach suggested in the link above, would they need to create two dedicated query builders for their project: one extending the original Eloquent builder, and another one extending the package's builder? To avoid it as dependency on production?

Note that per this discussion, some other developers also use the same approach:

https://github.com/laravel/framework/pull/42111#issuecomment-1108939741

Regarding connections being resolved on runtime, a package that provides an unsupported connection class, such as connecting to Oracle or DB2, usually register a new connection with the connection manager, and the developer should opt-in explicitly to use this connection on their project's config file. Also each model can use a different connection if needed by the developer explicitly setting their `$connection` property.

In your proposal, a user would not be able to use two different packages providing different query builder implementations that are supposed to enhance the query builder feature set. Beyond the fact on limiting using dedicated query builders as already mentioned.

Or even, they install in a project which has one or two models using dedicated eloquent query builders, and then wonder why the mock fakes work for some models and not for others.

Please note I do like your package's proposal, I just don't think that allowing automatic implicit under-the-hood engine swapping is the best way to go.

imanghafoori1:
@rodrigopedra Thanks for sharing your pro opinions.
For custom Eloquent query builders it does not matter, since we have to `Builder` classes. `Eloquent\Builder` and `Query\Builder`. The `Eloquent\Builder` uses `Query\Builder` as a lower lever injected dependency to achieve its goals.

I found that the way to swap the `Query\Builder` in Laravel currently exists and the change is not really needed. We can define a subclass of Connection and register and then swap it by changing configs with the original one. The Connection class can be overridden to resolve a different `Query\Builder`.
Hopefully, I can finish the `eloquent-mockery`  package in a proper manner.

Thanks for everyone being here.

PR #43783, opened 2022-08-19
Add ability to hidden Eloquent Model 

milwad-dev:
Sometimes you like to remove some attributes from query.
Actually you add hidden in model but with this method you can remove attributes like this :
```
$users = User::query()->latest()->get();
$users->hidden(['email', 'name']);
```

is very easy :)

### DIFF
--- a/src/Illuminate/Database/Eloquent/Collection.php
+++ b/src/Illuminate/Database/Eloquent/Collection.php
@@ -772,7 +772,7 @@ class Collection extends BaseCollection implements QueueableCollection
      */
     public function hidden($columns)
     {
-        $this->collect()->each(function ($item, $key) use ($columns) {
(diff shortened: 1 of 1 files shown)


ankurk91:
https://laravel.com/docs/9.x/eloquent-serialization#temporarily-modifying-attribute-visibility

milwad-dev:
@ankurk91 
Sometimes you can write like this:
```
$users = User::query()->with('roles')->get();
```
But write like this:
```
$users = User::query()->get();
$users->load('roles');
```

Sometimes need to hidden.

igorgaming:
Try use this https://laravel.com/docs/9.x/eloquent-collections#method-makeHidden

taylorotwell:

PR #43784, opened 2022-08-19
[9.x] Adds closure support to dispatch conditionals.

DarkGhostHunter:
## What?

Enables usage of closures to dispatch a job conditionally. The closure will receive the job arguments.

```php
MyQueuableJob::dispatchIf(Something::check(...), $name);
```

Also adds tests both `dispatchIf` and `dispatchUnless`.

### DIFF
--- a/src/Illuminate/Foundation/Bus/Dispatchable.php
+++ b/src/Illuminate/Foundation/Bus/Dispatchable.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Foundation\Bus;
 
+use Closure;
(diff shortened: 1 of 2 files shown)


taylorotwell:
@DarkGhostHunter 

IMO the entire job instance should be passed to the closure. I have updated the PR. Thoughts?

DarkGhostHunter:
I don't have too much of maneuverability if I get the job instance full of non-public members, so it would force me to:

a) make a public method for additional checks.
b) make its properties public. 

Since these are small chores, I can see it can be more convenient and forces me to push part of the check in the instance itself.

PR #36948, opened 2021-04-12
[8.x] Handle concurrent asynchronous requests in the HTTP client

cerbero90:
This PR aims to bring concurrency while sending asynchronous requests with the Laravel HTTP client.

Laravel Octane is not required to make this feature work, however Octane should be compatible with this feature as the latest version of [Swoole supports curl_multi operations](https://www.swoole.co.uk/article/v4-6-5-released).

The PR introduces:
- asynchronous requests
- a pool to handle asynchronous requests concurrently

Asynchronous requests can be created with the following syntax:
```php
$promise = Http::async()->get($url);
```
The result is an instance of `GuzzleHttp\Promise\Promise`.

Based on whether a promise is fulfilled or rejected, the `then()` method of the promise receives:
- An instance of `Illuminate\Http\Client\Response` if the promise is fulfilled
- An instance of `Illuminate\Http\Client\Response` if the promise is rejected but a response was provided (4xx, 5xx errors)
- An instance of `GuzzleHttp\Exception\TransferException` if the promise is rejected with no response (e.g. timeout)

```php
$promise = Http::async()->get($url)->then(fn (Response|TransferException $result) => $this->handleResult($result));
```

The pool collects asynchronous requests and send them concurrently, so that the slowest request determines the maximum waiting time for all promises to be fulfilled.

```php
// the waiting time will be ~3 seconds instead of 15

$responses = Http::pool(fn (Pool $pool) => [
    $pool->get('https://httpbin.org/delay/3')->then(/* ... */),
    $pool->get('https://httpbin.org/delay/3')->then(/* ... */),
    $pool->get('https://httpbin.org/delay/3')->then(/* ... */),
    $pool->get('https://httpbin.org/delay/3')->then(/* ... */),
    $pool->get('https://httpbin.org/delay/3')->then(/* ... */),
]);

$responses[0]->ok();
$responses[1]->successful();
// ...
```
Asynchronous requests can also be added to the pool with a key:

```php
$responses = Http::pool(fn (Pool $pool) => [
    $pool->add('foo')->get('https://httpbin.org/delay/3')->then(/* ... */),
    $pool->add('bar')->get('https://httpbin.org/delay/3')->then(/* ... */),
    $pool->add('baz')->get('https://httpbin.org/delay/3')->then(/* ... */),
]);

$responses['foo']->ok();
$responses['bar']->successful();
$connectionFailed = $responses['baz'] instanceof GuzzleHttp\Exception\ConnectException;
```
[...]

PR #43789, opened 2022-08-20
[9.x] Fix HTTP client pool

wfeller:
This PR fixes 2 problems:

1. Middleware not being sent when registered inside the pool callback (1st commit):
```php
// The middleware is never being called
Http::pool(fn (Pool $pool) => [
    $pool->withMiddleware($middleware)->get('https://example.com'),
]);
```
This is because the Guzzle client is created in the `setHandler` method on `PendingRequest`. Because of this, middleware are never registered on the handler stack when calling `send` on the request since the Guzzle client is not re-created.

2. Initial setup ignored (2nd commit):
```php
// The token (or any other setup done on the pending request before the pool() method call) is
// not passed to the final requests in the callback
Http::withToken('secret-key')->pool(fn (Pool $pool) => [
    $pool->post('https://example.com/api', $payload_1),
    $pool->post('https://example.com/api', $payload_2),
]);
```

### DIFF
--- a/src/Illuminate/Http/Client/Pool.php
+++ b/src/Illuminate/Http/Client/Pool.php
@@ -17,13 +17,6 @@ class Pool
     protected $factory;
 
     /**
-     * The handler function for the Guzzle client.
-     *
-     * @var callable
-     */
-    protected $handler;
-
-    /**
      * The pool of requests.
      *
      * @var array
--- a/src/Illuminate/Http/Client/Pool.php
+++ b/src/Illuminate/Http/Client/Pool.php
@@ -5,14 +5,14 @@ namespace Illuminate\Http\Client;
 use GuzzleHttp\Utils;
 
 /**
- * @mixin \Illuminate\Http\Client\Factory
+ * @mixin \Illuminate\Http\Client\PendingRequest
  */
 class Pool
 {
     /**
      * The factory instance.
      *
-     * @var \Illuminate\Http\Client\Factory
+     * @var \Illuminate\Http\Client\PendingRequest
      */
     protected $factory;
 
@@ -33,18 +33,18 @@ class Pool
     /**
      * Create a new requests pool.
      *
-     * @param  \Illuminate\Http\Client\Factory|null  $factory
+     * @param  \Illuminate\Http\Client\PendingRequest|null  $factory
      * @return void
      */
-    public function __construct(Factory $factory = null)
+    public function __construct(PendingRequest $factory = null)
     {
-        $this->factory = $factory ?: new Factory();
-
         if (method_exists(Utils::class, 'chooseHandler')) {
-            $this->handler = Utils::chooseHandler();
+            $handler = Utils::chooseHandler();
         } else {
-            $this->handler = \GuzzleHttp\choose_handler();
+            $handler = \GuzzleHttp\choose_handler();
         }
+
+        $this->factory = ($factory ?? new Factory())->async()->setHandler($handler);
     }
 
     /**
@@ -55,7 +55,7 @@ class Pool
      */
     public function as(string $key)
     {
-        return $this->pool[$key] = $this->asyncRequest();
+        return $this->pool[$key] = $this->newPendingAsyncRequest();
     }
 
     /**
@@ -63,9 +63,9 @@ class Pool
      *
      * @return \Illuminate\Http\Client\PendingRequest
      */
-    protected function asyncRequest()
+    protected function newPendingAsyncRequest()
     {
-        return $this->factory->setHandler($this->handler)->async();
+        return clone $this->factory;
     }
 
     /**
@@ -87,6 +87,6 @@ class Pool
      */
     public function __call($method, $parameters)
     {
-        return $this->pool[] = $this->asyncRequest()->$method(...$parameters);
+        return $this->pool[] = $this->newPendingAsyncRequest()->$method(...$parameters);
     }
 }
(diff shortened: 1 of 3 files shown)


taylorotwell:
It seems like there are code changes here that could be breaking to other applications / use cases?

src/Illuminate/Http/Client/Pool.php
```diff
@@ -55,17 +55,17 @@ public function __construct(Factory $factory = null)
...
     /**
      * Retrieve a new async pending request.
      *
      * @return \Illuminate\Http\Client\PendingRequest
      */
-    protected function asyncRequest()
+    protected function newPendingAsyncRequest()
     {
-        return $this->factory->setHandler($this->handler)->async();
+        return clone $this->factory;
```
taylorotwell, line 68: This doesn't seem to return a request instance?

src/Illuminate/Http/Client/Pool.php
```diff
@@ -5,14 +5,14 @@
...
  */
 class Pool
 {
     /**
      * The factory instance.
      *
-     * @var \Illuminate\Http\Client\Factory
+     * @var \Illuminate\Http\Client\PendingRequest
      */
     protected $factory;
```
taylorotwell, line 17: If this is a request, why is the variable $factory?

wfeller: It's actually a PendingRequest template that will be cloned in each pool request. I kept the name since it's like a template for each created request. Lmk if there's a more adequate name.

wfeller:
> It seems like there are code changes here that could be breaking to other applications / use cases?

@taylorotwell, I don't think so. Do you have anything in mind?

src/Illuminate/Http/Client/Pool.php
```diff
@@ -33,18 +26,18 @@ class Pool
     /**
      * Create a new requests pool.
      *
-     * @param  \Illuminate\Http\Client\Factory|null  $factory
+     * @param  \Illuminate\Http\Client\PendingRequest|null  $factory
      * @return void
      */
-    public function __construct(Factory $factory = null)
+    public function __construct(PendingRequest $factory = null)
```
driesvints, line 32: This is a major breaking change.

src/Illuminate/Http/Client/Pool.php
driesvints: I agree with @taylorotwell, this is highly confusing. We shouldn't change all of this imo in 9.x but in 10.x at the earliest (if we're going to do this).

src/Illuminate/Http/Client/Pool.php
wfeller: Only if people are directly creating pools without going through the `Http::pool()` method. Tbh, we could also just remove the type in the constructor and it would work the same. In the end, `$this->factory = ($factory ?? new Factory())->async()->setHandler($handler);` will always be a `PendingRequest`, whether `$factory` is a `Factory` or `PendingRequest`.

If you want, I can split this PR into 2 PR:
1. Resolving the issue with the `setHandler` method on the `PendingRequest` (which prevents middleware from being registered inside pools)
2. Pool setup not being passed to the requests created from the pool

henzeb: Just one question, this change of yours won't break the multiCurlHandler functionality where all pending requests are executed concurrently, right?

taylorotwell:
One thing I want to get to the bottom of - when Http::pool and async requests were first added by @cerbero90 ... why was it a goal to use the same client instance. It's explicitly stated in the original PR that it was intended for some reason but never explained why:

https://github.com/laravel/framework/pull/36948

Any ideas?

cerbero90:
@taylorotwell the reason behind the design choice of using the same client instance is related to the usage of the `CurlMultiHandler`.

Under the hood, async requests are handled as sub-processes of the main process started by the `CurlMultiHandler`.

If we were instantiating a different client for each request, we would also have one instance of `CurlMultiHandler` per request, making each request - in fact - synchronous.

On the other hand, having only one client instance makes sure that the same `CurlMultiHandler` handles all the requests as sub-processes of the same main process.

wfeller:
Currently (without this PR), all pending requests in a pool have their own Client instance, but a shared Handler instance.

The only difference with this PR is **when** the guzzle client is instantiated in the pending request.

See the changes to the `setHandler` and `buildHandlerStack` methods on the `PendingRequest`.

cerbero90:
Several changes have been applied to my original design.

Haven't kept track of them but as long as the underneath handler is the same, the requests should be handled asynchronously.

An easy way to manually test whether async requests are handled correctly is hitting endpoints delayed on purpose:

```php
// the total waiting time should be ~3 seconds instead of 15

    $pool->get('https://httpbin.org/delay/3')->then(/* ... */),
    $pool->get('https://httpbin.org/delay/3')->then(/* ... */),
    $pool->get('https://httpbin.org/delay/3')->then(/* ... */),
    $pool->get('https://httpbin.org/delay/3')->then(/* ... */),
    $pool->get('https://httpbin.org/delay/3')->then(/* ... */),
]);
```

taylorotwell:
I've merged part 1 of this PR here: https://github.com/laravel/framework/pull/44179

I need to address the second issue (initial setup ignored) as a separate PR.

I have confirmed that part 1 does not break the async nature of pools.

wfeller:
Thanks, @taylorotwell!

PR #43792, opened 2022-08-20
[9.x] add die directive

faissaloux:
Hello 👋🏻 

In this PR I have added a die directive so we can use `@die()` or `@die($var)` on blade


rodrigopedra:
What is the advantage over the `@dd()` directive?

faissaloux:
@rodrigopedra Actually we can use both the same.
I have just tried to use it after dumping on a loop and found that it doesn't exist, sooo I thought to add it to the framework

taylorotwell:

PR #43796, opened 2022-08-21
Uses `\Illuminate\Support\Stringable` instead `string`

mahmoudmohamedramadan:
This PR aims to make the checking process about request info clear, for illuminate, take a look over the next code

## In the old version

```PHP
if (str_contains($request()->url(), "create")) {
   // do something here...
}
```

## After this PR gets merged

```PHP
if ($request->url()->contains("create")) {
    // do something here...
}
```

I see that the second way is clear, because it makes the code more readable, and so on with the others methods.


taylorotwell:
I don't think I will make this breaking change.

PR #43797, opened 2022-08-21
[9.x] feature: `min_digits` and `max_digits` validation

danharrin:
As per the [`size` rule](https://laravel.com/docs/9.x/validation#rule-size), the number of digits in numeric input cannot be validated traditionally using `min` and `max`.

Currently, we have, `digits` to validate the exact length of a number, and `digits_between` to validate the length within two boundaries.

However, as far as I can see there is not yet a way to validate that a number has a minimum OR maximum number of digits, just both.

This PR introduces two validation rules - `min_digits` and `max_digits` for this use:

```php
Validator::validate([
    'number' => 1000,
], [
    'number' => [
        'min_digits:3', 'max_digits:5', // Passes as `1000` has 4 digits
        'min:3', 'max:5', // Fails as `1000` is greater than 5
    ],
])
```

### DIFF
--- a/src/Illuminate/Validation/Concerns/ReplacesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ReplacesAttributes.php
@@ -127,6 +127,20 @@ trait ReplacesAttributes
     }
 
     /**
+     * Replace all place-holders for the min digits rule.
+     *
+     * @param  string  $message
+     * @param  string  $attribute
+     * @param  string  $rule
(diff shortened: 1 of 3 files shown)


danharrin:
The failing 8.1 test is unrelated and needs rerunning.

danharrin:
Hey @taylorotwell do you need laravel/laravel and laravel/docs PRs for this or are you handling that?

PR #43801, opened 2022-08-21
[9.x] Restores original table name guessing for Morph Pivot model

### LINKED ISSUE #43800: Morph pivot table name conventions wrong when using MorphPivot model
- Laravel Version: 9.25.1
- PHP Version: 8.1.9

### Description:

You can tell a `morphToMany` or `morphedByMany` relations to use a Morph Pivot model by using a third parameter as the Model class name. This will also set the pivot table from it.

```php
return $this->morphToMany(Tag::class, 'taggable', Taggable::class).
```

The problem is that Eloquent tries to find the `subscription` table, where as it should be `subscriptions`. This is becase [`getTable()` in a `Pivot|MorphPivot`](https://github.com/laravel/framework/blob/b8db0d27f7dbefa0f066a7ca6472ea85e8fe2acc/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php#L155-L169) model doesn't returns the model plural, but rather the snake-singular.

```php
class Post extends \Illuminate\Database\Eloquent\Model {
    public function tags()
    {
        return $this->morphToMany(Tag::class, 'taggable', Taggable::class);
    }
}

Post::find(1)->tags()->dd();

// "select * from "tags" 
//    inner join "taggable" on "tags"."id" = "taggable"."tag_id" 
//    where 
//        "taggable"."taggable_id" = ? 
//    and 
//        "taggable"."taggable_type" = ?"
```

Here, the table should be `taggables` instead of `taggable`. This would force the developer to, unexplainably, set the table name property in the pivot model.

### What it should happen?

If the developer is setting the morph pivot model, it should always use the plural instead of a two-model-junction to guess the table. This shouldn't affect the normal pivot model (for belongs to many).

DarkGhostHunter:
## What?

When calling a many-to-many relations using morph pivot model, the table that Laravel guesses is snake singular from the pivot model name. For example, "Taggable" will point to `taggable` instead of `taggables`.

```php
use Illuminate\Database\Eloquent\Relations\MorphPivot;

class Post extends Model
{
    public function tags()
    {
        return $this->morphToMany(Tag::class, 'taggable', Taggable::class);
    }
}

class Tag extends Model
{
    // ...
}

class Taggable extends MorphPivot
{
    // ...
}

Post::find(1)->tags()->dd();

// "select * from "tags" 
//    inner join "taggable" on "tags"."id" = "taggable"."tag_id" 
//    where 
//        "taggable"."taggable_id" = ? 
//    and 
//        "taggable"."taggable_type" = ?
```

This patch adds back the plural table name guesser to `MorphPivot`, which will point to the plural name of the model rather than a the singular snake (which doesn't make sense). It will also avoid the developer to set the table manually when is the plural of the model name.

> **Note** I decided against removing the method from the `AsPivot` trait because it would break something, but it could be done for 10.x I supposse.

This doesn't affect monomorphic Pivot, as is inferred the developer is using the convention `{Bar}_{Foo}`, which if not it should point the table name manually.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/MorphPivot.php
+++ b/src/Illuminate/Database/Eloquent/Relations/MorphPivot.php
@@ -2,7 +2,6 @@
 
 namespace Illuminate\Database\Eloquent\Relations;
 
-use function class_basename;
 use Illuminate\Support\Str;
 
 class MorphPivot extends Pivot
--- a/src/Illuminate/Database/Eloquent/Relations/MorphPivot.php
+++ b/src/Illuminate/Database/Eloquent/Relations/MorphPivot.php
@@ -2,9 +2,8 @@
 
 namespace Illuminate\Database\Eloquent\Relations;
 
-use Illuminate\Support\Str;
 use function class_basename;
-use function str_replace;
+use Illuminate\Support\Str;
 
 class MorphPivot extends Pivot
 {
--- a/src/Illuminate/Database/Eloquent/Relations/MorphPivot.php
+++ b/src/Illuminate/Database/Eloquent/Relations/MorphPivot.php
@@ -2,6 +2,10 @@
 
 namespace Illuminate\Database\Eloquent\Relations;
 
+use Illuminate\Support\Str;
+use function class_basename;
+use function str_replace;
+
 class MorphPivot extends Pivot
 {
     /**
@@ -106,6 +110,16 @@ class MorphPivot extends Pivot
         $this->morphClass = $morphClass;
 
         return $this;
+    }
+
+    /**
+     * Get the table associated with the model.
+     *
+     * @return string
+     */
+    public function getTable()
+    {
+        return $this->table ?? Str::snake(Str::pluralStudly(class_basename($this)));
     }
 
     /**
--- a/tests/Database/DatabaseEloquentPivotTestModelPluralDummies.php
+++ b/tests/Database/DatabaseEloquentPivotTestModelPluralDummies.php
@@ -6,6 +6,7 @@ use Illuminate\Database\Connection;
 use Illuminate\Database\ConnectionResolverInterface;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Pivot;
+use Illuminate\Database\Eloquent\Relations\MorphPivot;
 use Illuminate\Database\Query\Grammars\Grammar;
 use Illuminate\Database\Query\Processors\Processor;
 use Mockery as m;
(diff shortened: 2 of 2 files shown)


driesvints:
@DarkGhostHunter please do not link to a issue and provide a thorough explanation in your PR description (see the PR template).

DarkGhostHunter:
> More info? When was this bug introduced? Tests?

Added info. Dunno when was introduced, I just picked it up yesterday when I saw failing queries on staging. Added Tests.

> @DarkGhostHunter please do not link to a issue and provide a thorough explanation in your PR description (see the PR template).

// @see above

taylorotwell:
I don't think I want to change this now since it could break applications.

rodrigopedra:
One workaround for now, is to call `BelongsToMany@using` on the relation definition, and don't pass the `$table` name argument as a FQCN:

~~~php
class Post extends Model
{
    public function tags()
    {
        return $this->morphToMany(Tag::class, 'taggable')->using(Taggable::class);
    }
}
~~~

The relation above will generate a SQL query using the `taggables` table name for the pivot table, and the results will be hydrated into a `Taggable` class instance.

Also the behavior of allowing a FQCN into the `$tableName` parameter to automatically guess the pivot table name and set the pivot model is not documented (I learned it from this PR)

https://laravel.com/docs/9.x/eloquent-relationships#defining-custom-intermediate-table-models

One last thing you can try, is to explicitly set the `Taggable` class' `$table` property

rodrigopedra:
> When was this bug introduced? Tests?

It seems this bug is present since the beginning. 

From looking into git blames on `BelongsTo Many` class and `AsPivot` trait, it seems passing a `MorphPivot` instead of a `Pivot` as the `$tableName` parameter was overseen.

I couldn't find no tests related on using a `MorphPivot` FQCN as the third parameter to define a relation, nor a test generating a SQL directly from a `MorphPivot` instance.

As this is not documented behavior, I would send it to master and add a note on the upgrade guide. I agree changing a behavior that was always present on a patch release seems not to be a good idea.

DarkGhostHunter:
> > When was this bug introduced? Tests?
> 
> It seems this bug is present since the beginning.
> 
> From looking into git blames on `BelongsTo Many` class and `AsPivot` trait, it seems passing a `MorphPivot` instead of a `Pivot` as the `$tableName` parameter was overseen.
> 
> I couldn't find no tests related on using a `MorphPivot` FQCN as the third parameter to define a relation, nor a test generating a SQL directly from a `MorphPivot` instance.
> 
> As this is not documented behavior, I would send it to master and add a note on the upgrade guide. I agree changing a behavior that was always present on a patch release seems not to be a good idea.

I agree. This should be documented for the next version because it was never clear.

PR #43803, opened 2022-08-22
Bug fix Paginator hasMorePages()

s-damian:
Bug fix for **hasMorePages()** of Paginator for simplePaginate() if items are not a Collection.

**Bug fix:**
(This bug exists when items are not a Collection)
There was a bug in **testSimplePaginatorReturnsRelevantContextInformation()** of **PaginatorTest**.
Because in this test we simulate: there are 3 items, 2 items per page = there are 2 pages. And we are on page 2 = we are on the last page.
I fixed this bug. And I added the test "testPaginatorisOnFirstAndLastPage()" in order to better perpetuate it.

### DIFF
--- a/src/Illuminate/Pagination/Paginator.php
+++ b/src/Illuminate/Pagination/Paginator.php
@@ -65,9 +65,17 @@ class Paginator extends AbstractPaginator implements Arrayable, ArrayAccess, Cou
      */
     protected function setItems($items)
     {
-        $this->items = $items instanceof Collection ? $items : Collection::make($items);
+        if ($items instanceof Collection) {
+            $this->items = $items;
 
-        $this->hasMore = $this->items->count() > $this->perPage;
+            $this->hasMore = $this->items->count() > $this->perPage;
+        } else {
+            $this->items = Collection::make($items);
+
+            $lastPage = ceil($this->items->count() / $this->perPage);
+
+            $this->hasMore = $this->currentPage < $lastPage;
+        }
 
         $this->items = $this->items->slice(0, $this->perPage);
     }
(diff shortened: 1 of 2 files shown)


s-damian:
@taylorotwell 
Hello.
I just fixed you run bug.
Hope this helps you.
Thanks.

dennisprudlo:
Correct me if I am wrong but I think the Paginator class does what it is supposed to. The `$items` property seems to be the items visible **on that page**. So when passing `[ 'a', 'b', 'c' ]` – _doesn't matter whether it's an array or a collection_ – like this:

```php
new Paginator([ 'a', 'b', 'c' ], perPage: 2, currentPage: 2);
```

the Paginator slices the items so it's only a set of `perPage`-items and the rest is being discarded.
https://github.com/laravel/framework/blob/24e3442e7617a9d04f604c68e3e489754b8a2322/src/Illuminate/Pagination/Paginator.php#L80

So It's not like `a` and `b` are on page 1 and `c` is on page 2, but rather `a` and `b` are on page 2 and since there are more items in the `$items` property the paginator knows it `$hasMore` items 😉. This is what you see, when we dump that paginator:

```php
Illuminate\Pagination\Paginator {#2601
  #items: Illuminate\Support\Collection {#2600
    #items: array:2 [
      0 => "a"
      1 => "b"
    ]
    #escapeWhenCastingToString: false
  }
  #perPage: 2
  #currentPage: 2
  ...
  #hasMore: true
}
```

We are **on page 2** and the items on that page are `a` and `b`. But since the initial items array **had more items** the `$hasMore` value is `true`

Tofandel:
> visible on that page. ... Paginator slices the items so it's only a set of perPage-items and the rest is being discarded.

That is counterintuitive, but yes that slice doesn't lie, this is expected behavior

taylorotwell:

PR #43805, opened 2022-08-22
Make string padding UTF-8 safe (bugfix)

martenkoetsier:
# The issue

In the `Str::pad*` functions, the pad string could be truncated in the middle of a unicode code point. E.g. (in Tinker):
```
for ($i = 3; $i < 7; $i++) echo Str::padRight('abc', $i, '❤'), PHP_EOL;
```
results in:
```
abc
abc<E2>
abc<E2><9D>
abc❤
```
This PR aims to fix this for the `padRight`, `padLeft` and `padBoth` functions.

# Tests

As the unit test does not yet test for these cases (UTF-8 characters in the `$pad` string), this PR also adds assert tests for this.

# Notes

As [noted](https://www.php.net/manual/en/function.str-pad.php#116164) on the PHP documentation "User Contributed Notes":

> since the default pad_type is STR_PAD_RIGHT. using STR_PAD_BOTH were always favor in the right pad if the required number of padding characters can't be evenly divided.

This PR takes this behavior into account.

# Impact

This PR should not impact code using the `Str::pad*` functions with a single-byte-coded pad-string (or the default string).
Users who used a UTF-8 code point in the pad string, should have encountered an issue and either avoided it or repaired it themselves in their own project.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -608,7 +608,13 @@ class Str
      */
     public static function padBoth($value, $length, $pad = ' ')
     {
-        return str_pad($value, strlen($value) - mb_strlen($value) + $length, $pad, STR_PAD_BOTH);
+        $short = max(0, $length - mb_strlen($value));
+        $short_left = floor($short / 2);
+        $short_right = ceil($short / 2);
+        return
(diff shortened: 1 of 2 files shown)


taylorotwell:
Merged.

Issue #43804, opened 2022-08-22
cookie()->forever(...) create cookie at most for one year


Saeed-Pooyanfar:
- Laravel Version: 9.25.1
- PHP Version: 8.1.9
- Browser: Google chrome 104.0.5112.101 (Official Build) (64-bit)

### Current behavior:

cookie()->forever('MyCookie', 'Test', secure: true) will create `MyCookie` in browser at most for 1 year

### Expected behavior:

cookie()->forever('MyCookie', 'Test', secure: true) must create `MyCookie` in browser for 5 years:
[2628000 minutes => 5 years](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Cookie/CookieJar.php#L88)

### Steps To Reproduce:

1- Create a resource controller
2- Assign `GET /test` route to show method
3- In show method:
```php
public function show()
{
    return response()->json(['test' => 'yes'])
        ->withCookie(cookie()->forever('MyCookie', 'Test', secure: true));
}
```
4- Open `/test` in google chrome
5- `MyCookie` with `Test` value will create in browser for only 1 year instead of 5 years

driesvints:
Seems like 400 days is the maximum amount of time so I guess Chrome limits this automatically: https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#section-4.1.2.1

driesvints:
Sent in a PR to update the comment. Thanks for making us aware.

PR #43806, opened 2022-08-22
[9.x] Adjust forever time for cookies

driesvints:
The maximum amount of days for an Expiration date of a cookie is apparently limited to 400 days.

See https://httpwg.org/http-extensions/draft-ietf-httpbis-rfc6265bis.html#section-4.1.2.1

Also https://github.com/laravel/framework/issues/43804


Tofandel:
Just changing a comment is not enough to change the value 😅 

`2628000` needs to be changed to `576000` (in forget as well I guess)

driesvints:
Done thanks. I left the `forgot` value for BC reasons if there are some browsers that still have set cookies with 5 years expiry.

PR #43814, opened 2022-08-22
[9.x] Load relationships using LazyCollection

hschimpf:
Hello everyone,
This will be my first contribution to the Laravel Framework.

This PR adds the ability to enable loading a relation using LazyCollection, why?

## Motivation
I'm working on a project that has a lot of data, our database has data since 1970.

On the table where we are fetching the main info has almost 300k records, over that we can load using the standard `lazy()` method of the model, but when we add the lazy loading of relations through `with(...)` we were facing memory issues, because we are loading relations up to 3 levels depth. For example, the first relation has between 130 and 150 records per parent, below that we have the second relation that has 25-50 records.

On some processes that runs once per month, we process a bit chunk of data from the database for reporting and other stuff. We tried to optimize the loading times using LazyCollection on the referenced relations. We see that loading a big resultset of data takes more time than loading in chunks (it's not much, but the CPU is happy that can wait waiting for next chunk 😄). Also as mentioned, we were getting memory.

## Examples
Instead of changing the core functionality of [Relation](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Relations/Relation.php#L158), I added a modifier to enable lazy on a single relation on demand:
```php
// get all Lands that where sold in the last month
$lands = SoldLand::with([
    // load all installments that were pay
    'installments' => fn($installment) => $installment->isPay()->with([
        // load payments of installments (load relation using LazyCollection)
        'payments' => fn($payment) => $payment->asLazy(),
    ]),
])->soldInLastMonth()->lazy();
```

## Breaking changes
This PR will break libraries that extends functionality over Illuminate/Database.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/BelongsTo.php
+++ b/src/Illuminate/Database/Eloquent/Relations/BelongsTo.php
@@ -8,6 +8,7 @@ use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Concerns\ComparesRelatedModels;
 use Illuminate\Database\Eloquent\Relations\Concerns\InteractsWithDictionary;
 use Illuminate\Database\Eloquent\Relations\Concerns\SupportsDefaultModels;
+use Illuminate\Support\LazyCollection;
 
 class BelongsTo extends Relation
 {
@@ -160,11 +161,11 @@ class BelongsTo extends Relation
      * Match the eagerly loaded results to their parents.
      *
      * @param  array  $models
-     * @param  \Illuminate\Database\Eloquent\Collection  $results
+     * @param  \Illuminate\Database\Eloquent\Collection | \Illuminate\Support\LazyCollection  $results
      * @param  string  $relation
      * @return array
      */
-    public function match(array $models, Collection $results, $relation)
+    public function match(array $models, Collection | LazyCollection $results, $relation)
     {
         $foreign = $this->foreignKey;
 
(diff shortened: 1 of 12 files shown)


hschimpf:
Sorry about the force push, I had a wrong GPG key configured

src/Illuminate/Database/Eloquent/Relations/Relation.php
```diff
@@ -69,6 +70,13 @@ abstract class Relation implements BuilderContract
      */
     protected static $selfJoinCount = 0;
 
+    /**
+     * Flag to enable loading this relation as a LazyCollection.
+     *
+     * @var false|int Chunk size for the lazy collection, false to load relation as normal collection
+     */
+    private false|int $lazy_chunkSize = false;
```
danilopinotti, line 78: camelCase?

```suggestion
    private false|int $lazyChunkSize = false;
```

hschimpf: Sure! Change was applied

src/Illuminate/Database/Eloquent/Relations/BelongsTo.php
```diff
@@ -160,11 +161,11 @@ public function initRelation(array $models, $relation)
      * Match the eagerly loaded results to their parents.
      *
      * @param  array  $models
-     * @param  \Illuminate\Database\Eloquent\Collection  $results
+     * @param  \Illuminate\Database\Eloquent\Collection|\Illuminate\Support\LazyCollection  $results
      * @param  string  $relation
      * @return array
      */
-    public function match(array $models, Collection $results, $relation)
+    public function match(array $models, Collection|LazyCollection $results, $relation)
```
michael-rubel, line 168: Perhaps it should be sent to the `10.x` because of this BC.

taylorotwell:

PR #43823, opened 2022-08-23
[9.x] Improve test for set method in RepositoryClass

moharami:


tests/Config/RepositoryTest.php
```diff
@@ -155,9 +155,20 @@ public function testSetArray()
         $this->repository->set([
             'key1' => 'value1',
             'key2' => 'value2',
+            'key3',
```
carlos-ea, line 158: `key3` here is not really a key but a value, this is the same as doing `[0 => 'key3']`

PR #43040, opened 2022-07-02
[9.x] Bind a Vite Null Object to the Container instead of a Closure

mabdullahsari:
Closes #43039

PR #43826, opened 2022-08-23
[9.x] Fix withoutVite

iamgergo:
This PR fixes the `withoutVite` method.

-----

With the latest release some `Vite` methods were added, however currently they are not compatible with the `withoutVite` testing method. 

For example in an app.blade.php:

```blade
    {{
        Vite::withEntryPoints('resources/js/app.js')
            ->useBuildDirectory('vendor/package-name/build')
            ->useHotFile(public_path('vendor/package-name/hot'))
    }}
```

The following error is coming up: 

```
Error: Call to a member function useBuildDirectory() on string in ...
```

The methods that are returning other value type than a string for example `$this` must be defined. Also, the `__toString` is added to the anonymus class to make sure blade can render it.

------

I was not sure how to test it since no other tests were added to the similar fake methods.

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/InteractsWithContainer.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/InteractsWithContainer.php
@@ -141,17 +141,17 @@ trait InteractsWithContainer
                 return $this;
             }
 
-            public function useBuildDirectory($path)
+            public function useBuildDirectory()
             {
                 return $this;
             }
 
-            public function useHotFile($path)
+            public function useHotFile()
             {
                 return $this;
             }
 
-            public function withEntryPoints($entryPoints)
+            public function withEntryPoints()
             {
                 return $this;
             }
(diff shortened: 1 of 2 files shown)


mabdullahsari:
I opened the PR for the [very first fix](https://github.com/laravel/framework/pull/43040).

I think that it'd be much better if an instance of `$this` is returned from `__call`. Then the `__toString()` can take on the responsibility for rendering nothing in the blade views. 

Thoughts?

iamgergo:
Yes, I think you are right, it's a clever solution, thanks! 

So, only `__call`, `__invoke` and `__toString` should stay, right? `__call` and `__invoke` should return with `$this` and `__toString` with `''`. It is correct?

iamgergo:
Okay, so for now, I leave it as is, because I'm not sure what side effects this refactoring would have.

Thank you!

mabdullahsari:
> Yes, I think you are right, it's a clever solution, thanks!
> 
> So, only `__call`, `__invoke` and `__toString` should stay, right? `__call` and `__invoke` should return with `$this` and `__toString` with `''`. It is correct?

Yup, that's correct.

PR #43838, opened 2022-08-24
[9.x] Improve @return value for `Illuminate\Database\Query\Builder::get()`

zoispag:
## Change

`Illuminate\Database\Query\Builder::get()` should return a `\Illuminate\Database\Eloquent\Collection` instance.

## Motivation

I have a class property with the type `Illuminate\Database\Eloquent\Collection`.

In the constructor, I populate it with `Book::all()`, and works as expected.

Recently I had to add one condition to my query, so I replaced `Book::all()` to `Book::whereNull('column')->get()` and I get the following complain from PHPStorm:

image

Both statements return objects of `Illuminate\Database\Eloquent\Collection`.

## Any considerations? 

- Should I completely drop `\Illuminate\Support\Collection` or the Union is actually fine?

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -2610,7 +2610,7 @@ class Builder implements BuilderContract
      * Execute the query as a "select" statement.
      *
      * @param  array|string  $columns
-     * @return \Illuminate\Support\Collection
+     * @return \Illuminate\Support\Collection|\Illuminate\Database\Eloquent\Collection
      */
     public function get($columns = ['*'])
     {


Jubeki:
For me it seems like this always calls `collect()` which always returns a `Illuminate\Support\Collection`

Please take a look at the Builder for Eloquent Models: https://github.com/laravel/framework/blob/cde1943666f6dccd9ddf9bbb639a1ce2844f541c/src/Illuminate/Database/Eloquent/Builder.php#L676-L688
which should return the `Illuminate\Database\Eloquent\Collection` instance.

taylorotwell:
That method doesn't return Eloquent collections.

zoispag:
> That method doesn't return Eloquent collections.

@taylorotwell what am I seeing in tinker/Tinkerwell then?

image

image

hamedghaderi:
@taylorotwell Hasn't `get` supposed to return an Eloquent\Collection? 🤔

dennisprudlo:
@zoispag @hamedghaderi the eloquent builder returns an eloquent collection, that's right. However, this PR updates the get method on the base query builder which returns a regular collection.

Just look into the method where you updated the return type:
```php
return collect(...)
```

The collect helper always returns a regular collection.

canvural:
> > That method doesn't return Eloquent collections.
> 
> @taylorotwell what am I seeing in tinker/Tinkerwell then?
> 
> image
> 
> image

You are seeing this: https://github.com/laravel/framework/blob/d23de5f9af1bd253a9d8b823cec9bf37987facf2/src/Illuminate/Database/Eloquent/Builder.php#L676

PR #43843, opened 2022-08-24
[9.x] Direct stream download

deleugpn:

The fact that the `ResponseFactory` class only accepts a `callback` is slightly annoying as we have to manage the process of dumping the contents of the stream onto the response inside project code. This small change enhances Laravel to take on a PSR Stream object and directly output it as a download onto the browser.

### DIFF
--- a/src/Illuminate/Routing/ResponseFactory.php
+++ b/src/Illuminate/Routing/ResponseFactory.php
@@ -167,7 +167,6 @@ class ResponseFactory implements FactoryContract
      *
      * @param  StreamInterface $stream
      * @param  string $name
-     *
      * @return  StreamedResponse
      */
     public function directStreamDownload(StreamInterface $stream, string $name)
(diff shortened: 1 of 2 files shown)


src/Illuminate/Routing/ResponseFactory.php
```diff
@@ -161,6 +162,20 @@ public function streamDownload($callback, $name = null, array $headers = [], $di
...
+     * Directly output stream as an HTTP Response.
+     *
+     * @param  StreamInterface  $stream
+     * @param  string  $name
+     * @return StreamedResponse
+     */
+    public function directStreamDownload(StreamInterface $stream, string $name)
+    {
+        return $this->streamDownload(function () use ($stream) {
+            echo $stream;
```
rodrigopedra, line 175: If the file is too big, wouldn't this load it all into memory? As pointed on the PSR docs?

https://github.com/php-fig/http-message/blob/efd67d1dc14a7ef4fc4e518e7dee91c271d524e4/src/StreamInterface.php#L20

Isn't the idea of outputting a stream the possibility of chunking its output? Something like this:

~~~php
return $this->streamDownload(function () use ($stream) {
    if ($stream->isSeekable()) {
        $stream->rewind();
    }

    while (! $stream->eof()) {
        echo $stream->read(4096);
    }
});
~~~

rodrigopedra: Or with `fpassthru($stream->detach())` instead of the `while` loop.

taylorotwell:

PR #43844, opened 2022-08-24
[9.x] Concurrent Safe firstOrCreate

deleugpn:

As pointed out on this post a few years back https://freek.dev/1087-breaking-laravels-firstorcreate-using-race-conditions, Eloquent's `firstOrCreate` is subject to race-condition between two separate requests. The summary of the problem is:

- Request 1 starts
- Request 2 starts
- Request 1 asks for firstOrCreate
- Request 2 asks for firstOrCreate
- Request 1 does not find record in DB
- Request 2 does not find record in DB
- Request 1 creates record
- Request 2 crashes due to duplicate unique key

`concurrentlyFirstOrCreate` will catch the database exception and attempt to load the record from the DB one more time, giving it a chance to find the record that was concurrently created by a separate process.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -12,6 +12,7 @@ use Illuminate\Database\Eloquent\Concerns\QueriesRelationships;
 use Illuminate\Database\Eloquent\Relations\BelongsToMany;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Query\Builder as QueryBuilder;
+use Illuminate\Database\QueryException;
 use Illuminate\Database\RecordsNotFoundException;
 use Illuminate\Pagination\Paginator;
 use Illuminate\Support\Arr;
@@ -554,6 +555,22 @@ class Builder implements BuilderContract
         return tap($this->newModelInstance(array_merge($attributes, $values)), function ($instance) {
             $instance->save();
         });
+    }
+
+    /**
+     * Concurrent-safe load first record matching the attributes or create it.
+     *
+     * @param  array  $attributes
+     * @param  array  $values
+     * @return \Illuminate\Database\Eloquent\Model|static
+     */
+    public function concurrentlyFirstOrCreate(array $attributes = [], array $values = [])
+    {
+        try {
+            return $this->firstOrCreate($attributes, $values);
+        } catch (QueryException $e) {
+            return $this->where($attributes)->firstOrFail();
+        }
     }
 
     /**


fgaroby:
So, we have to explicitly call the `concurrentlyFirstOrCreate` method to use it?
Why didn't you add the `try/catch` block directly in the `firstOrCreate`method?

deleugpn:
Because that could be slightly construed as a breaking change and this is rather an opt-in format for it

taylorotwell:
It seems like we would rather want firstOrCreate to just be as good as possible rather than introduce a new method?

src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -556,6 +557,22 @@ public function firstOrCreate(array $attributes = [], array $values = [])
...
+     *
+     * @param  array  $attributes
+     * @param  array  $values
+     * @return \Illuminate\Database\Eloquent\Model|static
+     */
+    public function concurrentlyFirstOrCreate(array $attributes = [], array $values = [])
+    {
+        try {
+            return $this->firstOrCreate($attributes, $values);
+        } catch (QueryException $e) {
```
jbrooksuk, line 571: The problem with catching `QueryException` is that we'll end up trying to re-do something that should have failed.

For example;

```php
User::firstOrCreate(['id' => 11]);
```

In the Laravel skeleton, the `name` field is also required, so now `firstOrCreate` will throw `QueryException` and we'll try again to find a matching model.

jbrooksuk (commented):
I'm not sure that this is a complete solution at the moment...

src/Illuminate/Database/Eloquent/Builder.php
jbrooksuk: > https://stackoverflow.com/questions/27878719/laravel-catch-eloquent-unique-field-error

This will change per database engine.

mikeydevelops:
So the main problem with adding this to the main `firstOrCreate` is that we could not be sure that the duplicate exception would be thrown. Maybe some data is missing or maybe the server went away and the DBMS complains about it.

I propose to extend `QueryException` and add a method (`isDuplicate` maybe?) to check if the exception code matches the error code for duplicate entries or if not supported, regex the thrown message and detect the unique syntax for each DBMS. Of course, this now means the `QueryException` should be aware of what driver is used to properly compare the error codes and messages.

If that is too much work, maybe just do it inline in the `firstOrCreate` method, but I would suggest against that and go with the previous suggestion because it would be useful for the developer and it allows for extension in the future for more similar cases.

bert-w:
Interesting, I'm using this snippet in my code:
```php
DB::transaction(function() use ($request) {
    $ip = Ip::query()->lockForUpdate()->firstOrCreate([
        'ip' => $request->ip()
    ]);
}, 3);
```
`'ip'` is a unique column in this example.

I now realize that this `lockForUpdate()` doesn't do anything because if an ip doesn't exist, there is nothing to row-lock. So the sole reason this snippet works is because of the `3` attempts parameter? The error never re-appeared after that attempts-parameter addition.

antonkomarev:
@bert-w you can use application-level advisory locks to prevent duplicates.

mikeydevelops:
> @bert-w you can use application-level advisory locks to prevent duplicates.

But then wouldn't this make the illuminate/database dependent on illuminate/cache for use of the locks? Also, we could not possibly know which cache driver a developer has chosen to use (I mean we can, if the code will be only in Laravel context and not only in illuminate/database).
Maybe we use the filesystem driver, there are deployments on slow drives which take time to create files which in turn slow down the request even more.
In short, this solution would tightly couple illuminate/database to the whole application which would mean that the feature cannot be used outside Laravel.

antonkomarev:
@mikeydevelops all popular RDBs has advisory locks functionality

antonkomarev:
I'm not proposing to use Advisory Locks in the framework, just pointing out that you can do it in your application.

garygreen:
This PR introduces a new method which fixes a race condition for just ONE function in Eloquent. The thing is, there is potentially race conditions everywhere... you could do a `find` and then `create` after manually and still have that exact same race condition.

So this could get messy real quick if Laravel introduces this, then others start contributing race-aware functions for everything.

I would be supportive of updating the `QueryException` to provide easier to interpret context as to why it failed, e.g. was it due to missing required value, or duplicate key. Some kind of helper functions on the exception like:

```
$queryException->wasDuplicate();
$queryException->wasMissingDefault();
```

That way it'll be easier to wrap your own eloquent functions (that may have race conditions) and catch the exception to then work out what action best to take.

DarkGhostHunter:
> This PR introduces a new method which fixes a race condition for just ONE function in Eloquent. The thing is, there is potentially race conditions everywhere... you could do a `find` and then `create` after manually and still have that exact same race condition.
> 
> 
> 
> So this could get messy real quick if Laravel introduces this, then others start contributing race-aware functions for everything.
> 
> 
> 
> I would be supportive of updating the `QueryException` to provide easier to interpret context as to why it failed, e.g. was it due to missing required value, or duplicate key. Some kind of helper functions on the exception like:
> 
> 
> 
> ```php
> 
> $queryException->wasDuplicate();
> 
> $queryException->wasMissingDefault();
> 
> ```
> 
> 
> 
> That way it'll be easier to wrap your own eloquent functions (that may have race conditions) and catch the exception to determine the best action to take.

This. Before this PR can have any leverage, the QueryException must be able to identify the error cause is a duplicate primary/unique key.

driesvints:
Closing this for now. Feel free to resend when you have more time.

PR #43845, opened 2022-08-24
[9.x] Compressed JSON Response

deleugpn:

This is particularly useful for projects running inside AWS Lambda. When an HTTP Response is bigger than 1MB, it causes AWS Lambda internal runtime to crash with a 502 which is hard to track down. The most convenient way to overcome that limitation is to gzip the response, which is widely supported by browsers and make the whole process transparent while still providing AWS with a smaller response size.

### DIFF
--- a/src/Illuminate/Routing/ResponseFactory.php
+++ b/src/Illuminate/Routing/ResponseFactory.php
@@ -91,13 +91,12 @@ class ResponseFactory implements FactoryContract
     /**
      * Create a new compressed JSON response instance. Requires php-zip extension.
      *
-     * @param array|string $data
-     * @param $status
-     * @param array $headers
-     * @param $options
+     * @param  array|string $data
+     * @param  $status
+     * @param  array $headers
      * @return Response
      */
-    public function compressedJson(array|string $data = [], $status = 200, array $headers = [], $options = 0)
+    public function compressedJson(array|string $data = [], $status = 200, array $headers = [])
     {
         if (! extension_loaded('zip')) {
             throw new RuntimeException('Gzip extension is not available.');
--- a/src/Illuminate/Routing/ResponseFactory.php
+++ b/src/Illuminate/Routing/ResponseFactory.php
@@ -9,6 +9,7 @@ use Illuminate\Http\Response;
 use Illuminate\Routing\Exceptions\StreamedResponseException;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
+use RuntimeException;
 use Symfony\Component\HttpFoundation\BinaryFileResponse;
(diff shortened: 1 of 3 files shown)


src/Illuminate/Routing/ResponseFactory.php
```diff
@@ -87,6 +88,33 @@ public function view($view, $data = [], $status = 200, array $headers = [])
...
+     * Create a new compressed JSON response instance. Requires php-zip extension.
+     *
+     * @param  array|string  $data
+     * @param  int  $status
+     * @param  array  $headers
+     * @return Response
+     */
+    public function compressedJson(array|string $data = [], $status = 200, array $headers = [])
+    {
+        if (! extension_loaded('zip')) {
```
rodrigopedra, line 101: Isn't the extension for GZip `ext-zlib` instead of `ext-zip`?

src/Illuminate/Routing/ResponseFactory.php
```diff
@@ -87,6 +88,33 @@ public function view($view, $data = [], $status = 200, array $headers = [])
...
+            throw new RuntimeException('Gzip extension is not available.');
+        }
+
+        if (is_array($data)) {
+            $data = json_encode($data);
+        }
+
+        $data = gzencode($data, 9);
```
rodrigopedra, line 109: Would you consider dropping the hard coded `$level` parameter? 

I usually never use a level greater than 5 as the output difference is negligible, but the CPU work required is not. Specially for large files.

If one prefer to always use level 9, they could configure it on their server's `php.ini` (`zlib.output_compression_level`)

src/Illuminate/Routing/ResponseFactory.php
jbrooksuk: Yeah, this needs to be `zlib`.

fgaroby: Same question [here](https://github.com/laravel/framework/pull/43845/files#diff-f69e5177a6a5562428bf2fa61f2175c94b7b13a398ac6aa65bc071ebf082ba83R34).

garygreen:
Isn't it better to just have this at the server level, e.g. compressing in nginx? Would be a lot faster too.

```php
gzip on;
gzip_types application/json
```

mikeydevelops:
> Isn't it better to just have this at the server level, e.g. compressing in nginx? Would be a lot faster too.
> 
> ```
> gzip on;
> gzip_types application/json
> ```

As I'm aware that on shared hosting you do not have access to server-level config.

driesvints:
What's the status here? If for now no work is being done it might be best to resubmit at a later time.

driesvints:
Closing this for now. Feel free to resend when you have more time.

PR #43846, opened 2022-08-24
[9.x] Feature whereLike added to Mathcing trait of AssertableJson + tests

hamidhaghdoost:
### Description 
The PR implements `$assertableJson->whereLike()` method for `Illuminate\Testing\Fluent\Concerns\Matching` trait, which gets the `key` and `expected` pattern and returns a boolean representing matched or not...

### Problem
In most of cases we need to check a particular pattern on a json response and we have to check it in callbacks maybe. This PR simply added a dedicated method for this purpose.
#### Sample usage
```php
$this->post('test-uri', [])->assertStatus(429)->assertJson(fn(AssertableJson $json) =>
    $json->whereLike('message', 'Try again * seconds later!')
        ->missing('data')
        ->etc();
);
```

Let me know your comments! 

Thanks <3

### DIFF
--- a/src/Illuminate/Testing/Fluent/Concerns/Matching.php
+++ b/src/Illuminate/Testing/Fluent/Concerns/Matching.php
@@ -5,6 +5,7 @@ namespace Illuminate\Testing\Fluent\Concerns;
 use Closure;
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Support\Collection;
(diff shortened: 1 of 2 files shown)


taylorotwell:
You can do this:

```php
$response->assertJson(function ($json) {
    $json->where('name', fn ($x) => str($x)->is('Taylor Otwell'));
});
```

PR #43852, opened 2022-08-25
Fixes the typos of `method` annotation

mahmoudmohamedramadan:
This PR updates the `method` annotation.


taylorotwell:
The router get method does return a Route though?

PR #43857, opened 2022-08-25
Add ListCommand with an option to hide vendor commands

KKSzymanowski:
Currently there's no way to hide vendor commands in `php artisan list` output.
When the application's commands aren't namespaced, it's difficult to see exactly what custom commands are available.
When the application's commands are however namespaced it's easy to forget about the namespace and adding the namespace every time one runs it gets annoying.

This aims to fix that by adding a `--no-vendor` option to the `php artisan list` command.

Output without `--no-vendor` option:
```
Available commands:
  about                  Display basic information about your application
  clear-compiled         Remove the compiled class file
  completion             Dump the shell completion script
  db                     Start a new database CLI session
  docs                   Access the Laravel documentation
  down                   Put the application into maintenance / demo mode
  env                    Display the current framework environment
  generate-some-report   Generate some report
  help                   Display help for a command
  list                   List commands
  migrate                Run the database migrations
  optimize               Cache the framework bootstrap files
  quick-command          Closure command defined in routes/console.php
  serve                  Serve the application on the PHP development server
  sync-some-data         Sync some data
  test                   Run the application tests
  tinker                 Interact with your application
  up                     Bring the application out of maintenance mode
 auth
  auth:clear-resets      Flush expired password reset tokens
...
```

Output with `--no-vendor` option:
```
Available commands:
  generate-some-report  Generate some report
  quick-command         Closure command defined in routes/console.php
  sync-some-data        Sync some data
```

As seen above this also correctly recognizes Closure Commands created in routes/console.php by checking the file where the Closure on which the command is based has been defined.

### DIFF
--- a/src/Illuminate/Foundation/Console/ClosureCommand.php
+++ b/src/Illuminate/Foundation/Console/ClosureCommand.php
@@ -79,4 +79,14 @@ class ClosureCommand extends Command
 
         return $this;
     }
+
+    /**
+     * Get the command callback.
+     *
+     * @return Closure
+     */
+    public function getCallback()
+    {
+        return $this->callback;
+    }
 }
--- a/src/Illuminate/Foundation/Console/ListCommand.php
+++ b/src/Illuminate/Foundation/Console/ListCommand.php
@@ -1,0 +1,49 @@
+<?php
+
+namespace Illuminate\Foundation\Console;
+
+use Illuminate\Console\Command;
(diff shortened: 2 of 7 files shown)


KKSzymanowski:
I don't think this breaks BC but I'm not 100% certain so I created this PR to the master branch. If you want I can move it over to the 9.x branch.

taylorotwell:
Honestly I think you should just namespace your app commands and do `php artisan list app`

PR #43859, opened 2022-08-25
[9.x] Fix Session missing() logic to match with description in documentation.

kenfai:
In the documentation under Session > https://laravel.com/docs/9.x/session#determining-if-an-item-exists-in-the-session, the `missing()` method has the following description:

> To determine if an item is not present in the session, you may use the `missing` method. The `missing` method returns `true` if the item is `null` or if the item is not present:

However, it cannot be correct if the `missing()` method is the inverse of `exists()`, since `exists()` will also return `true` if the item is `null`.

The `missing()` method should be defined as the inverse of `has()` instead, since `has()` will only return `true` if the item is present AND is not `null`.

The test cases have also been fixed to match with the described behavior in the documentation.

### DIFF
--- a/src/Illuminate/Session/Store.php
+++ b/src/Illuminate/Session/Store.php
@@ -265,7 +265,7 @@ class Store implements Session
      */
     public function missing($key)
     {
-        return ! $this->exists($key);
+        return ! $this->has($key);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
Feels like a breaking change on a patch release.

kenfai:
Should this be sent to the `master` branch then?

PR #43860, opened 2022-08-25
[9.x] Fix `is()` related model check

ildyria:
https://github.com/laravel/framework/blob/d23de5f9af1bd253a9d8b823cec9bf37987facf2/src/Illuminate/Database/Eloquent/Relations/Concerns/ComparesRelatedModels.php#L16-L19
In this part if the `id` of the model is `0`, then `compareKeys()` will return `false` even if the key matches.

```php
    /**
     * Compare the parent key with the related key.
     *
     * @param  mixed  $parentKey
     * @param  mixed  $relatedKey
     * @return bool
     */
    protected function compareKeys($parentKey, $relatedKey)
    {
        if (empty($parentKey) || empty($relatedKey)) {
            return false;
        }

        if (is_int($parentKey) || is_int($relatedKey)) {
            return (int) $parentKey === (int) $relatedKey;
        }

        return $parentKey === $relatedKey;
    }
```

This is because `empty(0)` is `true`. In this specific case, this is not the expected behavior.
This pull request simply replace the `empty($var)` check by ~~`$var === null || $var === ''`~~ `blank($var)`, cornering the two edge cases while leaving 0 as a valid id.

[Phpstan strict rules](https://github.com/phpstan/phpstan-strict-rules), while opinionated, discourage the use of `empty()`. :)

-----
### Rationale for this fix:
I am currently working on upgrading [Lychee](https://github.com/LycheeOrg/Lychee) to Laravel 9 and migrating from [Larapass](https://github.com/DarkGhostHunter/Larapass) to [WebAuthn](https://github.com/Laragear/WebAuthn). The later makes use of the `isNot()` function. In Lychee, currently user with id = `0` is the Admin, hence how I ended up finding this bug.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/ComparesRelatedModels.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/ComparesRelatedModels.php
@@ -64,7 +64,10 @@ trait ComparesRelatedModels
      */
     protected function compareKeys($parentKey, $relatedKey)
     {
-        if (empty($parentKey) || empty($relatedKey)) {
+        if ($parentKey === null
+            || $parentKey === ''
+            || $relatedKey === null
+            || $relatedKey === '') {
             return false;
         }
 
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/ComparesRelatedModels.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/ComparesRelatedModels.php
@@ -64,10 +64,7 @@ trait ComparesRelatedModels
      */
     protected function compareKeys($parentKey, $relatedKey)
     {
-        if ($parentKey === null
-            || $parentKey === ''
-            || $relatedKey === null
-            || $relatedKey === '') {
+        if (blank($parentKey) || blank($relatedKey)) {
             return false;
         }
 
--- a/tests/Database/DatabaseEloquentBelongsToTest.php
+++ b/tests/Database/DatabaseEloquentBelongsToTest.php
@@ -284,6 +284,69 @@ class DatabaseEloquentBelongsToTest extends TestCase
         $this->assertTrue($relation->is($model));
     }
 
+    public function testIsModelWithZeroParentKey()
+    {
+        $parent = m::mock(Model::class);
+
+        // when addConstraints is called we need to return the foreign value
+        $parent->shouldReceive('getAttribute')->once()->with('foreign_key')->andReturn('foreign.value');
+        // when getParentKey is called we want to return an integer
+        $parent->shouldReceive('getAttribute')->once()->with('foreign_key')->andReturn(0);
+
+        $relation = $this->getRelation($parent);
+
+        $this->related->shouldReceive('getConnectionName')->once()->andReturn('relation');
+
+        $model = m::mock(Model::class);
+        $model->shouldReceive('getAttribute')->once()->with('id')->andReturn('0');
+        $model->shouldReceive('getTable')->once()->andReturn('relation');
+        $model->shouldReceive('getConnectionName')->once()->andReturn('relation');
+
+        $this->assertTrue($relation->is($model));
+    }
+
+    public function testIsModelWithZeroRelatedKey()
+    {
+        $parent = m::mock(Model::class);
+
+        // when addConstraints is called we need to return the foreign value
+        $parent->shouldReceive('getAttribute')->once()->with('foreign_key')->andReturn('foreign.value');
+        // when getParentKey is called we want to return a string
+        $parent->shouldReceive('getAttribute')->once()->with('foreign_key')->andReturn('0');
+
+        $relation = $this->getRelation($parent);
+
+        $this->related->shouldReceive('getConnectionName')->once()->andReturn('relation');
+
+        $model = m::mock(Model::class);
+        $model->shouldReceive('getAttribute')->once()->with('id')->andReturn(0);
+        $model->shouldReceive('getTable')->once()->andReturn('relation');
+        $model->shouldReceive('getConnectionName')->once()->andReturn('relation');
+
+        $this->assertTrue($relation->is($model));
+    }
+
+    public function testIsModelWithZeroKeys()
+    {
+        $parent = m::mock(Model::class);
+
+        // when addConstraints is called we need to return the foreign value
+        $parent->shouldReceive('getAttribute')->once()->with('foreign_key')->andReturn('foreign.value');
+        // when getParentKey is called we want to return an integer
+        $parent->shouldReceive('getAttribute')->once()->with('foreign_key')->andReturn(0);
+
+        $relation = $this->getRelation($parent);
+
+        $this->related->shouldReceive('getConnectionName')->once()->andReturn('relation');
+
+        $model = m::mock(Model::class);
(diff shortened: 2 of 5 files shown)


ildyria:
Note that the issue is also present in 8.x.

src/Illuminate/Database/Eloquent/Relations/Concerns/ComparesRelatedModels.php
```diff
@@ -64,7 +64,10 @@ abstract protected function getRelatedKeyFrom(Model $model);
      */
     protected function compareKeys($parentKey, $relatedKey)
     {
-        if (empty($parentKey) || empty($relatedKey)) {
+        if ($parentKey === null
+            || $parentKey === ''
+            || $relatedKey === null
+            || $relatedKey === '') {
```
rodrigopedra, line 70: Did you consider to use the `blank()` helper? 

I guess these checks are fine, but `blank()` would keep it shorter and more legible, in my opinion

ildyria: I did not know that blank existed. I agree with you. :)

driesvints:
> In this part if the id of the model is 0

I don't think it's reasonable to assume 0 is a valid identifier.

mikeydevelops:
> I don't think it's reasonable to assume 0 is a valid identifier.

I don't it's reasonable for the framework to assume 0 is not a valid identifier. 0 is an integer the framework checks only the type of the identifier and not its value. What is a valid value is left for the developer to decide. If they want 0 they should be able to use 0.

driesvints:
https://stackoverflow.com/questions/8927761/why-is-negative-id-or-zero-considered-a-bad-practice

mikeydevelops:
> https://stackoverflow.com/questions/8927761/why-is-negative-id-or-zero-considered-a-bad-practice

I get your point and I am all for following good practices. But if the developer wants to do something, it is good to know that the framework they use is able to support it even if it is a bad practice. I assume OP used 0 for their admin user because it is not possible for a duplicate id to be created in the future since auto increment always increments the row id.

bert-w:
I would strongly advise to exclude 0 for use with key columns, but then again it is a valid query so it should work nonetheless.

Funny sidenote: a micro-optimization is to simply use `if(!$parentKey || !$relatedKey)` since it is functionally the same to the `empty()` counterpart (both variables are always defined in this case since they are function parameters). The exclamation mark is considerably faster but the practical benefit is little since we are talking about a 10 microsecond optimization (1/100 of a millisecond)

taylorotwell:

ildyria:
Hi @taylorotwell ,

this pull request does not add code to Laravel but:
- provide a **one line fix** which correct what I consider a bug
- **add tests** for this specific edge case in order to avoid regressions.

It can therefore not be provided by an external package.

As per requested, please find the above mentioned description:

In [Illuminate/Database/Eloquent/Relations/Concerns/ComparesRelatedModels.php](https://github.com/laravel/framework/blob/d23de5f9af1bd253a9d8b823cec9bf37987facf2/src/Illuminate/Database/Eloquent/Relations/Concerns/ComparesRelatedModels.php#L16-L19)
```php
 public function is($model) 
 { 
     $match = ! is_null($model) && 
            $this->compareKeys($this->getParentKey(), $this->getRelatedKeyFrom($model)) && 
            ...
```

In this part if the `id` of the model is `0`, then `compareKeys()` will return `false` even if the key matches.

```php
    /**
     * Compare the parent key with the related key.
     *
     * @param  mixed  $parentKey
     * @param  mixed  $relatedKey
     * @return bool
     */
    protected function compareKeys($parentKey, $relatedKey)
    {
        if (empty($parentKey) || empty($relatedKey)) {
            return false;
        }

        if (is_int($parentKey) || is_int($relatedKey)) {
            return (int) $parentKey === (int) $relatedKey;
        }

        return $parentKey === $relatedKey;
    }
```

This is because `empty(0)` is `true`. In this specific case, this is not an expected behavior.
This pull request simply replace the `empty($var)` check by blank($var)`, cornering the two edge cases while leaving `0` as a valid id.

Some concerns where raised about the bad practice of using `0` as an id. I don't think this is ground for rejection of this pull request, especially given that Laravel supports `0` keys pretty much everywhere e.g. relations...

[Phpstan strict rules](https://github.com/phpstan/phpstan-strict-rules), while opinionated, discourage the use of `empty()`. :)

-----
### Rationale for this fix:
I am currently working on upgrading [Lychee](https://github.com/LycheeOrg/Lychee) to Laravel 9 and migrating from [Larapass](https://github.com/DarkGhostHunter/Larapass) to [WebAuthn](https://github.com/Laragear/WebAuthn). The later makes use of the `isNot()` function. In Lychee, currently user with id = `0` is the Admin, hence how I ended up finding this bug.

Hopefully this mail finds you well.

Kind regards

ildyria:
@taylorotwell I did as you requested, and provided you with the details. Could you please check again. This is not adding functionality, but fixing a bug and adding test cases.

ildyria:
Hi @taylorotwell ,

this pull request does not add code to Laravel but:
- provide a **one line fix** which correct what I consider a bug
- **add tests** for this specific edge case in order to avoid regressions.

It can therefore not be provided by an external package.

As per requested, please find the description:

In [Illuminate/Database/Eloquent/Relations/Concerns/ComparesRelatedModels.php](https://github.com/laravel/framework/blob/d23de5f9af1bd253a9d8b823cec9bf37987facf2/src/Illuminate/Database/Eloquent/Relations/Concerns/ComparesRelatedModels.php#L16-L19)
```php
 public function is($model) 
 { 
     $match = ! is_null($model) && 
            $this->compareKeys($this->getParentKey(), $this->getRelatedKeyFrom($model)) && 
            ...
```

In this part if the `id` of the model is `0`, then `compareKeys()` will return `false` even if the key matches.

```php
    /**
     * Compare the parent key with the related key.
     *
     * @param  mixed  $parentKey
     * @param  mixed  $relatedKey
     * @return bool
     */
    protected function compareKeys($parentKey, $relatedKey)
    {
        if (empty($parentKey) || empty($relatedKey)) {
            return false;
        }

        if (is_int($parentKey) || is_int($relatedKey)) {
            return (int) $parentKey === (int) $relatedKey;
        }

        return $parentKey === $relatedKey;
    }
```

This is because `empty(0)` is `true`. In this specific case, this is not an expected behavior.
This pull request simply replace the `empty($var)` check by `blank($var)`, cornering the two edge cases while leaving `0` as a valid id.

Some concerns where raised about the bad practice of using `0` as an id. I don't think this is ground for rejection of this pull request, especially given that Laravel supports `0` keys pretty much everywhere e.g. relations...

[Phpstan strict rules](https://github.com/phpstan/phpstan-strict-rules), while opinionated, discourage the use of `empty()`. :)

-----
### Rationale for this fix:
I am currently working on upgrading [Lychee](https://github.com/LycheeOrg/Lychee) to Laravel 9 and migrating from [Larapass](https://github.com/DarkGhostHunter/Larapass) to [WebAuthn](https://github.com/Laragear/WebAuthn). The later makes use of the `isNot()` function. In Lychee, currently user with id = `0` is the Admin, hence how I ended up finding this bug.

Hopefully this mail finds you well, I would really appreciate an answer.

Kind regards

PR #43862, opened 2022-08-25
[9.x] Flush stale cache data

### LINKED ISSUE #43850: Tags not cleaned up from Redis cache
- Laravel Version: 9.11.0
- PHP Version: 8.1.6

### Description:

When an item is saved into Redis cache with tags and an expiry (`Cache::tags(['one', 'two'])->put('key', 'value', 3600)`) the following happens:
 1. A value is saved under a generated key name with an expiry date: `laravel:6c017031f987fbd8828ea55044d67f6bb9dd6316:key` -> `s:5:"value";`
 2. A set of key names is saved under tag name without expiration: `laravel:6307213384953462982890:standard_ref` -> `[laravel:6c017031f987fbd8828ea55044d67f6bb9dd6316:key]`
 3. A generated string value is saved under tag name without expiration: `laravel:tag:one:key` -> `6307213384953462982890`

The value expires as it should, and Laravel handles it correctly. The problem are the other two keys which **don't** expire ever, because they do also take up space. This caused our production Redis to use more than a gigabyte of space in a week (just in stale tags alone, _without the values_) when we tried to use them to cache data on per-user basis.

Since Redis does not allow individual values in a set to have an expiry date, the [usual approach is to run a Lua script](https://github.com/stackify/redis-tagging-locking-sample/blob/master/src/RedisWithTaggingAndLocking/Scripts/CleanupTags.lua) to cleanup those tags manually. Laravel should introduce a command to run on a schedule that cleans up those leftover tag values.

### Steps To Reproduce:

1. Use `CACHE_DRIVER=redis`
2. Create tagged values that will expire shortly:
```php
Cache::tags(['one'])->put('key2', 'value', 2);
Cache::tags(['one', 'two'])->put('key1', 'value', 2)
```
3. After 2 seconds pass, execute `KEYS *` on your Redis and see the leftover tag keys (note the keys that hold the values are gone):
```
0)	laravel:tag:two:key
1)	laravel:tag:one:key
2)	laravel:6307238514508800660377:standard_ref
3)	laravel:63072380e1b1c629296883:standard_ref
```

oprypkhantc:
Resolves #43850

The "flush" naming was chosen over "clear" as this is the naming consistently used internally by Laravel (e.g. Cache::flush(), 
queue:flush), but I can change it to "clear" if that makes any difference.

Otherwise this should be a 100% backwards compatible.

### DIFF
--- a/src/Illuminate/Cache/Console/FlushStaleCommand.php
+++ b/src/Illuminate/Cache/Console/FlushStaleCommand.php
@@ -1,0 +1,97 @@
+<?php
+
+namespace Illuminate\Cache\Console;
+
+use Illuminate\Cache\CacheManager;
+use Illuminate\Console\Command;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Input\InputArgument;
+
+#[AsCommand(name: 'cache:flush-stale')]
+class FlushStaleCommand extends Command
+{
+    /**
+     * The console command name.
+     *
+     * @var string
+     */
+    protected $name = 'cache:flush-stale {store? : The store to clean up}';
+
+    /**
+     * The name of the console command.
+     *
+     * This name is used to identify the command during lazy loading.
+     *
+     * @var string|null
+     *
+     * @deprecated
+     */
+    protected static $defaultName = 'cache:flush-stale';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Flushes any stale data from the cache';
+
+    /**
+     * The cache manager instance.
+     *
+     * @var \Illuminate\Cache\CacheManager
+     */
+    protected $cache;
+
+    /**
+     * Create a new cache clear command instance.
+     *
+     * @param  \Illuminate\Cache\CacheManager  $cache
+     * @return void
+     */
+    public function __construct(CacheManager $cache)
+    {
+        parent::__construct();
+
+        $this->cache = $cache;
+    }
+
+    /**
+     * Execute the console command.
+     *
+     * @return void
+     */
+    public function handle()
+    {
+        $this->laravel['events']->dispatch(
+            'cache:flushing-stale', [$this->argument('store')]
+        );
+
+        $repository = $this->cache->store($this->argument('store'));
+
+        // Adding a flushStale() empty method into Store contract would be a breaking change,
+        // so instead we're just checking whether it implements the method directly.
+        if (!$repository->supportsFlushingStale()) {
+            return $this->warn('Given store does not support flushing stale data. Make sure the correct store name was given.');
+        }
+
+        $repository->flushStale();
+
+        $this->laravel['events']->dispatch(
+            'cache:flushed-stale', [$this->argument('store')]
+        );
+
+        $this->info('Flushed stale cache data successfully.');
+    }
+
+    /**
+     * Get the console command arguments.
+     *
+     * @return array
+     */
+    protected function getArguments()
+    {
+        return [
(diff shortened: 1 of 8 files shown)


driesvints:
Remember that draft PR's aren't reviewed.

driesvints:
@oprypkhantc will need to see what @taylorotwell says

oprypkhantc:
Sure, thanks. This is the second implementation (non-blocking but slow) I used is the following. It's not complete and ugly and is there just to give you an idea of what it would roughly look like:

```php
$cursor = $defaultCursorValue = '0';
$del = [];

do {
	[$cursor, $valuesChunk] = $redis->scan(
		$cursor, ['match' => 'laravel:tag:*', 'count' => 1000]
	);

	// PhpRedis client returns false if set does not exist or empty.
	if (is_null($valuesChunk)) {
		break;
	}

	$valuesChunk = array_unique($valuesChunk);

	if (empty($valuesChunk)) {
		continue;
	}

	$tagValues = $redis->mget($valuesChunk);

	foreach ($tagValues as $tagValue) {
		$tagValue = is_numeric($tagValue) ? $tagValue : unserialize($tagValue);

		$members = $redis->smembers("laravel:$tagValue:standard_ref");

		$result = $redis->pipeline(function ($pipe) use ($members) {
			foreach ($members as $member) {
				$pipe->exists($member);
			}
		});

		foreach ($members as $i => $member) {
			if (!$result[$i]) {
				$del[] = $member;
			}
		}
	}
} while (((string) $cursor) !== $defaultCursorValue);

// todo: del from $del
// todo: cleanup tag references themselves
```

taylorotwell:

driesvints:
@oprypkhantc looks like Taylor doesn't want to take any action here for now. We'd welcome a PR to the docs to warn about this if you're willing.

oprypkhantc:
It's strange Laravel doesn't want to be production-ready on this matter. I'll leave it at that and follow the advice to make it a package.

Thanks @driesvints.

telkins:
> It's strange Laravel doesn't want to be production-ready on this matter. I'll leave it at that and follow the advice to make it a package.
> 
> Thanks @driesvints.

Did you ever create a package...?

We've noticed in some early testing of working with tags that in addition to potential performance issues, it also seems like it orphans set entries.  That's what led me to this PR in the first place.  😅

telkins:
Yep...I looked at your PR and can now see that you tried to fix the problem we noticed.  I'm also surprised that the Laravel team doesn't see this as a problem worth solving.  It's not exactly a bug, but it's leaving data that should otherwise be deleted.  This can add up quite quickly, especially for systems that use a large number of tags.

oprypkhantc:
@telkins Yes, the package is publicly available. We've had it deployed for three weeks now, see my reply in the issue: https://github.com/laravel/framework/issues/43850#issuecomment-1241904258

We were able to free nearly 1GB of stale tags alone, which was more than the tagged data itself.

oprypkhantc:
I've decided not to continue with PR's implementation because it was nearly 100x slower than ZSET tags implementation used in the package. This reduced the time of flush-stale command from a couple of minutes down to seconds.

taylorotwell:
@oprypkhantc where is the package?

oprypkhantc:
@taylorotwell See my reply: https://github.com/laravel/framework/issues/43850#issuecomment-1241904258

Packagist: https://packagist.org/packages/tenantcloud/laravel-better-cache

telkins:
@oprypkhantc Thx for your replies/links.  I'll check it out.  We will definitely benefit from what your package fixes.  👍

PR #43880, opened 2022-08-26
[9.x] Conditionally escape the PHP binary path

### LINKED ISSUE #43878: Artisan about - no composer version displayed when using composer.phar
- Laravel Version: 9.26.1
- PHP Version: 8.1.7
- Database Driver & Version: MySql 8.0.16
- Windows/10 latest version

### Description:

The `artisan about` does not show the correct composer version when using `composer.phar`.

image

### Steps To Reproduce:

- Have a composer.phar in the root of the project.
- Don't have composer installed otherwise.
- Execute `php artisan about`
- Observe there is no composer version displayed

jbrooksuk:
Fixes https://github.com/laravel/framework/issues/43878

It appears that escaping the full PHP path (for example; `'/opt/homebrew/Cellar/php/8.1.9/bin/php' composer.phar -V --no-ansi`) when running the Composer PHAR seems to break.

### DIFF
--- a/src/Illuminate/Support/Composer.php
+++ b/src/Illuminate/Support/Composer.php
@@ -68,7 +68,7 @@ class Composer
     public function findComposer()
     {
         if ($this->files->exists($this->workingPath.'/composer.phar')) {
-            return [$this->phpBinary(), 'composer.phar'];
+            return [$this->phpBinary(false), 'composer.phar'];
         }
 
         return ['composer'];
@@ -77,11 +77,18 @@ class Composer
     /**
      * Get the PHP binary.
      *
+     * @param  bool  $escape
      * @return string
      */
-    protected function phpBinary()
+    protected function phpBinary($escape = true)
     {
-        return ProcessUtils::escapeArgument((new PhpExecutableFinder)->find(false));
+        $binary = (new PhpExecutableFinder)->find(false);
+
+        if (! $escape) {
+            return $binary;
+        }
+
+        return ProcessUtils::escapeArgument($binary);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


mikeydevelops:
What if there are spaces in the path where the PHP binary is stored, as is likely on windows? Say PHP is in `C:/Program Files/php/8.1.3/php.exe'? Since you made it not escaped by default, the space means different arguments in the cli, no?

jbrooksuk:
@mikeydevelops that's a great point, and probably why it was escaped by default. I _think_, in that case, that we need to use `escapeshellcmd` instead?

MircoBabin:
As commented in https://github.com/laravel/framework/issues/43878#issuecomment-1229800454 this PR works on Windows, even when php.exe is in a path with spaces.

I can't test this on Linux/Mac, I don't have access to such a device.

@mikeydevelops In https://github.com/laravel/framework/blob/86f0c5f55f675d61c5a00f473ede998925f70b08/src/Illuminate/Support/Composer.php#L125-L127  the Process component ( https://symfony.com/doc/current/components/process.html#using-features-from-the-os-shell ) is used. The Process component states:

```
Using an array of arguments is the recommended way to define commands. This saves you from any escaping and allows sending signals seamlessly (e.g. to stop processes while they run):
```

So in theory, the `phpBinary()` should never have been escaped at all.

jbrooksuk:
Hmm, this won't work on macOS since spaces are escaped with `\ `.

taylorotwell:
Eh - I dunno if I really want to mess with these methods just to fix that one edge case.

PR #43891, opened 2022-08-27
Change constructor syntax

idanieldrew:
Change cunstructor property promotion syntax in Repository.php file in Cache service(folder).


rodrigopedra:
Why did you change the property visibility from `protected` to `private`? Most tests are failing due to this changes as subclasses cannot access this property anymore.

Also, I don't think promoted properties are still used throughout the framework yet.

PR #43892, opened 2022-08-27
[9.x] Add `makeOr` method to `Container`

michael-rubel:
## About
This method will help us ensure our code will be executed if container resolution fails. This way we can bind stuff to the container "on-the-fly" even if it wasn't bound previously, or do something else.

For example:
```php
app()->makeOr('company', function ($app) {
    $company = Company::firstWhere('uuid', request()->segment(1));

    $app->scoped('company', fn () => $company);

    return $company;
});
```

If we're resolving the service with parameters, we can do this:
```php
app()->makeOr(Service::class, ['dependency' => $dependency], function ($app) {
    //
});
```

There are a lot more use cases though. Optionally, it would have been binding results of executed queries to the container to avoid query duplication in the scope of the request lifecycle, if someone doesn't want to deal with the cache/invalidation issues, etc.

---

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -702,7 +702,7 @@ class Container implements ArrayAccess, ContainerContract
      *
      * @throws \Illuminate\Contracts\Container\BindingResolutionException
      */
-    public function makeOr($abstract, $parameters = [], $callback = null)
+    public function makeOr($abstract, $parameters = [], Closure $callback = null)
     {
         try {
             if ($parameters instanceof Closure) {
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -696,23 +696,22 @@ class Container implements ArrayAccess, ContainerContract
      * Resolve the given type from the container or execute the callback.
      *
      * @param  string  $abstract
+     * @param  mixed|array  $parameters
      * @param  Closure|array  $callback
      * @return mixed
      *
      * @throws \Illuminate\Contracts\Container\BindingResolutionException
      */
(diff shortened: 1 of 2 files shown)


garygreen:
This can easily be expressed in out the box PHP syntax:

```php
$company = app('company') ?? Company::makeOne([
	//
]);
```

Also I'd recommend not resolving directly out the IOC because it makes it difficult for IDEs to introspect. Better to do:

```php
<?php

class Company {
   public static function instance(): self
   {
      return app('company');
   }
}

Company::instance() ?? Company::makeOne([
]);
```

Or find some other way of abstracting it.

Just my two cents.

foremtehan:
> This can easily be expressed in out the box PHP syntax:
You cant. `Target class [company] does not exist.`

src/Illuminate/Container/Container.php
```diff
@@ -678,6 +678,31 @@ public function makeWith($abstract, array $parameters = [])
...
+     * @param  \Closure|null  $callback
+     * @return mixed
+     *
+     * @throws \Illuminate\Contracts\Container\BindingResolutionException
+     */
+    public function makeOr($abstract, $parameters = [], Closure $callback = null)
+    {
+        try {
+            if ($parameters instanceof Closure) {
+                $callback = $parameters;
```
bert-w, line 695: what does this do?

michael-rubel: @bert-w It allows you to pass the callback as a second argument preserving the method signature and avoiding parsing the args list.

bert-w: Oh I see now; this can fail and the catch-clause could be called

taylorotwell:

PR #43893, opened 2022-08-27
[9.x] Add Eloquent mode to prevent silently discarding fills for attributes not in `$fillable`

ralphjsmit:
One very irritating thing for me is to forget adding a new property to the `$fillable` array (because it's not my default way of working) and later discover that fills have been silently discarded, thus causing bugs.

Personally I always unguard my models using `$guarded = []`, because I know from myself that I never do anything like `$model->update($request->all())`. However, sometimes you join a new project where it's the convention to always use the `$fillable` array.

Currently Eloquent only throws an exception when the model is "totally guarded". This PR adds a method that is similar to the `Model::preventLazyLoading()` to also throw exceptions when a model change was discarded:

```php
Model::preventDiscardingGuardedAttributeFills();
```

It also works with a boolean:
```php
Model::preventDiscardingGuardedAttributeFills(! app()->isProduction());
```

_Naming ideas are appreciated. I tried to come up with a shorter name, but I couldn't yet think of a name that could convey the behaviour in a better and shorter way._

People can also add a callback to handle the exception. The callback receives the model, the key and the value of the attribute:

```php
Model::handleMassAssignmentViolationUsing(function(Model $model, string $key, mixed $value) {
    //
});
```

I think this would be very useful to prevent bugs on local environments when the developer forgets to add an attribute to `$fillable`. Thanks!

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -175,18 +175,11 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     protected static $lazyLoadingViolationCallback;
 
     /**
-     * Indicates whether discarding attribute fills is allowed.
+     * Indicates if an exception should be thrown instead of silently discarding non-fillable attributes.
      *
      * @var bool
      */
-    protected static $modelsShouldPreventDiscardingGuardedAttributeFills = false;
-
-    /**
-     * The callback that is responsible for handling mass assignment violations.
-     *
-     * @var callable|null
-     */
-    protected static $massAssignmentViolationCallback;
+    protected static $modelsShouldPreventSilentlyDiscardingAttributes = false;
 
     /**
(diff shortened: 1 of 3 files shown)


foremtehan:
add tests to your codes, thank you.

ralphjsmit:
> add tests to your codes, thank you.

I always add tests. Please double check before you say that.

Schermafbeelding 2022-09-01 om 16 48 25

Schermafbeelding 2022-09-01 om 16 47 39

edalzell:
Love this one, thanks @ralphjsmit!

tillkruss:
This should be the the Laravel default.

ralphjsmit:
@tillkruss Agree! You could PR it though, perhaps Taylor is open to making it the default in 9.x or 10.x. 👊

taylorotwell:
@ralphjsmit this PR actually doesn't even work for all cases... even a simple case like:

```
$user = new User(['unfillable' => 'foo']);
```

PR #43896, opened 2022-08-28
[9.x] Introduces `expectsAllOutputToContain` and `doesntExpectAllOutputToContain` for PendingCommand

sethsandaru:
## What

`expectsOutputToContain` and `doesntExpectOutputToContain` have their limitations. Real case scenario below:

I have a command that write a multiple lines in 1 go `$this->info($strContainsMultileLines)`. 

```php
@property int $id
@property string $first_name
```

Test result: 
```php
$command->artisan(...)
    ->execute();
```

```php
foreach ($lines as $line) {
    $this->info($line);
```

## Solution

Introducing `expectsAllOutputToContain` and `doesntExpectAllOutputToContain` for `PendingCommand`, which gonna help us to assert **all outputs** of the command, which gonna help for:

- Assert the string on all the outputs
- Doesn't really care about the order, I just wanna to validate the inner texts after the command ran, whether it's from the start, middle or end.

## Breaking change?

No

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/InteractsWithConsole.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/InteractsWithConsole.php
@@ -16,6 +16,13 @@ trait InteractsWithConsole
     public $mockConsoleOutput = true;
 
     /**
+     * All of the output that printed out by the command
+     *
+     * @var array
+     */
+    public $allOutputs = [];
+
+    /**
      * All of the expected output lines.
      *
      * @var array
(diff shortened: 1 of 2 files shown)


taylorotwell:
Honestly feels weird to have ANOTHER method for this. We should just fix the ones we have if they don't work correctly?

sethsandaru:
> Honestly feels weird to have ANOTHER method for this. We should just fix the ones we have if they don't work correctly?

Thanks @taylorotwell , let me submit the fix for the old methods then 👍

PR #43898, opened 2022-08-28
[9.x] Add support for `USING` clause in query join condition

Persaeus:
Currently, it is not trivial to make a join clause use a `USING` clause instead of an `ON` clause in its join condition.

While `ON` is generally more versatile, there are cases where `USING` is more useful. For example, when joining tables on columns that share the same name across the two tables. Additionally, it removes the need to fully qualify the joining columns. More information from this answer on stackoverflow: https://stackoverflow.com/a/11367066.

This PR seeks to add support for the `USING` clause in join conditions.

It may be set on the `JoinClause` builder object via a new `using` method. Multiple columns may be set as multiple arguments `$join->using('first, 'second')`, or by passing an array as the first argument `$join->using(['first', 'second'])`.

Since `USING` and `ON` cannot be used simultaneously, we will compile a `USING` clause if `$joinClause->using` is set, or otherwise continue with normal `ON` compilation.

Not needing to fully qualify columns allows us to join rows to an existing query without having to change the columns in any existing where conditions. This is not possible using only an `ON` clause, as we would get query exceptions like `ERROR 1052 (23000): Column 'xyz' in where clause is ambiguous`:

```php
use Illuminate\Database\Query\JoinClause;

$query = DB::table('users')->whereEmail('foo@bar.com');

$query
    ->rightJoin(
        'clients',
        fn (JoinClause $join) => $join->using('email')
    )
    ->toSql();
````

```SQL
SELECT * FROM `users`
RIGHT JOIN `clients` USING (`email`)
WHERE `email` = ?
```

The `USING` clause is valid syntax in MySql, PostgreSql and SQLite. It is not supported in SQL Server (thanks for clarifying @rodrigopedra). For SQL Server, we could convert the `USING` clause to a comparable `ON` clause (https://github.com/laravel/framework/pull/43898#issuecomment-1229539022), or throw an exception. I have left out this part of the PR, pending comment from the maintainers.

Also, this change does not break any existing features, as `USING` will only be compiled if it's been explicitly set via the `using` method on the `JoinClause` object.

### DIFF
--- a/src/Illuminate/Database/Query/Grammars/Grammar.php
+++ b/src/Illuminate/Database/Query/Grammars/Grammar.php
@@ -194,7 +194,7 @@ class Grammar extends BaseGrammar
     /**
      * Compile a join clause's USING (...) subclause.
      *
-     * @param  \Illuminate\Database\Query\JoinClause $query
+     * @param  \Illuminate\Database\Query\JoinClause  $query
      * @return string
      */
     protected function compileJoinUsingClause(JoinClause $query)
--- a/src/Illuminate/Database/Query/Grammars/Grammar.php
+++ b/src/Illuminate/Database/Query/Grammars/Grammar.php
@@ -203,7 +203,7 @@ class Grammar extends BaseGrammar
             ->map(fn ($column) => $this->wrap($column))
             ->implode(',');
 
-        return "using ($columns)";
+        return 'using ('.$columns.')';
     }
 
     /**
--- a/src/Illuminate/Database/Query/Grammars/Grammar.php
+++ b/src/Illuminate/Database/Query/Grammars/Grammar.php
@@ -180,8 +180,30 @@ class Grammar extends BaseGrammar
 
             $tableAndNestedJoins = is_null($join->joins) ? $table : '('.$table.$nestedJoins.')';
 
-            return trim("{$join->type} join {$tableAndNestedJoins} {$this->compileWheres($join)}");
+            // Determine if the join condition should be expressed
+            // with an ON subclause or a USING subclause, and compile
+            // the corresponding join condition
+            $joinCondition = $join instanceof JoinClause && ! is_null($join->using)
+                ? $this->compileJoinUsingClause($join)
+                : $this->compileWheres($join);
+
+            return trim("{$join->type} join {$tableAndNestedJoins} {$joinCondition}");
         })->implode(' ');
+    }
+
+    /**
+     * Compile a join clause's USING (...) subclause.
+     *
+     * @param  \Illuminate\Database\Query\JoinClause $query
+     * @return string
+     */
+    protected function compileJoinUsingClause(JoinClause $query)
+    {
+        $columns = collect($query->using)
+            ->map(fn ($column) => $this->wrap($column))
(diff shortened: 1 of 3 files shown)


src/Illuminate/Database/Query/Grammars/Grammar.php
```diff
@@ -180,10 +180,32 @@ protected function compileJoins(Builder $query, $joins)
...
+     * @param  \Illuminate\Database\Query\JoinClause  $query
+     * @return string
+     */
+    protected function compileJoinUsingClause(JoinClause $query)
+    {
+        $columns = collect($query->using)
+            ->map(fn ($column) => $this->wrap($column))
+            ->implode(',');
+
+        return "using ($columns)";
```
michael-rubel, line 206: Maybe it's better to concatenate instead of using interpolation?

rodrigopedra:
> As far as I'm aware USING is basic SQL, so this PR should work for all grammars.

`USING` is not available in SQLite and SQL Server. So I guess this should be added only to MySQL and Postgres grammars. Or throw an exception on SQLite and SQL Server ones.

references: 

- https://docs.microsoft.com/en-us/sql/t-sql/queries/from-transact-sql?view=sql-server-ver16
- https://www.sqlite.org/syntax/join-clause.html

nihilsen:
> `USING` is not available in SQLite and SQL Server.

You're right, it appears `USING` isn't available in SQL Server. Thank you for clarifying that.

According to the reference you linked, `USING` *is* available in SQLite: https://www.sqlite.org/syntax/join-constraint.html.

nihilsen:
> I guess this should be added only to MySQL and Postgres grammars. Or throw an exception on ~~SQLite~~ and SQL Server ones.

Good point.

I'm not sure what would be the best approach in this situation, when the syntax is supported by all except one of the SQL grammars.

We could also convert the `USING` clause to an approximately equivalent `ON` clause for SQL Server, like so:

```SQL
select * from a inner join b using (c)    # MySql, Postgres, SQLite
select * from a inner join b on a.c = b.c # SQL Server
```

src/Illuminate/Database/Query/Grammars/Grammar.php
nihilsen: You're right. Fixed.

rodrigopedra:
> According to the reference you linked, USING is available in SQLite

You are right! I overseen this as it was a bit late down here =)

Thanks for the heads up

taylorotwell:

nihilsen:
Thank you @taylorotwell, I understand the decision not to merge this code for now. There were more nuances to the implementation than I had initially anticipated, namely that `USING` isn't supported by SQL Server.

I hope that support for the `USING` directive, which has been part of the SQL standard since [SQL-92](http://www.java2s.com/Tutorial/Oracle/0140__Table-Joins/SimplifyingJoinswiththeUSINGKeyword.htm), is something that could be reconsidered for future versions of Laravel.

I've gone ahead and converted this PR into a package that adds support for `USING` in join constraints, for if anyone (like me) needs this functionality right now: https://github.com/nihilsen/laravel-join-using

PR #43911, opened 2022-08-29
Update Builder.php

HiHub-Online:
Fix setted new select columns.

Some stupid mistake.
A query with new columns was executed only if there were no selectors before.


driesvints:
Breaks the test suite.

Issue #43889, opened 2022-08-27
[9.x] Return data type of the `GeneratorCommand::handle()` method


imdhemy:
Using `bool|null` as a return type from `GeneratorCommand::handle()` method is
a bit confusing. The returned value is used by Symfony command `run()` method 
that expects `1` for FAILURE and `0` for SUCCESS.

Here, In an _error case_, `false` is returned (_a failure status  `1`  is intended_), but a success 
status (`0`) is returned by Symfony command `run` method.

https://github.com/laravel/framework/blob/2a1a55caf3c65a74f173f47be594587cb931f22c/src/Illuminate/Console/GeneratorCommand.php#L149-L153

The same applies to the following code:

https://github.com/laravel/framework/blob/2a1a55caf3c65a74f173f47be594587cb931f22c/src/Illuminate/Console/GeneratorCommand.php#L162-L168

Note: This may exist in other commands, but I started with this one.

---

If my understanding is correct, I would be happy to submit a PR to fix this.

driesvints:
I don't see anything wrong here.

imdhemy:
@driesvints 
Could you please have a look at this test?

https://github.com/imdhemy/laravel.framework/blob/generator-command/tests/Console/GeneratorCommandTest.php#L39

This following assertion fails:
```php
$this->assertSame(Command::FAILURE, $sut->run($input, $output));

// Failed asserting that 0 is identical to 1.
```

driesvints:
If you want to change anything feel free to send in a PR 👍

PR #43915, opened 2022-08-29
[9.x] Return exit status instead of booleans from `GeneratorCommand::handle()`

imdhemy:
This PR updates the `GenertorCommand::handle()` method to return exit code instead of boolean and adds 3 new tests. The unit tests help to understand why this update is needed. 

#43889 

Besides, it replaces a couple of calls to the application instance as an ArrayAccess with a `make()` method calls. This practice makes it easier to be tested. It does not change the behavior as the `\Illuminate\Container\Container::offsetGet()` uses the same logic.

https://github.com/laravel/framework/blob/45cd804928f4c1a42fafd2b78c7fa023a59c542a/src/Illuminate/Container/Container.php#L1413-L1416

### DIFF
--- a/src/Illuminate/Console/GeneratorCommand.php
+++ b/src/Illuminate/Console/GeneratorCommand.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Console;
 
 use Illuminate\Console\Concerns\CreatesMatchingTest;
+use Illuminate\Contracts\Container\BindingResolutionException;
 use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Str;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Reverted - breaks other commands that extend GeneratorCommand and check === false from parent::handle. See ComponentMake.

imdhemy:
Thanks @taylorotwell 

I'll work on it and resend the PR.

rodrigopedra:
@imdhemy I guess you are going to send it to 10.x, as there are third-party packages which use this behavior, for example https://github.com/spatie/laravel-view-models

https://github.com/spatie/laravel-view-models/blob/9c1877f4f99e2c373ac3f432fe6313f09323c674/src/Console/ViewModelMakeCommand.php#L16-L23

This way a note should be added to the upgrade guide so these packages can update accordingly without any breaking changes.

imdhemy:
@rodrigopedra 

Exactly. 👍

PR #43916, opened 2022-08-29
[9.x] Fixed expectsOutputToContain & doesntExpectOutputToContain of PendingAction

sethsandaru:
All the details go here #43896

Basically `expectsOutputToContain` and `doesntExpectOutputToContain` didn't assert against **all output** . 

So if you write an output that contains a lot of info, you can't assert it multiple times for the same output.

### Partial content from #43896

```php
@property int $id
@property string $first_name
```

Test result: 
```php
$command->artisan(...)
    ->execute();
```

```php
foreach ($lines as $line) {
    $this->info($line);
```

### DIFF
--- a/tests/Integration/Testing/ArtisanCommandTest.php
+++ b/tests/Integration/Testing/ArtisanCommandTest.php
@@ -34,7 +34,9 @@ class ArtisanCommandTest extends TestCase
         });
 
         Artisan::command('contains', function () {
+            $this->line('Hello everyone!');
             $this->line('My name is Taylor Otwell');
+            $this->line('I bet you can assert this line');
         });
     }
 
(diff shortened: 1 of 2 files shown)


tests/Integration/Testing/ArtisanCommandTest.php
```diff
@@ -128,6 +130,8 @@ public function test_console_command_that_passes_if_the_output_contains()
     {
         $this->artisan('contains')
              ->expectsOutputToContain('Taylor Otwell')
+             ->expectsOutputToContain('I bet you can assert this line')
+             ->expectsOutputToContain('I bet you can assert this line')
```
sethsandaru, line 134: To ensure I can assert the output twice (and more)

sethsandaru:
Hi @taylorotwell, could you please review it again? Thanks

taylorotwell:
Instead of risking breaking anything on a patch release I suggest just expecting multiple lines?

CleanShot 2022-09-08 at 09 24 02@2x

PR #43918, opened 2022-08-29
Disallow PHP 8.1 support for 6.x

stepananchugov:
This fixes PHP constraint to `"8.0"` for 6.1 since PHP 8.1 gives this:

> During inheritance of ArrayAccess: Uncaught Return type of Illuminate\Container\Container::offsetExists($key) should either be compatible with ArrayAccess::offsetExists(mixed $offset): bool, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice

Maybe it's worth to also port these changes to 7.x and 8.x

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "^7.2.5|^8.0",
+        "php": "^7.2.5|8.0",
         "ext-json": "*",
         "ext-mbstring": "*",
         "ext-openssl": "*",


GrahamCampbell:
Moreover, this only allows 8.0.0, and not 8.0.1!

stepananchugov:
Hm, sorry, this should be probably converted to a draft. Anyways, to give you more context, I have a requirement like this:
```json
{
    "php": "^7.4|^8.0|^8.1",
    "laravel/framework": "^6.0|^7.0|^8.53",
    ...
}
```
And I run `composer update --prefer-lowest` since this is a requirement for a lib. Could you please suggest any other solutions?

GrahamCampbell:
`^8.0|^8.1` is the same thing as `^8.0`.

GrahamCampbell:
There is nothing you can do to stop Laravel 6 from being installed on PHP 8.1.

stepananchugov:
Thanks @GrahamCampbell! Closing this then, thanks for helping out!

PR #43920, opened 2022-08-29
Add identifier to schedule event

fragkp:
**Problem**
Currently, it's not possible to track schedule events. This makes it hard for package creators to build something like Laravel Horizon for schedule tasks. Most packages out there, add a dynamic property (UUID, microtime, etc) to handle that case.

**Solution**
This PR adds an UUID to the schedule event.

**Example**
```php
// -> not needed anymore
// $this->app->resolving(Schedule::class, function (Schedule $schedule) {
//     foreach ($schedule->events() as $event) {
//         $event->taskIdentifier = Str::uuid();
//     }
// });

Event::listen(ScheduledTaskStarting::class, function ($event) {
    $event->task->uuid; // trackable
});
```


taylorotwell:

PR #43923, opened 2022-08-29
[9.x] Add view data assertions to `TestView`

browner12:
- `assertViewHas()`
- `assertViewHasAll()`
- `assertViewMissing()`

these methods are all available on the `TestResponse`, but not currently `TestView`.  These assertions can be more helpful to test the workings of the component, and not reliant on how the data actually gets displayed on the view side of things.

### DIFF
--- a/src/Illuminate/Testing/TestView.php
+++ b/src/Illuminate/Testing/TestView.php
@@ -2,6 +2,10 @@
 
 namespace Illuminate\Testing;
 
+use Closure;
+use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Traits\Macroable;
 use Illuminate\Testing\Assert as PHPUnit;
 use Illuminate\Testing\Constraints\SeeInOrder;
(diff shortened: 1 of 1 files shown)


taylorotwell:
If this is a test view object, is it a bit redundant to say `assertViewHas` instead of just `assertHas`?

jasonmccreary:
Tiny, tiny papercut, but if this is new code, I'd love to be able to check a piece of view data is explicitly `null` (as you can't in the current assertions since `null` is the default value and ignored).

browner12:
this just got deployed today, so unfortunately can't change it now.

I'm trying to think how this would work, and AFAIK in PHP you can't tell if an argument is passed, or if it's using the default value.  I feel like we'd have to separate out a separate method for checking simply if the variable is passed, and one for if the variable has a specific value.

jasonmccreary:
Yeah, no worries. Just something in the rare times I've actually passed in `null` as view data I thought could be improved.

I've seen other languages use a constant in this case. For example, Objective-C uses a `NOT_FOUND` constant when searching strings. Mimicing that, there could something like a `TestView::IGNORE_VALUE` which is the default - allowing `null` to be passed in separately.

browner12:
very interesting.... I feel like this could be super beneficial in numerous places on the framework.

how do you check for the constant? aren't you still comparing by value? what value do you assign the constant?

jasonmccreary:
Yes, its actual value could be a bit tricky as a simple value could pose the same problem. I suppose that's the benefit of a language constant.

Nonetheless, something random would have a lower hit rate than `null`. For example, `LARAVEL_CONSTANT_256`. 😂

PR #43925, opened 2022-08-30
[9.x] Allow invokable rules to specify custom messsages

nshiro:
From Laravel 9.2 we have a simple way to specify a custom message when validating with a rule object. #41145
But it doesn't seem to work with invokable rules.
This PR makes invokable rules to work with #41145.


timacdonald (approved):
Seems like a clean fix. Tested locally and confirmed.

Thanks @nshiro for the PR ❤️

nshiro:
Thank you for reviewing my code.😀 @timacdonald

PR #43929, opened 2022-08-30
[10.x] Return exit status instead of booleans from `GeneratorCommand::handle()`

imdhemy:
This PR reopens #43915 and updates the commands that use the `handle()` method.

::: Note
A note should be added to the upgrade guide.

### DIFF
--- a/src/Illuminate/Console/GeneratorCommand.php
+++ b/src/Illuminate/Console/GeneratorCommand.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Console;
 
 use Illuminate\Console\Concerns\CreatesMatchingTest;
+use Illuminate\Contracts\Container\BindingResolutionException;
 use Illuminate\Filesystem\Filesystem;
 use Illuminate\Support\Str;
 use Symfony\Component\Console\Input\InputArgument;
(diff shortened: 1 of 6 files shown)


src/Illuminate/Console/GeneratorCommand.php
```diff
@@ -258,12 +261,14 @@ protected function alreadyExists($rawName)
      *
      * @param  string  $name
      * @return string
+     *
+     * @throws BindingResolutionException
```
driesvints, line 265: Can you please remove all these `@throws` annotations you added everywhere? These methods don't throw this exception, they only cascade.

driesvints (changes requested):

src/Illuminate/Console/GeneratorCommand.php
imdhemy: Removed...

driesvints: @imdhemy there's still some in this PR (including this one).

tests/Console/GeneratorCommandTest.php
```diff
@@ -0,0 +1,113 @@
...
+use Symfony\Component\Console\Exception\RuntimeException;
+use Symfony\Component\Console\Input\ArrayInput;
+use Symfony\Component\Console\Output\NullOutput;
+
+class GeneratorCommandTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function itShouldThrowIfNameIsNotPassedAsArgument(): void
```
driesvints, line 23: Please use the same `test_` syntax as the rest of the test suite.

driesvints (changes requested):

tests/Console/GeneratorCommandTest.php
imdhemy: @driesvints updated...

taylorotwell:
This doesn't do anything to actually improve Laravel for end users, which is what I want to focus on.

PR #43930, opened 2022-08-30
[9.x] Introduce `artisan warm`

timacdonald:
> **Note** This description is a WIP.

This PR introduces a new `artisan warm` command that allows applications to specify how their application should be "warmed" before deployment. It also allows packages to hook into the process.

This command would enable the Framework, 3rd party packages, and the application to all warm via a single command. This command would be called before or just after bringing the application out of maintenance mode (depending on the requirements):

```sh
php artisan down

# ...

php artisan migrate

php artisan warm

php artisan up
```

Warming an application is all about ensuring that users do not suffer a slow experience when any cache is stale. A "cache" in this context is any kind of cache, but it Redis or a file that is expected on disk, etc.

An application initially might cache things via the `Cache::remember()` function when the resource is first required for a user request - but this means that the user will be impacted by this stale cache.

Solutions to this might include keeping the cache warm on a schedule:

```php
class Kernel extends ConsoleKernel
{
    protected function schedule(Schedule $schedule)
    {
        $schedule->call(function () {
            Cache::put('forge.ips', Http::get('https://forge.laravel.com/ips-v4.txt'), now()->addHours(2));
        })->hourly();
    }
}
```

The user still pays the price for caching:

1. On the initial deploy.
2. When the cache is cleared.
3. When things go wrong and the downtime creates a stale entry.

This command could be used for application specific warming - or package specific warming. Given a package that gives you currency conversion rates, the package could register a warmer in their service provider that downloads and caches the initial rates.

```php
/**
 *
 * @return void
 */
public function boot()
{
    $this->registerWarmer('currency/converter', fn () => UpdateRates::dispatchNow());
}
```

Screen Shot 2022-08-30 at 5 18 55 pm

An application may specify how it is warmed in the Kernel alongside the `schedule` method.

```php
class Kernel extends ConsoleKernel
{
    protected function schedule(Schedule $schedule)
    {
        $schedule->call(CacheForgeIPs::class)->hourly();
    }

    protected function warm()
    {
        CacheForgeIPs::dispatchNow();
    }
}
```

### DIFF
--- a/src/Illuminate/Foundation/Console/WarmCommand.php
+++ b/src/Illuminate/Foundation/Console/WarmCommand.php
@@ -1,0 +1,95 @@
+<?php
+
+namespace Illuminate\Foundation\Console;
+
+use Illuminate\Console\Command;
+use Illuminate\Contracts\Console\Kernel;
+use Illuminate\Support\WarmerCollection;
+use Symfony\Component\Console\Attribute\AsCommand;
+
+#[AsCommand(name: 'warm')]
+class WarmCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'warm {--exclude= : The registered warmers to exclude} {--only= : Only run the specified registered warmers}';
+
+    /**
+     * The name of the console command.
+     *
+     * This name is used to identify the command during lazy loading.
+     *
+     * @var string|null
+     *
+     * @deprecated
+     */
+    protected static $defaultName = 'warm';
+
+    /**
+     * The console command description.
+     *
+     * @var string
+     */
+    protected $description = 'Warm the application.';
+
+    /**
+     * Execute the console command.
+     *
+     * @param  \Illuminate\Support\WarmerCollection  $warmers
+     * @param  \Illuminate\Contracts\Console\Kernel  $kernel
+     * @return int
+     */
+    public function handle(WarmerCollection $warmers, Kernel $kernel)
+    {
+        $warmers
+            ->forget($this->excluded())
+            ->only($this->only())
+            ->when(method_exists($kernel, 'warmResolver'))->merge([
+                'The application: '.config('app.name') => $kernel->warmResolver()
+            ])
+            ->each(fn ($callable, $name) => $this->warm($name, $callable));
+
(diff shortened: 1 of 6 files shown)


src/Illuminate/Foundation/Console/WarmCommand.php
```diff
@@ -0,0 +1,95 @@
...
+
+#[AsCommand(name: 'warm')]
+class WarmCommand extends Command
+{
+    /**
+     * The name and signature of the console command.
+     *
+     * @var string
+     */
+    protected $signature = 'warm {--exclude= : The registered warmers to exclude} {--only= : Only run the specified registered warmers}';
```
driesvints, line 18: Maybe move the flags on different lines here?

src/Illuminate/Foundation/Console/WarmCommand.php
```diff
@@ -0,0 +1,95 @@
...
+     */
+    protected function excluded()
+    {
+        return explode(',', $this->option('exclude') ?? '');
+    }
+
+    /**
+     * The registered warmers that should be run.
+     *
+     * @return null|array
```
driesvints, line 85:
```suggestion
     * @return array|null
```

src/Illuminate/Foundation/Console/WarmCommand.php
```diff
@@ -0,0 +1,95 @@
...
+            ])
+            ->each(fn ($callable, $name) => $this->warm($name, $callable));
+
+        $this->components->info('Your application has finished warming.');
+
+        return self::SUCCESS;
+    }
+
+    /**
+     * Run the wamer.
```
siarheipashkevich, line 61: Please fix typo.

jbrooksuk, line 61:
```suggestion
     * Run the warmer.
```

DarkGhostHunter: Unusable

timacdonald:
Gonna close this one off until I have time to come back to it and finish it up. Will re-open it once it is ready to go.

timacdonald:
@julienbourdeau just FYI I don't think we are gonna pursue this one. We didn't feel it was compelling enough to bring it into core right now.

PR #43933, opened 2022-08-30
[9.x] Introducing Signal Traps 🚦

nunomaduro:
Documentation: https://github.com/laravel/docs/pull/8178.

Many operating systems allow signals to be sent to running processes. As an example, when typing `Control + C` on MacOS, a `SIGINT` will initiate an Artisan command shutdown.

Artisan commands shutdowns may lead to unexpected state, as the user may have defined an Artisan command that takes more than expected to run, and ending the command without any sort of clean up may leave temporary files, temporary resources, and more.

To mitigate this issue, this pull request introduces [Signal Traps](https://www.linuxjournal.com/content/bash-trap-command) on Laravel, a new concept that allows you to catch process signals and execute code when they occur:

```php
Artisan::command('export:sales', function () {
    $this->continue = true;

    $this->trap(SIGINT, function () {
        // End the command's job gracefully ...
        // E.g: clean created files, etc ...

        $this->continue = false;
    });

    while($this->continue) {
        dump('Exporting batch of sales...');
    }
});
```

With this technique, Laravel developers can perform any clean-up tasks / or potentially revert operations when an Artisan command is ended unexpectedly. The given API, sits on top of a new `trap` method, on the base `Command` class of Laravel.

Now, because is very common wanting to subscribe to multiple signals at the same time, you may pass an `array` of signals in those cases:
```php
$this->trap([SIGTERM, SIGINT], function () {
    ExportSales::markAsNotReady();


    throw new Exception('Command ended unexpectedly via signal');
}
```

In addition, `traps` get `untrap` once the command is "done". This is useful, and intentionally behavior, as commands may get called within commands, or within "queue:jobs":

```php
# Use Case 1: Command "1" is still running, command "2" is done:
Artisan::command('1', function () {
    $this->trap(SIGINT, function () {
        dump('trap #1');
    });

    Artisan::call('2');

    sleep(10); // Control + C here ...
});

Artisan::command('2', function () {
    $this->trap(SIGINT, function () {
        dump('trap #2');
    });
});

// Displays: 
// #1
```

```php
# Use Case 2: Command "2" and command "1" are still running;
Artisan::command('1', function () {
    $this->trap(SIGINT, function () {
        dump('trap #1');
    });

    Artisan::call('2');
});

Artisan::command('2', function () {
    $this->trap(SIGINT, function () {
        dump('trap #2');
    });

    sleep(10); // Control + C here ...
});

// Displays: 
// #2
// #1
```

```php
# Use Case 3: Command "1" is still running, and command "2" is done, on a dispatched job.

dispatch(function () {
    Artisan::call('1');
});

Artisan::command('1', function () {
    $this->trap(SIGINT, function () {
        dump('trap #1');
    });

    Artisan::call('2');

    sleep(10); // Control + C here ...
});

Artisan::command('2', function () {
    $this->trap(SIGINT, function () {
        dump('trap #2');
    });
});

// Displays: 
// #1
```

Finally, two important things that not be trivial at first glance:

1. **by defining a `trap` method, you intentionally override PHP's native behavior for that signal**. For example, if you define a `trap` for `SIGINT`, PHP won't end the process as it would do natively, therefore is up to the user to end the process gracefully or not.
2. **Traps won't be run on non-UNIX platforms / or platforms that don't support the `pcntl` extension** - in fact, trying to use `SIGINT` on an Windows environment, results on the following error:

Screenshot 2022-09-01 at 12 05 25

Package developers should use `extension_loaded('pcntl')` before even calling traps, as their users may use Windows.

---

For future development considerations:

1. As mentioned in point 1) above, signals that typically would shut down the process, won't anymore. While this may be expected, it may not also be expected for some users. Therefore, if people get confused by this, we can future develop an API as the following:
```php
Artisan::command('2', function () {
    $this->trap(SIGINT, function () {
        $this->quit(); //  this...
    })->shouldQuit(); // or this...
});
```


src/Illuminate/Console/Signals.php
```diff
@@ -0,0 +1,127 @@
...
+    protected function getHandlers()
+    {
+        return (fn () => $this->signalHandlers)
+            ->call($this->registry);
+    }
+
+    /**
+     * Sets the signal's default callback.
+     *
+     * @return void
```
driesvints, line 113: The return type needs to be updated here.

src/Illuminate/Console/Signals.php
```diff
@@ -0,0 +1,132 @@
...
+     */
+    protected $previousHandlers;
+
+    /**
+     * Create a new Signals instance.
+     *
+     * @param  \Symfony\Component\Console\SignalRegistry\SignalRegistry  $registry
+     * @return void
+     */
+    public function __construct($registry)
```
Wirone, line 29: @nunomaduro current Laravel version requires PHP8+, so why don't you use strict types here (and many other places)? Why not:

```suggestion
    public function __construct(SignalRegistry $registry)
```

nunomaduro: No plans on adding a type-hint here. 👍🏻

src/Illuminate/Console/Signals.php
```diff
@@ -0,0 +1,132 @@
...
+        return (fn () => $this->signalHandlers)
+            ->call($this->registry);
+    }
+
+    /**
+     * Gets the signal's default callback on the array format.
+     *
+     * @return [callable]
+     */
+    protected function initializeSignal($signal)
```
nunomaduro, line 120: @taylorotwell @driesvints @GrahamCampbell Undecided about this peace of code, how it should behave in different signals. For the context, by default, PHP ends the script when receiving a SIGTERM or similar. Yet, when you override the signal's handler, PHP actually does not end the script unless you call an `exit()`. 

Yet, as a user, my expectation when using something like `$this->trap(SIGTERM, fn () => @unlink($file));` is that PHP ends the script after my handler is done. Therefore, the code bellow helps here.

What are your thoughts?

src/Illuminate/Console/Signals.php
Wirone: But why? I'm just curious, because there are multiple places in this PR where language-level types (arguments, properties, returns) could be added which would make this code just better (safer). Is it Laravel's convention 🤔?

src/Illuminate/Console/Signals.php
joedixon: Like you, my expectation is that I want the script to end.

src/Illuminate/Console/Signals.php
foremtehan: Backward compatibility i guess ? i dont know...

Wirone: In brand new files 😉? That's the point, I saw Nuno's tweet, looked and the changes out of curiosity and saw this. It does not make sense to me, so I asked to understand 🤷

PR #43934, opened 2022-08-30
[10.x] Support calling whereExists with another builder

axlon:
Eloquent provides multiple ways to deal with subqueries. Most methods that deal with subqueries support passing one through a `Closure`, an eloquent builder or a (base) query builder. `whereExists()` has always seemed like an odd duck to me because it forces the developer to work with a closure. I couldn't find any reason why this works the way it does, so I decided to try and remove this limitation.

This would open Eloquent up to syntax like:

```php
SomeModel::where(...)->whereExists(OtherModel::where(...));
```

PS. This PR is breaking because it alters 4 method signatures

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1607,19 +1607,24 @@ class Builder implements BuilderContract
     /**
      * Add an exists clause to the query.
      *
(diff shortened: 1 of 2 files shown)


taylorotwell:
I think I'll just avoid the breaking change here.

PR #43937, opened 2022-08-30
[9.x] Add default to request enum

keithbrink:
This PR adds support to the `enum` method on the request for setting a default value, similar to the `input` method.

```php
$enum = $request->enum('value', Enum::class, Enum::Value);
```

I could add a check that the default enum value matches the enum class, but I thought I would keep it simple for now.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -350,16 +350,16 @@ trait InteractsWithInput
      * @param  string  $enumClass
      * @return mixed|null
      */
-    public function enum($key, $enumClass)
+    public function enum($key, $enumClass, $default = null)
     {
         if ($this->isNotFilled($key) ||
             ! function_exists('enum_exists') ||
             ! enum_exists($enumClass) ||
             ! method_exists($enumClass, 'tryFrom')) {
-            return null;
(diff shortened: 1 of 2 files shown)


src/Illuminate/Http/Concerns/InteractsWithInput.php
```diff
@@ -350,13 +350,13 @@ public function date($key, $format = null, $tz = null)
      * @param  string  $enumClass
      * @return mixed|null
```
fgaroby, line 351: Missing `@param`for this third parameter

src/Illuminate/Http/Concerns/InteractsWithInput.php
```diff
@@ -350,13 +350,13 @@ public function date($key, $format = null, $tz = null)
...
      */
-    public function enum($key, $enumClass)
+    public function enum($key, $enumClass, $default = null)
     {
         if ($this->isNotFilled($key) ||
             ! function_exists('enum_exists') ||
             ! enum_exists($enumClass) ||
             ! method_exists($enumClass, 'tryFrom')) {
-            return null;
+            return $default;
```
fgaroby, line 359: Isn't it a BC?

taylorotwell:
You can just do $request->enum() ?: 'default';

PR #43938, opened 2022-08-30
[9.x] Add getIntendedUrl method to redirector

iamgergo:
This PR adds the `getIntendedUrl` method the `Redirector` class. It feels more comfortable, than checking out the session directly.

### DIFF
--- a/src/Illuminate/Routing/Redirector.php
+++ b/src/Illuminate/Routing/Redirector.php
@@ -114,29 +114,6 @@ class Redirector
     }
 
     /**
-     * Set the intended url.
-     *
-     * @param  string  $url
-     * @return $this
-     */
-    public function setIntendedUrl($url)
-    {
-        $this->session->put('url.intended', $url);
-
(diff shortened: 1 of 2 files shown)


src/Illuminate/Routing/Redirector.php
```diff
@@ -126,6 +126,16 @@ public function setIntendedUrl($url)
         return $this;
     }
 
+    /**
+     * Get the intended url.
+     *
+     * @return string|null
+     */
+    public function getIntendedUrl()
```
akiyamaSM, line 134: Would be better to add type hinting

iamgergo: Yes, I would love that, but for now I think it's better be consistent. If most of the methods don't have typehints, I'd stick with that.

driesvints: Right now, we don't add type hints like these in the framework. We aim for consistency.

PR #38046, opened 2021-07-17
[8.x] Display unexpected validation errors when asserting status

jessarcher:
This PR expands on #38025 to also show any unexpected validation errors that occurred when asserting a status code other than 422.

Currently the only way to see the validation errors is by dumping the response, running the test again, and then removing the dump. With this PR, the unexpected validation errors will be displayed straight away.

If the response being asserted against is JSON, the validation errors will be also be displayed as JSON. Otherwise the errors will be displayed as a simple list.

PR #38025, opened 2021-07-15
[8.x] Track exceptions and display them on failed status checks for dx

taylorotwell:
This feature is inspired by feedback from @jessarcher and my own poor experience with testing endpoints that throw unexpected exceptions.

This problem typically arises when asserting against the status of a test response and receiving a failed assertions indicating the expected status code did not match the actual status code of `500`. You must then examine your logs to hopefully determine what the exception message was. Or, you can call `$this->withoutExceptionHandling()->get(...)` to disable exception handling for the test.

Ideally, Laravel would just show you the exception. This PR takes care of that by tracking exceptions while running unit tests and then displaying the exception failed status checks where the actual status code is `500`.

image

image

PR #43943, opened 2022-08-31
[9.x] Improve test failure output

jessarcher:
In https://github.com/laravel/framework/pull/38025 and https://github.com/laravel/framework/pull/38046, we improved the test output for the `assertStatus` methods to display any exceptions or errors that may have caused the status assertion to fail.

These improvements almost removed the need to call `withoutExceptionHandling`. Almost...

This PR expands on the previous work to display exceptions or errors on *any* failed assertion, not just status assertions, and not even just `TestResponse` assertions.

A common example is when asserting Inertia redirects, and an unexpected validation error occurs. The current output doesn't give any clues that the redirect assertion failure was due to a validation error:

image

Under the hood, this calls `assertStatus` and `assertLocation`, but it's only the `assertLocation` that fails, which doesn't currently display exceptions or errors.

With the changes in this PR, the error now looks like this:

image

> **Note** Unfortunately, it doesn't seem possible to inject the exception/error *after* the ComparisonFailure diff, so the output is not 100% what I'd want.

The existing `assertStatus` behaviour is now handled by this new logic, so there's no duplication of code. There is a minor change in output because the extra content is now appended _after_ PHPUnit's raw assertion error instead of before it:

**Before:**
image

**After:**
image

We'll now also see this extra context even if we don't assert against the TestResponse. For example, imagine this test where we expect a user record to be created:

```php
$this->post('/users', []);

$this->assertSame(1, User::count());
```

If a validation error or exception occurred during the last request, we'll get the extra context on PHPUnit's built-in assertions like `assertSame`:

image

The only scenario I can think of where this feature is undesirable is the `assertJson` assertion. If a validation error occurs, the `assertJson` assertion will already show the full response body. To prevent duplicate output, I've ensured that we don't append the JSON error response if the existing message already includes it.

If the user makes multiple requests in one test, only errors/exceptions from the latest request will be displayed, hopefully preventing red herrings.

The extra context will only be displayed when an assertion fails, so you won't see any output if you expect an error or exception as long as your assertions pass.

Developers may opt-out of this feature by overriding the `onNotSuccessfulTest` method in their base test case to throw the given exception like PHPUnit does by default.

### DIFF
--- a/src/Illuminate/Foundation/Testing/TestCase.php
+++ b/src/Illuminate/Foundation/Testing/TestCase.php
@@ -271,7 +271,8 @@ abstract class TestCase extends BaseTestCase
     /**
      * This method is called when a test method did not execute successfully.
      *
-     * @throws Throwable
+     * @param  Throwable  $exception
+     * @return void
      */
     protected function onNotSuccessfulTest(Throwable $exception): void
     {
--- a/src/Illuminate/Foundation/Testing/TestCase.php
+++ b/src/Illuminate/Foundation/Testing/TestCase.php
@@ -6,14 +6,19 @@ use Carbon\CarbonImmutable;
 use Illuminate\Console\Application as Artisan;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Foundation\Bootstrap\HandleExceptions;
+use Illuminate\Http\RedirectResponse;
 use Illuminate\Queue\Queue;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\Facade;
 use Illuminate\Support\Facades\ParallelTesting;
 use Illuminate\Support\Str;
+use Illuminate\Testing\AssertableJsonString;
 use Mockery;
 use Mockery\Exception\InvalidCountException;
+use PHPUnit\Framework\ExpectationFailedException;
 use PHPUnit\Framework\TestCase as BaseTestCase;
+use ReflectionProperty;
 use Throwable;
 
 abstract class TestCase extends BaseTestCase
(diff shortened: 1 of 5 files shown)


src/Illuminate/Foundation/Testing/TestCase.php
```diff
@@ -262,4 +267,115 @@ protected function callBeforeApplicationDestroyedCallbacks()
...
+        $exceptionToAppend = (string) $exceptionToAppend;
+        $message = <<<"EOF"
+            The following exception occurred during the last request:
+
+            $exceptionToAppend
+
+            ----------------------------------------------------------------------------------
+
+            $exceptionMessage
+            EOF;
```
jbrooksuk, line 330: I think we should put the `EOF` at the start of the line?

```suggestion
EOF;
```

src/Illuminate/Foundation/Testing/TestCase.php
```diff
@@ -262,4 +267,115 @@ protected function callBeforeApplicationDestroyedCallbacks()
...
+        // JSON error messages may already contain the errors, so we shouldn't duplicate them.
+        if (str_contains($exception->getMessage(), $errors)) {
+            return $exception;
+        }
+
+        $message = <<<"EOF"
+            The following errors occurred during the last request:
+
+            $errors
+            EOF;
```
jbrooksuk, line 358: Same here.

jbrooksuk (commented):
@jessarcher you may have put `EOF` where it is for indentation, so feel free to ignore this.

src/Illuminate/Foundation/Testing/TestCase.php
jessarcher: If we put it at the start of the line then we'll also need to move all of the other text inside it to the left as well, which I'm not a fan of.

I'm using the PHP 7.3+ [flexible heredoc syntax](https://laravel-news.com/flexible-heredoc-and-nowdoc-coming-to-php-7-3) here so the indentation starts from the closing tag (i.e. no indentation in this case)

jessarcher: Looks like GitHub syntax highlighting might not be sure how to handle it though :woman_facepalming:

jbrooksuk: Ah, gotcha!

jessarcher:
@nunomaduro I'd appreciate your thoughts on this one.

Do you see any issues hooking into `onNotSuccessfulTest` and appending to the `message` property on `ExpectationFailedException` exceptions?

Would this feature also be beneficial for Pest or can you think of any issues it might cause?

taylorotwell:
@jessarcher marking as draft until we get feedback from @nunomaduro

src/Illuminate/Foundation/Testing/TestCase.php
```diff
@@ -262,4 +269,116 @@ protected function callBeforeApplicationDestroyedCallbacks()
...
+                parent::onNotSuccessfulTest($this->appendErrorsToException($session->get('errors')->all(), $exception));
+
+                return;
+            }
+        }
+
+        if ($this->latestResponse->baseResponse->headers->get('Content-Type') === 'application/json') {
+            $testJson = new AssertableJsonString($this->latestResponse->getContent());
+
+            if (isset($testJson['errors'])) {
```
nunomaduro, line 304: @jessarcher Should we check that the status code is also 422?

src/Illuminate/Foundation/Testing/TestCase.php
```diff
@@ -262,4 +269,116 @@ protected function callBeforeApplicationDestroyedCallbacks()
...
+     * @return \Throwable
+     */
+    protected function appendErrorsToException($errors, $exception, $json = false)
+    {
+        $errors = $json
+            ? json_encode($errors, JSON_PRETTY_PRINT)
+            : implode(PHP_EOL, Arr::flatten($errors));
+
+        // JSON error messages may already contain the errors, so we shouldn't duplicate them.
+        if (str_contains($exception->getMessage(), $errors)) {
```
nunomaduro, line 353: Does this works? Asking because it seems we change the format of the JSON string by using the `JSON_PRETTY_PRINT` flag.

nunomaduro:
@jessarcher I've rebased, added minor changes, and made a few comments. Generally, this pull request seems OK to me. In addition, I've tested this pull request against Vapor test suite - more than 600 test cases - and seems to be all working.

src/Illuminate/Foundation/Testing/TestCase.php
jessarcher: The existing behaviour doesn't check this and I think that's okay. We don't mention anything about validation in the messaging, so I see it as an added bonus that this will work for users returning their own error payload for scenarios other than validation.

src/Illuminate/Foundation/Testing/TestCase.php
jessarcher: Yeah - the existing output in the exception is also pretty printed

PR #43952, opened 2022-08-31
[9.x] New flag: artisan make:test --model=XXXXX

PovilasKorop:
I want to help newcomers to testing and generate sample "smoke tests" for them, to test the typical resource controller routes.

The basic "version 1" idea is in this PR: new flag `php artisan make:test UserTest --model=User` would generate the feature UserTest from a specific stub, containing 7 methods, corresponding to exactly the methods of resource controllers: https://laravel.com/docs/9.x/controllers#resource-controllers

So, if the controller methods are implemented, the only thing to change in the generated test would be to enable `RefreshDatabase` after making sure you're working on the testing dababase and not on the production one.

Loom video with a 30-second demo: https://www.loom.com/share/d25be7d688eb44898cdabe40ba7c9d28

Now, of course, the Feature Test content is a little opinionated and assumes a few things:

- Factories are created
- All 7 resourceful controller methods exist and return the things they are typically supposed to return (status 200 or redirect)

But I still think it can be a massive helper to __newcomers__ to testing, to give them an example test, which they would then modify as they want, if some method doesn't pass initially.

If you don't think this should be a part of the core, no problem, I would create a separate package, with potentially more customizations.

- - - - - 

Open questions for discussions:

1. If I get a green light with this, I will also make a PEST implementation, with `php artisan make:test UserTest --pest --model=User`
2. Should RefreshDatabase be enabled by default, or it is too risky?
3. The test creates a `setUp()` method with `$this->followingRedirects()` to match two cases in post/put/delete requests: that those controllers return 2xx code, or that they redirect to success/index page. Not sure if it's the most elegant solution.
4. I wanted to extend that into `php artisan make:model --test` and `php artisan make:controller --test` but apparently those flags already exist there, just generate **empty** tests. If you give me green light, I could override those empty tests with the model-stub ones? Or should I create different flags for those like "--route-test" or "--resource-test" or something.

Any other suggestions or feedback are welcome. Thanks!

### DIFF
--- a/src/Illuminate/Foundation/Console/TestMakeCommand.php
+++ b/src/Illuminate/Foundation/Console/TestMakeCommand.php
@@ -49,7 +49,13 @@ class TestMakeCommand extends GeneratorCommand
      */
     protected function getStub()
     {
-        $suffix = $this->option('unit') ? '.unit.stub' : '.stub';
+        $suffix = '.stub';
+        if ($this->option('unit')) {
+            $suffix = '.unit.stub';
+        }
+        if ($this->option('model') && !$this->option('pest')) {
+            $suffix = '.model.stub';
+        }
 
         return $this->option('pest')
             ? $this->resolveStubPath('/stubs/pest'.$suffix)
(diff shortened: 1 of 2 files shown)


ousid:
When I saw the `--model` flag, I thought it's a model related test, is it best to switch the keyword into `resource`, or I am missing something?

PovilasKorop:
> When I saw the `--model` flag, I thought it's a model related test, is it best to switch the keyword into `resource`, or I am missing something?

Yeah, I was struggling with the naming. 

Resource refers to "API Resource" in other places of the framework, and if we mean resource controller then wouldn't it mean we need to specify the full controller? Like `--resource=TaskController` or something?

Also, the way the other generators are structured in Laravel, the model is the core fundamental name to then generate the controller/migration/other names/contents of other files. My long-shot goal was, in fact, to have `make:model --test` but, as I mentioned in the description, for now, I postponed it with the question, as that flag is already taken.

Rizky92:
I think this is better for the controller than the model, no? Since laravel already has test for controllers as well?  
`php artisan make:controller ModelController --test`  

Maybe if adding `--resource` flag alongside `--test` to generate resourceful test?

ousid:
> > When I saw the `--model` flag, I thought it's a model related test, is it best to switch the keyword into `resource`, or I am missing something?
> 
> Yeah, I was struggling with the naming.
> 
> Resource refers to "API Resource" in other places of the framework, and if we mean resource controller then wouldn't it mean we need to specify the full controller? Like `--resource=TaskController` or something?
> 
> Also, the way the other generators are structured in Laravel, the model is the core fundamental name to then generate the controller/migration/other names/contents of other files. My long-shot goal was, in fact, to have `make:model --test` but, as I mentioned in the description, for now, I postponed it with the question, as that flag is already taken.

`--resource=TaskController` make sense Indeed, but I don't know how you would control the `{ model}` and the `{route}`, because sometimes, devs name the controllers differently, something like `UserHomeController` or `UserAdminController`, but the model remain the same as `User`, and when I mentioned `--resource`, it's like following the controller `--resource` convention, something like `php artisan make:controller UserController --resource`

PovilasKorop:
> I don't know how you would control the { model} and the {route}

@ousid exactly the reason why I chose MODEL as the base name, from the model I can build route and other variables I need.

PovilasKorop:
> I think this is better for the controller than the model, no? Since laravel already has test for controllers as well? `php artisan make:controller ModelController --test`
> 
> Maybe if adding `--resource` flag alongside `--test` to generate resourceful test?

Yes, that also makes sense, but it would still then call the `make:test` under the hood, cause that's how `make:controller --test` works.

The parameter `--resource` is reserved, for generating the resource __controller__ itself, so then would need to rename the new flag to, don't know, `--test --resource-test`? Not sure, the word "resource" means so many things that I doubt that it would be clear for everyone.

PovilasKorop:
Another idea: execute this only if the controller is created with resource and model flags, and name it `--featuretest` or something

`php artisan make:controller TaskController --resource --model=Task --featuretest`

taylorotwell:
This feels like the type of opinionated scaffolding that should maybe be in a package? Thanks for the suggestion!

PR #43959, opened 2022-09-01
[9.x] Support objects like GMP for custom Model casts

### LINKED ISSUE #43870: Unable to cast database attributes to objects using custom casts
- Laravel Version: 9.26.1
- PHP Version: 8.1
- Database Driver & Version: Doesn't matter

### Description:

It is not possible to cast database attributes to objects using custom cast classes and then convert them to an array.

### Steps To Reproduce:

Models class:

```php
<?php

namespace App\Models;

use App\Casts\GMPCast;

class Balance extends Model
{
    protected $casts = [
        'amount' => GMPCast::class,
    ];
}

```

and our custom cast class:

```php
<?php

namespace App\Casts;

use GMP;
use Illuminate\Contracts\Database\Eloquent\SerializesCastableAttributes;

class GMPCast implements CastsAttributes, SerializesCastableAttributes
{
    {
        return gmp_init($value, 10);
    }

    {
        return gmp_strval($value, 10);
    }

    public function serialize($model, string $key, $value, array $attributes)
    {
        return gmp_strval($value, 10);
    }
}
```

The code:

```php
Balance::all()->toArray();
```

The error that happens:

```
Object of class GMP could not be converted to bool
```

The reason is obvious, here: [HasAttributes.php#L285](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L285)
we first cast the `amount` attribute to GMP class using custom cast, then inside the next `if` statements ([HasAttributes.php#L292](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L292), [HasAttributes.php#L296](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L296), [HasAttributes.php#L306](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L306) and [HasAttributes.php#L933](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L933)) we try to convert it to boolean, so the error happens.
The solution is to try to check the `$attributes[$key]` by not converting it to boolean implicitly.

driesvints:
This PR fixes an issue where some objects like GMP objects couldn't be used as custom casts. The reason is that these objects cannot be converted to booleans so an `if ($attribute[$key])` check will hard-fail. By using `isset` we can just check if the attribute is present and has a value.

Fixes https://github.com/laravel/framework/issues/43870

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -930,7 +930,7 @@ trait HasAttributes
         // If an attribute is listed as a "date", we'll convert it from a DateTime
         // instance into a form proper for storage on the database tables using
         // the connection grammar's date format. We will auto set the values.
-        elseif ($value && $this->isDateAttribute($key)) {
+        elseif (isset($value) && $this->isDateAttribute($key)) {
             $value = $this->fromDateTime($value);
         }
 
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -930,7 +930,7 @@ trait HasAttributes
         // If an attribute is listed as a "date", we'll convert it from a DateTime
         // instance into a form proper for storage on the database tables using
         // the connection grammar's date format. We will auto set the values.
-        elseif (isset($value) && $this->isDateAttribute($key)) {
+        elseif (! is_null($value) && $this->isDateAttribute($key)) {
             $value = $this->fromDateTime($value);
         }
 
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -289,11 +289,11 @@ trait HasAttributes
             // If the attribute cast was a date or a datetime, we will serialize the date as
             // a string. This allows the developers to customize how dates are serialized
             // into an array without affecting how they are persisted into the storage.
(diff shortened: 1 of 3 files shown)


src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
```diff
@@ -930,7 +930,7 @@ public function setAttribute($key, $value)
         // If an attribute is listed as a "date", we'll convert it from a DateTime
         // instance into a form proper for storage on the database tables using
         // the connection grammar's date format. We will auto set the values.
-        elseif ($value && $this->isDateAttribute($key)) {
+        elseif (isset($value) && $this->isDateAttribute($key)) {
```
sharifzadesina, line 933: Maybe it is better to use `! is_null($value)` here instead of `isset`, it is cleaner and consistent with [HasAttributes.php#L949](https://github.com/laravel/framework/blob/af9d6f3df83e7f5f4763902985bc5d02d6dad662/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L949) and [HasAttributes.php#L960](https://github.com/laravel/framework/blob/af9d6f3df83e7f5f4763902985bc5d02d6dad662/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L960).

driesvints:
```suggestion
        elseif (! is_null($value) && $this->isDateAttribute($key)) {
```

driesvints: Done, thanks.

PR #43963, opened 2022-09-01
[10.x] Support for BackedEnum in value helper

foremtehan:
when we use `value()` function we excepted this function return the 'value' of the argument we pass, so i think it make sense it return the value of BackedEnum instead of enum itself


taylorotwell:
I'm not sure I want to change this behavior. The purpose of the `value` helper is **only** to resolve closures. Otherwise, it returns **exactly** what you pass it.

PR #43965, opened 2022-09-01
[9.x] Support arrays on Enum validation rule

gdebrauwer:
The `in` and `exists` validation rules can be used to validate all values of an array, but the `Enum` rule does currently not support this. This PR fixes that.
```php
// Before
return [
    'enum_values_array' => [
        'required',
        'array',
        'min:1',
    ],
    'enum_values_array.*' => [
        'required',
        new Enum(SomeEnum::class),
    ]
]

// After
return [
    'enum_values_array' => [
        'required',
        'array',
        'min:1',
        new Enum(SomeEnum::class),
    ],
]
```


taylorotwell:
This feels a bit breaking? For example if I just had `['required', new Enum(...)]` before I could expect that it would ensure the incoming value was a string but now it might be an array?

PR #43971, opened 2022-09-01
[10.x] Added `Rule::boolean()` method and improved handling of `boolean` validation rule

andrey-helldar:
The main problem with the "boolean" string validation rule is that it ignores the textual representation of the `"true"` and `"false"` boolean values sent via [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append) JS.

My PR https://github.com/laravel/framework/pull/43955 was closed, but in return, @taylorotwell offered to move the functionality into a separate method.

In the course of development, I found out that the scheme [`proposed`](https://github.com/laravel/framework/pull/43955#issuecomment-1234411787) by Taylor will mislead developers because by writing, for example, `Rule::boolean(['foo', 'bar'])` we will get `true` for the `foo` value from the request, which is wrong.

Therefore, this method does not take external arguments, checking the field itself against certain values passed to the `in_array` function.

This is a breaking change because the values `"true"` and `"false"` are now "boolean".

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -435,7 +435,7 @@ trait ValidatesAttributes
      */
     public function validateBoolean($attribute, $value)
     {
-        $acceptable = [true, false, 0, 1, '0', '1'];
+        $acceptable = [true, false, 'true', 'false', 1, 0, '1', '0'];
 
         return in_array($value, $acceptable, true);
     }
--- a/src/Illuminate/Validation/Rule.php
+++ b/src/Illuminate/Validation/Rule.php
@@ -4,6 +4,7 @@ namespace Illuminate\Validation;
 
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Support\Traits\Macroable;
+use Illuminate\Validation\Rules\Boolean;
 use Illuminate\Validation\Rules\Dimensions;
 use Illuminate\Validation\Rules\ExcludeIf;
 use Illuminate\Validation\Rules\Exists;
@@ -137,5 +138,15 @@ class Rule
     public static function unique($table, $column = 'NULL')
     {
         return new Unique($table, $column);
+    }
+
+    /**
+     * Get a boolean constraint builder instance.
+     *
+     * @return \Illuminate\Validation\Rules\Boolean
+     */
+    public static function boolean()
+    {
+        return new Boolean();
     }
 }
--- a/src/Illuminate/Validation/Rules/Boolean.php
+++ b/src/Illuminate/Validation/Rules/Boolean.php
@@ -1,0 +1,53 @@
+<?php
+
+namespace Illuminate\Validation\Rules;
+
+use Illuminate\Contracts\Validation\Rule;
+
+class Boolean implements Rule
+{
+    /**
+     * The name of the rule.
+     */
+    protected string $rule = 'boolean';
+
+    /**
+     * The values to check.
+     *
+     * @var array
+     */
+    protected array $acceptable = [true, false, 'true', 'false', 1, 0, '1', '0'];
+
+    /**
+     * Determine if the validation rule passes.
+     *
+     * @param string $attribute
+     * @param mixed $value
+     *
+     * @return bool
+     */
+    public function passes($attribute, $value): bool
+    {
+        return in_array($value, $this->acceptable, true);
+    }
+
+    /**
+     * Get the validation error message.
+     *
+     * @return string
+     */
+    public function message(): string
+    {
+        return trans('validation.boolean');
+    }
+
+    /**
+     * Convert the rule to a validation string.
+     *
+     * @return string
+     */
+    public function __toString(): string
+    {
+        return $this->rule;
+    }
+}
--- a/src/Illuminate/Validation/Rules/Boolean.php
+++ b/src/Illuminate/Validation/Rules/Boolean.php
@@ -21,9 +21,8 @@ class Boolean implements Rule
     /**
      * Determine if the validation rule passes.
      *
-     * @param string $attribute
-     * @param mixed $value
-     *
+     * @param  string  $attribute
+     * @param  mixed  $value
      * @return bool
      */
     public function passes($attribute, $value): bool
--- a/tests/Validation/ValidationBooleanRuleTest.php
+++ b/tests/Validation/ValidationBooleanRuleTest.php
@@ -1,0 +1,112 @@
+<?php
+
+namespace Illuminate\Tests\Validation;
+
+use Illuminate\Container\Container;
+use Illuminate\Support\Facades\Facade;
+use Illuminate\Translation\ArrayLoader;
+use Illuminate\Translation\Translator;
+use Illuminate\Validation\Rule;
+use Illuminate\Validation\ValidationServiceProvider;
+use Illuminate\Validation\Validator;
+use PHPUnit\Framework\TestCase;
+
+class ValidationBooleanRuleTest extends TestCase
+{
+    public function testCorrectValues()
+    {
+        $this->passes(true);
+        $this->passes(false);
+
+        $this->passes('true');
+        $this->passes('false');
+
+        $this->passes(1);
+        $this->passes(0);
+
+        $this->passes('1');
+        $this->passes('0');
+
+        $this->passes('');
+    }
+
+    public function testIncorrectValues()
+    {
+        $this->fails('True');
+        $this->fails('False');
+
+        $this->fails('on');
+        $this->fails('off');
+
+        $this->fails('On');
+        $this->fails('Off');
+
+        $this->fails('yes');
+        $this->fails('no');
+
+        $this->fails('Yes');
+        $this->fails('No');
+
+        $this->fails(null);
+
+        $this->fails(12);
+        $this->fails('12');
+
+        $this->fails('custom');
+
+        $this->fails([]);
+
+        $this->fails(new Foo());
+        $this->fails(Foo::class);
+    }
+
+    protected function fails($value)
+    {
+        $this->assertValidationRules($value, false, ['value' => ['validation.boolean']]);
+    }
+
+    protected function passes($value)
+    {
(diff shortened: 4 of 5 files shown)


andrey-helldar:
In this case, you can use a more explicit check:

```php
$values = [true, false, 'true', 'false', 1, 0, '1', '0'];

return in_array($value, $values, true);
```

I think that would be more correct. I will make changes to the PR.

dennisprudlo:
I prefer the option proposed by Taylor. I think **not** having to specify the additional values will mislead developers because there is now a `boolean` and `Rule::boolean()` rule and both of them work differently.

andrey-helldar:
@dennisprudlo, Taylor's proposal has a drawback:

```php
$v = Validator::make(
    ['value' => 'foo'],
    ['value' => Rule::boolean(['foo', 'bar'])]
);

return $v->passes();
// is true
```

Thus, having the opportunity to pass custom values, we will turn this field into `Rule::in()`.

dennisprudlo:
@andrey-helldar I mean.. that's correct, right? `Rule::boolean()` will validate whether the value is a boolean or not. And you specified, that `'foo'` and `'bar'` should also be considered booleans. So it passes, because you told the validator, that `foo` is also a boolean.

Of course in the end `'foo'` is not really considered a boolean for PHP but for the validator it is. If you don't want `'foo'` to be considered a boolean, then it shouldn't be in that validation rule.

andrey-helldar:
@dennisprudlo, exactly. As well as the fact that it should not be possible to specify the `foo` value to check, since this would already be wrong.

taylorotwell:
`Rule::boolean` should work exactly like the `boolean` rule IMO. Only if you provide additional values should it work differently.

andrey-helldar:
@taylorotwell, I agree that the rules should work the same way. But I do not agree with giving the opportunity to independently determine which values will be booleans. After all, if you specify `Rule::boolean(['foo'])`, then when checking the field with the value `foo`, we will get `true`, which is wrong.

For example:

```
### HTTP Client from phpStorm
POST http://localhost/api/store
Content-Type: multipart/form-data; boundary=WebAppBoundary

--WebAppBoundary
Content-Disposition: form-data; name="is_foo"
Content-Type: text/plain

true
--WebAppBoundary
Content-Disposition: form-data; name="is_bar"
Content-Type: text/plain

bar
--WebAppBoundary
```

or

```javascript
const data = new FormData()

data.append('is_foo', true)
data.append('is_bar', 'bar')

axios.post(http://localhost/api/store, data)
```

Because the [`FormData`](https://developer.mozilla.org/en-US/docs/Web/API/FormData/append) object converts everything to strings.

And two cases:

```php
class StoreRequest extends FormRequest
{
    {
        return [
            'is_foo' => ['boolean'],
        ];
    }
}

// Getting a validation error because "true" is a string, not a boolean.
```

```php
class StoreRequest extends FormRequest
{
    {
        return [
            'is_bar' => [Rule::boolean(['foo', 'bar'])],
            // Validation won't return an error because `bar` value is in values.
        ];
    }
}

class StoreController extends Controller
{
    public function store(StoreRequest $request)
    {
        $value  = $request->get('is_bar');     // bar
        $is_bar = $request->boolean('is_bar'); // false
    }
}
```

Because:

```php
/**
 * Retrieve input as a boolean value.
 *
 * Returns true when value is "1", "true", "on", and "yes". Otherwise, returns false.
 *
 * @param  string|null  $key
 * @param  bool  $default
 * @return bool
 */
public function boolean($key = null, $default = false)
{
}
```

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Http/Concerns/InteractsWithInput.php#L311-L323

dennisprudlo:
> After all, if you specify `Rule::boolean(['foo'])`, then when checking the field with the value `foo`, we will get `true`, which is wrong.

I don't really think this is an issue. When someone specifies it that way, the validator will pass it. I mean why would anyone **explicitly(!)** tell the validator that `'bar'` is a boolean, then complain about `'bar'` passing the boolean validation?

andrey-helldar:
@dennisprudlo, ok. I'll ask the question differently: if I specify `foo` and it passes the validation check. Will it be `true` or `false`?

dennisprudlo:
Neither, it will be `'foo'`.

```php
// request data
[ 'somekey' => 'foo' ];

$validator = validator($request->all(), [
    'somekey' => Rule::boolean([ 'foo' ])
]);

$validator->passes(); // true, because we said 'foo' is a valid value for the boolean rule.

$request->somekey; // 'foo', it's neither true or false. The value is simply 'foo'

$request->boolean('somekey'); // false, because we force cast the value to a boolean.
```

Calling `$request->boolean()` works with any value. `$request->boolean('firstname')` will also be false if we try force cast the firstname to a boolean.

Now what I am trying to say is bluntly: "It's your own fault". You know you will cast a value to a boolean but you let a non-boolean value pass in your validation.

andrey-helldar:
@dennisprudlo, it's not my fault. This is a lack of technology. Read how FormData works:

image

https://developer.mozilla.org/en-US/docs/Web/API/FormData/append

```js
const data = new FormData()

data.append('is_foo', true)
data.append('is_bar', 'bar')

axios.post('http://localhost/api/store', data)
```

In this case, in the backend we get:

```php
[
    'is_foo' => 'true',
    'is_bar' => 'bar',
]
```

If the request does not plan to send a file, then there is no problem - we send `Content-Type: application/json` and work as it should, but if the request contains the `Content-Type: multipart/form-data` header, then we are faced with a nuance of the technology in the form of automatic everything in string.

That's what I'm talking about, to add not random words to the call method, but to use specific ones - `"true"` and `"false"`, since this is obvious.

Now you have to manually create your own boolean validation rule, so the boxed rule is never used:

```php
class BooleanRule implements Rule
{
    public function passes($attribute, $value): bool
    {
        $acceptable = [true, false, 'true', 'false', 1, 0, '1', '0'];

        return in_array($value, $acceptable, true);
    }

    public function message()
    {
        return trans('validation.boolean');
    }
}

class CustomRequest
{
    {
        // ...
        'is_admin' => ['required', new BooleanRule()]
    }
}
```

I use this custom rule in several projects.

This is because the box rule does not consider true and false to be boolean values.

Yes, from the point of view of the PHP language, these are strings that are not cast to a boolean type, but `yes`, `no`, `on` and `off` are also strings, but the same `filter_var` can convert them to booleans.

dennisprudlo:
What I mean is: Just don't add `'foo'` as an additional boolean value – It's your fault if you do. It doesn't really matter how FormData works. In the end – serverside – the only thing that matters is what data is received.

You say that the `Rule::boolean([ 'foo' ])` is logically wrong and that `'foo'` is not really a boolean. Well if `'foo'` is not a valid boolean, why do you add it as a possible boolean value here? Just don't do it. It's that simple.

FormData turns every boolean to a string? That's fine. Use `Rule::boolean([ 'true', 'false' ])` as your validation rule. Then `'foo'` will not pass the validation. Just `'true'` and `'false'`.

andrey-helldar:
@dennisprudlo, I'll try to explain it easier.

Let's say we have a `Rule::boolean([ 'true', 'false' ])` method.

Questions:
1. What prevents a developer from adding any other value to it, such as `Rule::boolean([ 'true', 'false', 'qwerty' ])`?
2. How will this rule differ from `Rule::in([true, false, 'true', 'false', 1, 0, '1', '0'])`?

dennisprudlo:
1. Nothing. But why should we prevent a developer from doing so?

Counter question: What prevents a developer from doing this:
```php
Rule::in([ app(), $this, new class extends User {}, [[[[[[]]]]]] ])
```

2. It won't, it's more or less an alias so we don't have to type it out every time

Counter question: How do these rules differ:
```php
[
    'alpha',
    'regex:/[A-Za-z]+/'
]
```

andrey-helldar:
I understand. You want to let developers decide which values are booleans, and at the same time you don't want to admit that `"true"` is `true` and `"false"` is `false`.

On this discussion I will stop. I have nothing more to say.

dennisprudlo:
@andrey-helldar Now you are just putting words in my mouth. I never said `'true'` is not `true` and `'false'` is not `false`. The only concern in this PR is – *and that is why I am against this change* – that the `boolean` and `Rule::boolean()` rule will validate differently and you won't be able to use them interchangeably.

`boolean` does not include the string versions of true and false and that's why `Rule::boolean()` shouldn't either – **by default**.

I am not against passing `'true'` and `'false'` as booleans, but then this change should also be done for the regular `boolean` validation rule. Of course this will result in a breaking change and should be PR'd for 10.x

With the option to specify custom values in `Rule::boolean([ ... ])` this PR could be merged into 9.x

andrey-helldar:
@dennisprudlo, when I wanted to fix the string rule (https://github.com/laravel/framework/pull/43955), Taylor suggested creating a new method with accepted parameters instead. When I started doing this, I realized that the new rule copies the behavior of the `Rule::in`. It is not right.

I agree that the rules should work the same way, but I don't agree that developers should decide which values ​​are booleans.

That's what I've been talking about all this time.

Judging by the objections, your words are interpreted both in favor of passing arguments and in favor of not adding string equivalents of boolean values.

To be honest, I'm already tired of discussing the obvious drawback of FormData technology with an attempt to optimize the backend for it.

Sorry for wasting time.

andrey-helldar:
And yet, the thought of doing the right thing does not leave me.

I'll update the PR tonight and rebase to the master branch. I will bring all methods to a single view.

andrey-helldar:
The static analyzer error is not mine - it's a https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Factories/Factory.php#L478-L493 file.

driesvints:
Can you rebase with master to fix the static analysis test?

andrey-helldar:
@driesvints, ready

taylorotwell:
I don't plan to break the boolean rule. Maybe I'll introduce Rule::boolean how I intended on 9.x.

andrey-helldar:
@taylorotwell, it makes no sense. `Rule::in()`'s rule already exists.

garygreen:
Yeah I don't see any reason to add this or a `Rule::Boolean` - `Rule::in` has got that handled already.

PR #43972, opened 2022-09-02
[9.x] Add strict mode for when() conditionals

krzysztofrewak:
Recently I and @michalmyskow found out that `when()` conditional method is resolving parameters to "trurthy" and "falsy" accordingly to the documentation. For example you cannot find any user with `"0"` phrase, because `when()` converts string zero to boolean false:

```php
$search = $request->get("search"); // "0"

User::query()
    ->when($search, function(Builder $query) => $query->where("name", "LIKE", "%$search%"))
    ->toSql();
// "SELECT * FROM users;"
```

This change adds non-breaking change of parameters of `when()` and `unless()` methods. With this:

```php
$search = $request->get("search"); // "0"

User::query()
    ->when($search, function(Builder $query) => $query->where("name", "LIKE", "%$search%"), strict: true)
    ->toSql();
// "SELECT * FROM users WHERE name LIKE '%0%';"
```

### DIFF
--- a/src/Illuminate/Conditionable/Traits/Conditionable.php
+++ b/src/Illuminate/Conditionable/Traits/Conditionable.php
@@ -8,7 +8,7 @@ use Illuminate\Support\HigherOrderWhenProxy;
 trait Conditionable
 {
     /**
-     * Apply the callback if the given "value" is (or resolves to) truthy.
+     * Apply the callback if the given "value" is (or resolves to) truthy or true with strict mode.
      *
      * @template TWhenParameter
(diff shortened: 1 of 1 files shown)


dennisprudlo:
Wouldn't a simple `blank()`, 'filled()` or `strlen()` solve the issue?

```php
User::query()
    ->when(strlen($search), fn (Builder $query) => $query->where("name", "LIKE", "%$search%"))
    ->toSql();
```

taylorotwell:
Yeah I would do something like @dennisprudlo suggests.

PR #43974, opened 2022-09-02
[9.x] Added support for notification based events

tkaratug:
This PR adds support for executing any kind of action while sending a notification or after a notification is sent. And those actions are defined in notifications.

This can be done already using notification events but those events are dispatched for all notifications in the project. With this PR it will be possible to execute any kind of action only for a specific notification.

1- Actions that need to be executed while sending a notification can be added in `onSending()` method of the notification.
```php
public function onSending($channel): void
{
    logger('The notification is being sent via '.$channel);
}
```

2- Actions that need to be executed after a notification is sent can be added in `onSent()` method of the notification.
```php
public function onSent($channel): void
{
    logger('The notification has been sent via '.$channel);
}
```

The final notification will be like following:
```php
class FooNotification extends Notification
{
    public function __construct()
    {
        //
    }

    public function via($notifiable): array
    {
        return ['database', 'mail'];
    }

    public function toMail(mixed $notifiable): MailMessage
    {
        return (new MailMessage)
            ->subject('foo')
            ->markdown('emails.foo_notification');
    }

    public function toDatabase(mixed $notifiable): array
    {
        return [
            'message' => 'Lorem ipsum dolor sit amet',
        ];
    }

    public function onSent($channel): void
    {
        logger($this::class.' has been sent via '.$channel);
    }

    public function onSending($channel): void
    {
        logger($this::class.' is being sent via '.$channel);
    }
}
```

### DIFF
--- a/src/Illuminate/Notifications/NotificationSender.php
+++ b/src/Illuminate/Notifications/NotificationSender.php
@@ -147,6 +147,10 @@ class NotificationSender
 
         $response = $this->manager->driver($channel)->send($notifiable, $notification);
 
+        if (method_exists($notification, 'onSent')) {
+            $notification->onSent($channel);
+        }
+
         $this->events->dispatch(
             new NotificationSent($notifiable, $notification, $channel, $response)
         );
--- a/src/Illuminate/Notifications/NotificationSender.php
+++ b/src/Illuminate/Notifications/NotificationSender.php
@@ -171,6 +171,10 @@ class NotificationSender
             return false;
         }
 
+        if (method_exists($notification, 'onSending')) {
+            $notification->onSending($channel);
+        }
+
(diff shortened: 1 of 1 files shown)


bert-w:
How about https://laravel.com/docs/9.x/notifications#notification-events

tkaratug:
> How about https://laravel.com/docs/9.x/notifications#notification-events

Yes, you may be right but notification events are dispatched for all notifications in the project. My point is to execute any kind of action only for a specific notification.

taylorotwell:

tkaratug:
Hi,

I have released a package for the need mentioned in this PR. Hope it will be useful for the community.
https://github.com/tkaratug/laravel-notification-event-subscriber

PR #43975, opened 2022-09-02
[9.x] Remove mail fake failures method

crissi:
Mail::failures does not exists anymore, so should also be removed from the MailFake.


taylorotwell:
I don't really want to remove methods on a patch release. Can remove on 10.x.

PR #43977, opened 2022-09-02
[9.x] Adds `Process` convenience layer 🧘🏻

nunomaduro:
This pull request proposes an expressive, minimal API around Symfony's Process, allowing you to quickly run processes in your Laravel application. Just like the `HTTP` facade, this pull request introduces a `Process` facade that is focused on the most common use cases, testing, and a **wonderful developer experience**.

usage

## Running Processes

To run a process, you may use the `run` method provided by the `Process` facade. Let's examine how to run a basic `ls` command:

```php
use Illuminate\Support\Facades\Process;

$result = Process::run('ls');

$result->ok(); // true
$result->output(); // my-file-1, my-file-2
```

The `run` method returns an instance of `Illuminate\Console\Contracts\ProcessResult`, which provides a variety of methods that may be used to inspect the process result:

```php
$result = Process::run('ls');

$result->output(): string;
$result->errorOutput(): string;
$result->exitCode(): int;
$result->ok(): bool;
$result->failed(): bool;
```

The `Illuminate\Console\Contracts\ProcessResult` object also implements the PHP `ArrayAccess` interface, allowing you to iterate over each line of the output. The output, is "exploded" via the `\n` character: 

```php
$files = Process::run('ls');

foreach ($files as $file) {
    dd($file); my-file-1
}

$files[0]; // my-file-1
$files[1]; // my-file-2
```

### Accessing process's output at real-time

By default, when using the `output` method, Laravel waits for the process to be finished before giving you the entire process's output:

```php
$result = Process::run('echo 1; sleep 1; echo 2');

dd($result->output()); // 1, 2
```

If you wish to access the process's output at real-time, you may use a closure as second argument of the `run` method:

```php
$result = Process::run('echo 1; sleep 1; echo 2', function ($output) {
    dump($output); // Runs twice, outputting 1 and 2 respectively ...
});

dd($result->output()); // 1, 2
```

In addition, if you with to know if the given output is from the type `stderr` or `stderr`, you may use the second argument of given closure:

```php
Process::run('echo 1; sleep 1; echo 2', function ($output, $type) {
    dump($type == "out"); // true
    dump($type == "err"); // false
});
```

### Dumping processes

If you would like to dump the outgoing process instance before it is sent and terminate the script's execution, you may add the `dump` or `dd` method to the beginning of your process definition:

```php
Process::dd()->run('ls');

// ^ Illuminate\Console\Process^ {#662
//  -commandline: "ls"
//  -cwd: "/Users/nunomaduro/Work/Code/laravel"
// ...
```

### Options

Of course, it is common when running processes to specify multiple command arguments, configure the path, timeouts, and more.

#### Process Command

When running processes, you may pass an string or an array of strings as the first argument to the `run` method:

```php
Process::run('ls -la');
Process::run(['ls', '-la', 'other-option']);
```

#### Process Path

You may use the `path` method if you would like to specify the working directory / base path of the process:

```php
$path = storage_path('images');

Process::path($path)->run('ls'); // image-1.jpg, image-2.jpg
```

#### Process Timeout

The `timeout` method may be used to specify the maximum number of seconds to wait for a process:

```php
Process::timeout(3)->run('sleep 2');
```

If the given timeout is exceeded, an instance of `Illuminate\Console\Process\Exceptions\ProcessTimedOutException` will be thrown.

```php
try {
    Process::timeout(3)->run('sleep 4');
} catch (ProcessTimedOutException $e) {
    dd($e->result()->exitCode()); // 143
}
```

If you don't care about the time a process takes to run, and you would just like a process to run forever, you may use the `forever` method:

```php
Process::forever()->run('something that takes minutes');
```

#### Process Exceptions

If you have a process result and would like to throw an instance of `Illuminate\Console\Exceptions\ProcessFailedException` if the exit code indicates an error, you may use the `throw`, `throwIf`, or `throwUnless` methods:

```php
$result = Process::run('exit 1');
$result->exitCode(); // 1 

// Throw an exception if an error occurred...
try {
    $result->throw();
}  catch (ProcessFailedException $e) {
    dd($e->result()->exitCode()); // 1
}
 
$result->throwIf($condition);
 
$result->throwUnless($condition);
```

The `throw` method returns a regular result instance if no error occurred, allowing you to chain other operations onto the `throw` method:

```php
$result = Process::run('ls');

$result->throw()->output(); // my-file-1, my-file-2
```

If you would like to perform some additional logic before the exception is thrown, you may pass a closure to the `throw`, `throwIf`, or `throwUnless`  methods. The exception will be thrown automatically after the closure is invoked, so you do not need to re-throw the exception from within the closure:

```php
return Process::run('exit 1');->throw(function (ProcessFailedException $e) {
    // Perform some clean-up ...
})->ok();
```

## Concurrent Processes

Sometimes, you may wish to run multiple processes requests concurrently. In other words, you want several processes to be dispatched at the same time instead of issuing the processes sequentially. This can lead to substantial performance improvements when interacting with slow processes.

Thankfully, you may accomplish this using the `pool` method. The pool method accepts a closure which receives an `Illuminate\Console\Process\Pool` instance, allowing you to easily add `processes` to the process pool for dispatching:

```php
$results = Process::pool(fn (Pool $pool) => [
    $pool->run('sleep 1'),
    $pool->run('sleep 1'),
    $pool->run('sleep 1'),
]); // 1 second

return $results[0]->ok() &&
       $results[1]->ok() &&
       $results[2]->ok();
```

## Asynchronous Processes

By default, processes are synchronous, meaning the method `run` will automatically wait for the process to be finished before returning the process's result. Yet, if you need to run processes asynchronously while performing other tasks in your code, you may use the `async` method:

```php
$result = Process::async()->run(/* ... */);

// Do other tasks...

return $result->wait()->output();
```

## Macros

The Process's factory is also "macroable", meaning that you may define the macro within the boot method of your application's `App\Providers\AppServiceProvider` class:

```php
use Illuminate\Support\Facades\Process;
 
/**
 *
 * @return void
 */
public function boot()
{
    Process::macro('ls', fn () => Process::command('ls'));
}
```

Once your macro has been configured, you may invoke it from anywhere in your application to create a pending process with the specified configuration:

```php
Process::ls()->run();
```

## Testing

The `Process` facade's fake method allows you to instruct Laravel to return stubbed / dummy results when processes are made.

For example, to instruct the `Process` to return successful empty result, you may call the `fake` method with no arguments:

```php
use Illuminate\Support\Facades\Process;
 
Process::fake();
 
$result = Process::run(/* ... */);
$result->ok(); // true
$result->output(); // ''
```

### Faking specific processes

Alternatively, you may pass an array to the `fake` method. The array's keys should represent command patterns that you wish to fake and their associated results. The `*` character may be used as a wildcard character. Any processes made to Process commands that have not been faked will actually be executed. You may use the Process facade's `result` method to construct stub / fake results for these processes:

```php
Process::fake([
    'ls' => Process::result(['file-1.php', 'file-2.php'),
    'curl *' => Process::result(json_encode(['item'])),
]);
```

If you would like to specify a fallback command pattern that will stub all unmatched commands, you may use a single `*` character:

```php
Process::fake([
    'ls' => Process::result(['file-1.php', 'file-2.php'),

    '*' => Process::result(),
]);
```

#### Fake callback

If you require more complicated logic to determine what results to return for certain processes, you may pass a closure to the fake method. This closure will receive an instance of `Illuminate\Console\Process` and should return a result instance. Within your closure, you may perform whatever logic is necessary to determine what type of result to return:

```php
Process::fake(fn () => Process::result('An error message', 1));
```

### Inspecting Processes

When faking processes, you may occasionally wish to inspect the processes in order to make sure your application is sending the correct commands. You may accomplish this by calling the `Process::assertRan` method after calling `Process::fake`.

The `assertRan` method accepts a closure which will receive an `Illuminate\Console\Process` instance and should return a boolean value indicating if the process matches your expectations. In order for the test to pass, at least one request must have been issued matching the given expectations:

```php
use Illuminate\Support\Facades\Process;
 
Process::fake();
 
Process::timeout(50)->run('sleep 30');
 
Process::assertRan(function (\Illuminate\Console\Process $process) {
    return $process->command() == 'sleep 30') && $process->timeout() == 30.0;
});

Process::assertRan('sleep 30');
```

If needed, you may assert that a specific process was not ran using the `assertNotRan` method:

```php
use Illuminate\Support\Facades\Process;
 
Process::fake();
 
Process::run('ls');
 
Process::assertNotRan(function (\Illuminate\Console\Process $process) {
    return $process->command() == 'ls -la');
});

Process::assertNotRan('ls -la');
```

You may use the `assertRanCount` method to assert how many processes were "ran" during the test:

```php
Process::fake();
 
Process::assertRanCount(5);
```

Or, you may use the `assertNothingRan` method to assert that no processes were ran during the test:

```php
Process::fake();
 
Process::assertNothingRan();
```

---

@taylorotwell Something to keep in mind while reviewing this pull request:

As mentioned https://github.com/laravel/framework/issues/44278, currently processes inherit all the env vars defined in your application / system. In theory, this seems no deal, yet it can cause stuff like this to happen:

```php
// On app A, runs command on app B
Process::path('app-b-path')
    ->run('php artisan migrate:refresh'); // refreshes database of app A with migrations of app B.
```


src/Illuminate/Console/Process/PendingProcess.php
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Create a new Pending Process instance.
+     *
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        // ..
```
rodrigopedra, line 35: I don't know any other class in Laravel which have an empty constructor. I am not against it as the docs are useful, but I wonder about consistency.

Also doc block has 2 empty lines instead of one.

src/Illuminate/Console/Process/PendingProcess.php
```diff
@@ -0,0 +1,74 @@
...
+
+    /**
+     * Sets the process's path.
+     *
+     * @param  string  $path
+     * @return $this
+     */
+    public function path($path)
+    {
+        return tap($this, fn () => $this->path = $path);
```
rodrigopedra, line 57: I do use `tap(...)` a lot on my projects, and the usage in `run()` method is something I think `tap()` brings simplicity.

But here and in the `withArgument` method, I don't see the advantage of using it. No temporary variables would be needed, both are simple assignments, and I know discussing micro-optimization is out of question, but it adds two function calls for no benefit (in my opinion).

Also I think in such cases it is actually more complex to skim through, and I don't know any other *"wither"* or fluent setter in the framework code which use this code style.

timacdonald:
@nunomaduro loving this.

Can we get a `throw` function up in here 🙏

```php
return Process::run('ls')->throw()->output();
```

src/Illuminate/Console/Process/PendingProcess.php
nunomaduro: This PR is a work in progress, ignore the code for now.

src/Illuminate/Console/Process/PendingProcess.php
nunomaduro: This PR is a work in progress, ignore the code for now.

whoami-idk:
add some methods to run tasks in the background or get the proccess pid.

ralphjsmit:
First of all, I just saw this PR and I really love it. Thanks for the great implementation!

You showed that the `->throw()` method receives a closure to handle some addition logic before an exception is thrown. Developers don't need to re-throw the exception again, because it is automatically thrown after the closures has been called. 

Personally I think that it would be somewhat more flexible to not automatically throw the exception after executing the closure. That would allow developers to provide just a closure and handle the exception themselves, without throwing (e.g. using `report()`). Sometimes a command isn't really that important. That's why I'd propose 1) to not automatically throw the exception when there is a closure passed, or 2) to add a method like `->catch()` that receives a closure as well, but doesn't throw the exception after executing the closure.

ralphjsmit:
Could be you already planned on doing this, but interesting idea would be to add a `Process::sequence()` method to fake a sequence of process results, similar to the `Http::sequence()`.

src/Illuminate/Console/Process/PendingProcess.php
```diff
@@ -0,0 +1,239 @@
...
+
+    /**
+     * Sets the process's timeout.
+     *
+     * @param  float|null  $timeout
+     * @return $this
+     */
+    public function timeout($timeout)
+    {
+        return tap($this, fn () => $this->timeout = $timeout);
```
rodrigopedra, line 155: Now this is not a draft anymore, could you revisit it? I am copying my last comment below to avoid explaining it again. Thanks =)

---

I do use `tap(...)` a lot on my projects, and the usage in `run()` method is something I think `tap()` brings simplicity.

But here and in the `withArgument` method, I don't see the advantage of using it. No temporary variables would be needed, both are simple assignments, and I know discussing micro-optimization is out of question, but it adds two function calls for no benefit (in my opinion).

Also I think in such cases it is actually more complex to skim through, and I am not aware of any other *"wither"* or fluent setter in the framework code which use this code style.

nunomaduro: Thanks for your comment. We are going to keep the `tap` for now.

DarkGhostHunter:
1. I would change "path" for "atPath". 
2. Can a process receive a signal?

src/Illuminate/Console/Contracts/ProcessResult.php
```diff
@@ -0,0 +1,109 @@
...
+     * @throws \Illuminate\Console\Exceptions\ProcessFailedException
+     */
+    public function throwUnless($condition);
+
+    /**
+     * Get the process's output.
+     *
+     * @return string
+     */
+    public function toString();
```
rodrigopedra, line 108: There is already `output()`, `__toString()` (due to implementing `\Stringable`), why does this 3rd option is needed?

src/Illuminate/Console/Process/PendingProcess.php
```diff
@@ -0,0 +1,260 @@
...
+     * Register a stub callable that will intercept processes and be able to return stub process result.
+     *
+     * @param  array<int, callable(\Illuminate\Console\Process): (\Illuminate\Console\Contracts\ProcessResult|null)>  $callbacks
+     * @return $this
+     */
+    public function stubs($callbacks)
+    {
+        $this->stubs = $callbacks;
+
+        return $this;
```
rodrigopedra, line 130: `return tap($this, fn () => $this->stubs = $callbacks);` to be consistent with the other setters/withers below?

Just for the record I prefer this style over the other, but as you already said you will keep the ones using `tap(...)` I guess it is better to be consistent, at least on the same class, don't you think?

timacdonald:
@ralphjsmit changing the throwing logic would make it inconsistent with the Http client. I think it might be nice if it was consistent and perhaps had additional affodances to handle other cases.

aneesdev:
Name a better love story than @nunomaduro x artisan console <3

IServerNetworkable:
amazing pr. Great work, hope to see this merged

src/Illuminate/Console/Exceptions/ProcessAlreadyStarted.php
```diff
@@ -0,0 +1,18 @@
...
+class ProcessAlreadyStarted extends LogicException
+{
+    /**
+     * Creates a new Process Exception instance.
+     *
+     * @return void
+     */
+    public function __construct()
+    {
+        parent::__construct('The process has already been started.');
```
deleugpn, line 16:
```suggestion
        parent::__construct('The process has alread started.');
```

gofish543:
Is there plans on adding the ability to interface with the "Prepared Commands" of Symfony's Process module?
https://symfony.com/blog/new-in-symfony-4-1-prepared-commands

driesvints:
@nunomaduro tests still fail here.

src/Illuminate/Console/Process/PendingProcess.php
```diff
@@ -0,0 +1,272 @@
...
+namespace Illuminate\Console\Process;
+
+use Illuminate\Console\Exceptions\ProcessAlreadyStarted;
+use Illuminate\Console\Process;
+use Illuminate\Console\Process\Results\Result;
+use Illuminate\Support\Traits\Macroable;
+
+class PendingProcess
+{
+    use Macroable;
```
gofish543, line 12: Can we include the Conditional trait?

WendellAdriel:
Checking the conversation I'm not really sure, but is there anything that I could help with here?

timacdonald:
It would be nice if we could, like we can with the HTTP client, enforce that faking process calls are faked throughout the whole testsuite.

For example, the following test fails...

```php
class ProcessTest extends TestCase
{
    public function test_all_process_calls_must_be_faked_throughout_the_testsuite()
    {
        // Imagine this is in the global TestCase::setUp()...

        Process::fake(function () {
            throw new \Exception('You should fake all proces calls in your tests');
        });

        // In this test...

        Process::fake(['echo *' => Process::result('fake output')]);

        $this->assertSame("fake output", Process::run('echo "hello world"')->output());

        // ❌ Exception is thrown
    }
}
```

I solved this on the HTTP client with a new method `Http::preventStrayRequests()`.

https://github.com/laravel/framework/blob/2641bc172640ec625d84a0e66b0fd6caf112fb2a/src/Illuminate/Support/Facades/Http.php#L101-L111

So it would be nice if we could do the same here...


```php
class ProcessTest extends TestCase
{
    public function test_all_process_calls_must_be_faked_throughout_the_testsuite()
    {
        // Imagine this is in the global TestCase::setUp()...

        Process::preventStrayCommandExecutions();

        // In this test...

        Process::fake(['echo *' => Process::result('fake output')]);

        $this->assertSame("fake output", Process::run('echo "hello world"')->output());

        // ✅

        Process::run('ls');
 
        // ❌ Throw an exception.
    }
}
```

PR #43980, opened 2022-09-02
[9.x] Add missing /DateInterval typehint to `expireAfter`

iBotPeaches:
Screen Shot 2022-09-02 at 9 28 30 AM

Saw this warning, but `expireAfter` calls `secondsUntil` which takes - `@param  \DateTimeInterface|\DateInterval|int  $delay` which allows it work from the Interval. So a simple update to docblock.


GrahamCampbell:
@iBotPeaches Note that Laravel does not support named params/have a BC promise on them.

iBotPeaches:
> @iBotPeaches Note that Laravel does not support named params/have a BC promise on them.

Okay. This was about a wrong docblock though. So don't understand the ping.

PR #43981, opened 2022-09-02
[9.x] Allow validator messages to use nested arrays

ollieread:
This PR allows for validator messages to be nested without the dot notation.

Here's an example of the current way of providing custom messages.

```php
[
   'attribute.rule1' => 'message1',
   'attribute.rule2' => 'message2',
   'attribute.rule3' => 'message3',
   'attribute.rule4' => 'message4',
   'attribute.rule5' => 'message5',
]
```

This PR makes it possible to nest the `rule => message` elements.

```php
[
   'attribute' => [
       'rule1' => 'message1',
       'rule2' => 'message2',
       'rule3' => 'message3',
       'rule4' => 'message4',
       'rule5' => 'message5',
    ]
]
```

The idea is that it makes it easier to read and identify.

### DIFF
--- a/src/Illuminate/Validation/Concerns/FormatsMessages.php
+++ b/src/Illuminate/Validation/Concerns/FormatsMessages.php
@@ -115,6 +115,8 @@ trait FormatsMessages
                         if (is_array($message) && isset($message[$lowerRule])) {
                             return $message[$lowerRule];
                         }
+
+                        return $message;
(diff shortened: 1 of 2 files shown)


ollieread:
> @ollieread will it work for `FormRequest` attributes?

That's specifically why I created it, working for all validators was a side effect.

PR #43987, opened 2022-09-02
[9.x] Automatically convert field's values to json string when using Model::upsert

bardiz12:
This PR adds a feature to automatically **json_encode** array/object value when using **Model::upsert** method. 

If we have a JSON column in the database, when we want to do batch inserting/updating using the **Model::upsert method** we have to manually json_encode it's value. i think we can handle that in the upsert method automatically based on **$casts** property that defined in the Model's class.

usually we will do this to do upsert query to a table with JSON field : 
```php
$values = [
    [
        'name' => 'J.K. Rowling',
        'books' => [
            [
                'title' => "Harry Potter and the Philosopher's Stone",
                "year" => "1997"
            ],
            ....
        ]
    ],
    [
        'name' => 'J. R. R. Tolkien',
        'books' => [
            [
                'title' => "The Lord of the Rings",
                "year" => "1954"
            ],
            ...
        ]
    ],
];

$values = collect($values)->map(function ($item) {
    return [
        'name' => $item['name'],
        'books' => json_encode($item['books'])
    ];
})->toArray();

Author::query()
    ->upsert($values, ['name']);
```

with this PR, if in the model we **cast** the field to one of these : 
- `object`, 
- `array`, 
- `\Illuminate\Database\Eloquent\Casts\AsArrayObject::class`, or 
- `\Illuminate\Database\Eloquent\Casts\AsCollection::class` 

those fields values would be converted to JSON string automatically without the needs to do json_encode manually again in the user's codebase

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1022,8 +1022,12 @@ class Builder implements BuilderContract
             $update = array_keys(reset($values));
         }
 
+
+        $modifiedUpdateValues = $this->addTimestampsToUpsertValues($values);
+        $modifiedUpdateValues = $this->convertArrayObjectToJsonString($modifiedUpdateValues, $update);
+
         return $this->toBase()->upsert(
-            $this->addTimestampsToUpsertValues($values),
+            $modifiedUpdateValues,
             $uniqueBy,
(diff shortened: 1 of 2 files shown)


taylorotwell:

taylorotwell:
If we go down this road I'm not sure it makes sense to only partially support some casts and not all.

PR #43991, opened 2022-09-03
[9.x] Add previous exception

nagmat84:
This PR ensures that the original (or previous) exception is not lost, when the exception handler wraps the exception into a `HttpException` in `prepareResonse`.

Knowing the original exception is helpful (or even mandatory) if the exception shall be handled in a special way in `renderHttpRepsonse` later down in the call stack. Moreover it is best practice to keep the chain of exception intact.

Note, this also does not introduce any security risk as it is the responsibility of `renderHttpRepsonse` anyway to only include information into the final HTML response which are deemed safe.

### DIFF
--- a/src/Illuminate/Foundation/Exceptions/Handler.php
+++ b/src/Illuminate/Foundation/Exceptions/Handler.php
@@ -536,7 +536,7 @@ class Handler implements ExceptionHandlerContract
         }
 
         if (! $this->isHttpException($e)) {
-            $e = new HttpException(500, $e->getMessage());
+            $e = new HttpException(500, $e->getMessage(), $e);
         }
 
         return $this->toIlluminateResponse(


taylorotwell:
You would have already logged / reported the exception in the `reportable` callback so it is not lost.

nagmat84:
The point is _not_ about _logging_ the exception (but even there it might turn out to be helpful to have access to the original exception.)

The point is about handling the exception in a special way. In order to decide whether an exception requires special treatment it is more reliable to check for the exception type with `instanceof` than falling back to some string comparison of the message. But in order to do so, the original type of the exception must be preserved.

Moreover, I don't get the argument. Only because the exception has already been logged (or reported in Laravel terms), is not a good reason to break the chain of exceptions. The whole point of chaining exceptions together via the `$previous` attribute is to preserve the lineage and chain of provenance.

The whole PR only changes a single line and only adds a single argument to a constructor call to fix something which IMHO looks rather like an oversight than an intentional decision.

nagmat84:
The only change of this PR is
```php
$e = new HttpException(500, $e->getMessage());
```
vs.
```php
$e = new HttpException(500, $e->getMessage(), $e);
```

@taylorotwell Any chance that this gets still merged?

antonkomarev:
Exception chains must always be preserved

nagmat84:
> Exception chains must always be preserved

That my intention. That is what this PR is all about. Currently it is not.

PR #26108, opened 2018-10-13
[5.7] Re-add mixins for models

ZacharyDuBois:
Re-add the mixins for the Model abstract to facilitate IDEs code inspections and autocompletion.

They were originally deleted by commit 6f518a8a534702da9166ef2fc5e297a06ed5cc18. 

Without these mixins, the following occurs:
screen shot 2018-10-12 at 21 27 00

With the mixins:
screen shot 2018-10-12 at 21 29 04

As you can see, it reduces the code inspection error and also facilitates auto-competing the methods.

PR #43994, opened 2022-09-04
[9.x] Update Model.php to assist in IDE detection of Builder methods

BrekiTomasson:
Adding the `@mixin` docblock helps various IDE:s automatically detect that `Builder` methods are permitted on Models.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -23,6 +23,9 @@ use Illuminate\Support\Traits\ForwardsCalls;
 use JsonSerializable;
 use LogicException;
 
+/**
+ * @mixin \Illuminate\Contracts\Database\Eloquent\Builder
+ */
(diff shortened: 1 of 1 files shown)


rodrigopedra:
See discussion in: #42778 and #26108

I am all in for adding it, as the main argument is to avoid breaking a third-party code, which I actually currently use, but I don´t agree that the core should refrain from adding such simple life-improvement changes as this based on third-party impact.

driesvints:
Add them to your own model or use Laravel IDE Helper please.

BrekiTomasson:
> Add them to your own model or use Laravel IDE Helper please.

This is a very disappointing answer, unfortunately.

I already use my own model, but that doesn't help me when using third party packages which include models of their own. Also, the Laravel IDE Helper is not a good solution as cmd-clicking a method takes me to the Laravel IDE Helper file rather than the "real" declaration of the method in question, a problem that does not occur when simply adding the `@mixin` docblock.

PR #43999, opened 2022-09-04
[9.x] Allow injecting tagged instances via attribute

abenerd:
This PR add the `Tagged` attribute which can be added to a constructor argument that is explicitly typehinted as `iterable`, this way tagged servives can be required without having to explicitly create a custom binding.

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -1006,6 +1006,19 @@ class Container implements ArrayAccess, ContainerContract
             return $parameter->getDefaultValue();
         }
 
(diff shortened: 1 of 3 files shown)


driesvints:
Heya. I don't feel like we should complicate things here. I'm not sure why this is an improvement over a custom binding.

abenerd:
It allows one to omit having to maintain the custom binding when for example the dependencies change and with the attribute the configuration of the service is all within one file, at least this part of the configuration. I feel like creating the custom binding is a bit of boilerplate code that can be omitted by using the attribute.

taylorotwell:

PR #44002, opened 2022-09-05
Option to move null values at the end for Ascending Sort/Order

Afcyy:
I've faced cases when I wanted to sort/order values ascendingly, but I didn't want null values at the start.

When I googled I noticed that many users face the same issue so I decided to make it somewhat easier for everyone to sort ascendingly but move null values at the end of the results

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -1411,6 +1411,23 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
+     * Sort the collection in ascending order using the given callback with the option to move null values at the end.
+     *
+     * @param array<array-key, (callable(TValue, TKey): mixed)|array<array-key, string>|(callable(TValue, TKey): mixed)|string  $callback
+     * @param int $options
+     * @return static
+     */
+    public function sortByAsc($callback, $options = SORT_REGULAR)
+    {
+        define("SORT_ASC_NULLS_LAST", 6);
+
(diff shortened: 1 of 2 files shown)


Afcyy:
I decided to instead of being optional these functions should move null values at the and since orderBy/sortBy already exists

taylorotwell:

KasparRosin:
Hey @taylorotwell !
I could consider this a bug, though it's not exactly laravels fault.
This is because `postgres` and `mysql` differentiate in the way they order results when null values are involved.

Take a simple query like
```sql
SELECT id, parent_id FROM pages ORDER BY parent_id DESC
```

Here are the two results in `postgres` and in `mysql`.
This has caused a lot of headache within packages that are supposed to support both `mysql` and `postgres`.

Postgres
| parent_id | id |
|-----------|----|
| null      | 1  |
| null      | 2  |
| 2         | 3  |
| 1         | 4  |

Mysql
| parent_id | id |
|-----------|----|
| 2      | 3  |
| 1      | 3  |
| null         | 2  |
| null         | 1  |

KasparRosin:
Here is how we've usually solved this.
This query works (tested) the same for postgres and mysql.

```sql
SELECT parent_id, id FROM pages ORDER BY parent_id IS NULL, parent_id DESC
```

Here is what a laravel implementation could look like.
Maybe there are alternative ways two write the orderBy part, but this has been the most readable for me.
```php
Page::select(['id', 'parent_id'])->orderBy(DB::raw('parent_id IS NULL, parent_id'), 'desc');
```

PR #44003, opened 2022-09-05
[10.x] remove mail fake failures method

crissi:
Mail::failures does not exists anymore, so should also be removed from the MailFake.


driesvints (approved):

driesvints:
The static analysis failure has nothing to do with this PR. It's fixed on 9.x and master.

PR #44005, opened 2022-09-05
[9.x] Add $instance->withoutTimestamps()

timacdonald:
This feature has been PR'd before, but I hoping this simplified version may be more palatable for core. This PR adds the ability to turn off timestamps of a single model for the duration of a given callback.

```php
$user = User::first();


// `updated_at` is not changed...

$user->withoutTimestamps(
    fn () => $user->update(['reserved_at' => now()])
);
```

A lot of these `without*` helpers are global / static to the Model, but this is on an instance by instance basis.

This, for example, is **not** how you use this helper...

```php
$users = User::get();

$user[0]->withoutTimestamps(function () use ($users) {

    // timestamp is not updated...

    $users[0]->update(['reserved_at' => now()]);

    // timestamp is updated because `withoutTimestamps` was called on $users[0]...

    $users[1]->update(['reserved_at' => now()]);
});
```

Instead you would do the following for a collection of models...

```php
User::get()->each->withoutTimestamps(
    fn ($user) => $user->update(['reserved_at' => now()])
);
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
@@ -170,6 +170,10 @@ trait HasTimestamps
 
         $this->timestamps = false;
 
-        return tap($callback($this), fn () => $this->timestamps = true);
+        try {
+            return $callback($this);
+        } finally {
+            $this->timestamps = true;
+        }
     }
 }
--- a/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
@@ -155,4 +155,21 @@ trait HasTimestamps
     {
         return $this->qualifyColumn($this->getUpdatedAtColumn());
     }
+
+    /**
+     * Run the given callable without timestamping the model.
+     *
+     * @param  callable  $callback
+     * @return mixed
+     */
+    public function withoutTimestamps(callable $callback)
+    {
+        if (! $this->usesTimestamps()) {
+            return $callback();
+        }
+
+        $this->timestamps = false;
+
+        return tap($callback(), fn () => $this->timestamps = true);
+    }
 }
--- a/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
@@ -165,11 +165,11 @@ trait HasTimestamps
     public function withoutTimestamps(callable $callback)
     {
         if (! $this->usesTimestamps()) {
-            return $callback();
+            return $callback($this);
         }
 
         $this->timestamps = false;
 
-        return tap($callback(), fn () => $this->timestamps = true);
+        return tap($callback($this), fn () => $this->timestamps = true);
     }
 }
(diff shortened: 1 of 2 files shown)


ralphjsmit:
Just curious, was there a particular reason for not making it static?

Tofandel:
@ralphjsmit Because the `$timestamps` property on a model is not static (as are none of the others like `$fillable`, `$hidden` etc)

And except constants, laravel tries to never use static as it's an anti pattern

ralphjsmit:
> @ralphjsmit Because the `$timestamps` property on a model is not static (as are none of the others like `$fillable`, `$hidden` etc)

I was more thinking of stuff like `User::withoutEvents(/** */)`. It feels somewhat inconsistent to now switch to a non-static method for `withoutTimestamps`. 
 
> And except constants, laravel tries to never use static as it's an anti pattern and instead uses singletons

I know about singletons and that Laravel tries to not use static properties for storing stuff. It was mainly that we have a static `withoutEvents` function, so why not a static `withoutTimestamps()` function to preserve the consistency.

timacdonald:
Mainly because a recent attempt to add the static version failed. I would be happy to refactor and make this static while approaching it slightly differently to the previous attempt and see if I can simplify that as well.

See: https://github.com/laravel/framework/pull/43441

src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
```diff
@@ -155,4 +155,21 @@ public function getQualifiedUpdatedAtColumn()
...
+     */
+    public function withoutTimestamps(callable $callback)
+    {
+        if (! $this->usesTimestamps()) {
+            return $callback($this);
+        }
+
+        $this->timestamps = false;
+
+        return tap($callback($this), fn () => $this->timestamps = true);
```
derekmd, line 173: Copying `withoutEvents()` behaviour, I think this should handle callback throwables like `QueryException`. Otherwise `$timestamps = true` may not be restored.

```suggestion
        try {
            return $callback($this);
        } finally {
            $this->timestamps = true;
        }
```

timacdonald: Updated and tested. Thanks!

johanrosenson:
> > was there a particular reason for not making it static 
> Because a [recent attempt](https://github.com/laravel/framework/pull/43441) to add a static version failed. I would be happy to refactor and make this static while approaching it slightly differently to the previous attempt if Taylor would be interested in that.

IMHO don't refactor and make it static, having it accept a callback is perfect, this puts the responsibility on whether to use timestamps or not on the saver and not the retriever.

timacdonald:
The static version would also accept a Closure.

rodrigopedra:
> Because a recent attempt to add a static version failed.

That attempt only target the `updated_at`, and by doing so it changed a lot of methods. I guess doing it statically, but keeping it simple and straightforward would be fine, and more consistent with the other `without*()` methods.

By the way, nice PR, I've been doing this manually very often

taylorotwell:
Probably if we do this feature we would want it to be static so it applies to operations within a given callback similar to withoutEvents, etc. Would feel a bit more consistent with the other "without" methods that way. I do think the other PR got closed because it was just a little overwhelming - but maybe there is no simpler way to do it.

timacdonald:
I'll take a go at it. I think there is potential to simplify while keeping is static.

ralphjsmit:
Was afk today, will check this tomorrow!

PR #44007, opened 2022-09-05
[9.x] Fix documented return type of Database\Concerns\BuildsQueries::first()

martin-heralecky:
This changes PhpDoc-documented return type of method `Database\Concerns\BuildsQueries::first()` to `mixed`. The reason is that `Database/Query/Builder::get()` calls a `Database\Query\Processors::processSelect()`, which may be extended to return anything. Therefore, `Database\Concerns\BuildsQueries::first()` may return anything as well.


martin-heralecky:
Can I get a reason for closing? @taylorotwell

PR #44012, opened 2022-09-05
[9.x] Support Enumerable in Stringable

erikgaal:
This PR adds simple quality of life improvement which allows us to pass `Enumerable` objects such as `Collection` to the `Stringable` helper functions that already accept `array` as a parameter.

For example, in a project I'm working on, we're doing the following:

```php
/** @var Collection<array-key, string> $months */
$months = months();

Str::remove($months->toArray(), $someString);
```

With this PR we can skip the `->toArray` call.

```php
/** @var Collection<array-key, string> $months */
$months = months();

Str::remove($months, $someString);
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -218,12 +218,16 @@ class Str
      * Determine if a given string contains a given substring.
      *
      * @param  string  $haystack
-     * @param  string|string[]  $needles
+     * @param  string|string[]|Enumerable<array-key, string>  $needles
      * @param  bool  $ignoreCase
      * @return bool
      */
     public static function contains($haystack, $needles, $ignoreCase = false)
     {
+        if ($needles instanceof Enumerable) {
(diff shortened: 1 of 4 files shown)


src/Illuminate/Support/Str.php
```diff
@@ -242,12 +246,16 @@ public static function contains($haystack, $needles, $ignoreCase = false)
      * Determine if a given string contains all array values.
      *
      * @param  string  $haystack
-     * @param  string[]  $needles
+     * @param  string[]|Enumerable<array-key, string>  $needles
```
rodrigopedra, line 249: `iterable` maybe?

https://www.php.net/manual/en/language.types.iterable.php

erikgaal:
The code works, but PHPStan throws an error now even when passing an `array`. Will investigate and send a PR to fix.

https://phpstan.org/r/363ce01b-696f-4e26-8c90-ba9f93b943e8

PR #44022, opened 2022-09-06
Prunable contracts

markwalet:
**What**
I added 2 contracts to improve the typehinting and static analysis experience for prunable model.

In my previous  PR regarding pruning, I did an update to the PruneCommand to make it better extendible. Now, I'm getting phpstan errors when trying to call `pruneAll()` manually. Since traits are not type-hintable, I've decided to add 2 interfaces for this.

**Why**
I figured this change is welcome because we also use it on the `Authenticatable` trait. Type hinting is used more frequently on that interface ofcourse, but I figured we should give the developer the choice here. If this is not the way to go I could always just suppress the code style warning.

**No breaking changes**
I purposefully kept the support for using the trait without the interface to prevent breaking changes.

### DIFF
--- a/src/Illuminate/Contracts/Eloquent/MassPrunable.php
+++ b/src/Illuminate/Contracts/Eloquent/MassPrunable.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Illuminate\Contracts\Eloquent;
+
+interface MassPrunable
+{
+    /**
+     * Prune all prunable models in the database.
+     *
+     * @param  int  $chunkSize
+     * @return int
+     */
+    public function pruneAll(int $chunkSize = 1000);
+
+    /**
+     * Get the prunable model query.
+     *
+     * @return \Illuminate\Database\Eloquent\Builder
+     */
+    public function prunable();
+}
--- a/src/Illuminate/Contracts/Eloquent/Prunable.php
+++ b/src/Illuminate/Contracts/Eloquent/Prunable.php
@@ -1,0 +1,28 @@
+<?php
+
+namespace Illuminate\Contracts\Eloquent;
+
+interface Prunable
+{
+    /**
+     * Prune all prunable models in the database.
+     *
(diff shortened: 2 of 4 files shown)


driesvints:
Heya. I don't feel this provides any value. It's just dead code?

markwalet:
I understand why you feel this way. Let me explain a little further.

It was mainly meant to make type hinting possible for this trait, or even provide your own implementation for prunable models instead of using the given one. Just like authenticatable models work right now. 

In the current state you are required to use the Prunable trait because else the command will not pick up on them.

If you don't agree we can close the merge request :)

markwalet:
Looks like the pipeline failed, but I'm not sure if it is my fault. Can you try to rerun the workflow?

driesvints:
Done.

markwalet:
@driesvints 

> Done.

Thanks! Let me know what you think

<hr>
@ollieread 

> @markwalet could you define "pruning" in this sense? I'm trying to understand the PR but it's not entirely clear from the code changes.

I'm talking about the [Laravel model pruning functionality](https://laravel.com/docs/9.x/eloquent#pruning-models).

I have several reasons for adding the interface:
- Interface checking is faster than trait checking (doesn't really count I guess, wouldn't shave off more than a ms)
- Unlike traits, interfaces can be type-hinted. Which improves static code analysis.
- You can now implement the interface your own way, without using the trait. The `php artisan model:prune` command will still work.
- This change unifies the behaviour compared with the `Authenticable` trait & contract.

Does this clarifies things for you?

driesvints:
> Thanks! Let me know what you think

I still don't see much value in this PR, sorry :-/

markwalet:
Alright, no problem. I can also add them to my own repositories ;)

Is there a difference between the Authenticatable contracts and this? Or should I create a PR to cleanup/deprecate those interfaces as well?

taylorotwell:

PR #44023, opened 2022-09-06
Cache - support for dependencies between cached items

MarekVikartovsky:
In this PR request I would like to introduce cache dependencies. The main purpose of cache dependencies is to create connection of various cache items with each other by specific identifiers. This functionality is intended to help with simple work with the invalidation of cache records that chain each other. The programmer does not have to remember all the names of the cache keys or the exact sequence of cache tags from which he wants to delete cached items. You just have to set dependencies and invalidate them. Cache dependencies are not supported in database cache store.

Cache dependencies usage example:

When storing records in the cache, you can set the dependencies like:

```
 Cache::dependencies('my_cache_dependency')->set('test_item', 1); 
Cache::tags(['taggroup_1', 'taggroup_2'])->dependencies('my_cache_dependency')->set('test_item', 1);
```

Also setting more dependencies for cached item is supported as well, you just have to set them as an array: 

```
Cache::dependencies(['my_cache_dependency', 'another_cache_dependency'])->set('test_item', 1);
```

Or you can set dependencies as key, value pair: (this is useful if you want to make cache dependency by specific identifier/s from database etc.) 

```
Cache::dependencies(['product' => [1]])->set('test_item', 1); 
Cache::dependencies(['product' => [1, 2, 3, ..etc]])->set('test_item2’, 1);
```

If you want to invalidate (remove) all the records from the cache that have set specific dependency or dependencies, you can use one of the following codes: 

```
Cache::dependencies('my_cache_dependency')->invalidate();
Cache::dependencies(['my_cache_dependency', 'another_cache_dependency'])->invalidate();
Cache::dependencies(['product' => [1]])->invalidate();
```

If you have any questions or suggestions, let me ask. Thank you!

### DIFF
--- a/src/Illuminate/Cache/ApcStore.php
+++ b/src/Illuminate/Cache/ApcStore.php
@@ -119,6 +119,20 @@ class ApcStore extends TaggableStore
     }
 
     /**
+     * Begin executing a new dependency operation.
+     *
+     * @param  array|mixed  $dependencies
+     * @return \Illuminate\Cache\DependencyCache
+     */
+    public function dependencies($dependencies)
+    {
+        return new DependencyCache(
+            $this,
+            is_array($dependencies) ? $dependencies : func_get_args(),
+        );
+    }
+
+    /**
      * Get the cache key prefix.
      *
      * @return string
--- a/src/Illuminate/Cache/ArrayStore.php
+++ b/src/Illuminate/Cache/ArrayStore.php
@@ -160,6 +160,20 @@ class ArrayStore extends TaggableStore implements LockProvider
     }
 
     /**
+     * Begin executing a new dependency operation.
+     *
+     * @param  array|mixed  $dependencies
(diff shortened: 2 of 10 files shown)


taylorotwell:
I'm not sure I can really take on maintenance of this atm. Maybe you could release a package for this? 🙂

jackie111222:
Thanks for question. Yes, I can release it as a separate package but I thought, if there is a support for cache tags, it could be nice to have cache dependencies support as native as well.

Thanks for an answer. Have a nice day :)

jackie111222:
> I'm not sure I can really take on maintenance of this atm. Maybe you could release a package for this? 🙂

Thanks for question. Yes, I can release it as a separate package but I thought, if there is a support for cache tags, it could be nice to have cache dependencies support as native as well.

Thanks for an answer. Have a nice day :)

PR #44030, opened 2022-09-06
[9.x] Fix Stringable typehints with Enumerable

erikgaal:
Fixes a bug with static analysis using PHPStan that was introduced in #44012. I've taken the liberty of making the `Str` and `Stringable` helpers work with iterators where possible, and broadened the types as well.

Below is an example of the errors and how they are fixed now.
https://phpstan.org/r/a8f703b0-5d8c-470e-8b7f-2ce3adbcdf1a

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -218,22 +218,21 @@ class Str
      * Determine if a given string contains a given substring.
      *
      * @param  string  $haystack
-     * @param  string|string[]|Enumerable<array-key, string>  $needles
+     * @param  string|iterable<string>  $needles
      * @param  bool  $ignoreCase
      * @return bool
      */
     public static function contains($haystack, $needles, $ignoreCase = false)
     {
-        if ($needles instanceof Enumerable) {
-            $needles = $needles->toArray();
-        }
-
         if ($ignoreCase) {
             $haystack = mb_strtolower($haystack);
-            $needles = array_map('mb_strtolower', (array) $needles);
         }
 
         foreach ((array) $needles as $needle) {
+            if ($ignoreCase) {
+                $needle = mb_strtolower($needle);
+            }
+
             if ($needle !== '' && str_contains($haystack, $needle)) {
                 return true;
             }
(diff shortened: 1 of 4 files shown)


src/Illuminate/Support/Str.php
```diff
@@ -813,9 +797,9 @@ public static function replaceArray($search, $replace, $subject)
...
      *
-     * @param  string|string[]|Enumerable<array-key, string>  $search
-     * @param  string|string[]|Enumerable<array-key, string>  $replace
-     * @param  string|string[]|Enumerable<array-key, string>  $subject
+     * @param  string|array<string>|Enumerable<array-key, string>  $search
+     * @param  string|array<string>|Enumerable<array-key, string>  $replace
+     * @param  string|array<string>|Enumerable<array-key, string>  $subject
      * @return string
      */
     public static function replace($search, $replace, $subject)
```
erikgaal, line 805: We could also make this work with `iterable`, but then we need to call `iterator_to_array`, which might not be as efficient? Not sure if that's true though.

erikgaal: Made it work with `iterable` now. Did some benchmarks and tried to keep the same performance by not over-naively passing everything to `collect`.

src/Illuminate/Support/Str.php
```diff
@@ -341,21 +340,19 @@ public static function finish($value, $cap)
...
         $value = (string) $value;
 
-        if (empty($patterns)) {
-            return false;
+        if (! is_iterable($pattern)) {
+            $pattern = [$pattern];
         }
 
-        foreach ($patterns as $pattern) {
+        foreach ($pattern as $pattern) {
```
u01jmg3, line 355: This doesn't look to be correct.

nbayramberdiyev:
This makes sense. [PHPStan documentation](https://phpstan.org/writing-php-code/phpdoc-types#iterables) says:

> If `Collection|Foo[]` means “`Collection` or `array`” in your case even if `Collection` implements `Traversable`, you need to disambiguate the type by using `Collection|array<Foo>` instead.

vlakoff:
Code style:

In `contains()` and `endsWith()` you used:
```php
        if (! is_iterable($needles)) {
            $needles = (array) $needles;
        }
```

Whereas in `startsWith()` you used:
```php
        if (! is_iterable($needles)) {
            $needles = [$needles];
        }
```

Additionally, in `is()` and `isMatch()` there are:
```php
        if (! is_iterable($pattern)) {
            $pattern = [$pattern];
        }
```

Suggestions:
* At least, `contains()`, `startsWith()` and `endsWith` should be harmonized.
* To improve further, the above five methods should be harmonized.

vlakoff:
Note that, provided `$needles` equals null, `(array) $needles` gives `[]` whereas `[$needles]` gives `[null]`. The former doesn't enter into the `foreach`, the latter relies on `(string) $needle !== ''`.

In the case at hand, for converting non-iterable to array, I would definitely pick `[$needles]`:
* If provided `$needles` null, the most probable case if that the method dynamically receives _**one**_ needle (which may be undefined, expecting the method to return false).
* More consistent with, provided `$needles` empty string `''`, `(array) $needles` and `[$needles]` both of which give `['']`.
* In any case, we can't remove the string cast in `(string) $needle !== ''`, as the method may be provided array containing null (e.g. `['foo', null, 'bar']`).
* Proper handling if provided a Stringable object (casting object to array would be wrong).
* `[$value]` is more descriptive, less surprising, has less pitfalls than these `(array) $value` casts.

PR #44035, opened 2022-09-07
[9.x] Add RequiredIfAccepted validation rule

pascalbaljet:
This PR adds a `required_if_accepted` rule. The field under validation must be present and not empty if the *anotherfield* field is accepted (`"yes"`, `"on"`, `1`, or `true`).

```php
Validator::make([
    'is_company'   => '1',
    'company_name' => 'Apple',
], [
    'is_company'   => 'required|boolean',
    'company_name' => 'required_if_accepted:is_company',
]);
```


pascalbaljet:
The PHP 8.0 test fails because of an unrelated Redis error, but I can't restart the job :)

garygreen:
Couldn't you just do:

```
required_if:is_company,on
```

This PR adds a bool-liker checker which I'm not personally a fan of, as it causes issues with having to handle "on" "yes" which is quirky. Often better to just use `0` `1` for true/false on/off as it's easier handled.

pascalbaljet:
> This PR adds a bool-like checker which I'm not personally a fan of, as it causes issues with having to handle "on" "yes" which is quirky. Often better to just use very explicit defines values - I prefer `0` `1` for true/false on/off as it's easier to handle.

You could argue the same for the `accepted` rule. This new rule combines the `accepted` and `required` rules. It's just a little shortcut.

garygreen:
> You could argue the same for the accepted rule. This new rule combines the accepted and required rules. It's just a little shortcut.

I'm not a fan of the `accepted` style rules either - what does it mean to be "accepted"? "true" "false" "approved" "accepted" ? This PR reminds me of https://github.com/laravel/framework/pull/43955 - these general validation rules have a great degree of feature creep over time and can become a source of weird bugs. Honestly, it's better to just define explicitly what values you expect.

PR #44037, opened 2022-09-07
[9.x] Add Vite asset path generation method

wilsenhc:
This PR adds a method to the `Vite` class so the the way the asset URLs are generated can be separated from the main class.

This would make it so Teams/Projects that have their own way of loading assets can extend the functionality of the `Vite` class and add their own method of loading assets. 

This would ease the path of migrations from Laravel-Mix to Vite for those teams.

The `assetPath()` method has the same signature as the global `asset()` helper

### DIFF
--- a/src/Illuminate/Foundation/Vite.php
+++ b/src/Illuminate/Foundation/Vite.php
@@ -219,7 +219,7 @@ class Vite implements Htmlable
 
                     $tags->push($this->makeTagForChunk(
                         $partialManifest->keys()->first(),
-                        asset("{$buildDirectory}/{$css}"),
+                        $this->assetPath("{$buildDirectory}/{$css}"),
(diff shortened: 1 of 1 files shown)


timacdonald:
Just note that the `vite-plugin` uses the `ASSET_URL` under the hood as well, so swapping this may not work as intended.

I'd recommend just setting the `ASSET_URL` in your different teams as needed.

peterthomson:
> I'd recommend just setting the `ASSET_URL` in your different teams as needed.

Just FYI, a delta between the asset URL and the physical folder can be weird, annoying and confusing. But it's the whole point of this (and similar optional customisations) in webpack / mix. For example shared hosting often has weird symlinks and rules around public folders. These types of customisations help the build pipeline for those developers.

PR #44040, opened 2022-09-08
[9.x] Adds interval casting

DarkGhostHunter:
## What?

Allows a model cast to support datetime intervals through the [`CarbonInterval`](https://carbon.nesbot.com/docs/#api-interval).

```php
class Subscription extends Model
{
   protected $casts = ['period' => 'interval'];
}

$subscription = Subscription::create(['period' => '1 month 1 hour']);

DB::table('subscriptions')->value('period'); // "P1MT1H"
```

A Carbon Interval can be used to manipulate Carbon instances and calculate periods of time, and the cast allows to safely store that interval. I usually use this for subscriptions, to extend a subscription cycle.

## The standard way

The cast is handled by `Illuminate\Support\CarbonInterval` instead of the base class because it needs to be `Arrayable` and override the `toArray()` to return a string on array serialization (overridable by the user).

The cast will create an interval from a human period like `1 month` or `yesterday, next friday`, spec periods like `P1MT1H`, and `DateInterval`, `CarbonInterval` or `Closure` instances. This is all handled internally by `CarbonInterval`.

The cast, as you expect, will be persisted using the ISO 8601 spec (format with designators). Since is a string, it can be safely stored as an string column.

## Caveats

- The object is not immutable, as PHP doesn't have an immutable version for intervals.
- Builder methods helpers no included, so nothing like `$builder->whereInterval('period', $interval)`.  Comparisons in SQL are tricky for intervals and are out of the scope for this PR.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -21,6 +21,7 @@ use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\LazyLoadingViolationException;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Carbon;
+use Illuminate\Support\CarbonInterval;
 use Illuminate\Support\Collection as BaseCollection;
 use Illuminate\Support\Facades\Crypt;
 use Illuminate\Support\Facades\Date;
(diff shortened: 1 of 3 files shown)


ankurk91:
I would suggest to name the cast as `datetime_interval` for more clarity

DarkGhostHunter:
> I would suggest to name the cast as `datetime_interval` for more clarity 

Dunno. Which other type of interval exist?

taylorotwell:

PR #44041, opened 2022-09-08
[9.x] Do not trim the request URL in testing

browner12:
URLs with and without a trailing slash are technically different routes according to the browser, so we should not be auto trimming the trailing slash here.

I've recently been trying to setup a middleware to redirect all non-root URLs that end in a trailing slash to be redirected to their non-trailing-slash counterpart. While trying to setup feature test for this behavior, I discovered this was impossible due to the fact all test URLs are having their trailing slashes trimmed automatically.

Looking back through the history of this line of code, it appears the `trim()` has been there since the beginning, so it's difficult to discern the reasoning behind having the `trim()`.

I'm not sure how other people feel if this is a breaking change or not, but it's currently targeting 9.x, so let me know if that's not correct.

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
@@ -559,7 +559,7 @@ trait MakesHttpRequests
             $uri = substr($uri, 1);
         }
 
-        return trim(url($uri), '/');
+        return url($uri);
     }
 
     /**


browner12:
Well, turns out this doesn't actually solve the problem. The `url()` function also does its own trimming of trailing slashes, but changing that has farther reaching implications.  I think this PR is still okay, since the `trim()` I removed is repetitive.

One interesting thing to note about the `url()` function, if you provide a fully valid absolute URL it will not trim the trailing slash, but if you provide a relative path it will.

```php
url('testing/');                          // http://mysite.test/testing

url('http://mysite.test/testing/');       // http://mysite.test/testing/
```

taylorotwell:
I'm not sure I want to mess with this on a patch release.

PR #44046, opened 2022-09-08
[9.x] Added return type declaration for stubs

andrii-trush:
Added return type declaration. Each time after class creation from artisan command requires manually add return type for each method.


jbrooksuk:
You can publish and customise the stubs to your liking.

andrii-trush:
@jbrooksuk Yeah and each project to start with copying with stubs and edit them. Nice solution, answer NO, when you need default return types like in PHPDoc

PR #44050, opened 2022-09-08
[10.x] Do not trim the request URL

browner12:
this is a resubmission of #44041 to the master branch.

URLs with and without a trailing slash are technically different routes according to the browser, so we should not be auto trimming the trailing slash here.

the `url()` function also does its own trimming, so this commit doesn't actually solve the problem, but it does get rid of the duplicate trimming. changing `url()` has wider reaching effects, so will have to look closer into that change.

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
@@ -566,7 +566,7 @@ trait MakesHttpRequests
             $uri = substr($uri, 1);
         }
 
-        return trim(url($uri), '/');
+        return url($uri);
     }
 
     /**


garygreen:
It's going in 10.x so isn't breaking. Also it's just for testing. Seems like a sensible change to me, though would need to be documented in upgrade guide.

driesvints:
Right, maybe so.

browner12:
FYI, the only users it will actually affect is people who are writing absolute URLs in their tests. The `url()` function does not trim these. If people are using relative URLs (as I'm assuming most people do), they will still be unaffected due to `url()` doing its own trimming.

taylorotwell:
I don't think I'm going to merge this atm.

You should be able to test your middleware by instantiating it and calling the `handle` method directly then inspecting the response you get back?

browner12:
yah, I can _unit_ test it okay, but I can't _feature_ test it with current framework code.

PR #44053, opened 2022-09-08
[9.x] Prevent workflows from running twice

Jubeki:
### Explanation

When creating a PR from a branch within the same repo (i.e. not a fork), the current run-test workflows get triggered twice - for both push and pull_request:

Screenshot 2022-09-08 at 18 08 50

This PR adds constrains so that the workflow only runs once.
(Push Event will only be triggered on the `master` and `*.x` branch)

### Sources

Idea and Changes are taken from https://github.com/spatie/package-skeleton-laravel/pull/119
Thanks and Credits to @jessarcher 

Branch names can use patterns for Pull Requests:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet

### Side note
I think it would be good to change the default branch to main per
https://github.com/github/renaming#new-repositories-use-main-as-the-default-branch-name

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -2,6 +2,9 @@ name: static analysis
 
 on:
   push:
+    branches:
+      - master
+      - '*.x'
   pull_request:
   schedule:
     - cron: '0 0 * * *'
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -6,8 +6,6 @@ on:
       - master
       - '*.x'
   pull_request:
-  schedule:
-    - cron: '0 0 * * *'
 
 jobs:
   src:
--- a/.github/workflows/databases.yml
+++ b/.github/workflows/databases.yml
@@ -1,6 +1,11 @@
 name: databases
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - master
+      - '*.x'
+  pull_request:
 
 jobs:
   mysql_57:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,6 +2,9 @@ name: tests
 
 on:
   push:
+    branches:
+      - master
+      - '*.x'
   pull_request:
   schedule:
     - cron: '0 0 * * *'


GrahamCampbell:
I think that's actually desirable that they run twice, allowing testing before cutting a PR.

Jubeki:
> I think that's actually desirable that they run twice, allowing testing before cutting a PR.

I am sorry, but could you please explain what you mean with "cutting a PR"? 😅

GrahamCampbell:
Opening a pull request.

garygreen:
It would be better to only listen on [`pull_request`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) because by default it should trigger whenever any changes occur to a pull request because it has the default types of `opened` `synchronize` and `reopened`:

> By default, a workflow only runs when a pull_request event's activity type is opened, synchronize, or reopened. To trigger workflows by different activity types, use the types keyword

garygreen:
Another project with related discussion: https://github.com/open-telemetry/opentelemetry-python/issues/1370

driesvints:
Bit torn. I've thought about this a lot in the past already and my gut feeling says it'll be annoying to not be able to build on individual branches anymore before sending in a PR. 

> It would be better to only listen on [pull_request](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) because by default it should trigger whenever any changes occur to a pull request because it has the default types of opened synchronize and reopened:

That would mean directly pushed code to existing stable branches isn't tested anymore?

driesvints:
@Jubeki @GrahamCampbell @taylorotwell maybe we should just give this a try to see how it goes and then evaluate if we want to keep or ditch it.

driesvints (approved):

Jubeki:
> I think that's actually desirable that they run twice, allowing testing before cutting a PR.

If you fork the repo then tests will be not run by default on your fork. Also because there is the `cronjob` event in the workflow that means that you have to specifically enable them.

> Bit torn. I've thought about this a lot in the past already and my gut feeling says it'll be annoying to not be able to build on individual branches anymore before sending in a PR.

I am also a bit torn about it, but most features are worked on in a Draft PR, so that would mean the tests would still run.
(My main reasons are that it is annoying to see always two tests, double the run time and for environmental reasons)

> > It would be better to only listen on [pull_request](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) because by default it should trigger whenever any changes occur to a pull request because it has the default types of opened synchronize and reopened:
> 
> That would mean directly pushed code to existing stable branches isn't tested anymore?

I also think that would be the case, but I am not completely sure.

.github/workflows/static-analysis.yml
```diff
@@ -2,6 +2,9 @@ name: static analysis
 
 on:
   push:
+    branches:
+      - master
+      - '*.x'
   pull_request:
   schedule:
     - cron: '0 0 * * *'
```
driesvints, line 10: Btw, maybe also remove the cron check here. I don't think it's important to let that one run regularly.

Jubeki: I have removed the static analysis cron, but will leave this review open, so that other people can see it.

driesvints:
Thanks @Jubeki

PR #44057, opened 2022-09-09
[9.x] Fixed expectsOutputToContain & doesntExpectOutputToContain of PendingAction

sethsandaru:
All the details go here #43896 and resends #43916

Basically `expectsOutputToContain` and `doesntExpectOutputToContain` didn't assert against **all output** . 

So if you write an output that contains a lot of info, you can't assert it multiple times for the same output.

### Partial content from #43896

```php
@property int $id
@property string $first_name
```

Test result: 
```php
$command->artisan(...)
    ->execute();
```

```php
foreach ($lines as $line) {
    $this->info($line);
```

### DIFF
--- a/src/Illuminate/Testing/PendingCommand.php
+++ b/src/Illuminate/Testing/PendingCommand.php
@@ -67,6 +67,13 @@ class PendingCommand
     protected $hasExecuted = false;
 
     /**
+     * All outputs that printed out by the command.
+     *
+     * @var array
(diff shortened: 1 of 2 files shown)


sethsandaru:
Hi @taylorotwell , I applied the https://github.com/laravel/framework/pull/43916#issuecomment-1240790556 feedback here, could you review it again please? Thanks

PR #44058, opened 2022-09-09
use rpush directly instead of using luascript (to get thrown exception if there's an error on push)

edricwilliem:
this PR will throw an exception when there's error on `Queue::push` since `LuaScripts::push()` not throwing any error

```
public function pushRaw($payload, $queue = null, array $options = [])
{
        $this->getConnection()->eval(
            LuaScripts::push(), 2, $this->getQueue($queue),
            $this->getQueue($queue).':notify', $payload
        ); // not throwing any error on redis oom

        return json_decode($payload, true)['id'] ?? null;
}
```


taylorotwell:
Lua scripts are atomic. This isn't.

PR #44062, opened 2022-09-09
Update Builder.php

EG-Mohamed:
Add WhereLike function to short-cut Like query, I can't make test case can any one do it

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -298,7 +298,7 @@ class Builder implements BuilderContract
     
     public function whereLike($column, $value = null)
     {
-        $this->where($column,'Like','%'.$value.'%');
+        $this->where($column,'LIKE','%'.$value.'%');
 
         return $this;
     }
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -289,6 +289,21 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Add a basic where Like clause to the query, and return the All Like results.
+     *
+     * @param  \Closure|string|array|\Illuminate\Database\Query\Expression  $column
+     * @param  mixed  $value
+     * @return \Illuminate\Database\Eloquent\Model|static|null
+     */
+    
+    public function whereLike($column, $value = null)
+    {
+        $this->where($column,'Like','%'.$value.'%');
+
+        return $this;
+    }
+    
+    /**
      * Add a basic where clause to the query, and return the first result.
      *
      * @param  \Closure|string|array|\Illuminate\Database\Query\Expression  $column


src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -288,6 +288,21 @@ public function where($column, $operator = null, $value = null, $boolean = 'and'
...
+     * Add a basic where Like clause to the query, and return the All Like results.
+     *
+     * @param  \Closure|string|array|\Illuminate\Database\Query\Expression  $column
+     * @param  mixed  $value
+     * @return \Illuminate\Database\Eloquent\Model|static|null
+     */
+    
+    public function whereLike($column, $value = null)
+    {
+        $this->where($column,'Like','%'.$value.'%');
```
GrahamCampbell, line 301: This needs to escape the value.

GrahamCampbell (changes requested):

foremtehan:
What about other drivers.

rodrigopedra:
I guess this should be sent to master, as the query builder allows for dynamic where clauses, and someone could be using `->whereLike(...)` to filter a column named `like`

EGYWEB-Mohamed:
sorry am not expert enough, so if anyone has any update more then welcome I just thinking why not a method like this not exists yet so I decided to add it

src/Illuminate/Database/Eloquent/Builder.php
EGYWEB-Mohamed: any note how to !

kayw-geek, line 301: Using double% to match directly, isn't there a performance problem?

kayw-geek, line 301: In MySQL all the operator should be uppercase

EGYWEB-Mohamed: I don't think so, but let's wait for other opinion

src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -288,6 +288,21 @@ public function where($column, $operator = null, $value = null, $boolean = 'and'
...
+     * Add a basic where Like clause to the query, and return the All Like results.
+     *
+     * @param  \Closure|string|array|\Illuminate\Database\Query\Expression  $column
+     * @param  mixed  $value
+     * @return \Illuminate\Database\Eloquent\Model|static|null
+     */
+    
+    public function whereLike($column, $value = null)
+    {
+        $this->where($column,'LIKE','%'.$value.'%');
```
Jubeki, line 301:
```suggestion
        $this->where($column, 'LIKE', $value);
```

I would remove the `%` because sometimes you want to search only for `xxx%` or `%xxx`. I would expect that the `LIKE` Operator is set only and not that the value is changed.

EGYWEB-Mohamed: so let's figure a way to let us decide from the method argument

taylorotwell:

PR #44064, opened 2022-09-09
FIX Resource class name was not derived correctly from the Collection…

allanlaal:
… name

this only worked if the Resource class did not have a Resource suffix (like in the first class_exists check) However the code clearly intends to support the Resource suffix aswell

does not break existing usage! fixes expected functionality

### DIFF
--- a/src/Illuminate/Http/Resources/CollectsResources.php
+++ b/src/Illuminate/Http/Resources/CollectsResources.php
@@ -52,8 +52,8 @@ trait CollectsResources
         if ($this->collects) {
             $collects = $this->collects;
         } elseif (str_ends_with(class_basename($this), 'Collection') &&
-            (class_exists($class = Str::replaceLast('Collection', '', get_class($this))) ||
-             class_exists($class = Str::replaceLast('Collection', 'Resource', get_class($this))))) {
+            (@class_exists($class = Str::replaceLast('Collection', '', get_class($this))) ||
+             @class_exists($class = Str::replaceLast('Collection', 'Resource', get_class($this))))) {
             $collects = $class;
         }
 


src/Illuminate/Http/Resources/CollectsResources.php
```diff
@@ -52,8 +52,8 @@ protected function collects()
         if ($this->collects) {
             $collects = $this->collects;
         } elseif (str_ends_with(class_basename($this), 'Collection') &&
-            (class_exists($class = Str::replaceLast('Collection', '', get_class($this))) ||
-             class_exists($class = Str::replaceLast('Collection', 'Resource', get_class($this))))) {
+            (@class_exists($class = Str::replaceLast('Collection', '', get_class($this))) ||
+             @class_exists($class = Str::replaceLast('Collection', 'Resource', get_class($this))))) {
```
driesvints, line 56: We should not suppress errors like this.

taylorotwell:

Thanks!

src/Illuminate/Http/Resources/CollectsResources.php
allanlaal: without supression this throws an exception and never reaches the second replacement (Collection → Resource)

allanlaal: @taylorotwell class_exists() throws an error here, because its called with autoload.. and that fails if the class name is not a real name for a class. When you wrote this, you forgot to test for the second case ;)

this only works for `..\Resource\SomeCollection` using `..\Resource\Some`
and will fail when I am using `..\Resource\SomeResource`

allanlaal: @driesvints without suppression this throws an exception and never reaches the second replacement (Collection → Resource)

PR #44069, opened 2022-09-09
[8.x] Patch for timeless timing attack vulnerability in user login

JensJI:
## Timebox class
This new Timebox class makes a callable execute for at least the supplied amount of time.

This helps us guard against timing attacks at the application.

## Timeless timing attacks

The authentication method is currently vulnerable to user enumeration via timeless timing attacks.
This is caused by the early return inside the `hasValidCredentials` method in the `Illuminate\Auth\SessionGuard` class.

If the user does not exist most of the code in that method will not be called and thus the execution time will be a tiny bit shorter.

With traditional timing attacks this would not be practical to utilize because of the large sample sizes needed, but [timeless timing attacks](https://tom.vg/papers/timeless-timing-attack_usenix2020.pdf) which uses the HTTP/2 multiplexing protocol can with high accuracy measure timing differences between two calls to a remote server on 20 microseconds with a sample size of only 6 request pairs.

This means that most throttling/max attempts/DoS attack protection etc will not be triggered, and it is suddently very practical to harvest existing emails for a site (user enumeration).

User enumeration in itself is a security problem for some sites (where users dont want others to know they are using that site), but in general user enumeration can be used in tandem with other attacks (e.g. brute-forcing passwords or using previously leaked passwords).

## The patch

That is why the new Timebox class is also implemented inside the `hasValidCredentials` method in this PR.

A demo script that can be used to exploit the user enumeration can be [found here](https://github.com/ephort/laravel-user-enumeration-demo).

The changes in this PR add a minimum execution time for the `hasValidCredentials` method of 200 milliseconds.
But if the credentials are correct the timebox will be escaped and the user would not have to wait.

So this change only affects users typing the wrong credentials.

This pull request is opened with permission from Taylor via e-mail.

More in depth explanation of timeless timing attacks [can be found here](https://ephort.dk/blog/laravel-timing-attack-vulnerability/).


tests/Support/SupportHelpersTest.php
```diff
@@ -552,7 +552,7 @@ public function testRetry()
         $this->assertEquals(2, $attempts);
 
         // Make sure we waited 100ms for the first attempt
-        $this->assertEqualsWithDelta(0.1, microtime(true) - $startTime, 0.02);
+        $this->assertEqualsWithDelta(0.1, microtime(true) - $startTime, 0.03);
```
GrahamCampbell, line 555: microtime is not appropriate to use for timing. hrtime needs to be used, otherwise these tests will be flakey.

GrahamCampbell (changes requested):

src/Illuminate/Auth/SessionGuard.php
```diff
@@ -904,6 +919,16 @@ public function getSession()
         return $this->session;
     }
 
+    /**
+     * Get the timebox instance used by the guard.
+     *
+     * @return \Illuminate\Support\Timebox
+     */
+    public function getTimebox()
```
X-Coder264, line 927: It looks like this getter was added in the guard just because of tests. IMO this should be removed from this class and the test should be refactored so that it does not need this getter (changing "production" code just because of tests does not seem like a good idea).

src/Illuminate/Auth/SessionGuard.php
```diff
@@ -419,13 +430,17 @@ public function attemptWhen(array $credentials = [], $callbacks = null, $remembe
...
-        }
+            if ($validated) {
+                $timebox->returnEarly();
+
+                $this->fireValidatedEvent($user);
+            }
 
-        return $validated;
+            return $validated;
+        }, 200000);
```
rodrigopedra, line 443: what do you think about making this magic number configurable?

src/Illuminate/Auth/SessionGuard.php
```diff
@@ -88,6 +89,13 @@ class SessionGuard implements StatefulGuard, SupportsBasicAuth
      */
     protected $events;
 
+    /**
+     * The timebox implementation.
+     *
+     * @var \Illuminate\Support\Timebox
```
Jubeki, line 95:
```suggestion
     * @var \Illuminate\Support\Timebox|null
```
Timebox can be null (see constructor)

src/Illuminate/Auth/SessionGuard.php
```diff
@@ -419,13 +430,17 @@ public function attemptWhen(array $credentials = [], $callbacks = null, $remembe
      */
     protected function hasValidCredentials($user, $credentials)
     {
-        $validated = ! is_null($user) && $this->provider->validateCredentials($user, $credentials);
+        return $this->timebox->make(function ($timebox) use ($user, $credentials) {
```
Jubeki, line 433: Please add a check if the timebox is not `null` because the timebox can be null (see constructor)

Jubeki: There should probably also be a test if timebox is null and the expected behaviour

src/Illuminate/Auth/SessionGuard.php
Jubeki: It should be configurable, because some services (like AWS) will bill you for the time.

Also there is a concern of DoS because your service will block for a certain amount of time, which means some connections like cache, database and etc. will be kept open for the same amount of time which will increase the amount of concurrent connections open. (Some services have limitations for the amount of concurrent connections open)

src/Illuminate/Auth/AuthManager.php
```diff
@@ -122,7 +123,7 @@ public function createSessionDriver($name, $config)
     {
         $provider = $this->createUserProvider($config['provider'] ?? null);
 
-        $guard = new SessionGuard($name, $provider, $this->app['session.store']);
+        $guard = new SessionGuard($name, $provider, $this->app['session.store'], null, new Timebox);
```
Admiral-Enigma, line 126: Maybe use the 'timebox' named parameter and remove the null parameter for better readability

JensJI: Thanks for the suggestion! Unforturnatly security PR's must go to the 8.x branch which goes down to php7.3, so can't use that in this case.

src/Illuminate/Auth/SessionGuard.php
JensJI: Good idea. Could it be `auth.validate_credentials_minimum_time` or do you have better suggestions for an config key?

src/Illuminate/Auth/SessionGuard.php
JensJI: Thanks for the comment!

I was looking at the other tests and how that was done. The getSession() method which is also a getter on the SessionGuard class is also only used in the tests.

I somehow agree with you about creating code to only satisfy the tests. If there were checks in the production code that checked if you were running tests and then behaved differently I would also make an objection. But changing the production code to be more testable is for me one of the primary reasons for writing tests, so dont think that is a bad thing.

I will resolve the conversion now, but let me know if you disagree.

X-Coder264: You already know the timebox instance in the test as you are passing it through the constructor to the guard so I'm not sure how adding a getter makes this "more testable". I don't have a strong preference either way, but that's just my 2 cents.

src/Illuminate/Auth/SessionGuard.php
JensJI: I have placed the config inside the auth.guards.session.validate_credentials_minimum_time now.
The SessionGuard operates in that scope so thought it made sense to place it there. The default will still be 200000 microseconds.

src/Illuminate/Auth/SessionGuard.php
JensJI: I have now made the constructor fallback to setting the "timebox" property to an instance of the Timebox class if it was not provided in the constructor parameter.
It was done like that a couple of other places in the code.

Will resolve convo now, but let me know if you disagree.

src/Illuminate/Auth/SessionGuard.php
JensJI: See above comment. Now the property cannot be null.

driesvints:
I merged this into 9.x as well so this will be included in the next release.

lazerg:
Thank you for the PR.

mxp100:
May be correctly create this timeout in AuthManager and fix this attack for all drivers (not only Session) ?

pratikkuikel:
This PR is gold 🏅

PR #44074, opened 2022-09-09
[9.x] UUID and ULID support for Eloquent

driesvints:
This PR brings UUID & ULID support for Eloquent as primary keys. By simply applying a trait, any Eloquent model will have a UUID or ULID generated as their primary key.

For ULID support, an optional `symfony/uid` dependency will need to be pulled in.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasPrimaryUlid.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasPrimaryUlid.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Illuminate\Database\Eloquent\Concerns;
+
+use Illuminate\Support\Str;
+
+trait HasPrimaryUlid
+{
+    /**
+     * Generate a primary ULID for the model.
+     *
+     * @return void
+     */
+    public static function bootHasPrimaryUlid()
+    {
+        static::creating(function (self $model) {
+            if (empty($model->{$model->getKeyName()})) {
+                $model->{$model->getKeyName()} = $model->generatePrimaryUlid();
+            }
+        });
+    }
+
+    /**
+     * Generate the primary ULID key for the model.
+     *
+     * @return string
+     */
+    public function generatePrimaryUlid()
+    {
+        return (string) Str::ulid();
+    }
+
+    /**
+     * Get the auto-incrementing key type.
+     *
+     * @return string
+     */
+    public function getKeyType()
+    {
+        return 'string';
+    }
+
+    /**
+     * Get the value indicating whether the IDs are incrementing.
+     *
+     * @return bool
+     */
+    public function getIncrementing()
+    {
+        return false;
+    }
+}
--- a/src/Illuminate/Database/Eloquent/Concerns/HasUlidPrimaryKey.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasUlidPrimaryKey.php
@@ -15,7 +15,7 @@ trait HasUlidPrimaryKey
     {
         static::creating(function (self $model) {
             if (empty($model->{$model->getKeyName()})) {
-                $model->{$model->getKeyName()} = $model->generatePrimaryUlid();
+                $model->{$model->getKeyName()} = $model->generatePrimaryKey();
             }
         });
     }
@@ -25,7 +25,7 @@ trait HasUlidPrimaryKey
      *
      * @return string
      */
-    public function generatePrimaryUlid()
+    public function generatePrimaryKey()
     {
         return (string) Str::ulid();
     }
--- a/src/Illuminate/Database/Eloquent/Concerns/HasUlidPrimaryKey.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasUlidPrimaryKey.php
@@ -4,30 +4,30 @@ namespace Illuminate\Database\Eloquent\Concerns;
 
 use Illuminate\Support\Str;
 
-trait HasPrimaryUuid
+trait HasUlidPrimaryKey
 {
     /**
-     * Generate a primary UUID for the model.
+     * Generate a primary ULID for the model.
      *
      * @return void
      */
-    public static function bootHasPrimaryUuid()
+    public static function bootHasUlidPrimaryKey()
     {
         static::creating(function (self $model) {
             if (empty($model->{$model->getKeyName()})) {
-                $model->{$model->getKeyName()} = $model->generatePrimaryUuid();
+                $model->{$model->getKeyName()} = $model->generatePrimaryUlid();
             }
         });
     }
 
     /**
-     * Generate the primary UUID key for the model.
+     * Generate the primary ULID key for the model.
      *
      * @return string
      */
-    public function generatePrimaryUuid()
+    public function generatePrimaryUlid()
     {
-        return (string) Str::orderedUuid();
+        return (string) Str::ulid();
     }
 
     /**
--- a/src/Illuminate/Database/Eloquent/Concerns/HasUlids.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasUlids.php
@@ -4,30 +4,42 @@ namespace Illuminate\Database\Eloquent\Concerns;
 
 use Illuminate\Support\Str;
 
-trait HasUuidPrimaryKey
+trait HasUlids
 {
     /**
-     * Generate a primary UUID for the model.
+     * Boot the trait.
      *
      * @return void
      */
-    public static function bootHasUuidPrimaryKey()
+    public static function bootHasUlids()
     {
         static::creating(function (self $model) {
-            if (empty($model->{$model->getKeyName()})) {
-                $model->{$model->getKeyName()} = $model->generatePrimaryKey();
+            foreach ($model->uniqueIds() as $column) {
+                if (empty($model->{$column})) {
+                    $model->{$column} = $model->newUniqueId();
+                }
             }
         });
     }
 
     /**
-     * Generate the primary UUID key for the model.
+     * Generate a new ULID for the model.
      *
      * @return string
      */
-    public function generatePrimaryKey()
+    public function newUniqueId()
     {
-        return (string) Str::orderedUuid();
+        return strtolower((string) Str::ulid());
+    }
+
+    /**
+     * Get the columns that should receive a unique identifier.
+     *
+     * @return array
+     */
+    public function uniqueIds()
+    {
+        return [$this->getKeyName()];
     }
 
     /**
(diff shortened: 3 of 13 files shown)


X-Coder264:
@driesvints Awesome PR - I was just thinking about using ULIDs in my next project.

Yesterday Symfony added UUID v7 (which is similar to ULID) support (https://github.com/symfony/symfony/pull/47525) to `symfony/uid`. Would it be possible to add an option to choose whether we want UUID v4 or v7 (as currently `Str::orderedUuid()` always returns a v4 UUID) ?

tests/Integration/Database/EloquentPrimaryUlidTest.php
```diff
@@ -0,0 +1,36 @@
...
+{
+    protected function defineDatabaseMigrationsAfterDatabaseRefreshed()
+    {
+        Schema::create('users', function (Blueprint $table) {
+            $table->char('id', 26)->primary();
+            $table->timestamps();
+        });
+    }
+
+    public function testUserWithPrimaryUuidCanBeCreated()
```
siarheipashkevich, line 21: please change to `Ulid`

tests/Integration/Database/EloquentPrimaryUuidTest.php
```diff
@@ -0,0 +1,36 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Support\Str;
+
+class EloquentPrimaryUuidTest extends DatabaseTestCase
+{
+    protected function defineDatabaseMigrationsAfterDatabaseRefreshed()
+    {
+        Schema::create('users', function (Blueprint $table) {
+            $table->uuid('id')->primary();
```
siarheipashkevich, line 16: what about different name of the primary key? for example `uuid` then we should override `$primaryKey` property in the model to this name?

driesvints: This is just an example. You can still adjust the primary key name to whatever you want.

siarheipashkevich: If I use different name, then I'll add `HasPrimaryUuuid` and override `$primaryKey` property for model, yes?

SjorsO:
`HasUlidPrimaryKey` and `HasUuidPrimaryKey` might be better names than `HasPrimaryUlid` and `HasPrimaryUuid`.

src/Illuminate/Database/Eloquent/Concerns/HasPrimaryUlid.php
```diff
@@ -0,0 +1,52 @@
...
+            }
+        });
+    }
+
+    /**
+     * Generate the primary ULID key for the model.
+     *
+     * @return string
+     */
+    public function generatePrimaryUlid()
```
BrandonSurowiec, line 28: I wonder if `generatePrimaryUlid()` and `generatePrimaryUuid()` should be generalized and renamed to both be `generatePrimaryKey()`. If accessing these methods manually, it will remove some friction if different models are using different variations of the uuid.

tests/Integration/Database/EloquentPrimaryUlidTest.php
driesvints:
```suggestion
    public function testUserWithPrimaryUlidCanBeCreated()
```

driesvints: Thanks!

tests/Integration/Database/EloquentPrimaryUuidTest.php
driesvints: Yep.

driesvints:
> Yesterday Symfony added UUID v7 (which is similar to ULID) support (https://github.com/symfony/symfony/pull/47525) to symfony/uid. Would it be possible to add an option to choose whether we want UUID v4 or v7 (as currently Str::orderedUuid() always returns a v4 UUID) ?

I'm not sure. We don't use `symfony/uid` for UUID's but `ramsey/uuid`. It does not seem that library has support for v7 yet. It also does not seem `Str::orderedUuid()` has support for specifying the type of UUID so I don't think this is possible?

src/Illuminate/Database/Eloquent/Concerns/HasPrimaryUlid.php
driesvints: Done.

driesvints:
> `HasUlidPrimaryKey` and `HasUuidPrimaryKey` might be better names than `HasPrimaryUlid` and `HasPrimaryUuid`.

Done

src/Illuminate/Database/Eloquent/Concerns/HasUlidPrimaryKey.php
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /**
+     * Generate the primary ULID key for the model.
+     *
+     * @return string
+     */
+    public function generatePrimaryKey()
+    {
+        return (string) Str::ulid();
```
driesvints, line 30: I didn't use `Str::orderedUlid()` as I'm not even sure that's possible with ULID?

mabdullahsari: [ULID : Universally Unique Lexicographically Sortable Identifiers.](https://github.com/ulid/spec)

The abbreviation itself already suggests that it is ordered / sortable. 👍

src/Illuminate/Database/Eloquent/Concerns/HasUlidPrimaryKey.php
```diff
@@ -0,0 +1,52 @@
...
+{
+    /**
+     * Generate a primary ULID for the model.
+     *
+     * @return void
+     */
+    public static function bootHasUlidPrimaryKey()
+    {
+        static::creating(function (self $model) {
+            if (empty($model->{$model->getKeyName()})) {
```
DarkGhostHunter: Shouldn't be `blank($model->getKey())` ?

DarkGhostHunter:
Just came here to say that ULID will be saved into the database as a 26 character string, which occupies 208 bits, almost double than an UUID (128 bit).

If the column is an UUID, most database engines will return an error.

What it could work is setting the key type to "uuid", so then the primary key is detected as an ULID, it will change to RFC4122.

Jubeki:
> > Yesterday Symfony added UUID v7 (which is similar to ULID) support ([symfony/symfony#47525](https://github.com/symfony/symfony/pull/47525)) to symfony/uid. Would it be possible to add an option to choose whether we want UUID v4 or v7 (as currently Str::orderedUuid() always returns a v4 UUID) ?
> 
> I'm not sure. We don't use `symfony/uid` for UUID's but `ramsey/uuid`. It does not seem that library has support for v7 yet. It also does not seem `Str::orderedUuid()` has support for specifying the type of UUID so I don't think this is possible?

It seems like `ramsey/uuid` now has support for v7: https://github.com/ramsey/uuid/releases/tag/4.5.0

driesvints:
I'll look into it 👍

larskoole:
Maybe creating a `ulid()` macro for the migrations would be nice too.

Currently I'm adding the following macro to the AppServiceProvicer:
```PHP
Blueprint::macro('ulid', function (string $column = 'ulid') {
    return $this->addColumn('char', $column, ['length' => 26]);
});
```
Which enables me to do this in my migrations:
```PHP
    $table->ulid('id')->primary();
    ...
});
```

tomb1n0:
Awesome! Just this week i'd created a trait that used `Str::orderedUuid()` exactly like this. Great to have this officially supported inside Laravel

driesvints:
@larskoole I'll also check into that.

mfmarques:
@driesvints  Could we have an option to set which column to be the uuid? 

For example, we could keep the default behaviour of looking at the primary key but it would be nice to have something like `getUuidColumn` function or something like that inside each Model to override that default behaviour.

Potential usage:

table: `users`
`id` (auto increment to be used for relations) | `uuid`

larskoole:
> @driesvints Could we have an option to set which column to be the uuid?
> 
> For example, we could keep the default behaviour of looking at the primary key but it would be nice to have something like `getUuidColumn` function or something like that inside each Model to override that default behaviour.
> 
> Potential usage:
> 
> table: `users` `id` (auto increment to be used for relations) | `uuid`

Already there, you can override the `uniqueIds()` method from the trait inside of your model.

https://github.com/laravel/framework/pull/44074/files#diff-0dbc036e590312e780e3fac6936e57aa9628b31679dd979e098cd73fc34d85cfR35-R43

BrandonSurowiec:
If someone wanted to use this for columns other than the primary `id` and updates `uniqueIds()` to not contain the primary key, the `getKeyType()` and `getIncrementing()` shouldn't be overriding. We'll need to PR a fix for that.

driesvints:
> If someone wanted to use this for columns other than the primary `id` and updates `uniqueIds()` to not contain the primary key, the `getKeyType()` and `getIncrementing()` shouldn't be overriding `$this->keyType` and `$this->incrementing`. We'll need to PR a fix for that.

I'll think about that a bit more as well.

driesvints:
Worked on some of the above stuff here: https://github.com/laravel/framework/pull/44146. I'll check into UUID v7 support later.

src/Illuminate/Support/composer.json
```diff
@@ -46,6 +46,7 @@
         "league/commonmark": "Required to use Str::markdown() and Stringable::markdown() (^2.0.2).",
         "ramsey/uuid": "Required to use Str::uuid() (^4.2.2).",
         "symfony/process": "Required to use the composer class (^6.0).",
+        "symfony/uid": "Required to generate ULIDs for Eloquent (^6.0).",
```
tanthammar, line 49: This is misspelled. Should be symfony/uuid else pkg must be manually installed

martio:
@driesvints @taylorotwell I added missing morphs methods for ULID support :point_right: PR [44364](https://github.com/laravel/framework/pull/44364) :blush:

src/Illuminate/Database/Eloquent/Concerns/HasUlids.php
```diff
@@ -0,0 +1,64 @@
+<?php
```
mikebrandl, line 1: Is this now built in functionality as of 9.30? 

https://github.com/laravel/framework/pull/44074

src/Illuminate/Database/Eloquent/Concerns/HasUuids.php
```diff
@@ -0,0 +1,64 @@
+<?php
```
mikebrandl, line 1: Again, now built in? 
https://github.com/laravel/framework/pull/44074

PR #44078, opened 2022-09-10
[9.x] Add getContainer method to the Route class

iamgergo:
This PR allows to have access to the previously set container instance on a route. It can be useful when doing some custom route model binding for example.

```php
App::make('router')->bind('post', function ($value, $route) {
    if ($route->getContainer()->make(CustomValueBinder::class)->passes($route, $value)) {
        // doing some custom resolution...
    }

    return (new Post())->resolveRouteBinding($value);
});
```

### DIFF
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -1267,6 +1267,16 @@ class Route
     }
 
     /**
+     * Get the container instance.
+     *
+     * @return \Illuminate\Container\Container|null
+     */
+    public function getContainer()
+    {
+        return $this->container;
+    }
+
+    /**
      * Prepare the route instance for serialization.
      *
      * @return void


iamgergo:
I don't think the failing tests are related to the PR.

X-Coder264:
Not sure if I'm missing something but the logic for customizing the model binding resolution logic should be placed in the `RouteServiceProvider` class as per the documentation: https://laravel.com/docs/9.x/routing#customizing-the-resolution-logic

There you already have access to the container via `$this->app`.

iamgergo:
Yes, most of the cases that is the place for it. But recently I faced some situations, where I had to put it somewhere else, so I found useful to have access to the container.

So, I still prefer using the RouteServiceProvider, but I think it's nice to have a bit more flexibility if needed.

iamgergo:
@driesvints I don't know why the tests are failing. Could you make a rerun for me, please? Thank you!

driesvints:
Re-triggered them

taylorotwell:
Just use the App facade?

PR #44079, opened 2022-09-10
[9.x] Add read-only filesystem adapter decoration as a config option

frankdejonge:
This PR adds the ability to configure a disk to operate in `read-only` mode. When this option is set to `true`, the underlying adapter is decorated with an implementation that ensures no write operation is possible on the disk. This is useful when you're receiving files through a shared storage, but you shouldn't manipulate any of the files stored.

If interested, I can also add a PR to include a path prefixing decorator: https://flysystem.thephpleague.com/docs/adapter/path-prefixing/

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -151,6 +151,7 @@
         "laravel/tinker": "Required to use the tinker console command (^2.0).",
         "league/flysystem-aws-s3-v3": "Required to use the Flysystem S3 driver (^3.0).",
         "league/flysystem-ftp": "Required to use the Flysystem FTP driver (^3.0).",
+        "league/flysystem-read-only": "Required to use read-only disks (^3.3)",
         "league/flysystem-sftp-v3": "Required to use the Flysystem SFTP driver (^3.0).",
         "mockery/mockery": "Required to use mocking (^1.4.4).",
         "nyholm/psr7": "Required to use PSR-7 bridging features (^1.2).",
--- a/composer.json
+++ b/composer.json
@@ -87,6 +87,7 @@
         "guzzlehttp/guzzle": "^7.2",
         "league/flysystem-aws-s3-v3": "^3.0",
         "league/flysystem-ftp": "^3.0",
+        "league/flysystem-read-only": "^3.3",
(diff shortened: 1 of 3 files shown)


tests/Filesystem/FilesystemManagerTest.php
```diff
@@ -2,11 +2,16 @@
...
+use function fclose;
+use function file_put_contents;
+use function fopen;
+use function fwrite;
 use Illuminate\Contracts\Filesystem\Filesystem;
 use Illuminate\Filesystem\FilesystemManager;
 use Illuminate\Foundation\Application;
 use InvalidArgumentException;
 use PHPUnit\Framework\TestCase;
+use function unlink;
```
driesvints, line 14: No need to import functions like this.

driesvints (changes requested):

tests/Filesystem/FilesystemManagerTest.php
frankdejonge: I've removed the function imports.

driesvints (approved):

composer.json
```diff
@@ -87,6 +87,7 @@
         "guzzlehttp/guzzle": "^7.2",
         "league/flysystem-aws-s3-v3": "^3.0",
         "league/flysystem-ftp": "^3.0",
+        "league/flysystem-read-only": "^3.3",
```
ziadoz, line 90: Does this need adding to the `suggest` block?

frankdejonge: I've added the suggests entry.

taylorotwell:
@frankdejonge is there any need for us to pass any of these options like `temporary-url`?

CleanShot 2022-09-12 at 09 01 37@2x

driesvints:
@taylorotwell yes that part should remain and doesn't have anything to do with this PR.

driesvints:
@taylorotwell those options are still passed as the new adapter is just a decorator.

taylorotwell:
Ah, sorry - read the code wrong.

PR #44080, opened 2022-09-10
[9.x] Passing event into `viaQueue` and `viaConnection` of Queued Listener

sethsandaru:
Proposal this (if got approved & merged, I'll submit a doc PR)

For our project, we have a lot of Queued Listeners. At the moment, for some critical endpoints (that need to respond as fast as they can), we need to push it to a `redis` connection instead of `sqs` based on the events's data (which is impossible at the moment)

At my first try (which didn't work), I used the `shouldQueue($event)` to set the `$connection` dynamically.

After spending a bit of time reading the logic. I see that before Laravel invoking the `shouldQueue`, it will create a new concrete listener class to and use it to check. 

And before the queue push happen, Laravel create another new concrete listener class via Reflector.  Thus my `$connection` will never be there.

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Events/Dispatcher.php#L560

With this patch, we would be able to do something like this:

```php
public function viaConnection($event): string
{
    if ($event->useRedisQueue) {
        return 'redis';
    } 

    return config('queue.default');
}
```

Code & unit testing added, no breaking changes!

### DIFF
--- a/src/Illuminate/Events/Dispatcher.php
+++ b/src/Illuminate/Events/Dispatcher.php
@@ -579,11 +579,11 @@ class Dispatcher implements DispatcherContract
         [$listener, $job] = $this->createListenerAndJob($class, $method, $arguments);
 
         $connection = $this->resolveQueue()->connection(method_exists($listener, 'viaConnection')
-                    ? $listener->viaConnection()
+                    ? $listener->viaConnection($arguments[0])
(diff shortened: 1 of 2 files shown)


sethsandaru:
Hi @taylorotwell @driesvints could you please review to see if this one makes sense? 

Thanks

PR #44083, opened 2022-09-10
[9.x] Fix QueryBuilder whereNot with array conditions 

Propaganistas:
```php
$query->whereNot([
    'foo' => 1, 
    'bar' => 2
]);
```

currently yields `not ( not foo = 1 and not bar = 2 )`
while I would expect this to implictly become a nested clause `not ( foo = 1 and bar = 2 )`.

The commits in this PR are deliberately threefold:

1. the actual bugfix + tests
2. the consequence of reusing `whereNested`: aesthetically superfluous parentheses (but sql doesn't care).
3. a missing variable interpolation I encountered on the go

---
Note: the PR only tackles `Query\Builder`. The same principle probably applies for `Eloquent\Builder`

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -800,7 +800,7 @@ class Builder implements BuilderContract
                 if (is_numeric($key) && is_array($value)) {
                     $query->{$method}(...array_values($value));
                 } else {
-                    $query->$method($key, '=', $value, $boolean);
+                    $query->{$method}($key, '=', $value, $boolean);
                 }
             }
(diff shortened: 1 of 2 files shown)


BrandonSurowiec:
Since it has been like this for a long time this seems like a breaking change. Should probably target 10.x as someone out there is depending on the current behavior.

rodrigopedra:
`whereNot` was introduced earlier this year by PR #41096.

I agree with @Propaganistas the expected behavior when passing an array should not be to negate it twice.

taylorotwell:
Made the change only apply to arrays.

PR #44084, opened 2022-09-10
Adds the ability for generating a resource route

mahmoudmohamedramadan:
This PR allows us to generate a resource route for the created resource controller class

## After this PR gets merged

```SHELL
php artisan make:controller UserController -r --route
```

As we notice, there is a new option that has been added is `--route`, and after that command gets executed, the `Route::resource('users', App\Http\Controllers\UserController::class);` will be appended into the `web.php` routes file.


taylorotwell:

PR #44096, opened 2022-09-12
[9.x] Support preloading assets with Vite

timacdonald:
Preloading is a mechanism that informs the browser which assets are required for the current page load. You can read more about [preloading on web.dev](https://web.dev/preload-critical-assets/).

Using this PR on a real-world application increased the Lighthouse score from `59` to `90`. These were the average results from a controlled environment. Your mileage may vary, but this PR should generally work for the 95% usecase (IMO).

Preloading is not something to opt into. With this PR it becomes the default behaviour for applications using Vite, however the middleware does need to be applied manually if you want to have the preloading headers sent with the initial Laravel HTTP response.

I'll outline the problem we are looking to solve: for SPA applications building with Vite, or any modern build tool using code splitting, the resulting assets look similar to the following:

- `app.123.js`
- `Welcome.123.js`
- `Login.123.js`
- `Register.123.js`
- ...

Each of these top-level "pages" will often have imports:

- `Login.123.js`
  - `TextInput.123.js`
  - `InputLabel.123.js`
  - `PrimaryButton.123.js`

"Code splitting" like this is a performance optimisation so you load only the JavaScript required for the current page, which makes the initial page load faster.

Currently in Laravel, when loading the very first page of an application, we specify the following entry point:

```blade
@vite(['resources/js/app.js'])
```

This results in the following _waterfall_ of requests:

1. `/login.html`...
2. `app.js`...
3. `Login.js`, `TextInput.js`, `InputLabel.js`, `PrimaryButton.js`...

This waterfall of requests is, of course, less than ideal. Working through this waterfall:

1. The browser must first download the HTML and parse it to know what CSS and JavaScript the document has specified. Our document has specified "app.js", so the browser proceeds to fetch it.
2. Once `app.js` js fetched and parsed, the JavaScript kicks in an informs the browser that it should be loading the `Login.js` file and it's imports.
3. The browser begins fetching, parsing, and executing the required imports.

Possible solutions are:

1. Bundle everything into a single `app.js` with "eager globing"
4. Preloading

Before I dig into these solutions, let me post the results of my Lighthouse performance testing against a real-world application. These tests were run against "mobile".

I acknowledge that this is a single application and need further verification.

| With Preloading | Configuration | Lighthouse score |
|-----------------|---------------|------------------|
| YES | Code split + vendor split | 90 |
| YES | Code split | 90 |
| NO | Code split + vendor split | 85 |
| NO | Code split | 85 |
| YES | Eager glob + vendor split | 65 |
| NO | Eager glob + vendor split | 63 |
| YES | Eager glob | 61 |
| NO | Eager glob | 59 |

We can see in all instances, preloading had a positive impact on the Lighthouse scores.

<details>

<summary>Lighthouse screenshots</summary>

Although all these screenshots are of the same page, I found similar results across other pages with the application.

## Preloading w/ Code split + vendor split

1  Code split + vendor split w: Preloading

## Preloading w/ Code split

2  Code split w: Preloading

## Not Preloading w/ Code split + vendor split

3  Code split + vendor split

## Not Preloading w/ Code split

4  Code split

## Preloading w/ Eager glob + vendor split

5  Eager glob + vendor split w: Preloading

## Preloading w/ Eager

6  Eager glob w: Preloading

## Not preloading w/ Eager glob + vendor split

7  Eager glob + vendor split

## Not preloading w/ Eager glob

8  Eager glob

</details>

Although eager globing does solve the waterfall issue, I believe that the Vite setup for Laravel should also support code splitting the best it can. Preloading is utilised by Vite itself when doing things "the Vite way". This is the mechanism that all code splitting usually works. Additionally, if you utilise vendor splitting with eager globing, you are back to the original waterfall problem, as the `app.js` has to be fetched and parsed before `vendor.js` is requested and parsed.

With this PR in place, we are able to, at best, remove the waterfall entirely for all configurations:

1. `/login.html`, `app.js`, `Login.js`, `TextInput.js`, `InputLabel.js`, `PrimaryButton.js`...

and at worst, the non-SPA waterfall:

1. `/login.html`
2. `app.js`, `Login.js`, `TextInput.js`, `InputLabel.js`, `PrimaryButton.js`...

This PR works by adding "preload" `<link` tags into the HTML for any asset required by the current page and if used the middleware will send preload `Link: ` headers to the browser.

Preload tags are generated for the JavaScript and CSS entry points and their imports.

To take full advantage of this, applications do need to specify the page / chunk they are currently trying to load in the Vite tag:

```blade
@vite(["resources/js/Pages/{$page['component']}.vue"])
```

Applications do not have to update the Vite directive may still find this PR beneficial, as it can send the headers with the document instructing the browser to start fetching the `app.js` and any additional CSS it imports.

With that in mind, preloading is beneficial to all configurations: eager glob, vendor split, code split, etc.

> **Note** Applications that are vendor splitting will still need to add the `app.js` into the Vite directive:

```blade
@vite([
    'resources/js/app.js',
    "resources/js/Pages/{$page['component']}.vue",
])
```

## Browser support

For CSS files, we have [great browser support](https://caniuse.com/link-rel-preload).

For JavaScript "Module preloading" unfortunately out of the main browsers Firefox and Safari do not currently support this feature. According to [caniuse](https://caniuse.com/link-rel-modulepreload) 73% of browsers in use globally support this feature. Their stats are provided by [statscounter](https://gs.statcounter.com/).

A lot of JavaScript eco-system is relying on this feature and we are _all_ waiting for Firefox and Safari to roll the feature out.

The good news is that this feature has no negative impacts on un-supprted browsers. They simply ignore the tag and move on with their existing behaviour.

However, because CSS preloading is supported, applications will still likely get a boost in Firefox and Safari.

### DIFF
--- a/src/Illuminate/Foundation/Vite.php
+++ b/src/Illuminate/Foundation/Vite.php
@@ -360,7 +360,7 @@ class Vite implements Htmlable
     }
 
     /**
-     * Make preload tag for the given chunk.
+     * Make a preload tag for the given chunk.
      *
      * @param  string  $src
      * @param  string  $url
--- a/src/Illuminate/Foundation/Vite.php
+++ b/src/Illuminate/Foundation/Vite.php
@@ -63,11 +63,35 @@ class Vite implements Htmlable
     protected $styleTagAttributesResolvers = [];
 
     /**
+     * The preload tag attributes resolvers.
+     *
+     * @var array
+     */
+    protected $preloadTagAttributesResolvers = [];
+
+    /**
+     * The preloaded assets.
+     *
+     * @var array
+     */
+    protected $preloadedAssets = [];
+
+    /**
      * The cached manifest files.
      *
      * @var array
      */
     protected static $manifests = [];
+
+    /**
+     * Get the preloaded assets.
+     *
+     * @var array
+     */
+    public function preloadedAssets()
+    {
+        return $this->preloadedAssets;
+    }
 
     /**
      * Get the Content Security Policy nonce applied to all generated tags.
@@ -187,6 +211,23 @@ class Vite implements Htmlable
     }
 
     /**
+     * Use the given callback to resolve attributes for preload tags.
+     *
+     * @param  (callable(string, string, ?array, ?array): array)|array  $attributes
+     * @return $this
+     */
+    public function usePreloadTagAttributes($attributes)
+    {
+        if (! is_callable($attributes)) {
+            $attributes = fn () => $attributes;
+        }
+
+        $this->preloadTagAttributesResolvers[] = $attributes;
+
+        return $this;
+    }
+
+    /**
      * Generate Vite tags for an entrypoint.
      *
      * @param  string|string[]  $entrypoints
@@ -212,13 +253,35 @@ class Vite implements Htmlable
         $manifest = $this->manifest($buildDirectory);
 
         $tags = collect();
+        $preloads = collect();
 
         foreach ($entrypoints as $entrypoint) {
             $chunk = $this->chunk($manifest, $entrypoint);
 
+            $preloads->push($this->makePreloadTagForChunk(
+                $chunk['src'],
+                $this->assetPath("{$buildDirectory}/{$chunk['file']}"),
+                $chunk,
+                $manifest
+            ));
+
             foreach ($chunk['imports'] ?? [] as $import) {
+                $preloads->push($this->makePreloadTagForChunk(
+                    $import,
+                    $this->assetPath("{$buildDirectory}/{$manifest[$import]['file']}"),
+                    $manifest[$import],
+                    $manifest
+                ));
+
                 foreach ($manifest[$import]['css'] ?? [] as $css) {
                     $partialManifest = Collection::make($manifest)->where('file', $css);
+
+                    $preloads->push($this->makePreloadTagForChunk(
+                        $partialManifest->keys()->first(),
+                        $this->assetPath("{$buildDirectory}/{$css}"),
+                        $partialManifest->first(),
+                        $manifest
+                    ));
 
                     $tags->push($this->makeTagForChunk(
                         $partialManifest->keys()->first(),
@@ -239,6 +302,13 @@ class Vite implements Htmlable
             foreach ($chunk['css'] ?? [] as $css) {
                 $partialManifest = Collection::make($manifest)->where('file', $css);
 
+                $preloads->push($this->makePreloadTagForChunk(
+                    $partialManifest->keys()->first(),
+                    $this->assetPath("{$buildDirectory}/{$css}"),
+                    $partialManifest->first(),
+                    $manifest
+                ));
+
                 $tags->push($this->makeTagForChunk(
                     $partialManifest->keys()->first(),
                     $this->assetPath("{$buildDirectory}/{$css}"),
@@ -250,7 +320,9 @@ class Vite implements Htmlable
 
         [$stylesheets, $scripts] = $tags->partition(fn ($tag) => str_starts_with($tag, '<link'));
 
-        return new HtmlString($stylesheets->join('').$scripts->join(''));
+        $preloads = $preloads->sortByDesc(fn ($link) => str_contains($link, 'rel="preload"'));
+
+        return new HtmlString($preloads->join('').$stylesheets->join('').$scripts->join(''));
     }
 
     /**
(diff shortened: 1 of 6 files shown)


taylorotwell:
@timacdonald 

How would you picture doing this in an Inertia app? Do you put this in the main layout? Having trouble picturing it.

```php
@vite(["resources/js/Pages/{$page['component']}.vue"])
```

timacdonald:
That is correct. I would suggest we make that the default for Inertia across the started kits.

Take a fresh Breeze application for example:

## Before

**LAYOUT**

```blade
@vite('resources/js/app.js')
```

**HEADERS**

```
n/a
```

**HTML**

```html
<link rel="stylesheet" href="http://127.0.0.1:8000/build/assets/app.f31a18ec.css" />
<script type="module" src="http://127.0.0.1:8000/build/assets/app.9bf14e2c.js"></script>
```

**Waterfall**

1. login.html
2. app.js, app.css
3. Login.js, PrimaryButton.js, TextInput.js, ApplicationLogo.js


## After

**LAYOUT**

```
@vite("resources/js/Pages/{$page['component']}.vue")
```

**HEADERS**

```
Link: <http://127.0.0.1:8000/build/assets/app.f31a18ec.css>; rel="preload"; as="style", 
      <http://127.0.0.1:8000/build/assets/Login.5e7e8a19.js>; rel="modulepreload", 
      <http://127.0.0.1:8000/build/assets/app.9bf14e2c.js>; rel="modulepreload", 
      <http://127.0.0.1:8000/build/assets/PrimaryButton.62d440f3.js>; rel="modulepreload", 
      <http://127.0.0.1:8000/build/assets/TextInput.d97ddb4c.js>; rel="modulepreload", 
      <http://127.0.0.1:8000/build/assets/ApplicationLogo.fc364aed.js>; rel="modulepreload"
```

**HTML**

```html
<link rel="preload" as="style" href="http://127.0.0.1:8000/build/assets/app.f31a18ec.css" />
<link rel="modulepreload" as="script" href="http://127.0.0.1:8000/build/assets/Login.5e7e8a19.js" />
<link rel="modulepreload" as="script" href="http://127.0.0.1:8000/build/assets/app.9bf14e2c.js" />
<link rel="modulepreload" as="script" href="http://127.0.0.1:8000/build/assets/PrimaryButton.62d440f3.js" />
<link rel="modulepreload" as="script" href="http://127.0.0.1:8000/build/assets/TextInput.d97ddb4c.js" />
<link rel="modulepreload" as="script" href="http://127.0.0.1:8000/build/assets/ApplicationLogo.fc364aed.js" />
<link rel="stylesheet" href="http://127.0.0.1:8000/build/assets/app.f31a18ec.css" />
<script type="module" src="http://127.0.0.1:8000/build/assets/Login.5e7e8a19.js"></script> 
```

**Waterfall**

1. login.html
2. app.js, app.css, Login.js, PrimaryButton.js, TextInput.js, ApplicationLogo.js

However with the headers, the browser is able to prioritise and connect etc. in advance of parsing the DOM.

taylorotwell:
In your after example. How do we actually get `$page['component]` to the layout? And doesn't that page component change in between page navigations? Is the entire layout re-rendered on those navigations?

timacdonald:
> In your after example. How do we actually get $page['component] to the layout?

`$page['component]` is injected by Inertia for us. Don't need to do anything to make it work.

I believe it is handled here: https://github.com/inertiajs/inertia-laravel/blob/540b953ec383364264f9bd633849db16560a4461/src/Response.php#L99-L110

> And doesn't that page component change in between page navigations?

It does, but future navigations are done in SPA land and the blade layout is not re-rendered.

```blade
@php
    logger()->info('rendering blade');
@endphp
```

You only get one log statement on initial load. Future navigations do not log.

Additionally, what I am proposing we do on initial load is what already happening for us on subsequent navigations, if the components required for the page have not already be preloaded before...

https://user-images.githubusercontent.com/24803032/195722527-270e34ff-03c1-40b2-8fbe-e5a568da10c0.mov

If we are keen on this in general, I would also like to propose that we optimize _future navigations_ as well - but I feel that is fine for a different PR, which is why it isn't included here. The full story I have in my head is that we remove the waterfall from the initial load, but then in the background we offer the option to preload all (or some) of the sites assets, which will mean that even on future navigations, the only request to the server is the Inertia request.

timacdonald:
Just some more info on this one:  these link headers are what would power the recent “Early Hints” feature in some caches.

Cloudflare’s implementation is based on these headers.

2F5C1E1C-CBDC-40B9-A74D-D15396203835

timacdonald:
- Jetstream PR: https://github.com/laravel/jetstream/pull/1169
- Breeze PR: https://github.com/laravel/breeze/pull/196

driesvints:

lk77:
Is there any way to disable this ?

it preloads every thing, even modules that 99% of users will most likely never need

dennisprudlo:
@lk77 I also get warnings in the console that scripts were preloaded that haven't been used later on. I was able to disable it using the following line in a service provider's boot method:

ˋˋˋphp
Vite::usePreloadTagAttributes(fn () => false);
ˋˋˋ

timacdonald:
If you can provide a reproduction repository that shows assets being preloaded that are not used by the current page, I'm happy to look into this for you.

I'm unable to replicate the issue myself.

lk77:
@timacdonald i don't see how you can know in advance if the asset will be needed or not, for me it's more business logic than anything.

PR #44098, opened 2022-09-12
[9.x] Improve Artisan Commands Readability

AamirSohailKmAs:
Hi developers

I also love the new artisan look like you (Special Thanks @nunomaduro ), but I was having readability issues, I have to track the dots from one end to another, This PR will improve a little bit by shuffling between dots and dashes. here is an example of `about` command 

### Before

before

### After

after

I tried to use different symbols and also more than two symbols, but they reduce the beauty instead of increasing readability 😉

### DIFF
--- a/src/Illuminate/Console/View/Components/Task.php
+++ b/src/Illuminate/Console/View/Components/Task.php
@@ -16,6 +16,13 @@ class Task extends Component
     private static $counter = 0;
 
     /**
+     * The list of separators
+     *
+     * @var array $separators
+     */
+    private $separators = ['.', '-'];
+
+    /**
      * Renders the component using the given arguments.
      *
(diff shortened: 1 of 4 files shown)


driesvints:
I personally don't think this is an improvement over how it was before, sorry.

AamirSohailKmAs:
Okay, maybe more readable.

nunomaduro:
Sorry, I also don't think this is an improvement.

nunomaduro:
Ups. Re-opening, so we can have Taylor's opinion.

AamirSohailKmAs:
Okay sure, maybe I was having the issue with reading,

with-equal

taylorotwell:
Not really feeling this one.

PR #44101, opened 2022-09-12
[9.x] Enhance column modifying

hafezdivandari:
This PR does:
- Add support for changing column type to `double`
- Add support for changing column type to `tinyText`.
- Fix a issue when changing column type to `timestamp`:
  * The SQL contained charset/collate: `TIMESTAMP SET utf8mb4 COLLATE utf8mb4_unicode_ci`
- ~~Fix a issue when changing column type to `binary`:~~
  * ~~on MySQL, the type was `LONGBLOB` instead of expected `BLOB`~~

PS: I added tests for all adds and changes.
PS: changes to `binary` reverted.

### DIFF
--- a/src/Illuminate/Database/DBAL/TimestampType.php
+++ b/src/Illuminate/Database/DBAL/TimestampType.php
@@ -41,12 +41,6 @@ class TimestampType extends Type implements PhpDateTimeMappingType
             $columnType = 'TIMESTAMP('.$fieldDeclaration['precision'].')';
         }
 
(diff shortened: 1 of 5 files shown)


driesvints:
Won't this change the behavior for apps who have already run their migrations in production?

driesvints:
Their migrations which are part of their app will still be run in development. If they run them in development after this PR is merged, their development databases will be out if sync with production.

hafezdivandari:
OK, I reverted changes on `binary` I'll send that to master later.

PR #44104, opened 2022-09-12
[9.x] Fake Batches

taylorotwell:
It's currently hard to test things like if a batch was cancelled by a job or if a job added additional jobs to a batch. You have to create a FakeBatch manually and override the cancel / add methods, etc.

This solves that.

```php
[$job, $batch] = (new TestJob)->withFakeBatch();

$job->handle();

$this->assertTrue($batch->cancelled());
$this->assertNotEmpty($batch->added);
```

### DIFF
--- a/src/Illuminate/Bus/Batchable.php
+++ b/src/Illuminate/Bus/Batchable.php
@@ -84,7 +84,9 @@ trait Batchable
                                   int $failedJobs = 0,
                                   array $failedJobIds = [],
                                   array $options = [],
-                                  CarbonImmutable $createdAt = null)
+                                  CarbonImmutable $createdAt = null,
+                                  ?CarbonImmutable $cancelledAt = null,
+                                  ?CarbonImmutable $finishedAt = null)
     {
         $this->fakeBatch = new BatchFake(
             empty($id) ? (string) Str::uuid() : $id,
@@ -94,7 +96,9 @@ trait Batchable
             $failedJobs,
             $failedJobIds,
             $options,
-            $createdAt ?? CarbonImmutable::now()
+            $createdAt ?? CarbonImmutable::now(),
+            $cancelledAt,
+            $finishedAt,
         );
 
         return $this->fakeBatch;
--- a/src/Illuminate/Bus/Batchable.php
+++ b/src/Illuminate/Bus/Batchable.php
@@ -100,6 +100,6 @@ trait Batchable
             $finishedAt,
         );
 
-        return $this->fakeBatch;
+        return [$this, $this->fakeBatch];
     }
 }
--- a/src/Illuminate/Bus/Batchable.php
+++ b/src/Illuminate/Bus/Batchable.php
@@ -4,7 +4,6 @@ namespace Illuminate\Bus;
 
 use Carbon\CarbonImmutable;
 use Illuminate\Container\Container;
-use Illuminate\Support\Carbon;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Tempting to make `withFakeBatch` `return [$job, $batch]` so that you can do:

```php
[$job, $batch] = (new ExampleJob)->withFakeBatch();
```

Wulfheart:
Maybe adding an array shape would help with autocomplete in IDEs?

bhaustein:
What about good old call by reference:

```php
$batch = new TestBatch();
$job = (new TestJob)->withFakeBatch($batch);

$job->handle();

$this->assertTrue($batch->cancelled());
$this->assertNotEmpty($batch->added);
```

jasonmccreary:
Not to be grammar police, but isn't the job _within_ a batch? So maybe `withinFakeBatch($batch)`

bhaustein:
> Not to be grammar police, but isn't the job _within_ a batch? So maybe `withinFakeBatch($batch)`

Yes, you are right. That Sounds better.

njames:
> Not to be grammar police, but isn't the job _within_ a batch? So maybe `withinFakeBatch($batch)`

Grammar police are the best kind of police 🚓

jasonmccreary:
I don't think I can edit this directly, but the [BusFake](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/Testing/Fakes/BusFake.php#L664) and [BatchRepositoryFake](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/Testing/Fakes/BatchRepositoryFake.php#L52) should be updated to use this new `BatchFake`.

It'd also be nice if the `Bus` had a similar `dispatchFakeBatch()` helper method on it to easily create fake batches for testing.

src/Illuminate/Bus/Batchable.php
```diff
@@ -49,4 +63,43 @@ public function withBatchId(string $batchId)
...
+     * Indicate that the job should use a fake batch.
+     *
+     * @param  string  $id
+     * @param  string  $name
+     * @param  int  $totalJobs
+     * @param  int  $pendingJobs
+     * @param  int  $failedJobs
+     * @param  array  $failedJobIds
+     * @param  array  $options
+     * @return \Illuminate\Support\Testing\BatchFake
```
crynobone, line 77:
```suggestion
     * @return array{0: $this, 1: \Illuminate\Support\Testing\BatchFake}
```

gustavo-ticto:
🚀

RTippin:
Seems since this update, I have a few tests that fail now, due to calling `$batch->add(...)`.

> `TypeError : array_merge(): Argument #2 must be of type array, Illuminate\Support\LazyCollection given`

https://github.com/laravel/framework/blob/abd4787bcaf0bb55dd4be35bf15f8985ee06bbcc/src/Illuminate/Support/Testing/Fakes/BatchFake.php#L81

I have a batch created, and then use a lazy collection/chunk to add jobs to the batch.

```php
private function addToBatch(Model $model, Batch $batch): void
{
    $model->relations()
        ->cursor()
        ->map(fn (Relation $relation): ThumbnailJob => $this->createThumbnailJob($relation))
        ->chunk(500)
        ->each(fn (LazyCollection $jobs) => $batch->add($jobs));
}
```

Wondering if I am going about this the wrong way, or should the `BatchFake` be updated?

PR #44105, opened 2022-09-12
[9.x] Added scoped filesystem driver

frankdejonge:
Often, files are placed on a single filesystem (like S3) but at different paths. This base path needs to be kept in sync all across the codebase. The scoped driver makes it easy to re-use disk configurations, and decorated the internal adapter.

Consider the following situation, when I have an existing `s3` disk:

```php
        's3' => [
            'driver' => 's3',
            'key' => env('AWS_ACCESS_KEY_ID'),
            'secret' => env('AWS_SECRET_ACCESS_KEY'),
            'region' => env('AWS_DEFAULT_REGION'),
            'bucket' => env('AWS_BUCKET'),
            'url' => env('AWS_URL'),
            'endpoint' => env('AWS_ENDPOINT'),
            'use_path_style_endpoint' => env('AWS_USE_PATH_STYLE_ENDPOINT', false),
            'throw' => false,
        ],
```

Using the new `scoped` driver, I can create a version of the existing `s3` driver with a deeper base path:

```php
        's3_videos' => [
            'driver' => 'scoped',
            'prefix' => 'path/for/videos',
            'disk' => 's3',
        ],
```

---

Possible optimisation, since this is a decorated adapter, we could re-use the inner adapter. I didn't find an elegant way to do so, but I'm open to suggestions.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -88,6 +88,7 @@
         "guzzlehttp/guzzle": "^7.2",
         "league/flysystem-aws-s3-v3": "^3.0",
         "league/flysystem-ftp": "^3.0",
+        "league/flysystem-path-prefixing": "^3.3",
         "league/flysystem-read-only": "^3.3",
         "league/flysystem-sftp-v3": "^3.0",
         "mockery/mockery": "^1.4.4",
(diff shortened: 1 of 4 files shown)


iotronlab:
I have created an issue with details.
https://github.com/laravel/framework/issues/44327

PR #44106, opened 2022-09-12
[9.x] Allow to define which jobs should be actually dispatched when using `Bus::fake`

mateusjunges:
### The problem

Currently, we can use the `Bus@fake` method to determine which jobs should be faked when executing tests.

The problem with it is that if I have 100 jobs that will be dispatched during one request and I want only one of them to be actually executed, then I have to explicit define which 99 jobs should be faked, instead of just defining which job should be dispatched.

### Solution
This PR adds the option to determine which jobs should be executed when using `Bus::fake`, like this example:

```php
Bus::fake()->except(MyJob::class);

//Or:

Bus::fake()->except([
    MyJob::class,
    AnotherJob::class,
]);
```

Now, all jobs dispatched in the test case except the `MyJob::class`, in the first example, or `MyJob::class` and `AnotherJob::class` in the second one, will be faked.

If this PR gets merged, I can open another one adding this functionality to the `Queue` and `Event` classes. I can also add it to this PR if Taylor think that is that way to go.


dwightwatson:
Perhaps it would be clearer to chain `except` onto the existing `fake` method?

```php
Bus::fake()->except(MyJob::class)
```

mateusjunges:
@dwightwatson I liked it, just updated the PR

jasonmccreary:
I have it on my list to do this for the `Queue::fake()` as well. Being able to specify both exclusion and inclusion provides the most flexibility when testing.

mateusjunges:
@jasonmccreary I'm waiting for Taylor's to review this PR before implementing it on other fake classes, so I don't waste time if it gets rejected. But yeah, I agree with you :slightly_smiling_face:

jasonmccreary:
@mateusjunges, I'll take it off my list then. ;)

mateusjunges:
I'm opening a follow up PR to implement this feature in other fake classes as well.

jasonmccreary:
@jasonmccreary, yay! Please at me for the `Queue::fake()` as I had some ideas there too.

mateusjunges:
Hey @jasonmccreary you can check it out here #44117

PR #30880, opened 2019-12-18
[6.x] Create Factory Primers

browner12:
### Problem

Let's take a typical factory with a relationship. We'll use the example right off the Laravel docs.

```php
$factory->define(App\Post::class, function ($faker) {
    return [
        'title' => $faker->title,
        'content' => $faker->paragraph,
        'user_id' => factory(App\User::class),
    ];
});
```

When we create a `Post`, we will also create a `User`. At the surface this seems okay, but let's think what happens when we create 50 `Post`s.

```php
factory(Post::class, 50)->create();
```

Without the relationship we'd be running 50 inserts, but with the relationship we double that to 100 queries!  Now imagine a factory that has multiple relationships. The number of database queries (and time) grows with each new relationship handled this way.

---

What if we assume that `User`s have already been created, and we'll randomly select 1 to attach the `Post` to?

```php
$factory->define(App\Post::class, function ($faker) {
    return [
        'title' => $faker->title,
        'content' => $faker->paragraph,
        'user_id' => App\User::all()->random(),
    ];
});
```

Unfortunately, this does not cut down on our query count, since the `all()` query is executed for *every single* new `Post`.

---

We can also not pass it as an override attribute.

```php
factory(Post::class, 50)->create([
    'user_id' => App\User::all()->random(),
]);
```

This will cut down our query count, but since it is only executed once, all 50 `Post`s will receive the same User ID, which is not what we want.

### Solution

The solution to this is a feature I call "primers". At the time of building the factories, you can inject key/value pairs into the builder that the factories can then use to generate their data.

Let's first look at how we call it. Simply chain the `prime()` method onto your factory prior to calling `make()` or `create()`, and pass it a string key and a value of any type you like.

```php
factory(Post::class, 50)->prime('users', User::all())->create();
```

With this call we've made 1 query to get all the `User`s.  Now let's see how to use primers in the factories.

```php
$factory->define(App\Post::class, function ($faker, $attributes, $primers) {
    return [
        'title' => $faker->title,
        'content' => $faker->paragraph,
        'user_id' => $primers['users'] ? $primers['users']->random() : factory(App\User::class),
    ];
});
```

Here we can see the primers are passed into the factory definition as a new 3rd parameter.  For properties that we expect primers for, we should first check to see the primer exists. If the primer exists, we can use it as needed, but also fall back to the old way of determining the relationship if the primer was not provided.

With this new feature, and a primed factory, we've now cut our queries down from 100 to 51!

Primers can also be used with factory states.

```php
[...]

PR #44107, opened 2022-09-13
[9.x] Allow using a model instance in place of nested model factories

jessarcher:
For applications with nested model factories, tests and seeders can get a little overwhelming when you want the nested models to all be related to a specific model.

For example, imagine you have `User`, `Team`, and `Post` models where posts belong to a user and a team, and the team also belongs to a user.

The following factory call would create two user records. One for the post itself, and one for the team that the post factory created.

```php
$post = Post::factory()->create();
```

To get around this, you might write something like the following:

```php
$post = Post::factory()
    ->for($user)
    ->for(Team::factory()->for($user))
    ->create();
```

That's still pretty nice! But maybe your posts have images that also belong to a team and/or a user. It can get out of control pretty quickly, and this is far from the most complex nesting I've seen.

This PR introduces a new `using` method for factories that allows you to provide one or more model instances to be used in place of any nested factories for the given models. What's more, it works recursively, automatically passing the model instances to any nested factories.

For our example above, your code would now look something like this:

```php
$post = Post::factory()
    ->using($user)
    ->create();
```

In this example, the post _and_ the nested team would all be owned by the provided user instance, with no additional records created. Depending on your application, this will either tidy up your factory calls, or it will prevent superfluous models from being created.

Of course, to take advantage of this feature, your factories must specify other factory instances in their definitions. If your factory definition calls the `create` method on any nested factories, then the created model will be used instead of anything passed to the `using` method.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/BelongsToManyRelationship.php
+++ b/src/Illuminate/Database/Eloquent/Factories/BelongsToManyRelationship.php
@@ -62,13 +62,13 @@ class BelongsToManyRelationship
     /**
      * Specify the model instances to always use when creating relationships.
      *
-     * @param  \Illuminate\Support\Collection  $using
+     * @param  \Illuminate\Support\Collection  $recycle
      * @return $this
      */
-    public function using($using)
+    public function recycle($recycle)
     {
         if ($this->factory instanceof Factory) {
-            $this->factory = $this->factory->using($using);
+            $this->factory = $this->factory->recycle($recycle);
         }
 
         return $this;
--- a/src/Illuminate/Database/Eloquent/Factories/BelongsToManyRelationship.php
+++ b/src/Illuminate/Database/Eloquent/Factories/BelongsToManyRelationship.php
@@ -43,15 +43,6 @@ class BelongsToManyRelationship
         $this->relationship = $relationship;
     }
 
-    public function using($using)
-    {
-        if ($this->factory instanceof Factory) {
-            $this->factory = $this->factory->using($using);
-        }
-
-        return $this;
-    }
-
     /**
      * Create the attached relationship for the given model.
      *
@@ -66,5 +57,20 @@ class BelongsToManyRelationship
                 is_callable($this->pivot) ? call_user_func($this->pivot, $model) : $this->pivot
             );
         });
+    }
+
+    /**
+     * Specify the model instances to always use when creating relationships.
+     *
+     * @param  \Illuminate\Support\Collection  $using
+     * @return $this
+     */
+    public function using($using)
+    {
+        if ($this->factory instanceof Factory) {
+            $this->factory = $this->factory->using($using);
+        }
+
+        return $this;
     }
 }
--- a/src/Illuminate/Database/Eloquent/Factories/BelongsToManyRelationship.php
+++ b/src/Illuminate/Database/Eloquent/Factories/BelongsToManyRelationship.php
@@ -43,6 +43,15 @@ class BelongsToManyRelationship
         $this->relationship = $relationship;
     }
 
+    public function using($using)
+    {
+        if ($this->factory instanceof Factory) {
+            $this->factory = $this->factory->using($using);
+        }
+
+        return $this;
+    }
+
     /**
      * Create the attached relationship for the given model.
      *
--- a/src/Illuminate/Database/Eloquent/Factories/BelongsToRelationship.php
+++ b/src/Illuminate/Database/Eloquent/Factories/BelongsToRelationship.php
@@ -41,15 +41,6 @@ class BelongsToRelationship
         $this->relationship = $relationship;
     }
 
-    public function using($using)
-    {
-        if ($this->factory instanceof Factory) {
-            $this->factory = $this->factory->using($using);
-        }
-
-        return $this;
-    }
-
     /**
      * Get the parent model attributes and resolvers for the given child model.
      *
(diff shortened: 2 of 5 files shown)


dillingham:
This is a huge win for cleanliness. 

One thought, maybe "using" could be called "provide". Being a vue user of [provide](https://vuejs.org/guide/components/provide-inject.html#provide), I personally wouldn't overlook "using" and realize "provide" was for all children.

rodrigopedra:
I prefer to keep `->using()` to be consistent on how we define a custom pivot model when defining a belongs to many relationship.

But I agree this PR is a huge time saver.

taylorotwell:
I agree that `using` usually has a different meaning in Laravel.

What do you think about:

```php
Post::factory()
        ->recycle($user)
        ->create();
```

sebdesign:
Recycle is not very clear IMO, it could mean that something is trashed and then repurposed, It could be confused with delete/restore. Maybe `reuse` is more accurate?

WAZIRI123:
> 

i think reuse is more clear

basepack:
@jessarcher it won't work when following this approach: https://laravel.com/docs/9.x/eloquent-factories#defining-relationships-within-factories

```
// PostFactory
public function definition()
{
    return [
        'user_id' => User::factory(), 
    ];
}

// in your test
Post::factory()->recycle($user)->create();
Post::factory()->recycle($user)->create();

ray(User::count()); // 2
```

martin-ro:
Post Model:
```php
class Post extends Model
{
    use HasFactory;

    public function user(): BelongsTo
    {
        return $this->belongsTo(User::class);
    }
}
```

Post Factory:
```php
class PostFactory extends Factory
{
    protected $model = Post::class;

    public function definition(): array
    {
        return [
            'title' => fake()->words(5, true),
            'user_id' => User::factory()->create(),
        ];
    }
}
```

DatabaseSeeder:
```php
public function run()
    {
        // Creates 1 posts + 1 users
        // Post::factory()->create();

        // Creates 10 posts + 10 users
        // Post::factory(10)->create();

        // Creates 1 posts + 2 users
        // Post::factory()
        //    ->recycle(User::factory()->create())
        //    ->create();

        // Creates 10 posts + 11 users
        Post::factory(10)
            ->recycle(User::factory()->create())
            ->create();
    }
```

Am I using it wrong?

jessarcher:
Hi @basepack, I can't replicate this. Using your code in a fresh Laravel project, I only have one User at the end.

Are you sure there's only 1 user in your database just before you call the Post factory? And are you sure that your PostFactory is just returning the User factory, without calling the `create` method on it?

martin-ro:
> without calling the `create` method on it?

That threw me off at first.

basepack:
Hi @jessarcher, @martin-ro, no I do not call `->create()`, I will post my complete setup here, maybe that will uncover the issue:

```php
// SiteFactory
namespace Database\Factories;

use App\Models\User;
use Faker\Provider\Uuid;
use Illuminate\Database\Eloquent\Factories\Factory;

/**
 * @extends \Illuminate\Database\Eloquent\Factories\Factory<\App\Models\Site>
 */
class SiteFactory extends Factory
{

    public function definition()
    {
        return [
            'uuid' => Uuid::uuid(),
            'user_id' => User::factory()->create()->id,
            'domain_name' => $this->faker->domainName,
        ];
    }
}
````

```php
// EmailaddressFactory
namespace Database\Factories;

use App\Models\Site;
use Illuminate\Database\Eloquent\Factories\Factory;

/**
 * @extends \Illuminate\Database\Eloquent\Factories\Factory<\App\Models\Emailaddress>
 */
class EmailaddressFactory extends Factory
{
    public function definition()
    {
        return [
            'emailaddress' => $this->faker->unique()->safeEmail,
            'site_id' => Site::factory(),
        ];
    }

    public function withEmailaddress(string $emailaddress): self
    {
        return $this
            ->state([
                'emailaddress' => $emailaddress,
            ])
            ->for(
                Site::factory([
                    'domain_name' => str($emailaddress)->after('@')->toString(),
                ])
            );
    }
}
````

```php
it('test', function () {
    $site = Site::factory()->create([
        'domain_name' => 'example.com',
    ]);

    $adminEmail = Emailaddress::factory()
        ->recycle($site)
        ->withEmailaddress('info@example.com')
        ->create();

    $userEmail = Emailaddress::factory()
        ->recycle($site)
        ->withEmailaddress('user@example.com')
        ->create();

    ray(Site::count()); // 3 <-- should be 1
    ray(Emailaddress::count()); // 2
});
```

jessarcher:
Thanks @basepack, I was able to replicate it with this setup. It looks like it's the `for` call in the `withEmailaddress` state that's triggering the extra site(s) to be created.

I'll dig into this properly tomorrow.

michal3377:
This is a great feature, making factories much easier to write.
However, could we maybe go 1 step further, allowing for providing multiple models of a given type, which would be then randomly picked? This would allow developers to provide their own, already created pool of models, so they could be reused in all nested relations. 
For example, we often see that the User model relates to everything - as an author of some content, comments, reactions, performed operations etc. In more complex applications, that nesting can go deep. In that scenario, much better would be seeding a number of users once and then reusing them in other relations (so we don't create a user only to be the author of a single comment). If that nested reuse could be achieved with a one-liner (which seems to be not far away from what this PR achieves), that would be a game changer.

In my opinion, Eloquent Factories is a great concept and the only thing it lacks is better support for model reuse. That enhancement would fix that and make them much more versatile.

jessarcher:
Hey @basepack, I've created #44265 to address your issue :+1:

jessarcher:
@michal3377 Feel free to attempt a PR. I can potentially see the benefit in seeders where you might want a handful of users to be used throughout instead of either recycling a single user, or creating a new user for every relationship.

basepack:
@jessarcher thanks for that, looks good! 👍

browner12:
today I realized `recycle` is basically the "primer" feature I tried to get merged 5 years ago 😂

https://github.com/laravel/framework/pull/30880

PR #44110, opened 2022-09-13
[9.x] Allow any kind of whitespace in cron expressions

onlime:
Be more tolerant on task scheduling cron expression spacing, allowing any whitespace as separator. Previously, only a single space was allowed as separator, resulting in the following exception if e.g. a tab was used instead:

```php
$schedule->command('foobar:sample-task')->cron('5,35 *	* * *'); // "5,35 *\t* * *"
```

```bash
$ pa schedule:list

   ErrorException 

  Undefined array key 4

  at vendor/laravel/framework/src/Illuminate/Console/Scheduling/ScheduleListCommand.php:165
    161▕     {
    162▕         $expressions = explode(' ', $expression);
    163▕ 
    164▕         return collect($spacing)
  ➜ 165▕             ->map(fn ($length, $index) => str_pad($expressions[$index], $length))
    166▕             ->implode(' ');
    167▕     }
    168▕ 
    169▕     /**
```

### DIFF
--- a/src/Illuminate/Console/Scheduling/ManagesFrequencies.php
+++ b/src/Illuminate/Console/Scheduling/ManagesFrequencies.php
@@ -546,7 +546,7 @@ trait ManagesFrequencies
      */
     protected function spliceIntoPosition($position, $value)
     {
-        $segments = explode(' ', $this->expression);
+        $segments = preg_split("/\s+/", $this->expression);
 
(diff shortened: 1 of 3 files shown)


jbrooksuk (approved):
According to `man crontab`

> The fields may be separated  by  spaces or tabs.

PR #44115, opened 2022-09-13
Fixes `make:model` issues

mahmoudmohamedramadan:
When I tried to run the next command as exists in [docs](https://laravel.com/docs/9.x/eloquent#generating-model-classes), I noticed that the model and `controller` **ONLY** are created, and when I analyzed the code I discovered that the request class did not create because it will be created **ONLY** in case we pass a valid model name to `make:controller` command as exists in [ControllerMakeCommand](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Routing/Console/ControllerMakeCommand.php#L123), and because I did not pass `resource`, or `api` options the **NULL** value is passed to the `--model` option so, I have passed the given model name directly which passed to `make:model` option.

```SHELL
php artisan make:model Flight -c -R
```

- I removed `array_filter` method because it does nothing where we did not pass a closure.
- I removed `$this->option('all')` with `--requests` option also because it is meaningless as all model classes are created when we pass `--all` option.

### DIFF
--- a/src/Illuminate/Foundation/Console/ModelMakeCommand.php
+++ b/src/Illuminate/Foundation/Console/ModelMakeCommand.php
@@ -145,12 +145,12 @@ class ModelMakeCommand extends GeneratorCommand
 
         $modelName = $this->qualifyClass($this->getNameInput());
 
-        $this->call('make:controller', array_filter([
+        $this->call('make:controller', [
             'name' => "{$controller}Controller",
-            '--model' => $this->option('resource') || $this->option('api') ? $modelName : null,
+            '--model' => $modelName,
             '--api' => $this->option('api'),
-            '--requests' => $this->option('requests') || $this->option('all'),
-        ]));
+            '--requests' => $this->option('requests'),
+        ]);
     }
 
     /**


taylorotwell:
Some of these changes I don't understand. I'll have to try it myself and am a bit overwhelmed with PRs and Laracon right now.

mahmoudmohamedramadan:
Sorry for that sir, let me explain again in detail to save your time.

## Before this PR gets merged

ModelMakeCommand

According to the previous figure, we can do these actions:

- We can remove `array_filter` method because it does nothing as it did not have a closure to iterate over each value.
- We can remove the ternary operator which exists in the `--model` option because the model name is actually passed to `make:model` command, that's why we can pass it directly to that option.
- We can remove `$this->option('all')` that exists in `--requests` option because it meaningless.

**As we noticed in the previous figure that the request class did not create, and I'll illustrate that in the next figure**

ControllerMakeCommand

According to this figure, we check for the given model option to execute `buildModelReplacements` and [`buildFormRequestReplacements`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Routing/Console/ControllerMakeCommand.php#L174) and  when we passed `-c` to create a controller the `--model` option gets **NULL** value because we did not pass `resource` or `api` options so, the `buildModelReplacements` and `buildFormRequestReplacements` did not execute that's why the request class did not be created

I wish I explained that precisely 😃.

dennisprudlo:
Well this will definitely change behavior for the `make:model` command.

> We can remove the ternary operator which exists in the --model option because the model name is actually passed to make:model command, that's why we can pass it directly to that option.

This will always create resource controllers when using `make:model SomeModel --controller` but not everyone may want a resource controller. If someone wants a resource controller while creating a model they should user `make:model SomeModel --resource`. Thats exactly why the `resource`-option exists.

> We can remove $this->option('all') that exists in --requests option because it meaningless.

Using `make:model SomeModel --all` will not create FormRequest-classes anymore.

> We can remove array_filter method because it does nothing as it did not have a closure to iterate over each value.

This is only true if we want to apply the changes above. When `make:model SomeModel --controller` is used the --model, --api and --requests options are either null or false. The array_filter will remove these options from the array.

I'm neither against nor for this PR, just saying that this heavily changes the behavior of the `make:model` command.

mahmoudmohamedramadan:
> This will always create resource controllers when using make:model SomeModel --controller but not everyone may want a resource controller. If someone wants a resource controller while creating a model they should user make:model SomeModel --resource. Thats exactly why the resource-option exists.

Hmm- The reason that makes me remove that ternary operator, is the request class did not be created because the given value to `--model` option is **NULL** as I did not pass `resource` or `api` options

> Using make:model SomeModel --all will not create FormRequest-classes anymore.

You are right, but let us concentrate on that approach is invalid, is I create a request class when I pass `--all` option, I can set the `--request` options on the `handle` method

tontonsb:
cc @PovilasKorop

It seems natural to expect that `php artisan make:model Post -R` will generate the requests even if no controller is created.

PR #44117, opened 2022-09-13
[9.x] Implement `except` method for fake classes to define what should not be faked

mateusjunges:
This PR is a follow up on #44106 and implements the same `except` method for the `Event` and `Queue` fake classes.

@jasonmccreary I'd appreciate your thoughts here :smiley:

### DIFF
--- a/src/Illuminate/Support/Testing/Fakes/QueueFake.php
+++ b/src/Illuminate/Support/Testing/Fakes/QueueFake.php
@@ -29,6 +29,13 @@ class QueueFake extends QueueManager implements Queue
     protected $jobsToFake;
 
     /**
+     * The job types that should be pushed to the queue and not intercepted.
+     *
+     * @var Collection
+     */
+    protected $jobsToBeQueued;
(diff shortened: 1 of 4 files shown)


src/Illuminate/Support/Testing/Fakes/QueueFake.php
```diff
@@ -323,6 +331,10 @@ public function push($job, $data = '', $queue = null)
      */
     public function shouldFakeJob($job)
```
mateusjunges, line 332: Shouldn't this method be `protected`? I created `shouldDispatchJob` as public to be consistent with this one, let me know what's the correct approach here.

jasonmccreary:
Awesome. I needed this the other week when testing a job dispatched another job. I wanted to actually dispatch that job in my test, but then fake anything else. Having `Queue::fake()->except(JobImTesting::class)` would have been awesome, instead of my hack of `(new JobImTesting)->handle();`. 👍

PR #44119, opened 2022-09-13
[9.x] Add $added array to Translator to store added lines

npostman:
Adding lines to the Translator prevents the Translator from loading the translation files, as it thinks it has already loaded. By storing the added lines in a separate array, and merging the 'loaded array' and the 'added array' when getting a line, you can prevent that behavior.

Steps to reproduce:
Set a different language (eg. nl) as main locale, and English as fallback. Duplicate the lang/en/passwords.php file to the new language dir and update the values to something different. In your view, echo
```
{{ __('passwords.reset') }}
{{ __('passwords.sent') }}
```

It should show your new language output.

If you add a line to the translator before the files are loaded (eg. in a service provider):

```
    /**
     *
     * @return void
     */
    public function boot()
    {
        app('translator')->addLines(['passwords.reset' => 'UPDATED LINE'], 'nl');
    }
```

Now, if you open your view again, it shows the UPDATED LINE for the 'passwords.reset' entry, but the 'passwords.sent' entry shows the English fallback line. This is because after adding the line to the nl locale, the Translator->isLoaded returns true for the current locale, as it is set by adding the line.

This update will fix that, as it stores added lines in a separate array.

### DIFF
--- a/src/Illuminate/Translation/Translator.php
+++ b/src/Illuminate/Translation/Translator.php
@@ -41,6 +41,13 @@ class Translator extends NamespacedItemResolver implements TranslatorContract
      * @var array
      */
     protected $loaded = [];
+    
+    /**
+     * The array of added translation lines.
+     *
+     * @var array
+     */
+    protected $added = [];
 
     /**
      * The message selector.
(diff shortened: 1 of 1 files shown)


npostman:
To work around this issue, I forced the loading of the language files before adding the lines, like so:
```
app('translator')->has('passwords.reset')
```
But doing this in the AppServiceProvider defeats the purpose of lazy loading the language files, and in many cases it does not need to load the files at all...

taylorotwell:
One issue here is this is technically a change in behavior and could break existing applications relying on the current behavior for whatever reason. Since you can work around it relatively easily I'm going to leave it be for now.

PR #44122, opened 2022-09-14
[9.x] Request lifecycle duration handler

timacdonald:
This PR is in a similar spirit the the [cumulative query duration PR](https://github.com/laravel/framework/pull/42744). It allows developers to easily handle when an application request lifecycle exceeds a certain threshold.

```php
use Carbon\CarbonInterval as Interval;
use Illuminate\Contracts\Http\Kernel;


public function boot()
{
    if ($this->app->runningInConsole()) {
        return;
    }

    $this->app[Kernel::class]->whenRequestLifecycleIsLongerThan(
        Interval::seconds(1),
        fn ($startedAt, $request, $response) => /* ... */
    );

}
```

The feature includes the duration of the request handle method and the terminating of the middleware and application.

This gives developers a nice hook around the entire lifecycle of the request, as seen: https://github.com/laravel/laravel/blob/c1dc4199b83466a3a6a8c70953250b0e2ec70001/public/index.php#L51-L55

This is an easy starting point for tracking slow requests and not a full blown replacement for APM, but gives developers some quick insight into their system with little to no extra knowledge / setup.

```php
$this->app[Kernel::class]->whenRequestLifecycleIsLongerThan(
    Interval::seconds(1),
    fn ($startedAt, $request, $response) =>
        Log::notice('Request exceeded duration threshold.', [
            'user' => $request->user()?->id,
            'url' => $request->fullUrl(),
            'duration' => $startedAt->floatDiffInSeconds(),
        ])
);
```

The handler is invoked after the request has been returned to the user.

Command compliment: https://github.com/laravel/framework/pull/44125

### DIFF
--- a/src/Illuminate/Foundation/Http/Kernel.php
+++ b/src/Illuminate/Foundation/Http/Kernel.php
@@ -2,18 +2,24 @@
 
 namespace Illuminate\Foundation\Http;
 
+use Carbon\CarbonInterval;
+use DateTimeInterface;
 use Illuminate\Contracts\Debug\ExceptionHandler;
 use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Contracts\Http\Kernel as KernelContract;
 use Illuminate\Foundation\Http\Events\RequestHandled;
 use Illuminate\Routing\Pipeline;
 use Illuminate\Routing\Router;
+use Illuminate\Support\Carbon;
 use Illuminate\Support\Facades\Facade;
+use Illuminate\Support\InteractsWithTime;
 use InvalidArgumentException;
 use Throwable;
 
 class Kernel implements KernelContract
 {
+    use InteractsWithTime;
+
     /**
      * The application implementation.
      *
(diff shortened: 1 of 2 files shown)


christophrumpel:
Hey, just gave this a try. Is there a reason this does NOT work, when I use the HTTP Kernel namespace, and not the kernel contract? (with the contract it works)

timacdonald:
Hey christoph,

Laravel only [binds the Kernel against the contract](https://github.com/laravel/laravel/blob/9.x/bootstrap/app.php#L29-L43), so resolving the contract is the only way (that I know of) to get resolve the actual instance Laravel is using.

If you resolve the "HTTP Kernel namespace" instance, you are resolving a new instance and not the one that is handling the incoming request.

If you wanted to be able to resolve the HTTP namespace Kernel, you would need to bind it to the same instance that the contract uses...

Like:

```php
$this->app->singleton(\App\Http\Kernel::class, fn ($a) => 
    $a[\Illuminate\Contracts\Http\Kernel::class)
);
```

Which feels a little backwards. Alternatively, we could PR a helper to the service provider that handles the resolution under the hood:

```php
public function boot()
{
    $this->whenRequestLifecycleIsLongerThan(/* ... *);
}
```

which proxies the call to the resolved contract instance, but I don't know that we need it IMO.

christophrumpel:
Hey Tim, thanks for clearing this up. I don't think there is a need currently to change it. It's just something people have to be aware of when using it.

PR #44130, opened 2022-09-14
Sets `requests` options as true

mahmoudmohamedramadan:
This PR sets the `requests` options as true in case we pass `--all` option.


taylorotwell:
Why the model changes?

taylorotwell:
I still don't have a good understand of what we are fixing here. What is wrong with this command for end users?

PR #44132, opened 2022-09-14
[9.x] Adds ShouldNotify contract

alkin:
Broadcasting events is fairly easy with Laravel. All you have to do is to add the `ShouldBroadcast` interface to the event and implement the `broadcastOn` method.

Another feature that commonly uses events as a triggers is the Notification. Although broadcast and notifications serve the same purpose (notifying some third party about some event), dispatching notifications requires extra steps. People must create a Notification object, listen to the event and manually dispatch the notification, like the following.

```php
Event::listen(IssueClosedEvent::class, function ($event) {
    $event->issue->author->notify(new IssueClosedNotification($event->issue));
});
```

This PR adds a `ShouldNotify` interface that can be added to any event, which automatically dispatches a notification when the event is dispatched, similar with how broadcasting works.

When the following event is dispatched, the author of the Issue will be automatically notified.

```php
class IssueClosedEvent extends Notification implements ShouldNotify
{
    public function __construct(
        public Issue $issue
    ){
    }

    public function notifyTo()
    {
        return [
            $this->issue->author,
        ];
    }

    {
        return ['mail'];
    }

    public function toMail()
    {
    }
}

```

### DIFF
--- a/src/Illuminate/Contracts/Notifications/ShouldNotify.php
+++ b/src/Illuminate/Contracts/Notifications/ShouldNotify.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Illuminate\Contracts\Notifications;
+
+interface ShouldNotify
+{
+    /**
+     * Get the notifiables the event should notify.
+     *
+     * @return object|object[]
+     */
+    public function notifyTo();
+}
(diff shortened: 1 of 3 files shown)


taylorotwell:
I have sort of had ideas like this before - let me explore it a bit more.

PR #44134, opened 2022-09-14
[9.x] Improve Blade compilation exception messages

danilopinotti:
# Problem

Currently, when we write a `@foreach` or `@forelse` the wrong way, the exception message is not descriptive. e.g:

Code:
```
@foreach
TEST
@endforeach
```

Exception:
image

The main objective of this PR is to improve this message:
image

In an actual situation, it occurred for a beginner developer to take hours to discover the problem since the stacktrace also doesn't make it easy to debug issues in rendering views.

### DIFF
--- a/src/Illuminate/Contracts/View/ViewCompilationException.php
+++ b/src/Illuminate/Contracts/View/ViewCompilationException.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace Illuminate\Contracts\View;
+
+use Exception;
+
+class ViewCompilationException extends Exception
+{
+    //
+}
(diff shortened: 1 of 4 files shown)


maxddev:
@danilopinotti 

Because the regex quantifier after "as" was changed from * to +, if the foreach statement is multiline, this now throws an error:

```
@foreach ($foo as 
$bar)
```

Should this be considered?

danilopinotti:
> @danilopinotti 
> 
> Because the regex quantifier after "as" was changed from * to +, if the foreach statement is multiline, this now throws an error:
> 
> ```
> @foreach ($foo as
> $bar)
> ```
> 
> Should this be considered?

I wasnt considering. I will submit a PR fixing it

PR #44138, opened 2022-09-15
[9.x] Add Model::withoutTimestamps(...)

timacdonald:
Re-submission of #44005 as static methods.

API now matching the static nature of `withoutTouching`, but hopefully with a simpler implementation than previous attempts at this functionality.

```php
$user = User::first();


// `updated_at` is not changed...

User::withoutTimestamps(
    fn () => $user->update(['reserved_at' => now()])
);
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
@@ -14,7 +14,7 @@ trait HasTimestamps
     public $timestamps = true;
 
     /**
-     * The list of models classes that should have timestamps disabled.
+     * The list of models classes that have timestamps temporarily disabled.
      *
      * @var array
      */
@@ -164,7 +164,7 @@ trait HasTimestamps
     }
 
     /**
-     * Disables timestamps for the current class during given callback scope.
+     * Disable timestamps for the current class during the given callback scope.
      *
      * @param  callable  $callback
      * @return void
@@ -175,7 +175,7 @@ trait HasTimestamps
     }
 
     /**
-     * Disables timestamps for the given model classes during given callback scope.
+     * Disable timestamps for the given model classes during the given callback scope.
      *
      * @param  array  $models
      * @param  callable  $callback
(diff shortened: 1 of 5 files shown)


src/Illuminate/Database/Eloquent/Concerns/HasTimestamps.php
```diff
@@ -113,7 +120,7 @@ public function freshTimestampString()
      */
     public function usesTimestamps()
     {
-        return $this->timestamps;
+        return $this->timestamps && ! static::isIgnoringTimestamps($this::class);
```
wotta, line 123: According to all other parts of you PR shouldn't you use `static::class` here?

marcovo:
I suspect this implementation will not work as expected with nested calls to `withoutTimestamps()`. E.g.:

```
User::withoutTimestamps(function () use ($user) {

    User::withoutTimestamps(
        fn () => $user->update(['reserved_at' => now()])
    );
    
    $user->update([...]); // <-- Will update timestamps
});
```

You can imagine one would call `withoutTimestamps()` e.g. in a controller and then deeper somewhere in a model, in different contexts, which would lead to this situation. I think it should be fixed to be able to cope with this situation.

timacdonald:
I did not intend for it to be used this way. Would recommend not nesting calls to this or the other methods that work this way.

jasonmccreary:
Yeah, I've run into this before. Disabling timestamps is a tricky beast and heavily dependent on the use case.

moisish:
I had my own implementation of withoutTimestamps on trait that was working differently. 
The withoutTimestamps was a non static method that I was able to use on the existing model I am using or on an Eloquent Builder instance

On a single model
```
$model
    ->withoutTimestamps()
    ->update([
           'included' => true
      ]);
```

Updating all models by using a relationship
```
$tasks
            ->subTasks()
            ->tap(function (Builder $query) {
                $query->getModel()->withoutTimestamps();
            })
            ->update([
                'included' => false,
            ]);
```

dennisprudlo:
Wouldn't both example be possible like this?
```php
Model::withoutTimestamps(fn () => $model->update([ 'included' => true ]));
```

```php
SubTask::withoutTimestamps(fn () => $task->subTasks()->update([ 'included' => false ]));
```

timacdonald:
See @dennisprudlo suggestion above, and also see the previous PR where the method was not static for future discussion: https://github.com/laravel/framework/pull/44005

PR #44141, opened 2022-09-15
[9.x] Optimize code style for `originalIsEquivalent`

kayw-geek:
At present, we can think, the original and attribute are not equivalent when the `$original` is null and the `$attribute` is not null.


taylorotwell:
No plans to change this on a patch release.

PR #44146, opened 2022-09-15
[9.x] Improve UUID and ULID support for Eloquent

driesvints:
Improves upon #44074 by:

- Allowing to more easily use the trait to just be used on non-primary key columns
- Adds a special ulid column on Blueprint that creates a char(26) behind the scenes
- Automatically sets `incrementing` to false when a `string` type primary key is used
- Adds more tests

I was doubting on checking to see if I could add support to use both traits at the same time (so you could both fill in UUID and ULID columns) but I guess that scenario isn't so common.


src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -1851,6 +1851,10 @@ public function setKeyType($type)
      */
     public function getIncrementing()
     {
+        if ($this->getKeyType() === 'string') {
+            return false;
+        }
+
```
driesvints, line 1857: This might need to be rethought. In my mind a string key type column can never be incrementing but I might be wrong.

tests/Database/DatabaseEloquentModelTest.php
```diff
@@ -2214,8 +2214,10 @@ public function testIntKeyTypePreserved()
     public function testStringKeyTypePreserved()
     {
         $model = $this->getMockBuilder(EloquentKeyTypeModelStub::class)->onlyMethods(['newModelQuery', 'updateTimestamps', 'refresh'])->getMock();
+        $model->id = 'string id';
+
         $query = m::mock(Builder::class);
-        $query->shouldReceive('insertGetId')->once()->with([], 'id')->andReturn('string id');
+        $query->shouldReceive('insert')->once()->with(['id' => 'string id']);
         $query->shouldReceive('getConnection')->once();
         $model->expects($this->once())->method('newModelQuery')->willReturn($query);
```
driesvints, line 2222: I had to rewrite the test here because of the change I did in the `getIncrementing` method. I think the current test wasn't realistic as I can't see how you can increment with string-based keys. With string-based keys you always need to set them beforehand afaik (like I adjusted the test to).

src/Illuminate/Database/Eloquent/Concerns/HasUuids.php
```diff
@@ -49,16 +49,10 @@ public function uniqueIds()
      */
     public function getKeyType()
     {
-        return 'string';
-    }
+        if (in_array($this->getKeyName(), $this->uniqueIds())) {
```
antonkomarev, line 52:
```suggestion
        if (\in_array($this->getKeyName(), $this->uniqueIds())) {
```

antonkomarev: I might be wrong, but this method FQN will work faster.

driesvints: This is not necessary. We don't do this anywhere else.

taylorotwell:
Why was the change to `getKeyType` not needed on HasUlids too?

driesvints:
@taylorotwell whoops, you're right. Pushed.

I first was thinking of doing it through a property but when you apply a trait with a property directly on a model you cannot overwrite it.

martio:
Good job! 👍

martio:
Maybe it's nice to add support for a binary column type for UUID? :smirk:

The-Hasanov:
This update break jenssegers/laravel-mongodb
``` php
public function getIncrementing()
    {
        if ($this->getKeyType() === 'string') {
            return false;
        }

        return $this->incrementing;
    }
```

driesvints:
@The-Hasanov I replied on the related PR fix: https://github.com/jenssegers/laravel-mongodb/pull/2452

PR #44151, opened 2022-09-16
[9.x] Make \Illuminate\Routing\Router redirect method accept several URIs

brunoalod:
This PR makes the _\Illuminate\Routing\Router@redirect_ method accept an array of URIs instead of a single URI. 

I found myself having to redirect several routes to a single one and thought this would be useful. 

This should not be a breaking change.

```PHP
// Current
Route::redirect('contact-us', 'contact');
Route::redirect('contact-us-2', 'contact');
Route::redirect('another-contact-us', 'contact');

// Proposal
Route::redirect(['contact-us', 'contact-us-2', 'another-contact-us'], 'contact');
```

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -238,18 +238,28 @@ class Router implements BindingRegistrar, RegistrarContract
     }
 
     /**
-     * Create a redirect from one URI to another.
+     * Create a redirect from one or many URIs to another.
      *
-     * @param  string  $uri
+     * @param  string|string[]  $uri
      * @param  string  $destination
      * @param  int  $status
-     * @return \Illuminate\Routing\Route
+     * @return \Illuminate\Routing\Route|\Illuminate\Routing\Route[]
      */
     public function redirect($uri, $destination, $status = 302)
     {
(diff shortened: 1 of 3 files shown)


src/Illuminate/Routing/Router.php
```diff
@@ -238,18 +238,28 @@ public function fallback($action)
...
     public function redirect($uri, $destination, $status = 302)
     {
-        return $this->any($uri, '\Illuminate\Routing\RedirectController')
+        if(is_string($uri)) {
+            $uri = [$uri];
+        }
+
+        $routes = [];
+
+        foreach ($uri as $value) {
```
BrandonSurowiec, line 256: You can do:

```php
foreach ((array) $uri as $value) {
```

To ensure that the string will be cast to an array instead of the longer if check above.

brunoalod: Now that's cool. Didn't know. Thanks.

taylorotwell:
I would just do something like...

```php
collect(['uri1', 'uri2')->each(fn ($uri) => Route::redirect($uri, 'contact');
```

PR #44153, opened 2022-09-16
[9.x] Prompt to create MySQL db when migrating

timacdonald:
This PR is the MySQL counterpart to https://github.com/laravel/framework/pull/43867

Screen Shot 2022-09-16 at 4 34 49 pm

```
# No prompt, just attempts to create the file.
# If it fails the exception is shown.

php artisan migrate --force

# No prompt and the exception is shown.
# This is the same as the current behaviour.

php artisan migrate --force
```

### DIFF
--- a/src/Illuminate/Database/Console/Migrations/MigrateCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/MigrateCommand.php
@@ -9,6 +9,8 @@ use Illuminate\Database\Events\SchemaLoaded;
 use Illuminate\Database\Migrations\Migrator;
 use Illuminate\Database\SQLiteDatabaseDoesNotExistException;
 use Illuminate\Database\SqlServerConnection;
+use PDOException;
+use Throwable;
 
 class MigrateCommand extends BaseCommand
 {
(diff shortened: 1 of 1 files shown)


jbrooksuk:
@timacdonald what happens when the user does not have permissions to create additional databases?

src/Illuminate/Database/Console/Migrations/MigrateCommand.php
```diff
@@ -134,26 +136,88 @@ protected function prepareDatabase()
...
-            if ($this->option('force')) {
-                return touch($e->getPrevious()->path);
-            }
+                $connection = $this->migrator->resolveConnection($this->option('database'));
 
-            if ($this->option('no-interaction')) {
+                if (
+                    $e->getPrevious() instanceof PDOException &&
+                    $e->getPrevious()->getCode() === 1049 &&
+                    $connection->getDriverName() === 'mysql') {
```
fgaroby, line 149: Why don't you create a `MySQLDatabaseDoesNotExistException`, instead of this `if` with three conditions (like `SQLiteDatabaseDoesNotExistException` on [L.140](https://github.com/laravel/framework/pull/44153/files#diff-28caf967243e756318d62361df72ec340f0b96c96f439c0bc800360733fee304R140))?

timacdonald:
If any failure occurs the original exception message would be shown.

amaelftah:
when i was moving from rails to learn laravel I was missing this suggestion ... thanks so much for adding it ❤️

KhalidEdaig:
It's finally come true, I love that, thanks laravels

PR #44155, opened 2022-09-16
[9.x] using better variable names & improve comments for request input methods

omalizadeh:
Request input checking methods accept multiple values & array, but variable name and php docs were not written accordingly.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -77,14 +77,14 @@ trait InteractsWithInput
     }
 
     /**
-     * Determine if the request contains a given input item key.
+     * Determine if the request contains given input item keys.
      *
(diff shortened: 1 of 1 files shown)


omalizadeh:
all of these methods have just one defined argument. why someone use named arguments for such methods?
besides, laravel docs says:

> [Named arguments](https://www.php.net/manual/en/functions.arguments.php#functions.named-arguments) are not covered by Laravel's backwards compatibility guidelines. We may choose to rename function arguments when necessary in order to improve the Laravel codebase. Therefore, using named arguments when calling Laravel methods should be done cautiously and with the understanding that the parameter names may change in the future.

@brunoalod

taylorotwell:

omalizadeh:
this is certainly a simple improvement to laravel codebase. is there a good reason to reject it? i explained named arguments problem already. we have also an option to merge this on master for that matter

@taylorotwell

Issue #44157, opened 2022-09-16
QueryBuilder ignores microseconds when processing DateTime arguments


paulandroshchuk:
- Laravel Version: 9.30.1
- PHP Version: 8.1.9
- Database Driver & Version: MySQL 8.0.30

### Description:

I came across this issue when I was writing tests for a "reporting" feature which goal was to count all records between two dates (on a high precision date time column). And I just accidentally noticed that not all records were taken into account so I started digging what could've went wrong and noticed an interesting article that explained how MySQL works with such columns: https://medium.com/bumble-tech/high-precision-time-how-to-use-fractional-seconds-in-mysql-and-php-fd9b89afe9ad

### Steps To Reproduce:
1. Create a table with [high precision timestamps](https://laravel.com/docs/9.x/migrations#column-method-timestamp)
2. [Set the corresponding date format](https://laravel.com/docs/9.x/eloquent#timestamps) in the model
3. Perform a (where/whereBetween) query using DateTime objects

```php
User::whereBetween('created_at', [
  today(),
  today()->endOfDay(),
])->count();

User::where('created_at', today()->endOfDay())->count();
```

| Created At  | Counted / Ignored |
| ------------- | ------------- |
| 2022-01-01 23.59.**58**.999999  | Counted  |
| 2022-01-01 23.59.**59**.111111  | Ignored  |
| 2022-01-01 23.59.**59**.999999  | Ignored  |

### Links
* Bug-report repository: https://github.com/paulandroshchuk/bug-report
* My attempt to fix this issue: https://github.com/laravel/framework/pull/44158

driesvints:
This is not needed. It's intended for microseconds not to be saved here.

paulandroshchuk:
> This is not needed. It's intended for microseconds not to be saved here.

@driesvints what's the reason to ignore microseconds here? Just curious..

driesvints:
Because it just doesn't works the same across all engines.

trevorgehman:
> Because it just doesn't works the same across all engines.

@driesvints For SQL Server the datetime format is overridden: 

https://github.com/laravel/framework/blob/74cad576dbc3f48c6f529950cee52bc948371f63/src/Illuminate/Database/Query/Grammars/SqlServerGrammar.php#L535-L543

[...]

PR #44158, opened 2022-09-16
[9.x] Update the format for database stored dates to include microseconds

paulandroshchuk:
This is an attempt to fix this issue: https://github.com/laravel/framework/issues/44157


driesvints:
This is a breaking change unfortunately. It's also intentionally that microseconds are not saved here.

PR #44160, opened 2022-09-16
[9.x] Require symfony/uid to generate Ulid's

vincentvankekerix:
The ```symfony/uid``` is required in order to use ```Str::ulid()```. Without it, an ```Class "Symfony\Component\Uid\Ulid" not found``` error gets triggered.


driesvints:
Please see the suggests list when installing dependencies.

PR #44161, opened 2022-09-16
[9.x] TestResponse new assertion about rendered views

nuernbergerA:
Currently, it's not possible to test if the response includes a specific view partial or not, as `assertSee*` is not the best fit for this case.

My Solution:
* Listen to all `composing` events (which are dispatched during the view rendering)
* Store the view names and count how often they were called

Possible problems:
* false positives due to `View` usage during the response from pdf/mail generation

Alternativ solution:
* use a rendering engine (during tests) to wrap blade directives and components with HTML comments and search for them in the actual rendered result


taylorotwell:
Hmm, not sure I want to go this route atm. 🤔

PR #44164, opened 2022-09-16
[9.x] Container - Set tag on object based on instanceof

wimulkeman:
Allow tagging an object based on its parent(s), interface(s) or its own FQCN.

This enables a developer to tag multiple objects dynamically at once without having to manually add new objects when they are added and extend a certain interface or parent.

Dynamically tagging can be an useful addition when you use a strategy pattern in your code and want, for example, register a new handler.

An developer can influence the behaviour of the objects being considered for tagging by filling the classmap property in the container before calling the tagInstanceof method.

Its usage would result in line with the following sample code.

```php
$this->app->tagInstanceof(BaseServiceTypeInterface::class, 'my-tag');

$this->app
    ->when(MyDynamicHandlerService::class)
    ->needs(BaseServiceTypeInterface::class)
    ->giveTagged('my-tag');

// ...

class MyDynamicHandlerService
{
    public function __construct(BaseServiceTypeInterface ...$handlers)
    {
        $this->handlers = $handlers;
    }

    public function handle(string $type, array $config)
    {
        foreach ($this->handlers as $handler) {
            if (! $handler->supports($type)) {
                continue;
            }

            return $handler->handle($config);
        }

        throw new \LogicException('Missing handler for type '.$type);
    }
}
```

This update has been inspired by the Symfony methology to tag services through an instanceof provided value. @see https://symfony.com/doc/current/service_container/tags.html#autoconfiguring-tags

Resolves #42499

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -158,7 +158,7 @@ class Container implements ArrayAccess, ContainerContract
     protected $afterResolvingCallbacks = [];
 
     /**
-     * The classes loaded from the composer classmapper
+     * The classes loaded from the composer classmapper.
      *
      * @var array<int, string>
      */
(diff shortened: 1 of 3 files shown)


wimulkeman:
That's correct

taylorotwell:
There are some things that make me a bit uneasy merging this for now (global `app` usage in container component) - however, it seems like this could be a package or a user land defined function since it just reads the class map and calls `tag` as necessary?

Thanks!

wimulkeman:
Do I understand userland  defined function correctly if you mean by that that it would be a custom definition in a project using the Laravel frmework, so not in the framework itself?

Could this be handy as a helper function in Laravel? Or maybe better suited in the Application layer because of the app usage?

PR #44168, opened 2022-09-16
[9.x] Use `match` expression for Email Validation drivers

sethsandaru:
Hi team,

Since Laravel requires PHP 8.0.2 and above, thus adding `match` expression here will provide higher readability, less lines & ` { } ` for Email Validation.

Thanks 😄


ankurk91:
Off the topic but It looks like `filter` and `filter_unicode` is not documented.

PR #44181, opened 2022-09-17
[9.x] Adds new method Arr::coalesce

oyepez003:
This small PR adds the **Arr::coalesce** method:  Return the first non-null value in the array value.

### How to use:

```php
use Illuminate\Support\Arr;

$result = Arr::coalesce([null, null, null, 'W3Schools.com', null, 'Example.com']);
//result = 'W3Schools.com'

// Use case
$name = Arr::coalesce([$user->nickname, $user->name, 'Taylor']);
```
The name **coalesce** was taken from the [MySQL function](https://www.w3schools.com/sql/func_mysql_coalesce.asp) but the name could be changed if necessary for one more appropriate.

The PR includes tests.

Thanks!.

### DIFF
--- a/tests/Support/SupportArrTest.php
+++ b/tests/Support/SupportArrTest.php
@@ -1116,8 +1116,11 @@ class SupportArrTest extends TestCase
     public function testCoalesce()
     {
         $this->assertEquals('W3Schools.com', Arr::coalesce([null, null, null, 'W3Schools.com', null, 'Example.com']));
-        $this->assertNotNull(Arr::coalesce([null, null, null, new \DateTime()]));
+        $this->assertEquals('foo', Arr::coalesce(['foo' => null, 'bar' => 'foo']));
+        $this->assertNull(Arr::coalesce(['foo' => null, 'bar' => null]));
         $this->assertNull(Arr::coalesce(null));
+        $this->assertNull(Arr::coalesce([null]));
         $this->assertNull(Arr::coalesce([]));
+        $this->assertNotNull(Arr::coalesce([null, null, null, new \DateTime()]));
     }
 }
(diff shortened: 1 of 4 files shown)


ankurk91:
No more global function please, can you move this method to `Arr` class instead?

oyepez003:
> No more global function please, can you move this method to `Arr` class instead?

Done @ankurk91

tests/Support/SupportArrTest.php
```diff
@@ -1112,4 +1112,12 @@ public function testPrependKeysWith()
             ],
         ], Arr::prependKeysWith($array, 'test.'));
     }
+
+    public function testCoalesce()
+    {
+        $this->assertEquals('W3Schools.com', Arr::coalesce([null, null, null, 'W3Schools.com', null, 'Example.com']));
+        $this->assertNotNull(Arr::coalesce([null, null, null, new \DateTime()]));
+        $this->assertNull(Arr::coalesce(null));
+        $this->assertNull(Arr::coalesce([]));
```
antonkomarev, line 1121: Add one test with array of one null value
`[null]`

And what will happen if string keyed array will be passed?

crynobone:
Why not just 

```php
$name =  $user->nickname ?? $user->name ?? 'Taylor';
```

bert-w:
> Why not just
> 
> ```
> $name =  $user->nickname ?? $user->name ?? 'Taylor';
> ```
> 
> Even using collection should be simple:
> 
> ```
> $result = collect([null, null, null, 'W3Schools.com', null, 'Example.com'])->filter()->first();
> ```
Nitpick but your example is 2n running time vs n running time:
```php
collect()->first(fn ($value) => ! is_null($value));
```

Idk if we need to add this function...

oyepez003:
Hi guys! thanks all of you for your feedback.

More tests were added, I just keep watching for any other feedback.

Regards!.

taylorotwell:

Issue #44159, opened 2022-09-16
Eloquent query builder does not work correct for morph to relation


siarheipashkevich:
- Laravel Version: 9.30.1
- PHP Version: 8.1.9
- Database Driver & Version: MySQL 8.0.29

### Description:

Hi all.

I'm using [Constraining Eager Loads With Relationship Existence](https://laravel.com/docs/9.x/eloquent-relationships#constraining-eager-loads-with-relationship-existence) but getting unexpected result. Please see my code:

In `Comment` model I have morph to relation:
```
public function commentable(): MorphTo
{
    return $this->morphTo();
}
```

`Country` model use `SoftDeletes` trait.

And test:
```
public function testWithWhereHas()
{
    $user = User::factory()->create();

    Comment::factory()
        ->for($user)
        ->for(Country::factory()->trashed(), 'commentable') // trashed country <--
        ->create();

    Comment::factory()
        ->for($user)
        ->for(Country::factory(), 'commentable')
        ->create();

    // I want to get all users which has comments with not deleted `commentable` relation, but I'm getting
    // comments with trashed commentable relation also. why?
    User::query()
        ->withWhereHas('comments.commentable')
        ->each(function (User $user) {
            // should be 1
            dump($user->comments->count());

            $user->comments->each(function (Comment $comment) {
                dump(['comment_id' => $comment->id, 'commentable_id' => $comment->commentable?->id]);
            });
        });
}
```
![image](https://user-images.githubusercontent.com/7753600/190609050-155636b1-f67d-43fe-873e-00c1e877f236.png)

Why `withWhereHas` method returns comment which has **trashed** `commentable` relation?

### Steps To Reproduce:

Please see description above.

driesvints:
Hey there,

Can you first please try one of the support channels below? If you can actually identify this as a bug, feel free to open up a new issue with a link to the original one and we'll gladly help you out.

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

Thanks!

siarheipashkevich:
@driesvints it's obviously a bug, I've described all steps how you can reproduce it and attach link to the documentation where described how it's should work.

[...]

PR #44189, opened 2022-09-18
[9.x] Constraint withWhereHas on nested relationships

cosmeoes:
Related to this issue: https://github.com/laravel/framework/issues/44159

Calling `withWherehas` with a nested relationship and without a callback, the method returns all related models from the parent relationship even if it has no children.

```php
SomeModel::withWhereHas('parent.child');
```
In that case, it will return only `SomeModel` where the `parent` has `child` but it would eagerly load every `parent` related to `SomeModel` not only those that have a `child`. 

Another less abstract example:
```php

    Comment::factory()
        ->for($user)
        ->for(Thread::factory()->trashed()) // soft deleted thread
        ->create();

    Comment::factory()
        ->for($user)
        ->for(Thread::factory())
        ->create();

    User::query()
        ->withWhereHas('comments.thread')
        ->get();
```

Here it would be expected for the query to return the `User` model with exactly one comment (the one that has the non-deleted thread). But, instead, we get the `User` with the two comments, one with the thread as `null`.

You can circumvent this by passing a callback that calls `withWhereHas` with the nested relationship:
```php
    User::query()
        ->withWhereHas('comments', function ($query) => {
            $query->withWhereHas('thread');
        })
        ->get();
```

With this pull request that is the generated query when calling `withWhereHas` with a nested relationship and no callback.

If you think the current behavior makes more sense feel free to close this pull request.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -165,8 +165,41 @@ trait QueriesRelationships
      */
     public function withWhereHas($relation, Closure $callback = null, $operator = '>=', $count = 1)
     {
+        $relations = explode('.', $relation);
+        if (! $callback && sizeof($relations) > 1) {
+            return $this->withWhereHasNested($this, $relations);
+        }
+
         return $this->whereHas($relation, $callback, $operator, $count)
             ->with($callback ? [$relation => fn ($query) => $callback($query)] : $relation);
+    }
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
withWhereHasNested has operator and count arguments which are never used?

siarheipashkevich:
@taylorotwell why these arguments is never used?
Please check the code.

We have a bug but you are closing the PR without any reasons.

robsontenorio:
That would be nice :(

PR #44190, opened 2022-09-18
[9.x] Can add `after()` method in policy

juliomotol:
This PR aims to allow to define a `after()` method in policies which will be called when the ability method returns `null`. Much like the `before()` method, the `after()` method wont be called if the ability method is not defined.


taylorotwell:

juliomotol:
Gotch, no worries, I understand. I hope to see this in a future release. Thanks!

PR #44194, opened 2022-09-18
[9.x] Refactor validation closures  to arrow functions

Nothehi:
Refactor `Validation` closures to arrow functions

### DIFF
--- a/src/Illuminate/Validation/Rules/DatabaseRule.php
+++ b/src/Illuminate/Validation/Rules/DatabaseRule.php
@@ -69,13 +69,10 @@ trait DatabaseRule
                 return $table;
             }
 
-            return implode(
-                '.',
-                array_map(
-                    fn (string $part) => trim($part, '.'),
-                    array_filter([$model->getConnectionName(), $model->getTable()])
-                )
-            );
+            return implode('.', array_map(
+                fn (string $part) => trim($part, '.'),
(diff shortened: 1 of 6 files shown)


henzeb:
Why?

Nothehi:
From PHP 7.4, We can make closure functions with the fat arrow which is now more readable, shorter, and cleaner. I thought it would be a good idea to convert these closures to arrow functions.

src/Illuminate/Validation/Rules/DatabaseRule.php
```diff
@@ -69,9 +69,13 @@ public function resolveTableName($table)
...
-            return implode('.', array_map(function (string $part) {
-                return trim($part, '.');
-            }, array_filter([$model->getConnectionName(), $model->getTable()])));
+            return implode(
+                '.',
+                array_map(
+                    fn (string $part) => trim($part, '.'),
+                    array_filter([$model->getConnectionName(), $model->getTable()])
+                )
+            );
```
driesvints, line 78:
```suggestion
            return implode('.', array_map(
                fn (string $part) => trim($part, '.'),
                array_filter([$model->getConnectionName(), $model->getTable()])
            ));
```

driesvints (changes requested):

taylorotwell:
It's OK.

PR #44196, opened 2022-09-18
[9.x] `InteractsWithDatabase::castAsJson($value)` incorrectly handles SQLite Database

### LINKED ISSUE #43863: [9.x] `InteractsWithDatabase::castAsJson($value)` incorrectly handles SQLite Database
- Laravel Version: 
- PHP Version: v9.26.0
- Database Driver & Version: MySQL 8.0 | SQLite

### Description:

When using SQLite to test locally, calling `InteractsWithDatabase::castAsJson` in a test incorrectly casts the value to a raw DB expression, containing `CAST($value AS JSON)`.

### Steps To Reproduce:

```php
// SomeTest extends TestCase

$this->assertDatabaseHas(TriggerListenerCondition::class, [
    'group' => 0,
    'operand' => 'and',
    'type' => 'array',
    'attribute' => 'call.answered_by',
    'operator' => 'has_any',
    'value' => $this->castAsJson([$user->id]),
]);
```

```
Failed asserting that a row in the table [trigger_listener_conditions] matches the attributes {
    "group": 0,
    "operand": "and",
    "type": "array",
    "attribute": "call.answered_by",
    "operator": "has_any",
    "value": "CAST('[1]' AS JSON)"
}.

Found similar results: [
    {
        "group": "0",
        "operand": "and",
        "type": "array",
        "attribute": "call.answered_by",
        "operator": "has_any",
        "value": "[1]"
    },
```

When running this with a MySQL connection, the assertion passes fine.

This causes issues since I run my tests locally under SQLite for speed, and then in GitHub Actions with MySQL when pushing to master.

---

Reference:

https://github.com/laravel/framework/blob/d23de5f9af1bd253a9d8b823cec9bf37987facf2/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php#L173-L184

Proposed solution:

```php
public function castAsJson($value)
{
    if ($value instanceof Jsonable) {
        $value = $value->toJson();
    } elseif (is_array($value) || is_object($value)) {
        $value = json_encode($value);
    }

    if (DB::connection() instanceof SQLiteConnection) {
        return $value;
    }

    $value = DB::connection()->getPdo()->quote($value);

    return DB::raw("CAST($value AS JSON)");
}
```

stevebauman:
Closes https://github.com/laravel/framework/issues/43863

This PR implements a `compileJsonCast()` method on the base query grammar instance to allow the transformation of a raw JSON string into a "cast" operation compatible with its database.

This new method is then called in the `castAsJson` test utility method to generate a proper raw query expression into its compatible operation. Previously, `castAsJson` would only work with MySQL, as other databases do not have the same operation and will either fail, or (in SQLite's case) simply query for the whole operation's text. 
 
I've tested these cast functions in where clauses using [SQL Fiddle](http://sqlfiddle.com). SQLite and Postgres do not appear to require cast functions to be called on the raw JSON to properly query.

@browner12 Let me know your thoughts on this implementation 👍


src/Illuminate/Database/Query/Grammars/MySqlGrammar.php
```diff
@@ -113,6 +113,17 @@ protected function compileJsonContainsKey($column)
...
 
+    /**
+     * Compile a "JSON cast" statement into SQL.
+     *
+     * @param  string  $value
+     * @return string
+     */
+    public function compileJsonCast($value)
+    {
+        return "cast($value as json)";
```
browner12, line 124: for code consistency, I would switch to single quotes and concatenation, rather than double quotes and interpolation.

src/Illuminate/Database/Query/Grammars/SqlServerGrammar.php
```diff
@@ -154,6 +154,17 @@ protected function compileJsonContains($column, $value)
...
 
+    /**
+     * Compile a "JSON cast" statement into SQL.
+     *
+     * @param  string  $value
+     * @return string
+     */
+    public function compileJsonCast($value)
+    {
+        return "json_query($value)";
```
browner12, line 165: for code consistency, I would switch to single quotes and concatenation, rather than double quotes and interpolation.

src/Illuminate/Database/Query/Grammars/MySqlGrammar.php
stevebauman: Gotcha, will do!

src/Illuminate/Database/Query/Grammars/SqlServerGrammar.php
stevebauman: Noted! Will update later tonight 👍

tests/Testing/Concerns/InteractsWithDatabaseTest.php
```diff
@@ -0,0 +1,145 @@
...
+    protected function tearDown(): void
+    {
+        m::close();
+    }
+
+    public function testCastToJsonSqlite()
+    {
+        $grammar = new SQLiteGrammar();
+
+        $this->assertEquals(<<<'TEXT'
```
browner12, line 34: is there a reason you're using Nowdocs everywhere in the tests? Could you just use a simple string here to make it more readable?

```php
$this->assertEquals('["foo", "bar"]', $this->castAsJson(['foo', 'bar'], $grammar));
```

browner12 (commented):
Same comment on all the Nowdocs in the test.

tests/Testing/Concerns/InteractsWithDatabaseTest.php
stevebauman: Shoot your right -- thanks @browner12, I'll update this 👍

stevebauman: Scratch that -- forgot these are intended, because the actual output of the string contains the quoted string. Using `$this->assertEquals('["foo", "bar"]', ...)` asserts that the output is `["foo", "bar"]`, not the fully quoted `'["foo", "bar"]'`.

browner12: gotcha, thanks for clarifying.

PR #44199, opened 2022-09-19
[9.x] Implement markdown as built-in model casting type

magdicom:
Hello,

as I used to write an attribute accessor whenever I needed to cast value as markdown, even years before `Str::markdown` was implemented, today I wrote this custom casting type, and with some effort I was able to include it as a native one.

it would be very helpful to simply write `$casts = ['column_name' => 'markdown']` and let laravel do the heavy task as usual.

+ tests included
+ I can write the documentation once approved
+ open to any modification request

### DIFF
--- a/src/Illuminate/Database/Eloquent/Casts/Markdown.php
+++ b/src/Illuminate/Database/Eloquent/Casts/Markdown.php
@@ -64,8 +64,19 @@ class Markdown implements Castable
                         return [];
                     }
 
-                    // Properly type cast option integer value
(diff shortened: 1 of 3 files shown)


crynobone:
Wouldn't this mean that each time you want to edit the attribute you need to convert the value from casted Markdown to raw Markdown?

magdicom:
> Wouldn't this mean that each time you want to edit the attribute you need to convert the value from casted Markdown to raw Markdown?

your point is correct, but there is few scenario where I would use it to display only.

taylorotwell:

PR #44200, opened 2022-09-19
[8.x] Reusable logging context into custom channels

s008nyx:
When initializing the logging channel, a new Logger constructor is called. This leads to the reset of the globally defined context and does not allow it to be thrown into the channel


driesvints:
8.x is closed for new features. Furthermore, this is a breaking change to the property.

PR #44201, opened 2022-09-19
[9.x] Return `true` from policy stub methods by default

danharrin:
This is just a proposal. I welcome discussion from others to decide if this would be a better default. I am aware that people can customise the stubs themselves if they wish.

Laravel's `make:policy` method currently generates empty stub files. For example:

```php
public function update(User $user, Post $post)
{
    //
}
```

However, this doesn't feel very "batteries included" to me.

Often, these policies are generated when the rest of the authorisation system has not yet been set up, and we then have to deliberately go through 7 separate methods in the file, to `return true`. If not, then any policy checks will fail, which is unexpected behaviour.

This is especially a "gotcha" when you use `make:model --all` which generates a policy amongst other files. People often forget about the policy though, and then are confused why their app isn't working. In particular, some packages like Nova and Filament which read policies will simply not show a resource at all, as the policy method exists but doesn't `return true`.

This PR simply changes the stub that is used to generate policies, so that all methods `return true`:

```php
public function update(User $user, Post $post)
{
    return true;
}
```

## The better (breaking change) alternative

Ideally, we'd keep the stubs as they are, and just tweak Laravel's authorisation system to allow access when a policy method returns `null` instead of `false`. Then people still clearly know which methods they haven't implemented in their policy, but aren't surprised by policy checks returning `false`. Of course though, this is a breaking change, so it's not gonna be merged.

### DIFF
--- a/src/Illuminate/Foundation/Console/stubs/policy.stub
+++ b/src/Illuminate/Foundation/Console/stubs/policy.stub
@@ -18,7 +18,7 @@ class {{ class }}
      */
     public function viewAny({{ user }} $user)
     {
-        //
+        return true;
     }
 
     /**
@@ -30,7 +30,7 @@ class {{ class }}
      */
     public function view({{ user }} $user, {{ model }} ${{ modelVariable }})
     {
-        //
+        return true;
     }
 
     /**
@@ -41,7 +41,7 @@ class {{ class }}
      */
     public function create({{ user }} $user)
     {
-        //
+        return true;
     }
 
     /**
@@ -53,7 +53,7 @@ class {{ class }}
      */
     public function update({{ user }} $user, {{ model }} ${{ modelVariable }})
     {
-        //
+        return true;
     }
 
     /**
@@ -65,7 +65,7 @@ class {{ class }}
      */
     public function delete({{ user }} $user, {{ model }} ${{ modelVariable }})
     {
-        //
+        return true;
     }
 
     /**
@@ -77,7 +77,7 @@ class {{ class }}
      */
     public function restore({{ user }} $user, {{ model }} ${{ modelVariable }})
     {
-        //
+        return true;
     }
 
     /**
@@ -89,6 +89,6 @@ class {{ class }}
      */
     public function forceDelete({{ user }} $user, {{ model }} ${{ modelVariable }})
     {
-        //
+        return true;
     }
 }


awcodes:
I totally understand how this can cause confusion for people, but I'd be hesitant to default anything around authorization to  'true'. I'd rather figure it out as developer as to why something isn't working, than to give user's access by accident.

If devs forget about the policy now, they will forget about it with this change too. Except defaulting to true in this case can have worse consequences. 

Just my 2 cents.

danharrin:
> I'd be hesitant to default anything around authorization to 'true'. I'd rather figure it out as developer as to why something isn't working, than to give user's access by accident.
> 
> If devs forget about the policy now, they will forget about it with this change too. Except defaulting to true in this case can have worse consequences.

Yes, that's a very valid argument. I'll run through what I considered before making this PR on that topic.

In my opinion, there are two cases whereby someone will generate a policy:

- The first is while they are setting up authorization in their app. They use `make:policy` and then immediately add permissions to the policy. They do not encounter this issue, as their policy is always correct. In this case, they aren't granting users access by accident.
- The second is while they are simply setting up their models. They use `make:model -all` as they need a migration, factory and seeder, but they forget about the policy that gets generated. In this case, they are setting up their authorization and app logic later, and will therefore revisit their policy files and add the correct checks in. Therefore, they aren't granting users access by accident either.

So personally I think this is a pretty safe change to make 🤷 Laravel is a sharp tool and people need to be careful with other things too.

Jubeki:
I also think it should be a "Deny by Default" instead of "Allow by Default".

See: https://owasp.org/Top10/A01_2021-Broken_Access_Control/

Instead you could also return an error that the authorisation has failed at x in a local context.

stevebauman:
Hmmm, maybe the stub methods should throw an exception by default to ensure the policy methods are touched by the developer?

```php
public function update(User $user, Post $post)
{
    throw new Exception('Unable to authorize.');
}
```

Thoughts?

danharrin:
> Hmmm, maybe the stub methods should throw an exception by default to ensure the policy methods are touched by the developer?

I like this, @stevebauman. What do you think @taylorotwell?

cheesegrits:
Personally I'd like to see an argument on make:policy like --default=true (or false), and if that isn't provided, then have the methods throw an exception saying "policy not yet configured".

Issue #44167, opened 2022-09-16
Str::ulid() not working


tanthammar:
There is something wrong with the Illuminate Support `symfony/uid` requirement.
I had to manually `composer require symfony/uid` to get `Str::ulid()` to work.

_Originally posted by @tanthammar in https://github.com/laravel/framework/pull/44074#discussion_r973140010_

Laravel v 9.30.1
```
Class "Symfony\Component\Uid\Ulid" not found
```

huangdijia:
composer r symfony/ulid

tanthammar:
@huangdijia Sorry, I don't understand your comment?

I did manually install the dependency with composer require, but what i want to say is that it should be required by Laravel Framwork in order for the Str::ulid() to work properly.

Or, do I misunderstand you?

Also see where I commented in the pr.

rodrigopedra:
The PR #44074 is unreleased. So its changes are not available yet.

Should be released on Tuesday as usually.

rodrigopedra:
`symfony/uid` is in the `suggests` session of `illuminate/support`'s `composer.json`. 

So to use its functionality you should manually install this dependency.

tanthammar:
@rodrigopedra 
Aha, then I get it :) 
Thought it was released with in Laravel 9.30.1, as the method Str::ulid() exists.

rodrigopedra:
No, I actually just saw they have released it after I sent my first answer (now deleted) about the release date. 

I didn't think they would have introduced a new public method on a patch release. As I had the same method macro'ed I was keeping an eye on the next minor release to test it out before updating.

The `Str::ulid()` is already available, and you can use it. 

But as some other optional features in the framework, you must install the suggested dependency manually if you plan on using this method. It is an opt-in feature.

This is similar on using S3 storage. All code is available in the framework, but you need to manually install AWS SDK compose package to use it.

So, to summarize:

- the PR is merged and released
- code for `Str::ulid()` is released and available to use
- To make use of it you need to opt-in by manually requiring the suggested `symfony/uid` package

Hope it is clearer now. Sorry for the confusion.

tanthammar:
Thank you @rodrigopedra 
Never came across a "Str" helper before that requires manually installing a package .
Being spoiled by the framework :) 

Anyway, it was easy to figure out, and I guess it will be added to the docs that it requires manual installation

DarkGhostHunter:
Added method to documentation, along with a not to require `symfony/ulid` so it can work.

[...]

PR #44202, opened 2022-09-19
[9.x] Require symfony/uid

driesvints:
I see people struggling with this so let's just require the package as it's not that large anyway. Still optional for `illuminate/support` like `ramsey/uuid` is.

https://github.com/laravel/framework/issues/44167


DarkGhostHunter:
Cool, I'll update the docs PR to remove the references since is now included in the framework.

PR #44208, opened 2022-09-19
Arr::get by sending keys as array

akshit-arora:
Current functionality of `Arr::get` supports only dot notation. when trying to fetch the value from the multidimensional array.

But it creates a little problem to create a dot notation string when we have to deal with dynamic keys

`$key = 'parent.'. $dynamicKey.'.final_key;`

It hardens the readability of the code with too many dots when using dynamic keys (one for concatenation of strings, another defining the depth of my key)

But it will be easier if we can pass the depth directly in as the key

`$key = ['parent', $dynamicKey, 'final_key'];`

and then running `Arr::get($myArray, $key, 'default');`


taylorotwell:

PR #44221, opened 2022-09-20
fixed: multiple route cache loading

ulcuber:
Used this:
```php
protected function loadCachedRoutes()
    {
            $key = 'Routes: ' . static::class;
            \BeyondCode\ServerTiming\Facades\ServerTiming::start($key);
            require $this->app->getCachedRoutesPath();
            \BeyondCode\ServerTiming\Facades\ServerTiming::stop($key);
        });
    }
```
And discovered that some packages use extended `RouteServiceProvider` therefore route cache included multiple times:
image

Since route cache uses heavy function calls better to include it only once.


taylorotwell:
Need more info here - also worried this would break Octane or other environments in some scenarios.

PR #44222, opened 2022-09-20
flushable routes cache

ulcuber:
Prevents multiple loading when application has multiple `RouteServiceProvider` classes.

Related with #44221

@taylorotwell what about this approach? This way should be possible to reload routes cache.

### DIFF
--- a/src/Illuminate/Contracts/Foundation/CachesRoutes.php
+++ b/src/Illuminate/Contracts/Foundation/CachesRoutes.php
@@ -17,4 +17,18 @@ interface CachesRoutes
      * @return string
      */
     public function getCachedRoutesPath();
+
+    /**
+     * Check if the routes cache file loaded.
+     *
+     * @return bool
+     */
+    public function isCachedRoutesLoaded();
+
+    /**
+     * Load the routes cache file.
+     *
+     * @return void
+     */
+    public function loadCachedRoutes();
 }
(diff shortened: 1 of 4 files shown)


taylorotwell:
I can't really add to contracts on a patch release - I'm not sure I even totally understand the problem you are describing. Can you give example packages, etc.

ulcuber:
@taylorotwell for example:
https://github.com/area17/twill/blob/2.x/src/RouteServiceProvider.php
it extends from `Illuminate\Foundation\Support\Providers\RouteServiceProvider`
so `register` is called twice: https://github.com/laravel/framework/blob/1d8e86193fd1740606a836f25043e84fe78c562d/src/Illuminate/Foundation/Support/Providers/RouteServiceProvider.php#L37
then exists two boot callback for routes cache:
https://github.com/laravel/framework/blob/1d8e86193fd1740606a836f25043e84fe78c562d/src/Illuminate/Foundation/Support/Providers/RouteServiceProvider.php#L108

ulcuber:
Now I was trying another workaround:
```php
protected function loadCachedRoutes()
{
        if ($this->app['router']->getRoutes()->count() === 0) {
            require $this->app->getCachedRoutesPath();
        }
    });
}
```
then encountered this:
https://github.com/Intervention/image/pull/1184
Not every package/application respects route caching to use this approach.

PR #44226, opened 2022-09-21
[9.x] Adds Eloquent User Provider query handler

DarkGhostHunter:
## What

Allows the user to register a query handler for retrieving the user from the `EloquentUserProvider`. This allows to tap into the query and modify columns to retrieve, like bypassing columns that may have too much data or are unused.

```php
Auth::guard('web')->getProvider()->setQueryHandler(function ($query) {
    // ...
});
```

Since it works a `newModelQuery()` level, any change be overridden through callbacks when retrieving the user.

```php
Auth::guard('web')->attempt([
    'email' => $email,
    'password' => $password,
    fn($query) => ...
});
```

Also the `setQueryHandler()` can be used to erase it with `null` and it can be retrieved for further inspection (or even wrapping).

### DIFF
--- a/src/Illuminate/Auth/EloquentUserProvider.php
+++ b/src/Illuminate/Auth/EloquentUserProvider.php
@@ -25,11 +25,11 @@ class EloquentUserProvider implements UserProvider
     protected $model;
 
     /**
-     * The callback to modify the query to find the user.
+     * The callback that may modify the user retrieval queries.
      *
      * @var (\Closure(\Illuminate\Database\Eloquent\Builder):mixed)|null
      */
-    protected $queryHandler;
+    protected $queryCallback;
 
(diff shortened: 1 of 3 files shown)


taylorotwell:
Why did you need this in your application?

DarkGhostHunter:
> Why did you need this in your application?

I cannot alter which columns to retrieve from the table, or even use eloquent relations, without also altering all other queries in the app.

For example, I hooked a Laravel app into an existing DB where someone decided it was a good idea to have an embedded image in a column `users` table. 

Since I cannot remove it, or move it, my only chance is to tell the User Provider to not load a column that contains 2~5 mb of data.

sabanworld:
Much appreciated!

PR #44227, opened 2022-09-21
[9.x] Share WithoutOverlapping key across jobs

timacdonald:
This PR allows `WithoutOverlapping` to apply across different job classes. Currently this feature only allows without overlapping to be applied to instances of a single class.

```php
use Illuminate\Queue\Middleware\WithoutOverlapping;

class ProviderIsDown
{
    // ...


    public function middleware()
    {
        return [
            (new WithoutOverlapping("provider-status:{$this->provider}"))->shareKey(),
        ];
    }
}

class ProviderIsUp
{
    // ...


    public function middleware()
    {
        return [
            (new WithoutOverlapping("provider-status:{$this->provider}"))->shareKey(),
        ];
    }
}
```

These two jobs will now not overlap even though they are different classes. They both utilise the same lock key (assuming the provider is the same) `"provider-status:{providerName}".

Thanks @andrewbroberg for raising and chatting to me about this.

Documentation PR: https://github.com/laravel/docs/pull/8240

### DIFF
--- a/src/Illuminate/Queue/Middleware/WithoutOverlapping.php
+++ b/src/Illuminate/Queue/Middleware/WithoutOverlapping.php
@@ -39,6 +39,13 @@ class WithoutOverlapping
     public $prefix = 'laravel-queue-overlap:';
 
     /**
+     * Share the key across different jobs.
+     *
+     * @var bool
+     */
+    public $shareKey = false;
+
+    /**
      * Create a new middleware instance.
      *
      * @param  string  $key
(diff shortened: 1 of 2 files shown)


rihardssceredins:
Hah, I ran into the same constraint earlier last week and wanted to contribute but you beat me to it. 😄
Thanks!

timacdonald:
No troubles! Glad to hear it will be useful :)

PR #44228, opened 2022-09-21
[fix]: swap redis zrevrangebyscore max and min position

var618:


src/Illuminate/Redis/Connections/PhpRedisConnection.php
```diff
@@ -260,7 +260,7 @@ public function zrangebyscore($key, $min, $max, $options = [])
      * @param  array  $options
      * @return array
      */
-    public function zrevrangebyscore($key, $min, $max, $options = [])
+    public function zrevrangebyscore($key, $max, $min, $options = [])
     {
```
BrandonSurowiec, line 264: Breaking change to modify the method signature. Would revert and only modify the `command` call order.

PR #44229, opened 2022-09-21
[8.x] Fix: Swap Redis zrevrangebyscore min and max position

var618:

### DIFF
--- a/src/Illuminate/Redis/Connections/PhpRedisConnection.php
+++ b/src/Illuminate/Redis/Connections/PhpRedisConnection.php
@@ -255,13 +255,16 @@ class PhpRedisConnection extends Connection implements ConnectionContract
     /**
      * Return elements with score between $min and $max.
      *
+     * [Since Redis 2.2.0]
+     * ZREVRANGEBYSCORE key max min [WITHSCORES] [LIMIT offset count]
+     *
      * @param  string  $key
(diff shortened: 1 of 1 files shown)


src/Illuminate/Redis/Connections/PhpRedisConnection.php
```diff
@@ -255,13 +255,16 @@ public function zrangebyscore($key, $min, $max, $options = [])
...
+     * ZREVRANGEBYSCORE key max min [WITHSCORES] [LIMIT offset count]
+     *
      * @param  string  $key
      * @param  mixed  $min
      * @param  mixed  $max
      * @param  array  $options
      * @return array
      */
-    public function zrevrangebyscore($key, $min, $max, $options = [])
+    public function zrevrangebyscore($key, $max, $min, $options = [])
```
BrandonSurowiec, line 267: Breaking change to modify the method signature. Would revert and only modify the `command` call order.

var618: I try to only swap max and min params in `this->command()` method. 
But the GitHub Action tests failed.
Then I check the code of **5.x, 7.x, 8.x**, they are the same problem.
So, It is not a bug. 
The developer should take care of this, otherwise the return result would be empty array.

PR #44232, opened 2022-09-21
[9.x] feat: use model values in upsert

jaulz:
Sometimes you already know the ID of the model but don't know whether it exists in the database or not. This is what `upsert` was designed for but right now it only allows explicit values. This PR uses the dirty values and the key name of the model in case the `upsert` method was called on the instance of a model.

If you think this is a good idea, I am happy to create the relevant tests as well.


taylorotwell:

PR #44239, opened 2022-09-21
[9.x] Add methods to get request data as integer or float

jasonmccreary:
This adds methods to conveniently retrieve and cast request data as an `int` and `float`. A `bool` cast already existed. So why not have other common scalar types.

**Before**
```php
intval($request->input('some_int_value'));
floatval($request->input('some_float_value'));
```

**After**
```php
$request->integer('some_int_value');
$request->float('some_float_value');
```

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -353,6 +353,8 @@ trait InteractsWithInput
      * @param  string|null  $format
      * @param  string|null  $tz
      * @return \Illuminate\Support\Carbon|null
+     *
+     * @throws \Carbon\Exceptions\InvalidFormatException
      */
     public function date($key, $format = null, $tz = null)
     {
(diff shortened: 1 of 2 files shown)


jasonmccreary:
Something that came up in my live stream and separately on Twitter is the integer conversion of `'2_000'` to `2`. While a literal `2_000` is equivalent to `2000` in PHP, the string `2_000` is not.

Someone is welcome to support this edge case, but I felt it better to align with PHP. There's an argument for PHP to update the behavior of `intval()`. In which case, this would just work.

ejunker:
@jasonmccreary why not use casting instead of `intval()` and `floatval`? According to [this](https://tonyshowoff.com/articles/casting-int-faster-than-intval-in-php/) it is 3 times faster.

jasonmccreary:
Clarity and consistency with other parts of the framework. However, there are instances of both types of casting. So feel free to open a follow-up PR. 👍

PR #44241, opened 2022-09-21
[9.x] fix TrustHost middleware

esolo:
Fixes returning an array of hosts. Without this fix, the single host returns unchanged. Then all subdomains are automatically subject to returns. Without a fix, the shouldSpecifyTrustedHosts() method makes no sense.


taylorotwell:
Tests?

Issue #44212, opened 2022-09-19
`model:show` does not honor composer autoload


CalebDW:
- Laravel Version: 9.30.1
- PHP Version: 8.1.9
- Database Driver & Version: pgsql 14

### Description:
First off, thanks @jessarcher for your work on the model show command! It's very convenient and I expect to get a lot of mileage from it.

`model:show` seems to only search for models in the `App\` directory (unless there's a config variable I missed) instead of honoring composer's autoload. I'm not sure if this is the desired behaviour or not (for instance, Tinker will auto-alias a model found in autoload), but my application is organized by (sub)domains and it quickly becomes inconvenient to type out the full namespace that's several levels deep.

If this is not a bug, would you please consider honoring autoload (like Tinker) for convenience? 

Thanks!

### Steps To Reproduce:
Try to list the details for any model not in `App\` but available through composer autoloading.
`artisan model:show <Model>`

driesvints:
Hi there. Right now this is indeed limited to the App namespace. Feel free to attempt a PR to change that.

jbrooksuk:
It's worth noting here that you may pass through the entire namespace, for example, `artisan model:show Laravel\\Cashier\\Subscription`

PR #44243, opened 2022-09-21
[9.x] Add composer autoload and wildcard support to `model:show`

calebdw:
This PR adds composer autoload and wildcard support to `model:show` as suggested by @driesvints in #44212.

### Previous Behaviour
`model:show` only qualifies model in the `App\` (or `App\Models\` if exists) directory. It is possible to pass the full model namespace but this quickly becomes inconvenient as the number of sub-namespace levels increase (i.e., `App\Foo\Bar\Baz\...\Model`).

### Current Behaviour
`model:show` now uses the composer autoload classmap (adapted from `Laravel\Tinker\ClassAliasAutoLoader`) as well as wildcard matching to qualify model names. In addition, when multiple models are matched the user is presented with the choices and asked to select the desired model.

This allows the user to more quickly select and view the details of a desired model, especially when dealing with models several levels deep.

#### Examples
```php
// matches model basenames that equal `FooBar`
//   - App\Models\FooBar
//   - App\Baz\FooBar
model:show FooBar

// matches model basenames that start with `FooBar`
//   - App\Models\FooBarFlight
//   - App\Baz\FooBarTicket
model:show FooBar*

// matches model basenames that contain `FooBar`
//   - App\Models\PlaneFooBar
//   - App\Baz\UserFooBarTicket
model:show *FooBar*

// matches models that end in `Bar\FooBar`
//   - App\Foo\Bar\FooBar
model:show Bar\\FooBar

// matches model basenames that start with `FooBar` in 
// sub-namespaces that contain `Bar`
//   - App\Foo\Bar\Baz\FooBarFlight
model:show Bar*\\FooBar*

// matches model basenames === `FooBar` in 
// sub-namespaces that contain `Bar\Very`
//   - App\Foo\Bar\VeryLongName\FooBar
model:show Bar\\Very*\\FooBar
```

### DIFF
--- a/src/Illuminate/Foundation/Console/ShowModelCommand.php
+++ b/src/Illuminate/Foundation/Console/ShowModelCommand.php
@@ -6,10 +6,11 @@ use BackedEnum;
 use Doctrine\DBAL\Schema\Column;
 use Doctrine\DBAL\Schema\Index;
 use Doctrine\DBAL\Types\DecimalType;
-use Illuminate\Console\Command;
 use Illuminate\Contracts\Container\BindingResolutionException;
 use Illuminate\Database\Console\DatabaseInspectionCommand;
+use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Relation;
+use Illuminate\Support\Env;
 use Illuminate\Support\Str;
 use ReflectionClass;
 use ReflectionMethod;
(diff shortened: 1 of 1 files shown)


taylorotwell:

CalebDW:
@taylorotwell,

I'm sorry to see this closed, could you possibly reconsider? If you try the new `model:show` I think you'll like it---everyone I've shown it to loves the new search capabilities.

@jessarcher adapted the `ShowModel` qualification code from `GeneratorCommand` which makes sense when *creating* a model (i.e., place the new class in a default location if a full path is not given), however this does not make sense when *searching* for a model---any application of sufficient complexity would (hopefully) not have all of their models in a default location but rather organized throughout a directory structure. In my particular case, none of my models reside in the default search location (`App\` or `App\Models`) and it became very tedious to type the full path for every model I wanted to see (often several levels deep).

This PR just tries to make `ShowModel` a bit more usable / user friendly and I think many developers will appreciate it.

Thanks!

PR #44245, opened 2022-09-22
[9.x] Fixes order of migrations in StatusCommand

brunoalod:
The `migrate:status` command doesn't sort the migrations by batch when outputting to console.

I've noticed this issue by running an old migration I forgot to migrate after having ran other migrations.

This can be easily reproduced running the following commands on a fresh Laravel installation.

```
// Commands
php artisan migrate --path=/database/migrations/2014_10_12_100000_create_password_resets_table.php
php artisan migrate --path=/database/migrations/2014_10_12_000000_create_users_table.php 
php artisan migrate --path=/database/migrations/2019_08_19_000000_create_failed_jobs_table.php
php artisan migrate:status
```

```
// Expected
2014_10_12_100000_create_password_resets_table ... [1] Ran
2014_10_12_000000_create_users_table ............. [2] Ran
2019_08_19_000000_create_failed_jobs_table ....... [3] Ran
```

```
// Actual
2014_10_12_000000_create_users_table ............. [2] Ran
2014_10_12_100000_create_password_resets_table ... [1] Ran  
2019_08_19_000000_create_failed_jobs_table ....... [3] Ran
```

Note the batch numbers are not sorted. 

This actually can happen quite often when working with a team on different branches.

Hopefully this PR fixes it 🙂

### DIFF
--- a/src/Illuminate/Database/Console/Migrations/StatusCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/StatusCommand.php
@@ -85,20 +85,17 @@ class StatusCommand extends BaseCommand
      */
     protected function getStatusFor(array $ran, array $batches)
     {
-        return Collection::make($this->getAllMigrationFiles())
-                    ->map(function ($migration) use ($ran, $batches) {
-                        $migrationName = $this->migrator->getMigrationName($migration);
-
-                        $status = in_array($migrationName, $ran)
-                            ? '<fg=green;options=bold>Ran</>'
-                            : '<fg=yellow;options=bold>Pending</>';
-
-                        if (in_array($migrationName, $ran)) {
-                            $status = '['.$batches[$migrationName].'] '.$status;
-                        }
-
-                        return [$migrationName, $status];
-                    });
+        return Collection::make($batches)
+            ->map(function ($batch, $migration) {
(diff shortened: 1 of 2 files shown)


fgaroby:
Why do you think the migration `2014_10_12_100000_xxx` would be run before the migration `2014_10_12_000000_yyy`?

brunoalod:
> Why do you think the migration `2014_10_12_100000_xxx` would be run before the migration `2014_10_12_000000_yyy`?

You can create a migration file and run it at later date.

brunoalod:
> Yes: by renaming it.

I'm not sure that would be a good idea. 
When working with different branches and tons of migrations this issue can happen quite often.

Also, migrations already migrated would still be shown in the incorrect order.

dennisprudlo:
> Yes: by renaming it.

@fgaroby It is common that migrations are not always executed in order when multiple devs are working on a project. Image two dev branch off of the main branch with "feature-1" and "feature-2". Migrations are being created on feature-1 first, then feature-2. However, the feature-2 branch gets merged and deployed first.

However I don't think this command should order the migrations by their batch number. The command lists all existing migrations ordered by how they are stored in the migrations folder with some meta information about them (like batch number and if they have been run). It is not a "Migration History" command. IMO, if at all this should be an optional flag to allow sorting by batch number.

taylorotwell:

PR #44250, opened 2022-09-22
[9.x] Added `events` and `views` to `optimize` command

siarheipashkevich:
Added `events` and `views` to `optimize` command


taylorotwell:
It's a breaking change for existing applications.

PR #44251, opened 2022-09-22
[9.x] Fix bug in Fluent Class with named arguments in migrations

sneycampos:
When used named arguments on migration methods, e.g: `$table->boolean('isActive')->default(value: true); `the Fluent Class throws a Exception with "Undefined array key"

image

This fixes the way to get the first value on $parameter from array index to reset()


driesvints:
Why would you use a named argument there?

sneycampos:
why not?

driesvints:
It's a one argument method. There is no point?

PR #44252, opened 2022-09-22
[9.x] Adds `Benchmark::measure` ⏱

nunomaduro:
This pull request introduces a new `Benchmark::measure` facade that allows you quickly measure and compare Laravel code performance. Here is an example:

```php
use Illuminate\Support\Facades\Benchmark;

Benchmark::measure([
    fn () => User::count(),
    fn () => User::all()->count(),
]);
```

Using the following code, and if the application is running on the console, Artisan will output the following content:
console

As you can see, this report shows the elapsed real time, and thehe unit of time is miliseconds. In addition, Ten "repeats" were used by default. Meaning that both the given callbacks were executed ten times, and the value on the right is the average time elapsed of those ten executions.

Also, the fastest callback gets highlighted by using the green color. And, the left content contains an "description" based on the given closure's code.

Next, besides passing a list of callbacks, you can also pass a single callback. And of course, you can equally specify the number of desired repetitons:

```php
Route::get('/', function () {
    Benchmark::repeat(100)->measure(function () {
        $user = User::first();

        $team = $user->team;

        // More code...
    });
});
```

When visiting this route, the output will be rendered on the browser like so:
http

---

**As future scope**, would be interesting to provide certain methods, in special places of the framework, so we can easily benchmark routes, commands, and other framework resources without having to put the entire code of those resources within a closure for the `Benchmark::measure` method. In concrete terms, this would mean something like this:

```php
// Not this...
Route::get('/', function () {
    Benchmark::measure(function () {
        // My code ...
    });
});

// But this...
Route::get('/', function () {
    // ...
})->benchmark();
```

### DIFF
--- a/src/Illuminate/Console/View/Components/Factory.php
+++ b/src/Illuminate/Console/View/Components/Factory.php
@@ -12,6 +12,7 @@ use InvalidArgumentException;
  * @method void error(string $string, int $verbosity = \Symfony\Component\Console\Output\OutputInterface::VERBOSITY_NORMAL)
  * @method void info(string $string, int $verbosity = \Symfony\Component\Console\Output\OutputInterface::VERBOSITY_NORMAL)
  * @method void line(string $style, string $string, int $verbosity = \Symfony\Component\Console\Output\OutputInterface::VERBOSITY_NORMAL)
+ * @method void newLine(int $verbosity = \Symfony\Component\Console\Output\OutputInterface::VERBOSITY_NORMAL)
  * @method void task(string $description, ?callable $task = null, int $verbosity = \Symfony\Component\Console\Output\OutputInterface::VERBOSITY_NORMAL)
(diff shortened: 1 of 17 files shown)


Valence-Kajuna:
This is awesome @nunomaduro  🔥👏. I can't wait to use this feature to measure and compare db queries in terms of time used to retrieve data.

dennisprudlo:
This is perfect when trying to show laravel newcomers the difference between those database calls. Whether its 0.8ms or 51.6ms you doesn't really *feel* the difference. But showing `User::all()->count()` takes about 65 times longer than `User::count()` hits very differently 😅

fgilio:
This is amazing, I've done similar things in the past but always had to code it again

taylorotwell:
Discussed with @nunomaduro I want to keep this a bit simpler and lighter for now.

https://github.com/laravel/framework/commit/b4293d7c18b08b363ac0af64ec04fb1d559b4698

Can expand on it later if necessary.

fgilio:
There could be an extended version as a package laravel-benchmark

foremtehan:
Yeah, like: composer require lib/laravel-benchmark --dev

PR #44254, opened 2022-09-22
[9.x] Allow signed URLs with custom key resolver

bonroyage:
In case there are multiple applications communicating together, you don't want to share the same encryption key throughout all applications, but you may have a different pre-shared secret to sign URLs with. With this new `withKeyResolver` method on the UrlGenerator, a clone of the current UrlGenerator will be created where a different key can be used to validate the signature or generate signed URLs 

If you were to call `url()->setKeyResolver()`, then this will change the key for the remainder of the request, potentially generating incorrect signed URLs where this custom key wasn't needed.

Two very basic examples of how it could be used:

It can be used in middleware to validate with custom key:
```php
public function handle(Request $request, Closure $next)
{
    $urlGenerator = url()->withKeyResolver(fn() => config('services.my-service.key'));
    
    if(!$urlGenerator->hasValidSignature($request)) {
       throw new InvalidSignatureException;
    }
   
    return $next($request);
}
```

It can be used in a controller to generate a URL with custom key:
```php
public function __invoke()
{
    $urlGenerator = url()->withKeyResolver(fn() => config('services.my-service.key'));

    return response()->json([
        'return_url' => $urlGenerator->signedRoute('test')
    ]);
}
```

### DIFF
--- a/tests/Routing/RoutingUrlGeneratorTest.php
+++ b/tests/Routing/RoutingUrlGeneratorTest.php
@@ -867,6 +867,41 @@ class RoutingUrlGeneratorTest extends TestCase
 
         $this->assertSame('http://www.foo.com/foo/fruits', $url->route('foo.bar', CategoryBackedEnum::Fruits));
     }
+
+    public function testSignedUrlWithKeyResolver()
+    {
+        $url = new UrlGenerator(
+            $routes = new RouteCollection,
+            $request = Request::create('http://www.foo.com/')
+        );
+        $url->setKeyResolver(function () {
+            return 'secret';
+        });
+
+        $route = new Route(['GET'], 'foo', ['as' => 'foo', function () {
+            //
+        }]);
+        $routes->add($route);
+
+        $request = Request::create($url->signedRoute('foo'));
+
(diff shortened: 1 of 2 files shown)


tests/Routing/RoutingUrlGeneratorTest.php
```diff
@@ -867,6 +867,41 @@ public function testRouteGenerationWithBackedEnums()
...
+        $route = new Route(['GET'], 'foo', ['as' => 'foo', function () {
+            //
+        }]);
+        $routes->add($route);
+
+        $request = Request::create($url->signedRoute('foo'));
+
+        $this->assertTrue($url->hasValidSignature($request));
+
+        $request = Request::create($url->signedRoute('foo').'?tempered=true');
```
vr00, line 890: I think there is a typo, it should probably say `tampered`

driesvints: Can you send in a PR?

Krisell: PR for fixing the typo: https://github.com/laravel/framework/pull/51238

Issue #44249, opened 2022-09-22
Enum route binding: Default values not working


levu42:
- Laravel Version: 9.31.0
- PHP Version: 8.1.10
- Database Driver & Version: irrelevant

### Description:

Binding enums to route parameters works fine (yay!) but it doesn't work with default values given in the controller function signature:

`public function index(PostsStatus $status = PostsStatus::published)`. See the steps to reproduce below.

I found a (not-nice) workaround, by doing `app()->bind(PostsStatus::class, fn() => PostsStatus::published)` - but this is then a global default for the whole app instead of a default parameter for a route.

### Steps To Reproduce:

When creating a route with enum route binding ([see example repository](https://github.com/levu42/enum-route-binding-bug-report/commit/6372f55aee727b39dc8d5f501881f177fc55c2ec)):

```php
// routes/web.php
Route::get('/posts/{status?}', [PostsController::class, 'index']);

//app/Controllers/PostsController.php
public function index(PostTypeEnum $status = PostTypeEnum::Published) {}
```

It works fine with `/posts/drafts` - `$status` is correctly `PostTypeEnum::Drafts`. But when I try to call `/posts/`, it gives me an `TypeError` with the message

> App\Http\Controllers\PostsController::index(): Argument #1 ($status) must be of type App\PostTypeEnum, null given, ...

gthedev:
I am getting the same result as well.

In `src/Illuminate/Routing/RouteDependencyResolverTrait.php` on line `80` you can see it resolving the value to `null` if default value is is available. Replacing that `null` with `$parameter->getDefaultValue()` fixes this issue, however, might have side-effects in Model Route Binding as I haven't tested that yet.

driesvints:
Hi there. This is unsupported at the moment. Feel free to attempt a PR but indeed keep in mind other route model binding situations like @gthedev mentioned.

levu42:
@driesvints I attempted a PR in #44255 that shouldn't have any influence on route model binding.

PR #44255, opened 2022-09-22
[9.x] Allow enum route bindings to have default values

levu42:
It's possible already to bind route parameters to enums:

```php
enum Status: string {
  case Published = 'published';
  case Draft = 'draft';
}

Route::get('/posts/{status}', function(Status $status) {});
```

However, it's not yet possible to have default values for optional parameters, as I discovered in #44249:

```php
Route::get('/posts/{status?}', function(Status $status = Status::Published) {});
```

As discussed in #44249, I made the PR in a way, that it only applies, if the parameter is an enum, so it doesn't interfere with model binding in any way.

I'm not fully happy with the nested ternary operator, making it a bit more complex - I'm happy to change it so some other structure if you think something else is more readable.

I tried to keep the code change minimal to support this while having extensive tests, for them I looked at the existing backed enum binding tests and extended those.

### DIFF
--- a/src/Illuminate/Routing/RouteDependencyResolverTrait.php
+++ b/src/Illuminate/Routing/RouteDependencyResolverTrait.php
@@ -78,8 +78,10 @@ trait RouteDependencyResolverTrait
         // the list of parameters. If it is we will just skip it as it is probably a model
         // binding and we do not want to mess with those; otherwise, we resolve it here.
         if ($className && ! $this->alreadyInParameters($className, $parameters)) {
+            $isEnum = method_exists(ReflectionClass::class, 'isEnum') ? (new ReflectionClass($className))->isEnum() : false;
+            
             return $parameter->isDefaultValueAvailable()
-                ? ((new ReflectionClass($className))->isEnum() ? $parameter->getDefaultValue() : null)
+                ? ($isEnum ? $parameter->getDefaultValue() : null)
                 : $this->container->make($className);
         }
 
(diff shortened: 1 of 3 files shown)


src/Illuminate/Routing/RouteDependencyResolverTrait.php
```diff
@@ -77,7 +78,11 @@ protected function transformDependency(ReflectionParameter $parameter, $paramete
         // the list of parameters. If it is we will just skip it as it is probably a model
         // binding and we do not want to mess with those; otherwise, we resolve it here.
         if ($className && ! $this->alreadyInParameters($className, $parameters)) {
-            return $parameter->isDefaultValueAvailable() ? null : $this->container->make($className);
+            $isEnum = method_exists(ReflectionClass::class, 'isEnum') ? (new ReflectionClass($className))->isEnum() : false;
+
```
BrandonSurowiec, line 82: You could tighten it up with:

```php
$isEnum = method_exists(ReflectionClass::class, 'isEnum') && (new ReflectionClass($className))->isEnum();
```

I think your nested ternary is understandable as-is.

levu42: @BrandonSurowiec Thanks for the idea, just changed it!

BrandonSurowiec (commented):
Looks good to me!

taylorotwell:
Thanks 👍

PR #44261, opened 2022-09-22
[9.x] Add deferrable option to make:provider artisan command

salkuadrat:
Request to add deferrable option to `make:provider` artisan command for generating a deferrable Service Provider class.


taylorotwell:
I'm not sure we need a separate stub for this.

PR #44263, opened 2022-09-22
Feature: Preventing Missing Translation Keys Flag

yob-yob:
## Description

Feature is inspired by `Model::preventLazyLoading()`

This will benefit the end user when trying to develop locally and makes it easier when trying to figure out which translations are missing.

## How to use

on your `AppServiceProvider.php`

```php
use Illuminate\Translation\Translator;

public function boot()
{
   Translator::preventMissingTranslationKey(! app()->isProduction());
}
```

you can also disallow fallbacks...

> Normal Flow: if key does not exists in main locale it will try to look for the same key in your fallback locale

disallowing fallback will disrupt this flow:

> Disruptive Flow: if key does not exists in main locale, it will throw an error, skipping fallback checks

### How to disallow fallback

```php
use Illuminate\Translation\Translator;

public function boot()
{
   Translator::preventMissingTranslationKey(
       value: ! app()->isProduction(), 
       allow_fallback: false
   );
}
```

You can also disallow replace...

> Normal Flow: if `__("foo :message", ['message' => 'bar'])` does not exists in main json locale it will try to make replacements directly from the `key` string

disallowing replace will disrupt this flow:

> Disruptive Flow: if `__("foo :message", ['message' => 'bar'])` does not exists in main json locale it will throw an error, skipping replacements..

### How to disallow replacements

```php
use Illuminate\Translation\Translator;

public function boot()
{
   Translator::preventMissingTranslationKey(
       value: ! app()->isProduction(), 
       allow_fallback: false,
       allow_replace: false
   );
}
```

### DIFF
--- a/src/Illuminate/Translation/MissingTranslationKeyViolationException.php
+++ b/src/Illuminate/Translation/MissingTranslationKeyViolationException.php
@@ -7,11 +7,18 @@ use RuntimeException;
 class MissingTranslationKeyViolationException extends RuntimeException
 {
     /**
-     * The name of the affected Eloquent model.
+     * The Key of the missing translation.
      *
      * @var string
      */
     public $key;
+
+    /**
+     * The default locale that was used.
+     *
+     * @var array
+     */
+    public $locales;
 
     /**
      * Create a new exception instance.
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #44264, opened 2022-09-22
Keep url query string clean when user visits first page again

vmiguellima:
Keep the query string clean when a user navigates to first page using the pagination.

**Problem:** when navigating using pagination, when a user goes back to the first page the url keeps the query string `?page=1`.

**Goal:**  Keep it clean like: `https://blog.laravel.com/posts`, instead of `https://blog.laravel.com/posts/?page=1`.


ankurk91:
It can be a breaking change for many, for example Intertia users

mabdullahsari:
No, this will break my existing Inertia app.

PR #44265, opened 2022-09-23
[9.x] Allow factories to recycle models with `for` method

jessarcher:
This PR addresses https://github.com/laravel/framework/pull/44107#issuecomment-1249158498 where using the Factory `for` method instead of specifying a `'model_id' => Model::factory()` definition does not use recycled models. I *think* this makes sense, although typically you'd call `for` from your test where you could just pass the model directly, so I think it's more of an edge case.

I considered adding support for the `has` method as well, however, I don't think that one makes as much sense. If the recycled model _belongs to_ another model, then I don't think it's parent relationship should be re-assigned, especially if it's part of a _has many_ relationship where more than one model might be needed. I could make it attach the model in a _many-to-many_ relationship, but it would probably need to be conditional on whether one or many models were requested. In any case, I can't think of a scenario where this would be a papercut so I'm not inclined to look into it further right now.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/BelongsToRelationship.php
+++ b/src/Illuminate/Database/Eloquent/Factories/BelongsToRelationship.php
@@ -69,7 +69,9 @@ class BelongsToRelationship
     {
         return function () use ($key) {
             if (! $this->resolved) {
-                $instance = $this->factory instanceof Factory ? $this->factory->create() : $this->factory;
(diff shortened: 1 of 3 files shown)


jessarcher:
I wasn't too keen on making the `recycle` property public as it doesn't match the other properties. But I thought it might also be overkill to create a public method to access it.

PR #44266, opened 2022-09-23
[9.x] Improve generator command output

michaeldyrynda:
This PR augments the output of Laravel's generator command in order to include the generated file path in the output, consistent with the `make:migration` command.

Particularly useful [for users new](https://twitter.com/cjav_dev/status/1570855550640066561) to the framework who may not know where the generated files end up by convention.

I haven't modified the output of the `make:migration` command, however, as this has an explicit option to show the full path else the abbreviated migration filename. The added complication with this command being that it is possible to pass `--realpath`, such that the path is outside of the conventional location. Happy to amend this PR to include, but felt like changing that would be a BC break at this point.

CleanShot 2022-09-23 at 12 11 54

### DIFF
--- a/src/Illuminate/Console/GeneratorCommand.php
+++ b/src/Illuminate/Console/GeneratorCommand.php
@@ -182,7 +182,7 @@ abstract class GeneratorCommand extends Command
             }
         }
 
-        $this->components->info($info.' created successfully.');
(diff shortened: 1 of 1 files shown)


driesvints (approved):
I agree, this lgtm 👍

nunomaduro (approved):

PR #44275, opened 2022-09-23
[9.x] Adds `retryQuietly` method to the HTTP Client

joedixon:
### TL;DR

This PR proposes a new method for the HTTP Client - `retryQuietly`. 

It's simply a wrapper around the existing `retry` method, setting `retryThrow` to false.

_____

### Background

Recently when working on a project, I needed to add the `retry` method to an existing implementation of the HTTP Client which looked a bit like this.

```php
$response = Http::get('https://example.com');

if ($response->failed()) {
    return;
}
```

Adding the retry method was as simple as you would expect:

```php
$response = Http::retry(3, 100)
    ->get('https://example.com');

if ($response->failed()) {
    return;
}
```
However, this came with the unexpected side effect of throwing an exception when encountering a server or client error.

When using the HTTP client without `retry`, the default is not to throw, but when using `retry` the default is to throw.

I was able to get around this with the below, but that didn't feel very Laravel.

```php
$response = Http::retry(3, 100, null, false) // :-(
    ->get('https://example.com');

if ($response->failed()) {
    return;
}
```

Using `retryQuietly`, the same can be achieved like so:

```php
$response = Http::retryQuietly(3, 100)
    ->get('https://example.com');

if ($response->failed()) {
    return;
}
```

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -542,6 +542,19 @@ class PendingRequest
     }
 
     /**
+     * Retry the request the specified number of times without throwing an exception on server or client errors.
+     *
+     * @param  int  $times
+     * @param  int  $sleepMilliseconds
+     * @param  callable|null  $when
+     * @return $this
+     */
+    public function retryQuietly(int $times, int $sleepMilliseconds = 0, ?callable $when = null)
+    {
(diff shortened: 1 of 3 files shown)


taylorotwell:
CleanShot 2022-09-23 at 10 02 07@2x

PR #44277, opened 2022-09-23
Check and get nested relations for Model

sirmathays:
This addition to Model class allows us to check and access nested relations with two new methods.

See following example:

```php
$u = User::first();

if ($u->nestedRelationLoaded('posts.comments')) { 
    return $u->getNestedRelation('posts.comments'); // Illuminate\Database\Eloquent\Collection<App\Models\Comment>
}
```

This pull request also introduces a change to JsonResources as well. When before, `whenLoaded` method didn't support nested relations, with this change it would.

Although this approach now supports doing API resources not in the most conventional way, this comes in handy especially in when creating index resources where you might want to keep the amount of items per row minimal.

```php
class UserResource extends JsonResource
{
    {
        return [
            // ...
            'comments' => new CommentResourceCollection($this->whenLoaded('posts.comments')),
            // ...
        ];
    }
}
```

vs

```php
class UserResource extends JsonResource
{
    {
        return [
            // ...
            'posts' => new PostResourceCollection($this->whenLoaded('posts')),
            // ...
        ];
    }
}

class PostResource extends JsonResource
{
    {
        return [
            // ...
            'comments' => new CommentResourceCollection($this->whenLoaded('comments')),
            // ...
        ];
    }
}
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
@@ -792,6 +792,37 @@ trait HasRelationships
     }
 
     /**
+     * Get a specified nested relationship.
+     *
+     * @param  string  $relation
+     * @return mixed
+     */
+    public function getNestedRelation($relation)
+    {
+        // If the specified relation is not nested, let's just fetch the
+        // relation.
+        if (! Str::contains($relation, '.')) {
+            return $this->getRelation($relation);
+        }
+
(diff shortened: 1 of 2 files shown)


taylorotwell:

sirmathays:
@taylorotwell Hey! I accidentally sent the pull request before completing my full explanation for this enhancement :)

This addition would be especially helpful in JsonResources, described in the examples above. And because this has changes to the Model and JsonResource classes themselves, it wouldn't work as a package.

PR #44279, opened 2022-09-23
[9.x] Allow each model to have different parent in Model Factories

bastien-phi:
Sometimes we need to create models having unicity constraint on parent id.

When using `for()`, all addresses belongs to the same user.

Now, we need to explicit the usage of foreign key

```php
Address::factory(3)->state(['user_id' => User::factory->registered()])->create();
```

This PR adds the ability te declare that each address will have its own user without messing with foreign keys : 
```php
Address::factory(3)->eachFor(User::factory()->registered())->create();
```


taylorotwell:

Issue #44285, opened 2022-09-23
Unable to run aggregate queries (withSum, etc.) for pivot table columns on self-referencing many-to-many relations


kayrunm:
- Laravel Version: 9.31.0 (from what I can see, also on 8.x)
- PHP Version: 8.1.10 (also on 7.4.x)
- Database Driver & Version: MySQL 8.0.30, SQLite 3.39.3

### Description:

An SQL error occurs when trying to add an aggregate (e.g. `withSum`) for a pivot column on a many-to-many relation that has both sides of the relationship from the same table.

For example, in an application which contains a `transactions` table, I may be able to have positive (+) and negative (-) transactions and I am able to allocate positive transactions to a negative transaction. Take the following list of transactions

| ID | Description |   Amount |
|----|-------------|---------:|
| 1  | Invoice 1   |  -500.00 |
| 2  | Invoice 2   | -1000.00 |
| 3  | Invoice 3   |  -200.00 |
| 4  | Payment 1   |  1700.00 |

In this example, I want to allocate:
* 500 from Payment 1 to Invoice 1
* 1000 from Payment 1 to Invoice 2
* 200 from Payment 1 to Invoice 3

I can achieve this through a `transaction_transaction` pivot table with an `amount` column signifying how much was allocated from one transaction to another. 

### Steps To Reproduce:

In my use case, I want to be able to allocate different amounts of Payment 1 across the invoices. I would end up with an intermediate table like this:

| allocated_from | allocated_to |  amount |
|----------------|--------------|--------:|
| 4              | 1            |  500.00 |
| 4              | 2            | 1000.00 |
| 4              | 3            |  200.00 |

And two relationships on my `Transaction` model like so:
```php
    public function allocatedTo(): BelongsToMany
    {
        return $this->belongsToMany(
            Transaction::class,
            'transaction_transaction',
            'allocated_from_id',
            'allocated_to_id',
        )->withPivot('amount');
    }

    public function allocatedFrom(): BelongsToMany
    {
        return $this->belongsToMany(
            Transaction::class,
            'transaction_transaction',
            'allocated_to_id',
            'allocated_from_id',
        )->withPivot('amount');
    }
```

When running the following:
```php
$result = Transaction::query()
    ->withSum('allocatedTo as total_allocated', 'transaction_transaction.amount')
    ->first();
```

I get the following error in SQLite (I get a similar one in MySQL as well)
```
[...]

PR #44286, opened 2022-09-23
[9.x] Fix issue with aggregates (withSum, etc.) for pivot columns on self-referencing many-to-many relations

kayrunm:
An SQL error currently occurs when trying to run any `withX` aggregate functions on a column from a pivot table for a many-to-many relation that references the same table. The error that occurs is similar to the below:

```
Illuminate\Database\QueryException : SQLSTATE[HY000]: General error: 1 no such column: laravel_reserved_0.transaction_transaction.amount (SQL: select "transactions".*, (select sum("laravel_reserved_0"."transaction_transaction"."amount") from "transactions" as "laravel_reserved_0" inner join "transaction_transaction" on "laravel_reserved_0"."id" = "transaction_transaction"."allocated_to_id" where "transactions"."id" = "transaction_transaction"."allocated_from_id") as "total_allocated" from "transactions" limit 1)
```

The issue at hand is from [lines 625–627 in QueriesRelationships](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php#L625):

```php
$hashedColumn = $this->getQuery()->from === $relation->getQuery()->getQuery()->from
     ? "{$relation->getRelationCountHash(false)}.$column"
     : $column;
```

The condition in this ternary checks whether the current query references the same table as the relation's query, and if so, it prepends a "relationCountHash" which in the above error is `"laravel_reserved_0"`. In a many-to-many relationship which references the same table, for example a "friends" relationship which links from a `users` table to the `users` table, this would return true and it would make referencing columns from a pivot table impossible in this situation.

My proposed fix for this is to move the above logic into a new method which checks for a period (`.`) in the `$column`, to see if the referenced column is on a different table to the original query. If this is the case, it just returns the `$column`, without the reserved table alias.

This shouldn't, to my knowledge, introduced any breaking changes with  current versions as any columns passed in with a period are automatically escaped into `"table_name"."column"` syntax, rather than treated as a single column (e.g.  `"table_name.column"`.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -678,6 +678,13 @@ trait QueriesRelationships
         return $this;
     }
 
+    /**
+     * Get the relation hashed column name for the given column and relation.
+     *
+     * @param  string  $column
+     * @param  \Illuminate\Database\Eloquent\Relations\Relationship  $relation
+     * @return string
+     */
     protected function getRelationHashedColumn($column, $relation)
     {
         if (str_contains($column, '.')) {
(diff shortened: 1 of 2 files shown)


driesvints:
Please add a thorough explanation to your PR and don't link to an issue (see the pr description).

kayrunm:
Apologies @driesvints, sorted now. 

For more information on the issue, you can view the issue at #44285.

PR #44289, opened 2022-09-23
Fixes uploading a multi-files issue

mahmoudmohamedramadan:
When we validate uploading **ONLY** one file, will be no validation error, like so

```HTML
<input type="file" name="image" class="form-control">
```

```PHP
$request->validate([
    'image' => ['mimes:png'],
]);
```

## Before this PR gets merged

When we try to upload multi-files that fit validation, the validation will fail, for instance

```HTML
<input type="file" name="images[]" class="form-control" multiple>
```

```PHP
$request->validate([
    'images' => ['mimes:png'],
]);
```

uplaoding-mutli-file-issue

## After this PR gets merged

The issue has been solved 🎉

fixes-uplaoding-mutli-file-issue

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1308,6 +1308,10 @@ trait ValidatesAttributes
      */
     public function validateMax($attribute, $value, $parameters)
     {
+        if(is_array($value)) {
+            return $this->checkMultiValues(__FUNCTION__, $attribute, $value, $parameters);
+        }
+        
         $this->requireParameterCount(1, $parameters, 'max');
 
         if ($value instanceof UploadedFile && ! $value->isValid()) {
@@ -1376,7 +1380,7 @@ trait ValidatesAttributes
         if(is_array($value)) {
             return $this->checkMultiValues(__FUNCTION__, $attribute, $value, $parameters);
         }
-        
+
         if (! $this->isValidFileInstance($value)) {
             return false;
         }
@@ -1422,6 +1426,10 @@ trait ValidatesAttributes
      */
     public function validateMin($attribute, $value, $parameters)
     {
+        if(is_array($value)) {
+            return $this->checkMultiValues(__FUNCTION__, $attribute, $value, $parameters);
+        }
+
         $this->requireParameterCount(1, $parameters, 'min');
 
         return $this->getSize($attribute, $value) >= $parameters[0];
@@ -2267,12 +2275,12 @@ trait ValidatesAttributes
      */
     protected function checkMultiValues($functionName, $attribute, $value, $parameters)
     {
-        $isOk = true;
-
         foreach ($value as $file) {
-            $isOk = $isOk && $this->$functionName($attribute, $file, $parameters);
+            if (!$this->{$functionName}($attribute, $file, $parameters)) {
+                return false;
+            }
         }
 
-        return $isOk;
+        return true;
     }
 }
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1344,6 +1344,10 @@ trait ValidatesAttributes
      */
     public function validateMimes($attribute, $value, $parameters)
     {
+        if(is_array($value)) {
+            return $this->checkMultiValues(__FUNCTION__, $attribute, $value, $parameters);
+        }
+
         if (! $this->isValidFileInstance($value)) {
             return false;
         }
@@ -1369,6 +1373,10 @@ trait ValidatesAttributes
      */
     public function validateMimetypes($attribute, $value, $parameters)
     {
+        if(is_array($value)) {
+            return $this->checkMultiValues(__FUNCTION__, $attribute, $value, $parameters);
+        }
+        
         if (! $this->isValidFileInstance($value)) {
             return false;
         }
@@ -2246,5 +2254,25 @@ trait ValidatesAttributes
         if (is_numeric($this->getValue($attribute))) {
             $this->numericRules[] = $rule;
         }
+    }
+
+    /**
+     * Check the given value and loop over each file then return the result.
+     *
+     * @param  string  $functionName
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @param  array<int, int|string>  $parameters
+     * @return bool
+     */
+    protected function checkMultiValues($functionName, $attribute, $value, $parameters)
+    {
+        $isOk = true;
+
+        foreach ($value as $file) {
+            $isOk = $isOk && $this->$functionName($attribute, $file, $parameters);
+        }
+
+        return $isOk;
     }
 }


morloderex:
@mahmoudmohamedramadan please take a look at.

https://laravel.com/docs/9.x/validation#validating-arrays

I believe this should be closed.

rodrigopedra:
Adding up to @morloderex comment:

@mahmoudmohamedramadan multi-file validation is already supported, try this:

~~~php
$request->validate([
    'images.*' => ['mimes:png'],
]);
~~~

And read the docs linked by @morloderex

mahmoudmohamedramadan:
@morloderex This differs from what I talking about. The link that you provide is talking about **Validating Arrays**. As docs say **the array rule accepts a list of allowed array keys. If any additional keys are present within the array, validation will fail**, which means, if the given array doesn't have these values then trigger a validation error, and I don't mean that.

mahmoudmohamedramadan:
> Adding up to @morloderex comment:
> 
> @mahmoudmohamedramadan multi-file validation is already supported, try this:
> 
> ```
> $request->validate([
>     'images.*' => ['mimes:png'],
> ]);
> ```
> 
> And read the docs linked by @morloderex

This does not work perfectly when trying to upload `.png` with `.txt` whereas works perfectly with my PR.

rodrigopedra:
Well it won't work as it is validating only for PNG. If in your it is also accepting TXT files, then there is an implementation problem.

Look, see the code regarding validating arrays, it basically does what yours does, iterate over the input array and apply the same rules individually.

But it also takes care of other stuff: 

- allows to specify the item index in the validation message
- allow validating minimum and maximum array items (see example below)
- will keep the same index order when sending validation error messages, so a developer can easily iterate over individual error messages
- allows nested array items
- and other details that have been added since Laravel 5.3, when input array validation was first introduced.

Are you planning on handling those too?

It seems you faced this problem, wasn't aware Laravel already had this feature, and tried to solve your particular problem with an implementation that in the end is less capable compared to what we currently have.

Don´t take me wrong, I see you've been contributing a lot lately with some degree of success. And that is great, Laravel is a framework built from contributions like yours.

But unless you detail what new problem this PR solves that is not already handled by the framework, or what is the new issue the current feature set does not cover, I guess this is very likely not to be merged.

Also, if someone points out there is a similar feature already in place, and you believe it is still not covering the issue you want to handle, why not trying to fix the current feature to solve your issue instead of insisting on introducing new code to new places based "my solution works, current code does not"?

Imagine if someone is not aware of queue batches and then they introduced a half-baked solution to an particular queue driver that only encompasses their single use-case, and when told to use the built-in feature they insist their approach is better? If the framework accepts all new submitted code for each individual use case it would become a nightmare to maintain it.

I am not trying to let you down, on the contrary, I am suggesting you to be more open-minded and see responses like the ones you got here as an incentive and not just critical.

This is the snippet with array size validation, I mentioned above.

~~~php
$request->validate([
    'images' => ['array', 'min:1', 'max:3'],
    'images.*' => ['mimes:png'],
]);
~~~

src/Illuminate/Validation/Concerns/ValidatesAttributes.php
```diff
@@ -2247,4 +2255,24 @@ protected function shouldBeNumeric($attribute, $rule)
...
+    protected function checkMultiValues($functionName, $attribute, $value, $parameters)
+    {
+        $isOk = true;
+
+        foreach ($value as $file) {
+            $isOk = $isOk && $this->$functionName($attribute, $file, $parameters);
+        }
+
+        return $isOk;
+    }
```
rodrigopedra, line 2277: You can skip the temporary variable:

~~~php
protected function checkMultiValues($functionName, $attribute, $value, $parameters)
{
    foreach ($value as $file) {
        if (! $this->{$functionName}($attribute, $file, $parameters)) {
            return false;
        }
    }

    return true;
}
~~~

src/Illuminate/Validation/Concerns/ValidatesAttributes.php
```diff
@@ -1344,6 +1344,10 @@ public function validateMaxDigits($attribute, $value, $parameters)
      */
     public function validateMimes($attribute, $value, $parameters)
     {
+        if(is_array($value)) {
+            return $this->checkMultiValues(__FUNCTION__, $attribute, $value, $parameters);
+        }
+
```
rodrigopedra, line 1350: Why were the array checks only added to the mime rules? 

What if someone wants to validate max file size, required, minimum file size?

mahmoudmohamedramadan:
**"If in your solution it is also accepting TXT files, then there is an implementation problem"**

nope, my PR solves that issue with a correct validation message error as exists in the gif in addition, this PR can support all features that you mentioned easily by including the next check in `validateMin` and `validateMax` as you have just said.

```PHP
if(is_array($value)) {
    return $this->checkMultiValues(__FUNCTION__, $attribute, $value, $parameters);
}
```

**"Don´t take me wrong, I see you've been contributing a lot lately with some degree of success. And that is great, Laravel is a framework built from contributions like yours."**

Thanks ❤

src/Illuminate/Validation/Concerns/ValidatesAttributes.php
mahmoudmohamedramadan: Nope, because when `$this->{$functionName}($attribute, $file, $parameters)` returns **false** this will exit from the loop then will return **true** and this is invalid approach so, I created a temporary variable to store all check value then return the final result.

rodrigopedra:
@mahmoudmohamedramadan I guess you didn't get what I meant by individual validation messages.

Take this HTML:

~~~blade
<input type="file" name="images[]" class="form-control">
{{-- error message for first input only --}}

<input type="file" name="images[]" class="form-control">
{{-- error message for second input only --}}

<input type="file" name="images[]" class="form-control">
{{-- error message for third input only --}}
~~~

Instead of using the `multiple` attribute, a developer lists individual inputs. This is very common when a developer wants to outline the allowed maximum input.

How, in your solution, would each individual error message be displayed? 

For example, if only the second input exceeds the maximum file sizes, how, in your solution, a developer would display that the user only needs to fix the second file?

rodrigopedra:
> his PR can support all features that you mentioned easily by including the next check in `validateMin` and `validateMax` as you have just said.

That is what I meant when I said you were only targeting your use case. A PR should consider the global impacts to the framework and not only look at an individual issue.

Imagine one issue being opened every time an user realizes their preferred validation rule is not yet array-aware? It would surely be a nice way to increase merged PR counts.

src/Illuminate/Validation/Concerns/ValidatesAttributes.php
rodrigopedra: No it won't. A `return` will always exit the function it belongs to. Maybe you are confusing with a `break` statement.

This approach is used in many places on the framework:

https://github.com/laravel/framework/blob/278c6d7fd04be005c1175a7eaea868976177600c/src/Illuminate/Auth/SessionGuard.php#L457-L466

https://github.com/laravel/framework/blob/278c6d7fd04be005c1175a7eaea868976177600c/src/Illuminate/Auth/Access/Gate.php#L109-L120

https://github.com/laravel/framework/blob/278c6d7fd04be005c1175a7eaea868976177600c/src/Illuminate/Collections/Collection.php#L541-L552

https://github.com/laravel/framework/blob/278c6d7fd04be005c1175a7eaea868976177600c/src/Illuminate/Collections/Collection.php#L560-L575

these were the first results from searching Laravel code base for `return false`

rodrigopedra:
> this PR can support **all features that you mentioned** easily by including the next check in validateMin and validateMax as you have just said.

If you add the same code to `validateMin` and `validateMax` how you'd allow a developer to developer max/min file sizes and max/min items in the array (for example a post can have at most 5 images). 

For a single file upload max/min are usually coerced to its file size. How to allow restricting file count?

src/Illuminate/Validation/Concerns/ValidatesAttributes.php
rodrigopedra: Returning in first failure will also skip checking the remaining items of the array. So if the first check fails and you have 30 items in your array, you won't need to check all of them.

mahmoudmohamedramadan: You are right maybe I am confused BETWEEN `return` and `break`.

mahmoudmohamedramadan:
> @mahmoudmohamedramadan I guess you didn't get what I meant by individual validation messages.
> 
> Take this HTML:
> 
> ```
> <input type="file" name="images[]" class="form-control">
> {{-- error message for first input only --}}
> 
> <input type="file" name="images[]" class="form-control">
> {{-- error message for second input only --}}
> 
> <input type="file" name="images[]" class="form-control">
> {{-- error message for third input only --}}
> ```
> 
> Instead of using the `multiple` attribute, a developer lists individual inputs. This is very common when a developer wants to outline the allowed maximum input.
> 
> How, in your solution, would each individual error message be displayed?
> 
> For example, if only the second input exceeds the maximum file sizes, how, in your solution, a developer would display that the user only needs to fix the second file?

I've tested this and it has worked successfully.

## Show validation message error for the first input with a correct index

first-not-valid


validation-first-not-valid

## Show validation message error for the second input with a correct index

second-not-valid

validation-second-not-valid

rodrigopedra:
@mahmoudmohamedramadan and when the user uploads no files, and both are required?

Do both messages are shown?

- "The avatar.0 must be a file of type jpeg"
- "The avatar.1 must be a file of type jpeg"

This is currently possible with existing array validation feature.

mahmoudmohamedramadan:
yup, still working perfectly 😃

both-input-validation

taylorotwell:

PR #44291, opened 2022-09-24
[9.x] Add `even` and `odd` method to collection

Nothehi:
Receiving values that are in the even or odd position is one of those tasks that are very widely used, and I think it is better to create semantic functions for them.

example:
```php
collect(['a', 'b', 'c', 'd', 'e', 'f', 'g'])->even()
// output: b, d, f

collect(['a', 'b', 'c', 'd', 'e', 'f', 'g'])->odd()
//output: a, c, e, g
```


JayBizzle:
Naming might be confusing.

```php
collect([1, 3, 5, 7, 9])->even();

// Should it return empty, or the even keys?
```

Nothehi:
It actually returns items that are in even or odd position, and it doesn't matter what key or value the items have, the only thing that matters is the item's position in the collection.

It will return:
```
// 3, 7
```

But I agree with you might be confusing, Have you any suggestion?

mahmoudmohamedramadan (commented):
```PHP
/**
 * Get the values that are in even positions.
 *
 * @return static
 */
public function even()
{
   return $this->nth(2, 1);
}
```

taylorotwell:

PR #44295, opened 2022-09-25
[9.x] Improve type definitions of the enum() method

supun-io:
This pull request improves the types of the `$request->enum()` method. 

Currently, in the following code, `$type` is `mixed`.

```php
$type = $request->enum('key', MyEnum::class);
// $type is mixed
```

With this pull request, `$type` will have `MyEnum|null` type.

```php
$type = $request->enum('key', MyEnum::class);
// $type is now MyEnum|null
```

[Here's](https://phpstan.org/r/86115cd3-052d-4a50-b0fe-71fa8799d6b1) a PHPStan playground to play with the generics of this. 

I am not sure if the place where I added type-level tests is the best place (`/types/Http/Request.php`)

This pull request does not change any actual functionalities - just the PHPDoc types.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -372,9 +372,10 @@ trait InteractsWithInput
     /**
      * Retrieve input from the request as an enum.
      *
+     * @template T
      * @param  string  $key
-     * @param  string  $enumClass
-     * @return mixed|null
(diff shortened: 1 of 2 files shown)


supun-io:
Ah, static analyzing for types fails because it runs on PHP 8.0. Should it be changed to 8.1 in the workflow?

taylorotwell:
@nunomaduro

PR #44298, opened 2022-09-26
[9.x] Add `before` option to make:migration command

Nothehi:
There are times when we need to create a migration before another migration, so what we're doing at this point is just changing the timestamps of the file, but It is very confusing looking at these timestamps and changing them to move.

I suggest having commands like the following to make this simple:
```php
php artisan make:migration  create_roles_table --before=users
```


taylorotwell:

PR #44303, opened 2022-09-26
[9.x] Introduce Arr::firstNotEmpty helper

linaspasv:
This PR introduces a new ``Arr`` helper function which allows you to get the first non-empty item from an array using a list of keys.

Example usage:

```php
$array = [
    'name' => null, 
    'owner' => [
        'name' => 'John Smith'
    ]
];

$value = Arr::firstNotEmpty($array, ['name', 'owner.name']);

// 'John Smith'

```

This is particularly helpful when you want to quickly extract non-empty value from a nested array structure based on a priority defined through keys.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -205,7 +205,7 @@ class Arr
 
         return value($default);
     }
-    
+
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #44309, opened 2022-09-26
[9.x] Add `soleValue` method to Collections

ash-jc-allen:
Hey! This PR proposes a small `soleValue` method that can be added to `Collection`s and `LazyCollection`s.

The new method tries to mimic the `soleValue` that's available in the query builder (added in this PR: https://github.com/laravel/framework/pull/41368).

The idea behind using this is that it gives you more confidence that:
1. Exactly one items in the collection
2. The key definitely exists in the item that we are accessing

As a basic example, before this method, I might have written something similar to this:

```php
$item = $collection->where('name', 'foo')->sole();

if (!array_key_exists('bar', $item)) { 
    throw new Exception('The "bar" field does not exist in the item');
}

return $item->bar;
```

But with the new method, I could write it like this:

```php
return $collection->where('name', 'foo')->soleValue('bar');
```

If the item doesn't exist, an `Illuminate\Support\InvalidKeyException` exception will be thrown.

If you think this might be something worth merging and you need any changes making to it, please let me know and I'll sort it ASAP 😄


driesvints:
@ash-jc-allen can you rebase with 9.x? We fixed the failing build there.

ash-jc-allen:
@driesvints Sweet! I've done that now 🙂

taylorotwell:
Think I'm going to hold off on this one for now. You can macro it in.

ash-jc-allen:
That's no problem at all, I'll use a macro 🙂

PR #44311, opened 2022-09-26
[10.x] Stick with ordered uuid

tillkruss:
This is a follow-up to #44210.

- Stick with `orderedUuid()` in `HasUuids::newUniqueId()`
- Revert changed from v4 to v7 in `Str::uuid()`

IMO it makes no sense to change the default `Str::uuid()` to v7, which will always reveal the timestamp (minor data leak), especially since the `Str::orderedUuid()` still exists.

I'd expect `Str::uuid()` to be fully random and not an alias of `Str::orderedUuid()`.


taylorotwell:
I basically agree with this.

driesvints:
Thanks @tillkruss

PR #44312, opened 2022-09-26
[9.x] Get the validated data Except keys from the request

BNhashem16:
Sometimes we want to validate the inputs and make sure that are correct, but when retrieve data in the controller `$request->validated() ` all keys will return! 
so I added function will filter keys with values and with be applicable with one key or array of keys `$request->validatedExcept(['keys'])` 
-> now, the validated keys will return except keys I passed to the function

### DIFF
--- a/src/Illuminate/Foundation/Http/FormRequest.php
+++ b/src/Illuminate/Foundation/Http/FormRequest.php
@@ -10,6 +10,7 @@ use Illuminate\Contracts\Validation\ValidatesWhenResolved;
 use Illuminate\Contracts\Validation\Validator;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Redirector;
+use Illuminate\Support\Arr;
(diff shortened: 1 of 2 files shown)


habibalkhabbaz:
Hello @BNhashem16 

I think you can use `exclude` or `exclude_if` instead.

https://laravel.com/docs/9.x/validation#rule-exclude

BNhashem16:
Hello @habibalkhabbaz 

can you read my description again, please?

this functions you mentioned is not related to my function!

please check the files I changed, and you will understand my idea :)

[](https://github.com/laravel/framework/pull/44312/files)

habibalkhabbaz:
It is exactly what you need.

> `exclude`

The field under validation will be excluded from the request data returned by the **validate** and **validated** methods.

alirezasalehizadeh:
```
// Retrieve a portion of the validated input data...
$request->safe()->except();
```
https://laravel.com/docs/9.x/validation#creating-form-requests

PR #44313, opened 2022-09-27
[9.x] Update console stub to use constant over magic number

jarrodtomholt:
Small pain point for our developers (and new candidates) remembering what 0,1,2 mean when returning Command status'.
This PR updates the console command stub to return the constant configured in the parent `Symfony\Component\Console\Command` over magic numbers.


antonkomarev:
There were several attempts to start using constant. No luck.

jarrodtomholt:
> 

thanks heaps for the feedback @antonkomarev, can for sure appreciate that. it shouldn’t stop us progressing and no harm in adding the PR if others find the same benefit our developers and candidates have. if not, then no harm in trying

taylorotwell:
I guess I'll just give in and merge this. Tired of PRs about it.

PR #44314, opened 2022-09-27
[9.x] Faking specific methods in the `Http` client

ash-jc-allen:
Hey! This PR proposes a few new changes for using in the `Http` client when testing.

It allows you to define the specific HTTP methods that should be accepted when sending requests to faked URLs.

For some context, I've been working on a project recently where I give users the option to define webhooks that get sent out from my application. They can also choose whether to send their request as a `GET` or `POST` request. So, in my tests, I've been asserting that the webhooks are being sent correctly.

As a basic example, if I wanted to test that my code is sending a `POST` request to `https://example-api.com`, my test and app code (mixed together for the sake of the example) looks something like this:

```php
// In the test...
Http::fake([
    'https://example-api.com' => Http::response('ok'),
]);

Http::preventStrayRequests();

// App code
Http::post('https://example-api.com');

// Back in the test...
Http::assertSent(function (Request $request): bool {
    return $request->method() === 'POST';
});
```

But with my proposed change, I could use the `Http::fakePost()` method to specify that the `https://example-api.com` route can only accept a `POST` request. If you send anything else, the `Http::preventStrayRequests()` method will throw an error and fail the test. So, for instance, sending a `GET` request would fail.

For example, using the new method, we could rewrite it like so:

```php
// In the test...
Http::fake([
    Http::fakePost('https://example-api.com') => Http::response('ok'),
]);

Http::preventStrayRequests();

// App code
Http::post('https://example-api.com');
```

The biggest benefit that I see to this is that we don't need to explicitly add a new assertion to check the request method is what was intended, because the request will fail (for lack of better term) if we use any other method.

An added bonus is that we can use this approach to define different responses for matching routes depending on the HTTP method being used. For example:

```php
Http::fake([
    Http::fakePost('https://example-api.com') => Http::response('post'),
    Http::fakeGet('https://example-api.com') => Http::response('get'),
    Http::fakePatch('https://example-api.com') => Http::response('patch'),
    Http::fakePut('https://example-api.com') => Http::response('put'),
    Http::fakeDelete('https://example-api.com') => Http::response('delete'),
]);
```

Under-the-hood, to achieve this, the `fakePost` etc methods are just prepending a prefix to the URLs so that we can determine whether we should only accept a specific method when running the callbacks during testing. For example, if we used `Http::fakePost('https://example-api.com')`, we'd just be adding a new faked URL called `LARAVEL_FAKE_REQUEST:POST:https://example-api.com`.

I don't know if this approach is suitable, but it seems to do the job without having to make a huge amount of changes to the code or without breaking the existing implementation. I also chose a pretty verbose prefix to add to the beginning of the URLs to avoid any clashes with URLs that devs might have set. But maybe it's a bit overkill?

Apologies if something like this is already possible and I didn't already realise.

If it's something that you think might be helpful to other devs, please let me know if you need any changes making 😄


taylorotwell:
Since it's pretty easy to test your POST as in your first example, I think I'll just hold off adding more code to the framework for this for now.

PR #44328, opened 2022-09-27
[9.x] Allow factories to recycle multiple models of a given type

michal3377:
This PR is an enhancement to recently added feature, Eloquent Factory recycling (#44107) and adds support for recycling multiple models of the same type.
Model recycling is a great feature, making factories much cleaner and easier to write. While recycling a single model has its valid use cases, it doesn't really help if we want to create multiple models (`->count(10)`) with our factory - all of them will be related to that same model.

This PR allows developers to provide their own, already created pool of models (collections), so they could be reused in all nested relations. For each relation, factory class **picks a random model** from a provided collection of models, instead of depending on a single one.
Sample usage:

        $users = User::all();
        $companies = Company::all();

        Posts::factory()
            ->recycle($users)
            ->recycle($companies)
            ->count(10)
            ->create();

For example, we often see that the User model relates to everything - as an author of some content, comments, reactions, performed operations etc. In more complex applications, that nesting can go deep. In that scenario, much easier would be to seed a number of users once and then reuse them in all nested relations (so we don't create a user only to be the author of a single comment). It's then closer to a real-life case, where we could have a set number of users in the system and all the actions performed and content created is within that group.

This approach leads to cleaner factories, less database pollution and overall higher seeding quality - created models will be much more related to each other. 

New recycle method still allows for providing a single model so the PR is backward-compatible, without any breaking changes.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/BelongsToRelationship.php
+++ b/src/Illuminate/Database/Eloquent/Factories/BelongsToRelationship.php
@@ -70,7 +70,7 @@ class BelongsToRelationship
         return function () use ($key) {
             if (! $this->resolved) {
                 $instance = $this->factory instanceof Factory
-                    ? ($this->factory->recycle->get($this->factory->modelName()) ?? $this->factory->create())
+                    ? ($this->factory->getRandomRecycledModel($this->factory->modelName()) ?? $this->factory->create())
                     : $this->factory;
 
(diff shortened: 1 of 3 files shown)


taylorotwell:
Revert all spacing changes so I can read the diff please.

taylorotwell:
Please mark as ready for review when that is done.

michal3377:
Sure! Sorry about that

PR #44329, opened 2022-09-27
[10.x] Prevent silently discarding model attributes by default

tillkruss:
It makes sense to be verbose as the default, especially for beginners.

Replaces https://github.com/laravel/laravel/pull/5993.
Follow-up to #43893.


taylorotwell:
But this shouldn't be enabled in production IMO?

tillkruss:
Personally, I'd want Laravel to be explicit and never silently discard rows I expected to be inserted. But maybe that's me 🤷‍♂️

dennisprudlo:
Wouldn't insertions/updates with `$request->all()` always throw when some user tries to sneak in a field?

PR #44333, opened 2022-09-27
[9.x] Adds `@var` statement

tenmajkl:
# `@var` statement

This pull request adds `@var` into blade. Sometimes you want to create variables and using `@php` might not be the best option. It contains simple syntax checker which lets you write only variable assignment inside.

```blade
@var($foo = 0)
@var($foo++)
```
etc.

It lives in its own Concern as it doesn't fit the others.


taylorotwell:

dennisprudlo:
@TENMAJKL Isn't this already possible with `@php($var = 5)` or does the wrapping in parenthesis have any side effects?

TENMAJKL:
@dennisprudlo it is, this is just syntax sugar that allows only variable assignment.

PR #44334, opened 2022-09-27
Fixed #44332: dd shows compiled view not the original view file

### LINKED ISSUE #44332: `dd` source file shows compiled view file instead of original one.
- Laravel Version: 9.32.0
- PHP Version: 8.1.8

### Description:

When using `dd` in a view file. the source file line of it is displayed of a compiled view file instead of the original one:

image
 
### Steps To Reproduce:

1. Make sure you are on Laravel 9.32.0 version.
2. Add dd in a view file like: 
`@php 
dd('hello from view'); 
@endphp`
3. See the output

baselrabia:
Allow dd to print out the correct path for the view file that being dump not the compiled one, This fixes https://github.com/laravel/framework/issues/44332.
Screenshot from 2022-09-28 00-55-59

### DIFF
--- a/src/Illuminate/Foundation/Concerns/ResolvesDumpSource.php
+++ b/src/Illuminate/Foundation/Concerns/ResolvesDumpSource.php
@@ -37,12 +37,12 @@ trait ResolvesDumpSource
             $relativeFile = substr($file, strlen($this->basePath) + 1);
         }
 
-        if(str_starts_with($relativeFile,"storage/framework/views/")){
+        if (str_starts_with($relativeFile, 'storage/framework/views/')) {
             $fileArr = file($file);
             $lastLine = end($fileArr);
 
-            $result = str_replace("<"."?php /**PATH ", "", $lastLine);
-            $result = str_replace(" ENDPATH**/ ?>", "", $result);
+            $result = str_replace('<' . '?php /**PATH ', '', $lastLine);
+            $result = str_replace(' ENDPATH**/ ?>', '', $result);
             $relativeFile = substr($result, strlen($this->basePath) + 1);
         }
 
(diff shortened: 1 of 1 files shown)


src/Illuminate/Foundation/Concerns/ResolvesDumpSource.php
```diff
@@ -37,13 +37,22 @@ public function resolveDumpSource()
             $relativeFile = substr($file, strlen($this->basePath) + 1);
         }
 
+        if (str_starts_with($relativeFile, 'storage/framework/views/')) {
```
X-Coder264, line 40: @baselrabia This cannot be hardcoded like this. The view storage path should be fetched from the config.

https://github.com/laravel/laravel/blob/v9.3.8/config/view.php#L20-L34

src/Illuminate/Foundation/Concerns/ResolvesDumpSource.php
```diff
@@ -34,16 +34,38 @@ public function resolveDumpSource()
...
+
+        $viewPath = $this->getAbsolutePath(config('view.compiled'));
+
+        if (str_starts_with($relativeFile, $viewPath)) {
+            $fileArr = file($file);
+            $lastLine = end($fileArr);
+
+            $result = str_replace('<'.'?php /**PATH ', '', $lastLine);
+            $result = str_replace(' ENDPATH**/ ?>', '', $result);
+            $relativeFile = $this->getAbsolutePath($result);
```
xiCO2k, line 48: I would use a regex instead of `str_replace`.

Something like this will work `/\/\*\*PATH\s(.*)\sENDPATH/`.

nunomaduro:
Thank you. I've created this https://github.com/laravel/framework/pull/44347, that's inspired by this pull request, yet it contains some changes we've discussed internally. 

Also, I've added you as co-author on the pull request.

PR #44343, opened 2022-09-28
Add ```spread``` method to Collection

rick20:
This pull request adds ```spread()``` method to Collections.

### Why?
There are many cases where we need to produce more collection items from each item in the given collection.

For example, given we have the following notes:
```php
collect(['C', 'D', 'E', 'F', 'G', 'A', 'B']);
```

Now we want to get all possible notes including sharps.

### Solution

```php
collect(['C', 'D', 'E', 'F', 'G', 'A', 'B'])
    ->spread(fn ($note) => in_array($note, ['E', 'B']) ? $note : [$note, $note . '#']);

// output: ['C', 'C#', 'D', 'D#', 'E', 'F', 'F#', 'G', 'G#', 'A', 'A#', 'B']
```

Edit:
Alternatively, the method can also be named as ```expand()```

### DIFF
--- a/src/Illuminate/Collections/Enumerable.php
+++ b/src/Illuminate/Collections/Enumerable.php
@@ -915,6 +915,14 @@ interface Enumerable extends Arrayable, Countable, IteratorAggregate, Jsonable, 
     public function split($numberOfGroups);
 
     /**
+     * Spread a collection into more members using given callback
+     * 
+     * @param callable $callback
+     * @return static
+     */
+    public function spread(callable $callback);
+
+    /**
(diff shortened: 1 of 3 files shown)


taylorotwell:
```php
collect([1, 3, 4])->map(function ($x) {
  return $x == 1 ? [1, 2] : $x;
})->flatten();
```

rick20:
Yes, but I think it’s more straightforward and gives the collection more contextual power that easily pop in mind of what the collection is capable of, which is not only “shrink”, or “one on one mapping”, but also to “grow” its members.

But thanks anyway for looking it.

PR #44348, opened 2022-09-28
[9.x] Fixed #44344: Fix RedisStore value types validation inconsistencies

### LINKED ISSUE #44344: RedisStore serializer type inconsistencies check
- Laravel Version: >=6
- PHP Version: >=4.2

### Description:
The [validation for serializer ](https://github.com/laravel/framework/blob/5dd58d428c64811b70ecba57a75e77c136578410/src/Illuminate/Cache/RedisStore.php#L332) is type inconsistent.  The `is_numeric` function validates if a value is numeric without checking for a specific type, the accepted type is mixed. This means that a string representing a number will return true. However, the `is_nan` check is only accepting float. Currently, the serialization is passing and not throwing an error because strict types are not enforced.

### Steps To Reproduce:
Have a numeric value represented as a string, ex: `$value = "123";'

`is_numeric($value) && ! in_array($value, [INF, -INF]) && ! is_nan($value)` will return true, because it is not enforcing types, however this can break if `strict_types` are set or in possible future PHP versions. 

### Proposed solutions:
Convert the value to float if it is a numeric value. See https://github.com/laravel/framework/pull/44348.

danielpopdan:


taylorotwell:
PR doesn't describe what bug is being fixed in real applications.

PR #44352, opened 2022-09-28
[9.x] Fix support path

GhostWalkers:


taylorotwell:
To support legacy factories.

PR #44353, opened 2022-09-28
illuminate/console/parser.php from switch to match php8

ceyhun-celik:
Laravel framework require php 8.0.2 version so i want to help to adapt Laravel to the new version. This is my first pull request and I made "match"  expression rather than "switch" expression. If you not think this unnecessary, my second pr will contains all "switch" expression.


GrahamCampbell:
This change is not necessary and breaks the build.

ceyhun-celik:
Okay Graham thank you for your response

PR #44359, opened 2022-09-29
[9.x] Add Custom Path to Env Encryption Command and Add Test

afiqiqmal:
This PR proposes 1 new argument for commands; env:encrypt and env:decrypt which provide custom path of environment file. 

Why? because sometime developers located their .env for staging / production for CI on another folders instead of base project path.

We may utilise our own path by using the --path option.
```
php artisan env:encrypt --path=docker/dev/environment --key=abcdefghijklmnopabcdefghijklmnop
```

Screenshot 2022-09-29 at 11 50 34 AM

php artisan env:decrypt --path=docker/dev/environment

```
php artisan env:decrypt --path=docker/dev/environment --key=abcdefghijklmnopabcdefghijklmnop
```

Screenshot 2022-09-29 at 11 51 38 AM

I added tests for this PR

### DIFF
--- a/src/Illuminate/Foundation/Console/EnvironmentDecryptCommand.php
+++ b/src/Illuminate/Foundation/Console/EnvironmentDecryptCommand.php
@@ -22,6 +22,7 @@ class EnvironmentDecryptCommand extends Command
                     {--key= : The encryption key}
                     {--cipher= : The encryption cipher}
                     {--env= : The environment to be decrypted}
+                    {--path= : Where is the environment file path is located}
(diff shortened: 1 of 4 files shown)


driesvints:
> Why? because sometime developers located their .env for staging / production for CI on another folders instead of base project path.

That's not supported.

taylorotwell:
I don't think this approach is necessary - the command could be updated to call `$app->environmentPath()` to get the path instead of hard-coding `base_path` though. 

CleanShot 2022-09-29 at 08 47 35@2x

PR #44371, opened 2022-09-29
[9.x] Testing methods. Making error messages with json_encode more readable.

ar2r:
Added JSON_UNESCAPED_UNICODE in json_encode to make error messages more readable.

Before patching, error messages looks like this (look at Unicode \u0000):
```
Failed asserting that exception message 'Unable to find JSON fragment: 

[{"id":1}]

within

[[{"foo":"bar","id":10},{"foo":"\u041c\u0438\u0440","id":30},{"foo":"\u041f\u0440\u0438\u0432\u0435\u0442","id":20}]].
```

After patching, error messages become more readable:
```
Failed asserting that exception message 'Unable to find JSON fragment: 

[{"id":1}]

within

[[{"foo":"bar","id":10},{"foo":"Мир","id":30},{"foo":"Привет","id":20}]].
```

This feature allows reading Unicode text without converting it in external tools or sites.

### DIFF
--- a/src/Illuminate/Testing/AssertableJsonString.php
+++ b/src/Illuminate/Testing/AssertableJsonString.php
@@ -113,11 +113,12 @@ class AssertableJsonString implements ArrayAccess, Countable
      */
     public function assertSimilar(array $data)
     {
-        $actual = json_encode(Arr::sortRecursive(
-            (array) $this->decoded
-        ));
+        $actual = json_encode(
+            Arr::sortRecursive((array)$this->decoded),
+            JSON_UNESCAPED_UNICODE
+        );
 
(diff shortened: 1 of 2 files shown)


jessarcher:
@taylorotwell makes sense to me - those escape strings are pretty nasty to look at. I haven't tested this locally though.

ar2r:
> A lot of irrelevant spacing changes here. Please don't make those.

I removed redundant spaces. Do you have any other comments?

ar2r:
> @taylorotwell makes sense to me - those escape strings are pretty nasty to look at. I haven't tested this locally though.

As for me it makes me feed bad if i read in logs a lot of escaped strings :-(  What do I need to improve?

PR #44355, opened 2022-09-28
[9.x] Update SQL Server Offset

dunhamjared:
This PR helps reduce query select conflicts and removes unexpected data generated by laravel. 

Changes:

- Renamed temporary column `row_num` to `temp_row_num` to help reduce conflicts 
- Removes the `temp_row_num` column from the results if offset used 

Fixes https://github.com/laravel/framework/issues/44354

Issue #20814, opened 2017-08-29
'chunk loop' loop is adding a 'row_num' property


mojosef:
- Laravel Version: 5.4.32
- PHP Version: 7.1

### Description:
If you are on the last iteration of a 'chunk loop' loop, a 'row_num' property is added. If you then go to update and save that model, it fails as the query attempts to update a column 'row_num' on the model.

### Steps To Reproduce:
`\App\SomeModel::chunk(200, function($things){
    foreach ($things as $key => $thing) {
        $thing->foo = 'bar';
        $thing->save();
    }
});
Column not found: 1054 Unknown column 'row_num' in 'field list' (SQL: update some_model set foo = 'bar' updated_at = 2017-08-29 13:14:04, row_num = 201 where id = 201)`

mnabialek:
What engine do you use? Are you ussing `sqlsrv`?

mnabialek:
Yes, I think that's the reason, but I don't use that driver so cannot fix this easily.

xorock:
I have the same problem. Temporary solution is to add 'row_num' to list of hidden fields in Your model.

```php
protected $hidden = [
        'row_num',
    ];
```

Issue #7568, opened 2015-02-23
[Proposal] SqlServer2012 grammar file


crissi:
That implements limit and offset with sql server 2012's: **OFFSET** AND **FETCH** instead of **row_number() over**

GrahamCampbell:
Any thoughts?

twaileit:
Well, that "should" be pretty easy.

In SqlServerGrammar.php:
```php
    protected function compileOffset(Builder $query, $offset)
    {
        return '"OFFSET {$offset} ROWS";
    }

    protected function compileLimit(Builder $query, $limit)
    {
        return "FETCH {$limit} ROWS ONLY";
    }
```

GrahamCampbell:
We're open to PRs.

yansern:
I attempted to implement OFFSET...FETCH but it only take effect when `->orderBy` is used. Otherwise it will fallback to `ROW_NUMBER()`.

Gist: https://gist.github.com/yansern/4d7017351314eb37fa0df284ccd51c40

This is for the file: `src/Illuminate/Database/Query/Grammars/SqlServerGrammar.php`.

Perhaps this can be a starting point for others to follow.

PR #44377, opened 2022-09-29
[10.x] Possible fix for unexpected column in SQL Server offset

### LINKED ISSUE #44354: Offset for sqlsrv breaks queries with row_num column
- Laravel Version: 9.x
- PHP Version: 8.1
- Database Driver & Version: SQLServer

### Description

PR https://github.com/laravel/framework/pull/39863 introduces two issues.

If offset is used for SQL Server:

 1. You cannot use `row_num` as a column name or alias
 2. An unexpected column `row_num` is included in the results

#### Scenario 1

If you have a `row_num` column or alias, SQL Server will throw:

```
SQLSTATE[42000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]The column 'row_num' was specified multiple times for 'temp_table'
```

#### Scenario 2

If you try to insert the offset results into the same table by using `chunk`:

```php
db()->connection('db_1')->table('example')->chunk(500, function($chunk){
    db()->connection('db_2')->table('example')->insert($chunk);
});
```

`row_num` is added to the results and is an unexpected column, causing SQL server to throw:

```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'row_num' in 'field list'
```

dunhamjared:
Looking for feedback on this PR for resolving this issue:

Currently, if offset is used for SQL Server:

 1. You cannot use `row_num` as a column name or alias
 2. An unexpected column `row_num` is included in the results

This PR helps reduce query select conflicts and removes unexpected data generated by laravel. 

Changes:

- Renamed temporary column `row_num` to `temp_row_num` to help reduce conflicts 
- Removes the `temp_row_num` column from the results if offset used 

Fixes https://github.com/laravel/framework/issues/44354

#### Scenario 1

If you have a `row_num` column or alias, SQL Server will throw:

```
SQLSTATE[42000]: [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]The column 'row_num' was specified multiple times for 'temp_table'
```

#### Scenario 2

If you try to insert the offset results into the same table by using `chunk`:

```php
db()->connection('db_1')->table('example')->chunk(500, function($chunk){
    db()->connection('db_2')->table('example')->insert($chunk);
});
```

`row_num` is added to the results and is an unexpected column, causing SQL server to throw:

```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'row_num' in 'field list'
```

Original PR: https://github.com/laravel/framework/pull/44355

### DIFF
--- a/src/Illuminate/Database/Query/Grammars/SqlServerGrammar.php
+++ b/src/Illuminate/Database/Query/Grammars/SqlServerGrammar.php
@@ -289,7 +289,7 @@ class SqlServerGrammar extends Grammar
      */
     protected function compileOver($orderings)
     {
-        return ", row_number() over ({$orderings}) as row_num";
+        return ", row_number() over ({$orderings}) as temp_row_num";
     }
 
     /**
@@ -333,7 +333,7 @@ class SqlServerGrammar extends Grammar
     {
         $constraint = $this->compileRowConstraint($query);
 
-        return "select * from ({$sql}) as temp_table where row_num {$constraint} order by row_num";
+        return "select * from ({$sql}) as temp_table where temp_row_num {$constraint} order by temp_row_num /* remove_temp_row_num */";
     }
 
     /**
--- a/src/Illuminate/Database/Query/Processors/SqlServerProcessor.php
+++ b/src/Illuminate/Database/Query/Processors/SqlServerProcessor.php
@@ -9,6 +9,24 @@ use Illuminate\Database\Query\Builder;
 class SqlServerProcessor extends Processor
 {
     /**
+     * Process the results of a "select" query.
+     *
+     * @param  \Illuminate\Database\Query\Builder  $query
+     * @param  array  $results
+     * @return array
+     */
+    public function processSelect(Builder $query, $results)
+    {
+        if(str_ends_with($query->toSql(), '/* remove_temp_row_num */')){
+            array_map(function($row){
+                unset($row->temp_row_num);
+            }, $results);
+        }
+
+        return $results;
+    }
+
+    /**
      * Process an "insert get ID" query.
      *
      * @param  \Illuminate\Database\Query\Builder  $query
--- a/src/Illuminate/Database/Query/Processors/SqlServerProcessor.php
+++ b/src/Illuminate/Database/Query/Processors/SqlServerProcessor.php
@@ -17,8 +17,8 @@ class SqlServerProcessor extends Processor
      */
     public function processSelect(Builder $query, $results)
     {
-        if(str_ends_with($query->toSql(), '/* remove_temp_row_num */')){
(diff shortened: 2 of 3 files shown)


dunhamjared:
Hi @staudenmeir, @yansern and @garygreen -- do you both have any input on this issue and solution? 

Or should we instead use `OFFSET` and `FETCH` instead of the current `row_number()` implementation? 
This should remove the need for a `row_num` column as well. 

From the community: 

- https://github.com/laravel/ideas/issues/1006 
- https://github.com/laravel/framework/issues/20814 - Same issue given in example "Scenario 2".
- https://github.com/staudenmeir/laravel-cte/issues/35 - soultion it so only select columns from the table the person is trying to query so `row_num` will not be returned
- https://github.com/laravel/framework/issues/7568 -

staudenmeir:
HI @dunhamjared,

> Renamed temporary column row_num to temp_row_num to help reduce conflicts
> Removes the temp_row_num column from the results if offset used

Renaming the column definitely makes sense. Not sure if we do it consistently, but we use `laravel_` as a prefix for internal columns in other places.

I would say that this is a breaking change and needs to go into the next major release. There could be people that actually use the `row_num` in their applications. Definitely an edge case, but we can't rule it out completely.

> Or should we instead use OFFSET and FETCH instead of the current row_number() implementation?

What's the minimum required SQL Server version for that?

dunhamjared:
@staudenmeir 

> I would say that this is a breaking change and needs to go into the next major release. There could be people that actually use the row_num in their applications. Definitely an edge case, but we can't rule it out completely.

Completely agree, I will create a new PR to master. 

> What's the minimum required SQL Server version for that?

2012 ([source](https://dbadiaries.com/new-t-sql-features-in-sql-server-2012-offset-and-fetch)), below our minimum support version. Woo! 

> Renaming the column definitely makes sense. Not sure if we do it consistently, but we use laravel_ as a prefix for internal columns in other places.

Good to know, thanks!

It also seems that `OFFSET/FETCH` is generally faster and less expensive than `row_number()` - [source](https://www.mssqltips.com/sqlservertip/2696/comparing-performance-for-different-sql-server-paging-methods/)

staudenmeir:
> 2012 ([source](https://dbadiaries.com/new-t-sql-features-in-sql-server-2012-offset-and-fetch)), below our minimum support version. Woo!

That's good. Laravel only supports 2017+, but I’m sure that quite a few people use Laravel with older versions (since almost all query features still work there). So Laravel is technically “allowed” to rely on 2017+ features, but it still needs to be considered that people aren't happy when pagination no longer works on their SQL Server 2008/2012 app (even though it happened because of a major update).

dunhamjared:
Working on new PR for master branch.

PR #44381, opened 2022-09-29
Uploaded the router, made the env function on the route to specify the environments the user wants the route to run in

AhmedSalheia:
I made this to make it easier to the developer to use multi-stagging on the router package ... as I do it mostly in every project I make, so instead of doing it like this 
`if (config('app.env')==='local'||config('app.env')==='testing'|) Route::get('/', fn () => view('welcome'));`
I make it like this
`Route::get('/', fn () => view('welcome'))->env('local','testing');`
and made it also on route groups:
`Route::group([
    'env'  => [ 'local', 'testing' ],
], function () {
    ....
});`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ Thumbs.db
 /.idea
 /.vscode
 .phpunit.result.cache
+.gitCommitMessage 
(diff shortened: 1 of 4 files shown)


taylorotwell:

Issue #44380, opened 2022-09-29
PostgreSQL updateFrom() is not applying global scopes


peanutsmcgee:
- Laravel Version: 9.22.1
- PHP Version: 8.0.21
- Database Driver & Version: PostgreSQL 13.2

### Description:
When performing an update using `->updateFrom()` (as opposed to `->update()`), global scopes are not being applied to the query.

It looks like normally the eloquent method `Illuminate\Database\Eloquent\Builder::update()` is called, which applies scopes and then forwards to `Illuminate\Database\Query\Builder::update()`. But there is no `Illuminate\Database\Eloquent\Builder::updateFrom()` method, so calling `->updateFrom()` goes directly to `Illuminate\Database\Query\Builder::updateFrom()`, without having the global scopes applied.

### Steps To Reproduce:

1. Define a global scope on a model.
2. Run an `updateFrom` query using that model.

driesvints:
Thanks. I sent in a PR for this: https://github.com/laravel/framework/pull/44391

PR #39151, opened 2021-10-08
[8.x] Re-add update from support for PostgreSQL

driesvints:
This PR re-introduces "update ... from ..." support for the PostgreSQL grammar. Since there wasn't a good option or straight way to re-implement the behavior after the refactor in #29393 I opted to add a new dedicated `updateFrom` method specifically for the PostgreSQL grammar.

This implementation is a bit cumbersome and crude as people will need to specifically use `updateFrom` instead of `update` if they want the correct behavior. Ideally this would be done automatically.

I've tested this with the queries from #33265 and got them back working with this. I also re-added the original tests that were removed.

Before merging this I'd like to:

- Get a review from @staudenmeir
- Get feedback from people who reported #33265
- Maybe clean up the code in the PostgreSQL grammar

If we get feedback from someone who can think of a better way to solve this then gladly. Otherwise I think this is the only solution for the problem at hand.

Fixes #33265

PR #44391, opened 2022-09-30
[9.x] Apply scopes for updateFrom

driesvints:
This PR implements updateFrom which got re-introduced in https://github.com/laravel/framework/pull/39151 in a similar way on the Eloquent Builder. This way, any global scope will be applied.

See https://github.com/laravel/framework/issues/44380

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1001,6 +1001,17 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Update records in the database.
+     *
+     * @param  array  $values
+     * @return int
+     */
+    public function updateFrom(array $values)
+    {
+        return $this->toBase()->updateFrom($this->addUpdatedAtColumn($values));
+    }
+
+    /**
      * Insert new records or update the existing ones.
      *
      * @param  array  $values


peanutsmcgee:
I've run into an issue with bindings being incorrectly ordered here, using the following test case.
```
public function testUpdateFromMethodWithSubJoinsOnPostgres()
{
	$builder = $this->getPostgresBuilder();
	$builder->getConnection()->shouldReceive('update')->once()->with('update "users" set "email" = ?, "name" = ? from (select * from "orders" where "orders"."id" = 1) as "sub" where "name" = ? and "users"."id" = "sub"."user_id"', [1, 'foo', 'bar', 'baz'])->andReturn(1);
	$result = $builder->from('users')
		->joinSub(function ($q) {
			$q->from('orders')
				->where('orders.id', 1);
		}, 'sub', 'users.id', '=', 'sub.user_id')
		->where('name', 'baz')
		->updateFrom(['email' => 'foo', 'name' => 'bar']);
	$this->assertEquals(1, $result);
}
```



based off of `testUpdateFromMethodWithJoinsOnPostgres()` and `testJoinSub()` in `Illuminate\Tests\Database\DatabaseQueryBuilderTest.php`

The test is returning the bindings in the order `['foo', 'bar', 'baz', 1]` (notice the placement of the "1").

driesvints:
Ow, I think I remember that we specifically didn't want global queries to apply here. I'm sorry but this isn't supported.

peanutsmcgee:
@driesvints I've found a solution that's been working for what I need, using an overwritten `PostgresGrammar.php` file. I'm not sure if it's good enough for a PR, but I figured I'd share just in case it's useful to others.

In short, instead of re-ordering the bindings, we can re-order the `where` statements.
```
// this bit from your commit
public function updateFrom(array $values)
{
	return $this->toBase()->updateFrom($this->addUpdatedAtColumn($values));
}


// no longer needs the re-ordering of bindings, which makes it identical to prepareBindingsForUpdate()
// I could also just change the method updateFrom() in Builder.php, but it's convenient for me to keep my changes to a single file
public function prepareBindingsForUpdateFrom(array $bindings, array $values)
{
        return $this->prepareBindingsForUpdate($bindings, $values);
}


// same general idea as the original function, but places $joinWheres before $baseWheres
protected function compileUpdateWheres(Builder $query)
{
	$baseWheres = $this->compileWheresToArray($query); // method from Grammar.php

	$joinWheres = $this->compileUpdateJoinWheresToArray($query); // new method (alternative to compileUpdateJoinWheres())

	$allWheres = array_merge($joinWheres, $baseWheres);

	return $this->concatenateWhereClauses($query, $allWheres); // method from Grammar.php
}

// replacement method for compileUpdateJoinWheres()
// basically the same, just without imploding the results
protected function compileUpdateJoinWheresToArray(Builder $query)
{
	$joinWheres = [];

	foreach ($query->joins as $join) {
		foreach ($join->wheres as $where) {
			$method = "where{$where['type']}";

			$joinWheres[] = $where['boolean'] . ' ' . $this->$method($query, $where);
		}
	}

	return $joinWheres;
}

```

driesvints:
Maybe yeah. I just remember a lot of apps broke last time we modified this method.

PR #44384, opened 2022-09-29
[9.x] Make `$path` optional in Filesystem methods

innocenzi:
This PR makes the `$path` parameter optional in the following methods:
- `FilesystemAdapter#putFile`
- `FilesystemAdapter#putFileAs`
- `UploadedFile#store`
- `UploadedFile#storeAs`
- `UploadedFile#storePublicly`
- `UploadedFile#storePubliclyAs`

Since https://github.com/laravel/framework/pull/44105 (thanks Frank), I've been using the `scoped` driver in order to be able not to repeat paths throughout my codebases. It allows me to do the following:

```php
// Before
Storage::disk('s3')->putFile('chat/attachments', $uploadedFile);
//                           ~~~~~~~~~~~~~~~~~~  This could be repeated in some places

// After
Storage::disk(Disk::ChatAttachments)->putFile('', $uploadedFile)
```

With this PR, it's possible to omit the first argument, so one could do this instead:

```php
Storage::disk(Disk::ChatAttachments)->putFile($uploadedFile)
```

PR #44395, opened 2022-09-30
[10.x] Make $path optional in Filesystem methods

innocenzi:
Follow-up of https://github.com/laravel/framework/pull/44384, since this contains breaking changes to method signatures, this PR is targeted at `master` instead of `9.x`.

---

This PR makes the `$path` parameter optional in the following methods:
- `FilesystemAdapter#putFile`
- `FilesystemAdapter#putFileAs`
- `UploadedFile#store`
- `UploadedFile#storeAs`
- `UploadedFile#storePublicly`
- `UploadedFile#storePubliclyAs`

Since https://github.com/laravel/framework/pull/44105 (thanks Frank), I've been using the `scoped` driver in order to be able not to repeat paths throughout my codebases. It allows me to do the following:

```php
// Before
Storage::disk('s3')->putFile('chat/attachments', $uploadedFile);
//                           ~~~~~~~~~~~~~~~~~~  This could be repeated in some places

// After
Storage::disk(Disk::ChatAttachments)->putFile('', $uploadedFile)
```

With this PR, it's possible to omit the first argument, so one could do this instead:

```php
Storage::disk(Disk::ChatAttachments)->putFile($uploadedFile)
```

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemAdapter.php
+++ b/src/Illuminate/Filesystem/FilesystemAdapter.php
@@ -329,17 +329,13 @@ class FilesystemAdapter implements CloudFilesystemContract
     /**
      * Write the contents of a file.
      *
-     * @param  \Psr\Http\Message\StreamInterface|\Illuminate\Http\File|\Illuminate\Http\UploadedFile|string|resource|string  $path
-     * @param  \Psr\Http\Message\StreamInterface|\Illuminate\Http\File|\Illuminate\Http\UploadedFile|string|resource|array|null  $contents
+     * @param  string  $path
+     * @param  \Psr\Http\Message\StreamInterface|\Illuminate\Http\File|\Illuminate\Http\UploadedFile|string|resource  $contents
      * @param  mixed  $options
      * @return string|bool
      */
-    public function put($path, $contents = null, $options = [])
+    public function put($path, $contents, $options = [])
     {
-        if (is_null($contents) || is_array($contents)) {
-            [$path, $contents, $options] = ['', $path, $contents ?? []];
-        }
-
         $options = is_string($options)
                      ? ['visibility' => $options]
                      : (array) $options;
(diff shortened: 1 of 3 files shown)


ar2r:
It will be more usefull like this, but i dont need that:
```
Storage::disk('s3')->putFiles('chat/attachments', [$uploadedFile1,$uploadedFile2]);
```

Your code allows you to upload files without controlling the directory and the level of access to this file. the path can be defined in a constant and you can reuse it every time, instead of a string.

innocenzi:
I'd rather the constant be the disk, that's why I do that. The path is defined in the disk definition, so, yes, I do control the directory.

taylorotwell:
Just curious - why would the `put` method not allow this?

innocenzi:
That's an omission. Added it 👍

src/Illuminate/Filesystem/FilesystemAdapter.php
```diff
@@ -329,13 +329,17 @@ protected function fallbackName($name)
     /**
      * Write the contents of a file.
      *
-     * @param  string  $path
-     * @param  \Psr\Http\Message\StreamInterface|\Illuminate\Http\File|\Illuminate\Http\UploadedFile|string|resource  $contents
+     * @param  \Psr\Http\Message\StreamInterface|\Illuminate\Http\File|\Illuminate\Http\UploadedFile|string|resource|string  $path
+     * @param  \Psr\Http\Message\StreamInterface|\Illuminate\Http\File|\Illuminate\Http\UploadedFile|string|resource|mixed  $contents
```
driesvints, line 333: `mixed` can never be combined with another type.

innocenzi: Sorry about that - fixed.

milwad-dev (commented):
Good

PR #44401, opened 2022-09-30
[8.x] Remove Google Fonts from default exception views

joshuablum:
Same as https://github.com/laravel/framework/pull/40926 but for Laravel 8.

From the Statamic Discord ([link to message](https://discord.com/channels/489818810157891584/489828514648817668/1025399669347594311)):

> Laravel 8 Default Error Pages load Google Fonts:
> A client of mine got a cease and desist letter from a law firm today - for using Google Fonts. I was pretty certain that we only use local fonts. 
> TLDR: The Laravel error:minimal layout loads Google Fonts. That means: As soon as there is an error on your site, you are liable to this. Cost: 170.00 EUR and the client is extremely unsettled of course.

Translated from 🇩🇪


driesvints:
Laravel 8.x is closed for anything but security fixes.

joshuablum:
Thanks, for the quick response! I understand the decision. Nonetheless this will be a huge pain for a lot of users (who are going to complain to the devs, agencies etc). Sketchy law firms will take advantage of this (like they do when you have no site notice or privacy statement; required by law in Germany, Austria, and a lot of other EU countries)

PR #44404, opened 2022-09-30
check event listeners for shouldQueue{trait} methods

coleshirley:
I found in my project I was duplicating code to determine whether the listener `shouldQueue`. I started looking into adding this duplicated code to traits and thought this might be a useful dx feature. Essentially I am copying this livewire functionality https://laravel-livewire.com/docs/2.x/traits but for event listeners
```php
class DoThingListener implements ShouldQueue {
    use MyCondition;

    public function shouldQueue($event) {
        // my specific condition
    }
}

trait MyCondition {
    public function shouldQueueMyCondition($event) {
        // my trait condition
    }
}
```


taylorotwell:

PR #44412, opened 2022-10-01
[9.x] Add ext-gmp to composer suggest

krsriq:
Running the tests in `Illuminate\Tests\Integration\Database\EloquentModelCustomCastingTest` without the `ext-gmp` extension installed leads to multiple `Call to undefined function Illuminate\Tests\Integration\Database\gmp_init()` errors.

This PR adds `ext-gmp` to composer's suggested packages.


taylorotwell:
Feels weird to suggest to end users an extension only needed to run tests.

PR #44421, opened 2022-10-02
[9.x] Add a new function `withSoftDeletes` in routes

michaelnabil230:
Laravel allows models to have soft delete to not remove from the row form database. to add the date the resource is deleted.
This contribution allows a developer to create a resource route that supports soft delete routes.

For ex:
```php
Route::resource('users', UserController::class)->withSoftDeletes();
```


siarheipashkevich:
Please check https://laravel.com/docs/9.x/routing#implicit-soft-deleted-models

michaelnabil230:
@siarheipashkevich 

‎‏In fact it doesn’t look like that feature. In short, it creates the links you want to restore or trashed in your data.

ankurk91:
https://laravel-code.tips/you-can-call-withtrashed-on-a-route-to-accept-soft-deleted-models/
Can you rename this method to `withTrashed()` ?

michaelnabil230:
@ankurk91
But there will be problems in the freamwork. And also my PR creates its own urls trashed and restore.

PR #44422, opened 2022-10-02
[9.x] Add a new cast for `time`

michaelnabil230:
This PR is a new type for the cast.

As Laravel support `datetime` or `date` casting, it seems reasonable to add support for `time`.

HOW TO USE:

```php
class User extends Model
{
    /**
     *
     * @var array
     */
    public $casts = [
        'start_at' => 'time:H:i',
        // or
        'end_at' => 'time',
    ];
}
```


antonkomarev:
Not sure it is correct to use DateTime object for the time. It's better to use special value object LocalTime for this case. For example from brick datetime package.

PR #44424, opened 2022-10-02
[9.x] Fix Precognition headers for Symfony responses

timacdonald:
The `$response->header()` function is only available on Laravel responses. This PR fixes usage with Symfony responses, such as `BinaryFileResponse`.


bert-w:
But the docblock says ` @param  \Illuminate\Http\Request  $request`? Is it intended to somehow use this middleware with Symfony requests?

driesvints:
@bert-w a symfony response can still be sent out instead of a Illuminate http response.

PR #44429, opened 2022-10-03
[9.x] Generalizes the output result

mahmoudmohamedramadan:
This PR generalizes the number format of all methods in the `Benchmark` class.

## BEFORE this PR gets merged

```PHP
$data = [];

return Benchmark::measure(function () use ($data) {
    User::create($data);
});
```

The output will be **4.9183**

## AFTER this PR gets merged

The output for the previous code will be **3.343**

### DIFF
--- a/src/Illuminate/Support/Benchmark.php
+++ b/src/Illuminate/Support/Benchmark.php
@@ -11,9 +11,9 @@ class Benchmark
      *
      * @param  \Closure|array  $benchmarkables
      * @param  int  $iterations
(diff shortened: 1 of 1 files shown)


nunomaduro:
@taylorotwell will give his opinion, yet, I think we already format things on the `dd` method. So, we could leave the `measure` method returning an `float` so it can be used for more complex use cases.

taylorotwell:
Nuno is correct I think.

PR #44437, opened 2022-10-03
[9.x] remove extra foreach in method in RouteCollection

moharami:


moharami:
there are two foreach in this method that iterate over one array two time, and i think we could convert to one foreach without any condition

donnysim:
It changes the logic, it's more of a sort operation where it puts fallback routes last.

PR #44441, opened 2022-10-03
[9.x] Adds `Route::environment` 🐝

nunomaduro:
This pull request is a work in progress, and it proposes the addition of a minimal `environment` method to the route definition that allows the creation routes for certain environments only. In other words, sometimes, we may have those "test" routes that we want to use locally — or in staging — but not in production:

```php
Route::get('/playground', function () {
    // My test code...
})->environment('local');

Route::get('/playground', function () {
    // My test code...
})->environment(['local', 'staging']);
```

Of course, when trying to visit these routes from `production` the response back will be an `404`.

### DIFF
--- a/src/Illuminate/Http/Middleware/VerifyEnvironment.php
+++ b/src/Illuminate/Http/Middleware/VerifyEnvironment.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Illuminate\Http\Middleware;
+
+use Closure;
+use Illuminate\Contracts\Foundation\Application;
+
+class VerifyEnvironment
+{
+    /**
+     * The application instance.
+     *
+     * @var \Illuminate\Contracts\Foundation\Application
+     */
+    protected $app;
+
+    /**
+     * Create a new middleware instance.
+     *
+     * @param  \Illuminate\Contracts\Foundation\Application  $app
+     * @return void
+     */
+    public function __construct(Application $app)
+    {
+        $this->app = $app;
+    }
+
+    /**
+     * Handle an incoming request.
+     *
+     * @param  \Illuminate\Http\Request  $request
+     * @param  \Closure  $next
+     * @param  string  ...$environments
+     * @return mixed
+     */
+    public function handle($request, Closure $next, ...$environments)
+    {
+        if (! in_array($this->app->environment(), $environments, true)) {
+            return response(status: 404);
+        }
+
+        return $next($request);
+    }
+}
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -5,6 +5,7 @@ namespace Illuminate\Routing;
 use Closure;
 use Illuminate\Container\Container;
 use Illuminate\Http\Exceptions\HttpResponseException;
+use Illuminate\Http\Middleware\VerifyEnvironment;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Contracts\CallableDispatcher;
 use Illuminate\Routing\Contracts\ControllerDispatcher as ControllerDispatcherContract;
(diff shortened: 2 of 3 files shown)


mpociot:
Wouldn't it be better to not register those routes on unwanted environments in the first place, instead of always registering them but filtering them out later on?

src/Illuminate/Http/Middleware/VerifyEnvironment.php
```diff
@@ -0,0 +1,44 @@
+<?php
+
+namespace Illuminate\Http\Middleware;
+
+use Closure;
+use Illuminate\Contracts\Foundation\Application;
+
+class VerifyEnvironment
```
mateusjunges, line 8: My personal opinion but I'd use `EnvironmentVerifier` as the name of this class

jasonvarga: The names of middleware are usually English readable actions.

"Check response for modifications", "Handle CORS", "Set cache headers", etc
See: https://github.com/laravel/framework/tree/9.x/src/Illuminate/Http/Middleware

michaelnabil230:
I have been able to do such a feature and the team did not accept its integration. I don't know why, but it is very important in the coming period for developers so that they do not create links to the experience and forget it. #42723

src/Illuminate/Http/Middleware/VerifyEnvironment.php
mateusjunges: Oh yeah, I totally missed that, thanks! :sweat_smile:

timacdonald:
I'm wondering if the feature could be...

```php
Route::environmentGroup('local', function () {
    //
});
```

which would just be a shortcut for...

```php
if ($this->container->environment($environments)) {
    return Route::group($closure);
}

return new Fluent; // nop
```

Then we just don't execute the closure and there isn't a performance penalty for de-serializing etc.

A single route would still be easily registered...

```php
Route::environmentGroup('local', fn () => Route::get('playground', function () {
    // ..
});
````

**Comment by decocodes** [comment]
I am up to it. Nice addition. 😀

**Comment by MircoBabin** [comment]
What would `artisan route:cache` do with regard to this PR, when executed in a `local` environment ? 

```
APP_NAME="Laravel Project"
APP_DEBUG=true
APP_ENV=local
APP_URL=http://localhost:6001
```

I build the route cache on my own development machine (with `local` environment). And then deploy the `bootstrap/cache/routes-v7.php` to production. Because I have no `ssh` access to the production site, which is on shared hosting.

driesvints:
What about a `local.php` routes file instead?

RhysLees:
> Wouldn't it be better to not register those routes on unwanted environments in the first place, instead of always registering them but filtering them out later on?

With you on this. is better IMO to just register a separate routes file in local if you only need it for local

ksassnowski:
I think filtering out routes by using the middleware is fine during local development. But I think the `route:cache` command should be updated, too, so it doesn't include any routes that were registered for a different environment. Maybe we could add some kind of `--environment` option to `route:cache` command so the user doesn't have to muck around with their environment variables.

MircoBabin:
> I think filtering out routes by using the middleware is fine during local development. But I think the `route:cache` command should be updated, too, so it doesn't include any routes that were registered for a different environment. Maybe we could add some kind of `--environment` option to `route:cache` command so the user doesn't have to muck around with their environment variables.

I think when you issue a `route:cache` command you are always preparing for production. 
- So the default should be `--environment=production`. 
- If some truely wants to cache routes on the local development machine `--environment=local` can be used. I normally use `route:clear` on my local development machine.
- `--environment=staging` is a bit odd. Because if you forget to change the routes to `production` and go into production it won't work. And as a sidenote, changing the routes to `production` should undertake another test and staging run.

This is getting very confusing!

I think `environment` is not a good property to test. What you possible may want is `APP_DEBUG` routes, independend of the `environment`. But enabling `APP_DEBUG` on a production system is not really recommended.

devscript-abdo:
@nunomaduro it's amazing faeature

PR #44442, opened 2022-10-03
Added types for Translation

SuperDJ:
Added types to Translation. Adding types should help not only with autocomplete but with preventing errors and debugging as well.


taylorotwell:
We wouldn't really add types to one contract like this - if we were going to do this we would do it more thoroughly.

SuperDJ:
@taylorotwell I added this PR to see if there was any interest in adding types.

But would you suggest start adding types in Contracts and go from there?

PR #44453, opened 2022-10-04
[9.x] Fix `Route::withoutMiddleware()` docblock

axlon:
Tiny doc block fix


axlon:
Failing CI seems unrelated

PR #14628, opened 2016-08-04
[5.2] Allow collections to be created from objects that implement Traversable

timsweb:
As the title suggests. I'm making use of the Collection code via [tightenco/collect](/tightenco/collect), and I had a use case where my data was contained within an object that implements Iterator.

PR #44456, opened 2022-10-04
[9.x] Traversable should have priority over JsonSerializable in EnumerateValues

brunoalod:
Using an unrelated third party package I came across this problem, this exact same issue happened to someone else, as per (https://github.com/laravel/ideas/issues/2556) and was introduced in #14628.

**Problem**

EnumerateValues::getArrayableItems checks for JsonSerializable before checking for Traversable. 

This is a problem because a class implementing both JsonSerializable and Traversable will have its JSON representation used in the Collection instead of its actual iterable representantion, which should have priority. There is no need to transform a Traversable object to a JSON.

This is not an expected behaviour since a Traversable object can be traversed (no pun intended).

**Consider the following:**

```PHP
// Person.php

class Person
{
    public string $firstName;
    public string $lastName;

    public function __construct(string $firstName, string $lastName)
    {
        $this->firstName = $firstName;
        $this->lastName = $lastName;
    }

    public function getFullName(): string
    {
        return $this->firstName . ' ' .  $this->lastName;
    }
}
```

```PHP
// MyIterator.php

use ArrayIterator;
use IteratorAggregate;
use JsonSerializable;
use Traversable;

class MyIterator implements IteratorAggregate, JsonSerializable
{
    public function __construct(
        public $items,
    ) {
    }

    public function getIterator(): Traversable
    {
        return new ArrayIterator($this->items);
    }

    public function jsonSerialize(): mixed
    {
        return array_map(fn ($item) => (array) $item, $this->items);
    }
}
```

```PHP
// Example.php

$iterator = new MyIterator([
    new Person('John', 'Doe'),
    new Person('Jane', 'Smith'),
]);

$collection = collect($iterator);

$collection->first()->getFullName();
```

EnumerateValues::getArrayableItems will use MyIterator JSON representantion, and now the items in the Collection will **no longer be of type Person**, meaning that $collection->first()->getFullName() will throw `Call to a member function getFullName() on array`.

That's assumming that the implementation of jsonSerialize returns an array of the items, but it can be worse if it returns other data such as:

```PHP
public function jsonSerialize(): mixed
{
    return [
        'otherData' => ['my', 'other', 'data'],
        'data' => array_map(fn ($item) => (array) $item, $this->items),
    ];
}
```

Although this can be solved in user-land, EnumerateValues should accurately parse the items given the fact that Traversable is a native PHP feature.

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -980,12 +980,12 @@ trait EnumeratesValues
             return $items->all();
         } elseif ($items instanceof Arrayable) {
             return $items->toArray();
+        } elseif ($items instanceof Traversable) {
+            return iterator_to_array($items);
         } elseif ($items instanceof Jsonable) {
             return json_decode($items->toJson(), true);
         } elseif ($items instanceof JsonSerializable) {
             return (array) $items->jsonSerialize();
-        } elseif ($items instanceof Traversable) {
-            return iterator_to_array($items);
         } elseif ($items instanceof UnitEnum) {
             return [$items];
         }
--- a/tests/Support/SupportCollectionTest.php
+++ b/tests/Support/SupportCollectionTest.php
@@ -16,12 +16,14 @@ use Illuminate\Support\LazyCollection;
 use Illuminate\Support\MultipleItemsFoundException;
 use Illuminate\Support\Str;
 use InvalidArgumentException;
+use IteratorAggregate;
 use JsonSerializable;
 use Mockery as m;
 use PHPUnit\Framework\TestCase;
 use ReflectionClass;
 use stdClass;
 use Symfony\Component\VarDumper\VarDumper;
+use Traversable;
 use UnexpectedValueException;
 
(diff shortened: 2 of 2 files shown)


georgefromohio:
This somehow breaks when converting a ResourceCollection because it is both Traversable and JsonSerializable but we don't want it get the iterator_to_array result of a ResourceCollection... T or F?

Geoffry304:
This is giving problems when you want to collect a resource that extends JsonResource and use the toArray collect function. Only the top level is changed to an Array but the items underneath are staying as  a resource.

`<?php

namespace App\Http\Resources;

use Illuminate\Http\Resources\Json\JsonResource;

class TestResource extends JsonResource
{

    /**
     * Transform the resource into an array.
     *
     * @param Request $request
     * @return array
     */
    {
        return [
            'id' => $this->id,
            'name' => $this->name
            
        ];
    }

}
`
This is not giving the correct format:
`$collection = collect(TestResource::collection($newArray))->toArray();`

georgefromohio:
@Geoffry304, correct… I had to override this with my own function because it broke my collections.

PR #44457, opened 2022-10-04
[9.x] Have `Model::withoutTimestamps()` return the callback's return value

axlon:
https://github.com/laravel/framework/pull/44138 introduced the ability to perform actions on models without timestamps.

This PR makes `Model::withoutTimestamps()` act the same as `Model::withoutEvents()`, `Model::withoutTimestampsOn()` and various other "without" methods, by having it return the callbacks return value.

This would then allow e.g.:

```php
$user = User::withoutTimestamps(fn () => User::create($request->validated());
```


internetbuilder (commented):
Xxx

PR #44459, opened 2022-10-04
Improve return type for `Eloquent\Collection::find()`

sebdesign:
Thanks to PHPStan's conditional return types, we can now infer what does the `find` method of the Eloquent Collection returns.
When the `$key` argument is an `array` or an `Arrayable` object, it returns a new collection with the found items.
In all other cases, it returns a single item, or the `$default` value.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Collection.php
+++ b/src/Illuminate/Database/Eloquent/Collection.php
@@ -24,7 +24,7 @@ class Collection extends BaseCollection implements QueueableCollection
      *
      * @param  mixed  $key
      * @param  TFindDefault  $default
-     * @return static<TKey|TModel>|TModel|TFindDefault
+     * @return ($key is array|Arrayable ? static<TKey|TModel> : TModel|TFindDefault)
      */
     public function find($key, $default = null)
     {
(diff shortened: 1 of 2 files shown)


nunomaduro:
@sebdesign Does PHPStorm already support conditional return types?

sebdesign:
Apparently not yet: https://youtrack.jetbrains.com/issue/WI-66465/Support-parsing-of-PhpStans-conditional-return-types.
Is this a good  idea to submit to Larastan?

nunomaduro:
@sebdesign Feel free to submit to Larastan. Meanwhile, we can't break autocompletion on PHPStorm in favour of having better PHPStan support.

bastien-phi:
@nunomaduro what about using `@phpstan-return` ?

```php
     * @param  mixed  $key
     * @param  TFindDefault  $default
     * @return static<TKey|TModel>|TModel|TFindDefault
     * @phpstan-return ($key is array|Arrayable ? static<TKey|TModel> : TModel|TFindDefault)
     */
    public function find($key, $default = null)
```
This will provide both completion in IDEs and better static analysis

Issue #44463, opened 2022-10-05
Forgot password notification email is not sending if field name is not email in users table


rahulshinde19:
https://github.com/laravel/framework/blob/df474f63f5411cd178258e5333d6d997df38591b/src/Illuminate/Notifications/RoutesNotifications.php#L48

I think at this line ```$this->getEmailForPasswordReset()``` function should be called instead of ```$this->email```

rahulshinde19:
Looks like I have to add custom field email (getEmailAttribute()) in my users table.
```getEmailForPasswordReset()``` function available in other trait ```CanResetPassword``` not in ```RoutesNotifications```

**Comment by driesvints** [comment]
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

However, this issue will not be locked and everyone is still free to discuss solutions to your problem!

Thanks.

**Comment by rahulshinde19** [comment]
Found solution for this

```
public function routeNotificationForMail($notification) {
    return $this->personal_email;
}
```
adding this to module will fix this issue and will return custom email field value to send an email 
thanks to @crynobone for the solution

PR #44464, opened 2022-10-05
$this->getEmailForPasswordReset() should be called instead of $this->email

rahulshinde19:
$this->getEmailForPasswordReset() function should be called instead of $this->email

Referring to this issue #44463

I think at this line $this->getEmailForPasswordReset() function should be called instead of $this->email if the driver is 'mail' then it should return email field value but in my users table the name of email field is different. returning ```$this->email``` is restricting notify to use "email" as email field name.

### DIFF
--- a/src/Illuminate/Notifications/RoutesNotifications.php
+++ b/src/Illuminate/Notifications/RoutesNotifications.php
@@ -45,7 +45,7 @@ trait RoutesNotifications
 
         return match ($driver) {
             'database' => $this->notifications(),
-            'mail' => $this->email,
+            'mail' => $this->getEmailForPasswordReset(),
             default => null,
         };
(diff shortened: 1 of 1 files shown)


crynobone:
This is only correct for forgot password notification but wrong for everything else.

rahulshinde19:
> This is only correct for forgot password notification but wrong for everything else.

I got that, thanks for the reply.

Looks like I have to add custom field email (```getEmailAttribute()```) in my users table.
```getEmailForPasswordReset()``` function available in other trait ```CanResetPassword``` not in ```RoutesNotifications```

rahulshinde19:
@crynobone I have made some changes and raised new pull request #44466 would you like to review it?

PR #44466, opened 2022-10-05
Provide option to choose different mail field name to notify using mail driver

rahulshinde19:
Referring to issue #44463 and cancelled previous pull request #44464

New function ```$this->getEmailForMailNotification()``` should be called instead of ```$this->email```

Returning ```$this->email``` is restricting notify to use "email" as email field name.

If user want to notify through mail driver and email is stored in different field (other than email) like "personal_email"
OR
If "email" field is exist but user want to notify to some other email field like "personal_email" then he can override function ```getEmailForMailNotification``` in his model and that field will be used to notify through "mail" driver.

### DIFF
--- a/src/Illuminate/Notifications/RoutesNotifications.php
+++ b/src/Illuminate/Notifications/RoutesNotifications.php
@@ -31,6 +31,18 @@ trait RoutesNotifications
     }
 
     /**
+     * Get the e-mail address field to notify through mail.
+     * This function will provide option to choose different field name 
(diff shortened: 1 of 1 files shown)


crynobone:
Why would you need this when you can already do:

```php
public function routeNotificationForMail($notification) {
    return $this->email;
}
```

rahulshinde19:
My bad, I haven't realized that.
thanks again for the way to fix my issue

PR #44473, opened 2022-10-05
[9.x] Fixes issue using static class properties as blade attributes

### LINKED ISSUE #44465: New blade syntax breaks passing in static variables as attribute.
- Laravel Version: 9.34.0
- PHP Version: 8.10.0

### Description:

There is a backwards compatibility issue with the new blade syntax. https://twitter.com/taylorotwell/status/1577388846806929415

Using static class variables as blade attributes is no longer working. 

A work around:
Use a getter method on the class instead of the static variable.
``:options="\App\Actions\CreateNewUser::userRanges()"``

### Steps To Reproduce:

```blade
        <x-select
          name="users"
          :label="Estimated number users"
          :value="old('users')"
          :options="\App\Actions\CreateNewUser::$userRanges"
        />
```

joedixon:
The PR resolves #44465 which raises a breaking change where static class properties can no longer be passed to blade attributes since v9.32.0 when short syntax was introduced.

The fix updates the regex used to determine whether short syntax is being used.

Currently the regex looks for the sequence `:$` within the component string. This successfully determines the short syntax, but will also match calls to a static class property, for example User: _**:$**_ id

To resolve this issue, `\s` has been added to the beginning of the regex to ensure there is whitespace before the `:` character.

### DIFF
--- a/src/Illuminate/View/Compilers/ComponentTagCompiler.php
+++ b/src/Illuminate/View/Compilers/ComponentTagCompiler.php
@@ -567,7 +567,7 @@ class ComponentTagCompiler
      */
     protected function parseShortAttributeSyntax(string $value)
     {
-        $pattern = "/\:\\\$(\w+)/x";
+        $pattern = "/\s\:\\\$(\w+)/x";
 
         return preg_replace_callback($pattern, function (array $matches) {
             return ":{$matches[1]}=\"\${$matches[1]}\"";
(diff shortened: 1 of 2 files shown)


PerryvanderMeer:
Seems like this breaks a component with multiple attributes, because it removes the leading white-space character.
```blade
{{-- works --}}
<x-alert :$title/>

{{-- broken --}}
<x-alert type="info" :$title/>
```

joedixon:
> Seems like this breaks a component with multiple attributes, because it removes the leading white-space character.

Great spot @PerryvanderMeer! Just pushed an update - is the working for you now?

PerryvanderMeer:
Looks good to me @joedixon 🚀 
Maybe we should add some tests for components with multiple attributes, since all current tests rely on only one attribute?

PR #44474, opened 2022-10-05
[9.x] Partition support for migrations. (PostgreSQL supported)

ArdaKaraderi:
Additions:
- Added logic to PostgresGrammar.php for partition creation while generating tables. 
- Added logic for compile command creation. 
- Added builder link to support facade.
- Added extra parameters for dynamic partition creation while defining the migration.

In order to work with partitions, PostgreSQL requires tables to be created as "partitioned table", and it's impossible to convert a regular table into a partitioned one. Solutions on the internet show that people use raw queries to make their tables when they are using  Laravel and PostgreSQL. They lose the beauty of Laravel and its features. If this feature is implemented, users can use the new Schema static method to create their partitioned table.

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -79,6 +79,27 @@ class Blueprint
     public $after;
 
     /**
+     * Column key one for creating a composite key.
+     *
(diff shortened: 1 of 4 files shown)


taylorotwell:

PR #44480, opened 2022-10-05
[9.x] Makes Blade inline components 10x faster. ⛽️

nunomaduro:
This pull request is a work in progress, and it makes rendering Blade inline components nearly **10x faster**. So, assuming the following environment:

```php
class Button extends Component
{
    function __construct(public $number) {}

    function render() { return <<<'blade'
        <div class="alert alert-danger">
            {{ $number }}
        </div>
    blade; }
}

// and welcome.blade.php
@for ($i = 0; $i < $numberOfComponents; $i++)
    <x-button :number="$i" />
@endfor

// before:
- 1 component: 3.332ms
- 10 components: 5.010ms
- 100 components: 47.670ms
- 1000 components: 482.629ms // 0.5 seconds ...
- 10000 components: 5,046.893ms // 5 seconds ...

// after:
- 1 component: 3.296ms
- 10 components: 0.875ms
- 100 components: 7.285ms
- 1000 components: 65.646ms // 0,06 seconds
- 10000 components: 642.549ms // 0,6 seconds
```

~~Important note: this pull request assumes static content being returned from the Component::render. If we think this may not be the case for some users. We will need to adjust this pull request.~~

### DIFF
--- a/src/Illuminate/View/Component.php
+++ b/src/Illuminate/View/Component.php
@@ -99,12 +99,14 @@ abstract class Component
      */
     protected function extractBladeViewFromString($factory, $contents)
     {
-        if (isset(static::$bladeViewCache[$contents])) {
-            return static::$bladeViewCache[$contents];
+        $key = sprintf('%s::%s', static::class, $contents);
+
+        if (isset(static::$bladeViewCache[$key])) {
+            return static::$bladeViewCache[$key];
         }
 
         if (strlen($contents) <= PHP_MAXPATHLEN && $factory->exists($contents)) {
-            return static::$bladeViewCache[$contents] = $contents;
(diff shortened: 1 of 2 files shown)


JayBizzle:
Just tested this against an issue we were having with https://github.com/blade-ui-kit/blade-icons/issues/190

Based on the repo in that issue we are seeing the following improvement when rendering 1000 icons with this PR in place.

```
Before - 280.601ms
After - 173.016ms
```

👍

nunomaduro:
@JayBizzle Can you try again with the latest version of this pull request?

JayBizzle:
No difference really with the latest change, if anything, about `10ms` slower

PR #44481, opened 2022-10-05
[9.x] Add validation rules to check the between, maximum and minimum of the…

youssofOkiel:
### Our Chalange

We wanted to validate that the total size of an array of files should be less than or equal to 10 MegaBytes but there was only the max validation which is apply only to the size of one file, not an array so we came up with these brand new validation features to support this kind of validation (dealing with an array of files):

### Features

- **Min Multiple Files Size**

Parameter: min size (in bytes)
ex: if you have an array of files and you want to check that the total size of these files should be greater than or equal to size ( 2 MB) and there is no validation rule to handle this case but now you could use the "minFilesSize" function rule.

` Rule::minFilesSize(2048)`

- **Max Multiple Files Size**

Parameter: max size (in bytes)
ex: if you have an array of files and you want to check that the total size of these files should be less than or equal to size ( 10 MB) and there is no validation rule to handle this case but now you could use the "`maxFilesSize`" function rule.

` Rule::maxFilesSize(10240)`

- **Between Multiple Files Size**

Parameter: min size (in bytes), max size (in bytes)
ex: if you have an array of files and you want to check that the total size of these files should be between two sizes the min (2 MB) and max( 10 MB) and there is no validation rule to handle this case but now you could use the "`betweenFilesSize`" function rule.

`Rule::betweenFilesSize(2048, 10240)`
### Benefits

- You will find it as built-in validation rules features and you don't have to create them as custom rules.
- Now you could deal easily with an array of multi files and validate the size of these files by using one of the three rule functions

### Author
Youssof Okiel

### Co-author
Fady Ibrahim @FadySobhy

### DIFF
--- a/src/Illuminate/Validation/Rule.php
+++ b/src/Illuminate/Validation/Rule.php
@@ -8,6 +8,9 @@ use Illuminate\Validation\Rules\Dimensions;
 use Illuminate\Validation\Rules\ExcludeIf;
 use Illuminate\Validation\Rules\Exists;
 use Illuminate\Validation\Rules\In;
+use Illuminate\Validation\Rules\MaxFilesSize;
+use Illuminate\Validation\Rules\MinFilesSize;
+use Illuminate\Validation\Rules\BetweenFilesSize;
 use Illuminate\Validation\Rules\NotIn;
 use Illuminate\Validation\Rules\ProhibitedIf;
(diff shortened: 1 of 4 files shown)


taylorotwell:
I suggest just creating a custom rule - or publishing a package.

PR #44482, opened 2022-10-06
Add except method to HasAttributes trait.

franksteinberg:
Adds a method `except($attributes)` to HasAttributes.

This is roughly an inverse to the existing method, `only($attributes)` with a nearly identical signature.

It is useful in cases where you have a Model and would to get the fields as an array, however there are a limited number of fields that would otherwise be returned when calling `toArray()` that you wish to exclude.

This method can be used  in lieu of temporarily adding fields to the `$hidden` fields on the Model, and has a cleaner appearance than existing methods of performing the described filtering.

Example:

```php
// Usage
$user->except('id', 'email');

// Or
$user->except(['id', 'email']);

// Is equivalent to 
Arr::except($user->toArray(), ['id', 'email']);
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1825,6 +1825,21 @@ trait HasAttributes
     }
 
     /**
+     * Get all of the model's visible attributes except specified attributes.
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hesitant to add a common method name like this on a patch release. Feel free to add it to your own base model.

PR #44462, opened 2022-10-04
[9.x] Alternative Mailable Syntax

taylorotwell:
Currently, mailables are "configured" via a `build` method. Within this method, various methods on the parent mailable class are invoked to specify the structure of the mailable:

```php
public function build()
{
    return $this->from('example@example.com', 'Example')
                ->subject('Order Shipped')
                ->view('emails.orders.shipped');
}
```

While this works fine, over the last couple of years it has always rubbed me the wrong way. It feels a lot like an initialization method or a two-step constructor. The mailable object can't be inspected in any useful way until that method is invoked. There aren't really any other types of objects in Laravel user-land that work this way or follow this pattern.

This PR offers an alternative syntax for defining mailables in which several discrete methods may be defined that each return configuration style objects that describe the structure of the mailable:

```php
<?php

namespace App\Mail;

use Illuminate\Bus\Queueable;
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Mail\Mailable;
use Illuminate\Mail\Mailables\Address;
use Illuminate\Mail\Mailables\Attachment;
use Illuminate\Mail\Mailables\Content;
use Illuminate\Mail\Mailables\Envelope;
use Illuminate\Queue\SerializesModels;

class InvoicePaid extends Mailable
{
    use Queueable, SerializesModels;

    /**
     * Create a new message instance.
     *
     * @return void
     */
    public function __construct()
    {
        //
    }

    /**
     * Get the message envelope.
     *
     * @return \Illuminate\Mail\Mailables\Envelope
     */
    public function envelope()
    {
        return new Envelope(
            subject: 'Invoice Paid',
            cc: [new Address('foo@example.com', 'Example Name')],
            tags: [],
            metadata: [],
        );
    }

    /**
     * Get the message content definition.
     *
     * @return \Illuminate\Mail\Mailables\Content
     */
    public function content()
    {
        return new Content(
            view: 'html-view-name',
            text: 'text-view-name',
        );
    }

    /**
     * Get the attachments for the message.
     *
     * @return \Illuminate\Mail\Mailables\Attachment[]
     */
    public function attachments()
    {
        return [
            Attachment::fromPath('/path/to/file'),
        ];
    }
}
```

**Note:** Only the `content` method is really required when defining mailables in this fashion.

A more minimal example would look like the following:

```php
<?php

namespace App\Mail;

use Illuminate\Bus\Queueable;
use Illuminate\Contracts\Queue\ShouldQueue;
[...]

PR #44492, opened 2022-10-06
build Mailables when mocking mails

### LINKED ISSUE #44486: From address is not set for Mailables when using Mail:fake()
- Laravel Version: 9.34.0
- PHP Version: 8.1.10
- Database Driver & Version: MySQL 5.7 (But no database is used in the issue)

### Description:
When testing the sending of mailables using `Mail::fake()` the `from` address (and other properties) configured in `build()` is not set.
i.e it seems that `build()` is not called at all when using Mail::fake().

Is that expected behavior that `build()` is not called in test?
if that is the case the documentation at https://laravel.com/docs/9.x/mocking#mail-fake where it shows asserting from address and subjects etc does not work. 

I hope my explanation makes sense. And let me know if I can supply another example or similar.

### Steps To Reproduce:

I have a test repo here: https://github.com/hmazter/laravel-mail-test-fail
With a mail here: https://github.com/hmazter/laravel-mail-test-fail/blob/main/app/Mail/WelcomeMail.php
And a test here: https://github.com/hmazter/laravel-mail-test-fail/blob/main/tests/Feature/ExampleTest.php

TheClassified:
According to the [documentation](https://laravel.com/docs/9.x/mocking#mail-fake) you should be able to fetch the variables (from, cc, etc) defined during the build of the Mailable, however this never gets run when mocking the mail.

This PR should fix https://github.com/laravel/framework/issues/44486 and might also be relevant for https://github.com/laravel/framework/pull/44462

First time submitting a PR to Laravel - so please let me know if there's something more I can do.

### DIFF
--- a/src/Illuminate/Support/Testing/Fakes/MailFake.php
+++ b/src/Illuminate/Support/Testing/Fakes/MailFake.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Support\Testing\Fakes;
 
 use Closure;
+use Illuminate\Container\Container;
 use Illuminate\Contracts\Mail\Factory;
 use Illuminate\Contracts\Mail\Mailable;
 use Illuminate\Contracts\Mail\Mailer;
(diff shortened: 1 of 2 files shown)


taylorotwell:
No plans to change this logic on a patch release. People will soon be able to test this stuff easier.

PR #44500, opened 2022-10-07
Add support to collections for using dot notation in get functions

grantholle:
This adds support for using dot notation on `$collection->get()` functions.

```php
$collection = collect([
    'a' => [
        'b' => 'c',
    ],
]);
$collection->get('a.b'); // c
```

For normal collections, it's pretty straightforward. My hesitation is with lazy collections. I'm not sure if my implementation is good enough. Looking forward to feedback!


grantholle:
This isn't ready.

stephan-v:
Has this been picked up in another PR? This sounds like something which should have been in there ages ago.

grantholle:
I don't think so. I kind of forgot and gave up...

PR #44506, opened 2022-10-07
Add loginOnce

jreason:
Once a user has been manually authenticated log the user in once using the current AuthenticatableContract.

Otherwise, if a user has been manually authenticated, using onceUsingId() will make another database call.


jreason:
Having looked at the code in more detail, i guess that setUser is the same as logging in once?

PR #44509, opened 2022-10-07
[9.x] use `Str::of` method instead of re-calling the Str class multiple

michaelnabil230:
I believe this PR is better than before, especially since this is a Laravel core method,
also its shorter and more readable this way,

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/Factory.php
+++ b/src/Illuminate/Database/Eloquent/Factories/Factory.php
@@ -572,7 +572,7 @@ abstract class Factory
      */
     protected function guessRelationship(string $related)
     {
-        $guess = Str::camel(Str::plural(class_basename($related)));
+        $guess = Str::of(class_basename($related))->plural()->camel();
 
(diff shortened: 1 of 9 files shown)


src/Illuminate/Database/Eloquent/Factories/Factory.php
```diff
@@ -591,11 +591,13 @@ public function hasAttached($factory, $pivot = [], $relationship = null)
             'has' => $this->has->concat([new BelongsToManyRelationship(
                 $factory,
                 $pivot,
-                $relationship ?? Str::camel(Str::plural(class_basename(
+                $relationship ?? Str::of(class_basename(
```
Jubeki, line 594: You could use `Str::of(...)->classBasename()`

https://laravel.com/docs/9.x/helpers#method-fluent-str-class-basename
(There are also some other places)

taylorotwell:
Nothing is broken.

michaelnabil230:
@taylorotwell I'm working to check these issues. Give me some time

PR #44511, opened 2022-10-07
Adds `strong` password validation rule

stebogit:
I think it would be beneficial - as well as convenient - to provide a "recommendable" strong password rule (that hopefully most people already implement anyway), that would probably evolve in time as technology change. Nowadays an 8-characters password for example is just [not OK anymore](https://www.techrepublic.com/article/how-an-8-character-password-could-be-cracked-in-less-than-an-hour) for any reasonably secure system.


taylorotwell:
Feel free to define in your own application. Will be a honeypot of opinionated PRs for me.

PR #44513, opened 2022-10-07
[9.x] Throw IllegalOperatorAndValueException

imanghafoori1:
Throw a special type of exception instead of the generic `\InvalidArgumentException`.
The second commit converts closures to more compact arrow functions in the same test file.
- It tries to be fully backward-compatible by keeping the same message and extending the original exception class.


taylorotwell:
Not sure this is necessary?

PR #44514, opened 2022-10-07
[9.x] Register `cutInternals` casters for particularly noisy objects

inxilpro:
This is a much smaller version of #44408 that addresses the major pain point of using `dd()` in Laravel—dumping an object that holds a reference to the container (or another potentially large object that you don't actually care about from a debugging context).

By default, Symfony registers the `StubCaster::cutInternals` caster for many internal classes, including the Symfony container and event dispatcher. This caster cuts down the object when dumping, but only if it's nested. This PR registers that same caster for a couple of Laravel interfaces that tend to be very large.

For example, currently if you were to call `dd()` on an instance of the `Router`, your output will include include about 500 lines showing the `Dispatcher` and another 500 lines showing the `Container` before you get anywhere near the `$routes` or other values that you're likely trying to view.

After this PR, those 1000 lines are collapsed to:

```
Illuminate\Routing\Router {
  #events: Illuminate\Events\Dispatcher { …20 }
  #container: Illuminate\Container\Container { …19 }
}
```

The `cutInternals` caster only applies if the object is nested, though. So if you call `dd()` directly on the container or event dispatcher, the entire object will be dumped.

This feels like 70% of the benefits of [Laravel Dumper](https://github.com/glhd/laravel-dumper) for about 1% of the maintenance cost.

### DIFF
--- a/src/Illuminate/Foundation/Providers/FoundationServiceProvider.php
+++ b/src/Illuminate/Foundation/Providers/FoundationServiceProvider.php
@@ -81,11 +81,11 @@ class FoundationServiceProvider extends AggregateServiceProvider
      */
     public function registerDumper()
     {
+        AbstractCloner::$defaultCasters[ConnectionInterface::class] = [StubCaster::class, 'cutInternals'];
         AbstractCloner::$defaultCasters[Container::class] = [StubCaster::class, 'cutInternals'];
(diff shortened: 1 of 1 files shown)


taylorotwell:
Thanks!

PR #44524, opened 2022-10-09
[9.x] Add value directive to Blade

danie1net0:
This PR adds a short diretive `value` to assign values to the 'input' and 'option' elements in Blade.

Current sintaxy:
`<input type="text" value="{{ $someValue }}"/>`

New possible sintaxy:
`<input type="text" @value($someValue)/>`

or

Current sintaxy:
```
<option value="{{ $someValue }}"/>
  Foo Bar
</option>
```

New possible sintaxy:
```
<option @value($someValue)/>
  Foo Bar
</option>
```

If it's merged, I'll create the PR for the docs.

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesInputs.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesInputs.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Illuminate\View\Compilers\Concerns;
+
+trait CompilesInputs
+{
+    /**
+     * Compile the value of 'input' or 'select' HTML elements.
+     *
+     * @param $value
+     * @return string
+     */
+    protected function compileValue($value)
+    {
+        $expression = $this->stripParentheses($value);
+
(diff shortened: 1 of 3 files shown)


src/Illuminate/View/Compilers/Concerns/CompilesInputs.php
```diff
@@ -0,0 +1,19 @@
...
+     * Compile the value of 'input' or 'select' HTML elements.
+     *
+     * @param $value
+     * @return string
+     */
+    protected function compileValue($value)
+    {
+        $expression = $this->stripParentheses($value);
+
+        return "<?php echo 'value=\"' . $expression . '\"'; ?>";
```
rodrigopedra, line 17: should wrap `$expression` into the `e()` helper to avoid XSS

rodrigopedra:
- Should we add one directive to each HTML attribute? 
- What about data attributes?

medilies:
Why not a shorter directive `@v`

taylorotwell:

PR #44528, opened 2022-10-10
 [9.x] Ability to disable route scope bindings

stevethomas:
Per the [docs](https://laravel.com/docs/9.x/routing#implicit-model-binding-scoping):

> When using a custom keyed implicit binding as a nested route parameter, Laravel will automatically scope the query to retrieve the nested model by its parent using conventions to guess the relationship name on the parent

The problem is that there are situations where the models are unrelated, but we would still prefer to use a custom key. 

Consider an eCommerce app with the route `stores/{seller}/{category:slug}`; a seller store can be browsed by category, but the`$seller->categories` relationship does not exist, because categories are shared amongst all sellers. 

I came across another completely different [use case here](https://github.com/laravel/framework/discussions/39643).

The two workarounds i've found are:
- define a custom route parameter like `categoryForStore` and have that query remove the parent/child scope
- ~~pass the `scope_bindings` attribute manually like `Route::get('foo', ['uses' => ..., 'scope_bindings' => false]`~~ (UPDATE - actually providing a false value to `scope_bindings` only started working with this PR due to the change to this if statement: https://github.com/laravel/framework/pull/44528/files#diff-614b34c8eaac77de1d4faf100a7133e526cfe1350ce4f02e9f60c24a88179c77L49)

This PR provides a more fluent `withoutScopeBindings()` method which disables child parameter scoping logic. It does this by setting `scope_bindings` to false in the route `$action` array, and then checking whether scope bindings has been disabled in the `ImplicitRouteBinding` class. 

This shouldn't break anything, because `scope_bindings` can only be set using `->scopeBindings()` on the route, so by internally checking whether it is set first, we can ascertain whether the user is explicitly wanting to modify the behaviour of scope bindings. 

With this change, the store route above can now be defined as:

```php
Route::get('stores/{seller}/{category:slug}', ShowStoreComponent::class)
    ->name('stores.show')
    ->withoutScopeBindings();
```

I also believe I fixed a bug, which was a reference to `scoping` instead of `scope_bindings` in an integration test; I couldn't find any reference to a `scoping` attribute anywhere in routing.

### DIFF
--- a/src/Illuminate/Routing/ImplicitRouteBinding.php
+++ b/src/Illuminate/Routing/ImplicitRouteBinding.php
@@ -46,7 +46,7 @@ class ImplicitRouteBinding
                         ? 'resolveSoftDeletableRouteBinding'
                         : 'resolveRouteBinding';
 
-            if ($parent instanceof UrlRoutable && $route->getAction('scope_bindings') !== false && ($route->enforcesScopedBindings() || array_key_exists($parameterName, $route->bindingFields()))) {
+            if ($parent instanceof UrlRoutable && ! $route->preventsScopedBindings() && ($route->enforcesScopedBindings() || array_key_exists($parameterName, $route->bindingFields()))) {
                 $childRouteBindingMethod = $route->allowsTrashedBindings() && in_array(SoftDeletes::class, class_uses_recursive($instance))
                             ? 'resolveSoftDeletableChildRouteBinding'
                             : 'resolveChildRouteBinding';
--- a/src/Illuminate/Routing/ImplicitRouteBinding.php
+++ b/src/Illuminate/Routing/ImplicitRouteBinding.php
@@ -46,7 +46,7 @@ class ImplicitRouteBinding
                         ? 'resolveSoftDeletableRouteBinding'
                         : 'resolveRouteBinding';
 
-            if ($parent instanceof UrlRoutable && ($route->enforcesScopedBindings() || array_key_exists($parameterName, $route->bindingFields()))) {
+            if ($parent instanceof UrlRoutable && $route->getAction('scope_bindings') !== false && ($route->enforcesScopedBindings() || array_key_exists($parameterName, $route->bindingFields()))) {
                 $childRouteBindingMethod = $route->allowsTrashedBindings() && in_array(SoftDeletes::class, class_uses_recursive($instance))
                             ? 'resolveSoftDeletableChildRouteBinding'
                             : 'resolveChildRouteBinding';
--- a/src/Illuminate/Routing/ImplicitRouteBinding.php
+++ b/src/Illuminate/Routing/ImplicitRouteBinding.php
@@ -46,7 +46,9 @@ class ImplicitRouteBinding
                         ? 'resolveSoftDeletableRouteBinding'
                         : 'resolveRouteBinding';
 
-            if ($parent instanceof UrlRoutable && ! $route->preventsScopedBindings() && ($route->enforcesScopedBindings() || array_key_exists($parameterName, $route->bindingFields()))) {
+            if ($parent instanceof UrlRoutable &&
+                ! $route->preventsScopedBindings() &&
+                ($route->enforcesScopedBindings() || array_key_exists($parameterName, $route->bindingFields()))) {
                 $childRouteBindingMethod = $route->allowsTrashedBindings() && in_array(SoftDeletes::class, class_uses_recursive($instance))
                             ? 'resolveSoftDeletableChildRouteBinding'
                             : 'resolveChildRouteBinding';
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -1124,6 +1124,18 @@ class Route
     }
 
     /**
+     * Indicate that the route should not enforce scoping of multiple implicit Eloquent bindings.
+     *
+     * @return $this
+     */
+    public function withoutScopeBindings()
+    {
+        $this->action['scope_bindings'] = false;
+
+        return $this;
+    }
+
+    /**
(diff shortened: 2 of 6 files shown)


src/Illuminate/Routing/ImplicitRouteBinding.php
```diff
@@ -46,7 +46,7 @@ public static function resolveForRoute($container, $route)
                         ? 'resolveSoftDeletableRouteBinding'
                         : 'resolveRouteBinding';
 
-            if ($parent instanceof UrlRoutable && ($route->enforcesScopedBindings() || array_key_exists($parameterName, $route->bindingFields()))) {
+            if ($parent instanceof UrlRoutable && ! $route->preventsScopedBindings() && ($route->enforcesScopedBindings() || array_key_exists($parameterName, $route->bindingFields()))) {
```
rodrigopedra, line 49: can't we just add the logic for preventing scopeBinding to `enforcesScopedBinding()`?

From my understanding of below implementation both would never be true the same time, right?

implementation can be almost the same as you have for `preventsScopedBindings()`, only by changing the equality comparison from `false` to `true`

rodrigopedra:
I like the fluent addition. And I think there is more good and no harm on this addition.

But, from the current alternatives you listed as a justification, there are other already clean options you might not be aware of.

If `Category` is always routable by its `slug` field, you can add this method to the `Category` class:

~~~php
public function getRouteKeyName()
{
    return 'slug';
}
~~~

Then in the routes you want to have scoped binding, you can keep using `{category:slug}`. But for routes you don't want it you can just use `{category}` and it will use the `slug` field by default, while not scoping it to any parent model.

reference: https://laravel.com/docs/9.x/routing#customizing-the-key

Another one, similar to your first bullet item, is to use explicit binding.

I am listing it here due to the final part of the item, where you say *"...and have that query remove the parent/child scope"*. This statement didn't make it clear to me if you were referring to explicit binding, please apologize if I misunderstood it.

Add this to the beginning of  your `RouteServiceProvider@boot` method:

~~~php
Route::model('categoryBySlug', fn($value) => Category::firstWhere('slug', $value));
~~~

Then, just for completeness, you can use the `{categoryBySlug}` route parameter and have this custom resolution logic applied.

reference: https://laravel.com/docs/9.x/routing#explicit-binding

src/Illuminate/Routing/ImplicitRouteBinding.php
stevethomas: > From my understanding of below implementation both would never be true the same time, right?

Don't think so because `enforcesScopedBindings()` is going to return false if `scope_bindings` has not been set (the default), but `array_key_exists($parameterName, $route->bindingFields())` generally returns true so it still evaluates with child binding. 

Are you suggesting something like:
```php
public function enforcesScopedBindings()
{
    return ! isset($this->action['scope_bindings']) || $this->action['scope_bindings'] === true;
}

// ...

if ($parent instanceof UrlRoutable && $route->enforcesScopedBindings() && array_key_exists($parameterName, $route->bindingFields()))) {
```

That might work and allow the removal of `preventsScopedBindings`.

stevethomas:
thanks @rodrigopedra, I did consider both of those options and still felt this addition is more useful overall. 

> ```
> public function getRouteKeyName()
> {
>     return 'slug';
> }
> ```
I find hardcoding this in the model to be limiting / requiring other workarounds downstream. For example, a web app would typically want to use `slug`, whereas the API would typically want to use `id`. 

> ```
> Route::model('categoryBySlug', fn($value) => Category::firstWhere('slug', $value));
> ```
I agree this is pretty easy, but it is not as easy / obvious as specifying a custom key in the route. The other annoyance (unless i'm missing something) is that `public function __construct(Category $category)` now must be `public function __construct(Category $categoryBySlug)`.

src/Illuminate/Routing/ImplicitRouteBinding.php
rodrigopedra: You are right. I oversaw this `OR` condition on the `if` clause.

It still feels strange to have both methods that should just be the other's inverse.

Maybe you can propose this new implementation of `Route@enforcesScopedBindings()` :

~~~php
public function enforcesScopedBindings($parameterName = null)
{
    if (isset($this->action['scope_bindings'])) {
        return $this->action['scope_bindings'] === true;
    }
    
    if ($parameterName) {
        return array_key_exists($parameterName, $this->bindingFields());
    }
    
    return false;
}
~~~

Then you can simplify the `if` condition to:

~~~php
if ($parent instanceof UrlRoutable && $route->enforcesScopedBindings($parameterName)) {
~~~

But this would be a breaking change (public method signature/arity changed) and thus would need to be sent to master instead.

Otherwise, I think your current implementation is the best solution -- for adding this feature -- for now.

rodrigopedra:
> The other annoyance (unless i'm missing something) is that `public function __construct(Category $category)` now must be `public function __construct(Category $categoryBySlug)`

Indeed, parameter name should match when using explicit binding, which is an annoyance.

src/Illuminate/Routing/ImplicitRouteBinding.php
stevethomas: I'm still getting green tests when combining `enforcesScopedBindings()` and `preventsScopedBindings()` as detailed in my previous comment, and it does feel more intuitive to have the single method. 

My concern is that `enforcesScopedBindings()` would now return `true` if `scope_bindings` is not set, whereas previously it would return `false`. 

That won't matter for the single usage in the framework, but i'm not sure whether people would be using this method for some reason in the wild. 

For this reason I think the extra method ensures backwards-compatibility; but happy to refactor to the cleaner version if requested.

PR #44530, opened 2022-10-10
[9.x] Ability to guess policy from container

michael-rubel:
## About

When using authorization, you often end up writing `@can` like this:

```php
@can('useDashboard', App\Models\Company::class)
    ...
@endcan
```

But currently, it's impossible to guess policy based on the container binding. This PR adds the possibility to do this:
```php
@can('useDashboard', Company::class)
    ...
@endcan
```

If you have aliases for models, you'll be able to use them when resolving policies.
It shouldn't break anything, because in case of failed resolution we just do nothing.

---


taylorotwell:

michael-rubel:
Hey @taylorotwell, I just sent an improved implementation to the current fork. We can just check if it has `app()->bound()` in this case. Of course, if you're not completely against this feature.

PR #44531, opened 2022-10-10
[9.x] Correct channel matching

taylorotwell:
Laravel previously had some unexpected behavior (imo) when matching channels. This was raised to me via email by a community member.

If you have:

```php
Broadcast::channel(
    'App.Interaction.{uuid}',
    static function ($user, $uuid) {
        return true;
    },
);
```

And you attempt to subscribe to channel `App.Interaction.some-uuid.Internal`, the channel will match; however, the injected parameter will just be "some-uuid"... ".Internal" will be removed. This is because the channel matching in Laravel was using a fairly naive Str::is check which did not match the regular expression logic of the parameter injection. IMO the logic should be the same between the two. So, I have updated the channel matching regular expression to match the parameter injection regular expression.

In theory this could break applications that expected this route to match; however, those routes were already receiving a sliced parameter that may not be what they expect. So, I am viewing this as a bug. Will revisit if community feedback pushes back on release.

There was a test that showed this broken behavior but it appears to be added just out of thoroughness by a community member 4 years ago when the test was first added. Not for a specific use case.

### DIFF
--- a/src/Illuminate/Broadcasting/Broadcasters/Broadcaster.php
+++ b/src/Illuminate/Broadcasting/Broadcasters/Broadcaster.php
@@ -372,6 +372,8 @@ abstract class Broadcaster implements BroadcasterContract
      */
     protected function channelNameMatchesPattern($channel, $pattern)
     {
-        return Str::is(preg_replace('/\{(.*?)\}/', '*', $pattern), $channel);
+        return preg_match('/'.preg_replace('/\{(.*?)\}/', '([^\.]+)', $pattern).'$/', $channel);
+
+        // return Str::is(preg_replace('/\{(.*?)\}/', '*', $pattern), $channel);
     }
 }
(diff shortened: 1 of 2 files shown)


PhiloNL:
This did "break" in one of my apps by throwing 403 errors, but I agree this is a bug.
I did not fully understand the description at first so if anyone is looking for another example:

```php
Broadcast::channel('team.{team}', function ($user, App\Models\Team $team) {
    return (int) $user->currentTeam->id === (int) $team->id;
});

// Subscribing to the following channel would work in < v9.35.0
// team.{$teamId}.products.{$productId}
// As a partial match was found for "team.{$teamId}" and ".products.{$productId}" is ignored.

// In => v9.35.0, you have to explicitly define the full route, so besides the channel above, you would need the following:
Broadcast::channel('team.{team}.products.{product}', function ($user, App\Models\Team $team, \App\Models\Product $product) {
    return (int) $user->currentTeam->id === (int) $product->team_id;
});
```

timacdonald:
I have created a PR to address this: https://github.com/laravel/framework/pull/45692

PR #44533, opened 2022-10-10
[9.x] Delete commented out code introduced in #44531

BrandonSurowiec:
Also removed the `Illuminate\Support\Str` import as there are no more references to that class.


BrandonSurowiec:
Test failures are unrelated to the PR.

PR #44536, opened 2022-10-10
[9.x] Add Ask component for termwind way terminal commands

stickpro:
Created a component for ask
use code to use

` $this->components->ask('Question?'); `

in your Command class

image


nunomaduro:
The `ask` function will be removed from Termwind in the future, and we don't have plans ( at least for now ) to add this feature to the core of these new components. Thank you for your contribution!

PR #44538, opened 2022-10-11
[9.x] Add validation for object

ben221199:
This pull request adds support for validating objects using the validator.


ben221199:
Could be if the input is JSON with no associative array and you receive the following:
```json
{
	"myStringField": "string",
	"myIntegerField": 1234,
	"myFloatField": 1.234,
	"myArrayField": [],
	"myObjectField": {},
}
```

I had this problem once, when receiving such data and I had to explicitly check on being an (empty) object. Maybe this validator isn't used by many, but it seems reasonable to me to have this function available for the ones that need it.

PR #44541, opened 2022-10-11
[9.x] Update schedule:work usleep from 0.1 seconds to 0.5 seconds

michaelr0:
Howdy,

Its a minor thing, but do we need to poll every 0.1 second in `artisan schedule:work`? there could be a specific reason that I'm not aware of?

Its not a super scientific test, but I tested with a schedule to run `$schedule->command('inspire')->everyMinute();`, then tested the following sleep configurations for a total of 10 minutes.

```
100 = 0.1 seconds
500 = 0.5 seconds
1000 = 1 second
```

There is a minor decrease in CPU utilization shown in `htop` for both the `0.5 second` and `1 second` test
image

Personally I would be inclined to shoot for a straight 1 second sleep in the loop, but I figured there is probably a reason why this was initially not set as such, in which case 0.5 might be a nice middle ground.

Having said that, I'd understand if this PR gets knocked back because of how minor 0.1% cpu usage is, at the very least I figure its worth a shot to see if the sleep timer is worth discussion.

### DIFF
--- a/src/Illuminate/Console/Scheduling/ScheduleWorkCommand.php
+++ b/src/Illuminate/Console/Scheduling/ScheduleWorkCommand.php
@@ -47,7 +47,7 @@ class ScheduleWorkCommand extends Command
         [$lastExecutionStartedAt, $keyOfLastExecutionWithOutput, $executions] = [null, null, []];
 
         while (true) {
-            usleep(100 * 1000);
+            usleep(500 * 1000); // 0.5 seconds
 
             if (Carbon::now()->second === 0 &&
                 ! Carbon::now()->startOfMinute()->equalTo($lastExecutionStartedAt)) {


src/Illuminate/Console/Scheduling/ScheduleWorkCommand.php
```diff
@@ -47,7 +47,7 @@ public function handle()
         [$lastExecutionStartedAt, $keyOfLastExecutionWithOutput, $executions] = [null, null, []];
 
         while (true) {
-            usleep(100 * 1000);
+            usleep(500 * 1000); // 0.5 seconds
```
rodrigopedra, line 50: I guess this is a good opportunity to move to "Numeric Literal Separator" instead of using a multiplication for thousands notation.

~~~php
usleep(500_000);
~~~

reference: https://wiki.php.net/rfc/numeric_literal_separator

By the way, I guess it is less than a second so the next if clause can be more precise as it checks for the "zero-th" second.

bert-w:
Kinda hard to gauge what the performance benefit is if 0.1% CPU could be 0.05 (rounded up) and 0.0% CPU could be 0.049 (rounded down). That being said I dont know what the reason is for 100ms intervals.

taylorotwell:
Just going to leave it as is if it's not broken and not hurting anything severely.

PR #44543, opened 2022-10-11
[10.x] Remove option from `make:rule` re-added by mistake

rodrigopedra:
PR #44016 re-added the `--invokable` option to the `MakeRuleCommand` after it was removed by PR #43868

As Invokable rules will be the default in 10.x, this PR removes this option again.


rodrigopedra:
By the way, would a PR allowing current/regular class rule creation be considered?

I have many rules where I use the translator within the `message()` method, and when having multiple "failure" points (think of early returns for multiple assessment), repeating the failure message + translation boilerplate when it has placeholders is cumbersome.

I think everyone will use the newer invokable design more, but I guess keeping the old-school style as an easy to create alternative presents no harm.

If this would be considered, I can send a PR.

rodrigopedra:
Don´t know why tests are failing. This PR just removes a line for an unused command option.

PR #44545, opened 2022-10-11
[10.x] Uses PHP Native Type Declarations 🐘

nunomaduro:
Please check https://github.com/laravel/laravel/pull/6010 for details about this pull request.

### DIFF
--- a/src/Illuminate/Contracts/Auth/UserProvider.php
+++ b/src/Illuminate/Contracts/Auth/UserProvider.php
@@ -19,7 +19,7 @@ interface UserProvider
      * @param  string  $token
      * @return \Illuminate\Contracts\Auth\Authenticatable|null
      */
-    public function retrieveByToken($identifier, string $token);
+    public function retrieveByToken($identifier, $token);
 
     /**
      * Update the "remember me" token for the given user in storage.
--- a/src/Illuminate/Contracts/Auth/UserProvider.php
+++ b/src/Illuminate/Contracts/Auth/UserProvider.php
@@ -10,7 +10,7 @@ interface UserProvider
      * @param  mixed  $identifier
      * @return \Illuminate\Contracts\Auth\Authenticatable|null
      */
-    public function retrieveById($identifier);
+    public function retrieveById(mixed $identifier);
 
     /**
      * Retrieve a user by their unique identifier and "remember me" token.
@@ -19,7 +19,7 @@ interface UserProvider
      * @param  string  $token
      * @return \Illuminate\Contracts\Auth\Authenticatable|null
      */
-    public function retrieveByToken($identifier, $token);
+    public function retrieveByToken(mixed $identifier, string $token);
 
     /**
      * Update the "remember me" token for the given user in storage.
@@ -28,7 +28,7 @@ interface UserProvider
      * @param  string  $token
      * @return void
      */
-    public function updateRememberToken(Authenticatable $user, $token);
+    public function updateRememberToken(Authenticatable $user, string $token);
 
     /**
      * Retrieve a user by the given credentials.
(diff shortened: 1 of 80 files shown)


SuperDJ:
@nunomaduro I might have some that help you: https://github.com/SuperDJ/framework/tree/master-contract-types and https://github.com/laravel/framework/pull/44442

src/Illuminate/Contracts/Auth/UserProvider.php
```diff
@@ -10,7 +10,7 @@ interface UserProvider
      * @param  mixed  $identifier
      * @return \Illuminate\Contracts\Auth\Authenticatable|null
      */
-    public function retrieveById($identifier);
+    public function retrieveById(mixed $identifier);
```
browner12, line 13: Wouldn't this more accurately be typed as `string|int`?

I think if we're doing this, now's a good time to get slightly more explicit about what types we're accepting, where applicable.

There are probably a number of cases where `mixed` is accurate because we're accepting everything under the sun, we should take a closer look at some of these.

nunomaduro: We will consider it. Thanks!

JredLewis (commented):
The markdown mail stub should also be updated with this return doc block for consistency.

src/Illuminate/Contracts/Database/Eloquent/CastsAttributes.php
```diff
@@ -2,27 +2,19 @@
...
 interface CastsAttributes
 {
     /**
      * Transform the attribute from the underlying model values.
-     *
-     * @param  \Illuminate\Database\Eloquent\Model  $model
-     * @param  string  $key
-     * @param  mixed  $value
-     * @param  array  $attributes
-     * @return mixed
```
driesvints, line 14: The `@return mixed` needs to be kept here.

src/Illuminate/Contracts/Validation/DataAwareRule.php
```diff
@@ -10,5 +10,5 @@ interface DataAwareRule
      * @param  array  $data
```
driesvints, line 10: Remove this line.
```suggestion
```

src/Illuminate/Contracts/Validation/InvokableRule.php
```diff
@@ -12,5 +14,5 @@ interface InvokableRule
      * @param  \Closure(string): \Illuminate\Translation\PotentiallyTranslatedString  $fail
```
driesvints, line 14: Remove these params.

src/Illuminate/Contracts/Validation/ValidatorAwareRule.php
```diff
@@ -10,5 +12,5 @@ interface ValidatorAwareRule
      * @param  \Illuminate\Validation\Validator  $validator
```
driesvints, line 12: Remove this param.
```suggestion
```

driesvints (changes requested):
Highlighted a few places where docblocks still need to be updated but there seem to be many more in this PR yet. So best to go through those once more.

src/Illuminate/Foundation/Console/stubs/event.stub
```diff
@@ -27,10 +25,12 @@ class {{ class }}
     /**
      * Get the channels the event should broadcast on.
      *
-     * @return \Illuminate\Broadcasting\Channel|array
+     * @return array<int, \Illuminate\Broadcasting\Channel>
      */
-    public function broadcastOn()
+    public function broadcastOn(): array
```
RTippin, line 30: I believe this should be a return type of `array|Channel` as the current `BroadcastEvent` still wraps the return of this method.

https://github.com/laravel/framework/blob/b856dc5e8bd65a4d9ef1afada998855b52388c30/src/Illuminate/Broadcasting/BroadcastEvent.php#L71

Haythamasalama:
I love it ❤️❤️

foremtehan:
https://github.com/laravel/framework/pull/45314 (Does not have types)

PR #44547, opened 2022-10-11
[9.x] Add helpers for accessor and mutator

percymamedy:
Add two new helper methods to facilitate creating accessors and mutators in Eloquent models.

Example:

```php
<?php
 
namespace App\Models;
 
use App\Support\Address;
 
class User extends Model
{
    /**
     * Interact with the user's full name.
     *
     * @return \Illuminate\Database\Eloquent\Casts\Attribute
     */
    protected function fullName(): Attribute
    {
        return accessor(fn ($value, $attributes) => $attributes['first_name'].' '.$attributes['last_name']);
    }

    /**
     * Interact with the user's address.
     *
     * @return \Illuminate\Database\Eloquent\Casts\Attribute
     */
    protected function address(): Attribute
    {
        return mutator(
            get: fn ($value, $attributes) => new Address(
                $attributes['address_line_one'],
                $attributes['address_line_two'],
            ),
            set: fn (Address $value) => [
                'address_line_one' => $value->lineOne,
                'address_line_two' => $value->lineTwo,
            ],
        );
    }
}
```

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -12,6 +12,7 @@ use Illuminate\Contracts\Routing\UrlGenerator;
 use Illuminate\Contracts\Support\Responsable;
 use Illuminate\Contracts\Validation\Factory as ValidationFactory;
 use Illuminate\Contracts\View\Factory as ViewFactory;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Define in your own application if you wish.

PR #44555, opened 2022-10-12
Improve `Request` instance datetime parsing by introducing default value

rognales:
### Why 
Recently I bumped into the needs to add fallback value to the $request->date() helper.

As mentioned here https://github.com/laravel/framework/pull/39945#issuecomment-1059835329, there's a need for a default/fallback value which will then make it consistent with the rest of the parser (`boolean()`,` float()`, `integer()`, `string()` & `input()` )

### Usage

Previously 

````php
$startDate = Request::date('from', 'Ymd') ?? now()
````

Now

````php
$startDate =  $request->date('from', null, null, now())
````

Note: this is my first open source contributions, in the spirit of hacktoberfest.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -352,14 +352,15 @@ trait InteractsWithInput
      * @param  string  $key
      * @param  string|null  $format
      * @param  string|null  $tz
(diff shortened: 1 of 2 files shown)


ankurk91:
Should type hint with native Datetime interfere instead

taylorotwell:

PR #44438, opened 2022-10-03
[9.x] Register `MigrationServiceProvider` in `DatabaseServiceProvider`

jerodev:
Currently, it is not possible to remove the `DatabaseServiceProvider` in `app/config.php` because the `MigrationServiceProvider` requires `db` to be registered in the container, and thus an error is thrown.  
Right now, `MigrationServiceProvider` is loaded and resolved through `ConsoleSupportServiceProvider`.

This PR moves the registration of `MigrationServiceProvider` to `DatabaseServiceProvider`, making it possible to remove the `DatabaseServiceProvider` in projects that do not require a database.

PR #44559, opened 2022-10-12
[10.x] Move MigrationServiceProvider to DatabaseServiceProvider

jerodev:
Currently, it is not possible to remove the `DatabaseServiceProvider` in `app/config.php` because the `MigrationServiceProvider` requires `db` to be registered in the container, and thus an error is thrown.  
Right now, `MigrationServiceProvider` is loaded and resolved through `ConsoleSupportServiceProvider`.

This PR moves the registration of `MigrationServiceProvider` to `DatabaseServiceProvider`, making it possible to remove the `DatabaseServiceProvider` in projects that do not require a database.

Ref #44438


taylorotwell:
Hmm, I'm not sure I am totally compelled by this atm. It makes MigrationServiceProvider an eagerly loaded provider when it used to be deferred and lazy loaded.

PR #44568, opened 2022-10-13
Update: Fix Error spatie/laravel-missing-page-redirector

vdvcoder:
Since update from Laravel 9.32 to 9.33 the package spatie/laravel-missing-page-redirector did not work anymore.

This should fix it.

### DIFF
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -18,6 +18,7 @@ use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
 use Laravel\SerializableClosure\SerializableClosure;
 use LogicException;
+use ReflectionFunction;
 use Symfony\Component\Routing\Route as SymfonyRoute;
 
 class Route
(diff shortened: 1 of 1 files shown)


rodrigopedra:
Sent PR https://github.com/spatie/laravel-missing-page-redirector/pull/78 to `spatie/laravel-missing-page-redirector` as an alternative to address this same issue.

@timacdonald you may want to look on the reasoning on that PR. 

For short, as PR #44339 introduced the `Illuminate/Routing/Contracts/CallableDispatcher`, empty containers -- such as the default to `Illuminate\Routing\Router` when no container instance is provided to its constructor -- would fail to dispatch closure based routes as they won't know how to build this interface.

I believe this is kind of a breaking change, as maybe other packages using private Router instances, for similar reasons as the package linked above, could start failing to dispatch closure based routes.

As this ship has been sailed already, maybe we could consider changing this line on `Illuminate\Routing\Router`'s constructor:

https://github.com/laravel/framework/blob/a1611c76ffd9221978378f361af00fac2fe0b4aa/src/Illuminate/Routing/Router.php#L138

To 

~~~php
$this->container = $container ?: Container::getInstance();
~~~

So at least when no container is provided it uses the global one instead, and avoid this issue. What do you think?

rodrigopedra:
Or the alternative would be dropping the new `Illuminate/Routing/Contracts/CallableDispatcher` interface, as the introduced `Illuminate\Foundation\Routing\PrecognitionCallableDispatcher` is a sub-class for this new interface only subclass `Illuminate\Routing\CallableDispatcher`

Instead of binding to the interface, and dispatching an interface that might be unbounded on the container, we ask the container for the concrete class, and PrecognitionServiceProvider can alias/rebind its subclass when needed.

taylorotwell:
Looks like an alternative was merged to the package itself.

PR #44573, opened 2022-10-13
fix $cast defined at Model class does not use timezone defined at config/app.php

usuyuki:
> **Note**  
>  I am contributing for the first time.

# Abstruct
Regardless of timezone in config/app.php, it converted into string specified in $casts on Model with utc when toJson method use on Illuminate\Database\Eloquent\Model instance.

# Details
I know it is intended to use UTC inside of the framework.
But I think it should converted into string specified in $casts on Model with utc when toJson method use on Illuminate\Database\Eloquent\Model instance.

And I know we should use UTC basically as written in the docuement.
But the value is different when converting to 'Y-m-d' like this and when calling with ModelInstance->dateValue.

## Example

config/app.php
```
'timezone' => 'Asia/Tokyo'
``` 

database/migrations/xxxxx_create_model_names_table.php
```
$table->date('dateValue')->comment('Save in Y-m-d format');
```

in the datetimeValue column stored on the database
```
2022-09-30
```

In this situation, datetimeValue will be "2022-09-30 00:00:00.0 Asia/Tokyo (+09:00)" with the Carbon instance.
And after using toJson method on Illuminate\Database\Eloquent\Model, it become "2022-09-29T15:00:00.000000Z" on default.
At this point, it is suitable data.

But when we defined in $casts in app/Models/ModelName.php like this
```
 protected $casts = [
        'dateValue' => 'datetime:Y-m-d',
    ];
```
After using toJson method on Illuminate\Database\Eloquent\Model, it become "2022-09-29".

But when I write code like below.
```
use App\Models\ModelName;

ModelName::where('id',1)->dateValue;
```
It become "2022-09-30".

It's not only this difference, but also formated by toJson method data cannot be changed to the correct timezone because it's not contain hours.
So, toJson method with defined in $casts should use timezone defined in config/app.php

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -296,7 +296,7 @@ trait HasAttributes
 
             if (isset($attributes[$key]) && ($this->isCustomDateTimeCast($value) ||
                 $this->isImmutableCustomDateTimeCast($value))) {
-                $attributes[$key] = $attributes[$key]->format(explode(':', $value, 2)[1]);
+                $attributes[$key] = $attributes[$key]->timezone(config('app.timezone'))->format(explode(':', $value, 2)[1]);
             }
 
             if ($attributes[$key] instanceof DateTimeInterface &&


jmarcher:
This is a huge breaking change. I do not see the gain in changing this on the current release (or even in future releases) UTC datetimes are the way to go, you could implement an attribute mutator for the needed models.

GrahamCampbell:
I don't think this change is correct, and either way, we can't just jam a foundation helper function call into the middle of the code.

Usuyuki:
I'm so sorry for my lack of consideration toward this change.

PR #44577, opened 2022-10-13
[9.x] Fix error message for incorrect rendering of components due to missing closing tag

### LINKED ISSUE #44553: Misleading error message for incorrect rendering of  components.
Screenshot 2022-10-12 at 08 19 34

### Steps To Reproduce:

1. Create a component are described in [docs](https://laravel.com/docs/9.x/blade#anonymous-index-components)

Example:

`resources/views/components/reuse.blade.php`
```html
<div>
	Hello! I'm a Component!
</div>
```
2. Render the components in another view.
Example:
`resources/views/welcome.blade.php`
```html

<x-reuse  > 
<!-- Instead of <x-reuse  /> --!>
```

3. The message display when the view is return. 

Screenshot 2022-10-12 at 08 28 51

I think a better error message explaining that there is a missing character on the syntax could be more helpful.

AamirSohailKmAs:
This PR will close #44553 

While compiling blade components tags to valid PHP it checks for opening and closing tags, If a component has an opening tag But missing a closing tag then it'll through an expectation saying `Missing Closing Tag for component [foo]`.

### DIFF
--- a/src/Illuminate/View/Compilers/ComponentTagCompiler.php
+++ b/src/Illuminate/View/Compilers/ComponentTagCompiler.php
@@ -86,6 +86,15 @@ class ComponentTagCompiler
      */
     public function compileTags(string $value)
     {
+        $hasOpeningTag = preg_match($this->getOpeningTagsPattern(), $value, $matches);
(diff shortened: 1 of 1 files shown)


taylorotwell:

PR #44584, opened 2022-10-14
[9.x] Adds `inverse` method

mahmoudmohamedramadan:
This PR introduces an `inverse` method that enables us to inverse the boolean column value with a given condition, for instance, If we have a system that switches the user status if was an `enabled` or `disabled`, the implementation of the previous scenario would be like so,

```PHP
/**
 * Update the specified resource in storage.
 *
 * @param  \App\Models\User  $user
 */
public function update(Request $request, User $user)
{
    $data = $request->validate([
        'name' => ['required'],
        'email' => ['required', 'unique:users,email,' . $user->id],
    ]);

    $toUpdate = array_merge($data, ['status' => !$user->status]);

    $user->update($toUpdate);

    return redirect()->back()->with('success', 'Data was updated.');
}
```

## AFTER this PR gets merged

We can easily use the `inverse` method to inverse the user's status with a given column's name that we want to inverse its value in addition, we can provide the `condition` to inverse the status with some `extra` columns to be updated with the status.

```PHP
/**
 * Update the specified resource in storage.
 *
 * @param  \App\Models\User  $user
 */
public function update(Request $request, User $user)
{
    $data = $request->validate([
        'name' => ['required'],
        'email' => ['required', 'unique:users,email,' . $user->id],
    ]);

    $user->inverse('status', $data);

    return redirect()->back()->with('success', 'Data was updated.');
}
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1083,6 +1083,21 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Inverse the column's value by a given condition.
+     *
+     * @param  string|\Illuminate\Database\Query\Expression  $column
+     * @param  bool|null  $condition
+     * @param  array  $extra
+     * @return int
+     */
+    public function inverse($column, $condition = null, array $extra = [])
+    {
+        return $this->toBase()->inverse(
+            $column, $condition, $this->addUpdatedAtColumn($extra)
+        );
+    }
+
+    /**
      * Add the "updated at" column to an array of values.
      *
      * @param  array  $values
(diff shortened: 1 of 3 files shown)


dennisprudlo:
You could also do this, it is a bit cleaner than in the example:
```php
$user->update($data + ['status' => !$user->status]);
```

I don't know how to feel about the fact that `$user->inverse(...)` (I think `$user->toggle(...)` would be a better name) saves the updated value instantly. What about toggling multiple fields? You will have to call this method multiple times resulting in multiple queries.

mahmoudmohamedramadan:
> ```
> $user->update($data + ['status' => !$user->status]);
> ```

From my point of view, the way that I used is more practical and clear, and according to the method name, I think that @taylorotwell decides that.

BrandonSurowiec:
Usually toggle-like methods are not a good idea and can have some side-effects. Would look at the comments of this past PR: https://github.com/laravel/framework/pull/41958

taylorotwell:

PR #44217, opened 2022-09-20
[9.x] Short attribute syntax for Blade Components

pascalbaljet:
This PR implements [Caleb's proposal](https://twitter.com/calebporzio/status/1568700635683627008) to optionally allow a shorter syntax for passing attributes to Blade Components.

```blade
<!-- current syntax -->
<x-profile :user-id="$userId"></x-profile>

<!-- short syntax -->
<x-profile :$userId></x-profile>
```

PR #44589, opened 2022-10-14
[9.x] Support spread operator in component tag attributes

pboivin:
This PR is inspired by https://github.com/laravel/framework/pull/44217. It is a first attempt to add support for the spread operator (`...`) in component attributes:

```blade
@php
    $header = [
        'title' => 'Lorem ipsum',
        'secondaryTitle' => 'Dolor sit amet',
    ];
@endphp

<x-header ...$header class="foo" />
```

I believe this would be equivalent to:
```blade
<x-header
    :title="$header['title']"
    :secondary-title="$header['secondaryTitle']"
    class="foo"
/>
```

Under the hood, it compiles to something closer to:
```blade
@component('components.header', array_merge($header, ['class' => 'foo'])) @endcomponent
```

I'm sending it as a draft PR because I'm not 100% confident about this implementation. It's my first time digging into the Blade internals and I'm looking to validate (or invalidate... 🙈) the idea.

I've added a few tests for the following use cases:
- slot
- component
- self-closing component
- anonymous component

Looking forward to any thoughts and suggestions!

Best,

P.

### DIFF
--- a/src/Illuminate/View/Compilers/ComponentTagCompiler.php
+++ b/src/Illuminate/View/Compilers/ComponentTagCompiler.php
@@ -258,7 +258,6 @@ class ComponentTagCompiler
                 : "'$class'";
 
             $parameters = [
-                'view' => "'$class'",
                 'view' => $view,
                 'data' => $this->mergeAttributeSpreads($this->attributesToString($data->all(), $escapeBound = false)),
             ];
@@ -730,7 +729,9 @@ class ComponentTagCompiler
     protected function mergeAttributeSpreads(string $attributeString)
     {
         if ($this->attributeSpreads) {
-            return "array_merge(".implode(",", $this->attributeSpreads).",[$attributeString])";
+            $spreads = implode(',', $this->attributeSpreads);
+
+            return "array_merge({$spreads},[{$attributeString}])";
         }
 
         return "[$attributeString]";
--- a/src/Illuminate/View/Compilers/ComponentTagCompiler.php
+++ b/src/Illuminate/View/Compilers/ComponentTagCompiler.php
@@ -48,6 +48,13 @@ class ComponentTagCompiler
     protected $boundAttributes = [];
 
     /**
+     * The "unpack:" attributes that have been compiled for the current component.
+     *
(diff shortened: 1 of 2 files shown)


pboivin:
Digging a bit deeper, I was reminded that array unpacking with string keys is a PHP 8.1 feature ([RFC](https://wiki.php.net/rfc/array_unpacking_string_keys)). Would it be worth rebasing onto `master`, for `10.x`?

pboivin:
Taking inspiration from [`spatie/laravel-blade-x`](https://github.com/spatie/laravel-blade-x/blob/master/src/Compiler.php#L286), I reworked the attribute merging using `array_merge()` instead of the `...` operator. This enables support for PHP < 8.1.

Also, this revised solution is perhaps cleaner than my first draft, because it fully extracts the `...$var` expressions from the parsed attributes instead of adding a special `unpack:` prefix. I'll mark this as ready for review, as I'm a bit more confident in the implementation details.

taylorotwell:
I think this is kinda I'm just kinda wondering if it would be all that common of a use case. I can't think of a situation where I have really needed this. What prompted it?

pboivin:
Hi @taylorotwell, thanks for the feedback. I agree that it's a bit niche... The idea came when working with a library of reusable UI components. We found ourselves preparing the data for each component at the controller level:

```php
class CareersController extends Controller
{
    public function index()
    {
        $page = CareersListing::first();

        return view('site.careersListing', [
            'hero' => $this->transformHero($page),
            'intro' => $this->transformIntro($page),
            'blocks' => $this->transformBlocks($page->blocks),
            'footer' => $this->transformFooter($page),
            // ...
        ]);
    }
}
```

and then "unpacking" the props for each component:

```blade
<x-hero.generic
    :title="$hero['title']"
    :subtitle="$hero['subtitle']"
    :media="$hero['media']"
/>
```

It's probably not the most convincing example but this pattern emerged on a few projects. My colleague tried the `...$header` syntax and reported that it did not work with component tags. The idea stuck with me.

taylorotwell:

PR #44592, opened 2022-10-14
[9.x] Make functions static improving performance and scoping

SuperDJ:
Make functions static improving performance and scoping according to https://plugins.jetbrains.com/plugin/7622-php-inspections-ea-extended-


mateusjunges (commented):
@SuperDJ The chance of performance related PRs being merged without benchmarks to prove  are not good, I think

PR #44595, opened 2022-10-14
[9.x] Add `Arr:fill()` method.

xiCO2k:
Hey,

This PR adds the capability to use `array_fill` with a callback to fill the values.

## Before

If you need to fill an array with **10** items, and change every key on it, you need to use a `foreach` to loop over each item and change the value on it.

```php
$array = collect(array_fill(1, 10, ''));
foreach ($array as $key => $value) {
    $array[$key] = $key;
}
```

## After

```php
Arr::fill(0, 10, fn ($key) => $key);
```

With this approach you would archive the same result as before, with an array that as the same `key` and `value`.

### Example

```php
Arr::fill(0, 10, 'string');
```

This will have an `array` with **10** items with the value of `string`.

Thanks,
Francisco

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -224,7 +224,7 @@ class Arr
     }
 
     /**
-     * Fill an array with values
+     * Fill an array with values.
      *
      * @param  int  $start
      * @param  int  $count
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -224,6 +224,29 @@ class Arr
     }
 
     /**
+     * Fill an array with values
+     *
+     * @param  int  $start
+     * @param  int  $count
+     * @param  mixed  $value
+     * @return array
+     */
+    public static function fill($start, $count, $value)
+    {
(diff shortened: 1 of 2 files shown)


xiCO2k:
The failing test does not look related with this feature.

taylorotwell:
I think this is basically the same thing?

```php

Collection::times(10)->mapWithKeys(
  fn ($value, $key) => [$key => $value * 2]
)->all();
```

xiCO2k:
Not the same.

If you need the `key` to start on `10` that would not work out of the box.

xiCO2k:
It will be the same if:

```php

$start = 10;
Collection::times(10)->mapWithKeys(
    fn ($value, $key) => [$key + $start => $key + $start]
 )->all();
```

but I prefer the `Arr::fill()` method since it's bound to the original `array_fill`.

rozsazoltan:
Using with `Collection::rangeByKey` (if accepted #44606):

```php
public static function fill($start, $count, $value = null)
{
  if (! is_null($value) && ! is_callable($value)) {
    $value = fn () => $value;
  }

  return Collection::rangeByKey($start, $count, $value)->toArray();
}
```

PR #44598, opened 2022-10-15
added data type returned by functions, added "static" on callback

nick-rashkevich:
added several fixes to migration templates:)


driesvints:
Feel free to export these stubs and adjust to your liking.

nick-rashkevich:
according to my proposed changes, the code will become cleaner:)

PR #44599, opened 2022-10-15
[9.x] First Data from DB get in array format

tr-shishir:
Hi there,

When we get data from $query->get(), its returning data in collection formate with toArray() we convert it into Array formate, but why we can't do the same with $query->first()?
I faced this several times and thought about the solution $query->firstToArray() and I added it here.

Thanks

### DIFF
--- a/src/Illuminate/Database/Concerns/BuildsQueries.php
+++ b/src/Illuminate/Database/Concerns/BuildsQueries.php
@@ -297,6 +297,17 @@ trait BuildsQueries
     }
 
     /**
+     * Execute the query and get the first result in Array.
+     *
+     * @param  array|string  $columns
+     * @return \Illuminate\Database\Eloquent\Model|array|static|null
+     */
+    public function firstToArray($columns = ['*'])
+    {
+        return (array)$this->take(1)->get($columns)->first();
+    }
+
+    /**
      * Execute the query and get the first result if it's the sole matching record.
      *
      * @param  array|string  $columns
--- a/src/Illuminate/Database/Concerns/BuildsQueries.php
+++ b/src/Illuminate/Database/Concerns/BuildsQueries.php
@@ -304,7 +304,7 @@ trait BuildsQueries
      */
     public function firstToArray($columns = ['*'])
     {
-        return (array)$this->take(1)->get($columns)->first();
+        return (array) $this->first($columns);
     }
 
     /**


rozsazoltan:
Current:
```php
return (array) $this->take(1)->get($columns)->first();
```

Can use current `first()` function instead of duplication:
```php
return (array) $this->first($columns);
```

tr-shishir:
> Current:
> 
> ```
> return (array) $this->take(1)->get($columns)->first();
> ```
> 
> Can use `first()` function instead of duplication:
> 
> ```
> return (array) $this->first($columns);
> ```

Thanks for the instruction :)

ankurk91:
No more helper method to builder please

zunaidmiah:
Hi,
Great idea!
Hopefully it'll really good idea. It'll help us if your solution will be merged.
Thanks

zunaidmiah:
> No more helper method to builder please

Hi,
Do you have any solution about it?
Actually From my side, Several time I also tried to get first row from database expected array type but always it return collection so then I decided to convert this to array using json_encode and decode function sometimes I made it collection then convert to array. So it's really so horrible. So if we something like that it'll really useful
Thanks Man!

dennisprudlo:
@zunaidmiah I think its because the helper method is very trivial.

> Several time I also tried to get first row from database expected array type but always it return collection

If the return value is a collection then a simple `->toArray()` will give you the array representation of the collection. If its not a collection but a model you could use the array cast just like in the implementation.

Instead of
```php
User::where(...)->firstToArray()
```
use either
```php
User::where(...)->first()->toArray()
(array) User::where(...)->first()
```

However I don't think it is a good idea to use `(array)` when the item is a model. 
```php
$model = User::where(...)->first(); // \App\Models\User

$model->toArray();
// [
//   "id" => 1,
//   "name" => "Admin",
//   "email" => "admin@example.com",
//   ...
// ]


(array) $model;
// [
//   "\0*\0connection" => "mysql",
//   "\0*\0table" => "users",
//   "\0*\0primaryKey" => "id",
//   "\0*\0keyType" => "int",
//   ...
// ]
```

src/Illuminate/Database/Concerns/BuildsQueries.php
```diff
@@ -296,6 +296,17 @@ public function first($columns = ['*'])
         return $this->take(1)->get($columns)->first();
     }
 
+    /**
+     * Execute the query and get the first result in Array.
+     *
+     * @param  array|string  $columns
+     * @return \Illuminate\Database\Eloquent\Model|array|static|null
```
dennisprudlo, line 303: Due to the `(array)` cast the function will always return an array. Model and static types are not possible here. Even `(array) null;` will result in an empty array.

PR #44600, opened 2022-10-15
[9.x] Add viteNonce directive

joostdebruijn:
This PR adds a `@viteNonce` directive to easily add the Vite nonce to script, style and link tags that are not managed by Vite.


timacdonald:
heads up this can already be achieved with…

```
<script nonce="{{ Vite::cspNonce() }}">
    //
```

PR #44604, opened 2022-10-15
[9.x] Check pivot keys existence on pivot models

imanghafoori1:
When we have a model extending `MorphPivot` and we forget to specify default values for the `foreignKey` and the `relatedKey` properties on the class, like below:

----------------------

image

The exception thrown while deleting the model (`$myModel->delete()`) will be like this: (which is not clear)

image

----------------------

- **After this change the message will be:**

image

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php
@@ -4,6 +4,7 @@ namespace Illuminate\Database\Eloquent\Relations\Concerns;
 
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Str;
+use LogicException;
 
 trait AsPivot
 {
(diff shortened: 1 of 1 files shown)


imanghafoori1:
The two properties declared on the parent model have no default values, so if they are not overridden like below it would not work properly.
image

taylorotwell:
So, with your new exception message, would the user actually have any idea what properties to add?

imanghafoori1:
I think it is visible in the stack trace, but yeah it is a good bet to include them in the message since for APIs it can be blurry.
image

taylorotwell:
But, is this documented anywhere? Even I would have no idea what to do. Why is it necessary in the first place?

imanghafoori1:
Honestly, the project I was working on was written by someone else before me, I just tried to understand the solution by studying the source code.
(The problem was : `Reaction::query()->first()->delete();` throwing an stange exception, mentioned above.)

These properties come from the `AsPivot` trait
image

taylorotwell:
If this is a problem - it is something that should be explained and documented.

PR #44606, opened 2022-10-15
[9.x] Add `Collection::rangeByKey` method

rozsazoltan:
****Can generate collection by adding range of keys.****
Inspired by: PR #44595

In mentioned PR tried to add a new array generation function that is too similar to `Collection::times` and `Collection::range` functions, but there request only values. So need a function that can generate a new Collection by taking into range of keys.

The `rangeByKey()` is more like the `range()` function, but I added a value modifier opportunity (maybe unnecessary?) as in the `times()` method, because `rangeByKey()` does not request values, just a range of keys.

```php
public static function rangeByKey($start, $number, callable $callback = null)
{ 
    return static::range($start, $number)
        ->keyBy(fn ($item) => $item)
        ->unless($callback == null)
        ->map($callback);
}
```

[Tests / Examples](https://github.com/laravel/framework/pull/44606/files)

<details>
  <summary>Collection::range() - keys can only start with 0</summary>

  First key always is 0 with `range()`:
  (values always equal our range)
  ```php
  Collection::range(-1, 1)->all();
  
  /*
    return: [
      0 => -1,
      1 => 0,
      2 => 1
    ]
  */
  ```
</details>

<details>
  <summary>Collection::rangeByKey() - keys can start with any number</summary>

  Can generate Collection with unique first key with `rangeByKey`:
  (values always equal our range, but values always equal their keys after generation)
  ```php
  Collection::rangeByKey(1, 3)->all();
  
  /*
    return: [
      1 => 1,
      2 => 2,
      3 => 3
    ]
  */
  ```
</details>

<details>
  <summary>Collection::rangeByKey() - unique starting key with manipulated values</summary>

  Can manipulate values:
  (if you need different range of values)
  ```php
  Collection::rangeByKey(1, 3, fn ($key) => $key - 2)->all();
  
  /*
    return: [
      1 => -1,
      2 => 0,
      3 => 1
    ]
  */
  ```
</details>

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -55,7 +55,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
-     * Create a new collection with default starting key by invoking the callback a given range.
+     * Create a new collection with range of key.
      *
      * @param  int  $start
      * @param  int  $number
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -55,6 +55,24 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
+     * Create a new collection with default starting key by invoking the callback a given amount of times.
+     *
+     * @template TTimesValue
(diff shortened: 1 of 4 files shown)


taylorotwell:

PR #44607, opened 2022-10-15
[9.x] Add setIf and forgetIf methods to the Arr class

olekjs:
Sometimes I need to modify an array using the Arr class.
There are cases where modifying an array depends on some condition. 

Example:
```php
$isOwner = $role === 'OWNER';

if ($isOwner) {
    Arr::set($user, 'role', $role);
}
```

It would be good if it was possible to use the if method to make the code more clean and readable. 

```php
Arr::setIf($isOwner, $user, 'role', $role);
```

Also for the basic forget method I added an option with "if"

```php
Arr::forgetIf(
    $request->boolean('remove_attributes'),
    $product,
    'attributes'
);
```

What do you think about it? Hope these methods will be useful.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -822,4 +822,39 @@ class Arr
 
         return is_array($value) ? $value : [$value];
     }
+
+    /**
+     * Set an array item to a given value using "dot" notation if the given condition is true.
+     *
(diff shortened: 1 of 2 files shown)


huangdijia:
setIf/forgetIf？？

olekjs:
@huangdijia thanks for your time on review, I have made name corrections

taylorotwell:

PR #44608, opened 2022-10-15
[9.x] TransactionCommitting

rez1dent3:
Hello. 

There was a need for a committing event, it would be great to add to the framework.

Thanks.

### DIFF
--- a/src/Illuminate/Database/Concerns/ManagesTransactions.php
+++ b/src/Illuminate/Database/Concerns/ManagesTransactions.php
@@ -43,6 +43,7 @@ trait ManagesTransactions
 
             try {
                 if ($this->transactions == 1) {
+                    $this->fireConnectionEvent('committing');
                     $this->getPdo()->commit();
                 }
 
@@ -188,7 +189,8 @@ trait ManagesTransactions
      */
     public function commit()
     {
-        if ($this->transactions == 1) {
+        if ($this->transactionLevel() == 1) {
+            $this->fireConnectionEvent('committing');
             $this->getPdo()->commit();
         }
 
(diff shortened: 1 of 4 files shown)


ankurk91:
Naming can be improved to align with other similar events

taylorotwell:

rez1dent3:
@taylorotwell Hello. The event must be done on the side of the framework. A standalone package would be difficult to maintain and keep ManagesTransactions up to date. Any change to the framework will break the package. I believe adding an event won't harm your product. For example, this event can be used in logging, in profiling tools, etc.

For example, event sourcing model. Inside the transaction, we will fill some of our state and, before the last commit, write logs with one big insert request, inside the same transaction. These are all examples, but they are very important, I think.

rez1dent3:
@taylorotwell I will describe in detail my case of using this functionality. I am developing a package for working with virtual wallets. Now I have a custom service for working with transactions in the database. When developers call `DB::beginTransaction()` services in their code, I throw an exception. Package users ask to remove this exception (for example, they have problems with laravel nova https://github.com/bavix/laravel-wallet/issues/455 https://github.com/bavix/laravel-wallet/issues/569 https://github.com/bavix/laravel-wallet/issues/570). I can't use native laravel transactions right now because before `DB::commit()`, I need to collect all requests and execute them in batches (to do this, I need to subscribe to the event before commit). To be sure of the atomicity of the written data, in case of an error, all data will be rolled back. Adding this event will help me a lot, without it there is no point in continuing to move away from our own transactions towards standardization. I understand that this is a rather narrow case. But I think that the event will not be superfluous in the framework.

PR #44612, opened 2022-10-16
[9.x] Allow clearing of stub responses of HTTP client

kielabokkie:
I think it’s quite common that when dealing with an external API, that’s used throughout your app, that you set a standard fake response for it in your tests by adding something like below to your `tests/CreatesApplication.php` file:

```php
Http::fake(['https://api.example.com/*' => Http::response(['success' => true], 200)]);
```

This way your external API always responds with something expected in your tests.

The problem is if you want to test a non-successful response (maybe `401` or `500`) you can’t overwrite this fake response so you always get the default `200` response back. This PR adds a function to clear the stub callbacks of the HTTP client that you can call before testing any response that’s different from the default fake that has been set.

For example when you want to test that your application handles a `500` response from an external API correctly in one of your tests you can do:

```php
Http::clearStubs()->fake(['https://api.example.com/some-endpoint' => Http::response(['success' => false], 500)]);
```

### DIFF
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -231,6 +231,18 @@ class Factory
     }
 
     /**
+     * Clear all the stub callbacks.
+     *
+     * @return $this
+     */
+    public function clearStubs()
+    {
+        $this->stubCallbacks = collect();
+
+        return $this;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
I would just put this in the setUp of the classes that actually fake HTTP calls.

PR #44617, opened 2022-10-16
[9.x] Adds Custom Components to Query Builder

DarkGhostHunter:
## What?

Allows to register custom instance-living components to the Query Builder, which allows to save data state or functionality, and expand on them. This implementation is in light of Deprecated Dynamic Properties in PHP 8.2, will be enforced in PHP 9.0.

For example, a macro setting a dynamic property will now have the option to use the internal `customComponent` array and helpers.

```php
use Illuminate\Database\Query\Builder;

Builder::macro('isCool', function () {
    // Before
    // return $this->coolness ?? false;

    // After
    return $this->getComponent('coolness', false);
});
```

> Considering the impact of macros setting dynamic properties in the future, a `Composable` trait could be available framework-wide for PHP 8.2 >= compatibility, or include it on the `Macroable` trait.

### DIFF
--- a/src/Illuminate/Database/Concerns/CustomComponents.php
+++ b/src/Illuminate/Database/Concerns/CustomComponents.php
@@ -1,0 +1,50 @@
+<?php
+
+namespace Illuminate\Database\Concerns;
+
+trait CustomComponents
+{
+    /**
+     * An array of custom components for the current instance.
+     *
+     * @var array<string, mixed>
+     */
+    protected $customComponents = [];
+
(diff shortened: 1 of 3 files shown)


axlon:
If this really is a problem a lot of people face, a better way to fix this would be to add the `AllowDynamicProperties` attribute, because that way theres no changes necessary for the people affected until PHP 9.

Though personally I would extend or decorate Laravel's query builder if I ever needed this.

PR #44619, opened 2022-10-17
[9.x] Add `assertServerError` to `TestResponse`.

xiCO2k:
This PR adds the `assertServerError` to the `TestResponse` class.

This will allow to use:

```php
$response->assertServerError();
```

instead of:

```php
$response->assertStatus(500);
```

It also allows for other server errors between `500` to `599`, based on here https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#server_error_responses.


lloricode:
how about 503? is it consider as server error?

PR #44620, opened 2022-10-17
add new test, that a value for a castable date can be an empty string or null

comes:
This PR fixes a bug, I found in custom castables for dates.

## Given

Imagine you have a model with a custom cast property to date. The code below will demonstrate it:

```php
class Foo extends \Illuminate\Database\Eloquent\Model {
  protected $guarded = [];
  protected $casts = ['something' => 'datetime:Y-m-d\TH:i',];
}

$foo = new Foo(['something' => '']);
```

## Expectation

I expect that I convert my model to an array or to json and the property `something` that has an empty string as a value, will be empty after conversion.

```php
dump([$foo, $foo->something, $foo->toJson()]);
```

The output I expect is:
```
[
  Foo {#2298
    something: "",
  },
  null,
  "{"something":""}",
]
```

## Bug

Instead of getting an empty value, I got the current datetime in my castable property, which is clearly wrong and manipulates the integrity of my data. The least expectation will be, that an empty field will get a random datetime, instead of an empty value

Let us add a dump to the given code above:

```php
dump([$foo, $foo->something, $foo->toJson()]);
```

The output is:
```
[
  Foo {#2298
    something: "",
  },
  Illuminate\Support\Carbon @1666016419 {#2294
    value: "2022-10-17 14:20:19",
  },
  "{"something":"2022-10-17T14:20"}",
]
```

## Solution

instead of using `Carbon::parse()` which always returns an instance of it self and in case it fails, it returns now (https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Traits/Creator.php#L211).
We switch to the more safe method `Carbon::make()`, which returns the value we gave in, when it is not possible to transform into datatime. https://github.com/briannesbitt/Carbon/blob/master/src/Carbon/Traits/Creator.php#L899-L919

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -891,10 +891,7 @@ trait HasAttributes
     protected function serializeClassCastableAttribute($key, $value)
     {
         return $this->resolveCasterClass($key)->serialize(
-            $this,
-            $key,
-            $value,
-            $this->attributes
+            $this, $key, $value, $this->attributes
         );
     }
 
@@ -1092,9 +1089,7 @@ trait HasAttributes
         [$key, $path] = explode('->', $key, 2);
 
         $value = $this->asJson($this->getArrayAttributeWithValue(
-            $path,
-            $key,
-            $value
+            $path, $key, $value
         ));
 
         $this->attributes[$key] = $this->isEncryptedCastable($key)
@@ -1229,9 +1224,7 @@ trait HasAttributes
 
         if ($value === false) {
             throw JsonEncodingException::forAttribute(
-                $this,
-                $key,
-                json_last_error_msg()
+                $this, $key, json_last_error_msg()
             );
         }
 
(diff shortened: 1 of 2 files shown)


MohammadAlhallaq:
I was really confused trying to understand how the date_field and datetime_field properties are being casted in the test class unit I notcied and TestModel1 defined at the bottom of the class 😅

taylorotwell:
Not changing any of this on a patch release.

comes:
Where should it then be changed or solved @taylorotwell? Please let me know how to address that bug correctly!

bilfeldt:
> Not changing any of this on a patch release.

@taylorotwell would it be an option to introduce a new _nullable datetime cast_ (`?datetime` for example) which returns `null` if an empty string or `null` is provided otherwise a Carbon object (which was the expected behaviour for many of us) and then merge this into the next major release so that the two different casts are essentially the same?

mateusjunges:
>Not changing any of this on a patch release.

@comes it means it will not be changed in minor releases (any 9.x release). Try sending it to `10.x`

PR #44630, opened 2022-10-18
[9.x] Allow database connection to be specified for test migrations

ziadoz:
I'm working on a project with two database connections: one is elevated with specific grants (e.g. CREATE, ALTER, DROP), whereas the other is limited (e.g. SELECT, INSERT, UPDATE, DELETE). 

The limited database connect is the default, so adding `use RefreshesDatabase` to test cases blows up, because it doesn't have the grants to migrate. 

This PR adds a `connectionToMigrate` property that allows the database connection for test migrations to be specified.


taylorotwell:
Just overwrite the `migrateFreshUsing` method.

ziadoz:
@taylorotwell I have things configured that way. But it didn't feel like a stretch to me to make it configurable using a property, seeing as `connectionsToTransact` does the same for `DatabaseTransactions`.

PR #44638, opened 2022-10-18
Enhance key:generate artisan command

imanghafoori1:
Running the `key:generate` command is usually the first thing after cloning a repo.
but if we forget to first run `cp .env.example .env` we face an ugly error like below:

image

- With this change, it will first try to create the `.env'`file by copying the `.env.example` first 
image

- And if it can not, it will print a nice little message:
image

So this way, the installation steps of laravel 9 projects will have one less step (`cp .env.example .env`) in their readme files.


taylorotwell:
I no longer accept PRs to this command as it just became a time-sink.

PR #44640, opened 2022-10-18
[9.x] add method performCommit

rez1dent3:
Hello. 

I proposed an additional event (committing #44608) in laravel, but the PR was closed. Maybe then we can at least move the commit logic into one protected method? It will greatly simplify the application logic for me.

Thanks

### DIFF
--- a/src/Illuminate/Database/Concerns/ManagesTransactions.php
+++ b/src/Illuminate/Database/Concerns/ManagesTransactions.php
@@ -42,15 +42,7 @@ trait ManagesTransactions
             }
 
             try {
-                if ($this->transactions == 1) {
-                    $this->getPdo()->commit();
-                }
-
-                $this->transactions = max(0, $this->transactions - 1);
-
-                if ($this->afterCommitCallbacksShouldBeExecuted()) {
-                    $this->transactionsManager?->commit($this->getName());
-                }
+                $this->performCommit();
             } catch (Throwable $e) {
                 $this->handleCommitTransactionException(
                     $e, $currentAttempt, $attempts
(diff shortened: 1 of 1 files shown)


taylorotwell:
What problem are you trying to solve in your application?

rez1dent3:
I am developing a package for virtual wallets. For the duration of the transaction, wallets are blocked. In order not to make unnecessary requests to the database to update the balance, I need to update wallets before commit (those inside the transaction).

Here is the pseudocode:
```php
DB::beginTransaction();
$wallet1->deposit(100);
$wallet2->deposit(100);
$wallet5->deposit(100);
DB::commit();
```

Not to do after each insert query update I need access to commit. Then the code will look like this:
```sql
start transaction;
insert into trans (wallet_id, amount, ...) values (1, 100, ...); // q1
insert into trans (wallet_id, amount, ...) values (2, 100, ...); // q2
insert into trans (wallet_id, amount, ...) values (5, 100, ...); // q3

... update case when ... then...
 commit;
```

My case is quite narrow, but I really need to have access before the commit operation.

rez1dent3:
The performCommit method implemented is an analogue of the performRollBack method, which has been in the framework for a long time.

rez1dent3:
Specifically, at the moment, I want to abandon the "own" implementation of transactions (which internally uses the standard one) in favor of the standard ones. With this change, I can add full support for laravel nova, but now there are only problems. https://github.com/bavix/laravel-wallet/issues/455 https://github.com/bavix/laravel-wallet/issues/569

taylorotwell:
Reopening other PR.

PR #44641, opened 2022-10-18
[9.x] Fix ViewException: File does not exist at path

erikn69:
**Illuminate\View\ViewException: File does not exist at path**
Maybe it was a typo
https://github.com/spatie/laravel-permission/actions/runs/3276093280/jobs/5391731843


taylorotwell:
I don't think your logic makes sense. The whole point of this check is to not check if the file is expired every time we need to render the view. Your change removes that performance benefit.

erikn69:
@taylorotwell yes, i change that now

taylorotwell:
The current state of the logic you have in this PR still does not make sense.

PR #44643, opened 2022-10-18
[9.x] Fixes blade not forgetting compiled views on `view:clear`

nunomaduro:
This pull request addresses Spatie's Permission package test suite, as they need to run the `view:clear` at runtime between assertions.

https://github.com/spatie/laravel-permission/actions/runs/3276093280/jobs/5391731843


erikn69:
Thanks

PR #44654, opened 2022-10-19
[9.x] Prevents booting providers when running `env:decrypt`

joedixon:
This PR fixes an issue with the `env:decrypt` command which can occur when a service provider relies on the presence of a value in the environment file. 

For example, the Pusher broadcast driver relies on the API token set in the environment and that token doesn't exist until the environment file is decrypted. However, when running the command, the `BroadcastServiceProvider` is booted which attempts to instantiate Pusher and an exception is thrown.

This PR aims to resolve that issue by omitting `\Illuminate\Foundation\Bootstrap\BootProviders::class` from the kernel bootstrappers **only** when invoking the `env:decrypt` command.

This functionality has been added to both the `handle` and `call` methods.

As part of this, I have also added the option to decrypt a file to a different path.

```shell
php artisan env:decrypt --path=/tmp
```

Running the command above would result decrypting a file called `.env.encrypted` to `/tmp/.env`

Of course, you may combine this with the other options:

```shell
php artisan env:decrypt --env=production --filename=.env --path=/tmp
```

The above command will decrypt `.env.production.encrypted` to `/tmp/.env`

### DIFF
--- a/src/Illuminate/Foundation/Console/EnvironmentDecryptCommand.php
+++ b/src/Illuminate/Foundation/Console/EnvironmentDecryptCommand.php
@@ -24,7 +24,7 @@ class EnvironmentDecryptCommand extends Command
                     {--env= : The environment to be decrypted}
                     {--force : Overwrite the existing environment file}
                     {--filename= : Filename to write the decrypted file contents}
-                    {--path= : Path to write the decrypted file}';
+                    {--path= : Path to write the decrypted file contents}';
 
     /**
      * The name of the console command.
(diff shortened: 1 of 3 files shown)


morloderex:
Any reason you didn't opt-in for making the BroadcastServiceProvider a deferred provider?

https://laravel.com/docs/9.x/providers#deferred-providers

joedixon:
There are a couple of reasons for that:

1. `BroadcastServiceProvider` is just one example where this issue can arise. Making it deferred doesn't resolve the issue for user providers which rely on environment values.
2. I don't actually think deferring it would resolve the issue as it's immediately used when registering the broadcast channels

PR #44657, opened 2022-10-19
[9.x] Reduce amount of iterations when searching a view in paths.

rschoonheim:
Hello Everyone,

I've noticed the following method `Illuminate\View\FileViewFinder::findInPaths()`. This function is used to determine if a given name relates to a view. When it does the path is returned, else a `InvalidArgumentException` is thrown.

This process involves 3 loops. The first loop iterates over every configured view path, the second loop appends supported file extensions to the given name, the third loop validates if a view exists. According to my calculations this will result in a total of max 16 iterations per view path the max total iterations will double.

Current iteration count

### My Calculation 
amount of view paths * amount of extensions * amount of files to validate

**Amount of view paths**
count of paths defined in the config at view.paths

**Amount of extensions**
when appending file extensions to the name we iterate over all supported file extensions

**Amount of files to validate**
size of the results from “amount of extensions”

### Suggested Change
With this PR the nesting of the `Illuminate\View\FileViewFinder::findInPaths()` method is changed. This will reduce the amount of total iterations but doesn’t change the behavior. We start by generating possible view file names (4 iterations). The next step is to iterate over the configured view paths (1 - *). 

This halves the number of iterations when a single view path is configured (from 16 to 8 iterations). When the number of view paths grows this number will change (at 7 paths from 112 to 32 iterations).

New iteration count

### My Calculation 
amount of file names * amount of view paths + iterations for generating the file names

**Amount of file names**
An array containing possible file names

**Amount of view paths** 
count of paths defined in the config at view.paths

**Iterations for generating the file names**
When generating the `Amount of file names` array a array_map is used.  So it iterates over something..  

### Expected impact
None, the methods behavior is the same. The performance impact is really low shouldn't be noticeable.  

You can find me in the discord server username: _bromano

### DIFF
--- a/src/Illuminate/View/FileViewFinder.php
+++ b/src/Illuminate/View/FileViewFinder.php
@@ -45,9 +45,9 @@ class FileViewFinder implements ViewFinderInterface
     /**
      * Create a new file view loader instance.
      *
-     * @param  \Illuminate\Filesystem\Filesystem  $files
-     * @param  array  $paths
-     * @param  array|null  $extensions
+     * @param \Illuminate\Filesystem\Filesystem $files
+     * @param array $paths
+     * @param array|null $extensions
      * @return void
      */
     public function __construct(Filesystem $files, array $paths, array $extensions = null)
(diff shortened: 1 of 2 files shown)


rschoonheim:
:warning: I assume this change doesn't have impact on behavior of the "view component". Someone with more project knowledge should help validating it..

taylorotwell:
That's a change of logic. The order of the paths is supposed to have priority.

PR #44658, opened 2022-10-19
[9.x] Add `assertBadRequest` to the `TestResponse`.

xiCO2k:
This PR adds the `assertBadRequest` to the `TestResponse` class.

This will allow to use:

```php
$response->assertBadRequest();
```

instead of:

```php
$response->assertStatus(400);
```

This one was also submitted in the past by @michael-rubel on https://github.com/laravel/framework/pull/43047.

### DIFF
--- a/tests/Testing/TestResponseTest.php
+++ b/tests/Testing/TestResponseTest.php
@@ -564,6 +564,22 @@ class TestResponseTest extends TestCase
         $response->assertUnauthorized();
     }
 
+    public function testAssertBadRequest()
+    {
+        $statusCode = 500;
+
+        $this->expectException(AssertionFailedError::class);
+
+        $this->expectExceptionMessage('Expected response status code');
+
+        $baseResponse = tap(new Response, function ($response) use ($statusCode) {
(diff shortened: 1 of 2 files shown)


rschoonheim:
Are there any other common response status assertions missing?

tests/Testing/TestResponseTest.php
```diff
@@ -564,6 +564,22 @@ public function testAssertUnauthorized()
         $response->assertUnauthorized();
     }
 
+    public function testAssertBadRequest()
+    {
+        $statusCode = 500;
```
rschoonheim, line 569: Shouldn't this be 400 to match HTTP status `400 - bad request`?

xiCO2k: for this tests no, check the other `400` methods.

rschoonheim: > for this tests no, check the other `400` methods.

Ah! Didn't noticed it. Do you know why it should be `500` and not `400`? I think it is confusing..

xiCO2k:
Got it

PR #44660, opened 2022-10-19
[9.x] Checksums for filesystem disks

frankdejonge:
Flysystem 3.8 has shipped a new capability; file checksums. This PR exposes this that functionality for Laravel ⚡️

Discussion point: This method is not added to the Filesystem interface as it would be a breaking change. Flysystem had the same problem, where the method was added to the interface using a docblock annotation, which could help with IDE discovery.

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemAdapter.php
+++ b/src/Illuminate/Filesystem/FilesystemAdapter.php
@@ -556,6 +556,24 @@ class FilesystemAdapter implements CloudFilesystemContract
     }
 
     /**
+     * Get the checksum for a file.
+     *
+     * @return string|false string checksum or false on error
+     *
+     * @throws UnableToProvideChecksum
+     */
+    public function checksum(string $path, array $options = []): string|false
+    {
+        try {
+            return $this->driver->checksum($path, $options);
+        } catch (UnableToProvideChecksum $e) {
+            throw_if($this->throwsExceptions(), $e);
+
+            return false;
+        }
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


src/Illuminate/Filesystem/FilesystemAdapter.php
```diff
@@ -842,6 +843,17 @@ public function deleteDirectory($directory)
         return true;
     }
 
+    public function checksum(string $path, array $options = []): string|false
```
michael-rubel, line 846: Docblock is missing?

frankdejonge: Added one :)

src/Illuminate/Filesystem/FilesystemAdapter.php
```diff
@@ -842,6 +843,24 @@ public function deleteDirectory($directory)
...
+     * @throws UnableToProvideChecksum
+     */
+    public function checksum(string $path, array $options = []): string|false
+    {
+        try {
+            return $this->driver->checksum($path, $options);
+        } catch (UnableToProvideChecksum $e) {
+            throw_if($this->throwsExceptions(), $e);
+
+            return false;
```
michael-rubel, line 860: I'm also wondering why to return `false`? Wouldn't it be more convenient to return `null`?
This way you can also use `?string` return type instead of union.

frankdejonge: False is in line with the other failure responses when configured not to throw exceptions.

PR #44661, opened 2022-10-19
[9.x] change status code assertions on TestResponseTest.

rschoonheim:
Hello People, 

Earlier today I noticed a pull request (#44658) containing a test case I didn't understand. The author copied a test case that was testing another assertion on the TestResponse object. I noticed there were many test cases that were expecting a status code `500 - server error` while their assertions expected something else and were passing.

This pull request changes the status codes of the test cases and removes redundant exception expectations.

### DIFF
--- a/tests/Testing/TestResponseTest.php
+++ b/tests/Testing/TestResponseTest.php
@@ -487,11 +487,7 @@ class TestResponseTest extends TestCase
 
     public function testAssertOk()
     {
-        $statusCode = 500;
-
-        $this->expectException(AssertionFailedError::class);
-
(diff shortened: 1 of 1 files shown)


taylorotwell:
Those tests are testing that assertions fail no a non matching status code though?

rschoonheim:
> Those tests are testing that assertions fail on a non matching status code though?

Didn't thought that way. My expectation by `testAssertOk` is testing of the `AssertOk` functions passes I would expect another function `testAssertOkFails` the failing behaviour. 

Might be a personal preference making it a none issue. 

Thanks @taylorotwell!

PR #44662, opened 2022-10-19
[9.x] Improve input argument parsing for commands

bert-w:
I noticed that whenever I would listen to the `CommandStarting`/`CommandFinished` events, there was some functionality missing with the `input` property on this event (`\Symfony\Component\Console\Input\InputInterface`). It appears that the input definition is not loaded at all and it causes the `getArguments()`/`getOptions()` functions to be empty. So the following code would never work:

```php
// Keep track of called commands and their arguments.
Event::listen(function (CommandStarting $event) {
    Log::info('starting ' . $event->command);
    Log::info($event->input->getOptions()); // Always `[]` even though the command has options.
    Log::info($event->input->getArguments()); // Always `[]` even though the command has arguments.
});
```

Therefore I've added a few lines of code to `Illuminate\Console\Application` to load the input definition directly from the command that is being called. The `catch` block around calling `$input->bind()` appears to be a common thing judging by the occurrences in the Symfony codebase, I'm not sure why this is necessary.

I've also just noticed by looking at this application structure, that Laravel is basically overriding some Symfony events (which DO work properly). Laravel calls its own `CommandStarting`/`CommandFinished` events while there already is existing code in Symfony that uses their own (for instance https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Console/Application.php#L1006 and https://github.com/symfony/symfony/blob/6.2/src/Symfony/Component/Console/Application.php#L1034). However, the Symfony commands cannot be listened to because there apparently is no event dispatcher attached in the Laravel override so these events are never sent.

I'm not sure of the downsides of my addition so I've submitted this as a draft. Someone with more knowledge about the bootstrapping logic should have a look at this.

### DIFF
--- a/src/Illuminate/Console/Application.php
+++ b/src/Illuminate/Console/Application.php
@@ -95,10 +95,9 @@ class Application extends SymfonyApplication implements ApplicationContract
         );
 
         try {
-            // Use command's input definition to properly parse the arguments.
             $input->bind($this->find($commandName)->getDefinition());
         } catch (ExceptionInterface) {
-            // Ignore invalid options/arguments for now.
+            // ...
         }
 
         $this->events->dispatch(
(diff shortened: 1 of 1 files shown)


driesvints:
Hi there. We fixed the failing tests on 9.x which are causing this PR to fail as well. Can you please rebase your PR with 9.x? Thanks

driesvints:
I've decided to revert this PR. Please see https://github.com/laravel/framework/pull/44888 for more info.

PR #44664, opened 2022-10-19
[9.x] Adds handlers for silently discarded and missing attribute violations

aarondfrancis:
This PR adds the ability to register handlers for two of the strict mode violations: silently discarded and missing attributes.

This ability is currently available for only the lazy load violations, this PR basically replicates that behavior for the other 2/3rds of Eloquent's new strict mode.

I think this will make it easier to adopt strict mode, as people can enable strict mode in production, but instead of throwing violations, they can log them.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -467,6 +467,11 @@ trait HasAttributes
         if ($this->exists &&
             ! $this->wasRecentlyCreated &&
             static::preventsAccessingMissingAttributes()) {
+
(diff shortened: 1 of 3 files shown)


deKuddus:
@aarondfrancis,  maybe this is going to be a good feature but did you see that a lots of changes are showing just due to changes of a single space. I think it was happened due to IDE configuration, so that this type of unnecessary changes are showing. Could you please recheck the changes again please??

aarondfrancis:
Fixed. I kinda thought that would happen automatically, which is why I left it. Thanks @deKuddus. 

And thanks to @timacdonald for this [incredible tip](https://twitter.com/timacdonald87/status/1564770267687432195) on how to fix this!

PR #44667, opened 2022-10-20
[9.x] Use latest PHPRedis extension

it-can:


ankurk91:
Can we use wild cards in version here?

PR #44487, opened 2022-10-06
[9.x] Makes blade components blazing fast ⛽️

nunomaduro:
## Technical Details 

This pull request addresses the performance of blade view components, and it mainly performs:

1. Components are only resolved from the IOC container if necessary. This is done by checking if the component `$data` matches exactly the `Component::__construct()` method parameters names, and if not, we proxy for the container. 

2. The computation of the `Component::render`, when using inline components, is only computed once. So, assuming that the return is `components.button`, we only check if `components.button` is a view once, and following calls will use this cached result.

3. We provide a new method, called `Component::view`, that should be documented instead of having people using the simple `view` helper.

4. Events on views, related to creators and composers, are only fired if there is actually creators or composers for the given view.

5. Checking if views are expired, only happens once for the given view, during the lifecycle of the Application instance. 

## How To Test This Pull Request

If you would like to test this pull request locally, simply run the command above.

```
composer require laravel/framework:"dev-feat/view-components-performance as 9.34" --with-all-dependencies
php artisan clear
php artisan view:clear
```

After, run your test suite, navigate in your application, and see if things are working as normal.

## Results
**Anonymous components — up to 52% faster:** 
```
@for ($number = 0; $number < $numberOfComponents; $number++)
    <x-button: :$number /> 
@endfor

Before:
- 100 components: 7.203ms
- 1000 components: 67.764ms
- 10000 components: 666.120ms
After:
- 100 components: 3.649ms
- 1000 components: 34.204ms
- 10000 components: 326.806ms
```

**Class View components - up to 39% faster:** 
```
@for ($number = 0; $number < $numberOfComponents; $number++)
    <x-button: :$number /> 
@endfor

class Button extends Component
{
    function __construct(public $number) {}

    function render() {
        return 'components.button';
    }
}

Before:
- 100 components: 6.946ms
- 1000 components: 69.657ms
- 10000 components: 670.511ms
After:
- 100 components: 4.441ms
- 1000 components: 42.266ms
- 10000 components: 413.125ms
```

**Class Inline components — up to 92% faster:** 
```
@for ($number = 0; $number < $numberOfComponents; $number++)
    <x-button: :$number /> 
@endfor

class Button extends Component
{
    function __construct(public $number) {}

    function render() {
        return '{{ $number }}';
    }
}

Before:
- 100 components: 66.426ms
- 1000 components: 560.284ms" // 0.5 seconds ..
- 10000 components: 5,080.315ms" // 5 seconds ..
After:
[...]

PR #31206, opened 2020-01-22
[7.x] Allow disabling checks for expired views

browner12:
This is a resubmission of #31195 with some performance results.

I did my best to isolate this change, so we could see the performance impact specifically from this change, so I kept the request as simple as possible.

Route:
```php
Route::get('/', 'MainController@index');
```

Controller:
```php
class MainController
{
    public function index(Request $request)
    {
        return view ('parent');
    }
}
```

Parent View:
```blade
@for($a=1; $a<=50; $a++)
    @include('child');
@endfor
```

Child View:
```blade
dummy content
```

The execution savings are entirely dependent on how many views need to be rendered for a given request. In the example above it is set at 51 (1 parent + 50 children).  I varied this number to see savings for different scenarios.

---

The following is the **inclusive time** of the `isExpired()` method:

| View Count    | Sample 1 | Sample 2 | Sample 3 | Average |
| ------------- | --------- | ---------- | --------- | -------- |
| 10 | 7.95 ms | 7.34 ms | 7.36 ms | 7.55 ms |
| 25  | 17 ms | 20 ms | 18.5 | 18.50 ms |
| 50  | 35.5 ms | 38.5 ms | 33.3 ms | 35.77 ms |
| 100 | 73.6 ms | 80.3 ms | 62.8 ms | 72.23 ms |

---

The following are the **total times** of the request:

| View Count    | Sample 1 | Sample 2 | Sample 3 | Average |
| ------------- | --------- | ---------- | --------- | -------- |
| 10 | 410 ms | 346 ms | 331 ms | 362 ms |
| 25  | 343 ms | 382 ms | 375 ms | 367 ms |
| 50  | 398 ms | 414 ms | 387 ms | 400 ms |
| 100 | 478 ms | 489 ms | 408 ms | 458 ms |

---

Based on these numbers our average savings per request by implementing this feature are:

| View Count | Inclusive Time | Total Time | Savings |
| ------------- | --------- | ---------- | --------- |
| 10 | 7.55 ms | 362 ms | 2.08% |
| 25 | 18.50 ms | 367 ms | 5.05% |
| 50 | 35.77 ms | 400 ms | 8.95% |
| 100 | 72.23 ms | 458 ms | 15.76% |

---

So we see, as expected, the more views that need to be rendered, the bigger percentage it takes up of your request.  Looking through some of our sites, it's not uncommon to see some pages with 40-50 views rendered.

---

[...]

PR #44674, opened 2022-10-21
[9.x] Allow disabling checks for expired views

browner12:
One of the changes @nunomaduro made in #44487 reminded me of a similar PR (#31206) I'd done about 2 years ago that was merged and then reverted due to a reported edge case bug.

The reported bug was that the `view:cache` was not correctly caching registered vendor views. I wish I'd have looked closer 2 years ago, because I've been testing all day, and everything seems to be working well to me, and I can't find any noticeable change in the framework that would've changed this behavior.

With that said, this is a resubmission of that original PR (#31206), with essentially no changes.  This change takes Nuno's optimization on compiled views a step further. The root of the problem is the `$this->compiler->isExpired($path)` method is a relatively expensive call, and it can get called a **large** number of times if:

1. you have a large number of views/components necessary for the page rendering
2. you render the same view/components multiple times on the page 

So the goal is to avoid unnecessary calls to `isExpired()`. Nuno's change in #44487 solves scenario 2 by building up a local cache of file paths already compiled, and then checking for the path value in the array on subsequent calls.  The changes in this PR actually solve both scenarios because you can now explicitly tell the `CompilerEngine` to never check if the path is expired and to assume the file is compiled validly.  This is intended to be a ***production only*** feature, and will only work if you run `php artisan view:cache` as part of your deployment.

So to start, could I first get a little confirmation from others that vendor registered views are being compiling with `artisan view:cache`. After that the PR should be good to review.

Check out the previous PR for the performance improvement data.

### DIFF
--- a/src/Illuminate/Support/ServiceProvider.php
+++ b/src/Illuminate/Support/ServiceProvider.php
@@ -172,7 +172,7 @@ abstract class ServiceProvider
                 }
             }
 
-            $view->addNamespace($namespace, $path);
+            $view->addNamespace($namespace, realpath($path) ?: $path);
         });
     }
 
--- a/src/Illuminate/View/Engines/CompilerEngine.php
+++ b/src/Illuminate/View/Engines/CompilerEngine.php
@@ -31,17 +31,26 @@ class CompilerEngine extends PhpEngine
     protected $compiledOrNotExpired = [];
 
     /**
+     * Flag to check expired views.
+     *
+     * @var bool
+     */
+    protected $checkExpiredViews = true;
+
+    /**
      * Create a new compiler engine instance.
      *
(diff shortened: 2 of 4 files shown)


taylorotwell:
How does this address the issues we had with the original PR?

milwad-dev (commented):
👍

browner12:
Temporarily converting to draft while I figure out the best course forward for the edge cases.

driesvints:
Hi there. We fixed the failing tests on 9.x which are causing this PR to fail as well. Can you please rebase your PR with 9.x? Thanks

driesvints:
@browner12 feel free to resend this once you've figured out a way forward here. Thanks.

PR #44676, opened 2022-10-21
[9.x] Add Str::in() helper

intrepidws:
This adds the `Str::in()` helper that will determine whether the given value is contained within the provided array of options. This expands upon the existing `Str::is()` helper and is consistent with other parts of the framework like the `whereIn()` Eloquent and DB methods.

While it's certainly possible to simply use `in_array()` in your code, I find myself reaching for this when using fluent strings where I am performing other operations first. Being able to chain `in()` on to the end of this transformation string would be helpful.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -351,6 +351,18 @@ class Str
     }
 
     /**
+     * Determine if a given string matches any of the given values.
+     *
+     * @param  string  $value
+     * @param  array  $sequence
+     * @return bool
+     */
+    public static function in($value, array $options = [])
+    {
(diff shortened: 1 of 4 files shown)


MohammadAlhallaq:
I'm a bit curious why you added it to the Str class and not the Arr class

intrepidws:
> I find myself reaching for this when using fluent strings where I am performing other operations first. Being able to chain in() on to the end of this transformation string would be helpful.

So something like this:

`Str::of($value)->trim()->in(['values', 'more'])`

MohammadAlhallaq:
hmm seems legit to me!

PR #44679, opened 2022-10-21
Adds `getExcept` method

mahmoudmohamedramadan:
This PR adds the ability to get the columns except we provide to the method, and I think this method is IMPORTANT something in some cases, and to illustrate that let's take a glance over these examples.

## BEFORE this PR gets merged

This code returns all columns except `created_at` and `updated_at`.

```PHP
use App\Models\User;

return User::get(['name', 'email', 'email_verified_at', 'status']);
```

The previous code is quite bad and gets worse if we have more columns. **I've faced the same problem before and this is what makes me Introduce this method**

## AFTER this PR gets merged

This code does the same thing that the previous code does but in an elegant way.

```PHP
use App\Models\User;

return User::getExcept(['created_at', 'updated_at']);
```

**I do not have a good annotation description for this method, so anyone who has a good one can leave a preview.**

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -689,6 +689,23 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Execute the query as a "select" statement.
+     *
+     * @param  array  $except
+     * @return \Illuminate\Database\Eloquent\Collection|static[]
+     */
+    public function getExcept(array $except)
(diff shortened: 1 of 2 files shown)


jmarcher:
It seems a little bit counterproductive to me to query the schema for a command in order to exclude a few columns. 

Yes, your code will get longer but at the same time it is clear which columns will be selected, and the other thing is that this method will perform worse than the normal `get`

PR #44686, opened 2022-10-21
[9.x] Testing: Allow users to determine the schema file used when migrating the database for tests

didac-adria:
If you are running tests, chances are that you are using a different database connection for your tests than your normal connection. In fact, the default phpunit.xml file provided by Laravel already switches to a "testing" connection. However, if you are [squashing migrations](https://laravel.com/docs/9.x/migrations#squashing-migrations) you will provably want to use the same schema file as in your normal database, rather than using a copied file with the name of the testing connection (for instance: testing-schema.dump).

This PR adds the option to easily choose your desired schema file when running tests which require running migrations. The way to specify the file, is exactly the same as the one used to indicate that you want to seed the database before each test, or to specify the database seeder: https://laravel.com/docs/9.x/database-testing#running-seeders. So, you just need to add a `protected $dump = 'database/schema/mysql-schema.dump';` property in your testCase class.

The PR also updates the CanConfigureMigrationCommandsTest, to ensure that functionality works well. In addition, this new functionality is being documented in a PR I'm creating after this in the [Laravel docs project](https://github.com/laravel/docs). I will be adding the PR link in the comments.

### DIFF
--- a/src/Illuminate/Foundation/Testing/Traits/CanConfigureMigrationCommands.php
+++ b/src/Illuminate/Foundation/Testing/Traits/CanConfigureMigrationCommands.php
@@ -2,6 +2,8 @@
 
 namespace Illuminate\Foundation\Testing\Traits;
 
+use Illuminate\Support\Collection;
+
 trait CanConfigureMigrationCommands
 {
     /**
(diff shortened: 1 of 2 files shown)


didac-adria:
Here is the related documentation PR: https://github.com/laravel/docs/pull/8301

taylorotwell:
Hey there - you can just overwrite this method in your class and return whatever options you want.

didac-adria:
Hey @taylorotwell!

Yes, that's right, but then, if you just want to add the schema path, you will need to override also the other functionality the migrateFreshUsing() method is giving. Or call the parent and extend it. This is actually what I have done on our project.

However, I do think Laravel is built in a way in which the code is self-documented, and the aim of that PR was to contribute in that direction. In fact, I consider the way Laravel finds the schema file (using the connection name) can be a little bit opaque. I had to spend some time to be able to know that, because it's not documented anywhere. I consider I can help others by documenting it + adding some syntactic sugar by allowing developers to add one single line to choose the schema file picked. In fact, this is exactly what Laravel currently does with the seeders in: `protected $seed = true;` and `protected $seeder = OrderStatusSeeder::class;`. Plus, if you see that in the documentation, you don't have to loose time by seeing how laravel works to be able to know that you have to override the migrateFreshUsing() method.

What do you think?

ziadoz:
I opened a related PR here: https://github.com/laravel/framework/pull/44630 

I ended up rolling my own trait to solve the problem: https://gist.github.com/ziadoz/db4277e16647981cac91e428a9c8e2a9

didac-adria:
> I opened a related PR here: #44630
> 
> I ended up rolling my own trait to solve the problem: https://gist.github.com/ziadoz/db4277e16647981cac91e428a9c8e2a9

That's a good point. In my case though, I don't want to change the connection. I just want to be able to use the "testing" connection (which in fact the default phpunit.xml file gives) in a project we are using database dumps. I think it can be a common case: you just need to do tests in a project where you do database dumps.

In my case, I also solved it by overriding the migrateFreshUsing() method, but I don't feel that's "the Laravel way". I thought I could contribute to the community with that PR.

PR #44690, opened 2022-10-22
Use `array_is_list` to determine if it is list

mahdiaslami:
In this PR `Arr::isAssoc()` and `Arr::isList()` refactored to use `array_is_list` function which introduced in php 8.0.


huangdijia:
https://www.php.net/manual/zh/function.array-is-list.php

array_is_list required 8.1

mahdyaslami:
@huangdijia
thanks

shaedrich:
Now that Laravel 12.x has
```json
        "php": "^8.2",
```
This could be done

PR #44693, opened 2022-10-22
[9.x] Refactor use syntax sugar in belongsTo method in HasRelationships trait

usuyuki:
With the greatest respect, I found an area where the belongsTo method in HasRelationships allows me to write shorter code.

$foreignKey of the default argument is null. So, I think this syntax run same behavior.
In addition, this syntax is also used by other methods such as hasOne, so it provides a sense of uniformity!

This is a result of null.
https://3v4l.org/lN6qY

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
@@ -206,9 +206,7 @@ trait HasRelationships
         // If no foreign key was supplied, we can use a backtrace to guess the proper
         // foreign key name by using the name of the relationship function, which
         // when combined with an "_id" should conventionally match the columns.
-        if (is_null($foreignKey)) {
-            $foreignKey = Str::snake($relation).'_'.$instance->getKeyName();
-        }
+        $foreignKey = $foreignKey ?: Str::snake($relation).'_'.$instance->getKeyName();
 
         // Once we have the foreign key names we'll just create a new Eloquent query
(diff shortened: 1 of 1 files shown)


src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
```diff
@@ -206,9 +206,7 @@ public function belongsTo($related, $foreignKey = null, $ownerKey = null, $relat
         // If no foreign key was supplied, we can use a backtrace to guess the proper
         // foreign key name by using the name of the relationship function, which
         // when combined with an "_id" should conventionally match the columns.
-        if (is_null($foreignKey)) {
-            $foreignKey = Str::snake($relation).'_'.$instance->getKeyName();
-        }
+        $foreignKey = $foreignKey ?: Str::snake($relation).'_'.$instance->getKeyName();
```
crissi, line 209: ?: = non falsey, so not the same as a null check

Usuyuki: Thank you for reviewing.

I know that ?: behaves differently than is_null.
https://www.php.net/manual/en/types.comparisons.php

But, in this method, $foreignKey only become string or null.
https://3v4l.org/8WYJ7

Even if an empty string is passed, it cannot be determined as is_null, but if it is ?:, it can be determined as false, and the foregin key can be guessed and assigned.

I am a beginner of the PHP.
Im sorry if there are mistakes.

driesvints:

PR #44698, opened 2022-10-23
[9.x] Add a generator for canonical URLs

tontonsb:
This adds a method for canonical links to the URLGenerator class.

```php
<link rel="canonical" href="{{url()->canonical()}}">
```

It will generate URLs to the current resource, but on the `app.url` root, which is the de facto canonical root already used in emails.

```php
Config::set('app.url', 'http://example.com');

// reqeust url: https://www.example.net/some/path?page=999
url()->canonical(); // returns 'http://example.com/some/path?page=999'
```

### Options

When the same page is available on multiple paths, one can manually specify the canonical one. In this case we will only canonicize the domain name.

```php
Config::set('app.url', 'http://example.com');

// reqeust url: https://www.example.net/some/path?page=999
url()->canonical('custom/path'); // returns 'http://example.com/custom/path'
```

The domain can also be specified in the code for some projects where different sections should have different canonical domains:

```php
// reqeust url: https://www.example.net/some/path?page=999
url()->canonical(domain: 'http://example.org');  // http://example.org/some/path?page=999
```

### Side effects and implementation details

This PR also adds an optional `root` argument to `URL::to()` as that felt the easiest way to accomplish this. This might also be useful for other uses.

I noticed that Laravel [is moving towards](https://github.com/laravel/framework/pull/44545) native PHP types, so I declared native types for the **new** things that I added. Please let me know if something about types and/or newer PHP features should be done differently on Laravel PRs.

### DIFF
--- a/src/Illuminate/Routing/RoutingServiceProvider.php
+++ b/src/Illuminate/Routing/RoutingServiceProvider.php
@@ -64,7 +64,8 @@ class RoutingServiceProvider extends ServiceProvider
             return new UrlGenerator(
                 $routes, $app->rebinding(
                     'request', $this->requestRebinder()
-                ), $app['config']['app.asset_url']
+                ), $app['config']['app.asset_url'],
+                $app['config']['app.url'],
             );
         });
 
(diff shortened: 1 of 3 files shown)


tontonsb:
Tests are failing now for seemingly unrelated reasons and other PRs apparently have the same problem.

timacdonald:
This should also support specifying which query parameters are meaningful in some way, e.g.

```
Request:   https://example.net/some/path?page=999&fbclid=123asd123asd

Canonical: https://example.net/some/path?page=999
```

driesvints:

taylorotwell:

PR #44703, opened 2022-10-24
[9.x] Add raw content property for mailables

Xammie:
This is a follow up from https://github.com/laravel/framework/pull/44696#issuecomment-1288210815. 

This PR introduces a new property and method on the mailable content class to supply a raw html string.

```php
public function content() {
    return new Content(
        raw: 'raw html string'
    );
}
```

The classic mailable approach already has this feature.
```php
public function build() {
    return $this->html('raw html string');
}
```

### DIFF
--- a/src/Illuminate/Mail/Mailables/Content.php
+++ b/src/Illuminate/Mail/Mailables/Content.php
@@ -39,7 +39,7 @@ class Content
     public $markdown;
 
     /**
-     * The raw html of the message
+     * The raw html of the message.
      *
      * @var string|null
      */
--- a/src/Illuminate/Mail/Mailables/Content.php
+++ b/src/Illuminate/Mail/Mailables/Content.php
@@ -39,11 +39,11 @@ class Content
     public $markdown;
 
     /**
-     * The raw html of the message.
+     * The pre-rendered HTML of the message.
      *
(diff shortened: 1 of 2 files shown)


src/Illuminate/Mail/Mailables/Content.php
```diff
@@ -52,16 +59,18 @@ class Content
      * @param  string|null  $html
      * @param  string|null  $text
      * @param  string|null  $markdown
+     * @param  string|null  $raw
      * @param  array  $with
      *
      * @named-arguments-supported
      */
-    public function __construct(string $view = null, string $html = null, string $text = null, $markdown = null, array $with = [])
+    public function __construct(string $view = null, string $html = null, string $text = null, string $markdown = null, string $raw = null, array $with = [])
```
driesvints, line 67: This is a major breaking change. You can't add a new argument here in the middle.

driesvints:

ascentcreative:
Are there plans for a rawText() method at all? 

I'm using soundasleep/html2Text to convert my HTML view to text-only, and it would be great to be able to pass that value into the Content. At the moment I can fudge it by passing the text via a blade, but it's a bit clunky...

PR #44715, opened 2022-10-24
Fixes incorrect resource name and parameters

mahmoudmohamedramadan:
## BEFORE this PR gets merged

I was working on a project and I wanted to add a new route to the `resource` routes like so,

```PHP
Route::resource('users', UserController::class);
Route::delete('/users/multi_delete', [UserController::class, 'multiDelete'])->name('users.multiDelete');
```

The result was as shown in the next figure

before_prefix

But I was would like to use the `prefix` route method to be more clear like so,

```PHP
Route::prefix('users')->group(function () {
    Route::resource('/', UserController::class);
    Route::delete('/multi_delete', [UserController::class, 'multiDelete'])->name('users.multiDelete');
});
```

The result was as shown in the next figure

before-fixing-issue

As we noticed, the resource routes with **NULL** segments and incorrect routes names.

## AFTER this PR gets merged

after-fixing-issue

The issue is solved now 🎉

> In fact, I tried to use `getGroupStack` method, but I realized that its a painful way because it returns an array with every `middleware`, `prefix`, `name`...etc in the route file and there is no distinctive key to catch the prefix that I need, so I decide to guess the routes segments and names with the given controller name.

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -332,9 +332,9 @@ class Router implements BindingRegistrar, RegistrarContract
             $controllerName = last(explode("\\", $controller));
             $name = explode("controller", strtolower($controllerName))[0];
 
-            return new PendingResourceRegistration(
-                $registrar, "{$name}s", $controller, $options
-            );
+            return (new PendingResourceRegistration(
+                $registrar, "/", $controller, $options
+            ))->names("{$name}s")->parameters([null => $name]);
         }
 
         return new PendingResourceRegistration(
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -327,6 +327,15 @@ class Router implements BindingRegistrar, RegistrarContract
         } else {
             $registrar = new ResourceRegistrar($this);
         }
+        
+       if($name == "/") {
+            $controllerName = last(explode("\\", $controller));
+            $name = explode("controller", strtolower($controllerName))[0];
+
+            return new PendingResourceRegistration(
+                $registrar, "{$name}s", $controller, $options
+            );
+        }
 
         return new PendingResourceRegistration(
             $registrar, $name, $controller, $options


PerryvanderMeer:
In my opinion, this resource-thing is much broader than just to the `prefix` function.

## Example
Beside the default resource actions, i need some pages for `delete`, `history` and `rights`.

### Current setup 
afbeelding

```php
Route::prefix('users')->controller(UserController::class)->name('users.')->group(function ()
{
    Route::get('{user}/rights', 'rights')->name('rights');
    Route::get('{user}/history', 'history')->name('history');
    Route::get('{user}/delete', 'delete')->name('delete');
});

Route::resource('users', UserController::class);
```

### Current problems
In the current situation, if i move the `resource` into the `group` , the `prefix` and `name` are duplicated.
The prefix-issue could be resolved by this PR.

afbeelding

```php
Route::prefix('users')->controller(UserController::class)->name('users.')->group(function ()
{
    Route::resource('users', UserController::class);

    Route::get('{user}/rights', 'rights')->name('rights');
    Route::get('{user}/history', 'history')->name('history');
    Route::get('{user}/delete', 'delete')->name('delete');
});
```

### Ideal situation
Ideally, i want to move the `resource` into the `group` and make the `$controller` parameter optional, since it's redundant.

```php
Route::prefix('users')->controller(UserController::class)->name('users.')->group(function ()
{
    Route::resource('/');

    Route::get('{user}/rights', 'rights')->name('rights');
    Route::get('{user}/history', 'history')->name('history');
    Route::get('{user}/delete', 'delete')->name('delete');
});
```

mahmoudmohamedramadan:
> In my opinion, this resource-thing is much broader than just to the `prefix` function.
> 
> ## Example
> Beside the default resource actions, i need some pages for `delete`, `history` and `rights`.
> 
> ### Current setup
> afbeelding
> 
> ```
> Route::prefix('users')->controller(UserController::class)->name('users.')->group(function ()
> {
>     Route::get('{user}/rights', 'rights')->name('rights');
>     Route::get('{user}/history', 'history')->name('history');
>     Route::get('{user}/delete', 'delete')->name('delete');
> });
> 
> Route::resource('users', UserController::class);
> ```
> 
> ### Current problems
> In the current situation, if i move the `resource` into the `group` , the `prefix` and `name` are duplicated. The prefix-issue could be resolved by this PR.
> 
> afbeelding
> 
> ```
> Route::prefix('users')->controller(UserController::class)->name('users.')->group(function ()
> {
>     Route::resource('users', UserController::class);
> 
>     Route::get('{user}/rights', 'rights')->name('rights');
>     Route::get('{user}/history', 'history')->name('history');
>     Route::get('{user}/delete', 'delete')->name('delete');
> });
> ```
> 
> ### Ideal situation
> Ideally, i want to move the `resource` into the `group` and make the `$controller` parameter optional, since it's redundant.
> 
> ```
> Route::prefix('users')->controller(UserController::class)->name('users.')->group(function ()
> {
>     Route::resource('/');
> 
>     Route::get('{user}/rights', 'rights')->name('rights');
>     Route::get('{user}/history', 'history')->name('history');
>     Route::get('{user}/delete', 'delete')->name('delete');
> });
> ```

You still had not solved the issue!

mahmoudmohamedramadan:
I've noticed duplicating path `/users` value in the URL as you can see in the next figure

before-fixing-issue


**The issue has been completely solved now as shown in the next figure 🎉**

after-fixing-issue

taylorotwell:
I don't think we should solve this by reading the controller segments and sticking an "s" after it.

PR #44716, opened 2022-10-24
early return model class name

medilies:
**Suggestion:** An explicit exception can be thrown if `$model` class doesn't exist!


medilies:
Hi @taylorotwell, I hope you doing well. Can get feedback about discarding my PR?

If it is related to forcing pushes. Then the first one was to add 2 spaces to fix styling issues. The second one was because the test failed due to some database operation duration that wasn't related to my commits.

PR #44717, opened 2022-10-24
Add hasAttribute method to Eloquent Models

geisi:
Hello Laravel Team,

With Laravel's new strict ``` Model::preventAccessingMissingAttributes``` Eloquent configuration, there are some problems with 3rd party packages when these packages want to access optional attributes.

To prevent ``` MissingAttributeException```, it would be handy to have a central way to check if an Eloquent Model attribute exists or not.

This PR introduces a new ```hasAttribute(string $key): bool``` method on Eloquent models to check if a model has an attribute configured. This also could be archived in the 3rd party package code. But I think it would be helpful if this gets added to the core. So it will be aligned with upcoming new framework features.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -268,7 +268,7 @@ trait HasAttributes
     /**
      * Check if model has a given attribute.
      *
-     * @param  string $key
+     * @param  string  $key
      * @return bool
      */
     public function hasAttribute($key)
@@ -465,7 +465,7 @@ trait HasAttributes
             return;
         }
 
-        if($this->hasAttribute($key)) {
+        if ($this->hasAttribute($key)) {
             return $this->getAttributeValue($key);
         }
 
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -266,6 +266,40 @@ trait HasAttributes
     }
 
     /**
+     * Check if model has a given attribute.
+     *
+     * @param  string $key
+     * @return bool
+     */
+    public function hasAttribute($key)
+    {
+        // If the attribute exists in the attribute array or has a "get" mutator we will
+        // return the attribute's value. Otherwise, we will proceed as if the developers
+        // are asking for a relationship's value. This covers both types of values.
+        if (array_key_exists($key, $this->attributes)) {
+            return true;
+        }
+
+        if (array_key_exists($key, $this->casts)) {
+            return true;
+        }
+
+        if ($this->hasGetMutator($key)) {
+            return true;
+        }
+
+        if ($this->hasAttributeMutator($key)) {
+            return true;
+        }
+
+        if ($this->isClassCastable($key)) {
+            return true;
+        }
+
+        return false;
+    }
+
+    /**
      * Add the casted attributes to the attributes array.
      *
      * @param  array  $attributes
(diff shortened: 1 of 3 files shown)


taylorotwell:
Can you do `isset($model->foo)`?

geisi:
Isset indeed works for Attributes, GetMutators, Relations, and AttributeMutators. Sadly isset does not work for Castables.

src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
```diff
@@ -265,6 +265,40 @@ protected function addMutatedAttributesToArray(array $attributes, array $mutated
...
+     * Check if model has a given attribute.
+     *
+     * @param  string $key
+     * @return bool
+     */
+    public function hasAttribute($key)
+    {
+        // If the attribute exists in the attribute array or has a "get" mutator we will
+        // return the attribute's value. Otherwise, we will proceed as if the developers
+        // are asking for a relationship's value. This covers both types of values.
```
m-englisch, line 278: I think that comment doesn't make any sense here and should be moved back to where it was.

donnysim:
Why would

```php
if (array_key_exists($key, $this->casts)) {
    return true;
}
```
mean the attribute exists? This will result in always returning true because a cast exists even though the attribute wasn't selected - doesn't exist.

m-englisch:
I guess if a casts exists, the model must has that attribute at some point. Even if that means that that the attribute is null or has no allocated value at all.

donnysim:
But isn't the whole point of hasAttribute to check if it exists now not if it will exist at some point? Because what does it actually solve then? It will say true and the next thing you get will be a missing attribute exception because it lied?

geisi:
If there is a cast no missingAttributException will be thrown. The ```hasAttribute()``` method abstracts the checks which are done previously done in the ```getAttribute()``` method. So if hasAttribute() is true you can be sure that no missingAttributeException will be thrown.

donnysim:
Oh, so the whole attribute strict mode is even more crooked than I imagined 😂. It shouldn't matter if a cast is defined or not, if the attribute does not exist for casting it should throw an error. Isn't it the whole point of it?

```php
class TestModel extends Model
{
    protected $casts = [
        'parent_id' => 'int',
    ];
}

TetsModel::first(['id'])->parent_id // <- should throw not silently return null
```

Sadly the more I follow the whole strict mode changes the more I'm inclined to just never use the built in way because it's a minefield in itself, so I'll just leave you do your things and carry on.

geisi:
This PR is not a general discussion regarding Laravel's strict mode. 
But in my opinion, it's the right way how it is currently implemented. The only reason when a ModelAttributeMissing Exception should be thrown is when the attribute is not defined anywhere (Database, Casts, GetMutators, Relations ...). And this is what the current implementation does.

src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
geisi: Fixed

inxilpro:
Personally, I think that `hasAttribute` would have a place if #44642 hadn't been merged. My preference would be that we revert #44642 as it complicates the semantics of `preventAccessingMissingAttributes` and add a `hasAttribute` method that essentially behaves like the current `offsetExists` implementation.

Either way, I don't think `hasAttribute()` should return `true` just because a cast exists. And if we don't revert #44642, I think the current `offsetExists` behavior covers this use-case.

taylorotwell:
I honestly am not sure I plan to mess with this anymore. I'm not super happy with how "strict mode" has panned out and it has felt a bit more trouble than it's worth.

royduin:
It's added in Laravel 11.3: https://github.com/laravel/framework/pull/50909

PR #44725, opened 2022-10-25
Add insertGetIds method to Query Builder

NerijusBartosevicius:
This PR proposes the addition of a insertGetIds() method to  query builder get values from inserted record result. By default returning id column, you can select any column you want, in certain cases you can get all the columns data by specifying '*'.


NerijusBartosevicius:
Got it, I created new package for this request https://packagist.org/packages/neriba/laravel-insert-update-delete-ids

PR #44730, opened 2022-10-25
[9.x] add dictionary to slug helper

imdhemy:
The current implementation of the `Str::slug()` method already replaces the `@` with `at`. The keyboard still has some other replaceable characters.

E.g `500$ bill` becomes `500-dollar-bill` instead of `500-bill`.

Besides, some languages have different rules for the replacement of special characters.
With the current implementation, we can have something like:

```
أحمد-at-المدرسة
```

which should be:

```
أحمد-في-المدرسة
```

This PR adds a new parameter to the `Str::slug()` method to allow the developer to specify the replacement character for the special characters.

It does not break the current implementation, as the default value for the new parameter is `['@' => 'at']`.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -1023,15 +1023,14 @@ class Str
 
         $title = preg_replace('!['.preg_quote($flip).']+!u', $separator, $title);
 
-        // add separator before and after dictionary words
+        // Replace dictionary words
         foreach ($dictionary as $key => $value) {
             $dictionary[$key] = $separator.$value.$separator;
         }
 
-        // replace dictionary words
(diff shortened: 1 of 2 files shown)


milwad-dev (commented):
is useful

taylorotwell:
Suggest just writing your own slug utility function if you need customization here.

driesvints:
@milwad-dev please only comment when you have something to contribute to the discussion.

shaedrich:
Wouldn't it be better to have a more diverse method signature like this:

```php
Str::slug('500$ bill', [ '$' => fn($value) => /* do something */]);
Str::slug('500$ bill', fn($value) => /* do something */);
Str::slug('500$ bill', [ '/$|€/' => fn($value) => /* do something */]);
```

RegExp would even work quite well, since it is already used in that method.

PR #44736, opened 2022-10-25
[9.x] Add Illuminate\Session\Store::setHandler()

### LINKED ISSUE #547: Run tenant aware query from central/tenants context
#### Describe the bug
- When we use $tenant->run() to perform query on another tenants from central/tenants context it should work as expected.
for example : 
`$tenant->run(function () {
    User::create(...);
});`

Above code works fine when we have session driver set to `SESSION_DRIVER=files`.
But if we set session driver to `SESSION_DRIVER=database` . it fails and give `Call to a member function prepare() on null.`

#### Steps to reproduce
- Make env file to use `SESSION_DRIVER=database`
- Run tenants specific query.

#### Expected behavior
- It should run the tenant specific query and return to central or whatever tenants we calling from.

#### Your setup
 - Laravel version: 6.2.x
 - stancl/tenancy version: 3.4.0
 - Storage driver: DB
- Session Driver : DB

stancl:
This adds a simple setter for the handler used by `Illuminate\Session\Store`. The PR doesn't introduce any breaking changes.

The use case is that when you're changing database connections on the fly, the database session driver can run into issues where it tries to run queries on a DB connection that no longer exists.

The session-related middleware instantiates `Illuminate\Session\Store` for the database driver which uses `DatabaseSessionHandler` as the `$handler`. That class stores the connection and is used at the end of the request to write the session data. At that point, the connection that was injected initially may not exist anymore.

By adding the setter, I can reconstruct the `DatabaseSessionHandler` on the fly and make it use the correct DB connection: https://gist.github.com/stancl/eb6e210b553bf697c13f19ddaf80f539#file-sessiontenancybootstrapper-php-L18-L38

The addition will let me fix https://github.com/archtechx/tenancy/issues/547.

### DIFF
--- a/src/Illuminate/Session/Store.php
+++ b/src/Illuminate/Session/Store.php
@@ -733,6 +733,17 @@ class Store implements Session
     }
 
     /**
+     * Set the underlying session handler implementation.
+     *
+     * @param  \SessionHandlerInterface  $handler
+     * @return void
+     */
+    public function setHandler(SessionHandlerInterface $handler)
+    {
+        return $this->handler = $handler;
+    }
+
+    /**
      * Determine if the session handler needs a request.
      *
      * @return bool


stancl:
Thanks :)

PR #44739, opened 2022-10-26
Remove comma from mail template

askdkc:
Some languages, like Japanese, don't put comma at the end of sentence. For example:

image

But Laravel mail template including comma outside of @lang section like below.

`@lang('Regards'),<br>`

This should be

`@lang('Regards,')<br>`
or
`@lang('Regards')<br>`

With this fix, you can translate Mail notification with LANG.json file without having this end-comma problem. Further more, it helps Devs to create Laravel app to support multi-languages easily. (Just make as many lang.json files as you want and swap them in config/app.php 'locale' => 'yourLang' )

### DIFF
--- a/src/Illuminate/Notifications/resources/views/email.blade.php
+++ b/src/Illuminate/Notifications/resources/views/email.blade.php
@@ -39,7 +39,7 @@
 @if (! empty($salutation))
 {{ $salutation }}
 @else
-@lang('Regards'),<br>
+@lang('Regards')<br>
 {{ config('app.name') }}
 @endif
 
--- a/src/Illuminate/Notifications/resources/views/email.blade.php
+++ b/src/Illuminate/Notifications/resources/views/email.blade.php
@@ -39,7 +39,7 @@
 @if (! empty($salutation))
 {{ $salutation }}
 @else
-@lang('Regards')<br>
+@lang('Regards,')<br>
 {{ config('app.name') }}
 @endif
 


driesvints:
This is unfortunately a breaking change because the string changed here and thus existing app language files won't be able to match it anymore.

askdkc:
@driesvints 
Hmm, how about just removing ','?
That way, translating "Regards" to other languages still works.

dennisprudlo:
This would remove the comma from all apps that are counting on that comma. The change would be better for 10.x I guess.

driesvints:
I don't think we'll take any action on this right now, sorry. You can still publish the template and modify it to your liking: https://laravel.com/docs/9.x/notifications#customizing-the-templates

askdkc:
@dennisprudlo 
I see, but not all languages have comma. Like, Japanese for example, it looks just so weird to have a comma at the end of the sentence. People might think it as some kind of SPAM or phishing email.

If it was included in `@lang('Regards,')` not outside of `@lang()`, customization was so easy and useful.

dennisprudlo:
I get that. However we can't simply take away the comma from all other languages that require one there. As Dries said the templates can be changed.

askdkc:
Yeah, like you said "The change would be better for 10.x". I get that too.

Laravel makes translating app so easy! Fantastic!! That makes me wonder why oh why this comma thing is just HERE...

PR #44742, opened 2022-10-26
Works on fixing incorrect resource name and parameters 🕸

mahmoudmohamedramadan:
**This PR continues the working of my old closed [PR](https://github.com/laravel/framework/pull/44715)**

As I said before, I did not use `ControllerName` to guess the resource `name` and `parameters`, also @taylorotwell did not prefer that way, so I decided to use `getGroupStack` method to get the `prefix` value, and then use it in the resource `name` and `parameters`

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -334,9 +334,13 @@ class Router implements BindingRegistrar, RegistrarContract
                 ->each(function ($groupStack) use ($registrar, $controller, $options) {
                     $name = explode("/", $groupStack["prefix"])[1];
 
-                    return (new PendingResourceRegistration($registrar, "/", $controller, $options))
-                        ->names($name)
-                        ->parameters([null => $name]);
+                    if(str_ends_with($name, "s")) {
+                        $name = substr($name, 0, -1);
+                    }
+
+                    return (new PendingResourceRegistration(
+                        $registrar, "/", $controller, $options
+                    ))->names($name)->parameters([null => $name]);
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not going to mess with this behavior. Just register a resource route at the top level.

mahmoudmohamedramadan:
@taylorotwell  Hmm - I have a certainty that is an issue and I do not know why you sir do not need to accept this PR.

dennisprudlo:
@mahmoudmohamedramadan There are different reasons why this fix shouldn't be merged as is:

1. The condition checks for `$name == "/"`, however an empty string is also a valid endpoint for registering a route.
```php
Route::resource('', ...);
```
2. Registering such a route will return `null` instead of a `PendingResourceRegistration` as per the docblock and will prevent from further chaining methods such as `only` or `except`.
```php
Route::resource('users', ...)->only('index'); // works
Route::resource('/', ...)->only('index'); // PHP Error:  Call to a member function only() on null
```
3. You use `str_ends_with($name, "s")` to check if the route ends with an `s`. If so it gets cut off for the route name. This may work for `users` => `user` but not for example `anamneses` => `anamnesis` or `buses` => `bus`
```php
Route::resource('users', ...); // -> "user"
Route::resource('anamneses', ...); // -> "anamnese"
Route::resource('buses', ...); // -> "buse"
```

mahmoudmohamedramadan:
@dennisprudlo Thanks for these notes. I'll try to fix these issues

PR #44743, opened 2022-10-26
[9.x] Make `migrate` command isolated

olivernybroe:
This PR makes the `migrate` command isolated which limits the migration command to only have one process active at once. Meaning two servers cannot both run `migrate` at the same time.

## Problem
When deploying on multiple servers at once, it needs to be decided which of these servers have the responsibility for running migrations, however this means that all the deployments to all the servers are not identical, to simplify deployments.

For example when using k8s, you would use something like `initContainers` for running things like migrations. However as this will happen on each replica it could result in two servers running the same migration at the same time, as there is no locking so only one server will run one migration. 

Other people talking about this issue
- https://github.com/laravel/framework/discussions/41559
- https://github.com/laravel/ideas/issues/1645
- https://github.com/laravel/ideas/issues/1970

## Solution
Make migrate command isolated. This could in theory still result in the command running twice in the same deployment, however not at the same time, so it won't cause any issues, it will however result in queries to the database to check if migrations are missing where we already know there are none missing.

This solution works by adding the `Isolated` interface to a command which marks the command as an isolated command and it will now only run on one server. This implementation would allow to easily make other commands run in isolated when needed.

```php
class MyCommand extends Command implements Isolated
{
   ...
}
```

An alternative solution could be to instead make this a flag called `isolated` so the user can choose when running the command to run it isolated or not. This solution could be added as a parameter available for all commands by default.

### DIFF
--- a/src/Illuminate/Console/CacheCommandMutex.php
+++ b/src/Illuminate/Console/CacheCommandMutex.php
@@ -8,18 +8,18 @@ use Illuminate\Contracts\Cache\Factory as Cache;
 class CacheCommandMutex implements CommandMutex
 {
     /**
-     * The cache store that should be used.
-     *
-     * @var string|null
-     */
-    public $store = null;
-
-    /**
      * The cache factory implementation.
      *
      * @var \Illuminate\Contracts\Cache\Factory
      */
     public $cache;
+
+    /**
+     * The cache store that should be used.
+     *
+     * @var string|null
+     */
+    public $store = null;
 
     /**
      * Create a new command mutex.
--- a/src/Illuminate/Console/CacheCommandMutex.php
+++ b/src/Illuminate/Console/CacheCommandMutex.php
@@ -24,9 +24,9 @@ class CacheCommandMutex implements CommandMutex
     public $cache;
 
     /**
-     * Create a new command mutex
+     * Create a new command mutex.
      *
-     * @param \Illuminate\Contracts\Cache\Factory $cache
+     * @param  \Illuminate\Contracts\Cache\Factory  $cache
      */
     public function __construct(Cache $cache)
     {
@@ -50,7 +50,7 @@ class CacheCommandMutex implements CommandMutex
     }
 
     /**
-     * @param Command $command
+     * @param  Command  $command
      * @return string
      */
     protected function commandMutexName($command): string
@@ -61,7 +61,7 @@ class CacheCommandMutex implements CommandMutex
     /**
      * Specify the cache store that should be used.
      *
-     * @param string|null $store
+     * @param  string|null  $store
      * @return $this
      */
     public function useStore($store): static
--- a/src/Illuminate/Console/CacheCommandMutex.php
+++ b/src/Illuminate/Console/CacheCommandMutex.php
@@ -1,7 +1,5 @@
 <?php
 
-declare(strict_types=1);
-
 namespace Illuminate\Console;
 
 use Carbon\CarbonInterval;
@@ -33,6 +31,12 @@ class CacheCommandMutex implements CommandMutex
         $this->cache = $cache;
     }
 
+    /**
+     * Attempt to obtain a command mutex for the given command.
+     *
+     * @param  \Illuminate\Console\Command  $command
+     * @return bool
+     */
     public function create($command)
     {
         return $this->cache->store($this->store)->add(
@@ -42,13 +46,12 @@ class CacheCommandMutex implements CommandMutex
         );
     }
 
-    public function exists($command)
-    {
-        return $this->cache->store($this->store)->has(
-            $this->commandMutexName($command)
-        );
-    }
-
+    /**
+     * Release the mutex for the given command.
+     *
+     * @param  \Illuminate\Console\Command  $command
+     * @return bool
+     */
     public function release($command)
     {
         return $this->cache->store($this->store)->forget(
@@ -57,10 +60,23 @@ class CacheCommandMutex implements CommandMutex
     }
 
     /**
-     * @param  Command  $command
+     * Determine if a command mutex exists for the given command.
+     *
+     * @param  \Illuminate\Console\Command  $command
+     * @return bool
+     */
+    public function exists($command)
+    {
+        return $this->cache->store($this->store)->has(
+            $this->commandMutexName($command)
+        );
+    }
+
+    /**
+     * @param  \Illuminate\Console\Command  $command
      * @return string
      */
-    protected function commandMutexName($command): string
+    protected function commandMutexName($command)
     {
         return 'framework'.DIRECTORY_SEPARATOR.'command-'.$command->getName();
     }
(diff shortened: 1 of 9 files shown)


src/Illuminate/Console/CacheCommandMutex.php
```diff
@@ -0,0 +1,47 @@
...
+        return $this->cache->store($this->store)->has(
+            $this->commandMutexName($command)
+        );
+    }
+
+    protected function commandMutexName(Command $command): string
+    {
+        return 'framework'.DIRECTORY_SEPARATOR.'command-'.$command->getName();
+    }
+    public function useStore(string|null $store): static
```
driesvints, line 41: Can you remove all typehints and add DocBlocks like in the rest of the framework please? Thanks!

src/Illuminate/Console/CommandMutex.php
```diff
@@ -0,0 +1,17 @@
...
+use DateTimeInterface;
+
+interface CommandMutex
+{
+    /**
+     * Attempt to obtain a command mutex for the given command
+     */
+    public function create(Command $command): bool;
+
+    public function exists(Command $command): bool;
```
driesvints, line 16: Please add proper DocBlocks like in the rest of the framework.

driesvints (changes requested):

src/Illuminate/Console/CacheCommandMutex.php
olivernybroe: whoops, sorry, force of habit :)

src/Illuminate/Console/CacheCommandMutex.php
```diff
@@ -0,0 +1,80 @@
+<?php
+
+declare(strict_types=1);
```
driesvints, line 3: Remove this again please.

src/Illuminate/Console/CacheCommandMutex.php
```diff
@@ -0,0 +1,80 @@
...
+     * The cache factory implementation.
+     *
+     * @var \Illuminate\Contracts\Cache\Factory
+     */
+    public $cache;
+
+    /**
+     * Create a new command mutex.
+     *
+     * @param  \Illuminate\Contracts\Cache\Factory  $cache
```
driesvints, line 29:
```suggestion
     * @param  \Illuminate\Contracts\Cache\Factory  $cache
     * @return void
```

src/Illuminate/Console/CacheCommandMutex.php
```diff
@@ -0,0 +1,80 @@
...
+     */
+    public function __construct(Cache $cache)
+    {
+        $this->cache = $cache;
+    }
+
+    public function create($command)
```
driesvints, line 36: All these methods in this class still need DocBlocks.

src/Illuminate/Console/CacheCommandMutex.php
```diff
@@ -0,0 +1,80 @@
...
+
+    public function release($command)
+    {
+        return $this->cache->store($this->store)->forget(
+            $this->commandMutexName($command)
+        );
+    }
+
+    /**
+     * @param  Command  $command
```
driesvints, line 60: We use FQN everywhere in DocBlocks.
```suggestion
     * @param  \Illuminate\Console\Command  $command
```

src/Illuminate/Console/CacheCommandMutex.php
```diff
@@ -0,0 +1,80 @@
...
+        return 'framework'.DIRECTORY_SEPARATOR.'command-'.$command->getName();
+    }
+
+    /**
+     * Specify the cache store that should be used.
+     *
+     * @param  string|null  $store
+     * @return $this
+     */
+    public function useStore($store): static
```
driesvints, line 74:
```suggestion
    public function useStore($store)
```

src/Illuminate/Console/CacheCommandMutex.php
```diff
@@ -0,0 +1,80 @@
...
+        return $this->cache->store($this->store)->forget(
+            $this->commandMutexName($command)
+        );
+    }
+
+    /**
+     * @param  Command  $command
+     * @return string
+     */
+    protected function commandMutexName($command): string
```
driesvints, line 63:
```suggestion
    protected function commandMutexName($command)
```

src/Illuminate/Console/CommandMutex.php
```diff
@@ -0,0 +1,32 @@
...
+declare(strict_types=1);
+
+namespace Illuminate\Console;
+
+interface CommandMutex
+{
+    /**
+     * Attempt to obtain a command mutex for the given command.
+     *
+     * @param  Command  $command
```
driesvints, line 12: Use FQN everywhere.

src/Illuminate/Console/CommandMutex.php
```diff
@@ -0,0 +1,32 @@
+<?php
+
+declare(strict_types=1);
```
driesvints, line 3: Remove this everywhere.

tests/Console/CacheCommandMutexTest.php
```diff
@@ -0,0 +1,80 @@
...
+use Illuminate\Console\Command;
+use Illuminate\Contracts\Cache\Factory;
+use Illuminate\Contracts\Cache\Repository;
+use Mockery as m;
+use PHPUnit\Framework\TestCase;
+
+class CacheCommandMutexTest extends TestCase
+{
+    /**
+     * @var CacheCommandMutex
```
driesvints, line 17: Please use FQN everywhere.

driesvints (changes requested):

deleugpn:
Interesting overall, but when you look at a generic `Isolated` interface without knowing the underlying `Command`, I'm not sure how I feel about exit successfully. I keep imagining a process executing `php artisan something` and getting a successful response back "because I was not able to acquire a lock". In HTTP world, we have the 3xx but in CLI we unfortunately only have 0 and non-zero. Maybe I'm chasing ghosts, but I see this end up getting evolved into "let's ask the underlying command what I should do if I can't acquire the lock" and then it just feels easier if you just wrap the migrator yourself?

olivernybroe:
@deleugpn Would you feel better about it if it was a flag? 

So you say `php artisan migrate --isolated` and then it is enabled. This way you actively choose that the command can finish early because another server is running it already.

olivernybroe:
Alright, I have added a flag named `isolated` to it instead. 👍

taylorotwell:
Added ability to customize the lock expiration time. Also added ability to specify the exit code (`--isolated=12`) per @deleugpn's comment.

taylorotwell:
Interface renamed to `Isolatable`.

bytestream:
Why does this use `store()->add()` instead of `store()->lock()`? 

`store->add()` is affected by `cache:clear` while `->lock()` is not.

PR #44749, opened 2022-10-26
[9.x] Add prompt to optimize command in non-production environments

martinbean:
Various places in the Laravel docs say that the caching of routes etc should only be done as part of a deployment process. Yet various developers get tripped up after inadvertently running any of the `cache:*` or `optimize` commands and then wonder why things aren‘t working when they’re changing source files.

This PR adds a prompt when trying to run `php artisan optimize` in **non-production** environments. This means that any one running `php artisan optimize` during a deployment to their production environments shouldn’t be impacted (so long as they have set their `APP_ENV` value to `production`); but any one trying to run it in other environments (say `local`) will be prompted to actually confirm they want to run the command.

### DIFF
--- a/src/Illuminate/Foundation/Console/OptimizeCommand.php
+++ b/src/Illuminate/Foundation/Console/OptimizeCommand.php
@@ -3,11 +3,14 @@
 namespace Illuminate\Foundation\Console;
 
 use Illuminate\Console\Command;
+use Illuminate\Console\ConfirmableTrait;
 use Symfony\Component\Console\Attribute\AsCommand;
 
 #[AsCommand(name: 'optimize')]
 class OptimizeCommand extends Command
 {
+    use ConfirmableTrait;
+
     /**
      * The console command name.
      *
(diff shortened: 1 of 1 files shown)


taylorotwell:
Couldn't this break applications deploying to staging environments where they actually do want `optimize` to run?

IMO this prompt should only display if the environment is *actually* `local`.

martinbean:
> IMO this prompt should only display if the environment is _actually_ `local`.

Can certainly change the prompt callback to that case if you prefer as yeah, the prompt would be displayed if the environment was set to something like `staging`, `qa`, etc.

taylorotwell:
Honestly I think I would just maybe add a comment to the output if the environment is not production - no prompt.

MircoBabin:
I rebuild the cache a lot from scripting in my local environment. I know I have to do that when I change routes and when I introduce a new class. This additional confirmation message would only be annoying to me. At least it should be suppressable by a commandline switch.

So I agree with @taylorotwell that a message would be better.

taylorotwell:
Yeah - I think just a comment would be sufficient.

martinbean:
@taylorotwell Can you re-open this and I’ll change it to just an informational message instead?

N30:
haha , I ended up here from here: https://didyoucacheyourconfig.com/ this is funny

PR #44751, opened 2022-10-26
Eloquent forceDelete shouldn't ignore global scopes, it should ignore onDelete

jwadhams:
The root issue is that if you prepare an Eloquent query that uses global scopes, the results you get with `count` or `get` are not the same as the things that will be deleted by `forceDelete`.  

We had a production incident where we carefully prepared a query in Tinker, `get` and `count` were spot on, but when we ran `forceDelete` it affected 13000 more rows, destroying data for hundreds of accounts who have paused service.

Comments in `\Illuminate\Database\Eloquent\Builder::forceDelete` suggest it needs to ignore global scopes to work with soft deletes, but that's not supported by the unit tests. As long as it avoids the `onDelete` behavior that turns DELETE sql into UPDATE sql, then all existing Eloquent unit tests pass unmodified.

(The 2014 comment on `Builder::forceDelete` seemed to assume the only global scope is Soft Deletes, but we encountered this bug in the way `nanigans/single-table-inheritance` uses global scopes, and the unit test is based on the [simple documentation case for all global scopes](https://laravel.com/docs/9.x/eloquent#applying-global-scopes).)

The weird side effect of this change is that MassPrunable was documented and unit tested to affect soft deleted records, _even if soft delete scopes weren't specified in the user's pruneable query_.  My approach in this PR was to preserve that behavior, but I _personally_ think the change introduced on lines 25-28 should be removed in a future version of Laravel (at a time when semver says we can introduce breaking behavior changes) and docs should be updated that `pruneable` needs to manage `withTrashed` or `onlyTrashed` at the author's discretion.

### DIFF
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 down=false
-php="8.0"
+php="8.1"
 
 while true; do
   case "$1" in
--- a/bin/test.sh
+++ b/bin/test.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 down=false
-php="8.1"
+php="8.0"
 
 while true; do
   case "$1" in
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1182,13 +1182,14 @@ class Builder implements BuilderContract
     /**
      * Run the default delete function on the builder.
      *
-     * Since we do not apply scopes here, the row will actually be deleted.
+     * Unlike `delete` this does not use onDelete behavior, so traits that `extend` Builder
+     * (like SoftDeletingScope) can't intercept the DELETE and turn it into an UPDATE.
      *
      * @return mixed
      */
     public function forceDelete()
     {
-        return $this->query->delete();
+        return $this->toBase()->delete();
     }
 
     /**
--- a/src/Illuminate/Database/Eloquent/MassPrunable.php
+++ b/src/Illuminate/Database/Eloquent/MassPrunable.php
@@ -15,16 +15,21 @@ trait MassPrunable
      */
     public function pruneAll(int $chunkSize = 1000)
     {
+        $usesSoftDeletes = in_array(SoftDeletes::class, class_uses_recursive(get_class($this)));
         $query = tap($this->prunable(), function ($query) use ($chunkSize) {
             $query->when(! $query->getQuery()->limit, function ($query) use ($chunkSize) {
                 $query->limit($chunkSize);
             });
         });
 
+        if ($usesSoftDeletes) {
+            $query->withTrashed();
+        }
+
         $total = 0;
 
         do {
-            $total += $count = in_array(SoftDeletes::class, class_uses_recursive(get_class($this)))
+            $total += $count = $usesSoftDeletes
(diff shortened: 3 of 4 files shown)


bin/test.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 down=false
-php="8.0"
```
jwadhams, line 4: Sorry, this is my first PR to this project, but I couldn't get tests to run at all without this change, `bin/test.sh` would fail immediately with 
```
Your Composer dependencies require a PHP version ">= 8.1.0". You are running 8.0.24.
```

driesvints: This needs to be reverted. Laravel 9 still works on PHP 8.0 and needs to be tested as such with the lowest version.

taylorotwell:
Just curious - what custom global scopes did you have defined on the model. What did they do?

jwadhams:
We use the heck out of `nanigans/single-table-inheritance`. So we have a model `AccountIntegrations` that has dozens of sub-classes. The subclasses are all B2B partnerships we maintain, for example `class AccuTrade extends AccountIntegrations` and `class TradePending extends AccountIntegrations`  --which under the covers uses global scopes.  This is a really great fit for us, the relational table needs are very consistent across different AccountIntegrations, so there's no compelling reason to have dozens of tables and dozens of relationships.

Accounts are businesses that pay us for access to those AccountIntegrations. One Account will have a relationship to ~6 AccountIntegrations that represents both that Account uses that partnership, and account-specific settings. It's extremely common for an Account to pause then restore services, so we use soft deletes on Accounts and AccountIntegrations.

So I was doing final cleanup on an ending B2B partnership.  The command I ran was:
```
TradePending::onlyTrashed()->forceDelete();
```
Which I **wanted** to run SQL like
```
DELETE from account_integrations WHERE type='TradePending' AND deleted_at IS NOT NULL;
```
but because of the no-global-scopes thing the SQL it **actually** ran was more like
```
DELETE from account_integrations WHERE deleted_at IS NOT NULL;
```

This removed **all** the AccountIntegrations from all our Accounts who have paused services. So we had to scramble to restore that from backups before any of those lapsed Accounts tried to restore service.

bin/test.sh
jwadhams: Thanks, have done.

taylorotwell:
I feel like this may be something we need to address on 10.x if we want to take action here. There could be apps out there depending on this behavior not using global scopes and any change here would break those applications.

jwadhams:
Yeah, that's a tough thing to balance, and obvs I trust you to find that balance for the wider ecosystem.

I guess the consideration I'd weigh is: before this fix, it's possible to accidentally *delete* a lot more information than you intend to, and with the fix it's possible to *keep* more information than you intend to--but with the fix, the rows returned by `count()` and `forceDelete()` will match so you can diagnose that for yourself.

taylorotwell:
I just don't think I can make this change on 9.x. We can consider it for 10.x and update the docs for 9.x if necessary. Not a huge fan of global scopes in general 😬

jwadhams:
@taylorotwell Should I resubmit this in some way to be considered for 10.x? 
Personally I love _using_ global scopes, although I confess this was my first dive anywhere near internals. Is there anything I could do to increase overall confidence in them? More test coverage, for instance?

PR #44759, opened 2022-10-27
[9.x] Adds `response()->ok()`

nunomaduro:
I've always found confusing the away of returning an **typed `Illuminate\Http\Response` instance** with empty 200 status code in Laravel. While I understand that `200` is the default status code, and there is a couple of ways of making this possible, none of the existing ways looks clear to me.

```php
return response('');
return response('', 200);
return response(status: 200); // This one is actually OKish...
```

This pull request introduces an `Response::ok()` method, that can be used in this very simple way:

```php
public function something(Request $request): Response
{
    // $request->user()

    return response()->ok();
}
```

Of course, this combines well with the `assertOk()` response test method, as the route can be tested as such:

```php
it('works', function () {
    $this->get('/something')->assertOk();
});
```

### DIFF
--- a/src/Illuminate/Routing/ResponseFactory.php
+++ b/src/Illuminate/Routing/ResponseFactory.php
@@ -117,6 +117,18 @@ class ResponseFactory implements FactoryContract
     }
 
     /**
+     * Create a new "ok" response.
+     *
+     * @param  mixed  $content
+     * @param  array  $headers
+     * @return \Illuminate\Http\Response
+     */
+    public function ok(mixed $content = '', array $headers = [])
+    {
+        return $this->make($content, 200, $headers);
+    }
+
+    /**
      * Create a new streamed response instance.
      *
      * @param  \Closure  $callback
--- a/tests/Integration/Http/ResponseTest.php
+++ b/tests/Integration/Http/ResponseTest.php
@@ -9,6 +9,42 @@ use Orchestra\Testbench\TestCase;
 
 class ResponseTest extends TestCase
 {
+    public function testResponseOk()
+    {
+        Route::get('/response', function () {
+            return response()->ok();
+        });
+
+        $this->get('/response')
+            ->assertOk()
+            ->assertContent('');
+    }
+
+    public function testResponseOkWithContent()
+    {
+        Route::get('/response', function () {
+            return response()->ok('Hello World');
+        });
+
+        $this->get('/response')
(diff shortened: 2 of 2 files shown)


peterfox:
When would you normally want a 200 OK and an empty response body? Most people building APIs would return a [204 No Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/204) instead?

ahinkle:
Suitable for the idea of it being easier to read, but in your example, `response()->noContent();` would be the more acceptable response; why should we return a 200?

I think "nudging" developers would be best to return a valid response.  I also don't see the benefit or reason for using this in a natural response, but I would love to hear your thoughts.

```php
// Sure, but shouldn't this be a 200?
return response()->ok();
```

```php
// 200, but the "200" can be omitted.
return response()->json([
    ...
], 200);
```

```php
// returns 200.
return response()->json([
    ...
]);
```

```php
//  A bit difficult to read...
return response()
    ->ok()
    ->json([
        ...
    ]);
```

ju5t:
@driesvints not sure if your question was directed at me. 

I think the best way to answer it is as follows:

- Should the browser or API parse the response? It's a 200 OK.
- Is the content not important for the browser or the API? It's a 204 No Content.

For 'request successful', I would replace it with a 204.

deleugpn:
> What if the method returned a body of "Request successful"?

Would the code inspect the `Accept` headers and return JSON instead of plain string for API-based projects?

driesvints:
> > What if the method returned a body of "Request successful"?
> 
> Would the code inspect the `Accept` headers and return JSON instead of plain string for API-based projects?

Yeah, should be.

ashleyevans:
I think the wording of the ok() function is pretty ambiguous and not obvious what is returned in this instance for anyone else reading your code. ok() could return a string ‘OK’, an empty string, a true Boolean, or something else. It could also be confused with the Response::successful() check. As @ahinkle suggests ->noContent() would be the cleaner approach for me

I’d personally stick with using 204 headers for these instances, or return some documented success format

taylorotwell:
Think I'll wait on this for now since it seems pretty hotly debated.

For the record, I just don't return anything from a controller that I want to return an empty response. 🙂 An empty 200 response will be returned automatically. You can also do:

```php
use Illuminate\Http\Response;

return new Response;
```

PR #44761, opened 2022-10-27
Make prevent lazy loading behave as expected

dreadkopp:
As stated in the doc https://laravel.com/docs/9.x/eloquent-relationships#preventing-lazy-loading

Model::preventLazyLoading() should prevent lazy loading.

A change introduced in https://github.com/laravel/framework/commit/f26816df096b873c5ec7220bc3aa7634e3e8c4fa changes that behavior and ejects the lazy-loading prevention when there is only one model queried.

While the current implentation will not lead to N+1 problem performance impact it still:

- is unexpected behavior by name
  -  " i explicitly forbid lazy loading. it should always throw an exception when attempted " 
- is unexpected behavior in regards of documented behavior
  - "After preventing lazy loading, Eloquent will throw a Illuminate\Database\LazyLoadingViolationException exception when your application attempts to lazy load **any** Eloquent relationship."
- to my knowledge might degrade performance since i.e. database connection must be re-established

the current implentation shouldn't be triggered via ::preventLazyLoading (if at all). If an option as currently implented should exist it should be triggerd by maybe a method ::migitateNPlusOneQueryIssuesByPreventingLAzyLoadingWhenThereIsMoreThanOnItemInTheQueryResult() ... or something like that... dunno... naming is hard :)

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -391,12 +391,9 @@ class Builder implements BuilderContract
     {
         $instance = $this->newModelInstance();
 
-        return $instance->newCollection(array_map(function ($item) use ($items, $instance) {
+        return $instance->newCollection(array_map(function ($item) use ($instance) {
             $model = $instance->newFromBuilder($item);
-
-            if (count($items) > 1) {
-                $model->preventsLazyLoading = Model::preventsLazyLoading();
-            }
+            $model->preventsLazyLoading = Model::preventsLazyLoading();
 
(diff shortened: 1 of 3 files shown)


taylorotwell:
Sorry, I'm not changing behavior that's been around for almost a year on a patch release at this point.

dreadkopp:
@taylorotwell 

understood. 

before sending another dozen pull requests:

how do you feel about a second parameter to the preventLazyLoading() and shouldBeStrict() methods to indicate it will be 'as strict as it gets ?'

PR #44769, opened 2022-10-28
Enforce lazyLoading Prevention to be strict via flag

dreadkopp:
As stated in the doc https://laravel.com/docs/9.x/eloquent-relationships#preventing-lazy-loading

Model::preventLazyLoading() should prevent lazy loading.

A change introduced in https://github.com/laravel/framework/commit/f26816df096b873c5ec7220bc3aa7634e3e8c4fa changes that behavior and ejects the lazy-loading prevention when there is only one model queried.

This current behavior should not be altered thus in this PR a second parameter is introduced to 

Model::preventLazyLoading()
and
Model::shouldBeStrict()

to indicate that lazyLoadingPrevention will be enforced.

(imho the expected behavior and more matching to the documentation of those methods)

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -391,12 +391,9 @@ class Builder implements BuilderContract
     {
         $instance = $this->newModelInstance();
 
-        return $instance->newCollection(array_map(function ($item) use ($items, $instance) {
+        return $instance->newCollection(array_map(function ($item) use ($instance) {
             $model = $instance->newFromBuilder($item);
-
-            if (count($items) > 1) {
-                $model->preventsLazyLoading = Model::preventsLazyLoading();
-            }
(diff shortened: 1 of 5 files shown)


ankurk91:
There must be a reason behind the change made by themsaid, and you should also preserve that change

dreadkopp:
@ankurk91 

my first Pull request did actually change the default behavior

this one preserves the default behavior but making it more strict if enabled

to my understanding (and that of my colleagues when reading the docs) is that preventLazyLoading() will **always** prevent lazy loading

however in the current implementation it does not by default

in case a single result is returned by the builder it will be disabled since no n+1 query problems should araise.. n is 1 in that case.

however the method name says 'preventLazyLoading' not 'preventN+1Problems' :)

also, depending to the database setup, it might still lead to performance degredation since on (allowed) lazy loading relations on a single model instance the connection to the database might need to be re-established

taylorotwell:

PR #44770, opened 2022-10-28
[9.x] Accept body in getJson() method of MakesHttpRequests.php

didac-adria:
The getJson() method of  MakesHttpRequests.php doesn't accept a body. That method is widely used when doing tests for api routes. It is interesting to accept a body although being a get method especially for API calls which get a list of resources but filtered with multiple filters, which you can pass them through the body when the filter information is too big to include it in the url.

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/MakesHttpRequests.php
@@ -317,11 +317,12 @@ trait MakesHttpRequests
      *
      * @param  string  $uri
      * @param  array  $headers
+     * @param  array  $body
      * @return \Illuminate\Testing\TestResponse
      */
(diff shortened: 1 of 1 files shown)


driesvints:
Heya. This doesn't makes sense imo. A GET request is to retrieve content. You don't "post" a body with it. This is also a breaking change because the method signature got changed.

didac-adria:
Hey @driesvints thanks for your prompt response. As I described above, it could be interesting to pass a body when you are doing a get request with filters data which is big. Like an array of id's + multiple other values.

I can correct the breaking change if we agree on the above.

taylorotwell:

didac-adria:
Hey guys!

Thanks for your responses. I think we can leave it like this. It's not a super interesting thing that worths a debate.

See you!

PR #44772, opened 2022-10-28
[9.x] Added generic typing to EnumeratesValues::flatMap method.

mad-briller:
Improves the generic typing of `EnumeratesValues` by providing them for the `flatMap()` method.

Using `flatMap` right now causes all type information to be lost, and explicitly states there will only be integer keys, when `collapse()` can produce both `string` and `int` keys thanks to it's use of `array_merge()`.

ref: https://github.com/nunomaduro/larastan/discussions/1427

### DIFF
--- a/src/Illuminate/Collections/Traits/EnumeratesValues.php
+++ b/src/Illuminate/Collections/Traits/EnumeratesValues.php
@@ -384,8 +384,11 @@ trait EnumeratesValues
     /**
      * Map a collection and flatten the result by a single level.
      *
-     * @param  callable(TValue, TKey): mixed  $callback
-     * @return static<int, mixed>
+     * @template TFlatMapKey of array-key
+     * @template TFlatMapValue
+     *
+     * @param  callable(TValue, TKey): (\Illuminate\Support\Collection<TFlatMapKey, TFlatMapValue>|array<TFlatMapKey, TFlatMapValue>)  $callback
+     * @return static<TFlatMapKey, TFlatMapValue>
      */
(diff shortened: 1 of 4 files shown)


canvural:
`Enumerable` interface needs to be changed also.

src/Illuminate/Collections/Traits/EnumeratesValues.php
```diff
@@ -384,8 +384,11 @@ public function mapToGroups(callable $callback)
     /**
      * Map a collection and flatten the result by a single level.
      *
-     * @param  callable(TValue, TKey): mixed  $callback
-     * @return static<int, mixed>
+     * @template TFlatMapKey of array-key
+     * @template TFlatMapValue
+     *
+     * @param  callable(TValue, TKey): array<TFlatMapKey, TFlatMapValue>  $callback
```
tyabu12, line 390: I think that the following would be nice because `Arr::collapse()` allow also `Collection`.

https://github.com/laravel/framework/blob/2794f8139c46036db22054c0d3459badadbf3f9d/src/Illuminate/Collections/Arr.php#L42-L62

```suggestion
     * @param  callable(TValue, TKey): \Illuminate\Support\Collection<TFlatMapKey, TFlatMapValue>|array<TFlatMapKey, TFlatMapValue>  $callback
```

mad-briller: good spot, thanks

taylorotwell:
@nunomaduro can you take a look at this?

PR #44777, opened 2022-10-29
Add required if size validations

aligulzar729:
This PR adds the ability to make a field required if the size of the given field is equal to, less than (max) or greater that (min) the given size. This is quiet convenient, readable, shorter and doesn't need importing Rule class.
Also, I'm currently working on the same thing for exclude if validation.
carbon.


igorgaming:
Why not just use [sometimes](https://laravel.com/docs/9.x/validation#validating-when-present)?

taylorotwell:

igorgaming:
> > Why not just use [sometimes](https://laravel.com/docs/9.x/validation#validating-when-present)?
> 
> Because it's just a different story. Consider this scenario you want to validate attachments only when there are more than 2 attachment element in attachments array. How do you write the validation rule for this scenario? Is it doable with sometimes?

It sounds weird, do you have a real-life example?

agulzar-netlinks:
> > > Why not just use [sometimes](https://laravel.com/docs/9.x/validation#validating-when-present)?
> > 
> > 
> > Because it's just a different story. Consider this scenario you want to validate attachments only when there are more than 2 attachment element in attachments array. How do you write the validation rule for this scenario? Is it doable with sometimes?
> 
> It sounds weird, do you have a real-life example?

```php
  'attachments' => [
      'array',
      Rule::requiredIf(count(request('attachments', [])) > 0),
  ],
  'attachments.*' => [
      'required_with:attachments',
      'integer',
      'exists:attachments,id'
  ],
```

PR #44782, opened 2022-10-30
[9.x] Add option to allow raw output in php artisan serve

### LINKED ISSUE #44781: [9.x] php artisan serve issue printing arrays to console via print_r
- Laravel Version: 9.36.4
- PHP Version: 8.1.12

### Description:

Previously in Laravel 8.x and below, the output buffer using `php artisan serve` didn't had a custom handler. This handler was introduced in this commit a52f623cf6e7ed9be7f6e87f5eb411a383ad42fc via `handleProcessOutput`. 

The issue is that the current `handleProcessOutput` doesn't seem to handle well when `error_log` or `Log::` is used to print complex objects (e.g. via `print_r`).

Previous to this commit in 9.x, printing an array on a server running under `php artisan serve` would look like this:

```php
[Some Date] Array
(
    [0] => Foo
    [1] => Bar
)
```

After the commit, printing an array on a server running under `php artisan serve` would look like this (verbatim):

```php
   WARN  Array.  

   WARN  (.  

   WARN  => Foo.  

   WARN  => Bar.  

   WARN  ).  
```

As you can see, the `handleProcessOutput` is mangling the desire output on this scenario, something that didn't used to happen before.

### Steps To Reproduce:

Use `php artisan serve` and on any desired controller route use `Log::info(print_r($complexArray, true));` or `error_log(print_r($complexArray, true));`

### Next Steps:

Currently I'm manually changing back `./vendor/laravel/framework/src/Illuminate/Foundation/Console/ServeCommand.php`

```diff
-$process->start($this->handleProcessOutput());
+$process->start(function ($type, $buffer) {
+    $this->output->write($buffer);
+});
```

An ideal solution would be to introduce maybe a new flag to allow the end-user to customize this, such as:

```php
if ($this->option('no-output-handling')) {
    $process->start(function ($type, $buffer) {
        $this->output->write($buffer);
    });
} else {
    $process->start($this->handleProcessOutput());
}
```

or in a smaller form-factor

```php
$outputHandler = $this->option('no-output-handling') ? fn($type, $buffer) => $this->output->write($buffer) : $this->handleProcessOutput();
$process->start($outputHandler);
```

```diff
    protected function getOptions()
    {
        return [
            ['host', null, InputOption::VALUE_OPTIONAL, 'The host address to serve the application on', Env::get('SERVER_HOST', '127.0.0.1')],
            ['port', null, InputOption::VALUE_OPTIONAL, 'The port to serve the application on', Env::get('SERVER_PORT')],
            ['tries', null, InputOption::VALUE_OPTIONAL, 'The max number of ports to attempt to serve from', 10],
            ['no-reload', null, InputOption::VALUE_NONE, 'Do not reload the development server on .env file changes'],
+           ['no-output-handling', null, InputOption::VALUE_NONE, 'Do not process server output buffer'],
        ];
    }
```

I'd be happy to open a PR if you think this is something that could be added and you'd think it's not more complicated than what I depicted above.

Very open to feedback if I missed something that might resolve this. I'm trying to avoid using the built-in php dev server as `php artisan serve` provides me with a lot of goodies. I could not find any other alternate way to make it print like it used to before without extensive pre-formatting of the output string on the application side.

Other ideas for flag names: `no-output-processing`

I opened this as an issue as I didn't think this would be a "new feature" given that it used to work in previous versions.

samypr100:
Closes #44781

### DIFF
--- a/src/Illuminate/Foundation/Console/ServeCommand.php
+++ b/src/Illuminate/Foundation/Console/ServeCommand.php
@@ -148,7 +148,7 @@ class ServeCommand extends Command
             return in_array($key, static::$passthroughVariables) ? [$key => $value] : [$key => false];
         })->all());
 
-        $outputHandler = $this->option('no-output-processing') ? fn($type, $buffer) => $this->output->write($buffer) : $this->handleProcessOutput();
+        $outputHandler = $this->option('no-output-processing') ? fn ($type, $buffer) => $this->output->write($buffer) : $this->handleProcessOutput();
         $process->start($outputHandler);
 
         return $process;
--- a/src/Illuminate/Foundation/Console/ServeCommand.php
+++ b/src/Illuminate/Foundation/Console/ServeCommand.php
@@ -148,7 +148,8 @@ class ServeCommand extends Command
             return in_array($key, static::$passthroughVariables) ? [$key => $value] : [$key => false];
         })->all());
 
-        $process->start($this->handleProcessOutput());
+        $outputHandler = $this->option('no-output-processing') ? fn($type, $buffer) => $this->output->write($buffer) : $this->handleProcessOutput();
+        $process->start($outputHandler);
 
         return $process;
     }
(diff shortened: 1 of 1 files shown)


antonkomarev:
--output=raw

taylorotwell:
Use `dd` or `ray`

samypr100:
> Use `dd` or `ray`

@taylorotwell Thanks for the suggestions

I'm actually trying to see the unformatted output in the console rather than back in a reponse. I still want the response to contain the typical data (not die halfway), rather than a formatted `dd` output html page. This previously used to work in 8.x, hence why I opened the issue 😢 

Using `dd` does give you a nice page with colored output, but a dump and die is not what I'm looking for given above

```
array:2 [ // path/to/file.php
  0 => "Foo"
  1 => "Bar"
]
```

Using `ray` would mean integrating an external service and I was hoping to keep this local/offline given it's primarily for personal home projects

PR #44784, opened 2022-10-30
[10.x] database expressions with grammar-specific formatting

tpetry:
Last year I proposed PR #40115 to change database expressions to be grammar-enhanced. But my PR was changing too much, so I redid the whole concept to change as least as possible to be easier to merge for the next Laravel major release.

# What is the problem?

Laravel's database implementation provides us a good way of working with multiple databases while abstracting away the inner workings. We don't have to think about small syntax differences when using a query builder or how column names are escaped to not interfere with reserved keywords.

However, when we want to use more database functionality we have to fall back to raw database code and writing database-specific code to e.g. quote column names or column aliases:
```php

DB::table('table')
    ->when(isPostgreSQL(), fn ($query) => $query->select(DB::raw('coalesce("user", "admin") AS "value"')))
    ->when(isMySQL(), fn ($query) => $query->select(DB::raw('coalesce(`user`, `admin`) AS `value`')))
```

# How can this be solved?

Currently the `Expression` class returned by `DB::raw` is just a container to indicate that the value should not be further processed. My implementation changes that approach a little bit by declaring an expression contract `DB::raw` uses and custom classes can implement:
```php
interface Expression
{
    /**
     * Get the value of the expression.
     *
     * @param  \Illuminate\Database\Grammar  $grammar
     * @return scalar
     */
    public function getValue(Grammar $grammar);
}
```

Following that idea, the operations normally done by raw expressions and statements can be created as reusable expression classes. The former example could be written like this, which is totally database-agnostic for the one using those classes instead of raw expressions:
```php
DB::table('table')
    ->select(new Alias(new Coalesce(['user', 'admin']), 'count'));


class Alias implements Expression
{
    public function __construct(
    public readonly Expression|string $expression,
    public readonly string $name,
    ) { }

    public function getValue(Grammar $grammar): string
    {
        return match ($grammar->isExpression($this->expression)) {
            true => "{$grammar->getValue($this->expression)} as {$grammar->wrap($this->name)}",
            false => $grammar->wrap("{$this->name} as {$this->name}"),
        };
    }
}

class Coalesce implements Expression
{
    public function __construct(
        public readonly array $expressions,
    ) { }

    public function getValue(Grammar $grammar): string
    {
        $expressions = array_map(function ($expression) use($grammar): string {
            return match ($grammar->isExpression($expression)) {
                true => $grammar->getValue($expression),
                false => $grammar->wrap($expression),
            };
        }, $this->expressions);
        $expressions = implode(', ', $expressions);

        return "coalesce({$expressions})";
    }
}
```

# But isn't it much work to write those classes?

Kind of. But keep in mind these have to be written only once and you don't have to be the one. Like with every package created by the community anyone can start and write those SQL expressions for anyone to use. When this PR will be merged into 10.x I will start by writing the most common ones at [tpetry/laravel-query-expressions](https://github.com/tpetry/laravel-query-expressions).

# But I don't like the syntax

That's ok as I don't propose any syntax, I propose a system of how this can work. You can start a package that will provide a macro-able facade to do the same with static function calls as long as those calls return expression objects:

```php
Expr::alias(Expr::coalesce(['user', 'admin']), 'count')
```

An implementation even doesn't have to use this wrapping. I can imagine a parser could transform a special little syntax to create those object chains from a string:

```php
DB::table('table')
    ->select(exprParse('[user], [admin] ~> coalesce ~> alias:count'));
```

As said, this PR is not about providing syntax. It is about adding a building block packages can use to build something great

# Outlook

In future PRs I would like to work on more things useful for database expressions:

* **Access to the database type:**. Some functions are implemented differently by the databases. When the `Grammar` would provide methods to identify it as e.g. a MySQL grammar, the expression could use a different function for regex replacement than for e.g. PostgreSQL.
* **Value bindings:** Right now request-provided values can't be added to expressions as the value could contain a SQL injection. When the `Grammar` would provide method to `quote()` a value these could be safely injected into an expression.
* **Database version:** Some features are easily done on recent database versions as critical features have been added while a lot of boilerplate is needed for older releases. When the `Grammar` would provide the database type and version a grammar class could use best implementation for every database and version.

### DIFF
--- a/src/Illuminate/Database/Grammar.php
+++ b/src/Illuminate/Database/Grammar.php
@@ -2,7 +2,7 @@
 
 namespace Illuminate\Database;
 
-use Illuminate\Database\Query\Expression;
+use Illuminate\Contracts\Database\Query\Expression;
 use Illuminate\Support\Traits\Macroable;
 use RuntimeException;
 
@@ -31,7 +31,7 @@ abstract class Grammar
     /**
      * Wrap a table in keyword identifiers.
      *
-     * @param  \Illuminate\Database\Query\Expression|string  $table
+     * @param  \Illuminate\Contracts\Database\Query\Expression|string  $table
      * @return string
      */
     public function wrapTable($table)
@@ -46,7 +46,7 @@ abstract class Grammar
     /**
      * Wrap a value in keyword identifiers.
      *
-     * @param  \Illuminate\Database\Query\Expression|string  $value
+     * @param  \Illuminate\Contracts\Database\Query\Expression|string  $value
      * @param  bool  $prefixAlias
      * @return string
      */
@@ -208,14 +208,19 @@ abstract class Grammar
     }
 
     /**
-     * Get the value of a raw expression.
+     * Transforms expressions to their scalar types.
      *
-     * @param  \Illuminate\Database\Query\Expression  $expression
-     * @return mixed
+     * @param  scalar|\Illuminate\Contracts\Database\Query\Expression  $expression
+     * @return scalar
      */
     public function getValue($expression)
     {
-        return $expression->getValue();
+        $value = $expression;
+        if ($this->isExpression($value)) {
+            $value = $this->getValue($value->getValue($this));
+        }
+
+        return $value;
     }
 
     /**
--- a/src/Illuminate/Database/Grammar.php
+++ b/src/Illuminate/Database/Grammar.php
@@ -210,8 +210,8 @@ abstract class Grammar
     /**
      * Transforms expressions to their scalar types.
      *
-     * @param  scalar|\Illuminate\Contracts\Database\Query\Expression  $expression
-     * @return scalar
+     * @param  string|int|float|\Illuminate\Contracts\Database\Query\Expression  $expression
+     * @return string|int|float
      */
     public function getValue($expression)
     {
--- a/src/Illuminate/Database/Grammar.php
+++ b/src/Illuminate/Database/Grammar.php
@@ -210,17 +210,16 @@ abstract class Grammar
     /**
      * Transforms expressions to their scalar types.
      *
-     * @param  string|int|float|\Illuminate\Contracts\Database\Query\Expression  $expression
+     * @param  \Illuminate\Contracts\Database\Query\Expression|string|int|float  $expression
      * @return string|int|float
      */
     public function getValue($expression)
     {
-        $value = $expression;
-        if ($this->isExpression($value)) {
-            $value = $this->getValue($value->getValue($this));
+        if ($this->isExpression($expression)) {
+            return $this->getValue($expression->getValue($this));
         }
 
-        return $value;
+        return $expression;
     }
 
     /**
--- a/src/Illuminate/Database/Query/Expression.php
+++ b/src/Illuminate/Database/Query/Expression.php
@@ -10,14 +10,14 @@ class Expression implements ExpressionContract
     /**
      * The value of the expression.
      *
-     * @var scalar
+     * @var string|int|float
      */
     protected $value;
 
     /**
      * Create a new raw query expression.
      *
-     * @param  scalar  $value
+     * @param  string|int|float  $value
      * @return void
      */
     public function __construct($value)
@@ -29,7 +29,7 @@ class Expression implements ExpressionContract
      * Get the value of the expression.
      *
      * @param  \Illuminate\Database\Grammar  $grammar
-     * @return scalar
+     * @return string|int|float
      */
     public function getValue(Grammar $grammar)
     {
--- a/src/Illuminate/Database/Query/Expression.php
+++ b/src/Illuminate/Database/Query/Expression.php
@@ -2,19 +2,22 @@
 
 namespace Illuminate\Database\Query;
 
-class Expression
+use Illuminate\Contracts\Database\Query\Expression as ExpressionContract;
+use Illuminate\Database\Grammar;
+
+class Expression implements ExpressionContract
 {
     /**
      * The value of the expression.
      *
-     * @var mixed
+     * @var scalar
      */
     protected $value;
 
     /**
      * Create a new raw query expression.
      *
-     * @param  mixed  $value
+     * @param  scalar  $value
      * @return void
      */
     public function __construct($value)
@@ -25,20 +28,11 @@ class Expression
     /**
      * Get the value of the expression.
      *
-     * @return mixed
+     * @param  \Illuminate\Database\Grammar  $grammar
+     * @return scalar
      */
-    public function getValue()
+    public function getValue(Grammar $grammar)
     {
         return $this->value;
-    }
-
-    /**
-     * Get the value of the expression.
-     *
-     * @return string
-     */
-    public function __toString()
-    {
-        return (string) $this->getValue();
     }
 }
--- a/src/Illuminate/Contracts/Database/Query/Expression.php
+++ b/src/Illuminate/Contracts/Database/Query/Expression.php
@@ -1,0 +1,16 @@
+<?php
+
+namespace Illuminate\Contracts\Database\Query;
+
+use Illuminate\Database\Grammar;
+
+interface Expression
+{
+    /**
+     * Get the value of the expression.
+     *
+     * @param  \Illuminate\Database\Grammar  $grammar
+     * @return scalar
+     */
+    public function getValue(Grammar $grammar);
+}
--- a/src/Illuminate/Contracts/Database/Query/Expression.php
+++ b/src/Illuminate/Contracts/Database/Query/Expression.php
@@ -10,7 +10,7 @@ interface Expression
      * Get the value of the expression.
      *
      * @param  \Illuminate\Database\Grammar  $grammar
-     * @return scalar
+     * @return string|int|float
      */
     public function getValue(Grammar $grammar);
 }
--- a/src/Illuminate/Database/Concerns/BuildsQueries.php
+++ b/src/Illuminate/Database/Concerns/BuildsQueries.php
@@ -423,7 +423,7 @@ trait BuildsQueries
      */
     protected function getOriginalColumnNameForCursorPagination($builder, string $parameter)
     {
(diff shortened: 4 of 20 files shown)


Wulfheart:
This would be really nice in my opinion!

michaeldyrynda:
This is pretty slick, nicely done. 

Two things:

First, is the `new 'admin'` in your example a typo, or some new PHP 8 syntax?

Second, should we preserve the `__toString()` in `src/Illuminate/Database/Query/Expression.php` for backwards compatibility?

tpetry:
> First, is the `new 'admin'` in your example a typo, or some new PHP 8 syntax?
That is a typo. I simplified the example to make it easier to understand and missed the new keyword.

> Second, should we preserve the `__toString()` in `src/Illuminate/Database/Query/Expression.php` for backwards compatibility?
That's not easily possible. We could do that but then we have two different expression implementations that need to be handled differently. And I guess almost no one really extended the `Expression` class.

deleugpn:
I can see how this PR seems like a good compromise. It basically expands the `Expression` class to be able to take a Grammar as dependency in order to get help from the Grammar to generate the underlying DB expression.
Personally I don't like any potential syntax that will arise from this due to how one expression needs to wrap another one, but at least that's no longer the Framework's concern and will be handled exclusively outside of the framework.
Another thing that I think is extremely edge-case is the need to write a single raw expression compatible with multiple databases. Sure, things like Laravel Nova may theoretically benefit from such a feature, but truth is most applications will target only one database and when doing raw expressions, doing the raw expression of your database will likely yield a better syntax.

In summary, I don't particularly like the code that will be written as a result of this feature, but this feature in itself isn't that much harmful to the framework. I just wonder whether there's a simpler approach to provide the Grammar to the Expression class with less changes, but that requires digging deeper...

taylorotwell:
Can you provide the breaking changes and the upgrade steps we would need to include in the 10.x upgrade guide if this is merged?

tpetry:
Sure.

# Breaking Changes

The `Illuminate\Database\Query\Expression` class now implements the `Illuminate\Contracts\Database\Query\Expression` contract. An expression can now return any `scalar` value instead of the former requirement to return `string`s.

# Upgrade Guide

If you use `DB::raw` or create instances from `Illuminate\Database\Query\Expression` you don't have to change anything. All changes are transparent to you. But if you do extend the `Illuminate\Database\Query\Expression` class you have to change your implementation. The SQL code you returned in the `__toString` method needs to be moved to the `getValue` method.

morloderex:
@tpetry It might also be worth noting that because of the removal from `__toString()` method towards `getValue()` that any expression cannot be converted to a string ny typecasting anymore due to the now missing Grammar context in that method.

I wonder if we could add that back somehow? I am however not even sure if doing `(string) new \Illuminate\Database\Query\Expression('some expression')` is something we want to have?

tpetry:
I am not sure whether that counts as a breaking change for users of Laravel, as only the Laravel core is transforming those expressions into strings. I did choose against both ways to get the string for an expression because it increases the complexity of handling those expressions. One way is more straightforward than two ways. And as I said, I don't expect anyone having extended those classes. I only found the idea of doing it in a tweet which I used as an idea to work on this PR.

src/Illuminate/Database/Grammar.php
```diff
@@ -208,14 +208,19 @@ public function isExpression($value)
...
      *
-     * @param  \Illuminate\Database\Query\Expression  $expression
-     * @return mixed
+     * @param  scalar|\Illuminate\Contracts\Database\Query\Expression  $expression
+     * @return scalar
      */
     public function getValue($expression)
     {
-        return $expression->getValue();
+        $value = $expression;
```
rodrigopedra, line 218: Is this just to rename the variable?

If so, isn't it easier to rename the method's parameter?

tpetry: The parameter should be an expression as defined by the variable name. But some core function of Laravel ignores that and calls it with a value. So it exists to make it more clear.

nevadskiy:
@tpetry That is fantastic. What do you think about adding a separate `GrammarAwareInterface` that contains a single `setGrammar` method? I think that would allow to keep the original `toString` method and avoid the breaking change.

taylorotwell:
I think this look fairly good to me if we can get the conflicts fixed @tpetry 👍

tpetry:
@taylorotwell Done. Rebased it on the newest `10.x` commit.

src/Illuminate/Database/Grammar.php
```diff
@@ -208,14 +208,19 @@ public function isExpression($value)
     }
 
     /**
-     * Get the value of a raw expression.
+     * Transforms expressions to their scalar types.
      *
-     * @param  \Illuminate\Database\Query\Expression  $expression
-     * @return mixed
+     * @param  scalar|\Illuminate\Contracts\Database\Query\Expression  $expression
+     * @return scalar
```
driesvints, line 214: This doesn't seem correct. I can't think of any place where we used `scalar` before. Please add the correct types like int, string, or whatever's needed here.

src/Illuminate/Database/Query/Expression.php
```diff
@@ -2,19 +2,22 @@
...
+use Illuminate\Contracts\Database\Query\Expression as ExpressionContract;
+use Illuminate\Database\Grammar;
+
+class Expression implements ExpressionContract
 {
     /**
      * The value of the expression.
      *
-     * @var mixed
+     * @var scalar
```
driesvints, line 13: Please use the proper types here instead of "scalar". Do this for all other occurrences in this PR.

driesvints (changes requested):

src/Illuminate/Database/Grammar.php
tpetry: The correct type is now `int|float|string|bool` because the expression can transform these values to SQL. And that is exactly the definition of `scalar`. Do you want me to change it nevertheless?

driesvints: Yes please use those.

tpetry: Done

src/Illuminate/Database/Query/Expression.php
tpetry: Done

taylorotwell:
Thanks! 👍

willrowe:
@tpetry I'm curious, in what case would an expression ever return anything but a SQL string from `getValue`? I thought it was supposed to represent a SQL expression.

tpetry:
An integer or float is still representable as string. And it was completely valid to use them before, so I doesnt want to break this.

willrowe:
@tpetry I know it allowed those types previously, I was just wondering if there was an example of a situation in which something other than a string would need to be returned.

taylorotwell:
Hey @tpetry - I've finally finished my review of this. Thanks for your patience!

I wanted to get some further elaboration on your "Outlook" section regarding adding bindings / quoted values from the expression. What was your thoughts there on how that would work. While this PR definitely provides an improvement over our current expression capabilities, I can see people quickly asking for the ability for their expressions to be able to safely add bindings to the query. Any thoughts on how that would look or be implemented?

tpetry:
I did that in my old PR. Basically we have to the grammar. In the future (in my opinion) the query grammar instance should contain a database connection object (PDO). That way it could provide escaping of raw content based on the current connection settings (the charset encoding is important with mysql to prevent sql injections!) or export the version number of the database (to use newer sql features when available).

I could work backport this one tomorrow or tuesday if this merged. And we‘ll check it for 10.x or better plan it for 11.x.

taylorotwell:
Now that I have a pretty decent grasp of this PR we can just try to expose the PDO connection to the grammar with another commit on this PR if you don't mind?

tpetry:
I can do that, on tomorrow. If those changes are then too much for you, feel free to just revert the additional commit.

tpetry:
While thinking about it, I would like to delay that change. I don‘t want to break the 10.x release. Let me work on the grammar thing in a backward compatible way within L10. I guess other db providers (like mine, staudenmeier or singlestore) may break with such a change? Its a little bit late for that in the 10.x process. I‘ll find a way shortly after the 10.x release for grammars to opt-in for that feature if they want support it.

tpetry:
@taylorotwell Sorry, I did write the last message misleading last night. I am ok with merging this PR. Adding the connection to grammars I would like to work on shortly after the 10.x release to do it in a safe backwards-compatible way. I don‘t want to break other non-core drivers so shortly before the final release.

danrichards:
Was there a reason for not just passing in an extension of `Expression` that implements your own `__toString()` method.

https://www.php.net/manual/en/class.stringable.php

I might be missing something?

joelharkes:
@tpetry Why did you not add the classes like `Alias` in the PR? i would be nice if they were generally available to anyone.

garygreen:
I think the impact this has on developers trying to upgrade to Laravel 10 has been a understated.

It was described as a breaking change when developers have "manually" casted database attributes, or extend the QueryExpression, but the reality is when things are type-casted e.g. in arguments it automatically has been handled in Laravel up until this PR.

For example, this is now broken in Laravel 10. Essentially *all* references to `DB::raw()` will need to be updated.

**Doesn't work in Laravel 10:**

```php
$users = DB::select(DB::raw('SELECT username FROM users LIMIT 2'));
$user = DB::selectOne(DB::raw('SELECT username FROM users LIMIT 1'));
```

image



It also affects if you do something like this:

```php
$pushNotification->update([
	'fail_count'   => 0,
	'last_sent_at' => now(),
	'sent_count'   => \DB::raw('sent_count + 1'), // needs updating
]);
```

And if your using various methods to inject a custom bit of sql, this needs updating:

```php
$query->addSelect(\DB::raw(...));
$query->leftJoin(\DB::raw(...));
$query->orderBy(\DB::raw(...));
$query->groupBy(\DB::raw(...));
...lots of other methods
```

It's fairly trivial to update all references to `DB::raw()` by appending `->getValue()`, but I do think this level of breaking change is far reaching and extensive, which is unfortunate as this PR wasn't absolutely essential. Many developers will be caught out by this change, including the mature vast ecosystem of Laravel packages, migrations, anybody using custom queries.

PR #44790, opened 2022-10-31
call custom job in queue config for customize fire method and other

mohammadgh1370:
customize job queue access change fire method in job

### DIFF
--- a/src/Illuminate/Queue/BeanstalkdQueue.php
+++ b/src/Illuminate/Queue/BeanstalkdQueue.php
@@ -31,6 +31,13 @@ class BeanstalkdQueue extends Queue implements QueueContract
     protected $timeToRun;
 
     /**
+     * The job handler name.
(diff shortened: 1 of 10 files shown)


src/Illuminate/Queue/BeanstalkdQueue.php
```diff
@@ -43,20 +50,23 @@ class BeanstalkdQueue extends Queue implements QueueContract
...
      * @param  int  $timeToRun
+     * @param  string|null  $job
      * @param  int  $blockFor
      * @param  bool  $dispatchAfterCommit
      * @return void
      */
     public function __construct(Pheanstalk $pheanstalk,
                                 $default,
                                 $timeToRun,
+                                $job = null,
```
driesvints, line 61: Adding a new argument like this is a breaking change unfortunately.

PR #44791, opened 2022-10-31
call custom job in queue config for customize fire method and other

mohammadgh1370:
customize queue job and change fire method


driesvints:
I think I closed a similar or just now? Please don’t resend this in.

PR #44792, opened 2022-10-31
[9.x] Fix `whenPivotLoaded(As)` api resource methods when using Eloquent strict mode

gdebrauwer:
When the "Prevent accessing missing attributes on retrieved models" feature of Eloquent strict mode is enabled, the code `$this->resource->$accessor` will cause an unwanted exception. Using `isset` method fixes this.


taylorotwell:
So tired of strict mode 😢

PR #44793, opened 2022-10-31
[9.x] Add source file to Collection's dd method output

hasyirin:
This pull request further extends #44211 by @nunomaduro to work on `Collection` by adding the source file/line to the `dd` output of the `Collection` class method.

After using the improved `dd`, I found it odd that `Collection` class `dd` method did not give the same output as the current improved `dd` function. The workaround was to wrap the collection variable with `dd` instead of using the built in `dd` method of `Collection` class.

### DIFF
--- a/src/Illuminate/Foundation/Concerns/ResolvesDumpSource.php
+++ b/src/Illuminate/Foundation/Concerns/ResolvesDumpSource.php
@@ -56,12 +56,17 @@ trait ResolvesDumpSource
         $sourceKey = null;
 
         foreach ($trace as $traceKey => $traceFile) {
-            if (isset($traceFile['file']) && str_ends_with(
-                $traceFile['file'],
-                'dump.php'
-            )) {
-                $sourceKey = $traceKey + 1;
+            if (! isset($traceFile['file'])) {
+                continue;
+            }
 
+            if (str_ends_with($traceFile['file'], 'dump.php')) {
+                $sourceKey = $traceKey + 1;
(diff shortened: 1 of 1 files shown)


nunomaduro:
@AhrimFakhriy you think you can create a static property of the trait, that key value map with (name => traceKey), so we can extend this work to other files?

AhrimFakhriy:
@nunomaduro will this suffice?

nunomaduro (changes requested):

src/Illuminate/Foundation/Concerns/ResolvesDumpSource.php
```diff
@@ -29,6 +29,16 @@ trait ResolvesDumpSource
...
     ];
 
+    /**
+     * All of the trace names and its keys.
+     *
+     * @var array<string, int>
+     */
+    protected static $traceNames = [
+        'symfony/var-dumper/Resources/functions/dump.php' => 1,
+        'Illuminate/Collections/Traits/EnumeratesValues.php' => 4,
```
nunomaduro, line 39: This changes work on Windows due the different separator?

AhrimFakhriy:
@nunomaduro Im sorry, I didnt know I'm not supposed to click on the "Ready for review" button myself.
As for your question, I added in a `str_replace` which replace the slash to `DIRECTORY_SEPARATOR` in the loops which should make it work for Windows, and I've tried on both Linux & Windows.

PR #44796, opened 2022-10-31
[9.x] Allow component DI through `render()`

browner12:
Currently the component will try to resolve dependencies that are passed via the constructor, which is not always ideal. This change allows you to specify on your component that you would like dependencies resolved on the `render()` method instead of the constructor.

For example, let's say you have a `UserCard` component that has simple output for anonymous users, and a fancy output for registered ones.

```php
class UserCard extends Component
{
    public function __construct(
        protected User|null $user = null,
        protected string $name = null,
        protected string $email = null,
    ) {}

    public function render()
    {
        $user = $this->user;
        $name = $this->name;
        $email = $this->email;

        return view('components/user-card', compact('user', 'name', 'email');
    }
}
```

and the view looks something like this:

```blade
<div id="user.card">
    @if(! is_null($user))
         //fancy output
    @else
        //plain output
    @endif
</div>
```

If we call the component via `<x-user-card name="Andy" email="andy@email.com"></x-user-card>`, we might think we'd get the plain output, but Laravel dependency injects an empty `User` into our component, making it not null, and rendering the fancy output.  Yes, this can be handled, but I believe it's an annoyance that can be avoided.

---

This also gives us the ability to use native PHP to enforce required parameters. Let's say we have a `ProductCard` component.

```php
class ProductCard extends Component
{
    public function __construct(
        protected Product $product,
    ) {}
}
```

Without this PR, you could make a `<x-product-card>`, and barring any view errors, you might not know that you forgot to pass a required parameter, because Laravel will have automatically injected an empty `Product` in for you.

### Precedence

There is also precedence for this switch, as Jobs are handled the same way.  Jobs do not do any dependency injection via the constructor, but rather via the `handle()` method. This gives the developer better control over how to build their Job, but still gives them the easy DI that we're used to.

### Future Thoughts

IMO, this is how components should behave by default. If there's interest, I'd love to send a PR through to 10.x as well.

### DIFF
--- a/src/Illuminate/View/Component.php
+++ b/src/Illuminate/View/Component.php
@@ -76,11 +76,11 @@ abstract class Component
     protected static $constructorParametersCache = [];
 
     /**
-     * Get the view / view contents that represent the component.
+     * Toggle to resolve the via the constructor or the render method.
      *
-     * @return \Illuminate\Contracts\View\View|\Illuminate\Contracts\Support\Htmlable|\Closure|string
+     * @var bool
      */
-    abstract public function render();
+    protected static $shouldResolveConstructor = true;
 
     /**
      * Resolve the component instance with the given data.
@@ -98,7 +98,7 @@ abstract class Component
 
         $dataKeys = array_keys($data);
 
-        if (empty(array_diff($parameters, $dataKeys))) {
+        if (! static::$shouldResolveConstructor || empty(array_diff($parameters, $dataKeys))) {
             return new static(...array_intersect_key($data, array_flip($parameters)));
         }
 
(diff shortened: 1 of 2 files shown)


browner12:
Apparently my brain was not working yesterday. Since I set this as a static property, this changes applies globally to all components. So I created a public static method that can be called via a service provider, mimicking the behavior of the new Model "strict" options.

taylorotwell:

browner12:
Can I get an explanation for this one?  This seems like a pretty solid change, as it cleans up some clearly laid out oddities with components.

PR #44799, opened 2022-11-01
[9.x] Allow toggle off of public property and method injection

browner12:
The components currently use reflection to automatically include all public properties and public methods in the component's rendered view.  This change allows developers to toggle this functionality off.

While I definitely value a lot of the magic that Laravel puts into the framework, when it's easy to be explicit I prefer to be explicit. We currently **never** use `public` properties on our components and force all data to be explicitly passed.  This change would allow us to enforce this a little more.

I would also guess there's a small performance benefit by avoiding all the reflection, Collection methods, and loops, but it would depend on the number of components rendered per request. I don't have any data to back this up, purely "a priori".

### DIFF
--- a/src/Illuminate/View/Component.php
+++ b/src/Illuminate/View/Component.php
@@ -76,6 +76,11 @@ abstract class Component
     protected static $constructorParametersCache = [];
 
     /**
+     * @var bool
+     */
+    protected static $shouldInjectPublic = true;
+
+    /**
      * Get the view / view contents that represent the component.
      *
      * @return \Illuminate\Contracts\View\View|\Illuminate\Contracts\Support\Htmlable|\Closure|string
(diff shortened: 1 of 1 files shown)


dennisprudlo:
I think if this switch is available in view components it should also be available in mailables where public properties are auto-injected as well. Or simply as a framwork-wide setting.

schonhoff:
Would love to see this! But I have some comment to the subject in general.

Wouldn't it be great if this is a static flag like the new strict security features Laravel added lately? In my opinion it would be great to toggle the behavior on the app level instead on the component level. (See this PR https://github.com/laravel/framework/pull/44283) Or maybe we need both?

And as @dennisprudlo said it would be nice if the mailables had the same behavior.

browner12:
I think that's a great idea @schonhoff. I was definitely thinking about the new "strict" feature when I built this. Let me make a couple tweaks to this PR.

As for the mailables, if this PR gets accepted we could definitely look into other places to add similar functionality.

browner12:
Test failure appears to be unrelated to change.

@schonhoff, let me know what you think of these changes.

schonhoff:
This changes looks good to me. Hopefully someone from the Laravel team will review the change.

taylorotwell:
Not sure I want to go down this road. Third party components / tools would have no idea what they can depend on in terms of component behavior. Facing similar problems with Eloquent "strict mode" option.

browner12:
Thanks for the explanation.

Will keep noodling on other ways we can maybe improve this.

PR #44816, opened 2022-11-02
[9.x] Add CSP nonce to Vite reactRefresh inline script

andyvenus:
This PR adds the Laravel Vite CSP nonce to the inline javascript generated by the Blade directive `@reactRefresh` (`Vite::reactRefresh`).

This allows hot reloading to work for React projects where unsafe inline javascript is disabled via the CSP.


timacdonald (approved):
Tested and working as expected. Thanks for the improvement @andyvenus

PR #44821, opened 2022-11-02
Fix InteractsWithContainer::withoutMix

alies-dev:
Original `mix()` returns `HtmlString` instance, fake-mix should return the same in case if a developer uses `toHtml()` explicitly. Example:

```php
mix('asset.css')->toHtml()
```


driesvints:
It seems that this breaks the test suite.

PR #44823, opened 2022-11-02
[8.x] Fix too many channels with pusher broadcasting

laabidi1990:
This PR fixes a problem with the Pusher channel, so that it sends events in chunks of 100, because Pusher does not allow more than that number of channels.


driesvints:
Laravel 8 is closed for bug fixes. Feel free to attempt 9.x but make sure tests are passing.

PR #44825, opened 2022-11-02
[9.x] Allow route group method to be chained

ockle:
This allows the `group` method to be chain called when defining routing. An example of where this would be useful:

```php
// In RouteServiceProvider
Route::middleware('web')
    ->domain('foo.example.com')
    ->group(base_path('routes/common.php')
    ->group(base_path('routes/foo.php');

Route::middleware('web')
    ->domain('bar.example.com')
    ->group(base_path('routes/common.php')
    ->group(base_path('routes/bar.php');
```


PerryvanderMeer:
@ockle dockblocks should be updated too 🚀

ockle:
@PerryvanderMeer Updated, thanks 👍

PR #44832, opened 2022-11-03
PendingDispatch can be dispatched manually

NilLlisterri:
The PendingDispatch was dispatched when it was destroyed, and it was not possible to retrieve the generated job ID (when using queues). This change allows dispatching the job like this, obtaining the new job ID:  
```$job_id = ProcessPodcast::dispatch()->onQueue('emails')->now();```

It should be backwards-compatible. The dispatch logic in the `__destruct` method has been moved to the new `dispatch` method, which executes the previous statements if the new property `dispatched` is `false`, the default value.
This property will only be changed to `true` when dispatched to ensure the job is not dispatched twice.

I have not found any other way to obtain the queued job ID without having to extend large parts of the framework, but there might be another way. If that is the case, please forgive me and close the request.

Thanks!

### DIFF
--- a/src/Illuminate/Foundation/Bus/PendingDispatch.php
+++ b/src/Illuminate/Foundation/Bus/PendingDispatch.php
@@ -23,6 +23,13 @@ class PendingDispatch
      * @var bool
      */
     protected $afterResponse = false;
+    
+    /**
+     * Indicates if the job has been already dispatched
+     *
+     * @var bool
+     */
+    protected $dispatched = false;
 
     /**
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by taylorotwell** [comment]
`now` to me sounds like it will use the `sync` queue.

You can just pass a job directly to `Bus::dispatch` if you really need the job ID?

```php
$jobId = Bus::dispatch(new Job);
```

NilLlisterri:
Thanks for your response @taylorotwell, with this approach I can also send it to a specific queue, no caveats.

PR #44835, opened 2022-11-03
Customizable blade component tag prefix

tombanksme:
This pull request allows developers to customize the component tag prefix Laravel uses to identify blade components.

In my projects, I have a Laravel backend with a Vue frontend. For my Vue components, I use the `x-` prefix. When I tried to use these components in a Blade file, I received `InvalidArgumentException: Unable to locate a class or view for component [header].`

I have resolved the issue using `Blade::withoutComponentTags()` in a service provider. It would be nice to have the option to use blade components in the future, so this pull request adds a `Blade::setComponentTagPrefix($prefix)` option to the Blade compiler. I've added a test case to verify that the `ComponentTagCompiler` uses the prefix. It should be sufficient but shout if I need to add more tests (and what you'd like tested).

```php
Blade::setComponentTagPrefix("blade"); // <blade-component-name />
```

This issue could be resolved by "just using a different prefix for your Vue components"; but alas, I'm stubborn, and don't want to. And that was a good enough reason for me to write the code & this PR.

Let me know what you think!

### DIFF
--- a/src/Illuminate/View/Compilers/BladeCompiler.php
+++ b/src/Illuminate/View/Compilers/BladeCompiler.php
@@ -152,6 +152,13 @@ class BladeCompiler extends Compiler implements CompilerInterface
     protected $compilesComponentTags = true;
 
     /**
+     * The prefix for component tags.
+     *
+     * @var string
+     */
+    protected $componentTagPrefix = 'x';
+
+    /**
      * Compile the view at the given path.
      *
      * @param  string|null  $path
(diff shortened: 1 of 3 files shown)


driesvints:
I'm really sorry but I want to push back against this in the highest way. `<x-` is currently the main way in the Laravel ecosystem to identify Blade components. Making this configurable will just confuse people when reading tutorials and code as you wouldn't be able to distinct anymore between Blade, Vue, etc. If I'd have to be honest I'd say that if you wanted to use Laravel you'd have to choose a different prefix for your components to not collide with them.

taylorotwell:
No plans to allow this.

PR #44836, opened 2022-11-03
[10.x] Validate objects with & without PHP attributes

imdhemy:
This PR provides a new feature to the Validator class. It allows object validation against a set of assertions. Asserts can be added by using a PHP attribute.

#### Using the Assert attribute

Add the `Assert` attribute to a property of a class. The attribute takes an array of rules as a parameter.

```php
class User
{
    #[\Illuminate\Validation\Assert(['required', 'email'])]
    public string $email;
}

$user = new User();
$user->email = 'invalid-email';

$validator = \Illuminate\Support\Facades\Validator::make($user, []);

$validator->fails(); // true
```

#### Override Assert attributes

The `Assert` attributes could be overridden by the classic rules array:

```php
class User
{
    #[\Illuminate\Validation\Assert(['required', 'email'])]
    public string $email;
}

$user = new User();
$user->email = 'invalid-email';

$validator = \Illuminate\Support\Facades\Validator::make($user, [
    'email' => ['required'],
]);

$validator->fails(); // false
```

#### Validate objects

The new implementation allows object validation against the classic rules array without the need to add the `Assert` attribute to the object.

```php
class User
{
    public string $email;
}

$user = new User();
$user->email = 'invalid-email';

$validator = \Illuminate\Support\Facades\Validator::make($user, [
    'email' => ['required', 'email'],
]);

$validator->fails(); // true
```

---

## Thinking out loud

This is a **draft**, it would be nice to have some feedback on the implementation.

- Should the rules parameter of the `Validator` be optional?
- The `parseData` method uses getters if they are available, how will this affect the usage with Models?
- Still not sure about the naming, is `Assert` suitable for that functionality?

### DIFF
--- a/src/Illuminate/Contracts/Validation/Factory.php
+++ b/src/Illuminate/Contracts/Validation/Factory.php
@@ -7,13 +7,13 @@ interface Factory
     /**
      * Create a new Validator instance.
      *
-     * @param  array  $data
+     * @param  array|object  $data
      * @param  array  $rules
      * @param  array  $messages
      * @param  array  $customAttributes
      * @return \Illuminate\Contracts\Validation\Validator
      */
-    public function make(array $data, array $rules, array $messages = [], array $customAttributes = []);
+    public function make(array|object $data, array $rules, array $messages = [], array $customAttributes = []);
 
     /**
      * Register a custom validator extension.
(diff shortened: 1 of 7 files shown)


driesvints:
Heya thanks. But I'd personally really not go down this path. Just complicates things with attributes from what you can already do :/

imdhemy:
Thank you Dries for your feedback. Apart from the attributes, what do you think about allowing validating objects, not just arrays?

driesvints:
I don't think we should complicate the validation component like that.

PR #44838, opened 2022-11-04
[9.x] Introduced flag without-interactive-output

ramasofficial:
We have a DataDog logs parsing issues, because when it was added without a new line, we should have to change our Logger monolog format, something like this:
`'format' => "\n%channel%.%level_name%: %message% %context% %extra%\n", // New line before`

I think that others also have the same issue.
I fixed it into this PR.

### DIFF
--- a/src/Illuminate/Queue/Console/WorkCommand.php
+++ b/src/Illuminate/Queue/Console/WorkCommand.php
@@ -250,13 +250,13 @@ class WorkCommand extends Command
             return;
         }
 
-        if (! $useLineBreak) {
-            $this->output->write($output);
+        if ($useLineBreak) {
+            $this->output->writeln($output);
 
             return;
         }
 
-        $this->output->writeln($output);
+        $this->output->write($output);
     }
 
     /**
--- a/src/Illuminate/Queue/Console/WorkCommand.php
+++ b/src/Illuminate/Queue/Console/WorkCommand.php
@@ -214,28 +214,45 @@ class WorkCommand extends Command
 
             $formattedStartedAt = Carbon::now()->format('Y-m-d H:i:s');
 
-            $this->output->writeln("  <fg=gray>{$formattedStartedAt}</> {$job->resolveName()}");
-
-            return;
+            return $this->writePrettyOutput("  <fg=gray>{$formattedStartedAt}</> {$job->resolveName()}");
         }
 
         if ($this->latestStatus && $this->latestStatus != 'starting') {
             $formattedStartedAt = Carbon::createFromTimestamp($this->latestStartedAt)->format('Y-m-d H:i:s');
 
-            $this->output->writeln("  <fg=gray>{$formattedStartedAt}</> {$job->resolveName()}");
+            $this->writePrettyOutput("  <fg=gray>{$formattedStartedAt}</> {$job->resolveName()}");
         }
 
         $runTime = number_format((microtime(true) - $this->latestStartedAt) * 1000, 2).'ms';
-        $dots = max(terminal()->width() - mb_strlen($runTime) - 8, 0);
+        $dots = max(terminal()->width() - mb_strlen($job->resolveName()) - mb_strlen($runTime) - 31, 0);
 
-        $this->output->write('  '.str_repeat('<fg=gray>.</>', $dots));
-        $this->output->write(" <fg=gray>$runTime</>");
+        $this->writePrettyOutput(' '.str_repeat('<fg=gray>.</>', $dots));
+        $this->writePrettyOutput(" <fg=gray>$runTime</>");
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm sorry - can you explain this issue a little better. Your change makes the output worse IMO:

CleanShot 2022-11-04 at 16 04 08@2x

ramasofficial:
Imagine what we're logging to stdout and DataDog reads stdout and process every line by line, so when we're logging something to stdout it appears directly in console after Job Class name and it is in the same line with Job Class without break. So first log in job is always unprocessed.

ramasofficial:
@taylorotwell fixed it, please check it, should show dots now through all width.

nunomaduro:
We can't please both pretty / interactive output and logging services.

@ramasofficial Can you consider listen the queue events, and create your own custom entry on the logs that get sent to DataDog? Here are the events you can listen to:

```
Illuminate\Queue\Events\JobExceptionOccurred
Illuminate\Queue\Events\JobQueued
Illuminate\Queue\Events\QueueBusy
Illuminate\Queue\Events\JobFailed
Illuminate\Queue\Events\JobReleasedAfterException
Illuminate\Queue\Events\WorkerStopping
Illuminate\Queue\Events\JobProcessed
Illuminate\Queue\Events\JobRetryRequested
Illuminate\Queue\Events\JobProcessing
```

crazynds:
I dont think this change is good for output to the console . For every line you had before now you will have two of them with the same information.

ramasofficial:
@crazynds why 2, didn't understand what you had in mind with this, arguments please. What will I have same?

ramasofficial:
> We can't please both pretty / interactive output and logging services.
> 
> 
> 
> @ramasofficial Can you consider listen the queue events, and create your own custom entry on the logs that get sent to DataDog? Here are the events you can listen to:
> 
> 
> 
> ```
> 
> Illuminate\Queue\Events\JobExceptionOccurred
> 
> Illuminate\Queue\Events\JobQueued
> 
> Illuminate\Queue\Events\QueueBusy
> 
> Illuminate\Queue\Events\JobFailed
> 
> Illuminate\Queue\Events\JobReleasedAfterException
> 
> Illuminate\Queue\Events\WorkerStopping
> 
> Illuminate\Queue\Events\JobProcessed
> 
> Illuminate\Queue\Events\JobRetryRequested
> 
> Illuminate\Queue\Events\JobProcessing
> 
> ```

We're logging these events not when some events happening so this way is not suit for us. You can have a logging just in service and the main thing was that first log is always unparsed with DataDog, because it's in the same line as a logged Job Class name. So when it was released and we upgraded Laravel, we had an issue.

crazynds:
> @crazynds why 2, didn't understand what you had in mind with this, arguments please. What will I have same?



As you can see in the image you have a line for ```2022-11-06 09:29:17 App\Jobs\TestJobs\TestJob``` and a line for ```...........(...) 13.92ms DONE```

Without this change you have only 5 lines for 5 jobs, after you have 10 for the same 5 jobs, that's my point.

ramasofficial:
@crazynds yes, but now we and others who have DataDog and using DataDog recommended installing approach are having issues, because of this. In our case about 20 projects at least.

taylorotwell:
Could we provide some sort of alternative format flag? `--json` or something that would output the logs as JSON?

taylorotwell:
@ramasofficial feel free to give it a go. thanks.

ramasofficial:
@taylorotwell It's already finished. Can you check or everything is okay and suits for you?

src/Illuminate/Queue/Console/WorkCommand.php
```diff
@@ -214,26 +215,48 @@ protected function writeOutput(Job $job, $status)
...
+            return;
+        }
+
+        if (! $useLineBreak) {
+            $this->output->write($output);
+
+            return;
+        }
+
+        $this->output->writeln($output);
```
StyleShit, line 259:
```suggestion
        if ($useLineBreak) {
            $this->output->writeln($output);

            return;
        }

        $this->output->write($output);
```

I would swap those, just so the condition will be truthy. WDYT?

ramasofficial: Good point, already did it!

StyleShit:
@ramasofficial 

LOL I'm just some random reviewer, I'm not authorized for anything...
Taylor will review it anyway...

ramasofficial:
> @ramasofficial
> 
> LOL I'm just some random reviewer, I'm not authorized for anything... Taylor will review it anyway...

Didn't know it 😄

taylorotwell:
If you don't want those message logged just use the existing `php artisan queue:work --quiet`?

PR #44841, opened 2022-11-04
[9.x] Fixed 404 internal message output error in production

andrey-helldar:
## Current state

Laravel always returns a reference to the model despite changing the `APP_DEBUG` option (`true` or `false`).

image

### `APP_DEBUG = false`

image

### `APP_DEBUG = true`

image

### Additional functionality

The text of the message may be subject to translation:

```php
__('No query results for model [:model]:ids', [...])
__('Not Found')
```


andrey-helldar:
For now, I'll cancel this PR. After passing the test, I saw information that the exception file does not know and should not know anything about the application configuration.

I'm still thinking about how to solve this problem.

PR #44845, opened 2022-11-04
[9.x] Improve missing attributes functionality

bert-w:
This is an improvement for the recently merged https://github.com/laravel/framework/pull/44283.

The following changes are included:

1. I got an error in my own project because I was using the `HasAttributes` trait for a custom class. However, the following piece of code makes a call to the properties `$this->exists` and `$this->wasRecentlyCreated` first which did not exist in my implementation, therefore throwing an "Undefined property" ErrorException:

https://github.com/laravel/framework/blob/0702f556f72745f34570fbdfa8e37b8ea0b7ffa5/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L465-L478

That if-statement has now been reordered to first call the `static::preventsAccessingMissingAttributes()`, and since this is `false` by default the code will work correctly.

2. I moved all associated code to a new concern so this functionality becomes slightly more scoped instead of being added to the 2k+ lines monsterclass `HasAttributes`.
3. I updated a couple calls in the Model class where these functions were used, so it can now work without the trait.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -451,7 +451,7 @@ trait HasAttributes
         // Here we will determine if the model base class itself contains this given key
         // since we don't want to treat any of those methods as relationships because
         // they are all intended as helper methods and none of these are relations.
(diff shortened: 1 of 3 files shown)


taylorotwell:

Thanks!

Issue #44697, opened 2022-10-23
Can't prevent error 500 using Route::resource/apiResource in pgsql


crazynds:
- Laravel Version: 9.35.1
- PHP Version: 8.1.4
- Database Driver & Version: pgsql 13

### Description:

I came across the problem started while testing a route generated using Route::apiResource, and noticed that when any string is sent as a parameter, an error 500 occurs due to Laravel performing a database query.

When an integer is set as the parameter, it finds the model by its id and binds it properly:
![image](https://user-images.githubusercontent.com/36674096/197398300-73bee07b-2523-496c-895a-a102b7bd7201.png)

On the other hand, it throws a status of 500 when the parameter is a string:
![image](https://user-images.githubusercontent.com/36674096/197398351-489206ef-b38b-40a8-a3be-6a06b9ccb062.png)

The route is defined as follow:
```php
<?php

use Illuminate\Support\Facades\Route;

// /api/...

Route::apiResource('client', 'ClientController');
```
Some solutions I thought of:

- Catch the exception and ignore it. However, this solution prevents other database exceptions from being logged that have not yet been identified in the system.
- Apply a regex to the parameter and laravel just ignores the route and displays a 404. However, it is not possible to apply a regex to a resource or an apiResource. *

Without the regex rule, the server receiving a string when it expects to receive an integer in the parameter returns an error 500, which should not happen. It should just show that the route was not found (404) or that the request is invalid (400 or 422), since its a client error, not the other way around.

*There is a way to define a pattern in a route parameter, and this pattern applies to resources as well, as explained in this [thread](https://stackoverflow.com/questions/22622629/regular-expression-route-constraint-for-a-resource-route), however, it does not scale.

### Why shouldn't we let error 500 happen?

Code 500 is defined as "The server has encountered a situation it does not know how to handle.". And in this case we know exactly what the problem is and if we can ignore the request, or respond with a status of an invalid request (400-499), so if we want to keep track of all the server errors 500 to be able to fix them in the future, the system will be filled with false errors. 

### Steps To Reproduce:

- Create a complete CRUD of one model and use resource to bind the routes.
```php
<?php

use Illuminate\Support\Facades\Route;

// /api/...

Route::apiResource('client', 'ClientController');
```
- Send a request with string instead of integer in the parameter
- Error 500 happened 😔

Edit:

As mentioned by @dwightwatson, this error only happens when using postgresql.

mateusmdr:
[...]

PR #44846, opened 2022-11-04
[9.x] Added multi-resource support to PendingResourceRegistration class

crazynds:
The problem started with issue #44697 where it was impossible to add a regular expression to a resource, however when checking the source code there was already implementation a support for some functions that change the resource options (where, only, name, middleware, etc...).

But I saw that the problem I was having was when using the ```resources()``` function (multiple resource) which returns a void instead of a manipulable object.

This pull request solves this by changing the way the PendingResourceRegistration class works to work with multiple resources instead of just one and allowing to manipulate ```resources``` function options as well

Therefore, the following code snippet will work as expected:

```php

use Illuminate\Support\Facades\Route;


Route::apiResources([
     'client' => 'ClientController',
     'client.contact' => 'ContactController',
])->where(['client' => '[0-9]+', 'contacts' => '[0-9]+']);
```

I would really like a function in which I could constrain the format of all the parameters of a resources in just one line, like the example below:

```php

use Illuminate\Support\Facades\Route;


Route::apiResources([
     'client' => 'ClientController',
     'client.contact' => 'ContactController',
])->whereAll('[0-9]+');
```
But I don't want to implement a code for nothing, so I want to know your opinion if I can create another pull request for this function or if this will not be accepted from the concept?


src/Illuminate/Routing/PendingResourceRegistration.php
```diff
@@ -17,11 +17,11 @@ class PendingResourceRegistration
...
 
     /**
-     * The resource name.
+     * The array of resources containing the name => controller.
      *
-     * @var string
+     * @var array
      */
-    protected $name;
+    protected $resources;
```
driesvints, line 24: This is a breaking change.

src/Illuminate/Routing/PendingResourceRegistration.php
```diff
@@ -48,17 +48,15 @@ class PendingResourceRegistration
...
      *
      * @param  \Illuminate\Routing\ResourceRegistrar  $registrar
-     * @param  string  $name
-     * @param  string  $controller
+     * @param  array  $resources
      * @param  array  $options
      * @return void
      */
-    public function __construct(ResourceRegistrar $registrar, $name, $controller, array $options)
+    public function __construct(ResourceRegistrar $registrar, array $resources, array $options)
```
driesvints, line 55: This is a breaking change as well unfortunately...

driesvints:
Can you please attempt this to master instead? Although I don't feel like we should change anything here.

PR #44847, opened 2022-11-05
[9.x] Remove `__sleep()` & `__wakeup()` from SerializesModels trait.

r-kujawa:
This PR removes support for the `__sleep()` & `__wakeup()` methods in the `SerializesModels` trait, since they have become irrelevant in Laravel 9.x as the minimum requirements are PHP 8.0. In this version of PHP the `__serialize()` method will be called instead of the `__sleep()` method and the `__unserialize()` method will be called instead of the `__wakeup()` method.

I believe the reason why both magic methods were still present in this trait was to keep supporting PHP 7.3 in Laravel 8.x.

[Documentation reference](https://www.php.net/manual/en/language.oop5.magic.php#object.serialize)

### DIFF
--- a/src/Illuminate/Queue/SerializesModels.php
+++ b/src/Illuminate/Queue/SerializesModels.php
@@ -10,44 +10,6 @@ trait SerializesModels
     use SerializesAndRestoresModelIdentifiers;
 
     /**
-     * Prepare the instance for serialization.
-     *
-     * @return array
-     */
-    public function __sleep()
-    {
-        $properties = (new ReflectionClass($this))->getProperties();
-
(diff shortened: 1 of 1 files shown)


adrianb93:
Really wish this was targeted at Laravel 10 given methods were removed. 🤦

driesvints:
@adrianb93 but these methods aren’t used anymore.

adrianb93:
Just a bit of a deployment panic. A package we use referenced it and some of our code also used this trait and had it's own `__sleep` method.

driesvints:
It was never meant for these methods to be overwritten. We can't protect against everything people do in userland.

PR #44848, opened 2022-11-05
[9.x] Prevent model serialization on sync connection.

r-kujawa:
It prevents additional queries from being made by **not** serializing models that are meant to be executed synchronously.

It might not be a big optimization but it can help reduce the request execution time, especially when the properties are `QueueableCollection`s or `QueueableEntity`s that have many loaded relations.

### DIFF
--- a/src/Illuminate/Queue/SerializesAndRestoresModelIdentifiers.php
+++ b/src/Illuminate/Queue/SerializesAndRestoresModelIdentifiers.php
@@ -19,6 +19,10 @@ trait SerializesAndRestoresModelIdentifiers
      */
     protected function getSerializedPropertyValue($value)
     {
+        if (isset($this->connection) && $this->connection === 'sync') {
+            return $value;
+        }
+
         if ($value instanceof QueueableCollection) {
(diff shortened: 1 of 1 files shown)


axlon:
This could introduce subtle bugs into applications as sync queue is often used when testing jobs.

r-kujawa:
> This could introduce subtle bugs into applications as sync queue is often used when testing jobs.

I understand your concern, one of the reasons I am proposing this change is to reduce execution time of tests. For one test the difference is irrelevant, but if you have thousands of tests that send models to different queues it could make slight difference in your pipeline.

We use the sync queue in our testing jobs to allow them to be tested, not to make sure that the sync connection serializes and deserializales our models correctly. We should trust that is a framework feature that has already been tested.

Maybe I am mistaken, but shouldn't the model/collection being passed into the job be exactly the same as the one that was loaded miliseconds before in the same request lifecycle? So why reload it along with all its relationships?

taylorotwell:
This is to ensure that queues behave exactly the same.

PR #44850, opened 2022-11-05
[9.x] Introducing `Gate::fake()`

CWDN:
This adds the ability to allow you to fake out calls to the `Gate` so that you can easily bypass authorisation checks in your tests.

# Problem
Over the last months we have been moving our self made authorisation system across to use Laravel Gates. While going through our existing code and updating we found that being able to mock/ignore certain gates didn't have the cleanest experience. 

For example if we have a Gate defined for checking user has access to an account and we want to ignore it then we would have to do the below.

```php
Gate::before(fn (User $user, $ability) => $ability !== 'canAccessAccount');
```  

# Proposal

To hopefully improve developer's experience and give developers control and assertions around their Gates in their tests this  implements a Gate fake object into Laravel.

To fake out the Gates you run a similar function to all the other fakes within the system `Gate::fake()`. This swaps the default implementation with the fake one. From there all checks to any gate will return true as the response. 

## Failing a Gate

To fail a gate you simply call `fail()` on the Gate facade after you have called `fake()`. This by default will return false as the response to the Gate.

### Possible responses

You can provide a few different responses to the fail method on how you want your gate to fail. These are a Boolean, Response or a Sequence of the previous two.

```php
Gate::fake();

// Boolean
Gate::fail('canAccessAccount', false);

// Response
Gate::fail('canAccessAccount', Response::denyAsNotFound());

// Sequence
Gate::fail('canAccessAccount', new Sequence([false, Response::denyAsNotFound(), true));
```

## Failing a Policy

Very similar to failing a Gate you can perform all the same actions for a Policy. However you need to provide the Policy and the "ability" you want to fake. e.g.

```php
Gate::fake();

Gate::fail(UserPolicy, 'update', false);
```

## Assertions

We've added a couple of assertions to start this off but this is by no means an extensive list of assertions. We kept this simple so that we can organically grow what assertions are needed by the community. 

### `assertChecked`

This allows you to assert that the give gate or policy ability has been ran, you can also provide a closure to this function to perform a more detail assertion of the checks that have been made. Very similarly behaviour to the `assertDispatched` on the BusFake, this is called for every check that has been made on the gate/policy ability. The arguments passed into the closure are the same arguments that were passed into the `Gate::allows`/`Gate::denies`...etc call.

```php
Gate::fake();

// Run user code

Gate::assertChecked('accessAccount');

Gate::assertChecked('accessAccount', function (Account $account) {
    return $account->name === 'Example';
});
```

### `assertCheckedTimes`

Again similar to the Bus fake's `assertDispatchedTimes` this behaves like `assertChecked` however allows you to specify the number of times that check should of been called. 

## Calling the original method

We found In some instances you still want to execute the original gate/policy ability as it might trigger side effects or perform some DB check that you were relying on. In this use case we have provided a method to execute the original method as well as allowing a custom response.

```php
Gate::checkOriginalGate();
```

## Not faking certain gates

Final feature in this PR is the ability to just allow certain gates/abilities to execute as normal with no modification. To do this you call `except` on the Gate fake to ignore certain methods.

```php
Gate::fake()->except('accessAcount`);
```

## Other Notes

As part of the faking logic we've introduced a PolicyFake object, this is a decorated around the Developer's original policy and helps track any calls made to the specific policy. However we hit against a problem that because we are using `__call` to forward on the call to the original policy Laravel was unable to determine the parameters the faked policy abilities had. To get around the Reflection parameter checking on Policy functions we've implemented a new interface that requires you to return all the abilities that a policy requires. These "abilities" are just an array of `ReflectionFunctionAbstract`'s that Laravel can then use to determine if the ability can be ran as a guest and has a user parameter.

I felt this was a clean way to get around this problem and also provided developers was for them to provide abilities for their policies where they might being doing similar things.

Any questions or feedback is greatly appreciated. Thanks.

### DIFF
--- a/src/Illuminate/Auth/Access/DefinesAbilities.php
+++ b/src/Illuminate/Auth/Access/DefinesAbilities.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Illuminate\Auth\Access;
+
+interface DefinesAbilities
+{
+    /**
+     * @return array<string, callable>
+     */
+    public function abilities();
+}
--- a/src/Illuminate/Auth/Access/Gate.php
+++ b/src/Illuminate/Auth/Access/Gate.php
@@ -12,8 +12,10 @@ use Illuminate\Support\Arr;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Str;
 use InvalidArgumentException;
+use Laravel\SerializableClosure\Support\ReflectionClosure;
 use ReflectionClass;
 use ReflectionFunction;
+use ReflectionFunctionAbstract;
 
 class Gate implements GateContract
 {
@@ -479,7 +481,22 @@ class Gate implements GateContract
         try {
             $reflection = new ReflectionClass($class);
 
-            $method = $reflection->getMethod($method);
+            if ($reflection->implementsInterface(DefinesAbilities::class)) {
+                $object = is_string($class) ? $this->container->make($class) : $class;
+                $abilities = $object->abilities();
+
+                if (isset($abilities[$method])) {
+                    $method = $abilities[$method];
+
+                    if ($method instanceof Closure) {
+                        $method = new ReflectionClosure($method);
+                    }
+                }
+            }
+
+            if (! $method instanceof ReflectionFunctionAbstract) {
+                $method = $reflection->getMethod($method);
+            }
         } catch (Exception $e) {
             return false;
         }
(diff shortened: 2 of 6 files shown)


taylorotwell:
Honestly I recommend just creating a user that passes / fails certain gates. You can even have factory methods to create these types of situations if you need to.

PR #44863, opened 2022-11-07
feat: add support for nullable passwords for users

owenvoke:
This adds support for passwords that are `null` in the database. We are currently doing this, however it results in `strlen()` deprecation notices when the hash is checked.

If this isn't the correct way to handle this, and Laravel isn't intended to have nullable passwords, we can just overload `getAuthPassword()` on our user class. 👍🏻

### DIFF
--- a/src/Illuminate/Auth/Authenticatable.php
+++ b/src/Illuminate/Auth/Authenticatable.php
@@ -44,7 +44,7 @@ trait Authenticatable
     /**
      * Get the password for the user.
      *
-     * @return string
+     * @return string|null
      */
     public function getAuthPassword()
     {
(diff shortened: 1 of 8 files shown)


lk77:
I don't see any reason to have nullable passwords

olivernybroe:
@ikari7789 many systems allows signing up and logging in with Google. When you do that you don't need a password, hence the password can be nullable in those situations.

lk77:
@olivernybroe but in that case you will not validate user crendentials ?

olivernybroe:
@lk77 if your user which only have Google login hence password being nullable tries to login with email/password combination because they forgot that they have not set a password yet.

driesvints:
Why don't you just generate a password for every user?

olivernybroe:
@driesvints generating a password means we are adding another way to login for a user that doesn't need it. 
Adding unneeded logic for making secure password generation seems a waste when that logic is not needed.

olivernybroe:
@lk77 true, that might make more sense to reduce the extra time spend on hashing. 

But seems like people didn't really agree. So we are closing and either just living with an extra query to fetch the user first and check if password is nullable or extending the guard to do the early exit in our own codebase.

Issue #44849, opened 2022-11-05
Hard coupling of Console (and indirectly Eloquent) on the framework


wouterj:
- Laravel Version: 9.21.0+
- PHP Version: n/a (but for the sake of completness: 8.1.12)
- Database Driver & Version: n/a

### Description:

As you might know, I'm maintaining (on low pace) an Eloquent bundle for Symfony. The recent releases (Laravel 7/8/9) have been an incredibly smooth ride, thanks for your hard work!
However, the console output restyling (https://github.com/laravel/framework/pull/43065) introduced the use of the `resolve()` function in the Console component, creating a hard dependency between the component and the framework.

This hard dependency has been identified before, as it also broke Lumen. Since https://github.com/laravel/framework/pull/43312 , it uses the `app()` function. While no longer having a hard dependency on the full framework, this still makes it impossible to use the Console  component (and indirectly Eloquent's `Migrator` class) outside of the framework.

Relevant code: https://github.com/laravel/framework/blob/c76c3bdddf3bac41f92bbcaddb50d1053d0011f2/src/Illuminate/Console/View/Components/Component.php#L81-L94

My bundle already has some plumbing to work around [`Seeder::resolve()`](https://github.com/wouterj/WouterJEloquentBundle/blob/10bbbad94c6eafa3004db08dee6d4b332f75e0f8/src/Seeder.php#L62-L78). I'm not too happy about it, but it is acceptable given creating decoupled components isn't your top priority (and I'm probably one of a few weirdo's attempting this). However, this newly used function makes it a lot harder to work around. I currently believe the best solution is to fully reimplement all illuminate/console's components to make this work: https://github.com/wouterj/WouterJEloquentBundle/commit/b2276e9687f1bba5a8a17e73ef7e04f4daf7a018 This is very fragile, given it breaks whenever you introduce a new cool component.

Is this coupling expected (and should I perhaps reconsider my bundle)? Or can there be something done in the console package to make it possible to work around this (without the need to reimplement critical components like `Task`)?
I don't know much about Laravel's internals, but I'm happy to work together with you to make sure Eloquent remains usable outside the Laravel framework if you wish.

henzeb:
I hook in on this as I have a package that may get affected by this if anything gets rewritten as I have some plumbing in there as well. 
https://packagist.org/packages/henzeb/laravel-console-facade

driesvints:
[...]

PR #44864, opened 2022-11-07
[9.x] De-couple Console component from framework

driesvints:
This PR decouples the Console component further from the framework. https://github.com/laravel/framework/pull/43065 introduced the concept of mutators which are getting resolved through the container before they're invoked. But since none of these mutators have dependencies this isn't necessary. 

I've also added a check for the `app` function in the `EnsureRelativePaths` mutator so it's not invoked when called from outside the framework.

See https://github.com/laravel/framework/issues/44849


wouterj:
Thanks for the quick actions! I've applied this patch locally and I can confirm that this fixes my use-case :+1:

nunomaduro (approved):

PR #44865, opened 2022-11-07
[9.x] Fix Collection::combine return docblock

Josh-G:
Previously, the combine docblock always used the template type of the first collections keys. It should be `TValue` since we're [combining the values of both collections](https://laravel.com/docs/9.x/collections#method-combine).

Added a type assertion as well, without this change it fails with
```
  :605   Expected type Illuminate\Support\Collection<string, string>, actual: Illuminate\Support\Collection<int, string>  
```


nunomaduro:
Thank you!

PR #44875, opened 2022-11-08
Add helper methods to `Schedule` to test against jobs and commands

orkhanahmadov:
This PR adds 2 public methods to `Illuminate\Console\Scheduling\Schedule` to test if a given job/command with a given expression is scheduled. Comes in handy on unit tests. 

Assuming we have a job `CheckProductAvailability` which we schedule to run daily.

```php
$schedule->job(CheckProductAvailability::class)->daily();
```

In units tested with the new `hasJob` method, we can test against this scheduled job.

```php
$this->assertTrue(
  app(Schedule::class)->hasJob(CheckProductAvailability::class, '0 0 * * *')
);
```

Testing against command is also easy:

```php
$schedule->command('foo:bar')->daily();

// test case
$this->assertTrue(
  app(Schedule::class)->hasCommand('foo:bar', '0 0 * * *')
);
```

`Schedule` is macroable, we can also add these methods as local macros, but I think makes sense to have them on the class directly, makes it easier to test


taylorotwell:

PR #44876, opened 2022-11-08
[9.x] Refactor for missing attributes

bert-w:
The changes below allow developers to use the traits `HasAttributes`, `RestrictsAttributes`, `UsesTimestamps` themselves in custom classes, therefore creating a model with a subset of components. For example, I am using this in my own project to create a `Country` model disconnected from a database (the country attributes are loaded in-code since it never changes in my situation) while still allowing all the handy stuff from `HasAttributes` like mutators.

- Moved functionality of https://github.com/laravel/framework/pull/44283 (Opt-in Model::preventAccessingMissingAttributes() option) to a new trait `RestrictsAttributes` instead of it being spread out in `HasAttributes` and `Illuminate\Database\Eloquent\Model`. This increases modularity greatly and allows developers to use these traits themselves.

- Changed `HasAttributes` trait so it has no dependencies on the `RestrictsAttributes` trait (that is: functions from this trait have to be checked first using `method_exists()` before being called. Not having the `RestrictsAttributes`trait thus reverts the behaviour to before PR #44283 ).

- Changed `HasAttributes` trait to have no dependencies on the `UsesTimestamps` trait.

- Moved `__get()` and `__set()` from `Eloquent\Model` to the very top of the `HasAttributes` trait since it calls exactly those functions contained in that trait.

- Changed the order of the if-statement (`a && b && c`) in `throwMissingAttributeExceptionIfApplicable()` so it doesn't break on the calls `$this->exists && !$this->wasRecentlyMerged` since these properties are not part of the trait (and therefore may be undefined depending on where you include this trait). Now, the `self::preventsAccessingMissingAttributes()` is called first which by default returns `false` therefore ending the if-statement. This only increases stability since `a && b && c` can be rearranged in any way, but in this situation there is 1 case that is better being called first.

- Included tests to show that these traits can be used mostly independently. There are still some dependencies from `HasAttributes` to `HasRelations` left in the code (and probably some other traits as well) but I have not included those changes in this PR.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -451,7 +451,7 @@ trait HasAttributes
         // Here we will determine if the model base class itself contains this given key
         // since we don't want to treat any of those methods as relationships because
         // they are all intended as helper methods and none of these are relations.
-        if (method_exists($this, $key)) {
+        if (method_exists(self::class, $key)) {
             return $throwMissingAttributeExceptionIfApplicable($key);
         }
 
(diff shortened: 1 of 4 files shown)


taylorotwell:

PR #44885, opened 2022-11-09
Str startsWith and endsWith with case insensitive param

ricventu:
I added param to make Str::startsWith and Str::endsWith case insensitive, same like Str::contains

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -1087,15 +1087,23 @@ class Str
      *
      * @param  string  $haystack
      * @param  string|iterable<string>  $needles
+     * @param  bool  $ignoreCase
      * @return bool
      */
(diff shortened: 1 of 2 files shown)


ricventu:
tests / PHP 8.0 - prefer-lowest (pull_request)
fails but its an error in docker workflow, not related to the code

taylorotwell:
I think you could just strtolower your parameters before calling this method.

ricventu:
sure,
but therefore also Str::contains should not have the $ignoreCase parameter

Thanks for your attention )

caendesilva:
> I think you could just strtolower your parameters before calling this method.

@taylorotwell Would you be open to reconsidering this? `Str::contains` supports this parameter even though you can call strtolower beforehand, so merging this PR would both add value and consistency.

Issue #44877, opened 2022-11-08
CommandStarting event input yields incorrect values for getArgument.


mad-briller:
- Laravel Version: 9.39.0
- PHP Version: 8.1.11

### Description:

When listening for the `CommandStarting` event and reading the input using the `$event->input->getArgument` method, the returned values are incorrect and off by one.
it seems the input considers the command name as the first positional argument.

### Steps To Reproduce:

setup a reproduction here:

https://github.com/mad-briller/laravel-bug/commit/0883a9dcf3c8cc840af90cb103b0c965ccefe4d7

driesvints:
Thanks for your report. We've decided to revert the PR that introduced these arguments in the first place because it's too brittle. Please see the full reasoning here: https://github.com/laravel/framework/pull/44888

PR #44888, opened 2022-11-09
[9.x] Remove argument assignment for console

driesvints:
This PR reverts https://github.com/laravel/framework/pull/44662 by @bert-w. It seems this doesn't exactly works as expected because incorrect arguments are assigned: https://github.com/laravel/framework/issues/44877. I spent a great deal today looking into a fix but always managed to run into failing integration tests. Since this part of Laravel is too brittle it seemed to me to be the best choice to revert the PR to get back to the state it was instead of having a half baked implementation.

If we ever want to pursue this we'll need to make sure that the arguments of the input doesn't contains the command name as the issue in https://github.com/laravel/framework/issues/44877 describes.


bert-w:
Ok reverting is the best option then. I'll take a look again when I have time; i agree that some of the current implementation seems a bit brittle. So does some of the code in Symfony because they also seem to just ignore any exception following from `try { $input->bind(...) } catch (ExceptionInterface) {...}`, on multiple occasions in their code.

PR #44893, opened 2022-11-10
Loops through stack to determine the token for request.

SameOldNick:
Previously, it was a bunch of if statements. This is simply a refactor (does the same thing, but in less code).


SameOldNick:
Also adds getToken method to Request instances. Getting the token from the request (query, body, and headers) is universal and may very well be needed elsewhere (not just the TokenGuard instance). It would be good if this was available (and didn't depend on the 'token' guard being set) but if not, developers could just add it as a macro.

PR #44904, opened 2022-11-11
[9.x] Adds new ucwords method to the Str and Stringable classes

LNCH:
This PR adds a new `ucwords` method to the Str helper function, and also a chainable `ucwords()` method to the Stringable class.

Rather than utilising the PHP core `ucwords` function, I've chosen to hook in to the existing `ucfirst` method, which takes into account other characters such as the Cyrillic characters, meaning we're able to offer the same functionality for more languages, in the same way Laravel already does.

I was recently working on a personal project, and I found myself wanting to chain a ucwords method to the end of a string manipulation. Obviously this can be achieved by wrapping the existing code within the PHP function, but this PR will allow us to maintain the chainable nature of the new Stringables available in Laravel.

This should allow users to utilise the functionality of ucwords now while keeping the fluent syntax Laravel is known for.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -1212,6 +1212,17 @@ class Str
     }
 
     /**
+     * Makes the first character of each word in a string uppercase.
+     *
+     * @param $string
+     * @return string
+     */
+    public static function ucwords($string)
+    {
+        return collect(explode(' ', $string))->map(fn ($word) => static::ucfirst($word))->join(' ');
+    }
+
+    /**
      * Get the number of words a string contains.
      *
      * @param  string  $string
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -1214,7 +1214,7 @@ class Str
     /**
      * Makes the first character of each word in a string uppercase.
      *
-     * @param $string
+     * @param  string  $string
      * @return string
      */
     public static function ucwords($string)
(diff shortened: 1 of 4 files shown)


src/Illuminate/Support/Str.php
```diff
@@ -1211,6 +1211,17 @@ public static function ucsplit($string)
         return preg_split('/(?=\p{Lu})/u', $string, -1, PREG_SPLIT_NO_EMPTY);
     }
 
+    /**
+     * Makes the first character of each word in a string uppercase.
+     *
+     * @param $string
```
driesvints, line 1217:
```suggestion
     * @param  string  $string
```

src/Illuminate/Support/Str.php
```diff
@@ -1211,6 +1211,17 @@ public static function ucsplit($string)
...
 
+    /**
+     * Makes the first character of each word in a string uppercase.
+     *
+     * @param  string  $string
+     * @return string
+     */
+    public static function ucwords($string)
+    {
+        return collect(explode(' ', $string))->map(fn ($word) => static::ucfirst($word))->join(' ');
```
vdauchy, line 1222: Using native functions `explode/array_map/implode` directly might be slightly more performant without loosing much readability.

dennisprudlo:
Is there any reason not to use `Str::title('laravel framework')`?

LNCH:
> Is there any reason not to use `Str::title('laravel framework')`?

Nope, only that I had completely missed this when I was looking for it. That would work fine, I guess this would just provide an alternative but realistically this PR can probably be closed

PR #44905, opened 2022-11-11
[9.x] Added `Auth::userOrFail()` function

kw-pr:
I propose to have a `Auth::userOrFail()` method. Returning only the current user but never not null. Instead it throws a UnauthorizedException.

With `Auth::user()` you always have to check if there is a user. In every route and where every you use it. Often this is already checked by a guarded route. This can be solved by userOrFail() similar to firstOrFail() an other ...OrFail() methodes we already have.

Please let me now if this PR needs improvement.


taylorotwell:
Auth::user() ?? throw new Exception

PR #44906, opened 2022-11-11
[9.x] Add zero-width space to trimmed characters in TrimStrings middleware

tomcodes:
I had a weird bug recently when adding a job to the queue when the content had a zero-width space present somewhere.

The bug I had was: `Unable to JSON encode payload. Error code: 5`

It took me a while to corner this issue to a zero-width space present in a user firstname.

I was wondering also if would be interesting to add a middleware to actually remove those characters also from within the fields instead of just the side. I am willing to create another pull request if that need can benefit more people.

Github does not show the diff, so here is a visual diff:

**Before**  
image

**After**  
image

### DIFF
--- a/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
+++ b/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
@@ -53,7 +53,7 @@ class TrimStrings extends TransformsRequest
             return $value;
         }
 
-        return preg_replace('~^[\s﻿]+|[\s﻿]+$~u', '', $value) ?? trim($value);
+        return preg_replace('~^[\s﻿​]+|[\s﻿​]+$~u', '', $value) ?? trim($value);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tomcodes:
Also, I can try and add a test for this if necessary.

taylorotwell:
Yeah we would definitely want a test for this I think.

taylorotwell:
Please mark as ready for review when it has a test.

tomcodes:
I tried to think of multiple situations with the tests, I hope it's fine like this!

tomcodes:
woops, I have noticed a tiny typo issue in the last test, should I open a new pull request?

The title should not be `This title contains a zero-width non-breakable space at the end`, it should be `This title contains a combination of zero-width non-breakable space and zero-widh spaces characters at the begining and the end` instead.

The test is still valid, it's just the title that does not fit the test name.

driesvints:
@tomcodes yes feel free to send in a new pr

PR #44911, opened 2022-11-11
[9.x] add url_protocol validation rule

Ahmad-Mohammad-Kouja:
this rule validates that the value is a URL
and its protocol is one of the provided protocols
helping the user to accept URLs with specific protocols like accepting only secure URLs
or HTTP/HTTPS only URLs


Ahmad-Mohammad-Kouja:
> This PR introduces a new validation rule "url_protocol" that helps the user determine which protocols for URLs want to accept.

> 
> Example:
> 
> ```
> 'url' => ['required', 'url_protocol:https'];
> ```

> https://laravel.com //pass
> http://laravel.com //not pass

dennisprudlo:
Couldn't you just use the `starts_with:http,https`?

This PR checks for a "valid" scheme but why should the validation rule throw an exception? Someone may use custom protocols and this rule wouldn't allow it.

taylorotwell:

PR #44915, opened 2022-11-12
[9.x] Extract child instructions

maoleng:
We can extract the statement here (clearer intention, lower complexity numbers)


driesvints:
Thanks!

PR #44918, opened 2022-11-12
[9.x] Add `uppercase` validation rule

michaelnabil230:
This PR introduces an "uppercase" validation rule.

We have a current want for this rule to create some standardized data formatting without changing the case on the user silently.

For example, we want a user to be able to create a database user, but we want to force the user to uppercase.

Usage:

```php
Validator::make(['name' => 'ADMIN'], ['name' => ['required', 'string', 'uppercase']);
```

I don't have a real use case for an "uppercase" rule, so I did not include it. If we want it for consistency, I can follow up with one (or someone else is welcome to add it).

Documentation: https://github.com/laravel/docs/pull/8336

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1172,6 +1172,19 @@ trait ValidatesAttributes
     }
 
     /**
+     * Validate that an attribute is uppercase.
+     *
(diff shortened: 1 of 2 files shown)


MaRmARk0:
I recently wanted to validate whether incoming language identifier is an uppercase string, e.g. "SK", "CZ", "EN" etc. so this comes handy now.

PR #44919, opened 2022-11-12
[9.x] Move function `withoutTrashed` into `DatabaseRule`

michaelnabil230:
ُThis PR move `withoutTrashed` into `DatabaseRule` makes more every sense

### DIFF
--- a/src/Illuminate/Validation/Rules/DatabaseRule.php
+++ b/src/Illuminate/Validation/Rules/DatabaseRule.php
@@ -170,6 +170,19 @@ trait DatabaseRule
     }
 
     /**
+     * Ignore soft deleted models during the existence check.
+     *
+     * @param  string  $deletedAtColumn
+     * @return $this
+     */
+    public function withoutTrashed($deletedAtColumn = 'deleted_at')
+    {
+        $this->whereNull($deletedAtColumn);
+
+        return $this;
+    }
+
+    /**
      * Register a custom query callback.
      *
      * @param  \Closure  $callback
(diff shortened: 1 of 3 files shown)


src/Illuminate/Validation/Rules/DatabaseRule.php
```diff
@@ -169,6 +169,19 @@ public function whereNotIn($column, $values)
         });
     }
 
+    /**
+     * Ignore soft deleted models during the existence check.
```
amims71, line 173: I believe, it would work with unique check as well with the existence check.

driesvints:
This is unfortunately a breaking change for third party packages that use these traits separately.

michaelnabil230:
But there will be no problem with this change.

michaelnabil230:
@driesvints 
But there will be no problem with this change.

michaelnabil230:
@BrandonSurowiec This is what you said to me @driesvints, what do you think?

driesvints:
People using the trait directly will see a method suddenly missing from the Exists rule. This is a breaking change.

driesvints:
@michaelnabil230 feel free to re-attempt this

michaelnabil230:
@driesvints It seems to me that you didn't see the all changes.
I merged `withoutTrashed' into '‏DatabaseRule'

michaelnabil230:
@driesvints 
@BrandonSurowiec 

The new PR https://github.com/laravel/framework/pull/44938

PR #44920, opened 2022-11-12
Convert closures to arrow functions

michaelnabil230:


driesvints:
This (although probably non-breaking) changes the behavior with returning the query adjustment from the short closure.

PR #44921, opened 2022-11-12
[9.x] improve model functions `findOrFail` and `firstOrFail`

michaelnabil230:


driesvints:
If nothing's broken it's best to leave things be.

PR #44027, opened 2022-09-06
[9.x] Don't use locks for queue job popping for PlanetScale's MySQL-compatible Vitess engine

rihardsgrislis:
PlanetScale provides great serverless MySQL-compatible database and guide on how to integrate it in Laravel (https://planetscale.com/docs/tutorials/connect-laravel-app) but under hood uses Vitess.

However it looks that Vitess doesn't support "skip" queries:
```SQLSTATE[HY000]: General error: 1105 syntax error at position 185 near 'SKIP' (SQL: select * from `jobs` where `queue` = default and ((`reserved_at` is null and `available_at` <= 1662479913) or (`reserved_at` <= 1662479823)) order by `id` asc limit 1 FOR UPDATE SKIP LOCKED)```

Following https://github.com/laravel/framework/issues/31536 I've added a check for Vitess engine/version parsing so it wouldn't use locks for popping:

```
>>> DB::connection()->getPdo()->getAttribute(PDO::ATTR_DRIVER_NAME)
=> "mysql"

>>> DB::connection()->getPdo()->getAttribute(PDO::ATTR_SERVER_VERSION)
=> "8.0.23-vitess"
```

PR #44925, opened 2022-11-13
[9.x] Update database version check for lock popping for PlanetScale

narciero:
PlanetScale (Vitess) does not support `SKIP LOCKED`.

This issue was addressed previously in #44027 however PlanetScale seems to have changed their version string suffix from `-vitess` to `-PlanetScale`.

```
> DB::connection()->getPdo()->getAttribute(PDO::ATTR_SERVER_VERSION)
= "8.0.23-PlanetScale"
```

### DIFF
--- a/src/Illuminate/Queue/DatabaseQueue.php
+++ b/src/Illuminate/Queue/DatabaseQueue.php
@@ -261,7 +261,7 @@ class DatabaseQueue extends Queue implements QueueContract, ClearableQueue
         if (Str::of($databaseVersion)->contains('MariaDB')) {
             $databaseEngine = 'mariadb';
             $databaseVersion = Str::before(Str::after($databaseVersion, '5.5.5-'), '-');
-        } elseif (Str::of($databaseVersion)->contains('vitess') || Str::of($databaseVersion)->contains('PlanetScale')) {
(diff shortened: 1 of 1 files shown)


src/Illuminate/Queue/DatabaseQueue.php
```diff
@@ -261,7 +261,7 @@ protected function getLockForPopping()
         if (Str::of($databaseVersion)->contains('MariaDB')) {
             $databaseEngine = 'mariadb';
             $databaseVersion = Str::before(Str::after($databaseVersion, '5.5.5-'), '-');
-        } elseif (Str::of($databaseVersion)->contains('vitess')) {
+        } elseif (Str::of($databaseVersion)->contains('vitess') || Str::of($databaseVersion)->contains('PlanetScale')) {
```
fragkp, line 264: You could use to simplify:
```php
Str::of($databaseVersion)->contains(['vitess', 'PlanetScale])
```

Docs: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Support/Stringable.php#L170

narciero: Nice, didn't know you could pass an array there...fixed!

fragkp (commented):
👍

PR #44926, opened 2022-11-13
[9.x] Add `assertRedirectToRoute` to `TestResponse`

zaherg:
`Illuminate\Testing\TestResponse` has several helpers to assert the response has a redirect of various types. There is `assertRedirect(string $uri = null)`, `assertRedirectContains(string $uri)`, and `assertRedirectToSignedRoute(string $name = null, array $parameters = [])`.

This PR add a missing assertion `assertRedirectToRoute`, which in theory does not differ much from `assertRedirectToSignedRoute`.

Used as the following:

```php
$this->get('test-route')
    ->assertRedirectToRoute('named-route');
```

and in case of passing parameters we use it as the following:

```php
$this->get('test-route')
    ->assertRedirectToRoute('named-route-with-param', 'hello');

$this->get('test-route-with-extra-param')
    ->assertRedirectToRoute('named-route-with-param', [
        'param' => 'foo',
        'extra' => 'another',
    ]);
```

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -291,6 +291,33 @@ class TestResponse implements ArrayAccess
     }
 
     /**
+     * Assert whether the response is redirecting to a given route.
+     *
+     * @param  string|null  $name
+     * @param  mixed  $parameters
+     * @return $this
+     */
+    public function assertRedirectToRoute($name = null, $parameters = [])
+    {
+        if (! is_null($name)) {
+            $uri = route($name, $parameters);
+        }
+
+        PHPUnit::assertTrue(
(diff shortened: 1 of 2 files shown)


gms8994:
The `assertRedirectToRoute` function suggests that it can take a `null` value, but passing in a null value throws an error:

See this test case for PoC:

```
public function testAssertRedirectToRouteWithoutRouteName()
{
    $this->router->get('test-route', function () {
        return new RedirectResponse($this->urlGenerator->route('named-route'));
    });

    $this->get('test-route')
        ->assertRedirectToRoute();
}
```

zaherg:
I'll push new code in a few min.

thanks @gms8994 for catching this.

PR #32267, opened 2020-04-07
[7.x] Updated observer.stub to handle all model events

berdjabrahamian:
Updated the observer.stub that is created when a model class is provided to handle all model events

- retrieved (new)
- creating (new)
- created
- updating (new)
- updated
- saving (new)
- saved (new)
- deleting (new)
- deleted
- restoring (new)
- restored
- force deleted

If the user is already in a position to creating a observer for there model class, they should have access to all model events within the observer, this will stop the user adding them manually later 
- especially for new users, who aren't aware what events are available for them, or how to go about it.

PR #29489, opened 2019-08-10
[5.8] Update observer.stub to handle all events

iamoat:
Update observer.stub to handle all events
- retrieved
- creating
- created
- updating
- updated
- saving
- saved
- deleting
- deleted
- restoring
- restored
- force deleted

PR #44928, opened 2022-11-13
Change observer.stub to include all possible event methods + sort methods alphabetically

robertgdev:
Just a short patch for observer.stub to

1. Include all possible methods
2. Sort methods alphabetically 

I recently generated an observer and was surprised that it did not contain all possible methods. This patch fixes this.


Rizky92:
See #32267, #29489

driesvints:
Please see the PR's from above.

PR #44933, opened 2022-11-14
Improve setAttribute Mutators return consistency

rhurling:
This change returns the current object from `setAttributeMarkedMutatedAttributeValue` which allows using `$model->setAttribute('key', 'value')->save()` when using Attribute Mutators. As far as I can tell the setAttribute is otherwise intended to return `$this` and only returns `mixed` (in phpdoc) since it is unknown what is returned in a custom mutator function via `setMutatedAttributeValue`.

Since `setAttributeMarkedMutatedAttributeValue` currently returns `void` this change should have no impact on anything else.


taylorotwell:
I don't think I'll change this behavior for now. You can use `tap`.

PR #44934, opened 2022-11-14
Added missing docblock params and parameter types to Str::slug and Str::wrap

mirzazeyrek:
I've noticed Str::slug method is updated recently and the custom implementation that I was using started to give extra warnings on Larastan. So I decided to check the class and a small fix for missing docblock params and added types to parameters.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -1012,9 +1012,10 @@ class Str
      * @param  string  $title
      * @param  string  $separator
      * @param  string|null  $language
+     * @param  array<string, string> $dictionary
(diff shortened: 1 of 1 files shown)


taylorotwell:
Please just update the docblocks.

mirzazeyrek:
I've sent another PR with only docblock changes. It looks like docblocks are not matching with test expectations when implementing strict types. :thinking: 

`    public function testAsciiNull()
    {
        $this->assertSame('', Str::ascii(null));
        $this->assertTrue(Str::isAscii(null));
        $this->assertSame('', Str::slug(null));
    }
`

PR #44937, opened 2022-11-14
[10.x] Update SQL Server to FETCH and OFFSET for queries that do not include an order by

dunhamjared:
## Overview

SQL Server 2012 introduced the fetch and offset feature.

Laravel started using FETCH and OFFSET in 2021 -- but only when you include an order by. 
See PR: https://github.com/laravel/framework/pull/39863

Laravel also only officially supports 2017 and up: 
https://laravel.com/docs/9.x/database#introduction

This PR updates the query builder to also use FETCH and OFFSET for queries that do not include an order by.

Others have voiced support for this PR: 
https://www.reddit.com/r/laravel/comments/ypooqe

This update improves the speed by 33% and requires less execution steps!
https://gist.github.com/dunhamjared/cb40bbf294ed6f8e48bd60010a31b4f3

## Changes

- Updated `row_number()` logic to match the current use of `OFFSET/FETCH`
- Consolidated the `OFFSET/FETCH` logic into built in functions
- Ordered the `$selectComponents` in the correct order for SQL Server

## Example

Currently, if you include an order by, it uses offset and fetch:

```php
$builder->select('*')->from('users')->skip(11)->take(10)->orderBy('email', 'desc');
```

```sql
select * from [users] order by [email] desc offset 11 rows fetch next 10 rows only
```

However, if order by is not included, it falls back to the old method of offsetting the results:

```php
$builder->select('*')->from('users')->skip(10)->take(10);
```

```sql
select * from (select *, row_number() over (order by (select 0)) as row_num from [users]) as temp_table where row_num between 11 and 20 order by row_num
```

This PR would instead produce:

```sql
select * from [users] order by (SELECT 0) offset 10 rows fetch next 10 rows only
```

## Edge cases 

- There could be people that actually use the `row_num` in their applications when order by is not used, so this could be considered a breaking change.

### DIFF
--- a/src/Illuminate/Database/Query/Grammars/SqlServerGrammar.php
+++ b/src/Illuminate/Database/Query/Grammars/SqlServerGrammar.php
@@ -47,7 +47,7 @@ class SqlServerGrammar extends Grammar
     public function compileSelect(Builder $query)
     {
         // An ORDER BY clause is required for offset to work
-        if($query->offset && empty($query->orders)){
+        if ($query->offset && empty($query->orders)) {
             $query->orders[] = ['sql' => '(SELECT 0)'];
         }
 
@@ -309,7 +309,7 @@ class SqlServerGrammar extends Grammar
      */
     protected function compileLimit(Builder $query, $limit)
     {
-        if($limit && $query->offset > 0){
+        if ($limit && $query->offset > 0) {
             return "fetch next {$limit} rows only";
         }
 
@@ -325,7 +325,7 @@ class SqlServerGrammar extends Grammar
      */
     protected function compileOffset(Builder $query, $offset)
     {
-        if($offset){
+        if ($offset) {
             return "offset {$offset} rows";
         }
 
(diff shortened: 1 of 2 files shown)


joelharkes:
@dunhamjared  this introduced a big bug!

Page 1 has no offset and thus is not explicitly sorted by column 0 and thus has another sort order, this must be fixed!!

dunhamjared:
Hi @joelharkes -- that sounds very odd, I haven't experienced this issue myself. I also tried quickly replicating the issue you described with no success. 

Could you create an issue for this and provide a query example?

joelharkes:
@dunhamjared 
The problem arrives when you don't use incrementing ID's


We use guids as primary key and generating the guids in laravel (creates alphabetically data time sorted guis).

But SQL server orders differently.

This means the inserted record might be on the last 'page' in the databse but when you sort on ID it is actually not the last record in the the list.

see also: https://stackoverflow.com/questions/7810602/sql-server-guid-sort-algorithm-why

dunhamjared:
Nice, glad to see it was a simple fix. :)

joelharkes:
@dunhamjared i broke distincts + limits but i can't get it to work, see: https://github.com/laravel/framework/pull/47766

Issue #36376, opened 2021-02-24
ArrayObject cast does not serialize empty array to empty object


danilopolani:
- Laravel Version: 8.29.0
- PHP Version: 8.0.2
- Database Driver & Version: N/A

### Description:
Using the `AsArrayObject` cast I expect an empty array to be JSON serialized into an empty object, like the native behaviour, but instead it serialize into an empty array.

For example with a given `Model` with the `AsArrayObject` cast to `field` attribute:

```php
json_encode(['field' => new ArrayObject([])]); // {"field":{}}
(new Model(['field' => []))->toJson(); // {"field":[]}
```

However passing MANUALLY the `ArrayObject` to wrap the field content, it works.

```php
(new Model(['field' => new ArrayObject([])))->toJson(); // {"field":{}}
```

### Steps To Reproduce:

> <small>A demo repository is available here: https://github.com/danilopolani/laravel-arrayobject. Just run it with `sail up` and go to `http://localhost` to see dumps.</small>

1. Create a Model assigning `AsArrayObject` cast to a field:

```php
<?php

namespace App\Models;

use Illuminate\Database\Eloquent\Casts\AsArrayObject;
use Illuminate\Database\Eloquent\Model as BaseModel;

class Example extends BaseModel
{
    /**
    * The attributes that are mass assignable.
    *
    * @var array
    */
    protected $fillable = [
        'content',
    ];

    /**
    * The attributes that should be cast to native types.
    *
    * @var array
    */
    protected $casts = [
        'content' => AsArrayObject::class,
    ];
}
```

2. Create a Model with some data and JSON-serialize it.

```php
$model = new Model([
    'content' => [],
]);

dd($model->json()); // {"content":[]}
```

browner12:
Is it as simple as adding a `JSON_FORCE_OBJECT` to the `json_encode` call here:

https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Eloquent/Casts/AsArrayObject.php#L26

danilopolani:
Ok, going in depth on this I may have found the issue: `\Illuminate\Database\Eloquent\Casts\ArrayObject::toArray()`.

It has been added (I guess) to let users do `$model->content->toArray()` as a shortcut to `$model->options->getArrayCopy()`, but it is invoked in the Jsonify too and what happens is that `json_encode` encodes a basic array and not a `ArrayObject`.

[...]

PR #44939, opened 2022-11-14
Eloquent Cast : Add new cast AsObject

squiaios:
Add new cast AsObject.

I got the same problem as described here https://github.com/laravel/framework/issues/36376, so I create a new cast class


taylorotwell:
I think you can just include this in your own project if you need it?

Headscracher:
@taylorotwell so you're concluding based on you're not using it, that this wouldn't be a valuable default?
I've ran into this exact same problem now multiple times and whilst I obviously can add that cast myself, I could also say that about any other feature the framework provides?

ncphillips:
I am also running into this issue

PR #44940, opened 2022-11-14
[9.x] Translate Based on Object Property

devajmeireles:
## What

This PR introduces the capability to pass an entire object to the translator to make translations based on property bind.

## Why

In some cases we need to translate something based on a property of an object, so it is easier to pass the object to the translator to be able to do the translation based on the property.

## Example:

```php
//before this PR

@lang('Welcome, :name!', ['name' => $user->name])
```
```php
//with the proposal of this PR

@lang('Welcome, :name!', $user)
```

### DIFF
--- a/src/Illuminate/Translation/Translator.php
+++ b/src/Illuminate/Translation/Translator.php
@@ -99,12 +99,12 @@ class Translator extends NamespacedItemResolver implements TranslatorContract
      * Get the translation for the given key.
      *
      * @param  string  $key
-     * @param  array  $replace
+     * @param  mixed  $replace
      * @param  string|null  $locale
      * @param  bool  $fallback
      * @return string|array
      */
-    public function get($key, array $replace = [], $locale = null, $fallback = true)
+    public function get($key, mixed $replace = null, $locale = null, $fallback = true)
     {
         $locale = $locale ?: $this->locale;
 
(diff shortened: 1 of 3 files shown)


src/Illuminate/Translation/Translator.php
```diff
@@ -99,12 +99,12 @@ public function has($key, $locale = null, $fallback = true)
...
      *
      * @param  string  $key
-     * @param  array  $replace
+     * @param  mixed  $replace
      * @param  string|null  $locale
      * @param  bool  $fallback
      * @return string|array
      */
-    public function get($key, array $replace = [], $locale = null, $fallback = true)
+    public function get($key, mixed $replace = null, $locale = null, $fallback = true)
```
driesvints, line 107: This is a breaking change to the method signature.

ajmeireles: > This is a breaking change to the method signature.

Hey! What do you propose?

driesvints: Best to send this to master

ajmeireles: > Best to send this to master

What will be the difference to send it to the master?

driesvints: New major versions allow for breaking changes. But I'm not sure we should do these changes. Up to you to try to send it to master or not.

taylorotwell:

PR #44941, opened 2022-11-14
Populate attributes on new instance after foreign attributes

jonathanroy:
I am submitting this feature request after encountering an issue where I was assuming the foreign model had been already associated in my mutator methods.

For example, let's say I have the following code:

`$child = $parent->children()->make($attributes)`

I might have one attribute called `last_name` that I want to set by default to the parent's last name. However, at the moment of initializing the new Child model, the parent is not yet associated.

Up until now, I have been resorting to doing the following:

`$child = $parent->children()->make()->fill($attributes)`

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
@@ -50,8 +50,10 @@ abstract class HasOneOrMany extends Relation
      */
     public function make(array $attributes = [])
     {
-        return tap($this->related->newInstance($attributes), function ($instance) {
+        return tap($this->related->newInstance(), function ($instance) use ($attributes) {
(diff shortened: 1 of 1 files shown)


driesvints:
As you can see, this breaks the existing test suite.

jonathanroy:
Thanks, I will investigate.
I'm curious to hear what you think of the behaviour.

jonathanroy:
@driesvints I just investigated the tests and they will always fail.

Take a look at this test: https://github.com/laravel/framework/blob/9.x/tests/Database/DatabaseEloquentHasManyTest.php#L20

My PR, on purpose, will add the relationship foreign key to the instance. Hence, it won't only be `['name' => 'taylor']` but also `['parent_key' => ...]`.

How should I proceed from here?

PR #44945, opened 2022-11-15
[9.x] Validate uuid before route binding query

benbjurstrom:
This pull request overrides the resolveRouteBinding method in the HasUuids trait to ensure the given URI segment is a valid UUID before passing it to the database.

This check is necessary when using a Postgres database with the UUID datatype as passing anything other then a valid UUID will throw a QueryException. This check is also useful more generally to avoid unnecessary database queries.

### DIFF
--- a/tests/Integration/Routing/ImplicitModelRouteBindingTest.php
+++ b/tests/Integration/Routing/ImplicitModelRouteBindingTest.php
@@ -4,12 +4,10 @@ namespace Illuminate\Tests\Integration\Routing;
 
 use Illuminate\Database\Eloquent\Concerns\HasUuids;
 use Illuminate\Database\Eloquent\Model;
-use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Database\Eloquent\SoftDeletes;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Facades\Route;
 use Illuminate\Support\Facades\Schema;
-use Illuminate\Support\Str;
 use Orchestra\Testbench\Concerns\InteractsWithPublishedFiles;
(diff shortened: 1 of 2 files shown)


driesvints (approved):
Good idea 👍

driesvints:
We might want to do the same for the ULID trait?

michaelnabil230:
> We might want to do the same for the ULID trait?

I created it
https://github.com/laravel/framework/pull/44995

taylorotwell:
I don't think this properly addresses child route bindings.

benbjurstrom:
It looks like moving the check down to resolveRouteBindingQuery catches the child bindings as well. I updated the pull request. Let me know if you'd like to see any other changes.

taylorotwell:
Don't you actually need to check the `$field` parameter if it's present?

tests/Integration/Routing/ImplicitModelRouteBindingTest.php
```diff
@@ -40,6 +41,13 @@ protected function defineDatabaseMigrations(): void
...
+        Schema::create('comments', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+            $table->string('slug');
+            $table->integer('user_id');
+            $table->timestamps();
+        });
+
         $this->beforeApplicationDestroyed(function () {
             Schema::dropIfExists('users');
             Schema::dropIfExists('posts');
```
michaelnabil230, line 53: Please add the drop method of the table `comments` in `beforeApplicationDestroyed`

michaelnabil230, line 53: Please add the drop method of the table `comments` in `beforeApplicationDestroyed`

PR #25203, opened 2018-08-14
[5.6] Use write connection on Schema::hasColumn() and Schema::hasTable()

crynobone:
This solves issue when using `Schema::hasColumn()` or `Schema::hasTable()`
on a read write connection with huge latency.

Signed-off-by: Mior Muhammad Zaki <crynobone@gmail.com>

PR #44946, opened 2022-11-15
[9.x] Use write connection on Schema::getColumnListing() and Schema::hasTable() for MySQL and PostgreSQL

taka-oyama:
This is a follow up PR to https://github.com/laravel/framework/pull/25203.

That PR made it so that the `Illuminate\Database\Schema\Builder` would do a `select` on the write connection.

The change was made to prevented migrations from getting `PDOException` when the read connection had not received the preceding migration changes due to replication lag.

The problem with that PR was that it applied the fix towards the generic schema builder which is overridden for [MySQL](https://github.com/laravel/framework/blob/213e07ef363bb6f2355254a5a9a0da33dcd692ad/src/Illuminate/Database/Schema/MySqlBuilder.php#L43) and [PostgreSQL](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Schema/PostgresBuilder.php#L51) specific builders.

So the fix was not applied for MySQL and PostgreSQL. This PR is an attempt to fix that.

This change will not break any existing features.

### DIFF
--- a/src/Illuminate/Database/Schema/MySqlBuilder.php
+++ b/src/Illuminate/Database/Schema/MySqlBuilder.php
@@ -40,7 +40,7 @@ class MySqlBuilder extends Builder
     {
         $table = $this->connection->getTablePrefix().$table;
 
-        return count($this->connection->select(
+        return count($this->connection->selectFromWriteConnection(
             $this->grammar->compileTableExists(), [$this->connection->getDatabaseName(), $table]
         )) > 0;
     }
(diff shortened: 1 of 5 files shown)


taka-oyama:
I just realized there were no tests for this case. I will add some.

taka-oyama:
AWS's S3 library seems to be failing but that is unrelated to this fix.

I submitted a [PR to aws-sdk-php](https://github.com/aws/aws-sdk-php/issues/2558) to fix the failing test.

driesvints (approved):

taka-oyama:
Sorry, you're right, I meant I sent an issue.

PR #44948, opened 2022-11-15
[9.x] Extract statement `else then return`

maoleng:
Extract these statements for clearer intention and lower complexity numbers


taylorotwell:
It's fine.

PR #44950, opened 2022-11-15
[9.x] fix foreach loop in Str::is method

mirzazeyrek:
There is a typo in foreach loop for the pattern variable which modifies the pattern value probably unintentionally. It's a small fix for that.


taylorotwell:
If there is a bug send a failing test with your fix.

PR #44959, opened 2022-11-15
[10.x] Add the PresentIf validation rule

neemspees:
Like the `required_if` and `prohibited_if` validation rules I would like to add the `present_if` validation rule.

This rule allows you to conditionally constrain an attribute to be present.

I also added the `PresentIf` rule class.

This PR only adds features so it is backwards compatible.

PR for documentation: `TODO`
PR for translation: `TODO`


taylorotwell:

neemspees:
@taylorotwell I needed this rule in multiple projects and I think it should be included in the framework. When will Laravel be open for new features like this?

PR #44962, opened 2022-11-16
[9.x] Fix passing null to `assertRedirectToRoute`

zaherg:
A few days ago, I added this PR https://github.com/laravel/framework/pull/44926 which introduced a new assertion function called `assertRedirectToRoute`.

as noted by @gms8994 here I have forgot to check when passing `null` as the name for the route.

This PR should fix this, and include the test code which @gms8994 provided.

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -270,11 +270,13 @@ class TestResponse implements ArrayAccess
             $this->statusMessageWithDetails('201, 301, 302, 303, 307, 308', $this->getStatusCode()),
         );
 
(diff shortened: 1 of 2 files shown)


simensen:
I think rather than testing for null and adding a test for the null case we instead follow `route()`'s lead and make `$name` not nullable.

```php
function route($name, $parameters = [], $absolute = true)
```

```php
public function assertRedirectToRoute($name, $parameters = [])
```
Since this is already merged, I don't know what the BC rules are since it was merged *so recently*. Can we still make changes to the call signature at this point?

taylorotwell:
Made name required.

zaherg:
will do, PR coming in a few min

PR #44972, opened 2022-11-16
Make success message of migration creation similar to other makers

ajaxray:
The success message pattern of creating a migration is different from other entity related classes. 

Generally it's not a noticeable issue. But when we run `make:model {name} -a`, it creates bunch of classes and the message looks odd in that situation. Also, it doesn't display the file path like others in `make:model {name} -a`.

This tiny PR will make the message pattern similar with other makers.

**Current success message:**
image

**This PR will make it:**
image

### DIFF
--- a/src/Illuminate/Foundation/Console/ModelMakeCommand.php
+++ b/src/Illuminate/Foundation/Console/ModelMakeCommand.php
@@ -117,6 +117,7 @@ class ModelMakeCommand extends GeneratorCommand
         $this->call('make:migration', [
             'name' => "create_{$table}_table",
             '--create' => $table,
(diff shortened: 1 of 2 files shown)


src/Illuminate/Foundation/Console/ModelMakeCommand.php
```diff
@@ -117,6 +117,7 @@ protected function createMigration()
         $this->call('make:migration', [
             'name' => "create_{$table}_table",
             '--create' => $table,
+            '--fullpath' => true,
```
taylorotwell, line 120: Why is this part of this PR?

PR #44967, opened 2022-11-16
Include Eloquent Model Database name in model:show command

muath-ye:
This PR builds on the new `artisan model:show` command to show the database name of the Model. It could be handy to see on the Model information as it's not always obvious the name of the database of a Model.

It handles cases where multiple databases connection are used.

image

PR #44976, opened 2022-11-17
[9.x] Include Eloquent Model Database name in model:show command to distinguish between connection and database names

muath-ye:
@taylorotwell actually this is fixing for "capitalization" on PR #44967 and I did it because of confusing between database name and connection name.
This PR builds on the new `artisan model:show` command to show the database name of the Model. It could be handy to see on the Model information as it's not always obvious the name of the database of a Model.

It handles cases where multiple databases connection are used.

_________

It is originally called the connection name as database, see bellow.
image

But this PR is differentiate between database name and connection name
image


driesvints:
These are technically breaking changes to the method signatures.

muath-ye:
@driesvints Thanks for your response.
Can I add this PR to [master] branch (Laravel 10)? Or what do you think?

driesvints:
Best sent to master yeah.

PR #44979, opened 2022-11-17
Strip key, secret and token from root config options on aws clients

### LINKED ISSUE #44977: Error with new aws/aws-sdk-php
- Laravel Version: 9.40.1
- PHP Version: 8.1.5
- Database Driver & Version: Mysql 8.0.28

### Description:
We are currently using a filesystem that is hosted at DigitalOcean (space), we are using the AWS adapter requiring a few packages, one of the packages: aws/aws-sdk-php

By upgrading from 3.242.1 to 3.243.0 we suddenly got issues in our system (a nasty Xdebug has detected a possible infinite loop, and aborted your script with a stack depth of '256' frames)
image

We tested by pinning in our composer.json
"aws/aws-sdk-php": "3.242.1", -> OK
"aws/aws-sdk-php": "3.243.0", -> NOT OK

After investigation, we identified this bug reported on that repo:
https://github.com/aws/aws-sdk-php/issues/2563

We tried to apply the patch proposed on src/Illuminate/Filesystem/FilesystemManager.php (before $client = new S3Client($s3Config)):
`$s3Config['endpoint_provider'] = PartitionEndpointProvider::defaultProvider(); `
and that fixes the issue...

Seems like it's linked to a file  that starts with / (example : /private/image.gif )

georgeboot:
Fixes #44977

Related to https://github.com/aws/aws-sdk-php/issues/2567

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemManager.php
+++ b/src/Illuminate/Filesystem/FilesystemManager.php
@@ -265,7 +265,7 @@ class FilesystemManager implements FactoryContract
             $config['credentials'] = Arr::only($config, ['key', 'secret', 'token']);
         }
 
-        return Arr::except($config, ['key', 'secret', 'token']);
+        return Arr::except($config, ['token']);
     }
 
     /**
--- a/src/Illuminate/Filesystem/FilesystemManager.php
+++ b/src/Illuminate/Filesystem/FilesystemManager.php
@@ -265,7 +265,7 @@ class FilesystemManager implements FactoryContract
             $config['credentials'] = Arr::only($config, ['key', 'secret', 'token']);
         }
 
-        return $config;
+        return Arr::except($config, ['key', 'secret', 'token']);
     }
 
     /**
--- a/src/Illuminate/Mail/MailManager.php
+++ b/src/Illuminate/Mail/MailManager.php
@@ -253,7 +253,7 @@ class MailManager implements FactoryContract
             $config['credentials'] = Arr::only($config, ['key', 'secret', 'token']);
         }
 
-        return $config;
+        return Arr::except($config, ['key', 'secret', 'token']);
     }
 
     /**
--- a/src/Illuminate/Mail/MailManager.php
+++ b/src/Illuminate/Mail/MailManager.php
@@ -253,7 +253,7 @@ class MailManager implements FactoryContract
             $config['credentials'] = Arr::only($config, ['key', 'secret', 'token']);
         }
 
-        return Arr::except($config, ['key', 'secret', 'token']);
+        return Arr::except($config, ['token']);
     }
 
     /**
--- a/src/Illuminate/Queue/Connectors/SqsConnector.php
+++ b/src/Illuminate/Queue/Connectors/SqsConnector.php
@@ -24,7 +24,7 @@ class SqsConnector implements ConnectorInterface
 
         return new SqsQueue(
             new SqsClient(
-                Arr::except($config, ['key', 'secret', 'token'])
+                Arr::except($config, ['token'])
             ),
             $config['queue'],
             $config['prefix'] ?? '',
--- a/src/Illuminate/Queue/Connectors/SqsConnector.php
+++ b/src/Illuminate/Queue/Connectors/SqsConnector.php
@@ -23,7 +23,9 @@ class SqsConnector implements ConnectorInterface
         }
 
         return new SqsQueue(
-            new SqsClient($config),
+            new SqsClient(
+                Arr::except($config, ['key', 'secret', 'token'])
+            ),
             $config['queue'],
             $config['prefix'] ?? '',
             $config['suffix'] ?? '',


driesvints:
I'm super scared that this will break apps on older versions of the AWS PHP SDK. Is there any way we can verify this on a range of AWS PHP SDK versions?

georgeboot:
AFAIK the credentials have always only been inside the `credentials` key. They were also supplied at root level, and ignored by the sdk as they were not valid keys. Until now.

SamRemis:
Hi there :) 
Here from the PHP SDK- 'key' and 'secret' shouldn't be an issue, just 'token'; some new services are adding it as an authentication scheme this week, so we need to give customers some way of passing in their own.  There's no valid reason that I could think of right now where we would need you to remove the 'key' and 'secret' from the base level.  Shouldn't harm anything either though since those values would just be ignored right now.

driesvints:
Thanks @SamRemis. If there's also no reason to keep them then we might be able to just accept the entire PR. Thanks @georgeboot for looking into it.

georgeboot:
@driesvints revert last commit if you want

SamRemis:
> I'm super scared that this will break apps on older versions of the AWS PHP SDK. Is there any way we can verify this on a range of AWS PHP SDK versions?

It's used [here](https://docs.aws.amazon.com/aws-sdk-php/v2/guide/credentials.html) in V2 of the SDK as a nested array, and that's quite old.  I can look into the history of it later today to confirm, but it'll take some time.  If it is, then we can definitely push out something to fix this.


> @SamRemis how big is the chance that `key` or `secret` ever get added as parameters and this becomes an issue again?
The only use case I can see for this is if for some reason customers start asking for it.  There's no reason I can think of that either of these would become a valid base level config value, but I don't have a crystal ball.

driesvints:
> It's used [here](https://docs.aws.amazon.com/aws-sdk-php/v2/guide/credentials.html) in V2 of the SDK

We only allow v3 in Laravel v9 so that should be okay.

driesvints:
@SamRemis just found another issue with the token PR on aws php sdk. Left a comment on it: https://github.com/aws/aws-sdk-php/pull/2517

Mesuva:
Would be nice to have a patch for V8. 

For live production systems running serverless we're not going to constantly be upgrading our major Laravel version, but still might want to do updates to pull in security or bugfixes across libraries.

dennisprudlo:
There's no need to upgrade constantly. Laravel 8 security fixes were supported 2-3 years. Big fixes ended half a year ago.

mmehmet:
@georgeboot I'm experiencing this issue using the DynamoDB Cache driver - which I note when I look at the `Illuminate\Support\Cache\CacheManager` class I see a method `newDynamodbClient()` which appears to want to pass in a token value into the constructor of a new `DynamoDbClient` (which extends `AwsClient`)

```
    protected function newDynamodbClient(array $config)
    {
        $dynamoConfig = [
            'region' => $config['region'],
            'version' => 'latest',
            'endpoint' => $config['endpoint'] ?? null,
        ];

        if (isset($config['key'], $config['secret'])) {
            $dynamoConfig['credentials'] = Arr::only(
                $config, ['key', 'secret', 'token']
            );
        }

        return new DynamoDbClient($dynamoConfig);
    }
```

I could be reading this wrong, but I believe that 'token' should be removed from this array?

for ref: I'm on Laravel v9.52.10

anuragnandan:
Hey @mmehmet I tried removing key and secret from cache config for dynamodb and Im still getting the same error. Did you had to do anything more to make it work ?

mmehmet:
> Hey @mmehmet I tried removing key and secret from cache config for dynamodb and Im still getting the same error. Did you had to do anything more to make it work ?

@anuragnandan sorry I should have clarified, this was for an AWS instance - so authentication happens via IAM and credentials are unused/ignored, within the AWS-to-AWS ecosystem

if you actually do need to authenticate using credentials (eg from an instance running on a dev machine and not within AWS) then this will not work - since you do actually need a key and secret...

in that case, I would recommend building your own client, which extends the above or at least also extends an `AwsClient` - and just leave out the `token` value in your implementation:

`
        if (isset($config['key'], $config['secret'])) {
            $dynamoConfig['credentials'] = Arr::only(
                $config, ['key', 'secret']
            );
        }
`

anuragnandan:
@mmehmet  I'm having an issue running the app on a Lambda which uses AWS Role. It seems to have started when https://github.com/laravel/framework/commit/909ea24597b86cfc7a49ce41074280d0347bf6a3 was merged and we upgraded our framework. I have tried removing key and secret from `cache.stores.dynamodb` but it didnt seem to work as you suggested.

mmehmet:
did you _only_ remove `key` and `secret` - as in you did not also remove `token`?

because the code will try to use that if it finds one

```
if (! empty($config['token'])) {
    $dynamoConfig['credentials']['token'] = $config['token'];
}
```

anuragnandan:
Yes, I tried removing just key and secret first and then key, secret and token from config and have the same issue.

anuragnandan:
@mmehmet Nothing seem to work on Lambda, Im applying below patch to make it work for now. Let me know if I can try anything else.
```
--- vendor/laravel/framework/src/Illuminate/Cache/CacheManager.php	2023-09-19 15:25:04.000000000 +0000
+++ CacheManager.php	2023-11-01 20:02:45.693625146 +0000
@@ -92,7 +92,7 @@
             return $this->callCustomCreator($config);
         }

-        $driverMethod = 'create'.ucfirst($config['driver']).'Driver';
+        $driverMethod = 'create' . ucfirst($config['driver']) . 'Driver';

         if (method_exists($this, $driverMethod)) {
             return $this->{$driverMethod}($config);
@@ -260,14 +260,11 @@

         if (! empty($config['key']) && ! empty($config['secret'])) {
             $dynamoConfig['credentials'] = Arr::only(
-                $config, ['key', 'secret']
+                $config,
+                ['key', 'secret', 'token']
             );
         }

-        if (! empty($config['token'])) {
-            $dynamoConfig['credentials']['token'] = $config['token'];
-        }
-
         return new DynamoDbClient($dynamoConfig);
     }
```

PR #44983, opened 2022-11-17
[10.x] Include Eloquent Model Database name in model:show command to distinguish between connection and database names

muath-ye:
@driesvints https://github.com/laravel/framework/pull/44976#issuecomment-1318437753
This PR builds on the new `artisan model:show` command to show the database name of the Model. It could be handy to see on the Model information as it's not always obvious the name of the database of a Model.

It handles cases where multiple databases connection are used.

_________

It is originally called the connection name as database, see bellow.
image

But this PR is differentiate between database name and connection name
image

### DIFF
--- a/src/Illuminate/Foundation/Console/ShowModelCommand.php
+++ b/src/Illuminate/Foundation/Console/ShowModelCommand.php
@@ -228,38 +228,38 @@ class ShowModelCommand extends DatabaseInspectionCommand
      * Render the model information.
      *
      * @param  string  $class
-     * @param  string  $database
+     * @param  string  $connection
      * @param  string  $table
      * @param  \Illuminate\Support\Collection  $attributes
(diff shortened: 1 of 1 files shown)


WatheqAlshowaiter:
@muath-ye 

Why don't you name 'Database' instead of 'DatabaseName'? 

keep it as it is and add the "Connection" to the list.

muath-ye:
> Also think Database is probably a nicer looking name 👍

@driesvints I have change it to name database like as @WatheqAlshowaiter suggested.
image

crynobone:
Screenshot 2022-11-17 at 8 42 16 PM

Why not utilise the existing structure. Also believe `database` should also include "driver" info if connection name != driver

taylorotwell:
I'm not making any changes to this right now.

PR #44986, opened 2022-11-17
[9.x] Fix deprecation warning when comparing a password against a `NULL` database password

MaxGiting:
This used to be dealt with via a `strlen(null)` check added in this commit 
https://github.com/laravel/framework/commit/a36f1fe931abd07460f49875491ee066e319a64a
@GrahamCampbell I'd appreciate if you could cast your eye over this change.

`strlen(null)` used to return 0 as shown in [this PHP docs comment](https://www.php.net/manual/en/function.strlen.php#:~:text=%3C%3Fphp%20strlen(null)%3B%20%3F%3E%20will%20return%200), however since PHP 8.0, passing null to strlen() has been deprecated.

I think this check has always been a null check as passing an empty string as the second parameter to `password_verify()` has always been permitted and throws no level of warning or error. 

To be on the safe side I have also added tests for when `$hashedValue` is passed in as an empty string.

We're are seeing a lot of deprecation notices as in our system it is possible for users to have a NULL password in the database until they have activated their account and setup a password.

### DIFF
--- a/src/Illuminate/Hashing/AbstractHasher.php
+++ b/src/Illuminate/Hashing/AbstractHasher.php
@@ -19,13 +19,13 @@ abstract class AbstractHasher
      * Check the given plain value against a hash.
      *
      * @param  string  $value
-     * @param  string  $hashedValue
+     * @param  string|null  $hashedValue
      * @param  array  $options
      * @return bool
      */
     public function check($value, $hashedValue, array $options = [])
     {
-        if (strlen($hashedValue) === 0) {
+        if (is_null($hashedValue)) {
             return false;
         }
 
--- a/src/Illuminate/Hashing/Argon2IdHasher.php
+++ b/src/Illuminate/Hashing/Argon2IdHasher.php
@@ -10,7 +10,7 @@ class Argon2IdHasher extends ArgonHasher
      * Check the given plain value against a hash.
      *
      * @param  string  $value
-     * @param  string  $hashedValue
+     * @param  string|null  $hashedValue
      * @param  array  $options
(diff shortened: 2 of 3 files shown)


driesvints:
A password in the database should never be null. Just generate passwords for all users.

JayBizzle:
> A password in the database should never be null. Just generate passwords for all users.

In that case, should passwords in the database ever be empty strings?

JayBizzle:
We have a system that allows users to invite other users. When the user is invited, an account is created with a `NULL` password, and the user creates their own password when they accept the invitation.

If passwords in the database should never be `NULL`, surely they should never be empty strings, but there an existing check to make sure the password in the database is no an empty string?

driesvints:
By default, a password is not nullable in Laravel: 

https://github.com/laravel/laravel/blob/9.x/database/migrations/2014_10_12_000000_create_users_table.php#L21

I can't speak for empty strings.

MaxGiting:
I believe this commit https://github.com/laravel/framework/commit/a36f1fe931abd07460f49875491ee066e319a64a was added to fix errors if null was passed in as the hashed password.
The commit message was "Fixed errors for empty password hashes". No errors would have been thrown from passing an empty string in as the second parameter to `password_verify()` whereas passing `null` does.

Would be good to hear from @GrahamCampbell  to know if that was the case.

taylorotwell:
I don't understand why your tests were using `assertTrue`. Should they not be `assertFalse`? I have changed them.

MaxGiting:
Yes they should have been assertFalse

PR #44988, opened 2022-11-17
Do not invoke Attribute method if it has parameters

Cellard:
In some cases Model may has helper Attribute method with some parameters. This pr suggests to isolate such helper methods from being invoked while building array of attributes.


Cellard:
It was a bug fix 😢

PR #44989, opened 2022-11-17
[9.x] Add a new function `onlyTrashed`

michaelnabil230:
This PR introduces an "onlyTrashed" validation rule in `Exists` and `Unique`.

For example, we want a user to be able to create a new row in the database user, but we want to search in the database where the `deleted_at` is not null.
And the same since the same function is in the same trait `SoftDeletes`

Usage:

```php
Validator::make(
    [
        'name' => 'Michael Nabil',
    ],
    [
        'name' => [
            'required', 
            Rule::exists('users')->where('id', $this->id)->onlyTrashed(),  
            Rule::unique('users')->where('id', $this->id)->onlyTrashed(),  // Or
        ],
    ]
);
```

### DIFF
--- a/src/Illuminate/Validation/Rules/DatabaseRule.php
+++ b/src/Illuminate/Validation/Rules/DatabaseRule.php
@@ -183,7 +183,7 @@ trait DatabaseRule
     }
 
     /**
-     * Do not ignore soft deleted models during the existence check.
+     * Only include soft deleted models during the existence check.
      *
      * @param  string  $deletedAtColumn
      * @return $this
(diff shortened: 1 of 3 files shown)


martinbean:
The naming seems off on this for me. “Only trashed” reads to me as if it’ll only check soft-deleted records; not _omit_ soft-deleted (trashed) records when doing the exists/unique checking 😕

s4muel:
> The naming seems off on this for me. “Only trashed” reads to me as if it’ll only check soft-deleted records; not _omit_ soft-deleted (trashed) records when doing the exists/unique checking 😕

Well, it reads to you just fine, because it really checks only within soft-deleted items 🤯

PR #44994, opened 2022-11-17
Fix typo in `ValidationEnumRuleTest`

michaelnabil230:


driesvints:
There's no reasoning here why this is needed.

michaelnabil230:
@driesvints 
And why use `Password::$defaultCallback` in validation enum rule

michaelnabil230:
@driesvints It is fix typo only

PR #44998, opened 2022-11-17
[9.x] Add some class rules in class `Rule`

michaelnabil230:
It is the best way to call these functions `password` and `imageFile` and `file`, `enum`.

And don't call in Controller or FormValidation all namespace

Before PR:

```php
use Illuminate\Http\Request;
use Illuminate\Validation\Rule;
use Illuminate\Validation\Rules\Password;
use Illuminate\Validation\Rules\File;
use Illuminate\Validation\Rules\Enum;

class WelcomeController
{
    public function store(Request $request)
    {
        $request->validate([
            'password' => Password::default(),
            'image' => File::default(),
            'status' => new Enum(StatusEnum::class),
            'name' => Rule::requiredIf(function () {
                return true;
            })
        ]);

        // 
    }
}
``` 

After PR:

```php
use Illuminate\Http\Request;
use Illuminate\Validation\Rule;

class WelcomeController
{
    public function store(Request $request)
    {
        $request->validate([
            'password' => Password::default(),
            'image' => Rule::file()->default(),
            'status' => Rule::enum(StatusEnum::class),
            'name' => Rule::requiredIf(function () {
                return true;
            })
        ]);

        // 
    }
}
```

### DIFF
--- a/src/Illuminate/Validation/Rule.php
+++ b/src/Illuminate/Validation/Rule.php
@@ -12,6 +12,10 @@ use Illuminate\Validation\Rules\NotIn;
 use Illuminate\Validation\Rules\ProhibitedIf;
 use Illuminate\Validation\Rules\RequiredIf;
 use Illuminate\Validation\Rules\Unique;
+use Illuminate\Validation\Rules\Password;
+use Illuminate\Validation\Rules\ImageFile;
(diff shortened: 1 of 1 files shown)


taylorotwell:
Removed the password one. The rest seems fine.

PR #45004, opened 2022-11-18
Unification ResetPassword email builder

injitools:
Currently in VerifyEmail notification class: if you set toMailUsing and createUrlUsing handlers then when you call toMail - VerifyEmail first using your url handler and send his return to body creator.

But in ResetPassword createUrlUsing ignoring while using toMailUsing.

This changes will unificate both handlers

### DIFF
--- a/src/Illuminate/Auth/Notifications/ResetPassword.php
+++ b/src/Illuminate/Auth/Notifications/ResetPassword.php
@@ -59,11 +59,13 @@ class ResetPassword extends Notification
      */
     public function toMail($notifiable)
     {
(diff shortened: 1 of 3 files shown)


crynobone:
This was attempted multiple times and the failing tests is there to prevent another attempt and showcase the breaking change.

injitools:
> This was attempted multiple times and the failing tests is there to prevent another attempt and showcase the breaking change.

Precisely because these are breaking changes, they are brought into the master branch, and not the current version as described in the documentation.
I corrected the tests.

Issue #45002, opened 2022-11-18
Vite.php is not using custom manifest file generated by laravel-vite-plugin


ngstwr:
- Laravel Version: 9.32
- PHP Version: 8.1

### Description:
Vite.php has 'manifest.json' hardcoded in manifestPath method!

protected function manifestPath($buildDirectory)
{
    return public_path($buildDirectory.'/manifest.json');
}

### Steps To Reproduce:
1. Install laravel-vite-plugin
yarn add laravel-vite-plugin -D

2. Change manifest file name to 'custom-manifest.json' in vite.config.js
build: {
    manifest: 'custom-manifest.json',
}

3. Build the application
yarn prod

driesvints:
Seems like this is a feature request. Feel free to send in a PR 👍

ngstwr:
I don't understand how is this a feature request! 

An official plugin 'laravel-vite-plugin', is generating the manifest file and Vite.php facade is using the same manifest file to map and provide resources. 

Now, laravel-vite-plugin provides an option to generate manifest, named as per user provided configuration. However Vite.php hasn't been updated to allow this.

Anyway, I don't have a problem sending in a PR, but I am new to Laravel and have no idea on how to write tests!

PR #45007, opened 2022-11-18
[9.x] Adding option for custom manifest filename on Vite Facade

ngstwr:
Option added to pass the name of the manifest file to be used while using the Vite Facade.

Before:
Vite::asset(['resources/sass/app.scss'], 'dist')

Now:
Vite::asset(['resources/sass/app.scss'], 'dist', 'app-manifest.json')

### DIFF
--- a/src/Illuminate/Foundation/Vite.php
+++ b/src/Illuminate/Foundation/Vite.php
@@ -49,6 +49,13 @@ class Vite implements Htmlable
     protected $buildDirectory = 'build';
 
     /**
+     * The name of the manifest file.
+     *
+     * @var string
+     */
+    protected $manifestFile = 'manifest.json';
+
+    /**
      * The script tag attributes resolvers.
      *
      * @var array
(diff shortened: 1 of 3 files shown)


ngstwr:
In response to this issue: #45002

timacdonald:
Hey @ngstwr,

Thanks for the PR. I've tweaked the implementation slightly. In a global service provider, route based middleware, or inline in a view, you may now specify the manifest file with the following API... 

Service provider / middleware example...

```php

Vite::useManifestFilename('custom-manifest.json');
```

```blade
{{-- Uses the custom manifest filename --}}

@vite(['resources/js/app.js'])
```

Inline example...

```blade
{{ Vite::useManifestFilename('custom-manifest.json')->withEntryPoints(['resources/js/app.js']) }}
```

timacdonald (approved):

ngstwr:
@timacdonald This is indeed a better way!

Somewhat related to this PR:

Shouldn't we do this with the 'buildDirectory' parameter also? 

If we can use the new API (useManifestFilename) via a global service provider or route-based middleware, it doesn't make sense to pass the custom build directory path as a parameter with every @vite() call.

Let me know what you think!

timacdonald:
@ngstwr that is already available with `Vite::useBuildDirectory($path)`.

See: https://laravel.com/docs/9.x/vite#advanced-customization

ngstwr:
Oh, thanks, somehow I missed that section on the documentation!

Will wait for this PR to be merged!

PR #45017, opened 2022-11-18
[9.x] Use the LARAVEL_ENV_ENCRYPTION_KEY to encrypt when it is present

ianfortier:
Use the `LARAVEL_ENV_ENCRYPTION_KEY` to encrypt when it is present.

The functionality is already there on `env:decrypt` but not on `env:encrypt`

Thanks!

### DIFF
--- a/src/Illuminate/Foundation/Console/EnvironmentEncryptCommand.php
+++ b/src/Illuminate/Foundation/Console/EnvironmentEncryptCommand.php
@@ -69,7 +69,7 @@ class EnvironmentEncryptCommand extends Command
     {
         $cipher = $this->option('cipher') ?: 'AES-256-CBC';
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
It shouldn't be needed on encrypt. The purpose of this env variable is so you can decrypt via env variables without putting your environment variable secret in your deployment script?

phadaphunk:
@taylorotwell If someone on your team needs to change the `.env.staging` file and re-encrypt it with the **same** key (so we don't have to change them on all target machines) they have to use the key locally in the prompt via `--key` so why not make it take a default value in the `.env` file you can set as `LARAVEL_ENV_ENCRYPTION_KEY` (as it does already for decrypt)?

PR #45019, opened 2022-11-18
[9.x] Add whenMissing() method to InteractsWithInput trait 

phh:
The `InteractsWithInput` methods `has()` and `filled()` both have a syntactic sugar method `whenHas()` and `whenFilled()` that accepts a callback to avoid adding `if's`.

This PR adds the method `whenMissing()` that is basically the opposite of `whenHas()`.

Today I was working on a middleware that takes care of both mappings and default fields. That looked something like this:

```php
$request->mergeWhenMissing([
    'type' => 'default',
]);
```

However, sometimes you need to add a little more than just the additional fields. Could be something like notifying a 3rd part service that this part of the API is still getting used:

```php
if ($request->whenMissing('type') {
    $this->deprecated('type');

    $request->merge(['type' => 'default']);
}
```

I just think it's a nice little syntactic sugar.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -225,6 +225,27 @@ trait InteractsWithInput
     }
 
     /**
+     * Apply the callback if the request is missing the given input item key.
+     *
+     * @param  string  $key
+     * @param  callable  $callback
+     * @param  callable|null  $default
+     * @return $this|mixed
+     */
+    public function whenMissing($key, callable $callback, callable $default = null)
+    {
+        if ($this->missing($key)) {
+            return $callback(data_get($this->all(), $key)) ?: $this;
+        }
+
+        if ($default) {
+            return $default();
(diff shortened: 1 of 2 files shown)


src/Illuminate/Http/Concerns/InteractsWithInput.php
```diff
@@ -224,6 +224,27 @@ public function missing($key)
...
+     * @return $this|mixed
+     */
+    public function whenMissing($key, callable $callback, callable $default = null)
+    {
+        if ($this->missing($key)) {
+            return $callback(data_get($this->all(), $key)) ?: $this;
+        }
+
+        if ($default) {
+            return $default();
```
foremtehan, line 242: `return value($default);` and eliminate callable type hint.

phh: The current solution is closer to how it's done in `whenHas()` and `whenFilled()` so at this stage I would prefer it like that.

Thanks for the feedback!

PR #45024, opened 2022-11-19
Outlook web dark mode email layout fix

0x1stvan:
In the web version of outlook (outlook.office.com) there is a visible border inside the email template layout when dark mode is enabled, this small change will hide this border, since outlook comments out the style tag, it only works with inline code.

The change working in version: 9.x and 8.x, older versions not tested.

Email layout in dark mode outlook web:
laravel_outlook_web_dark_email_layout

Email layout in dark mode outlook web (fixed)
laravel_outlook_web_dark_email_layout_fixed

### DIFF
--- a/src/Illuminate/Mail/resources/views/html/layout.blade.php
+++ b/src/Illuminate/Mail/resources/views/html/layout.blade.php
@@ -33,7 +33,7 @@ width: 100% !important;
 
 <!-- Email Body -->
 <tr>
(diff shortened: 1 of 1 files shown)


0x1stvan:
Outlook web comments out the style tag, but the selectors still working, so adding the bellow code will work like the inline CSS.

```
td.body {
border: hidden !important;
}
```

PR #45031, opened 2022-11-21
[9.x] ignore ECONNRESET in sqs:ReceiveMessage

Tietew:
`artisan queue:work sqs` frequently (a few times per week) dies with `Aws\Sqs\Exception\SqsException`:
> Aws\Sqs\Exception\SqsException: Error executing "ReceiveMessage" on "https://(queue url)"; AWS HTTP error: cURL error 35: OpenSSL SSL_connect: Connection reset by peer in connection to sqs.(region).amazonaws.com:443

This error may be caused by reusing HTTP connection after keepalive timeout.
It can be ignored safely and we should retry.

It is very noisy when we define custom `ErrorHandler` to report exceptions, but ignoring all `SqsException` may suppress other important errors.

This PR just ignores `Connection reset by peer` error in `sqs:ReceiveMessage` call.

### DIFF
--- a/src/Illuminate/Queue/SqsQueue.php
+++ b/src/Illuminate/Queue/SqsQueue.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Queue;
 
+use Aws\Sqs\Exception\SqsException;
 use Aws\Sqs\SqsClient;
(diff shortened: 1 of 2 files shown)


taylorotwell:
Register a custom `reportable` callback on SqsException that handle this for you.

PR #45051, opened 2022-11-22
make DB select work in migration pretend mode

### LINKED ISSUE #45050: DB Select statement do not return data in migration pretend mode
- Laravel Version: 9.0
- PHP Version: 8.0.21
- Database Driver & Version:
Mysql
### Description:
In database Laravel migration pretend mode `Select` statement do not return data which leads to errors in those migrations that read database. For example it could be a migration that extends values list in database field of `enum` type. To avoid hard-coded values lists such migration could first read `enum` type values list from database, then extend it and write the result back to database. Which works correctly in normal database migration mode but gives error when migration is launched in `pretend` mode.

### Steps To Reproduce:
Create a database migration that launch `Select` query and then uses its result in some way - for example to compose then next database query. This gives when migration is launched in `pretend` mode because in that mode `Select` query returns empty data.

lebedevsergey:
Allows to launch Laravel database migrations that read database in pretend mode without errors
The more detailed description is in in issue that this fixes: https://github.com/laravel/framework/issues/45050

### DIFF
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -403,10 +403,6 @@ class Connection implements ConnectionInterface
     public function select($query, $bindings = [], $useReadPdo = true)
     {
         return $this->run($query, $bindings, function ($query, $bindings) use ($useReadPdo) {
-            if ($this->pretending()) {
-                return [];
-            }
-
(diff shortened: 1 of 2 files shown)


taylorotwell:
Would break other applications.

PR #45052, opened 2022-11-22
[9.x] Create new `schema:load` Artisan command

calebdw:
Hello,

Currently any existing schema files are loaded by the `migrate`/`migrate:fresh` commands when creating a new database. However, I have a couple of  non-default connections that need their schema files to be loaded but *without performing any migrations on those connections*---which is not possible at the moment.

This PR plucks the schema loading logic out of the `migrate` command and creates the sister command to `schema:dump`---namely, it loads/restores a schema file for a given database, but *does not perform any migrations*.

Thanks!

### DIFF
--- a/src/Illuminate/Database/Console/LoadCommand.php
+++ b/src/Illuminate/Database/Console/LoadCommand.php
@@ -1,0 +1,154 @@
+<?php
+
+namespace Illuminate\Database\Console;
+
+use Illuminate\Console\Command;
+use Illuminate\Console\ConfirmableTrait;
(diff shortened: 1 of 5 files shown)


taylorotwell:
I don't really want to make these changes on a patch release. You could likely write your own custom command in your application to do this?

CalebDW:
@taylorotwell,

Yes, I could move the `schema:load` command into my own application but I thought it would be useful to the community at large.

Would you be willing to consider adding this in the next major release?

driesvints:
@CalebDW I think it's best to just write your own custom command for now. Thanks

PR #45066, opened 2022-11-24
Add `--rest` option to `queue:listen`

PHPGuus:
When running the `queue:listen` command as a daemon one might want to wait between jobs for a short time, just to make sure that one can dispatch a lot of jobs, without them overwhelming a 3rd party API or a rate-limited service.

To that effect, I've made the following changes:
+ `sleep()` if there was a rest option
+ Add `rest` to the `ListenerOptions`
+ Add `--rest` to the `ListenCommand` and store the value in the `ListenerOptions`

Since I'm quite unsure how this could be tested, I have not prepared / written any test cases. Sorry for that!

### DIFF
--- a/src/Illuminate/Queue/Console/ListenCommand.php
+++ b/src/Illuminate/Queue/Console/ListenCommand.php
@@ -25,7 +25,8 @@ class ListenCommand extends Command
                             {--queue= : The queue to listen on}
                             {--sleep=3 : Number of seconds to sleep when no job is available}
                             {--timeout=60 : The number of seconds a child process can run}
-                            {--tries=1 : Number of times to attempt a job before logging it failed}';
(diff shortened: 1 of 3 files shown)


taylorotwell:
Merged.

tomswinkels:
Why add a sleep, I think it is better to add a delay to the next job?

PHPGuus:
@tomswinkels The difficulty, really, is that the listener conjures up a `--once` worker. And a `--once` worker does not consider the rest or delay options (since it runs only once...) We could pass `--rest` from the listener to the `--once` worker, but then we'd need to add code and a flag for the worker to consider `--rest` *if* called from the listener master process.

Just letting the listener sleep if `--rest` is given seemed far easier.

If you can find a way to make it work so that the listener propagates the `--rest` delay to the worker, I am more than willing to learn!!

PR #45068, opened 2022-11-24
[9.x] Sending boolean in get request

MostafaRabia:
When sending get request like:
`&is_user_paid=true`
He is reading `true` as string, and can't pass the validation

This fix it for get request


driesvints:
Please see many previous attempts.

PR #45070, opened 2022-11-24
[9.x] Add `response()->accepted()`

tecbay:
Whenever I dispatch a job from the controller. I returned 202 as [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202) suggested.
```
public function something(Request $request): Response
{
    PatiendDataImportingJob::dispatch($media);

    return response()->noContent(202);
}
```

So, I have to return `noContent` with the 202 code. But it would make a lot more sense If we could do something like
```
  return response()->accepted();
  or
  return response()->processing();
```

### DIFF
--- a/src/Illuminate/Routing/ResponseFactory.php
+++ b/src/Illuminate/Routing/ResponseFactory.php
@@ -70,6 +70,18 @@ class ResponseFactory implements FactoryContract
     }
 
     /**
+     * Create a new "accepted" response.
+     *
+     * @param  mixed  $content
+     * @param  array  $headers
+     * @return \Illuminate\Http\Response
+     */
+    public function accepted(mixed $content = '', array $headers = [])
+    {
+        return $this->make($content, 202, $headers);
+    }
+
+    /**
      * Create a new response for a given view.
      *
      * @param  string|array  $view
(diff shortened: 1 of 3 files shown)


src/Illuminate/Routing/ResponseFactory.php
```diff
@@ -69,6 +69,18 @@ public function noContent($status = 204, array $headers = [])
...
     }
 
+    /**
+     * Create a new "accepted" response.
+     *
+     * @param  mixed  $content
+     * @param  array  $headers
+     * @return \Illuminate\Http\Response
+     */
+    public function accepted(mixed $content = '', array $headers = [])
```
BrandonSurowiec, line 79: Need to remove the `mixed` type hint here. The other methods don't do that. The `array` is good.
```
public function accepted($content = '', array $headers = [])
```

taylorotwell:
We can't add to the response factory interface on a patch release. For now, you can just do:

```php
return response('', 202);
```

Or, define a macro:

```php
use Illuminate\Support\Facades\Response;

Response::macro('accepted', function ($content = '') {
    return Response::make($content, 202);
});

return response()->accepted();
```

PR #45072, opened 2022-11-24
Option to stop serve command if environment file has changed.

sdapkus:
Idea came while using Sail - when updating environment file, `serve` command restarts underlying php server and most of the values are getting picked up after update. But for example `APP_ENV` change is not picked up by restarted server. Underlying command is not restarted, and it is seems that it is passing original `APP_ENV` value.

This change unlocks ability, to stop server process and exit command if environment file has been updated. Then Sail supervisor will start fresh `serve` command with updated environment file.

I will make PR to Sail repository to start serve command with new option if this will be merged.


taylorotwell:

PR #45079, opened 2022-11-25
fix docblock

michaelnabil230:


michaelnabil230:
I think is array not mixed but I need confirmation for anyone.

driesvints:
Mixed is correct here, thanks!

PR #45082, opened 2022-11-25
[9.x] Call object's dump method when passed to global dump

avrahamappel:
Many Laravel classes have `dump` and `dd` methods.

It would be nice if these methods would be called when an object of this sort is passed to `dd` or `dump`, instead of the default behavior (which can get pretty ugly, especially with things like QueryBuilder, see screenshot below).

**Screenshot**
Screen Shot 2022-11-25 at 11 18 14 AM

 This would be especially useful when passing multiple values to the dump methods, i.e.
```php
dump($query1, $query2); // I still want nice output here
```

This PR adds a new contract, `Dumpable`, which requires the presence of the `dump` and `dd` methods on the implementing class. It also adds a new set of `dump` and `dd` functions which replace Symfony's while still being completely backwards compatible.

**Note:**
In order to get Composer to load `Illuminate\Support\helpers.php` before `vendor/symfony/var-dumper/Resources/functions/dump.php` while running tests, I had to install [`funkjedi/composer-include-files`](https://github.com/funkjedi/composer-include-files) (in require-dev only, hope that's ok). In a real application this will probably work out of the box, since both functions will be loaded by Composer, and "L" comes before "S".

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -86,6 +86,7 @@
         "aws/aws-sdk-php": "^3.235.5",
         "doctrine/dbal": "^2.13.3|^3.1.4",
         "fakerphp/faker": "^1.9.2",
+        "funkjedi/composer-include-files": "^1.1",
         "guzzlehttp/guzzle": "^7.5",
         "league/flysystem-aws-s3-v3": "^3.0",
         "league/flysystem-ftp": "^3.0",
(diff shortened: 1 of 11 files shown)


avrahamappel:
Looks like using `funkjedi/composer-include-files` broke an unrelated test. 😞 

Also, on second thought maybe we don't want this behavior, as we should leave a way for developers to actually dump the full object. Unless we add that option as a second parameter? 🤔

taylorotwell:

PR #45083, opened 2022-11-25
Add valuesRecursive Collection method

tavsec:
I added the `valuesRecursive` to `Collection` and `LazyCollection` classes. This method will allow developers to reset indexes recursively through multidimensional colletions, like so:

```php
collect([
    "a" => [
        "x" => 1,
        "y" => [
             "0"
          ]
      ],
      "b" => 1
])->valuesRecursive();

/* Will return 
[
    [
        1,
        [
            "0"
        ]
    ],
    1
]
*/
```

It will not break any current features, as it is completely new method, which does not interfere with any other Collection methods.


taylorotwell:

PR #45095, opened 2022-11-26
Fix docblock

michaelnabil230:


browner12:
Rather than removing the import, should we type-hint the parameter? We already do it in a couple middleware, and that's the direction the framework is moving anyway.

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Http/Middleware/TrustHosts.php

browner12:
my fault. though it was the `Request`, not the `Response`. probably better to leave that for Nuno to update.

michaelnabil230:
‏I’ll really leave that to Nuno.

PR #45103, opened 2022-11-26
Add morph helper to Request

hebinet:
This PR adds a new `morph` request helper.

**Usage**
```php
$contactable = request()->morph('contactable');

\\ $contactable is now an instance of the given morph.
```
The `morph` helper looks if `{$key}_id` and `{$key}_type` exists in the request and tries to instanciate a model instance with the given data.
If the fields aren't present or the model was not found it returns `null`.

**Why do I think this is usefull?**
Lately I had to deal a lot with morph models in an REST-API and I had to go the hard way like this:
```php
$id = request()->input('contactable_id');
$type = request()->input('contactable_type');

$contactable = null;
$class = Relation::getMorphedModel($type);
if ($class) {
  $contactable = $class::find($this->input($id))
}
```
This has to be done in multiple controllers so I extracted it and then I thought, maybe others have the same requirements so lets create a PR :)

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -2,7 +2,6 @@
 
 namespace Illuminate\Http\Concerns;
 
-use Illuminate\Contracts\Database\Eloquent\Builder;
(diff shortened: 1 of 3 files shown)


taylorotwell:
I suggest you macro this into your own project.

PR #45107, opened 2022-11-26
[9.x] Add method missingAny to request

fernandokbs:
This PR adds the method missingAny() that is basically the contrary of hasAny()

Since Laravel has the methods `has`, `hasAny`, `whenHas`, and `missing`, `whenMissing`, I thought it was a good idea to add this method.

Before
```php
if(! $request->hasAny(['name','email'])) {

}
```

After

```php
if($request->missingAny(['name','email'])) {

}
```

it adds syntactic sugar.


taylorotwell:
Logic looks more like `missingAll` unless I'm mistaken.

PR #45111, opened 2022-11-27
[10.x] Add Route::json() method

Sultenhest:
This PR adds a new `Route::json()` method making it possible to register a new route that returns json.

Before:
```php
Route::get('/', function() {
});
```

After:
```php
Route::json('/', ['foo' => 'bar']);
```

Arguments that can be passed are:
1. The route
2. The data (\Illuminate\Contracts\Support\Arrayable|array)
3. A status code
4. Headers
5. Options

Long version:
```php
Route::json('/', ['foo' => 'bar'], 418, ['Framework' => 'Laravel'], JSON_PRESERVE_ZERO_FRACTION);
```

### DIFF
--- a/src/Illuminate/Routing/JsonController.php
+++ b/src/Illuminate/Routing/JsonController.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace Illuminate\Routing;
+
+use Illuminate\Contracts\Routing\ResponseFactory;
+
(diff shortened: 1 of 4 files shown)


taylorotwell:
I would probably just do:

```
Route::get('/foo', fn () => [1, 2, 3]);
```

PR #45112, opened 2022-11-27
Align method visibility in HandlesAuthorization

Propaganistas:
Not sure why these two methods are public while their regular counterparts (`deny` / `allow`) are protected.


taylorotwell:
I see no reason to change this in a patch release.

PR #45117, opened 2022-11-28
[9.x] Allow set command description via `AsCommand` attribute

dinhquochan:
Currently, if we use the `AsCommand` attribute, it will not be set to `$description` of the class, because Laravel command overrides this property.

This PR will allow us to use `AsCommand` to set the description of the command class.

ExampleCommand.php using `AsCommand`:

```php
<?php

use Symfony\Component\Console\Attribute\AsCommand;

#[AsCommand('example', 'This is an example command')]
class ExampleCommand extends Command
{
    // 
}

```

**Before the PR**

image

**After the PR**

image

Thanks.

### DIFF
--- a/src/Illuminate/Console/Command.php
+++ b/src/Illuminate/Console/Command.php
@@ -79,7 +79,11 @@ class Command extends SymfonyCommand
         // Once we have constructed the command, we'll set the description and other
         // related properties of the command. If a signature wasn't used to build
         // the command we'll set the arguments and the options on this command.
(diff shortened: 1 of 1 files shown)


xuandung38:
Nice bro

vigstudio:
Great 👯

datlechin:
What a great PR

huynt57:
Nice idea

driesvints:
Please keep the conversation on-topic and use the 👍  if you like the PR.

PR #45118, opened 2022-11-28
[9.x] Prevent test issues with relations with the $touches property

### LINKED ISSUE #45064: Relationship is empty after using `hasAttached()` with model that `$touches` its parent
- Laravel Version: 9.41.0
- PHP Version: 8.1.11
- Database Driver & Version: SQLite 3.37.0 and MySQL 8.0.30

### Description:

Using the `hasAttached()` factory method to relate models, if either model `$touches` the relationship, the model instances will not be related correctly after being created—their relations to each other will be loaded but empty.

### Steps To Reproduce:

- New Laravel app demonstrating this issue: https://github.com/bakerkretzmar/laravel-has-attached-touches-bug (see [this test](https://github.com/bakerkretzmar/laravel-has-attached-touches-bug/blob/main/tests/Feature/ExampleTest.php))

- Failing `laravel/framework` test case: https://github.com/laravel/framework/compare/9.x...bakerkretzmar:framework:fix-factory-relations-with-touches

You have two Eloquent models, e.g. `User` and `Group`, connected with a BelongsToMany relationship. We want users to have their timestamps updated any time any of their groups are updated, so the `Group` model contains `protected $touches = ['users']`.

In a test, you create a user and group and attach them:

```php
$group = Group::factory()->hasAttached($user)->create();

$this->assertCount(1, $group->users);
```

That assertion fails.

The `users` relation is set on the new `$group` model, _as if_ it was just loaded from the database, but it's an empty collection (output from `dump($group)` inside the above test):

image

This only happens when one of the models has the relationship listed in its `$touches` property.

WendellAdriel:
As discussed on #45064 this is a change that will prevent the issue of an empty relationship when testing models with the `$touches` property defined.

TBH IDK if this is the expected behaviour so if this is not the best solution, feel free to close the PR or let me know what should I change so the PR can be accepted.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/Factory.php
+++ b/src/Illuminate/Database/Eloquent/Factories/Factory.php
@@ -328,24 +328,14 @@ abstract class Factory
 
             $model->save();
 
-            $this->removeEmptyRelations($model);
+            foreach ($model->getRelations() as $name => $items) {
+                if ($items->isEmpty()) {
+                    $model->unsetRelation($name);
+                }
+            }
+
             $this->createChildren($model);
         });
-    }
-
-    /**
-     * Remove the empty relations to avoid issues with touch when testing.
-     *
-     * @param  \Illuminate\Database\Eloquent\Model  $model
-     * @return void
-     */
-    protected function removeEmptyRelations(Model $model)
-    {
-        foreach ($model->getRelations() as $name => $items) {
-            if ($items->isEmpty()) {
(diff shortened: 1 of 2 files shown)


axlon:
@taylorotwell @WendellAdriel this PR breaks factories, it assumes a relation can only ever be a collection, but it can also be a single model.

This results in errors like:

```
BadMethodCallException : Call to undefined method App\Models\FooModel::isEmpty()
```

PR #45119, opened 2022-11-28
[9.x ] Updated Mailable to prevent duplicated recipients

### LINKED ISSUE #45042: Duplicate replyTo when running render() prior to send
- Laravel Version: 9.40.1
- PHP Version: 8.1.10
- Database Driver & Version: not relevant

### Description:
If you run `render()` prior to sending the email, it will double `replyTo`.
A bit related to #30857

### Steps To Reproduce:

The below example will add `info@example.com` 2 times to the reply_to field.
```
use App\Mail\Question;

$mailTemplate = new Question( 'Hello, I have a question' );
$mail = $mailTemplate->render();
// ...store $mail in the database

Mail::to( config( 'mail.from.address' ) )
    ->locale( 'nl' )
    ->send( $mailTemplate );
```

```
<?php
namespace App\Mail\Patient;

use App\Models\Patient;
use Illuminate\Bus\Queueable;
use Illuminate\Mail\Mailable;

class Question extends Mailable implements shouldQueue
{

    /**
     *
     * @param Patient $patient
     * @param string  $question
     */
    public function __construct( public string $question )
    {
        $this->queue = config( 'config.queue-email' );
    }

    /**
     * Build the message.
     *
     * @return $this
     */
    public function build()
    {
        return $this->from( [
            'address' => config( 'mail.from.address' ),
            'name'    => config( 'mail.from.name' ),
        ] )
                    ->replyTo( 'info@example.com' )
                    ->subject( 'someone has a question' )
                    ->view( 'mail.question' );
    }
}
```

WendellAdriel:
As discussed on #45042 these changes prevent the use of duplicated recipients in the `to`, `from`, `cc`, `bcc` and `replyTo` properties of the mailable.

### DIFF
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -689,6 +689,11 @@ class Mailable implements MailableContract, Renderable
             ];
         }
 
+        $this->{$property} = collect($this->{$property})
+            ->unique('address')
+            ->values()
+            ->all();
+
         return $this;
     }
 
--- a/tests/Mail/MailMailableTest.php
+++ b/tests/Mail/MailMailableTest.php
@@ -71,7 +71,7 @@ class MailMailableTest extends TestCase
         $mailable->assertHasTo('taylor@laravel.com');
 
         $mailable = new WelcomeMailableStub;
-        $mailable->to(collect([new MailableTestUserStub, new MailableTestUserStub, new MailableDumbUserStud()]));
+        $mailable->to(collect([new MailableTestUserStub, new MailableTestUserStub, new MailableTestUserStub2]));
         $this->assertEquals([
             ['name' => 'Taylor Otwell', 'address' => 'taylor@laravel.com'],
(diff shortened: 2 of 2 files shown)


driesvints (approved):
This *looks* okay to me. At least worth a try.

WendellAdriel:
Thank you so much @driesvints!!!

taylorotwell:
But now these tests no longer check that two "to" addresses can actually be added.

WendellAdriel:
That’s true @taylorotwell, sorry!
I’ll update the tests and the PR ASAP

AAllport:
It looks like this may have introduced a bug/breaking change...

If you have the following: `Mail::to($recipient)->send(new FooMail($recipient))`
You can no longer set the "name" of the recipient in the build method when calling `->to($reciepent->mail, $recipient->full_name)`

It looks like the call to `->unique()` is removing the instance of the address with the name.

Was this intended behaviour?
It feels like if you are calling `->to()`, then the most recent addition should take priority.

Issue #44741, opened 2022-10-26
[9.x] Dynamic Relation Resolver is inconsistent when extending Eloquent models 


vmcvlad:
- Laravel Version: 9.36.1
- PHP Version: 8.1.11
- Database Driver & Version: MySQL 8.0.19

### Description:

This [PR](https://github.com/laravel/framework/pull/33025) has added the `resolveRelationUsing()` method which facilitates dynamically defining relationships on models.

The way it works is by binding the fully-qualified class name inside the `$relationResolvers` array to the closure that accepts the model instance and returns the Eloquent relationship definition.

With the current implementation, whenever a parent model has a dynamically registered relationship, attempting to call the same dynamic relationship from a child instance will fail due to the fact that there is no key with the child's class name in the `$relationResolvers` array unless obviously it was also explicitly bound with the same `resolveRelationUsing()` method.  

```
    public function __call($method, $parameters)
    {
        ...
        if ($resolver = (static::$relationResolvers[get_class($this)][$method] ?? null)) {
            return $resolver($this);
        }
        ...
    }
```

The expected behaviour when extending models which have dynamically resolved relations would be to also have access to the parent's dynamic relationships and call them "as usual" from within child instances. 

However, the current behaviour is inconsistent, since relationships defined as methods inside an Eloquent model class will be inherited. There seems to be no apparent reason as to why the same principle shouldn’t apply to the dynamically defined relationships.

### Steps To Reproduce:

1. Define three distinct Eloquent classes, having one of them extend one of the others

```
class Foo extends Illuminate\Database\Eloquent\Model
{

}
```

```
class Bar extends Illuminate\Database\Eloquent\Model
{

}
```

```
class Baz extends Foo
{

}
```

2. Register a dynamic relationship between the extended Eloquent model class and the unextended one.

```
Foo::resolveRelationUsing(
  'bar', fn($model) => $model->belongsTo(Bar::class),
);
```

3. Create an instance of the related Eloquent model subclass and attempt to call it's parent's dynamically registered relationship in the previous step.

`(new Baz())->bar();`

`BadMethodCallException with message 'Call to undefined method Baz::bar()'`

### Possible fix:

A possible fix would be to recursively loop through each of the child model's parent classes and looking for the first dynamic relation registered. Perhaps a pattern like this:

```
    public function __call($method, $args)
    {
       ...
        if ($resolver = (self::relationResolver(static::class, $method))) {
            return $resolver($this);
        }
       ...
    }

[...]

PR #45122, opened 2022-11-28
[9.x] Address Dynamic Relation Resolver inconsiency issue with extended Models

vmcvlad:
This PR aims to address the inconsistency issues with child Eloquent Models not inheriting their parents' dynamically registered relations, an issue which I attempted to explain in detail in #44741.

I have also added an in-line comment with an alternative approach to implementing the `relationResolver()` method.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
@@ -70,6 +70,26 @@ trait HasRelationships
     }
 
     /**
+     * Return a dynamic relation resolver if defined or inherited, null otherwise
+     *
+     * @param  string  $class
+     * @param  string  $key
+     * @return mixed
+     */
+    public function relationResolver($class, $key)
+    {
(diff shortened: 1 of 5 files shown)


src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
```diff
@@ -69,6 +69,26 @@ public static function resolveRelationUsing($name, Closure $callback)
...
 
+    /**
+     * Return a dynamic relation resolver if defined or inherited, null otherwise
+     *
+     * @param  string  $class
+     * @param  string  $key
+     * @return mixed
+     */
+    public function relationResolver($class, $key)
+    {
```
vmcvlad, line 80: Alternative implementation:

```
public function relationResolver($key)
{
    $class = static::class;

    do {
        $resolver = static::$relationResolvers[$class][$key] ?? null;
    } while (!$resolver && $class = get_parent_class($class));

     return $resolver;
}
```

PR #45123, opened 2022-11-28
[9.x] Created the Clock facade to implement PSR-20

WendellAdriel:
I know that `Laravel` already has ways to deal with dates on it, but since `PRS-20` was accepted some days ago I thought on working on a simple implementation of it as a `facade`.

This PR introduces the `Clock` facade that implements the `ClockInterface` from the `PSR` implementing the `now` method and also adds two other methods: `withTimezone` and `createFromFormat`.

IDK if this is something that you want to add to the framework and if this is not the correct way of trying to add a new feature to the framework I apologize, I'm starting now to get involved with contributing to it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
         "monolog/monolog": "^2.0",
         "nesbot/carbon": "^2.62.1",
         "nunomaduro/termwind": "^1.13",
+        "psr/clock": "^1.0",
         "psr/container": "^1.1.1|^2.0.1",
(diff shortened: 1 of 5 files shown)


taylorotwell:
I appreciate your efforts - but not sure we really need this in the core at the moment. Thank you.

WendellAdriel:
@taylorotwell no problems at all, it was something I thought of but I was not sure as well if that's something you had in mind for the Framework. But thanks for taking some time to check on this! 😉

Issue #45091, opened 2022-11-25
Anonymous Global Scopes does not apply after $model->refresh()


amshehzad:
- Laravel Version: 9.41.0
- PHP Version: 8.1.2
- Database Driver & Version:

### Description:
When a refresh method is called on a model,  the Anonymous Global Scope does applied on model

### Steps To Reproduce:
In User Model, apply a global scope to select a column like this:

 ```
  protected static function booted()
  {
      static::addGlobalScope(fn(Builder $builder) => $builder
          ->when(
              is_null($builder->getQuery()->columns),
              fn(Builder $builder) => $builder->addSelect('*')
          )
          ->selectRaw("34 AS age")
      );
  }
```

Then in UserController:

```
$user = \App\Models\User::first();
$user->age;  // 34
$user->refresh();
$user->age;  // null
```

**Comment by WendellAdriel** [comment]
I think the same can be achieved using the `$attributes` array in the model instead of the global scope or am I missing something @amshehzad?

**Comment by amshehzad** [comment]
@WendellAdriel Yes that would be right if I had the same exact same attribute. I just added it here as an example, 
Actual code was this:
```
static::addGlobalScope(fn(Builder $builder) => $builder
            ->when(
                is_null($builder->getQuery()->columns),
                fn(Builder $builder) => $builder->addSelect('*')
            )
            ->selectRaw('IFNULL((SELECT SUM(amount) FROM leave_transactions WHERE user_id = users.id), 0) AS leave_balance')
        );
```

**Comment by WendellAdriel** [comment]
@amshehzad oh got it!
Can’t you do that with an `$appends` attribute using the relation between the two models?

**Comment by mattford** [comment]
I'm not sure you're using global scopes as intended, it looks like perhaps you need an accessor:

```
public function getAgeAttribute()
{
    return 34;
}
```

taylorotwell:
I don't think we would global scopes to apply on a `refresh` call - you could fail to retrieve the model entirely this way.

PR #45124, opened 2022-11-28
[9.x] Apply global scopes when refreshing model

WendellAdriel:
As discussed on #45091 this PR changes the `refresh` method to apply the `global scopes` when calling `refresh` on a model. This made some tests fail, so I updated those.

Another approach that can be done if this is not the expected behaviour and I can work on it is to add an optional parameter on `refresh` that defaults to `false` but when set to `true` it will apply the `global scopes` when refreshing or to create a new method `refreshWithScopes` per example.

Let me know your thoughts on this.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1688,7 +1688,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
         }
 
         $this->setRawAttributes(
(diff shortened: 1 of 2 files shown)


taylorotwell:
I dunno if we really want to change behavior like this on a patch release. IMO this could prevent the model from being retrieved at all.

WendellAdriel:
Yeah, I was afraid of that too! If I create a new PR creating a new refreshWithScopes would be considered for merge @taylorotwell?

Issue #45043, opened 2022-11-21
Model `increment()`/`decrement()` is unstable and may update everything


bert-w:
- Laravel Version: 9.x
- PHP Version: 8.x
- Database Driver & Version: MySQL/MariaDB

### Description:
I ran into yet another issue with the current increment/decrement functions on an Eloquent model. The issue with Eloquent's increment/decrement is that it seems to follow a completely different route than an update-statement would, resulting in weird behaviour like the following:

```php
// Assume a random $user (id:1) with a hasMany relation named `userConfig()`,
// which simply allows you to save custom properties attached to a user:
$cfg = $user->userConfig()->firstOrNew([
    // 'user_id' => 1, // (we don't need to set this explicitly because of the relation, which is correct)
    'name' => 'some_custom_property',
    'value' => 100,
]);

// Lets increment the value of this property with 200 (atomic query):
$cfg->increment('value', 200);
```

Expected outcome:
- Updated `value=300` for user_config `name=some_custom_property` for `user_id=1`

Actual outcome:
- If the userConfig record already exists -> expected outcome (correct behaviour)
- If the userConfig record does not exist (thus `$cfg` is a new userConfig object in the example above) -> ALL values in the user_config table are updated, not just for `user_id=1` and not just for `name=some_custom_property`. This is obviously not correct.

# Test case branch
https://github.com/bert-w/framework/tree/bug-increments-everything has both scenarios quickly hacked into a unit test. 

# Solution
See my previous PR: https://github.com/laravel/framework/pull/43646. This PR streamlines the `increment()` function to use the existing `update()` logic. This PR has some other nice possibilities as well like the ability to do:
```php
$model->update([
    'internet_points' => $model::expectable('+', 10),
    'coins' => $model::expectable('-', 5),
]);
```
This `expectable` is a wrapper class to help with determining the outcome of a query, based on the value the Eloquent model currently knows about a column (e.g. `internet_points + 10` would result in `15` if the `internet_points` in memory is equal to `5`). A hack for this is already used in Laravel at the moment where the new value is guesstimated in the model's `updated` event.

# Related
#43113, #32567

EDIT: changed the actual outcome description since the query is actually worse than I described: it updates the complete table with no where-statements whatsoever.

bert-w:
https://github.com/laravel/framework/blob/daf094441404d0c51084abb6c03a64da1391a798/src/Illuminate/Database/Eloquent/Model.php#L951-L957

[...]

PR #45127, opened 2022-11-28
[9.x] Prevent increment/decrement on non-existing models

bert-w:
This fixes the issue as described here: https://github.com/laravel/framework/issues/45043

Calling `increment` or `decrement` on a non-existing Eloquent model previously defaulted to executing it on a new query, effectively running the incrementation on the complete table (e.g. `UPDATE users SET value = value + 1`). This will unintentionally change the whole table. Furthermore, there were no tests that described a use case for this previous implementation, so I have added a test to explicitly disallow this behaviour.

Note: this probably needs a documentation section right after https://laravel.com/docs/9.x/eloquent#upserts named "Increment / Decrement". There is currently no mention of this functionality in the docs, only at https://laravel.com/docs/9.x/queries#increment-and-decrement which is the query builder itself.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -946,14 +946,14 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
      * @param  float|int  $amount
      * @param  array  $extra
      * @param  string  $method
-     * @return int
+     * @return int|false
      */
     protected function incrementOrDecrement($column, $amount, $extra, $method)
     {
         $query = $this->newQueryWithoutRelationships();
 
         if (! $this->exists) {
-            return $query->{$method}($column, $amount, $extra);
+            return false;
         }
 
         $this->{$column} = $this->isClassDeviable($column)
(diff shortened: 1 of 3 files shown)


tests/Integration/Database/EloquentUpdateTest.php
```diff
@@ -112,7 +112,7 @@ public function testIncrement()
             'counter' => 0,
         ])->delete();
 
-        TestUpdateModel3::increment('counter');
+        TestUpdateModel3::query()->increment('counter');
```
bert-w, line 115: I changed this test because the behaviour has changed now. Calling `increment` on a blank model created from scratch should `return false`. We already don't allow `TestUpdateModel3::update([...]);` either so that aligns properly with the expected

src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -2310,10 +2310,6 @@ public function __unset($key)
      */
     public function __call($method, $parameters)
     {
-        if (in_array($method, ['increment', 'decrement'])) {
-            return $this->$method(...$parameters);
-        }
```
bert-w, line 2315: Anyone knows what the pattern for this seems to be? Why not make the actual methods public?

Issue #45033, opened 2022-11-21
Pre-rendered Maintenance Mode is not rendered as if it was in Maintenance Mode


Cbrad24:
- Laravel Version: v9.38.0
- PHP Version: v8.1.12
- Database Driver & Version: Postgres

### Description:
The pre-rendered maintenance mode using the `php artisan down --render="errors::503"` command does not show up as if it was actually in maintenance mode.

### Steps To Reproduce:
1. Create a `resources/views/errors/503.blade.php` with the following contents:
```blade
@extends('errors::minimal')

@if(App::isDownForMaintenance())
    @section('title', __('Down for Maintenance'))
    @section('code', '503')
    @section('message', __('Down for Maintenance'))
@else
    @section('title', __('Service Unavailable'))
    @section('code', '503')
    @section('message', __('Service Unavailable'))
@endif
```
2. Run the command `php artisan down --render="errors::503"`.
3. View the application in a browser.
4. You will see 'Service Unavailable' instead of desired output 'Down for Maintenance'. This only happens with the pre '--render' flag.

It would be nice to be able to differentiate between maintenance mode and otherwise to display a more detailed reason for a 503 error to the user without having to boot the entire framework.

lk77:
When the view is rendered, the framework is not yet in maintenance mode

driesvints:
As said on https://github.com/laravel/framework/pull/45131 this is a bit of an edge case and thus we're not going to take any action here, sorry.

PR #45131, opened 2022-11-28
[9.x] Updated maintenance mode to be able to pre-render views usin…

WendellAdriel:
As discussed on #45033 when pre-rendering custom views for Maintenance mode, the app is not set to maintenance when the Framework renders the view. This PR updates this behaviour by adding a flag to check if the maintenance mode needs pre-render and if so, the isDownForMaintenance will return true when pre-rendering those custom views.

This PR also includes a new Blade directive maintenance that can be used instead of the isDownForMaintenance method, so instead of doing:

```
@extends('errors::minimal')

@if(App::isDownForMaintenance())
    @section('title', __('Down for Maintenance'))
    @section('code', '503')
    @section('message', __('Down for Maintenance'))
@else
    @section('title', __('Service Unavailable'))
    @section('code', '503')
    @section('message', __('Service Unavailable'))
@endif
```

We can do:

```
@extends('errors::minimal')

@maintenance
    @section('title', __('Down for Maintenance'))
    @section('code', '503')
    @section('message', __('Down for Maintenance'))
@else
    @section('title', __('Service Unavailable'))
    @section('code', '503')
    @section('message', __('Service Unavailable'))
@endmaintenance
```

### DIFF
--- a/src/Illuminate/Contracts/Foundation/MaintenanceMode.php
+++ b/src/Illuminate/Contracts/Foundation/MaintenanceMode.php
@@ -5,6 +5,21 @@ namespace Illuminate\Contracts\Foundation;
 interface MaintenanceMode
 {
     /**
+     * Marks the application in "pre-maintenance" mode so views can be pre-render checking for maintenance mode.
+     *
+     * @param  bool  $prerender
+     * @return void
+     */
+    public function prerenderMaintenance(bool $prerender): void;
+
+    /**
+     * Determine if the application needs to pre-render views for maintenance mode.
+     *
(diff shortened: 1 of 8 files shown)


taylorotwell:
Hey there - unfortunately we can't add methods to interfaces on patch releases. As this is a bit of an edge case issue - I wouldn't worry about it too much for now.

WendellAdriel:
Oh ok, got it!
I didn’t know, @taylorotwell!
Since you said it’s an edge case no need to open this to v10 right?

PR #45133, opened 2022-11-29
fix: drop column while droping the foreign key

anilkumarthakur60:
43d222d9e44524db93fbef9f4ced4b9a04806ed3

```
  public function dropForeignIdFor($model, $column = null)
    {
        if (is_string($model)) {
            $model = new $model;
        }

        $this->dropForeign([$column ?: $model->getForeignKey()]);

        return $this->dropColumn([$column ?: $model->getForeignKey()]);
    }
```


ankurk91:
Bro, Your IDE is touching too many files, Dont any kind of formatter plugin in vs-code.

PR #45135, opened 2022-11-29
[9.x] Fix factory breaking when trying to determine whether a relation is empty

axlon:
This fixes a bug introduced in https://github.com/laravel/framework/pull/45118. When a model's singular relation is loaded before the model is stored (e.g. in a `afterMaking` callback). The factory will try to call `isEmpty()` on it which will cause an error.

```
BadMethodCallException : Call to undefined method App\Models\OtherModel::isEmpty()
```


WendellAdriel:
@axlon that's a nice catch, I didn't expect this since it's something that was not covered in the tests.
I'm not sure if that's mandatory but I would say that adding a test case covering this issue that you fixed would be a good idea.
Also, sorry for the bug introduced with my PR 😅

PR #45136, opened 2022-11-29
[9.x] Add `redirect_if` and `redirect_unless` helpers

jbrooksuk:
This PR introduces two new helpers:

- `redirect_if`
- `redirect_unless`

They are analogous to the `abort_if` and `abort_unless` helpers in that they'll only redirect if a truthy boolean is passed.

Both helpers can be used to refactor something like:

```php
public function meta(Server $server)
{
  if (! auth()->user()->hasPermission($server, 'server:meta')) {
    return redirect('/servers');
  }

  return view('meta');
}
```

Into something much nicer and modern:

```php
public function meta(Server $server)
{
  redirect_unless(auth()->user()->hasPermission($server, 'server:meta'), '/servers');

  return view('meta');
}
```

These methods work by throwing a new `HttpRedirectException` (similar to the `abort*` methods) and then handling this in the exceptions `Handler`.

### DIFF
--- a/src/Illuminate/Foundation/Exceptions/Handler.php
+++ b/src/Illuminate/Foundation/Exceptions/Handler.php
@@ -15,6 +15,7 @@ use Illuminate\Contracts\Support\Responsable;
 use Illuminate\Database\Eloquent\ModelNotFoundException;
 use Illuminate\Database\MultipleRecordsFoundException;
 use Illuminate\Database\RecordsNotFoundException;
+use Illuminate\Http\Exceptions\HttpRedirectException;
 use Illuminate\Http\Exceptions\HttpResponseException;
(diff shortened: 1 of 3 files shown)


devfrey:
This already works 👇 
```php
abort_unless(auth()->user()->hasPermission($server, 'server:meta'), redirect('/servers'));
````

Also, I think a simple `if` statement is easier to understand. I wish we'd stop reinventing language constructs through global helpers.

jbrooksuk:
> ```
> abort_unless(auth()->user()->hasPermission($server, 'server:meta'), redirect('/servers'));
> ```

I'll be honest, I didn't know that worked. It makes sense though, now I've looked at it.

PR #45137, opened 2022-11-29
[9.x] Fix (widen) PHPDoc parameter typehint for `EnumeratesValues::reject()`

devfrey:
In my tests I often use `Collection::reject()` to select a random value from an enum _except_ for a specific value. Consider this example:

```php
$notAdminRole = Collection::make(Role::cases())
    ->reject(Role::Admin)
    ->random();
```

However, the PHPDoc parameter type hint for `EnumeratesValues::reject()` doesn't "allow" this (which triggers errors in PHPStan) – even though this use case is specifically covered in the tests:

https://github.com/laravel/framework/blob/9.x/tests/Support/SupportCollectionTest.php#L3803


driesvints:
This seems to break the test suite. Please resubmit this with passing tests.

PR #45144, opened 2022-11-30
Cache PDO in RefreshDatabase trait

mzur:
This caches the PDO object in the `RefreshDatabase` trait for tests.

Originally, I implemented this for my own tests to fix the ["too many connections"](https://github.com/laravel/framework/issues/18471) error which can occur if a database other than in-memory SQLite is used in the tests. But I noticed that this also substantially sped up my tests, too. I have 1152 feature tests and most of these use the database. With the default `RefreshDatabase` trait, these finish after an average 62 seconds. With the updated trait, the tests finish in an average 41 seconds, which is 65 % of the previous time!

This also may make the change from https://github.com/laravel/framework/pull/22569 obsolete but I left the diconnect there because it had no noticable negative effect on the runtime.

### DIFF
--- a/src/Illuminate/Foundation/Testing/RefreshDatabase.php
+++ b/src/Illuminate/Foundation/Testing/RefreshDatabase.php
@@ -4,10 +4,13 @@ namespace Illuminate\Foundation\Testing;
 
 use Illuminate\Contracts\Console\Kernel;
 use Illuminate\Foundation\Testing\Traits\CanConfigureMigrationCommands;
+use Illuminate\Support\Facades\DB;
 
 trait RefreshDatabase
 {
     use CanConfigureMigrationCommands;
+
+    public static \PDO $cachedPdo;
 
     /**
      * Define hooks to migrate the database before and after each test.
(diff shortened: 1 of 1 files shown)


mzur:
Somebody please rerun the failed tests. The error is unrelated to my changes.

taylorotwell:
This feels pretty risky to throw on a patch release. Feel free to add it to your own project if you find it helps.

mzur:
Fair enough! If anyone is interested, this can be done directly in the `TestCase` class as well:

```php
<?php

use App\Tests\CreatesApplication;
use Illuminate\Foundation\Testing\TestCase as BaseTestCase;

class TestCase extends BaseTestCase
{
    use CreatesApplication, RefreshDatabase;

    public static $cachedPdo;

    protected function beforeRefreshingDatabase()
    {
        if (static::$cachedPdo) {
            DB::setPdo(static::$cachedPdo);
        } else {
            static::$cachedPdo = DB::getPdo();
        }
    }
}
```

PR #45152, opened 2022-12-01
[9.x] cache repository put methond should pass correct ttl to putmany method

andiLeong:
look like the put mtthod inside cache repository class does not pass the correct ttl to the put many method 

thank you


driesvints:
This isn't correct. If you pass an array as values to the put method then the second argument becomes the ttl, not the third.

PR #45153, opened 2022-12-01
[9.x] Add Policies to Model Show Command

ahinkle:
This PR adds registered policies to the `artisan model:show` command.  This can be useful to see what policies are registered on a model.

Screen Shot 2022-11-30 at 10 09 29 PM


taylorotwell:
Since it's very atypical I think to have more than one policy per model (correct me if I'm wrong), should this be something we just put at the top ... just below "Table" maybe?

CleanShot 2022-12-01 at 09 47 14@2x

ahinkle:
Agreed. Thanks for the code change. 👍

PR #45154, opened 2022-12-01
[9.x] Auto-generate Facade Docblocks

timacdonald:
🤖 `Beep::boop()` 🤖

This PR intends to introduce a script to auto-generate / lint Facade docblocks to improve the developer experience.

This seems to be the main complaint is see in the community (from those of us that enjoy using Facades).

It's general usage is:

```
php -f bin/facade.php
```

The script will exit with a status of `0` on success or `1` on failure. In the case of a failure, the expected docblock will also be output.

I'm sure there are lots of improvements that could be made to the script, but I didn't want to overcook the thing. It moves the needle massivly improving the experience for the community and reducing docblock PRs for maintainers.

We should run the script on each PR that comes in. @driesvints will want your input on where you would like that to happen. As a new job in the static analysis CI?

## Success output

Screen Shot 2022-12-05 at 5 28 57 pm

## Failure output

Screen Shot 2022-12-05 at 5 29 23 pm

## Correctness

I'm not claiming that the docblocks generated at 100% correct, but they _should_ be 100% matching whatever the underlying code has typed. So if there is a mistake in the Facade docblock, there should be a corresponding mistake in the underlying implementation / contract docblock.

## The docblocks might lie

Because of the nature of Manager classes, it is possible that the Facade has methods on it that are not actually accessible.

Take the `Auth` facade for example: `Auth::setRememberDuration()` is only accessible on the `SessionGuard`. It is documented on the Auth facade. If you are using the token guard, that method will not be accessible.

## Template tags

For now I am resolving the template tags and attaching them to the class. In theory this should work, however I need to test this again tomorrow to see where things are falling down. Worst case, we can strip our the template tags.

## Target branch

I’m currently targeting 9.x, however we may want to consider this for the 10.x release. From what I can see this fixes things, however it does have the _potential_ to break app builds if  static analysis tooling is expecting other types (even if incorrect).

## Type expansion

I experimented with introducing https://github.com/phpDocumentor/TypeResolver into the script to ensure that classes were resolved to their FQCN representation.

However, I was able to just update the framework's types to reference FQCN. This is nice because we don't introduce a new dependency. 

This could be something introduced once we have a need for it, however.

## Question: Additional methods to include

When there are potentially varying implementations, do we want to include them all. Currently, the included methods are some-what generic. Take for example the `Cache` facade.

It includes the methods from:

1. `\Illuminate\Cache\CacheManager`
2. `\Illuminate\Cache\Repository`
3. `\Illuminate\Contracts\Cache\LockProvider`

It does not include methods from all the `LockProvider` _implementations_. Thus the `Illuminate\Cache\FileStore::getFilesystem()` method is not generated.

This gets tricky because different implementations may share the same method name, but have differing return types. For now I've included what I think makes sense and is sensible. If we feel this needs to change, we can always do that in the future anyway.


src/Illuminate/Support/Facades/Response.php
```diff
@@ -5,20 +5,20 @@
...
+ * @method static \Symfony\Component\HttpFoundation\BinaryFileResponse download(\SplFileInfo|string $file, string|null $name = null, array $headers = [], string|null $disposition = 'attachment')
+ * @method static \Symfony\Component\HttpFoundation\BinaryFileResponse file(\SplFileInfo|string $file, array $headers = [])
+ * @method static \Illuminate\Http\RedirectResponse redirectTo(string $path, int $status = 302, array $headers = [], bool|null $secure = null)
+ * @method static \Illuminate\Http\RedirectResponse redirectToRoute(string $route, mixed $parameters = [], int $status = 302, array $headers = [])
+ * @method static \Illuminate\Http\RedirectResponse redirectToAction(string $action, mixed $parameters = [], int $status = 302, array $headers = [])
+ * @method static \Illuminate\Http\RedirectResponse redirectGuest(string $path, int $status = 302, array $headers = [], bool|null $secure = null)
+ * @method static \Illuminate\Http\RedirectResponse redirectToIntended(string $default = '/', int $status = 302, array $headers = [], bool|null $secure = null)
  *
  * @see \Illuminate\Contracts\Routing\ResponseFactory
```
timacdonald, line 23: Still a contract.

nunomaduro:
@canvural Can you check the consequences of this pull request on Larastan?

canvural:
> @canvural Can you check the consequences of this pull request on Larastan?

Current test suit passes with no errors 👍🏽 We have stubs for some of the facades with just `@mixin ...` anyway. So Laravel's docblocks are ignored for those files.

As a side note latest version of PHPStorm supports auto completion for static methods just from `@mixin` tag, so all the `@method` tags on facades can be dropped. Unless support for other IDEs also wanted.

timacdonald:
I think we 100% want support for other IDEs / editors / language servers and SA tools.

timacdonald:
After some testing, it seems that generics are not supported with the @method docblock, so rather than including them and handling them "correctly" I'm opting to strip them out and replace them with `mixed`.

This does result in some `mixed|null` situations, but this just boils down to `mixed` anyway, so I believe this is the "best" solution for us right now.

.github/workflows/static-analysis.yml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 8.0
+          php-version: 8.1
```
timacdonald, line 28: @nunomaduro is there any reason we shouldn't bump the "static analysis" GitHub action for the framework to run on PHP 8.1?

I've done it here as I use some PHP 8.1 syntax in the script.

I figure this is static analysis, so it may as well run alongside this one to save introducing another job.

driesvints:
> This does result in some mixed|null situations, but this just boils down to mixed anyway, so I believe this is the "best" solution for us right now.

Can't we strip out the `|null` from `mixed|null`?

.github/workflows/static-analysis.yml
driesvints: Unsure as Laravel v9 still relies on PHP 8.0 so maybe it's best to rely on the lower bound?

Static Analysis is probably the right workflow but I'd still opt for a different job as it's checking something differently. I'll work on it.

.github/workflows/static-analysis.yml
```diff
@@ -38,3 +38,38 @@ jobs:
...
+          command: composer update --prefer-stable --prefer-dist --no-interaction --no-progress
+
+      - name: Lint facade docblocks
+        run: php -f bin/facades.php
+
+      # - name: Commit facade docblocks
+      #   uses: stefanzweifel/git-auto-commit-action@v4
+      #   with:
+      #     commit_message: Update facade docblocks
+      #     file_pattern: src/
```
driesvints, line 75: @timacdonald for some reason I thought your script also modified the DocBlocks automatically. I commented out this but if you ever opt to auto-fix them then we can use this part.

taylorotwell:
@driesvints @timacdonald so this doesn't actually modify the docblock annotations in the facade files? Is the purpose then mainly for just throwing warnings on PRs?

timacdonald:
@taylorotwell @driesvints

1. Wanted to ship fast and then iterate, so didn't see a need for it.
2. Figured we wouldn't want them to auto-commit as we don't currently do that for things like Style CI.


I'll tweak to auto-update the docblocks. We should still review all changes to the docblocks to ensure they are correct.

We should keep in mind that this may interfere with contributions, however. e.g. if someone is working on a contribution and then we auto-commit, they would have to pull before they can push again. That is fine for many of us, however I'm sure there are a lot of first-time contributors that might not know, at first, how to deal with a new commit on remote. Just a note more than anything.

timacdonald:
> > This does result in some mixed|null situations, but this just boils down to mixed anyway, so I believe this is the "best" solution for us right now.
> 
> Can't we strip out the `|null` from `mixed|null`?

Done - but only for the places were the script itself introduces these typings.

PR #45156, opened 2022-12-01
fix getInstance in Container for better performance

amirfaramarzi:
reasons:

isset is faster than is_null
Because the "instance" property is returned at the first opportunity, it improves performance

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -1379,11 +1379,11 @@ class Container implements ArrayAccess, ContainerContract
      */
     public static function getInstance()
     {
-        if (is_null(static::$instance)) {
-            static::$instance = new static;
+        if (isset(static::$instance)) {
+            return static::$instance;
         }
 
-        return static::$instance;
+        return static::$instance = new static;
     }
 
     /**


amirfaramarzi:
Hello @taylorotwell 

I have used the following package for review.
https://github.com/phpbench/phpbench

source codes:
codes

benchmark:
resualt

bert-w:
@amirfaramarzi I'd resubmit it as the following line:
```php
return static::$instance ?? static::$instance = new static;
```

Here's a benchmark (it might take a couple seconds since it does 10 iterations of the test):
```php
<?php

require __DIR__ . '/vendor/autoload.php';

class SomeClass
{
    public static $instance;

    public function getInstanceMine()
    {
        return static::$instance ?? static::$instance = new static;
    }
    
    public function getInstanceOriginal()
    {
        if (is_null(static::$instance)) {
            static::$instance = new static;
        }
        return static::$instance;
    }

    public function getInstanceYours()
    {
        if (isset(static::$instance)) {
            return static::$instance;
        }
        return static::$instance = new static;
    }

}

\Illuminate\Support\Benchmark::dd([
    'mine' => function () {
        $x = new SomeClass();
        SomeClass::$instance = null;
        for ($i = 0; $i < 10e5; $i++) {
            $x->getInstanceMine();
        }
    },
    'original' => function () {
        $x = new SomeClass();
        SomeClass::$instance = null;
        for ($i = 0; $i < 10e5; $i++) {
            $x->getInstanceOriginal();
        }
    },
    'yours' => function () {
        $x = new SomeClass();
        SomeClass::$instance = null;
        for ($i = 0; $i < 10e5; $i++) {
            $x->getInstanceYours();
        }
    },

], 10);
```
```
array:3 [
  "mine" => "373.640ms"
  "original" => "427.554ms"
  "yours" => "413.579ms"
]
```

PR #45159, opened 2022-12-01
Avoid registering multiple event listeners

dalholm:
When adding listeners dynamically in a queue worker

To simulate what will happen if you register events dynamically with queues, you can simply register events in a loop and the same issue will occur.
 
```php
for ($i = 0; $i < 5; $i++) {
    Event::listen($event, $listener);
}
```
The problem with events registers multiple times is that the listener also will fire multiple times.


driesvints:
This seems to break the test suite. Please resubmit this with passing tests. Thanks

Issue #44812, opened 2022-11-02
Can not use request() to get manual set input parameter


lvqingan:
- Laravel Version: 9.38.0
- PHP Version: 8.1.9
- Octane: 1.2.5

### Description:
Can not use request() to get manual set input parameter from a post json request.

### Steps To Reproduce:
Custom request class and manual set a extra parameter
```php
<?php

namespace App\Http\Requests;

use Illuminate\Foundation\Http\FormRequest;

class ManualRequest extends FormRequest
{
    public function rules(): array
    {
        return [];
    }

    protected function prepareForValidation()
    {
        $this->getInputSource()->set('foo', 1);
    }
}
```

```php
Route::post('manual/change', function (\App\Http\Requests\ManualRequest $request) {
    //return request()->input('foo'); // not work
    return $request->input('foo'); // it works
});
```

Is it related with #44671 ?

ankurk91:
```php
protected function prepareForValidation()
    {
        $this->merge(['foo'=> 1]);
    }
```

driesvints:
I don't think changing request params at runtime is a sensible thing to do, sorry.

florianoversmake:
We got the same problem as the original author.

Our application is relying on input manipulation for API Migrations ( renaming attributes etc. ).

I agree that manipulating input data at runtime is hacky and we wouldn't chose the same implementation again but this seems to be a **big change for a minor version change.**

The laravel documentation itself mentions "input transformations" here:
https://laravel.com/docs/9.x/requests#input-trimming-and-normalization

We will use the same approach ( early middlewares ) until we can get rid of our "API Migrations" entirely.

meyer59:
Hello
We found out our application is broken because of this change. We were relying on `prepareForValidation()` for `mergingIfMissing()` attribute with the `request()` helper.  
If you use the injected object  request `$request` this is working fine.
A few questions:
- Should we consider in the future that `request()` is not equal to `$request` ?
- Is this will be fixed ?

Thank you

meyer59:
@driesvints any consideration to fixing this ? As it seems to be an unwanted breaking change

kohlerdominik:
Hi all

I made a bugfix, and I can see now that it might be a breaking change if you relied on the behaviour of the bug. To write back parameters to the request only works for the `$json` ParameterBag. If you would have a `$query` ParameterBag or a `$request` ParameterBag, it wouldn't work. That's why I proposed to make the behaviour consistent.

---

[...]

PR #45164, opened 2022-12-02
[10.x] Create new Json ParameterBag Instance when cloning Request

kohlerdominik:
When creating a new Request instance by `createFrom`, all attributes of the request get copied to a new `InputBag`-instance ([Request:444-450](https://github.com/laravel/framework/blob/103ae1e4c83682b22f7eb6b8778e1c4e2f854d75/src/Illuminate/Http/Request.php#L444-L450)) 

The only exception is the `$json`-attribute: there, the object of the old instance (`$from`) get referenced. This means, if the ParameterBag is modified in the new instance, this change also applies in the old instance. I do not think, that this behavior is intended. For me, it was certainly unexpected.

This PR would fix this issue quite straight forward by simply cloning the ParameterBag instance. If another type of implementation is desired to fix this, please let me know.

---

This issue was merged already in 9.x in #44671. In #44812 we realized, that this is a breaking change, therefore it was reverted.

I suggest to release this change with the next major, and having a note about the change the upgrade guide. 

Applications, that break because of this change are mixing instances of `Illuminate\Http\Request` with others of `Illuminate\Foundation\Http\FormRequest`. They should consistenly either use `Illuminate\Http\Request` (which is a singleton), or pass the same instance of `Illuminate\Foundation\Http\FormRequest` forward. See detailed explaination with examples [here](https://github.com/laravel/framework/issues/44812#issuecomment-1332461961). It's important to point out, that they are not desigend to be mixed. Because of the bug, it was possible to mix them and use parameters from the `$json`-ParameterBag within each other. It didn't work for other ParameterBags.

### DIFF
--- a/src/Illuminate/Http/Request.php
+++ b/src/Illuminate/Http/Request.php
@@ -456,7 +456,7 @@ class Request extends SymfonyRequest implements Arrayable, ArrayAccess
 
         $request->setDefaultRequestLocale($from->getDefaultLocale());
 
-        $request->setJson($from->json());
+        $request->setJson(clone $from->json());
 
         if ($from->hasSession() && $session = $from->session()) {
             $request->setLaravelSession($session);


driesvints (approved):

taylorotwell:
I don't think I'm going to pursue this. It's a breaking change that potentially would affect a lot of applications in subtle ways.

Issue #45146, opened 2022-11-30
Laravel Form Validation Inconsistency


sedattcom:
- Laravel Version: 9.42.0
- PHP Version: 8.1.13
- Database Driver & Version:

### Description:
Before laravel 9 update, form rules were working as I wanted. It started behaving differently after the update.

While the same code and the same request occur without an error in laravel 8, error code 422 is returned in laravel 9.

### Steps To Reproduce:
```
Route::post('test', function (\Illuminate\Http\Request $request) {
    $request->validate([
        'type' => [
            'bail',
            'exclude_if:type,',
            'in:individual,corporate',
        ],
        'name' => [
            'bail',
            'exclude_if:type,',
            'required_with:type',
            'min:3',
            'max:49',
        ],
        'id' => [
            'bail',
            'exclude_if:type,',
            'required_with:type',
            'digits:10',
        ],
        'tax_office' => [
            'bail',
            'exclude_unless:type,corporate',
            'required_if:type,corporate',
            'min:2',
            'max:99',
        ],
    ], $request->all());
});
```

Request:
```
curl --location --request POST 'https://site.dev/api/test' \
--header 'Accept: application/json' \
--header 'X-Requested-With: XMLHttpRequest' \
--header 'Content-Type: application/json;charset=UTF-8' \
--form 'type=""' \
--form 'name=""' \
--form 'id=""' \
--form 'tax_office=""'
```

Laravel 8.21.0 response status code: 200
Laravel 9.42.0 response status code: 422

driesvints:
Hey there,

Can you first please try one of the support channels below? If you can actually identify this as a bug, feel free to open up a new issue with a link to the original one and we'll gladly help you out.

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

Thanks!

sedattcom:
Hello again,

this is definitely a bug.
After "exclude_if:type," in the first rule (in $this->data used on line 1730 in ValidatesAttributes.php in v9.42.0), the type is deleted, returning true here when checking the second rule. But here it checks the name field even though it returns true.

sedattcom:
[...]

PR #37262, opened 2021-05-04
[8.x] Fix fields not required with required_unless

dshoreman:
In v8.40, given a form with this data:
```
['foo' => true]
```

...and this rule in the FormRequest: 
```
['bar' => 'required_unless:baz,true'],
```

The short-circuit introduced in #37128 prevents 'validateRequired' from being called on *bar* when *baz* is omitted from data - even though the very lack of it might mean "bar" *should* be required.

Removing the short-circuit from `required_unless` *seems* to solve the issue without breaking anything else. I tried `false` first, but that just broke it differently so I added cases to cover that too.

PR #45167, opened 2022-12-02
 fix for #45146

sedattcom:
It is explained in detail in issue #45146.


driesvints:
I added this behavior in https://github.com/laravel/framework/pull/37128. Later on we changed this very part for `required_unless` in https://github.com/laravel/framework/pull/37262 (and that was backported to 6.x). [We also updated the docs for `required_unless`](https://laravel.com/docs/9.x/validation#rule-required-unless) to better explain the behavior. I'm not sure if we should do the same here... 

@taylorotwell what are your thoughts?

Btw: why are you sending an XMLHttpRequest against an API endpoint?

sedattcom:
XMLHttpRequest is not a value I specifically added, postman must have been left over from previous queries.

taylorotwell:
No tests.

PR #45169, opened 2022-12-02
[9.x] fix `Observer attribute changes ignored when using increment()` #45163

### LINKED ISSUE #45163: Observer attribute changes ignored when using increment()
- Laravel Version: 9 (latest)
- PHP Version: 8.2 (latest)
- Database Driver & Version: Mysql 8 (latest)

### Description:

When using `increment()`, it is not possible to have an Observer update any other values.

Not sure if this is intentional but if it is, it might be worth putting it as a caveat in the docs because I wasted quite a bit of time trying to work out what the problem was.

### Steps To Reproduce:

`$model->increment('column');`

then in the observer:

```
public function updating($model): void
{
        $model->another_column = 'example';
}
```

The `another_column` while it gets set in the Model instance, it isn't propagated in the database.

amjdhsan:
fixing `Observer attribute changes ignored when using increment()` by sync changes and appends them to $extra array before applying the changes


taylorotwell:
We would need tests.

PR #45175, opened 2022-12-02
[9.x] Remove redundant types on mixed

PerryvanderMeer:
This PR removes redundant types on `mixed` types, [since `mixed` already accepts these types](https://www.php.net/manual/en/language.types.mixed.php).


PerryvanderMeer:
Something different i came across while updating these docblocks. For nullable types, `string|null` and `?string` are used through eachother. For example:

https://github.com/laravel/framework/blob/37bbb8c299b51195fc38f845dae847726d982522/src/Illuminate/Auth/Access/HandlesAuthorization.php#L22

https://github.com/laravel/framework/blob/37bbb8c299b51195fc38f845dae847726d982522/src/Illuminate/Auth/Access/HandlesAuthorization.php#L35

I'm not sure if we want these internal inconsistencies. I'd prefer `?string` for variables with a single type, since it improves readability.

bert-w:
`array|mixed` could indicate that it's generally supposed to be an array, but other types are possible. I think we lose some clarity by making everything`mixed` only.

I think in a lot of cases it probably isn't really correct to define something as `mixed` since it probably only accepts a subset like `string|null|boolean|float|integer`, which for simplicity's sake is defined as `mixed`. So there are definitely some improvements to be made.

jmarcher:
I think it is more clear the way it is now.

PerryvanderMeer:
I agree with you both, `mixed` types would need more improvements for clarifying.
How about the [internal inconsistencies](https://github.com/laravel/framework/pull/45175#issuecomment-1336134424) for `string|null` and `?string`?

/cc @taylorotwell

PR #45178, opened 2022-12-03
[9.x] Add method missingAll to request

fernandokbs:
This PR adds the method missingAll()

I think it should be a good idea to add a method to check if all params are missing, since missing() returns true even if there is a present param.

```php
if($request->missingAll(['name','email'])) {

}
```

it adds syntactic sugar.


antonkomarev:
What is the use case?

PR #45179, opened 2022-12-03
The New String methode to get a recipient from email 

saravanasai:
This PR adds the method **Str::recipientFromEmail()** 

I think it should be a good idea to add a method to get a username from a email used to sign up . This just a quick helper to get name from email with good syntax. 

` Str::recipientFromEmail('TaylorOtwell@laravel.com')  // TaylorOtwell  `


Rizky92:
I would do it like this.
```php
use Illuminate\Support\Str;

return Str::before('username@example.com', '@');
// username
```

PR #45181, opened 2022-12-04
[10.x] Calculable expressions

bert-w:
# Calculable Expressions
Calculable expressions is a direct extension to existing database expressions (`DB::raw('...')`), of which the expected outcome can be directly applied onto the attributes of the model. This allows developers to use an improved syntax to deal with queries that have atomic calculations in them, for example:

```
`UPDATE `users` SET `counter` = `counter` + 1 WHERE `id` = 1;
```

These expressions run atomically, meaning they run as one unit. The value of `counter` in the equation is guaranteed to have the current value for that column. Running this query 1000 times in parallel will make sure that the counter is updated 1000 times correctly, whereas a PHP solution (dont do this) like `$user->update(['counter' => $user->counter + 1]);` may result in an incorrect endresult because the value `$user->counter` is retrieved from the PHP runtime, possibly referring to an outdated value.

## Current implementation
Currently, there is a solution already to deal with such queries. These solutions include:
- `Illuminate\Database\Eloquent\Model` `increment()`/`decrement()` use a raw expression internally with this syntax
- `Illuminate\Database\Eloquent\Model` `update()` with manual raw expressions like:
  -  ``$user->update(['counter' => DB::raw('`counter` + 1')]);``

However, these functions do not work nicely with setting the new attributes on a model. For instance you want the `$user->counter` variable updated after you execute a query that modifies this value. A hacky workaround that enables this is currently implemented for the `increment()`/`decrement()` functions:

https://github.com/laravel/framework/blob/1e51a5ad2f70007f7be028bc198aeb52e0264bcb/src/Illuminate/Database/Eloquent/Model.php#L959-L961

As you can see these lines only support addition and subtraction and only the "simple" case of increment/decrement (with one column and one number, i.e. `` `counter` + 1 ``).

## More complex SQL statements
By leveraging the `symfony/expression-language` package, new functionality comes available in this PR where developers can now use arbitrary SQL expressions that are automatically applied to the model attributes after the query has completed, without the need of an extra query:

```php
$user = User::create([
    'counter' => 5,
    'bonus' => 2,
]);

$user->update([
    'counter' => DB::calculate('(`counter` + `bonus`) * 2);
]);

// $user->counter === 20
```

During the `updating` and `saving` hooks, the attribute `$user->counter` will remain a `CalculableExpression` object and is sent as such to the query. Since it extends `\Illuminate\Database\Query\Expression`, the query is correctly parsed to:
```
UPDATE `users` SET `counter` = (`counter` + `bonus`) * 2 WHERE `id` = 1
```
Then, in the `updated` and `saved` hooks the calculated value will have been applied to the model's attributes, so `$user->counter` will resolve to `20`.

**The result of this expression is estimated based on the current model values** and will be calculated the moment a `CalculableExpression` is assigned to an attribute.

The expression ``(`counter` + `bonus`) * 2)`` is parsed using a regex (see code comments) so that the variables can be replaced with the current attribute values, resulting in the expression `(5 + 2) * 2`. This expression is parsed by `symfony/expression-language` (new dependency) so that no `eval()` is required. This ofcourse also supports arbitrary nesting of parentheses and a variety of operators including `+` (addition), `-` (subtraction), `/` (division), `*` (multiplication) and `%` (modulo).

## Unsupported operators
Operators like `^` (power) have not been included in the tests since they require a different SQL syntax (namely `POWER(2,2)` meaning `2^2`) and therefore don't streamline nicely with an equation parseable by `symfony/expression-language`. In the cases where these operators are required, you cannot leverage `DB::calculate()` so you will have to resort to:
```
$user->update(['counter' => DB::raw('POWER(`counter`, 2)')]);
$user->refresh();
// $user->counter === 25 (= 5^2)
```

which does not apply any sort of result to the model, unless you manually call a `$user->refresh()` like in the example.

## Other details
- Test cases are included for many different expressions. Table columns can be defined with backticks (`` ` ``) and without. A mix of these is also possible.
- Test cases are included for custom casts. In these cases, a `CalculableExpression` or `Expression` statement will always take precedence since they are meant to be sent as-is to the actual query.

## Future work
- Replace the `increment()`+`decrement()` functions to reroute to `update()` with a `DB::calculate()` statement.


src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
```diff
@@ -2059,7 +2077,7 @@ protected function transformModelValue($key, $value)
         // If the attribute exists within the cast array, we will convert it to
         // an appropriate native PHP type dependent upon the associated value
         // given with the key in the pair. Dayle made this comment line up.
-        if ($this->hasCast($key)) {
+        if (! $value instanceof Expression && $this->hasCast($key)) {
```
bert-w, line 2080: This change also makes sure that expressions work with casts (i.e. Expression objects are always used if assigned).

src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
```diff
@@ -944,6 +947,15 @@ public function setAttribute($key, $value)
             return $this->setAttributeMarkedMutatedAttributeValue($key, $value);
         }
 
+        if ($value instanceof Expression) {
+            if ($value instanceof CalculableExpression) {
+                $value->setExpectedResult($this->calculateResult($value));
```
bert-w, line 952: At property assignment (`$user->counter = DB::calculate(...)`), the expression is set as the property and if it is a Calculable Expression, the result is calculated and added to the `CalculableExpression` object so we can later assign it. The actual assignment happens later on in the routine after the query has completed (just before the `updated`/`saved`/`created` hooks).

taylorotwell:

PR #45182, opened 2022-12-04
[9.x] Fix aliasing with cursor pagination

maximepvrt:
This fixes an issue if  "as" is present in cursor pagination alias for casting a column value.

Example :
```

 \App\User::selectRaw('cast(created_at as DATE) as foo')->orderBy('foo')->cursorPaginate(1)->toArray();
```


driesvints:
Please resubmit this with tests.

PR #45187, opened 2022-12-05
[9.x] Update the changes that made within `updating` event

mtawil:
This PR is fixing this issue https://github.com/laravel/framework/issues/45163.

Here is what I've done as follows:
- Get the changes that were only made by model event `updating` and add them to the `$extra` variable.
- Sync the original attributes to both `[$column] + $extra` because it doesn't make sense that the `category` is still dirty even after reflecting it on SQL, and that's why I fixed the test assertions.

Please let me know what you think.


taylorotwell:
We're not going to change any of this behavior on a patch release, sorry.

PR #45188, opened 2022-12-05
[9.x] Fix aliasing with cursor pagination

maximepvrt:
This fixes an issue if "as" is present in cursor pagination alias for casting a column value.


taylorotwell:
Please describe the issue in detail. Mark as ready for review when done please. Thanks!

maximepvrt:
Currently the getOriginalColumnNameForCursorPagination function find the first occurrence of "as" of a dynamic column with an alias.
The correct alias is the last "as" found in a column expression.
Use explode function is bad, if multiple "as" are presents, it's not cut the dynamic column in 2.

I found this issue using cursorPaginate with an column dynamic expression that differentiated between two dates which I then cast to a float.

PR #45191, opened 2022-12-06
[9.x] Improved error logging for unmatched routes and route not found

i386:
This change adds some improved error messages so that poor client usages (for example, older versions of mobile clients) can be identified via sentry logging etc. At the moment this is impossible to do without logging every request.

The change targets the following:
* handleMatchedRoute's NotFoundHttpException
* methodNotAllowed's MethodNotAllowedHttpException

### DIFF
--- a/src/Illuminate/Routing/AbstractRouteCollection.php
+++ b/src/Illuminate/Routing/AbstractRouteCollection.php
@@ -41,7 +41,10 @@ abstract class AbstractRouteCollection implements Countable, IteratorAggregate, 
             return $this->getRouteForMethods($request, $others);
         }
 
-        throw new NotFoundHttpException;
+        throw new NotFoundHttpException(sprintf(
+            'The route %s could not be found.',
+            $request->path()
+        ));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Illuminate/Routing/AbstractRouteCollection.php
```diff
@@ -101,24 +104,26 @@ protected function getRouteForMethods($request, array $methods)
...
      *
+     * @param  \Illuminate\Http\Request  $request
      * @param  array  $others
      * @param  string  $method
      * @return void
      *
      * @throws \Symfony\Component\HttpKernel\Exception\MethodNotAllowedHttpException
      */
-    protected function methodNotAllowed(array $others, $method)
+    protected function methodNotAllowed($request, array $others, $method)
```
driesvints, line 120: This is a breaking change to the method signature.

i386: We can have two of them. One for the old behaviour and one for the new behaviour.

i386:
Recreated https://github.com/laravel/framework/pull/45206 based on feedback

PR #45193, opened 2022-12-06
added isNumeric method

ahmetbarut:
Sometimes we need to check whether the value we give is numeric or not.

```php
   str("123 333 A")->isNumeric() // bool: false
   str("123-333")->replace("-", "")->isNumeric() // bool: true
```


taylorotwell:

ahmetbarut:
Greetings, this code must be in the kernel. Why should I need to publish a package for a single method?
@taylorotwell

PR #45206, opened 2022-12-06
[9.x] Improved error logging for unmatched routes and route not found

i386:
Recreating #45191 as it was closed a little too early for me to provide an update based on feedback. 

This change adds some improved error messages so that poor client usages (for example, older versions of mobile clients) can be identified via sentry logging etc. At the moment this is impossible to do without logging every request.

The change targets the following:
* handleMatchedRoute's NotFoundHttpException
* methodNotAllowed's MethodNotAllowedHttpException


i386:
updated the change to fix the style issue but not sure why its not running again.

PR #45214, opened 2022-12-07
Support custom field names when using explicit route model binding

### LINKED ISSUE #33079: Explicit Model Binding Cannot Pass Column Name
- Laravel Version: 7.14
- PHP Version: 7.4.6
- Database Driver & Version: Irrelevant

### Description:
Using explicit model binding, I cannot pass in an optional parameter

### Steps To Reproduce:
RouteServiceProvider.php
`Route::model('page', \App\Page:class)`

web.php
`Route::post('pages/{page:slug}/publish', 'PageController@publish')->name('admin.pages.publish');`

I expect to route to use the `slug` column to fetch the page, but instead it does not pass that value through to the `\Illuminate\Routing\RouteBinding.php@forModel(...)` function. This causes the resolveRouteBinding to fallback to the key name for the model, which I do not want to change for a simple route binding.
```
    public function resolveRouteBinding($value, $field = null)
    {
        return $this->where($field ?? $this->getRouteKeyName(), $value)->first();
    }
```
Specifically to the line
```
70    if ($model = $instance->resolveRouteBinding($value)) {
71         return $model;
72    }
```
I'd expect it to be something like
```
70    if ($model = $instance->resolveRouteBinding($value, $column)) {
71         return $model;
72    }
```
with `$column = 'slug'`

As a result, I can either use explicit model binding, and change getRouteKeyName to return `slug` or I can use implicit model binding (which I don't want to because I need explicit route binding for the update and delete functions)

jacovdbergh:
Consider the following code snippet:

```php
Route::model('tenant', Tenant::class);

Route::get('/{tenant:subdomain}/show', fn (Tenant $tenant) => response($tenant->name));
```

Currently, it will try to resolve the Tenant using the `id` field, ignoring the custom `subdomain` field specified.

This pull requests fixes this behaviour to be the same as the implicit model binding behaviour, which resolves the model using the custom field name.

Fixes #33079

### DIFF
--- a/src/Illuminate/Routing/RouteBinding.php
+++ b/src/Illuminate/Routing/RouteBinding.php
@@ -55,9 +55,9 @@ class RouteBinding
      *
      * @throws \Illuminate\Database\Eloquent\ModelNotFoundException<\Illuminate\Database\Eloquent\Model>
      */
-    public static function forModel($container, $class, $callback = null)
+    public static function forModel($container, $class, $callback = null, $routeParameter = null)
     {
-        return function ($value) use ($container, $class, $callback) {
+        return function ($value) use ($container, $class, $callback, $routeParameter) {
             if (is_null($value)) {
                 return;
             }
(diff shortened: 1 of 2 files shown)


taylorotwell:
If you are going to do explicit binding I suggest you define a full custom callback and handle this in there.

georgeboot:
@taylorotwell any specific reasons not to merge this? This currently is unexpected behaviour and the fix seems to have a low impact.

Specifying the resolution logic works, but now you end up `{tenantById}` in your admin routes and `{tenentBySubdomain}` in the frontend routes etc.

Additionally, when using Livewire, you can't use `->can()` on routes, because the Livewire controller is doesn't typehint the argument type, and thus the router won't be able to bind it.

Using explicit binding fixes that because the router no longer looks at the controllers arguments.

PR #45216, opened 2022-12-07
[9.x] fix deprecation in is_file function

medeirosinacio:
Hello, I saw this warning in my application and I am sending a PR to correct it.

What I did was make sure that the given `$filename` is a string before calling the `is_file `function

`is_file(): Passing null to parameter #1 ($filename) of type string is deprecated in /app/vendor/laravel/framework/src/Illuminate/Filesystem/Filesystem.php on line 539`

image


driesvints:
@medeirosinacio you need to typecast before you pass in the value. I've sent in a PR to revert this one.

PR #45227, opened 2022-12-08
[8.x] Fix email verification request

axlon:
This PR is a redo of https://github.com/laravel/framework/pull/45226

For `hash_equals()` to effectively prevent timing attacks, the known value must be the first parameter, and the user supplied input must be the second parameter. Laravel currently does this the other way around in email verification requests, defeating the purpose of using `hash_equals()` over `===`.

This is the only instance in the framework (that I could find) where the parameter order is incorrect.


driesvints:
Laravel 8 isn't maintained anymore.

PR #45229, opened 2022-12-08
Revert "[9.x] fix deprecation in is_file function"

driesvints:
Reverts laravel/framework#45216

The docblock clearly states you can only pass a string. Like decided in the past, we won't be typecasting inside methods as this could lead to an exponential list of new PRs similar to this one.


medeirosinacio:
> Reverts #45216
> 
> The docblock clearly states you can only pass a string. Like decided in the past, we won't be typecasting inside methods as this could lead to an exponential list of new PRs similar to this one.

I understood. Thanks

PR #45231, opened 2022-12-08
[9.x] Added `whenIsJson` to Stringable.

michaelnabil230:
This PR is to add the method `whenIsJson` in the stringable class

Ex:

```php
str('[1,2,3]')->whenIsJson(function () {
    // Your code
});
```

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -1006,6 +1006,18 @@ class Stringable implements JsonSerializable
     }
 
     /**
+     * Execute the given callback if the string is Json.
+     *
+     * @param  callable  $callback
+     * @param  callable|null  $default
+     * @return static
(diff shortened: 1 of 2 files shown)


ankurk91:
Please stop adding such methods, you already do like

```php
Str::of($string)->when(function ($str) {
  //
})->value()
```

michaelnabil230:
> Please stop adding such methods, you already do like
> 
> 
> 
> ```php
> 
> Str::of($string)->when(function ($str) {
> 
>   //
> 
> })->value()
> 
> ```

‏I think it will be the last added.

michaelnabil230:
@ankurk91 ‏I think it will be the last added.

dinhquochan:
Don't add such methods. I prefer using `when()` to do this.

:-1: this PR.

PR #45232, opened 2022-12-08
Add docblock deprecated method in `InteractsWithInput`

michaelnabil230:
The function `str` is calling function `string` and makes the same functionality


taylorotwell:
No need to deprecate - it's just an alias.

PR #45233, opened 2022-12-08
[9.x] Added `ContainsWith` and `DoesntContainsWith` in validate rules

michaelnabil230:
This PR adds the ability to check if the string contains the given values, and I think this validation is important and makes the validation easier and we will not resort to using something else such as Regex.

```php
$request->validate([
    'comment' => ['required', 'contains_with:f***,s***'],
    'note' => ['required', Rule::contains(['f***', 's***'])],
]);
```

```php
$request->validate([
    'comment' => ['required', 'doesnt_contains_with:other words'],
    'note' => ['required', Rule::doesntContains(['other words'])],
]);
```


taylorotwell:
Can easily just use a custom Closure rule here.

PR #45235, opened 2022-12-08
Return 500 http error, instead of 200, when dotenv fails to load

cosminardeleanu:
When dotenv fails, the response on http is always 200.

Because of 200 response, alerts might not be triggered, and data will be cached with whatever laravel decides to output.


GrahamCampbell:
I wonder if we'd be better to just throw the exception and have things crash in the usual way.

PR #45237, opened 2022-12-08
[9.x] Improved error logging for unauthenticated routes

i386:
This change aims to improve observability through error reporting. 

When the Authenticate middleware throws a AuthenticationException it sends the method and path in the exception message.


i386:
Fixed style check

PR #45238, opened 2022-12-08
[9.x] Improved error logging for unauthorized routes

i386:
This change aims to improve observability through error reporting by logging method and path for unauthorised routes

### DIFF
--- a/src/Illuminate/Auth/Access/Response.php
+++ b/src/Illuminate/Auth/Access/Response.php
@@ -148,7 +148,11 @@ class Response implements Arrayable
     public function authorize()
     {
         if ($this->denied()) {
-            throw (new AuthorizationException($this->message(), $this->code()))
+            $message = $this->message();
+            if (!$message && request()) {
+                $message = sprintf('The action %s %s is unauthorized', request()->method(), request()->path());
+            }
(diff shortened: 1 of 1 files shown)


src/Illuminate/Auth/Access/Response.php
```diff
@@ -148,7 +148,11 @@ public function code()
     public function authorize()
     {
         if ($this->denied()) {
-            throw (new AuthorizationException($this->message(), $this->code()))
+            $message = $this->message();
```
i386, line 151: I realise this might not be the best way of going about it. Can someone more experienced offer some advice here?

i386: Would it be acceptable to use $_SERVER here?

axlon:
I feel like the previous PR made sense because those exceptions were about routes.

These exceptions have nothing to do with routing at all, for instance they could be thrown during a console command.

I think if you want to track what route an exception happens on in your logging, the best place to do it would probably be your application's exception handler, not the framework. That way exceptions don't have to be changed all over the framework and your application can add the extra context that you might need.

This also applies to #45237

taylorotwell:
I think the current code is ok.

PR #45250, opened 2022-12-09
[9.x] Add sort routes by occurrence

halloei:
I'd like to refactor my routes from one huge `web.php` file to multiple files (in a `web` subdirectory).

To ensure I'm making no mistakes, I exported the initial routes (`./artisan route:list --verbose > old.txt`) to be able to compare them against the refactored ones (`./artisan route:list --verbose > new.txt`).

Anyway, this does not protect me from unintentionally changing the order of the routes, which could cause serious bugs:
```php
Route::get('foo/{bar}', function () { return 'foo'; });
Route::get('foo/baz', function () { return 'foo/baz'; }); // This route is not reachable
```

In order to check the occurrence, this PR adds the possibility to sort by occurrence: `./artisan route:list --sort occurrence`

### DIFF
--- a/src/Illuminate/Foundation/Console/RouteListCommand.php
+++ b/src/Illuminate/Foundation/Console/RouteListCommand.php
@@ -121,8 +121,8 @@ class RouteListCommand extends Command
      */
     protected function getRoutes()
     {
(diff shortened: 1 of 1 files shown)


taylorotwell:

Issue #45254, opened 2022-12-10
Unexpected type change when renaming a column


hafezdivandari:
- Laravel Version: 9.43.0
- PHP Version: 8.0
- Database Driver & Version: MySQL 8.0

### Description:
Some unexpected type changes occurs when trying to [rename a column](https://laravel.com/docs/9.x/migrations#renaming-columns) with one of the following types:

```php
Schema::create('bars', function (Blueprint $table) {
    $table->mediumInteger('foo');
});

Schema::table('bars', function (Blueprint $table) {
    $table->renameColumn('foo', 'bar'); // changes the name but also changes its type to INT
});
```

| Column type | Expected | Actual | Notes |
| --- | --- | --- | --- |
| `timestamp('foo', 2)` | `TIMESTAMP(2)`  | `DATETIME` | Changes type and removes precision |
| `mediumInteger('foo')` | `MEDIUMINT`  | `INT` | Changes type |
| `tinyInteger('foo')` | `TINYINT`  | `TINYINT(1)` | Changes type to Boolean |
| `dateTime('foo', 2)` | `DATETIME(2)`  | `DATETIME` | Removes precision |
| `time('foo', 2)` | `TIME(2)`  | `TIME` | Removes precision |
| `year('foo')` | `YEAR`  | `DATE` | Changes type #38506 |
| `double('foo')`  | `DOUBLE(8, 2)` | `DOUBLE PRECISION`  | Removes total and places |
| `float('foo')`  | `DOUBLE(8, 2)` | `DOUBLE PRECISION`  | Removes total and places |
| `set('foo', ['bar', 'qux'])`  | `SET('bar', 'qux')` | `LONGTEXT COMMENT '(DC2Type:simple_array)'`  | Changes the type and adds a comment |
| `enum('foo', ['bar', 'qux'])`  | `ENUM('bar', 'qux')` | `VARCHAR(255)`  | Documented on Laravel docs as not supported |

#### Additional notes:
1. I just tested MySQL, but I believe there are similar issues on other databases.
1. Some modifiers are also get removed when renaming a column like `useCurrentOnUpdate` and `invisible`. I didn't check all of them.
1. The `nullable` modifier compiles to `default null` when renaming. I'm not sure if there is a difference between `INT NULL` and `INT DEFAULT NULL` on MySQL. Also when using `nullable()->default(4)` modifiers together, we expect `INT NULL DEFAULT 4` but we get `INT DEFAULT 4` instead.
[...]

PR #45258, opened 2022-12-10
[9.x] Add support for native rename/drop column commands

hafezdivandari:
Related to #45254

Currently it is required to install `doctrine/dbal` to be able to [rename a column](https://laravel.com/docs/9.x/migrations#renaming-columns) on all databases and [drop a column](https://laravel.com/docs/9.x/migrations#dropping-columns) on SQLite.
But all databases now have native support for renaming and dropping columns on their latest versions:

| Database | Supported version on Laravel | Native rename column support | Native drop column support |
| --- | --- | --- | --- |
| MariaDB | 10.3+ | 10.5.2+ ([Docs](https://mariadb.com/kb/en/alter-table/#rename-column)) | ✓ |
MySQL | 5.7+ | 8.0+ ([Docs](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html#alter-table-redefine-column)) | ✓ |
PostgreSQL | 10.0+ | ✓ ([Docs](https://www.postgresql.org/docs/10/sql-altertable.html)) | ✓  |
| SQLite | 3.8.8+ | 3.25.0+ ([Docs](https://www.sqlite.org/lang_altertable.html#alter_table_rename_column)) ([Release logs](https://www.sqlite.org/releaselog/3_25_0.html)) | 3.35.0+ ([Docs](https://www.sqlite.org/lang_altertable.html#alter_table_drop_column)) ([Release logs](https://www.sqlite.org/releaselog/3_35_0.html)) |
SQL Server | 2017+ | ✓ ([Docs](https://learn.microsoft.com/en-us/sql/relational-databases/tables/rename-columns-database-engine?view=sql-server-2017#rename-a-column)) | ✓  |

This PR adds support for native rename column command on all databases and drop column command on SQLite **without any breaking changes**.

This means if a user has already installed `doctrine/dbal`, they see no change, but for those users with a database version that supports native rename/drop column commands (as listed on the above table) there is no need to install `doctrine/dbal` anymore.

**But what if I have already installed `doctrine/dbal` and still want to use the native rename/drop column commands?**
This PR adds a new static method to `Schema` facade, `Schema::useNativeSchemaOperationsIfPossible()`. For example if you have already installed `doctrine/dbal` library but you don't want to use Doctrine for renaming/dropping columns you may call `Schema::useNativeSchemaOperationsIfPossible()` method within the `boot` method of your `App\Providers\AppServiceProvider` class (we already have similar methods [like this one](https://laravel.com/docs/9.x/migrations#index-lengths-mysql-mariadb)) or maybe inside your migration files directly.

Finally, this PR causes Laravel to support native rename/drop column commands and sees the usage of `doctrine\dbal` as a workaround for users that prefer Doctrine or have an old DB version.

PS1: I'll send a PR to add/edit Laravel docs if it merged (laravel/docs#8396).
PS2: I'll send a PR with the same approach for [changing/modifying columns](https://laravel.com/docs/9.x/migrations#modifying-columns) if this one merged (#45487).


hafezdivandari:
Thank you @driesvints, You are right, I agree that it's compilated now, but I think just checking if DBAL is installed or not is not enough, we would need some kind of config for that. do you have any suggestion?

For example we currently force user to install `doctrine/dbal` to be able to use `artisan model:show` command (Although I think it's not necessary) but when user agrees to install this package, it would cause different behavior on compiling migration commands.

driesvints:
>  it would cause different behavior on compiling migration commands.

Are you saying the behavior between using DBAL and native commands is different? If so then I think we might better hold off with this entire PR since it would be too much of a breaking change for many users. I don't think that the `Schema::useDoctrineToDropColumns` etc is the way forward.

hafezdivandari:
@driesvints it would be different behavior (breaking change) if I apply your suggestion into this PR: 
> I'd make the behavior so it detects if DBAL is installed and use that and if not that it uses the native column renaming and dropping.

What am I saying that is we need a switch/config instead of `Schema::useDoctrineToRenameColumns` or `Schema::useDoctrineToDropColumns`. For example I can simplify that to `Schema::useDoctrine`, is it better you think?

hafezdivandari:
@driesvints This PR, as is, is not a breaking change at all.

I was explaining that it would be a breaking change **IF** I apply your suggestion and remove `Schema::useDoctrineToRenameColumns` or `Schema::useDoctrineToDropColumns` from this PR.

hafezdivandari:
I just merged `Schema::useDoctrineToRenameColumns` and `Schema::useDoctrineToDropColumns` into one negated method `Schema::preventDoctrineUsageIfPossible`.

taylorotwell:
@hafezdivandari but I don't think Dries or I understand WHY it is a breaking change to remove those configuration methods. 

Just use Doctrine if it is installed and use the native methods if Doctrine is not installed. Why is that a breaking change? Please explain.

hafezdivandari:
@taylorotwell I think I was wrong using the term "braking change" here, we might see that as an "unexpected behavior". Assume this scenario after this PR, for example on MySQL 8.0:
1. I run migrations to rename a column (it compiles to native rename command as expected).
2. After a while I install DBAL, for example to be able to use `artisan model:show`.
4. I run migrations again (this time it compiles to a different query using DBAL or even an exception e.g. if there are 2 renames)
5. I have no way to get the previous migration without that config method. Isn't this an unexpected behavior?

I added that config method on schema for user to be able to control these situations regardless of DBAL availability.

I'm sorry for the confusion, please let me know if you prefer this PR without any config method.

driesvints:
But the migrations will already have run? You won't run migrations twice.

hafezdivandari:
@driesvints I agree but you once said: https://github.com/laravel/framework/pull/44101#issuecomment-1245431357
> Won't this change the behavior for apps who have already run their migrations in production?
> Their migrations which are part of their app will still be run in development. If they run them in development after this PR is merged, their development databases will be out if sync with production.

taylorotwell:
A Doctrine rename could be "different" than a native one since on SQLite I'm pretty sure Doctrine drops the column and re-adds it with a new name after storing the data in a temporary table?

The final outcome should be the same but the road taken to get there is different. Is that what you mean @hafezdivandari?

hafezdivandari:
@driesvints @taylorotwell 

> Why would there be two renames? What exception would occur?

Doctrine can not handle multiple rename/drop, check this tests please but SQLite has no problem with multiple native rename/drop: https://github.com/laravel/framework/blob/26c476db8aed75dbe390fd43ea84fa9a08a97691/tests/Database/DatabaseSchemaBlueprintIntegrationTest.php#L465-L506

> A Doctrine rename could be "different" than a native one since on SQLite I'm pretty sure Doctrine drops the column and re-adds it with a new name after storing the data in a temporary table?

Yes, exactly, and also on MySQL the doctrine renames column using `alter table change column` instead of `alter table rename column` that causes this issue: #45254

Finally is it that bad to use a single config method to control such behavior changes? We are using many config methods on Eloquent Model for example.

taylorotwell:
Renamed to `useNativeSchemaOperationsIfPossible`.

EdgarSedov:
Sorry, could u please elaborate further, what means 10.0- version for Rename column support for postgres? Does that mean what i can rename natively without doctrine only for version 10.0 and below? Kinda confused here

hafezdivandari:
@EdgarSedov sorry for confusion, I wanted to declare that the PostgeSQL and SQL server support native renaming **even** before that version. It was just a simple comparison to current supported DB version on Laravel. It's corrected now.

As stated on the updated [Laravel docs on renaming columns](https://laravel.com/docs/9.x/migrations#renaming-columns) you  won't need to install DBAL on PostgreSQL at all.

IP-Developer:
Will this change also fix those long-standing issues with renaming columns when an enum column is present in the table? https://github.com/laravel/framework/issues/1186

IP-Developer:
@hafezdivandari in this case I guess the same code can be reused in some other places were the enum issue still persists (like db:show command etc.)

hafezdivandari:
@IP-Developer True. I'm working on it, waiting for this to be reviewed #45598 as a prerequisite.

IP-Developer:
@hafezdivandari those are good news for the Laravel community :) Thanks for your work

PR #45264, opened 2022-12-11
[9.x] Granular notifications queue connections

JamesHemery:
Implement viaConnections (see #45262 suggestion)

### DIFF
--- a/tests/Notifications/NotificationSenderTest.php
+++ b/tests/Notifications/NotificationSenderTest.php
@@ -189,7 +189,7 @@ class DummyNotificationWithViaConnections extends Notification implements Should
     public function viaConnections()
     {
         return [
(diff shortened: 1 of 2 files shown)


tests/Notifications/NotificationSenderTest.php
```diff
@@ -138,7 +161,7 @@ public function via($notifiable)
     }
 }
 
-class DummyNotificationWithDatabaseVia extends Notification
+class DummyNotificationWithDatabaseVia extends Notification implements ShouldQueue
```
driesvints, line 164: You can't change existing tests like this.

JamesHemery: Oups, fail during my tests.

PR #45267, opened 2022-12-12
Correct PHPDoc for the HTTP-client transferStats property

AJenbo:
IDEs and analyzers are unable to determine the type of the property. Correcting the PHPDoc provides better feedback for developers. This is a simple change to a comment block and should not break anything.

Failed test not related to these PR changes, see #45251


driesvints:
Please fix the related facade docblock as well (see failing build).

driesvints:
Thanks!

PR #45271, opened 2022-12-12
[9.x] Added updateBlanks function for models

LTKort:
Regarding the discussion #45230 

You can update a model with an array that only fills the fields that do not have a value, 

So for example I have user model that has these fields;
| id | first_name | last_name | email                |  
|----|------------|-----------|----------------------|
| 1  |            | Kort      | lars@marshmallow.dev |  

And I have an array like; 
```php
 $form_data = [
    'first_name' => 'Lars',
    'last_name' => 'Ananas',
    'phone_number' => '1234567890',
];
```

If you call the `$user->update($form_data)` function with the data, it will overwrite the last_name & give an error on phone_number.

So now I propose a function like `$user->updateBlanks($form_data)`.
Which will check if the attributes exists in the model & updates only the values of the empty keys. Which will result in:
| id | first_name | last_name | email                |  
|----|------------|-----------|----------------------|
| 1  | Lars      | Kort      | lars@marshmallow.dev |  

In addition I'm working on a `$user->fillBlanks()`, which will only fill the fields & not save them:
[Added fillBlanks #45272 ](https://github.com/laravel/framework/pull/45272). 
I'm not sure about the locations of the functions, which are now located in `Illuminate\Database\Eloquent\Model.php`

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1038,7 +1038,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
             return false;
         }
 
-        $attributes = collect($attributes)
+        $attributes = (new Collection($attributes))
             ->filter(fn ($value, $key) => Arr::has($this->attributesToArray(), $key))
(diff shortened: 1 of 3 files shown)


WendellAdriel:
Nice addition!

PR #45272, opened 2022-12-12
[9.x] Added updateBlanks function (extension of #45271)

LTKort:
Regarding the discussion https://github.com/laravel/framework/discussions/45230
And extension of the PR #45271 

You can update / fill a model with an array that only fills the fields that do not have a value, 

So now I propose a function like `$model->updateBlanks($attributes)` & `$model->fillBlanks($attributes)`
Which will check if the attributes exists in the model & updates or fills only the values of the empty keys. 

~~I moved the tests to a separate location, which I don't think is correct, but they are working.~~

**Edit**: I moved & fixed the tests

Only not sure about the location of the functions `updateBlanks()` & `fillBlanks()`, which are both located in `Illuminate\Database\Eloquent\Model.php`


WendellAdriel:
Nice addition!

src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -546,6 +546,22 @@ public function fill(array $attributes)
...
+     * Fill only the missing with an array attributes of the model.
+     *
+     * @param  array  $attributes
+     * @return $this
+     */
+    public function fillMissing(array $attributes)
+    {
+        $attributes = collect($attributes)
+            ->filter(fn ($value, $key) => Arr::has($this->attributesToArray(), $key))
+            ->reject(fn ($value, $key) => filled($this->$key))
```
jmarcher, line 559: Helper functions inside the Database namespace is not good

src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -546,6 +546,22 @@ public function fill(array $attributes)
         return $this;
     }
 
+    /**
+     * Fill only the missing with an array attributes of the model.
+     *
+     * @param  array  $attributes
+     * @return $this
+     */
+    public function fillMissing(array $attributes)
```
jmarcher, line 555: Maybe `fillBlanks` is a better name as we have the filled() & blank() helpers, this way we can follow a "convention"

src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -546,6 +546,22 @@ public function fill(array $attributes)
...
+    {
+        $attributes = collect($attributes)
```
jmarcher, line 557: Do not use helpers here `new Collection` is better

src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -1025,6 +1041,22 @@ public function updateQuietly(array $attributes = [], array $options = [])
...
     }
 
+    /**
+     * Update only the missing values of the model in the database.
+     *
+     * @param  array  $attributes
+     * @param  array  $options
+     * @return bool
+     */
+    public function updateMissing(array $attributes = [], array $options = [])
```
jmarcher, line 1051: Maybe `updateBlanks` is a better name as we have the filled() & blank() helpers, this way we can follow a "convention"

src/Illuminate/Database/Eloquent/Model.php
LTKort: Thanks, I updated the name to `fillBlanks`

src/Illuminate/Database/Eloquent/Model.php
LTKort: Thanks, I updated the name to `updateBlanks`

LTKort:
@jmarcher  I unintentionally closed the PR, reopened it again with changing the names from `Missing` to `Blanks`

src/Illuminate/Database/Eloquent/Model.php
LTKort: @jmarcher What would be a better solution? 
This is also used in the `replicate` function, should it be changed there too?

taylorotwell:
Hey there - I think this might be something better off defined in a base model class in your own application where you can put your own helpers, etc.

PR #45275, opened 2022-12-12
[9.x] Auto fix docblock changes

driesvints:
Right now, builds on 9.x are failing because a PR was merged without updating docblocks. This is because of the new Facade DocBlocks job that was merged here: https://github.com/laravel/framework/pull/45154. Because of this all new PR's will now fail.

I don't think it's feasible that we keep this linting job as people will not update these docblocks themselves. Instead we should automate this. This new PR changes the job to only run on the Laravel organization and on the latest stable branches so not on PR's. People can change code as they want and when it gets merged, this GHA will automatically update DocBlocks accordingly. 

This PR removes the GHA for PR's but I think it's a good thing as contributors now will not have to worry anymore about these changes/failures.

### DIFF
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -47,6 +47,8 @@ jobs:
 
     name: Facade DocBlocks
 
+    if: ${{ github.repository_owner == 'laravel' && (github.ref_name == 'master' || github.ref_name == '9.x') }}
+
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/static-analysis.yml
```diff
@@ -47,6 +47,8 @@ jobs:
 
     name: Facade DocBlocks
 
+    if: ${{ github.repository_owner == 'laravel' && (github.ref_name == 'master' || github.ref_name == '9.x') }}
```
driesvints, line 50: I wish I could have used a regex here to update the branch checking to `*.x` but I'm unsure how to do this in an if statement: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif

driesvints:
Marking this as ready.

taylorotwell:
We have a slight problem with the script I mentioned in our internal Slack we would need to look at before merging this.

timacdonald:
I've fixed the issue in https://github.com/laravel/framework/pull/45352

I'm happy to automate this further if it is a blocker. Wasn't sure how much it would be triggered, but now we know it is pretty frequently! Enough to be a pain.

I'll try and keep an eye on the auto-commits still, as there could be some more edge-cases or introduction of new Generics that we need to manually address.

timacdonald (approved):

driesvints:
Thanks @timacdonald!

PR #45277, opened 2022-12-12
[9.x] Implementing `clear` method for collection

megasteve19:
Hello, I think `clear` method would be very useful for collections. The main reason we need this method is `readonly` properties, without internal clear method I guess It's impossible to clear readonly collection. There might be a ways to clear collection but I don't think It would be convinient as `clear` method.

It's my first contribution to a open source code base. Sorry if I have done something wrong.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -160,6 +160,18 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
         return new static(Arr::collapse($this->items));
     }
 
+	/**
+	 * Clear all items in the collection.
+	 * 
+	 * @return $this
+	 */
+	public function clear()
+	{
+		$this->items = [];
+
+		return $this;
+	}
+
     /**
      * Determine if an item exists in the collection.
      *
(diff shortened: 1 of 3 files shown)


taylorotwell:
Hey there - I would just make it not read only in this case. Adding a `clear` would defeat the purpose of making it read only in the first place I think so I'm not sure I see the difference? Let me know if I'm wrong!

megasteve19:
@taylorotwell Yes you are kinda of rigth, but if `readonly` worked as you said. What I want to achieve is something like Vue does;

```js
const foo = ref();
```

`foo` can not be assigned again, but still containing value can be modified.

A read only property can not be assigned again but you can still manipulate the object it self. If I would like to strictly make `readonly` the array, I could have been used plain `array` to do it. So the user of the class can collect them selfs. But I don't think It's a good idea to do it.

Also there is a still problem if I make it public and without readonly;

```php
<?php

class SomeClass
{
	public Collection $collection;

}
```

We can assign any class to $collection that extends the Collection class. That's not a behaviour I want to allow.

As I said previously, there might be ways to do it another way, but another way is not convinient as `clear`.

To sum it up, single instance, `public` `readonly` property, no getter and setter and we should be able to still manipulate collection.

Thanks for taking your time 😃

taylorotwell:
Hi there - I would just suggest rejecting all items. 🙂

CleanShot 2022-12-13 at 16 06 22@2x

taylorotwell:
Hmm, I think I will hold off on this for now. You can always macro it into your application though!

```php
Collection::macro('clear', function () {
    $this->items = [];

    return $this;
});

$collection = collect([1, 2, 3]);

$collection->clear();
```

PR #45279, opened 2022-12-12
Specify the table's primary key as a UUID or a ULID

talelmishali:
This PR adds two new methods to specify the table's primary key as a UUID or a ULID.

#### Before
```php
$table->uuid('id')->primary();
// OR
$table->ulid('id')->primary();
```

#### After
```php
$table->uuidPrimary();
// OR
$table->ulidPrimary();
```

If this PR would merge, I suggest reflecting it on the UUID/ULID section of the docs, where it instructs the user to ensure the model has an adequate primary key column.

#### References
[UUID & ULID Keys](https://laravel.com/docs/9.x/eloquent#uuid-and-ulid-keys)
>  Of course, you should ensure that the model has a [UUID equivalent primary key column](https://laravel.com/docs/9.x/migrations#column-method-uuid):

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -1514,6 +1514,17 @@ class Blueprint
     }
 
     /**
+     * Specify the primary key for the table as UUID.
+     *
+     * @param  string  $column
+     * @return void
+     */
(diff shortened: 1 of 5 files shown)


taylorotwell:
I think the before code is fine. 😇

talelmishali:
You are the boss 😅; thanks!

PR #45281, opened 2022-12-12
[9.x] Add support for native column modifying

hafezdivandari:
Moved to 10.x
* #45487


morloderex:
@hafezdivandari It's it a pretty big breaking change sense i would assume that just changing the coumn type would keep any other attributes as it currently does when using doctrine?

hafezdivandari:
@morloderex It won't be a breaking change because it only applies to those who doesn't use DBAL currently or has called `Schema::useNativeSchemaOperationsIfPossible` config method explicitly.

morloderex:
> @morloderex It won't be a breaking change because it only applies to those who doesn't use DBAL currently or has called `Schema::useNativeSchemaOperationsIfPossible` config method explicitly.

@hafezdivandari okay, but what would happen if I already have doctrine installed and then I decide to drop it from my project.

Isn't it a breaking change in that regard or would that be considered unexpected behavior?

hafezdivandari:
@morloderex it's not a breaking change nor an **unexpected** behavior in my opinion, you would expect a change when you add/drop a package. if you want to see an exception when DBAL is not available, like before this PR, you may call `Schema::useNativeSchemaOperationsIfPossible(false)` .

src/Illuminate/Database/Schema/Blueprint.php
```diff
@@ -243,13 +256,7 @@ public function addFluentCommands(Grammar $grammar)
...
         foreach ($this->columns as $column) {
             foreach ($grammar->getFluentCommands() as $commandName) {
-                $attributeName = lcfirst($commandName);
-
-                if (! isset($column->{$attributeName})) {
-                    continue;
-                }
-
-                $value = $column->{$attributeName};
+                $value = $column->{lcfirst($commandName)};
```
taylorotwell, line 259: Please explain the reason for this change. Seems breaking.

src/Illuminate/Database/Schema/Blueprint.php
```diff
@@ -188,6 +197,10 @@ protected function commandsNamed(array $names)
      */
     protected function addImpliedCommands(Grammar $grammar)
     {
+        if ($this->hasAutoIncrementColumn() && ! $this->creating()) {
+            array_unshift($this->commands, $this->createCommand('autoIncrementStartingValues'));
+        }
+
```
taylorotwell, line 203: Please explain this change.

src/Illuminate/Database/Schema/Blueprint.php
```diff
@@ -1794,7 +1801,7 @@ public function getChangedColumns()
      */
     public function hasAutoIncrementColumn()
     {
-        return ! is_null(collect($this->getAddedColumns())->first(function ($column) {
+        return ! is_null(collect($this->columns)->first(function ($column) {
```
taylorotwell, line 1804: Why no longer using `getAddedColumns`?

src/Illuminate/Database/Schema/Grammars/SQLiteGrammar.php
```diff
@@ -727,7 +728,11 @@ protected function typeTimeTz(Fluent $column)
      */
     protected function typeTimestamp(Fluent $column)
     {
-        return $column->useCurrent ? 'datetime default CURRENT_TIMESTAMP' : 'datetime';
+        if ($column->useCurrent) {
+            $column->default(new Expression('CURRENT_TIMESTAMP'));
+        }
+
+        return 'datetime';
```
taylorotwell, line 735: Please explain need for this change.

src/Illuminate/Database/Schema/Blueprint.php
hafezdivandari: It's to handle auto-increment starting value using `from` modifier, before this PR `from` was only supported when adding a column and creating a table. I removed this from `compileAdd` and add it here to handle both add and change. now it is supported to use `from` modifier with `change`.

src/Illuminate/Database/Schema/Blueprint.php
hafezdivandari: It now supports both added and modified columns (looking for `from` modifier on all column`. `$this->columns = $this->getAddedColumns() + $this->getChangedColumns()`

src/Illuminate/Database/Schema/Grammars/SQLiteGrammar.php
hafezdivandari: to be consistent with other Grammars, I changed SQLite too to use the same syntax for handling `useCurrent`.

src/Illuminate/Database/Schema/Blueprint.php
hafezdivandari: It's not breaking, many tests would fail if it was. I removed this `if` condition from here and added to L132-L140 of this file. Because unfortunately this method doesn't have `$connection` argument. I didn't want to change the signature of this method (I'm not quit sure if passing the `$connection` to this method and changing it's signature is a breaking change). Please let me know if the comment on L135-L137 is clearly explains why this change is needed.

hafezdivandari: TLDR; it's removed from here and added (with more conditions) to its parent method.

hafezdivandari: It's for handling `comment` on PostgreSQL and `default` on SQL Server`.

src/Illuminate/Database/Schema/Grammars/SQLiteGrammar.php
hafezdivandari: there are 2 different statements to change column type and default value on PostgreSQL and SQL server.

src/Illuminate/Database/Schema/Blueprint.php
hafezdivandari: Actually it's much cleaner to handle auto-increment starting value as a fluent command for MySQL and PostgreSQL. but I didn't want to change a lot of code at once and make it hard to review.

hafezdivandari:
@taylorotwell I think it is much cleaner to send this PR to master instead, the `Blueprint` and `Grammar` classes need some cleanup after years, do you agree? if yes, please let me know if you prefer series of little PRs or all changes at one PR?

taylorotwell:
Sure, you can send a single PR to master.

PR #45285, opened 2022-12-13
[9.x] Fix order of array_keys() vs Arr::except() to exclude modelKey

dbakan:
The previous implementation didn't actually exclude the modelKey because `Arr::except()` works on array keys, not values.


taylorotwell:
Hmm, can you submit a test that was failing before your fix but now works?

dbakan:
Turns out there are other errors when doing something as wild as `$posts->loadCount('comments as id');`.

PR #45289, opened 2022-12-13
[10.x] Clean up DBAL deprecations

driesvints:
This PR fixes some deprecations from the move to Doctrine DBAL v3 (since we're dropping support for DBAL v2). Most likely these will be removed in Doctrine v4. There's still a couple of left which I haven't determined the right path to migrate to yet. Left some notes on other quirks I found below.

### DIFF
--- a/src/Illuminate/Database/Connectors/Connector.php
+++ b/src/Illuminate/Database/Connectors/Connector.php
@@ -66,18 +66,6 @@ class Connector
     }
 
     /**
-     * Determine if the connection is persistent.
-     *
-     * @param  array  $options
-     * @return bool
-     */
-    protected function isPersistentConnection($options)
-    {
-        return isset($options[PDO::ATTR_PERSISTENT]) &&
-               $options[PDO::ATTR_PERSISTENT];
-    }
-
-    /**
      * Handle an exception that occurred during connect execution.
      *
      * @param  \Throwable  $e
--- a/src/Illuminate/Database/Connectors/Connector.php
+++ b/src/Illuminate/Database/Connectors/Connector.php
@@ -2,7 +2,6 @@
 
 namespace Illuminate\Database\Connectors;
 
-use Doctrine\DBAL\Driver\PDOConnection;
 use Exception;
 use Illuminate\Database\DetectsLostConnections;
 use PDO;
@@ -63,10 +62,6 @@ class Connector
      */
     protected function createPdoConnection($dsn, $username, $password, $options)
     {
-        if (class_exists(PDOConnection::class) && ! $this->isPersistentConnection($options)) {
-            return new PDOConnection($dsn, $username, $password, $options);
-        }
-
         return new PDO($dsn, $username, $password, $options);
     }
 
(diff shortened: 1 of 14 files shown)


src/Illuminate/Database/Connectors/Connector.php
```diff
@@ -63,10 +62,6 @@ public function createConnection($dsn, array $config, array $options)
      */
     protected function createPdoConnection($dsn, $username, $password, $options)
     {
-        if (class_exists(PDOConnection::class) && ! $this->isPersistentConnection($options)) {
-            return new PDOConnection($dsn, $username, $password, $options);
-        }
-
```
driesvints, line 69: This was the weirdest quirk I found. This part is checking for `Doctrine\DBAL\Driver\PDOConnection` which doesn't exists anymore in Doctrine DBAL v3 (it's `Doctrine\DBAL\Driver\PDO\Connection`). So this has never worked when people installed Doctrine DBAL v3 (on Laravel v9). Given that we never have gotten reports about it I feel like it's safe to remove this part.

src/Illuminate/Database/Events/NoPendingMigrations.php
```diff
@@ -2,7 +2,9 @@
 
 namespace Illuminate\Database\Events;
 
-class NoPendingMigrations
+use Illuminate\Contracts\Database\Events\MigrationEvent as MigrationEventContract;
+
+class NoPendingMigrations implements MigrationEventContract
```
driesvints, line 7: For some reason this event didn't have the `MigrationEvent` contract which seemed odd to me. I can't think of a reason why it shouldn't have this.

tests/Database/DatabaseSchemaBlueprintIntegrationTest.php
```diff
@@ -164,7 +164,7 @@ public function testChangingColumnWithCollationWorks()
             [
                 'CREATE TEMPORARY TABLE __temp__users AS SELECT age FROM users',
                 'DROP TABLE users',
-                'CREATE TABLE users (age INTEGER NOT NULL COLLATE "NOCASE")',
+                'CREATE TABLE users (age INTEGER NOT NULL)',
```
driesvints, line 167: I guess these now will take the default collation.

tests/Database/DatabaseSchemaBlueprintIntegrationTest.php
```diff
@@ -284,11 +284,6 @@ public function testAddUniqueIndexWithoutNameWorks()
 
         $expected = [
             [
-                'CREATE TEMPORARY TABLE __temp__users AS SELECT name FROM users',
-                'DROP TABLE users',
-                'CREATE TABLE users (name VARCHAR(255) DEFAULT NULL COLLATE "BINARY")',
-                'INSERT INTO users (name) SELECT name FROM __temp__users',
-                'DROP TABLE __temp__users',
```
driesvints, line 291: It seems DBAL is now smarter here and doesn't creates temporary tables anymore when only a unique constraint or an index is added to the table.

PR #45290, opened 2022-12-13
WIP: [9.x] Adding a "unixtime" Eloquant type and cast

jkoop:
This PR adds the `$table->unixTime();` migration table method and `"unixtime"` cast to Models.

This doesn't break any existing features. `$table->timestamps();`, `"datetime"`, `"timestamp"`, etc. still behave as they did.

MySQL (and by extension, MariaDB) does not yet support [>2038-01-19 03:14:07 UTC](https://en.wikipedia.org/wiki/Year_2038_problem), and it is slowly but increasingly becoming a problem. Many developers have resorted to using `BIGINT` and writing their own casts (including myself) to store datetimes beyond 2038.

This PR adds a convenient migration table method and an Eloquent cast to make this easier out-of-the-box.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -107,6 +107,7 @@ trait HasAttributes
         'real',
         'string',
         'timestamp',
+        'unixtime',
     ];
 
     /**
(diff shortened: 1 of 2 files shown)


bert-w:
Isn't the `timestamp` cast already a unix timestamp cast?

taylorotwell:

jkoop:
For anyone who finds this, you can just add this to your Models to make them send and accept unix timestamps to and from the database:

```php
protected $dateFormat = 'U';
```

(this is what happens when you don't read [the docs](https://laravel.com/docs/9.x/eloquent#timestamps))

PR #45297, opened 2022-12-14
Add session directive

MrMohamedAbdallah:
With `@session` we can replace `if(session()->has('...'))`, which makes the code more readable.

For example:

```blade
<!-- Before -->
@if(session()->has('success'))
<div class="...">{{ session()->get('success') }}</div>
@endif

<!-- After -->
@session('success')
<div class="...">{{ $message }}</div>
@endsession
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,10 +18,10 @@ services:
       MYSQL_DATABASE: "forge"
       MYSQL_ROOT_HOST: "%"
     ports:
-      - "3306:3306"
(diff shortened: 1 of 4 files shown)


taylorotwell:
You can just do:

```php
@if (session('success'))
    {{ session('success') }}
@endif
```

You don't have to use `has` and `get`.

MrMohamedAbdallah:
Thanks for letting me know that 👍

PR #45304, opened 2022-12-14
[9.x] Fix condition in AsPivot::hasTimestampAttributes()

webberwu:
## Description

When the pivot table only has a `created_at` column but not a  `updated_at` column,  the `updated_at` column will appear in the SQL query generated by executing `attach()`.

### Post model
```php
public function tags()
{
    return $this
        ->belongsToMany(
            Tag::class,
            'post_tag',
            foreignPivotKey: 'post_id',
            relatedPivotKey: 'tag_id'
        )
        ->using(PostTagPivot::class)
        ->withPivot('created_at');
}
```

### PostTagPivot model
```php
class PostTagPivot extends Pivot
{
    protected $table = 'post_tag';

    public $timestamps = false;

    protected $casts = [
        'created_at' => 'timestamp',
    ];
}
```

### Function call
```php
$post->tags()->attach($tag)
```

get the error as the following:
```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'updated_at' in 'field list' (SQL: insert into `post_tag` (`post_id`, `tag_id`, `created_at`, `updated_at`) values (1, 22, 2022-12-14 09:32:31, 2022-12-14 09:32:31))
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php
@@ -222,7 +222,10 @@ trait AsPivot
      */
     public function hasTimestampAttributes($attributes = null)
     {
-        return array_key_exists($this->getCreatedAtColumn(), $attributes ?? $this->attributes);
+        $attributes ??= $this->attributes;
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
I think because this could be a breaking change for now I would suggest you put both columns on your pivot model.

PR #45306, opened 2022-12-14
[9.x] Fix schedule:list crash when call() is given class-string

odinns:
When creating schedule of the form:

```php
$this->schedule->call(FooCall::class);
$this->schedule->call([FooCall::class, 'fooFunction']);
```

Then the `ScheduleListCommand::getClosureLocation()` method would crash with:

```
TypeError

Cannot use "::class" on value of type string
```

This PR fixes it by adding checks for string `$callback` and returning that string instead of adding `::class` to it.

Tests for both single string and array callable syntax included.

### DIFF
--- a/src/Illuminate/Console/Scheduling/ScheduleListCommand.php
+++ b/src/Illuminate/Console/Scheduling/ScheduleListCommand.php
@@ -225,11 +225,9 @@ class ScheduleListCommand extends Command
         }
 
         if (is_array($callback)) {
-            if (is_string($callback[0])) {
-                return sprintf('%s::%s', $callback[0], $callback[1]);
-            }
+            $className = is_string($callback[0]) ? $callback[0] : $callback[0]::class;
 
-            return sprintf('%s::%s', $callback[0]::class, $callback[1]);
+            return sprintf('%s::%s', $className, $callback[1]);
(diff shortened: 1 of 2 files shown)


odinns:
Thank you to @lukeraymonddowning and @olivernybroe for help with the solution! 😊👍

src/Illuminate/Console/Scheduling/ScheduleListCommand.php
```diff
@@ -220,7 +220,15 @@ private function getClosureLocation(CallbackEvent $event)
             );
         }
 
+        if (is_string($callback)) {
+            return $callback;
+        }
+
         if (is_array($callback)) {
+            if (is_string($callback[0])) {
```
lukeraymonddowning, line 228: Perhaps rather than repeating `sprintf` we could extract a variable?

```php
if (is_array($callback)) {
  $className = is_string($callback[0]) ? $callback[0] : $callback[0]::class;

  return sprintf('%s::%s', $className, $callback[1]);
}
```

odinns: Good point @lukeraymonddowning! - Change made now 👍

PR #45307, opened 2022-12-14
Fix SerializesAndRestoresModelIdentifiers trait with model repository

clementbirkle:
Hi,

Currently, the trait "SerializesAndRestoresModelIdentifiers" via "SerializesModels" doesn't support model repository like that:

```
<?php

namespace App\Mail;

use App\Models\Group;
use Illuminate\Bus\Queueable;
use Illuminate\Container\Container;
use Illuminate\Mail\Mailable;

class UserInvitation extends Mailable
{

    protected readonly Group $groupRepository;

    /**
     *
     * @return void
     */
        $this->groupRepository = Container::getInstance()->make(Group::class);
    }
}
```

This PR fix that.


taylorotwell:

Thanks!

PR #45309, opened 2022-12-14
[9.x] bug fix, change `array_merge` to `array_replace` to prevent reindex

kevindees:
Fixes issue where array [506 => null, 'name' => 'kevin'] is reindexed as [0 => null, 'name' => 'kevin']

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1119,7 +1119,7 @@ trait HasAttributes
     {
         $caster = $this->resolveCasterClass($key);
 
-        $this->attributes = array_merge(
+        $this->attributes = array_replace(
             $this->attributes,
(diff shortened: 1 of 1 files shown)


kevindees:
This fix does not provide any new functionality. The primary purpose is to prevent confusion when a database column has a numeric value that will be reindexed by PHP's `array_merge`. Adds:

1. Quality of life.
2. Removes unexpected/buggy behavior. 

Example, previous behavior: array `[506 => null, 'name' => 'kevin']` is reindexed as `[0 => null, 'name' => 'kevin']`. New behavior, array `[506 => null, 'name' => 'kevin']` results in `[506 => null, 'name' => 'kevin']`.

taylorotwell:
Hey there - so your problem is caused by having a column in your database with a purely numeric name? Is that correct?

kevindees:
@taylorotwell 

That is correct. I know numeric names are not accessible at the object level, but our legacy DB is used a numeric column name. If anything removing the reindex allows the issue to bubble to a more user-land-accessible level.

PR #45313, opened 2022-12-14
[9.x] Fixed errors occurring when encrypted cookies has been tampered with

vbezaras-leasingmarkt:
Dear maintainers,

we have faced several errors, while receiving seemingly malicious requests with tampered cookie data. 
`ErrorException: hash_equals(): Expected user_string to be a string, array given`
`ErrorException: base64_decode() expects parameter 1 to be string, array given`
`ErrorException: Array to string conversion`

After investigation, we came to a conclusion, that the validation of encrypted payload is insufficient, i.e. not checking if the value is scalar. In this PR I have extended the validation and added a test to cover the case.

As per requirement I send the PR to 9.x branch, but I would greatly appreciate it the fix would be backported to version 8.x. I can not estimate if this poses a significant security threat, yet I am concerned, that this behavior can be abused in some way.

### DIFF
--- a/src/Illuminate/Encryption/Encrypter.php
+++ b/src/Illuminate/Encryption/Encrypter.php
@@ -229,10 +229,24 @@ class Encrypter implements EncrypterContract, StringEncrypter
      */
     protected function validPayload($payload)
     {
-        return is_array($payload) && isset($payload['iv'], $payload['value'], $payload['mac']) &&
-            strlen(base64_decode($payload['iv'], true)) === openssl_cipher_iv_length(strtolower($this->cipher));
-    }
+        if (!is_array($payload)) {
+            return false;
+        }
 
+        // Required values, must be strings
+        foreach (['iv', 'value', 'mac'] as $item) {
+            if (!isset($payload[$item]) || !is_string($payload[$item])) {
+                return false;
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


vbezaras-leasingmarkt:
Oh sorry, it went so fast :) I forgot to commit a little fix to the test:

```
diff --git a/tests/Encryption/EncrypterTest.php b/tests/Encryption/EncrypterTest.php
index c989a5d9f1..b31e8584a9 100755
--- a/tests/Encryption/EncrypterTest.php
+++ b/tests/Encryption/EncrypterTest.php
@@ -209,10 +209,9 @@ public function provideTamperedData()
     {
         return [
             [['iv' => ['value_in_array'], 'value' => '', 'mac' => '']],
-            [['iv' => '', 'value' => '', 'mac' => '']],
             [['iv' => '', 'value' => ['value_in_array'], 'mac' => '']],
             [['iv' => '', 'value' => '', 'mac' => ['value_in_array']]],
-            [['iv' => '', 'value' => '', 'mac' => ['value_in_array'], 'tag' => ['value_in_array']]],
+            [['iv' => '', 'value' => '', 'mac' => '', 'tag' => ['value_in_array']]],
         ];
     }
```

So that each of the 4 values would be tested separately.

PR #45318, opened 2022-12-15
Fix Parameter Type Doc

xuanyanwow:
```php
ModelName::query()
            ->when($start, fn ($query, $start) => $query->where('id', '<', $start));
```

If `$start` is string or int , IDE give me a error tips: `should be \Closure|null`.


driesvints:
Static analysis seems to be failing here.

PR #45321, opened 2022-12-15
[9.x] Update EnumeratesValues docblocks to indicate nullable keys

abenerd:
When filtering based on the immediate elements in the array one must provide a key of `null` , this change updates the docblocks show this is allowed.


taylorotwell:
Can you give an example of this?

abenerd:
@taylorotwell That is fair and basically the implementation of `whereNotIn` but even if that were not the case I feel this should be added as providing `null` as a key is still valid input. Other methods like for example `whereNotNull` also already allow `null` as a key so this change would make the behavior consistent.


PS: sorry if I tagged you twice, I don't know if you receive notifications from closed PRs.

PR #45325, opened 2022-12-15
Add generic hints to Cache

AJenbo:

This PR is opened against master as generic annotations on the cache component is meant to come after 9.x.

P.s. @nunomaduro could you please share some more information regarding the plans for introducing generic annotations on the framework as it would make it easier to work on :)

### DIFF
--- a/src/Illuminate/Cache/Repository.php
+++ b/src/Illuminate/Cache/Repository.php
@@ -84,9 +84,11 @@ class Repository implements ArrayAccess, CacheContract
     /**
      * Retrieve an item from the cache by key.
      *
+     * @template TCacheValue
+     *
      * @param  array|string  $key
-     * @param  mixed  $default
-     * @return mixed
+     * @param  TCacheValue|(\Closure(): TCacheValue)  $default
+     * @return TCacheValue
      */
     public function get($key, $default = null): mixed
     {
@@ -175,9 +177,11 @@ class Repository implements ArrayAccess, CacheContract
     /**
      * Retrieve an item from the cache and delete it.
      *
+     * @template TCacheValue
+     *
      * @param  string  $key
-     * @param  mixed  $default
-     * @return mixed
+     * @param  TCacheValue|(\Closure(): TCacheValue)  $default
+     * @return TCacheValue
      */
     public function pull($key, $default = null)
     {
@@ -372,10 +376,12 @@ class Repository implements ArrayAccess, CacheContract
     /**
      * Get an item from the cache, or execute the given Closure and store the result.
      *
+     * @template TCacheValue
+     *
      * @param  string  $key
      * @param  \Closure|\DateTimeInterface|\DateInterval|int|null  $ttl
-     * @param  \Closure  $callback
-     * @return mixed
+     * @param  \Closure(): TCacheValue  $callback
+     * @return TCacheValue
      */
     public function remember($key, $ttl, Closure $callback)
     {
@@ -396,9 +402,11 @@ class Repository implements ArrayAccess, CacheContract
     /**
      * Get an item from the cache, or execute the given Closure and store the result forever.
      *
+     * @template TCacheValue
+     *
      * @param  string  $key
-     * @param  \Closure  $callback
-     * @return mixed
+     * @param  \Closure(): TCacheValue  $callback
+     * @return TCacheValue
      */
     public function sear($key, Closure $callback)
     {
@@ -408,9 +416,11 @@ class Repository implements ArrayAccess, CacheContract
     /**
      * Get an item from the cache, or execute the given Closure and store the result forever.
      *
+     * @template TCacheValue
+     *
      * @param  string  $key
-     * @param  \Closure  $callback
-     * @return mixed
+     * @param  \Closure(): TCacheValue  $callback
+     * @return TCacheValue
      */
     public function rememberForever($key, Closure $callback)
     {
--- a/types/Support/Cache.php
+++ b/types/Support/Cache.php
@@ -1,0 +1,37 @@
+<?php
+
+use Illuminate\Cache\Repository;
+use Illuminate\Contracts\Cache\Repository as RepositoryInterface;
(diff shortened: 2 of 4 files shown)


driesvints:
Can't this be sent to 9.x instead?

AJenbo:
> Can't this be sent to 9.x instead?

I was told it coudn't

driesvints:
Who did? @nunomaduro just said that we can't add these in general:

https://github.com/laravel/framework/pull/40759#issuecomment-1028237457

It's been over a year so maybe we can reconsider but it's up to @taylorotwell to decide.

AJenbo:
Ok, I'm not privy to what the reasoning is. But I'm happy to have it go in 9.x if possible.

P.s. I'm pretty sure it's still 2022, or I forgot to shop up for xmas which would be pretty bad :D

axlon:
These seem wrong to me, just because we pass a default doesn't mean that the value in the cache is of that type as well:

```
cache()->put('foo', 'bar');
cache()->get('foo', 123);
```

This will return a string but according to this PR it will return an int.

taylorotwell:
Thanks for sending this in. I'm going to mark it as draft for now while people discuss it and also while we wait for @nunomaduro to return from vacation.

AJenbo:
> These seem wrong to me, just because we pass a default doesn't mean that the value in the cache is of that type as well:
> 
> ```
> cache()->put('foo', 'bar');
> cache()->get('foo', 123);
> ```
> 
> This will return a string but according to this PR it will return an int.

I guess this is true, thought this feels like somewhat dubious usage.

The same issue could technically also occur with `remember()`:

```php
cache()->remember('foo', 60, function (): string { return 'bar'});
cache()->remember('foo', 60, function (): int { return 123});
```

For such cases I think it would be fair to still warn the user that they are likely doing something wrong.

src/Illuminate/Cache/Repository.php
```diff
@@ -175,9 +177,11 @@ protected function handleManyResult($keys, $key, $value)
     /**
      * Retrieve an item from the cache and delete it.
      *
+     * @template TCached of mixed
```
nunomaduro, line 180: Please use `TCacheValue` everywhere, and remove `of mixed`.

src/Illuminate/Cache/Repository.php
```diff
@@ -175,9 +177,11 @@ protected function handleManyResult($keys, $key, $value)
...
+     *
      * @param  string  $key
-     * @param  mixed  $default
-     * @return mixed
+     * @param  TCached  $default
+     * @return null|TCached
```
nunomaduro, line 184: No need for `null` here.

src/Illuminate/Cache/Repository.php
```diff
@@ -175,9 +177,11 @@ protected function handleManyResult($keys, $key, $value)
     /**
      * Retrieve an item from the cache and delete it.
      *
+     * @template TCached of mixed
+     *
      * @param  string  $key
-     * @param  mixed  $default
-     * @return mixed
+     * @param  TCached  $default
```
nunomaduro, line 183: This should consideration that the `$default` value may be a closure that returns `TCacheValue`. And this comment can applied to other arguments annotations in this pull request.

types/Support/Cache.php
```diff
@@ -0,0 +1,28 @@
+<?php
```
nunomaduro, line 1: You will need to add tests for the cases when a closure is given for the $default arguments.

nunomaduro (changes requested):
I've made some comments that must be addressed before merging. Also, can you rebase this on `v9.x` please?

AJenbo:
> I've made some comments that must be addressed before merging. Also, can you rebase this on v9.x please?

Thanks! I'll look in to it :+1:

src/Illuminate/Cache/Repository.php
AJenbo: Oh neat I wasn't aware of that short hand

src/Illuminate/Cache/Repository.php
AJenbo: Wasn't aware of this possibility in Laravel, hopefully this will help more discover it as well :)

AJenbo:
@nunomaduro the issues should now have been addressed and the PR has been rebased and targeted towards `9.x`

AJenbo:
@nunomaduro the issues should now have been addressed and the PR has been rebased and targeted towards `9.x`

driesvints:
@AJenbo can you maybe also fix the failing docblock check?

driesvints:
Just copy paste the output of the build.

AJenbo:
Well I have, and then added the following bit to make it actually valid:
```
 * @template TCacheValue
 *
```

@driesvints are you saying that we should go with the non functional generated one for now?

P.s. I think it would also be good to have the tool generate a diff to make the discrepancies more clear. But that is a bit out of scope here.

driesvints:
@AJenbo just leave it for now then. We're automating this soon: https://github.com/laravel/framework/pull/45275

timacdonald:
I'll fix up the generics Facade docblocks stuff once we merge this one. Don't let it be a blocker.

nunomaduro:
Currently working on something else. But I will look at this week at this.

nunomaduro:
Made a few adjustments, and added as you as co-author: https://github.com/laravel/framework/pull/45467.

Tklaversma:
@nunomaduro Since this update, doing the following will give the error `'void' method 'rememberForever' result used`.

```php
    public function doStuff(): string
    {
        return Cache::rememberForever('key', function (): string {
            return 'example';
        });
    }
```

In other words, it sees the return type `TCacheValue` on `rememberForever()` as a void. Probably, because `TCacheValue` has no declaration.

mvynhb:
Using template return types does not work for all IDE's intelephense! It might work for PHPStorm, but not for vscode and others. These changes were merged as if we all used PHPStorm...

PR #45337, opened 2022-12-16
[10.x] Improve error messages in the assertSee test method

nshiro:
This PR improves error messages in `assertSee`, `assertSeeText`, `assertDontSee`, `assertDontSeeText` methods in Illuminate\Testing\TestResponse.

## Problem
The current problem of these methods is when you fail tests, all the html is displayed.
In most cases they are not needed. We often just scroll all the way up.
This lowers the development experience, IMHO.

## How to solve
So instead of displaying all the html, we just display simple error messages.
This is the same way as `assertSeeInHtml` in Illuminate\Mail\Mailable.

### Sample output
2022-12-16_09h35_27

But sometimes we may want to see the html for debugging. That's why I added `debug` option.

We can use this third argument or named arguments to see the html. Though named arguments are not supported by the framework, I don't think this will be a big problem.

```
$response->assertSee('Hello', true, true);
// or
$response->assertSee('Hello', debug: true);
```
I don't think this is a breaking change, but I aimed at Laravel 10 because this may surprise some developers.

As for `assertSeeInOrder` and `assertSeeTextInOrder`, they are somehow different. So I kept them as they are.

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -590,17 +590,12 @@ class TestResponse implements ArrayAccess
      *
      * @param  string|array  $value
      * @param  bool  $escape
+     * @param  bool  $debug
      * @return $this
      */
-    public function assertSee($value, $escape = true)
+    public function assertSee($value, $escape = true, $debug = false)
     {
-        $value = Arr::wrap($value);
-
-        $values = $escape ? array_map('e', ($value)) : $value;
-
-        foreach ($values as $value) {
-            PHPUnit::assertStringContainsString((string) $value, $this->getContent());
-        }
+        $this->responseContains($this->getContent(), $value, $escape, $debug, true);
 
         return $this;
     }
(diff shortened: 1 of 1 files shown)


driesvints:
I'm much in favor of changing this behavior to just this by default but it might be best that we think this through thoroughly. Ideally I also think it's best that we change the default behavior so we can get the benefit of more readable errors immediately without having users to update all of their tests or third party packages like pest add extra methods to hide the html. 

Is it maybe an idea that we make the HTML output part of the verbose flag? Unsure what level it would be (`-v`, `-vv` or `-vvv`).

nshiro:
> Is it maybe an idea that we make the HTML output part of the verbose flag? Unsure what level it would be (`-v`, `-vv` or `-vvv`).

I prefer this idea to mine.👍

nshiro:
@taylorotwell
So this PR is closed and no merge or anything ?
Just wondering.

PR #45347, opened 2022-12-18
[10.x] Add `make:enum` command with enum creation standard

RussianFrank:
Hi, everyone! This request add a new command to laravel console.

I'm refactoring a feature in one of my projects and adding Enum. Since laravel 10 uses php 8.1 as a prerequisite, can we consider standardizing the creation of enums?

I added this functionality in this PR and hope we can open a discussion about it.

I added 2 methods inside the enum as I didn't find native support


SilvioDoMine:
This is a amazing feature that would improve coding time, I have something that you should consider: Laravel [9.x] is supported by `php > =8.0`, and ENUM feature is only available at `php >= 8.1`. You should make this PR to Laravel [10.x]

RussianFrank:
@SilvioDoMine 
True, I will create a new PR for this. Thanks!

src/Illuminate/Foundation/Console/EnumMakeCommand.php
```diff
@@ -0,0 +1,94 @@
...
+
+    /**
+     * Get the console command options.
+     *
+     * @return array
+     */
+    protected function getOptions()
+    {
+        return [
+            ['int', 'i', InputOption::VALUE_NONE, 'Create a integer backed enum'],
```
pyrou, line 91: Other make commands also provide following option: 

```php
['force', 'f', InputOption::VALUE_NONE, 'Create the class even if the enum already exists'],
```

RussianFrank: Thanks! I will add this option in the new PR, as @SilvioDoMine pointed out, I will have to adapt it to Laravel 10

brenolves:
Looks good! i've been waiting for a **make:enum** command for quite a while now, it'd be nice to have boilerplates for enums since not every developer does them the same way. I'm hoping this gets merged!

taylorotwell:

megasteve19:
Too bad. This command would be awesome. I was about to implement same command but there was already a pr for it. It's very sad that it's not merged.

pyrou:
@megasteve19 Possibly "useful", but "awesome" might be a bit overrated.

But I understand the choice here. Enums are nothing Laravel-specific, and generating a new enum is something already native in popular IDE, such as PHPStorm (Ctrl+N > PHP Class > Template: Enum).

dennisprudlo:
Yeah I agree, this _may_ be useful for _some_ developers but as pyrou said it's not a laravel specific construct. Otherwise we would have to create `make:trait`, `make:interface`, `make:class` etc.

And I think the stubs are too opinionated. Some developers (me included) don't use `PascalCase` for enum cases and don't need additional static names and values functions.

pyrou:
Looks like I've hurt you. I apologize, wasn't the point.

You are free to create a package (public or private) to include all these command (make trait, interface, enum, etc..) and add this package to all your laravel projects as a `composer require --dev` dependency.

There is a lot of advantage of doing this:
- package won't hit production (because included as a dev dependency)
- you are free to add enum cases in stub the way YOU prefer (pascal, upper, kebab, whatever case, no enum body at all)
- you can also customize (override) existing stubs for controller, model, tests, etc..

IMHO, `make:exception` doesn't deserve it place in the `laravel/framework` package

dennisprudlo:
> IMHO, make:exception doesn't deserve it place in the laravel/framework package

Same here – I don't think this adds much value either.

megasteve19:
At the end, the decision has already been made. I believe that if any `make` command has a direct relationship with the framework, it is valuable to me. If you look at them individually, they actually mean nothing; they are just here for productivity.

@pyrou, I apologize for overreacting earlier. If I have hurt you, I sincerely apologize.

PR #45348, opened 2022-12-18
[9.x] Add "snakecase" validation rule

tanakaphper:
we can validate whether requested values are snake or not.

## Usage
`Validator::make(['property' => 'zipCode'], ['property' => 'required|string|snakecase']);`

this pull request were inspired by [Add "lowercase" validation rule](https://github.com/laravel/framework/pull/44883).


ankurk91:
Please stop adding such rules.

This way, You can send a PR for each method in `Str` class .

PR #45349, opened 2022-12-18
fix: Double slash after domain in UrlGenerator->asset()

### LINKED ISSUE #45346: Bug: View::asset() returns double slash after domain in production mode
* Laravel version 9.33.0
* PHP version 8.1

#### Issue
When calling `Vite::asset()` in a blade file, the returned url has a double forward slash after the domain.

For example: https:\/\/domain.com **//** build/assets/script.js

The strange thing is, this only happens in a production build.

#### Settings
1. In my setup, both `.env` and `.env.production` have the same `APP_URL`s, etc (without trailing slash):
```
// .env
APP_ENV=local
APP_URL=https://domain.dev
FRONTEND_URL=https://domain.dev:5173
ASSET_URL=https://domain.dev

// .env.production
APP_ENV=production
APP_URL=https://domain.com
FRONTEND_URL=https://domain.com
ASSET_URL=https://domain.com
```

2. All file names in the created manifest are also without leading and trailing slash (keys as well as values).

#### Bug tracing

So, I'm not exactly sure where the inconsistency starts, but I could trace it back to the following line:

https://github.com/laravel/framework/blob/1ff0379d203ac836c3eeae567cc07b99c352b1e7/src/Illuminate/Routing/UrlGenerator.php#L258

- At this point, `$path` is trimmed of leading/trailing slashes
- But `$root` and `$this->assetRoot` have a trailing slash, as well as the returned value from `$this->removeIndex($root)`. So there is the double slash.
- Force-returning the `$this->formatRoot()` call above that, returns the domain url without trailing slash. So that's working correctly. (Is this what is used in dev mode?)
- So, the problem is how `app('url')` is constructed, I think. (Since `$this->assetRoot` is directly filled from the UrlGenerator's construct parameter.)

#### Quick fix
`rtrim($this->removeIndex($root), '/'))` fixes this issue.

**However**, I'm not sure if the inconsistency is also problematic elsewhere? And maybe there is bug somewhere that produces this that should be fixed?

vHeemstra:
Fixes: #45346 (quick fix, originating problem might still be there)

### DIFF
--- a/src/Illuminate/Routing/UrlGenerator.php
+++ b/src/Illuminate/Routing/UrlGenerator.php
@@ -255,7 +255,7 @@ class UrlGenerator implements UrlGeneratorContract
         // for asset paths, but only for routes to endpoints in the application.
         $root = $this->assetRoot ?: $this->formatRoot($this->formatScheme($secure));
 
-        return $this->removeIndex($root).'/'.trim($path, '/');
+        return rtrim($this->removeIndex($root), '/').'/'.trim($path, '/');
     }
 
     /**


timacdonald:
I'm not able to re-produce this issue this PR is attempting to fix on a fresh Laravel project.

We should include a test for the issue that this is resolving.

Using the following scripts will result in the expected URL generation without the double slashes mentioned.

```php
# Local .env

composer create-project laravel/laravel url-issue \
  && cd url-issue \
  && npm install \
  && npm run build \
  && echo "APP_ENV=local
APP_URL=https://domain.dev
FRONTEND_URL=https://domain.dev:5173
ASSET_URL=https://domain.dev" > .env \
  && php artisan tinker --execute="echo Vite::asset('resources/js/app.js')"
```

Output is `https://domain.dev/build/assets/app-2e5bf673.js`

```
# Prod .env

composer create-project laravel/laravel url-issue \
  && cd url-issue \
  && npm install \
  && npm run build \
  && echo "APP_ENV=production
APP_URL=https://domain.com
FRONTEND_URL=https://domain.com
ASSET_URL=https://domain.com" > .env \
  && php artisan tinker --execute="echo Vite::asset('resources/js/app.js')"
```

Output is `https://domain.com/build/assets/app-2e5bf673.js`

taylorotwell:
Marking this as draft until we have a clear reproduction of this issue.

vHeemstra:
I'll have another look and will try to provide a working example.

vHeemstra:
Bug was my own mistake. So fix is no longer needed/redundant.

PR #45352, opened 2022-12-19
[9.x] Support dynamic params for Facade docblock generation

timacdonald:
This PR adds "dynamic parameter" support for the facade docblock generator.

```php
/**
 * @param  string  $concrete
 * @param  int|null  $dynamic
 * @return string
 */
public function foo($concrete)
{
    $dynamic = func_get_args()[1] ?? null;

    // ...
}
```

```
/**
 * @method static string foo(string $concrete, int|null $dynamic = null)
 */
```


driesvints (approved):

taylorotwell:
Thanks!

Issue #45353, opened 2022-12-19
Illuminate\Support\Facades\URL::forceScheme() has no effect on request()->url()


miloslavkostir:
- Laravel Version: 9.44.0
- PHP Version: 8.1.8
- Database Driver & Version: no database

### Description: 
Behind reverse proxy there are all requests through HTTP. 
![behind proxy server](https://cylab.be/storage/blog/122/files/eSf2nkfQqEn8Z06zhAoCQYlIV3KU8r3ZsvHqnxRW.png)
https://cylab.be/blog/122/using-https-over-a-reverse-proxy-in-laravel

And `url()->current()` generates URL `http://example.com` instead of real `https://example.com`. If I use `\Illuminate\Support\Facades\URL::forceScheme('https');` in `AppServiceProvider::boot()` URL generator generates right URL address. But `request()->url()` generates wrong (http) URL:
```php
dump(url()->current());  // https://example.com
dump(request()->url());  // http://example.com
```
The same problem is with method `\Illuminate\Support\Facades\URL::forceRootUrl();`

### Steps To Reproduce:
1. install clean Laravel
2. Add `\Illuminate\Support\Facades\URL::forceScheme('https');` to `App\Providers\AppServiceProvider::boot()`
3. In `routes/web.php` create route:
```php
Route::get('/https-test', function () {
    dump(url()->current());
    dump(request()->url());
});
``` 
4. Ensure you are behind reverse proxy and go to URL `/https-test`  

If you can't try it behind proxy, try force whatever:
```php
\Illuminate\Support\Facades\URL::forceScheme('foo-bar');
dump(url()->current());  // foo-bar://example.com
dump(request()->url());  // http(s)://example.com
```

driesvints:
Did you set the correct url for `APP_URL`?

miloslavkostir:
> Did you set the correct url for `APP_URL`?

Yes, it has no effect. I think the problem is:
- `Illuminate\Http\Request` extends `Symfony\Component\HttpFoundation\Request`
- `Illuminate\Http\Request::url()` uses `Symfony\Component\HttpFoundation\Request::getUri()`
- `getUri()` -> `getSchemeAndHttpHost()` -> `getScheme()` -> `isSecure()`
- see `Symfony\Component\HttpFoundation\Request::isSecure()`. There is trusted proxy settings. I can't use trusted proxy because of Apache module `mod_remoteip`

[...]

PR #45354, opened 2022-12-19
[9.x] Fix issue in scheme in `request()->url()`

michaelnabil230:
Fix issue: https://github.com/laravel/framework/issues/45353


michaelnabil230:
The tests are failing but not related to my adding

taylorotwell:
The correct way to solve this is via trusted proxies:

https://laravel.com/docs/9.x/requests#trusting-all-proxies

miloslavkostir:
> The correct way to solve this is via trusted proxies:
> 
> https://laravel.com/docs/9.x/requests#trusting-all-proxies

Yes it works fine, thank you. But if the methods `\Illuminate\Routing\UrlGenerator::forceScheme()` and `\Illuminate\Routing\UrlGenerator::forceRootUrl()` exist, they should works well, shouldn't it? Otherwise, they should be marked as deprecated.

PR #45356, opened 2022-12-19
[9.x] Add `decimal` validation rule

RaazPuspa:
This PR adds `decimal` validation rule to check that an attribute is numeric and consists of decimal part digits between a set of values.

### Usage

`decimal:min,max`: parameter(s) defines minimum and maximum (optional) number of acceptable digits (both inclusive) in decimal part

### Examples

- `decimal:1,3`: requires values to have one, two or three digits in fractional part
- `decimal:0,3`: requires values to have at most three digits in fractional part
- `decimal:3`: requires values to have at least three digits in fractional part

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -555,6 +555,34 @@ trait ValidatesAttributes
     }
 
     /**
+     * Validate that an attribute has decimal figures between a set of values.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @param  array<int, int|string>  $parameters
+     * @return bool
+     */
+    public function validateDecimal($attribute, $value, $parameters)
+    {
+        if (!$this->validateNumeric($attribute, $value)) {
+            return false;
+        }
+
+        $this->requireParameterCount(1, $parameters, 'decimal');
+
+        $matches = [];
+        preg_match('/^\d*.(\d*)$/', $value, $matches);
+
+        $decimals = strlen(end($matches));
+
+        if ($decimals < $parameters[0]) {
+            return false;
+        }
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
tweaked behavior of rule so that `decimal:2` means exactly 2 decimal places.

taylorotwell:
Added translation line and replacements.

willrowe:
Does this allow negative numbers?

RaazPuspa:
@willrowe, I tested the case and found that current implementation does not support validation of signed numbers.

I have updated the validation rule logic allowing signed numeric values. Look #45407 for the changes.

vlakoff:
The dot is unescaped, I think it is unintentional:

```php
preg_match('/^\d*.(\d*)$/', $value, $matches);
```

However, by just escaping that dot, the case `42` (no decimal part) would not work anymore (it currently works by luck, because the dot makes a backtrack and consumes the last digit).

After some thoughts, I would suggest the following code (includes #45407):
```php
preg_match('/^[+-]?\d*(\.\d*)?$/', $value, $matches);

$decimals = isset($matches[1]) ? (strlen($matches[1]) - 1) : 0;
```

Also note the following cases are supported by the underlying [`is_numeric()`](https://www.php.net/manual/en/function.is-numeric.php), and also by this validation rule currently: `42.` and `.43` (also `-42.` and `-.43`). Maybe tests should be added for these.

vlakoff:
And for consistency with the Numeric rule, maybe also support leading/trailing spaces:

```php
preg_match('/^\s*[+-]?\d*(\.\d*)?\s*$/', $value, $matches);

preg_match('/^[+-]?\d*(\.\d*)?$/', trim($value, " \t\n\r\v"), $matches); // variant
```

(the null-byte character `\x0` is not supported by [`is_numeric()`](https://www.php.net/manual/en/function.is-numeric.php), hence excluding it from the [`trim()`](https://www.php.net/manual/en/function.trim.php) in the variant)

PR #45357, opened 2022-12-19
[10.x] Macro mixins DX improvement

NoelDeMartin:
This is a follow-up from #45343.

TLDR: Using the [mixin](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Macroable/Traits/Macroable.php#L40) method for setting up macros is cumbersome, this PR improves it (and it's backwards compatible).

Before:

```php
// To set up the mixin...
TestMacroable::mixin(new TestMixin);

// To define the mixin...
class TestMixin
{
    public function methodOne()
    {
        return function ($value) {
            return $this->methodTwo($value);
        };
    }

    protected function methodTwo()
    {
        return function ($value) {
            return $this->protectedVariable.'-'.$value;
        };
    }

    protected function methodThree()
    {
        return function () {
            return 'foo';
        };
    }
}
```

After:

```php
// To set up the mixin...
TestMacroable::mixin(TestMixin::class);

// To define the mixin...
class TestMixin extends MacroMixin
{
    public function methodOne($value)
    {
        return $this->methodTwo($value);
    }

    protected function methodTwo($value)
    {
        return $this->protectedVariable.'-'.$value;
    }

    protected function methodThree()
    {
        return 'foo';
    }
}
```

PS: I am aware that [there are some issues with PHPStan](https://github.com/laravel/framework/actions/runs/3734333962/jobs/6336262745) and I'm just ignoring the problematic line, which is not the ideal solution. I hope that someone more knowledgeable can help me with that. I don't think it's the point of the PR, so I don't want to spend a lot of time with that if I don't know if this is going to be merged anyways.


taylorotwell:

NoelDeMartin:
Ok, I don't think this warrants creating a new package; but if anyone wants to use it you can just copy this folder in your own project: [app/Support/Macros](https://github.com/NoelDeMartin/vocab/tree/main/app/Support/Macros).

And use it like this:

```php
BetterMacros::mixin(TestMacroable::class, TestMixin::class);
```

The API is not so nice as the one I added in the PR, but I don't think it's possible to override the framework method from the outside (or maybe it is possible, but it probably requires too much magic and I don't want to spend any more time looking into this).

PR #45371, opened 2022-12-20
[10.x] Use faster hash algorithm (xxh128)

KennedyTedesco:
As we are requiring PHP 8.1 or higher, what about starting to use xxh128 hashing on views?

Are there any further implications to consider with this change that I'm not aware of? 

Twig has made this move successfully. See the following references for more information:

https://github.com/twigphp/Twig/pull/3588

They are also using it for http cache store key:

https://github.com/symfony/symfony/pull/47094

About the hash:

https://php.watch/versions/8.1/xxHash

### DIFF
--- a/src/Illuminate/View/Compilers/Compiler.php
+++ b/src/Illuminate/View/Compilers/Compiler.php
@@ -81,7 +81,7 @@ abstract class Compiler
      */
     public function getCompiledPath($path)
     {
-        return $this->cachePath.'/'.sha1('v2'.Str::after($path, $this->basePath)).'.'.$this->compiledExtension;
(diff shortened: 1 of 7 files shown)


driesvints (approved):
Lgtm!

KennedyTedesco:
Specifically, about `sha1` vs `xxh128`, the second one is way faster.

image

Reference: https://php.watch/articles/php-hash-benchmark

I haven't benchmarked the View component, and I don't have a starting point to do it yet (tips are welcome! :D). I'm willing to work on this by the end of the week.

KennedyTedesco:
I agree. Worth to document this in future release notes. Since we're working on a major version, I think we're fine.

PR #45376, opened 2022-12-21
[9.x] Add `whenHas` to JsonResource

michaelnabil230:
This PR adds a `whenHas` method to the JSON resource class giving us the ability to conditionally include attributes in a response when an attribute has in the model.

In some time to need to make many resources for return data to API because we do need to return for example in the function index only `id` and `name` for the user. But in the show function, you need all data for the user. 

To fix this problem need to make many resources for resolving it.

And I think all developers need this feature

***Note: I make some clear and enhancement `whenCounted` to `whenHas` function***

Before:

```php
class UserController
{
    public function index()
    {
        $users = User::select(['id', 'name'])->get();

        return UserResource::collection($users);
    }

    public function show(User $user)
    {
        return SingleUserResource::make($user);
    }
}
```

```php
class UserResource extends Resource
{
    {
        return [
            'id' => $this->id,
            'name' => $this->name,
        ];
    }
}
```

```php
class SingleUserResource extends Resource
{
    {
        return [
            'id' => $this->id,
            'name' => $this->name,
            'email' => $this->email,
        ];
    }
}
```

Now with magic function with the same `Resource`:

```php
class UserController
{
    public function index()
    {
        $users = User::select(['id', 'name'])->get();

        return UserResource::collection($users);
    }

    public function show(User $user)
    {
        return UserResource::make($user);
    }
}
```

```php
class UserResource extends Resource
{
    {
        return [
            'id' => $this->id,
            'name' => $this->name,
            'email' => $this->whenHas('email'),
        ];
    }
}
```


taylorotwell:
Why can you not just use `whenNotNull('email')` in your example?

michaelnabil230:
@taylorotwell At some times, the default value is null for example `email_verified_at`
Not can use the `whenNotNull($this->email_verified_at)` in this case but if use `whenHas($this->email_verified_at)` return null if the attribute in resource attributes

michaelnabil230:
When if you select `email_verified_at` from the `User` model and the default value is null to return a null in the user response JSON does not return the `MissingValue` class. And on the other hand when using whenNotNull return the `MissingValue` class.

MasterRO94:
As for me naming is inconvenient to other Laravel classes. It would be better name it `whenExists` or `whenPresent`

PR #45383, opened 2022-12-21
[9.x] Str::after() minor tweaks

KennedyTedesco:
In the original version of the `after()` method, `$subject` is split into an array then reversed and the first element of the resulting array is returned.

```php
return $search === '' ? $subject : array_reverse(explode($search, $subject, 2))[0];
```

So, a temporary array is created every time.

In the modified version we use the same logic used in `afterLast()`, which means, we only use string search and manipulation to return the substring.

Also, the behavior of `after()` has been modified to allow it to return the remainder of a string after the first **or** last occurrence (controlled by the new `$last` parameter).

Stopping using array manipulation was the main goal of this change. The overall behavior of both functions wasn't changed.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -73,7 +73,7 @@ class Str
      *
      * @param  string  $subject
      * @param  string  $search
-     * @param  bool    $last
+     * @param  bool  $last
      * @return string
      */
     public static function after($subject, $search, $last = false)
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -69,15 +69,26 @@ class Str
     }
 
     /**
-     * Return the remainder of a string after the first occurrence of a given value.
+     * Return the remainder of a string after the first (or last) occurrence of a given value.
      *
      * @param  string  $subject
      * @param  string  $search
+     * @param  bool    $last
      * @return string
      */
-    public static function after($subject, $search)
+    public static function after($subject, $search, $last = false)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I'm not sure I really see this as a big improvement? It's more code?

KennedyTedesco:
Yeah, this change is not intended to be a major improvement.

Both versions have the same amount of code. The differences are:

**Current version:**

```php
public static function after($subject, $search)
{
    return $search === '' ? $subject : array_reverse(explode($search, $subject, 2))[0];
}

public static function afterLast($subject, $search)
{
    if ($search === '') {
        return $subject;
    }

    $position = strrpos($subject, (string) $search);

    if ($position === false) {
        return $subject;
    }

    return substr($subject, $position + strlen($search));
}
```

`afterLast()` and `after` can share almost the same algorithm, but they do the task differently. The `after()`  explodes as an array, reverses it, etc.

On the other side, the `afterLast()` uses another way of doing the same thing, but without creating unnecessary arrays.

The proposed change makes the two functions **more consistent** with each other (in my point of view).

**Proposed version:**

```php
public static function after($subject, $search, $last = false)
{
    if ($search === '') {
        return $subject;
    }

    $position = $last ? mb_strrpos($subject, (string) $search) : mb_strpos($subject, (string) $search);

    if ($position === false) {
        return $subject;
    }

    return static::substr($subject, $position + static::length($search));
}

public static function afterLast($subject, $search)
{
    return static::after($subject, $search, true);
}
```

Issue #45388, opened 2022-12-21
Problem directive @endphp after upgrade version to 9.45


massimodellarovere:
- Laravel Version: 9.45.0
- PHP Version: 8.1.13
- Database Driver & Version: MYSQL 5.7

### Description:
After upgrade version 9.44 to 9.45 my rendere view script have problem. This result screenshot

![image](https://user-images.githubusercontent.com/5731295/208987688-f1e576e0-a997-4071-be4b-2755cbdafd13.png)

If execute downgrade to versione 9.4 is OK

![image](https://user-images.githubusercontent.com/5731295/208987939-1e9d81b2-3415-4036-b1a5-865780d2389b.png)

This is example code

```
@php (
    $classes = [
        'admin-font-mono', // Font weight
    ])
@endphp

<span class="{{ eHtml::classes($classes) }}">{!! $value !!}</span>
```

taylorotwell:
Reverted problem PR.

PR #45390, opened 2022-12-21
fix for single line @php statements but also for multi-line @php with parentheses and @endphp

rikvdh:
I've changed the regex to work a bit different. At first we had a negative lookahead on the parentheses. But as seen in #45388, people actually use parentheses which makes it look like a single-line statement instead, but then include the `@endphp`. Since the  regex was greedy and ate other `@php`-statements in the process of searching for full blocks it didn't work.

I've added another test and change the regex to be omitting possible other `@php`-blocks making only the `@php/@endphp` pairs really working.

### DIFF
--- a/src/Illuminate/View/Compilers/BladeCompiler.php
+++ b/src/Illuminate/View/Compilers/BladeCompiler.php
@@ -380,7 +380,7 @@ class BladeCompiler extends Compiler implements CompilerInterface
      */
     protected function storePhpBlocks($value)
     {
-        return preg_replace_callback('/(?<!@)@php(.*?)@endphp/s', function ($matches) {
+        return preg_replace_callback('/(?<!@)@php(?! ?\()(.*?)@endphp/s', function ($matches) {
             return $this->storeRawBlock("<?php{$matches[1]}?>");
         }, $value);
     }
--- a/src/Illuminate/View/Compilers/BladeCompiler.php
+++ b/src/Illuminate/View/Compilers/BladeCompiler.php
@@ -380,7 +380,7 @@ class BladeCompiler extends Compiler implements CompilerInterface
      */
     protected function storePhpBlocks($value)
     {
-        return preg_replace_callback('/(?<!@)@php(?! ?\()(.*?)@endphp/s', function ($matches) {
+        return preg_replace_callback('/(?<!@)@php((?:(?!(?<![@<])@php\b).)*?)@endphp/s', function ($matches) {
             return $this->storeRawBlock("<?php{$matches[1]}?>");
         }, $value);
     }
--- a/tests/View/Blade/BladePhpStatementsTest.php
+++ b/tests/View/Blade/BladePhpStatementsTest.php
@@ -84,4 +84,11 @@ class BladePhpStatementsTest extends AbstractBladeTestCase
 
         $this->assertEquals($expected, $this->compiler->compileString($string));
     }
+
+    public function testCompilationOfMixedPhpStatements()
+    {
+        $string = '@php($set = true) @php ($hello = \'hi\') @php echo "Hello world" @endphp';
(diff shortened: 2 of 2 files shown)


taylorotwell:
Tests failing.

rikvdh:
Sorry I noticed. I need to continue breaking my head some more on that regex..

donnysim:
To be fair I think https://github.com/laravel/framework/issues/45388 uses the `@php` incorrectly, if you put brackets you don't put `@endphp`

```blade
@php (
    $classes = [
        'admin-font-mono', // Font weight
    ])
@endphp
```
should be 
```blade
@php
    $classes = [
        'admin-font-mono', // Font weight
    ]
@endphp
```
with his version even the IDE complains that something is wrong and it's missing an opening statement because of how it perceives the usages

rikvdh:
@donnysim I agree, but it breaks past behavior and this makes all cases work. If you put `@endphp` or not.

donnysim:
But is it worth fixing framework for someones bad code? Maybe this could be moved to 10.x as a breaking change?

rikvdh:
@donnysim I leave that up to maintainers to decide. I would like a fix in 9.x as well since I've now faced my earlier issue a couple of times. The past behaviour of `@php(..)` was a bit buggy depending on its surroundings.

donnysim:
@rikvdh I'm kind of confused here, I just tried
```blade
@php(
    $hello = 'Hello World'
)
@endphp

{{ $hello }}
```

and got 
```
Illuminate\View\ViewException : Undefined variable $hello (View: F:\Laragon\boilerplate\resources\views\render.blade.php)
```

how is his code working then? 🤔 if I use that variant without echo it prints me the php code instead of a blank string as it should

rikvdh:
> @rikvdh I'm kind of confused here, I just tried
> 
> ```
> @php(
>     $hello = 'Hello World'
> )
> @endphp
> 
> {{ $hello }}
> ```
> 
> and got
> 
> ```
> Illuminate\View\ViewException : Undefined variable $hello (View: F:\Laragon\boilerplate\resources\views\render.blade.php)
> ```
> 
> how is his code working then? 🤔 if I use that variant without echo it prints me the php code instead of a blank string as it should

Woah.. I was going to say: 'I dont know'. But if you put a space before the opening parenthesis ot works. ;-)

Apparently there is a difference between
```php
<?php (
    $hello = 'Hello World'
)
?>

<?php echo e($hello); ?>

and 
```php
<?php(
    $hello = 'Hello World'
)
?>

<?php echo e($hello); ?>
```

probably some weird PHP scoping thing. I've not bothered to investigate any further, but with the parenthesis adjacent it seems that the variables are scoped. This is also current behavior without my fix.

donnysim:
Yeah, the space makes it a full php opening and closing match 😅 so the whole statement got confusing. IDE also doesn't recognize it as a separate case which got me more confused that it should. Kind of seems like a gnarly case to crack using regex.

rodrigopedra:
If you always keep a space after the PHP opening tag, it should work.

```php
return $this->storeRawBlock("<?php {$matches[1]} ?>");
```

*added a space before the closing tag for symmetry only*

rikvdh:
@taylorotwell Why did you mark it as a draft? imo this is done.

taylorotwell:
Honestly I don't have any appetite to change this and make the regex more complicated and potentially have another breaking change. The issue isn't that important to me nor is it a pressing issue in the Laravel community.

rikvdh:
@taylorotwell I don't see how this is breaking in its current form. imo the current behavior is buggy and needs fixing.

caendesilva:
Just chiming in here that I also think this feels like a bug. It's rather unexpected to me that you can't use both single-line and multi-line directives. Maybe a note should be added to the docs. I'd be happy to do it if it's desired.

PR #45394, opened 2022-12-22
[10.x] Fix: Macroable - Parent's macros get overridden by descendants

StyleShit:
See the discussion here: #45392

Not sure that I like this approach, any ideas?


taylorotwell:
I don't think we'll really address this. It has functioned this way for years and I don't have a big appetite for diving into changing it.

PR #45398, opened 2022-12-22
[9.x] Http client (a)sync method return types

erikgaal:
Since #45236 the return types of the methods on the Http client have a union of `Illuminate\Http\Client\Response` and `GuzzleHttp\Promise\PromiseInterface`. If you're using PHPStan in your project on level 7 or higher, this results in errors when calling method that only exist on a part of the union. That's mostly all methods that exist on `Illuminate\Http\Client\Response`.

```
 ------ ---------------------------------------------------------------------------------------------------------- 
  Line   example.php                                
 ------ ---------------------------------------------------------------------------------------------------------- 
  1      Call to an undefined method GuzzleHttp\Promise\PromiseInterface|Illuminate\Http\Client\Response::json().  
 ------ ---------------------------------------------------------------------------------------------------------- 
```

By adding specific methods for `async` and `sync` we can narrow the return types. Ideally we would only have `get` and `getAsync` for each HTTP method, but that would be a BC.


src/Illuminate/Http/Client/PendingRequest.php
```diff
@@ -692,12 +716,36 @@ public function head(string $url, $query = null)
...
+        return $this->async()->head($url, $query);
+    }
+
     /**
      * Issue a POST request to the given URL.
      *
      * @param  string  $url
      * @param  array  $data
-     * @return \Illuminate\Http\Client\Response|\GuzzleHttp\Promise\PromiseInterface
+     * @return \Illuminate\Http\Client\Response
```
erikgaal, line 748: Unintended change, should revert this one.

taylorotwell:
I would honestly rather just revert the PR than have to add all these methods.

PR #45399, opened 2022-12-22
Add migration name to migration events

morloderex:
Inspired a bit by  https://twitter.com/taylorotwell/status/1387766514674192384.

I went ahead and added a lookup table for each migration name and if a name is within that list i would like to follow that up by running some arbitrary method call based on the migrationEnded event up method in order to figure out if additional data should be added.

This simple pull request make it possible for me to remove a lot of reflection in the lookup class sense it adds the migration's name that i am using in my lookup table.

With this change this code:
```php 
<?php

namespace App\Listeners\Migrations;

use App\Listeners\Migrations\DataMigrations\BranchData;
use Illuminate\Database\Events\MigrationEnded;

class MigratedListener
{
    protected array $migrationsWithDataFiles = [
        '2022_07_23_193537_create_branches_table' => Some::class,
    ];

    public function handle(MigrationEnded $event)
    {
        $reflection = new \ReflectionClass($event->migration);
        $migrationFile = basename($reflection->getFileName(), '.php');

        if (! array_key_exists($migrationFile, $this->migrationsWithDataFiles)) {
            return;
        }

        $class = new $this->migrationsWithDataFiles[$migrationFile];

        match (true) {
            $event->method === 'up' && $class->shouldAddData() => $class->addData(),
            $event->method === 'down' && $class->shouldRemoveData() => $class->removeData(),
            default => false,
        };
    }
}
``` 

Can become:
```php
<?php

namespace App\Listeners\Migrations;

use App\Listeners\Migrations\DataMigrations\BranchData;
use Illuminate\Database\Events\MigrationEnded;

class MigratedListener
{
    protected array $migrationsWithDataFiles = [
        '2022_07_23_193537_create_branches_table' => Some::class,
    ];

    public function handle(MigrationEnded $event)
    {
        if (! array_key_exists($event->migrationName, $this->migrationsWithDataFiles)) {
            return;
        }

        $class = new $this->migrationsWithDataFiles[$event->migrationName];

        match (true) {
            $event->method === 'up' && $class->shouldAddData() => $class->addData(),
            $event->method === 'down' && $class->shouldRemoveData() => $class->removeData(),
            default => false,
        };
    }
}
```

### DIFF
--- a/src/Illuminate/Database/Migrations/Migrator.php
+++ b/src/Illuminate/Database/Migrations/Migrator.php
@@ -206,7 +206,7 @@ class Migrator
             return $this->pretendToRun($migration, 'up');
         }
 
-        $this->write(Task::class, $name, fn () => $this->runMigration($migration, 'up'));
+        $this->write(Task::class, $name, fn () => $this->runMigration($migration,'up', $name));
 
         // Once we have run a migrations class, we will log that it was run in this
         // repository so that we don't try to run it next time we do a migration
@@ -371,7 +371,7 @@ class Migrator
             return $this->pretendToRun($instance, 'down');
         }
 
-        $this->write(Task::class, $name, fn () => $this->runMigration($instance, 'down'));
+        $this->write(Task::class, $name, fn () => $this->runMigration($instance, 'down', $name));
 
         // Once we have successfully run the migration "down" we will remove it from
         // the migration repository so it will be considered to have not been run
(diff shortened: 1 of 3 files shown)


src/Illuminate/Database/Migrations/Migrator.php
```diff
@@ -734,7 +734,7 @@ public function setOutput(OutputInterface $output)
      */
     protected function write($component, ...$arguments)
     {
-        if ($this->output && class_exists($component)) {
+        if ($this->output && $this->output  && class_exists($component)) {
```
taylorotwell, line 737: Is this necessary? Seems like `$this->output && $this->output`?

taylorotwell:
@morloderex why don't we just add this method to the base `Migration` class since the migration instance is already included in these events?

Feel free to PR it if it works for you:

```php
/**
 * Get the name of the migration.
 *
 * @return string
 */
public function migrationName()
{
    return str_replace('.php', '', basename((new ReflectionClass(static::class))->getFileName()));
}
```

PR #45405, opened 2022-12-23
[9.x] Precognitive validation with nested arrays doesn't throw validation error

morloderex:
Hello everybody,

I belive i have found a bug on the way the `filterPrecognitiveRules` method on the `CanBePrecognitive` trait works. The reason is that the string `nested.*.name` have not being parsed to the corisponding `nested.0.name` yet because the validator instance only looks after `nested.*.name` and builds the rules array based on the incoming data turning `nested.*.name` into n number of elements. This means that when we try to filter for only keys containing `nested` and `nested.0.name` within `filterPrecognitiveRules` our rules haven't been parsed yet which mean that the rule for key `nested.0.name` just get dropped. And never actually validated.

I have attached a test case for this bug.

### DIFF
--- a/src/Illuminate/Validation/Validator.php
+++ b/src/Illuminate/Validation/Validator.php
@@ -1098,9 +1098,11 @@ class Validator implements ValidatorContract
      */
     public function getRulesWithoutPlaceholders()
     {
-        return collect($this->rules)->mapWithKeys(fn ($value, $key) => [
-            $this->replacePlaceholderInString($key) => $value,
-        ])->all();
+        return collect($this->rules)
+            ->mapWithKeys(fn ($value, $key) => [
+                $this->replacePlaceholderInString($key) => $value,
+            ])
+            ->all();
     }
 
     /**
--- a/src/Illuminate/Validation/Validator.php
+++ b/src/Illuminate/Validation/Validator.php
@@ -1092,6 +1092,18 @@ class Validator implements ValidatorContract
     }
 
     /**
+     * Get the validation rules with key placeholders removed.
+     *
+     * @return array
+     */
+    public function getRulesWithoutPlaceholders()
+    {
+        return collect($this->rules)->mapWithKeys(fn ($value, $key) => [
+            $this->replacePlaceholderInString($key) => $value,
+        ])->all();
+    }
+
+    /**
      * Set the validation rules.
      *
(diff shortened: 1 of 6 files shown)


timacdonald:
## WIP Notes

- [ ] Should we "filterConditionalRules"?
- [ ]

timacdonald:
Note to self: consider the filtering of conditional rules.

timacdonald:
Thanks so much for providing a failing test @morloderex. I've addressed the issue and will summarise things here.

This PR now fixes 2 issues.

## Issue 1.

It is possible with Precognition to specify the validation rules to run from the client, via a `Precognition-Validate-Only` header.

```
POST /posts HTTP/1.1
Precognition true
Precognition-Validate-Only members,members.0.email
```

The backend then runs a filter over the validation rules,  something along the lines of the following.

```php
$initialRules = $formRequest->rules();

$only = explode($request->header('Precognition-Validate-Only'));

$rules = collect($initialRules)->only($only)->all();
```

The problem is that `$initialRules` is populated directly from the form request's `rules()` method. If a rule contains a `*`, it has not yet been expanded.

```php
public function rules()
{
    return [
        'members.*.name' => 'required',
    ];
}
```

So we are unable to filter `members.1.email`.

The fix for this is:

1. Retrieve the rules from the validator after they have been parsed and expanded.
2. Run the filter
3. Set the rules back on the validator after filtering.

src/Illuminate/Validation/Validator.php
```diff
@@ -1091,6 +1091,20 @@ public function getRules()
...
+    /**
+     * Get the validation rules with key placeholders removed.
+     *
+     * @return array
+     */
+    public function getRulesWithoutPlaceholders()
+    {
+        return collect($this->rules)
+            ->mapWithKeys(fn ($value, $key) => [
+                str_replace($this->dotPlaceholder, '\\.', $key) => $value,
```
timacdonald, line 1103: This is not using the existing method for replacing placeholders, as it does not restore the backslash.

timacdonald (commented):
Ready for review.

morloderex:
@timacdonald thank a lot for taking a look at this. This bug has bitten me a bit.

I hope it can be fixed with the next release.

taylorotwell:
Thanks!

PR #45406, opened 2022-12-23
[9.x] Allow returning an array from `Collection::keyBy(Func)` like is currently possible in `Collection::groupBy(Func)`

Thomas-Rosenkrans-Vestergaard:
Currently the `Collection::groupBy(Func)` method allows grouping a collection by multiple keys by returning an array from the function given to `groupBy`. I suggest that we allow the same from the `keyBy` method. 

I myself have needed this type of functionality, and I therefor decided to create a PR. The following could be a case where I would use the feature:

```php
// A bunch of matches between some number of players (already sorted)
$matches = collect([
    ['id' => 1, 'players' => ['Player A', 'Player B']],
    ['id' => 2, 'players' => ['Player C', 'Player A']],
    ['id' => 3, 'players' => ['Player C', 'Player B']]
]);

// Find the latest match for each player
dump($matches->keyBy(fn(array $match) => $match['players'])->all());
```

Results in:

```
^ array:3 [
  "Player A" => array:2 [
    "id" => 2
    "players" => array:2 [
      0 => "Player C"
      1 => "Player A"
    ]
  ]
  "Player B" => array:2 [
    "id" => 3
    "players" => array:2 [
      0 => "Player C"
      1 => "Player B"
    ]
  ]
  "Player C" => array:2 [
    "id" => 3
    "players" => array:2 [
      0 => "Player C"
      1 => "Player B"
    ]
  ]
]
```

This change is not technically backwards-compatible, since previous code that returned an array would throw an error, and will now no longer throw an error. I'm not sure whether it is truly considered a BC-break, since the previous behavior wasn't ever useful.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -542,13 +542,13 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
         foreach ($this->items as $key => $item) {
             $resolvedKeys = $keyBy($item, $key);
 
-            if (!is_array($resolvedKeys)) {
+            if (! is_array($resolvedKeys)) {
                 $resolvedKeys = [$resolvedKeys];
             }
 
             foreach ($resolvedKeys as $resolvedKey) {
                 if (is_object($resolvedKey)) {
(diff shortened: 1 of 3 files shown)


taylorotwell:

PR #45407, opened 2022-12-23
Update `decimal` validation rule to allow validation of signed numbers

RaazPuspa:
This PR allows signed numbers to pass through `decimal` validation rule.

### Previously

- `decimal:2` will fail validation of `-1.23` even though it is numeric and has exactly two decimal places.

### Now

- `decimal:2` will pass the validation of `-1.23`

### Extension

Even though no positive numbers start with `+` sign, there might be a scenario where handling numeric values inside string quotes include the sign. This PR extends the existing validation rule to allow presence of it.

- `decimal:2` will pass validation of `+1.23`

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -572,7 +572,7 @@ trait ValidatesAttributes
 
         $matches = [];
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't understand why you modified any existing tests. Just add new test cases.

PR #45412, opened 2022-12-23
only unlink a file after checking if the file exists

micksp:
Using tinkerwell I always get the following error message: 
```
PHP Warning:  unlink(/......./storage/framework/sessions/{identifier}): No such file or directory in /....../vendor/laravel/framework/src/Illuminate/Filesystem/Filesystem.php on line 286
```
The tinkerwell team told me this was a bug in laravel :-) - I resolved this by checking if a file exists before trying to remove it.


taylorotwell:
Would need more information on how to recreate this in Tinkerwell? The code in Laravel has been like this for years without issue?

PR #45416, opened 2022-12-23
[10.x] Adds PHPUnit 10 support

nunomaduro:
This pull request adds PHPUnit 10 support to `laravel/framework` - while keeping support for PHPUnit 9. Note that this pull request, should test and released once PHPUnit 10 is out. In addition:

1. [x] Have @taylorotwell validate the change on [CacheFileStoreTest.php](https://github.com/laravel/framework/pull/45416/files/74db339d4f43c9bbc3ecb9a38dbd972ab8ebf43b#diff-2a01fbfb6576c310b4bc5f82d2e4b4b73fb1e998c206f105d0a6affc866476b7).
2. [x] Make Octane compatible with Collision v7 (https://github.com/laravel/octane/pull/643).
3. [x] Merge this pull request.

---

4. [x] Ask Mior to release : https://github.com/orchestral/testbench-core/pull/84.
5. [x] Update our `composer.json` with the latest version of Testbench.


composer.json
```diff
@@ -96,7 +96,7 @@
         "orchestra/testbench-core": "^8.0",
```
crynobone, line 96:
```suggestion
        "orchestra/testbench-core": "dev-phpunit10",
```

We should use this branch for now (alias to `master`)

src/Illuminate/Cache/FileStore.php
```diff
@@ -267,9 +267,15 @@ protected function getPayload($key)
             return $this->emptyPayload();
         }
 
+        $e = null;
+
         try {
-            $data = unserialize(substr($contents, 10));
+            $data = @unserialize(substr($contents, 10));
```
nunomaduro, line 273: @taylorotwell @driesvints Previously, on situations where the `unserialize` was failing because of bad string content, the `$data` variable contained the value `false` - and no exception would be thrown. This is caused because the `unserialize` functions returns `false` and not an exception in this cases.

In PHPUnit 9, we are converting notices to exceptions, causing the `CacheFileStoreTest::testUnserializableFileContentGetDeleted` test to pass. 

Yet, in PHPUnit 10, there is no such as feature that converts notices to exceptions, effectively making our tests to behave exactly like user-land code would behave. Therefore, had to adjust this code - in the process probably fixing a bug - for when the result `$data` is `false`, we also forget the original `$key`.

tests/Cache/CacheFileStoreTest.php
```diff
@@ -28,21 +28,6 @@ public function testNullIsReturnedIfFileDoesntExist()
         $this->assertNull($value);
     }
 
-    public function testUnserializableFileContentGetDeleted()
```
nunomaduro, line 31: @taylorotwell This test is does not actually represent what's happening on the FileStore class. So, it's happening was that PHPUnit 9 was overriding the exception handler, by converting notices to exceptions. Meaning that unserialize unserializable content, would result in an exception and consequentially removing the file content.

Now, with PHPUnit 10, this test would be fail, because notices would not be converted to exceptions, meaning that unserialize unserializable content would simply result in `false`.

src/Illuminate/Foundation/Testing/TestCase.php
```diff
@@ -114,6 +113,26 @@ protected function refreshApplication()
...
+            if (! is_null(static::$latestResponse)) {
+                static::$latestResponse->transformNotSuccessfulException($e);
+            }
+
+            throw $e;
+        }
+
+        return $result;
+    }
+
```
a-bashtannik: For all who are looking for PHPStorm `Jump to source` fixes because of this update and found my comment - just upgrade your deps to PHPUnit 10

PR #45419, opened 2022-12-24
[9.x] Add `unless` to JsonResource

michaelnabil230:
This PR adds an `unless` method to the JSON resource class for retrieving a value if the given "condition" is (or resolves to) falsy.

### DIFF
--- a/tests/Integration/Http/ResourceTest.php
+++ b/tests/Integration/Http/ResourceTest.php
@@ -168,7 +168,7 @@ class ResourceTest extends TestCase
         ]);
     }
 
-    public function testResourcesMayDosntHaveOptionalValues()
+    public function testResourcesMayHaveOptionalValues()
     {
         Route::get('/', function () {
             return new PostResourceWithUnlessOptionalData(new Post([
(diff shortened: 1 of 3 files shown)


tests/Integration/Http/ResourceTest.php
```diff
@@ -167,6 +168,31 @@ public function testResourcesMayHaveOptionalValues()
         ]);
     }
 
+    public function testResourcesMayDosntHaveOptionalValues()
```
kevinvanboeckholtz, line 171: Unreadable function name
MayDosntHave
- May Doesn't have
- May Have?
- Doesn't Have?

michaelnabil230: Sorry but I’m a little weak in English Can you correct it for me?

kevinvanboeckholtz:
```suggestion
    public function testResourcesMayHaveOptionalValues()
```

kevinvanboeckholtz:
```suggestion
    public function testResourcesMayHaveOptionalValues()
```

PR #45425, opened 2022-12-25
[9.x] Add test assertion to Macroable session

mehdirajabi59:
This PR adds test assertion to Macroable session

### DIFF
--- a/tests/Session/SessionStoreTest.php
+++ b/tests/Session/SessionStoreTest.php
@@ -634,6 +634,15 @@ class SessionStoreTest extends TestCase
         ]], $errors->getBags()['default']->getMessages());
     }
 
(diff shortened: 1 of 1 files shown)


tests/Session/SessionStoreTest.php
```diff
@@ -634,6 +634,15 @@ public function testValidationErrorsCanBeReadAsJson()
         ]], $errors->getBags()['default']->getMessages());
     }
 
+    public function testsItIsMacroable()
```
nunomaduro, line 637:
```suggestion
    public function testItIsMacroable()
```

nunomaduro (changes requested):

nunomaduro (approved):

PR #45430, opened 2022-12-26
Add deleteWhere() and forceDeleteWhere() to Eloquent Query Builder

ayzerobug:
Add deleteWhere() and forceDeleteWhere() to Eloquent Query Builder


lk77:
what's the point ? you can already do that using where and delete / force delete

ayzerobug:
Yea, it's a combination of those methods

ankurk91:
Please don't pollute the framework to save some keystrokes.

PR #45434, opened 2022-12-27
[9.x] Move facades job to separate workflow

driesvints:
Right now, all PR's have an open list of GHA checks because there's one check for facades that's being skipped. Since GitHub doesn't allows to auto-collapse this list I'm moving the job to a different workflow to circumvent this. This will clean up PRs and make it easier to see if a PR has passing builds or not.

Screenshot 2022-12-27 at 12 00 59

### DIFF
--- a/.github/workflows/facades.yml
+++ b/.github/workflows/facades.yml
@@ -1,22 +1,21 @@
-name: static analysis
+name: facades
 
 on:
   push:
     branches:
       - master
       - '*.x'
-  pull_request:
+
+if: ${{ github.repository_owner == 'laravel' }}
 
 jobs:
-  types:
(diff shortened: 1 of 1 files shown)


.github/workflows/facades.yml
```diff
@@ -0,0 +1,45 @@
+name: facades
+
+on:
+  push:
+    branches:
+      - master
+      - '*.x'
+
+if: ${{ github.repository_owner == 'laravel' }}
```
driesvints, line 9: I had to remove this if statement post-merge. I guess it can't hurt.

timacdonald:
Thanks @driesvints

PR #45441, opened 2022-12-27
[9.x] Better validation error massage for rules with list of values

hafezdivandari:
This PR fixes 2 issues:

First, it encloses values that contain `,` character within double quote `"` in the validation error message for all rules that accept list of values:

```php
Validator::make(['name' => 'bar'], ['name' => 'starts_with:foo,"b, ar",baz'])->validate();

// Before this PR:
// The name must start with one of the following: foo, b, ar, baz.

// After this PR:
// The name must start with one of the following: foo, "b, ar", baz.
```

Second, the `date_format` validation rule accepts list of formats (related to laravel/docs#8422), this PR adds `:formats` placeholder for this rule.
```php
Validator::make(['birthdate' => 'Tue, 27 Dec 2022'], ['name' => 'date_format:d M Y,Y-m-d'])->validate();

// Before this PR, using `:format` placeholder:
// The birthdate does not match the format d M Y.
```

### DIFF
--- a/src/Illuminate/Validation/Concerns/FormatsMessages.php
+++ b/src/Illuminate/Validation/Concerns/FormatsMessages.php
@@ -460,6 +460,26 @@ trait FormatsMessages
     }
 
     /**
+     * Get the displayable list of the values.
+     *
+     * @param  string  $attribute
+     * @param  array  $values
+     * @return string
+     */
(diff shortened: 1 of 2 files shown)


taylorotwell:
Honestly in these edge cases I suggest just explicitly passing the error message you want.

PR #45442, opened 2022-12-27
Added userIs() method to guards

stephanbouman:
More than once I compare the authenticated user against another user instance. It is code which repeats a lot. 

```php
if(auth()->user()->id === $post->user->id){
```
could be a lot easier to read:

```php
if(auth()->userIs($post->user)){
```
It makes the DX better with adding a bit syntactic sugar. We read actually what we are doing.

### DIFF
--- a/src/Illuminate/Contracts/Auth/Guard.php
+++ b/src/Illuminate/Contracts/Auth/Guard.php
@@ -2,6 +2,8 @@
 
 namespace Illuminate\Contracts\Auth;
 
+use Illuminate\Contracts\Auth\Authenticatable as AuthenticatableContract;
+
 interface Guard
 {
     /**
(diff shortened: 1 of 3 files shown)


src/Illuminate/Contracts/Auth/Guard.php
```diff
@@ -46,6 +48,13 @@ public function validate(array $credentials = []);
      * @return bool
      */
     public function hasUser();
+    
+    /**
+     * Compares the authenticated user with a given user.
+     *
+     * @return bool
+     */
+    public function userIs(AuthenticatableContract $user);
```
timacdonald, line 57: The change to the contract makes this a breaking change if we are targeting 9.x

We could:

1. Target 10.x
2. Remove the contract change and just implement the method on the `GuardHelpers` trait to make it available on implementations.

BrandonSurowiec: The Eloquent Model has an `is` method for this exact comparison:

````php
if (auth()->user()->is($post->user)) {
````

BrandonSurowiec:
The Eloquent Model has an `is` method for this exact comparison:

````php
if (auth()->user()->is($post->user)) {
````

stephanbouman:
Missed the is() method on the model. I was looking for it in the auth guard. Thanks!

Issue #45418, opened 2022-12-24
Increment on a field with custom cast checks for method on one class but calls another


Sjord:
- Laravel Version: illuminate/database v9.45.1
- PHP Version: 8.2.0
- Database Driver & Version: SQLite PDO

### Description:

I'm calling `increment` on a field that has a custom cast. The Laravel Model class checks whether a custom `increment` implementation exists. However, it checks on one class and then performs the call on another class.

### Steps To Reproduce:

[Example code here](https://github.com/Sjord/eloquent-deviable-bug/blob/main/index.php). The `Euro->increment` and `Euro->decrement` methods need to exist, but are never called.

I have a `Member` model with an `amount` field, which is of type `Euro`, and casting is provided by `EuroCaster`. The problem is that calling `increment` checks for `Euro->increment`, and if that exists calls `EuroCaster->increment`.

The problem is in [Model.incrementOrDecrement](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Model.php#L959-L961):

```
        $this->{$column} = $this->isClassDeviable($column)
            ? $this->deviateClassCastableAttribute($method, $column, $amount)
            : $this->{$column} + ($method === 'increment' ? $amount : $amount * -1);
```

The last line won't work, since adding `Euro` with `+` isn't possible. Our column should be "class deviable". The function [`isClassDeviable`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1613-L1618) checks whether the methods exist:

```
    protected function isClassDeviable($key)
    {
        return $this->isClassCastable($key) &&
            method_exists($castType = $this->parseCasterClass($this->getCasts()[$key]), 'increment') &&
            method_exists($castType, 'decrement');
    }
```

It checks this on the `Euro` class. If these methods exist, [`deviateClassCastableAttribute`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L882-L887) is called. This calls [`resolveCasterClass`](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1641), which actually checks whether the class implements Castable and calls `castUsing` to get `EuroCaster`.

So it checks for `Euro->increment`, and then calls `EuroCaster->increment`. I believe this is a bug.

WendellAdriel:
[...]

PR #45444, opened 2022-12-28
[9.x] fixed issue on which class to check increment and decrement methods for custom cast

WendellAdriel:
This fix is related to the issue mentioned on #45418 
The code was checking if the increment/decrement methods were available on one class but then using another class to actually call the methods. This fix is for the code to check the same class that will be called the methods.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1616,12 +1616,9 @@ trait HasAttributes
             return false;
         }
 
-        $castType = $this->parseCasterClass($this->getCasts()[$key]);
-        if (method_exists($castType, 'castUsing')) {
-            $castType = $castType::castUsing([]);
-        }
+        $castType = $this->resolveCasterClass($key);
 
-        return method_exists($castType, 'increment') && method_exists($castType, 'decrement');
+        return method_exists($castType::class, 'increment') && method_exists($castType::class, 'decrement');
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Sjord:
Wouldn't it be better to just call the existing [`resolveCasterClass`](https://github.com/laravel/framework/blob/3dff5847ef8b9033ab1a414829abb543f1a9038a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1641-L1663) method?

WendellAdriel:
I see your point @Sjord 

But the `resolveCasterClass` will resolve the class while in the code I wrote it's going to only check if the `castUsing` is present and get the class from it without creating an instance of the class.

I'm fine with both solutions tho. I'll just wait for someone to review it and if needed I can update it.
Thanks for the help 😉

taylorotwell:
I think a test would be good here.

src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
```diff
@@ -1612,9 +1612,16 @@ protected function isEnumCastable($key)
...
-        return $this->isClassCastable($key) &&
-            method_exists($castType = $this->parseCasterClass($this->getCasts()[$key]), 'increment') &&
-            method_exists($castType, 'decrement');
+        if (! $this->isClassCastable($key)) {
+            return false;
+        }
+
+        $castType = $this->parseCasterClass($this->getCasts()[$key]);
+        if (method_exists($castType, 'castUsing')) {
+            $castType = $castType::castUsing([]);
```
sumihiran, line 1621: Wouldn't calling the "castUsing" method with an empty array disregards any given cast parameters?

WendellAdriel:
@taylorotwell I'll work on the test
@sumihiran I'm going to check your observation

WendellAdriel:
@taylorotwell I made some changes:
- Added a test case following the example given on #45418 
- Changed the `isClassDeviable` to use the `resolveCasterClass` method because of the observation made by @sumihiran

PR #45447, opened 2022-12-29
[9.x] Added encoding in limit and large negative number fix

Illusionist3886:
1. Added encoding in limit()
2. Previously `Str::limit('Laravel', -10);` would cause an error due to out of range but now it will show the ending of the limit() rather showing error. Just like `Str::limit('Laravel', -10);` would result in full string (**Laravel**).

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -492,7 +492,7 @@ class Str
     {
         if (mb_strwidth($value, $encoding) <= $limit) {
             return $value;
-        } else if ($limit < 0 && mb_strwidth($value, $encoding) <= abs($limit)) {
+        } elseif ($limit < 0 && mb_strwidth($value, $encoding) <= abs($limit)) {
             return $end;
         }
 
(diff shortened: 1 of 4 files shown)


ankurk91:
This is a breaking change in parameters' order

Illusionist3886:
> This is a breaking change in parameters' order

So, what should be our steps?

Illusionist3886:
@ankurk91 Updated the files, won't break the parameter sequence. Can you confirm?

ankurk91:
It looks good, please update the doc block with new argument and its type

Illusionist3886:
I think we are ready to roll.

timacdonald:
Just a note for consideration: This is a _potentially_ breaking change for any application that has extended this class and re-implemented this function as the argument count is changing.

Illusionist3886:
> Just a note for consideration: As the argument count is changing, this is a _potentially_ breaking change for any application that has extended this class and re-implemented this function.

Indeed! Didn't think about that. So, in such case, what should be our procedure?

taylorotwell:
You can just define your own limit function if the current one doesn't meet your needs.

PR #45448, opened 2022-12-29
Fix: acceptsAnyContentType returns false even if */* is provided

yob-yob:
based on this [docs from mozilla](https://developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation/List_of_default_Accept_values)

the accept any value of `*/*` is not the first accept value...

this is also true from the return value of `getAcceptableContentTypes`

that means when using safari... 

which has a default accept header of `text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8`

the function `acceptsAnyContentType` will return `false` even thou `*/*` is provided 2nd to the last of the comma separated values of accept.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithContentTypes.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithContentTypes.php
@@ -126,6 +126,8 @@ trait InteractsWithContentTypes
                 return true;
             }
         }
+
+        return false;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


yob-yob:
@taylorotwell do I need to pull latest changes? tests are failing for some reason but on files that was not changed by this PR.

taylorotwell:
No plans to change this behavior on a patch release.

PR #45452, opened 2022-12-29
[9.x] Test for `whereIn` and `whereIntegerInRaw` 🔍

imanghafoori1:
These methods have no integration tests at all.
- `whereIn`,
- `orWhereIn`,
-  `whereIntegerInRaw`,
-  `whereNotIn`,
-  `whereIntegerNotInRaw`

It also adds a test for a non-documented feature of `whereIn` which allows to pass a "queriable" object as the `$values`.


imanghafoori1:
The failing tests on PHP 8.2 are not related to this change. It seems a sub-package is using a deprecated PHP function.

PR #45453, opened 2022-12-29
[9.x] Optimize `whereKey` method 🏃🏻‍♂️

imanghafoori1:
If `whereIntegerInRaw` performs better, why not use it when filtering the numeric Primary-key columns?
This change also improves higher-level methods like `findMany` and etc, so their tests are updated.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -232,7 +232,11 @@ class Builder implements BuilderContract
         }
 
         if (is_array($id) || $id instanceof Arrayable) {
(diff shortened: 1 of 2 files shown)


driesvints:
We're reverting this as it's causing a breaking change when using arrays: https://github.com/laravel/framework/issues/45582

imanghafoori1:
Can I seek a fix to the reported bug and avoid the revert?

driesvints:
You're free to attempt a new PR 👍

imanghafoori1:
I am seeking the root cause of that right now, Hopefully, I can find it within an hour or so.

PR #45455, opened 2022-12-29
[9.x] Add phpDoc to Eloquent Model

josuelrocha:
This PR adds phpDoc to Eloquent\Model, so IDE's and other analyzer tools can better understand the code, and it also increases developer productivity.

Before this PR:

After this PR:

This PR does not break the code. I tried to put only the methods that are not used internally by Laravel. :)

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -23,6 +23,40 @@ use Illuminate\Support\Traits\ForwardsCalls;
 use JsonSerializable;
 use LogicException;
 
+/**
+ * @method static \Illuminate\Database\Eloquent\Model|static make(array $attributes = [])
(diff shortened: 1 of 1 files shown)


taylorotwell:
Agree with @timacdonald

josuelrocha:
> I've experimented with this recently also, but via an automated process. If we go down this path, I think we should be using the same technique we are using to generate Facade docblocks automatically.

@timacdonald Would you be available to do it? The lack of type-hint has always bothered me, it would be awesome if Laravel had type-hint for Model.

timacdonald:
It is on my list to look at a bit more. I'll circle back to it shortly.

It is possible to use `Model::query()->etc()` which will give you better autocomplete for now.

josuelrocha:
@timacdonald Perfect and thanks for the tip.

PR #45457, opened 2022-12-30
[10.x] Improve decimal cast precision

timacdonald:
Remove manual manipulation and require `bcmath` to use the decimal cast.

This should be noted in the 10.x upgrade guide.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -138,7 +138,7 @@
         }
     },
     "suggest": {
-        "ext-bcmath": "Required to use the multiple_of validation rule.",
+        "ext-bcmath": "Required to use the multiple_of validation rule and the Eloquent model decimal attribute cast.",
         "ext-ftp": "Required to use the Flysystem FTP driver.",
         "ext-gd": "Required to use Illuminate\\Http\\Testing\\FileFactory::image().",
         "ext-memcached": "Required to use the memcache cache driver.",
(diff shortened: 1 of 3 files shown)


timacdonald:
On hold until https://github.com/laravel/framework/pull/45492 is merged.

driesvints:
@timacdonald I guess this can be continued to work on now that PR is merged?

driesvints:
@timacdonald I moved the bc-math requirement to "suggest" which is what you wanted to do I think.

driesvints (approved):

barryvdh:
If it's in the suggest, then it crashes? 
This is different then https://github.com/laravel/framework/pull/45457 which checks if the extension is loaded. In this version for 10.x it always used bcmath

driesvints:
@barryvdh it's in "suggest" because you don't have to use the decimal cast per se. It's suppose to crash if you don't have it installed. That's why it's a "suggest" and not a hard dependency.

Also, I think you wanted to link to a different PR?

barryvdh:
If you don't want to require it, maybe better to keep the fallback? Maybe add an exception if the float precision is to low? Something like https://github.com/laravel/framework/pull/45580/commits/f66db469ba740d1957a288b443a88f48c882a16c
(Removed the exception in https://github.com/laravel/framework/pull/45580/ for 9.x now)

PR #45461, opened 2022-12-30
Allow get object properties from "dot" notation

rianhe:
Right now, it is not possible to access an object or class from dot notation, sometimes it is necessary. From the "data_get" helper it is possible to do so, with this change the same functionality is achieved.

Example:
```
$test = [
    'code' => 'php',
    'list' => [
        (object)['name' => 'one'],
        ['name' => 'two']
    ]
];

// Fail
dump(Arr::get($test, 'list.0.name'));

// Done
dump(data_get($test, 'list.0.name'));
```


taylorotwell:
Use `data_get`.

PR #45463, opened 2022-12-30
[9.x] migrations stubs does not match PSR-12

danielsuguimoto:
Opening braces of migrations files does not follow guidelines of PSR-12 for anonymous classes. The opening brace must be placed on the same line of the class keyword.

https://www.php-fig.org/psr/psr-12/#8-anonymous-classes


driesvints:
We prefer the current way of doing things.

PR #45464, opened 2022-12-30
Add missing throwUnless function to the HTTP Client

robinnydahl:
## Problem
The current Laravel 9.x documentation gives the following examples for throwing exceptions on the HTTP Client response:
```
 
// Throw an exception if a client or server error occurred...
$response->throw();
 
$response->throwIf($condition);
 
// Throw an exception if an error occurred and the given closure resolves to true...
$response->throwIf(fn ($response) => true);
 
$response->throwUnless($condition);
 
// Throw an exception if an error occurred and the given closure resolves to false...
$response->throwUnless(fn ($response) => false);
 
return $response['user']['id'];
```

However, the `throwUnless` function does not exist and throws the following exception:
```
Call to undefined method GuzzleHttp\\Psr7\\Response::throwUnless()
```

Source: https://laravel.com/docs/9.x/http-client#throwing-exceptions

## Solution
Add a `throwUnless` function to the `Illuminate\Http\Client\Response` class that uses the already existing `throwIf` function with a negation unary operator to invert the condition.

A second argument has also been added to the `throwIf` to take in the `$throwCallback` and default it to `null` instead of using `func_get_args()[1] ?? null`.

### DIFF
--- a/src/Illuminate/Http/Client/Response.php
+++ b/src/Illuminate/Http/Client/Response.php
@@ -353,7 +353,7 @@ class Response implements ArrayAccess
     {
         return $this->throwIf(! $condition);
     }
-    
+
     /**
      * Determine if the given offset exists.
      *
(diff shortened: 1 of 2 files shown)


taylorotwell:

robinnydahl:
Thanks @taylorotwell, should we then update the documentation to reflect this?

PR #45465, opened 2022-12-30
Adds ability for creating observer 👀

mahmoudmohamedramadan:
This PR allows us to create the model's observer when creating a model by passing the `-o` instead of creating a model and then creating its observer in separate steps. To illuminate, take a look over the next scenario

## BEFORE this PR gets merged

To create an observer for a model we're running the next commands

```SHELL
php artisan make:model Post
```

```SHELL
php artisan make:observer PostObserver -m Post
```

## AFTER this PR gets merged

But now we can accomplish the previous scenario by running only one command 🚀

```SHELL
php artisan make:model Post -o
```

### DIFF
--- a/src/Illuminate/Foundation/Console/ModelMakeCommand.php
+++ b/src/Illuminate/Foundation/Console/ModelMakeCommand.php
@@ -61,6 +61,7 @@ class ModelMakeCommand extends GeneratorCommand
             $this->input->setOption('seed', true);
             $this->input->setOption('migration', true);
             $this->input->setOption('controller', true);
+            $this->input->setOption('observer', true);
             $this->input->setOption('policy', true);
             $this->input->setOption('resource', true);
         }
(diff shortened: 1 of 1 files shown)


src/Illuminate/Foundation/Console/ModelMakeCommand.php
```diff
@@ -61,6 +61,7 @@ public function handle()
             $this->input->setOption('seed', true);
             $this->input->setOption('migration', true);
             $this->input->setOption('controller', true);
+            $this->input->setOption('observer', true);
```
driesvints, line 64: I don't think this should be part of the "all" flag. I doubt many people are relying on observers.

mahmoudmohamedramadan: In my opinion, I have added this line with the `--all` option because the **Observer** must be related to a model so when I pass the `--all` everything related to the model must be created.

taylorotwell:
No plans to take any action on this right now.

PR #45466, opened 2022-12-30
Tests written for throwIf function on HTTP Client is not correctly entering throwIf function.

robinnydahl:
## Problem
When the test for the HTTP Client's `throwIf` function is run, it's entering the `throw` function and not the `throwIf` function as it should. This causes the test to pass without actually validating what it's supposed to.

## Solution
Move the `get` call before the `throwIf` as done in the other tests to allow the correct function to be called.

### DIFF
--- a/tests/Http/HttpClientTest.php
+++ b/tests/Http/HttpClientTest.php
@@ -1472,8 +1472,8 @@ class HttpClientTest extends TestCase
 
         try {
             $this->factory
-                ->throwIf(true)
-                ->get('http://foo.com/get');
+                ->get('http://foo.com/get')
+                ->throwIf(true);
         } catch (RequestException $e) {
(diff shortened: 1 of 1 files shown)


gdebrauwer:
The reason the `throwIf` happens before the `get` in the test, is to test that the throwIf works on a pending request as well (see name of test)

driesvints:
@gdebrauwer so you're saying we should revert this?

gdebrauwer:
Yes, cause the `throwIf` after the `get` method is already tested a bit further in the file:
https://github.com/robinnydahl/framework/blob/5c8ef368c334642a1de3750905788406fa7915ef/tests/Http/HttpClientTest.php#L1663-L1711

driesvints:
Reverted. Thanks

PR #45471, opened 2022-12-30
[9.x] Short conditional statement

michaelnabil230:


driesvints:
These are not the same: https://3v4l.org/1n1UX

PR #45474, opened 2023-01-01
[10.x] Ability to supply constraint name when creating a foreign key via the fluent builder

Riley19280:
Currently, when creating a table with a long name and long column names, the ability to use the `->foreignId(...)->constrained()` method goes away, because the auto-generated constraint name is too long. The current workaround for this is to use the non fluent builder method (see below), but becomes cumbersome to have to do. 

```php
 Schema::create('my_execptionally_long_table_name', function (Blueprint $table) {

            // This will fail because the constraint name is too long
            $table->foreignId('my_super_long_column_name')->constrained('my_other_really_long_table_name');

            // this is the current workaround to the issue
            $table->foreignId('my_super_long_column_name')
            $table->foreign('my_super_long_column_name', 'my_custom_short_constraint_name')->references('id')->on('my_other_really_long_table_name');

            // Proposed alternative short syntax
            $table->foreignId('my_super_long_column_name')->constrained('my_other_really_long_table_name', 'id', 'my_custom_short_constraint_name');
        });
```

This PR adds the constraint name as an argument to the `->constrained` function call, so that a custom name can be specified for the constraint instead. 

Ideally, the ability to chain a `->name('constraint_name`) on the `->constrained` call would be a more ideal solution, but that seemed non-trivial given how the current constrained helper is working. If that is also the preferred final solution, then  I can put some more time in to that.


taylorotwell:
I don't think it's really worth it to save just 1-2 characters from the more extended definition.

PR #45475, opened 2023-01-01
Proof-of-concept: Component attribute bag merge strategies.

tombanksme:
Hello, me again (And happy new year!)

I ran into an issue with the `ComponentAttributeBag` while working with a Tailwind project. The default strategy for merging the `class` attribute appends the prop value onto the default value.

I had a button component with the `bg-red-500` class. When I wanted to make the button blue and added `bg-blue-500` my button remained red; having two of the same utility and variant combinations with Tailwind can cause unexpected outcomes.

e.g:

```blade
<!-- components/button.blade.php -->
<button {{ $attributes->merge(["class" => "bg-red-500"]) }}>
    {{$slot}}
</button>
```

```blade
<!-- home.blade.php -->
<x-button class="bg-blue-500">
</x-button>
```

This is a proof-of-concept pull request that adds merge strategies to the `ComponentAttributeBag`. It contains three strategies:
- OverrideStrategy (default)
- AppendStrategy
- TailwindStrategy

By default the `AppendStrategy` is used for `class` attributes; `OverrideStrategy` is used for everything else. Users can change the strategy mapping using the `view.strategies` configuration option, or by passing an overriding mapping to the `merge` function.

Still needs some work (mainly tests and some decommissioning) but what do you guys think?

### DIFF
--- a/src/Illuminate/View/ComponentAttributeBag.php
+++ b/src/Illuminate/View/ComponentAttributeBag.php
@@ -10,6 +10,7 @@ use Illuminate\Support\HtmlString;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Conditionable;
 use Illuminate\Support\Traits\Macroable;
+use Illuminate\View\Strategies\{AppendStrategy, OverrideStrategy};
 use IteratorAggregate;
 use Traversable;
 
@@ -25,6 +26,14 @@ class ComponentAttributeBag implements ArrayAccess, Htmlable, IteratorAggregate
     protected $attributes = [];
 
     /**
+     * An array of default merge strategies.
+     *
+     * @var array
+     */
+
+    protected $strategies = [];
+
+    /**
      * Create a new component attribute bag instance.
      *
      * @param  array  $attributes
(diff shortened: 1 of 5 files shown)


aguingand:
Interesting addition. It also may be useful to pass a strategy to the `class()` method.

dennisprudlo:
> Interesting addition. It also may be useful to pass a strategy class to the `class()` method.

I love that idea. Sometimes there are classes I want to append while replacing others in the same class attribute. I end up with either a `@php` block or `!important` styles. A custom merge strategy class would be perfect in that case.

driesvints:
Please rebase with 9.x as we fixed the build on that branch.

tombanksme:
> Please rebase with 9.x as we fixed the build on that branch.

Done

driesvints:
Tests still seem to fail

tombanksme:
The failing tests are caused by my changes.

If this is something that the Laravel team would be interested in merging, I'll invest the time in fixing/writing the tests.

aguingand:
> > Interesting addition. It also may be useful to pass a strategy class to the class() method.
> 
> Are you talking about the `@class` blade directive?

No, I was talking about something like this:
```blade
<div {{ $attributes->class('bg-red-500', TailwindMergeStrategy::class) }}></div>
```

which overrides the strategy for the `class` attribute.

taylorotwell:
I'm pretty happy with the current approach.

royduin:
For reference; @mpociot tried to fix this earlier and made a video about it: https://twitter.com/marcelpociot/status/1310935864848117760

PR #45476, opened 2023-01-01
[9.x] Remove extra code in Model.php to optimize performance 🏃🏻‍♂️

imanghafoori1:
#### First Commit:
- The `whereKey` method can accept an array of ids so the check is not really needed.
- It also kinda optimizes since the `whereKey` method now uses the `whreIntegerInRaw` (instead of `whereIn`) for numeric ids.

#### Second commit:
- The second commit converts anonymous function to arrow functions.


imanghafoori1:
The failing test is not related to this change. The static analyzer test complains about something on the Cache facade.

PR #45482, opened 2023-01-02
[9.x] Fix method contextual binding

imdhemy:
To put it simply, try to:

1. Create a class that has a method with a parameter that has a type hint.
 * For instance, a command class with a handle method that has a parameter with a type hint of an interface.
2. Configure a contextual binding for that parameter to a concrete class.
3. Try to run the command.
4. You will get an error that the given interface is not instantiable.

This is because when the container tries to resolve the method dependencies, It encounters a `buildStack` that does not contain the class that contains the method to be called, as it was already popped from the build stack in a previous step.


taylorotwell:
Provide simple code to reproduce the problem in your PR description and re-open please.

PR #45484, opened 2023-01-02
[9.x] Fix method contextual binding

imdhemy:
This PR reopens #45482.

This commit reproduces the issue:


imdhemy:
I will try to find an alternative solution

PR #45487, opened 2023-01-02
[10.x] Add support for native column modifying

hafezdivandari:
Related to #45258, this PR adds support for native column modifying and drops necessity of installing `doctrine/dbal` when modifying columns using `change()` on MariaDB, MySQL, PostgreSQL and SQL Server.

### Supported databases, modifiers and types:

| Database | Native Support | Supported column modifiers | Supported column types |
| --- | --- | --- | --- |
| MariaDB  | ✅ ([docs](https://mariadb.com/kb/en/alter-table/#modify-column)) | _Same as MySQL_ | All types |
| MySQL | ✅ ([docs](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html#alter-table-redefine-column)) | `after`, `first`, `autoIncrement`, `from`, `storedAs`, `storedAsJson`, `virtualAs`, `virtualAsJson`, `invisible`, `unsigned`, `nullable`, `default`, `charset`, `collation`, `comment`, `useCurrent`, `useCurrentOnUpdate`, `srid`, `renameTo`<sup>2</sup> | All types |
| PostgreSQL | ✅ ([docs](https://www.postgresql.org/docs/current/sql-altertable.html)) | `autoIncrement`<sup>3</sup>, `from`, `storedAs`<sup>4</sup>, `generatedAs`, `always`, `nullable`, `default`, `collation`, `comment`, `useCurrent`, `isGeometry`, `projection` |  All types except `enum`<sup>5</sup> |
| SQLite | ❌ ([docs](https://www.postgresql.org/docs/current/sql-altertable.html)) | _N/A_ | _N/A_
| SQL Server | ✅ ([docs](https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-ver16)) | `autoIncrement`<sup>6</sup>, `nullable`, `default`, `collation`, `persisted`, `useCurrent` | All types except `enum`<sup>7</sup> |

### Notes:
1. According to [MySQL docs](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html#alter-table-redefine-column), for column definition changes using CHANGE or MODIFY, **the definition must include the data type and all attributes** that should apply to the new column, **other than index attributes** such as PRIMARY KEY or UNIQUE. Attributes present in the original definition but not specified for the new definition are not carried forward. Suppose that a column `col1` is defined as:
   ```php
   $table->integer('col1')->unsigned()->default(1)->comment('my column'); // `col1` int unsigned not null default '1' comment 'my column'
   ```
   and you modify the column as follows, intending to change only INT to BIGINT:
   ```php
   $table->bigInteger('col1')->change(); // alter table `t1` modify `col1` bigint not null
   ```
   That statement changes the data type from INT to BIGINT, but it also drops the UNSIGNED, DEFAULT, and COMMENT attributes. To retain them, the statement must include them explicitly:
   ```php
   $table->bigInteger('col1')->unsigned()->default(1)->comment('my column')->change(); // alter table `t1` modify `col1` bigint unsigned not null default '1' comment 'my column'
   ```
   The same approach is also used on **PostgreSQL** and **SQL Server**, you have to redeclare all the attributes you want to keep on the column definition.
2. Currently Laravel compiles added and modified column before other commands, It means to change and rename a same column you should call `renameColumn` command after `change` command:
   ```php
   $table->enum('difficulty', ['easy', 'hard'])->change();
   $table->renameColumn('difficulty', 'hardness');
   ```
   However, this PR adds `renameTo` column modifier on MySQL to rename and modify a column at one go:
   ```php
   $table->enum('difficulty', ['easy', 'hard'])->renameTo('hardness')->change();
   ```
3. Laravel compiles `autoIncrement` on PostgreSQL as `primary key` and also causes column type to be `serial`. This PR partially supports `autoIncrement`, it means the `autoIncrement` modifier still causes a column type to be `serial` type and you may use `from` modifier to set it's starting value, but it doesn't add or drop the primary key. You may use `primary()` and `primary(false)` commands explicitly to achieve this.
4. PostgreSQL doesn't have a native alter command for modifying the expression of a generated column or turning a regular column to a stored generated column, it means `storedAs` modifier is not supported on this PR. However, it is supported to set `storedAs(null)` explicitly on PostgreSQL >= 13.0 to drop the generated column expression (turns a stored generated column into a normal base column). 
5. PostgeSQL [has native `enum`](https://www.postgresql.org/docs/10/datatype-enum.html) but the syntax is different, Laravel uses a check constraint on a `varchar` when adding an `enum` type. This PR doesn't handle modifying `enum` type on PostgreSQL.
6. Laravel compiles `autoIncrement` modifier as `identity primary key` on SQL Server. SQL Server doesn't support adding / dropping `identity` and `primary key` when using alter column. So this PR also doesn't support that. You may use `primary()` and `primary(false)` commands explicitly to achieve this.
7. SQL server doesn't have a native `enum` and Laravel is using a check constraint on a `nvarchar` when adding an `enum` type. This PR doesn't support modifying `enum` on SQL Server.

### What if I have already installed `doctrine/dbal`?
Just like #45258, if you have already installed Doctrine DBAL, you have to call `Schema::useNativeSchemaOperationsIfPossible()` method within the `boot` method of your `App\Providers\AppServiceProvider` class or within your migrations files to be able to use native schema operations.

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -119,7 +119,7 @@ class Blueprint
      */
     public function toSql(Connection $connection, Grammar $grammar)
     {
-        $this->addImpliedCommands($grammar);
+        $this->addImpliedCommands($connection, $grammar);
 
         $statements = [];
 
@@ -183,10 +183,11 @@ class Blueprint
     /**
      * Add the commands that are implied by the blueprint's state.
      *
+     * @param  \Illuminate\Database\Connection  $connection
      * @param  \Illuminate\Database\Schema\Grammars\Grammar  $grammar
      * @return void
      */
-    protected function addImpliedCommands(Grammar $grammar)
+    protected function addImpliedCommands(Connection $connection, Grammar $grammar)
     {
         if (count($this->getAddedColumns()) > 0 && ! $this->creating()) {
             array_unshift($this->commands, $this->createCommand('add'));
@@ -198,7 +199,7 @@ class Blueprint
 
         $this->addFluentIndexes();
 
-        $this->addFluentCommands($grammar);
+        $this->addFluentCommands($connection, $grammar);
     }
 
     /**
@@ -236,12 +237,17 @@ class Blueprint
     /**
      * Add the fluent commands specified on any columns.
      *
+     * @param  \Illuminate\Database\Connection  $connection
      * @param  \Illuminate\Database\Schema\Grammars\Grammar  $grammar
      * @return void
      */
-    public function addFluentCommands(Grammar $grammar)
+    public function addFluentCommands(Connection $connection, Grammar $grammar)
     {
         foreach ($this->columns as $column) {
+            if ($column->change && ! $connection->usingNativeSchemaOperations()) {
+                continue;
+            }
+
             foreach ($grammar->getFluentCommands() as $commandName) {
                 $this->addCommand($commandName, compact('column'));
             }
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -243,17 +243,7 @@ class Blueprint
     {
         foreach ($this->columns as $column) {
             foreach ($grammar->getFluentCommands() as $commandName) {
-                $attributeName = lcfirst($commandName);
-
-                if (! isset($column->{$attributeName})) {
-                    continue;
-                }
-
-                $value = $column->{$attributeName};
-
-                $this->addCommand(
-                    $commandName, compact('value', 'column')
-                );
+                $this->addCommand($commandName, compact('column'));
             }
         }
     }
@@ -1785,35 +1775,5 @@ class Blueprint
         return array_filter($this->columns, function ($column) {
             return (bool) $column->change;
         });
-    }
-
-    /**
-     * Determine if the blueprint has auto-increment columns.
-     *
-     * @return bool
-     */
-    public function hasAutoIncrementColumn()
-    {
-        return ! is_null(collect($this->getAddedColumns())->first(function ($column) {
-            return $column->autoIncrement === true;
-        }));
-    }
-
-    /**
-     * Get the auto-increment column starting values.
-     *
-     * @return array
-     */
-    public function autoIncrementingStartingValues()
-    {
-        if (! $this->hasAutoIncrementColumn()) {
-            return [];
-        }
-
-        return collect($this->getAddedColumns())->mapWithKeys(function ($column) {
-            return $column->autoIncrement === true
-                        ? [$column->name => $column->get('startingValue', $column->get('from'))]
-                        : [$column->name => null];
-        })->filter()->all();
     }
 }
--- a/src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
@@ -34,7 +34,7 @@ class PostgresGrammar extends Grammar
      *
      * @var string[]
      */
-    protected $fluentCommands = ['Comment'];
+    protected $fluentCommands = ['Comment', 'AutoIncrementStartingValues'];
 
     /**
      * Compile a create database command.
@@ -91,15 +91,15 @@ class PostgresGrammar extends Grammar
      *
      * @param  \Illuminate\Database\Schema\Blueprint  $blueprint
      * @param  \Illuminate\Support\Fluent  $command
-     * @return array
+     * @return string
      */
     public function compileCreate(Blueprint $blueprint, Fluent $command)
     {
-        return array_values(array_filter(array_merge([sprintf('%s table %s (%s)',
+        return sprintf('%s table %s (%s)',
             $blueprint->temporary ? 'create temporary' : 'create',
             $this->wrapTable($blueprint),
             implode(', ', $this->getColumns($blueprint))
-        )], $this->compileAutoIncrementStartingValues($blueprint))));
+        );
     }
 
     /**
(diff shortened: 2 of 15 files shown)


src/Illuminate/Database/Schema/Blueprint.php
```diff
@@ -236,24 +237,19 @@ protected function addFluentIndexes()
...
      * @return void
      */
-    public function addFluentCommands(Grammar $grammar)
+    public function addFluentCommands(Connection $connection, Grammar $grammar)
     {
         foreach ($this->columns as $column) {
-            foreach ($grammar->getFluentCommands() as $commandName) {
-                $attributeName = lcfirst($commandName);
-
-                if (! isset($column->{$attributeName})) {
```
taylorotwell, line 248: Why is this check removed?

hafezdivandari: Before this before, we had only one fluent command which was comment on PostgreSQL, this PR does a clean up on auto-increment starting values and add that as fluent command for PostgreSQL and MySQL, also adds default as a fluent command for SQL Server. This Check is removed because it won't work for all these commands, instead of checking the value here, we check the situation on the command itself and decide whether to compile or not.

fluent commands before this PR: Comment (PostgreSQL)
fluent commands after this PR: Comment (PostgreSQL), Auto-increment starting values (PostgreSQL and MySQL) and Default (SQL Server)

src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
```diff
@@ -672,42 +719,7 @@ protected function typeTinyInteger(Fluent $column)
...
-        if (! is_bool($column->generatedAs) && ! empty($column->generatedAs)) {
-            $options = sprintf(' (%s)', $column->generatedAs);
-        }
-
-        return sprintf(
-            '%s generated %s as identity%s',
-            $type,
-            $column->always ? 'always' : 'by default',
-            $options
-        );
```
taylorotwell, line 710: This seems like a pretty huge change for Laravel? We are changing the entire way we define auto-incrementing IDs in Postgres? Is this necessary?

hafezdivandari: It is not a change at all, it is necessary because PostgreSQL has different `alter column` syntax for column type and other column attributes so we have to separate type definitions and modifiers to handle that in a most clear way.

for example:
```
alter table `t1` alter column `col1` type bigserial. alter column `col1`, add generated always as identity, ....
```

hafezdivandari: The `generatableColumn` method has changed to `modifyGeneratedAs` method.

taylorotwell:
Can you elaborate on #3 and #6 in your PR description? This only applies to changing columns? Can you give some examples of what would work and what wouldn't work?

> Laravel compiles autoIncrement on PostgreSQL as primary key and also causes column type to be serial. This PR partially supports autoIncrement, it means the autoIncrement modifier still causes a column type to be serial type and you may use from modifier to set it's starting value, but it doesn't add or drop the primary key. You may use primary and dropPrimary commands explicitly to achieve this.

hafezdivandari:
> Can you elaborate on point 3 and point 6 in your PR description? This only applies to changing columns? Can you give some examples of what would work and what wouldn't work?

@taylorotwell sure. It's not just about modifying columns, currently on creating a table or adding a column, the `autoIncrement` modifier means both "auto-increment" and "primary key". This may not cause a problem at most cases, but it does in some cases for example in MySql you can do this:

```sql
CREATE TABLE users (
     id bigint unsigned not null auto_increment,
     name varchar(255) not null,
     primary key (id, name)
);
```

but you can't do this on Laravel, because `autoIncrement` modifier compiles to `auto_increment primary key` instead of just `auto_increment`. (I wanted to send another PR maybe to fix this on 10.x)

Ok that was the whole picture, but when you want to modify a column with `autoIncrement` modifier on MySQL, this is still not a big problem because MySQL accepts this (although that `primary key` part may not be what the user intended to add):
```
alter table `users` modify `id` bigint unsigned not null auto_increment primary key
```

but PostgreSQL and SQL Server doesn't accept that when altering a column. it means you can't control the adding/dropping of a primary key when changing a column. As I said this needs another BC PR to change what `autoIncrement` modifier means in Laravel when we already have `->primary()` to handle that.

taylorotwell:
OK, but for adding columns everything still works as before on 9.x. However, when modifying columns, can you give an example of some actual migration code that would now throw an error after this PR?

hafezdivandari:
Yes, for adding **everything** works as before. And when modifying an auto-increment column, **no error would be thrown** but there are some differences in compare to when adding a column:

### Examples of `autoIncrement` modifier
#### MySQL / MariaDB
```php
// when adding
$table->integer('id')->autoIncrement(); // alter table `users` add `id` int not null auto_increment primary key
$table->increments('id'); // alias to above

// when modifying
$table->integer('id')->autoIncrement()->change(); // alter table `users` modify `id` int not null auto_increment primary key
$table->increments('id'); // alias to above
```
✅ Everything works as expected.

#### PostgreSQL
```php
// when adding
$table->integer('id')->autoIncrement(); // alter table "users" add column "id" serial not null primary key
$table->increments('id'); // alias to above

// when modifying
$table->integer('id')->autoIncrement()->change(); // alter table "users" alter column "id" type serial, alter column "id" set not null, alter column "id" drop default, alter column "id" drop identity if exists; comment on column "users"."id" is NULL
$table->increments('id'); // alias to above
```
⚠️ The column type is `serial` as expected, but we don't handle the primary key.

#### SQL Server
```php
// when adding
$table->integer('id')->autoIncrement(); // alter table "users" add "id" int not null identity primary key
$table->increments('id'); // alias to above

// when modifying
$table->integer('id')->autoIncrement()->change(); // alter table "users" alter column "id" int not null
$table->increments('id'); // alias to above
```
❌ Both `identity` and `primary key` are missing and `autoIncrement` is ignored.

taylorotwell:
OK - this is a pretty risky PR I think but I'm willing to take a chance on it since it would be nice to not have doctrine/dbal for changing columns.

PR #45489, opened 2023-01-02
Add missing return statement for getDateTime validation

joshuaruesweg:
Currently the method returns nothing, if the `Date::parse($value)` thrown any exception. The return type should be explicit `null`, as this is the same return type as the return type, which returns if the parse fails.


taylorotwell:
Not necessary.

Issue #45424, opened 2022-12-25
Blade exception for text


mxts:
- Laravel Version: 9.1.4
- PHP Version: 8.1.13
- MySQL version: 8.0.31-0ubuntu0.22.04.1
- MySQL Client version: mysqlnd 8.1.13

### Description:
```
@include('subview', [
    'test' => '(',
])

// subview.blade.php
{{ $test }}
```
Blade above produces the error `ParseError syntax error, unexpected token ","`, the cause being the unclosed `(` inside the text.

While I understand this is an edge case, the exception does not point directly to the problem and the solution can be quite hard to spot. In my case, the offending text was a frowning face `:(`

### Steps To Reproduce:

I do have a repository, but this should be quite straightforward to reproduce. Thank you.

bert-w:
https://github.com/laravel/framework/issues/18317

mxts:
Thanks, perhaps it helps to hint about limitations in the subview documentation? I could file a PR for that.

driesvints:
Looks like we won't be fixing this. Feel free to attempt to PR the docs 👍

imanghafoori1:
This can be considered to be fixed as of version 9.47.0 or above.
@mxts

PR #45490, opened 2023-01-02
[9.x] Fixes blade tags issue #45424  🔧

imanghafoori1:
solves #45424 

The solution seems a bit hacky, but the idea, however, is simple:
- Without changing the regex pattern, we perform our first search to get matches. 
- Then by using PHP tokenizer, we check if we are in trouble or not. 
- If we are in trouble, we try to fix it by adding the next candidate section to the matched string to see if we can reach the proper closing parenthesis or not.
- We try the same process until we find the proper match for the closing parenthesis of the blade tag.

```php
@include ( 
  'partial1', ['hell)o' => ')-)--)']  
)
```

in this example, regex finds the :
``` @include (   'partial1', ['hell) ``` first and the loop adds these portions until it reaches the closing:
```)o' => '```
```)-```
```)--```
```']  )```   <=== the closing one.

### Performance:
- Let me note that this part of the code is NOT run during each and every request/response life cycle since the compiled versions of blade files get cached.

- Some tests are included to prove the fix.

As I know, Taylor likes to find a proper regex for it and does not like such solutions, maybe the regex can be found somewhere in open-source IDE syntax highlighters like VS code, or people from Jetbrain can help since they can highlight the syntax properly.

### DIFF
--- a/src/Illuminate/View/Compilers/BladeCompiler.php
+++ b/src/Illuminate/View/Compilers/BladeCompiler.php
@@ -505,6 +505,7 @@ class BladeCompiler extends Compiler implements CompilerInterface
     protected function compileStatements($template)
     {
         preg_match_all('/\B@(@?\w+(?:::\w+)?)([ \t]*)(\( ( [\S\s]*? ) \))?/x', $template, $matches);
+
         for ($i = 0; isset($matches[0][$i]); $i++) {
             $match = [
                 $matches[0][$i],
@@ -516,9 +517,12 @@ class BladeCompiler extends Compiler implements CompilerInterface
 
             // Here we check to see if we have properly found the closing parenthesis by
             // regex pattern or not, and will recursively continue on to the next ")"
-            // then check again until the tokenizer confirms we found the right one
-            while (isset($match[4]) && Str::endsWith($match[0], ')') && ! $this->isProperMatch($match[0])) {
+            // then check again until the tokenizer confirms we find the right one.
+            while (isset($match[4]) &&
+                   Str::endsWith($match[0], ')') &&
+                   ! $this->hasEvenNumberOfParentheses($match[0])) {
                 $rest = Str::before(Str::after($template, $match[0]), ')');
+
                 $match[0] = $match[0].$rest.')';
                 $match[3] = $match[3].$rest.')';
                 $match[4] = $match[4].$rest;
@@ -528,6 +532,34 @@ class BladeCompiler extends Compiler implements CompilerInterface
         }
 
         return $template;
+    }
+
+    /**
+     * Determine if the given expression has the same number of opening and closing parentheses.
+     *
+     * @param  string  $expression
+     * @return bool
+     */
+    protected function hasEvenNumberOfParentheses(string $expression)
+    {
+        $tokens = token_get_all('<?php '.$expression);
+
+        if (Arr::last($tokens) !== ')') {
+            return false;
+        }
+
+        $opening = 0;
+        $closing = 0;
+
+        foreach ($tokens as $token) {
+            if ($token == ')') {
+                $closing++;
+            } elseif ($token == '(') {
+                $opening++;
+            }
+        }
+
+        return $opening === $closing;
     }
 
     /**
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by taylorotwell** [comment]
Not worth it.

**Comment by taylorotwell** [comment]
If we are going to attempt to solve that issue should we not add a new test ensuring that issue is actually fixed?

**Comment by imanghafoori1** [comment]
@taylorotwell I think I am done with this PR for now.

**Comment by iman-watchenterprise** [comment]
The first commit solves the bug for ")" character, (without touching the regex pattern.)
 like: ```@include('a', ['  )))) '] )```
The second commit solves the bug for "(" character (but by tweaking the regex pattern.)

like:  ```@include('a', ['  (((( '] )```


- So if you really want to be stay safe and avoid touching the regex pattern, the first commit can be considered a remedy on its own, so you can merge the first one and discard the second.

**Comment by imanghafoori1** [comment]
The first commit solves the problem for closing chars (without touching the original regex)
like: ```@include('a', ['  :)))) '])```

The second commit solves the problem for opening chars (but tweaks the original regex)
like: ```@include('a', ['  :(((( '])```

So if you want to stay really safe and avoid touching the regex, you can keep the first and discard the second one.

File: src/Illuminate/View/Compilers/BladeCompiler.php [hunk 1]
```diff
@@ -499,16 +499,35 @@ protected function compileExtensions($value)
...
+                $matches[2][$i],
+                $matches[3][$i] ?: null,
+                $matches[4][$i] ?: null,
+            ];
+
+            // Here we check to see if we have properly found the closing parenthesis by
+            // regex pattern or not, and will recursively continue on to the next ")"
+            // then check again until the tokenizer confirms we found the right one
+            while (isset($match[4]) && Str::endsWith($match[0], ')') && ! $this->isProperMatch($match[0])) {
+                $rest = Str::before(Str::after($template, $match[0]), ')');
```
taylorotwell, line 521: I'm pretty confused by this Str::after statement. Can't the same statement appear twice in the same `$template`? Therefore, how would you know which one you're extracting after?

taylorotwell:
Can you provide some insight about what happens if it never finds a proper closing parenthesis? 

Also, do you have some before / after benchmarks of running view:cache with this change on a code base with a fairly large amount of templates?

src/Illuminate/View/Compilers/BladeCompiler.php
imanghafoori1: I think the problem of having 2 exact same statements with different results is tested here: `testNestedForelseStatementsAreCompiled`, and that is why I used `Str::replaceFirst` and not `str_replace`.

Here `Str::after` gives us the rest of the template, starting at where the regex match has ended. (Str::after matches the first occurrence) then from that, we fetch the first closing parenthesis.

After we replace the statement, we go for the next one.

taylorotwell:
Alright. Can you maybe add some sort of test case or behavior description for what happens if there is never a closing parenthesis found?

imanghafoori1:
I added the test for the unclosed blade tag. It seems that they are considered a tag with no parenthesis.

timacdonald:
I spent some time trying to break this today, but it seems to be holding up.

I also booted up a decently sized project from a few years ago that uses all the older style blade bits (i.e. not <x-whatever) and it seemed happy.

I could break it by passing through `'<?php '`, but I think that is to be expected. It also breaks slots, as you would expect.

rubenvanerk:
I think I may have found an edge case. The code below is broken on `v9.47`+, but works on `v9.46`.

```blade
<div>
    <span @class([
        'lt-text-xs lt-inline-block lt-py-1 lt-px-2 lt-rounded',
        'lt-text-green-600 lt-bg-green-50' => !@empty($success),
        'lt-text-yellow-600 lt-bg-yellow-50' => !@empty(
            $warning
        ),
        'lt-text-red-600 lt-bg-red-50' => !@empty($danger),
    ])>
        {{ $value }}
    </span>
</div>
```

This gets compiled to:
```
<span class="<?php echo \Illuminate\Support\Arr::toCssClasses([
        'lt-text-xs lt-inline-block lt-py-1 lt-px-2 lt-rounded',
        'lt-text-green-600 lt-bg-green-50' => !@empty($success),
        'lt-text-yellow-600 lt-bg-yellow-50' => !<?php if(empty(
            $warning
        )): ?>,
        'lt-text-red-600 lt-bg-red-50' => !<?php if(empty($danger)): ?>,
    ]) ?>">
```

And throws the error `syntax error, unexpected token "<"`

Context: https://github.com/lunarphp/lunar/pull/822

imanghafoori1:
@rubenvanerk I think the fix is somewhat easy.
Thanks for you report.

PR #45498, opened 2023-01-03
[9.x] Alternative to fix method contextual binding

imdhemy:
## Issue description
1. Create a class that has a method with a parameter that has a type hint.
 * For instance, a command class with a handle method that has a parameter with a type hint of an interface.
2. Configure a contextual binding for that parameter to a concrete class.
3. Try to run the command.
4. You will get an error that the given interface is not instantiable.

## Recreation example
The following commit reproduces the issue on a fresh laravel project.

## Error message

## PR description
This is because when the container tries to resolve the method dependencies, It encounters a `buildStack` that does not contain the class that contains the method to be called, as it was already popped from the build stack in a previous step.

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -650,14 +650,14 @@ class Container implements ArrayAccess, ContainerContract
     {
         $pushedToBuildStack = false;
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
PR has no description and recreation example.

imdhemy:
@taylorotwell
Updated the description.

PR #45500, opened 2023-01-03
[9.x] Fix bound method contextual binding

imdhemy:
## Issue description

Using contextual bindings with a method parameter will result in an error. This is because when the container tries to resolve the method dependencies, It encounters a `buildStack` that does not contain the class that contains the method to be called, as it was already popped from the build stack in a previous step.

## Recreation example
The following commit reproduces the issue on a fresh laravel project.

## Error message

## PR description

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -650,8 +650,13 @@ class Container implements ArrayAccess, ContainerContract
     {
         $pushedToBuildStack = false;
 
-        if (is_array($callback) && ! in_array($className = get_class($callback[0]), $this->buildStack, true)) {
+        if (is_array($callback) && ! in_array(
+            $className = (is_string($callback[0]) ? $callback[0] : get_class($callback[0])),
+            $this->buildStack,
+            true
+        )) {
             $this->buildStack[] = $className;
+
             $pushedToBuildStack = true;
         }
 
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -650,14 +650,14 @@ class Container implements ArrayAccess, ContainerContract
     {
         $pushedToBuildStack = false;
 
-        if(is_array($callback) &&  ! in_array($className = get_class($callback[0]), $this->buildStack, true)) {
+        if (is_array($callback) && ! in_array($className = get_class($callback[0]), $this->buildStack, true)) {
             $this->buildStack[] = $className;
             $pushedToBuildStack = true;
         }
 
         $result = BoundMethod::call($this, $callback, $parameters, $defaultMethod);
 
-        if($pushedToBuildStack){
+        if ($pushedToBuildStack) {
             array_pop($this->buildStack);
         }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't think contextual binding has ever been designed to actually support specifying contextual bindings at the method level?

I think it would require actual first-party support of that feature, like this: https://github.com/laravel/framework/pull/30542/files

taylorotwell:
For example, I would expect it to support syntax like:

`$app->when('Foo@handle')->needs(...)->give(...)`

taylorotwell:
Linking this here for my own reference:

https://github.com/laravel/framework/pull/30542/files

taylorotwell:
Would actually like to try to solve this.

taylorotwell:
Pushed a fix for static methods, which would be strings in call and not objects in the callback array.

taylorotwell:
I think this PR is likely better than the current situation. Ideally, we would be able to specify different implementations for different methods on the class but I think that is likely an edge case.

imdhemy:
I'd be happy to work on this. I will check it in the morning.

taylorotwell:
No need to do anymore work I don't think.

imdhemy:
I spent some time working on this today. As you already mentioned, we are already in a better situation with the current implementation.

---

Regarding the edge case, I started with a **controller**, and managed to make it work with the following syntax:

```php
$this->app->when('MyController@<methodName>`)->needs('<abstract>')->give('<concrete>');
```

By pushing the `MyController@<methodName>` to the container's build stack on the [Controller dispatcher](https://github.com/laravel/framework/blob/c2383f4bb9a2c0e61e46679649fc6db394f12ffd/src/Illuminate/Routing/ControllerDispatcher.php#L57).

---

Regarding the same edge case, I also managed to make it work with the following syntax:

```php
$this->app->when(`<concrete>@<method>`)->needs('<abstract>')->give('<concrete>');

// should be called as follows:
$app->call('Foo@method');
```

by pushing the `<concrete>@<method>` to the container's build stack on the call method of the container.

[1]
```php
// targeting something like `Foo@method`
if (is_string($callback) && ! in_array($callback, $this->buildStack, true)) {
    $this->buildStack[] = $callback;

    $pushedToBuildStack = true;
}
```

IMO, For the context of this PR, after pushing the code related to targeting `$app->call('Foo@method')` [1] to the container, we can merge this PR. And then I can prepare another PR for the controller dispatcher. I think the same technique can be used for all dispatchers if there are other dispatchers that need to be targeted.

**What do you think?**

taylorotwell:
I honestly think we should just merge this PR as is and not make any further changes for the time being because this PR solves most every use case people have wanted. I haven't seen many requests for method level contextual injection where multiple methods on the same class receive different types. 

If you really need that behavior you can just make a separate class.

michael-rubel:
@taylorotwell this breaks the package I've made to let people [specify the context in any place of the app](https://github.com/michael-rubel/laravel-enhanced-container#resolve-contextual-binding-outside-of-constructor).

Looks like global binding is no longer available for the same class if you use contextual one.
Fixed this by relying on [BoundMethod::call()](https://github.com/michael-rubel/laravel-enhanced-container/pull/17/commits/748dd7fb683e1162a678a6f218a617baed776812) directly instead of `app()->call()`.

PR #45502, opened 2023-01-03
[9.x] Added support for testing authorization usages

tkaratug:
This PR adds support for testing policy usages in controllers easily.

This can be done using `expectsAuthorization()` method which is added to `MockApplicationServices` trait. With this method there is no need to write separate tests for unauthorized users.

Let's say we have a `PostController` and `index()` method to get all posts:

```php
class PostController extends Controller
{
    public function index()
    {
        $this->authorize('view-any', Post::class);

        return PostResource::collection(Post::paginate(50));
    }
}
```

Authorization can be tested in `PostControllerTest` as following:

```php
class PostControllerTest extends TestCase
{
    public function testUserCanViewPosts(): void
    {
        Post::factory()->count(5)->create();

        $this->expectsAuthorization('view-any', Post::class);

        $response = $this->postJson(route('posts.index'));

        $response->assertOk();
    }
}
```

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/MocksApplicationServices.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/MocksApplicationServices.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Foundation\Testing\Concerns;
 
+use Illuminate\Contracts\Auth\Access\Gate;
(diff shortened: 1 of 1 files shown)


taylorotwell:

taylorotwell:
Just create the user in such a state that the authorization passes / fails.

PR #37307, opened 2021-05-07
[6.x] Support mass assignment to SQL Server views

madbarron:
SqlServerGrammar assumes that the model is backed by a table, but some unfortunate souls attach to views in SQL Server. Since we now check mass assignment keys against the database columns, we need to be able to fetch the columns for views.

SQL Server will not allow a table and a view to have the same name in the same namespace, so this should not introduce any problems with matching multiple items.

Commit which introduced the problem: https://github.com/laravel/framework/commit/70d261d9832fc30575276fd9042b02f5debdad8f

PR #45503, opened 2023-01-04
[7.x] Support Views in SQLServerGrammer

JohnHHunter:
As @madbarron wrote in a [6.x](https://github.com/laravel/framework/pull/37307) and [8.x](https://github.com/laravel/framework/pull/37348) PR:
> SqlServerGrammar assumes that the model is backed by a table, but some unfortunate souls attach to views in SQL Server. Since we now check mass assignment keys against the database columns, we need to be able to fetch the columns for views.
> 
> This revision removes the "table" object type specification when hitting the DB for schema info, so that views also work. SQL Server will not allow multiple objects to have the same name in the same namespace, so this should not introduce any problems with matching multiple items.

Let's bring this change to Laravel Framework 7.x


driesvints:
7.x isn't maintained anymore sorry.

PR #45504, opened 2023-01-04
[9.x] Improve Facade docblock generator

timacdonald:
This PR improves the Facade docblock generator's capabilities by incorporating the PHPStan docblock parser. This gives us much more control and power.

Some generated types have changed to better support editors, as not everything supported in the standard class method docblock is supported in the class level `@method` docblocks.

## Question we need to answer

- [x] Does PHPStorm support Closure / callable params and return values in `@method` declarations?
- [x] Does PHPStorm support class-string in `@method` declarations?
- [x] Does PHPStorm support `string[]` array notation?

A note on redundant types: You may notice that some types are redundant, e.g. `mixed|null` or `mixed|\Closure`, where everything but mixed is redundant. This is intentional as it is something we do throughout the framework. I can easily strip out redundant types when `mixed` in present, but I don't think we want to do that, as per: https://github.com/laravel/framework/pull/45175


src/Illuminate/Support/Facades/Date.php
```diff
@@ -81,7 +81,7 @@
  * @method static \Illuminate\Support\Carbon today($tz = null)
  * @method static \Illuminate\Support\Carbon tomorrow($tz = null)
  * @method static void useMonthsOverflow($monthsOverflow = true)
- * @method static \Illuminate\Support\Carbon useStrictMode($strictModeEnabled = true)
+ * @method static void useStrictMode($strictModeEnabled = true)
```
timacdonald, line 84: All the changes to the `Date` facade are unrelated to this PR. The underlying class was updated.

src/Illuminate/Support/Facades/Cache.php
```diff
@@ -14,10 +14,10 @@
  * @method static \Illuminate\Cache\CacheManager extend(string $driver, \Closure $callback)
  * @method static bool has(array|string $key)
  * @method static bool missing(string $key)
- * @method static mixed get(array|string $key, mixed $default = null)
+ * @method static mixed get(array|string $key, mixed|\Closure $default = null)
  * @method static array many(array $keys)
  * @method static iterable getMultiple(iterable $keys, mixed $default = null)
- * @method static mixed pull(string $key, mixed $default = null)
+ * @method static mixed pull(array|string $key, mixed|\Closure $default = null)
```
timacdonald, line 20: This is an example of a redundant type that we are intentionally leaving in place.

src/Illuminate/Foundation/Vite.php
```diff
@@ -233,7 +233,7 @@ public function useStyleTagAttributes($attributes)
     /**
      * Use the given callback to resolve attributes for preload tags.
      *
-     * @param  (callable(string, string, ?array, ?array): array|false)|array|false  $attributes
+     * @param  (callable(string, string, ?array, ?array): (array|false))|array|false  $attributes
```
timacdonald, line 236: This change was required for PHPStan parser to correctly understand the return types of the callable.

bin/facades.php
```diff
@@ -185,70 +192,221 @@ function resolveDocParamType($method, $parameter)
...
+        $typeNode->name === 'TEnum' &&
+        $method->getDeclaringClass()->getName() === Request::class
+    ) {
+        return 'object';
+    }
+
+    echo 'Found unknown type: '.$typeNode->name;
+    echo PHP_EOL;
+    echo 'You will need to update the `handleUnknownIdentifierType` to handle this new type / generic.';
+    exit(1);
```
timacdonald, line 380: This will now error when encountering unknown generics / classes, which means we won't auto-commit wrong values.

bin/facades.php
```diff
@@ -185,70 +192,221 @@ function resolveDocParamType($method, $parameter)
...
+{
+    if (
+        in_array($method->getname(), ['pull', 'get']) &&
+        $method->getDeclaringClass()->getName() === Repository::class
+    ) {
+        return 'mixed';
+    }
+
+    echo 'Found unknown conditional type. You will need to update the `handleConditionalType` to handle this new conditional type.';
+    exit(1);
```
timacdonald, line 323: This will now error when encountering conditional return types, which means we won't auto-commit wrong values.

I don't think we can assume they should all default to `mixed` yet.

timacdonald:
@nunomaduro Do you have any insights into the question I have in the description? You've mentioned the `iterable<string>` issue, which we have addressed.

For now I have stripped them out - but if they are supported, I can restore them into the generated docblocks with ease.

canvural:
If PHPStorm support is the only concern, why not replace all docblocks with `@mixin ...`? Didn't test it, but maybe other editors also support auto-complete from `@mixin` tag.

nunomaduro:
> Does PHPStorm support Closure / callable params and return values in @method declarations?

No.

> Does PHPStorm support class-string in @method declarations?

No.

> Does PHPStorm support string[] array notation?

Yes.

---

@timacdonald Things may have changed, so it may be worth to check @canvural's comment regarding what's the state if `@mixin` on IDEs. I know for sure that, `@mixin` as supported in PHPStorm and PHPStan.

timacdonald:
Thank you for checking that out, @nunomaduro. I've just restored the `type[]` type notation based on your feedback.

Regarding only using `@mixin`: that would be great, but right now it isn't supported by Intelephense, which I would say is the most popular / mainstream LSP language server. If we don't also give sensible consideration to that, and probably some other tooling, we are making the experience worse for a huge audience (VSCode, Vim / NeoVim, Sublime, Nova, etc).

I feel the current approach is still valid until we have more general support across the ecosystem for using `@mixin` in the manner we want to use it in for Facades.

Note that `@mixin` _is supported_ with Intelephense and a lot of tooling, but their implementation (and my understanding of the tag) is that non-static methods are not made statically available.

So with intelephsense...

```php
/** @mixin \Illuminate\Database\DatabaseManager */
class Foo {}


Foo::conn
         ^
// `connections` is not suggested.

(new Foo)->conn
               ^
// `connections` is suggested.
```

Until more tooling interprets `@mixin` as _"make static and non-static methods available statically"_ (which, IMO, doesn't make sense) or we see a `@mixin-static` tag for that use-case, I think we need to stick with the generated docblock.

nunomaduro (changes requested):
This looks good to me.

nunomaduro (approved):
This looks good to me.

iwasherefirst2:
@timacdonald thank you for this feature. Is there a reason why the `Log` facade is missing here? Because the Log facade contains currently some wrong type hints. My PR to fix those was closed with the comment that those are now autogenerated https://github.com/laravel/framework/pull/45736#issuecomment-1398141305

timacdonald:
@iwasherefirst2 The log facade is correct as per the underlying classes that it reads to generate the Facade dockblocks.

If you want to have the Facade docblocks changed, you will need to change the `LogManager` docblocks.

e.g.

https://github.com/laravel/framework/blob/9.x/src/Illuminate/Log/LogManager.php#L689-L699

PR #45505, opened 2023-01-04
[9.x] Improve `JsonResource::class` values

dinhquochan:
This PR improves the value of the `JsonResource::class`.

### Resource:

```php
return JsonResource::make([
  'array' => [1, 2, 3],
  'object' => (object) ['a' => 'b'],
  'arrayable' => $arrayableObject,
  'htmlString' => $htmlStringObject,
  'stringable' => $stringableObject,
]);
```

### Before

image

### After

image

Thanks.

### DIFF
--- a/src/Illuminate/Http/Resources/ConditionallyLoadsAttributes.php
+++ b/src/Illuminate/Http/Resources/ConditionallyLoadsAttributes.php
@@ -2,8 +2,10 @@
 
 namespace Illuminate\Http\Resources;
 
+use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
+use Stringable;
 
 trait ConditionallyLoadsAttributes
 {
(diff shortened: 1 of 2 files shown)


axlon:
This introduces a problem; because filtering happens before serialization `Arrayable` and `Stringable` will now take precedence over `JsonSerializable`.

For example:

```php
class Foo implements Arrayable, JsonSerializable
{
    public function toArray()
    {
        return ['bar'];
    }

    public function jsonSerialize()
    {
        return ['baz'];
    }
}
```

Before:

```json
["baz"]
```

After:

```json
["bar"]
```

This could be fixed by also including `JsonSerializable` and `Jsonable` in the filter method, but personally I think only jsonserializable data should be returned by a resource. If you want an arrayable object to be represented as an array, either make it jsonserializable or call `toArray()` on it manually.

ankurk91:
This can be a breaking change for many.

dinhquochan:
Yes @axlon, you're right. I wait for @taylorotwell's opinion.

dinhquochan:
Extra issue, when I add casting `HtmlString::class` instance to a property of Model. It's still empty `{}` when encoding. I think it causes by `json_encode` in `Model` don't know a property is stringable.

taylorotwell:

PR #45510, opened 2023-01-04
Added year() and current_year() helpers

sfolador:
In this pull request new helpers have been added: `year()` and `current_year()`. 

Both helpers returns the current year and can be used as a shortcut to retrieve the current year:

```php

year(); // returns 2023

current_year(); // returns 2023

```

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -573,6 +573,32 @@ if (! function_exists('now')) {
     }
 }
 
+if (! function_exists('year')) {
+    /**
+     * Gets the current year.
+     *
+     * @param  \DateTimeZone|string|null  $tz
+     * @return int
+     */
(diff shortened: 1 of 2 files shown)


sfolador:
Hi @taylorotwell, no problem for the pull request. 
Is this something that is not interesting for the framework or did I write the pull request in a wrong format? 
Just asking so that I can avoid wrong pull requests in the future.
Thank you very much!

dennisprudlo:
Adding helpers is always a tricky one because those functions are global. So helpers are usually only added when they are really necessary or will be used by a lot of people and have a great value.

A `year()` helper adds only litte value over `now()-year` but blocks this global name for everyone using laravel. If this would have been merged people would create PRs for `month()`, `day()`, etc.

sfolador:
Understood, makes sense. Thank you very much for your explanation!

PR #45522, opened 2023-01-05
When & unless helpers

Propaganistas:
These are two little helpers I often use to get rid of intermediary variable assignments.

Example:

```php
$id = $model->getOriginal('related_id');
$related = $id ? Related::find($id) : null;

// becomes
$related = when($model->getOriginal('related_id'), function($value) {
    return Related::find($value)
});

// or using short syntax
$related = when($model->getOriginal('related_id'), fn($value) => Related::find($value));
```

The callback's value is returned if the provided value is truthy.

It has the same signature as other `when()` and `unless()` methods throughout the framework. So a default value can be specified (callable if you like):

```php
$related = when($model->getOriginal('related_id'), function($value) {
    return Related::find($value)
}, 'something else');

$related = when($model->getOriginal('related_id'), function($value) {
    return Related::find($value)
}, function($value) {
    return Related::first();
});
```

### DIFF
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -395,6 +395,50 @@ if (! function_exists('transform')) {
     }
 }
 
+if (! function_exists('unless')) {
+    /**
+     * Apply the callback if the given "value" is (or resolves to) falsy and return it.
+     *
+     * @param  mixed  $value
+     * @param  callable  $callback
+     * @param  mixed  $default
+     * @return mixed
+     */
+    function unless($value, callable $callback, $default = null) {
+        if (! $value) {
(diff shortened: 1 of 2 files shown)


crynobone:
already possible with `transform()`?

```php
$id = $model->getOriginal('related_id');
$related = $id ? Related::find($id) : null;

// becomes
$related = transform($model->getOriginal('related_id'), function($value) {
    return Related::find($value)
});

// or using short syntax
$related = transform($model->getOriginal('related_id'), fn($value) => Related::find($value));
```

Propaganistas:
You're totally right. Can't believe I overlooked that one for so long. Thanks

PR #45524, opened 2023-01-05
[9.x] Add Stringable surround method

btaskew:
Adds a `surround` method to the `Stringable` support class that will surround the string with the given values.

```php
return str('foo')->surround('bar')

// barfoobar
```

Basically a shorthand method for
```php
return str('foo')->append('bar')->prepend('bar')
```

I've found this particularly useful when building query params for a like query that allows matching either side:
```php
User::where('email', 'like', str('gmail')->surround('%'))
```

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -452,6 +452,17 @@ class Stringable implements JsonSerializable
     }
 
     /**
+     * Surround the string with the given values.
+     *
+     * @param  array  $values
+     * @return static
+     */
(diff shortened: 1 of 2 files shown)


dennisprudlo:
You can use `wrap` for that:
```php
str('gmail')->wrap('%')
```

btaskew:
Ah thanks @dennisprudlo. I thought there must already be a method but couldn't find it anywhere. `wrap` doesn't appear to be listed on https://laravel.com/docs/9.x/helpers

Closing PR as no longer needed

PR #45533, opened 2023-01-05
[9.x] Add Decimal to list of Numeric rules

### LINKED ISSUE #45532: The `size/max/min` validation rules perform incorrect validation with `decimal` rule
- Laravel Version: 9.26.1
- PHP Version: 8.1.13
- Database Driver & Version: Postgres 14

### Description:
Hello,

The docs for the `size` validation rule show that when used with `integer` or `numeric` it validates against an integer value:

> For numeric data, value corresponds to a given integer value (the attribute must also have the numeric or integer rule).

However, when `size` used with the new `decimal` rule it is using string validation (that is, number of characters). The expected behaviour is that `size/min/max` would be the same with `decimal` as with `numeric`.

### Steps To Reproduce:

```
Validator::make(['decimal' => 250.0], [
  'decimal' => ['required', 'decimal:0,2', 'max:200'], // passes
  'decimal' => ['required', 'numeric', 'max:200'], // fails
  ])->validate()
```

calebdw:
Fixes #45532

Added `Decimal` to list of numeric rules, added tests.


driesvints:
Thanks!

PR #45535, opened 2023-01-05
[9.x] Remove `__construct()` function to accommodate most use cases.

r-kujawa:
A small DX update that I feal might accomodate to most use cases.

When I generate a new event listener using the `php artisan make:listener` command, the first thing I do is remove the `__construct()` method from it as it is very rarely used. I believe that 90%+ of developers either do the same or leave it blank, while 10%- of devs would probobly make use of the method.

🔴  Haven't actually done the research, this is mostly based on my personal experience.

### DIFF
--- a/src/Illuminate/Foundation/Console/stubs/listener.stub
+++ b/src/Illuminate/Foundation/Console/stubs/listener.stub
@@ -9,16 +9,6 @@ use Illuminate\Queue\InteractsWithQueue;
 class {{ class }}
 {
     /**
-     * Create the event listener.
-     *
-     * @return void
-     */
-    public function __construct()
-    {
-        //
-    }
-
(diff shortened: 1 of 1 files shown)


yoeriboven:
If this doesn’t get merged you could do this on your project: https://laravel.com/docs/9.x/artisan#stub-customization

r-kujawa:
> If this doesn’t get merged you could do this on your project: https://laravel.com/docs/9.x/artisan#stub-customization

Nice! Didn't know this was a thing 😃 . I still believe this would be helpful for other devs, not a big DX improvement, but a simple nice to have (or not have 😅).

taylorotwell:
Feel free to remove from your own stubs.

PR #45536, opened 2023-01-05
[9.x] Add missing and missingAny Collection methods

denniseilander:
Hey there!

I'd like to propose adding the `missing()` and `missingAny()` methods to the Collection class in.

I found that I was using the inverse of the `has()` method (`! $collection->has()`) quite frequently in my code and thought it would be more convenient to have a dedicated method for this purpose.

Here's a simple example of how the `missing()` and `missingAny()` methods could be used:

**missing()**
```php
$collection = collect(['apple', 'banana', 'orange']);

if ($collection->missing('mango')) {
    // true, because mango is missing
}

if ($collection->missing(['orange', 'mango'])) {
    // false, because orange is present
}
```

**missingAny()**
```php
$collection = collect(['apple', 'banana', 'orange']);

if ($collection->missingAny(['banana', 'mango'])) {
    // true, because at least mango is missing
}

if ($collection->missingAny(['banana', 'orange'])) {
    // false, because both banana and orange are present
}
```
In addition to being more convenient to use, these methods fit nicely with the existing `has()` and `hasAny()` methods in terms of naming and functionality.

I'd love to see these methods added to the Collection class. Let me know what you think!

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -595,6 +595,48 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
+     * Determine if an item is missing from the collection by key.
+     *
+     * @param  TKey|array<array-key, TKey>  $key
+     * @return bool
+     */
+    public function missing($key)
+    {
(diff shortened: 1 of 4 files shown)


taylorotwell:

PR #45543, opened 2023-01-06
[9.x] Wrap the json_encode call with the echoFormat

### LINKED ISSUE #45542: @json directive doesn't render with the correct $echoFormat
- Laravel Version: 9.46.0
- PHP Version: 8.1.10
- Database Driver & Version: mysql 8.0.30

### Description:

The BladeCompiler class has the method `withDoubleEncoding`. This method configures the `$echoFormat` variable in such a way that subsequent calls to `htmlspecialchars` from the `e` echo helper method used by Blade, are called with the `$double_encode` argument set to `true`.

This works fine when using the regular echo syntax of Blade `{{ json_encode($data) }}`, however it doesn't apply at all when using the json directive `@json($data)`.

### Steps To Reproduce:

Let's say we have the given example data `$data`:

```php
$data = ['foo' => 'bar'];
```

And the following view that get's passed the `$data` variable:

```html
<some-vue-component v-model="@json($data)"></some-vue-component />
{{-- vs --}}
<some-vue-component v-model="{{ json_encode($data) }}"></some-vue-component />
```

I would expect both to produce the same output. However, the compiled output looks like the following:

```html
<some-vue-component v-model="{"foo":"bar"}"></some-vue-component />

<some-vue-component v-model="{&quot;foo&quot;:&quot;bar&quot;}"></some-vue-component />
```

thomas-alrek:
This commit updates the CompilesJson concern to use the BladeCompiler's current `echoFormat` when compiling the json expression. See https://github.com/laravel/framework/issues/45542

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesJson.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesJson.php
@@ -33,14 +33,15 @@ trait CompilesJson
     /**
      * Wraps the given value in a json_encode function call.
      *
-     * @param  string     $value
-     * @param  int        $options
-     * @param  int|string $depth
+     * @param  string  $value
+     * @param  int  $options
+     * @param  int|string  $depth
      * @return string
      */
     protected function wrapJsonHandler($value, $options, $depth)
     {
         $value = "json_encode($value, $options, $depth)";
+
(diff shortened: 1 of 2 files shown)


taylorotwell:
Could be a breaking change to existing applications. I suggest just echoing it directly.

PR #45546, opened 2023-01-06
[10.x] Revert UUID v7 changes

driesvints:


Bosphoramus:
so no UUID v7 support? 👀

driesvints:
@Bosphoramus You can easily opt into it yourself.

Bosphoramus:
oh i see @driesvints, so this pr only changes the default UUID version back to 4 but still allows v7 to be used?
how can we opt to use v7? will this be documented on the docs? 
Thank you!

driesvints:
Here's the docs on that: https://laravel.com/docs/9.x/eloquent#uuid-and-ulid-keys

PR #45553, opened 2023-01-07
change method to protected since its only use inside the middleware

andiLeong:


taylorotwell:
Potentially breaking change.

PR #45556, opened 2023-01-07
[9.x] Customize maintenance exception

abenevaut:
That PR allows to customize maintenance exception by implementing `public function abort(array $data): void;` in `PreventRequestsDuringMaintenance.php`.

Like this, the exception could mastered by developper and if customized the response could be too (add headers, choose view etc..) without risk to break the complete middleware logic.

My exception implementation to illustrate
```
<?php

namespace Exceptions;

use Illuminate\Http\Request;
use Inertia\Inertia;
use Symfony\Component\HttpKernel\Exception\ServiceUnavailableHttpException as ParentHttpException;

final class ServiceUnavailableHttpException extends ParentHttpException
{
    public function __construct(
        private Request $request,
        array $headers,
        string $message = 'Service Unavailable',
        \Throwable $previous = null
    ) {
        parent::__construct($headers['Retry-After'], $message, $previous, 503, $headers);
    }

    public function report()
    {
        Log::channel('info')->log($this->getMessage(), $this->context());
    }

    public function render()
    {
        $seconds = $this->getHeaders()['Retry-After'];

        return Inertia::render('Errors/503', [
            'message' => trans('exception.service_unavailable', [
                'seconds' => $seconds,
                'minutes' => ceil($seconds / 60),
            ]),
        ])
            ->toResponse($this->request)
            ->setStatusCode($this->getStatusCode())
            ->withHeaders($this->getHeaders());
    }

    public function context()
    {
        return [
            'sessions.id' => $this->request->session()->getId(),
        ];
    }
}

```

### DIFF
--- a/src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php
+++ b/src/Illuminate/Foundation/Http/Middleware/PreventRequestsDuringMaintenance.php
@@ -59,8 +59,8 @@ class PreventRequestsDuringMaintenance
 
             if (isset($data['redirect'])) {
                 $path = $data['redirect'] === '/'
-                            ? $data['redirect']
-                            : trim($data['redirect'], '/');
+                    ? $data['redirect']
+                    : trim($data['redirect'], '/');
 
                 if ($request->path() !== $path) {
                     return redirect($path);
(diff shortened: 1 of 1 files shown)


ankurk91:
Can you please remove all those unnecessary changes made by your IDE, specially indentation .

driesvints:
Please revert indentations etc and mark as ready.

PR #45558, opened 2023-01-09
[9.x] Add `setVisible` and `setHidden` to Eloquent Collection

jessarcher:
This PR adds `setVisible` and `setHidden` methods to Eloquent Collections.

We already have the `makeVisible` and `makeHidden` methods, but `makeVisible` is only useful when a property is usually hidden or when the model has a non-empty `$visible` property, and `makeHidden` is only useful when you want to be explicit about what is excluded, not what is included.

Example:

I want to return an array of users with just the ID and name, hiding other attributes such as `email` and any future attributes that may be added.

```php
// This is effectively a no-op because these fields are not typically hidden.
$users->makeVisible(['id', 'name'])->toArray(); ❌
/*
[
    [
        'id' => 1,
        'name' => 'Test User',
        'email' => 'test@example.com'
    ]
]
*/

// This only gives us the desired result until a new non-hidden attribute is added to the model.
$users->makeHidden(['email'])->toArray(); 😐️
/*
[
    [
        'id' => 1,
        'name' => 'Test User',
    ]
]
*/

// This allows us to be explicit about the data we want, and it won't leak as new attributes are added to the model.
$users->setVisible(['id', 'name'])->toArray(); ✅
/*
[
    [
        'id' => 1,
        'name' => 'Test User',
    ]
]
*/
```

This is especially useful when passing collections to Inertia.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Collection.php
+++ b/src/Illuminate/Database/Eloquent/Collection.php
@@ -528,7 +528,7 @@ class Collection extends BaseCollection implements QueueableCollection
     /**
      * Set the visible attributes across the entire collection.
      *
-     * @param  array<string>  $visible
+     * @param  array<int, string>  $visible
      * @return $this
      */
     public function setVisible($visible)
--- a/src/Illuminate/Database/Eloquent/Collection.php
+++ b/src/Illuminate/Database/Eloquent/Collection.php
@@ -539,7 +539,7 @@ class Collection extends BaseCollection implements QueueableCollection
     /**
      * Set the hidden attributes across the entire collection.
      *
-     * @param  array<string>  $hidden
+     * @param  array<int, string>  $hidden
      * @return $this
      */
     public function setHidden($hidden)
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Eloquent/Collection.php
```diff
@@ -525,6 +525,28 @@ public function makeVisible($attributes)
         return $this->each->makeVisible($attributes);
     }
 
+    /**
+     * Set the visible attributes across the entire collection.
+     *
+     * @param  array<string>  $visible
```
nunomaduro, line 531:
```suggestion
     * @param  array<int, string>  $visible
```

nunomaduro (changes requested):

src/Illuminate/Database/Eloquent/Collection.php
jessarcher: Should we also make this change to the underlying method in `HidesAttributes`?

https://github.com/laravel/framework/blob/a615378002db978339493e9af05be9c81cf3187f/src/Illuminate/Database/Eloquent/Concerns/HidesAttributes.php#L54-L60

nunomaduro: @jessarcher Yes please.

PR #45577, opened 2023-01-09
[9.x] Add "incrementEach" to the Query\Builder

imanghafoori1:
This is useful when the user wants to atomically increment multiple columns by a single query.
For example, subtract from `wallet1` and add the same amount to `wallet2` in order to transfer money.
```php
DB::table('users')->where('id', 2)->incrementColumns([
     'wallet_1' => -30,
     'wallet_2' => 30,
], ['updated_at' => now()]);
```
or keep a proper update of pre-calculated  columns:
```php
DB::table('products')->where('id', 2)->incrementColumns([
     'in_store' => 2,
     'in_stock' => 3,
     'total' => 5,
], ['updated_at' => now()]);
```

- The logic can be reused in `increment` to avoid code duplication.
- The method is named `incrementColumns` to indicate that it "increments many columns" (not sure if it is the best name).
- A comprehensive set of integration tests is included. Since the `increment` now uses `incrementColumns` the tests also cover it as well, plus some unit tests

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -3393,32 +3393,6 @@ class Builder implements BuilderContract
     }
 
     /**
-     * Atomically increments columns values by the given amounts.
-     *
-     * @param  array<string, float|int|numeric-string>  $columns
-     * @param  array<string, mixed>  $extra
-     * @return int
-     *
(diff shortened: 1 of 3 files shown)


taylorotwell:
Renamed to `incrementEach`.

PR #45578, opened 2023-01-09
[9.x] Add queue notification stubs

stfndamjanovic:
This is adding the possibility to generate queue notifications by command, just like we have for making listeners.
```
php artisan make:notification InvoicePaid --queued
```
I wasn't sure if there are any tests regarding stubs already. If yes, please let me know to update the PR.

### DIFF
--- a/src/Illuminate/Foundation/Console/NotificationMakeCommand.php
+++ b/src/Illuminate/Foundation/Console/NotificationMakeCommand.php
@@ -102,6 +102,12 @@ class NotificationMakeCommand extends GeneratorCommand
      */
     protected function getStub()
     {
+        if ($this->option('queued')) {
+            return $this->option('markdown')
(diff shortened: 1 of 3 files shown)


ankurk91:
What about making it queued by default? May be in v10.x?

What about doing the same for Mails ?

ajmeireles:
> This is adding the possibility to generate queue notifications by command, just like we have for making listeners.
> 
> ```
> php artisan make:notification InvoicePaid --queued
> ```
> 
> I wasn't sure if there are any tests regarding stubs already. If yes, please let me know to update the PR.

What is the effort to queue a notification? Just add `implements ShouldQueue`, so for me this PR is :-1: 

... Unless `--queue` is also implemented in all that could be queued (like listeners too). Why not? :rocket:

taylorotwell:

taylorotwell:
I think it can just be easily marked as queued after generation.

PR #45580, opened 2023-01-10
[9.x] Return rounding to decimal cast

barryvdh:
Alternative to https://github.com/laravel/framework/pull/45456

The initial problem was that large numbers/precision was not correct. This reverts the changes to rounding and use bcmath when preferred.
For 9.x just let it lose precision, for 10.x perhaps throw an exception, if you want to only suggest ext-bcmath instead of requiring.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1324,10 +1324,10 @@ trait HasAttributes
     {
         if (extension_loaded('bcmath')) {
             if (((string) $value)[0] != '-') {
-                return bcadd($value, '0.' . str_repeat('0', $decimals) . '5', $decimals);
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
```diff
@@ -1338,10 +1338,6 @@ protected function asDecimal($value, $decimals)
             throw new RuntimeException('The "decimal" model cast is unable to handle string based floats with exponents.');
         }
 
-        if ($decimals >= ini_get('precision')) {
-            throw new RuntimeException('$decimals precision is higher then floating point precision, please use ext-bcmath.');
```
barryvdh: Removed it for now in 9.x

timacdonald:
@barryvdh keen to hear your thoughts on: https://github.com/laravel/framework/pull/45602

barryvdh:
Looks a lot cleaner :)

PR #45583, opened 2023-01-10
Revert "[9.x] Optimize `whereKey` method 🏃🏻‍♂️"

driesvints:
Reverts laravel/framework#45453


imanghafoori1:
@driesvints I think we do not need to revert since the fix is fairly simple. We just have to flatten the array of values in `whereIntegerInRaw`

PR #45586, opened 2023-01-10
[9.x] add `alwaysBail` option for validator

emargareten:
By default Inertia shares only the first validation error, so there is no point in further validating the field if won't be displayed (especially if the following validation rules are querying the database).

This PR adds an option to always stop running the validation rules for the field after the first failure, without having to specify `bail` it in the validation rules.

Usage:
AppServiceProvider.php
```php 
use Illuminate\Support\Facades\Validator;

public function boot()
{
    Validator::alwaysBail();

    // or pass in a boolean
    
     Validator::alwaysBail(! Str::startsWith(request()->path(), 'api'));
}
```

### DIFF
--- a/src/Illuminate/Support/Facades/Validator.php
+++ b/src/Illuminate/Support/Facades/Validator.php
@@ -11,6 +11,7 @@ namespace Illuminate\Support\Facades;
  * @method static void replacer(string $rule, \Closure|string $replacer)
  * @method static void includeUnvalidatedArrayKeys()
  * @method static void excludeUnvalidatedArrayKeys()
+ * @method static void alwaysBail(bool $alwaysBail = true)
  * @method static void resolver(\Closure $resolver)
  * @method static \Illuminate\Contracts\Translation\Translator getTranslator()
  * @method static \Illuminate\Validation\PresenceVerifierInterface getPresenceVerifier()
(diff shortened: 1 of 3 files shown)


ankurk91:
I always want this, since 99% of time we are showing first error from attribute.

emargareten:
@morloderex this is done globally once (besides `stopOnFirstFailure` is to stop validation for all following attributes while bail is for this attribute only see https://laravel.com/docs/9.x/validation#rule-bail)

imanghafoori1:
This method should accept a `Closure` which returns a boolean, and should get executed at the time the validation happens. Since when the boot is called it is sometimes too early to decide.

taylorotwell:
This is already achievable via:

```php
Validator::resolver(function (...$arguments) {
    return (new Validator(...$arguments))->stopOnFirstFailure();
});
```

lazer-hybiz:
@taylorotwell won't `stopOnFirstFailure` stop the entire validation vs `bail` stops validation for this attribute only?

from the docs:

> While the `bail` rule will only stop validating a specific field when it encounters a validation failure, the `stopOnFirstFailure` method will inform the validator that it should stop validating all attributes once a single validation failure has occurred:

Issue #45593, opened 2023-01-10
BelongsToMany: custom pivot model with incrementing IDs doesn't select IDs by default


levu42:
- Laravel Version: 9.46.0
- PHP Version: 8.1.13

### Description:

When I create a many to many relationship with a custom pivot model, where I set `public $incrementing = true`, I expect to be able to access the ID using `$related->pivot->id`, but the `id` column isn't fetched unless I additionally explicitly call `->withPivot('id')` on the relation.

### Steps To Reproduce:

See [levu42/pivot-id-bug](https://github.com/levu42/pivot-id-bug/blob/main/tests/Unit/ExampleTest.php).

driesvints:
Looks like we won't be changing anything here for now. It's best that you create an actual model for this pivot.

PR #45594, opened 2023-01-10
[9.x] BelongsToMany with custom pivot model: load key column if model is incrementing

levu42:
With a custom pivot model that has `$incrementing = true`, the key column should be selected by default instead of having to do a manual `->withPivot('id')`. I described it more in #45593 with an example repository linked.


taylorotwell:

PR #45600, opened 2023-01-10
 [9.x] Allow modification of the enum validation method

phh:
The [current implementation of enum validation](https://github.com/hobbii/laravel-framework/blob/b6b4a19673d65401f99bee13b0db64b6c6492866/src/Illuminate/Validation/Rules/Enum.php) only relies on checking the value of the enum. However, a common pattern today is to use the value for storing in the database while using another static method for the actual lookup of the enum.

This should be used by client of the API sending a locale code (like `en`) instead of providing the value matched by the enum (`1`).

Example:

```php
enum LanguageEnum: int
{
    case English = 1;
    case German = 2;
    case French = 3;

    public function getCode(): string
    {
        return match ($this) {
            self::English => 'en',
            self::German => 'de',
            self::French => 'fr',
        };
    }

    public static function tryFromCode(string $code): ?LanguageEnum
    {
        return collect(LanguageEnum::cases())->first(fn (LanguageEnum $language) => $language->getCode() === $code);
    }

    public static function tryFromName(string $name): ?ContextTypeEnum
    {
        return collect(ContextTypeEnum::cases())->first(fn (ContextTypeEnum $language) => $language->name === $name);
    }
}
```

Before I could only rely on the integer values for the lookup:
```php
$rules = [new Enum(LanguageEnum::class)];
```

With this change I can instead rely on other methods for the lookup:

```php
$rules = [new Enum(LanguageEnum::class, 'tryFromCode')];
$rules = [new Enum(LanguageEnum::class, 'tryFromName')];
```

### DIFF
--- a/src/Illuminate/Validation/Rules/Enum.php
+++ b/src/Illuminate/Validation/Rules/Enum.php
@@ -15,14 +15,28 @@ class Enum implements Rule
     protected $type;
 
     /**
+     * The method used to translate a value into the enum.
+     *
+     * @var string
+     */
+    protected $method;
+
+    /**
      * Create a new rule instance.
      *
      * @param  string  $type
+     * @param  string  $method
      * @return void
      */
-    public function __construct($type)
+    public function __construct($type, $method = 'tryFrom')
     {
         $this->type = $type;
+        $this->method = $method;
+    }
+
+    public function method($method)
+    {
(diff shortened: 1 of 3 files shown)


taylorotwell:
I would just build a custom class based or closure validation rule. Should be pretty easy.

PR #45602, opened 2023-01-11
[9.x] Fix decimal cast, again

timacdonald:
For 9.x, alternative to: https://github.com/laravel/framework/pull/45580
For 10.x, Supersedes: https://github.com/laravel/framework/pull/45457

This implementation handles both rounding and arbitrary sized floats, restoring the rounding functionality while addressing the large float issue.

It supports systems with `bcmath` or `gmp`, and delegates to pure PHP implementations when these are not found, meaning that it does not introduce any hard extension requirements to the framework.

This PR introduces [`brick/math`](https://github.com/brick/math) in order to handle the rounding.

Some points to consider on introducing a new framework dependency...

### Adding to the dependency tree

`brick/math` is already a sub-dependency of `laravel/framework` via [`ramsey/uuid`](https://github.com/ramsey/uuid/blob/4a14ce0c626423054c2b45760ceee248d87475ae/composer.json#L14). [`brick/math` only requires the JSON extension](https://github.com/brick/math/blob/master/composer.json#L17-L18) and has no sub-dependencies.

### Package is pre-1.0

From the readme:

> While this library is still under development, it is well tested and considered stable enough to use in production environments.

This package is already used in `brick/money`, which is an extremely popular money handing library, second to `moneyphp/money`.

Screen Shot 2023-01-11 at 12 20 31 pm

### Used for only one method

Introducing an entire dependency for a single method feels like a lot, however this library could be used to improve other parts of the framework.

- The `multiple_of` rule could potentially drop it's dependency on the `bcmath` extension.
- Addition improvements I've mentioned in our internal Slack.

### Nice to have

I believe it is nice to have a strong math library in the framework anyway. Requiring it directly may lead others to the package and learn to improve their own handling of large numbers within their own applications.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -18,6 +18,7 @@
         "php": "^8.0.2",
         "ext-mbstring": "*",
         "ext-openssl": "*",
+        "brick/math": "^0.10.2",
         "doctrine/inflector": "^2.0",
         "dragonmantank/cron-expression": "^3.3.2",
         "egulias/email-validator": "^3.2.1",
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -33,8 +33,6 @@ use LogicException;
 use ReflectionClass;
 use ReflectionMethod;
 use ReflectionNamedType;
-use RuntimeException;
-use TypeError;
 
 trait HasAttributes
 {
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -3,6 +3,8 @@
 namespace Illuminate\Database\Eloquent\Concerns;
 
 use BackedEnum;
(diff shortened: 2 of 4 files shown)


barryvdh:
Yeah I think this is a lot cleaner, while not breaking BC :)

I understand the dependency concern, but if it's already loaded anyways.. Only thing I might consider is handling the exception differently, eg catch the bigmath exception and throw a Laravel exception. In that case you keep the implementation of bigmath internal, in case you need to swap it out later (or bigmath changes it).

timacdonald:
Good thinking. I have wrapped `brick/math` exceptions in a first party exception.

This does not meant it is not breaking if the underlying exception changes, as the "previous" exception will have changed and code may rely on its state, but at least it can continue to be caught via `try` / `catch`.

I've placed the exception in `Illuminate\Support` as the framework will need the exception in other parts of the framework, not just `Illuminate\Database`. `Illuminate\Database` required `Illuminate\Support` already.

timacdonald:
Thanks for the feedback and help on this @barryvdh

PR #45603, opened 2023-01-11
[9.x] Prevent queries during response preparation

timacdonald:
Related PR of Octane: https://github.com/laravel/octane/pull/631

## The problem

In many contexts, queries within a response object can be problematic and hard to detect. The response object may be a `view()`, a `JsonResource`, or something else entirely.

Two database queries are run in the following view. For some applications this may be totally fine, while others may find this problematic from a performance / resources / conventions perspective.

```blade
@foreach(['users' => 'Users', 'teams' => 'Teams'] as $table => $title)
  <dl>
      <dt>Total {{ $title }}</dt>
      <dd>{{ DB::table($table)->count() }}</dd>
  </dl>
@endforeach
```

## The solution

To help applications detect these issues, I propose we introduce a new opt-in mechanism that will throw exceptions when a query is run while the response is being prepared.

```php
// in a service provider or middleware

App::preventQueriesWhilePreparingResponse();
```

Response preparation involves the following code: https://github.com/laravel/framework/blob/c0affdde5074514655947ecaaecc2f58ef129623/src/Illuminate/Routing/Router.php#L885-L914

If an exception is thrown while the response is passing through this method, an exception will be thrown.

This means apps can throw if a query occurs in a blade view, as seen above, or while a `JsonResource` is being resolved. The following resource would throw an exception.

```php
class TeamResource extends JsonResource
{
    {
        return [
            'name' => $this->name,
            'member_count' => $this->members()->count(), // query
        ];
    }
}
```

As soon as this resource is a collection, like in a `UserResource` that includes all the users teams, this will trigger a query for every team included in the resource.

There are other places where "response preparation" occurs, such as the exception handler, but this feature does not concern itself with that. It is only related to values returned from a Controller.

```php
public function show(Team $team)
{
    return TeamResource::make($team);

    return view('teams.show', ['team' => $team]);
}
```

## Doing it manually

As part of this main feature is a lower level feature for preventing queries for certain parts of an application.

```php

DB::preventQueries(function () {
   // queries here will throw an exception.
});

// queries can run again.

DB::preventQueries();

// queries here will throw an exception.

DB::allowQueries();

// queries can run again.
```

This manual mechanism is handy when an application wants to prevent queries while a view is being prepared manually. Take the following example:

```php
public function update(UpdateTeamRequest $request, Team $team)
{
    $team->update($request->validated());

    return response()->json($team);
}
```

This will allow queries to run when the `toArray` method is called even with `preventQueriesWhilePreparingResponse` in use. In this type of scenario where you are preparing the response yourself you may opt-in to prevent queries:
```php
public function update(UpdateTeamRequest $request, Team $team)
{
    $team->update($request->validated());

    return DB::preventQueries(fn () => response()->json($team));
}
```

> **Note** This could impact things like dispatching jobs to the queue. This is a sharp knife. Use wisely.

### DIFF
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -211,6 +211,13 @@ class Connection implements ConnectionInterface
     protected static $resolvers = [];
 
     /**
+     * Indicate that queries should be prevented.
+     *
+     * @var bool
+     */
+    protected $preventQueries = false;
+
+    /**
      * Create a new database connection instance.
      *
      * @param  \PDO|\Closure  $pdo
@@ -705,6 +712,10 @@ class Connection implements ConnectionInterface
      */
     protected function run($query, $bindings, Closure $callback)
     {
+        if ($this->preventQueries) {
+            throw new RuntimeException("Queries are being prevented. Attempting to run query [{$query}].");
+        }
+
         foreach ($this->beforeExecutingCallbacks as $beforeExecutingCallback) {
             $beforeExecutingCallback($query, $bindings, $this);
         }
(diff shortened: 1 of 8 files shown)


bastien-phi:
Maybe a `App::allowQueriesWhilePreparingResponse()` is missing to rollback the behaviour.

imanghafoori1:
```php
App::preventBadPractices([
    'queriesInBlades',
    'envCallsOutsideConfig',
    'fullModelPathInPolymorphicRelation',
     .... 
]);
```

Rizky92:
Question, will this prevent unwanted extra queries in blade too? Something like this for example:

```blade
<ul>
    @foreach($someModelCollection as $model)
        <li>{{ $model->relationship()->count() }}</li>
    @endforeach
</ul>
```

timacdonald:
@Rizky92 yes it will.

Given your example, if you had `App::preventQueriesWhilePreparingResponse()` in a service provider an exception would be thrown when `$model->relationship()->count()` is invoked.

@imanghafoori1 I wouldn't say queries in the view are inherently a _bad_ practice. In certain teams, application sizes, and contexts they are perfectly fine. In others, not so much.

garygreen:
Does this prevent relationships from lazily loading? E.g.

```php
@foreach ($user->invoices as $invoice)
   ...
@endforeach
```

Would that throw if that relation wasn't eagerly loaded?

It is an interesting approach because it'll bring more awareness of db performance considerations when rendering views. I hope that this won't be frustrating to debug if it's throwing errors and you can't easily decipher where it came from exactly. This PR kinda reminds me of https://github.com/beyondcode/laravel-query-detector

PR #45604, opened 2023-01-11
[9.x] Log facade to accept any number of arguments of any type

dusterio:
Log facade is one of the crucial elements of Laravel development – this is how you get feedback when things go wrong, as well as when things go well or when you just want to get some extra visibility over your app.

Laravel framework binds LogManager class under the 'log' key in its container and currently it implements a PSR LoggerInterface:
```php
public function emergency(string|\Stringable $message, array $context = []): void;
```

Where we can pass a message as well as a context - any supplemental data that we need. If we refer to the [PSR documentation](https://www.php-fig.org/psr/psr-3/):

```
1.3 Context
Every method accepts an array as context data. This is meant to hold any extraneous information that does not fit well in a string. The array can contain anything. Implementors MUST ensure they treat context data with as much lenience as possible. A given value in the context MUST NOT throw an exception nor raise any php error, warning or notice.
```

Imagine a scenario like this:
```php
try {
    $details = $api->request(); // $api is some kind of a third-party SDK
} catch (ClientException $e) {
    // $e->getResponse() might have a nice JSON error message but might have a CDN's HTML error or something else
    Log::emergency("Oh no, we got a failed API request!", json_decode($e->getResponse()->getContent()));
}
```

A very usual situation indeed – somewhere in our code we have a failing third-party integration and we want to log the output from that API, perhaps there is something useful in there that's going to help us troubleshoot this later.

The problem here is that the second parameter of Log::emergency() call HAS to be an array and if json_decode() doesn't return an array for whatever reason, our Laravel application is going to crash!

```Illuminate\Log\LogManager::emergency(): Argument #2 ($context) must be of type array, bool given, called in vendor/laravel/framework/src/Illuminate/Support/Facades/Facade.php on line 338 in vendor/psy/psysh/src/Exception/TypeErrorException.php on line 53.```

Oh, that's not nice! Remember JavaScript's console.log() command? You can pass anything to it and it's never going to kill your application:
```js
console.log('plain message')
// plain message
console.log('plain message', 123, [])
// plain message 123 []
```

The idea is simple - because we use the logging capability to troubleshoot things, the last thing we want is for the application to crash BECAUSE of the logging. I believe I should be able to pass just about anything to Log facade and it should just work in one capacity or another.

This pull-request is one of the possible implementations where I removed the "array" type from the method signatures in LogManager class. This doesn't violate the LoggerInterface contract that the class implements. Moreover, it's been done before if we look closer in the Logger class:

```php
    /**
     * @param  \Illuminate\Contracts\Support\Arrayable|\Illuminate\Contracts\Support\Jsonable|\Illuminate\Support\Stringable|array|string  $message
     * @param  array  $context
     * @return void
     */
    public function info($message, array $context = []): void
    {
        $this->writeLog(__FUNCTION__, $message, $context);
    }
```

Compare it with the actual LoggerInterface signature:

```php
    /**
     * @param string|\Stringable $message
     * @param mixed[] $context
     *
     * @return void
     */
    public function info(string|\Stringable $message, array $context = []): void;
```

As we can see the string|\Stringable type was removed from the signature so that Laravel can use the formatMessage() method later to display Stringable and Jsonable objects nicely.

I propose we do the same with the $context with the same goal - make it a bit more flexible and convenient.

With the updates proposed, I can do stuff like this:
```php
Log::emergency("Hello", $results, 'Another message');
```

And it just all goes to the log driver, to the logs. I don't have to worry about the number of arguments I pass or the types of arguments I pass (which are hard to predict sometimes). I never have to worry about Log throwing an error again.

Any feedback, opinions, and alternative implementations are welcome! :)

### DIFF
--- a/src/Illuminate/Log/LogManager.php
+++ b/src/Illuminate/Log/LogManager.php
@@ -720,7 +720,7 @@ class LogManager
      */
     protected function mergeContexts($userPassedContext)
     {
-        return array_reduce($userPassedContext, function($carry, $argument) {
+        return array_reduce($userPassedContext, function ($carry, $argument) {
             if (is_array($argument)) {
                 $carry = array_merge($carry, $argument);
             } else {
--- a/src/Illuminate/Log/LogManager.php
+++ b/src/Illuminate/Log/LogManager.php
@@ -22,7 +22,7 @@ use Throwable;
 /**
  * @mixin \Illuminate\Log\Logger
  */
-class LogManager
+class LogManager implements LoggerInterface
 {
     use ParsesLogConfiguration;
 
@@ -595,10 +595,13 @@ class LogManager
      * System is unusable.
      *
      * @param  string  $message
+     * @param  array  $context
      * @return void
      */
-    public function emergency($message, ...$context): void
+    public function emergency($message, $context = []): void
     {
+        $context = array_slice(func_get_args(), 1);
+
         $this->driver()->emergency($message, $this->mergeContexts($context));
     }
 
@@ -609,10 +612,13 @@ class LogManager
      * trigger the SMS alerts and wake you up.
      *
      * @param  string  $message
+     * @param  array  $context
      * @return void
      */
-    public function alert($message, ...$context): void
+    public function alert($message, $context = []): void
     {
+        $context = array_slice(func_get_args(), 1);
+
         $this->driver()->alert($message, $this->mergeContexts($context));
     }
 
@@ -622,10 +628,13 @@ class LogManager
      * Example: Application component unavailable, unexpected exception.
      *
      * @param  string  $message
+     * @param  array  $context
      * @return void
      */
-    public function critical($message, ...$context): void
+    public function critical($message, $context = []): void
     {
+        $context = array_slice(func_get_args(), 1);
+
         $this->driver()->critical($message, $this->mergeContexts($context));
     }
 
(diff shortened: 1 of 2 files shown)


src/Illuminate/Log/LogManager.php
```diff
@@ -22,7 +22,7 @@
 /**
  * @mixin \Illuminate\Log\Logger
  */
-class LogManager implements LoggerInterface
+class LogManager
```
crynobone, line 25: Huge breaking change.

dusterio: @crynobone I thought so too but 1) couldn't find any evidence 2) all tests pass. Just to clarify, LogManager IS NOT a logger, it's a manager class.

There is one hacky way potentially to do this without removing the interface, and somebody did this with the $string parameter already which was supposed to be plain string but Laravel allows an object - we could keep the signature in place and then use func_get_args(), proof of concept:
```php
interface LoggerInterface
{
    public function emergency(string|\Stringable $message, array $context = []): void;
}


class SuperLogger implements LoggerInterface
{
   public function emergency($message, $context = []): void
   {
        print_r(func_get_args());
   }
}

(new SuperLogger)->emergency('Message', 123, 555);
```

just like Laravel already removed "string" from the $message type, if we remove "array" type from the second parameter - it's going to make it a bit more flexible

crynobone: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Foundation/Application.php#L1356

It is being used to resolve `log`, a huge breaking change.

dusterio: Hmm, I went through the code and it's just an alias - it doesn't actually employ PHP interface check, thats why everything works and tests pass.

But it's still kind of wrong I agree, maybe I should just quickly switch to the second option mentioned above

ankurk91:
The idea is good but it is breaking change.

PR #45607, opened 2023-01-11
[9.x] Fix session store not able to read number index array

andiLeong:
when we get the number index array from the sessionhandler eg FileSessionHandler
like this

    $data = [
        999 => 'bar',
        'foo' => 'bar2'
    ];
we are not able to get 999 out because in the store class loadSession method
we are using array_merge , array merge will reset the index
https://www.php.net/manual/en/function.array-merge.php

its still possible to put the number index key and value in the store put method

```
public function put($key, $value = null)
    {
        if (! is_array($key)) {
            $key = [$key => $value];
        }

        foreach ($key as $arrayKey => $arrayValue) {
            Arr::set($this->attributes, $arrayKey, $arrayValue);
        }
    }
```
so I can put
```
put('999',''value')
```
but inside put method we build an array, php will force to cast the string 999 to number index 999
```
$key = [$key => $value];
```
as describe here:
https://stackoverflow.com/questions/4100488/a-numeric-string-as-array-key-in-php

that's why later we trigger strore class save() method those array will save to the file
```
[999=> 'value'];
```

reason why I am not using the + operator
```
$this->attributes = $this->attributes + $this->readFromHandler();
```
because previously we are using array_merge 
the  + operator ($arr + $arr2) is not merging array

issue # 
45610

### DIFF
--- a/src/Illuminate/Session/Store.php
+++ b/src/Illuminate/Session/Store.php
@@ -98,22 +98,7 @@ class Store implements Session
      */
     protected function loadSession()
     {
-        if (empty($session = $this->readFromHandler())) {
-            return;
-        }
-
-        $attributes = [];
-        foreach ($session as $key => $value) {
-            if (array_key_exists($key, $this->attributes)) {
-                $attributes[$key] = $value;
-
-                continue;
-            }
-
-            $attributes[$key] = $value;
-        }
-
-        $this->attributes = $attributes;
+        $this->attributes = array_replace($this->attributes, $this->readFromHandler());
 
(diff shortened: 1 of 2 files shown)


lk77:
It's not supported by php

lk77:
Well for me it should be discarded in that case, and not put a `"0": "bar"` in the session

andiLeong:
@lk77 let see how it goes your suggestion could also be implemented

taylorotwell:
I would just give it a string key? `id-999` or something?

PR #45609, opened 2023-01-11
[9.x] Add alpha_num_ascii validation rule

fuwasegu:
The `alpha_num` rule must be used to validate that the field consists entirely of alphabetic and numeric characters.
However, in cultures such as Japan, where multibyte characters are used routinely, full-width alphanumeric characters such as `Ａ` (unicode: \uff21) or `１` (unicode: \uff11)  may be accepted as input instead of half-width alphanumeric characters such as `A` (unicode: A) or `1` (unicode: 1).
This can be converted to half-width alphanumeric characters programmatically, but it would be better if FormRequest could check it and return it as an error.

This PR implements a validation rule that checks for AlphaNum and single-byte alphanumeric characters (i.e., ASCII).

As an alternative, you can write `regex:/^[a-zA-Z0-9]$/`, but it would be easier to use if it is defined as a rule.

## Usage
```php
{
    return [
        'attribute_1' => ['alpha_num'], // 🙆‍♀ ABC123, 🙆‍♀ ＡＢＣ１２３
        'attribute_2' => ['alpha_num_ascii'], // 🙆‍♀ ABC123, 🙅‍♀ ＡＢＣ１２３
    ];
}
```

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -378,6 +378,22 @@ trait ValidatesAttributes
     }
 
     /**
+     * Validate that an attribute contains only one-byte alpha-numeric characters.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @return bool
+     */
+    public function validateAlphaNumAscii($attribute, $value)
+    {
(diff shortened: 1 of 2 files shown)


taylorotwell:
You can easily write a custom closure or class based rule for this. ✅

PR #45616, opened 2023-01-11
[9.x] Add `@includeWithOnly` blade directive

alies-dev:
There are times when we want to include a Blade partial, but we don't want to automatically pass all available variables to it. In other words, we don't want the Blade partial to have access to its surrounding context.  This is the default behaviour for the `@component` directive, but there is no such option for `@include*` directives.

```php
@includeWithOnly('greeting', [
    'name' => $user->name,
])
```

## Benefits
The `@includeWithOnly` directive will:
- Make it easier to move between views as the context is always passed explicitly.
- Help avoid name collisions, especially with globally available vars like `$errors`, `$app`, etc.
- Help avoid issues when `isset($variable)` is used in a view and a `$variable` with the same name is automatically and unintentionally passed from the parent context

## Why not just use `@component`?
- We do not always need slots
- `@include` syntax is simpler - it doesn't require a closing tag
- `@include` syntax has a better IDE support

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesIncludes.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesIncludes.php
@@ -29,6 +29,19 @@ trait CompilesIncludes
     }
 
     /**
+     * Compile the includeWithOnly statements into valid PHP.
+     *
+     * @param  string  $expression
+     * @return string
+     */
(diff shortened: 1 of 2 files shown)


taylorotwell:
Honestly I would just use a self-closing component 😬

```html
<x-greeting :$name />
```

PR #45622, opened 2023-01-12
[10.x] Update redirector to use status 303

emargareten:
Currently the Redirector class uses status code `302` for redirects (if no other status was passed in), this is an issue when making a `POST/PUT/PATCH/DELETE` request and the response is a redirect it will use the same request method for the redirect, this will cause a `405 Method not allowed`.

[Inertia already fixes this in most cases](https://github.com/inertiajs/inertia-laravel/blob/master/src/Middleware.php#L102) but it can still occur if the redirect runs from a middleware before the Inertia middleware e.g. `RedirectIfAuthenticated`.

To replicate, create a new breeze app, login, go to profile page and logout from different tab, now try to save profile details in the first tab.

### DIFF
--- a/src/Illuminate/Routing/Redirector.php
+++ b/src/Illuminate/Routing/Redirector.php
@@ -43,7 +43,7 @@ class Redirector
      * @param  mixed  $fallback
      * @return \Illuminate\Http\RedirectResponse
      */
-    public function back($status = 302, $headers = [], $fallback = false)
+    public function back($status = 303, $headers = [], $fallback = false)
     {
         return $this->createRedirect($this->generator->previous($fallback), $status, $headers);
     }
@@ -55,7 +55,7 @@ class Redirector
      * @param  array  $headers
      * @return \Illuminate\Http\RedirectResponse
      */
-    public function refresh($status = 302, $headers = [])
+    public function refresh($status = 303, $headers = [])
     {
         return $this->to($this->generator->getRequest()->path(), $status, $headers);
     }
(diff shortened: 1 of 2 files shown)


emargareten:
see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302
>Even if the specification requires the method (and the body) not to be altered when the redirection is performed, not all user-agents conform here - you can still find this type of bugged software out there. It is therefore recommended to set the 302 code only as a response for [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET) or [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD) methods and to use [307 Temporary Redirect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307) instead, as the method change is explicitly prohibited in that case.
In the cases where you want the method used to be changed to [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET), use [303 See Other](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303) instead. This is useful when you want to give a response to a [PUT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT) method that is not the uploaded resource but a confirmation message such as: 'you successfully uploaded XYZ'.

and here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303
>The HyperText Transfer Protocol (HTTP) 303 See Other redirect status response code indicates that the redirects don't link to the requested resource itself, but to another page (such as a confirmation page, a representation of a real-world object — see [HTTP range-14](https://en.wikipedia.org/wiki/HTTPRange-14) — or an upload-progress page). This response code is often sent back as a result of [PUT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT) or [POST](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/POST). The method used to display this redirected page is always [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/GET).

taylorotwell:
I'm pretty concerned about the amount of existing tests this would potentially break. IMO just pass the code you want for now.

PR #45625, opened 2023-01-12
[9.x] Fail queued job with a string message

pascalbaljet:
A shortcut so you can fail a job with a `string` message instead of manually instantiating an `Exception`.

```php
class MyJob implements ShouldQueue
{
    use Dispatchable, InteractsWithQueue, Queueable;

    public function handle()
    {
        // Before:
        $this->fail(new \Exception('Whoops!'));

        // After:
        $this->fail('Whoops!');
    }
}
```

### DIFF
--- a/src/Illuminate/Queue/InteractsWithQueue.php
+++ b/src/Illuminate/Queue/InteractsWithQueue.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Queue;
 
+use Exception;
 use Illuminate\Contracts\Queue\Job as JobContract;
 use InvalidArgumentException;
 use Throwable;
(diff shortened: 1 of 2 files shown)


WendellAdriel:
This seems nice but at least IMO it would be better to throw a custom exception instead of a generic one

SjorsO:
The `report()` and `throw_if()` helpers can also receive a string message (see https://github.com/laravel/framework/pull/35890). The only difference is they can also receive the FQN of an exception class, like this:

```php
throw_if($hasError, LogicException::class);
```

Supporting `$this->fail(LogicException::class)` is probably a good idea to keep things consistent.

PR #45627, opened 2023-01-13
[9.x] Add isLocked to Cache Facade

Riley19280:
Currently it is not possible to inquire about the lock status of a lock before attempting to acquire it. This is needed in order to perform conditional logic for dispatching unique jobs. For example, if I want to throw an error if an instance of a job already has a lock, that is not possible with the current implementation. 
```php
$myUniqueJob = new UniqueJob()

// Will dispatch, since no lock as been acquired yet
dispatch($myUniqueJob)

// Will not dispatch the job silently, since it cannot acquire a lock
dispatch($myUniqueJob)

// PR allows
throw_if($myUniqueJob->isLocked());
```

I have currently only implemented test cases for the array and redis drivers, but can handle the other drivers if this PR moves forward.

### DIFF
--- a/src/Illuminate/Support/Facades/Cache.php
+++ b/src/Illuminate/Support/Facades/Cache.php
@@ -49,6 +49,7 @@ namespace Illuminate\Support\Facades;
  * @method static string getPrefix()
  * @method static \Illuminate\Contracts\Cache\Lock lock(string $name, int $seconds = 0, string|null $owner = null)
  * @method static \Illuminate\Contracts\Cache\Lock restoreLock(string $name, string $owner)
+ * @method static bool isLocked(string $name)
  *
  * @see \Illuminate\Cache\CacheManager
  * @mixin \Illuminate\Cache\Repository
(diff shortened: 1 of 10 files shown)


src/Illuminate/Support/Facades/Cache.php
```diff
@@ -49,6 +49,7 @@
  * @method static string getPrefix()
  * @method static \Illuminate\Contracts\Cache\Lock lock(string $name, int $seconds = 0, string|null $owner = null)
  * @method static \Illuminate\Contracts\Cache\Lock restoreLock(string $name, string $owner)
+ * @method static bool isLocked(string $name)
```
timacdonald, line 52: As our facade docblocks are auto generated, I would recommend putting an `isLocked` method on the Cache Respository. It could throw an exception if the store doesn't implement `LockProvider`. This should mean that it will appear in the Facade docblock.

Riley19280: Would you also recommend adding the `isLocked` method to the `LockProvider`? Originally I avoided doing this for BC, but could re-target to 10.x if that is desired.

axlon:
Doesn't this just create race conditions? I always figured having to find out the hard way was by design

taylorotwell:
See above, would create race conditions in your code.

PR #45630, opened 2023-01-13
[9.x] Add DatabaseTruncates option for setting up test DB

patrickomeara:
This PR adds another option to handle database state in tests.

The `DatabaseTruncates` trait is a drop in replacement for `DatabaseMigrations` however it only truncates tables with data instead of dropping and migrating the whole database.

In my dusk test suite I have seen a 60-70% time reduction in database setup operations; migrate, seed, rollback/truncate.

7 dusk tests, that all change 12 of 91 tables the following time was spent on database setup:
`DatabaseMigrations` 12.06 seconds - 1.72s/t
`DatabaseTruncates` 4.61 seconds - 0.65s/t

### DIFF
--- a/src/Illuminate/Foundation/Testing/DatabaseTruncates.php
+++ b/src/Illuminate/Foundation/Testing/DatabaseTruncates.php
@@ -32,7 +32,17 @@ trait DatabaseTruncates
         }
 
         // Seed the database on subsequent runs.
-        $this->artisan('db:seed', ['--class' => $this->seeder()]);
+        if ($seeder = $this->seeder()) {
+            // Use a specific seeder class.
+            $this->artisan('db:seed', ['--class' => $seeder]);
+
+            return;
+        }
+
+        if ($this->shouldSeed()) {
+            // Use the default seeder class.
+            $this->artisan('db:seed');
+        }
     }
 
     protected function excludeTables()
--- a/src/Illuminate/Foundation/Testing/DatabaseTruncates.php
+++ b/src/Illuminate/Foundation/Testing/DatabaseTruncates.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace Illuminate\Foundation\Testing;
+
+use Illuminate\Foundation\Testing\Traits\CanConfigureMigrationCommands;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+trait DatabaseTruncates
+{
+    use CanConfigureMigrationCommands;
+
+    protected static array $allTables;
+
+    protected function truncateTables()
+    {
+        // Always remove any test data before the application is destroyed.
+        $this->beforeApplicationDestroyed(function () {
+            Schema::disableForeignKeyConstraints();
+            collect(static::$allTables ??= DB::connection()->getDoctrineSchemaManager()->listTableNames())
+                ->diff($this->excludeTables())
+                ->filter(fn ($table) => DB::table($table)->exists())
+                ->each(fn ($table) => DB::table($table)->truncate());
+        });
+
+        // Migrate and seed the database on first run.
+        if (! RefreshDatabaseState::$migrated) {
+            $this->artisan('migrate:fresh', $this->migrateFreshUsing());
+            RefreshDatabaseState::$migrated = true;
+
+            return;
+        }
+
+        // Seed the database on subsequent runs.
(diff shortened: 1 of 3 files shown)


src/Illuminate/Foundation/Testing/DatabaseTruncates.php
```diff
@@ -0,0 +1,44 @@
...
+        // Migrate and seed the database on first run.
+        if (! RefreshDatabaseState::$migrated) {
+            $this->artisan('migrate:fresh', $this->migrateFreshUsing());
+            RefreshDatabaseState::$migrated = true;
+
+            return;
+        }
+
+        // Seed the database on subsequent runs.
+        $this->artisan('db:seed', ['--class' => $this->seeder()]);
```
taylorotwell, line 35: What if someone doesn't want to seed?

patrickomeara: I've added a check around the `seeder()`, and `shouldSeeder()` methods.

src/Illuminate/Foundation/Testing/DatabaseTruncates.php
```diff
@@ -0,0 +1,54 @@
...
+        if ($this->shouldSeed()) {
+            // Use the default seeder class.
+            $this->artisan('db:seed');
+        }
+    }
+
+    protected function excludeTables()
+    {
+        return [
+            'migrations',
```
emargareten, line 51: You should use the config `database.migrations`

src/Illuminate/Foundation/Testing/DatabaseTruncates.php
```diff
@@ -0,0 +1,54 @@
...
+    protected function truncateTables()
+    {
+        // Always remove any test data before the application is destroyed.
+        $this->beforeApplicationDestroyed(function () {
+            Schema::disableForeignKeyConstraints();
+            collect(static::$allTables ??= DB::connection()->getDoctrineSchemaManager()->listTableNames())
+                ->diff($this->excludeTables())
+                ->filter(fn ($table) => DB::table($table)->exists())
+                ->each(fn ($table) => DB::table($table)->truncate());
+        });
```
emargareten, line 24: Don't you have to re-enable foreign keys after truncate?

emargareten:
I would extract the truncate logic into a separate class so it can be used elsewhere.

I am using cypress for e2e testing with the laravel-cypress package where you send requests to run console commands. so the only way to truncate is to have a console command to run it. (I actually have my own command for this but I would love to see it in the framework)

src/Illuminate/Foundation/Testing/DatabaseTruncates.php
patrickomeara: Only if the app uses them. I think this will need to use them by default but have the developer be able to configure it. I'll put something together.

patrickomeara: In all the testing I've done, the foreign keys constraints don't need to be manually re-enabled when the application is being destroyed. I'm happy to be proven wrong on this, as I have a configurable solution ready to go. However, for the sake of simplicity and speed I've left the `Schema::enableForeignKeyConstraints()` out.

To test: try to truncate a table with non-nullable foreign keys in a subsequent test.

```php
DB::table('users')->truncate();
```

src/Illuminate/Foundation/Testing/DatabaseTruncates.php
patrickomeara: I didn't know this existed. Thanks.

taylorotwell:
How come traits like `RefreshDatabase` and `DatabaseMigrations` have this line of code but this trait doesn't?

```php
$this->app[Kernel::class]->setArtisan(null);
```

taylorotwell:
I think there is quite a bit of customizability missing before this could be added. For example, which database connections to use? Right now it's hard-coded to use the DB facade's default connection, etc.

Note you can add this to your own project using the `setUp{Trait}` and `tearDown{Trait}` methods.

taylorotwell:
If you want to re-submit this I recommend looking at `RefreshDatabase`. How we allow the connections to be specified and then spin through those connections, etc.

patrickomeara:
Yep, fair enough. I'll look into this when I get a chance. Looks pretty straight forward. Cheers.

PR #28412, opened 2019-05-04
[5.9] Return uuids as string

SjorsO:
This PR makes `Str::uuid()` and `Str::orderedUuid()` return a string instead of an instance of `Ramsey\Uuid\Uuid`.

I ran into the issue that that when `Str::uuid()` is used in a factory, using that value in a test will fail the `uuid` validation rule:

image

The test above fails with the message `The user id must be a valid UUID.`, because the uuid attribute on the model is an object, not a string.

I couldn't think of a reason why you would want the object instead of the string when using `Str::uuid()`. All usages in the framework also instantly convert the object to a string. If you do actually need the uuid object, you can always directly call `\Ramsey\Uuid\Uuid::uuid4();`

This PR targets master because it is a breaking change. This PR will cause code that calls `Str::uuid()->toString()` or `Str::orderedUuid()->toString()` to fail.

PR #45634, opened 2023-01-13
[10.x] Fix `Str::isUuid()` returning false when given a `Str::uuid()`

SjorsO:
Currently, `Str::isUuid(Str::uuid())` returns false, because `isUuid()` only supports strings, and `Str::uuid()` returns a `UuidInterface`.

This fixes a bug described here: https://github.com/laravel/framework/pull/28412, because [the `uuid` validation rule](https://github.com/laravel/framework/blob/9d198bd947dbf1eb0a9499344754eb8426cb7a73/src/Illuminate/Validation/Concerns/ValidatesAttributes.php#L2205) uses `Str::isUuid()`.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -14,6 +14,7 @@ use Ramsey\Uuid\Codec\TimestampFirstCombCodec;
 use Ramsey\Uuid\Generator\CombGenerator;
 use Ramsey\Uuid\Uuid;
 use Ramsey\Uuid\UuidFactory;
+use Ramsey\Uuid\UuidInterface;
 use Symfony\Component\Uid\Ulid;
 use Traversable;
 use voku\helper\ASCII;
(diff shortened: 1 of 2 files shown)


GrahamCampbell:
I don't feel like this is a bug. The Str class should only work on strings, and nothing else.

SjorsO:
> The Str class should only work on strings, and nothing else.

I agree with this, but `Str::isUuid(Str::uuid())` returning `false` definitely isn't right

a-bashtannik:
I would say differently, `Str::uuid()` should return a string, simply because it's evident from the naming and usage form. Even considering `UuidInterface` has `__toString` method. For me changing `Str::uuid()` return looks better.

taylorotwell:
I dunno - this is the `Str` class. So, it feels like it should be focused on strings. Not sure I really want to introduce a breaking change to make it accept objects too.

Issue #45635, opened 2023-01-13
Calling a model factory within another factory breaks in a namespaced package 


cannycookie:
- Laravel Version: 9.47.0
- PHP Version: 8.1.11
- Database Driver & Version: mysql

### Steps To Reproduce:

In a package create a Factory and call another factory from within it.

```php
namespace Visualbuilder\Core\Database\Factories;

use Visualbuilder\Core\Models\Form;
use Visualbuilder\Core\Models\Layout;
use Illuminate\Database\Eloquent\Factories\Factory;

/**
 * @extends Factory<\Visualbuilder\Core\Models\Form>
 */
final class FormFactory extends Factory
{

	protected $model = Form::class;
	
	public function definition(): array
	{
		
		return [
			'name'      => $this->faker->name,
			'layout_id' => Layout::factory()
		];
	}
}

```

Run the seeder and this error is thrown:-
```bash
Class "App\Layout" not found
```

The namespace is lost due to a borked function modelName() in:-

```src/Illuminate/Database/Eloquent/Factories/Factory.php```

propose changing:- 
```php
 public function modelName()
    {
        $resolver = static::$modelNameResolver ?? function (self $factory) {
	        $namespacedFactoryBasename = Str::replaceLast(
		        'Factory', '', Str::replaceFirst(static::$namespace, '', get_class($factory))
	        );
	
	        $factoryBasename = Str::replaceLast('Factory', '', class_basename($factory));
	
	        $appNamespace = static::appNamespace();
	
	        return class_exists($appNamespace.'Models\\'.$namespacedFactoryBasename)
		        ? $appNamespace.'Models\\'.$namespacedFactoryBasename
		        : $appNamespace.$factoryBasename;
        };

        return $this->model ?? $resolver($this);
    }
```

to

```php
 public function modelName()
   {
       $resolver = static::$modelNameResolver ?? function (self $factory) {
	        $namespacedFactoryBasename = Str::replaceLast(
		        'Factory', '', Str::replaceFirst(static::$namespace, 'Models\\', get_class($factory))
	        );
	
	        $factoryBasename = Str::replaceLast('Factory', '', class_basename($factory));
			
	        $appNamespace = static::appNamespace();
	
	        return class_exists($namespacedFactoryBasename)
		        ? $namespacedFactoryBasename
		        : $appNamespace.$factoryBasename;
        };

        return $this->model ?? $resolver($this);
}
```

driesvints:
Heya, thanks for reporting.

[...]

PR #45636, opened 2023-01-13
[9.x] Fix Factory Model name resolution. #45635

cannycookie:
This bug is preventing deployment of a package, seeders are unable to be run.


driesvints:
This seems to break the test suite.

PR #45637, opened 2023-01-13
[11.x] Add missing() in Session contract

apih:
Add the `missing()` method in the `Session` interface contract, since my editor complained about undefined method when using it via a `Request` instance. For example: `$request->session()->missing('key')`.

### DIFF
--- a/src/Illuminate/Contracts/Session/Session.php
+++ b/src/Illuminate/Contracts/Session/Session.php
@@ -64,6 +64,14 @@ interface Session
     public function exists($key);
 
     /**
+     * Determine if the given key is missing from the session data.
+     *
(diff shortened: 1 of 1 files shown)


driesvints:
This is not needed on the contract per se. `exists` already is there for the reverse which you can easily implement in userland.

driesvints:
master is also 11.x, not 10.x

apih:
Ah, my mistake. I intended to target 10.x earlier.

driesvints:
Gonna close this one then. But I suggest you don't re-attempt because I don't see the need for this.

apih:
Yes, thanks for your reminder.

PR #45641, opened 2023-01-13
[9.x] Add option for not store meta data using session database driver

andiLeong:
when using the session driver database

in the DatabaseSessionHandler class constructor we can pass null as the container property.

if the container is null those **user_id , ip_address ,user_agent** will not persist in the db.

so we have all prepared for users to opt to not save those meta data.

as I know there is no easy way for users to decide not include those meta data through configuration.

because in the session manager we always pass the container to the DatabaseSessionHandler when getting the driver.

this pr is to allow user weather or not to include those meta thought a the session config

if pr approve I will create a new pr to include the config in other laravel repo


taylorotwell:

PR #45644, opened 2023-01-13
[9.x] Fix prohibited validator should accept empty value

aran112000:
Bug fix to match the behaviour shown in the Laravel documentation for the `prohibited` validation rule. Presently empty values weren't accepted which goes against the documented behaviour:

https://laravel.com/docs/master/validation#rule-prohibited

Functionality has been wrong since first introduced in Laravel 8.x

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -1689,7 +1689,7 @@ trait ValidatesAttributes
      */
     public function validateProhibited($attribute, $value)
     {
-        return empty($value);
+        return $value === '';
     }
 
     /**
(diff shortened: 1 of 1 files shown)


Quirinus:
Do not use `empty()`, as it returns `true` on `empty('0')`. Where `'0'` is obviously not an empty string.

aran112000:
> Do not use `empty()`, as it returns `true` on `empty('0')`. Where `'0'` is obviously not an empty string. It also casts the given value, so `empty(0)` would also give `true`, even though it's not a string.
> 
> Instead, explicitly check for an empty string with a strict comparison.

Good catch, now resolved

taylorotwell:
Better to just update the docs than make this breaking change. 👍🏻

PR #45647, opened 2023-01-13
Add interface that matches with the InteractsWithQueue trait

lunfel:
## Description

This pull request introduces 3 new interfaces - QueueableInterface, DispatchableInterface and InteractsWithQueueInterface. These interfaces provide a clear, consistent way to define the contracts for classes that interact with the queue system in Laravel.

## Benefit to the users

One of the main benefits of using interfaces in PHP over traits is that they can be type hinted. This means that when a method or function expects an object of a certain type, you can specify that it must implement a certain interface. This allows for better type safety and makes it easier to catch errors early in the development process. On the other hand, in PHP you cannot type hint a trait, so if you want to make sure a class implements certain methods you need to rely on documentation or conventions. By using interfaces, we can ensure that classes that interact with the queue system in Laravel have a clear and consistent set of methods and behavior and can be type hinted to make sure they are used correctly.

## The reasons it does not break any existing features

To avoid any breaking changes, I decided to implement the appropriate interfaces on all classes that already utilize the corresponding traits. This way, these classes will keep their current functionality while also adhering to a clear, consistent contract that aligns with that functionality. In summary, this change will not affect the behavior of existing code and only adds a new way of defining contracts using interfaces.

## Paving the way to type safety

I believe this change, the introduction of 3 new interfaces, is a step towards achieving more strongly typed functions in Laravel. In the next major release, developers will have the option to use these interfaces, which may involve breaking changes, to provide explicit types for function parameters, instead of relying on typeless parameters.

### DIFF
--- a/src/Illuminate/Broadcasting/BroadcastEvent.php
+++ b/src/Illuminate/Broadcasting/BroadcastEvent.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Broadcasting;
 
 use Illuminate\Bus\Queueable;
+use Illuminate\Bus\QueueableInterface;
 use Illuminate\Contracts\Broadcasting\Factory as BroadcastingFactory;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Contracts\Support\Arrayable;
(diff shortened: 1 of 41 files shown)


lunfel:
If this PR is accepted. I will need advise on how to apply those changes to the 10.x branch and master. Do I cherry-pick my commits into the 10.x branch or a 9.x to 10.x merge occurs once in a while? This part is unclear for me.

driesvints:
@lunfel 9.x is merged into 10.x which is merged into master. No extra action needed.

PR #45651, opened 2023-01-14
Add `assertBadRequest()` and `assertNotImplemented()` methods

senaranya:
Added assert-methods for HTTP response codes 400 and 501


driesvints:
Please see previous attempts.

PR #45652, opened 2023-01-14
[9.x] Add Method to get request data as an array

khaledw62:
This adds methods to conveniently retrieve and cast request data as an array. and even add the ability to cast a comma separated into an array

Before

```php
$input = [];

if (request()->filled('array')) {
    if (is_array(request()->get('array'))) {
        $input = request()->get('array');
    }else{
        $input = [request()->get('array')];
    }
}
```

After

```php
$input = request()->array('array');
```

if the value is not an array this method will cast the value as an array

```php
//https://example.com/?array=23

$input = request()->array('array');
//will results in [23]
``` 

if the value is comma separated this method will cast it into an array

```php
//https://example.com/?array=23,24,55,60

$input = request()->array('array');
//will results in [23,24,55,60]
```

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -374,7 +374,7 @@ trait InteractsWithInput
      * @param  array  $default
      * @return array
      */
-    public function array($key, $default = [])
+    public function asArray($key, $default = [])
     {
         if ($this->isNotFilled($key)) {
             return $default;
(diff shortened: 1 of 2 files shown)


ankurk91:
Are we going to add method for each of php scaler type?

khaledw62:
No, just array

driesvints:
`array` is a reserved keyword: https://www.php.net/manual/en/reserved.keywords.php

taylorotwell:

zorenkonte:
@khaledw62 you can use `Arr::wrap()`

[https://laravel.com/docs/9.x/helpers#method-array-wrap](url)

PR #45654, opened 2023-01-14
Add enum class to migration enum accepted values

m-mohammadi1:
When we want to add an enum column ,and have a corresponding enum to that column, it is very annoying to write all enum case's values one by one or with other tricks. with this new feature it is very easy and fast to do we just give enum class we have and then it adds the cases.
instead of
`
            $table->enum('test_column', [TestEnum::TEST1->value, TestEnum::TEST2->value, TestEnum::TEST3->value]);
`
we now have
`
            $table->enum('test_column', new \Illuminate\Database\Schema\Enum(TestEnum::class));
`

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -10,7 +10,6 @@ use Illuminate\Database\Schema\Grammars\Grammar;
 use Illuminate\Database\SQLiteConnection;
 use Illuminate\Support\Fluent;
 use Illuminate\Support\Traits\Macroable;
-use function PHPUnit\Framework\isInstanceOf;
 
 class Blueprint
 {
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -10,6 +10,7 @@ use Illuminate\Database\Schema\Grammars\Grammar;
 use Illuminate\Database\SQLiteConnection;
 use Illuminate\Support\Fluent;
 use Illuminate\Support\Traits\Macroable;
+use function PHPUnit\Framework\isInstanceOf;
 
 class Blueprint
 {
@@ -1033,11 +1034,15 @@ class Blueprint
      * Create a new enum column on the table.
      *
      * @param  string  $column
-     * @param  array  $allowed
+     * @param  array|Enum  $allowed
      * @return \Illuminate\Database\Schema\ColumnDefinition
      */
-    public function enum($column, array $allowed)
+    public function enum($column, array|Enum $allowed)
     {
+        if ($allowed instanceof Enum) {
+            $allowed = $allowed->getAcceptedValues();
+        }
+
         return $this->addColumn('enum', $column, compact('allowed'));
     }
 
--- a/src/Illuminate/Database/Schema/Enum.php
+++ b/src/Illuminate/Database/Schema/Enum.php
@@ -12,8 +12,7 @@ class Enum
 
     public function getAcceptedValues(): array
     {
-        $enumClass = new $this->type;
-        $cases = $enumClass->cases();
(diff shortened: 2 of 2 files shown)


henzeb:
Would it rather be helpful if the method accepted an array of enums and then transforms those to either values or keys? Or just the enum class name? This Enum class is just extra code to write for users and makes things unreadable in my opinion.

antonkomarev:
What if in future you will decide to deprecate and delete enum? Then old migration will become invalid

m-mohammadi1:
> Would it rather be helpful if the method accepted an array of enums and then transforms those to either values or keys? Or just the enum class name? This Enum class is just extra code to write for users and makes things unreadable in my opinion.

Actually we have this syntax in rules and I prefer it because of type hinting and making it readable in migration file.

dennisprudlo:
Yeah I agree with @antonkomarev. This change has also been rejected every time it comes up: https://github.com/laravel/framework/pull/43483.

The before example couples the migration to the enum as well. Such references shouldn't be used in migrations.

m-mohammadi1:
> What if in future you will decide to deprecate and delete enum? Then old migration will become invalid
> Yeah I agree with @antonkomarev. This change has also been rejected every time it comes up: #43483.
> 
> The before example couples the migration to the enum as well. Such references shouldn't be used in migrations.

I think this is a matter the programmer should consider it, even we use first syntax
`
$table->enum('test_column', [TestEnum::TEST1->value, TestEnum::TEST2->value, TestEnum::TEST3->value]);
`
after deleting the enum it would have errors.
and coupling issue is because we are using enums in migration file and why don't we use the enums?! it is a good feature in php and most of times it represent our column accepted values It could be better to use Enums in migration files.
The programmer decides to use it or not.

m-mohammadi1:
Due to last pull requests I this pull request should be closed

henzeb:
@dennisprudlo, a migration should never run again, once it ran.

But even so: wouldn't that be solved by https://laravel.com/docs/9.x/migrations#squashing-migrations?

also, if you remove an enum and you ever should run the migrations again, it's probably on a new database, without data. You'd only be worried when you would import data from a source that hasn't removed/migrated the value in the database, as a migration removing the the enum case from the database schema would also use this tactic and would do nothing.

dennisprudlo:
Migrations are about the database schema not the data that's in them. So it doesn't matter if new database is empty or not. Both databases will have a different schema and to ensure that's not happening we use migrations.

Imagine a DB1 and DB2, both are in sync. Now you add a migration adding an enum column with some enum values from php. You run this migration on DB1. After that you change one of those enum cases (different name, whatever) in your code and then run the migration on DB2. You have now two out of sync databases.

That's why we use static strings as table names and not the eloquent model name. Same thing. Changes in your application code should never affect the database schema.

henzeb:
> Imagine a DB1 and DB2, both are in sync. Now you add a migration adding an enum column with some enum values from php. You run this migration on DB1. After that you change one of those enum cases (different name, whatever) in your code and then run the migration on DB2. You have now two out of sync databases.

wouldn't you also create a migration that accompanies the change in php?

dennisprudlo:
Exactly. That's why you can always explicitly state the enum values and don't have to rely on the the php enum cases.

By the way it's still possible. No one prevents anyone from doing this: `$table->enum('foo', [SomeEnum::someCase->value]);`

henzeb:
I can also write 'Exactly'.

If I use PHP enum references, I am certain that what's in my database, matches my php enum object. adding a migration for a change may feel useless, but if you, everytime you need to change an database related enum, make sure everything before your change is dumped to a schema file, you won't have  worry about it. 

I am also wondering, does Open/Closed not apply to enum cases?

dennisprudlo:
As I said no one stops anyone from doing so. I'm just saying Laravel tries to decouple migrations and application logic.

PR #45658, opened 2023-01-15
[9.x] Add `dispatchWith` method to Dispatcher class

imanghafoori1:
The Dispatcher class allows referencing any method as a listener for a given event, but there is an important assumption on the listeners.
And that is they should not throw exceptions or otherwise the execution of the other listeners will stop, and there is nothing that the dispatcher of the event can do about it. In other words, it can not set an exception handler for each and every listener.

For example in HomeController:
```php
try {
    \Event::dispatch('user-registered);
} catch (\Exception $e) {
    // handle the exception
}
```
In the example above, if the first listener throws an exception, all the other listeners would not be executed and it goes to the catch block directly.

-----------------------

```php
 $wrapper = function ($listener) {
    try {
        return $listener();
    } catch (\Exception $e) {
        // report the exception...
    }
}

\Event::dispatchWith('user-registered', wrapper: $wrapper);
```
But this one allows the caller to set a try/catch for each and every listener.

- This is not needed if all the listeners are set by the person who fires the event, but for a package developer who is firing an event in the package code, it is always a matter of question if the listeners set by the users are healthy enough or not. In other words, you fire the events but you never know what is going to be executed as the listener.

- In this implementation, the `wrapper` can not know about the details of the listener since it is already wrapped in a closure and can only call it and get the results.

- You may also decide to add it to the `Dispatcher` interface as well.

### DIFF
--- a/src/Illuminate/Events/Dispatcher.php
+++ b/src/Illuminate/Events/Dispatcher.php
@@ -223,14 +223,15 @@ class Dispatcher implements DispatcherContract
     }
 
     /**
-     * Fire an event and call the listeners.
+     * Fire an event and call the listeners within the given wrapper.
      *
      * @param  string|object  $event
      * @param  mixed  $payload
      * @param  bool  $halt
+     * @param  \Closure|null  $wrapper
      * @return array|null
      */
-    public function dispatch($event, $payload = [], $halt = false)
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #45662, opened 2023-01-16
Cookie::queue facade method has invalid docblock

azvyae:
Cookie::queue facade method has invalid expected input in the dockblock. Currently it expects array of parameters in the framework, but in the Laravel documentation it has to be three parameters. String $key, string $value, and int $minutes.


taylorotwell:
These are automatically generated so will need to be corrected in the source location.

cc @timacdonald

PR #45666, opened 2023-01-16
[9.x] Replace raw invisible characters in regex expressions with counterpart Unicode regex notations

restuff:
#41971 introduced regex selector update for invisible Unicode characters with additions in #44906.

This PR updates regex expressions by replacing raw invisible characters with counterpart Unicode regex notations to increase visibility of source code in text editors that don't support display of invisible characters labels or in VCS editors.

It also expands matching characters from ﻿U+FEFF|U+200B (in case of \Illuminate\Foundation\Http\Middleware::transform()) or from U+FEFF (in case of \Illuminate\Support\Str::squish()) to all characters of same group (\p{Cf} - invisible formatting indicator).

### DIFF
--- a/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
+++ b/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
@@ -53,7 +53,7 @@ class TrimStrings extends TransformsRequest
             return $value;
         }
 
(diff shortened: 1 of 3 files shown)


taylorotwell:
Needs better explanation, code examples, screenshots, etc. to explain what you are referring to.

PR #45667, opened 2023-01-16
[9.x] "Unfreeze" uuids on test teardown

gdebrauwer:
Currently when you freeze the uuid generation in one test, then that uuid generation is still frozen in another test.
```php
public function first_test()
{
    dump(Str::freezeUuids()->toString());
    dump(Str::uuid()->toString());
}

public function second_test()
{
    // All dumps return the same uuid,
    // while in this second test we are not using Str::freezeUuids()
    dump(Str::uuid()->toString());
}
```
This seems weird behavior, as other test helpers (storage fake, event fake, queue fake, ...) are all contained within a single test. This PR fixes that by "unfreezing" the uuid generation in testcase teardown.

### DIFF
--- a/src/Illuminate/Foundation/Testing/TestCase.php
+++ b/src/Illuminate/Foundation/Testing/TestCase.php
@@ -223,6 +223,7 @@ abstract class TestCase extends BaseTestCase
         Component::forgetFactory();
         Queue::createPayloadUsing(null);
         HandleExceptions::forgetApp();
+        Str::createUuidsNormally();
 
         if ($this->callbackException) {
             throw $this->callbackException;


timacdonald:
I'd recommend running the whole test in a closure passed to the `freezeUuids` function:


```php
public function test_it_works()
{
  Str::freezeUuids(function () {
    $this->assertions(...);
  });
}

public function test_it_works_again()
{
  Str::freezeUuids(function () {
    $this->assertions(...);
  });
}
```

gdebrauwer:
@timacdonald Is it not a bit weird that you have to nest all code of a single test that uses `freezeUuids` to ensure other tests that do not use `freezeUuids` are not broken by that single test because the other tests assume every generated uuid will be unique? 🤷‍♂️

Could I create a PR to change this in 10.x instead?

dennisprudlo:
Can't you just unfreeze the UUIDs at the end of the test?

gdebrauwer:
Yes that is possible, but is it not a bit weird that this is the only Laravel test helper where you have to do that? 🤷‍♂️

PR #45671, opened 2023-01-16
[9.x] replace string|null with ?string

PerryvanderMeer:
Not sure if we want to go this way, but I'll give it a try.
Currently, `string|null` and `?string` are used through eachother. For example:

https://github.com/laravel/framework/blob/37bbb8c299b51195fc38f845dae847726d982522/src/Illuminate/Auth/Access/HandlesAuthorization.php#L22

https://github.com/laravel/framework/blob/37bbb8c299b51195fc38f845dae847726d982522/src/Illuminate/Auth/Access/HandlesAuthorization.php#L35

I'd prefer `?string` for variables with a single type, since it improves readability.
This PR replaces `string|null` with `?string` and corrects those internal inconsistencies.
If this PR gets merged, I'll work on the other types.


dammy001:
@PerryvanderMeer These changes are not needed since doctypes have been removed in L10

driesvints:
@dammy001 not exactly. They're only removed on stubs which are published as userland code. Any internal code keeps their docblocks.

driesvints:
@PerryvanderMeer I've sent in a PR to do the opposite: https://github.com/laravel/framework/pull/45677

In general it's good to assume that the least change needed is the way to go. Thanks

PerryvanderMeer:
@driesvints In the end that seems to be the best solution. Thanks!

PR #45673, opened 2023-01-17
[9.x] Fix docblock

timacdonald:
This was changed to `mixed` in https://github.com/laravel/framework/pull/45662, however variadic function should include the `...`.

The `$parameters` _argument_ will always be an array, however the use of `...` indicates that it accepts multiple _parameters_ that are of type `mixed`.

See: https://phpstan.org/writing-php-code/phpdocs-basics#variadic-functions


driesvints:
Thanks!

PR #45680, opened 2023-01-17
[9.x]  Replace raw invisible characters in regex expressions with counterpart Unicode regex notations

restuff:
It is always a good practice to avoid or at least minimize the usage of unescaped non-ascii characters in source code, especially if they are invisible characters.

#41949 introduced an addition to TrimStrings middleware's regex to filter out "Zero Width No-Break Space" (\x{FEFF}) characters, however it was implemented as a raw unescaped invisible character in the source code.

image

#44906 added one more invisible character "Zero Width Space" (\x{200B}) to the same regex.

image

However, these symbols aren't visible in GitHub diff

image

as well as in some text editors like Notepad++ (even with Show All Characters option turned on)

image

This PR adds more readability to the source code by replacing those invisible characters in both places where they are used with their counterpart Unicode regex notations without affecting the functionality.

 \Illuminate\Foundation\Http\Middleware::transform():
image

\Illuminate\Support\Str::squish():
image

There's no need to modify corresponding tests that use "real-life" unescaped raw invisible characters as functionality hasn't changed, however there were added some spelling corrections.

### DIFF
--- a/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
+++ b/src/Illuminate/Foundation/Http/Middleware/TrimStrings.php
@@ -53,7 +53,7 @@ class TrimStrings extends TransformsRequest
             return $value;
         }
 
-        return preg_replace('~^[\s﻿​]+|[\s﻿​]+$~u', '', $value) ?? trim($value);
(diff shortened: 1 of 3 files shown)


taylorotwell:
Thanks

PR #45683, opened 2023-01-17
[9.x] Added a bootstrapping and bootstrapped event for pre and post boostrapping

ollieread:
Laravel currently supports listening to pre- and post-events for individual bootstrappers, but if you want to do something after all the bootstrappers have been run, or before any have been, you have to hardcode references to the default bootstrappers and hope nothing has changed.

This PR adds the `bootstrapping` and `bootstrapped` events, following the theme of `bootstrapping: BootstrapperName` and `bootstrapped: BootstrapperName`. This should allow developers to react to bootstrapping without having to worry whether `Kernel::$bootstrappers` has been modified in some way. In fact, the very existence of these two new events may remove any need to modify the bootstrappers array.

I have also added tests for both of these.

## `bootstrapping`
This event is fired before `Application` loops through the bootstrappers and can be hooked into using:
```php
Application::preBootstrapping(Closure $callback)
```

## `bootstrapped`
This event is fired after `Application` loops through the bootstrappers and can be hooked into using:
```php
Application::postBootstrapping(Closure $callback)
```

### DIFF
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -236,6 +236,8 @@ class Application extends Container implements ApplicationContract, CachesConfig
     {
         $this->hasBeenBootstrapped = true;
 
+        $this['events']->dispatch('bootstrapping', [$this]);
+
         foreach ($bootstrappers as $bootstrapper) {
             $this['events']->dispatch('bootstrapping: '.$bootstrapper, [$this]);
 
@@ -243,6 +245,8 @@ class Application extends Container implements ApplicationContract, CachesConfig
 
             $this['events']->dispatch('bootstrapped: '.$bootstrapper, [$this]);
         }
+
+        $this['events']->dispatch('bootstrapped', [$this]);
     }
 
     /**
@@ -271,6 +275,17 @@ class Application extends Container implements ApplicationContract, CachesConfig
     }
 
     /**
+     * Register a callback to run before any bootstrapping.
+     *
+     * @param  \Closure  $callback
+     * @return void
+     */
+    public function preBootstrapping(Closure $callback)
+    {
+        $this['events']->listen('bootstrapping', $callback);
+    }
+
+    /**
      * Register a callback to run after a bootstrapper.
      *
      * @param  string  $bootstrapper
@@ -280,6 +295,17 @@ class Application extends Container implements ApplicationContract, CachesConfig
     public function afterBootstrapping($bootstrapper, Closure $callback)
     {
         $this['events']->listen('bootstrapped: '.$bootstrapper, $callback);
+    }
+
+    /**
+     * Register a callback to run after all bootstrapping.
+     *
+     * @param  \Closure  $callback
+     * @return void
+     */
+    public function postBootstrapping(Closure $callback)
+    {
+        $this['events']->listen('bootstrapped', $callback);
     }
 
     /**
--- a/tests/Foundation/FoundationApplicationTest.php
+++ b/tests/Foundation/FoundationApplicationTest.php
@@ -315,6 +315,26 @@ class FoundationApplicationTest extends TestCase
         $this->assertArrayHasKey(0, $app['events']->getListeners('bootstrapped: Illuminate\Foundation\Bootstrap\RegisterFacades'));
     }
 
+    public function testPreBootstrappingAddsClosure()
+    {
+        $app = new Application;
+        $closure = function () {
+            //
+        };
+        $app->preBootstrapping($closure);
+        $this->assertArrayHasKey(0, $app['events']->getListeners('bootstrapping'));
+    }
+
+    public function testPostBootstrappingAddsClosure()
+    {
+        $app = new Application;
+        $closure = function () {
+            //
+        };
+        $app->postBootstrapping($closure);
+        $this->assertArrayHasKey(0, $app['events']->getListeners('bootstrapped'));
+    }
+
     public function testTerminationCallbacksCanAcceptAtNotation()
     {
         $app = new Application;


WendellAdriel:
Pretty cool idea. 🥳 
IDK if anyone agrees with me but I would prefer using this wording instead:

```php
Application::beforeBootstrap(Closure $callback)
```

```php
Application::afterBootstrap(Closure $callback)
```

ollieread:
> Pretty cool idea. 🥳 
> IDK if anyone agrees with me but I would prefer using this wording instead:
> 
> ```php
> Application::beforeBootstrap(Closure $callback)
> ```
> 
> ```php
> Application::afterBootstrap(Closure $callback)
> ```

The names could be changed to that, though that feels a little too close to the two methods that already exist.

timacdonald:
I like this idea, but wondering where would you call `Application::beforeBootstrap(Closure $callback)`?

I assume you would need to put this in `bootstrap/app.php`, as I don't believe it can go in a service provider, as service providers are part of the bootstrapping process.

```php
$app = new Illuminate\Foundation\Application(
    $_ENV['APP_BASE_PATH'] ?? dirname(__DIR__)
);

$app->singleton(
    Illuminate\Contracts\Http\Kernel::class,
    App\Http\Kernel::class
);

$app->singleton(
    Illuminate\Contracts\Console\Kernel::class,
    App\Console\Kernel::class
);

$app->singleton(
    Illuminate\Contracts\Debug\ExceptionHandler::class,
    App\Exceptions\Handler::class
);

$app->beforeBootstrap(function () {
    //
});

return $app;
```

I don't think you would be able to use the `Application` facade

ollieread:
> I like this idea, but wondering where would you call `Application::beforeBootstrap(Closure $callback)`?
> 
> I assume you would need to put this in `bootstrap/app.php`, as I don't believe it can go in a service provider; the service providers are part of the bootstrapping process.
> 
> ```php
> $app = new Illuminate\Foundation\Application(
>     $_ENV['APP_BASE_PATH'] ?? dirname(__DIR__)
> );
> 
> $app->singleton(
>     Illuminate\Contracts\Http\Kernel::class,
>     App\Http\Kernel::class
> );
> 
> $app->singleton(
>     Illuminate\Contracts\Console\Kernel::class,
>     App\Console\Kernel::class
> );
> 
> $app->singleton(
>     Illuminate\Contracts\Debug\ExceptionHandler::class,
>     App\Exceptions\Handler::class
> );
> 
> $app->beforeBootstrap(function () {
>     //
> });
> 
> return $app;
> ```
> 
> The problem here is that the applications `'events'` instance has not yet been bound to the container, i.e. the `EventServiceProvider` has not yet run. I think that the instance it resolved in the above would then be replaced.

You raise a very good point. I did have a solution for this, but I dont have it to hand

timacdonald:
I think it would make sense to have the "after" hook. But if you need a before hook, you may as well just put that code in the `boostrap/app.php` file outside of a closure.

```php
// bootstrap/app.php

$app = new Illuminate\Foundation\Application(
    $_ENV['APP_BASE_PATH'] ?? dirname(__DIR__)
);

// ...

$app->instance('laravel.started_at', Carbon\Carbon::now());

return $app;
```

```php
// AppServiceProvider

public function register()
{
    $this->app->afterBootstrap(function ($app) {
        if ($app['laravel.started_at']->addSecond()->isAfter(now()) {
            report('Bootstrapping the app took longer than one second.');
        }
    });
}

```

ollieread:
> I think it would make sense to have the "after" hook. But if you need a before hook, you may as well just put that code in the `boostrap/app.php` file outside of a closure.
> 
> ```php
> // bootstrap/app.php
> 
> $app = new Illuminate\Foundation\Application(
>     $_ENV['APP_BASE_PATH'] ?? dirname(__DIR__)
> );
> 
> // ...
> 
> $app->instance('laravel.started_at', Carbon\Carbon::now());
> 
> return $app;
> ```
> 
> ```php
> // AppServiceProvider
> 
> public function register()
> {
>     $this->app->afterBootstrap(function ($app) {
>         if ($app['laravel.started_at']->addSecond()->isAfter(now()) {
>             report('Bootstrapping the app took longer than one second.');
>         }
>     });
> }
> 
> ```

Oh wait, sorry. The events handler is registered. It's registered as a core provider before bootstrapping even happens. It's how you can listen for particular bootstrappers.

crynobone:
What's would be the difference between the suggested `afterBootstrap` vs existing `booted` method?

ollieread:
> What's would be the difference between the suggested `afterBootstrap` vs existing `booted` method?

That happens later in the startup process, after providers have been registered and booted.

ollieread:
> I think it would make sense to have the "after" hook. But if you need a before hook, you may as well just put that code in the `boostrap/app.php` file outside of a closure.
> 
> ```
> // bootstrap/app.php
> 
> $app = new Illuminate\Foundation\Application(
>     $_ENV['APP_BASE_PATH'] ?? dirname(__DIR__)
> );
> 
> // ...
> 
> $app->instance('laravel.started_at', Carbon\Carbon::now());
> 
> return $app;
> ```
> 
> ```
> // AppServiceProvider
> 
> public function register()
> {
>     $this->app->afterBootstrap(function ($app) {
>         if ($app['laravel.started_at']->addSecond()->isAfter(now()) {
>             report('Bootstrapping the app took longer than one second.');
>         }
>     });
> }
> ```

My particular use case is a package, and I want to add the following to `bootstrap/app.php`, either automatically, or by having the user do it.

```php
$app->beforeBootstrapping(function () use($app) {
    $app->registerProvider(new MyProvider(require /file.php));
});
```

The amount of stuff this provider does would make the app file very large. All of the things it does also needs to happen before any ENV or Config is loaded.

crynobone:
```php
$app->beforeBootstrapping(function () use($app) {
    $app->registerProvider(new MyProvider(require /file.php));
});
```

If your intention to add this under `ServiceProvider::register()` method doesn't provide any different because the application is already bootstrapping, you might as well just write:

```php
$this->app->registerProvider(new MyProvider(require /file.php));
```

ollieread:
> ```
> $app->beforeBootstrapping(function () use($app) {
>     $app->registerProvider(new MyProvider(require /file.php));
> });
> ```
> 
> If your intention to add this under `ServiceProvider::register()` method it doesn't provide any difference because the application is already bootstrapping, you might as well just write:
> 
> ```
> $this->app->registerProvider(new MyProvider(require /file.php));
> ```
> 
> Even registering `afterBootstrap()` in a Service Provider is too late, because previous "bootstrap" classes have been executed.

 For my particular use case, I need to perform a bunch of actions before the env and config is loaded, as the result of these actions can and will affect what is loaded. You are absolutely correct that I could simply load the provider right there, but, adding a provider then will still happen before "bootstrapping".
 
 Here's the lifecycle process:
 
 - Create new instance of `Application` with base bindings in `bootstrap/app.php`
 - The `Application` constructor does the following:
    - Create base bindings for things like `Container` and `Mix`
    - Register and boot the following service providers in the following order:
        - `EventServiceProvider`
        - `LogServiceProvider`
        - `RoutingServiceProvider`
    - Register container aliases
 - **This is where any extra code in `bootstrap/app.php` would be run**
 - It then creates a new `Kernel`
 - It then creates a `Request` and passes it to `Kernel::handle()`
 - The `Kernel::handle()` method does the following:
    - Bind `Request` to the container
    - Clear the resolved instance for the request facade
    - Calls `Kernel::bootstrap()`, which:
        - **This is where the "pre-bootstrapping" event would be
        - Cycles through the bootstrappers
            - `LoadEnvironmentVariables`
            - `LoadConfiguration`
            - `HandleExceptions`
            - `RegisterFacades`
            - `RegisterProviders`
            - `BootProviders`
        - Before each bootstrapper it fires off an event `bootstrapping: {class}`
        - After each bootstrapper it fires off an event `bootstrapped: {class}`
        - This is where the "post-bootstrapping" event would be

There was a mistake in my previous code example, I would be using `Application::register()` not `Application::registerProvider()`, as the second doesn't exist.

You are correct that you can just run the code directly in `app.php`, but if in the process you need an instance of `Request`, it will not be available.

While I completely appreciate that the event is only slightly later during the framework booting process, I also think there's no harm in adding it as it complements the two `bootstrapping` and `bootstrapped` events that already exist. Sure I could use the `Application::beforeBootstrapping()` and `Application::afterBootstrapping()` methods, but that would only work as long as there aren't modifications to `Kernel::$bootstrappers`.

taylorotwell:
Can you explain a bit more what kind of package you are trying to write? 😅

ollieread:
> Can you explain a bit more what kind of package you are trying to write? 😅

It's a fully fledged all singing all dancing multitenancy package, as in, it does just about every combination of different features attributed to multitenancy.

This particular change would be for the static/modular style multitenancy that is common in applications built to serve a franchise, where each tenant will have custom templates/views, as well as possible their own config and/or environment variables. When dealing with the dynamic/SaaS style multitenancy, you create new database connections, storage drives, etc, etc, but with the static approach, the default driver should match their default settings.

The idea is that `config/` and `.env` contain the core default settings, but something like `tenant/ollieread.com` (which would like sort of like a `laravel/laravel` installation) would have its own `config/` and `.env` that would override the default. It would be the same with the default view path, etc.

I could absolutely achieve what I want to do without this, but I figured that this is a small enough addition, and it wouldn't hurt to have it there.

taylorotwell:
Since you can continue working on the package without this particular feature at the moment, could you come back to us when the package is totally finished and ready for release and then we can look at adding it? Or, is it totally finished now?

ollieread:
> Since you can continue working on the package without this particular feature at the moment, could you come back to us when the package is totally finished and ready for release and then we can look at adding it? Or, is it totally finished now?

It's not finished now, and it's going to be a closed source package. I'm more than happy to close this PR if you aren't convinced. I can make do.

taylorotwell:
OK - why don't you re-open this when the package is ready for release just so we can be certain it will actually be needed. 👍

ollieread:
> OK - why don't you re-open this when the package is ready for release just so we can be certain it will actually be needed. 👍

Will do, thanks.

PR #45684, opened 2023-01-17
[9.x] Allow to set HTTP client for mailers

driesvints:
This PR allows people to configure custom options for the HTTP client when using the Mailgun or Postmark mailers from Symfony.

All options to be configured can be found on Symfony's docs: https://symfony.com/doc/current/reference/configuration/framework.html#http-client

For example, setting a timeout:

```php
'mailgun' => [
    'transport' => 'mailgun',
    'client' => [
        'timeout' => 7.5 // Seconds
    ],
],
```

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -100,7 +100,8 @@
         "phpstan/phpstan": "^1.4.7",
         "phpunit/phpunit": "^9.5.8",
         "predis/predis": "^1.1.9|^2.0.2",
-        "symfony/cache": "^6.0"
+        "symfony/cache": "^6.0",
+        "symfony/http-client": "^6.0"
     },
     "provide": {
         "psr/container-implementation": "1.1|2.0",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -100,7 +100,8 @@
         "phpstan/phpstan": "^1.4.7",
         "phpunit/phpunit": "^9.5.8",
         "predis/predis": "^1.1.9|^2.0.2",
-        "symfony/cache": "^6.0"
+        "symfony/cache": "^6.0",
+        "symfony/http-client": "^6.0"
```
GrahamCampbell, line 104: This should have also been added to the mailer composer.json, at least as a suggest.

driesvints: @GrahamCampbell that's already the case: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Mail/composer.json#L41

PR #45685, opened 2023-01-17
[9.x] fix: dependency in `cache:forget` command

owenvoke:
Move dependencies to to handle method.

By moving the dependencies to the handle method, we are able to run help and list commands without having to instantiate all of the dependencies of the command before we need them.


taylorotwell:
You didn't update Artisan Service Provider.

PR #45693, opened 2023-01-18
[9.x] Ensure decimal rule handles large values

timacdonald:
The validator does not currently handle arbitrary length floats when comparing the **size** of decimal numbers.

The following code currently passes when it should not...

```php
Validator::validate([
    'foo' => '1.88888888888888888888'
], [
    'foo' => [
        'decimal:20',
        'min:1.88888888888888888889',
    ],
]);

// [
//    "foo" => "1.88888888888888888888",
//  ]

```

This PR leans on `brick/math` further whenever we are comparing a value's size.

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -3,8 +3,8 @@
 namespace Illuminate\Validation\Concerns;
 
 use Brick\Math\BigDecimal;
-use Brick\Math\Exception\MathException as BrickMathException;
 use Brick\Math\BigNumber;
+use Brick\Math\Exception\MathException as BrickMathException;
 use DateTime;
 use DateTimeInterface;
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/composer.json
```diff
@@ -17,6 +17,7 @@
     "require": {
         "php": "^8.0.2",
         "ext-json": "*",
+        "brick/math": "^0.10.2",
```
timacdonald, line 20: This was missed when fixing the decimal model cast.

src/Illuminate/Validation/Concerns/ValidatesAttributes.php
```diff
@@ -2321,7 +2323,7 @@ public function validateUuid($attribute, $value)
      *
      * @param  string  $attribute
      * @param  mixed  $value
-     * @return mixed
+     * @return int|float|string
```
timacdonald, line 2326: Tightening this return type, as I believe it can only return these types.

PR #45697, opened 2023-01-18
[9.x] Allow Artisan commands to be aliased with a class property 

ziadoz:
This PR adds a new `aliases` convenience property to Artisan commands for defining name aliases, as opposed to implementing Symfony's `getAliases()` method or adding their `AsCommand` annotation.

In my use case I need to change the prefix on lots of commands, however doing so will break staging and production applications, but I can avoid this by aliasing the old command names to the new ones temporarily.

### DIFF
--- a/src/Illuminate/Console/Command.php
+++ b/src/Illuminate/Console/Command.php
@@ -61,6 +61,13 @@ class Command extends SymfonyCommand
     protected $hidden = false;
 
     /**
+     * The console command name aliases.
+     *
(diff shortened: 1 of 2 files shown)


taylorotwell:
This would technically be a breaking change if someone was already using an `$aliases` property on their command for some other reason.

ziadoz:
@taylorotwell Ah yeah. Would it be worth me re-opening this against 10.x instead?

PR #45701, opened 2023-01-18
[9.x] Add assert methods for status code 400 and 408

WendellAdriel:
This is adding two new assert methods for the response:

```php
$this->assertBadRequest(); // Checks if status code is 400

$this->assertRequestTimeout(); // Checks if status code is 408
```

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -171,6 +171,26 @@ class TestResponse implements ArrayAccess
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
Honestly I find it clearer to just assert on the actual status code. 😬

WendellAdriel:
Got it! 😉 

I just thought about adding that because in one of the projects I'm working on I was using some calls to assert these codes, but I understand 💪🏼 

Thanks for taking some time to look into this either way! 🥳

PR #45703, opened 2023-01-18
fix: migration stub  method  type key as static  added

anilkumarthakur60:
fix: migration stub method type key as ```static``` added


ankurk91:
Why is labelled as fix ? 
What kind of problem does it solves?

anilkumarthakur60:
@ankurk91  it doesn't solve the problem ...it makes readability clear

PR #45704, opened 2023-01-18
[9.x] Add new throw helper methods to the HTTP Client

WendellAdriel:
The goal is to add four new methods to the HTTP Client:

```php

$response->throwIfStatus(500); // Only throws an exception if the HTTP response code is 500

$response->throwUnlessStatus(200); // Throws an exception if the HTTP response code is not 200

$response->throwIfClientError(); // Only throws an exception if the HTTP response code is >= 400 and < 500

$response->throwIfServerError(); // Only throws an exception if the HTTP response code is >= 500
```

Sometimes we want to only throw an exception in our application for specific codes of the response, so these method add an easy way to accomplish that.

The `throwIf` method only throws the exception if two conditions are `true`: The condition passed resolves to `true` **AND** if the HTTP response code is `>= 400`


WendellAdriel:
@driesvints @timacdonald could you take a look at this and give some feedback if I should continue working on this and if so, could you help me understand some things?

ankurk91:
This type of helper (alias) methods are making the framework bulky and increasing learning curve.

$response->throwIf($callback) 

is already sufficient.

WendellAdriel:
I managed to work on a simpler approach for the implementation that I wanted.
Now it's ready!

@ankurk91 I understand your point, but the methods I'm adding are to decrease work for the developers. It's a syntax sugar to make it easier to throw exceptions based on the response HTTP status code.

WendellAdriel:
If this has any chances to be approved and merged and additional work is needed for docs or something like that, just let me know that I'll be happy to work on the needed updates as well 😉 💪🏼

ankurk91:
Thanks, you gave me a reason.

taylorotwell:
What if you want to throw if the status is *any* 500 level status code?

WendellAdriel:
@taylorotwell we can do it on two ways. Adding additional parameters to the methods I created or create two other ones called `throwIfClientError` for 40X codes and `throwIfServerError` for 50X codes. What do you think?

WendellAdriel:
@taylorotwell I added a new commit adding the two methods I mentioned:

- `throwIfClientError`
- `throwIfServerError`

If you feel like it's needed I can add four more methods:

- `throwIfStatusIn` that will receive an array of codes to throw an exception
- `throwIfStatusBetween` that will receive two params and throw an exception for all codes between both of them
- `throwUnlessStatusIn` that will be the opposite of the `throwIfStatusIn`
- `throwUnlessStatusBetween` that will be the opposite of the `throwIfStatusBetween`

If you think that's not needed or that I can work on those other four methods on a new PR it's all good for me. I'll be happy to help in all cases. 😉 💪🏼

PR #45706, opened 2023-01-18
[9.x] Add Encrypted validation rule

hossammonir:
This PR add new encrypted validation rule, The goal of this rule to validate the string to make sure it's an valid encrypted .

Before 
If you want to validate any encrypted values you will use 

```
try {
    decrypt($value)
} (DecryptException $e) {
     return $e;
}
```

After
Simply use validation rules 

```
    {
        return [
            'secrets' => ['encrypted'],
        ];
    }
```


JurianArie:
What is your usecase? The app key is not a shared secret, so who is encrypting the values?

dennisprudlo:
@JurianArie 
You could encrypt data on your server, giving the encrypted payload to the user and the user may come back with the payload at some point. Just like email verification works with signed URLs.

However I don't think validating if a value is correctly encrypted with the app key is not a common task. So maybe a custom rule would be best.

taylorotwell:

PR #45709, opened 2023-01-18
[9.x] Optimize destroy method 🏃🏻‍♂️

imanghafoori1:
Since the `whereKey` is optimized to use the `whereIntegerRaw` whenever is possible, we can use it in the destroy method as well.

The only little change in the underlying select query is that now it uses the fully qualified key.
- **(where "table"."key" in ... )** instead of key alone **(where "key" in ... )**

- It also adds a few missing tests for edge cases.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1347,11 +1347,9 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
         // We will actually pull the models from the database table and call delete on
         // each of them individually so that their events get fired properly with a
         // correct set of attributes in case the developers wants to check these.
-        $key = ($instance = new static)->getKeyName();
-
         $count = 0;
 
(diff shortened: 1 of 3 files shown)


leonardocustodio:
I don't know if this is intended or not, but now I can't destroy an array of models created with factory, here is an example:

```
$collection = Collection::factory([
            'owner_wallet_id' => $this->wallet,
            'token_count' => 1,
            'attribute_count' => 1,
])->create();

Collection::destroy($this->collections);
```

Now we get the error: `ErrorException : Object of class Platform\Core\Models\Collection could not be converted to int`. This did not use to happen before this update.

imanghafoori1:
@leonardocustodio
Thanks for your report. I will look into it.
It seems that you are passing an array of model instances into the destroy method.

taylorotwell:
Reverting.

imanghafoori1:
@leonardocustodio 
It seems that you are passing an array of models into the destroy method.
Are you sure that it really deletes the model from the database?
And please also mention your database driver.

Thanks

PR #45710, opened 2023-01-18
[9.x] Migrator strict mode

faissaloux:
This PR introduces `Migrator::preventRollback($bool)` and `Migrator::shouldBeStrict($bool)`.

Both methods prevent database rollback for now, we can add more preventions into `shouldBeStrict` (Open to any preventions that will help) and you can specify the condition when you want the `rollback` to be prevented.

To activate `strict mode` you just need to call it in the `boot()` method in a registered service provider.

```php
Migrator::shouldBeStrict(App::environment('production'));
```

To prevent only the rollback you can use
```php
Migrator::preventRollback(App::environment('production'));
```

If `strict mode` is activated and you trying to rollback a `WARNING` will show up on your console without rolling back the last migration.

Screenshot 2023-01-18 171227

### DIFF
--- a/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/RollbackCommand.php
@@ -51,6 +51,11 @@ class RollbackCommand extends BaseCommand
      */
     public function handle()
     {
+        if ($this->migrator->preventsRollback()) {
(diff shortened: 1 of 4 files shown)


taylorotwell:
I don't really have a lot of appetite for more "strict mode" type features atm. Thanks though!

PR #45712, opened 2023-01-18
[9.x] Add noActionOnDelete method

erfanhemmati:
This PR will add a new method in the `ForeignKeyDefinition` class with this method clients are able to create a foreign key in their migrations with the `no action` property/option.

- example
`$table->foreign('user_id')->references('id')->on('users')->noActionOnDelete();`


erfanhemmati:
@taylorotwell

PR #45714, opened 2023-01-18
[10.x] Prevent stray requests by default

jasonmccreary:
This makes `preventStrayRequests` the default behavior in Laravel 10. Having **strong** tests is always a good default. Many newcomers to Laravel (or testing) may not be aware of `preventStrayRequests` and as such write tests which allow _real_ HTTP requests.

While this is a breaking change, it is targeted at Laravel 10. In addition, I added a top-level `Http::allowStrayRequests()` method which apps with an extensive test suite could quickly add to their test case `setUp` method to avoid changing each `Http::fake()` call. With that said, it's also a pretty easy search and replace.

### DIFF
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -60,7 +60,7 @@ class Factory
      *
      * @var bool
      */
-    protected $preventStrayRequests = false;
+    protected $preventStrayRequests = true;
 
     /**
      * Create a new factory instance.
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -60,7 +60,7 @@ class Factory
      *
      * @var bool
      */
-    protected $preventStrayRequests = true;
+    protected $preventStrayRequests = false;
 
     /**
      * Create a new factory instance.
--- a/src/Illuminate/Support/Facades/Http.php
+++ b/src/Illuminate/Support/Facades/Http.php
@@ -136,6 +136,8 @@ class Http extends Facade
      * Indicate that an exception should be thrown if any request is not faked.
      *
      * @return \Illuminate\Http\Client\Factory
+     *
+     * @deprecated
      */
     public static function preventStrayRequests()
     {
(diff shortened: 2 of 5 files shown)


jasonmccreary:
I'm not sure what these test failures are. They seem unrelated as the tests I wrote pass locally. However, I will try to run the entire test suite as well.

cbaconnier:
I like the idea to prevent stray requests by default when running tests. But doesn't this also prevent the requests in the application as well?

About the errors:
I think the two falling tests are because they actually make an HTTP request: https://github.com/laravel/framework/blob/af11be582547f503d5ccf2b26bc81eaa833d8241/src/Illuminate/Validation/NotPwnedVerifier.php#L90

But I could be wrong, I traced the error back from `testUncompromised` on github 

- `Illuminate\Tests\Validation\ValidationPasswordRuleTest::testUncompromised`
- `Illuminate\Tests\Validation\ValidationPasswordRuleTest::testMessagesOrder`

jasonmccreary:
@cbaconnier, interesting. I didn't realize the _fake_ wasn't its own class like other _fakes_. As such, I'll revert the property change and ensure the `fake()` call builds one by calling `preventStrayRequests()`.

jasonmccreary:
I have recreated this behavior through the facade as the marking the property in the client prevents requests in the app (as noted by @cbaconnier).

I also marked `preventStrayRequests` as deprecated since it is now the default behavior.

jasonmccreary:
Alright, this is passing and ready for review.

It comes with a caveat that is likely easier to document than it would be to solve via code given the static/singleton nature of facades. Let me explain:

Prevent stray requests is now the default simply by calling `Http::fake()`. However, If you have the following code:

```php
Http::allowStrayRequests();

// ...

Http::fake();
```

Then you may send _stray_ requests. But in order to then prevent stray requests again, you would need to explicitly call `preventStrayRequests()`.

```php
Http::allowStrayRequests();

// ...

Http::preventStrayRequests()->fake();
```

taylorotwell:
I think I will just table this for now.

PR #45715, opened 2023-01-18
[10.x] Add optional parameter for name

jasonmccreary:
I constantly find myself trying to remember the correct syntax for sending a mail to a raw email/name combination. Digging in the framework shows you may pass an array with a key/value pair. However, you also have to remember to pass this as a nested array.

This PR adds an optional second parameter for name, to allow conveniently sending mail to a email/name.

**Before:**
```php
Mail::to([['email' => 'jmac@example.com', 'name' => 'JMac']])->send($mailable);
```

**After:**
```php
Mail::to('jmac@example.com', 'JMac')->send($mailable);
```

Note, this parameter is only used when set and the first parameter is a `string`.

### DIFF
--- a/src/Illuminate/Mail/Mailer.php
+++ b/src/Illuminate/Mail/Mailer.php
@@ -155,7 +155,7 @@ class Mailer implements MailerContract, MailQueueContract
      * Begin the process of mailing a mailable class instance.
      *
      * @param  mixed  $users
-     * @param  string  $name
+     * @param  string|null  $name
      * @return \Illuminate\Mail\PendingMail
      */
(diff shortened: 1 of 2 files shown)


taylorotwell:
What about `cc` and `bcc` methods? Do the `PendingMail` methods allow this syntax?

jasonmccreary:
Yes, the underlying methods in Mailable all have a signature of `$address, $name = null`. I am glad to make all of these methods (`to`, `cc`, and `bbc`)symmetrical on the Mailer if you like.

taylorotwell:
Done!

PR #45717, opened 2023-01-19
[9.x] Add "missing" validation rules

timacdonald:
Missing is a strict version of the prohibits rule. Missing is implicit, so it runs even if the value is an empty string.

You can think of "missing" as running array_key_exists (but supporting dot notation).

Will follow up with documentation and skeleton PRs if we are happy with this one.

I have only added the permutations that I think have utility:

- `missing`
- `missing_if:attribute,value`
- `missing_unless:attribute,value`
- `missing_with:attribute1,attribute2`
- `missing_with_all:attribute1,attribute2`

Skeleton: https://github.com/laravel/laravel/pull/6078
Documentation: https://github.com/laravel/docs/pull/8465


taylorotwell:
Looks good, but needs corresponding method additions in `ReplacesAttributes.php`.

timacdonald:
Added.

PR #45718, opened 2023-01-19
[9.x] Add `conflict` method to http client response

lentex:
This PR adds a convenient method to check if a HTTP request resulted in a conflict (HTTP Status 409).

So instead of checking the status code manually, you can use this new method:

```php
// Before
$response = Http::post('https://www.laravel.com', []);

if ($response->status() === 409) {
    doStuff();
}

// After
$response = Http::post('https://www.laravel.com', []);

if ($response->conflict()) {
    doStuff();
}
```

### DIFF
--- a/src/Illuminate/Http/Client/Response.php
+++ b/src/Illuminate/Http/Client/Response.php
@@ -215,6 +215,16 @@ class Response implements ArrayAccess
     }
 
     /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
It's really fine to just check the status.

PR #45720, opened 2023-01-19
[9.x] Add listener wrapper ability for event objects

imanghafoori1:
### Problem:
The Dispatcher class allows referencing any method as a listener for a given event, but there is an important assumption on the listeners.
And that is: they should not throw exceptions or otherwise the execution of the other listeners will stop, and there is nothing that the dispatcher of the event can do about it. 
In other words, it can not set an exception handler for each and every listener.

For example, this does not work:
```php
try {
    \Event::dispatch('user-registered');
} catch (\Exception $e) {
    // handle the exception
}
```
In the example above, if the first listener throws an exception, all the other listeners would not be executed and it goes to the catch block directly.

-----------------------

### Solution:

```php
class MyEvent
{
    public function wrapListener(\Closure $listener)
    {
        try {
            return $listener( ); // <--- no need to pass anything, very simple.
       } catch (\Exception) {
           // handle
       }
    }
}
```

Dispatched as before:

```php
\Event::dispatch(MyEvent::class);
```
But this allows an event to set a try/catch for each and every listener.

- This is not needed if all the listeners are set by the person, who fires the event since the try/catch can set written in the 
listener.
But for a package developer who is defining an event in the package code, it is always a matter of question if the listeners set by the package users are healthy enough or not. In other words, you define and fire your events but you never know what is going to be executed as the listener.

### Other benefits:

- When having duplicated code in all the listeners at the start or the end, we can use these events to DRY that code on a nice wrapper.

- For testing:
Since the event object is resolved out of the container, it is possible to do swapping while testing:
```php
app()->bind('MyRealEvent', fn () => new MyFakyEvent);
```

```php
class MyFakyEvent
{
    public function wrapListener(\Closure $listener)
    {
         return random_int(1, 100); // avoids calling real listeners and return fake results.
    }
}
```

- In this implementation, the `wrapper` can not know about the details of the listener since it is already wrapped in a closure and can only call it and inspect or change the result of the call so it does NOT couple events to listeners.

### Backward-compatibility:
- This approach does not change any method signature or add any methods.
- You may also decide to add a new interface that the events implement to make it fully backward-compatible since there is a slight change that someone has this method already defined on the listener.

```php
interface WrapperEvent
{
  public function wrapListener($listener);
}
```

### DIFF
--- a/src/Illuminate/Events/Dispatcher.php
+++ b/src/Illuminate/Events/Dispatcher.php
@@ -232,6 +232,16 @@ class Dispatcher implements DispatcherContract
      */
     public function dispatch($event, $payload = [], $halt = false)
     {
+        if (method_exists($event, 'wrapListener')) {
+            if (is_string($event)) {
+                $event = $this->container->make($event);
+            }
+
+            $wrapper = fn ($listener) => $event->wrapListener($listener);
+        } else {
+            $wrapper = fn ($listener) => $listener();
+        }
+
         // When the given "event" is actually an object we will assume it is an event
         // object and use the class as the event name and this event itself as the
         // payload to the handler, which makes object based events quite simple.
@@ -246,7 +256,7 @@ class Dispatcher implements DispatcherContract
         $responses = [];
 
         foreach ($this->getListeners($event) as $listener) {
-            $response = $listener($event, $payload);
+            $response = $wrapper(fn () => $listener($event, $payload));
 
             // If a response is returned from the listener and event halting is enabled
             // we will just return this response, and not call the rest of the event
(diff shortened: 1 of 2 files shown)


ankurk91:
I wish if the syntax could looks like

```php
Event::dispatchWithErrorHandling(MyEvent::class)
// or
Event::continueOrError(MyEvent::class)
```

taylorotwell:
I'm just not convinced on this idea. Exceptions should only be thrown in exceptional circumstances where continuing could result in bad data, etc. In that case, you *want* the exception to be thrown and no more listeners to run. Something has gone fatally wrong.

imanghafoori1:
@taylorotwell 
Ok, something has gone fatally wrong but only in the context of a single listener so it should stop that particular listener in the middle of its execution.

The fact is, unlike the middlewares, the listeners are completely ignorant of each other. Each listener is doing its own thing independent of the other listeners. So the handler of a listener should be wrapped around that particular one.

PR #45727, opened 2023-01-19
Add fullUrlOnlyWithQuery request method

fleader64:
Add another variant of the request fullUrl method. It helps to get the full URL for the request ONLY with the given query string parameters.
For example, it helped me to get the canonical URL of the page with only allowed query parameters.


taylorotwell:
Hey there - you should be able to macro this into the request class for your own application.

PR #45731, opened 2023-01-20
[9.x] Adds `assertJsonIsArray` and `assertJsonIsObject` for TestResponse 

sethsandaru:
Hi team,

When interacting with Collection, after several actions (like `sort` or `filter`,...), the array won't be an indexed array like before (I think most of us here would probably encounter this at least once). 

And if we return the Collection immediately, it will result in a JS Object `{ ... }`. For listing endpoints, it would be wrong and result in **an error** from the **API consumers** (frontend, mobile,...) 

So to ensure we won't return the wrong data structure to the consumers, I added 2 assert methods to validate the data:

- `assertJsonIsArray`
- `assertJsonIsObject`

FYA: from `10.x` we can use `array_is_list` (PHP 8.1) to check against the data since it requires PHP 8.2 and above. 

## Usage
```php
$this->json('GET', 'countries')
    ->assertOk()
    ->assertJsonIsArray(); // [ {..}, {...}, ....]

$this->json('GET', 'users')
    ->assertOk()
    ->assertJsonIsArray('data'); // {'data': [...]}

$this->json('GET', 'countries/US')
    ->assertOk()
    ->assertJsonIsObject(); // {id: '...', name: '..'}

$this->json('GET', 'users/1')
    ->assertOk()
    ->assertJsonIsObject('data'); // {'data': {id: '..', name: '...'}}
```

## PR Information
- Implemented the methods
- Added tests and cover all cases
- No **breaking changes** - only introduces new methods

## Additional information
`assertJsonStructure` won't help us to validate that (check `TestResponseTest@testAssertJsonStructure`)

The wildcard `*` only validates the repeating structure. It doesn't care whether our data is a JS array or JS object.

Screenshot 2023-01-20 at 10 14 01

`assertJson` or `assertExactJson` might be a good alternative, but for listing endpoints, preparing a big list of data to assert would not so fun, right?

Thanks all!

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -967,6 +967,48 @@ class TestResponse implements ArrayAccess
     }
 
     /**
+     * Assert that the key that holding the data must be a json array
+     *
+     * @param $key
+     *
+     * @return $this
+     */
+    public function assertJsonIsArray($key = null)
+    {
+        $data = $this->json($key);
+        $encodedData = json_encode($data);
+
+        PHPUnit::assertTrue(
+            is_array($data)
+            && str_starts_with($encodedData, '[')
+            && str_ends_with($encodedData, ']')
+        );
+
+        return $this;
+    }
+
+    /**
+     * Assert that the key that holding the data must be a json object
+     *
+     * @param $key
+     *
+     * @return $this
+     */
+    public function assertJsonIsObject($key = null)
+    {
+        $data = $this->json($key);
+        $encodedData = json_encode($data);
+
+        PHPUnit::assertTrue(
+            is_array($data)
+            && str_starts_with($encodedData, '{')
+            && str_ends_with($encodedData, '}')
+        );
+
+        return $this;
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


timacdonald:
I don't think this would handle _empty_ objects / arrays well. It would just assume they are all arrays.

If the payload was...

```json
{ "users": {} }
```

When you json encode that in PHP it is going to have to cast it to an object or an array based on flags.vvfg

sethsandaru:
@timacdonald Not sure if I understand you correctly.

So for Laravel Response, **by default**, if you pass an empty array `response()->json([])` or `new JsonResponse([])` => `[]` will be rendered => assertJsonIsArray is working correctly?

But for some cases, you wanted the `users` to be an `object`, then in your controller, you would do something like `response()->json(new YourObject());` (if the object is empty => `{}` => assertJsonIsObject is working correctly?

timacdonald:
@sethsandaru I was concerned about how an empty hash in PHP is JSON encoded as an empty array, but I don't think that impacts this feature, and if anything this feature actually forces users to consider that.

Sorry for my rushed comment without further testing.

PR #45698, opened 2023-01-18
[9.x] Add `paymentRequired` helper to Http Client response

tjardoo:
This PR adds the `paymentRequired` to the Http Client Response class. The 402 error page was just merged into the framework [ [9.x] Add 402 exception view #45682](https://github.com/laravel/framework/pull/45682)

Currently we have to use the following code to check if a response is 402 Payment Required.

```php
$response = Http::get('https://laravel.com');

if ($response->status() === 402) {
    doSomething();
}
```

With this PR you can do the following, which is a bit cleaner.

```php
$response = Http::get('https://laravel.com');

if ($response->paymentRequired()) {
    doSomething();
}
```

Thanks!

PR #45734, opened 2023-01-20
[9.x] Adds magic status methods and assertions

timacdonald:
As seen with may recent PRs, there is a thirst from the community to be able to have named status code methods.

This PR aims to bring named methods to all possible places without the need for a gazillion individual methods.

## Test response

```php
$response = $this->get(/* ... */);

$response->assertConflict(); // ✅|❌
```

## Http client response

```php
$response = Http::get(/* ... */);

if ($response->paymentRequired()) {
    //
}
```

Recent attempts to add a variety of named methods:

- https://github.com/laravel/framework/pull/45718
- https://github.com/laravel/framework/pull/45701
- https://github.com/laravel/framework/pull/45698
- https://github.com/laravel/framework/pull/45681
- https://github.com/laravel/framework/pull/44658
- etc.

Edit: I've removed the ability to use named methods on the response factory, as we would have to handle standard responses, JSON responses, streamed responses, etc.,  and that feels like a lot.

### DIFF
--- a/src/Illuminate/Http/Client/Response.php
+++ b/src/Illuminate/Http/Client/Response.php
@@ -484,7 +484,7 @@ class Response implements ArrayAccess
             return $this->macroCall($method, $parameters);
         }
 
-        $statusConstant = HttpResponse::class.'::HTTP_'.Str::of($method)->after('assert')->snake()->upper();
+        $statusConstant = HttpResponse::class.'::HTTP_'.Str::of($method)->snake()->upper();
 
         if (defined($statusConstant) && array_key_exists($status = constant($statusConstant), HttpResponse::$statusTexts)) {
             return $this->status() === $status;
--- a/src/Illuminate/Http/Client/Response.php
+++ b/src/Illuminate/Http/Client/Response.php
@@ -484,9 +484,12 @@ class Response implements ArrayAccess
             return $this->macroCall($method, $parameters);
         }
 
-        $statusConstant = HttpResponse::class.'::HTTP_'.Str::of($method)->snake()->upper();
+        $status = collect(HttpResponse::$statusTexts)
+            ->map(fn ($status) => (string) Str::of($status)->slug()->camel())
+            ->flip()
+            ->get($method);
 
-        if (defined($statusConstant) && array_key_exists($status = constant($statusConstant), HttpResponse::$statusTexts)) {
+        if ($status !== null) {
             return $this->status() === $status;
         }
 
--- a/src/Illuminate/Http/Client/Response.php
+++ b/src/Illuminate/Http/Client/Response.php
@@ -3,7 +3,9 @@
 namespace Illuminate\Http\Client;
 
 use ArrayAccess;
+use Illuminate\Http\Response as HttpResponse;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
 use LogicException;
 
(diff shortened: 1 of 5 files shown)


andiLeong:
@timacdonald great catch :clap: :clap:

timacdonald:
Note to self: test dashes. ensure all response types make sense with standard parameter format.

WendellAdriel:
That's amazing @timacdonald! 💪🏼

bert-w:
I think I'd prefer `$response->assertStatus(451)` over `$response->assertUnavailableForLegalReasons()` because it uses a universal numeric code (I am not sure if the associated title/name for a status code is actually a strict spec?).

a-bashtannik:
What about code completion?

deleugpn:
Reading, understanding and debugging that magic is a lot harder than a few dozen of one-liner methods that are statically analysable without, err, more magic

thinkverse:
> I am not sure if the associated title/name for a status code is actually a strict spec?

It is, `451` is *Unavailable for legal reasons* as per [RFC 7725](https://www.rfc-editor.org/rfc/rfc7725.html). IANA has all status codes, their respective descriptions, and RFCs available in a [public registry](https://www.iana.org/assignments/http-status-codes/http-status-codes.xhtml).

thinkverse:
@timacdonald what about headers? Do they get proxied through `...$parameters` then? 🤔 Given `451` for example, as per [RFC 7725 Section 4](https://www.rfc-editor.org/rfc/rfc7725.html#section-4) should include a `Link` header to identify the blocking entity.

```php
return response()->unavailableForLegalReasons('Unavailable due to DMCA.', [
    'link' => '<https://legal.laravel.com>; rel="blocked-by"',
]);
```

Or would we have to use the `$headers` on the `response` function instead?

```php
return response(headers: [
    'link' => '<https://legal.laravel.com>; rel="blocked-by"',
])->unavailableForLegalReasons('Unavailable due to DMCA.');
```

ziming:
I think the reason why there is a flood of these status code PRs is because this got merged and people thought that would be an easy way to score a merged PR 😂. by doing it for other status codes

https://github.com/laravel/framework/pull/45681

thinkverse:
> Not really because there is a "thirst" from the community.

Possible, but I do see why that PR (**45681**) was made, and I can also see why it didn't get pushback from the community. Having helper functions for things that developers do often is great. Currently there are 4 helper functions for specific status codes, and those four make sense.

When calling an API you're very likely to check if the status is `ok`, or the API returns `unauthorized` because your API key is wrong, or `forbidden` if your permissions aren't correct, or a `notFound` if the endpoint is wrong.

This PR makes more sense from a DX point of view, it's easier to reason about a piece of code that reads `$response->paymentRequired()` than `$request->status() === 402`.

I think the reason why this got backlash is because for better or worse a large part of the community doesn't like "magic" methods. 🤷‍♂️

bert-w:
> Possible, but I do see why that PR (45681) was made, and I can also see why it didn't get pushback from the community. Having helper functions for things that developers do often is great. 

Imho that PR #45681 should be reverted. There was not even 3 hours between the PR and the merge which probably explains why not a single comment was made. I think every single one of those `status` wrappers should be removed. `$response->ok()` is not good code since that "ok" could mean anything when we consider the rest of the Laravel framework. It is so unintuitive compared to simply stating `$response->status === 200`.

The only ones that might have some merrit that currently exist are `successful()` and `failed()`.

(then again the whole Laravel wrapper around Guzzle is meh and just renames/adds some functions for slightly less code; I've been using the actual Guzzle package for ages and it works completely fine as is)

PR #45738, opened 2023-01-20
[9.x] Add dropForeignWithIndex method in Blueprint class

erfanhemmati:
This PR will add a new method in the `Blueprint` class with this method clients are able to drop a foreign key with the related index in their migrations

- example
```
  $table->dropForeignWithIndex('users_office_id_foreign');
});
```


driesvints:
Seems like branching went wrong here.

PR #45741, opened 2023-01-20
[9.x] Lottery chances float odds percentage

andiLeong:
In the lottery class , we can pass float to the constructor like this

```
Lottery::odds(5.8,, 10);
```

In the above example users expect a _true_ 58% odds percentage, but in our current calculation it is only 50% always.
because this is how we determine the winning

```
random_int(1, $outOf) <= $chances
```

the random_int always returns the integer so in this case it always gets one of numbers between 1 to 10.
That means it only has a 50% chance of winning.

**propose**
to solve this simply multiply the chances and outOf value
eg

```
Lottery::odds(5.8,, 10);
//become
Lottery::odds(58,, 100);

Lottery::odds(3.15,, 9);
//become
Lottery::odds(315,, 900);
``

@timacdonald what do you think? 
Do you think we need to take care of the float odds calculation or this is an edge case just ask users to pass a non-float number if they want accurate odds.

### DIFF
--- a/src/Illuminate/Support/Lottery.php
+++ b/src/Illuminate/Support/Lottery.php
@@ -50,13 +50,45 @@ class Lottery
      */
     public function __construct($chances, $outOf = null)
     {
-        if ($outOf === null && is_float($chances) && $chances > 1) {
-            throw new RuntimeException('Float must not be greater than 1.');
-        }
+        [$chances, $outOf] = $this->parseChancesAndOutOf($chances, $outOf);
 
         $this->chances = $chances;
 
         $this->outOf = $outOf;
+    }
+
+    /**
+     * parse and validate chances and outOf
+     *
+     * @param  int|float  $chances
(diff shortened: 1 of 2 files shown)


bert-w:
It appears that using `float` for the first parameter is meant for using as a single parameter only, so `::odds(0.4)` means a 40% chance of success. Combining it with the second parameter has untested behaviour.

I think the best solution therefore is to check that the second parameter is not given in case the first parameter is a float, and if it is then throw a `RuntimeException` like how it already happens in the lottery constructor.

bert-w:
https://github.com/laravel/framework/pull/45744

taylorotwell:
I would just use `Lottery::odds(58, 100)`

andiLeong:
@bert-w did not notice the previous pr haha :grin:

bert-w:
i made it after yours :)

PR #45745, opened 2023-01-20
[9.x] Optimize findMany of BelongsToMany 🏃🏻‍♂️

imanghafoori1:
Since `whereKey` uses the `whereIntegerInRaw` it would be better to use it. 
This PR applies it on `touch` and `findMany` methods.

- Also adds integration tests for models with stringy primary key.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php
@@ -688,8 +688,8 @@ class BelongsToMany extends Relation
             return $this->getRelated()->newCollection();
         }
 
-        return $this->whereIn(
-            $this->getRelated()->getQualifiedKeyName(), $this->parseIds($ids)
+        return $this->whereKey(
+            $this->parseIds($ids)
         )->get($columns);
     }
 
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Eloquent/Relations/BelongsToMany.php
```diff
@@ -688,8 +688,8 @@ public function findMany($ids, $columns = ['*'])
             return $this->getRelated()->newCollection();
         }
 
-        return $this->whereIn(
-            $this->getRelated()->getQualifiedKeyName(), $this->parseIds($ids)
+        return $this->getRelated()->whereKey(
```
taylorotwell, line 691: This looks like different logic to me. Call `whereKey` on `this`.

imanghafoori1: I wonder how the tests pass in both cases, but I thought the key name should be derived based on the related model.

imanghafoori1: I wonder how the tests pass in both cases, I thought that the key name should be derived from the related model, since we had ` $this->getRelated()->getQualifiedKeyName()` before, as the first argument to whereIn.

PR #45747, opened 2023-01-20
[9.x] Add tests for custom phpunit assertions

abenerd:
There are some custom phpunit assertion that have no tests for themselves apart from being used and working, this PR adds some tests to explicitly test that the assertions behave correctly.


taylorotwell:
I think it's OK.

PR #45750, opened 2023-01-21
[9.x] Add `getAddition` to JsonResource

michaelnabil230:
This PR adds a `getAddition` method to the JsonResource class to get the addition from the additional array.


taylorotwell:
The `$additional` data is already a public property. Just access it.

PR #45751, opened 2023-01-21
[9.x] Remove unnecessary returns

daftspunk:
These don't do anything and are not exposed anywhere.

### DIFF
--- a/src/Illuminate/Database/Eloquent/SoftDeletes.php
+++ b/src/Illuminate/Database/Eloquent/SoftDeletes.php
@@ -69,17 +69,17 @@ trait SoftDeletes
     /**
      * Perform the actual delete query on this model instance.
      *
-     * @return mixed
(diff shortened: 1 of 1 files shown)


imanghafoori1:
The class can be extended and the method can be called in the subclasses. So we can not say that no one needs the returned result.

daftspunk:
Doubtful since the base class already returns void for the same operation: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Eloquent/Model.php#L1445

Relying on this result would be unreliable as it stands. This looks more like an uncorrected oversight...

driesvints:
This will allow the process to continue past the if statement which wasn't the case before.

daftspunk:
Yes apologies missed an empty return inside the if statement. So long as you are aware of this. Thanks

PR #45752, opened 2023-01-21
[9.x] Add `Arrayble` in to `Arr`

michaelnabil230:
**NOTE**
If the idea is good for the laravel team, I can complete the code and add tests and all functions. I just want approval from the team to complete the code

This PR adds an `of` method to the `Arr` class so that we can create more than one process at the same time

For Example:

```php
$data = [
    'id' => 1,
    'name' => 'Michael Nabil',
    'role' => [
        'id' => 1,
        'name' => 'admin',
    ],
];

return (array) Arr::of($data)
    ->when(true, function ($arrayable) {
        $arrayable->add('age', 23);
    })
    ->undot()
    ->except('age');
```

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -4,12 +4,24 @@ namespace Illuminate\Support;
 
 use ArgumentCountError;
 use ArrayAccess;
+use Illuminate\Support\Arrayable;
 use Illuminate\Support\Traits\Macroable;
 use InvalidArgumentException;
 
 class Arr
 {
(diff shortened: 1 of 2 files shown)


andiLeong:
we can pass the array to collection if we want chaining method call :green_heart:

huangdijia:
What's the difference with collection?

ankurk91:
I would rather use collection.
There is no benifit of carrying 2 similar implementation in framework.

New feature also bring the cost of maintaining them.

taylorotwell:

PR #45754, opened 2023-01-21
[9.x] Add "addRestoreOrCreate" extension to SoftDeletingScope

orange-alien:
I would like to suggest adding addRestoreOrCreate extenstion to the IlluminateDatabaseEloquentSoftDeletingScope.

Following exampels are assumed that User model use SofteDelete trait.

## before this PR
* We need to check if user exists or not and do restore() or create().

```

$user = User::withTrashed()
    ->where(['email' => 'user@example.com'])
    ->first();

if($user) {
    $user->restore();
} else {
    $user = User::create([
        'email' => 'user@example.com',
        'name' => 'user name',
    ]);
}

```

## after this PR
* We don't have to worry about if user exist or not.
```

$user = User::restoreOrCretae(
    ['email' => 'user@example.com'],
    ['name' => 'user name'],
);

### DIFF
--- a/src/Illuminate/Database/Eloquent/SoftDeletingScope.php
+++ b/src/Illuminate/Database/Eloquent/SoftDeletingScope.php
@@ -86,10 +86,6 @@ class SoftDeletingScope implements Scope
             $builder->withTrashed();
 
             return tap($builder->firstOrCreate($attributes, $values), function ($instance) {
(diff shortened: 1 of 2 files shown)


rowdy4e:
Just a thought, shouldn't this be searched with onlyTrashed instead of withTrashed? Because otherwise the restore function is applied even to a record that hasn't really been deleted.

PR #45755, opened 2023-01-21
[9.x] Cover session store test on setExists method

andiLeong:
the session store test does not cover test for public method ```setExists``` therefore I added
thanks


taylorotwell:
We don't typically test simply setters.

PR #45760, opened 2023-01-21
Update user Provider

doxroot:
Based on https://github.com/laravel/fortify/issues/437


driesvints:
We can’t change contracts on patch releases

PR #45762, opened 2023-01-22
[9.x] Allow brick/math 0.11 also

barryvdh:
Allows brick/math version 0.11 besides version 0.10

See https://github.com/brick/math/releases/tag/0.11.0
Doesn't seem to contain breaking changes for Laravel.

0.10 is required for uuid, but will also allow 0.11 later: https://github.com/ramsey/uuid/pull/488 (if merged)

### DIFF
--- a/src/Illuminate/Database/composer.json
+++ b/src/Illuminate/Database/composer.json
@@ -17,6 +17,7 @@
     "require": {
         "php": "^8.0.2",
         "ext-json": "*",
+        "brick/math": "^0.10.2|^0.11",
         "illuminate/collections": "^9.0",
         "illuminate/container": "^9.0",
         "illuminate/contracts": "^9.0",
--- a/src/Illuminate/Validation/composer.json
+++ b/src/Illuminate/Validation/composer.json
@@ -16,6 +16,7 @@
     "require": {
         "php": "^8.0.2",
         "ext-json": "*",
+        "brick/math": "^0.10.2|^0.11",
         "egulias/email-validator": "^3.2.1|^4.0",
         "illuminate/collections": "^9.0",
         "illuminate/container": "^9.0",
--- a/composer.json
+++ b/composer.json
@@ -18,7 +18,7 @@
         "php": "^8.0.2",
         "ext-mbstring": "*",
         "ext-openssl": "*",
-        "brick/math": "^0.10.2",
+        "brick/math": "^0.10.2|^0.11",
         "doctrine/inflector": "^2.0",
         "dragonmantank/cron-expression": "^3.3.2",
         "egulias/email-validator": "^3.2.1|^4.0",


driesvints:
Can you update the other brick requirements in other composer.json files as well?

barryvdh:
It doesn't seem to be added to the other packages, but I think it probably should be?
I've added it the the Database and Validation package. cc @timacdonald as this is an addition to #45729

src/Illuminate/Database/composer.json
```diff
@@ -17,6 +17,7 @@
     "require": {
         "php": "^8.0.2",
         "ext-json": "*",
+        "brick/math": "^0.10.2|^0.11",
```
GrahamCampbell, line 20: I don't think this should have been added here. Only as a suggest.

src/Illuminate/Validation/composer.json
```diff
@@ -16,6 +16,7 @@
     "require": {
         "php": "^8.0.2",
         "ext-json": "*",
+        "brick/math": "^0.10.2|^0.11",
```
GrahamCampbell, line 19: I don't think this should have been added here. Only as a suggest.

timacdonald:
Thanks @barryvdh! I added these in https://github.com/laravel/framework/pull/45693 but didn't realise I left it in draft, so it has just been sitting there 🙈

src/Illuminate/Database/composer.json
timacdonald: This would have been a breaking change if it was only a suggest. Anyone currently using the `decimal` model cast would have had issues on update.

src/Illuminate/Validation/composer.json
timacdonald: This would have been a breaking change if it was only a suggest. Anyone currently using the `multiple_of` validation rule would have had issues on update.

Temepest74:
> Allows brick/math version 0.11 besides version 0.10
> 
> See https://github.com/brick/math/releases/tag/0.11.0 Doesn't seem to contain breaking changes for Laravel.
> 
> 0.10 is required for uuid, but will also allow 0.11 later: [ramsey/uuid#488](https://github.com/ramsey/uuid/pull/488) (if merged)

Yes it contains. Today I updated from v9.48.0 to v9.49.0, and I got this error from Brick Math:
The given value " 15" does not represent a valid number. {"userId":1,"exception":"[object] (Brick\\Math\\Exception\\NumberFormatException(code: 0): The given value \" 15\" does not represent a valid number. at /var/www/html/vendor/brick/math/src/BigNumber.php:71)

I got this error inside a FormRequest and to solve this error I had to change
```
                    'numeric',
                    'between:1, 15',
``` 
to 
```
                    'numeric',
                    'between:1,15',
```
Not a big deal, and idk if the space was a bug to begin with.

driesvints:
@Temepest74 I don't think we can consider that a valid use case sorry.

RikSomers:
Yes, we had the same error here @Temepest74 . Luckily like with your solution, we had to remove the space and it started working again.

rolfdenhartog:
Maybe I've got a valid use case. I ahve a test that is validating the value of a field, which should be a number of course. But the test is failing since the update of `brick/math`. I've included the most important parts.

The validation from the `FormRequest` class:
```php
'year_of_birth' => [
    'required',
    'numeric',
    'min:'.$yearOfBirthMin,
    'max:'.$yearOfBirthMax,
],
```

The data that's sent using a POST request:
```php
['year_of_birth' => '']
```

I could remove the test, but that's not what I want. For now I've configured `conflicts` in `composer.json`.

If there is anyway I can help, please let me know.

driesvints:
We fixed the spacing issue btw. @timacdonald ^

driesvints:
@rolfdenhartog could you add a very simple example with rules as well as values that the validator now fails for?

timacdonald:
FYI that I believe we fixed the spacing issue in https://github.com/laravel/framework/pull/45912. It has not yet been tagged, so this will likely still fail until we tag a new release.

timacdonald:
This should have been fixed with https://github.com/laravel/framework/pull/45912, however it has not been tagged.

If you can provide a reproduction case that still fails after that fix, I'll dig into it and get it sorted.

rolfdenhartog:
Okay. Please give some time to provide as much as possible info. #kids 😬 I will also try to give some code to reproduce.

rolfdenhartog:
My apologies! I had a very small bug in my code. Which was visible thanks to the update. I owe you guys a beer in Australia or Belgium. Australia will probably take a while, but in two months I'll be in Belgium 🍺

rolfdenhartog:
For your information: The application I'm maintaining contains almost 700 (feature) tests. I had to fix a few tests and everything is 🟢 again. I've tested it with the latest version 9 release and also the `9.x` branch.

timacdonald:
That is great to hear.

PR #45764, opened 2023-01-22
[9.x] Fix blade tag issue with nested calls 🔧

imanghafoori1:
Fixes the reported bug introduced in https://github.com/laravel/framework/pull/45490#issuecomment-1399482841

- It seems that having multiple calls to other blade tags in between the parenthesis of another call can create problem.
- The added tests also pass with the logic we had before the fix. #45490

### DIFF
--- a/tests/View/Blade/BladePhpStatementsTest.php
+++ b/tests/View/Blade/BladePhpStatementsTest.php
@@ -141,4 +141,35 @@ class BladePhpStatementsTest extends AbstractBladeTestCase
 
         $this->assertEquals($expected, $this->compiler->compileString($string));
     }
+
+    public function testNestedTagCalls()
+    {
+        $string = "<span @class(['k' => @empty(\$v)])></span>";
+        $expected = '<span class="<?php echo \Illuminate\Support\Arr::toCssClasses([\'k\' => @empty($v)]) ?>"></span>';
+        $this->assertEquals($expected, $this->compiler->compileString($string));
+
+        $string = "<span @class(['k))' => @empty(\$v)])></span>";
+        $expected = '<span class="<?php echo \Illuminate\Support\Arr::toCssClasses([\'k))\' => @empty($v)]) ?>"></span>';
+        $this->assertEquals($expected, $this->compiler->compileString($string));
+
+        $string = "<span @class(['k' => @empty(\$v), 't' => @empty(\$v1)])></span>";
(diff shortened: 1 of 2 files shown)


rubenvanerk:
This fixes my issue, thanks!

tests/View/Blade/BladePhpStatementsTest.php
```diff
@@ -141,4 +141,35 @@ public function testUnclosedParenthesisForBladeTags()
 
         $this->assertEquals($expected, $this->compiler->compileString($string));
     }
+
+    public function testNestedTagCalls()
+    {
+        $string = "<span @class(['k' => @empty(\$v)])></span>";
+        $expected = '<span class="<?php echo \Illuminate\Support\Arr::toCssClasses([\'k\' => @empty($v)]) ?>"></span>';
```
taylorotwell, line 148: I don't understand - why isn't the `@empty` compiled. What was the behavior before your original Blade PR?

imanghafoori1: It is because now the regex eagerly captures the first ')' it sees, and continues on to the following ")" if the first one was not the right one.
In this example:
```php
@class(['k' => @empty($v1), 't' => @empty($v2)])
```
regex captures the `@class(['k' => @empty($v1` and `@empty($v2` as the `$matches` variable. Then starts to inspect the first one. (`@class(...`) While inspecting the `@class it add forward the rest of the template to reach the correct closing ")" and will result in :
`@class(['k' => @empty($v1), 't' => @empty($v2)])`.
All good up until now.
But the second match of regex $matches[1] is now redundant (`@empty($v2)`) because it is covered by the while loop, so we unset it to avoid extra compilation.

imanghafoori1: The original code also passes the same tests. I mean the @empty do not get compiled.
This is the original code before the fix:
```php  
  protected function compileStatements($value)
    {
        return preg_replace_callback(
            '/\B@(@?\w+(?:::\w+)?)([ \t]*)(\( ( (?>[^()]+) | (?3) )* \))?/x', function ($match) {
                return $this->compileStatement($match);
            }, $value
        );
    }
```

imanghafoori1:
@taylorotwell 
The Blade has never supported using tags within other tag calls. like this:  `@class( @include())`
So the @empty must not get compiled. In fact, the reported issue is treating the `@empty()` as a raw PHP function call and the `@` here is working as a raw PHP operator to suppress the errors.
image

PR #45767, opened 2023-01-23
[Breaking Change] Refactor validation rules: alpha, alpha_dash, alpha_num

fuwasegu:
This PR is a redo of #45609

## Reason for re-submitting PR
The reason why I refactored this again is discussed at #45660 . Please see.
In the previous PR, we received a comment that custom rules should be defined individually by the developer.
However, as described below, there is a problem that the definition of "alphabet" is implemented differently from what is generally known, which is not good for a Web framework widely used in the world.
Also, since we received a lot of support in the above Discussion, we thought it would be better to address this issue as a framework rather than as individual developers.

## What is this?
The validation rules: alpha, alpha_dash, alpha_num have a problem.
The common definition of "alphabet" or "alphabetic characters" is the 26 types of characters [a-zA-Z] (52 types of uppercase and lowercase characters combined).
However, Laravel's validation rules for verifying whether an input character is an alphabetic character actually allowed characters that went well beyond the definition of an alphabetic character.

This is the actual "alpha" rule code
```php
public function validateAlpha($attribute, $value)
{
    return is_string($value) && preg_match('/^[\pL\pM]+$/u', $value) > 0;
}
```

The following is the meaning of the properties included in this regular expression （from https://www.php.net/manual/en/regexp.reference.unicode.php）
スクリーンショット 2023-01-23 13 37 25

That's why, I believe the following code is the more correct one
```php
public function validateAlpha($attribute, $value)
{
    return is_string($value) && preg_match('/^[a-zA-Z]+$/u', $value) > 0;
}
```

This overly broad definition is contrary to the developers' intentions in a language area such as Japan, where multibyte characters are used on a daily basis.
I shared this opinion in the discussion above and got over 30 Upvodes. This is the second most Upvodes of any discussion topic in the Laravel project.

## Changes
- Refactored existing rule logic (alpha, alpha_num, alpha_dash) 4441ec3ace6576a3a7c7ebedb9947dd14b8672c7
  - Changed the rule to not allow multibyte characters according to the original definition of "alpha".
- Renamed existing definitions 59c39e6707c2a952b0ad76192bf379a7c72eb89e
  - The rule originally defined as "alpha" has been renamed to indicate that it "contains multibytes".
- Wrote the test. fa7c923a84ef15e157d46c30f4db10aef6275fd2

## What you get from this change
- Correct validation rules based on the original definition
- Clear separation of validation rules when multibyte is allowed and when it is not
- Preventing the use of validation rules contrary to the developer's intentions

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -334,7 +334,7 @@ trait ValidatesAttributes
     }
 
     /**
-     * Validate that an attribute contains only alphabetic characters.
+     * Validate that an attribute contains only ascii alphabetic characters.
      *
      * @param  string  $attribute
      * @param  mixed  $value
@@ -342,11 +342,11 @@ trait ValidatesAttributes
      */
     public function validateAlpha($attribute, $value)
     {
-        return is_string($value) && preg_match('/^[\pL\pM]+$/u', $value);
+        return is_string($value) && preg_match('/^[a-zA-Z]+$/u', $value) > 0;
     }
 
     /**
-     * Validate that an attribute contains only alpha-numeric characters, dashes, and underscores.
+     * Validate that an attribute contains only ascii alpha-numeric characters, dashes, and underscores.
      *
      * @param  string  $attribute
      * @param  mixed  $value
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hi there - sorry to close this again but I would actually leave the current rules as they are and instead introduce `ascii_alpha` variants. It's too breaking to change these existing rules which have existed in this way for many years.

lunain84:
@taylorotwell 
Thanks for the response!
I see...
I understood that this PR was a rather disruptive change, since the alpha and other variation rules have been implemented for quite some time, and I had to understand that since Laravel is a widely used and long-established web framework.
Can I implement your proposed ascii_alpha implementation from now on?

lunain84:
I tried to impl ascii validation rules to 9.x
https://github.com/laravel/framework/pull/45769

PR #45768, opened 2023-01-23
Remove check for deprecated guzzle code

elvan:
The function `get_message_body_summary()` from guzzle is no longer exist in the newest version of the library


taylorotwell:
I think that's here in case people are using older Guzzle versions?

driesvints:
@taylorotwell 7.0 was the version this was removed from which is also the minimum version for Laravel v9.

driesvints:
@elvan Guzzle v7 still relies on Guzzle PSR 1.x: https://github.com/guzzle/guzzle/blob/master/composer.json#L57

So we have to support the alternative here.

PR #45773, opened 2023-01-23
[11.x] Change order of composite morphable index members

cosmastech:
Morphable columns are built with a composite index on (`morphable_type`, `morphable_id`) which is fine if queries most commonly target the type OR if there are more types than there are unique IDs.

However, in my experience, it's more common that applications query for items that belong to a specific model. From what I understand about database indexing, a composite index is most useful if the first member has a lower cardinality than the next member.

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -1475,7 +1475,7 @@ class Blueprint
 
         $this->unsignedBigInteger("{$name}_id");
 
-        $this->index(["{$name}_type", "{$name}_id"], $indexName);
+        $this->index(["{$name}_id", "{$name}_type"], $indexName);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


cosmastech:
> Hey, did you make any benchmark results?

I did not. This is my first time opening a PR for Laravel, and to be honest, I had no idea what was expected of me for it.

Will gladly provide a benchmark (and fix the failing tests) if I can get some guidance on what maintainers would like to see. It would be great if there was an accepted PR to maybe take a look at to see how they were benchmarking.

tpetry:
I don't expect any performance difference. And you're breaking the behaviour of dopping morph indexes as you now need to know whether the index was create with Laravel >= 11.x or before and need to drop completely different indexes.

tpetry:
> A production database I work on has millions of rows in a media table (from spatie/media-library) and querying for media that belongs to an individual user (or other entity) is more costly than it needs to be.

And your task is to find out why. And like you said before you didn't even benchmark whether the change makes any difference any at all. That is not how performance optimization are made. You change only something because of perf when you can prove it is effective.

cosmastech:
> And your task is to find out why. And like you said before you didn't even benchmark whether the change makes any difference any at all. That is not how performance optimization are made. You change only something because of perf when you can prove it is effective.
> 
I don't have benchmarks from my production DB (because I think that violates privacy).

But yes it does seem to be effective from my tests. Going to spin up an instance for testing with two tables with matching data but different indices. That is data I can share publicly.

> I am out.

Wow what a friendly place.

PR #45778, opened 2023-01-24
[10.x] Use PSR Container in Pipeline component

rodrigopedra:
This PR makes the Pipeline component to depend on the PSR Container instead of Illuminate container.

This allows the Pipeline component to be used with another container implementation in projects that do not require the full Laravel framework, and already use a PSR Container implementation.

### DIFF
--- a/src/Illuminate/Pipeline/Hub.php
+++ b/src/Illuminate/Pipeline/Hub.php
@@ -3,15 +3,15 @@
 namespace Illuminate\Pipeline;
 
 use Closure;
-use Illuminate\Contracts\Container\Container;
 use Illuminate\Contracts\Pipeline\Hub as HubContract;
+use Psr\Container\ContainerInterface;
 
 class Hub implements HubContract
 {
     /**
      * The container implementation.
      *
-     * @var \Illuminate\Contracts\Container\Container|null
+     * @var \Psr\Container\ContainerInterface|null
      */
     protected $container;
 
@@ -25,10 +25,10 @@ class Hub implements HubContract
     /**
      * Create a new Hub instance.
      *
-     * @param  \Illuminate\Contracts\Container\Container|null  $container
+     * @param  \Psr\Container\ContainerInterface|null  $container
      * @return void
      */
-    public function __construct(Container $container = null)
+    public function __construct(ContainerInterface $container = null)
     {
         $this->container = $container;
     }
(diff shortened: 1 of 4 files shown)


rodrigopedra:
I also examined other Laravel packages for Pipeline usage, and none of them use any container method other than the basic methods addressed by this PR.

These other components also only use `->get()/->has()` (alias to `->make()/->bound()`) container's method:

- `Illuminate\Auth`: only within the `Gate` class
- `Illuminate\Events`: only within the `CallQueuedListener` and `Dispatcher` classes
- `Illuminate\Session`: only within the `DatabaseSessionHandler` class
- `Illuminate\Validation`: only within the `Factory`, `Validator`, and `Password` classes

If desired, I can send a PR to also make use of PSR container on those components.

Maybe just doing it for the Events and Validation components, as Auth and Session are more likely to be used with Laravel or paired with components that would require the Container component already.

I've seen both Events (very few times) and Validation (occasionally) used in the wild without Laravel.

Some other components (Support, Database, Queue, and HTTP) also have classes that use only those basic container methods, but they also contain classes which need the full container, so I didn't include those as by using those packages a developer would get the whole container component already.

When investigating, I noticed the `Illuminate\Http\Middleware\HandleCors` class only inject the container to use the Config Repository. Maybe we could consider injecting `Illuminate\Contracts\Config\Repository` directly. I can send this on a different PR if desired.

taylorotwell:
Why I hesitate to do this: if I add a more complicated method on the container that I would like to use in any of these packages, I can't do that anymore because I'm bound by the PSR contract until the next major release.

X-Coder264:
A new method couldn't be added to the Illuminate container contract until the next major release either so that reasoning doesn't seem to be valid, unless I'm misunderstanding something.

I think this change to the pipeline component is safe to make as the entire component itself is pretty much the same as it was half a decade ago.

deleugpn:
Adding a method to the concrete class without adding it to the interface is not a breaking change. Users of the interface won't have that new functionality until the next release, but users of the concrete implementation can.

rodrigopedra:
I guess what he also implicitly concerns, is if he wants to use a Container's method not available on PSR's Container.

For example, I could see how `Container@call` could be used to allow dependencies to be injected on a pipe method.

That wouldn't be a breaking change if using Illuminate Container, as it is just an implementation change.

PR #45784, opened 2023-01-24
[9.x] Add support for multiline strings in Str::excerpt method

nikinoldin:
In my opinion the Str::excerpt method's main use-case is to create excerpts from longer texts including paragraphs aka line-breaks. Therefore, I found myself often wrapping the method with another that replaces new line characters with spaces. 

By adding the DOTALL Regex flag (`s`) to the used expression it will also match on multiline strings.


ankurk91:
You may want to Target v10 now

PR #45785, opened 2023-01-24
[9.x] Add equals methods to Collection

wimski:
I've created some helper methods for comparing Collections using the already present diff methods.

It's basically a short-hand for converting this...
```php
if (
    $collectionA->diff($collectionB)->isEmpty()
    && $collectionB->diff($collectionA)->isEmpty()
) {
    // do something if both collections contain the same values
}
```

...into this.
```php
if ($collectionA->equals($collectionB)) {
    // do something if both collections contain the same values
}
```

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -302,6 +302,81 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
+     * Determine if the items equal the given items.
+     *
+     * @param  \Illuminate\Contracts\Support\Arrayable<array-key, TValue>|iterable<array-key, TValue>  $items
+     * @return bool
+     */
+    public function equals($items)
+    {
+        return $this->diff($items)->isEmpty()
+            && (new static($items))->diff($this->items)->isEmpty();
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


wimski:
Why do the tests fail with PHP 8.2 on a test that is completely unrelated to my changes? :astonished:

marcovo:
I suspect this implementation would falsely regard the arrays `['a', 'a', 'b']` and `['a', 'b', 'b']` to be equal. This points to an underlying fundamental problem: your implementation using `diff()`s is probably correct if the arrays represent two `Set`s, but for other data structures like lists or hashmaps, there will be different rules for what it means to be 'equal'; e.g. for lists you may expect the ordering within the arrays to be the same, for hashmaps you may want to ensure the same key/value pairs are present.

wimski:
> I suspect this implementation would falsely regard the arrays `['a', 'a', 'b']` and `['a', 'b', 'b']` to be equal.

Ah yes, I totally missed that 🙈 

 > [...] for hashmaps you may want to ensure the same key/value pairs are present.

Isn't that what the 'assoc' version is for?

> While I think `equals()` methods could be useful, I think these nuances should be addressed first, and probably there are other edge cases as well that should be carefully examined.

You're quite right. I was too hasty with my implementation. Thanks for the review!

taylorotwell:

PR #45788, opened 2023-01-24
[9.x] Remove database dependency on symfony/console

alpha0010:
Placing a requirement on `symfony/console` makes it hard to use the database subpackage with any framework that has its own `symfony/*` requirements.

Dependency is only used by database commands, so removing this dependency does not break anything for use case of not using database commands.

To use database commands, one must install `illuminate/console` which will in turn require `symfony/console`, so use case of using database commands will still get the correct dependencies.

(The existence of this dependency has forced us to use the very old 6.x release with our Drupal installation.)

### DIFF
--- a/src/Illuminate/Database/composer.json
+++ b/src/Illuminate/Database/composer.json
@@ -22,8 +22,7 @@
         "illuminate/container": "^9.0",
         "illuminate/contracts": "^9.0",
         "illuminate/macroable": "^9.0",
-        "illuminate/support": "^9.0",
-        "symfony/console": "^6.0.9"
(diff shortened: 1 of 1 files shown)


driesvints:
The Symfony component is also used in the Migrator class. 

It's also best that this is done in a major ne release instead.

alpha0010:
Shall I open a new PR against the 10.x branch then?

driesvints:
I'll let @taylorotwell decide.

taylorotwell:
We can move it to the `suggest` section in 10.x.

Issue #45781, opened 2023-01-24
dispatch method does not respect ShouldBeUnique on jobs


ankurk91:
- Laravel Version: 9.48..0
- PHP Version: 8.2
- Database Driver & Version: MySql
- Cache and Queue driver - Redis v7

### Steps To Reproduce:

Create a test job:
```php
<?php

namespace App\Jobs;

use Illuminate\Bus\Queueable;
use Illuminate\Contracts\Queue\ShouldBeUnique;
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Foundation\Bus\Dispatchable;
use Illuminate\Queue\InteractsWithQueue;
use Illuminate\Queue\SerializesModels;

class TestJob implements ShouldQueue, ShouldBeUnique
{
    use Dispatchable, InteractsWithQueue, Queueable, SerializesModels;

    public function __construct()
    {
        //
    }

    public function handle()
    {
        info("Running the job");
        sleep(3);
    }
}
```
Then create a Controller and dispatch the job via `dispatch()` helper available in all Controllers
```php
<?php

namespace App\Http\Controllers;

use App\Http\Controllers\Controller;
use App\Jobs\TestJob;

class TestController extends Controller
{
    public function index()
    {
        // TestJob::dispatch(); // works as expected 
        // dispatch(new TestJob()); // works as expected
        
        $this->dispatch(new TestJob()); // causing the issue
    }
}

```
Start the worker
```bash
php artisan queue:work --sleep=3
```

Now hit the controller via a route twice, notice worker will process the job twice.

The problem lies here in this trait

https://github.com/laravel/framework/blob/247735e224d27898197699c73909ee5ee2f93e01/src/Illuminate/Foundation/Bus/DispatchesJobs.php#L17

This method is dispatching the jobs in a different way which does not respect the `ShouldBeUnique` contract

driesvints:
This is intentional I feel. The former methods you used are all actions on the jobs themselves while the Dispatcher just works differently. If you want to use the ShouldBeUnique you already have two ways of making that work.

driesvints:
Feel free to send in a pr 👍

driesvints:
I meant to say: send in a PR to the docs. We can't change the behavior here.

alies-dev:
[...]

PR #45793, opened 2023-01-25
[10.x] make dispatch() behaviour same across the framework

ankurk91:
Fixes the bug https://github.com/laravel/framework/issues/45781

The dispatch method inside `DispatchesJobs` trait does not check for Unique Jobs.

The global [dispatch](https://github.com/laravel/framework/blob/f801dd731592281a48e592b27960663bfe5fafab/src/Illuminate/Foundation/helpers.php#L386) helper functions does check it here

https://github.com/laravel/framework/blob/f801dd731592281a48e592b27960663bfe5fafab/src/Illuminate/Foundation/Bus/PendingDispatch.php#L163

This PR make these 2 methods act same.

### DIFF
--- a/src/Illuminate/Foundation/Bus/DispatchesJobs.php
+++ b/src/Illuminate/Foundation/Bus/DispatchesJobs.php
@@ -2,8 +2,6 @@
 
 namespace Illuminate\Foundation\Bus;
 
-use Illuminate\Contracts\Bus\Dispatcher;
-
 trait DispatchesJobs
 {
     /**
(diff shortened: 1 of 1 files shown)


driesvints:
This is a big change in behavior as you're now using the job itself to dispatch and not the dispatcher.

ankurk91:
@driesvints 

The global helper method is not using the Dispatcher already.

If you dont like this big change then,
The the dispatch method available to all Controllers via DispatchesJobs Trait must be renamed to `forceDispatch()` instead.

taylorotwell:
Fine with this change but IMO we should probably just remove this trait from the base controller in Laravel 10. It isn't necessary at all.

driesvints:
Thanks @ankurk91

browner12:
For reference: https://github.com/laravel/laravel/commit/f62d260c765251427032c1a97aa9d26111ef1178

PR #45797, opened 2023-01-25
[9.x] Add createManyQuietly to HasOneOrMany

xurshudyan:
This method makes it possible to create multiple related models without raising any events to the parent model.


taylorotwell:
There have been too many of these individual PRs lately. The next PR to add these types of "quietly" methods should just add *all* remaining corresponding methods in one PR.

PR #45798, opened 2023-01-25
[9.x] Add uuidPrimary and ulidPrimary methods to Blueprint class

taron96:
This PR adds uuidPrimary and ulidPrimary methods to the `Blueprint` class. It makes it possible to create UUID or ULID field as the primary key with one method call.

Examples.
```
Schema::create('user', function (Blueprint $table) {
    $table->uuidPrimary('id');
});
```
```
Schema::create('user', function (Blueprint $table) {
    $table->ulidPrimary('id');
});
```


taylorotwell:

PR #45803, opened 2023-01-25
[9.x] Add ```get_enum_values``` function

milwad-dev:
Most of the time work with enums is hard.
If you want to take all values of one enum, you may have to do a series of operations, for example:

1.Get enum with cases method.
2.Loop the enums and move values to variable.
3. ....

But you can use this method to give all enum values, for example:
```php
<?php

namespace App;

enum ProductStatusEnum:string
{
    case ACTIVE = "active";
    case IN_PROGRESS = "in-progress";
    case INACTIVE = "inactive";
}
```
When need to take values:
```php
$enums = get_enum_values(ProductStatusEnum::class);
dd($enums);
```
The result is:

🟢 Now you can do anything :)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "^8.0.2",
+        "php": "^8.1",
         "ext-mbstring": "*",
         "ext-openssl": "*",
         "brick/math": "^0.9.3|^0.10.2|^0.11",
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,7 @@
         }
     ],
     "require": {
-        "php": "^8.1",
+        "php": "^8.0.2",
         "ext-mbstring": "*",
         "ext-openssl": "*",
         "brick/math": "^0.9.3|^0.10.2|^0.11",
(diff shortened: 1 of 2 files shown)


henzeb:
I think this is library stuff. The name is not really 'helper' material either. I may be biased. 

but you should take a look at https://packagist.org/packages/henzeb/enumhancer

milwad-dev:
@henzeb 
If we want to install a package, well, there are a lot of packages for this, but it becomes difficult on a project with a lot of packages and a lot of dependencies are created, especially if the project is big.
Also i know some packages about enum like spatie, but some operation is better use helper or something else.

henzeb:
My package works with actual enums and has no dependencies at all. And I think this one is going to blow, it doesn't add much and it changes the php requirements for no good reason and will block a lot of installations for updates from here on.

milwad-dev:
@henzeb 
I know your packages is very good for enums, but your package is good for large work.
This helpers is good for most of the time.

taylorotwell:
Yeah I think this is something you could add to a package or your own application.

Issue #45782, opened 2023-01-24
PostgreSQL schema:dump includes data


robclancy:
- Laravel Version: 8.83.27 (but was looking on 9.x branch and see the issue there)
- PHP Version: 8.0
- Database Driver & Version: postgresql 12, pg_dump 12.13

### Description:
The `artisan schema:dump` command for postgresql is dumping the data from other postgres schemas. 

In the original pull request it had `--schema-only` but after refactors it is now using `--exclude-table-data` for each table on the selected connection. https://github.com/laravel/framework/blob/8.x/src/Illuminate/Database/Schema/PostgresSchemaState.php#L24

If you have other schemas on the database the data for it won't be ignored. The original `schema:dump` works how I would expect it to work, where it does all database schemas without data (and my use case). 
But I see there is a `--database` option so this fix shouldn't be to exclude all the extra tables but to restrict the `pg_dump` to only dump the schema for the current connection.

### Steps To Reproduce:

Can add later if needed.

driesvints:
That indeed doesn't seem correct to me. Would be happy to get some help here with investigating at how we can prevent other connections from being dumped. I'm actually stumped it dumps the other schemas.

robclancy:
It's because it uses pg_dump and specifies a database which will by default dump all schemas (a postgres "schema" being a separate things makes things confusing to talk about).  

`pg_dump` has the following option...
` -n, --schema=PATTERN         dump the specified schema(s) only` 

So really the fix is just getting the schema from the config and then using that option. I don't have time to make a PR until tomorrow.

eduance:
> That indeed doesn't seem correct to me. Would be happy to get some help here with investigating at how we can prevent other connections from being dumped. I'm actually stumped it dumps the other schemas.

This is due to PostgreSQL and MySQL having different definitions of what schemas are. 

For MySQL a schema is defined as:

Conceptually, a schema is a set of interrelated database objects, such as tables, table columns, data types, indexes and so on. In MySQL a schema is synonymous with a database: you can swap out the words schema and database.

For PostgreSQL a schema is defined as:

A schema is a namespace for SQL objects, which all reside in the same database. Each SQL object must reside in exactly one schema.

More generically, the term schema is used to mean all data descriptions.

A schema can be used to organize database objects, or even allow having one database identical tables for different schemas. If you have a database named 'laravel', with schema 'symfony' and schema 'laravel', this database can have two tables named users in a PostgreSQL database.

[...]

Issue #34694, opened 2020-10-05
Exception restoring psql dump in 8.8


matejvelikonja:
- Laravel Version: 8.8.0
- PHP Version: 7.4.11
- Database Driver & Version: postgres 12.4

### Description:
After upgrading to Laravel 8.8., migration failed to restore migration squash created in previous version of Laravel.

```
Error Output:
================
pg_restore: error: input file appears to be a text format dump. Please use psql.
```

### Steps To Reproduce:
- Run `php artisan schema:dump --prune` in 8.7
- Upgrade to 8.8
- Clear database and run migrations.

I think the issue happens in this commit fcf175b246a646e81c13fb0e6c212f0b21cdfe51.

The other question is, why changing from text dump to binary dump, as binary dump is not visible in VCS?

taylorotwell:
I think you will have to re-dump, because we had some problems with previous version of schema dump.

matejvelikonja:
But what's the reasoning to move from `plain` text dump to `custom-format archive suitable for input into pg_restore`?
Is it not better to have plain text dumps in VCS instead of a binary one?

taylorotwell:
I don't really know except that it was part of this PR that fixed an issue: 

https://github.com/laravel/framework/issues/34281

https://github.com/laravel/framework/pull/34293

d-stephane:
Install laravel with a version lower than 8.8 and start the migration.
Then do "php artisan schema: dump --prune" and look in the generated file. You will find at the end this:
```
INSERT INTO public.migrations VALUES (1, '2014_10_12_000000_create_users_table', 1);
INSERT INTO public.migrations VALUES (2, '2014_10_12_100000_create_password_resets_table', 1);
INSERT INTO public.migrations VALUES (3, '2019_08_19_000000_create_failed_jobs_table', 1);
```
If you have to install this project on another server you will have to start the migration.
It is the file obtained with the dump that will be loaded and not the migration files because "--prune" option used.
Until then everything works. If you run migration -> error !
Look in the database of your new project:
```
test3=# SELECT * FROM migrations_id_seq;
 last_value | log_cnt | is_called 
------------+---------+-----------
          1 |      32 | t
```
The value of last_value is incorrect.
With a pg_dump -Fc I don't have this problem.
I do not know if it's clear


PS : sorry for my bad English.....

matejvelikonja:
@d-stephane ah, I see. I've also encountered this issue, but I have fixed it manually.

[...]

Issue #34281, opened 2020-09-11
Incomplete migration using a database schema with existing migration


d-stephane:
- Laravel Version: 8.0.4
- PHP Version: 7.3.19
- Database Driver & Version: pgsql - 12.3

### Description:
Incomplete migration using a database schema with existing migration due to a sequence issue.

### Steps To Reproduce:
Create "test" project :
```laravel new test```
Create postgresql database :
```createdb test```
Migrate :
```php artisan migrate```
Dump and prune schema :
```php artisan schema:dump --prune```
Make new migration and run migration :
```php artisan make:migration add_table_test
php artisan migrate
- Migrating: 2020_09_11_064507_add_table_test
- Migrated:  2020_09_11_064507_add_table_test (14.05ms)
```
---
Create "test2" project :
```laravel new test2```
Create postgresql database :
```createdb test2```
Remove default migrations and copy schema directory from test/database to test2/database  
Copy test/database/migrations/2020_09_11_064507_add_table_test.php to test2/database/migrations  
Run migration :
```php artisan migrate```
Error :
```
Migration table created successfully.
Loading stored database schema: /www/a/test2/database/schema/pgsql-schema.sql
Loaded stored database schema. (118.41ms)
Migrating: 2020_09_11_064507_add_table_test

   Illuminate\Database\QueryException 

  SQLSTATE[23505]: Unique violation: 7 ERREUR:  la valeur d'une clé dupliquée rompt la contrainte unique « migrations_pkey »
DETAIL:  La clé « (id)=(1) » existe déjà. (SQL: insert into "migrations" ("migration", "batch") values (2020_09_11_064507_add_table_test, 2))

  at vendor/laravel/framework/src/Illuminate/Database/Connection.php:671
    667▕         // If an exception occurs when attempting to run a query, we'll format the error
    668▕         // message to include the bindings with SQL, which will make this exception a
    669▕         // lot more helpful to the developer instead of just the database's errors.
    670▕         catch (Exception $e) {
  ➜ 671▕             throw new QueryException(
    672▕                 $query, $this->prepareBindings($bindings), $e
    673▕             );
    674▕         }
    675▕ 

      +28 vendor frames 
  29  artisan:37
      Illuminate\Foundation\Console\Kernel::handle(Object(Symfony\Component\Console\Input\ArgvInput), Object(Symfony\Component\Console\Output\ConsoleOutput))
```

d-stephane:
[...]

PR #45805, opened 2023-01-26
[9.x] - 45782: Solve data to be dumped for separate schemes

eduance:
This issue refers to:
https://github.com/laravel/framework/issues/45782

---

For MySQL a schema is defined as:

Conceptually, a schema is a set of interrelated database objects, such as tables, table columns, data types, indexes and so on. In MySQL a schema is synonymous with a database: you can swap out the words schema and database.

For PostgreSQL a schema is defined as:

A schema is a namespace for SQL objects, which all reside in the same database. Each SQL object must reside in exactly one schema.

More generically, the term schema is used to mean all data descriptions.

A schema can be used to organize database objects, or even allow having one database identical tables for different schemas. If you have a database named 'laravel', with schema 'symfony' and schema 'laravel', this database can have two tables named users in a PostgreSQL database.

In a MySQL database, you would simply have a database named laravel and symfony.

PostgreSQL has a default public schema, but for consistency sake I would recommend exporting all schemas for one database, so the artisan dump command simply dumps out all tables.

What we would like to achieve is that we want to get all schemas, without any data, excluding the migrations table.

This would look like something like:

pg_dump [connection_options] -s -T 'migrations' [db_name]

**Why not the current approach?**

The current approach will dump all database objects, excluding the table data defined in the public scheme. As we have no way of knowing what schemes a user might be using, and we shouldn't want to know for the dump command, it could be a more general solution to simply dump all sql objects, exclude data, for all schemes.

Another approach would be to only dump out all data under the default public scheme, but this would make a dump useless for companies that have a logical grouping the database.

### DIFF
--- a/src/Illuminate/Database/Console/DumpCommand.php
+++ b/src/Illuminate/Database/Console/DumpCommand.php
@@ -94,7 +94,7 @@ class DumpCommand extends Command
      */
     protected function path(Connection $connection)
     {
-        return tap($this->option('path') ?: database_path('schema/'.$connection->getName().'-schema.dump'), function ($path) {
+        return tap($this->option('path') ?: database_path('schema/'.$connection->getName().'-schema.sql'), function ($path) {
             (new Filesystem)->ensureDirectoryExists(dirname($path));
         });
     }
--- a/src/Illuminate/Database/Schema/PostgresSchemaState.php
+++ b/src/Illuminate/Database/Schema/PostgresSchemaState.php
@@ -15,11 +15,16 @@ class PostgresSchemaState extends SchemaState
      */
     public function dump(Connection $connection, $path)
     {
-        $this->makeProcess(
-            $this->baseDumpCommand().' --file="${:LARAVEL_LOAD_PATH}" '
-        )->mustRun($this->output, array_merge($this->baseVariables($this->connection->getConfig()), [
-            'LARAVEL_LOAD_PATH' => $path,
-        ]));
+        $commands = collect([
+            $this->baseDumpCommand() . ' --schema-only > ' . $path,
+            $this->baseDumpCommand() . ' -t '. $this->migrationTable .' --data-only >> ' . $path,
+        ]);
+
+        $commands->map(function ($command, $path) {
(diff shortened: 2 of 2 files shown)


robclancy:
This fixes the data issue but re-introduces this issue (which created the data issue lol) https://github.com/laravel/framework/pull/35018.

eduance:
@robclancy 

Thanks for the quick replies. It looks like Taylor is trying to fix something else in that PR, by removing --schema-only and replacing it with excluding the table data. I just checked the issue that he was initially fixing: https://github.com/laravel/framework/issues/34281 and I'm not great at French, but after having built too many Laravel projects, that error is telling us that he already has a migrations table and he is adding another migrations table, so therefore removing the migrations table from the dump seems to fix more or less his issue. 

d-stephane explained the problem in more detail here:

https://github.com/laravel/framework/issues/34694#issuecomment-704363156

This issue seemed to be fixed after renaming to custom binary dumps, but that isn't true. What actually solves his issue was to exclude the migration data table which is by default always available in the public scheme, so to really fix his issue, we don't need a lot of black magic, we just needed the -T parameter which excluded the migrations table, and combine it with --schema-only.

I'll be up in case you have any additions, now getting some good ol' coffee.

robclancy:
The issue is that the original way it worked is to dump all the SQL and then append the migration data as SQL. Then someone made a change to use the custom postgres binary format so that data was no longer appended. The fix was then to instead exclude the data of all tables except the migration table, this only works on the current schema though so data is added for all other schemas.  

The migration table and its data is still needed.

eduance:
@robclancy Thanks for the explanation.

As Taylor wanted to get rid of the binary dumps as he stated back in 2020, I looked for a solution that was able to achieve the following:

- Replace binary dumps with SQL dumps.
- Get the schema from all tables, exclude data.
- Take migration data into consideration and include it aswell.

Another commit, this could be an initial solution to both your problem as d-stephanes problem.

eduance:
@taylorotwell / @driesvints Can you review this PR? You introduced this functionality back in 2020, this should fix that all schemes should be taken into consideration + this should get rid of binary dumps.

taylorotwell:
Thanks

PR #45815, opened 2023-01-26
[9.x] feat: add support for database custom init commands

Pr3d4dor:
We currently ran into the situation where we need to set the max_execution_time for the current session in our MySql Database.

We implemented it (using a Provider with DB::unprepared and didn't like it) and thought maybe it makes sense to bring that into laravel core, as laravel currently lacks that possibility.

We currently configure it in database.php
```
return [
    'connections' => [
          'default' => [
                ...
               'host' => env('DB_HOST'),
               ...
               'init' => ['set session max_execution_time = 20000'],
              ....
            ]
      ]
];
```

We use the MySql connector and so I am following the same concept for the other connectors.

If approved, I can make the appropriate PR for the docs too.

### DIFF
--- a/src/Illuminate/Database/Concerns/ConfiguresCustomInitCommands.php
+++ b/src/Illuminate/Database/Concerns/ConfiguresCustomInitCommands.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Illuminate\Database\Concerns;
+
+trait ConfiguresCustomInitCommands
+{
+    /**
+     * Configure custom database init commands.
+     *
+     * @param  \PDO  $connection
+     * @param  array  $config
+     * @return void
+     */
+    protected function configureCustomInitCommands($connection, array $config)
+    {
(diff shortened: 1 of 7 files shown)


brunozorzetto:
it would be a good solution instead of running raw sql before any database operation, nice ideia

julianopaulos:
that would be a great improvement!

thiaguera00:
good solution

felipecousin:
really helpful!

rochacaio:
very good, this would solve some implementation problems I already had in a project!

taylorotwell:
Added new ConnectionEstablished event you can listen for in order to do whatever you want on database connection initiation.

https://github.com/laravel/framework/commit/f850d99c50d173189ece2bb37b6c7ddcb456f1f9

Pr3d4dor:
@taylorotwell Thanks, we will use the event to change database initiation.

PR #45816, opened 2023-01-26
Default content type to JSON

jasonmccreary:
This addresses a paper cut of having to set the content type when sending raw JSON bodies via the `Http` client. Since the `Http` client sends data as JSON by default, it'd find it more intuitive if the `withBody` did so as well.

**Before**
```php
Http::withBody('{"foo": "bar"}', 'application/json')->post('https://example.com');
```

**After**
```php
Http::withBody('{"foo": "bar"}')->post('https://example.com');
```

I'm targeting Laravel 10 since this changes the signature of a `public` method.

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -250,7 +250,7 @@ class PendingRequest
      * @param  string  $contentType
      * @return $this
      */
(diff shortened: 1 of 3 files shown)


jasonmccreary:
Failure seems like a spurious floating-point issue. Maybe someone on the Laravel Team could retry the job?

PR #45820, opened 2023-01-26
[9.x] remove `HandlesAuthorization` trait from stubs

browner12:
the function in these traits are for the most part proxies to static methods on the `Response` class.

Even the [documentation](https://github.com/laravel/docs/commit/0682ca014e5e1768066ef93b4269a83d82978a70) doesn't use them, and instead goes directly to the `Response` class, so I think it would be best to remove them from the stubs.

I would say we should deprecate the trait as well, to encourage the documented use, but I'll hold on that.


taylorotwell:
Going to do this on 10.x instead.

PR #45824, opened 2023-01-27
[10.x] Save the types in enum and then give an error if you enter the wrong type

DevNull-IR:
Save the types in enum and then give an error if you enter the wrong type


taylorotwell:

DevNull-IR:
@taylorotwell 

Thankyou.

It has happened so far that there was a mistake in writing the name of the method, that I did not pay attention to the name of the method during debugging, and because of this simple thing, I got into trouble.

Therefore, it may be a good possibility to avoid these problems and only accept the `get`, `post`, `put`, `delete`, `options` and `patch` values.

PR #45826, opened 2023-01-27
[9.x] Fluent has-x-through definitions.

kayrunm:
The current `hasOneThrough` and `hasManyThrough` method definitions can be confusing, especially when having to modify the local & foreign keys, and—at least personally—has me reaching for the docs pretty much every time I need to use them.

For (an admittedly contrived) example, let's say I have `photos` and `albums` tables, and in my app I have decided to use `photo_id` and `album_id` as the respective primary keys, I would have to achieve that like so:
```php
class User
{
    public function photos(): HasManyThrough
    {
        return $this->hasManyThrough(Photo::class, Album::class, 'photo_id', 'album_id', 'photo_id', 'album_id');
    }

    public function profilePhoto(): HasOneThrough
    {
        return $this
            ->hasOneThrough(Photo::class, Album::class, 'photo_id', 'album_id', 'photo_id', 'album_id')
            ->where('is_profile', true);
    }
}
```

This isn't super clear on what we're achieving or what any of the final 4 arguments are. Granted, some IDEs might add inlays, or I could use named arguments to clear this up, but the definition doesn't feel obvious from a glance.

I am proposing that we add a `through` method onto the `HasOne` and `HasMany` relation classes. This would allow a more fluent definition of the relationship and is a lot easier to understand what is happening:
```php
class User
{
    public function photos(): HasManyThrough
    {
        return $this
            ->hasMany(Photo::class, 'photo_id', 'photo_id')
            ->through(Album::class, 'album_id', 'album_id');
    }

    public function profilePhoto(): HasOneThrough
    {
        return $this
            ->hasOne(Photo::class, 'photo_id', 'photo_id')
            ->through(Album::class, 'album_id', 'album_id')
            ->where('is_profile', true);
    }
}
```

--

Edit:
* I have only added a couple of tests so far, as I wanted to see what appetite there was for this — if we wanted I can add more tests.
* The old methods (`hasOneThrough`, `hasManyThrough`) remain intact. This is simply an alternative

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/HasMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/HasMany.php
@@ -3,6 +3,7 @@
 namespace Illuminate\Database\Eloquent\Relations;
 
 use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Support\Str;
 
 class HasMany extends HasOneOrMany
 {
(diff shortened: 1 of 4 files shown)


kayrunm:
Just realised my implementation might be slightly off, so I'll look into that. More tests would probably help confirm it.

ankurk91:
So the old way of defining relationship would still work?

kayrunm:
> So the old way of defining relationship would still work?

Yup, nothing has changed in that respect. The old way would still be available, so no breaking change. Just a more fluent (and in my opinion, readable) alternative of defining them.

taylorotwell:
I actually find the specific of keys in this PR *more* confusing than just using the original method. For example, here are both side by side:

CleanShot 2023-01-31 at 15 43 27@2x

Note that in your PR you have to specify `mechanic_id` with the call to `hasOne(Owner::class)` - but `mechanic_id` is not on the `owners` table at all - it's on the `cars` table. Likewise, `car_id` isn't on the `Car` class, it's on the `Owner` class. Everything feels reversed.

In the "non-fluent" definition (current state of Laravel), you specify the keys in a logical order... joining table foreign key, final table foreign key, local key to be used on joining table, local key to be used on final table. It goes from joining to final both times.

timacdonald:
I've proposed a slight alternative approach to this idea: https://github.com/laravel/framework/pull/45894

PR #45828, opened 2023-01-27
[9.x] Add mergeWithValidated method into Http Request

milwad-dev:
Sometimes need to add some input after validation, like this:
```php
  public function rules()
  {
      return [
          'email' => 'required|email'
      ];
  }

  public function prepareForValidation()
  {
      $this->merge([
          'code' => random_int(111111, 999999)
      ]);
  }
```
But the problem is that when you want to get this information in the controller.
Maybe you think you can write ```$request->all()``` to take all values like this:
```php
    public function store(ProductRequest $request)
    {
        return $request->all();
    }
```
But if the developer wants to use the ```validated``` method, the requested information will not be sent and he will be required to use the ```all``` method, and somehow it will cause problems!
Now we can use the method to combine external and internal information and access it with the ```validated``` method:
```php
Form Request:
    public function rules()
    {
        return [
            'email' => 'required|email'
        ];
    }

    public function prepareForValidation()
    {
        $this->mergeWithValidated([
            'code' => random_int(111111, 999999)
        ]);
    }
    
Controller:
    public function store(ProductRequest $request)
    {
        return $request->validated();
    }
```


taylorotwell:
I just don't find this kind of stuff necessary to be honest. Just merge it with the validated data in your controller?

milwad-dev:
@taylorotwell 
If the developer does not want to use the all method, but adds a series of values, what should he do?

milwad-dev:
@taylorotwell 
In my opinion, this is groundless strictness :)

PR #45829, opened 2023-01-27
[9.x] Add forceDeleting event

xurshudyan:
Before
``` php
    public static function booted()
    {
        static::deleting(function (User $user) {
            if ($user->isForceDeleting()) {
                //
            }
        });
    }
```
After
``` php
    public static function booted()
    {
        static::forceDeleting(function (User $user) {
            // do something
        });
    }
```


jmarcher:
This is a huge breaking change and for me the way it is now is enough.

PR #45831, opened 2023-01-27
[10.x] remove handles authorization

browner12:
Resubmitting to "master". Wasn't sure if you were gonna do this or wanted me to, so I figured I'd just help out.

the function in these traits are for the most part proxies to static methods on the `Response` class.

Even the [documentation](https://github.com/laravel/docs/commit/0682ca014e5e1768066ef93b4269a83d82978a70) doesn't use them, and instead goes directly to the `Response` class, so I think it would be best to remove them from the stubs.


taylorotwell:
Hey @browner12 - I already removed them right after your PR. Thanks! 👍

PR #45832, opened 2023-01-27
[10.x] Lockable Model Route Bindings

bert-w:
### Problem Explanation
This feature allows developers to apply a lock on implicitly resolved models. Consider the following example:

```php
Route::post('/user/{user}', function (ImplicitBindingModel $user) {
    if ($user->paid) {
        throw ValidationException::withMessages([...]);
    }

    // Pay user ...
    $user->update(['paid' => true]);

    return $user;
});
```

Since the controller action requires a read to `$user->paid` (some property indicating the user has been paid which must only happen once), there is a potential exploit by calling this function twice in quick succession. Since both calls happen at the same time, they retrieve `$user->paid === false` for both occasions, causing the code to be executed in both occasions as well. This requires a transaction in order to run it correctly.

### Solution: Route `withLocks()`
By using the `withLocks()` option on your route (requiring the `SubstituteBindings` middleware), you can now make sure that models are retrieved with a read-lock using Eloquent's `lockForUpdate()`, using:

```php
// New example
Route::post('/user/{user}', function (ImplicitBindingModel $user) {
    // Pay user etc.
})->middleware(['web'])->withLocks();
```

The controller action is now automatically wrapped in a transaction, with read-locks applied to all implicitly bound models. This also works with `withTrashed()` (this enables retrieval of soft-deleted models) so `->withLocks()->withTrashed()` will work as expected too.

### Old situation: `DB::transaction(...)` and manually retrieving models
An explicit way to do this currently (showing the current limitation) is:

```php
// Old example
Route::post('/user/{user}', function ($userId)
{
    return DB::transaction(function () {
        $user = User::query()->lockForUpdate()->findOrFail($userId);

        if ($user->paid) {
            throw ValidationException::withMessages([...]);
        }

        // Pay user...

    });
});
```
Implicit model bounding cannot take place in the old situation, since the model would have been retrieved before the transaction commenced.

### DIFF
--- a/src/Illuminate/Contracts/Routing/Registrar.php
+++ b/src/Illuminate/Contracts/Routing/Registrar.php
@@ -102,4 +102,12 @@ interface Registrar
      * @return void
      */
     public function substituteImplicitBindings($route);
+
+    /**
+     * Substitute the route bindings that were passed as a closure to allow for delayed execution.
+     *
+     * @param  \Illuminate\Routing\Route  $route
+     * @return void
+     */
+    public function substituteClosureBindings($route);
 }
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -119,6 +119,20 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     protected $escapeWhenCastingToString = false;
 
     /**
+     * Determines if route bindings should resolve trashed models as well.
+     *
+     * @var bool
+     */
+    public $resolveWithTrashed;
+
+    /**
+     * Determines if route bindings should resolve with a row-level lock.
+     *
+     * @var bool
+     */
+    public $resolveWithLock;
+
+    /**
      * The connection resolver instance.
      *
      * @var \Illuminate\Database\ConnectionResolverInterface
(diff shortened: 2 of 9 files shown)


kafkiansky:
A very, very bad suggestion. You shouldn't wrap a controller call in a transaction, because the controller may have network calls (http, message brokers) whose delays can lead to longer transaction times, the number of which is usually limited. Increasing the number of concurrent transactions will degrade the database. And you should certainly not use locks in such scenarios.

bert-w:
Yes it is obvious that this approach does not work if your controller actions are bound to take seconds. What would you suggest in a scenario were all actions are local (and fast) ? I guess using a cache lock has some merit but it seems much more logical to use database level tools like row level locks.

Verzonden vanaf Outlook voor Android<https://aka.ms/AAb9ysg>
________________________________

kafkiansky:
> Yes it is obvious that this approach does not work if your controller actions are bound to take seconds. What would you suggest in a scenario were all actions are local (and fast) ? I guess using a cache lock has some merit but it seems much more logical to use database level tools like row level locks. Verzonden vanaf Outlook voor Android<https://aka.ms/AAb9ysg>
> […](#)

I suggest using small explicit transactions where necessary. This is nothing new, though. You see a problem where there is no problem.

ankurk91:
Too much magic.
Too much coupling.

bert-w:
> Too much magic. Too much coupling.

Where is the magic exactly? I can agree with the coupling argument in that this uses database-level locks instead of other options like cache locks. But before attempting to code such an abstraction I'd like to know if it even is a desired feature to have.

PR #45841, opened 2023-01-28
[10.x] Change custom function with native

xurshudyan:
Instead of a custom function, use native


andiLeong:
for the 10.x its already use array_is_list
you want to merge to 9.x , array_is_list is introduced since php8.1 , but 9.x still support 8.0
correct me if I am wrong
thank you

:zero:

xurshudyan:
You are correct

PR #45843, opened 2023-01-28
[10.x] Validate new rules if previous attribute validated

saeedhosseiinii:
Sometimes we need to add a validation rule if the validation of other attributes is passed.

For example, when I create a custom rule that requires multiple request parameters and requires these parameters to be validated.

With this feature we can:

```php
 $validator->validateIfPasses(['user_id', 'post_id'], [
      'post_type' => [new CustomRule($request->user_id, $request->post_id)]
 ]);
```

### DIFF
--- a/src/Illuminate/Contracts/Validation/Validator.php
+++ b/src/Illuminate/Contracts/Validation/Validator.php
@@ -49,6 +49,17 @@ interface Validator extends MessageProvider
     public function sometimes($attribute, $rules, callable $callback);
 
     /**
+     * Validate the new rules if the previous attribute have been validated.
+     *
+     * @param  string|array  $attribute
+     * @param  array  $rules
+     * @param  array  $messages
+     * @param  array  $customAttributes
+     * @return $this
+     */
+    public function validateIfPasses($attributes, array $rules, array $messages = [], array $customAttributes = []);
+
+    /**
      * Add an after validation callback.
      *
      * @param  callable|string  $callback
--- a/src/Illuminate/Validation/Validator.php
+++ b/src/Illuminate/Validation/Validator.php
@@ -1201,6 +1201,29 @@ class Validator implements ValidatorContract
     }
 
     /**
+     * Validate the new rules if the previous attribute have been validated.
+     *
+     * @param  string|array  $attribute
+     * @param  array  $rules
+     * @param  array  $messages
+     * @param  array  $customAttributes
+     * @return $this
+     */
+    public function validateIfPasses($attributes, array $rules, array $messages = [], array $customAttributes = [])
+    {
+        $attributes = (array)$attributes;
+
+        $this->after(function ($validator) use ($attributes, $rules, $messages, $customAttributes) {
+            if ($validator->messages()->hasAny($attributes)) {
+                return;
+            }
+            $validator->messages()->merge((new self($this->translator, $this->data, $rules, $messages, $customAttributes))->messages());
+        });
+
+        return $this;
+    }
+
+    /**
      * Instruct the validator to stop validating after the first rule failure.
      *
      * @param  bool  $stopOnFirstFailure


henzeb:
How does this anything different than the DataAwareRule as explained here
https://laravel.com/docs/9.x/validation#using-rule-objects under "accessing additional data"?

saeedhosseiinii:
> How does this anything different than the DataAwareRule as explained here https://laravel.com/docs/9.x/validation#using-rule-objects under "accessing additional data"?

Suppose we have the following rules:

```php
 public function rules()
    {
        return [
            'user_id' => 'required|exist:users,id',
            'post_id' => 'required|posts,id',
            'post_type' => [new CustomRule]
        ];
    }
```

How to make sure user_id and post_id are valid in custom rule?

henzeb:
You could also add the ValidatorAwareRule. This would satisfy your needs:

```php
class CustomRule implements InvokableRule, ValidatorAwareRule, DataAwareRule
{

    protected bool $canValidate = false;
    protected ?int $userId = null;
    protected ?int $postId = null;


    {
        if ($this->canValidate) {
            // do something
            return;
        }
        $fail('Failed because required fields already invalid');
    }

    /**
     * Set the data under validation.
     *
     * @param array $data
     * @return $this
     */
    public function setData($data): void
    {
        $this->userId = $data['user_id'] ?? null;
        $this->postId = $data['post_id'] ?? null;
    }

    /**
     * Set the current validator.
     *
     * @param \Illuminate\Validation\Validator $validator
     * @return $this
     */
    public function setValidator($validator)
    {
        $messageBag = $validator->getMessageBag();

        $this->canValidate = !$messageBag->hasAny('user_id', 'post_id');

        return $this;
    }
}
```

src/Illuminate/Contracts/Validation/Validator.php
```diff
@@ -48,6 +48,17 @@ public function failed();
...
+    /**
+     * Validate the new rules if the previous attribute have been validated.
+     *
+     * @param  string|array  $attribute
+     * @param  array  $rules
+     * @param  array  $messages
+     * @param  array  $customAttributes
+     * @return $this
+     */
+    public function validateIfPasses($attributes, array $rules, array $messages = [], array $customAttributes = []);
```
driesvints, line 60: We cannot add methods to contracts on patch releases.

saeedhosseiinii:
> You could also add the ValidatorAwareRule. This would satisfy your needs:
> 
> ```
> class CustomRule implements InvokableRule, ValidatorAwareRule, DataAwareRule
> {
> 
>     protected bool $canValidate = false;
>     protected ?int $userId = null;
>     protected ?int $postId = null;
> 
> 
>     public function __invoke($attribute, $value, $fail)
>     {
>         if ($this->canValidate) {
>             // do something
>             return;
>         }
>         $fail('Failed because required fields already invalid');
>     }
> 
>     /**
>      * Set the data under validation.
>      *
>      * @param array $data
>      * @return $this
>      */
>     public function setData($data): void
>     {
>         $this->userId = $data['user_id'] ?? null;
>         $this->postId = $data['post_id'] ?? null;
>     }
> 
>     /**
>      * Set the current validator.
>      *
>      * @param \Illuminate\Validation\Validator $validator
>      * @return $this
>      */
>     public function setValidator($validator)
>     {
>         $messageBag = $validator->getMessageBag();
> 
>         $this->canValidate = !$messageBag->hasAny('user_id', 'post_id');
> 
>         return $this;
>     }
> }
> ```

But with this method it becomes much easier!

src/Illuminate/Contracts/Validation/Validator.php
saeedhosseiinii: I changed the base branch to 10.x

saeedhosseiinii: I changed the base branch to 10.x

driesvints:
For some reason tests aren't running on this PR. Please re-send this as a new PR, thanks.

PR #45846, opened 2023-01-28
Fix some incorrect method overrides 

joelharkes:
Fixes some code problems:
* Return type missing (ArraySubSet)
* Stricter types on override methods: FilterEmailValidation, FileSystemAdapter => 
* Missing parameters: Builder, HasManyThrough

For stricter types i could have fixed it by adding types to parent classes but this could lead to breaking changes.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -1178,15 +1178,16 @@ class Builder implements BuilderContract
     /**
      * Delete records from the database.
      *
-     * @return mixed
+     * @param  mixed  $id
+     * @return mixed|int
      */
-    public function delete()
+    public function delete($id = null)
     {
         if (isset($this->onDelete)) {
             return call_user_func($this->onDelete, $this);
         }
 
-        return $this->toBase()->delete();
+        return $this->toBase()->delete($id);
     }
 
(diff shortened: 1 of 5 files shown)


src/Illuminate/Database/Eloquent/Builder.php
```diff
@@ -1178,15 +1178,16 @@ protected function addUpdatedAtToUpsertColumns(array $update)
     /**
      * Delete records from the database.
      *
-     * @return mixed
+     * @param  mixed  $id
+     * @return mixed|int
      */
-    public function delete()
+    public function delete($id = null)
```
driesvints, line 1184: This is a breaking change on the method signature unfortunately.

driesvints:
There are several breaking changes to method signatures in this PR. Please attempt 10.x, thanks.

src/Illuminate/Database/Eloquent/Builder.php
joelharkes: @driesvints  how is this breaking? in case someone overrides the builder class?
But the parent class already had this method?

driesvints: Yes, if someone overwrite the builder class' `delete` method their code will break.

PR #45848, opened 2023-01-28
Add more suggestged php extensions

joelharkes:
Added some suggested php extensions with explanations where the are used for.


driesvints:
Heya, you can send this to 9.x if you wish. Can you also add these to their respective component `composer.json` files that they belong to? Thanks!

PR #45850, opened 2023-01-28
Bring back consistency in FirstOrNew/Create

joelharkes:
Was looking into this code and saw a lot of different ways of it being written in the different classes.

I tried to make it as concise as possible using modern php nullable operator makes the waterfall code easy readable.

extra fix: and added the missing $values parameter in HasManyThrough.


driesvints:
Tests seem to fail here. Please resend this with passing tests, thanks.

PR #45854, opened 2023-01-29
[9.x] Add whenHas and whenHasAny() methods to Collections.

xurshudyan:
This PR adds support for whenHas and whenHasAny() method to collections. If I did something wrong let me know

``` php
    collect(['foo', 'bar', 'baz'])->whenHas('bar', function ($collection) {
       // do something 
    });

    collect(['foo', 'bar', 'baz'])->whenHasAny(['foo', 'bar'], function ($collection) {
        // do something 
    });
```

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -595,6 +595,32 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
+     * Execute the given callback if an item exists in the collection by key.
+     *
+     * @param  TKey|array<array-key, TKey>  $key
+     * @param  callable  $callback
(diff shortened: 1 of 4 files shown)


src/Illuminate/Collections/Collection.php
```diff
@@ -594,6 +594,32 @@ public function hasAny($key)
         return false;
     }
 
+    /**
+     * Execute the given callback if an item exists in the collection by key.
+     *
+     * @param  TKey|array<array-key, TKey>  $key
+     * @param  callable  $callback
+     * @param  callable|null  $default
+     * @return Collection
```
driesvints, line 603: Please use FQN everywhere in your new DocBlocks.

driesvints (changes requested):

PR #45869, opened 2023-01-30
[9.x] Add skip cancelled middleware

morrislaptop:
This PR adds a small Job middleware which will skip `Batchable` jobs if the associated batch is cancelled.

This basically removes the need for an initial if statement from the handle method:

```php
/**
 * Execute the job.
 *
 * @return void
 */
public function handle()
{
    if ($this->batch()->cancelled()) {
        return;
    }
 
    // Continue processing...
}
```

To this:

```php
public function middleware()
{
    return [new SkipCancelled];
}
```

Which aligns with the purpose of middleware from the [docs](https://laravel.com/docs/9.x/queues#job-middleware) 

> Job middleware allow you to wrap custom logic around the execution of queued jobs, reducing boilerplate in the jobs themselves.

### DIFF
--- a/src/Illuminate/Queue/Middleware/SkipCancelled.php
+++ b/src/Illuminate/Queue/Middleware/SkipCancelled.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Illuminate\Queue\Middleware;
+
+class SkipCancelled
+{
+    /**
+     * Process the job.
+     *
+     * @param  mixed  $job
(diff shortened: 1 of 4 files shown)


driesvints:
Thanks! @morrislaptop can you also send a PR to the docs?

morrislaptop:
@driesvints https://github.com/laravel/docs/pull/8495

PR #45871, opened 2023-01-30
[9.x] Restart syscalls for SIGALRM when worker times out a job

### LINKED ISSUE #45867: PDOException gets thrown whenever a SIGALRM happens during a query
- Laravel Version: 9.47.0
- PHP Version: 8.1
- Database Driver & Version: MSSql 18

### Description:

Whenever a SIGALRM happens during an execution of a query a PDOException gets thrown from the MSSql server.

```
Illuminate\Database\QueryException in App\Jobs\Job
SQLSTATE[08S01]: [Microsoft][ODBC Driver 17 for SQL Server]TCP Provider: Error code 0x2714 (SQL: select * from [table] order by [id] asc offset 1000 rows fetch next 1000 rows only)
```

NOTE: I am in no way an expert in Process Control but I've been digging about this error for days, and found relevant info. [This issue](https://github.com/microsoft/msphpsql/issues/885) describes why this happens and how, specifically in [this comment](https://github.com/microsoft/msphpsql/issues/885) a `msphpsql` maintainer explains how to resolve this issue. Another relevant change is [this commit](https://github.com/php/php-src/commit/e98e1f92c98b7c8910c55835d8f67d0d9230cc8b) by Nikita Popov in PHP core.

Apparently, to solve this we need to add the third argument (`restart_syscalls`) as true, here:

https://github.com/laravel/framework/blob/b75b5740dd0ec909206f94d4c2d4282632ac0a33/src/Illuminate/Queue/Worker.php#L211-L227

I have no idea what are the ramifications of this change, that's why I'm writing an issue and not opening a PR. Feel free to educate me and/or give me a clue on how to solve this issue for me.

### Steps To Reproduce:

Have a job time out while running a query on the MSSql server.

amiranagram:
As mentioned in #45867 whenever a `SIGALRM` happens during an execution of a query a PDOException gets thrown from the MSSql server. In simpler terms, whenever `SIGALRM` arrives the MSSQL connection gets broken resulting in the query not being to execute itself, hence throwing the error.

Although `restart_syscalls` argument is true by default in `pcntl_signal` this will not work with `SIGALRM` signal due to the backwards compatibility left by Nikita Popov in [this commit](https://github.com/php/php-src/commit/e98e1f92c98b7c8910c55835d8f67d0d9230cc8b).

### DIFF
--- a/src/Illuminate/Queue/Worker.php
+++ b/src/Illuminate/Queue/Worker.php
@@ -224,7 +224,7 @@ class Worker
             }
 
             $this->kill(static::EXIT_ERROR, $options);
-        });
+        }, true);
 
         pcntl_alarm(
             max($this->timeoutForJob($job, $options), 0)


driesvints:
Please add a thorough description at what this fixes for the person that will review this PR and not link to an issue (see the PR template).

amiranagram:
Updated @driesvints

PR #45873, opened 2023-01-30
[10.x] Use unsigned closures for route caching

nunomaduro:
This pull request uses unsigned routes - for Laravel 10 - as discussed https://github.com/laravel/serializable-closure/pull/62. We need a release of https://github.com/laravel/serializable-closure/pull/64, to bump `composer.json`.

### DIFF
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -997,6 +997,7 @@ class Route
         return is_string($missing) &&
             Str::startsWith($missing, [
                 'O:47:"Laravel\\SerializableClosure\\SerializableClosure',
+                'O:55:"Laravel\\SerializableClosure\\UnsignedSerializableClosure',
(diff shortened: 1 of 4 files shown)


src/Illuminate/Routing/Route.php
```diff
@@ -997,6 +997,7 @@ public function getMissing()
         return is_string($missing) &&
             Str::startsWith($missing, [
                 'O:47:"Laravel\\SerializableClosure\\SerializableClosure',
+                'O:55:"Laravel\\SerializableClosure\\UnsignedSerializableClosure',
```
olivernybroe, line 1000: This seems duplicated in `RouteAction`. Could prob. be consolidated

nunomaduro: Indeed. Going to keep it like this for now.

PR #45876, opened 2023-01-30
Fail jobs on OOM errors

khepin:
Right now, creating a dummy job that just consumes memory, the job fails with an OOM error, but it stays on the queue forever. It's never marked as failed and never released.

This adds a shutdown handler that will mark the job as failed on OOM errors.

Without this, jobs that fail because of high memory usage all stay on the queue and accumulate there. If enough of them have accumulated, the workers keep spinning on jobs that can never go through preventing any other jobs from being worked on.

This now works similarly to what would happen on job timeouts.

This isn't the final PR, but I wanted to gauge interest in adding this before making a more complete PR.

### DIFF
--- a/src/Illuminate/Queue/Worker.php
+++ b/src/Illuminate/Queue/Worker.php
@@ -141,10 +141,10 @@ class Worker
 
         register_shutdown_function(function () use (&$job) {
             $e = error_get_last();
-            if (!$job instanceof Job) {
+            if (! $job instanceof Job) {
                 return;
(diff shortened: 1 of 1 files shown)


ankurk91:
We should add some comments around the new code, it is handling a special case.

taylorotwell:
I think we would need tests for this, and I don't think we should register a shutdown function from the Worker class - only from the `queue:work` Artisan command potentially.

PR #45878, opened 2023-01-30
Use arrow function

michaelnabil230:


taylorotwell:
Honestly I hate merging such big formatting PRs like this... if anything it should be sent to Laravel 10.x. Thanks.

PR #45879, opened 2023-01-31
model array set key support relation key

duncanxia97:
~~~php
// before
$model = TestModel::with('admin')->find(1);
$model['admin']['name'] = "test_".$model['admin']['name']
echo $model['admin']['name'];
// echo: adminName

// after
$model = TestModel::with('admin')->find(1);
$model['admin']['name'] = "test_".$model['admin']['name']
echo $model['admin']['name'];
// echo: test_adminName
~~~


ankurk91:
It can be breaking for many, target 10.x?
Can you include some test.

taylorotwell:

PR #45883, opened 2023-01-31
Adds more `Regular Expression` methods 🔎

mahmoudmohamedramadan:
This PR adds more useful `Regular Expression Route` methods that save many of lines code. Let's see the next scenario for more illustration 👀.

## BEFORE accepting the PR

```PHP
/**
 * Display the given integer value.
 *
 * @param  int  $id
 * @return int
 */
public function show($id)
{
    // Checking if the length of `$id` is 1
    if (strlen($id) != 1) {
        abort(404);
    }

    // Checking if the `$id` value equals 1
    if ($id != 1) {
        abort(404);
    }

    return $id;
}
```

Let's see the magic of the new two methods 🎉

## AFTER accepting the PR

```PHP
Route::get('users/{user}', function ($user) {
    return $user;
})
    ->whereLength('user', 1)
    ->whereEquals('user', 1);
```

### DIFF
--- a/src/Illuminate/Routing/CreatesRegularExpressionRouteConstraints.php
+++ b/src/Illuminate/Routing/CreatesRegularExpressionRouteConstraints.php
@@ -40,6 +40,30 @@ trait CreatesRegularExpressionRouteConstraints
     }
 
     /**
+     * Specify that the given route parameters must equals the given length.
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #45884, opened 2023-01-31
[9.x] Ability to attach multiple in-memory data in MailMessage

rudashi:
If we have an unknown number of elements to be attached, we have to do a foreach loop, which looks nasty.

The `attachManyData()` method is much cleaner.

```php
    return (new MailMessage())
           ->attachManyData([
                [
                    $first_file,
                    'filename.pdf',
                    ['mime' => 'application/pdf']
                ],
                [
                    $second_file,
                    'filename.jpeg',
                    ['mime' => 'image/jpg']
                ],
            ]);
````

### DIFF
--- a/src/Illuminate/Notifications/Messages/MailMessage.php
+++ b/src/Illuminate/Notifications/Messages/MailMessage.php
@@ -278,6 +278,21 @@ class MailMessage extends SimpleMessage implements Renderable
     }
 
     /**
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by taylorotwell** [comment]
I would probably just do something like this:

```php
return (new MailMessage)
        ->when(is_array($attachments), fn () => 
            collect($attachments)->each(fn ($a) => $this->attachData($a))
        );
```

PR #45885, opened 2023-01-31
[9.x] Ensure subsiquent calls to `Mailable->to()` overwrite previous entries

AAllport:
Fixes a bug introduced by #45119 where subsequent calls to `->to()` with a different name will not take effect.
We do this by reversing the collection before the call to `unique` and then reversing it back.

A test for this functionality has been added to `MailMailableTest::testMailableSetsRecipientsCorrectly()`

More context - https://github.com/laravel/framework/pull/45119#issuecomment-1410285596

### DIFF
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -695,7 +695,9 @@ class Mailable implements MailableContract, Renderable
         }
 
         $this->{$property} = collect($this->{$property})
+            ->reverse()
             ->unique('address')
+            ->reverse()
             ->values()
             ->all();
 
--- a/tests/Mail/MailMailableTest.php
+++ b/tests/Mail/MailMailableTest.php
@@ -30,9 +30,8 @@ class MailMailableTest extends TestCase
         $mailable->to('taylor@laravel.com', 'Taylor Otwell');
 
         // Add the same recipient again, but with a different name. This should set the name correctly.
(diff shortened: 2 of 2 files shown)


AAllport:
So, when adding additional addresses via `->to()` or any of the other methods, the entire array is passed through `->unique()`
`Collection::unique()` (when you provide a key) will iterate and keep a record of the keys it has seen.
Each iteration will check the key against the keys it's already seen and reject if it's already there.
As such, `unique` will keep the first instance of the array containing the key.

Before this change, calling the following:
```php
$mailable->to("taylor@laravel.com")->to("taylor@laravel.com", "Taylor Otwell")
```
Will result in the following to array:
```php
[
    ["email"=>"taylor@laravel.com", "name"=>null]
]
```
Whereas intuitively, the name should be overridden by the last provided instance.
```php
[
    ["email"=>"taylor@laravel.com", "name"=>"Taylor Otwell"]
]
``````php
[
    ["email"=>"taylor@laravel.com", "name"=>null]
]
```

In place of a `uniqueLast()` function, we have to reverse the array to keep the last.
This broke the tests that rely on the order of the to/cc/bcc array, so we flip it back afterwards.


Frustratingly, this caught us out when updating from an older version as `->assertMailTo($email,$name)` was throwing as the `name` was null, despite calling `->to($email,$name)` in our `build()` function.

AAllport:
Awesome! Thanks for merging 🙌

PR #38016, opened 2021-07-15
[8.x] `@class` Blade directive

danharrin:
Hey! 👋 

Conditional class syntax for Blade component attribute bags has been available in Laravel since v8.27.0. It can be used to conditionally merge attributes into the bag based on Boolean expressions, like so:

```blade
<span {{ $attributes->class([
    'text-sm',
    'font-bold' => $active,
]) }}>
    Dashboard
</span>
```

In this example, the `font-bold` class would be merged into this class list when the component is `$active`.

I would like to expand the scope of this syntax one step further, and make it available in a `@class` Blade directive. Similar functionality can be found in many frontend frameworks, including [Vue.js](https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax) and [Alpine.js](https://alpinejs.dev/directives/bind#class-object-syntax). This will allow Blade developers to use conditional classes on any HTML element, not just when merging them into the attribute bag. For example:

```blade
<div>
    <span @class([
        'text-sm',
        'font-bold' => $active,
    ])>
        Dashboard
    </span>

    <svg @class([
        'w-6 h-6',
        'text-indigo-600' => $active,
    ])>
        <path />
        <path />
        <path />
    </svg>
</div>
```

To keep this new directive DRY, I extracted the class array compilation logic into a new array helper, `Arr::toCssClasses()`. Let me know if there is a better place to put this logic!

### Alternative

Alternatively, I could remove the `@class` Blade directive and allow developers to use the raw `Arr::toCssClasses()` helper, like so:

```blade
<div>
    <span class="{{ Arr::toCssClasses([
        'text-sm',
        'font-bold' => $active,
    ]) }}">
        Dashboard
    </span>

    <svg class="{{ Arr::toCssClasses([
        'w-6 h-6',
        'text-indigo-600' => $active,
    ]) }}">
        <path />
        <path />
        <path />
    </svg>
</div>
```

PR #45887, opened 2023-01-31
[9.x] Add @style Blade directive

intrepidws:
Building upon the great work by @danharrin, who added the `@class` Blade directive in PR #38016, this PR adds support for a `@style` Blade directive. While my preference is always to use `@class`, I often find myself needing to conditionally use one-off styles in projects where I don't have access to Tailwind.

Usage is very similar to the `@class` blade directive, as follows:

```
<div>
    <span @style([
        'margin-top: 10px;',
        'font-weight: bold' => $active,
    ])>
        Dashboard
    </span>

    <svg @style([
        'font-weight: bold; text-decoration: underline;',
        'margin-top: 10px' => $active,
    ])>
        <path />
        <path />
        <path />
    </svg>
</div>
```

One thing to note is that when specifying the styles, it doesn't matter whether or not you end the line with a semi-colon - the logic will add it if it's missing.

If accepted, I will PR to the docs. I could also add `@style` support for blade components as well (similar to what PR #43140 did for the `@class` directive) - if that's something that's desired.

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -4,7 +4,6 @@ namespace Illuminate\Support;
 
 use ArgumentCountError;
 use ArrayAccess;
-use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
 use InvalidArgumentException;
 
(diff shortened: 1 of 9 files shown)


taylorotwell:
I would suggest going ahead and adding support for Blade components. I don't think we would want to merge support for this without also supporting Blade components since `@class` is supported on both.

taylorotwell:
Please mark as ready for review when done. Thanks!

intrepidws:
No problem, this is ready for review @taylorotwell

PR #45892, opened 2023-02-01
Built in Auth Notification classes implement ShouldQueue interface and Queueable trait

BoilingSoup:
I think the `ResetPassword` and `VerifyEmail` notifications should implement the `ShouldQueue` interface and `Queueable` trait by default. This does not effect the current default behavior because the default queue driver is set to `sync` in the `.env`. 

This benefits the end users because it maintains the current functionality, and also allows the user to seamlessly utilize a queue driver for the `ResetPassword` and `VerifyEmail` notifications.

I recently needed to use a background queue to send these notifications, and I had to dig inside the `vendor` directory to find these base classes, create new classes that extend these and manually add the `ShouldQueue` interface + `Queueable` trait. I think that friction is avoidable and it would simply be better to have that done in the core.

### DIFF
--- a/src/Illuminate/Auth/Notifications/ResetPassword.php
+++ b/src/Illuminate/Auth/Notifications/ResetPassword.php
@@ -2,12 +2,15 @@
 
 namespace Illuminate\Auth\Notifications;
 
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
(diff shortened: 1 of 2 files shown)


ankurk91:
I was going to send same PR.😅

taylorotwell:
Could you send this to 10.x (`master` branch) instead?

PR #45893, opened 2023-02-01
[10.x] Add support to handle exceptions on Timebox

rodrigopedra:
PR #44069 introduced the Timebox class, which "aims to guard against timing attacks at the application".

Unfortunately, if the Timebox callback throws an Exception, the Timebox guard is bypassed, even if the developer handles this exception gracefully somewhere else.

An example: when implementing a custom guard that verifies a remote service, this verification can fail, bypassing the timebox guard entirely.

This PR:

- Adds support to handle exceptions thrown during a Timebox's callback execution
- Adds relevant test cases

### DIFF
--- a/src/Illuminate/Support/Timebox.php
+++ b/src/Illuminate/Support/Timebox.php
@@ -22,14 +22,23 @@ class Timebox
      */
     public function call(callable $callback, int $microseconds)
     {
+        $exception = null;
         $start = microtime(true);
 
-        $result = $callback($this);
+        try {
+            $result = $callback($this);
+        } catch (\Throwable $caugth) {
+            $exception = $caugth;
+        }
 
         $remainder = intval($microseconds - ((microtime(true) - $start) * 1000000));
 
         if (! $this->earlyReturn && $remainder > 0) {
             $this->usleep($remainder);
+        }
+
+        if ($exception) {
+            throw $exception;
         }
 
         return $result;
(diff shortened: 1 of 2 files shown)


taylorotwell:
@rodrigopedra since this was sent to 10.x I assume there are some breaking changes? Can you outline what would need to be in the upgrade guide for this change?

rodrigopedra:
@taylorotwell 

Well, I sent it to the 10.x branch as before it didn't handle exceptions, and now it does. I don't think someone was relying on exceptions not being delayed.

If you prefer, I can send this to 9.x.

Otherwise, a suggestion to the update guide is the following:

### The `Timebox` class

**Likelihood Of Impact: Very Low**

Previously, the `Illuminate\Support\Timebox` class would not respect the configured delay if an exception was thrown within its callback. Now, this class will respect the defined delay to throw the exception.

If you want to keep the previous behavior, you can disable it by wrapping any clauses that could throw an exception within a `try...catch` and telling the timebox instance to return earlier.

```php
use Illuminate\Support\Timebox;
use Throwable;

(new Timebox)->call(function ($timebox) {
    try {
        // any potential faulty statements, such as accesing an external service
    } catch (Throwable $e) {
        $timebox->returnEarlier();
        throw $e;
    }
}, 10000);
```


P.S.: Please apologize in advance for any grammar mistakes.

rodrigopedra:
It is more like a behavior change, not a breaking change. But I can imagine someone noticing the delayed exceptions, specially in large test suits.

PR #45895, opened 2023-02-01
[10.x] Built in Auth Notification classes implement ShouldQueue interface and Queueable trait

BoilingSoup:
I think the `ResetPassword` and `VerifyEmail` notifications should implement the `ShouldQueue` interface and `Queueable` trait by default. This does not effect the current default behavior because the default queue driver is set to `sync` in the `.env`. 

This benefits the end users because it maintains the current functionality, and also allows the user to seamlessly utilize a queue driver for the `ResetPassword` and `VerifyEmail` notifications.

I recently needed to use a background queue to send these notifications, and I had to dig inside the `vendor` directory to find these base classes, create new classes that extend these and manually add the `ShouldQueue` interface + `Queueable` trait. I think that friction is avoidable and it would simply be better to have that done in the core.

### DIFF
--- a/src/Illuminate/Auth/Notifications/ResetPassword.php
+++ b/src/Illuminate/Auth/Notifications/ResetPassword.php
@@ -11,6 +11,7 @@ use Illuminate\Support\Facades\Lang;
 class ResetPassword extends Notification implements ShouldQueue
 {
     use Queueable;
+
     /**
      * The password reset token.
      *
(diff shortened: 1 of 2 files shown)


taylorotwell:
Just for my curiosity - why did you **need** to make these notifications queuable in your application?

BoilingSoup:
> Just for my curiosity - why did you **need** to make these notifications queuable in your application?

I was using a redis queue in my app to send emails via background worker & keep the UI snappy. Maybe "need" was not the right word.

PR #40943, opened 2022-02-10
[9.x] Allows tls encryption to be used with port different than 465 with starttls

nicolalazzaro:
Resolve this issue: https://github.com/laravel/framework/issues/40887

PR #45900, opened 2023-02-01
fix: allow SMTP TLS to be set for all ports

### LINKED ISSUE #45899: Cannot set TLS for email on port other than 465
- Laravel Version: 9.23.0
- PHP Version: 8.1.9
- Database Driver & Version: Postgres 14.4

### Description:
At the beginning of last year, someone was unable to connect to an email server (#40887). I cannot really follow what the underlying problem eventually is, but a pull request (#40943) is made that fixes the issue and everything is closed.

The problem is that in that PR, the configuration is locked down the `smtps` scheme for Symfony Mailer is only set in case `MAIL_ENCRYPTION` is set to `tls` and `MAIL_PORT` is set to `465`. The problem is that if you would like to connect to a SMTP server via TLS directly on a port other than 465, there is no way to configure this.

My hunch is that somewhere along the way TLS and STARTTLS got confused. As a recap ([per Mailtrap](https://mailtrap.io/blog/starttls-ssl-tls/)), TLS means that the whole SMTP connection is encapsulated in TLS. This is often called SMTPS. Whereas with STARTTLS, the connection will start as a regular SMTP connection and get upgraded to TLS during the handshake. 

The scheme Symfony Mailer refers to determines the TLS status, so `smtps` is used for SMTP over TLS, where as `smtp` is used for plain SMTP, with a potential upgrade using STARTTLS. The extra confusing thing is that Symfony Mailer [deprecated explicit STARTTLS options](https://github.com/symfony/mailer/blob/6.2/CHANGELOG.md) in version 4.4, meaning that from this version onwards it will automatically upgrade a connection with STARTTLS if it is available.

I would like to make it possible to use SMTPS for ports other than 465. This is a valid use case since some cloud providers block common SMTP ports (20, 465, 587) and thus offer alternative ports (e.g. Scaleway offers SMTP over TLS on port 2465.

Hence, I propose to revert PR #40943 and will submit a PR soon for doing so soon. Using STARTTLS was already perfectly possible before the PR, but this meant users had to set `MAIL_ENCRYPTION` to `null`. I get that this is counter-intuitive, so maybe it is a good idea to add some docs and/or consider dropping `MAIL_ENCRYPTION` so developers know how to deal with this.

Pinging @driesvints and @nicolalazzaro since you were involved with the mentioned issue and PR. 

### Steps To Reproduce:
Create a new Laravel application, and attempt to send out an email to a TLS SMTP-server on a port other than 465 (e.g. 2465). The connection will time-out and the following error will be thrown:
```
Symfony\Component\Mailer\Exception\TransportException with message 'Connection could not be established with host "server:2465": stream_socket_client(): Unable to connect to server:2465 (Operation timed out)'
```

leinelissen:
This reverts PR #40943
This resolves #45899

Basically, this allows for setting TLS settings on ports other than 465. If someone wants to use STARTTLS, they should set `MAIL_ENCRYPTION` to `null` and Symfony Mailer will automatically upgrade the connection if STARTTLS is available.

### DIFF
--- a/src/Illuminate/Mail/MailManager.php
+++ b/src/Illuminate/Mail/MailManager.php
@@ -172,7 +172,7 @@ class MailManager implements FactoryContract
         $factory = new EsmtpTransportFactory;
 
         $transport = $factory->create(new Dsn(
-            ! empty($config['encryption']) && $config['encryption'] === 'tls' ? (($config['port'] == 465) ? 'smtps' : 'smtp') : '',
+            ! empty($config['encryption']) && $config['encryption'] === 'tls' ? 'smtps' : '',
             $config['host'],
             $config['username'] ?? null,
             $config['password'] ?? null,


taylorotwell:
So how does this not just put right back in the situation that people were saying was broken?

leinelissen:
If you want I can add documentation that clarifies configuration for TLS and STARTTLS in the SMTP transporter. I've also looped in the author of the other PR to find out what the underlying problem was. 

Regardless, I feel the current situation hurts SMTP configurability so I do want to propose this as a necessary change.

taylorotwell:
Can't merge this PR if we're just going back to something people were complaining about.

taylorotwell:
It wasn't just the original author - there were lots of people in that thread?

taylorotwell:
Added this: https://github.com/laravel/framework/commit/68a8bfc3ab758962c8f050160ec32833dc12e467

leinelissen:
Yeah that's perfect, thanks so much!

PR #45904, opened 2023-02-01
[9.x] Add missing tests for destroy method

imanghafoori1:
I think the reported backward incompatibility here: https://github.com/laravel/framework/pull/45709#issuecomment-1412345778 is not really a valid one.

## Why:
The reported example is passing an array of eloquent models into the destroy method, which has never been supported or documented.

The added test shows that if we pass an array of models into the destroy method (the implementation after the revert), depending on the driver either nothing gets deleted (in MySQL, Maria) or an error will be thrown (in Postgres or SQL server).

### DIFF
--- a/tests/Integration/Database/EloquentDeleteTest.php
+++ b/tests/Integration/Database/EloquentDeleteTest.php
@@ -143,6 +143,17 @@ class EloquentDeleteTest extends DatabaseTestCase
         // Total of 3 queries.
         $this->assertCount(3, $logs);
 
+        // It can accept an array models
+        PostStringyKey::query()->delete();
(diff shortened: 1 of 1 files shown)


src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -1353,6 +1353,7 @@ public static function destroy($ids)
 
         foreach ($instance->whereIn($key, $ids)->get() as $model) {
             if ($model->delete()) {
+                dump($count);
```
driesvints, line 1356: Remove this dump

taylorotwell:
I think we can just make it support it.

PR #45911, opened 2023-02-01
[9.x] Fix typos

PerryvanderMeer:
[Since we use `middleware` as plural.](https://english.stackexchange.com/questions/257120/middleware-vs-middlewares)
Left some variable names unchanged, although we might want to change them from `$middlewares` to `$middleware`.


taylorotwell:
It's OK for now.

PR #45916, opened 2023-02-02
[9.x] Random function doesn't generate evenly distributed random chars

janh-kramer:
The random function does not generate evenly distributed random characters particularly for short strings (<12 chars). For example, if exactly 1 character is generated, the letters A, Q, g and w are significantly overrepresented.

This is due to the conversion of the randomly generated bytes using the base64 function. If the number of bytes is not divisible by three, zeros are added at the end in the algorithm (see [https://en.wikipedia.org/wiki/Base64](https://en.wikipedia.org/wiki/Base64)).

This leads to the increased frequency of A (000000), Q (010000), g (100000) and w (110000).

Evenly distributed random characters are obtained if it is ensured that the number of generated random bytes modulo 3 is equal to 0:

```jsx
$bytesSize = (int) ceil(($size) / 3) * 3;
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -735,7 +735,9 @@ class Str
             while (($len = strlen($string)) < $length) {
                 $size = $length - $len;
 
-                $bytes = random_bytes($size);
+                $bytesSize = (int) ceil(($size) / 3) * 3;
+
+                $bytes = random_bytes($bytesSize);
 
                 $string .= substr(str_replace(['/', '+', '='], '', base64_encode($bytes)), 0, $size);
             }
--- a/tests/Support/SupportStrTest.php
+++ b/tests/Support/SupportStrTest.php
@@ -492,7 +492,6 @@ class SupportStrTest extends TestCase
         $results = [];
         // take 6.200.000 samples, because there are 62 different characters
         for ($i = 0; $i < 6200000; $i++) {
-
             $random = Str::random(1);
             $results[$random] = ($results[$random] ?? 0) + 1;
         }
--- a/tests/Support/SupportStrTest.php
+++ b/tests/Support/SupportStrTest.php
@@ -487,22 +487,21 @@ class SupportStrTest extends TestCase
     }
 
     /** @test */
-    public function TestWhetherTheNumberOfGeneratedCharactersIsEquallyDistributed() {
+    public function TestWhetherTheNumberOfGeneratedCharactersIsEquallyDistributed()
+    {
         $results = [];
         // take 6.200.000 samples, because there are 62 different characters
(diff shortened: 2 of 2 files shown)


janh-kramer:
@taylorotwell, When changing the test, you forgot to adjust the delta, so now the test will always pass.

taylorotwell:
ok, what should I change the delta to?

janh-kramer:
At most 200, could even be less

janh-kramer:
Should I target this to 10.x?

taylorotwell:
Sure

giggsey:
It's a minor complaint, but can the comments be updated to match the change in the iterations in https://github.com/laravel/framework/pull/45916/commits/30681bcebba490b5bff2b84901cbfd12938d66df

```php
        // take 6.200.000 samples, because there are 62 different characters
        for ($i = 0; $i < 620000; $i++) {
```

driesvints:
@janh-kramer this is already merged into 10.x 👍

janh-kramer:
Is this PR then accepted as a candidate for the Bughunt?

driesvints:
It was sent to 9.x so no sorry.

janh-kramer:
come on ..

driesvints:
I'm very sorry @janh-kramer but the rules are very clear on this. Only bug fixes sent to 10.x are eligible. The contest is specifically to figure out breaking changes in Laravel v10. Bugs that already exist in 9.x need to be solved on that version.

Please see https://blog.laravel.com/laravel-v10-bug-hunt

janh-kramer:
no problem thanks, it was just a try. Glad to contribute the first time

ziming:
I dont see any harm in allowing. not like 1 counted PR will win the contest. plus 7 feb is near. but that's just an opinion

driesvints:
@janh-kramer we very much appreciate your contribution 👍 

@ziming the problem with allowing one now would be unfair to everyone else who has sent in bug fixes to 9.x in the past weeks.

ziming:
Ah i see. :)

GrahamCampbell:
This seems overly complex? Why not just base62 encode the output (by which i mean, divide and ciel the number of chars to determine the number of bytes we need, gen the bytes, base62 encode the output, then trim to the correct length - which will remove either 0 or 1 chars off of the end)?

GrahamCampbell:
The reason this is correct is that bsse62 can be thought of as stretching out the bytes over the charset we want, which means we get the perfect distribution we want.

taylorotwell:
@janh-kramer let us think about the bug hunt stuff - so far we only have 3 bug fixes contributed to 10.x which feels a bit lame to me and I think it would be better to include 9.x as well personally. Will revisit this.

PR #45923, opened 2023-02-02
Added @mixin annotation to the Model class for improved autocompletion for IDEs eg. PhpStorm.

ringunger:
## Introduction
[DocBlocks](https://en.wikipedia.org/wiki/PHPDoc) can be used by IDEs like PhpStorm, to improve autocomplete suggestions.

## Problem
When using IDE like PhpStorm some of methods on model are note autocompleted and are sometimes underlined as errors (Method not found in model). Example:
<pre>
$user = User::where('id', 1)->first();
</pre>
The `where` method on user is considered as not available on the model.

## Work-around
I normally add the @mixin to all of my models during development so that my autocompletes work perfectly and I get rid of lint errors. 
<pre>
namespace App\Models;

use Illuminate\Foundation\Auth\User as Authenticatable;

/**
 * @mixin Builder
 */
{
    // User methods
}
</pre>
The above solves the issue for the model.

## Suggestion
I Would like to see this set by default for all models. 
I Hope this is a the appropriate way to do it. Otherwise I would like to have it implemented the correct way.

## Benefits
With this fix, the IDE would render all Builder autosuggestions on models during their usage. This gives the developer more confidence and speeds up development.

## Breaking 
This change doesn't break anything as it is only an annotation and doe not modify any part of the existing code.
> Note: I could not write any test because this change does not alter any operation or function.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -23,6 +23,9 @@ use Illuminate\Support\Traits\ForwardsCalls;
 use JsonSerializable;
 use LogicException;
 
+/**
+ * @mixin Builder
+ */
 abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToString, HasBroadcastChannel, Jsonable, JsonSerializable, QueueableEntity, UrlRoutable
 {
     use Concerns\HasAttributes,


taylorotwell:
I don't think mixing builder into model makes sense. You can always call `Model::query` which has the correct return type annotated.

ringunger:
Thanks. Now I understand.
I was always calling the builder methods directly on the `Modal` class.
'Just because it works, doesn't mean it's right'. :)

PR #45927, opened 2023-02-02
[9.x] Adds ability to compiles new http verbs directives 🚦

mahmoudmohamedramadan:
This PR allows us to use the new short directives instead `@method` directive, let's examine the next code for more illumination.

## BEFORE

```BLADE
<form method="post" action="/users/{{ $user->id }}">
    @csrf
    @method('PATCH')
</form>
```

But after this PR gets merged, the code will be like so 🎉

## AFTER

```BLADE
<form method="post" action="/users/{{ $user->id }}">
    @csrf
    @patch
</form>
```

### DIFF
--- a/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
@@ -57,7 +57,7 @@ trait CompilesHelpers
      */
     protected function compilePut($arguments)
     {
-        if ($arguments === '()') {
+        if ($arguments === null) {
             return $this->compileMethod("('PUT')");
         }
 
@@ -72,7 +72,7 @@ trait CompilesHelpers
      */
     protected function compilePatch($arguments)
     {
-        if ($arguments === '()') {
+        if ($arguments === null) {
             return $this->compileMethod("('PATCH')");
         }
 
@@ -87,7 +87,7 @@ trait CompilesHelpers
      */
     protected function compileDelete($arguments)
     {
-        if ($arguments === '()') {
+        if ($arguments === null) {
             return $this->compileMethod("('DELETE')");
         }
 
--- a/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
+++ b/src/Illuminate/View/Compilers/Concerns/CompilesHelpers.php
@@ -50,6 +50,36 @@ trait CompilesHelpers
     }
 
     /**
+     * Compile the patch method statements into valid PHP.
+     *
+     * @return string
+     */
+    protected function compilePatch()
+    {
+        return $this->compileMethod("('PATCH')");
+    }
+
+    /**
+     * Compile the put method statements into valid PHP.
+     *
+     * @return string
+     */
+    protected function compilePut()
+    {
+        return $this->compileMethod("('PUT')");
+    }
+
+    /**
+     * Compile the delete method statements into valid PHP.
+     *
+     * @return string
+     */
+    protected function compileDelete()
+    {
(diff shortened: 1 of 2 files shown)


mahmoudmohamedramadan:
Update the comment.

browner12:
this seems excessive, IMO. we're not gaining anything with the switch, save a few keystrokes.

if this were to get merged, would it have to go to 10.x, in case someone is defining these in userland? I'm not sure the precedence when it comes to Blade.

timacdonald:
@browner12 userland directives will take precedence over framework ones. See https://github.com/laravel/framework/pull/28062#issuecomment-483498281

I like the idea, but wanted to also replace the need for `action`, `method`,  and patch `input` in one swoop.

I've managed to do this with the following API:


```blade
<!-- BEFORE -->

<form method="post" action="/users/{{ $user->id }}">
    @csrf
    @method('PATCH')
</form>

<!-- AFTER -->

<form @patch("/users/{$user->id}")">
    @csrf
</form>
```

This adds the `_method` input into the query parameter instead of the form body.

It is still possible to put it into the method body via @mahmoudmohamedramadan's original API.

No arguments: outputs the hidden input.
Argument: outputs the method and action form attribute.

FYI: a url can have more than one question mark, but the first one always indicates the start of the query parameters, hence the primitive string concatenation is possible without fully parsing into a URL structure.

mahmoudmohamedramadan:
@timacdonald The feature of canceling `method` and `action` attributes are very nice but, there is an issue of your code becuase, you send the `_method` as a query string and it must be sent in the form body.

mahmoudmohamedramadan:
I have updated the code according to the new code style that @timacdonald suggested but, I found some issues, so I have solved these issues and the code style has changed again 🚀.

## Before

```BLADE
<form method="post" action="/users/{{ $user->id }}">
    @csrf
    @method('PATCH')
</form>
```

## AFTER

```BLADE
<form @patch("/users/{$user->id})">
    @csrf
    @patch
</form>
```

taylorotwell:
@mahmoudmohamedramadan are you sure it can't be in the query string. Here is the source:

CleanShot 2023-02-02 at 21 00 30@2x

timacdonald:
Here is a quick script to demo the query parameter working.

```
composer create-project laravel/laravel patch-test \
  && cd patch-test \
  && echo "Route::patch('patch-route', fn () => 'You are on a PATCH route!');" >> routes/web.php \
  && echo "<form method='POST' action='/patch-route?_method=PATCH'>
  @csrf
  <button>Submit</button>
</form>" > resources/views/welcome.blade.php

## Open in browser
(sleep 1 && open http://127.0.0.1:8000/)&
php artisan serve
```

mahmoudmohamedramadan:
@taylorotwell Whenever I submit the form, the 419 expextion is thrown.

mahmoudmohamedramadan:
@taylorotwell Whenever I submit the form, the 419 expextion is thrown.

timacdonald:
Is the form using use the `@csrf` directive? That is still required.

mahmoudmohamedramadan:
@timacdonald Ooh, I'm forgotten to add `@csrf`, But there is another issue that is

```PHP
if ($arguments === "()") {
    return $this->compileMethod("('PATCH')");
}
```

must be like so

```PHP
if ($arguments === null) {
    return $this->compileMethod("('PATCH')");
}
```

taylorotwell:
Tabling this one for now.

mahmoudmohamedramadan:
OK, but why you closed the PR if it will be merged later?

mahmoudmohamedramadan:
@taylorotwell This pull request will never be merged?

PR #45928, opened 2023-02-02
[9.x] Fixes blade escaped tags issue

### LINKED ISSUE #45915: Escaping Blade directives with @@ is broken
- Laravel Version: 9.50.1
- PHP Version: 8.2.1
- Database Driver & Version: (irrelevant)

### Description:
[Escaping Blade directives](https://laravel.com/docs/9.x/blade#blade-and-javascript-frameworks) with `@` symbol is broken in some cases.

### Steps To Reproduce:

Consider this valid blade:
```
@php
    $columns = [];
    $locales = [];
@endphp

@foreach($columns as $col)
    @if($col == 'foo')
        <div></div>
    @elseif($col == 'bar')
        @@foreach($locales as $locale)
            <div></div>
        @@endforeach
    @else
        <div></div>
    @endif
@endforeach
```

Expected behaviour:
Blade will complile. Double `@@` should just produce one `@`, just like written in the documentation (and how it worked few weeks ago, something must have break just lately).

I'm getting:
```
syntax error, unexpected token "endforeach", expecting "elseif" or "else" or "endif"
```

One solution is to instead of using `@@` to use `{{'@'}}`. If you prefer this solution, then please update the documentation so this is the recommended way.

imanghafoori1:
This fixes #45915

In the added test there are 4 matches, in the second replacement `@@endforeach` with `@endforeach` in the second replacement.

Then the final replacement searches for `@endforeach` to replace it with the compiled version, but it finds what was produced by the previously escaped tag because it starts the search from the beginning of the string.

### Solution:
Logically when replacing the matches found by REGEX one after the other using `Str::replaceFirst`, we have to continue from where we last left off.

### DIFF
--- a/src/Illuminate/View/Compilers/BladeCompiler.php
+++ b/src/Illuminate/View/Compilers/BladeCompiler.php
@@ -507,6 +507,7 @@ class BladeCompiler extends Compiler implements CompilerInterface
     {
         preg_match_all('/\B@(@?\w+(?:::\w+)?)([ \t]*)(\( ( [\S\s]*? ) \))?/x', $template, $matches);
 
+        $offset = 0;
         for ($i = 0; isset($matches[0][$i]); $i++) {
             $match = [
                 $matches[0][$i],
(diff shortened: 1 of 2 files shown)


taylorotwell:
Are all these substr_replace, strlen calls going to work if there are multi-byte characters in the template?

imanghafoori1:
@taylorotwell I will design some tests for those cases.

imanghafoori1:
@taylorotwell It works ok with emojis or other 3-bit chars but it needs some testing, because if someone changes `strpos` to `mb_str_pos` it will break;

PR #45929, opened 2023-02-02
Easy getter and setter methods for Json columns

rana01645:
Already laravel is giving enough flexibility on getting and setting values on JSON columns, gues we have a JSON column named `meta` like this
` $table->json('meta')->nullable();`

Currently, we can interact with it like this,

We need to cast this column as in model

```
    protected $casts = [
        'meta' => 'json',
    ];
```

then we can get the value as 
`Arr::get($this->meta, $key, $default)`

Also, set the value as

```
$this->update([
            'meta->something' => $value
        ]);
```

What if we can use it directly as `getMeta('key')` and `setMeta('key')`

This PR is all about adding a setter and getter for JSON column which is helpful to interact with!

thanks for looking into this PR, any feedback/suggestion is highly appreciated

image

### DIFF
--- a/src/Illuminate/Database/Eloquent/InteractsWithJsonColumn.php
+++ b/src/Illuminate/Database/Eloquent/InteractsWithJsonColumn.php
@@ -80,8 +80,7 @@ trait InteractsWithJsonColumn
 
         return parent::__call($method, $arguments);
     }
-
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #45930, opened 2023-02-03
[10.x] use AWS SES v2 Client for `ses` mailer

jacobmllr95:
Using the AWS SES v2 Client automatically increases the 10 MB outbound and 30 MB inbound message size limit to 40 MB, as further explained in this blogpost: https://aws.amazon.com/de/about-aws/whats-new/2022/04/amazon-ses-v2-supports-email-size-40mb-inbound-outbound-emails-default/

This is quite a huge deal, especially when dealing with email attachments.

### DIFF
--- a/src/Illuminate/Mail/Transport/SesTransport.php
+++ b/src/Illuminate/Mail/Transport/SesTransport.php
@@ -3,7 +3,7 @@
 namespace Illuminate\Mail\Transport;
 
 use Aws\Exception\AwsException;
-use Aws\Ses\SesClient;
+use Aws\SesV2\SesV2Client;
 use Exception;
 use Symfony\Component\Mailer\Header\MetadataHeader;
 use Symfony\Component\Mailer\SentMessage;
@@ -15,7 +15,7 @@ class SesTransport extends AbstractTransport
     /**
      * The Amazon SES instance.
      *
-     * @var \Aws\Ses\SesClient
+     * @var \Aws\SesV2\SesV2Client
      */
     protected $ses;
 
@@ -29,11 +29,11 @@ class SesTransport extends AbstractTransport
     /**
      * Create a new SES transport instance.
      *
-     * @param  \Aws\Ses\SesClient  $ses
+     * @param  \Aws\SesV2\SesV2Client  $ses
      * @param  array  $options
      * @return void
      */
-    public function __construct(SesClient $ses, $options = [])
+    public function __construct(SesV2Client $ses, $options = [])
     {
         $this->ses = $ses;
         $this->options = $options;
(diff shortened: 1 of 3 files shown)


ziming:
currently I had to have multiple mailers and switch when the attachment size go above ses v1 limits. this will be a great improvement.

is there something breaking that makes u target v10 instead of v9?

src/Illuminate/Mail/Transport/SesTransport.php
```diff
@@ -57,17 +57,21 @@ protected function doSend(SentMessage $message): void
...
-                        'Destinations' => collect($message->getEnvelope()->getRecipients())
-                                ->map
-                                ->toString()
-                                ->values()
-                                ->all(),
-                        'RawMessage' => [
-                            'Data' => $message->toString(),
+                        'ReplyToAddresses' => [$message->getEnvelope()->getSender()->toString()],
+                        'Destination' => [
+                            'ToAddresses' => collect($message->getEnvelope()->getRecipients())
```
taylorotwell, line 65: What if they are `cc` addresses?

jacobmllr95:
@ziming I wasn't sure if switching a service API version is a breaking change or not, so I targeted v10.
If targeting v9 is fine, there should not be any breaking changes.

src/Illuminate/Mail/Transport/SesTransport.php
jacobmllr95: If there is a way to decide between `to`, `bcc` and `cc` from the `SentMessage`, I would love to implement it:
[[SentMessage]()](https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-sesv2-2019-09-27.html#sendemail)

taylorotwell: Have you tested sending emails with cc and bcc recipients with this PR in place? What is the behavior?

jacobmllr95: @taylorotwell I tested `bcc` and `cc` and they work as expected.

taylorotwell:
Personally I would make this an entire new transport alongside the current v1 transport... like `ses-v2` or something.

jacobmllr95:
@taylorotwell We can also go this route an target Laravel 9, if you would prefer that.
The thing about that is that it's quite a bit code duplication and would also required additional documentation.

taylorotwell:
The class is fairly small - I don't consider it a burdensome amount of code duplication.

jacobmllr95:
Don't get me wrong, it's not that much duplicated code.
But then we have two `sns` transports and the user has to decide which one to use.

When AWS releases v3, do we add a third `sns` transport?
When do we drop support for v1?
And so one...

But in the end it's up to you :)

deleugpn:
I think it's not about giving users too many options. In fact, it could very well be that only v2 becomes documented and v1 fades out. But with 2 classes there's no chance existing code would break because of this change in Laravel.

taylorotwell:
Feel free to open a PR as a separate transport as discussed.

PR #45934, opened 2023-02-03
[9.x] Add sugar method to the eloquent model

gocanto:
### Description

Hi Team, I hope this message finds you well. I wonder whether you would consider adding this method to the framework since I keep finding myself looking for it when joining multiple tables.

I know and understand we could have this at userland, but I just wanted to try. Furthermore, I recall seeing a PR in the past attempting this, but not sure where it went or what the team decided. Thus, trying for the fun of it.

If this method were to be considered, but it might clash with userland methods, I bet Tayloar can find a more appropriate name. 

### Todos
- [ ] Send to master since it might be a breaking change 💭 
- [ ] Add tests in case we wanna have it 😉 

Cheers and happy Friday 🍺

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1776,6 +1776,16 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     }
 
     /**
+     * Sugar method that allows to get the table associated with the model.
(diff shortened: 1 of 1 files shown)


joelharkes:
doesnt this already happen, same as the query() method?
you could just docType it if you want

taylorotwell:
Add this to your own projects if you wish.

PR #45939, opened 2023-02-03
[9.x] Fix docblock issue

fabio-ivona:
See #45749

`append` and `prepend` methods accept a variadic list of `string`s 

the docblock changed in #45749 should be `@param string ...$values` in order to be correct

otherwise IDE would mark an error like this one:

image

**note** this should be fixed also in 10.x, will open a new PR as soon as this is approved

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -57,7 +57,7 @@ class Stringable implements JsonSerializable
     /**
      * Append the given values to the string.
      *
-     * @param  array  ...$values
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks!

fabio-ivona:
Thanks @driesvints!, will fix also in 10.x

driesvints:
@fabio-ivona 9.x is merged into 10.x by us so we're good 👍

fabio-ivona:
Ok, sorry, did miss that!

PR #45942, opened 2023-02-03
[10.x] Fix docblock issue

fabio-ivona:
same fix as in #45939


driesvints:
Heya, as said, your commits will be merged in 10.x 👍

PR #45947, opened 2023-02-03
[10.x] Add creation and update datetime columns

Jim-Webfox:
Add datetimes function to create "created_at" and "updated_at" columns using "datetime" columns instead off "timestamps" columns to add further future proofing.

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -1202,6 +1202,19 @@ class Blueprint
 
         $this->timestampTz('updated_at', $precision)->nullable();
     }
+    
+     /**
+     * Add creation and update datetime columns to the table.
+     *
+     * @param  int|null  $precision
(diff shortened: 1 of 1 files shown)


ankurk91:
If this is better way then all of migration stub should be updated with this

taylorotwell:
Most information I find online still suggests using timestamp for thing like created_at and updated_at timestamps.

ankurk91:
https://stackoverflow.com/questions/409286/should-i-use-the-datetime-or-timestamp-data-type-in-mysql

https://bugs.mysql.com/bug.php?id=12654

Jim-Webfox:
@taylorotwell I fully agree for web applications that are going to be consistently maintained and updated. The reason for my PR was that I've come across many web applications that just aren't kept up to date, and I feel like that trend will continue. So for the sake of those web apps that probably wont be maintained but will continue to be used for many years I feel like this is a good option to ensure longevity.

PR #45948, opened 2023-02-03
[9.x] Fix `Collection::wrap()` docblock typing

ralphjsmit:
The `Arr::wrap()` and `Collection::wrap()` are very useful to wrap an item in an array/collection or, if it is already an array collection, don't apply a wrap.

I noticed that my IDE was complaining about the type of `Collection::wrap()`, because currently the typehint would only support iterables. That's weird, because the `wrap()` was specifically designed to accept anything (and wrap that anything into an array/collection):

Scherm­afbeelding 2023-02-03 om 18 59 57

The interesting thing is that `Arr::wrap()` already has the correct type, it's only an issue in `Collection::wrap()`:

Scherm­afbeelding 2023-02-03 om 19 02 19

Not sure if this counts as a bug fix 😜 Thanks!

### DIFF
--- a/src/Illuminate/Collections/Enumerable.php
+++ b/src/Illuminate/Collections/Enumerable.php
@@ -54,7 +54,7 @@ interface Enumerable extends Arrayable, Countable, IteratorAggregate, Jsonable, 
      * @template TWrapKey of array-key
      * @template TWrapValue
      *
-     * @param  iterable<TWrapKey, TWrapValue>  $value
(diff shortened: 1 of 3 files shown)


ralphjsmit:
I'm a bit unsure how to fix/update the type tests, perhaps @nunomaduro could have look? Thanks!

taylorotwell:
@nunomaduro can you mark as ready for review when this is good?

nunomaduro:
https://github.com/laravel/framework/pull/45948

PR #45949, opened 2023-02-03
[9.x] Command schedule:work minor features: schedule:run output file & environment specific verbosity

ralphschindler:
Minor features to make `schedule:work` less chatty in certain environments:

`schedule:work` command to accept option `--run-output-file=` to redirect schedule:run output. This would allow someone using `schedule:work` to have it run similar to [directions for cron](https://laravel.com/docs/9.x/scheduling#running-the-scheduler), this is intended to be used like `artisan schedule:work --run-output-file=/dev/null` which would result in run being called like in this process list:

```
php /app/artisan schedule:work --run-output-file=/dev/null
  sh -c '/usr/bin/php8.1' 'artisan' 'schedule:run' >> '/dev/null' 2>&1
```

Additionally, different default verbosity levels for this command depending on the environment: local normal) vs. production = verbose (`-v`)

### DIFF
--- a/src/Illuminate/Console/Scheduling/ScheduleWorkCommand.php
+++ b/src/Illuminate/Console/Scheduling/ScheduleWorkCommand.php
@@ -17,7 +17,7 @@ class ScheduleWorkCommand extends Command
      *
      * @var string
      */
-    protected $signature = 'schedule:work {--run-output-file=}';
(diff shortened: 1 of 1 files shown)


taylorotwell:
Thanks 👍

PR #45950, opened 2023-02-03
[9.x] Wrong or not enough accurate types

Shizzen83:
Hello there,

First of all, thanks for your hard work on this amazing framework! This is my first PR ever, so please let me know if there is anything wrong.

As every change targets only `@param`/`@return` comments, there is no breaking change.

Finally, I'd like to know if you are interested by a 10.x PR changing `void` types to `never` (PHP 8.1 feature) when suitable.

Thanks in advance!

### DIFF
--- a/src/Illuminate/Database/Grammar.php
+++ b/src/Illuminate/Database/Grammar.php
@@ -128,7 +128,7 @@ abstract class Grammar
      * Wrap the given JSON selector.
      *
      * @param  string  $value
-     * @return string
+     * @return void
      *
      * @throws \RuntimeException
      */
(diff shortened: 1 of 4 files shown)


driesvints:
Unfortunately we had to revert this. The return types are actually correct. The fact that these methods don't return strings is a side effect from the exceptions that are thrown to indicate the methods aren't implemented yet. If you look at the implementations of the methods you can see that they are actually expected to return strings.

The classes you added to the DocBlocks also didn't contain FQN's. 

Appreciate you sending this in but seems like these changes weren't necessary, sorry.

Shizzen83:
@driesvints Thanks for your feedback, I understand it. Though, how about the changes except not implemented methods?

For example, `compileCreateTable` method from `Illuminate\Database\Schema\Grammars\MySqlGrammar`? Its DocBlock shows `@return array` but its implementation obviously returns a string, further used by `compileCreateEncoding` which also expects a string. I overrided this method but my PHPStan complains about this wrong type.

Missing types seem also correct because of `instanceof` operator.

driesvints:
@Shizzen83 you can try a new PR 👍

PR #45956, opened 2023-02-03
Revert "[10.x] Built in Auth Notification classes implement ShouldQueue interface and Queueable trait"

taylorotwell:
Reverts laravel/framework#45895


BoilingSoup:
Can we know why this was reverted? 
This means I'd need to dig the base classes out the `/vendor` directory in order to make them queueable.

`/vendor` is gitignored, so many text editors will not find the base classes via a project-wide text search.

Thanks

PR #45959, opened 2023-02-04
[10.x] Add contract to Http Kernel Bootstrappers

milwad-dev:
Laravel has always been known for code style.
In http kernel all bootstraps must be have ```bootstrap``` method, but if some developer modify the method name of bootstrapper or add a new bootstrapper like this:
```php 
    protected $bootstrappers = [
        \Illuminate\Foundation\Bootstrap\LoadEnvironmentVariables::class,
        \Illuminate\Foundation\Bootstrap\LoadConfiguration::class,
        \Illuminate\Foundation\Bootstrap\LoadConfigProject::class, // Add new for test
        \Illuminate\Foundation\Bootstrap\HandleExceptions::class,
        \Illuminate\Foundation\Bootstrap\RegisterFacades::class,
        \Illuminate\Foundation\Bootstrap\RegisterProviders::class,
        \Illuminate\Foundation\Bootstrap\BootProviders::class,
    ];
```
After in browser the message error is:

 
Or Maybe change the ```bootstrapper ``` method name, the error message like this:

----------------------
Here when add contract to bootstrapper, the error message is very pretty:

After implement, There is not error in application.

### DIFF
--- a/src/Illuminate/Contracts/Foundation/IsBootstrapper.php
+++ b/src/Illuminate/Contracts/Foundation/IsBootstrapper.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Illuminate\Contracts\Foundation;
+
+interface IsBootstrapper
+{
+    /**
+     * Bootstrap the given application.
+     *
+     * @param  Application  $app
+     * @return void
+     */
+    public function bootstrap(Application $app);
+}
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -245,7 +245,7 @@ class Application extends Container implements ApplicationContract, CachesConfig
 
                 $this['events']->dispatch('bootstrapped: '.$bootstrapper, [$this]);
             } else {
-                throw new \RuntimeException('The bootstrapper must be implements of IsBootstrapper');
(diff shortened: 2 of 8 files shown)


henzeb:
A. Would potentially break current installations that do have added a bootstrapper. 
B. Let the interface be something like "Bootstrapper" 
C. I suggest the error message to read: "The bootstrapper must implement the 'Bootstrapper' interface"

milwad-dev:
@henzeb 
A.Why?

henzeb:
@milwad-dev, the bootstrappers array is protected, and thus accessible to the childclasses where you can configure your middleware. You can either override the entire property, or you could add extra bootstrappers through a constructor.

milwad-dev:
@henzeb 
Yes, is protected.
My talk is that add a new or modify bootstrapper.
Why make contract? Because the code will be the same and all ```bootstrappers``` are same.

milwad-dev:
This topic refers a little to contract based design

henzeb:
I know about contract based design, but I am worried about the fact that any existing userland implementations without the interface will fail when updating to the latest minor, since you added the explicit check if it's implementing the interface which is in fact a breaking change.

I suggest adding this to 10.x instead.

milwad-dev:
@henzeb 
I think no problem for 10.x :)

taylorotwell:

PR #45962, opened 2023-02-04
Add validatedExcept() method

HungryBus:
Add validatedExcept() method to the Illuminate\Foundation\Http\FormRequest to retrieve all validated data except the given keys


danilopolani:
Wouldn't be simpler to just use `$request->safe()->except('...')` rather than creating a new method? Or, at least, let it be something like:

```php
return $this->validator->safe()->except(Arr::wrap($keys));
```

milwad-dev:
Did you read this? 
https://laravel.com/docs/9.x/requests#retrieving-a-portion-of-the-input-data

taylorotwell:
See @danilopolani's comment.

PR #45964, opened 2023-02-04
[9.x] Add mergeIfMissing method to Collection

xurshudyan:
This method allows you to merge two collections, but only if the elements in the second collection do not already exist in the first collection. This is useful when you want to add new items to a collection, but do not want to overwrite existing items.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -846,7 +846,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     /**
      * Merge the collection with the given items only if the items are missing.
      *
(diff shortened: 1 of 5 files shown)


taylorotwell:
Can't you just do $collection2->merge($collection1);?

xurshudyan:
> Can't you just do $collection2->merge($collection1);?

No, because if a string key in the given items matches a string key in the original collection, the given item's value will overwrite the value in the original collection.
The mergeIfMissing method will not overwrite data.
The same method is available on request instance.

xurshudyan:
> > Can't you just do $collection2->merge($collection1);?
> 
> No, because if a string key in the given items matches a string key in the original collection, the given item's value will overwrite the value in the original collection. The mergeIfMissing method will not overwrite data. The same method is available on request instance.
> 
> image

@taylorotwell

xurshudyan:
@taylorotwell

PR #45965, opened 2023-02-04
Add MessageBag remove method

guizoxxv:
This PR adds the ability to remove an existing message from the MessageBag using the key as parameter.

Scenario: I have an unversioned API that needs to support different parameter names temporarilly. I wanted to use a combination of the [prepareForValidation](https://laravel.com/docs/9.x/validation#preparing-input-for-validation) and [after validation](https://laravel.com/docs/9.x/validation#after-validation-hook) in the FormRequest to achieve this, but I'm unable to replace the validation errors keys.

[Here](https://stackoverflow.com/questions/18592200/remove-message-from-messagebag) is a discussion around this topic.

### DIFF
--- a/src/Illuminate/Contracts/Support/MessageBag.php
+++ b/src/Illuminate/Contracts/Support/MessageBag.php
@@ -23,6 +23,14 @@ interface MessageBag extends Arrayable, Countable
     public function add($key, $message);
 
     /**
+     * Remove a message from the bag.
+     *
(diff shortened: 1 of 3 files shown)


taylorotwell:
Can't add a method to a contract on a point release.

guizoxxv:
OK. Is this something that can be added in another release? If so I could open this PR to master branch. Thanks.

PR #45966, opened 2023-02-05
[10.x] Add a dynamic update method to Eloquent builder

andiLeong:
often not always we need to perform update on one column of a model only
eg:

```
$user->update([ 'active' => true);
$post->update([ 'archive' => true);
$product->update([ 'cat_id' => 99);
```
would it be nice we have dynamic method calling on the update

```
$user->updateActive(true);
$post->updateArchive(true);
$product->updateCatId(99);
```

by upgrading to mthod not only it reads better than the array syntax, also give you a nicer interface.


ankurk91:
I have already stopped using the magic where methods.
Laravel is moving towards strict types.

I would encourage people to find unresolved reported issues in framework on GitHub and send PR to solve them.

taylorotwell:

PR #45969, opened 2023-02-05
[10.x] Add public path to application contract

jnoordsij:
Requires #45968 to be merged first.

This MR adds the `publicPath` function the the `Application` contract for consistency with other path-related helpers. It also updates the global helper function to use this function.

### DIFF
--- a/src/Illuminate/Contracts/Foundation/Application.php
+++ b/src/Illuminate/Contracts/Foundation/Application.php
@@ -54,6 +54,14 @@ interface Application extends Container
     public function langPath($path = '');
 
     /**
+     * Get the path to the public directory.
+     *
+     * @param  string  $path
+     * @return string
+     */
+    public function publicPath($path = '');
+
+    /**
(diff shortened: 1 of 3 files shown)


jnoordsij:
Tests suggest I've hit a weird spot here: `path.public` can be overridden as singleton in the `Applications` class, which does (before this MR) change the behavior of the `public_path` helper but not the `publicPath` function; this seems somewhat inconsistent. Moreover, there's various other paths that can be set, with some of them ignored in both the helper and `Application` function (e.g. `path.config`) and some of them being used in both places (e.g. `path.storage`).

Looking for a bit of input here, but I think the best take would be to create a `publicPath` property and setter, similar to those for `config`, and use those consistently.

jnoordsij:
I've limited this MR to just adding the method to the contract and moving the function definition to inside the `Application` class; working on consistency of path overriding in https://github.com/jnoordsij/laravel-framework/tree/rework-application-public-path-helper.

PR #45976, opened 2023-02-06
[10.x] `offsetExists` fetches attribute and not only checks

olivernybroe:
## Problem
When calling `offsetExists` on a model, it will call the underlying attribute and actually get the value. In the case where the getter is a bit more heavy this can result in unexpected extra load. `offsetGet` should be the method that actually fetches the attribute. 

```php
$model = new Model();

if(isset($model['name']) {
    return $model['name'];
}
```
The following code will result in two calls to the underlying attribute getter.  
This is mostly a problem when utilizing third party packages which accepts an `ArrayAccess` object and relies on `isset` logic. These packages now need custom logic for handling Laravel (for better performance), instead of just relying on the expected behaviour of `ArrayAccess`.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -2227,16 +2227,35 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     /**
      * Determine if the given attribute exists.
      *
+     * In the case of an attribute, it will fetch the underlying attribute
+     * to check if the value of it is `null`. Relationships will not be
+     * fetched.
+     *
      * @param  mixed  $offset
      * @return bool
      */
     public function offsetExists($offset): bool
     {
-        try {
-            return ! is_null($this->getAttribute($offset));
-        } catch (MissingAttributeException) {
+        if (! $offset) {
             return false;
         }
+
+        // Checks if the attribute exists.
+        if (array_key_exists($offset, $this->attributes) ||
+            array_key_exists($offset, $this->casts) ||
+            $this->hasGetMutator($offset) ||
+            $this->hasAttributeMutator($offset) ||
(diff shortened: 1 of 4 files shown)


driesvints:
There seems to be a specific test for this situation which is now breaking which leads me to believe the current behavior is intended.

taylorotwell:
Please mark as ready for review when you have a code change. Thanks.

olivernybroe:
@taylorotwell I have submitted a potential fix. 

My solution does include a breaking change as isset check on a model where the value of the attribute is `null` will now result in `true` instead of `false` (hence why I am targeting next major). 


This solution could be rewritten to not include a breaking change, but with the tradeoff that checking if an attribute exists on a model will still result in the mutator being called. However it would still fix the problem of doing `isset` on a relationship, so it won't actually run database queries on `isset` (unless you do it in inside your mutator).


I am proposing doing it with the breaking change as I think that is the most optimal solution, but if you would prefer the other solution, I can rewrite.

taylorotwell:
It seems like `isset` returning `true` on `null` is not desired? From the PHP docs on `isset`: "Determine if a variable is declared and **is different than null**."

spawnia:
I think the semantics of `isset()` pretty much force the existing implementation. It might just be impossible to optimize the code snippet you provided as an example.

However, I could see the addition of a method that contains the checks you put into `offsetExists()` here. Perhaps called `hasProperty()`?

olivernybroe:
@spawnia the improvements can still me made for relationships at least :) I will revert the attribute getters

spawnia:
So `isset()` for a relation that is defined but will return `null` when fetched will be `true`?

olivernybroe:
@spawnia you are right, I was only thinking about many to many relationships. but of course with belongs to relations it will be wrong. 

Closing this PR as my solution doesn't actually work.

PR #45982, opened 2023-02-06
[9.x] Add uri templates to `Http` client

jbrooksuk:
This PR introduces the ability to use URI templates on the `Http` client. We pull in `guzzlehttp/uri-templates` which follows the spec at https://www.rfc-editor.org/rfc/rfc6570

```php
Http::withUrlParameters([
    'endpoint' => 'https://laravel.com',
    'page' => 'docs',
    'version' => '9.x',
    'thing' => 'validation',
])->get('{+endpoint}/{page}/{version}/{thing}');

// https://laravel.com/docs/9.x/validation
```

This is using [`guzzlehttp/uri-template`](https://github.com/guzzlehttp/uri-template) behind the scenes, which provides a much wider set of support features when expanding templates.


src/Illuminate/Http/Client/PendingRequest.php
```diff
@@ -715,7 +736,7 @@ public function post(string $url, $data = [])
      * @param  array  $data
      * @return \Illuminate\Http\Client\Response
      */
-    public function patch($url, $data = [])
+    public function patch(string $url, $data = [])
```
jbrooksuk, line 739: The other methods were already type-hinted, so I updated these too. Let me know if this is wrong though.

src/Illuminate/Http/Client/PendingRequest.php
```diff
@@ -439,6 +447,19 @@ public function withUserAgent($userAgent)
...
+     * Specify the URL parameters that can be substituted into the request URL.
+     *
+     * @param  array  $parameters
+     * @return $this
+     */
+    public function withUrlParameters(array $parameters = [])
+    {
+        return tap($this, function () use ($parameters) {
+            $this->urlParameters = $parameters;
+        });
```
mathieutu, line 460: Is it the same than writing

```suggestion
        $this->urlParameters = $parameters;

        return $this;
```
Or am I missing something?

src/Illuminate/Http/Client/PendingRequest.php
```diff
@@ -785,6 +806,8 @@ public function send(string $method, string $url, array $options = [])
             $url = ltrim(rtrim($this->baseUrl, '/').'/'.ltrim($url, '/'), '/');
         }
 
+        $url = $this->expandUrlParameters($url);
```
crissi, line 809: would it make sense to validate the params, to make sure it match the url template?

timacdonald:
This is sick 🔥

milewski:
For those who are interested in understanding the syntax for this templating, the information can be found at https://www.rfc-editor.org/rfc/rfc6570.

Tvup:
@jbrooksuk 
What will this do if you have a url like
`http://endpoint.top/model?filter={"PriceType":"Expensive"}`
?

jbrooksuk:
@Tvup good question, with multiple answers.

1. If you're not composing a `Http` request with URL parameters, then nothing different happens.
2. `withUrlParameters` will only expand on defined parameters. Since you're not doing anything with URL parameters, nothing changes.
3. You should really move the filter into the second argument, `Http::get('http://endpoint.top/model?', ['filter' => json_encode(["PriceType" => "Expensive"])]);`.

GrahamCampbell:
You should not include the question mark on the end there, in bullet v3. It is added for you.

```
Http::get('http://endpoint.top/model', ['filter' => json_encode(['PriceType' => 'Expensive'])]);
```

jbrooksuk:
Whoops, typo!

PR #45983, opened 2023-02-06
[9.x] Added ability to assert that event Listeners were pushed to Queue

asantibanez:
## Summary

This PR allows asserting that event listeners are pushed into the queue. Currently, `Event::assertListening()` exists for this but this allows the developer to ensure that the listener was pushed into the queue.

`Queue::assertListenerPushed` allows asserting if listener was pushed, if it was pushed X times, and even accepts a truth-test callback similar to `Queue::assertPushed`

### DIFF
--- a/src/Illuminate/Support/Testing/Fakes/QueueFake.php
+++ b/src/Illuminate/Support/Testing/Fakes/QueueFake.php
@@ -5,6 +5,7 @@ namespace Illuminate\Support\Testing\Fakes;
 use BadMethodCallException;
 use Closure;
 use Illuminate\Contracts\Queue\Queue;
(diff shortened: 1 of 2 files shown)


taylorotwell:
I'm not sure I see the value of this test specifically? You can test if the listener is attaching to the event - as long as the listener has the `ShouldQueue` interface it will be queued. Am I wrong?

asantibanez:
Hi @taylorotwell 

When a Listener is queued, the listener class itself is not queued but instead a CallQueuedListener class which holds the Listener object. My reasoning is to simplify assertions when testing Listeners we want queued and not directly dispatched. 

For directly dispatched, we can use the Event::assertListening helper. 

Hope it makes sense.

PR #45988, opened 2023-02-06
[10.x] Passthru missing methods on `EventFake` & `MailFake` for macro accessibility

### LINKED ISSUE #45986: [9.x] Facade Fakes lose macros
- Laravel Version: v9.50.2 
- PHP Version: v8.1.14

### Description:

I'm attempting to extend Laravel's `Event` facade with additional macros for an application feature.

However, during testing, these macros are lost and cannot be called after calling `Event::fake()`, due to the instance being swapped out with a new implementation.

### Steps To Reproduce:

For example:

```php
Event::macro('bundle', fn (Bundleable $event) => (
    Bundle::add($event)
));

Event::macro('assertBundled', fn ($event) => (
    Bundle::assertExists($event)
));
```

In the test:

```php
Event::fake(SomeEventToFake::class);

// Do something...

Event::assertDispatched(SomeEventToFake::class);

Event::assertBundled(SomeOtherEvent::class); // Throws exception
```

### Solution

With all "fake" implementations, we could forward missing method calls back onto their root instances that are `Macroable`, which allows the macros to be accessible:

```php
// src/Illuminate/Support/Testing/Fakes/EventFake.php

class EventFake implements Dispatcher
{
    use ForwardsCalls;

    // ...

    /**
     * Handle dynamic method calls to the dispatcher.
     *
     * @param  string  $method
     * @param  array  $parameters
     * @return mixed
     */
    public function __call($method, $parameters)
    {
        return $this->forwardCallTo($this->dispatcher, $method, $parameters);
    }
}
```

I can submit a PR with this added, but wanted to get everyones thoughts on this, and whether this behaviour is intentional and whether I'm using macros in an incorrect way.

Thanks so much for your time! ❤️

stevebauman:
Closes #45986

This PR implements `ForwardsCalls` trait onto the `EventFake` and `MailFake` classes  so that macro'd methods are still accessible after calling `Event::fake()` or `Mail::fake()`.

Below are the list of facades that have the capability of being faked, along with their status of their underlying implementations supporting macro calls:

| Facade | Supports `macro`'s on fakes |
| --- | --- |
| `Illuminate\Support\Facades\Event` | ✅ (in this PR) |
| `Illuminate\Support\Facades\Http` | ✅  |
| `Illuminate\Support\Facades\Mail` | ✅ (in this PR)  |
| `Illuminate\Support\Facades\Storage` | ✅ |

Let me know your thoughts about this, and thanks so much for your time! ❤️

### DIFF
--- a/src/Illuminate/Support/Facades/Mail.php
+++ b/src/Illuminate/Support/Facades/Mail.php
@@ -65,7 +65,7 @@ class Mail extends Facade
      */
     public static function fake()
     {
-        static::swap($fake = new MailFake);
+        static::swap($fake = new MailFake(static::getFacadeRoot()));
 
         return $fake;
     }
--- a/src/Illuminate/Support/Facades/Mail.php
+++ b/src/Illuminate/Support/Facades/Mail.php
@@ -65,7 +65,7 @@ class Mail extends Facade
      */
     public static function fake()
     {
-        static::swap($fake = new MailFake(static::getFacadeRoot()));
+        static::swap($fake = new MailFake);
 
         return $fake;
     }
--- a/src/Illuminate/Support/Testing/Fakes/EventFake.php
+++ b/src/Illuminate/Support/Testing/Fakes/EventFake.php
@@ -13,8 +13,7 @@ use ReflectionFunction;
 
 class EventFake implements Dispatcher
 {
-    use ForwardsCalls;
-    use ReflectsClosures;
+    use ForwardsCalls, ReflectsClosures;
 
     /**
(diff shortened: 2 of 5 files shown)


ralphjsmit:
Like it! Would it make sense to still also include the Mail fake before the L10 release? I don't think too many people will manually instantiate fake mailers, so the impact of adding the constructor shouldn't be too big imho.

driesvints:
This seems to break the test suite on Jetstream. Given that we're so close to a release, I feel we at least should wait until Laravel v11.

PR #45991, opened 2023-02-07
[9.x] Removes duplicate returns in tests for pending has-many-through

rbarden:
Coming from #45894

There are multiple returns, and based on the test name and the co-tested method on line 190, I believe this was the intended behavior.


timacdonald:
Thanks @rbarden. I was testing the tests on that one and forgot to correct it.

Appreciate you fixing that up.

PR #45992, opened 2023-02-07
[9.x] fix groupBy and keyBy by Enum field

mistersaal:
REF #42507 (just fixed it like it was said in this PR)

When doing groupBy or keyBy with enum field that casted from a model, it returns an enum instance. It will cause an error 

> array_key_exists(): Argument #1 ($key) must be a valid array offset type.

for example like below
```php
enum CustomerStatus: string
{
    case Bussines = 'bussines';
    case Private = 'private';
}
class Customer  extends Model {
    protected $casts = [
    ];
}
```
then when we do
`Customer::all()->groupBy('status')`
it will cause an error
>array_key_exists(): Argument #1 ($key) must be a valid array offset type

We need to workaround with enum for example `Customer::all()->groupBy('status.value')` and `Customer::all()->keyBy('status.value')`
This PR will make accessing value of an enum directly like `Customer::all()->groupBy('status')` and `Customer::all()->keyBy('status')`

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -507,6 +507,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
                 $groupKey = match (true) {
                     is_bool($groupKey) => (int) $groupKey,
                     $groupKey instanceof \Stringable => (string) $groupKey,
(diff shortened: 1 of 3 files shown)


taylorotwell:
I dunno - enums are objects so I think it's find to just require users to be explicit about what they want to do.

PR #45996, opened 2023-02-07
[9.x] Adding Not Die and Dump option => without ending script

Nejcc:
Sometimes you may want to inspect the value of a variable in your code, but you don't want the script execution to stop immediately after. In such cases, you can use a new alias ndd (for "no die") which is a shorthand for the dump function in Laravel.

For example:

```php

$variable = [1, 2, 3];
ndd($variable);
```

This will output the contents of $variable ([1, 2, 3]) to the browser, but the script execution will continue.

Using an alias for dd or dump allows you to quickly switch between them depending on your debugging needs, without having to change the code each time.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -621,6 +621,17 @@ trait InteractsWithInput
     }
 
     /**
+     * Dump the request items and not end the script.
+     *
+     * @param  mixed  ...$keys
+     * @return never
+     */
+    public function ndd(...$keys)
+    {
+        $this->dump(...$keys);
+    }
+
+    /**
      * Dump the items.
      *
      * @param  mixed  $keys
--- a/src/Illuminate/Routing/Console/stubs/controller.model-prepared.stub
+++ b/src/Illuminate/Routing/Console/stubs/controller.model-prepared.stub
@@ -1,0 +1,116 @@
+<?php
+
+namespace {{ namespace }};
+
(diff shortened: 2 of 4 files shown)


henzeb:
we have dump() for that.

Nejcc:
> we have dump() for that. and i don´t see the need for a shorthand. dumpanddie, yes, but not for dump.

Well, that is true. I often see people using dd() to check something and exit the script, but the problem arises when they start using print_r() and then var_dump() because they don't want the script to be terminated, but instead they want to display all data while the script continues to execute to the end

Mausy1995:
I use this all the time in helpers.php, but i have it included in composer:
```json
    "files": [
            "app/helpers.php"
        ]
```        
All the time i need to do this in every project:
```php
if (!function_exists('ndd')) {
    /**
     * @param mixed ...$vars
     */
    function ndd(...$vars)
    {
        foreach ($vars as $v) {
            VarDumper::dump($v);
        }
    }
}
```

taylorotwell:

Nejcc:
Thank you  for your input @taylorotwell i think many users including myself this tiny features have great potential and a lot of programmers  make sort of their own implementations. this small but usefull feature dont break any code it's just an shortcut of dump and is very easy when you do dd(...) and add in front N or just take back N to toggle between (die and dump or just dump).  

Sure i will make separate package it this you consider that will not be part of the core.  As someone in the comment mention it and past his way of implementation. the code not breaking any rules as far i understand. Please prove me wrong.

PR #45998, opened 2023-02-07
[9.x] Remove deprecation warnings in test output

driesvints:
Right now, with the latest PHPUnit v9 update, we get some deprecation messages in our test output. I've resolved these so the output stays clean.

Screenshot 2023-02-07 at 13 55 02
Screenshot 2023-02-07 at 13 54 58

### DIFF
--- a/tests/Events/EventsDispatcherTest.php
+++ b/tests/Events/EventsDispatcherTest.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Tests\Events;
 
+use Error;
 use Exception;
 use Illuminate\Container\Container;
 use Illuminate\Events\Dispatcher;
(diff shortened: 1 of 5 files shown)


tests/Events/EventsDispatcherTest.php
```diff
@@ -571,13 +571,6 @@ public function testInvokeIsCalled()
         $d->dispatch('myEvent', 'somePayload');
         $this->assertEquals(['__construct', '__invoke_somePayload'], $_SERVER['__event.test']);
 
-        // It throws an "Error" when there is no method to be called.
-        $d = new Dispatcher;
-        $d->listen('myEvent', TestListenerLean::class);
-        $this->expectError();
-        $this->expectErrorMessage('Call to undefined method '.TestListenerLean::class.'::__invoke()');
-        $d->dispatch('myEvent', 'somePayload');
-
```
driesvints, line 580: Unfortunately we needed to remove this test because there's no way to assert Error exceptions anymore in PHPUnit. See https://github.com/sebastianbergmann/phpunit/issues/5062

driesvints: No because it would still throw the warning: https://github.com/sebastianbergmann/phpunit/blob/9.6/src/Framework/TestCase.php#L607

Expecting Error, Warning, Notice or Deprecation is simply not possible anymore.

henzeb: It's a bit hacky maybe, but you could override the exception handler and do-it-yourself.

driesvints: I don't want to go there. Think it's best to follow PHPUnit's defaults as much as possible.

driesvints: Nvm, seems I was looking at a wrong error type.

nunomaduro (approved):

PR #45999, opened 2023-02-07
[10.x] - Adding prepared Controller Resource stub as an alternative

Nejcc:
When you add a prepared resource controller stub, it helps you to quickly set up a basic CRUD (Create, Read, Update, Delete) functionality for a resource. This can save you time and effort as you don't have to write the basic logic and routes for each action yourself.

### DIFF
--- a/src/Illuminate/Http/Concerns/InteractsWithInput.php
+++ b/src/Illuminate/Http/Concerns/InteractsWithInput.php
@@ -621,6 +621,17 @@ trait InteractsWithInput
     }
 
     /**
+     * Dump the request items and not end the script.
+     *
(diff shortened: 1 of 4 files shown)


henzeb:
This commit now also exists in your other PR. Just got a message about that.

Nejcc:
> This commit now also exists in your other PR. Just got a message about that.

Yeah i push on both better to do just in 9 ?

henzeb:
> > This commit now also exists in your other PR. Just got a message about that.
> 
> Yeah i push on both better to do just in 9 ?

Better one pull request per subject. And I guess if it's in 9 it will be there in 10.

PR #46004, opened 2023-02-07
[10.x] Check reportCallbacks before dontReport

tontonsb:
This is a bit opinionated, but it seems natural to expect that explicitly registered listeners will work:

```php
// App\Exceptions\Handler
public function register()
{
    $this->reportable(fn(ModelNotFoundException $e) => $this->log404s($e));
}
```

Currently this callback would only be called if the developer overrides `$internalDontReport` or `shouldntReport()` neither of which is documented (which would be fine IMO).

The proposed change would make `$internalDontReport` only suppress the default reporting, but additional callbacks would work as documented.

### DIFF
--- a/src/Illuminate/Foundation/Exceptions/Handler.php
+++ b/src/Illuminate/Foundation/Exceptions/Handler.php
@@ -243,6 +243,12 @@ class Handler implements ExceptionHandlerContract
     {
         $e = $this->mapException($e);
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Rather than make the breaking change, added this convenient method which we can document: https://github.com/laravel/framework/commit/24e0fb839b0531553732d636263d4c6acd984325

PR #35139, opened 2020-11-06
[8.x] Add lock support for file and null cache drivers

paras-malhotra:
This PR adds a generic cache lock class that can implement a lock with any cache store implementation (kind of a bridge that takes a cache store and returns a cache lock). With this PR, we can now support locks for all cache drivers (including file, apc and null).

Laravel uses locks for several things including session blocking, `WithoutOverlapping` job middleware and unique jobs. This PR allows users to use these features for all cache drivers.

PR #46011, opened 2023-02-07
[9.x] Fix race condition in locks issued by the file cache driver

### LINKED ISSUE #43627: Atomic locks do not provide reliable mutual exclusion when using the file cache driver

bytestream:
Fixes https://github.com/laravel/framework/issues/43627

Before:

```bash
root@fa0e2058d0ad:/app# php artisan app:test-locks --workers=2 --iterations=1000 --cache-driver=file
Running 1000 iterations with 2 workers using the cache driver 'file' for locking.
Process 1/2 started.
Process 2/2 started.
Process 1/2 finished.
Process 2/2 finished.
Process 1/2 started at 0 and ended at 1938.
Process 2/2 started at 0 and ended at 1932.
Counted only to 1938 instead of the expected 2000.
```

After:

```bash
root@fa0e2058d0ad:/app# php artisan app:test-locks --workers=2 --iterations=1000 --cache-driver=file
Running 1000 iterations with 2 workers using the cache driver 'file' for locking.
Process 1/2 started.
Process 2/2 started.
Process 1/2 finished.
Process 2/2 finished.
Process 1/2 started at 0 and ended at 1922.
Process 2/2 started at 26 and ended at 2000.
Successfully counted to 2000.

```

### DIFF
--- a/src/Illuminate/Cache/CacheLock.php
+++ b/src/Illuminate/Cache/CacheLock.php
@@ -34,7 +34,9 @@ class CacheLock extends Lock
      */
     public function acquire()
     {
-        if (method_exists($this->store, 'add') && $this->seconds > 0) {
+        if ($this->store instanceof FileStore
+            || (method_exists($this->store, 'add') && $this->seconds > 0)
+        ) {
             return $this->store->add(
                 $this->name, $this->owner, $this->seconds
             );
--- a/src/Illuminate/Cache/CacheLock.php
+++ b/src/Illuminate/Cache/CacheLock.php
@@ -34,9 +34,7 @@ class CacheLock extends Lock
      */
     public function acquire()
     {
-        if ($this->store instanceof FileStore
-            || (method_exists($this->store, 'add') && $this->seconds > 0)
-        ) {
+        if (method_exists($this->store, 'add') && $this->seconds > 0) {
             return $this->store->add(
                 $this->name, $this->owner, $this->seconds
             );
--- a/src/Illuminate/Cache/FileLock.php
+++ b/src/Illuminate/Cache/FileLock.php
@@ -5,29 +5,6 @@ namespace Illuminate\Cache;
 class FileLock extends CacheLock
 {
     /**
-     * The cache store implementation.
-     *
-     * @var FileStore
-     */
-    protected $store;
-
-    /**
-     * Create a new lock instance.
-     *
-     * @param  FileStore  $store
-     * @param  string  $name
-     * @param  int  $seconds
-     * @param  string|null  $owner
-     * @return void
-     */
-    public function __construct(FileStore $store, $name, $seconds, $owner = null)
-    {
-        parent::__construct($name, $seconds, $owner);
-
-        $this->store = $store;
-    }
-
-    /**
      * Attempt to acquire the lock.
      *
      * @return bool
--- a/src/Illuminate/Cache/FileLock.php
+++ b/src/Illuminate/Cache/FileLock.php
@@ -1,0 +1,39 @@
+<?php
+
+namespace Illuminate\Cache;
+
+class FileLock extends CacheLock
+{
+    /**
+     * The cache store implementation.
+     *
+     * @var FileStore
+     */
+    protected $store;
+
+    /**
+     * Create a new lock instance.
+     *
+     * @param  FileStore  $store
+     * @param  string  $name
+     * @param  int  $seconds
+     * @param  string|null  $owner
+     * @return void
+     */
+    public function __construct(FileStore $store, $name, $seconds, $owner = null)
+    {
+        parent::__construct($name, $seconds, $owner);
+
+        $this->store = $store;
+    }
+
+    /**
+     * Attempt to acquire the lock.
+     *
+     * @return bool
+     */
+    public function acquire()
+    {
+        return $this->store->add($this->name, $this->owner, $this->seconds);
+    }
+}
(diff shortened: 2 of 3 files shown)


taylorotwell:
Please explain how this changes anything? The FileStore already has an `add` method so it would already be falling into this conditional block anyways?

bytestream:
@taylorotwell the problem is when `$this->seconds = 0` it falls to the `get` / `put` combination which isn't atomic. The `add` method in the file driver is able to handle when `$seconds = 0`, and treats it as cache forever

bytestream:
@taylorotwell [362fc83](https://github.com/laravel/framework/pull/46011/commits/362fc838c060aec78d992b3538fb4ae5c20e95f9) is a cleaner way of handling it while maintaining BC

taylorotwell:
Are other drivers not able to handle `$seconds = 0`?

taylorotwell:
I also don't agree your second commit is necessarily better. The first was simpler?

bytestream:
@taylorotwell `CacheLock` and `HasCacheLock` are only actually used by the file driver. They're a legacy from https://github.com/laravel/framework/pull/35139 which originally added APC driver support as well, but was removed before it was merged. The simplest fix may just be the one that was originally suggested in https://github.com/laravel/framework/pull/43661. If it were me though, I'd remove the `HasCacheLock` and rename `CacheLock` to `FileLock`. That's clearer and also follows the way other drivers are implemented.

taylorotwell:
Sorry - I'm not following at all but I don' have time to dig deeper atm. 

1) Can other drivers not handle $seconds = 0? 

2) Why is this not the simplest fix and easier?

CleanShot 2023-02-09 at 16 55 37@2x

taylorotwell:
For what it's worth: I've updated the docs to not even show an example of not providing an expiration time (in seconds) to the lock methods.

bytestream:
> Can other drivers not handle $seconds = 0?

TLDR; not relevant as `CacheLock` is **only** used by the `FileStore`. To directly answer the question though:

* APC - :x: , missing `add` method
* Array - :x: , missing `add` method
* Database - :ok: , providing `$seconds = 0` to `add()` results in immediate expiry, but `DatabaseLock` uses expiry of 1 day (86400 seconds) if `$seconds = 0`
* DynamoDB - :ok: , providing `$seconds = 0` to `add()` results in immediate expiry, but `DynamoDbLock` uses expiry of 1 day (86400 seconds) if `$seconds = 0`
* File - :ok: , expiry is set to `9999999999`
* Memcached - :ok: , `0` means never expire - https://www.php.net/manual/en/memcache.add.php
* Null - :x: , missing `add` method
* Redis - :ok: , providing `$seconds = 0` to `add()` expires after 1 second, but `RedisLock` correctly handles `$seconds = 0`

> Why is this not the simplest fix and easier?

Simplest fix in the sense that it's a 1 line change, but given `CacheLock` is **only** used by the `FileStore` it really makes sense to refactor the usage. https://github.com/laravel/framework/pull/43661 is the simplest fix IMO without any additional refactoring, and **has no breaking changes / consequences.**

Issue #38259, opened 2021-08-06
There is no active transaction error in PHPUnit


rome2o:
- Laravel Version: 8.12
- PHP Version: 8.0.1
- Database Driver & Version:
MySQL 8.0.22

### Description:
We are writing SQL test cases for our application and are using DatabaseTransactions in our BaseTestCase.php

The setUp function looks like this

```
    protected function setUp(): void
    {
        parent::setUp();
    }
```

A sample test case is as the following

```
    public function product_with_required_quantitiy_is_returned_in_service_code()
    {
            \App\Models\Inventory::create([
                'product_variant_id'=> $id,
                'warehouse_id'=> $warehouse,
                'qty_onhand'=> $qty,
                'qty_available'=> $qty,
            ]);
            $response = $this->call('POST', route('ecommerce.checkout'), $payload);
            $response
            ->assertStatus(200)
            ->assertJson(fn (AssertableJson $json) => $json->has('rates', 7));

            $data = $response->json();
            $codeDetails = (new \App\Models\ServiceCode($data['rates'][0]['service_code']))->getDetails();
            $this->assertFalse($codeDetails['makeRequest']);
    }
```

However upon execution of the test cases we get the following error

  There is no active transaction

```
  at vendor/laravel/framework/src/Illuminate/Database/Concerns/ManagesTransactions.php:279
    275▕      */
    276▕     protected function performRollBack($toLevel)
    277▕     {
    278▕         if ($toLevel == 0) {
  ➜ 279▕             $this->getPdo()->rollBack();
    280▕         } elseif ($this->queryGrammar->supportsSavepoints()) {
    281▕             $this->getPdo()->exec(
    282▕                 $this->queryGrammar->compileSavepointRollBack('trans'.($toLevel + 1))
    283▕             );
```

Any help would be much appreciated

Issue #36666, opened 2021-03-19
[Parallel testing] PDOException: There is already an active transaction


ChoKaPeek:
- Laravel Version: 8.33.1
- PHP Version: 8.0.3
- Database Driver & Version: MySQL 5.7.32

### Description:
Greetings, we want to test multiple database interactions, so we need isolated tests and a way to create / migrate / seed replicas of our mysql database (disk storage). We are using Sanctum for authentication.

I've been trying to use the new parallel testing functionality and read your different docs and issues about it.

My conclusions are:

- the DatabaseTransactions trait is needed, the others are "replaced" in whole or in part by ParallelTesting
- the new ParallelTesting::setUpTestDatabase() is exactly what I need to seed the databases at creation
- no test db definition required as ParallelTesting will already create copies of the form name_test_1, name_test_2
- without --recreate-databases, seeding won't be called again as it is only called after its test db is created

My seeds are a bunch of factories, DB::unprepared.

Running `art migrate:fresh` and `art db:seed` manually then `art test` without `--parallel`, tests work fine. However when running with `--parallel --recreate-databases`, the seeding does not work and I stumble upon this error:

> PDOException: There is already an active transaction
```
EE                                                                  2 / 2 (100%)

Time: 00:14.187, Memory: 20.00 MB

There were 2 errors:

1) Tests\Feature\AuthTest::testAuthLogin
PDOException: There is already an active transaction

[root]/vendor/laravel/framework/src/Illuminate/Database/Concerns/ManagesTransactions.php:137
[root]/vendor/laravel/framework/src/Illuminate/Database/Concerns/ManagesTransactions.php:113
[root]/vendor/laravel/framework/src/Illuminate/Foundation/Testing/DatabaseTransactions.php:21
[root]/vendor/laravel/framework/src/Illuminate/Foundation/Testing/TestCase.php:129
[root]/vendor/laravel/framework/src/Illuminate/Foundation/Testing/TestCase.php:90

2) Tests\Feature\AuthTest::testAuthGetUser
Illuminate\Database\QueryException: SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'company_id' cannot be null (SQL: insert into `users` (`company_id`, [cut]))

[root]/vendor/laravel/framework/src/Illuminate/Database/Connection.php:678
[root]/vendor/laravel/framework/src/Illuminate/Database/Connection.php:638
[root]/vendor/laravel/framework/src/Illuminate/Database/Connection.php:472
[...]

Issue #35380, opened 2020-11-26
PHP 8.0 PDO::rollback() now throws PDOException after implicit commit


stemis:
- Laravel Version: 8.x
- PHP Version: 8.0RC5
- Database Driver & Version: MariaDB 

### Description:

As this is a difficult subject I will try my best to explain it properly, it is an issue that needs to be addressed properly.

#### PHP 8.0 changes
In PHP 8.0, the pdo transaction logic was rewritten to some extend.
When issuing a PDO::rollback(), this will now throw a PDOException if the transaction was already closed.

_* I was not able to find any official documentation on this yet, but was able to confirm this with the different PHP versions_
#### Background Implicit Commits with Transactions
MySQL has a thing called [implicit commits](https://dev.mysql.com/doc/refman/8.0/en/implicit-commit.html), to demonstrate what this does I have created the following example. I have used MariaDB as this has access to the `in_transaction` session var to demonstrate the working.

```sql

create table Table1
(
    name varchar(10) null
);

START TRANSACTION;

SHOW VARIABLES WHERE Variable_name = 'in_transaction' # Returns 1

INSERT INTO Table1 (name) values ('before');

SHOW VARIABLES WHERE Variable_name = 'in_transaction' # Returns 1

create table Table2
(
    name varchar(10) null
);

SHOW VARIABLES WHERE Variable_name = 'in_transaction' # Returns 0

INSERT INTO Table1 (name) values ('after');

SHOW VARIABLES WHERE Variable_name = 'in_transaction' # Returns 0
ROLLBACK ;

# Table 1 now contains both rows.
```
So imagine a UnitTest executing any Implicit commit commands, then it commits the transaction. So when running Migration command in a test, or stored procedures, or seeders that have implicit commands in them, a rollback is no longer possible as the transaction was already closed.
In php <=7.4.12 no exception was thrown, so most of the times, you would have no idea why you have data leakage in your test
PHP 8.0 now throwns an exception.

#### Laravel
**How does this affect Laravel?**

1. Many applications use DatabaseTransactions or RefreshDatabase traits, which no longer work when implicit commits have been issued in a test.
2. When a rollback is happening with `DB::transaction`, a new `PDOException` is thrown.

Consider the following test file:

```php
use Illuminate\Foundation\Testing\DatabaseTransactions;
use Illuminate\Support\Facades\DB;
use Tests\TestCase;

class MyTest extends TestCase
{
    use DatabaseTransactions;
    
    public function testMe()
    {
[...]

Issue #35161, opened 2020-11-09
[8.x] After upgrade to PHP 8.0 transactions stop to work on seeders


eusonlito:
- Laravel Version: 8.13.0
- PHP Version: 8.0.0RC3
- Database Driver & Version: MySQL 8.0.22

### Description:

Executing seeders with transactions on PHP 8.0 are failing with exception:

```
$> php8.0 artisan migrate:refresh && php8.0 artisan db:seed --class=Database\\Seeds\\Configuration

Migration table not found.
Migration table created successfully.
Migrating: 2020_02_05_000000_base
Migrated:  2020_02_05_000000_base (1,408.24ms)

   PDOException 

  There is no active transaction

  at vendor/laravel/framework/src/Illuminate/Database/Concerns/ManagesTransactions.php:44
     40▕             }
     41▕ 
     42▕             try {
     43▕                 if ($this->transactions == 1) {
  ➜  44▕                     $this->getPdo()->commit();
     45▕                 }
     46▕ 
     47▕                 $this->transactions = \max(0, $this->transactions - 1);
     48▕             } catch (Throwable $e) {

      +3 vendor frames 
  4   database/Seeds/Configuration.php:17
      Illuminate\Support\Facades\Facade::__callStatic()

      +22 vendor frames 
  27  artisan:37
      Illuminate\Foundation\Console\Kernel::handle()
```

Same command with PHP 7.4 is working as expected:

```
$> php7.4 artisan migrate:refresh && php7.4 artisan db:seed --class=Database\\Seeds\\Configuration

Migration table not found.
Migration table created successfully.
Migrating: 2020_02_05_000000_base
Migrated:  2020_02_05_000000_base (1,432.37ms)
Database seeding completed successfully.
```

### Steps To Reproduce:

1. Create any seeder like:

```php
<?php declare(strict_types=1);

namespace Database\Seeds;

use Illuminate\Database\Seeder;
use Illuminate\Support\Facades\DB;

class Configuration extends Seeder
{
    /**
     * @return void
     */
    public function run()
    {
        DB::transaction(function () {
            DB::table('configuration')->truncate();
        });
    }
}
```

2. Execute seeder

eusonlito:
Some fast debug:

`src/Illuminate/Database/Concerns/ManagesTransactions.php`:
Method: `createTransaction`
Line: `126`

```php
dump($this->getPdo()->beginTransaction());
```

```
true
```

[...]

PR #46017, opened 2023-02-07
[9.x] fix pdo exception when rollbacking without active transaction

khalyomede:
This PR prevents an issue when trying to rollback when there is no active database transaction. 

I faced this issue when I converted my tests suites (which uses `LazilyRefreshDatabase` trait) from SQLite (in memory) to MySQL, which seems to behave differently regarding the transaction verification (SQLite is more permissive).


khalyomede:
This may fix:

- https://github.com/laravel/framework/issues/38259
- https://github.com/laravel/framework/issues/36666
- https://github.com/laravel/framework/issues/35380
- https://github.com/laravel/framework/issues/35161

PR #46018, opened 2023-02-08
[11.x] Add MessageBag remove method

guizoxxv:
This PR adds the ability to remove an existing message from the MessageBag using the key as parameter.

Scenario: I have an unversioned API that needs to support different parameter names temporarilly. I wanted to use a combination of the [prepareForValidation](https://laravel.com/docs/9.x/validation#preparing-input-for-validation) and [after validation](https://laravel.com/docs/9.x/validation#after-validation-hook) in the FormRequest to achieve this, but I'm unable to replace the validation errors keys. Although I was able to solve the situation in a different way, I think that it would be nice to have this ability.

[Here](https://stackoverflow.com/questions/18592200/remove-message-from-messagebag) is a discussion around this topic.

### DIFF
--- a/src/Illuminate/Contracts/Support/MessageBag.php
+++ b/src/Illuminate/Contracts/Support/MessageBag.php
@@ -23,6 +23,14 @@ interface MessageBag extends Arrayable, Countable
     public function add($key, $message);
 
     /**
+     * Remove a message from the bag.
+     *
+     * @param  string  $key
+     * @return $this
+     */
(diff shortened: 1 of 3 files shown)


ankurk91:
It can target v10 branch

taylorotwell:
See above.

guizoxxv:
I opened the PR to branch `10.x` [here](https://github.com/laravel/framework/pull/46043).

PR #46032, opened 2023-02-08
[9.x] Add whereNotInstanceOf method to Collections

xurshudyan:
This PR introduces a new method whereNotInstanceOf for Laravel Collections. The method filters the items in the collection and removes any items that match the given type(s), providing a convenient way to exclude certain types from a collection. This method is the inverse of the existing whereInstanceOf method in Laravel.

## Example

``` php
$collection = collect([
    new User,
    new User,
    new Post,
    new Comment
]);

$filtered = $collection->whereNotInstanceOf(User::class);

dd($filtered->all());

// [App\Models\Post, App\Models\Comment]

```

### DIFF
--- a/src/Illuminate/Collections/Enumerable.php
+++ b/src/Illuminate/Collections/Enumerable.php
@@ -467,6 +467,16 @@ interface Enumerable extends Arrayable, Countable, IteratorAggregate, Jsonable, 
     public function whereInstanceOf($type);
 
     /**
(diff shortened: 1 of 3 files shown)


xurshudyan:
Can someone tell me why my tests failed. I checked all tests and everything seems to be ok.

taylorotwell:
I feel like we are just adding methods nobody is really asking for at the moment?

PR #46036, opened 2023-02-08
[9.x] Add return null to the getCookie method of test response

salehhashemi1992:
This pull request adds a return statement for the case where no matching cookie is found in the getCookie method. 
The method was previously missing a return statement.


taylorotwell:
Unnecessary.

PR #46019, opened 2023-02-08
[9.x] - Command Input Validation

surgiie:
I attempted to add this feature (https://github.com/laravel/framework/pull/29976) a while ago and I believe Taylor had some interest to explore, but he changed his mind for some reason. I'm happy to close if this is truly not a desired feature for the framework, but I wanted to try once again as I still believe it would be really useful for artisan commands to have the ability to utilize the framework's validation features without manual implementation per command. This allows the developer to not worry about validating the input for their commands manually and allows for the command to be clean of validation logic. 

Using the example of a sending a password reset command:

```php
<?php

namespace App\Console\Commands;

use Illuminate\Console\Command;
use Illuminate\Contracts\Console\Validatable;

class SendPasswordResetCommand extends Command implements Validatable
{

    protected $signature = 'password:send-reset {email}{--expires-date=}';

    protected $description = 'Send a password reset link to the given email.';

    public function rules()
    {
        return [
            'email'=>'email|max:75',
            'expires-date'=>'bail|required|date|after:today'
        ];
    }

    public function handle()
    {
         // send email logic
    }
}

```

The following command call:

`php artisan password:send-reset notanemail  --expires-date="not a date"`

Would result with an exit and with the following errors displayed:

image

PR #46037, opened 2023-02-08
[10.x] Command Input Validation

surgiie:
As requested by @taylorotwell, I'm reopening this pull request (https://github.com/laravel/framework/pull/46019) against the `10.x` branch for discussion/next steps. As documented previously in #46019, this pull request adds the ability to make artisan command arguments and options "validatable" via a new interface `Illuminate\Contracts\Console\Validatable`. 

I understand that there are goals in mind for Laravel 10 in regards to type hints and various other things, but I'm not super familiar with those goals or expectations, so I'm opting in to opening the pull request as it was originally opened in the previous attempt without any type hints and will gladly make any type hint/10.x changes as requested by the maintainers. 

## Concerns/Discussion

I'm not sure the default validation messages for this work well given that the default messages for validation are tailored for "form fields", especially after https://github.com/laravel/framework/pull/45974 was merged and the word "field" was added in. 

This makes the default message poorly worded imo as this input is not "fields" , but either a "argument" or "option". So seeing a messages like `The <argument-or-option-name> field ....` is not great.

This can of course be addressed with a custom `messages` function that can allow the developer to use custom language/messages, but it feels like it shouldn't be needed and I personally would rather see something more default be implemented that could make the messages a bit more distinct on their types, such as:

`The --example option must be lowercase.` or `The example argument must be lowercase.`  

vs 

`The example field must be lowercase.`

I think I can somewhat make this a bit better by prefixing `--` to options before passing it to the validator, but the default messages would still lack clarity as far as input type (field vs command input type) goes, so I'll wait for some initial feedback to see what maintainers think is a clean way to move forward here and I'm requesting guidance, thanks.

### DIFF
--- a/src/Illuminate/Console/Command.php
+++ b/src/Illuminate/Console/Command.php
@@ -4,6 +4,7 @@ namespace Illuminate\Console;
 
 use Illuminate\Console\View\Components\Factory;
 use Illuminate\Contracts\Console\Isolatable;
+use Illuminate\Contracts\Console\Validatable;
 use Illuminate\Support\Traits\Macroable;
 use Symfony\Component\Console\Command\Command as SymfonyCommand;
 use Symfony\Component\Console\Input\InputInterface;
@@ -17,6 +18,7 @@ class Command extends SymfonyCommand
         Concerns\InteractsWithIO,
         Concerns\InteractsWithSignals,
         Concerns\PromptsForMissingInput,
+        Concerns\ValidatesInput,
         Macroable;
 
     /**
@@ -186,6 +188,16 @@ class Command extends SymfonyCommand
             return (int) (is_numeric($this->option('isolated'))
                         ? $this->option('isolated')
                         : self::SUCCESS);
+        }
+
+        if ($this instanceof Validatable) {
+            try {
+                $this->validateInput();
+            } catch (ValidationException $e) {
+                $this->displayFailedValidationErrors($e->getValidator());
+
+                return 1;
+            }
         }
 
         $method = method_exists($this, 'handle') ? 'handle' : '__invoke';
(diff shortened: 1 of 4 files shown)


dinhquochan:
If `illuminate/console` support Validation, it must be required `illuminate/translation` and `illuminate/validation`.

crynobone:
What will happen when the command is called from web request using `Artisan::call()`, does the validation error get ignored or converted to a different exception? It would be misleading if it gets handled as if there it is validation errors for web request (especially if the route has it's own validation). 

Secondly, I would expect `options()` to prefix the name with `--`, this has been used in `Artisan::call()` and `$this->artisan()` on tests.

henzeb:
> Secondly, I would expect `options()` to prefix the name with `--`, this has been used in `Artisan::call()` and `$this->artisan()` on tests.

exactly my thought. That way you can have options and arguments with the same name. henzeb/laravel-console-facade works that way. 

I also would expect the error message to be more aggregated. like my package does.

taylorotwell:
I honestly wouldn't expect any special or magic behavior from the validation if it is called via Artisan::call... it should behave just like a normal console command, writing the errors to output as normal.

taylorotwell:
Drafting this for now because it could be a bit before I get to it.

DarkGhostHunter:
Before putting my thumbs up on this, the Validation exception should be pretty-printable on console first. Is it?

surgiie:
@taylorotwell Not in any hurry to move forward, just pinging to bring to your attention again. Im happy to coordinate any changes as needed.

driesvints:
@surgiie Taylor doesn't reviews draft PRs. Please mark this as ready for review if you need another one.

PR #46044, opened 2023-02-09
[9.x] invalid json method extra parameter

salehhashemi1992:
there is an extra parameter in this method. 
we can delete that parameter or just use the $request instead of the request method.

I replaced the request() with the $request parameter throughout the method to avoid any confusion.


driesvints:
Tests are now failing.

PR #46045, opened 2023-02-09
[9.x] Replace array_search with in_array in Kernel.php

salehhashemi1992:
The original code used `array_search` to check if the middleware already exists in the stack, but this has been replaced with the `in_array` function for simplicity and improved readability.

This change does not affect the functionality of the code but results in improved code quality and readability.

### DIFF
--- a/src/Illuminate/Foundation/Http/Kernel.php
+++ b/src/Illuminate/Foundation/Http/Kernel.php
@@ -328,7 +328,7 @@ class Kernel implements KernelContract
      */
     public function prependMiddleware($middleware)
     {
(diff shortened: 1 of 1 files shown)


driesvints:
The code is already working fine 👍

salehhashemi1992:
Isn't it more readable? 
Half of the methods in this class are using array_search and other half are using in_array @driesvints

driesvints:
There's no need to change code that's working fine.

salehhashemi1992:
> There's no need to change code that's working fine.

So 'refactoring' doesn't have meaning at all

PR #46046, opened 2023-02-09
[9.x] use an instead of a

salehhashemi1992:
Since the abbreviation HTTP begins with a vowel sound, it takes the article "an".


driesvints:
This isn't needed.

driesvints:
Looks like there's several different usages. Let's leave this be for now.

PR #46048, opened 2023-02-09
[9.x] invalid json method extra parameter

salehhashemi1992:
there is an extra parameter in this method.
we can delete that parameter


driesvints:
This is a breaking change to the method signature. If people are overwriting this method, their app will break.

PR #46052, opened 2023-02-09
[9.x] Update `afterPromptingForMissingArguments` method

xurshudyan:
This pull request allows users to check if the argument name is a reserved word in the PHP language before proceeding to the next step.

Why go to the next step if the input is invalid?

## before

Screenshot_20230209_011901

## after

Screenshot_20230209_011957


henzeb:
remember that this method can also be used by others. Who maybe want to use words like 'self' or 'default' for entirely different purposes...

PR #46053, opened 2023-02-09
[10.x] Config path customization

rogervila:
This PR adds the ability to customize the config path on `Illuminate\Foundation\Application` via the `useConfigPath` method that works the same way as other customization path methods like `useDatabasePath` or `useLangPath`, among others.

### DIFF
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -118,18 +118,18 @@ class Application extends Container implements ApplicationContract, CachesConfig
     protected $appPath;
 
     /**
+     * The custom configuration path defined by the developer.
+     *
+     * @var string
+     */
+    protected $configPath;
+
+    /**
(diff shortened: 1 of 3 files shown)


driesvints:
Hey @rogervila. Why do you need this?

rogervila:
@driesvints I am working on a project with a customized laravel structure. 

Right now, the config path remains on the base path, but we would like to move it to another directory. This PR provides the ability to do so.

driesvints:
Okay but can you explain why you need a customized structure?

rogervila:
Yes.

The previously mentioned project is an old PHP project that is being partially migrated to Laravel.

Fitting the project into the laravel structure or refactoring it from scratch is not viable, so what we are doing is fitting laravel into the project. This is why we are customizing the structure to integrate it as smoothly as possible, avoiding collisions between the old and the new code.

PR #46056, opened 2023-02-09
fix: support previous timezones

hendrikheil:
Support timezones like `America/Buenos_Aires` which are often sent by browsers in validation.

`timezone_identifiers_list()` currently does not contain backward compatible timezone links (see https://en.wikipedia.org/wiki/List_of_tz_database_time_zones), while the Intl extension plugin does support those when choosing `TYPE_CANONICAL_LOCATION`.


driesvints:
This seems to break the test suite. Feel free to resend this once tests are passing 👍

PR #46058, opened 2023-02-09
[10.x] Add ```setKey``` and ```getKey``` in container

milwad-dev:
This PR is related to this PR(https://github.com/laravel/docs/pull/8532)

When you want to set key on container, You may use this:
```php
app()['middleware.disable'] = true;
```

This is a bit ugly, With ```setKey``` method you can set key fluently on container.
```php
app()->setKey('middleware.disable', false);
```

Or maybe you want to catch one key.
```php
app()->getKey('middleware.disable');
```

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -707,6 +707,18 @@ class Container implements ArrayAccess, ContainerContract
     }
 
     /**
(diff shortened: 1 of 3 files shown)


taylorotwell:
Not really needed I don't think.

Issue #46029, opened 2023-02-08
Memory Leak when running anonymous class migrations in PHPUnit test


dbohn:
- Laravel Version: 9.51.0
- PHP Version: 8.2.2
- Database Driver & Version: SQLite 3.37.2

### Description:

During the update of our codebase, we migrated our migration files to the new anonymous class system. When running our tests, we noticed, that our tests were consuming much more memory, to the point, that the tests started to fail.
When reverting the change, we noticed, that the memory consumption dropped to an acceptable level again.

### Steps To Reproduce:

You can find a reproduction here: https://github.com/dbohn/laravel-anonymous-class-migration-report

It runs a lot of migrations and a lot of test cases to illustrate the impact of the memory issue, in the first execution round using the new migration style, and afterwards the same test cases with the old migration style.

When executing the `run.sh` script in the repository, we observe the following results:

![image](https://user-images.githubusercontent.com/166164/217543713-4dedad4c-5726-48ee-a301-887dd1b2bfb2.png)

The top shows the anonymous class migrations, the bottom show the classic migrations. The only difference between the migrations is the definition style. Which shows, that the legacy syntax only consumes 15% of the memory.

**EDIT**: I noticed, that I still had xdebug enabled (it was a long debugging session to isolate this), so the numbers (especially the execution time) were a bit higher. Here is another screenshot with PHP 8.2.2 without xdebug:

![image](https://user-images.githubusercontent.com/166164/217546811-08fb3887-4095-412e-8cda-1e16d0002e1f.png)

I also switched to PHP 8.1 to check if it is an error with PHP 8.2, but there is a similar difference in memory consumption:

![image](https://user-images.githubusercontent.com/166164/217546227-15e662c9-4abc-4e42-8150-4cbd8bb896e9.png)

phh:
I changed my migrations back to named classes instead of anonymous and it was much faster and using less memory.

Before: 
`Time: 00:12.996, Memory: 116.50 MB`

After:
`Time: 00:07.089, Memory: 104.50 MB`

rojtjo:
The issue seems to be happening in the [`Migrator` on line 514](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Database/Migrations/Migrator.php#L514).

[...]

PR #46061, opened 2023-02-09
[10.x] Fixes memory leak on anonymous migrations

nunomaduro:
This pull request fixes a memory leak on Laravel test suites that are using anonymous migrations. 

// Work in progress.

```
// Before:
Time: 00:39.915, Memory: 324.00 MB

// After
Time: 00:37.882, Memory: 40.00 MB
```

Fixes https://github.com/laravel/framework/issues/46029.

### DIFF
--- a/src/Illuminate/Database/Migrations/Migrator.php
+++ b/src/Illuminate/Database/Migrations/Migrator.php
@@ -67,6 +67,13 @@ class Migrator
     protected $paths = [];
 
     /**
+     * The migrations that have already been required.
+     *
+     * @var array<string, \Illuminate\Database\Migrations\Migration>
+     */
+    protected static array $requiredMigrations = [];
+
+    /**
      * The output interface implementation.
      *
      * @var \Symfony\Component\Console\Output\OutputInterface
@@ -515,7 +522,11 @@ class Migrator
             return new $class;
         }
 
-        $migration = $this->files->getRequire($path);
+        if (! isset(static::$requiredMigrations[$path])) {
+            static::$requiredMigrations[$path] = $this->files->getRequire($path);
+        }
+
+        $migration = static::$requiredMigrations[$path];
 
         return is_object($migration) ? $migration : new $class;
     }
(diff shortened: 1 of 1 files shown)


src/Illuminate/Database/Migrations/Migrator.php
```diff
@@ -557,7 +568,9 @@ public function getMigrationFiles($paths)
     public function requireFiles(array $files)
     {
         foreach ($files as $file) {
-            $this->files->requireOnce($file);
```
miki131, line 560: `static::$requiredMigrations[$file] `??=` $this->files->requireOnce($file);`

Try this instead of IF

donnysim:
I wonder if php will ever fix the anonymous global function memory leaking 🙃

src/Illuminate/Database/Migrations/Migrator.php
```diff
@@ -557,7 +568,9 @@ public function getMigrationFiles($paths)
...
+            if (! isset(static::$requiredMigrations[$file])) {
+                static::$requiredMigrations[$file] = $this->files->requireOnce($file);
```
dbohn, line 572: If I remember correctly, require_once only returns `true` if the file was successfully required and does not return the return value of the file. That might be unintended at this stage

dbohn:
Thank you for looking into this so quickly!
I noticed, that there is no logic in place to reset the state of the migration class. Is this intentional or still WIP?
This might be a breaking change (as there probably are devs relying on this when running their tests), which might be why it's targeted to v10 :-)

rojtjo:
> I noticed, that there is no logic in place to reset the state of the migration class. Is this intentional or still WIP?

This can be fixed by cloning the cached migration before returning it. I've tested this using your (@dbohn) example repository and the memory seems to remain the same.

This would mean that this pull request could actually target the 9.x branch.

```
// without clone
Time: 01:29.299, Memory: 38.00 MB

// with clone
Time: 01:29.864, Memory: 38.00 MB
```

nunomaduro:
@rojtjo Excellent idea. Going to update the pull request.

nunomaduro:
Replaced by  https://github.com/laravel/framework/pull/46073.

PR #46063, opened 2023-02-09
[9.x] Add index hinting support to query builder

cwhite92:
This PR adds `useIndex()`, `forceIndex()` and `ignoreIndex()` methods to the query builder, to help MySQL's (and others) query planner use the indexes I want for a query. I've had to do this in the past to optimize slow queries where the planner didn't choose to use an index even though it was dramatically more performant, and today I came across @assertchris [doing the same thing](https://twitter.com/assertchris/status/1623730973534744578), so I thought it might be worth adding to the ORM for other people to use.

Support across engines is spotty, with only MySQL supporting all three hinting modes. Based on my research this seems to be the support:

| Database  | Index hinting | Index forcing | Index ignoring
| ---------- | ---------- | ---------- | ---------- |
| MySQL | ✅ Supported | ✅ Supported | ✅ Supported |
| SQLite | ❌ Not supported | ✅ Supported | ❌ Not supported (can only ignore all indexes) |
| PostgreSQL | ❌ Not supported | ❌ Not supported | ❌ Not supported |
| SQL Server | ❌ Not supported | ✅ Supported | ❌ Not supported (can only ignore all indexes) |

I have only implemented these where supported, for engines without support the query will be unchanged by the `useIndex()`, `forceIndex()` and `ignoreIndex()` methods.

I have also only added support for hinting one index. MySQL supports the use of multiple `USE INDEX` hints, and a combination of `USE INDEX` hints with `IGNORE INDEX` hints, but I have left that unimplemented for simplicity and to test the water on if this change is wanted. If we want to add support for hinting multiple indexes, I can add that functionality.

A 4th method called `ignoreIndexes()` could be added to support SQLite and SQL Server's ability to ignore all indexes and do full table scans, but I struggle to think of why you'd actually want to do that in a real environment.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -103,7 +103,7 @@ class Builder implements BuilderContract
     /**
      * The index hint for the query.
      *
-     * @var IndexHint
+     * @var \Illuminate\Database\Query\IndexHint
      */
     public $indexHint;
 
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -101,6 +101,13 @@ class Builder implements BuilderContract
     public $from;
 
     /**
+     * The index hint for the query.
+     *
+     * @var IndexHint
+     */
+    public $indexHint;
+
+    /**
      * The table joins for the query.
      *
      * @var array
@@ -454,6 +461,45 @@ class Builder implements BuilderContract
         }
 
         $this->from = $as ? "{$table} as {$as}" : $table;
+
+        return $this;
+    }
+
+    /**
+     * Add an index hint to suggest a query index.
+     *
+     * @param  string  $index
+     * @return $this
+     */
+    public function useIndex($index)
+    {
+        $this->indexHint = new IndexHint('hint', $index);
+
+        return $this;
+    }
+
+    /**
+     * Add an index hint to force a query index.
+     *
+     * @param  string  $index
+     * @return $this
+     */
+    public function forceIndex($index)
+    {
+        $this->indexHint = new IndexHint('force', $index);
+
+        return $this;
+    }
+
+    /**
+     * Add an index hint to ignore a query index.
+     *
+     * @param  string  $index
+     * @return $this
+     */
+    public function ignoreIndex($index)
+    {
+        $this->indexHint = new IndexHint('ignore', $index);
 
         return $this;
     }
--- a/src/Illuminate/Database/Query/Grammars/Grammar.php
+++ b/src/Illuminate/Database/Query/Grammars/Grammar.php
@@ -36,6 +36,7 @@ class Grammar extends BaseGrammar
         'aggregate',
         'columns',
         'from',
+        'indexHint',
         'joins',
         'wheres',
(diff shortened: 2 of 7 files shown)


decadence:
> As of MySQL 8.0.20, the server supports the index-level optimizer hints [JOIN_INDEX](https://dev.mysql.com/doc/refman/8.0/en/optimizer-hints.html#optimizer-hints-index-level), [GROUP_INDEX](https://dev.mysql.com/doc/refman/8.0/en/optimizer-hints.html#optimizer-hints-index-level), [ORDER_INDEX](https://dev.mysql.com/doc/refman/8.0/en/optimizer-hints.html#optimizer-hints-index-level), and [INDEX](https://dev.mysql.com/doc/refman/8.0/en/optimizer-hints.html#optimizer-hints-index-level), which are equivalent to and intended to supersede FORCE INDEX index hints, as well as the [NO_JOIN_INDEX](https://dev.mysql.com/doc/refman/8.0/en/optimizer-hints.html#optimizer-hints-index-level), [NO_GROUP_INDEX](https://dev.mysql.com/doc/refman/8.0/en/optimizer-hints.html#optimizer-hints-index-level), [NO_ORDER_INDEX](https://dev.mysql.com/doc/refman/8.0/en/optimizer-hints.html#optimizer-hints-index-level), and [NO_INDEX](https://dev.mysql.com/doc/refman/8.0/en/optimizer-hints.html#optimizer-hints-index-level) optimizer hints, which are equivalent to and intended to supersede IGNORE INDEX index hints. Thus, you should expect USE INDEX, FORCE INDEX, and IGNORE INDEX to be deprecated in a future release of MySQL, and at some time thereafter to be removed altogether.

taylorotwell:
Hmm, well, we probably should not have added this then.

cwhite92:
Damn, I wasn't aware of that depreciation. This complicates things for MySQL > 8.0.

We can still support forcing and ignoring indexes, but the syntax has now changed to:

```
select /*+ index(users users_email_unique) */ email
from users
where email = 'user@domain.com';

select /*+ no_index(users users_email_unique) */ email
from users
where email = 'user@domain.com';
```

(I attempted to move the index hinting to different parts of the query, but it only works if it's between the `select` and list of columns)

So to support all versions of MySQL without introducing a breaking change, I'd have to:

1. Add `compileAggregate()` and `compileColumns()` to the `MySqlGrammar` class to insert the index hint there for MySQL > 8.0; and
2. Keep the current `compileIndexHint()` method in `MySqlGrammar` for MySQL < 8.0; and
3. Check for the MySQL version in all of those three methods, which I don't seem to currently have access to inside the grammar classes.

I could simplify this and only support it on MySQL 8.0 and above, but that would be a breaking change since this PR is already released with Laravel 10, so I assume it'd go to `master` for Laravel 11?

Let me know how you'd like to handle this.

decadence:
@cwhite92 What decision have you made about it? Because current MySQL implementation will stop working at one point I think.

> I could simplify this and only support index hinting on MySQL 8.0 and above, but that would be a breaking change since this PR is already released with Laravel 10, so I assume it'd go to master for Laravel 11?

I think this is the most suitable for this situation.

PR #46067, opened 2023-02-10
[9.x] Add default eager loading of related column aggregates

cwilby:
Building on `$with` and `$withCount`, `$withAggregate` adds the ability to eager load related column aggregate values by default.

```php
class Product extends Model
{
    protected $with = ['orders'];
    protected $withCount = ['orders'];
    protected $withAggregate = [
        'sum' => [
            'orders:amount'
        ],

        'avg' => [
            'orders:amount'
        ],

        'min' => [
            'orders:amount'
        ],

        'max' => [
            'orders:amount'
        ]
    ];

    public function orders()
    {
        return $this->belongsToMany(Order::class);
    }
}
```

```php
foreach (Product::popular()->take(10)->get() as $product) {
    echo $product->orders_sum_amount;
    echo $product->orders_avg_amount;
    echo $product->orders_min_amount;
    echo $product->orders_max_amount;
}
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -84,6 +84,13 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
     protected $withCount = [];
 
     /**
+     * The relationship aggregates that should be eager loaded on every query.
+     *
+     * @var array
+     */
+    protected $withAggregate = [];
+
+    /**
      * Indicates whether lazy loading will be prevented on this model.
      *
      * @var bool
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -1525,9 +1532,32 @@ public function registerGlobalScopes($builder)
...
+        );
+    }
+
+    /**
+     * Add default model aggregates to a query builder.
+     *
+     * @param  \Illuminate\Database\Eloquent\Builder|static  $query
+     * @return \Illuminate\Database\Eloquent\Builder|static
+     */
+    protected function withAggregates(Builder $query)
```
cwilby, line 1548: Note:  I don't love this method, or how it's called in newQueryWithoutScopes.  Open for improvement.

taylorotwell:
I really just don't think this is a great practice 😬 I'm pretty strongly of the opinion you shouldn't even use `with` or `withCount`, so I don't really want to add `withAggregate` to further things.

cwilby:
Makes sense, I had a feeling that's why it wasn't there.

Thanks!

PR #46068, opened 2023-02-10
[9.x] Update `afterPromptingForMissingArguments` method

xurshudyan:
This PR continuation of this https://github.com/laravel/framework/pull/46052

## Before

Screenshot_20230210_010627

## After

Screenshot_20230210_010706

### DIFF
--- a/src/Illuminate/Console/GeneratorCommand.php
+++ b/src/Illuminate/Console/GeneratorCommand.php
@@ -457,7 +457,7 @@ abstract class GeneratorCommand extends Command implements PromptsForMissingInpu
      *
      * @return bool
      */
-    protected function isReservedAOrExistenceName()
+    protected function isReservedOrExistingName()
     {
         if ($this->isReservedName($this->getNameInput())) {
             return true;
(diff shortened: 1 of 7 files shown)


src/Illuminate/Console/GeneratorCommand.php
```diff
@@ -452,6 +452,24 @@ protected function isReservedName($name)
         return in_array($name, $this->reservedNames);
     }
 
+    /**
+     * Checks whether the given class name is reserved or already exists.
+     *
+     * @return bool
+     */
+    protected function isReservedAOrExistenceName()
```
driesvints, line 460: This method name doesn't feels correct.

driesvints (changes requested):

src/Illuminate/Console/GeneratorCommand.php
driesvints: `isReservedOrExistingName`

xurshudyan: good

henzeb:
Would it be more practical if we get to enter a new name instead?

taylorotwell:
I don't really want to keep nit-picking this.

PR #46072, opened 2023-02-10
[9.x] Don't expose model full path in model not found exception

andiLeong:
When The ModelNotFoundException class prepares for the exception message does not check for app.debug level so if we are in APP_DEBUG set to false (meaning in production) we still expose the model full path as the json response. 

example:
```
{
    "message": "No query results for model [App\\Models\\Video] 99"
}
```

which is not good. We are exposing our model path.
I propose just give the model short class name as the message, as below no matter what env

```
{
    "message": "No query results for model [Video] 99"
}
```

Thank you

:zzz: :zzz: :zzz: :zzz:

### DIFF
--- a/src/Illuminate/Database/Eloquent/ModelNotFoundException.php
+++ b/src/Illuminate/Database/Eloquent/ModelNotFoundException.php
@@ -36,6 +36,7 @@ class ModelNotFoundException extends RecordsNotFoundException
         $this->model = $model;
         $this->ids = Arr::wrap($ids);
 
+        $model = class_basename($model);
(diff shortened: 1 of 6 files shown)


ankurk91:
The Model class may or may not resides in the `app/Models` directory.

It may resides in the vendor folder as well.

I would suggest not to show the model class name in production at all.

andiLeong:
@ankurk91 
I think its not confused because the message already hint that that's a model.
no query results for model [Post]
unless the model class belongs to vendor, developer have hard time to track donw.

taylorotwell:

PR #46077, opened 2023-02-10
[10.x] Add missing void type

GrahamCampbell:
The other void functions in the stubs have these, but these were forgotten about.


nunomaduro:
As explaining in https://github.com/laravel/laravel/pull/6010, we are intentionally not adding return types to closures.

PR #46084, opened 2023-02-11
[9.x] Use ```shouldAllowMockingProtectedMethods``` for mocking in FilesystemAdapterTest

milwad-dev:
Need to use this method because protected methods cannot be mocked without calling ```Mock.shouldAllowMockingProtectedMethods```

### DIFF
--- a/tests/Filesystem/FilesystemAdapterTest.php
+++ b/tests/Filesystem/FilesystemAdapterTest.php
@@ -91,6 +91,7 @@ class FilesystemAdapterTest extends TestCase
         $files = m::mock(FilesystemAdapter::class, [$this->filesystem, $this->adapter])
             ->shouldAllowMockingProtectedMethods()
             ->makePartial();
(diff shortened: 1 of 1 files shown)


taylorotwell:
But our tests aren't currently failing?

milwad-dev:
@taylorotwell 
It covers more tests and runs properly.

milwad-dev:
@taylorotwell 
It covers more tests and runs properly.

tests/Filesystem/FilesystemAdapterTest.php
```diff
@@ -91,6 +91,7 @@ public function testFallbackNameCalledAlreadyProvidedToResponse()
         $files = m::mock(FilesystemAdapter::class, [$this->filesystem, $this->adapter])
             ->shouldAllowMockingProtectedMethods()
```
PerryvanderMeer, line 92: 👀

PR #46087, opened 2023-02-11
Pass resetUrl variable to user's toMail callback function instead of token

borisolhor:
I noticed, that in `ResetPassword.php` and `VerifyEmail.php` files in the `toMail` function used different parameters for the user's callback function: in the `VerifyEmail.php` file there are used the same variable `$verificationUrl` for both user callback and `buildMailMessage` function but in `ResetPassword.php` file in `toMail` function for user's callback function used token variable instead of URL, but to the `buildMailMessage` function there is passed URL.
I think it's will be better to pass the URL to the user callback function in both files. In this case, we don't need manually generate a URL, based on the passed token, when we use the `toMailUsing` callback.

### DIFF
--- a/src/Illuminate/Auth/Notifications/ResetPassword.php
+++ b/src/Illuminate/Auth/Notifications/ResetPassword.php
@@ -59,11 +59,13 @@ class ResetPassword extends Notification
      */
     public function toMail($notifiable)
     {
+        $resetUrl = $this->resetUrl($notifiable);
+
         if (static::$toMailCallback) {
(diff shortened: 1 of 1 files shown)


rojtjo:
This would break anyone's application that relies on the token being passed instead of the URL. It's also more flexible to receive the token by itself since you can easily generate the URL, but parsing the URL to retrieve the token is a bit more tedious.

It might be better to add it as an additional parameter. That way existing applications won't break and anyone who doesn't use it can just ignore the parameter all together.

PR #46091, opened 2023-02-13
[10.x] Adds Servers to HTTP Client Factory

DarkGhostHunter:
# HTTP Servers

This PR is meant to help on two scenarios

1) Having more than one HTTP API servers. For example, posting something to multiple social networks at the same time.
2) Having multiple verbose URIs for a given server, scattered through the app.

Let's start with the first problem. 

> Note, this PR doesn't uses  the Container because it's not a dependency for `Illuminate/Http`. Bummer :(

* **Problem 1: Multiple Servers**

Instead of creating raw HTTP Clients around controllers, we can move that logic into a single class that extends the `Server` class.

```php
namespace App\Http\Servers;

use Illuminate\Http\Client\Server;

class Chirper extends Server
{
    protected $baseUrl = 'https://chirper.app/api';
    protected $authToken = 'my-chirper-secret';
}
```

We can use that server with the HTTP Factory with  `server()` and the class name. A new instance will be created by gathering the server attributes.

```php
// With the server syntax.
$latest = Chirper::request()->get('/timeline')

// Same as:
$latest = Http::withToken('my-chirper-secret')->get('https://chirper.app/api/timeline')
```

* **Problem 2: Too many URIs**

The second problem is having multiple URIs for different things. For example, if we have multiple actions over one resource, the URIs to define can become hectic and scattered across the application. 

We can simplify this by defining "actions" for a server on the class with an array of action names that contain the HTTP Verb and the route. If no verb is prepended, `get` is inferred.

```php
class Chirper extends Server
{
    protected $actions = [
        'latest' => 'timeline', // sane as "get:timeline"
        'chirp'  => 'post:new',
        'delete' => 'delete:chirp/{id}'
    ];
};

// Arguments are passed down magically to the request.
Chirper::request()->chirp(['message' => 'This is awesome!']);

// You can set the URL parameters when invoking the server too.
Chirper::request(['id' => 46578])->delete();
```

Of course, you can also create a method that makes the request to the server directly with optional arguments. With some `_call()` magic, we can build upon the same request and chain multiple actions.

```php
class Chirper extends Server
{
    public function noReply()
    {
        return $this->header(['X-Reply' => false]);
    }

    public function chirp($message)
    {
        return $this->post('/new', [
            'message' => $message,
        ]);
    }
};

Chirper::request()->chirp('This is awesome!');

Chirper::request()->noReply()->chirp('You can not reply to this');
```

## Overriding

All methods not present in the class are forwarded to the underlying `PendingRequest`. You can override anything as usual since the build step happens on server invoke, and the last step is the action.

```php
$newChirp = Chirper::request()->withToken('user-secret')->chirp("You're awesome");
```

## Async (Concurrent)

When doing concurrent requests to a single or different servers, the `server()`  method must be used on the pool request rather than outside.

```php
Http::pool(fn ($pool) => [
    $pool->server(Chirper::class)->latest(),
    $pool->server(Facebook::class)->get('latest'),
]);
```

## Macros

If the Factory contains Macros, these will take precedence over the dynamic server name. The purpose of using `server()` is to avoid these collisions.

## Testing

Since the Servers defined use the HTTP Factory behind the scenes, there is no need to refactor tests that already exists.

### DIFF
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -403,7 +403,6 @@ class Factory
         return $this->dispatcher;
     }
 
-
     /**
      * Execute a method against a new pending request instance.
      *
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -10,7 +10,6 @@ use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
 use PHPUnit\Framework\Assert as PHPUnit;
-use RuntimeException;
 
 /**
  * @mixin \Illuminate\Http\Client\PendingRequest
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -107,7 +107,7 @@ class Factory
         $server = $this->servers[$name] ?? $this->servers[Str::camel($name)];
 
         return $server
-            ? tap(new $server($this, $parameters))
+            ? new $server($this, $parameters)
             : throw new RuntimeException("The server \"$name\" is not defined.");
     }
 
@@ -418,7 +418,7 @@ class Factory
         }
 
         if (isset($this->servers[$method])) {
-            return tap(new $this->servers[$method]($this, $parameters));
+            return new $this->servers[$method]($this, $parameters);
         }
 
         return tap($this->newPendingRequest(), function ($request) {
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -10,6 +10,7 @@ use Illuminate\Contracts\Events\Dispatcher;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
 use PHPUnit\Framework\Assert as PHPUnit;
+use RuntimeException;
 
 /**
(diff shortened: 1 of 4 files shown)


DarkGhostHunter:
Crap, wrong button.

oprypkhantc:
Aren't SDKs supposed to solve these? If you still have to define a class with all of the endpoints, you're much better off not extending anything and tying yourself to the HTTP client implementation and instead using it as a dependency:

```php
class Chirper {
    private readonly HttpClient $client;

        $this->client = Http::withBaseUri('https://chirper.app');
    }

    public function chirp(string $message): void
    {
        $this->client->get('new', ['message' => $message]);
    }
}
```

You get the same done with less hardcode dependency, no public facing HTTP clutter and the same amount of code.

DarkGhostHunter:
@oprypkhantc Not everything has an SDK, let alone a _good_ SDK.

> You get the same done with less hardcode dependency

The whole HTTP Client is hardcoded (remember, no Container). This is the same case as controllers classes: you can make your own, but it's more convenient to extend the `App\Http\Controller`.

taylorotwell:
This feels like forcing a new concept into the framework that doesn't really need to be there. I would just make an `App\Services\Chirper` class use it as a gateway for interacting with Chirper. Within that class, I can do whatever I want to give myself convenient affordances for interacting with that service, as well as expose convenient methods to the public.

oprypkhantc:
> Not everything has an SDK, let alone a _good_ SDK.

Well that's the point. It looks like an SDK, behaves similar and takes roughly the same effort to write one, but you're getting worse results.

PR #46098, opened 2023-02-13
[10.x] Default spaces=false when generating passwords

valorin:
I'd recommend defaulting to `$spaces = false` when generating passwords to avoid the situation where the password begins/ends with a space. 
Having a space at the start or end could be easily missed when a user copies the generated password from the page.

Having a `.` at the end can be confusing in some situations too, but that's less of an issue than space character and can be solved through formatting the password display.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -731,7 +731,7 @@ class Str
      * @param  bool  $spaces
      * @return string
      */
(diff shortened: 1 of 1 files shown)


vlakoff:
As a special case, it would be even better to ensure, when spaces are enabled, that there can't be spaces at start/end of the password. (But maybe it would complexify the code too much)

PR #46099, opened 2023-02-13
[10.x] Updated Eloquent\Collection callable docblocks

Ilyes512:
So while using `->load()` I got a PHPStan (Larastan) error. For example with something like this:

```php
<?php

use Illuminate\Database\Eloquent\Relations\HasMany;

$eloquentCollection->load([
    'relationOne' => static function (HasMany $q): void {
        $q->where('foobar', 1);
        $q->with(['relationTwo' => static function (HasMany $q2): void {
            $q2->limit(100);
        }]);
    },
    'anotherRelation',
]);
```

I get:

> Parameter #1 $relations of method Illuminate\Database\Eloquent\Collection<(int|string),App\Models\Eloquent\SomeEloquentModel>::load() expects array<(callable(Illuminate\Database\Eloquent\Builder): mixed)|string>|string, array{relationOne: Closure(Illuminate\Database\Query\Builder): void, 0: 'anotherRelation'} given.

This is because `HasMany` does not extend `Illuminate\Database\Eloquent\Builder`, but it does implement the contract `Illuminate\Contracts\Database\Eloquent\Builder`.

I replaced all Builder implementation with the contract within `Eloquent\Collection`, but I am not sure if this is correct.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Collection.php
+++ b/src/Illuminate/Database/Eloquent/Collection.php
@@ -52,7 +52,7 @@ class Collection extends BaseCollection implements QueueableCollection
     /**
      * Load a set of relationships onto the collection.
      *
-     * @param  array<array-key, (callable(\Illuminate\Database\Eloquent\Builder): mixed)|string>|string  $relations
+     * @param  array<array-key, (callable(\Illuminate\Contracts\Database\Eloquent\Builder): mixed)|string>|string  $relations
      * @return $this
      */
     public function load($relations)
@@ -73,7 +73,7 @@ class Collection extends BaseCollection implements QueueableCollection
     /**
      * Load a set of aggregations over relationship's column onto the collection.
      *
-     * @param  array<array-key, (callable(\Illuminate\Database\Eloquent\Builder): mixed)|string>|string  $relations
+     * @param  array<array-key, (callable(\Illuminate\Contracts\Database\Eloquent\Builder): mixed)|string>|string  $relations
      * @param  string  $column
      * @param  string|null  $function
      * @return $this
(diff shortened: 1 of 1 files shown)


taylorotwell:
I will need @nunomaduro or someone more familiar with Larastan to review this.

nunomaduro:
I've investigated this issue, and I am currently not sure the best way to address this. You can try to get help from the community in the Larastan repository: https://github.com/nunomaduro/larastan.

Ilyes512:
@nunomaduro 

This is not an issue with PHPStan/Larastan, but an issue with the defined types. Yes, one might be able to create a custom rule/types using stubs maybe, but solving this at the source would be WAY nicer.

So I am not sure if the above was clear enough so I try to summarise the issue here:

The closure receives an instance of `Illuminate\Database\Eloquent\Relations\HasMany`. The class `HasMany` does NOT extend `\Illuminate\Database\Eloquent\Builder`, but the type definitions states that it receives an instance of the `Builder` class.

```
...
* @param  array<array-key, (callable(\Illuminate\Database\Eloquent\Builder): mixed)|string>|string  $relations
...
```

What `HasMany` does do is implement the contract `Illuminate\Contracts\Database\Query\Builder`. So thats why I was suggesting the current change in this PR, by replace the hard instance type with the "softer"/"wider" `Illuminate\Contracts\Database\Query\Builder` contract.

In any case I am flexibly and willing to invest more time to resolve this. Any suggestions on how to proceed with this?

nunomaduro:
> Any suggestions on how to proceed with this?

Yes! You have to generate tests in this file: https://github.com/laravel/framework/blob/10.x/types/Database/Eloquent/Collection.php and update the PHPDocs accordingly, so they pass your new test without breaking the existing ones.

To run the "types" test suite, please run: `./vendor/bin/phpstan --configuration="phpstan.types.neon.dist" --debug`

PR #46100, opened 2023-02-13
[10.x] Update password generator to use `range()` and avoid `array_merge()`

browner12:
- rather than listing out all the alphabet characters, we can use `range()` to make this more succinct
- by using array unpacking, we can avoid the overhead of the `array_merge()` function call

https://wiki.php.net/rfc/spread_operator_for_array#advantages_over_array_merge

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -736,31 +736,34 @@ class Str
         $characters = [];
 
         if ($letters) {
-            $characters = array_merge($characters, [
-                'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k',
-                'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v',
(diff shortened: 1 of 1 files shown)


taylorotwell:
I just updated this function to use collections. 😅

src/Illuminate/Support/Str.php
```diff
@@ -736,31 +736,34 @@ public static function password($length = 32, $letters = true, $numbers = true,
...
+                ...range('a', 'z'),
+                ...range('A', 'Z'),
+            ];
         }
 
         if ($numbers) {
-            $characters = array_merge($characters, [
+            $characters = [
+                ...$characters,
                 '0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
```
valorin, line 749: I know this is closed, but I'm really curious why you didn't use `range(0, 9)` here?

browner12: that generates `int`s, not `string`s.

probably would've worked okay with PHP's type inference on concatenation, but Taylor specifically made these string integers, and I couldn't replicate that with `range()`.

PR #46101, opened 2023-02-13
[10.x] use array unpacking for Collection merging

browner12:
array unpacking should have better performance from avoiding the overhead of a method call, and apparently there are some runtime optimizations as well.

https://wiki.php.net/rfc/spread_operator_for_array https://www.php.net/releases/8.1/en.php#array_unpacking_support_for_string_keyed_arrays https://www.php.net/manual/en/migration56.new-features.php https://php.watch/versions/8.1/spread-operator-string-array-keys


taylorotwell:
I think current code is generally ok.

Issue #46085, opened 2023-02-11
HTTP client builds the wrong request


zareismail:
- Laravel Version: 9.51.0
- PHP Version: 8.2.0
- Database Driver & Version: Mysql

### Steps To Reproduce:

I think there is a bug in the HTTP client because that fills the request body by default at line `Illuminate/Http/Client/PendingRequest.php:727` as following

```
public function post(string $url, $data = [])
    {
        return $this->send('POST', $url, [
            $this->bodyFormat => $data,
        ]);
    }
```

this causes some errors in some matters. for example, at the last work, I try Paypal REST API to [generate client tokens](https://developer.paypal.com/docs/checkout/advanced/integrate/#link-generateclienttoken)  by the following request and received the 400: 

``` 
        Http::withHeaders(['Accept-Language' => 'en_US'])
            ->withToken($this->generateAccessToken())
            ->asJson()
            ->post($this->requestEndpoint().'/v1/identity/generate-token')
```
But when I sent the request by CURL I catch that is working:
```
curl -X POST https://api-m.sandbox.paypal.com/v1/identity/generate-token -H 'Content-Type: application/json' -H 'Authorization: Bearer A21AAJVJOCjtuu-dZZQEDwDo-b4BA3CeNiyV47jMDvz8tNjP37eFOyqqV7b5st0W6A1BQfKBmgbf4MKzpALWPh7w6AqgWqAMg' -H 'Accept-Language: en_US'
```
after debugging, I found that the HTTP client fills the request body with an empty array by default which causes the wrong response.
I solved this by the following solution:
```
Http::withHeaders(['Accept-Language' => 'en_US'])
            ->withToken($this->generateAccessToken())
            ->asJson()
            ->send('POST', $this->requestEndpoint().'/v1/identity/generate-token');
```

iazaran:
I think the solution like
```
empty($data) ? [] : [
    $this->bodyFormat => $data,
]
```
that is used in `delete` method in here: https://github.com/laravel/framework/blob/9.x/src/Illuminate/Http/Client/PendingRequest.php#L767-L772
is useful for other methods like `post` or `put`.

driesvints:
Given the failing tests from the above PR, I think this is working as intended. If you believe this needs to be changed, feel free to attempt a PR with passing tests 👍

PR #46108, opened 2023-02-14
[9.x] Check for empty array in request body

iazaran:
In _Http/Client/PendingRequest_ the helpers of `post`, `patch` and `put` don't check for empty array as request body and this PR will implement same solution that is used on `delete` method.

Related issue: https://github.com/laravel/framework/issues/46085


driesvints:
This seems to break the test suite. Please resend this with passing tests.

PR #46116, opened 2023-02-14
[10.x] Use `range` for alphabets

foremtehan:
@phptinker 
```
array_merge(range('a', 'z'), range('A', 'Z'));
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -735,7 +735,7 @@ class Str
     {
         return (new Collection)
                 ->when($letters, fn ($c) => $c->merge([
-                    array_merge(range('a', 'z'), range('A', 'Z'))
+                    array_merge(range('a', 'z'), range('A', 'Z')),
                 ]))
                 ->when($numbers, fn ($c) => $c->merge([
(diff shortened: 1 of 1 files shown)


src/Illuminate/Support/Str.php
```diff
@@ -735,11 +735,7 @@ public static function password($length = 32, $letters = true, $numbers = true,
     {
         return (new Collection)
                 ->when($letters, fn ($c) => $c->merge([
-                    'a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'i', 'j', 'k',
-                    'l', 'm', 'n', 'o', 'p', 'q', 'r', 's', 't', 'u', 'v',
-                    'w', 'x', 'y', 'z', 'A', 'B', 'C', 'D', 'E', 'F', 'G',
-                    'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R',
-                    'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z',
+                    array_merge(range('a', 'z'), range('A', 'Z')
```
datlechin: Just `range('a', 'Z')` :)

foremtehan: No. It's not working

PR #46117, opened 2023-02-14
Remove redundant PHPDoc

itxshakil:


driesvints:
We're not changing this on internal methods. Thanks

PR #46122, opened 2023-02-14
[10.x] Show queue connection in MonitorCommand

gustavosobrinho01:
Currently, the MonitorCommand command does not display the queue connection in question, only its name and size.

This change will allow the user who uses more than one connection to display multiple connections in the same command.

Ex: php artisan queue:monitor redis:events,database:events

Outuput: the command will display both queues and their connections in the listing.


renoirtech:
This is awesome, kudos @gustavosobrinho01 !

juhniorsantos:
For those who use many queues, it helps a lot in debugging and logging

PR #46129, opened 2023-02-15
[9.x] Support int backed enums in QueryBuilder::whereIntegerInRaw

leandro-hermes:
When using backed enum cases in `\Illuminate\Database\Query\Builder::whereIntegerInRaw`, we get the error:
```
ErrorException : Object of class App\Enums\EnumClass could not be converted to int
```

This PR fixes this by identifying when is an int backed enum and uses its raw value. When is a string backed enum, or even is not a backed enum, the behavior remains the same.

If this PR is accepted, I'll open one for the 10.x branch.


taylorotwell:
New behavior is no longer being accepted for the 9.x branch.

PR #46131, opened 2023-02-15
[10.x] Support int backed enums in QueryBuilder::whereIntegerInRaw

leandro-hermes:
When using backed enum cases in `\Illuminate\Database\Query\Builder::whereIntegerInRaw`, we get the error:
```
ErrorException : Object of class App\Enums\EnumClass could not be converted to int
```

This PR fixes this by identifying when is an int backed enum and uses its raw value. When is a string backed enum, or even is not a backed enum, the behavior remains the same.


taylorotwell:
Just pass integers to this method.

PR #46132, opened 2023-02-15
[10.x] Register policies automatically to the gate

ahmett:
This PR aims to enhance eventual consistency between parts of the framework.

Unlike the other features of the framework (observers, listeners, etc.), policies don't get automatically registered while AuthServiceProvider is booting. It is required to call the `registerPolicies` method. 

And I think this is redundant.

Logic preserves its natural behavior. It doesn't affect auto-discovery, defined with the `Gate::guessPolicyNamesUsing` method.

I also create a PR in the skeleton repo to remove redundant calling of the `registerPolicies` method.

### DIFF
--- a/src/Illuminate/Foundation/Support/Providers/AuthServiceProvider.php
+++ b/src/Illuminate/Foundation/Support/Providers/AuthServiceProvider.php
@@ -22,9 +22,29 @@ class AuthServiceProvider extends ServiceProvider
     public function register()
     {
         $this->booting(function () {
-            foreach ($this->policies as $model => $policy) {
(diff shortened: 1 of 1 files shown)


ahmett:
Docs updated according to this PR, see: laravel/docs#8584
Skeleton repo also updated, see: laravel/laravel#6108

taylorotwell:
Won't they be registered twice if we make this change and people don't update their skeleton?

ahmett:
@taylorotwell It won't because we are replacing that method. Existing codebases with `10.x` only need to delete that method call from the service provider. I already documented this in upgrade docs.

PR #46136, opened 2023-02-16
[10.x] Remove unused catch exception variable

andiLeong:
in php8.0 introduced Non-capturing catches

With non-capturing catches, we can omit the variable, I have clean up some catch statement remove unuse exception variable

```
try {
    // Something goes wrong
} catch (lException) {
    return false;
}
```

before php8.0 , we must keep the exception variable
```
try {
    // Something goes wrong
} catch (lException $e) {
    return false;
}
```

https://wiki.php.net/rfc/non-capturing_catches


ankurk91:
Tip: Laravel team can use rector to upgrade the whole codebase in one go.

driesvints (approved):

PR #46137, opened 2023-02-16
[10.x] fix `Expression` string casting

browner12:
The `Expression` class no longer provides a `__toString()` method, so we cannot cast it to a string here.

We'll now use the `->getValue()` method and pass the current database connection's query grammar.

#44784

strangely enough, the `HasInDatabase::toString()` method is only called to get the failure description, but I was getting the unavailable string casting error even without a test failure. this leads me to believe PHPUnit is (sometimes) calling the `failureDescription()` method even without a failure? Not sure on this.

### DIFF
--- a/src/Illuminate/Testing/Constraints/HasInDatabase.php
+++ b/src/Illuminate/Testing/Constraints/HasInDatabase.php
@@ -113,7 +113,7 @@ class HasInDatabase extends Constraint
     public function toString($options = 0): string
     {
         foreach ($this->data as $key => $data) {
-            $output[$key] = $data instanceof Expression ? (string) $data : $data;
+            $output[$key] = $data instanceof Expression ? $data->getValue($this->database->getQueryGrammar()) : $data;
         }
 
         return json_encode($output ?? [], $options);


src/Illuminate/Testing/Constraints/HasInDatabase.php
```diff
@@ -113,7 +113,7 @@ protected function getAdditionalInfo($table)
     public function toString($options = 0): string
     {
         foreach ($this->data as $key => $data) {
-            $output[$key] = $data instanceof Expression ? (string) $data : $data;
+            $output[$key] = $data instanceof Expression ? $data->getValue($this->database->getQueryGrammar()) : $data;
```
crynobone, line 116: Why not just 

```suggestion
            $output[$key] = $this->database->getQueryGrammar()->wrap($data);
```

https://github.com/laravel/framework/blob/3a46fa307c73c8371ea6c72129f7c23a2f52c7f3/src/Illuminate/Database/Grammar.php#L46-L74

browner12: that looks like you're jumping through a lot of hoops just to get back to what I've written here.

Issue #32198, opened 2020-04-02
When response() is returned in controller the framework tries and fails to convert Response Factory to string


petar-rankovic-logeecom:
- Laravel Version: v6.18.3
- PHP Version: 7.4.3
- Database Driver & Version:5.7.29-0ubuntu0.18.04.1

### Description:
When empty response with status 200 is returned the following exception is thrown:
[2020-04-02 09:51:56] local.ERROR: Object of class Illuminate\Routing\ResponseFactory could not be converted to string {"exception":"[object] (Symfony\\Component\\Debug\\Exception\\FatalThrowableError(code: 0): Object of class Illuminate\\Routing\\ResponseFactory could not be converted to string at /home/petar/Code/shopify/vendor/symfony/http-foundation/Response.php:409) 

### Steps To Reproduce:

1. Write a simple method in controller as follows:
public function test() {
return response();
}

2. Call the method from a web request.

3. The error is thrown as described.

driesvints:
Hi there,

Thanks for reporting but it looks like this is a question which can be asked on a support channel. Please only use this issue tracker for reporting bugs with the library itself. If you have a question on how to use functionality provided by this repo you can try one of the following channels:

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://webchat.freenode.net/?nick=laravelnewbie&channels=%23laravel&prompt=1)

However, this issue will not be locked and everyone is still free to discuss solutions to your problem!

Thanks.

devcircus:
That's not how you return an empty response. Check out the response helper definition. When used with no arguments, it returns a response factory.

Issue #18498, opened 2017-03-26
ResponseFactory could not be converted to string


vutaviva:
- Laravel Version: 5.4
- PHP Version: 7.1.1
- Database Driver & Version: MariaDB 10

### Description:
My log file announce error on ResponseFactory. However, I don't know where it comes from and how to fix it.

Object of class Illuminate\Routing\ResponseFactory could not be converted to string...

```
[2017-03-26 04:22:01] production.ERROR: ErrorException: Object of class Illuminate\Routing\ResponseFactory could not be converted to string in /var/www/example.com/vendor/symfony/http-foundation/Response.php:447
Stack trace:
#0 /var/www/example.com/vendor/symfony/http-foundation/Response.php(447): Illuminate\Foundation\Bootstrap\HandleExceptions->handleError(4096, 'Object of class...', '/var/www/example...', 447, Array)
#1 /var/www/example.com/vendor/laravel/framework/src/Illuminate/Http/Response.php(41): Symfony\Component\HttpFoundation\Response->setContent(Object(Illuminate\Routing\ResponseFactory))
#2 /var/www/example.com/vendor/symfony/http-foundation/Response.php(225): Illuminate\Http\Response->setContent(Object(Illuminate\Routing\ResponseFactory))
#3 /var/www/example.com/vendor/laravel/framework/src/Illuminate/Routing/Router.php(602): Symfony\Component\HttpFoundation\Response->__construct(Object(Illuminate\Routing\ResponseFactory))
#4 /var/www/example.com/vendor/laravel/framework/src/Illuminate/Routing/Router.php(559): Illuminate\Routing\Router->prepareResponse(Object(Illuminate\Http\Request), Object(Illuminate\Routing\ResponseFactory))
#5 /var/www/example.com/vendor/laravel/framework/src/Illuminate/Routing/Pipeline.php(30): Illuminate\Routing\Router->Illuminate\Routing\{closure}(Object(Illuminate\Http\Request))
#6 /var/www/example.com/vendor/laravel/framework/src/Illuminate/Routing/Middleware/SubstituteBindings.php(41): Illuminate\Routing\Pipeline->Illuminate\Routing\{closure}(Object(Illuminate\Http\Request))
#7 /var/www/example.com/vendor/laravel/framework/src/Illuminate/Pipeline/Pipeline.php(148): Illuminate\Routing\Middleware\SubstituteBindings->handle(Object(Illuminate\Http\Request), Object(Closure))
[...]

PR #46141, opened 2023-02-16
[10.x] Fix calling `response()` helper without args

mputkowski:
The `response()` helper as its first argument accepts `$content` which defaults to an empty string. Also, it could be called without providing any arguments, so we are able to easily return JSON from the endpoint (`response()->json(['foo' => 'bar'])`).

The problem occurs when the helper is called without args and we are not calling any `ReponseFactory`'s method (like `json()` from the previous example).

**Sure, there is a workaround by passing an empty string as a first argument, but since calling without args is allowed, it should be handled properly.**

Currently, it throws `Symfony\Component\HttpFoundation\Response::setContent(): Argument #1 ($content) must be of type ?string, Illuminate\Routing\ResponseFactory given` error.

In older versions: `Object of class Illuminate\Routing\ResponseFactory could not be converted to string`

I couldn't find a better solution that won't be a breaking change. It's not perfect but it's simple.

See: #18498 #32198

### DIFF
--- a/src/Illuminate/Routing/ResponseFactory.php
+++ b/src/Illuminate/Routing/ResponseFactory.php
@@ -273,4 +273,14 @@ class ResponseFactory implements FactoryContract
     {
         return $this->redirector->intended($default, $status, $headers, $secure);
     }
+
+    /**
+     * Get the string representation of the factory.
+     *
+     * @return string
+     */
+    public function __toString(): string
+    {
+        return '';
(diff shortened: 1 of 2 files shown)


taylorotwell:
We don't ever document returning `response()` from a route without any further methods or arguments. I don't think it makes sense to put a string cast on the response factory just to work around incorrect code?

PR #46143, opened 2023-02-16
Check if the file name of the file to be generated has the .php exten…

Jandelson:
Check if the file name to be generated has the .php extension
If so, remove.
impacts on the creation of files that are with php.php and also on the validation of reserved words.
For example, make:request array.php where array is a reserved word, but today you can create a file that is array.php.php.

modified getNameInput function


taylorotwell:

Jandelson:
Hi @taylorotwell 
Thank you your for answer.
Sorry for something about pull request.
Dont is a bug, however is an improvement to avoided the generate the files like for example array.php.php using some artisan make command.

PR #46146, opened 2023-02-16
[10.x] Added missing expression support for pluck

ahmett:
This PR fixes an unnoticed breaking change that was caused by the #44784 update.
For more explanation and example case please see: #46138


taylorotwell:
Thanks!

PR #46148, opened 2023-02-17
[10.x] Adds `date_json` validation

DarkGhostHunter:
# What?

Adds the `date_json` validation rule, which checks if the given date is one of ISO 8601 formats that Javascript or a Model JSON serialization generates by default.

```php
use Illuminate\Support\Facades\Validator;

// Before
Validator::make(['x' => '2020-01-01T19:30:51.321Z'], ['x' => 'date_format:Y-m-d\TH:i:s.vp,Y-m-d\TH:i:s.up']);

// Now
Validator::make(['x' => '2020-01-01T19:30:51.321Z'], ['x' => 'date_json']);
```

The main purpose of this is to validate ISO 8601 dates that come up from Javascript itself (milliseconds + timezone), or from an Model datetime (microsedons + timezone) that goes back to the application, without having to write the a cryptic PHP date format for a Vue/React/Angular/Livewire app.

* `1234-05-06T07:08:09.012Z`
* `1234-05-06T07:08:09.012Z+03:00`
* `1234-05-06T07:08:09.012345Z`
* `1234-05-06T07:08:09.012345+07:00`

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -576,6 +576,24 @@ trait ValidatesAttributes
     }
 
     /**
+     * Validate that an attribute matches a JSON date format.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @param  array<int, int|string>  $parameters
+     * @return bool
+     */
+    public function validateDateJson($attribute, $value, $parameters)
+    {
+        return $this->validateDateFormat($attribute, $value, [
+            'Y-m-d\TH:i:s.vp', 'Y-m-d\TH:i:s.v',
+            'Y-m-d\TH:i:s.up', 'Y-m-d\TH:i:s.u',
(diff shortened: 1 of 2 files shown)


src/Illuminate/Validation/Concerns/ValidatesAttributes.php
```diff
@@ -575,6 +575,19 @@ public function validateDateEquals($attribute, $value, $parameters)
...
 
+    /**
+     * Validate that an attribute matches a JSON date format.
+     *
+     * @param  string  $attribute
+     * @param  mixed  $value
+     * @param  array<int, int|string>  $parameters
+     * @return bool
+     */
+    public function validateDateJson($attribute, $value, $parameters)
```
ankurk91, line 586: $parameters is unused?

DarkGhostHunter: Oh snap! Yes, I will kill that.

taylorotwell:
I think this is a better candidate for a custom rule.

PR #46157, opened 2023-02-17
[10.x] Use typed properties and return types on Gate Contract

GeoSot:
On the mood of laravel 10, I am trying to apply some more descriptive typehint at Gate Contract & implementation


driesvints:
We have no plans right now to move to types properties and return types for non-userland code. Thanks

PR #46160, opened 2023-02-17
[10.x] Allow expressions in query builder

jerodev:
With the types being added, the query builder no longer allows expressions for _where_ and _aggregate_ functions. This now generates errors with phpstan or other similar tools, but works just fine.

This PR adds `ExpressionContract` typehint for these functions.

## Example
This allows using `DB::raw()` expressions for these functions

```php
Product::query()->whereIn(DB::raw('TRIM(LEADING \'0\' FROM upc)'), ['12345678', '87654321'])->get();
```
```php
Product::query()->where('category', 'toys')->avg(DB::raw('price / 100'));
```

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -3094,7 +3094,7 @@ class Builder implements BuilderContract
     /**
      * Retrieve the "count" result of the query.
      *
-     * @param  string  $columns
+     * @param  string|ExpressionContract  $columns
      * @return int
      */
     public function count($columns = '*')
@@ -3105,7 +3105,7 @@ class Builder implements BuilderContract
     /**
      * Retrieve the minimum value of a given column.
      *
-     * @param  string  $column
+     * @param  string|ExpressionContract  $column
      * @return mixed
      */
     public function min($column)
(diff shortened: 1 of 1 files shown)


src/Illuminate/Database/Query/Builder.php
```diff
@@ -745,7 +745,7 @@ public function mergeWheres($wheres, $bindings)
     /**
      * Add a basic where clause to the query.
      *
-     * @param  \Closure|string|array  $column
+     * @param  \Closure|string|array|ExpressionContract  $column
```
driesvints, line 748: Can you use FQN for all of these please?

driesvints:
@tpetry does this PR look good to you?

tpetry:
Looks good. Having expressions on `$table` is a niche use-case but I used it too for implementing lateral joins.

driesvints (approved):
Thanks @jerodev

madhurbhaiya:
Hi @jerodev Thanks for this PR

Few more type hints are missed

```joinWhere``` and ```leftJoinWhere```

In these, ```$first``` parameter can technically be a "Raw Expression" - same as ```where(``` method

tpetry:
The `newJoinClause` would be another candidate. I am currently silencing that error in PHPStan.

Issue #44430, opened 2022-10-03
Eloquent `upsert()` method doesn't work with UUID primary key


plakhin:
- Laravel Version: 9.33.0
- PHP Version: 8.1.10
- Database Driver & Version: mysql  Ver 8.0.30 for macos12.4 on arm64 (Homebrew)

### Description:
Users table migration creates the UUID `id` column: `$table->uuid('id')->primary();`.
Users model uses `Illuminate\Database\Eloquent\Concerns\HasUuids` trait.
When trying to upsert a number of records, getting the error: `PDOException: SQLSTATE[HY000]: General error: 1364 Field 'id' doesn't have a default value in...`

When using auto-incrementing `id` column all works fine.

driesvints:
Not sure this is compatible with upsert.

plakhin:
@driesvints some mention in the docs would be nice then, either on `upsert()` section, or `UUID` section, or both.

driesvints:
@plakhin we'd welcome a PR for that 👍

plakhin:
@driesvints please check https://github.com/laravel/docs/pull/8274

PR #46161, opened 2023-02-17
[10.x] Add Support for SaveQuietly and Upsert with UUID/ULID Primary Keys

TFSThiagoBR98:
Currently, the framework sets the unique ids before saving via the 'creating' event, however this does not work when trying to use saveQuietly which ignores all events or upsert which does not fire the 'creating' event, this PR changes it to not use events but be called directly in the performInsert and upsert methods, as it is done to set timestamp values.

Add some tests for this use case.

Related #44430

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1287,7 +1287,6 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
             $this->updateTimestamps();
         }
 
-        // Set the Unique IDs
         if ($this->usesUniqueIds()) {
             $this->setUniqueIds();
         }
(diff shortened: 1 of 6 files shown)


src/Illuminate/Database/Eloquent/Model.php
```diff
@@ -1286,6 +1287,10 @@ protected function performInsert(Builder $query)
             $this->updateTimestamps();
         }
 
+        if ($this->usesUniqueIds()) {
```
joelharkes, line 1290: i would expect this to fire before the creating events?

joelharkes:
The uuids are set later now in the events loop, i don't understand how this is not seen as a breaking change?

TFSThiagoBR98:
@joelharkes Good point, I'll send a PR to move the setUniqueIds before the creating event

PR #46162, opened 2023-02-18
[10.x] Add option to transform result in validation framework (for future type safety)

joelharkes:
Note: This an MVP, im willing to work out some more kinks.

Our biggest problem with the validation framework at the moment is that it does not restrict the types in the result, for example:

### Example old situation

```php
$validator = $this->getValidator(['input' => '1'], ['input' => 'required|int']);
$data = $validator->validate();
$data['input']; // = string '1';
```

Result: creating a lot of bugs in our code projects which could have been solved if the validation (as expected by any new developer to Laravel) is cast to an int type during the validation, which it is not.

You write your TestCases using integers but an external api developer might put integers or floats in strings, and thus breaking your code.

### New situation

With this RFC you can now make rules that also provide type safety, by making your own rules (I will be willing to make a library for all default types).

 for float for example:

```php
class ValidFloat implements Rules, TransformsResultRule {
    // removed passes() and message() function for readability.

    public function transform(string $attribute, mixed $value, array $context): mixed
    {
        return ((float) $value);
    }
}


$validator = $this->getValidator(['input' => '1'], ['input' => new ValidFloat()]);
$data = $validator->validate();
$data['input']; // = float 1.0;
```

### Other benefits

Rules can be changed and don't constantly have to cast values, especially usefull for date rules:

```php 
$this->validate($data, [
    'date' => [new ValidCarbonDate(), new MustBeInTheFuture()]
])
```

now `MustBeInTheFuture` in the `passes()` function can expect to always get a Carbon instance instead of a string which it will have to cast again, and the developer also has to cast again later in controller code.

### Considerations.

Does not work yet with `ValidationRule` but im willing to add support for this if necessary. shouldn't be too complicated.

### DIFF
--- a/src/Illuminate/Contracts/Validation/TransformsResultRule.php
+++ b/src/Illuminate/Contracts/Validation/TransformsResultRule.php
@@ -1,0 +1,15 @@
+<?php
+declare(strict_types=1);
+
+namespace Illuminate\Contracts\Validation;
+
+interface TransformsResultRule
+{
+    /**
+     * @param string $attribute the name of the attribute we are validating.
+     * @param mixed $value The value the attribute currently has
+     * @param array $context data context we are validating in (all other values), which will be updated as validation rules being applied.
+     * @return mixed The value we want to set for the attribute
+     */
+    public function transform(string $attribute, mixed $value, array $context): mixed;
+}
--- a/src/Illuminate/Validation/Validator.php
+++ b/src/Illuminate/Validation/Validator.php
@@ -5,6 +5,7 @@ namespace Illuminate\Validation;
 use BadMethodCallException;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Contracts\Translation\Translator;
+use Illuminate\Contracts\Validation\TransformsResultRule;
 use Illuminate\Contracts\Validation\DataAwareRule;
 use Illuminate\Contracts\Validation\ImplicitRule;
 use Illuminate\Contracts\Validation\Rule as RuleContract;
@@ -833,6 +834,11 @@ class Validator implements ValidatorContract
                 $this->messages->add($key, $this->makeReplacements(
                     $message, $key, $ruleClass, []
                 ));
+            }
+        } elseif($rule instanceof TransformsResultRule){
+            $newValue = $rule->transform($attribute, $value, $this->data);
+            if($newValue !== $value){
+                Arr::set($this->data,  $attribute, $newValue);
             }
         }
     }
--- a/tests/Integration/Validation/ValidatorTest.php
+++ b/tests/Integration/Validation/ValidatorTest.php
@@ -11,6 +11,7 @@ use Illuminate\Translation\ArrayLoader;
 use Illuminate\Translation\Translator;
 use Illuminate\Validation\DatabasePresenceVerifier;
 use Illuminate\Validation\Validator;
+use Orchestra\Testbench\TestCase;
 
 class ValidatorTest extends DatabaseTestCase
 {
--- a/tests/Integration/Validation/ValidatorTransformsResultTest.php
+++ b/tests/Integration/Validation/ValidatorTransformsResultTest.php
@@ -1,0 +1,81 @@
+<?php
+declare(strict_types=1);
+
+namespace Illuminate\Tests\Integration\Validation;
+
+use Illuminate\Contracts\Validation\Rule;
+use Illuminate\Contracts\Validation\TransformsResultRule;
+use Illuminate\Translation\ArrayLoader;
+use Illuminate\Translation\Translator;
+use Illuminate\Validation\DatabasePresenceVerifier;
+use Illuminate\Validation\Validator;
+use Orchestra\Testbench\TestCase;
+
+class ValidatorTransformsResultTest extends TestCase
+{
+    public function testRuleTransformsResult()
+    {
+        $validator = $this->getValidator(['value' => '1'], ['value' => new ValidFloat()]);
+        $data = $validator->validate();
+        $this->assertIsFloat($data['value']);
+        $this->assertSame(1.0, $data['value']);
+    }
+
+    public function testRuleWithoutTransformIsNotCalledTransformsResult()
+    {
+        $rule = \Mockery::mock(Rule::class);
+        $rule->shouldReceive('passes')->andReturn(true);
+        $validator = $this->getValidator(['value' => '1'], ['value' => $rule]);
+        $data = $validator->validate();
+        $this->assertIsString($data['value']);
+    }
+
+    public function testTransformIsNotCalledOnFailure()
+    {
+        $rule = \Mockery::mock(Rule::class, TransformsResultRule::class);
+        $rule->shouldReceive('passes')->andReturn(false);
+        $rule->shouldReceive('message')->andReturn('tset');
+        $rule->shouldNotReceive('transform');
+        $validator = $this->getValidator(['value' => '1'], ['value' => $rule]);
+        $this->assertTrue($validator->fails());
+    }
+
+    public function testSecondRuleGetsTransformedValueFromFirstRule()
+    {
+        $rule = \Mockery::mock(Rule::class);
+        $rule->shouldReceive('passes')->with('value', 1.0)->andReturn(true);
+        $rule->shouldNotReceive('transform');
+        $validator = $this->getValidator(['value' => '1'], ['value' => [new ValidFloat(), $rule]]);
+        $this->assertTrue($validator->passes());
+    }
+
+    protected function getValidator(array $data, array $rules): Validator
+    {
+        $translator = new Translator(new ArrayLoader, 'en');
+        $validator = new Validator($translator, $data, $rules);
+        $validator->setPresenceVerifier(new DatabasePresenceVerifier($this->app['db']));
+
+        return $validator;
+    }
+}
+
+
+class ValidFloatBase implements Rule {
+    public function passes($attribute, $value)
+    {
+        return true;
+    }
+
+    public function message()
+    {
+        return 'float';
+    }
+}
+
+
+class ValidFloat extends ValidFloatBase implements TransformsResultRule {
+    public function transform(string $attribute, mixed $value, array $context): mixed
+    {
+        return ((float) $value);
+    }
+}


src/Illuminate/Contracts/Validation/TransformsResultRule.php
```diff
@@ -0,0 +1,15 @@
+<?php
+declare(strict_types=1);
+
+namespace Illuminate\Contracts\Validation;
+
+interface TransformsResultRule
```
joelharkes, line 6: alternatively call it `ModifyValueRule` and call the method `modifyValue`?

ankurk91:
You can already add `passesValidation` method to your FormRequest  class and do your transformation.

joelharkes:
> You can already add `passesValidation` method to your FormRequest class and do your transformation.

Most people don't use FormRequests (to my experience) and also i would say it should be part of validation framework and not part of what a developer needs to typecast in every FormRequest.

henzeb:
So, if I want to validate if something is an int, but I want to go typesafe, I have to create an entirely new rule to do that, instead of using int? Or is this going to be part of all existing validation rules? That would be a breaking change. Especially for date/time and enum rules that suddenly return objects.

joelharkes:
> So, if I want to validate if something is an int, but I want to go typesafe, I have to create an entirely new rule to do that, instead of using int? Or is this going to be part of all existing validation rules? That would be a breaking change. Especially for date/time and enum rules that suddenly return objects. 

To avoid breaking changes it just provides us with the framework to make typesafe validation rules. 
I can make a separate library containing these rules so developer can decide to use Laravel default or my Library. Than we can see based on popularity if it's something Laravel can consider adopting internally?

joelharkes:
An alternative could be to use the With Validator interface and add the method "setData" on the Validator to update the data after passing in the rule itself

ankurk91:
https://github.com/WendellAdriel/laravel-validated-dto

donnysim:
I'd say this is a dangerous wilds to go into, the validator as is does not validate rules in order, and there's no strict type validation rules, for example `['string', 'integer']` you're expecting it to be a string consisting of numbers and it works, but now you might get an int and so on. I think the validator needs to embrace strict type rules like boolean-like and boolean-type, validation by rule order etc. before venturing into type casting.

joelharkes:
> g it to be a string consisting of numbers and it works, but now you might get an int and so on. I think the validator needs to embrace strict type rules like boolean-like and boolean-type, validation by rule order etc. before venturing into type casting.

Of course the test-case is not a complete and valid example of a real-world ValidFloat rule would look like. It's just to demonstrate how it would work. I would stick to the identical  as laravel: filter_var($value, FILTER_VALIDATE_FLOAT) 

I do agree, it could be an issue with re-running validation where you first cast to int and later to date (for example age to birthdate), which could lead to problems on the second run. The validator **does validate rules in order** which is actually beneficial so you can chain followup rules that do not have to do the type casting all over again.  as written in my example above.

joelharkes:
> https://github.com/WendellAdriel/laravel-validated-dto

This is a very boilerplatey library and could also benefit from this change. 

I try to keep best of both worlds: keep simple array input and output. (i would add typesafety to if it i could, like in TS validation libraries), but create the ability to make it **typesafe result**. Which currently was one of our most bug creating issue in Laravel.

We solved it in our own project by overriding the Validator ourselves and add this code. This saved us so many bugs! i'd like to add this to the repo so we can stop adapting and start using the framework :)

donnysim:
What I mean by validating in order is that:

```php
validator(['test' => '4'], ['test' => ['string', 'max:3', 'integer']])->validate()
```
You would expect it to validate string length to max 3, but it validates to integer max 3. Similar cases with date and date_format rule where date will pick the format no matter where the rule is, required rules don't matter where they are and so on.

joelharkes:
@donnysim  required rules indeed but it wouldn't matter in normal usecases.

honestly: validator(['test' => '4'], ['test' => ['string', 'max:3', 'integer']])->validate()

Would mess up any developer, simply your validation rule would be off here. I think its messed up now too, depending on wether you inject it with string '4' or integer 4.

I would expect integer, max:1000 that makes obviously much more sense.

donnysim:
@joelharkes the point is, if you remove the `integer` rule it will validate to string max length 3, but because integer is in the rules array, it validates to int max 3, So the string `'4`' in that case fails because the integer is higher than 3. The problem is, integer doesn't care that it must be a string, or int etc.. it will always pass as long as it looks like a integer - thus no strict type validation problem. 

For concrete real case example, you could take a numeric string that's held as machine a identifier, where the value must be a string and it's length cannot exceed e.g. say 5 - so it can be '12.34', '1.123' etc., you could think to write the rule as `['string', 'max:3', 'numeric']` as you wouldn't want to receive it as a float, but it would fail because `numeric` changes how max works no matter where it's specified and it fails because the `'12.34'` is higher than 3. If you remove `numeric` then the max validates against string length.

To clarify, I have nothing against this pull or anything, I'm just putting my 2 cents out there that from my perspective the validator is showing its age where in some cases we actually expect strict types, but it gives us no rules oob to deal with it. But again, that's just my perspective, we all have different opinions and that's fine, I had no intention on this to be offensive if this came across as such, which I apologies is it did.

rodrigopedra:
> if the validation (as expected by any new developer to Laravel) is cast to an int type during the validation

Maybe I am not new to Laravel anymore, but I'd never expect any validator to change the data I fed it to validate against. It is just counter-intuitive.

If you could point me to any poll you found to assess this statement, I got curious on the proportion on how many new, or experienced, developers expect a validation component to mutate their input data.

I am ignorant of any language, or framework, where validation (just validation, not sanitization) is implemented like this. Could you give an example of prior art for comparison?

> You write your TestCases using integers but an external api developer might put integers or floats in strings, and thus breaking your code.

Never trust user input. Even if the user is an API consumer. If unexpected input breaks your code, write new tests, and try to prepare for these cases.

---

It seems to me, I could be wrong about this, you are after sanitization. Which is not included by default. In `FormRequest`s classes, you can customize the `prepareForValidation` and `passedValidation` hooks to do some kind of sanitization on a request data. 

From my memory, sanitization was considered to be added to Laravel 5 beta, but was dropped before public release. I can't remember why, but you can surely find something on the issue tracker. From a quick search, I found the removal commit, with some comments.

https://github.com/laravel/framework/commit/924a7fcf21bbba4f4efc8e367f456cea5e4d25c1

I also have a different feeling about `FormRequest` usage. 

On my personal experience, since `FormRequest`s were introduced, the projects I found that didn't use them, were either very focused on scope (e.g. a microservices API); or legacy projects; or because the developer didn't know about them, and they start to use it immediately after being introduced.

But, as you stated, I am also talking on my own experience, and personal feeling. Of course at large it could be a different scenario, or maybe it is different on a different region.

As I can see from your PR's code, you are proposing to add an interface that would flag rules that in addition to have the responsibility to validate their data, they would also accumulate the responsibility to mutate the data they are fed with for the sake of sanitization. Is that correct?

Usually, the validator stops validating a single attribute when a preceding rule fails. When using the validator manually, by calling `Validator@passes`/`Validator@fails`, would the data be partially transformed, and partially not, if some attributes fail validation?

If the idea is just to add an escape hatch for developers who want to perform sanitization, alongside with validation, I still think it is adding a responsibility to the validator that shouldn't be there, but as it would be opt-in, I am not against it.

And if new rules would need to be added, how is that different from using an "After Validation Hook"? Couldn't sanitization be performed there, as a developer would need to manually select the transform rules anyway?

https://laravel.com/docs/10.x/validation#after-validation-hook

This would work even for the those who doesn't use `FormRequest`.

> An alternative could be to use the With Validator interface and add the method "setData" on the Validator to update the data after passing in the rule itself

`Validator@setData` is already a public method.

https://github.com/laravel/framework/blob/8f9196503151ac9b3c523e34a6c40c5671b60534/src/Illuminate/Validation/Validator.php#L1069

Maybe, an alternative, would be adding something that works similar to how the `Validator@safe` method works today. So the sanitization part is deferred to a different object that is responsible just for that. 

One unrelated note: no other files in the Laravel codebase use `declare(strict_types=1);`. You could consider sticking to the code style.

donnysim:
@rodrigopedra I've built some, wouldn't say big, but medium and high complexity projects and only in the first project used FormRequest and then moved to just having the validation in the controller. Reasons are, I just never actually need to reuse the request, in most cases that actually meant I was doing something wrong instead and in some cases it actually gets more complicated the deeper you go where having a FormRequest is actually painful, for example having to acquire a lock for a record before executing stock count query for 'max' rule etc. to further avoid deducting a quantity that no longer is available and logging the before and after change and so on. But that's just my experience.

From the validation perspective, I'd say I do not even trust anything the Laravel validator gives, even if validated, it's just too loose with what it accepts and I'd say it does a really poor job at it - you say `boolean`? then why do I get string `"1"`, you say `integer` then why do I have boolean `true` as a value? all of them passes. Using with strict type it's just a plain minefield waiting for someone to step into. Maybe I'm missing something about it that I don't know?

For being able to cast, while not the responsibility of the validator, I'd say that actually can be a really nice and useful addition, especially when working with FormData, where you could go say `['boolean_like', 'cast_to_boolean']` (<- just an improvisation) where all the variants like `'on'` are converted to bool instead and you know what you're actually dealing with instead of wishing it to be true.

joelharkes:
@rodrigopedra almost any framework does validation and sanitation at the same time:

* .net (using attributes): https://learn.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-7.0#validation-attributes
* symfony (using attributes):  https://symfony.com/doc/current/validation.html
* typescript: yup: (typesafe dynamic objects, next level stuff) https://github.com/jquense/yup

Agree with @donnysim about FormRequestUsage

joelharkes:
@rodrigopedra I do agree about  partial validation being not ideal, ideally i would also not make 'Data' part of the validator but instead being passed in every time. But i didn't want to completly refactor the Validator class.

joelharkes:
I'l see if i can rework my code to use the setData+RuleUsesValidator interface.
It could cover this solution as well.

henzeb:
> @rodrigopedra almost any framework does validation and sanitation at the same time:
> 
> * .net (using attributes): https://learn.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-7.0#validation-attributes
> * symfony (using attributes):  https://symfony.com/doc/current/validation.html
> * typescript: yup: (typesafe dynamic objects, next level stuff) https://github.com/jquense/yup
> 
> Agree with @donnysim about FormRequestUsage

This is not sanitation, laravel already does that , this is meant to transform values into their scalar types. Which is what formrequest already is doing. I still don't see why a validator should transform the values passed to it.

rodrigopedra:
> Laravel provides several different approaches to validate your **application's incoming data**. It is most common to use the validate method available on all incoming HTTP requests. 

from: https://laravel.com/docs/10.x/validation#introduction

@donnysim I guess it might come down to a preference matter. I usually use the validation method to validate if input data looks fine. Unique and exists rules are fine to me, as if they are wrong that is invalid input.

An out-of-stock rule, like the example you gave, would be a better fit for a domain validation, at least in my opinion. I would do it, wither on an action, or a service, or even in the model (for smaller less complex projects).

On an incoming request there could be several steps before checking the stock, such as validating a coupon code, customer membership, pending order placement authorization, and others. Having all this in the front door doesn't seem -- **to me** -- the right place to check for those.

Also, in the middle to high complex applications I worked on, there can be several entry-points of data, such as artisan integration, other HTTP entry-points, queued jobs that are lazy validated... So having domain logic validation close to where the data is stored seems a better fit **to me**. 

Whereas, input validation should always be done as soon as the data gets in your app. At least both of these **in my opinion**.

The `boolean` rule accepting `"1"` as a valid value is fine, as HTML checkboxes send this by default when no `value` attribute is present on them. It is well documented, and fits what the Validator proposes to do.

I can see some frustration with how the Validator currently works, but I still think it still fits well to its proposed usage. Incoming data is basically a string stream, you can deserialize it as JSON, use an XML parser, but the incoming data is still a string stream.

Maybe proposing improvements to how the Validator handle deserialized data, from different input sources, such when the input is from a JSON object, is something to be considered, to accommodate more modern usage when one does not expect a data they know is already decoded to consider `"1"` to be a boolean, and `true` to be an integer. Like how the HTTP client can handle different request or response data serialization. 

But I still don't think the discontentment on how it handles different input sources is sufficient to justify adding sanitization responsibility to it. I'd expect a validator component to validate data, not to mutate it.

On `FormRequest`s usage: I also don't reuse them a lot. Usually only among store/update endpoints of the same entity. But I still consider them very valuable on separating concerns on what should be done on each layer, and rest assured my controller's methods are very lean and always receive validated data. But it might also be a matter of personal preference.

This is from one of my own projects, a case management system:

```php
public function store(
    ManifestationRequest $request,
    Authenticatable $user,
    ResponseFactory $response,
): RedirectResponse {
    /** @var \Domain\Auth\Models\User $user */
    $ulid = ManifestationAggregate::retrieve(Str::ulid())
        ->createManifestation($request->dto(), $user)
        ->persist()
        ->uuid();

    $manifestation = Manifestation::findByUlid($ulid);

    // add pending attachments to the newly created manifestation
    PendingMedia::query()
        ->firstWithSession($request->session()->getId())
        ->moveTo($manifestation->openingActivity);

    if ($manifestation->isComplete()) {
        return $response->redirectToRoute('manifestations.show', ['manifestation' => $manifestation])
            ->with('success', 'Manifestação criada'); // manifestation created
    }

    return $response->redirectToRoute('manifestations.state', ['manifestationState' => new Pending()])
        ->with('success', 'Manifestação criada'); // manifestation created
}
```

I added the comments for clarity.

Its form request object validates 16 common attributes, and as many custom fields the user configured on their installation to have when creating a manifestation. It also gives me back my validated data as a DTO, this is where I sanitize the input regarding its data types.

On the aggregate entity (I am using spatie's event sourcing package), an opening activity is created (referenced just after), among other related entities, admins and the complainer are notified, some domain logic validation is performed, like if the manifestation is related to an older one and depending on the state should also reopen the former or just flag it, etc.

Having all this dynamic custom field validation on one layer, and having the controller just gluing together different pieces of the engine, and then deciding on the response, is a pattern I grew to like. It looks better, **to me**, than having a long sequence of steps together on the same place.

Manifestations can be created manually (this form), directly by a complainer on a public form that does not require a user, from each client website on a "contact us form" which they later call my system from an API endpoint, by e-mail ingestion, by spreadsheet ingestion (some of my clients are service providers to other companies and receive their manifestation in batches).

Each entry point has different input data shapes, needs different validation. But all of them ends up needing the same domain logic validation. So I keep those on the domain layer, and use `FormRequest`s and the validator just to check these several input sources. 

It also seems easier, **to me** at least, to test different parts of my application separately.

But as everything else, I guess it is a matter of personal preference. Also, I wouldn't consider this project neither medium nor high complexity, compared to some others I worked on.

rodrigopedra:
@joelharkes , thanks for the references. 

I skimmed through them and have some doubts you might be able to clarify.

#### .net Model Validation

It doesn't seem, to me, the validation performs type casting. Maybe type coercion is performed when assigning validated strings to the model, which would work with PHP with typed properties and no `declare(strict_types=1)` declaration.

For example, on their page on how to create a custom validation attribute:

https://learn.microsoft.com/en-us/aspnet/core/mvc/models/validation?view=aspnetcore-7.0#custom-attributes

In the code snippet, the value is coerced to a date object, before checking if the release date of a movie is within the allowed range. 

The coerced data inside the validation rule is not used anywhere else. It seems very similar to what we have today with Laravel custom rules, where one has to coerce the data to be validated themselves, and don't reuse it.

#### Symfony

On the URL linked, there is no mention input data is automatically converted/parsed to an expected data type.

Its `IsTrue` validation constraint accepts `true`, `1`, `"1"`, much like Laravel. From this quick skim, it seems to serve a similar purpose than Laravel's Validator.

https://symfony.com/doc/current/reference/constraints/IsTrue.html

For example, on all examples, at least on the link you sent, and the one above, all properties are untyped. Do you happen to know if they are coerced as the validated data-type or as strings?

#### yup

From their README they say they are *"a schema builder for runtime value parsing and validation"*, so besides doing input validation parsing/sanitization is already a goal they propose to achieve.

I liked the fluent way to define rules. Didn't know this package before, thanks. That is definitely a fine example on how sanitization and validation can live together.

---

As I said on my first comment ( https://github.com/laravel/framework/pull/46162#issuecomment-1437052266 )

> If the idea is just to add an escape hatch for developers who want to perform sanitization, alongside with validation, I still think it is adding a responsibility to the validator that shouldn't be there, but as it would be opt-in, I am not against it.

My 2 cents is that sanitization doesn't belong to the Validator component. But I don't oppose adding as is proposed in this PR. To me, personally, either a new Validation component, or a method like `Validator@safe`, or a dedicated Sanitization component, seems to be better ideas.

What triggered me to comment was adding to the discussion based on "developers expect this", or "most developers don't use that", which are not enough to justify its proposal, IMHO.

rodrigopedra:
@donnysim you are right, sorry for all the noise.

joelharkes:
Alternative solution would be to add an additional method on the validator:

```php
protected function setValue($attribute, $value) {
   return Arr::set($this->data, $attribute, $value);
}
```

i could use that  together with `ValidatorAwareRule` inteface to achieve what i need.

Would you prefer this way of going?

Alternativly i could both keep the 'originalData' and 'transformedData' attributes on the validator. so that every loop through the validator it starts with the 'originalData' but writes back to the transformedData and returns this. but That is a very opiniated solution, that i would almost suggest adding all the typesafe rules to the base Laravel framework as well.

rodrigopedra:
@joelharkes 

I guess you can already do something similar with the current interfaces

```php
<?php

namespace App\Rules;

use Closure;
use Illuminate\Contracts\Validation\DataAwareRule;
use Illuminate\Contracts\Validation\ValidationRule;
use Illuminate\Contracts\Validation\ValidatorAwareRule;
use Illuminate\Validation\Validator;

class BooleanCast implements ValidationRule, DataAwareRule, ValidatorAwareRule
{
    public readonly array $data;
    public readonly Validator $validator;

    public function validate(string $attribute, mixed $value, Closure $fail): void
    {
        $this->validator->setData([...$this->data, $attribute => boolval($value)]);
    }

    public function setData(array $data)
    {
        $this->data = $data;
    }

    public function setValidator(Validator $validator)
    {
        $this->validator = $validator;
    }
}
```

The validator and its data will be passed to the rule instance before it is checked. 

It also works with rules implementing either `Illuminate\Contracts\Validation\Rule` or, the newer `Illuminate\Contracts\Validation\ValidationRule`.

A simpler version, using only the `Illuminate\Contracts\Validation\ValidatorAwareRule` interface.

```php
<?php

namespace App\Rules;

use Closure;
use Illuminate\Contracts\Validation\ValidationRule;
use Illuminate\Contracts\Validation\ValidatorAwareRule;
use Illuminate\Validation\Validator;

class BooleanCast implements ValidationRule, ValidatorAwareRule
{
    public readonly Validator $validator;

    public function validate(string $attribute, mixed $value, Closure $fail): void
    {
        $this->validator->setData([...$this->validator->getData(), $attribute => boolval($value)]);
    }

    public function setValidator(Validator $validator)
    {
        $this->validator = $validator;
    }
}
```

joelharkes:
@rodrigopedra  `setData` does way more than replace the data, it would not be a performance viable option.

```php
    public function setData(array $data)
    {
        $this->data = $this->parseData($data);

        $this->setRules($this->initialRules);

        return $this;
    }
```

rodrigopedra:
@joelharkes neither setting an arbitrary attribute would be viable. 

As you can see from parseData implementation, dots and asterisks are replaced by placeholders before storing into Validators  data property, one of those is random generated.

You consider parsing the attribute the same way when setting it though.

joelharkes:
> @joelharkes neither setting an arbitrary attribute would be viable.
> 
> As you can see from parseData implementation, dots and asterisks are replaced by placeholders before storing into Validators data property, one of those is random generated.
> 
> You could consider parsing the attribute the same way when setting it, though.

The attribute name is ok i would just receive `users.1.name` for example but the validator would re-parse the whole rules if this happens and also re-iterate the data array. which just creates a quadratic increase of calculations.

taylorotwell:

PR #46164, opened 2023-02-18
People Who Return View will no need to declare Response Again

lwinmaungmaung:
Current Laravel Resource Controllers are only Response. When peoples are using return view(), they will need to redeclare the View, or they will cause an error. So, I added return **Response | View** then they don't need to redeclare or solve the error every time they create a resource controller.


driesvints:
Feel free to do this in your own app 👍

PR #46166, opened 2023-02-18
Use mixed return type on controller stubs

taylorotwell:
This pull request updates the controller stubs to use a `mixed` return type to accommodate the various types of return values a controller can return in Laravel.

This allows developers to return anything they want without an error by default, but they can easily tighten up the types as they decide which each controller endpoint will actually return.

### DIFF
--- a/src/Illuminate/Routing/Console/stubs/controller.api.stub
+++ b/src/Illuminate/Routing/Console/stubs/controller.api.stub
@@ -10,7 +10,7 @@ class {{ class }} extends Controller
     /**
      * Display a listing of the resource.
      */
-    public function index(): mixed
+    public function index()
     {
         //
     }
@@ -18,7 +18,7 @@ class {{ class }} extends Controller
     /**
      * Store a newly created resource in storage.
      */
-    public function store(Request $request): mixed
+    public function store(Request $request)
     {
         //
     }
@@ -26,7 +26,7 @@ class {{ class }} extends Controller
     /**
      * Display the specified resource.
      */
-    public function show(string $id): mixed
+    public function show(string $id)
     {
         //
     }
@@ -34,7 +34,7 @@ class {{ class }} extends Controller
     /**
      * Update the specified resource in storage.
      */
-    public function update(Request $request, string $id): mixed
+    public function update(Request $request, string $id)
     {
         //
     }
@@ -42,7 +42,7 @@ class {{ class }} extends Controller
     /**
      * Remove the specified resource from storage.
      */
-    public function destroy(string $id): mixed
+    public function destroy(string $id)
     {
         //
     }
--- a/src/Illuminate/Routing/Console/stubs/controller.api.stub
+++ b/src/Illuminate/Routing/Console/stubs/controller.api.stub
@@ -4,14 +4,13 @@ namespace {{ namespace }};
 
 use {{ rootNamespace }}Http\Controllers\Controller;
 use Illuminate\Http\Request;
-use Illuminate\Http\Response;
 
 class {{ class }} extends Controller
 {
     /**
      * Display a listing of the resource.
      */
-    public function index(): Response
+    public function index(): mixed
     {
         //
     }
@@ -19,7 +18,7 @@ class {{ class }} extends Controller
     /**
      * Store a newly created resource in storage.
      */
-    public function store(Request $request): Response
+    public function store(Request $request): mixed
     {
         //
     }
@@ -27,7 +26,7 @@ class {{ class }} extends Controller
     /**
      * Display the specified resource.
      */
-    public function show(string $id): Response
+    public function show(string $id): mixed
     {
         //
     }
@@ -35,7 +34,7 @@ class {{ class }} extends Controller
     /**
      * Update the specified resource in storage.
      */
-    public function update(Request $request, string $id): Response
+    public function update(Request $request, string $id): mixed
     {
         //
     }
@@ -43,7 +42,7 @@ class {{ class }} extends Controller
     /**
      * Remove the specified resource from storage.
      */
-    public function destroy(string $id): Response
+    public function destroy(string $id): mixed
     {
         //
     }
--- a/src/Illuminate/Routing/Console/stubs/controller.invokable.stub
+++ b/src/Illuminate/Routing/Console/stubs/controller.invokable.stub
@@ -4,14 +4,13 @@ namespace {{ namespace }};
 
 use {{ rootNamespace }}Http\Controllers\Controller;
 use Illuminate\Http\Request;
-use Illuminate\Http\Response;
 
 class {{ class }} extends Controller
 {
     /**
      * Handle the incoming request.
      */
-    public function __invoke(Request $request): Response
+    public function __invoke(Request $request): mixed
     {
         //
     }
--- a/src/Illuminate/Routing/Console/stubs/controller.invokable.stub
+++ b/src/Illuminate/Routing/Console/stubs/controller.invokable.stub
@@ -10,7 +10,7 @@ class {{ class }} extends Controller
     /**
      * Handle the incoming request.
      */
-    public function __invoke(Request $request): mixed
+    public function __invoke(Request $request)
     {
         //
     }
--- a/src/Illuminate/Routing/Console/stubs/controller.model.api.stub
+++ b/src/Illuminate/Routing/Console/stubs/controller.model.api.stub
@@ -11,7 +11,7 @@ class {{ class }} extends Controller
     /**
      * Display a listing of the resource.
      */
-    public function index(): mixed
+    public function index()
     {
         //
     }
@@ -19,7 +19,7 @@ class {{ class }} extends Controller
     /**
      * Store a newly created resource in storage.
      */
-    public function store({{ storeRequest }} $request): mixed
+    public function store({{ storeRequest }} $request)
     {
         //
     }
@@ -27,7 +27,7 @@ class {{ class }} extends Controller
     /**
      * Display the specified resource.
      */
-    public function show({{ model }} ${{ modelVariable }}): mixed
+    public function show({{ model }} ${{ modelVariable }})
     {
         //
     }
@@ -35,7 +35,7 @@ class {{ class }} extends Controller
     /**
      * Update the specified resource in storage.
      */
-    public function update({{ updateRequest }} $request, {{ model }} ${{ modelVariable }}): mixed
+    public function update({{ updateRequest }} $request, {{ model }} ${{ modelVariable }})
     {
         //
     }
@@ -43,7 +43,7 @@ class {{ class }} extends Controller
     /**
      * Remove the specified resource from storage.
      */
-    public function destroy({{ model }} ${{ modelVariable }}): mixed
+    public function destroy({{ model }} ${{ modelVariable }})
     {
         //
     }
--- a/src/Illuminate/Routing/Console/stubs/controller.model.api.stub
+++ b/src/Illuminate/Routing/Console/stubs/controller.model.api.stub
@@ -5,14 +5,13 @@ namespace {{ namespace }};
 use {{ namespacedModel }};
 use {{ rootNamespace }}Http\Controllers\Controller;
 use {{ namespacedRequests }}
-use Illuminate\Http\Response;
 
 class {{ class }} extends Controller
 {
     /**
      * Display a listing of the resource.
      */
-    public function index(): Response
+    public function index(): mixed
     {
         //
     }
@@ -20,7 +19,7 @@ class {{ class }} extends Controller
     /**
      * Store a newly created resource in storage.
      */
-    public function store({{ storeRequest }} $request): Response
+    public function store({{ storeRequest }} $request): mixed
     {
         //
     }
@@ -28,7 +27,7 @@ class {{ class }} extends Controller
     /**
      * Display the specified resource.
      */
-    public function show({{ model }} ${{ modelVariable }}): Response
+    public function show({{ model }} ${{ modelVariable }}): mixed
     {
         //
     }
@@ -36,7 +35,7 @@ class {{ class }} extends Controller
     /**
      * Update the specified resource in storage.
      */
-    public function update({{ updateRequest }} $request, {{ model }} ${{ modelVariable }}): Response
+    public function update({{ updateRequest }} $request, {{ model }} ${{ modelVariable }}): mixed
     {
         //
     }
@@ -44,7 +43,7 @@ class {{ class }} extends Controller
     /**
      * Remove the specified resource from storage.
      */
-    public function destroy({{ model }} ${{ modelVariable }}): Response
+    public function destroy({{ model }} ${{ modelVariable }}): mixed
     {
         //
     }
--- a/src/Illuminate/Routing/Console/stubs/controller.model.stub
+++ b/src/Illuminate/Routing/Console/stubs/controller.model.stub
@@ -5,15 +5,13 @@ namespace {{ namespace }};
 use {{ namespacedModel }};
 use {{ rootNamespace }}Http\Controllers\Controller;
 use {{ namespacedRequests }}
-use Illuminate\Http\RedirectResponse;
-use Illuminate\Http\Response;
 
 class {{ class }} extends Controller
 {
     /**
      * Display a listing of the resource.
      */
-    public function index(): Response
+    public function index(): mixed
     {
         //
     }
@@ -21,7 +19,7 @@ class {{ class }} extends Controller
     /**
      * Show the form for creating a new resource.
      */
-    public function create(): Response
+    public function create(): mixed
     {
         //
     }
@@ -29,7 +27,7 @@ class {{ class }} extends Controller
     /**
      * Store a newly created resource in storage.
      */
-    public function store({{ storeRequest }} $request): RedirectResponse
+    public function store({{ storeRequest }} $request): mixed
     {
         //
     }
@@ -37,7 +35,7 @@ class {{ class }} extends Controller
     /**
      * Display the specified resource.
      */
-    public function show({{ model }} ${{ modelVariable }}): Response
+    public function show({{ model }} ${{ modelVariable }}): mixed
     {
         //
     }
@@ -45,7 +43,7 @@ class {{ class }} extends Controller
     /**
      * Show the form for editing the specified resource.
      */
-    public function edit({{ model }} ${{ modelVariable }}): Response
+    public function edit({{ model }} ${{ modelVariable }}): mixed
     {
         //
     }
@@ -53,7 +51,7 @@ class {{ class }} extends Controller
     /**
      * Update the specified resource in storage.
      */
-    public function update({{ updateRequest }} $request, {{ model }} ${{ modelVariable }}): RedirectResponse
+    public function update({{ updateRequest }} $request, {{ model }} ${{ modelVariable }}): mixed
     {
         //
     }
@@ -61,7 +59,7 @@ class {{ class }} extends Controller
     /**
      * Remove the specified resource from storage.
      */
-    public function destroy({{ model }} ${{ modelVariable }}): RedirectResponse
+    public function destroy({{ model }} ${{ modelVariable }}): mixed
     {
         //
     }
--- a/src/Illuminate/Routing/Console/stubs/controller.model.stub
+++ b/src/Illuminate/Routing/Console/stubs/controller.model.stub
@@ -11,7 +11,7 @@ class {{ class }} extends Controller
     /**
      * Display a listing of the resource.
      */
-    public function index(): mixed
+    public function index()
     {
         //
     }
@@ -19,7 +19,7 @@ class {{ class }} extends Controller
     /**
      * Show the form for creating a new resource.
      */
-    public function create(): mixed
+    public function create()
     {
         //
     }
@@ -27,7 +27,7 @@ class {{ class }} extends Controller
     /**
      * Store a newly created resource in storage.
      */
-    public function store({{ storeRequest }} $request): mixed
+    public function store({{ storeRequest }} $request)
     {
         //
     }
@@ -35,7 +35,7 @@ class {{ class }} extends Controller
     /**
      * Display the specified resource.
      */
-    public function show({{ model }} ${{ modelVariable }}): mixed
+    public function show({{ model }} ${{ modelVariable }})
     {
         //
     }
@@ -43,7 +43,7 @@ class {{ class }} extends Controller
     /**
      * Show the form for editing the specified resource.
      */
-    public function edit({{ model }} ${{ modelVariable }}): mixed
+    public function edit({{ model }} ${{ modelVariable }})
     {
         //
     }
@@ -51,7 +51,7 @@ class {{ class }} extends Controller
     /**
      * Update the specified resource in storage.
      */
-    public function update({{ updateRequest }} $request, {{ model }} ${{ modelVariable }}): mixed
+    public function update({{ updateRequest }} $request, {{ model }} ${{ modelVariable }})
     {
         //
     }
@@ -59,7 +59,7 @@ class {{ class }} extends Controller
     /**
      * Remove the specified resource from storage.
      */
-    public function destroy({{ model }} ${{ modelVariable }}): mixed
+    public function destroy({{ model }} ${{ modelVariable }})
     {
         //
     }
--- a/src/Illuminate/Routing/Console/stubs/controller.nested.api.stub
+++ b/src/Illuminate/Routing/Console/stubs/controller.nested.api.stub
@@ -6,14 +6,13 @@ use {{ namespacedModel }};
 use {{ rootNamespace }}Http\Controllers\Controller;
 use {{ namespacedParentModel }};
 use Illuminate\Http\Request;
-use Illuminate\Http\Response;
 
 class {{ class }} extends Controller
 {
     /**
      * Display a listing of the resource.
      */
-    public function index({{ parentModel }} ${{ parentModelVariable }}): Response
+    public function index({{ parentModel }} ${{ parentModelVariable }}): mixed
     {
         //
     }
@@ -21,7 +20,7 @@ class {{ class }} extends Controller
     /**
      * Store a newly created resource in storage.
      */
-    public function store(Request $request, {{ parentModel }} ${{ parentModelVariable }}): Response
+    public function store(Request $request, {{ parentModel }} ${{ parentModelVariable }}): mixed
     {
         //
     }
@@ -29,7 +28,7 @@ class {{ class }} extends Controller
     /**
      * Display the specified resource.
      */
-    public function show({{ parentModel }} ${{ parentModelVariable }}, {{ model }} ${{ modelVariable }}): Response
+    public function show({{ parentModel }} ${{ parentModelVariable }}, {{ model }} ${{ modelVariable }}): mixed
     {
         //
     }
@@ -37,7 +36,7 @@ class {{ class }} extends Controller
     /**
      * Update the specified resource in storage.
      */
-    public function update(Request $request, {{ parentModel }} ${{ parentModelVariable }}, {{ model }} ${{ modelVariable }}): Response
+    public function update(Request $request, {{ parentModel }} ${{ parentModelVariable }}, {{ model }} ${{ modelVariable }}): mixed
     {
         //
     }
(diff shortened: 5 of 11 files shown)


datlechin:
Developers can do it on their own project, I think no need to change to `mixed` return type in stub

taylorotwell:
@datlechin I'm well aware they can, the question is whether the default (as it stands currently) should be to error when a developer creates a new controller and tries to return a `View`.

yungifez:
Noob question

Doesn't this kill the whole purpose of return types?

What's the difference between this and no return types at all

And cant we instead do something like this

Response|View etc

taylorotwell:
@yungifez one could definitely make that argument; however, we have added return types to all other parts of the skeleton so it may feel weird to not have them on controllers at all. Having the mixed type in place at least gives you some text to quickly replace with your own type hint if you wish.

martinbean:
@taylorotwell Do views and Eloquent API resources both implement the `Responsible` interface? Would that suffice as a “looser”—but better than `mixed`—return type?

taylorotwell:
@martinbean `Illuminate\Http\Response` itself isn't responsible - neither is `View`.

tontonsb:
I agree 100% — the default return type should reflect what the framework is willing to accept here. And it accepts all the views, arrays, strings and everything else.

@yungifez `mixed` does not include `void`, it's a bit stricter than nothing. And it's a decent placeholder :)

LucaRed:
I think the return type `mixed` doesn't add anything of value.

Personally, I always return views for normal web controllers, and responses for API controllers (isn't this the "Laravel way"?). If I had to choose, I'd go with `Response|View`.

jonathanmacgregor:
There could be an opportunity here to direct users to change the return types and educate them at the same time. Adding a note to the head of the stub could point them in the direction of what returns types mean.

They can always be removed in the stubs for users who don't need direction.

taylorotwell:
@LucaRed I return arrays from controllers pretty frequently. Wouldn't work with that type-hint you suggested.

I personally lean @tontonsb's direction in that the `mixed` type simply reflects reality. The "reality" is controllers can return a wide variety of things in Laravel - why not make the default type hint reflect that?

phpfour:
Returning `mixed` is a necessary compromise that we as a community will need to accept. It will allow newcomers to have a smooth experience, while experienced devs can (will) change them anyways for their use cases.

Saeed-Pooyanfar:
I don't know if this was possible, but can we extend `View` by `Response` class, or can we create an interface for different response types?
I don't like the `mixed` type 😔

ashabed:
> @martinbean `Illuminate\Http\Response` itself isn't responsible - neither is `View`.

How about changing that? Make Response and View implement Responsable, as they can both be used for returning from controllers. 
Then set the return type to Responsable instead of mixed. Seems like things would be more clear that way.

joniand:
Yeah for me too 99% of the times a View type would be more appropriate so having the default being Response is a bit annoying, as much sense as it makes. I think mixed would be better though, this way we also don't need to delete imports etc. if we want to change it...

martinbean:
> @martinbean `Illuminate\Http\Response` itself isn't responsible - neither is `View`.

Fair enough. Should the `Responsable` interface be applied to views in that case if they can, in fact, be used as responses?

yungifez:
I think looking at all these comments mixed is the way to go

LucaRed:
Are there any differences between specifying `mixed` and not specifying any return type at all? (other than than not allowing `void` as @tontonsb said)

If there are no differences, we could go with `mixed` for consistency reasons (and also to make users aware that they can return pretty much anything).

jorqensen:
Using `mixed` is like using TypeScript and throwing an `: any` in there, it's simply pointless. I am not a fan of this PR, even though I use Blade and can see the annoyance.

Hopefully a better solutions could come out of this, specifically something along of what @martinbean suggests / a common interface as the return type.

dinhquochan:
I think it should be a `Responable` interface and `Illuminate\View\View` or other stuff should be implemented `Responable` interface. `mixed` type is terrible, like `any` in TypeScript 🥲

jrd-lewis:
Food for thought, what about using the ```Illuminate\Contracts\Support\Renderable``` interface as the return type?

iksaku:
For the sake of further clarification, we can all refer to the [Documentation on `mixed` type](https://www.php.net/manual/en/language.types.mixed.php).

@LucaRed you’re correct. No return type and `mixed` mean basically the same from _the framework’s perspective_ that something must be returned from the controller, and as it is Laravel, _any value type_ that is returned from a controller function will properly be transformed into something usable by the framework.

I would argue that to the good initiative to strict-type anything, this is that little exception to the rule where we’re actually typing something for the _framework_ to process initially.

We can always adjust typings as must of us would be already doing, or we can update our projects stubs manually to be more strict.

Talking from a _community_ standpoint, there’s no majority that will use X or Y; every developer will be using their own conventions, and reflecting that from a _stubs_ seems really accurate as @tonto

manumrtf:
Not a fan of using Mixed, I would rather change the type manually, it's not that big of a deal.

igorkosteski:
I think that `mixed` is bead a idea. What about `Illuminate\Contracts\View\View`  to be default return type,
and if you use `--api` other then  excluding the create and edit, to change the return return type to all the method of the Controller to be `Illuminate\Http\Response`

yungifez:
> I think that `mixed` is bead a idea. What about `Illuminate\Contracts\View\View` to be default return type, and if you use `--api` other then excluding the create and edit, to change the return return type to all the method of the Controller to be `Illuminate\Http\Response`

I don't see what is so bad about returning mixed as a return type.

In my opinion, the framework has functioned without type hints for a long time. You are not accounting for cases like Inertia that does not make use of API controllers and I'm speculating is not of type`Illuminate\Contracts\View\View`

Mixed in this case just acts as a decorator to prevent a partially typed application and follow consistent design

ConsoleTVs:
This conversation is leading nowhere. The reality is:

**Laravel accepts any return types in controller handlers (strings, arrays, views, Response, etc.). The ideal return type, as for the current API is: `mixed`.**

On another hand, this provides little to no gain at all, makes static typing look useless, and that's no far from the truth. Reality is, that this API is not something I would use for future versions. Perhaps, if you want to keep a similar API, we could use mixed internally and leave the surface API fully typed. There's likely the possibility to do this:

```php
public function foo(): Responsable {
    return response('foo');
}

public function foo(): Responsable {
    return response()->noContent();
}

public function foo(): Responsable {
}

public function foo(): Responsable {
    return view('foo');
}

public function foo(User $user): Responsable {
    return new UserResource($user);
}

public function foo(): Responsable {
    return UserResource::collection(User::all());
}
```

In reality, the response helper will have mixed as it's parameter type, but on the surface we can be sure that the returned type is correct.

There seems to be no correct answer on this but rather multiple points of view.

inxilpro:
I personally don't find that return types are very useful on controllers actions, so my first choice would be to just leave them out of that stub (I'm in favor of return types in many other places—but only where they bring value). That said, `mixed` is a fine option if the consensus is that controller actions *should* have a return type by default.

mabdullahsari:
Honestly, since Laravel offers the option to return almost anything from a `Controller`, it'd be better to remove the type-hints altogether from the `Controller` stubs. There's just no way to cover every single use case and combination. Trying to do so will result in a Cartesian Explosion.

People can still add a return type of their own if they wish, or just ignore it.

taylorotwell:
Another option: https://github.com/laravel/framework/pull/46166

Use `Responsable` after adding it to more classes.

wlepinski:
Isn't there a common interface that both Response and View implements that can be used here as the return type?

The main problem I see is that, even with a common interface the only acceptable type here is `mixed` because the framework accepts also strings and other literals as valid returns from a controller method. 

I agree with the sentiment of adding a `mixed` to solve the problem, it feels like adding an `any` to a typescript file.

taylorotwell:
**Important Note:** any comment along the lines of "`mixed` is useless" or "you might as well not have a type hint" are thoroughly missing the point.

The question is not what should the type be in your final application you ship to production. We likely all agree `mixed` isn't useful in that scenario. The question is what should the type be in a **brand new** controller created by the `make` command that the developer can then easily modify. That is an important distinction.

wlepinski:
> The question is what should the type be in a **brand new** controller created by the `make` command that the developer can then easily modify.

A union type of all valid possible return types would be feasible?

yungifez:
> Another option: #46167
> 
> Use `Responsable` after adding it to more classes.

I thought controllers could return strings, wouldn't this PR prevent this.

Orrison:
I think 'mixed' is fine. But it may be worth setting the skeleton to the most common scenarios: 'Response|View'.

The user can then change it to what makes sense for their return. Removing the one that does not make sense, changing it to something else, etc. It feels like that may be better than 'mixed'.

I could see 'mixed' making sense too though. Almost like the framework skeleton is saying "you can return whatever here and it's up to you to specify if you want".

tontonsb:
@ConsoleTVs What's the benefit in making things like `return User::all();`, `return User::first();`, `return 'OK';` and `return ['status' => 'success'];` forbidden/not recommended?

@wlepinski the full list of [handled return types](https://github.com/laravel/framework/blob/master/src/Illuminate/Routing/Router.php#L885) would be too verbose. And somewhat useless as it accepts `stdClass` among others, so it's nearly equivalent to `mixed` anyways...

Wirone:
> The question is what should the type be in a brand new controller created by the make command that the developer can then easily modify.

I believe that question is different: what should the type be to be consistent with the consumer of the return type, which is framework itself. It does not make sense to narrow it to `Response` when framework accepts views, arrays, strings etc. It's a broken contract.

As far as I don't like `mixed` I think it's proper default type, that can be later changed by developer if needed. But I believe `Response` would be better in general (as a contract).

ConsoleTVs:
@tontonsb Well, all this move is basically to create a static-typing feeling. If we look at how this is solved in other frameworks of typed languages (eg. Rust, Swift, etc) we can see most create an interface (Responsable) or wrap all possible responses in a concrete type (Response).

Some frameworks in Rust, allow returning strings, or arrays but that's because you can actually implement interfaces on primitive types (you could implement Responsable for array for example), and you can't in PHP.

Regardless, I have always used explicit response(), redirect(), API Resources for years with Laravel, and has helped me keep return types correct. At the end of the day, I think it makes it a cleaner intention about what you are returning rather than "just give me whatever you can and I'll try to figure out how the return should look like" kind of thing that we have now.

**I wanna make it clear this is all just my opinion on this matter.**

wlepinski:
> And somewhat useless as it accepts `stdClass` among others, so it's nearly equivalent to `mixed` anyways...

I see your point, totally valid. I just think we should use a common interface here where all allowed response types implement.

jrd-lewis:
> The question is not what should the type be in your final application you ship to production. We likely all agree `mixed` isn't useful in that scenario. The question is what should the type be in a **brand new** controller created by the `make` command that the developer can then easily modify.
> 
> That is an important distinction. In a brand new controller that **literally** contains no code, `mixed` is potentially a viable candidate for the return type because it sends a signal to the developer - "Hey, I don't know what your final return type is going to be so I'm just `mixed` for now, but you're free to tighten this up once you start writing code."

The API controller stubs could probably remain with Response as the return type since they shouldn't be returning views at that point. For the other controller stubs, I would go either for the ```Responsable``` option you presented or do a union return type of ```Response|View```.

LucaRed:
@jrd-lewis API controllers can simply return arrays, which are then transformed to Json. A `Responsable` interface wouldn't cover this (IMO) common scenario.

yungifez:
> @jrd-lewis API controllers can simply return arrays, which are then transformed to Json. A `Responsable` interface wouldn't cover this (IMO) common scenario.

I agree, there are a lot of valid variations on how controllers can be used

Using mixed should be the way to go

It makes real world sense

jrd-lewis:
> @jrd-lewis API controllers can simply return arrays, which are then transformed to Json. A `Responsable` interface wouldn't cover this (IMO) common scenario.

Taylor added the interface to JsonResponse in the other PR; so, yes, it would cover that scenario.

Wirone:
> Taylor added the interface to JsonResponse in the other PR; so, yes, it would cover that scenario.

No, it wouldn't. Raw `array` does not implement `Responsable`, and you can return it from the controller.

yungifez:
> > @jrd-lewis API controllers can simply return arrays, which are then transformed to Json. A `Responsable` interface wouldn't cover this (IMO) common scenario.
> 
> Taylor added the interface to JsonResponse in the other PR; so, yes, it would cover that scenario.

Wouldn't it still throw an error as primitive arrays do not implement Responsable?

tabuna:
If you do not support the use of the "mixed" type and instead prefer to use the "Responsible" interface, it's important to note that this interface is likely to be widely accepted in almost every class.

For instance, with the Responsible interface, you can return a model, a collection, and other types of responses. This means that even your favorite packages will need to integrate this interface into their code. After all, no open-source author is likely to ask you to remove typing, so it will be more convenient for them to add the Responsible interface to their codebase.

In my opinion, I believe that using the "mixed" type or having no type at all is the most appropriate approach.

osuagwu:
> My experience of Laravel is that things work out of the box due to sensible defaults. In this case I feel that 'mixed' return type ensures that the generated controller works out of the box. We should make the most of our dynamic PHP when it makes sense.

MrWeb:
> **Important Note:** any comment along the lines of "`mixed` is useless" or "you might as well not have a type hint" are **thoroughly missing the point**.
> 

I agree, probably a better upgrade would be to return 'mixed' and extend Artisan with a flag to specify the return type, ie: `make:controller DemoController --type=response`

tweichart:
Not to make this too academic but from a SOLID perspective Responsable makes more sense than mixed, where the latter would give more options to return different types of course but then on the other hand too many I feel. Not sure when e.g. I'd want to return an int here, it should not be allowed by means of the return types definition already though, otherwise (as others mentioned already) types don't make much sense here.

Only valid argument against the Interface I've seen here was the array being returned, tbh I didn't even know that this is possible until now and should probably not be the case to start with. In this case I don't feel it's too much of a limitation though, this is not talking about a breaking change for existing classes but only a stub. Meaning, future controllers would be discouraged to return arrays, which again imo wouldn't be the worst idea. Having an typed return bound to an interface would make most sense therefore.

kevinvanboeckholtz:
make:controller NameController --response, --view, --responsable --mixed, --api

One command to rule them all ~ Lord of the rings

dillingham:
I think mixed makes the least friction out of the gate. Maybe laravel new can have a flag for changing the controller stubs for smoothing out the customization story. --controller=inertia

rodrigopedra:
Note that `Illuminate\Http\RedirectResponse` does not extend `Illuminate\Http\Response`, neither does `Illuminate\Http\Response`.

Having anything narrower than `mixed` could:

- be a pain point when one wants to quickly prototype something and have to remember changing the return type all the time when they are just testing an endpoint returning whatever they want to debug/build it, like strings, array, models, etc.
- lead to an avalanche of forums requests on row to return things that are supported, but the chosen return type does not encompass it, as many users assume the default is somewhat a "best practice"

As others have already said, a controller is fundamentally consumed by the framework, so its methods' return type should satisfy the framework expectations.

IMHO, `mixed` is the way to go, if having return types on all generated content is the desired goal.

taylorotwell:
Just going to remove return types on controller method stubs.

dillingham:
Wouldn't that make phpstan complain?

taylorotwell:
@dillingham complain about what? the stubs have no code in them and thus no return type.

PovilasKorop:
@taylorotwell thanks for taking the time to consider various approaches, and sorry if my tweet caused the headache here.

I agree that removing types is probably the best because historically it worked with almost whatever you would return:

- view() / response()->view()
- json
- collection/model/array - "magically" transformed into json
- string "It works"
- Inertia::render()
- redirect
- etc.

madebydanda:
I think the default code should be a good example from where anyone (junior or senior developers) can start from. If we use `mixed` as a return type we might "teach" juniors that `mixed` is a correct return type and that they can keep it as it is because it works. They might even think it's a good return type to use in other circumstances.

Sure getting an error when returning a `view` might be annoying but default stubs shouldn't please everyone or even anyone. We use controllers to handle a "Request" and therefore provide a "Response". This for me is the correct way to think.

If the "Response" causes an error I will learn to modify my response based on my needs and not just use `mixed` because it works.

Again, I understand the willingness to avoid confusion but I always considered the default code of Laravel as "good practice" to follow or learn from. Using `mixed` as a return type to avoid errors wouldn't be for me "good practice" or the "right" way to do things.

rodrigopedra:
@madebydanda on his last commit, Taylor actually removed any return types, you can see on the changes tab.

As the merge title didn't change, it can be misleading.

madebydanda:
@rodrigopedra Yes just saw that, well I guess my comment can also apply to the "no return type" solution.

rodrigopedra:
Well, `JsonResponse`, and `RedirectResponse`  don't inherit from `Response`. Neither `View` does. 

I would say in a usual Laravel app those are more common than a plain `Response`. Should we add those as union types?

Having `Response` as the default would be misleading in the very same sense some users would assume that is the recommended way to do things.

Also, Laravel strives for DX. Narrowing the return type just because it feels "correct", while the framework can handle many other return types, is neither correct, in the sense it could imply the framework expect you to return a specific data type, but also goes against its philosophy, of moving out the way to let an app code shine, not to make a developer worry about complying to a specification, IMO.

By the way, what is a response? The request object is a wrapper that flats out many PHP inconsistencies, but should a response be wrapped onto a specific class? 

Should we move away from model route binding, and have the request object retrieve the route parameters, as a controller method without a request wouldn't feel "correct"?

I understand your concern, but I personally prefer not having a return type suggested than having something like: `Response|JsonResponse|RedirectResponse|Responsable|static|iterable|\stdClass`, to highlight everything a request can return as a response. That would be more correct, in the sense the framework is ready to handle any of these return types for you.

madebydanda:
@rodrigopedra that wasn't really my point. In my opinion it's not an issue to have a specific return type that would throw an error when I return a different type. It "teaches" me something and makes me aware that I either need to update my return type or that I have an actual error in my code.

I would have left everything as it was but I understand the reason why this was modified.

ConsoleTVs:
> I understand your concern, but I personally prefer not having a return type suggested than having something like: `Response|JsonResponse|RedirectResponse|Responsable|static|iterable|\stdClass`, to highlight everything a request can return as a response. That would be more correct, in the sense the framework is ready to handle any of these return types for you.

That's exactly the purpose of an interface... `Illuminate\Contracts\Support\Responsable`.

You have to read it as, return me anything that can be turned into a response. However, as I pointed out in some of my comments, the issue will come with primitive types, as you can't implement an interface on them.

rodrigopedra:
> In my opinion it's not an issue to have a specific return type that would throw an error when I return a different type.  It "teaches" me something and makes me aware that I either need to update my return type or that I have an actual error in my code.

I got that in the first time. We just happen to have different opinions on this matter. 

I don't think it is up to this fundamental part of the framework, to get in front of a developer and try to *"teach"* them anything.

rodrigopedra:
>That's exactly the purpose of an interface... Illuminate\Contracts\Support\Responsable.
>
> **You have to read it as, return me anything that can be turned into a response**. However, as I pointed out in some of my comments, the issue will come with primitive types, as you can't implement an interface on them.

If I **have to** read something differently from what it is written is counter-productive, IMO.

Having an interface on the response is a hint for whoever will consume that method. 

And on this particular part of the request/response cycle, a controller method is only consumed by the framework, which can handle many other different types than this interface.

I would agree if Laravel allowed to swap the request dispatcher by something else, like a PSR-7, or PSR-15. In that case, using a contract would enforce compliance with unknown code that could consume that method. That is not the case here, though. 

A controller's method is not meant to be reused anywhere else, or consumed by any other part of the framework. It is not code meant to be interoperable, nor code written without knowing who will consume it.

If there is a single consumer, requiring a narrow interface, or contract -- when this consumer does already accept a wider domain of values -- adds little to no value, but unneeded noise and overhead.

I don't want to extend this discussion, as this seems to be a matter of opinion, or self-preference. But what would be the true benefit of enforcing any kind of contract on a controller's method return type, when the only consumer accepts anything from it?

khru:
I have to say that I am somewhat surprised, Laravel has always been a highly opinionated framework, with a very specific way of doing things. 

Finally, it has started to standardize things that are logical, like for example, a language based on request, response. End up by returning the obvious, a Response.

I think it is a change that breaks the current way of working, but I see a logical change and, as I mentioned before, consistent with the programming language.

Not only that, it's unnatural to fight against the very language that goes towards typed types.

ccmiller2018:
so how about a slightly different approach?

1) the default is to return as is now
2) if you want it to return an alternative, use a --return option for the stub?

so you might do artisan make:controller --invokable --return=Illuminate\Http\JsonResponse for example?

this feels like it might offer the best of both worlds without compromising the intended purpose of the original tag?

rodrigopedra:
>with a very specific way of doing things.

<details>
<summary>Not sure about that. (see expanded details)</summary>

```php
<?php

use App\Events\AEvent;
use App\Http\Controllers\AController;
use App\Http\Controllers\AnotherController;
use App\Jobs\AJob;
use App\Models\User;
use App\Notifications\ANotification;
use Auth as AuthAlias;
use Facades\Illuminate\Http\Request as RequestRealTimeFacade;
use Illuminate\Contracts\Auth\Authenticatable;
use Illuminate\Contracts\Notifications\Dispatcher as NotificationsDispatcher;
use Illuminate\Contracts\Bus\Dispatcher as BusDispatcher;
use Illuminate\Contracts\Events\Dispatcher as EventsDispatcher;
use Illuminate\Contracts\Foundation\Application;
use Illuminate\Contracts\Session\Session;
use Illuminate\Http\Request;
use Illuminate\Support\Facades\Auth;
use Illuminate\Support\Facades\Queue;
use Illuminate\Support\Facades\Request as RequestFacade;
use Illuminate\Support\Facades\Route;
use Illuminate\Support\Facades\Hash;
use Illuminate\Support\Facades\Session as SessionFacade;
use Illuminate\Support\Facades\Event;
use Illuminate\Support\Facades\Notification;
use Illuminate\Support\Str;
use Request as RequestAlias;
use Session as SessionAlias;
use Str as StrAlias;
use Hash as HashAlias;
use DB as DBAlias;
use Queue as QueueAlias;
use Bus as BusAlias;
use Event as EventAlias;
use Notification as NotificationAlias;

if (is_file(database_path('database.sqlite'))) {
    User::query()->truncate(); // query() + method
    User::truncate(); // auto delegate to query builder
    DB::query()->from('users')->truncate();
    DBAlias::table('users')->truncate();

    $aUser = User::factory()->createOne([
        'name' => Str::random(),
        'email' => StrAlias::random() . '@example.com',
        'password' => bcrypt('password'),
    ]);

    Auth::login($aUser);

    User::forceCreate([
        'id' => 2,
        'name' => Str::random(),
        'email' => StrAlias::random() . '@example.com',
        'password' => Hash::make('password'),
    ]);
}

Route::get('/foo', AController::class) // with preeciding forward slash
->middleware('web', 'auth'); // middleware as parameter list, auth without guard

// middleware can also be defined on a controller's constructor
// (not shown here)
Route::get('bar', [AnotherController::class, 'method']);

Route::get('baz', AnotherController::class . '@method')
    ->middleware(['web', 'auth:web']) // middleware as array, auth with guard
    ->name('a-name');

Route::get('qux', [
    'prefix' => '/boo',
    'uses' => AnotherController::class . '@method',
    'as' => 'boo.',
    'middleware' => ['auth:web'],
]); // classic, all the others get converted to something like this internally

// helper class, as ijecting Authenticatable
// does not work properly when also injecting User
class AUser extends Model {
    protected $table = 'users';
    protected $guarded = [];
}

// $router variable is injected on web.php
$router->get('/user/{user}', function (
    Application $app,
    Request $injectedRequest,
    Session $session,
    Authenticatable $authUser,
    BusDispatcher $bus,
    EventsDispatcher $events,
    NotificationsDispatcher $notifications,
    AUser $user,
) {
    $user->fill([
        'password' => HashAlias::make('password'),
    ])->save(); // fill + save

    $session->put('bar', 3);

    dispatch(new AJob());
    AJob::dispatch();
    Queue::push(new AJob());
    QueueAlias::push(new AJob());
    Bus::dispatch(new AJob());
    BusAlias::dispatch(new AJob());

    event(new AEvent());
    AEvent::dispatch();
    Event::dispatch(new AEvent());
    EventAlias::dispatch(new AEvent());

    /** @var  \App\Models\User $authUser */
    $authUser->notify(new ANotification());
    $notifications->send([$authUser], new ANotification());
    Notification::send([$authUser], new ANotification());
    NotificationAlias::send([$authUser], new ANotification());

    return [
        request('foo'),
        request()->input('foo'),
        app('request')->get('foo'),
        $app['request']->input(),
        $injectedRequest->query->get('foo'),
        RequestFacade::query('foo'),
        RequestAlias::input('foo'),
        RequestRealTimeFacade::get('foo'),

        '###################################',

        request()->user(),
        $authUser,
        auth()->user(),
        Auth::user(),
        AuthAlias::user(),
        app('auth')->user(),

        '###################################',

        $user,
        request('user'),
        request()->user,
        request()->route('user'),
        request()->route()->parameter('user'),
        $injectedRequest->route()->parameters['user'],
        $injectedRequest->route()->user,
        resolve('request')->route('user'), // resolve() is an alias to app()

        '###################################',

        $session->get('bar'),
        $injectedRequest->session()->get('bar'),
        SessionFacade::get('bar'),
        SessionAlias::get('bar'),
        session('bar'),
        session()->get('bar'),
        app('session')->get('bar'),

        '###################################',

        User::all(),
        User::query()->get(),
        DB::table('users')->get(),
    ];
})->middleware('auth:web');

// could keep on with how to define route groups with prefixes, aliases, ...
```
</details>


> Laravel has always been a highly opinionated framework

Agreed. 

But its strength is precisely getting out the way of the developer, so they have to worry mostly with their own business logic, and not with implementation details.

It provides sensible defaults, while still allowing a developer to stick with any style they want.

danabrey:
> Honestly, since Laravel offers the option to return almost anything from a `Controller`, it'd be better to remove the type-hints altogether from the `Controller` stubs. There's just no way to cover every single use case and combination. Trying to do so will result in a Cartesian Explosion.
> 
> People can still add a return type of their own if they wish, or just ignore it.

This is entirely sensible. You're stubbing out userland code, this isn't internal Laravel code, so adding useless return types that only have any meaning once the developer has changed them is arbitrary.

devellopah:
I propose to add property to `app.php`

` 'defaultControllerTemplate' => 'blade' // or 'inertia'`

If **blade** option is set then `php artisan make:controller MyController` should scaffold [this](https://gist.github.com/devellopah/60d6fbf080345da8aebf09664a1df664) template.
If **inertia** option is set then `php artisan make:controller MyController` should scaffold [this](https://gist.github.com/devellopah/2116ee70cbec8562cddbfe83c6a4c9bb) template.

So if you're building project with Inertia, you just change property value to **inertia**.

I also propose to add flags `--blade` and `--inertia` to `php artisan make:controller` command in order to manually scaffold templates.

`php artisan make:controller --blade` to scaffold [this](https://gist.github.com/devellopah/60d6fbf080345da8aebf09664a1df664) template
`php artisan make:controller --inertia` to scaffold [this](https://gist.github.com/devellopah/2116ee70cbec8562cddbfe83c6a4c9bb) template

rodrigopedra:
@devellopah 

One can already customize the stubs (templates) used by all `make:` commands, on a per-project basis, by exporting all stubs using this command:

```bash
php artisan stub:publish
```

reference: https://laravel.com/docs/10.x/artisan#stub-customization

Pluuk:
I feel like the main argument against an interface like ```Illuminate\Contracts\Support\Responsable``` is that return values like arrays and strings won't be supported.

What about extending ```Response``` with something like fromString() and fromArray() methods. Or maybe even something like fromNativeType(). Then, objects would be able to implement the Responsable interface, and native types should passed to the Response interface with the fromNativeType() function.

dennisprudlo:
@Pluuk this is already possible with `response('whatever')` or `response([ 1, 2, 3 ])` or `Response::make(...)` if you would like to use a facade. This however doesn't solve `return 'whatever';`

Although I am a heavy user of types I think no types by default as it is right now is the best option.

```php
// this is not an issue:
public function method()
{
    return view('foo');
}

// this however is:
public function method(): Responsable
{
    return [ 'bar' ];
}
```

the second option simply includes more use cases by default. And if anyone want's to lock down the first method to views, that it can be done either in the controller itself or in published stubs.

sadathimel:
> Noob question
> 
> Doesn't this kill the whole purpose of return types?
> 
> What's the difference between this and no return types at all
> 
> And cant we instead do something like this
> 
> Response|View etc

Thank you, you mentioned the obvious. Why hit zend_gc_collector with garbage it cannot even collect? The streets gonna stink like all the previous versions of php

PR #46167, opened 2023-02-18
Support responsable on more objects

taylorotwell:
Support `Responsable` on all response types + views so that the contract could potentially be used as an alternative to `Response` or `mixed` on controller stubs.

Reference: https://github.com/laravel/framework/pull/46166


ConsoleTVs:
Here's an example of multiple controller handlers this PR would allow:

```php
use Illuminate\Contracts\Support\Responsable;

public function foo(): Responsable {
    return response('foo');
}

public function foo(): Responsable {
}

public function foo(): Responsable {
    return redirect()->route('foo');
}

public function foo(): Responsable {
    return view('foo');
}

public function foo(User $user): Responsable {
    return new UserResource($user);
}

public function foo(): Responsable {
    return UserResource::collection(User::all());
}
```

PR #46169, opened 2023-02-19
[10.x] PHPDoc generics @template on Builder for better type analys

puggan:
Implemented the PHPDoc changes suggested in #40986.

Helps Static analysers and IDEs get the correct type from queries.

Exemple running dumpTypes on this model:

```
class User extends \Illuminate\Database\Eloquent\Model
{
    public static function example(): void
    {
        \PHPStan\dumpType(self::query());
        \PHPStan\dumpType(self::query()->first());
        \PHPStan\dumpType(self::query()->find(1));
        \PHPStan\dumpType(self::query()->findMany([1,2]));
        \PHPStan\dumpType(self::query()->get());
    }
}
```

PHPStan types with PR:

```
 ------ ----------------------------------------------------------------------------------------------------
  Line   User.php
 ------ ----------------------------------------------------------------------------------------------------
  7      Dumped type: Illuminate\Database\Eloquent\Builder<static(User)>
  8      Dumped type: static(User)|null
  9      Dumped type: Illuminate\Database\Eloquent\Collection<(int|string), static(User)>|static(User)|null
  10     Dumped type: Illuminate\Database\Eloquent\Collection<(int|string), static(User)>
  11     Dumped type: Illuminate\Database\Eloquent\Collection<(int|string), static(User)>
 ------ ----------------------------------------------------------------------------------------------------
```

PHPStan types without PR:

```
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  Line   User.php
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
  7      Dumped type: Illuminate\Database\Eloquent\Builder
  8      Dumped type: object|null
  9      Dumped type: Illuminate\Database\Eloquent\Builder|(Illuminate\Database\Eloquent\Collection&iterable<Illuminate\Database\Eloquent\Builder>)|Illuminate\Database\Eloquent\Model|null
  10     Dumped type: Illuminate\Database\Eloquent\Collection
  11     Dumped type: Illuminate\Database\Eloquent\Collection&iterable<Illuminate\Database\Eloquent\Builder>
 ------ ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Questions:
* was the removed `static`, `static[]` and `$this` there to try to solve the same problem? just looking at the function they made no sens. (should I kept them?)
* Do queries always return numeric index arrays, or do some implementations return other types of index? in #40986 the Collection uses int as index but I used type `array-key` instead as I wasn't sure.,

Tests: as it only changes PHPDocs, it have no effect on the code execution. (also reasons it does not break any existing features.)

benefit to end users: More narrow types when using static analyses and IDE autocompletes.

how it makes building web applications easier: better support in IDE autocomplete and static analyses.

### DIFF
--- a/src/Illuminate/Database/Concerns/BuildsQueries.php
+++ b/src/Illuminate/Database/Concerns/BuildsQueries.php
@@ -18,6 +18,9 @@ use Illuminate\Support\Traits\Conditionable;
 use InvalidArgumentException;
 use RuntimeException;
 
+/**
+ * @template TModel
+ */
 trait BuildsQueries
 {
     use Conditionable;
@@ -289,7 +292,7 @@ trait BuildsQueries
      * Execute the query and get the first result.
      *
      * @param  array|string  $columns
-     * @return \Illuminate\Database\Eloquent\Model|object|static|null
+     * @return TModel|null
      */
     public function first($columns = ['*'])
     {
@@ -300,7 +303,7 @@ trait BuildsQueries
      * Execute the query and get the first result if it's the sole matching record.
      *
      * @param  array|string  $columns
-     * @return \Illuminate\Database\Eloquent\Model|object|static|null
+     * @return TModel|null
      *
      * @throws \Illuminate\Database\RecordsNotFoundException
      * @throws \Illuminate\Database\MultipleRecordsFoundException
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -21,6 +21,8 @@ use ReflectionClass;
 use ReflectionMethod;
 
 /**
+ * @template TModel of Model
+ *
  * @property-read HigherOrderBuilderProxy $orWhere
  * @property-read HigherOrderBuilderProxy $whereNot
  * @property-read HigherOrderBuilderProxy $orWhereNot
@@ -43,7 +45,7 @@ class Builder implements BuilderContract
     /**
      * The model being queried.
      *
-     * @var \Illuminate\Database\Eloquent\Model
+     * @var TModel
      */
     protected $model;
 
@@ -146,7 +148,7 @@ class Builder implements BuilderContract
      * Create and return an un-saved model instance.
      *
      * @param  array  $attributes
-     * @return \Illuminate\Database\Eloquent\Model|static
+     * @return TModel
      */
     public function make(array $attributes = [])
     {
@@ -306,7 +308,7 @@ class Builder implements BuilderContract
      * @param  mixed  $operator
      * @param  mixed  $value
      * @param  string  $boolean
-     * @return \Illuminate\Database\Eloquent\Model|static|null
+     * @return TModel|null
      */
     public function firstWhere($column, $operator = null, $value = null, $boolean = 'and')
     {
@@ -431,7 +433,7 @@ class Builder implements BuilderContract
      *
      * @param  mixed  $id
      * @param  array|string  $columns
-     * @return \Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Collection|static[]|static|null
+     * @return TModel|\Illuminate\Database\Eloquent\Collection|null
      */
     public function find($id, $columns = ['*'])
     {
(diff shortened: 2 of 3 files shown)


puggan:
Test failed due too slow execution, a wait of 100ms [97-103ms] took 248ms :-(

nunomaduro:
Thank you for suggesting this generic improvements. However, at this time, editors are not ready for this level of auto-completion, and we don't want to take the risk of loosing auto-completion on PHPStorm / VS Code.

puggan:
For reference, current PHPStorm is 2022.3.2, and for the following code:

```php
class User extends \Illuminate\Database\Eloquent\Model
{
    public static function example(): void
    {
        /** @var \Illuminate\Database\Eloquent\Builder<self> $q */
        $q = self::query();
        $a = $q->first();
        $b = $q->find(1);
        $c = $q->findMany([1,2]);
        $d = $c[0];
        $e = $q->get();
        $f = $e[0];
    }
}
```

with this PR the type are identified as:

* `$q` is Builder, but TModel isn't set to User, so to test the rest the phpdoc setting the correct TModel is added.
* `$a` is correctly identified as `null|User`
* `$b` is correctly identified as `Collection|null|User`, could later be split with conditional return type
* `$c` is identified as `Collection` but lost it type so
* `$d` is identified as `mixed`
* `$e` is identified as `Collection` but lost it type so
* `$f` is identified as `mixed`

without this PR the types are identified as:

* `$a` identifies as `Builder|Model|null|object`
* `$b` identifies as `Builder|Builder[]|Collection|Model|null`
* `$e` identifies as `Builder[]|Collection`
* `$f` identifies as `Builder|mixed`

---

Both cases are far from as good as they could be, so personally i feel the `risk of loosing auto-completion on PHPStorm` is minimal as the auto-completion doesn't work to begin with. But maybe there is some workaround I'm not aware of that's works with the current odd declarations. I going to assume [nunomaduro](https://github.com/nunomaduro) knows what's he talking about. Thanks for the answer.

puggan:
For reference, I also installed latest VS Code 1.75.1 with PHP Intelephense 1.9.5:

This is what types I get in VS Code, the phpdoc from the phpstorm exemple wasn't needed.

* `$q` -> `\Illuminate\Database\Eloquent\Builder<\User>`
* `$a` -> `mixed|null`
* `$b` -> `\User|\Illuminate\Database\Eloquent\Collection<array-key, \User>|null $b`
* `$c` -> `\Illuminate\Database\Eloquent\Collection<array-key, \User>`
* `$d` -> `\User`
* `$e` -> `\Illuminate\Database\Eloquent\Collection<array-key, \User>`
* `$f` -> `\User`

and VS Code without the PR:

* `$q` -> `\Illuminate\Database\Eloquent\Builder`
* `$a` -> `\Illuminate\Database\Eloquent\Model|object|\Illuminate\Database\Eloquent\Builder|null`
* `$b` -> `\Illuminate\Database\Eloquent\Model|\Illuminate\Database\Eloquent\Collection<array-key, \Illuminate\Database\Eloquent\Builder>|\Illuminate\Database\Eloquent\Builder|null`
* `$c` -> `\Illuminate\Database\Eloquent\Collection`
* `$d` -> `\Illuminate\Database\Eloquent\Model`
* `$e` -> `\Illuminate\Database\Eloquent\Collection<array-key, \Illuminate\Database\Eloquent\Builder>`
* `$f` -> `\Illuminate\Database\Eloquent\Builder`

So VS Code got 6 of 7 types correct with PR, and only failed on the `@use BuildsQueries<TModel>`

JJWesterkamp:
@nunomaduro I'm curious what you mean by the "risk of loosing auto-completion on PHPStorm / VSCode". What is it specifically that may break in any way, other than assumptions in code about types that are no longer satisfied by narrower return types? Is there anything specific the community is required to do in order to mitigate those concerns, e.g. by improving support for generics in IDEs?

Can you consider re-opening this PR?

haringsrob:
Hey @nunomaduro quick question regarding this. I feel that mosts lsp's and editors have support for generics these days.

- phpstorm
- devsense
- intelephense
- phpactor

Just to name a few.

Will a new pr be approved for this? (or can this be reopened to continue @puggan's work?

PR #46172, opened 2023-02-19
[11.x] Route redirect with query string

guizoxxv:
This PR changes the `Illuminate/Routing/Router@redirect` method to maintain the query parameters from the source url in the target url. I think it is intuitive to expect the query parameters to be kept in a redirect. Without this, the user would need to use a closure or controller and add the query parameters to the redirect.


taylorotwell:

PR #46174, opened 2023-02-19
Allow Belongs to many relationship to sync id's with distinct pivot values

chibyk101:
The behavior of the method syncWithPivotValues() is to sync  all entered id's with a particular set of value(s) to pivot column(s)

Example
```php
$user->roles()->syncWithPivotValues([1,2,3],[ 'status' => 'active' ]);
//will produce this array after mapping
array(
  '1' => [ 'status' => 'active'],
  '2' => [ 'status' => 'active'],
  '3' => [ 'status' => 'active'],
)
```

But what I wanted was the ability to have different pivot column values for each role_id. More like so:

```php
//notice the different values for each role_id
array(
  '1' => [ 'status' => 'active'],
  '2' => [ 'status' => 'inactive'],
  '3' => [ 'status' => 'pending'],
)
```
Now for the user's roles to all have a different status in one short at syncing; This can be possible with a new method syncWithAssocPivotValues()

```php
$user->roles()->syncWithAssocPivotValues([1,2,3],[
           ['status' => 'active'],
           ['status' => 'inactive'],
           ['status' => 'pending'],
        ];

```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
@@ -141,23 +141,33 @@ trait InteractsWithPivotTable
         }), $detaching);
     }
     
-     /**
-     * Sync the intermediate tables with a list of IDs or collection of models with the given pivot values associatively.
-     *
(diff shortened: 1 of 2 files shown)


nk-d:
Why do you use 2-space tabs?

driesvints:
Please resubmit this with passing coding standards, thanks.

PR #46178, opened 2023-02-20
Add localization to throttle exception

mshamaseen:
Add localization to throttle exception.

Laravel returns English-only response on throttle middleware, which is not the correct thing for localized projects.

it doesn't break anything as it is just a localization. (unless someone matches the string of the response, which doesn't make sense :D)


driesvints:
Thank you but this isn't an error message that should ever reach an end user. It's purely informative to the system.

PR #46186, opened 2023-02-20
[10.x] Removes duplicate builds using PHPUnit 9

nunomaduro:
This pull request removes duplicate builds using PHPUnit 9, and effectively ensures our framework is only being tested by PHPUnit 10.


driesvints (approved):

crynobone:
Can we have phpunit 9 run only on tags or `workflow_dispatch`. This can allow us to troubleshoot if anything stop working on phpunit 9.

driesvints:
@crynobone we'd have to add the majority of code of this PR again this way. Maybe we can, for now, set up a separate GitHub Action yml file that runs once a week on a schedule?

crynobone:
That's another option too, it not something that we forsee would break but would be nice not to break any existing code on app still running phpunit 9.

driesvints:
Yeah, main reason I want to get rid of the PHPUnit 9 builds here is to not have so many builds running for every PR/commit push.

Issue #46180, opened 2023-02-20
Argument #1 ($manager) must be of type Illuminate\Mail\MailManager


alexgaal:
- Laravel Version: 10.0.3
- PHP Version: 8.2.0
- Database Driver & Version:

### Description:
We are using `Mail::fake` in a test and since we upgraded to Laravel 10 it's failing with following exception:

```
Illuminate\Support\Testing\Fakes\MailFake::__construct(): Argument #1 ($manager) must be of type Illuminate\Mail\MailManager, Illuminate\Support\Testing\Fakes\MailFake given, called in /var/www/vendor/laravel/framework/src/Illuminate/Support/Facades/Mail.php on line 68
```

```
Mail::fake();
Mail::assertNothingSent();
expect(fn () => $votingCard->approveRequest())->not()->toThrow(\Symfony\Component\Mime\Exception\LogicException::class);
```

driesvints:
This was changed here: https://github.com/laravel/framework/pull/46055 because of https://github.com/laravel/framework/pull/45988

@joelbutcher @stevebauman do you have any ideas here on how to fix this?

joelbutcher:
@driesvints I can't replicate this... I've [written a test](https://github.com/laravel/framework/compare/10.x...joelbutcher:framework:test-mailable-sent) that that should cover this in the framework

driesvints:
Thanks @joelbutcher. @alexgaal can you check this? ^

joelbutcher:
I've [raised a PR](https://github.com/laravel/framework/pull/46181) for this, too

joelbutcher:
@alexgaal @maloun96 if you run the following in tinker, what do you get?

```shell
> app('mail.manager')
```

The only thing I can think of is an incorrectly configured service provider binding that's causing `app('mail.manager')` to resolve to the `MailFake` class instead of `MailManager`.

maloun96:
I believe it's because we declared multiple times:
```php
    protected function setUp(): void
    {
        parent::setUp();
        Mail::fake();
    }

    public function test_123(): void
    {
        Mail::fake();
```
Before it worked. Now we should declare only once

driesvints:
@maloun96 that doesn't seems like valid code to me (even if it was working before).

maloun96:
@driesvints indeed, legacy code...

driesvints:
@alexgaal are you faking emails in a similar way?

maloun96:
Tested with a fresh laravel app and still error.
[...]

PR #46055, opened 2023-02-09
[10.x] Use `MailManager` as underlying passthru object for `MailFake`

joelbutcher:
This PR fixes a bug introduced in #45988 where calling `Mail::fake()` throws the following `TypeError`:

```
TypeError: Illuminate\Support\Testing\Fakes\MailFake::__construct(): Argument #1 ($mailer) must be of type Illuminate\Contracts\Mail\Mailer, Illuminate\Mail\MailManager given, called in /home/runner/work/jetstream/jetstream/vendor/laravel/framework/src/Illuminate/Support/Facades/Mail.php on line 68

/home/runner/work/jetstream/jetstream/vendor/laravel/framework/src/Illuminate/Support/Testing/Fakes/MailFake.php:53
/home/runner/work/jetstream/jetstream/vendor/laravel/framework/src/Illuminate/Support/Facades/Mail.php:68
```

I've added two new test cases to ensure the intended functionality for the original PR remains – please let me know if you want these moved to a better suited directory.

PR #46188, opened 2023-02-20
[10.x] Facade Fake Awareness

joelbutcher:
## Description

This PR attempts to address a [slight issue](https://github.com/laravel/framework/issues/46180) with Facade fakes. Currently, if a facade fake requires an instance of the root as a constructor argument. (e.g. `MailFake::__construct()` or `BusFake::__construct()`), then calling `fake()` more than once can cause tests to fail in userland.

Let's consider the following scenario that currently passes in 9.x:

```php
<?php

namespace Tests

use Illuminate\Support\Facades\Mail;
use Tests\TestCase;

class ExampleTest extends TestCase
{
    public function setUp(): void
    {
        // ...
        Mail::fake();
        // ...
    }

    // ...


    public function testItWorks(): void
    {
        Mail::fake();

        // ...
    }
}
```

This test case runs perfectly fine on Laravel < 10.0.0, but due to the merge of #45988 and #46055 will fail in Laravel >= 10.0.0. This is because we end up with a small loop:

`Mail::fake()` loads `Mail::getFacadeRoot()` (also resolvable as `app('mail.manager')`) as a constructor argument of `MailFake`. On the first attempt, this call resolves to the correct `MailManager`. However on the second pass, the binding to `mail.manager` has been swapped out with the new `MailFake` instance in the container. Causing errors like this one:

```bash
Illuminate\Support\Testing\Fakes\MailFake::__construct(): Argument #1 ($manager) must be of type 
Illuminate\Mail\MailManager, Illuminate\Support\Testing\Fakes\MailFake given, called in 
/var/www/vendor/laravel/framework/src/Illuminate/Support/Facades/Mail.php on line 68
```

---

## Solution

To solve this issue, I've created a new `\Illuminate\Support\Testing\Fakes\Fake` interface that is used to determine if the current facade root is being "faked". If that's the case, then in the facade itself I skip the call to `static::swap(...)`.

I've then taken inspiration from `isMock()` used for determining if the facade is a mockery instance during for testing and added a new `protected function isFake()` which is checked inside the `fake()`:

```php
$fake = // ...

return tap($fake, function ($fake) {
    if (!static::isFake()) {
        static::swap($fake);

        // ...
    }
});
```

### Facades updated

| Facade | Done |
| ---  | --- |
| Bus | ✅  |
| Event | ✅ |
| Mail  | ✅ |
| Notification  | ✅ |
| Queue  | ✅ |

### DIFF
--- a/src/Illuminate/Support/Facades/Facade.php
+++ b/src/Illuminate/Support/Facades/Facade.php
@@ -173,10 +173,9 @@ abstract class Facade
      * Hotswap the underlying instance behind the facade.
      *
      * @param  mixed  $instance
-     * @paramm  bool  $fake
      * @return void
      */
-    public static function swap($instance, $fake = false)
+    public static function swap($instance)
     {
 
         static::$resolvedInstance[static::getFacadeAccessor()] = $instance;
--- a/src/Illuminate/Support/Facades/Facade.php
+++ b/src/Illuminate/Support/Facades/Facade.php
@@ -177,7 +177,6 @@ abstract class Facade
      */
     public static function swap($instance)
     {
-
         static::$resolvedInstance[static::getFacadeAccessor()] = $instance;
 
         if (isset(static::$app)) {
--- a/src/Illuminate/Support/Facades/Facade.php
+++ b/src/Illuminate/Support/Facades/Facade.php
@@ -212,7 +212,7 @@ abstract class Facade
      *
      * @return string
      *
-     * @throws RuntimeException
+     * @throws \RuntimeException
      */
     protected static function getFacadeAccessor()
     {
--- a/src/Illuminate/Support/Facades/Facade.php
+++ b/src/Illuminate/Support/Facades/Facade.php
@@ -3,17 +3,13 @@
 namespace Illuminate\Support\Facades;
 
 use Closure;
-use Illuminate\Contracts\Foundation\Application;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Arr;
-use Illuminate\Support\Collection;
 use Illuminate\Support\Js;
 use Illuminate\Support\Str;
 use Illuminate\Support\Testing\Fakes\Fake;
 use Mockery;
-use Mockery\Expectation;
 use Mockery\LegacyMockInterface;
-use Mockery\MockInterface;
 use RuntimeException;
 
 abstract class Facade
@@ -21,7 +17,7 @@ abstract class Facade
     /**
      * The application instance being facaded.
      *
-     * @var Application
+     * @var \Illuminate\Contracts\Foundation\Application
      */
     protected static $app;
 
@@ -42,7 +38,7 @@ abstract class Facade
     /**
      * Run a Closure when the facade has been resolved.
      *
-     * @param  Closure  $callback
+     * @param  \Closure  $callback
      * @return void
      */
     public static function resolved(Closure $callback)
@@ -61,7 +57,7 @@ abstract class Facade
     /**
      * Convert the facade into a Mockery spy.
      *
-     * @return MockInterface
+     * @return \Mockery\MockInterface
      */
     public static function spy()
     {
@@ -77,7 +73,7 @@ abstract class Facade
     /**
      * Initiate a partial mock on the facade.
      *
-     * @return MockInterface
+     * @return \Mockery\MockInterface
      */
     public static function partialMock()
     {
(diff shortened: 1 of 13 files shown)


src/Illuminate/Support/Facades/Facade.php
```diff
@@ -176,13 +177,27 @@ protected static function getMockableClass()
      */
     public static function swap($instance)
     {
+
```
joelbutcher, line 180:
```suggestion
```

src/Illuminate/Support/Facades/Queue.php
```diff
@@ -76,9 +76,12 @@ public static function popUsing($workerName, $callback)
      */
     public static function fake($jobsToFake = [])
     {
-        static::swap($fake = new QueueFake(static::getFacadeApplication(), $jobsToFake, static::getFacadeRoot()));
+        tap(new QueueFake(static::getFacadeApplication(), $jobsToFake, static::getFacadeRoot()), function ($fake) {
+            if (! static::isFake()) {
+                static::swap($fake);
+            }
+        });
```
joelbutcher, line 84:
```suggestion
```

src/Illuminate/Support/Facades/Facade.php
```diff
@@ -198,7 +212,7 @@ public static function getFacadeRoot()
      *
      * @return string
      *
-     * @throws \RuntimeException
+     * @throws RuntimeException
```
driesvints, line 215:
```suggestion
     * @throws \RuntimeException
```

driesvints (approved):
Thanks @joelbutcher

tests/Integration/Queue/UniqueJobTest.php
```diff
@@ -48,9 +48,9 @@ public function testUniqueJobsAreNotDispatched()
             $this->app->get(Cache::class)->lock($this->getLockKey(UniqueTestJob::class), 10)->get()
         );
 
-        Bus::fake();
+        Bus::assertDispatchedTimes(UniqueTestJob::class);
         UniqueTestJob::dispatch();
-        Bus::assertNotDispatched(UniqueTestJob::class);
+        Bus::assertDispatchedTimes(UniqueTestJob::class);
```
taylorotwell, line 53: Why was this test changed?

joelbutcher: @taylorotwell because of the changes made here:

```php
// \Illuminate\Support\Facades\Bus.php

return tap(new BusFake(static::getFacadeRoot(), $jobsToFake, $batchRepository), function ($fake) {
    if (! static::isFake()) {
        static::swap($fake);
    }
});
```

Previously, this test:
1. Swapped the underlying facade root instance for `BusFake` with a `$dispatcher` property of the previous facade root (`\Illuminate\Bus\Dispatcher`) – Line 42
3. Swapped out the faked facade root for a new `BusFake` instance, now with a `$dispatcher` property of the previous `BusFake` instance from the facade root – Line 51.

As a result of the fix, calling `Bus::fake()` twice in the same test now causes `Bus::assertNotDispatched(...)` to fail, which is correct, as we're now using the same `\Illuminate\Contracts\Bus\QueueingDispatcher` instance for the `BusFake::$dispatcher` property.

This test is now correct because, despite executing `UniqueTestJob::dispatch();` twice (lines 44 and 52), we are now correctly verifying that the unique job was dispatched exactly once on the underlying `BusFake` facade instance

Unless we manually build a `QueueingDispatcher` instance within the `fake` method.

taylorotwell: I'm sorry - I can't merge this with this test change. I don't understand this test anymore and we've had too many breaking changes lately.

The test used to plain read that a unique job was dispatched and a lock acquired, then a second job couldn't be dispatched since a lock was in place for the unique job. The test no longer reads that way and is confusing. Please feel free to mark this as ready for review again when the test reads more plainly.

joelbutcher: It makes more sense if you disregard the "faked" element to this test. This test has been updated to reflect the REAL functionality of the `Bus` facade and underlying manager regardless of the call to to `Bus::fake()`.

In userland, you're not going to call `Bus::fake()`, so if you called `UniqueTestJob::dispatch()` twice in userland, like the test, then the the expectation is that **this job was not dispatched more than once**.

This PR identifies a problem with using `Bus::getFacadeRoot()` and passing the result to `BusFake` – calling `Bus::fake()` twice means you end up with the second`BusFake` instance deferring to the first `BusFake` instance rather than the underlying `Dispatcher` instance (which is what it should do). In any case, calling `Bus::fake()` (or any Facade::fake()`) more than once in a test should be considered bad practice as it leads to incorrect behaviour where `getFacadeRoot` is concerned.

If it helps readability, I can add an `assertDispatchedOnce` method to the `BusFake` class and update the test accordingly.

PR #46192, opened 2023-02-20
[10.x] Add `fragment*` signatures to view interface

jorqensen:
### Description

While working on a Blade & HTMX project, I've noticed my editor not being aware of the `fragment*` methods on the `view()` helper. The helper returns interfaces that does not implement the method signature for said methods.

During this PR I've noticed several public methods are not on the interface, I am unsure if this is by choice however, so I've left them out (for now).

If this is indeed a breaking change, perhaps another solution could be raised.

### Solution

Added signatures for all `fragment*` methods on `Illuminate\Contracts\View\View`.

### Benefit

Editor/IDE & end users aware of method being available.

### DIFF
--- a/src/Illuminate/Contracts/View/View.php
+++ b/src/Illuminate/Contracts/View/View.php
@@ -28,4 +28,38 @@ interface View extends Renderable
      * @return array
      */
     public function getData();
+
+    /**
+     * Get the evaluated contents of a given fragment.
+     *
+     * @param  string  $fragment
+     * @return string
+     */
+    public function fragment($fragment);
+
+    /**
+     * Get the evaluated contents for a given array of fragments.
+     *
+     * @param  array  $fragments
+     * @return string
+     */
+    public function fragments(array $fragments);
+
+    /**
+     * Get the evaluated contents of a given fragment if the given condition is true.
+     *
+     * @param  bool  $boolean
+     * @param  string  $fragment
+     * @return string
+     */
+    public function fragmentIf($boolean, $fragment);
+
+    /**
+     * Get the evaluated contents for a given array of fragments if the given condition is true.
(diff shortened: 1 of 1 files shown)


rodrigopedra:
An alternative solution would be type-hinting the concrete classes, instead of their contracts/interface, on the view helper.

https://github.com/laravel/framework/blob/9a409cfa38919efa22f8aba32f9bc1b78b329755/src/Illuminate/Foundation/helpers.php#L1003-L1010

I am not saying I think this would be a better solution, I am just suggesting as an alternative solution if adding public methods to the interface should be deferred to the 11.0 release.

jorqensen:
Updating the return type of the function was my initial thought, however looking at the other helpers returning interfaces it felt out of place to suddenly change that. So I opted towards maybe the interface should reflect the public API.

In theory updating the function does sound better, as I wonder if updating the interface would break potential implementations of it for other view engines, that users/package maintainers migh have.

rodrigopedra:
Let's see how the maintainers feel about it. 

An approach could be adding, not replacing, the concrete classes to the helper's return union type to Laravel 10.x, and add those to the interface in Laravel 11. And also on Laravel 11, remove the concrete classes from the helper return type.

So package authors could anticipate any additions, while this feature can play nice with IDE/static analysis tools.

It is cumbersome, but it is an option.

driesvints:
We can't add methods to contracts on a patch release sorry.

jorqensen:
@driesvints Sure, but could we update the return of the view helper then? Like the original post said, perhaps a better solution could be raised from the PR. It's insanely annoying.

driesvints:
@jorqensen it's best that you just attempt a PR so we can look at actual code.

PR #46197, opened 2023-02-21
[10.x] Added possibility to globally set timeout setting via env variable

oliverhkraft:
Having third-party libraries that depends on the HTTP Client, we found the need to be able to globally set a higher timeout threshold. 

This leaves the standard timeout as 30 seconds.


ankurk91:
You should not use `env()` helper beyond config folder.

As long as the HTTP client is being resolved via Container, you can easily set the global default options to it.

decadence:
> You should not use `env()` helper beyond config folder.

Yes. This will return `null` when config is cached.

taylorotwell:
See comments above.

PR #46198, opened 2023-02-21
[10.x] Added Pipeline Helper Function

JordanDalton:
Pipelines are quite powerful in Laravel. GIven how quick one may wish to utilize this feature I would like to propose the pipe() helper function so developers can quickly and cleanly leverage this feature.

### DIFF
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -427,3 +427,20 @@ if (! function_exists('with')) {
         return is_null($callback) ? $value : $callback($value);
     }
 }
+
+if( ! function_exists('pipe')) {
+    /**
+     * Pass a payload through a series of pipelines.
+     *
+     * @param $payload
+     * @param array $pipelines
+     * @return mixed
+     */
+    function pipe($payload, $pipelines = [])
+    {
+        return app(\Illuminate\Pipeline\Pipeline::class)
+            ->send($payload)
+            ->through($pipelines)
+            ->thenReturn();
+    }
+}
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -432,15 +432,14 @@ if (! function_exists('pipe')) {
     /**
      * Pass a payload through a series of pipelines.
      *
-     * @param $payload
-     * @param  array  $pipelines
-     * @return mixed
+     * @param  $payload
+     * @param  array $pipelines
+     * @return \Illuminate\Pipeline\Pipeline
      */
-    function pipe($payload, $pipelines = [])
+    function pipe($payload, $pipelines = []) : Illuminate\Pipeline\Pipeline
     {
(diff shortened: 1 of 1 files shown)


rodrigopedra:
May I suggest an alternative implementation?

```php
if (! function_exists('pipe')) {
    /**
     * Send the given payload through a pipeline, or return a pipeline instance.
     *
     * @param  $payload
     * @param  array  $pipes
     * @return mixed
     */
    function pipe()
    {
        $arguments = func_get_args();

        if (empty($arguments)) {
            return app(\Illuminate\Pipeline\Pipeline::class);
        }

        if (count($arguments) === 1) {
            return app(\Illuminate\Pipeline\Pipeline::class)
                ->send($arguments[0]);
        }

        $payload = array_shift($arguments);

        if (! is_array($arguments[1])) {
            throw new InvalidArgumentException(
                'When sending a value through a pipeline, you must pass an array of pipes.'
            );
        }

        return app(\Illuminate\Pipeline\Pipeline::class)
            ->send($arguments[0])
            ->through($arguments[1])
            ->thenReturn();
    }
}
```

Some other helpers use this approach of returning just a new instance when no arguments are passed. Check the `cache()` helper as an example.

This would allow usage like these:

```php
$pipes = [
    // some pipes to send the value through
];

function doSomethingWithValue ($value) {
    // do something with the value
    return $value;
}

$result = pipe()
    ->send($value)
    ->through($pipes)
    ->then(doSomethingWithValue(...)); // First class callable syntax

// or

$result = pipe($value)
    ->through($pipes)
    ->then(doSomethingWithValue(...));

// or, with your original syntax

$result = doSomethingWithValue(pipe($value, $pipes));
```

JordanDalton:
The effect I was looking to achieve was not having to call send(), through(), then(), and return() all the time.

imanghafoori1:
This may be helpful:
https://github.com/imanghafoori1/laravel-middlewarize

henzeb:
> The effect I was looking to achieve was not having to call send(), through(), and then() all the time.

The alternative solution achieves the same, but also gives other users some more control when needed.

JordanDalton:
calling app(), send(), and through() will always be true, it's just the return that's the question, which makes the following solution more appealing:

https://github.com/laravel/framework/discussions/46170#discussioncomment-5063459

rodrigopedra:
I'd prefer having it always return a `Pipeline` instance. But that makes me wonder, how is that much different from doing:

```php
app(Pipeline::class)->pipe([])->thenReturn();
```

Or from adding a static constructor to the Pipeline class itself. 

Aside of being shorter, of course. But from what I've been reading on the PR's since a few years ago, new global helpers are to be vetted very carefully, as often many complain when new ones are added due to global namespace conflict with locally defined helpers.

JordanDalton:
I've updated the PR so that it will return the Pipline so the user has choice if they want to use then() or thenReturn().

taylorotwell:

PR #46199, opened 2023-02-21
Add missing exception in phpdoc for ```transformDependency``` method

milwad-dev:


taylorotwell:
We don't document this kind of thing unless the method actually explicitly throws an exception.

Issue #46179, opened 2023-02-20
Mailable custom theme incorrectly applied


AEM5299:
- Laravel Version: 9.51.0
- PHP Version: 8.1.3
- Database Driver & Version: MySQL 5.7

### Description:
When a Mailable with a custom theme is sent before another Mailable with implicit default (`$theme = null`) in the same process (e.g., a queue), the second Mailable also gets the custom theme. As far as I can tell this happens because the [markdown renderer class is a singelton](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Mail/Mailable.php#L325) and the theme is only updated [if it's **not null**](https://github.com/laravel/framework/blob/9.x/src/Illuminate/Mail/Mailable.php#L327).

If someone can confirm this is indeed not the intended behaviour, I can probably send in a PR. I have been trying to debug this for hours at this point so might be missing something trivial here.

### Steps To Reproduce:

1. Create a Mailable `MailableA` and set `public $theme = 'custom';`
2. Create another Mailable `MailableB` and don't set a theme (or set it to null)
3. Queue `MailableA` first then `MailableB` in the same queue and run the queue
4. `MailableB` will have the `'custom'` theme apllied, when it should be the default

driesvints:
Heya, thanks for reporting.

We'll need more info and/or code to debug this further. Can you please create a repository with the command below, commit the code that reproduces the issue **as one separate commit** on the main/master branch and share the repository here? Please make sure that you have the [latest version of the Laravel installer](https://github.com/laravel/installer) in order to run this command. Please also make sure you have both Git & [the GitHub CLI tool](https://cli.github.com) properly set up.

```bash
laravel new bug-report --github="--public"
```

Please do not amend and create a separate commit with your custom changes. After you've posted the repository, we'll try to reproduce the issue.

Thanks!

AEM5299:
Bug report repo is here https://github.com/AEM5299/bug-report. If you run `php artisan serve` and visit the `/test` route, it will log both mails. In **both logged mails** you'll see that the body tag has a `style="background-color: salmon;"` while the `custom-theme` is **only** applied to `MailableA`

driesvints:
I can confirm this. 

[...]

PR #46200, opened 2023-02-21
[10.x] Fix custom themes not reseting on Markdown renderer

AEM5299:
This fixes the issue described in #46179. It should reset the theme to `'default'` on the Markdown renderer object if no `$theme` is set on the `Mailable` object. I've also added a test for this. This PR does not introduce any breaking changes.

There's a small inconsistency between `Mailable` and `Markdown` classes, the `Mailable` class has a default `$theme` value of `null` (implicitly the default theme), however `Markdown` has a default theme value of `'default'` ([here](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Mail/Mailable.php#L166) and [here](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Mail/Markdown.php#L28)). But fixing this inconsistency _might_ be a breaking change, so not sure if you'd want to fix it.

### DIFF
--- a/src/Illuminate/Mail/Mailable.php
+++ b/src/Illuminate/Mail/Mailable.php
@@ -324,9 +324,7 @@ class Mailable implements MailableContract, Renderable
     {
         $markdown = Container::getInstance()->make(Markdown::class);
 
-        if (isset($this->theme)) {
-            $markdown->theme($this->theme);
-        }
+        $markdown->theme($this->theme ?? 'default');
 
         $data = $this->buildViewData();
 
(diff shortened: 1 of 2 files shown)


pikseli:
Seems like it's setting the theme always to `default` unless it's explicitly defined in the `Mailable` class thus making the `markdown.theme` option in `config/mail.php` configuration file being ignored.

driesvints:
@AEM5299 we'll need to fix that. Can you provide a follow up PR?

AEM5299:
Opened PR #46317.

driesvints:
We've decided to revert this PR for now.

PR #46201, opened 2023-02-21
[10.x] Accept time when generating ULID

zepfietje:
Currently `Str::ulid()` doesn't accept a time, although the underlying Symfony class does have a `generate` method that does accept it.

When migrating models from integer IDs to ULIDs, it's nice if you can keep the creation timestamp in the ULID.
This PR enables that.

Not sure what laravel/framework uses for type declarations, so let me know if anything needs changing.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -1380,7 +1380,11 @@ class Str
      */
     public static function ulid($time = null)
     {
-        return Ulid::generate($time);
+        if ($time === null) {
+            return new Ulid();
+        }
+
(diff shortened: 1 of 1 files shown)


taylorotwell:
This makes the method return a string instead of a ULID object, which is a breaking change.

taylorotwell:
Mark this as ready for review if you want me to look again.

zepfietje:
Thanks, @taylorotwell, good catch!

I've updated this PR to always return a ULID object.
Still calling the constructor directly when no time is passed to prevent any performance degradation by the internal validity check.

PR #46206, opened 2023-02-21
[9.x] Tests for Multi byte string in blade files 🔍

imanghafoori1:
- If you change the `strpos` to `mb_strpos`, the functionality will break for blade files containing multi-byte strings, but no tests will fail. So it is safer to have some of those characters in those blade files.

- It also removes some extra code originally copied/pasted from illuminate/Str class.
Here we never can have an int or an empty string passed into this method.

image


taylorotwell:
Hey there - no need to update the code, just update the test.

imanghafoori1:
I reverted the code changes, but let's keep in mind that there is a compile-time error in there:
image
image

PR #46209, opened 2023-02-21
[10.x] Add orderByField query builder method

mikaelpopowicz:
This PR adds a `orderByField` query builder method that is only supported by MySQL grammar.

```php
$keys = [42, 1, 1000];

$models = Users::query()
    ->whereKey($keys)
    ->orderByField('id', $keys)
    ->get();

// models will be sorted by the `$keys` values
```


taylorotwell:

PR #46212, opened 2023-02-21
[10.x] Fix missing `pid` method

SupianIDz:
The [documentation](https://laravel.com/docs/10.x/processes#process-ids-and-signals) says to get the pid of a running process using the `pid` method but that method doesn't exist, it should be `id`.

Screenshot from 2023-02-22 01-14-54

But based on the use of a well-known term among developers and sysadmins and also in some well-known software the name is `pid`.

Look at the image below, marked in red

## lsof
Screenshot from 2023-02-22 01-34-28

## netstat
Screenshot from 2023-02-22 01-34-37

Based on that, I think the term `pid` is better and `id` can still be used.

### DIFF
--- a/src/Illuminate/Contracts/Process/InvokedProcess.php
+++ b/src/Illuminate/Contracts/Process/InvokedProcess.php
@@ -9,7 +9,7 @@ interface InvokedProcess
      *
      * @return int|null
      */
-    public function id();
+    public function pid();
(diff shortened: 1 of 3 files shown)


SupianIDz:
Sorry, some spaces were removed by PHPStorm without my noticing it.

PR #46219, opened 2023-02-22
[10.x] Alternative to the stringy middleware API

timacdonald:
This is simplified and framework first implementation of my [HasParameters](https://github.com/timacdonald/has-parameters/) package.

Removes the need to use string concatenation when specifying middleware. No need for middleware classes to extend a base class or use a trait. The framework just handles things.

> **Note** This is just a wrapper around the stringy API. The middleware classes still receive the string representation of the value.

## Using a middleware class

```php
// before

Route::etc()->middleware([
    MyMiddleware::class.':'.implode(',', [Foo::ONE, Foo::TWO]),
]);

// after

Route::etc()->middleware([
    MyMiddleware::class => [Foo::ONE, Foo::TWO],
]);
```

## Using a middleware alias

```php
// before

Route::etc()->middleware([
    'myAlias:'.implode(',', [Foo::ONE, Foo::TWO]),
]);

// after

Route::etc()->middleware([
    'myAlias' => [Foo::ONE, Foo::TWO],
]);
```

## Optional single value API

You may omit the array if it is a single value.

```php
// before

Route::etc()->middleware([
   'alias:'.Foo::ONE,
]);

// after

Route::etc()->middleware([
    'alias' => Foo::ONE,
]);
```

## Handles backed enums

Remember that all values are still cast to a string. The middleware does not receive the enum.

```php
// before

Route::etc()->middleware([
   'alias:'.Suit::Diamonds->value,
]);

// after

Route::etc()->middleware([
    'alias' => Suit::Diamonds,
]);
```

### DIFF
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -1437,6 +1437,22 @@ class Router implements BindingRegistrar, RegistrarContract
     }
 
     /**
+     * Parse an array of middleware.
+     *
+     * @param  array  $middleware
+     * @return array
+     */
+    public function parseMiddleware($middleware)
+    {
+        return Collection::make($middleware)
+            ->map(fn ($value, $key) => is_int($key)
+                ? (string) $value
+                : $key.':'.Collection::wrap($value)->implode(','))
+            ->values()
+            ->all();
+    }
+
+    /**
      * Set the container instance used by the router.
      *
      * @param  \Illuminate\Container\Container  $container
--- a/src/Illuminate/Routing/Router.php
+++ b/src/Illuminate/Routing/Router.php
@@ -1447,7 +1447,9 @@ class Router implements BindingRegistrar, RegistrarContract
         return Collection::make($middleware)
             ->map(fn ($value, $key) => is_int($key)
                 ? (string) $value
-                : $key.':'.Collection::wrap($value)->implode(','))
+                : $key.':'.Collection::wrap($value)
+                    ->map(fn ($v) => $v === false ? '0' : $v)
+                    ->implode(','))
             ->values()
             ->all();
     }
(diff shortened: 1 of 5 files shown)


src/Illuminate/Routing/Router.php
```diff
@@ -1436,6 +1436,24 @@ public static function uniqueMiddleware(array $middleware)
...
+     * @param  array  $middleware
+     * @return array
+     */
+    public function parseMiddleware($middleware)
+    {
+        return Collection::make($middleware)
+            ->map(fn ($value, $key) => is_int($key)
+                ? (string) $value
+                : $key.':'.Collection::wrap($value)
+                    ->map(fn ($v) => $v === false ? '0' : $v)
```
timacdonald, line 1451: This is a bit of an opinion I've put in to make the casting of `true` and `false` consistent.

Usually:

```php
(string) true;

// "1"

(string) false;

// ""
```

I wanted `false` to be cast to a zero for consistency.

emargareten:
> **Note** This is a wrapped around the stringy API. The middleware classes still receive the string representation of the value.

Why shouldn't we have the option to pass in any type?

rodrigopedra:
> Why shouldn't we have the option to pass in any type?

I guess due to route serialization, when caching routes.

timacdonald:
The root reason, IMO, it that it would be a breaking change for user-land and 3rd party middleware.

Middleware expect only strings, and may be typed as such. As much as I would like to investigate allow rich type serialization / deserialization, it fundamentally changes middleware and feels like a major version change.

src/Illuminate/Routing/Router.php
timacdonald: I removed this. After some thought I don't think it shouldn't act differently.

taylorotwell:
Keeping it real - I honestly find your package better. Or, adding context specific builders to the applicable middleware:

```php
->middleware([Authenticate::using('web')]);
```

timacdonald:
I appreciate realness being kept.

PR #46228, opened 2023-02-22
[9.x] Fix expectsDatabaseQueryCount() $connection parameter

giggsey:
The listen() function was assuming that the event was being dispatched for just that connection, rather than globally.

Instead, we now check the connection name matches if it was provided.

Fixes #45932


giggsey:
I think the checks failed because I created this PR against 10.x instead of 9.x. I'm not sure which version I should have targeted. My feeling was 9.x because the feature was only added 3 weeks ago, and didn't work as expected.

timacdonald:
@giggsey I personally think this is a bug with the connection, rather than the test helper. Going to open a different PR to address the bug.

timacdonald:
See https://github.com/laravel/framework/pull/46231

giggsey:
> @giggsey I personally think this is a bug with the connection, rather than the test helper. Going to open a different PR to address the bug.
> 

Whilst I agree, I feel that's a bigger backwards incompatible change that has more potential to break user's applications.

At the very least, should the additional test in this PR be added after your new PR?

PR #46230, opened 2023-02-22
[10.x] Refactor existing code to include function for dropping columns if th…

ahmetsabri:
For MySql and Postgres drivers, The dropColumn method will work like the dropIfExists method, if the column exists it will be deleted, if not it will not throw any errors.


driesvints:
I don't feel we should do this as people might expect the error to know when they're calling the method wrongly.

PR #46231, opened 2023-02-22
[10.x] Ensures connections only listen for their own events

timacdonald:
The `Connection::listen()` method currently listens to queries on any connection.

```php
DB::connection('mysql')->listen(function () {
    dump('mysql query run');
});

DB::connection('sqlite')->listen(function () {
    dump('sqlite query run');
});

DB::connection('sqlite')->table('foo')->select('*');

// output:
// "mysql query run"
// "sqlite query run"
```

This is a bug and unexpected behaviour that is potentially problematic. It also impacts other framework behaviour that depends on adding listeners to the specific connection, e.g. the query duration handlers.

### DIFF
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -108,7 +108,7 @@ class Connection implements ConnectionInterface
     /**
      * The event dispatcher instance.
      *
-     * @var \Illuminate\Contracts\Events\Dispatcher
+     * @var \Illuminate\Contracts\Events\Dispatcher|null
      */
     protected $events;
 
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Connection.php
```diff
@@ -965,7 +965,11 @@ public function beforeExecuting(Closure $callback)
      */
     public function listen(Closure $callback)
     {
-        $this->events?->listen(Events\QueryExecuted::class, $callback);
+        $this->events?->listen(Events\QueryExecuted::class, function ($event) use ($callback) {
+            if ($event->connection === $this) {
+                $callback();
+            }
+        });
```
timacdonald, line 972: This is an identity check. I felt this was better than checking the connection name.

If you change the driver instance, it feels like you would want to also remove the listeners.

PR #46235, opened 2023-02-23
[10.x] Convert closures to arrow functions

ryosuke1202:
Transforms closures into their shortened equivalent using arrow functions in Eloquent directory.


taylorotwell:
They are fine as is for now.

PR #46237, opened 2023-02-23
[10.x] Add custom recaller segments callback to SessionGuard

JurianArie:
#42316 added support for reading additional segments from the recaller cookie. This allows you to set the segments on creation without having to extend the SessionGuard.

This would allow me to replace my custom guard with just one line: `SessionGuard::$customRecallerSegments = fn () => now()->timestamp;`

```php
<?php

namespace App\Support\Auth;

use Illuminate\Auth\SessionGuard as Guard;
use Illuminate\Contracts\Auth\Authenticatable;
use Illuminate\Support\Facades\Date;

class SessionGuard extends Guard
{
    /**
     * Get the name of the cookie used to store the "recaller".
     *
     * @return string
     */
    public function getRecallerName(): string
    {
        return 'remember_' . $this->name . '_' . sha1(Guard::class);
    }

    /**
     * Queue the recaller cookie into the cookie jar.
     *
     * @param \Illuminate\Contracts\Auth\Authenticatable $user
     * @return void
     */
    protected function queueRecallerCookie(Authenticatable $user): void
    {
        $this->getCookieJar()->queue($this->createRecaller(vsprintf('%s|%s|%s|%s', [
            $user->getAuthIdentifier(),
            $user->getRememberToken(),
            $user->getAuthPassword(),
            Date::now()->timestamp,
        ])));
    }
}
```

### DIFF
--- a/src/Illuminate/Auth/SessionGuard.php
+++ b/src/Illuminate/Auth/SessionGuard.php
@@ -41,6 +41,13 @@ class SessionGuard implements StatefulGuard, SupportsBasicAuth
     public readonly string $name;
 
     /**
+     * Add additional segments to the recaller cookie.
+     *
+     * @var callable(\Illuminate\Contracts\Auth\Authenticatable): string
+     */
+    public static $customRecallerSegments;
+
+    /**
      * The user we last attempted to retrieve.
      *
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #46239, opened 2023-02-23
[10.x] Add pipe function to Process layer

WendellAdriel:
This PR adds a new function `pipe` to the **Process** layer that receives an array of processes and runs them in sequence passing the output of the previous process as the input for the next one.

The syntax will be like this:

```php
$pipe = Process::pipe([
    Process::command('cat test.txt'),
    Process::command('grep -i "foo"'),
]);

$pipe->run(); // "foo"
```

```
Hello, world
foo
```

This can be very handy when we need to chain processes that rely on the result of the previous one to run correctly.


taylorotwell:
I wonder if this would be better if it worked like Pool where you could configure the Processes fully?

```php
Process::pipe([
    Process::timeout(10)->command('foo'),
    Process::timeout(10)->command('bar'),
]);
```

WendellAdriel:
@taylorotwell oh, that's great!
I'll work on the needed changes and I'll update the PR ASAP

taylorotwell:
One thing I'm curious about is what should the behavior be if a process in the middle of the pipeline fails?

WendellAdriel:
Sorry for the delay, @taylorotwell 
I was on vacation.
I think that if any process fails, we should throw an exception since the expected result is the result after running all the processes in the pipe

WendellAdriel:
@taylorotwell do you think that we need to create a custom exception for this? Like wrapping to throw a custom one like this:

```php
  public function run(?callable $output = null)
  {
      call_user_func($this->callback, $this);

      return collect($this->pendingProcesses)
              ->reduce(function ($previousProcessResult, $pendingProcess) use ($output) {
                  if (! $pendingProcess instanceof PendingProcess) {
                      throw new InvalidArgumentException('Process pipe must only contain pending processes.');
                  }

                  try {
                      return $pendingProcess->when(
                          $previousProcessResult,
                          fn () => $pendingProcess->input($previousProcessResult->output())
                      )->run(output: $output);
                  } catch (Exception $e) {
                      throw new ProcessPipeException($e->getMessage());
                  }
              });
  }
```

Or can we leave as it is?

driesvints:
Seems like something went wrong with branching here. Please resubmit this anew with a clean commit timeline.

PR #46240, opened 2023-02-23
[10.x] Add ```get``` method to HTTP Client Response

milwad-dev:
When you want to get data from response, They didn't method to get a specific field.

Now you can use a specific field easy:
```php
$response = Http::get('https://jsonplaceholder.typicode.com/todos/1');
$response->get('title');
```

It is not over yet, You can access to value of associative array in response like this 🔥:
```php
$response = Http::get('https://jsonplaceholder.typicode.com/users/1');
$response->get('company.name');
```

### DIFF
--- a/src/Illuminate/Http/Client/Response.php
+++ b/src/Illuminate/Http/Client/Response.php
@@ -274,6 +274,17 @@ class Response implements ArrayAccess
     }
 
     /**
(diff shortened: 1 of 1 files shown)


huangdijia:
What's the difference with the `json` method?

PR #46241, opened 2023-02-23
[10.x] type bool also needs to be accepted

emtudo:


src/Illuminate/Validation/Rules/DatabaseRule.php
```diff
@@ -81,7 +81,7 @@ public function resolveTableName($table)
      * Set a "where" constraint on the query.
      *
      * @param  \Closure|string  $column
-     * @param  \Illuminate\Contracts\Support\Arrayable|array|string|int|null  $value
+     * @param  \Illuminate\Contracts\Support\Arrayable|array|string|int|null|bool  $value
```
driesvints, line 84:
```suggestion
     * @param  \Illuminate\Contracts\Support\Arrayable|array|string|int|bool|null $value
```

PR #46243, opened 2023-02-23
[10.x] Replace status codes with Symfony Response

milwad-dev:


driesvints:
It's fine, thank you.

PR #46248, opened 2023-02-23
[10.x] Add a "channel:list" command

ash-jc-allen:
Hey!

This PR proposes a new `php artisan channel:list` Artisan command that can be used to list the registered broadcasting channels in your app.

## What is it?

My idea behind this is that I find the `route:list` command really useful. I sometimes end up working on projects that have a big-ish number of channels registered in the `routes/channels.php` file. So I tried to mimic the concept of the `route:list` command for broadcasting channels.

I think this command could possibly make it easier to get a quick overview of which private channels are registered in the application (in alphabetical order).

One thing I also like about the command is that I've added a warning message that lets you know if you've not uncommented the `BroadcastServiceProvider` in your `config/app.php` file. I don't know about other people, but this is something I tend to forget to do quite often when adding WebSockets to my applications haha! I then spend quite a while trying to figure out why I keep getting a 404 when trying to authorise a private channel using Echo. I'd like to think that this would reduce the chance of that happening.

So far, I've only done a limited amount of testing, so there may be some things I've missed. I've also not written any tests for it just yet, just in case it wasn't something you thought would be worth merging. If it's something you think might be worthwhile, I'll be happy to put some tests together.

## Future and limitations

At the moment, the command is relatively basic and just outputs the name of the channel and the channel class or closure that is used to resolve the authorisation logic.

I think this command could be extended in the future to have things like filtering (similar to the `route:list` command).

It could maybe also list the parameters that are expected by the closure/class (e.g. - an `App\Models\Chat` model).

I'm a little biased, but I think this command could be a handy little tool to improve the DX when working with broadcasting. I think it has potential to be extended with useful features.

One limitation I'm aware of is that it doesn't list the public channels because they're not listed in the `routes/channels.php` file. I don't know if it's possible to extend the command in the future to also detect other public channels that are being used in the event classes?

I also guessed at the design for the command and used the same approach as the `route:list` command. But I'm wondering if it would be better to make it more like (with the dots separating and the "resolver" being on the right side of the terminal):

```text
channel name ............ resolver
channel name ............ resolver
channel name ............ resolver
```

## Example screenshots

This screenshot shows what happens if there aren't private channels registered in the app:

Screenshot 2023-02-23 at 17 53 29

This screenshot shows what happens if the provider isn't registered and there are no private channels registered:

Screenshot 2023-02-23 at 17 53 54

This screenshot shows what happens if there are 4 channels registered:

Screenshot 2023-02-23 at 17 53 06

If this might be something you'd be interested in accepting, please let me know if there's anything you'd like changing. Thanks! 😄

### DIFF
--- a/src/Illuminate/Broadcasting/Broadcasters/Broadcaster.php
+++ b/src/Illuminate/Broadcasting/Broadcasters/Broadcaster.php
@@ -97,6 +97,16 @@ abstract class Broadcaster implements BroadcasterContract
     }
 
     /**
+     * Return a Collection of registered private channels.
+     *
+     * @return \Illuminate\Support\Collection
+     */
+    public function getChannels()
+    {
+        return collect($this->channels);
+    }
+
+    /**
      * Authenticate the incoming request for a given channel.
      *
      * @param  \Illuminate\Http\Request  $request
--- a/src/Illuminate/Broadcasting/Broadcasters/Broadcaster.php
+++ b/src/Illuminate/Broadcasting/Broadcasters/Broadcaster.php
@@ -97,16 +97,6 @@ abstract class Broadcaster implements BroadcasterContract
     }
 
     /**
-     * Return a Collection of registered private channels.
-     *
-     * @return \Illuminate\Support\Collection
-     */
-    public function getChannels()
-    {
-        return collect($this->channels);
-    }
-
-    /**
      * Authenticate the incoming request for a given channel.
      *
      * @param  \Illuminate\Http\Request  $request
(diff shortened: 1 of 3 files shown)


peterfox:
I'll be honest my first thought when I saw the title was it was about notification channels. 😅

Maybe `broadcast:list` instead?

mgcodeur:
Good idea 😄

ash-jc-allen:
@peterfox Ahh yeah I hadn't considered that. I chose `channel:list` because the broadcast channels are registered in the `routes/channels.php` file and you can make new broadcast channel classes with `make:channel`.

If it's something that's going to get merged, I'm cool to change it if needed 😄

taylorotwell:
Thanks 🙏

PR #46251, opened 2023-02-23
Update Handler.php to logging Auth exception

jbidad:
some message formatter ignore empty array and no log recorded


ankurk91:
What was the actual issue?

driesvints:
Please resubmit this with a thorough explanation why you need this, not just one sentence.

jbidad:
If a custom driver with provider used in `auth.config` and it was not implemented in `AuthServiceProvider`, therefore, it was catch in context method and returning empty array. I faced with similar issue which was no log was recorded. For the reason of this issue I added error logger to make it reportable.

PR #46258, opened 2023-02-24
Adding Conditionable to Logger

aglipanci:
This PR will add the Conditionable on the Logger to allow conditional logging. There are situations where a user would want conditional logging, by adding the Conditionable trait we can allow the when/unless syntax to be used easily.

```php
Log::when(true, function($logger) {
  $logger->debug('An informational messsdsage.');
});
```

```php
Log::unless(false, function($logger) {
  $logger->debug('An informational messsdsage.');
});
```

```php
Log::channel('daily')->when(false, function($logger) {
  $logger->debug('This wont be logged.');
});
```

### DIFF
--- a/src/Illuminate/Log/Logger.php
+++ b/src/Illuminate/Log/Logger.php
@@ -12,6 +12,8 @@ use RuntimeException;
 
 class Logger implements LoggerInterface
 {
+    use Conditionable;
+    
     /**
      * The underlying logger implementation.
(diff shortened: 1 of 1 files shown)


taylorotwell:
You didn't import the class?

aglipanci:
> You didn't import the class?

Sorry about that. Should I create a new PR?

PR #46264, opened 2023-02-24
Fix docblock

michaelnabil230:


taylorotwell:
I personally do not like to add many @throws blocks for every possible exception. Thank you!

PR #46265, opened 2023-02-24
[10.x] Add "dot" method to Illuminate\Support\Collection class

medeirosinacio:
This PR adds the dot method to the Illuminate\Support\Collection class in Laravel.

Additionally, I recently had a need to perform the dot operation on an array, extract 5 items from the flattened result, and then convert it back to a multi-dimensional array using the undot method. Unfortunately, achieving this with the current implementation required me to write the following code:

```php
Arr::undot(array_slice(Arr::dot($keys), 0, 5))
```
With this PR, users can now easily achieve the same result using the following code:

```php
collect($keys)->dot()->take(5)->undot()
```
This simplifies the code and makes it more readable, providing a better developer experience for users of Laravel's Illuminate\Support\Collection class.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -1578,6 +1578,16 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
     }
 
     /**
+     * Flatten a multi-dimensional associative array with dots.
+     *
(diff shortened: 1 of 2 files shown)


BafS:
Me thinking `dot` would do a [dot product](https://en.wikipedia.org/wiki/Dot_product).

PR #46268, opened 2023-02-25
[10.x] remove unused local variable

salehhashemi1992:
there was an extra local variable in the return statement.


salehhashemi1992:
@taylorotwell may I ask why?

PR #46269, opened 2023-02-25
[10.x] remove redundant optional argument

salehhashemi1992:
There was an extra redundant optional argument that matches the parameter's default value


taylorotwell:
I think we just do this to be explicit.

yungifez:
According to PHP docs, when set to false, it returns an object which is what the method intends to do

So it therefore doesn't seem redundant

PR #46274, opened 2023-02-25
[10.x] Add batch callback event to the schedule

markuskooche:
The current way to call a batch via the scheduler is via a command.

This works as follows:
```php
class CommandName extends Command
{
    /**
     *
     * @var string
     */
    protected $signature = 'command:name';

    // ...

    /**
     *
     * @throws Throwable
     */
    {
        Bus::batch([
            new ImportCsv(1),
            new ImportCsv(2),
        ])->then(function (Batch $batch) {
            // All jobs completed successfully...
        })->catch(function (Batch $batch, Throwable $e) {
            // First batch job failure detected...
        })->finally(function (Batch $batch) {
            // The batch has finished executing...
        })->dispatch();
    }
}
```
```php
class Kernel extends ConsoleKernel
{
    /**
     * Define the application's command schedule.
     *
     * @param  Schedule  $schedule
     * @return void
     */
    protected function schedule(Schedule $schedule): void
    {
        $schedule->command('command:name')->everyFifteenMinutes();
    }

    // ...
}
```

---
But maybe you don't need or want a command. In addition, this creates unnecessary boilerplate.
This PR allows job batching also via the scheduler.

A batch can be passed to the scheduler as follows:
```php
class Kernel extends ConsoleKernel
{
    /**
     * Define the application's command schedule.
     *
     * @param  Schedule  $schedule
     * @return void
     */
    protected function schedule(Schedule $schedule): void
    {
        $schedule->batch(
            Bus::batch([
                new ImportCsv(1),
                new ImportCsv(2),
            ])->then(function (Batch $batch) {
                // All jobs completed successfully...
            })->catch(function (Batch $batch, Throwable $e) {
                // First batch job failure detected...
            })->finally(function (Batch $batch) {
                // The batch has finished executing...
            })
        )->everyFifteenMinutes();
    }

    // ...
}
```

The console output looks like this:
```
  2023-02-25 19:02:00 Running [Batch of 2 jobs] ................. 5,211ms DONE
```

### DIFF
--- a/src/Illuminate/Console/Scheduling/Schedule.php
+++ b/src/Illuminate/Console/Scheduling/Schedule.php
@@ -161,7 +161,7 @@ class Schedule
     /**
      * Add a new batch callback event to the schedule.
      *
-     * @param  \Illuminate\Bus\PendingBatch $batch
+     * @param  \Illuminate\Bus\PendingBatch  $batch
      * @return \Illuminate\Console\Scheduling\CallbackEvent
      *
      * @throws \RuntimeException
--- a/src/Illuminate/Console/Scheduling/Schedule.php
+++ b/src/Illuminate/Console/Scheduling/Schedule.php
@@ -159,6 +159,21 @@ class Schedule
     }
 
     /**
+     * Add a new batch callback event to the schedule.
+     *
+     * @param  \Illuminate\Bus\PendingBatch $batch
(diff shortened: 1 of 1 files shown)


taylorotwell:
You can just schedule a Closure?

```php
$schedule->call(function () { /* Dispatch batch... */ });
```

PR #46276, opened 2023-02-26
add custom public folder via .env

abunaum:


ankurk91:
You can not use `env` helper directly

antonkomarev:
`env` method call is allowed only in config files. It is a very expensive operation and method call will always return `null` if config caching is enabled.

ankurk91:
Btw, Laravel already allows you to customise the public path

```php
app()->usePublicPath(__DIR__.'/public');
```

https://laravel.com/docs/10.x/upgrade#public-path-binding

PR #46279, opened 2023-02-26
[10.x] Add ArrayAccess to Stringable

mathieutu:
Hi,

This little PR adds `ArrayAccess` to `Stringable`.

It allows to make its behavior more like a real string, by proxifying the calls.

```diff
- str('My StRinG')->lower()->value()[0]
+ str('My StRinG')->lower()[0]
```

Thanks,
Matt'

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -397,6 +397,7 @@ class Stringable implements JsonSerializable, ArrayAccess
     /**
      * Convert GitHub flavored Markdown into HTML.
      *
+     * @param  array  $options
      * @return static
      */
     public function markdown(array $options = [])
@@ -407,6 +408,7 @@ class Stringable implements JsonSerializable, ArrayAccess
     /**
      * Convert inline Markdown into HTML.
      *
+     * @param  array  $options
      * @return static
      */
     public function inlineMarkdown(array $options = [])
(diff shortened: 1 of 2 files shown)


src/Illuminate/Support/Stringable.php
```diff
@@ -396,7 +397,6 @@ public function lower()
     /**
      * Convert GitHub flavored Markdown into HTML.
      *
-     * @param  array  $options
```
rodrigopedra, line 399: Why this was removed? (and 4 other `@param` and one `@return` at `jsonSerialize`)

tests/Support/SupportStringableTest.php
```diff
@@ -1179,4 +1179,13 @@ public function testToDateThrowsException()
 
         $this->stringable('not a date')->toDate();
     }
+
+    public function testArrayAccess()
+    {
+        $str = $this->stringable('my string');
+        $this->assertSame('m', $str[0]);
+        $this->assertSame('s', $str[4]);
```
rodrigopedra, line 1187: This should be `$this->assertSame('t', $str[4]);` or `$this->assertSame('s', $str[3]);`.

|  index |  char |
|:-------:|:-----:|
| 0 | m |
| 1 | y |
| 2 | SPACE |
| 3 | s |
| 4 | t |
| 5 | r |
| 6 | i |
| 7 | n |
| 8 | g |

src/Illuminate/Support/Stringable.php
mathieutu: Hum weird indeed, I wrote that directly in github.dev, the ide should have removed them.

tests/Support/SupportStringableTest.php
mathieutu: Indeed, thanks for the catch.

PR #46281, opened 2023-02-27
[10.x] Do not mutate underlying values on redirect

### LINKED ISSUE #46273: Parameters in Route::redirect() are handled differently for implicit/explicit bindings
- Laravel Version: 10.1.3
- PHP Version: 8.2.0

### Description:

Consider the following:

```php
// web.php
Route::redirect('user/{user}', '/user/{user}/overview');
Route::get('user/{user}/overview', UserController::class);

// app/Http/Controllers/UserController.php
class UserController
{
    public function __invoke(User $user)
    {
          // ...
    }
}

```

When ran as-is and hence implicitly binding the User model, the redirect works as expected. The `redirect()` method receives a user id and splices it into the destination URL. As a consequence the `{user}` parameter is only substituted in the second request (after the redirect).

When explicitly defining the binding, things change:
```php
// app/Providers/RouteServiceProvider.php : boot()
Route::bind('user', function ($id) {
    return User::findOrFail($id); // probably extended with additional query logic.
});
```

The `{user}` parameter is now also substituted in the **first** request, meaning that the framework's  `RedirectController` receives a full user model which it'll be unable to easily splice into the destination URL. Ultimately the application breaks and an exception is thrown.

Writing logic in the binding to check for redirect routes feels rather cumbersome.
I think this is a valid use case for the framework to disable the `SubstituteBindings` middleware in case of redirect routes... Can't think of any scenario where raw request parameters shouldn't end up in the destination url as-is.

timacdonald:
Fixes #46273

Due of the use of `toArray` the underlying values, e.g. an Eloquent model, are also cast to an array if they implement the Arrayable contract.

This means that the following will fail, as the explicit route model binding will return an eloquent model and the model will then be cast to an array and finally the array will be used as the "user" parameter.

```php
Route::redirect('users/{user}', 'users/{user}/overview');

Route::get('users/{user}/overview', fn () => 'foo');

Route::bind('user', fn ($id) => User::findOrFail($id));
````

This should not be a breaking change, as the router can not handle parameters that are an array. The following will throw an "array to string exception"...

```php
Route::redirect('users/{user}', 'users/{user}/overview');

Route::get('users/{user}/overview', fn () => 'foo');

// bind to an array
Route::bind('user', fn ($id) => [$id]);
````

### DIFF
--- a/src/Illuminate/Routing/RedirectController.php
+++ b/src/Illuminate/Routing/RedirectController.php
@@ -31,7 +31,7 @@ class RedirectController extends Controller
 
         $parameters = $parameters->only(
             $route->getCompiled()->getPathVariables()
-        )->toArray();
+        )->all();
 
         $url = $url->toRoute($route, $parameters, false);
 
--- a/tests/Integration/Routing/RouteRedirectTest.php
+++ b/tests/Integration/Routing/RouteRedirectTest.php
@@ -2,6 +2,8 @@
 
 namespace Illuminate\Tests\Integration\Routing;
 
+use Illuminate\Foundation\Auth\User;
+use Illuminate\Routing\Middleware\SubstituteBindings;
 use Illuminate\Support\Facades\Route;
 use Orchestra\Testbench\TestCase;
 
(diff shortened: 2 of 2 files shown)


Propaganistas:
Seems like an elegant fix for my bug report. Thanks!

Issue #40380, opened 2022-01-13
Basic (not backed) native Enums are not supported in casting or validation


chojnicki:
- Laravel Version: 8.78.0
- PHP Version: 8.1
- Database Driver & Version: Mysql 8.0.27

### Description:

Backed enums are working just fine, like
```
enum Demo: int
{
    case AAA = 1;
    case BBB = 2;
    case CCC = 3;
}
```

or

```
enum Demo: string
{
    case AAA = 'AAA';
    case BBB = 'BBB';
    case CCC = 'CCC';
}
```

In my case I need to use strings, but second option is repetitive and easy for typo in long lists.

Simple (not backed) enum are simpler and suitable option:

```
enum Demo
{
    case AAA;
    case BBB;
    case CCC;
}
```

But based on changes in #39315 and #39608 it uses `from()` method on Enum to find value, which does not exist on non-backed enum.

`Call to undefined method App\\Enums\\Demo::from()`

Is there any reason for that? Is it on purpose or bug? Sure that I can just give up on enums and use strings, but non-backend enums are also helpful. In my case I want to create enum in Lighthouse to pass it directly to GraphQL schema. It's good practice to use same enum definition in both PHP and graphql schema, to avoid bugs, but currently it's not possible with Laravel/Lumen and non-backed enums.

For tests purposes I changed some lines here
_illuminate/database/Eloquent/Concerns/HasAttributes.php_
```
protected function getEnumCastableAttributeValue($key, $value)
{
    if (is_null($value)) {
        return;
    }

    $castType = $this->getCasts()[$key];

    if ($value instanceof $castType) {
        return $value;
    }

    // HERE
    return $castType::cases()[array_search($value, array_column($castType::cases(), 'name'))];
    // return $castType::from($value);
}
```

Sure it's just quick and ugly fix, code could by nicer and this will broke backed enums, but it worked just fine for my case. 

But problem will maintain also in other places, like validation:

_Illuminate\Validation\Rules\Enum_
```
public function passes($attribute, $value)
{
    if (is_null($value) || ! function_exists('enum_exists') || ! enum_exists($this->type) || ! method_exists($this->type, 'tryFrom')) {
        return false;
    }

    return ! is_null($this->type::tryFrom($value));
}
```

Validation uses `tryFrom`, so same problem.

I can make this on app Level by making Trait (and use it in every enum file) that will mimic `::tryFrom()` and `::from`, but trying to reproduce PHP methods feels to much like hacking for me...

[...]

PR #46293, opened 2023-02-28
[10.x] Using enum case's key for pure Enum validation

yeganemehr:
Related: #40380
This PR handle [pure enum](https://www.php.net/manual/en/language.enumerations.basics.php) validation.
If provided `$type` is a pure enum, the value checks against enum case keys.
There is no breaking change.

### DIFF
--- a/src/Illuminate/Validation/Rules/Enum.php
+++ b/src/Illuminate/Validation/Rules/Enum.php
@@ -2,8 +2,10 @@
 
 namespace Illuminate\Validation\Rules;
 
+use BackedEnum;
 use Illuminate\Contracts\Validation\Rule;
 use TypeError;
+use UnitEnum;
 
 class Enum implements Rule
 {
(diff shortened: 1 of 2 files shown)


henzeb:
If PR is not merged, one can use https://packagist.org/packages/henzeb/enumhancer for it. I will release a version supporting 10 today.

milwad-dev:
@yeganemehr 
What is the benefit?

henzeb:
> @yeganemehr What is the benefit?

The benefit of this is that we can use basic enumerations with this. Basic enumerations are designed for internal use where you do not store things in a database, but that's a bit overdone, as sometimes a value is literally the same as the key and storing a key as is is not unusual.

milwad-dev:
Oh. For basic validation of enum, you can use Enum class that exists in Core of Laravel.

henzeb:
> Oh. For basic validation of enum, you can use Enum class that exists in Core of Laravel.

That's the problem. It uses tryFrom. Meaning you have to add your own tryFrom implementation to create an enum from key, as basic enumerations doesn't have tryFrom.

taylorotwell:

PR #46300, opened 2023-03-01
[10.x] Added "isValidSignedUrl" method

ash-jc-allen:
Hey!

This PR is only a small one and proposes a new `URL::isValidSignedUrl` method for syntactic sugar.

I've been working with some code where we're storing signed URLs in the database. In my tests, I'm currently asserting that we're storing these signed URLs by using something along these lines:

```php
$this->assertTrue(
    URL::hasValidSignature(Request::create($model->signed_url))
);
```

With the new proposed method, this could be changed to something like so:

```php
$this->assertTrue(
    URL::isValidSignedUrl($model->signed_url)
);
```

I feel like the new method might be a bit easier to understand because you can pass a raw string to the method rather than needing to use `Request::create($url)`.

I know this is only a small change, but if it's something you think might be worthwhile, please let me know if you'd like anything changing 😄

### DIFF
--- a/src/Illuminate/Routing/UrlGenerator.php
+++ b/src/Illuminate/Routing/UrlGenerator.php
@@ -451,9 +451,9 @@ class UrlGenerator implements UrlGeneratorContract
     /**
      * Determine whether the given URL is a valid, signed URL.
      *
-     * @param string $url
-     * @param bool $absolute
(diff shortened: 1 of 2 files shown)


taylorotwell:

PR #46307, opened 2023-03-01
[10.x] Added whereTrue() and whereFalse() method to query builder

troccoli:
This PR is in response to a pool I took: #46071 .

It's not uncommon to have boolean fields in tables, and querying for them is only possible with something like
```php
$query->where('admin', true);
```

To make the code more readable, this PR adds two methods: `whereTrue()` and `whereFalse()`, plus the "or" versions. So now, the query above can be rewritten as
```php
$query->whereTrue('admin');
```

The SQL created is
```sql
"admin" = true
```

which is valid in MySQL, PostgreSQL, SQLite and SQL Server

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -2037,6 +2037,52 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Add a "where 'field' = true" clause to the query.
+     *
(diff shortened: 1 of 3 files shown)


milwad-dev:
https://github.com/laravel/framework/pull/43088

taylorotwell:

PR #46310, opened 2023-03-01
Update Handler.php to logging Auth exception 

jbidad:
If a custom driver with provider used in `auth.config` and it was not implemented in `AuthServiceProvider`, therefore, it was catch in context method and returning empty array. I faced with similar issue which was no log was recorded. For the reason of this issue I added error logger to make it reportable


lk77:
You can implement that function in app/Http/Exceptions/Handler.php

jbidad:
> You can implement that function in app/Http/Exceptions/Handler.php

It wrong way to fix something like this in app!

lk77:
@jbidad there is nothing to fix

PR #46313, opened 2023-03-01
[9.x] Make createIndexName function public

chrisflorin:
Having this function public is useful to create a default index name according to the standard without creating your own function.  This is useful for checking if an index already exists.

See https://stackoverflow.com/a/45883428/6383358 for an example of a use case.

Also cleaned up function definition for later versions of Php.


chrisflorin:
Should I retry merging this into 10.x?  I'd contend that this feature would be useful; otherwise anytime anybody needs the name of an index, that developer needs to create their own function that emulates this one.  Would be useful to use a Framework standard function for something like this.

PR #46317, opened 2023-03-02
[10.x] Config markdown theme ignored

AEM5299:
This is a bug introduced by my earlier PR #46200. It is always defaulting to `default` theme unless explicitly set in the Mailable class, which ignored the config option ([reported by @pikseli](https://github.com/laravel/framework/pull/46200#issuecomment-1448217021)). This should fix that bug.

I am not sure if this is the best approach, so open to any suggestions or other ideas.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,87 @@
 # Release Notes for 10.x
 
-## [Unreleased](https://github.com/laravel/framework/compare/v10.0.0...10.x)
+## [Unreleased](https://github.com/laravel/framework/compare/v10.1.5...10.x)
+
+
+## [v10.1.5 (2023-02-24)](https://github.com/laravel/framework/compare/v10.1.4...v10.1.5)
+
+### Fixed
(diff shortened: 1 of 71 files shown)


driesvints:
Thanks @AEM5299. But I'm afraid this PR can't be merged as it contains breaking changes to class constructs and replaces properties. I'll try to have a look myself.

AEM5299:
We can get the value from config instead of always setting it to `'default'` [here](https://github.com/laravel/framework/blob/3c5ae94660bc72e9527e96cbd263733d9c5f9d6f/src/Illuminate/Mail/Mailable.php#L327). We can also change `Markdown` to not be a singleton, since it correctly sets the config value on construction.

driesvints:
I think the former would work. You can maybe use ` Container::getInstance()` to resolve the config value.

driesvints (approved):

u01jmg3:
@AEM5299: not sure if it's intentional but your PR no longer has any test coverage.

taylorotwell:
Can you please fix conflicts?

taylorotwell:
Fixed here: https://github.com/laravel/framework/commit/791f8ea70b5872ae4483a32f6aeb28dd2ed4b8d7

PR #46319, opened 2023-03-02
[10.x] Remove obsolete function_exists('enum_exists') calls

jnoordsij:
The `function_exists('enum_exists')` call is used in various places for feature detection. As the minimum PHP version for Laravel 10 is 8.1, this should now always return true, and therefore can be safely removed.


driesvints (approved):
Thanks!

PR #46323, opened 2023-03-02
[10.x] Remove unnecessary arguments.

Tatsurou-Yajima:
The following two arguments were not used in the `execute` method.

`$input`, `$output`

Unnecessary arguments are confusing and should be removed.


driesvints:
As you can see, this breaks the test suite.

PR #46333, opened 2023-03-02
[10.x] Replace deprecated `Rule` class with `ValidationRule` in Enum validation

michaelnabil230:
This PR updates the Enum validation rules in Laravel to use the new `ValidationRule` class instead of the deprecation `Rule` class. This is important because the Rule class has been removed in the new version of Laravel.


taylorotwell:
The `Rule` class isn't being removed? It's just no longer the preferred choice.

PR #46337, opened 2023-03-03
[10.x] Add additional type to $app property to match app() helper.

jayanratna:
I am proposing this change after I added `preventLazyLoading` to my `AppServiceProvider` and realised that the code snippet suggested in the docs (shown below) shows an undefined method error in VSCode. This was mostly inspired by the addition of the same type in the `app()` helper, the commit for that can be found here: https://github.com/laravel/framework/commit/2c4a4e7338227bf6ba549402790ef5ba820873ce

Code snippet from docs:

```php
 
/**
 */
public function boot(): void
{
    Model::preventLazyLoading(! $this->app->isProduction());
}
```

### DIFF
--- a/src/Illuminate/Support/ServiceProvider.php
+++ b/src/Illuminate/Support/ServiceProvider.php
@@ -15,7 +15,7 @@ abstract class ServiceProvider
     /**
      * The application instance.
      *
-     * @var \Illuminate\Contracts\Foundation\Application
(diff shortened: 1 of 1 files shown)


driesvints:
This doesn't makes much sense to me. It should be one or the other I think. If you're concerned about the missing method you can also do:

```php
Model::preventLazyLoading(! $this->app->environment('production'));
```

PR #46344, opened 2023-03-03
[10.x] Add processors to logging (placeholders)

alanpoulain:
Maybe this PR should target `11.x`.

This PR adds the possibility to use Monolog processors and adds a default `PsrLogMessageProcessor` to all drivers.
Following @Crell's [post](https://peakd.com/hive-168588/@crell/using-psr-3-placeholders-properly), the logger should be able to use placeholders by default:
```php
Log::info('Showing the user profile for user: {id}', ['id' => $id]);
```
It's not the case today so I added it in this PR.
It can be disabled with `with_placeholders`:
```php
// logging.php

'channels' => [
    'single' => [
        'driver' => 'single',
        'path' => storage_path('logs/laravel.log'),
        'level' => env('LOG_LEVEL', 'debug'),
        'with_placeholders' => false,
    ],
],
```

This PR also adds the possibility to configure processors for the `monolog` driver:

```php
// logging.php

'channels' => [
    'memory' => [
        'driver' => 'monolog',
        'handler' => Monolog\Handler\StreamHandler::class,
        'level' => 'notice',
        'with' => [
            'stream' => 'php://stderr',
        ],
        'processors' => [
            // simple syntax
            Monolog\Processor\MemoryUsageProcessor::class,
            // complex syntax
            ['processor' => Monolog\Processor\PsrLogMessageProcessor::class, 'with' => ['removeUsedContextFields' => true]],
        ],
    ],
],
```

### DIFF
--- a/src/Illuminate/Log/LogManager.php
+++ b/src/Illuminate/Log/LogManager.php
@@ -295,7 +295,7 @@ class LogManager implements LoggerInterface
                     $config['bubble'] ?? true, $config['permission'] ?? null, $config['locking'] ?? false
                 ), $config
             ),
-        ], $config['with_placeholders'] ?? true ? [new PsrLogMessageProcessor()] : []);
+        ], $config['replace_placeholders'] ?? false ? [new PsrLogMessageProcessor()] : []);
     }
 
     /**
@@ -311,7 +311,7 @@ class LogManager implements LoggerInterface
                 $config['path'], $config['days'] ?? 7, $this->level($config),
                 $config['bubble'] ?? true, $config['permission'] ?? null, $config['locking'] ?? false
             ), $config),
-        ], $config['with_placeholders'] ?? true ? [new PsrLogMessageProcessor()] : []);
+        ], $config['replace_placeholders'] ?? false ? [new PsrLogMessageProcessor()] : []);
     }
 
     /**
@@ -335,7 +335,7 @@ class LogManager implements LoggerInterface
                 $config['bubble'] ?? true,
                 $config['exclude_fields'] ?? []
             ), $config),
-        ], $config['with_placeholders'] ?? true ? [new PsrLogMessageProcessor()] : []);
+        ], $config['replace_placeholders'] ?? false ? [new PsrLogMessageProcessor()] : []);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


ankurk91:
To preserve previous behaviour `with_placeholders` must be treated as false by framework, and let user opt in to this feature

alanpoulain:
Sure but if it's `false` by default, it contradicts what the post is saying, particularly in term of security:
> Passing potentially user-provided data into a log message is a security risk, and the escaping mechanism needed for them is not known at call time.

Crell:
The behavior change would be that, currently, you get the following written to the log:

```
Showing the user profile for user: {id} ['id' => 5]
```

whereas with the placeholders interpolated, you'd get this:

```
Showing the user profile for user: 5 ['id' => 5]
```

The degree to which that's a breaking change is not my call to make.  I will note, though, that whether you want the placeholders interpolated is going to vary by the handler.  For writing to syslog, probably yes.  For writing to a DB, almost certainly no.  For writing to a remote logging service, maybe, depending on if the service supports placeholders.

So the default should probably vary by the handler used.

dennisprudlo:
The default can be `false` for 10.x and `true` for 11.x and above.

Then apps using 10.x can opt in, and apps that don't want to use this can opt out in 11.x

alanpoulain:
It could, sure (and I can do it), but it's worth the (very small) risk to do it now in my opinion.
But I will comply to what Laravel mergers say.

kayw-geek:
> Maybe this PR should target `11.x`.
> 
> This PR adds the possibility to use Monolog processors and adds a default `PsrLogMessageProcessor` to all drivers. Following @Crell's [post](https://peakd.com/hive-168588/@crell/using-psr-3-placeholders-properly), the logger should be able to use placeholders by default:
> 
> ```
> Log::info('Showing the user profile for user: {id}', ['id' => $id]);
> ```
> 
> It's not the case today so I added it in this PR. It can be disabled with `with_placeholders`:
> 
> ```
> // logging.php
> 
> 'channels' => [
>     'single' => [
>         'driver' => 'single',
>         'path' => storage_path('logs/laravel.log'),
>         'level' => env('LOG_LEVEL', 'debug'),
>         'with_placeholders' => false,
>     ],
> ],
> ```
> 
> This PR also adds the possibility to configure processors for the `monolog` driver:
> 
> ```
> // logging.php
> 
> 'channels' => [
>     'memory' => [
>         'driver' => 'monolog',
>         'handler' => Monolog\Handler\StreamHandler::class,
>         'level' => 'notice',
>         'with' => [
>             'stream' => 'php://stderr',
>         ],
>         'processors' => [
>             // simple syntax
>             Monolog\Processor\MemoryUsageProcessor::class,
>             // complex syntax
>             ['processor' => Monolog\Processor\PsrLogMessageProcessor::class, 'with' => ['removeUsedContextFields' => true]],
>         ],
>     ],
> ],
> ```

The `with_placeholders` is wrong, we should  replace it using the `replace_placeholders`.

PR #46349, opened 2023-03-04
[10.x] Add charAt method to both Str and Stringable

localusercamp:
## What does this PR propose

This PR adds new `charAt` method to both `Illuminate\Support\Str` and `Illuminate\Support\Stringable`.
`charAt` method allows to get a character by index from a **multibyte** string.

If this PR is accepted, you can do the following:
```php
str('Hello, world!')->charAt(1); // (string) 'e'
str('Привет, мир!')->charAt(0); // (string) 'П'
str('「こんにちは世界」')->charAt(4); // (string) 'ち'
```
It also supports negative indexes:
```php
str('Hello, world!')->charAt(-1); // (string) '!'
str('Привет, мир!')->charAt(-12); // (string) 'П'
str('「こんにちは世界」')->charAt(-2); // (string) '界'
```
If index is greater then string length, or if negative index less then negative length, then `null` will be returned:
```php
str('Hello, world!')->charAt(100); // null
str('Привет, мир!')->charAt(-13); // null
str('「こんにちは世界」')->charAt(-2000); // null
```

## Why it do so

Str provides a better way to work with strings in Laravel, because it always work with multibyte strings.
For now there is no way to **simply** get character by index if you work not only with latin characters.

```php
$string = 'Привет';
$string[0]; // not a 'П'

// ArrayAccess was recently added for Stringable
str($string)[0] // not a 'П'

// To actually get a 'П' we need to do the following
$cyrillic_p = mb_substr($string, 0, 1); // 'П'

// I think the following is much better
Str::charAt($string, 0); // 'П'
// or
str($string)->charAt(0); // 'П'
```

## Behavior I would like to discuss
Current implementation will return `null` if index is greater then string length (or if negative index less then negative string length). `null` indicates that string does not contain character at that index. For me this behaviour seems to be more strict then returning empty string (`''`). If you see more meaning in returning empty strings, please describe why.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -217,7 +217,7 @@ class Str
     }
 
     /**
-     * Get character at the specified index.
+     * Get the character at the specified index.
      *
      * @param  string  $subject
      * @param  int  $index
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -217,6 +217,24 @@ class Str
     }
 
     /**
+     * Get character at the specified index.
+     *
+     * @param  string  $subject
+     * @param  int  $index
+     * @return string|null
+     */
+    public static function charAt($subject, $index)
+    {
+        $length = mb_strlen($subject);
+
+        if ($index < 0 ? $index < -$length : $index > $length - 1) {
+            return null;
+        }
+
+        return mb_substr($subject, $index, 1);
+    }
+
+    /**
      * Determine if a given string contains a given substring.
      *
      * @param  string  $haystack
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -100,17 +100,7 @@ class Stringable implements JsonSerializable, ArrayAccess
     }
 
     /**
-     * Get the basename of the class path.
-     *
-     * @return static
-     */
-    public function classBasename()
-    {
-        return new static(class_basename($this->value));
-    }
-
-    /**
-     * Get character at the specified index.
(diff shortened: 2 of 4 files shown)


rodrigopedra:
@taylorotwell 

Shouldn't this, as a method that return a string, return a new `Stringable` instance instead?

Someone could be iterating a list of names to grab its first letter and ensure it is uppercase, and try this:

```php
Str::of($name)->charAt(0)->upper();
```

Of course, fixing this is as simple as inverting the method's call, but I wonder if we should return a new instance for consistency with all the other methods which return a string.

Before this PR, other than `Stringable@value`, `Stringable@toString`, `Stringable@jsonSerialize`, and `Stringable@__toString`, where one would expect a static type to be returned, only the `Stringable@excerpt` method returns a raw string, which maybe could also be reviewed to keep it consistent with all the other methods.

localusercamp:
Recently ArrayAccess was added to Stringable. I think we could return `static` from `charAt` if ArrayAccess will use `charAt` under the hood but return string instead:
```php
public function offsetGet($offset)
{
    return $this->charAt($offset)->value();
}
```

Then it will be consistent. What do you think? @rodrigopedra

rodrigopedra:
Yes, I think returning a string from `offsetGet` is fine, as that method allows the user to use the instance as a raw string, and any user would be less surprised to find out they get a char back.

That would be a nice middle ground solution, keeping the method consistent with others, while allowing direct multibyte character access.

localusercamp:
@taylorotwell What do you think about that? Should I create new PR to allow ArrayAccess work with `charAt` and change return type of `charAt`?

PR #46354, opened 2023-03-05
allow serialization customization of relationships

browner12:
currently you can set which attributes are "visible" or "hidden" when an Eloquent Model is serialized, and you can also set whether a relationship is visible or hidden.  but you cannot set the visibility of the individual attributes of your relationships at call time.  this change allows users to do that.

this can be activated by using dot notation

```php
$product->setVisible(['id', 'name', 'price', 'manufacturer', 'manufacturer.id', 'manufacturer.name']);
```

Couple questions:

1.  Are relationships guaranteed to be Eloquent Models or Eloquent Collections? If so, I can probably remove the conditional check for that.
2. It would be nice to make this recursive so it could support any depth of relationships. Wanted to gauge interest first, though.
3. Not sure what the current policy is on typing in the framework, so I might have to make some adjustments with that.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -353,9 +353,7 @@ trait HasAttributes
         $attributes = [];
 
         foreach ($this->getArrayableRelations() as $key => $value) {
-
             if ($value instanceof Collection || $value instanceof Model) {
-
                 if (count($this->getVisibleForRelation($key))) {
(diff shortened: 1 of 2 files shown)


MohammadAlhallaq:
I think it would be nice if it gets merged, I was thinking of the same thing when I had to map over the result and unset unwanted data from the eager loaded relations,

taylorotwell:

browner12:
@taylorotwell are you open to this feature with a different implementation, or against the feature all together?

PR #46356, opened 2023-03-05
[10.x] Add string capability to `withTrashed` method

milwad-dev:
When we want to pass a function to the `withTrashed` method, it must be inside an array, but maybe just need to passed one function!
But now you can send string.

Compare 🔥:
```php
Route::resource('milwad', null)->withTrashed(['show']);
Route::resource('milwad', null)->withTrashed('show');
```

### DIFF
--- a/src/Illuminate/Routing/PendingResourceRegistration.php
+++ b/src/Illuminate/Routing/PendingResourceRegistration.php
@@ -230,11 +230,15 @@ class PendingResourceRegistration
     /**
      * Define which routes should allow "trashed" models to be retrieved when resolving implicit model bindings.
      *
-     * @param  array  $methods
+     * @param  array|string  $methods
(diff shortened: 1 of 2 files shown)


dennisprudlo:
I don't understand where the reduced time comes from. How many iterations did you do to test the performance?

I mean it's still an array, you just allow to pass a string you then manually cast to an array (so it's rather an extra step)

MohammadAlhallaq:
isn't it just based as a string and then manually casted to an array? how did the improvement happen?

```
 if (is_string($methods)) {
            $methods = (array) $methods;
        }    
```

milwad-dev:
@MohammadAlhallaq @dennisprudlo 
Yes, Yes.
I did a wrong test, actually this PR is to add string capability to the `withTrashed` method.

PR #46359, opened 2023-03-05
[10.x] db:seed with rollback on failure (rof) option 

Sairahcaz:
Many of our clients now want a working test suite, which is great!
So I lately quite often find me refactoring their **migrations**, **factories** and **seeders**.

So after tweaking around, you test if the migration and seeders work by hitting:
`migrate:fresh --seed`

Fine, migrations are finished, but ohh no the seeder throws an error.

Now your database has a total unknown state and you also don't exactly know which entries were seeded correctly.

If you have a lot of migrations you don't want to run `migrate:fresh --seed` again and again for each test, cause only the seeder don't worked and that would increase the execution of each "migration test" a lot...

Now a `db:truncate` command, which truncates all your tables, would be cool, so that you can just run `db:seed` afterwards.
But that would mean each time the migration fails you have to do `db:truncate && db:seed`

I therefor added the `db:seed --rof` option, which would **R**ollback **O**n **F**ailure.

In the second commit I made this also the default option for `migrate --seed` and `migrate:fresh --seed`, which would make a lot of sense in this context, but you can decide to maybe take only the first commit for some reason, this would also help a bit here.

Thanks a lot!

### DIFF
--- a/src/Illuminate/Database/Console/Migrations/FreshCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/FreshCommand.php
@@ -94,6 +94,7 @@ class FreshCommand extends Command
             '--database' => $database,
             '--class' => $this->option('seeder') ?: 'Database\\Seeders\\DatabaseSeeder',
             '--force' => true,
+            '--rof' => true,
         ]));
     }
 
--- a/src/Illuminate/Database/Console/Migrations/MigrateCommand.php
+++ b/src/Illuminate/Database/Console/Migrations/MigrateCommand.php
@@ -97,6 +97,7 @@ class MigrateCommand extends BaseCommand implements Isolatable
                 $this->call('db:seed', [
                     '--class' => $this->option('seeder') ?: 'Database\\Seeders\\DatabaseSeeder',
                     '--force' => true,
+                    '--rof' => true,
                 ]);
             }
         });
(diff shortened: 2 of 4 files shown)


MohammadAlhallaq:
it's a great idea but I don't think making it the default option is a good idea since it would feel like unexpected behavior or something odd, so documenting the new arg would be enough.

Sairahcaz:
I guess without this to be the default, we would need a **db:truncate** command, so that you can easily reset the database after a failed **migrate:fresh --seed**...

morloderex:
Mysql and maybe some other database engines uses [transactions that commits implicitly](https://dev.mysql.com/doc/refman/8.0/en/implicit-commit.html) so i don't think we can support this unless we look at what kind of sql statement is being ran which i believe would be impossible.

Sairahcaz:
MySQL works great 👍 😃

MohammadAlhallaq:
> I guess without this to be the default, we would need a **db:truncate** command, so that you can easily reset the database after a failed **migrate:fresh --seed**...

so if I get it right, when --rof is being set to true by default, the migration is going to get rolled back when an exception occurs by using this code:
`$this->resolver->connection()?->rollBack()`

AndreSchwarzer:
I agree with @MohammadAlhallaq to not introduce something new, which alters the current behavior. 

Furthermore are all flags you can set to the altered commands clear and speaking for them selfs. `--rof` on the other hand is not clear, **to me**, at all; not by looking at it.

You're saying "MySQL works great", but what about all the other, by default, supported database drivers?

ankurk91:
`rof` is not a good name

Sairahcaz:
> `rof` is not a good name

Kinda agree with that, maybe `safe` is better?

taylorotwell:

PR #46362, opened 2023-03-06
[10.x] Named static methods for middleware

timacdonald:
Continuation of https://github.com/laravel/framework/pull/46219. Framework first version of "[HasParameters](https://github.com/timacdonald/has-parameters/)".

This PR introduces a more "typed" API for all the first party middleware - which gives a nicer DX than the "stringy" API, in my opinion.

```php
Route::get('users', UserController::class)
    ->middleware([

        Authenticate::class, // default.
        Authenticate::using('web'), // specify a guard.
        Authenticate::using('web', 'another'), // specify multiple guards.

        AuthenticateWithBasicAuth::class, // default.
        AuthenticateWithBasicAuth::using('web'), // specify a guard.
        AuthenticateWithBasicAuth::using('web', 'field'), // specify field
        AuthenticateWithBasicAuth::using(field: 'field'), // supports named arguments

        Authorize::using('access-nova'), // specify an ability.
        Authorize::using('store', Post::class), // specify an ability with a model.
        Authorize::using('update', 'post', 'comment'), // specify multiple models.

        EnsureEmailIsVerified::class, // default.
        EnsureEmailIsVerified::redirectTo('route.name'), // specify a redirect.

        RequirePassword::class, // default.
        RequirePassword::using('route.name'), // specify a redirect.
        RequirePassword::using('route.name', 100), // specify both.
        RequirePassword::using(passwordTimeoutSeconds: 100), // supports named arguments

        SetCacheHeaders::using('max_age=120;no-transform;s_maxage=60'), // using a string.
        SetCacheHeaders::using([
            'max_age=120',
            'no-transform',
            's_maxage=60',
        ]), // using a list of values.
        SetCacheHeaders::using([
            'max_age' => 120,
            'no-transform',
            's_maxage' => '60',
        ]), // using a hash/list of values.

        ValidateSignature::class, // default (absolute).
        ValidateSignature::relative(), // relative URL.
        ValidateSignature::absolute(), // absolute URL. This is the default, but provided a named method for completeness.

        ThrottleRequests::using('gold-tier'), // named rate limiter
        ThrottleRequests::with(100, 1, 'foo'), // custom with attempts, decay, and prefix
        ThrottleRequests::with(prefix: 'foo'), // supports named arguments.
    ]);
```

If we made this the documented approach, we could potentially remove the middleware aliases from the kernel.

https://github.com/laravel/laravel/blob/5070934fc5fb8bea7a4c8eca44a6b0bd59571be7/app/Http/Kernel.php#L48-L66

### DIFF
--- a/src/Illuminate/Auth/Middleware/RequirePassword.php
+++ b/src/Illuminate/Auth/Middleware/RequirePassword.php
@@ -64,7 +64,7 @@ class RequirePassword
      * @param  \Illuminate\Http\Request  $request
      * @param  \Closure  $next
      * @param  string|null  $redirectToRoute
-     * @param  int|string|null  $passwordTimeoutSeconds
+     * @param  string|int|null  $passwordTimeoutSeconds
      * @return mixed
      */
     public function handle($request, Closure $next, $redirectToRoute = null, $passwordTimeoutSeconds = null)
--- a/src/Illuminate/Auth/Middleware/RequirePassword.php
+++ b/src/Illuminate/Auth/Middleware/RequirePassword.php
@@ -45,12 +45,26 @@ class RequirePassword
     }
 
     /**
+     * Specify the redirect route and timeout for the middleware.
+     *
+     * @param  string|null  $redirectToRoute
+     * @param  string|null  $passwordTimeoutSeconds
+     * @return string
+     *
+     * @named-arguments-supported
+     */
+    public static function using($redirectToRoute = null, $passwordTimeoutSeconds = null)
+    {
+        return static::class.':'.implode(',', func_get_args());
+    }
+
+    /**
      * Handle an incoming request.
      *
      * @param  \Illuminate\Http\Request  $request
      * @param  \Closure  $next
      * @param  string|null  $redirectToRoute
-     * @param  int|null  $passwordTimeoutSeconds
+     * @param  int|string|null  $passwordTimeoutSeconds
      * @return mixed
      */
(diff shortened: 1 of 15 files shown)


src/Illuminate/Auth/Middleware/RequirePassword.php
```diff
@@ -63,7 +77,7 @@ public function handle($request, Closure $next, $redirectToRoute = null, $passwo
             }
 
             return $this->responseFactory->redirectGuest(
-                $this->urlGenerator->route($redirectToRoute ?? 'password.confirm')
+                $this->urlGenerator->route($redirectToRoute ?: 'password.confirm')
```
timacdonald, line 80: `$redirectToRoute` may be `null` if no arguments, but it may also be an empty string when only `passwordTimeoutSeconds` is specified.

```php
"password.confirm:,300"
```

newtonjob:
I feel that `using` potentially works for all cases, and should probably be available for all the middleware classes.

`Authorize::using('admin')` may read nicer than `Authorize::can('admin')`.

Also, I think middleware aliases have their place, especially for simple middleware with few or no parameters.

lazer-hybiz:
I would suggest naming the `using()` method to `withParameters()`.

I think this functionality should either be implemented in a base middleware class which all middleware should extend or a trait like the original package, this will be helpful when creating for custom middleware with the `make:middleware` command to not have to write this functionality manually.

timacdonald:
I've renamed `Authorize::can(...)` to `Authorize::using(...)`.

Mastacow:

PR #46364, opened 2023-03-06
[10.x] Support FIFO type of AWS SQS queue driver

techieforfun:
AWS SQS has two types of queue. [`Standard`](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/standard-queues.html) type is supported by Laravel, however [`FIFO`](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/FIFO-queues.html) type is not, because it needs different set of arguments when pushing jobs otherwise AWS PHP SDK will throw an exception. We and [some others](https://stackoverflow.com/questions/40733957/how-to-accommodate-amazon-fifo-sqs-in-laravel-queue) had this need so I started develop a [package for it](https://github.com/CulturalInfusion/laravel-sqs-fifo) however thought the changes could be merged into Laravel framework itself to be accessible via convenient methods. 

This PR uses new index called `sqsfifo` inside `config/queue.php`.

FIFO type guarantees delivering in order, and this PR implements two main features of it which are `Message Group ID` and `Message Deduplication ID`.

`Message Group ID` is like a queue name but within a queue, means you can have multiple message group id inside a queue. For example there is a queue called `email` and you can push your message into multiple group ids like `sensitive`, `normal` and `low`. However when popping or reading message they just use same worker as queue.

`Message Deduplication ID` is a token used for deduplication of sent messages. If a message with a particular message deduplication ID is sent successfully, any messages sent with the same message deduplication ID are accepted successfully but aren't delivered during the 5-minute deduplication interval.

Also FIFO doesn't support delay per message, just per queue and can be configured via AWS SQS console or API. So there is another small change which is used for `delay` method when dispatching jobs, and that is a new key inside config called `allow_delay` that indicated job can use queue delay or not. If the value is false and delay method get called it throws an exception.

Also FIFO queue needs a `.fifo` posftix in its name when getting create inside AWS, and with this PR there is no need to add this postfix manually to queue name because it will handle it. The reason behind this is that developer can change queue driver easily without worrying about queue name which we've experienced multiple times.

Both of these parameters (`Message Group ID` and `Message Deduplican ID`) can be have default value inside `config/queue.php`:

```php
'sqsfifo' => [
    'driver' => 'sqsfifo',
    'key' => env('AWS_ACCESS_KEY_ID'),
    'secret' => env('AWS_SECRET_ACCESS_KEY'),
    'prefix' => env('SQS_PREFIX', 'https://sqs.us-east-1.amazonaws.com/your-account-id'),
    'queue' => env('SQS_QUEUE', 'default'),
    'suffix' => env('SQS_SUFFIX'),
    'region' => env('AWS_DEFAULT_REGION', 'us-east-1'),
    'message_group_id' => 'sensitive',
    'message_deduplication_id' => 'bf9ce2bf-9461-9960-7f92-61b32abf425e',
    'allow_delay' => 0
    'after_commit' => false,
],
```

and also can be defined per message inside the job class:

```php
<?php

namespace App\Jobs;

use Illuminate\Bus\Queueable;

class SendNotificationJob implements ShouldQueue
{

   /**
     * The message group id.
     *
     * @var string
     */
    public $messageGroupId;

    /**
     * The message deduplication id.
     *
     * @var string
     */
    public $messageDeduplicationId;

    /**
     * Create a new job instance.
     *
     * @return void
     */
    public function __construct($messageGroupId, $messageDeduplicationId)
    {
        $this->messageGroupId = $messageGroupId;
        $this->messageDeduplicationId = $messageDeduplicationId;
    }
```

**Docs**: https://github.com/laravel/docs/pull/8633
**Laravel Config**: https://github.com/laravel/laravel/pull/6137


ankurk91:
Having a separate  SqsQueue class for FIFO would be easy to maintain and debug.

Message group ID Must be configurable from outside of job, for example:

```php
CreateBackupJob::dispatch($server)->withMessageId("backups")
```

For those who are opting the FIFO queue, please read the [FAQ](https://aws.amazon.com/sqs/faqs/)

>By default, FIFO queues support up to 3,000 messages per second with batching or up to 300 messages per second (300 send, receive, or delete operations per second) without batching.

taylorotwell:
Agree, should be a separate class.

techieforfun:
- Added new class for queue and connector
- About the @ankurk91 suggestion on using a `withMessageGroupId` method instead of variable inside job, because it needs to add this specific method of SQS FIFO in `Illuminate\Foundation\Bus\PendingDispatch`, I wasn't sure if it's acceptable by Laravel team.

Mastacow:

taylorotwell:
This community package seems to have a bit more features and completeness: https://github.com/shiftonelabs/laravel-sqs-fifo-queue

Maybe users should just use that?

PR #46370, opened 2023-03-06
[10.x] Adds ModelsServiceProvider

imanghafoori1:
In laravel 10.x version models may have many things configured while booting.
The method call can get scattered in generic service providers and be eventually hard to find.

So we can have a brand new `ModelsServiceProvider.php` in the `app/providers` folder like this:

```php

use Illuminate\Foundation\Support\Providers\ModelsServiceProvider as CoreModelsServiceProvider;

class ModelsServiceProvider extends CoreModelsServiceProvider 
{
    /**
     * @var array<string, string>
     */
    protected $morphMap = [
         'comment' => 'App\Models\Comment',
         ...
    ];

    protected function preventLazyLoading()
    {
        return ! $this->app->isProduction();
    }

    protected function onLazyLoadingViolation(Model $model, string $relation)
    {
          // ...
    }
}

```

and you get the idea.

### DIFF
--- a/src/Illuminate/Foundation/Support/Providers/ModelsServiceProvider.php
+++ b/src/Illuminate/Foundation/Support/Providers/ModelsServiceProvider.php
@@ -1,0 +1,100 @@
+<?php
+
+namespace Illuminate\Foundation\Support\Providers;
+
+use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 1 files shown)


milwad-dev (commented):
If you can say some configure of models, it's better.

taylorotwell:

PR #46373, opened 2023-03-06
[10.x] Adds optional parameter to afterMaking/afterCreating factory methods to support handling complex state

adamthehutt:
When instantiating an Eloquent Factory, the configure() method is called immediately. That causes $this to be bound at the very beginning of the factory's life. 

Factory states are great for tracking data that corresponds neatly to model attributes. But sometimes state is more complex than that, and in those cases, allowing the factory itself to track some additional properties would be very useful. Currently that data is lost, however, due to the early binding of $this. Any properties (other than "state") that are manipulated prior to calling create() or make() are unavailable in the callbacks.

This PR passes the factory object to the "afterMaking" and "afterCreating" callbacks. This allows callbacks to access the factory itself _as it exists when the callbacks are executed_.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/Factory.php
+++ b/src/Illuminate/Database/Eloquent/Factories/Factory.php
@@ -686,7 +686,7 @@ abstract class Factory
     {
         $instances->each(function ($model) {
             $this->afterMaking->each(function ($callback) use ($model) {
-                $callback($model);
+                $callback($model, $this);
             });
         });
     }
@@ -702,7 +702,7 @@ abstract class Factory
     {
         $instances->each(function ($model) use ($parent) {
             $this->afterCreating->each(function ($callback) use ($model, $parent) {
-                $callback($model, $parent);
+                $callback($model, $parent, $this);
             });
         });
     }


taylorotwell:
Can you provide a code example of how this would be used? Thanks!

adamthehutt:
@taylorotwell Sure thing. 

A real use case of mine has to do with accounting transactions. A single financial transaction is represented by several different models (e.g., one for the transaction itself, one for each line of the transaction, one for the customer, N for the tags, etc.) There might be 10 different DB rows inserted when recording a single transaction.

What I'd like to be able to do is something like:

```Sale::factory()->toCustomer($customer)->withItem($item, 2)->create();```

If this was as simple as making "customer_id" an attribute of the sale, then I could easily use a factory state for that. But that doesn't work, because customer_id is actually a property of some (but not all) of the transaction lines, which are themselves separate models from the transaction itself. 

So what I initially tried was something like this:

```php
class SaleFactory 
{
    public $customer;

    public function toCustomer($customer) 
    {
        $this->customer = $customer;
        return $this;
    }

    public function configure()
    {
        return $this->afterMaking(function ($sale) {
            $customer = $this->customer ?? Customer::factory()->create();
            // code that associates the customer with the sale before saving the sale model            
        });
    }
}
```

But because of when configure() is called, $this->customer will never be populated. So instead, I'd like to be able to do:

```php
class SaleFactory 
{
    public $customer;

    public function toCustomer($customer) 
    {
        $this->customer = $customer;
        return $this;
    }

    public function configure()
    {
        return $this->afterMaking(function ($sale, $factory) {
            $customer = $factory->customer ?? Customer::factory()->create();
            // code that associates the customer with the sale before saving the sale model            
        });
    }
}
```

taylorotwell:
Hmm, I'm not sure I want to go down this road. I think if I were in your shoes I would create a separate class (not Laravel related at all) that is responsible for creating these financial transactions for testing / factory purposes. 

Something like:

```php
$transaction = TransactionFactory::createForCustomer($customer);
```

Let that class be responsible for orchestrating everything and making sure everything works correctly.

PR #46374, opened 2023-03-07
Add New Validation Rule AlphaDashChar

altwaireb:
Validate that an attribute contains only characters, dashes, and underscores.

The field under validation must be entirely Unicode  **characters** contained in \p{L}, \p{M}, as well as ASCII dashes (**-**) and ASCII underscores (**_**).

To restrict this validation rule to characters in the ASCII range (a-z and A-Z), you may provide the ascii option to the validation rule:

```php
'username' => 'alpha_dash_char:ascii',
```

Example:

```php
'username' => 'asls-_dlks', // true
'username' => 'aslsdlks', // true
'username' => 'asls dlks', // false
'username' => 'नमस्कार-_, // false
'username' => 'من-هنا', // false
```


taylorotwell:
Hey there - I would suggest making this a custom rule. 👍

Issue #46368, opened 2023-03-06
Model toJSON fails when there has been a previous JSON encode or decode error


manticorp:
- Laravel Version: 10.2.0
- PHP Version: 8.1.11
- Database Driver & Version: N/A

### Description:

When json encoding a model using toJson, if there was a previous json error, the model fails because of how it detects json errors.

Here is the toJson code in ```Illuminate\Database\Eloquent\Model.php```:

```php
    /**
     * Convert the model instance to JSON.
     *
     * @param  int  $options
     * @return string
     *
     * @throws \Illuminate\Database\Eloquent\JsonEncodingException
     */
    public function toJson($options = 0)
    {
        $json = json_encode($this->jsonSerialize(), $options);

        if (json_last_error() !== JSON_ERROR_NONE) {
            throw JsonEncodingException::forModel($this, json_last_error_msg());
        }

        return $json;
    }
```

The problem occurs if, somewhere else in the software, an error was thrown in a json_encode or json_decode call, even if called with the ```@``` operator.

I would suggest changing to something like this:

```php
    public function toJson($options = 0)
    {
        try {
            $json = json_encode($this->jsonSerialize(), $options & JSON_THROW_ON_ERROR);
        } catch (\JsonException $e) {
            throw JsonEncodingException::forModel($this, $e->getMessage());
        }

        return $json;
    }
```

or this:

```php
    public function toJson($options = 0)
    {
        if (json_last_error() !== JSON_ERROR_NONE) {
            json_encode("");
        }

        $json = json_encode($this->jsonSerialize(), $options);

        if (json_last_error() !== JSON_ERROR_NONE) {
            throw JsonEncodingException::forModel($this, json_last_error_msg());
        }

        return $json;
    }
```

### Steps To Reproduce:

Do the following:
```
composer create-project laravel/laravel laravel-json-bug
cd laravel-json-bug
php artisan make:model Test --all
```

Create a test route in ```web.php```:

```php
Route::get('/test', [TestController::class, 'index']);
```

Edit TestController.php:

```php
<?php

namespace App\Http\Controllers;

use App\Models\Test;

class TestController extends Controller
{
    public function index()
    {
        $testModel = new Test();
        $testModel->foo = 'bar';

        @json_decode('[this is invalid json]');

        return view('test', ['test' => $testModel]);
    }
}
```

Now create a view ```test.blade.php```

```php
<!DOCTYPE html>
<html>
    <head>
        <meta charset="utf-8">
[...]

PR #46375, opened 2023-03-07
[10.x] Fix toJson json_encode error handling

manticorp:
Fixes issue 46368

toJson would previously fail if there was any previous json_encode/decode errors, as the json_last_error would survive even through its own json_encode call, if JSON_THROW_ON_ERROR was passed as a flag.

BREAKING CHANGE: this will break workflows that relied on catching a JsonException rather than a JsonEncodingException

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1653,7 +1653,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
         } catch (\JsonException $e) {
             throw JsonEncodingException::forModel($this, $e->getMessage());
         }
-        if (!($options & JSON_THROW_ON_ERROR)) {
+        if (! ($options & JSON_THROW_ON_ERROR)) {
             if (json_last_error() !== JSON_ERROR_NONE) {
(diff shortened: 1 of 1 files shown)


manticorp:
Relevant issue:

https://github.com/laravel/framework/issues/46368

The way ```toJson``` in Eloquent models handles json encoding is in conflict with users that pass ```JSON_THROW_ON_ERROR``` as an option, which include the ```@js``` blade directive.

This PR stops that situation from happening, whilst still handling cases when `JSON_THROW_ON_ERROR` is not passed as an option.

However, it will break workflows that rely on the `JsonException` that used to be thrown when passing ```JSON_THROW_ON_ERROR``` as a parameter.

Perhaps also `JsonEncodingException `should extend from `JsonException`? That would solve the issue, but at the moment it extends `RuntimeException`, and I'm not sure what relies on that!

manticorp:
For googlers using the `@js` directive and receiving `toJson` errors, you will have to replace your usage of `@js` for the moment when encoding eloquent models.

ankurk91:
JSON_THROW_ON_ERROR is always available in php 8.1

taylorotwell:

Thanks!

PR #46377, opened 2023-03-07
[10.x] Add columns params to `firstWhere` method

milwad-dev:
When use `firstWhere`, you can't to select columns, ex:
```php
User::query()->firstWhere('name', 'Dr');
```

But now if you want to select specific columns you can do like this:
```php
User::query()->firstWhere('name', 'Dr', columns: 'name'); // Just (name)
User::query()->firstWhere('name', 'Dr', columns: ['name', 'email']); // (name, email)
User::query()->firstWhere('name', '!=', 'Dr', columns: ['name', 'email']); // (name, email)
```

> **_NOTE:_**  If column selection is not available in this method, the feature of `first` method in `firstWhere` will be lost ⚠️

<hr>

There is no breaking changes.
Added tests successfully ✅

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -306,7 +306,7 @@ class Builder implements BuilderContract
      * @param  mixed  $operator
      * @param  mixed  $value
      * @param  string  $boolean
-     * @param array|string  $columns
+     * @param  array|string  $columns
      * @return \Illuminate\Database\Eloquent\Model|static|null
      */
     public function firstWhere($column, $operator = null, $value = null, $boolean = 'and', $columns = ['*'])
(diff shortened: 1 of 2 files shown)


rodrigopedra:
```php
User::query()->select('name', 'email')->firstWhere('name', 'Dr');
```

rodrigopedra:
> There is no breaking changes.

Adding a parameter to a public method IS a breaking change on itself, for anyone extending that class.

taylorotwell:
I think you can just call `select` before calling `firstWhere`?

milwad-dev:
@taylorotwell @rodrigopedra 
Well, it can be used to make the work easier.
In a method, he can easily make his own query.
```php
User::query()->select('name', 'email')->firstWhere('name', 'Dr');
OR
User::query()->firstWhere('name', 'Dr', columns: ['name', 'email']);
```

milwad-dev:
@driesvints 
Can you give your opinion?

driesvints:
Looks like we're not considering this right now, sorry.

PR #43837, opened 2022-08-24
[9.x] Translation of the default message of the validator

andrey-helldar:
### Before:

```json
{
    "message": "The given data was invalid.",
    "errors": {
        "email": [
            "Taki adres e-mail już występuje."
        ]
    }
}

{
    "message": "Taki adres e-mail już występuje. (and 1 more error)",
    "errors": {
        "email": [
            "Taki adres e-mail już występuje."
        ],
        "name": [
            "Taki nazwa już występuje."
        ]
    }
}
```

### After

```json
{
    "message": "Podane dane były nieprawidłowe.",
    "errors": {
        "email": [
            "Taki adres e-mail już występuje."
        ]
    }
}

{
    "message": "Taki adres e-mail już występuje. (i jeszcze 1 błąd)",
    "errors": {
        "email": [
            "Taki adres e-mail już występuje."
        ],
        "name": [
            "Taki nazwa już występuje."
        ]
    }
}
```

On line 99, I deliberately put the space character outside the translation function, as it can be confusing for developers, namely, they can compose translations as `"(and :count more errors)"` instead of `" (and :count more errors)"` and will not understand why the translation does not work.

PR #46378, opened 2023-03-07
[11.x] Revert "[9.x] Translation of the default message of the validator"

GrahamCampbell:
Reverts laravel/framework#43837. This was a major mistake. Exceptions should not know about translation - it should be handled by the exception rendering layer. Moreover, the validator contract doesn't have a getTranslator method, so this will result in a crash, in general.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Release Notes for 9.x
 
-## [Unreleased](https://github.com/laravel/framework/compare/v9.52.3...9.x)
+## [Unreleased](https://github.com/laravel/framework/compare/v9.52.4...9.x)
+
+
+## [v9.52.4](https://github.com/laravel/framework/compare/v9.52.3...v9.52.4) - 2023-02-22
+
+### Fixed
+- Fixes constructable migrations ([#46223](https://github.com/laravel/framework/pull/46223))
 
 
 ## [v9.52.3](https://github.com/laravel/framework/compare/v9.52.2...v9.52.3) - 2023-02-22
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
 # Release Notes for 9.x
 
-## [Unreleased](https://github.com/laravel/framework/compare/v9.52.2...9.x)
+## [Unreleased](https://github.com/laravel/framework/compare/v9.52.3...9.x)
+
+
+## [v9.52.3](https://github.com/laravel/framework/compare/v9.52.2...v9.52.3) - 2023-02-22
+
+### Reverted
+- Revert changes from `Arr::random()` ([cf3eb90](https://github.com/laravel/framework/commit/cf3eb90a6473444bb7a78d1a3af1e9312a62020d))
 
 
 ## [v9.52.2](https://github.com/laravel/framework/compare/v9.52.1...v9.52.2) - 2023-02-21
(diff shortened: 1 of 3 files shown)


ankurk91:
Reverting a change after 6+ months 😅

rodrigopedra:
It is never late to do the right thing.

I have mixed feelings about this. I fully agree with @GrahamCampbell that localizing an exception message should be done on the exception rendering layer, or any other closer to the user layer, and that calling a method not present in the interface is a mistake.

But in this particular case there should be at least a way to assert the errors count, so one wanting to have a localized message, somehow, can easily assess this information, without defaulting to string parsing, that could break anytime the error message is changed.

Maybe adding the count as a property, and having a getter would be a good middle-ground solution.

taylorotwell:
I agree this change was bad but we can't revert it until the next major release.

GrahamCampbell:
let's re-target at 11.x ;)

driesvints:
@GrahamCampbell can you resolve the conflict?

GrahamCampbell:
The conflict can be resolved only with branch merges 9.x -> 10.x -> master. I'll see if I can do that now.

GrahamCampbell:
Done.

andrey-helldar:
I agree that the decision was suboptimal. But if a project provides translations of validator errors, then it should also provide a translation of this message out of the box without manual intervention from developers.

rodrigopedra:
@andrey-helldar, out of curiosity, when an error, such as `429 Too Many Requests`, happens, which a translation package could consider handling, as such error can occur during validation, and is a client error, not a server error, how are you handling its error's message translation?

The default error message on this exception is like a top-level error, and should be handled in a generic way.

Also, you can test for this exception instance with the `instanceof` operator and provide a proper translation.

The only piece I think is missing is a proper way to retrieve the exception count, without needing to parse the exception, as it is an added feature to the exception instance.

But this can be added in an additional PR.

rodrigopedra:
Or even (I was just looking into the exception's code)

~~~php
$count = count($messages = $exception->errors());

\dd($messages, $count);
~~~

https://github.com/laravel/framework/blob/10ba03e5541cceb66acf861ea56d7ce7a18c5078/src/Illuminate/Validation/ValidationException.php#L110-L113

andrey-helldar:
@rodrigopedra, if error 429 is detected, the message is returned by Ngink. So Laravel basically can't handle it. But if this error is returned by the result of calling the HTTP facade, then I simply implement translation when catching errors.

rodrigopedra:
@andrey-helldar , in the same measure, for `422` errors on API requests, the error message is also handled by nginx: `422 Unprocessable Content`, isn't it?

When giving the `429` example, I thought you were worried about the JSON payload, or any other text content returned in the response's body.

But if:

> by the result of calling the HTTP facade, then I simply implement translation when catching errors.

And you seem, by the response you gave, to ignore the `message` key on the JSON payload, and prefer to implement translation when "catching errors", then I don't see where the problem is.

For completeness, this is the output, through nginx, for a `429` error, and for a `422` error.



---



For 422 errors, you'll get all translated errors on the `errors` key, as you seem to ignore the `message` key, based on your `429` example, and prefer to handle error responses by their status code, I guess the current behavior suffices.

andrey-helldar:
@rodrigopedra, Not certainly in that way.  Code 422 is returned by Laravel when validating incoming data, while code 429 is returned by nginx after several redirects.  On your screenshots, you yourself return code 429 using the abort function.  This has nothing to do with "Too Many Redirects" being caught by the server.

driesvints:
Reverting this one: https://github.com/laravel/framework/pull/50546

PR #46381, opened 2023-03-07
Update ConsoleApplicationTest.php

amirsadeghi1:
i removed the $this->equalTo() check in the addToParent expectation,
as it is not necessary since we are already passing the exact same object instance to the method
and added a separate variable $application to store the expected argument for the setLaravel method
this makes the code more readable and easier to change if necessary.
also changed the assertion method from assertEquals to assertSame
as we want to ensure that the same object instance is returned from the add method


driesvints:
The test currently passes so this is fine, thanks.

Issue #46322, opened 2023-03-02
JsonResource wrapping where toArray contains same wrap key results in unexpected JSON output


require-dev:
- Laravel Version: 9.52.4
- PHP Version: 8.1.15
- Database Driver & Version: MySQL 8.0.32

### Description:
When working with JsonResource and creating it as a singular resource with `TestResource::make()` re-using the `$wrap` key inside the `toArray()` method, results in the outputted JSON not being wrapped anymore. 

### Steps To Reproduce:
```php
// in TestResource
namespace App\Http\Resources;

use Illuminate\Http\Resources\Json\JsonResource;

class TestResource extends JsonResource
{
    public static $wrap = 'test';

    public function toArray($request)
    {
        return [
            'id' => 'X',
            'test' => 'test'
        ];
    }
}

// in api.php
Route::get('/test', function () {
     return new TestResource([]);
});
```

I expect the endpoint `/test` to return the following JSON:
```json
{
    "test": {
        "id": "X",
        "test": "test"
    }
}
```

But get the following back:
```json
{
    "id": "X",
    "test": "test"
}
```

taylorotwell:
PRs welcome.

iazaran:
The PR will fix the issue in version 10. If that solution works fine, then we can apply it on version 9 too.

require-dev:
https://github.com/laravel/framework/pull/46383#issuecomment-1460343071

@taylorotwell, Although I do understand (and agree) that re-using the wrapper key inside a wrapped resource is not ideal, the behaviour that is described above, shouldn't happen or should be documented?

PR #46383, opened 2023-03-07
[10.x] Check only for first key in JSON response

iazaran:
Based on issue https://github.com/laravel/framework/issues/46322
The developer can set a wrapper for response and we don't need to check if that wrapper exists on the responses keys. Only check for first key is enough.
So developer can make something like this simply:
```
{
    "test": {
        "id": "X",
        "test": "test"
    }
}
```


taylorotwell:
I kinda think you should just not reuse your wrapper key in your data if possible. It would be pretty confusing for API consumers.

iazaran:
@taylorotwell ,
I got it.
Thanks

PR #46390, opened 2023-03-07
[10.x] use the same criteria StartSession uses on previous URL

### LINKED ISSUE #46388: return back() is broken in laravel 10.3 version
- Laravel Version: 10.3.0
- PHP Version: 8.2.3 with FPM apache
- Database Driver & Version: Ver 15.1 Distrib 10.2.44-MariaDB, for Linux (x86_64) using readline 5.1
- OS: CentOS 7 with Plesk

### Description:
In my controllers, return back() redirects to homepage instead of previous page. This bug only produced on server side for laravel 10.3 version. When I downgrade to laravel 10.2, there are no any mis redirect.

### Steps To Reproduce:
I change laravel/framework version to 10.2 from composer.json. There are no anything unusual.
Then, I change the version to 10.3 or ^10.0.
return back() function redirects to homepage.

But It only happens in my production server which is CentOS 7 with plesk.

rodrigopedra:
Closes https://github.com/laravel/framework/issues/46388

PR https://github.com/laravel/framework/pull/46234 introduced a change on how the previous URL is calculated when the previous URL matches the current URL.

Unfortunately, it breaks redirecting back to the same URL, which is a common pattern on PATCH/PUT, as the previous implementation didn't consider which request method is being used.

This PR

- Changes the previous URL calculation to use the same criteria `StartSession` middleware uses to check if it should update the previous URL in session

### DIFF
--- a/src/Illuminate/Routing/UrlGenerator.php
+++ b/src/Illuminate/Routing/UrlGenerator.php
@@ -158,7 +158,15 @@ class UrlGenerator implements UrlGeneratorContract
 
         $url = $referrer ? $this->to($referrer) : $this->getPreviousUrlFromSession();
 
-        if ($url && rtrim($url, '/') !== $this->request->fullUrl()) {
+        $isSame = $url &&
+            rtrim($url, '/') === $this->request->fullUrl() &&
+            $this->request->isMethod('GET') &&
+            $this->request->route() instanceof Route &&
+            ! $this->request->ajax() &&
+            ! $this->request->prefetch() &&
+            ! $this->request->isPrecognitive();
+
+        if ($url && !$isSame) {
             return $url;
         } elseif ($fallback) {
             return $this->to($fallback);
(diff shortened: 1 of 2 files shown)


rodrigopedra:
I guess we could extract this logic to the request object. I wanted to propose a quick fix for the introduced error

rodrigopedra:
Not sure why just one set of tests, failed, from the running time it seems it got stuck somewhere and ended up failing.

StyleCI cloudflare human verification keeps me on a verification loop, so I couldn't see what is wrong.

taylorotwell:
I had to revert the other PR and tag a patch, so I'm guessing this PR isn't complete anymore.

taylorotwell:
I'm also hesitant to keep chasing down this road.

rodrigopedra:
@taylorotwell sorry for the inconvenience caused. This PR was more like a hot-fix attempt.

When I saw the related issue, I just wondered to share what I use, which doesn't break as I only use it at blade templates. I didn't think on the broad usage.

I guess it is better to close this by now, and then propose a more thoughtful solution, considering those other caveats. Perhaps closer to 11.x.

PR #46391, opened 2023-03-07
fix: Eloquent model toJson fix for existing errors.

manticorp:
# The Problem

Related to this issue:

https://github.com/laravel/framework/issues/46368

When casting an eloquent model to JSON using the ```toJson``` method and passing in the `JSON_THROW_ON_ERROR` option, if there was a previous, unrelated JSON error, an erroneous exception would be thrown.

The problem related to this line here:

```php
public function toJson($options = 0)
{
    $json = json_encode($this->jsonSerialize(), $options);

    // This line will always trigger if JSON_THROW_ON_ERROR is passed
    // as an option, and an existing json_encode or json_decode error exists
    if (json_last_error() !== JSON_ERROR_NONE) { 
        throw JsonEncodingException::forModel($this, json_last_error_msg());
    }

    return $json;
}
```

The situation transpires when someone has a previous JSON error, but passes the JSON_THROW_ON_ERROR flag to the toJson function:

```php
// ... earlier on

$array = @json_decode('{some badly formed, but non essesntial, JSON'});

// now json_last_error() is not JSON_ERROR_NONE
// ... and then later in the code ...

$widget = new Widget(); // An Eloquent Model
$widget->name = 'bob';

try {
    $json = $widget->toJson(JSON_THROW_ON_ERROR);
} catch (JsonEncodingException $e) {
    var_dump($e->getMessage()); // Syntax error
} catch (\JsonException $e) {
}
```

This happened to me in developing using the `@js` directive from Livewire, which uses the ```JSON_THROW_ON_ERROR``` flag as a required option.

I think because PHP supports both ways of handling JSON errors, it should be handled both ways by Laravel as well.

## Why fix this?

A good example is the `@js` directive in Livewire. This will *always* fail if there has been *any* json error without the use of `JSON_THROW_ON_ERROR`.

Considering that handling json errors in both ways is still supported, Laravel should support this as well.

## Proposed solution

The solution is fairly simple - we see if `JSON_THROW_ON_ERROR` was passed as an option, and only throw on error in that case:

```php
public function toJson($options = 0)
{
    $json = json_encode($this->jsonSerialize(), $options);
    if (! ($options & JSON_THROW_ON_ERROR)) {
        if (json_last_error() !== JSON_ERROR_NONE) {
            throw JsonEncodingException::forModel($this, json_last_error_msg());
        }
    }

    return $json;
}
```

## Tests

Included in this PR are 7 tests appended to `DatabaseEloquentModelTest.php` to test various toJson parameters and outputs, including exception cases.

Integration tests are hard to write, because this is such a core feature.

## Related

This contrived example demonstrates the problem well:

```php
$object = @json_decode('[invalid json]');

try {
    $json = json_encode("\xB1\x31", JSON_THROW_ON_ERROR); // an invalid UTF-8 sequence
} catch (JsonException $e) {
    var_dump($e->getCode()); // 5
    var_dump($e->getMessage()); // Malformed UTF-8 characters, possibly incorrectly encoded 
}

var_dump(json_last_error()); // 4
var_dump(json_last_error_msg()); // Syntax error
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1648,10 +1648,15 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
      */
     public function toJson($options = 0)
     {
-        $json = json_encode($this->jsonSerialize(), $options);
-
-        if (json_last_error() !== JSON_ERROR_NONE) {
-            throw JsonEncodingException::forModel($this, json_last_error_msg());
+        try {
+            $json = json_encode($this->jsonSerialize(), $options);
+        } catch (\JsonException $e) {
+            throw JsonEncodingException::forModel($this, $e->getMessage());
+        }
+        if (! ($options & JSON_THROW_ON_ERROR)) {
+            if (json_last_error() !== JSON_ERROR_NONE) {
+                throw JsonEncodingException::forModel($this, json_last_error_msg());
+            }
         }
 
         return $json;
--- a/src/Illuminate/Database/Eloquent/Model.php
+++ b/src/Illuminate/Database/Eloquent/Model.php
@@ -1648,11 +1648,7 @@ abstract class Model implements Arrayable, ArrayAccess, CanBeEscapedWhenCastToSt
      */
     public function toJson($options = 0)
     {
-        try {
-            $json = json_encode($this->jsonSerialize(), $options);
-        } catch (\JsonException $e) {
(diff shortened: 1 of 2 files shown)


manticorp:
Just to note, this is no longer a breaking change (unlike my previous PR).

For future versions, though, I would consider getting rid of `JsonEncodingException` and just replacing it with JsonException.

taylorotwell:
No plans to change this right now.

manticorp:
This is a shame - it renders the @js directive in livewire virtually useless!

PR #46403, opened 2023-03-08
[10.x] Alternative database port in Postgres DSN

timvango:
PR #43542 has added support for alternative database names in the DSN for PostgreSQL.
This PR did not add support for an alternative port in the DSN, making it incompatible with the PgBouncer implementation of DigitalOcean (and most probably others).

This PR allows setting a different port that is used only for generating the PostgresConnector DSN, with the connect_via_port option. The port that should be used for the connect_via_database option should be specified here.

### DIFF
--- a/src/Illuminate/Database/Connectors/PostgresConnector.php
+++ b/src/Illuminate/Database/Connectors/PostgresConnector.php
@@ -167,13 +167,14 @@ class PostgresConnector extends Connector implements ConnectorInterface
         // name than the database used for "information_schema" queries. This is
         // typically the case if using "pgbouncer" type software when pooling.
         $database = $connect_via_database ?? $database;
+        $port = $connect_via_port ?? $port ?? null;
 
         $dsn = "pgsql:{$host}dbname='{$database}'";
 
         // If a port was specified, we will add it to this Postgres DSN connections
         // format. Once we have done that we are ready to return this connection
         // string back out for usage, as this has been fully constructed here.
-        if (isset($config['port'])) {
+        if ($port !== null) {
             $dsn .= ";port={$port}";
         }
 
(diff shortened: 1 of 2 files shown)


keithbrink:
@timvango  Lifesaver, thanks!

startork:
Hi @timvango,
Would you be able to clarify why this is necessary? I currently am using a Digital Ocean Connection Pool without this config but worried something's going to crop up :)
The database config option is used for other things such as setting table_catalog in migration queries, hence the need for the connection_via_database option. However, I can't imagine the port is used for much else other than the dsn. So would setting the port to the Connection Pool port not achieve the same result?

timvango:
Hey @startork,

I'm not entirely sure about the exact problem I ran into, but from what I can remember, the migrations were failing when I used a pooled connection, while they worked fine with a direct connection. Laravel already has the 'connection_via_database' option for situations like this, but DigitalOcean's PgBouncer implementation has a quirk: it uses different ports for pooled and direct connections. This causes an issue because trying to use the 'connection_via_database' option results in a connection rejection due to an invalid port for either the pooled or direct connection.

To work around this, I created a patch where I changed the connection port, and it seemed to fix the issue. I even used this patch in a production environment for a while, and it resolved the problems for me. Thinking that others might run into a similar situation, I decided to create this pull request. But it's possible that this issue might not be affecting many people. If everything is running smoothly on your end, then there's probably nothing to worry about.

startork:
Hi @timvango,

When using connect_via_database I do understand that you sometimes have to change the port, I am using DigitalOcean and had to do the same thing. 
However, this change suggests that laravel will sometimes use the `port` value and sometimes the `connect_via_port` value but I don't think that is ever the case. 
In practice if `connect_via_port` is set then Laravel will only ever use that value, `port` becomes obsolete. I do think having some clarity around this is helpful and maybe some dev's will prefer to have a `connect_via_port` value to make it very obvious that that port is specific to pgBouncer; but, in the same vein, it should be clear that setting `connect_via_port` has exactly the same affect as changing the `port` value itself.

PR #46405, opened 2023-03-09
[10.x] feat add `whereLike()` and `whereLikeContain()`

Haythamasalama:
add methods that handle queries for performing 'where like' searches in the database.

* `whereLike()`, `whereNotLike()` , `orWhereLike()`
* `whereLikeContain()` , `whereNotLikeContain()` , `orWhereLikeContain()`

```php
 // select * from `users` where `name` like laravel;
  
 User::whereLike('name', 'laravel')->get();
  
// select * from `users` where `name` like "%laravel%";

 User::whereLikeContain('name', 'laravel')->get();
```

```php  

 // select * from `users` where `name` like laravel or `name` like forge;
 User::whereLike('name', 'laravel')->orWhereLike('name','forge')->get();
 
 // select * from `users` where `name` like "%laravel%" or `name` like "%forge%";
 User::whereLikeContain('name', 'laravel')->orWhereLikeContain('name','forge')->get();
 
 // select * from `users` where `name` not like laravel and `name` not like "%forge%";
 User::whereNotLike('name', 'laravel')->whereNotLikeContain('name','forge')->get();
```

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1252,7 +1252,7 @@ class Builder implements BuilderContract
     {
         $type = 'like';
 
-        $this->wheres[] = compact('type', 'value', 'column', 'boolean','not');
+        $this->wheres[] = compact('type', 'value', 'column', 'boolean', 'not');
 
         $this->addBinding($value, 'where');
 
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1240,6 +1240,92 @@ class Builder implements BuilderContract
     }
 
     /**
+     * Add a where like clause to the query.
+     *
+     * @param  \Illuminate\Contracts\Database\Query\Expression|string  $column
+     * @param  string|null|int  $value
+     * @param  string  $boolean
+     * @param  bool  $not
+     * @return $this
+     */
+    public function whereLike($column, $value, $boolean = 'and', $not = false)
+    {
+        $type = 'like';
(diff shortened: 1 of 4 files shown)


ankurk91:
This has been attempted many times. 
We should stop adding more alias methods to Builder.

Btw it does not cover all use cases.

How to produce this query

```sql
select * from stores where name like "apple%"

select * from stores where name like "a%e"
```

There are many others cases

Haythamasalama:
> This has been attempted many times. We should stop adding more alias methods to Builder.
> 
> Btw it does not cover all use cases.
> 
> How to produce this query
> 
> ```sql
> select * from stores where name like "apple%"
> 
> select * from stores where name like "a%e"
> ```
> 
> There are many others cases

You Can Use like this or other cover all use cases
```php
Stores::whereLike('name',"apple%")->get();
Stores::whereLike('name',"a%e")->get();

```

Haythamasalama:
> Thank you. Please try to search PR's before submitting. @driesvints 

🧐 Hmm, I think it's useful, just like other query builders such as whereBetween() ...etc. and `whereLike()` it supports many other cases in the where clause.

newtonjob:
I'm surprised this is closed without any practical reason.

LIKE is a very common SQL operator and probably the only one that doesn't have a helper method yet. 
Sometimes I find myself reaching for it by default, only to realise such a method doesn't exist.

We have helper methods for several operators like IN, NULL, BETWEEN, DATE, FULLTEXT, etc.

Why is LIKE excluded? Are there any downsides?

PR #46412, opened 2023-03-09
[10.x] Make Route isControllerAction method public

jnoordsij:
Follow-up to https://github.com/laravel/framework/discussions/46332.

Currently `isControllerAction` is a `protected` method on the `\Illuminate\Routing\Route` class. This makes distinguishing between controller- and closure-based routes difficult and slightly cumbersome, as can be seen here:
https://github.com/laravel/framework/blob/38342ab7d7a4e52fe78109d37a145d8309641da1/src/Illuminate/Routing/RouteCollection.php#L86-L93

This PR makes this method public, to allow for more consistent checks on controller- vs. closure-based routes, both within the framework and in all applications using it.

### DIFF
--- a/src/Illuminate/Routing/Route.php
+++ b/src/Illuminate/Routing/Route.php
@@ -216,7 +216,7 @@ class Route
      *
      * @return bool
      */
-    protected function isControllerAction()
+    public function isControllerAction()
     {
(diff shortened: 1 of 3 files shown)


jnoordsij:
I'm gonna lay this off for now, appearently my assumptions about the `isControllerAction` method were a bit too optimistic.

PR #46418, opened 2023-03-09
[9.x] Add assertUnsupportedMediaType method for assertion Unsupported…

shamimulalam:
… Media Type


taylorotwell:
Wrong branch. Should be 10.x.

bert-w:
I swear I read somewhere that no more of these functions were going to be added.

PR #46422, opened 2023-03-10
Fix Str::isJson('123') issues

rshkabko:
When you call Str::isJson('123') it returns true, but 123 isn't JSON. Fix this issues

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -428,7 +428,7 @@ class Str
      */
     public static function isJson($value)
     {
-        if (! is_string($value)) {
+        if (! is_string($value) || is_numeric($value)) {
             return false;
         }
 


crynobone:
> Early versions of JSON (such as specified by RFC 4627) required that a valid JSON text must consist of only an object or an array type, which could contain other types within them. This restriction was dropped in RFC 7158, where a JSON text was redefined as any serialized value.

From the same Wikipedia page.

rshkabko:
If you're right, anything can be JSON, so why check at all then? And why 123 - not JSON, but the same value as string ('123') is json?))

X-Coder264:
> In a nutshell, this is not quite correct:
> 
> ```
> use Illuminate\Support\Str;
> dump(Str::isJson(123)); // false
> 
> dump(Str::isJson('123')); // true
> ```

This is correct. JSON can be only of the string type, raw integers are not JSON. The method would ideally have a string type declaration instead of doing a `is_string` check, but it was written a long time ago when PHP 7 wasn't a thing yet.

Once the min PHP version that Laravel supports become PHP 8.3 we'll be able to use the `json_validate` here instead and that method only accepts `string` data as the input -> https://wiki.php.net/rfc/json_validate

donnysim:
I think there's confusion here between `123` and string `'123'`, a string `'123'` is a valid json where a number `123` is not, which is why the original behavior works as intended.

PR #46426, opened 2023-03-10
[10.x] Add assertUnsupportedMediaType method for assertion Unsupported Media Type

shamimulalam:
Same [PR](https://github.com/laravel/framework/pull/46418#event-8714116179) for 10.x


GrahamCampbell:
You don't need to send PRs to 10.x. Anything that lands in 9.x will already end up in 10.x and 11.x.

GrahamCampbell:
I see this was reverted in 9.x. 10.x is the right branch. ;)

PR #46427, opened 2023-03-10
[10.x] fix namespace

macbookandrew:
Fixes the namespace for fake batches; see [the `use` statement](https://github.com/laravel/framework/blob/8c3a23d466debf20ff601eb373f9e60a9ed1fe86/src/Illuminate/Bus/Batchable.php#L8) for reference


driesvints:
I see you sent this in to 9.x so gonna close this one. We'll merge it into 10.x. Thanks

PR #46430, opened 2023-03-11
[10.x] Fix parsed input arguments for command events

bert-w:
This PR fixes the input arguments for `CommandStarting` and `CommandFinished` events, which were previously empty because the `InputDefinition` (Symfony) was bound later than the events.

The following will now work:
```php
Event::listen(function (CommandStarting $event) {
    Log::info('starting ' . $event->command);
    Log::info($event->input->getOptions()); // Previously `[]`, now filled.
    Log::info($event->input->getArguments()); // Previously `[]`, now filled.
});
```

This is an improvement for https://github.com/laravel/framework/pull/44662 which was later reverted (https://github.com/laravel/framework/pull/44888) because it did not function well when an Artisan command was run through the command line.

This case has now been covered (based on `\tests\Integration\Console\CommandSchedulingTest.php` which uses an actual `artisan` file to simulate such call) and also the in-process artisan calls using array  (`Artisan::call('cmd', ['arg'=> 'abc', '--def' => 123])`) and string (`Artisan::call('cmd abc --def=123')`) syntax have been covered.

### DIFF
--- a/src/Illuminate/Console/Application.php
+++ b/src/Illuminate/Console/Application.php
@@ -13,6 +13,7 @@ use Illuminate\Support\ProcessUtils;
 use Symfony\Component\Console\Application as SymfonyApplication;
 use Symfony\Component\Console\Command\Command as SymfonyCommand;
 use Symfony\Component\Console\Exception\CommandNotFoundException;
+use Symfony\Component\Console\Exception\ExceptionInterface;
 use Symfony\Component\Console\Input\ArgvInput;
 use Symfony\Component\Console\Input\ArrayInput;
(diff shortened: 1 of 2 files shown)


taylorotwell:
I'm not pursuing this again on a patch release.

bert-w:
ok any chance for master branch?

taylorotwell:
@bert-w I'll reopen it here for discussion, but this just seems like a big change? Overwriting a whole new method, etc.

bert-w:
This gives another solution to the problem which I think I like more: https://github.com/laravel/framework/pull/46442

PR #46439, opened 2023-03-12
Add noRetriesLeft method in Cache RateLimiter

senaranya:
This method can be used in place of the condition `Ratelimiter::retriesLeft($key, $maxAttempts) === 0`, making the following snippet easier to read:
```php
if (RateLimiter::noRetriesLeft($key, $maxAttempts)) {
    // Do something
}
```


taylorotwell:
How is this practically different then `RateLimiter::tooManyAttempts($key, $maxAttempts)` which also returns a boolean?

PR #46443, opened 2023-03-12
[10.x] Allow for converting a HasMany to HasOne && MorphMany to MorphOne

cosmastech:
### Problem
There are often times where I am building a HasMany relationship and I need to get just one item from the collection. This means I have to either define the relationship two times (one for HasMany and then one for HasOne->ofMany()).

```php
class User extends Model
{
    public function logins(): HasMany {
        return $this->hasMany(Login::class, 'some_id', 'the_other_id');
    }
    public function latestLogin(): HasOne {
        return $this->hasOne(Login::class, 'some_id', 'the_other_id')->latestOfMany();
    }
}
```

As you can imagine, if the design of the DB changes (like during early development of a project) or if the relationship isn't simple, this can be cumbersome or lead to duplication of work.
### Proposed Solution
What is added in this PR is the ability to convert a HasMany relationship to HasOne (and MorphMany to MorphOne). So the above code can become

```php
class User extends Model
{
    public function logins(): HasMany {
        return $this->hasMany(Login::class, 'some_id', 'the_other_id');
    }
    public function latestLogin(): HasOne {
        return $this->logins()->hasOne()->latestOfMany();
    }
}
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/HasMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/HasMany.php
@@ -7,6 +7,16 @@ use Illuminate\Database\Eloquent\Collection;
 class HasMany extends HasOneOrMany
 {
     /**
+     * Convert the relationship to a "has one" relationship.
+     *
+     * @return Illuminate\Database\Eloquent\Relations\HasOne
+     */
+    public function hasOne()
+    {
+        return new HasOne($this->getQuery(), $this->parent, $this->foreignKey, $this->localKey);
+    }
+
+    /**
      * Get the results of the relationship.
      *
      * @return mixed
@@ -45,15 +55,5 @@ class HasMany extends HasOneOrMany
     public function match(array $models, Collection $results, $relation)
     {
         return $this->matchMany($models, $results, $relation);
-    }
-
-    /**
-     * Convert the relationship to HasOne.
-     *
-     * @return HasOne
-     */
-    public function hasOne(): HasOne
-    {
-        return new HasOne($this->getQuery(), $this->parent, $this->foreignKey, $this->localKey);
     }
 }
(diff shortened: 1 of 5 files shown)


cosmastech:
If this is something that makes sense, will gladly add HasManyThrough@hasOneThrough() method as well

cosmastech:
@taylorotwell would you like me to create another PR for `HasManyThrough@one()`?

siarheipashkevich:
@cosmastech could you please check which sql query will be generated? I saw that the query has duplicated conditions...

cosmastech:
> @cosmastech could you please check which sql query will be generated? I saw that the sql query has duplicated conditions...

@siarheipashkevich could you provide a code sample the generated SQL?

siarheipashkevich:
I've terminal model with next relations:
image

In Shift model I've opened scope:
image

Get openedShift from terminal:
image

Logs:
```
[2023-03-24 11:41:38] local.INFO: select * from `shifts` where `shifts`.`terminal_id` = ? and `shifts`.`terminal_id` is not null and `shifts`.`terminal_id` = ? and `shifts`.`terminal_id` is not null and `closed_at` is null and `shifts`.`deleted_at` is null limit 1 [480,480] 
[2023-03-24 11:41:38] local.INFO: see logs above  
```

siarheipashkevich:
> > @cosmastech could you please check which sql query will be generated? I saw that the sql query has duplicated conditions...
> 
> @siarheipashkevich could you provide a code sample the generated SQL?
> 
> edit: I see now.
> 
> ```
> $user->userPreferences()->one()->toSql();
> // select * from `user_preferences` where `user_preferences`.`user_id` = ? and `user_preferences`.`user_id` is not null and `user_preferences`.`user_id` = ? and `user_preferences`.`user_id` is not null
> ```

PLease take a look at your sql. You also have diplicated conditions

```
`user_preferences`.`user_id` = ? and `user_preferences`.`user_id` is not null and `user_preferences`.`user_id` = ? and `user_preferences`.`user_id` is not null
```

cosmastech:
@siarheipashkevich thank you for reporting this! https://github.com/laravel/framework/pull/46575 should resolve this issue

PR #46448, opened 2023-03-13
[10.x] Fix route:cache from excluding octane routes

CharlesBilbo:
When running route:cache command with octane routes they are being excluded from the cached routes file. This causes octane routes to 404 when cached.

While caching the routes gives a negative performance to swoole (not sure about road runner) by increasing the amount of memory needed for each request this will at least prevent them prevent them from 404ing.

This will also prevent questions like this

https://github.com/laravel/octane/issues/596
https://github.com/laravel/octane/issues/658

**This PR is a required dependency for this PR.** 

https://github.com/laravel/octane/pull/659

Additonally it doesnt seem like Octane::route method doesn't work with road runner.

### DIFF
--- a/src/Illuminate/Foundation/Console/RouteCacheCommand.php
+++ b/src/Illuminate/Foundation/Console/RouteCacheCommand.php
@@ -86,12 +86,12 @@ class RouteCacheCommand extends Command
             $routes->refreshActionLookups();
         });
 
-        if($application->has('octane')) {
+        if( $application->has('octane')) {
             $warning = 'Running route:cache while running swoole decreases memory performance';
             $this->components->warn($warning);
             collect($application->octane->getRoutes())
-                ->each( function ($callback, $signature) use ($application) {
-                    list($method, $uri) = explode('/',$signature, 2);
+                ->each(function ($callback, $signature) use ($application) {
+                    [$method, $uri] = explode('/', $signature, 2);
                     $application['router']->addRoute($method, $uri, $callback);
                 });
         }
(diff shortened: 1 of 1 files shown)


src/Illuminate/Foundation/Console/RouteCacheCommand.php
```diff
@@ -78,10 +79,33 @@ public function handle()
...
+        return $routes;
+    }
+
+    /**
+     * @param Application $application
+     * @return bool
+     */
+    protected function hasOctane(Application $application): bool
+    {
+        return $application?->octane !== null;
```
cosmastech, line 108: Isn't this adding null-safety to $application (which cannot be null or the interpreter will raise a TypeError) and not to the Application@$octane property?

CharlesBilbo: yeah it did i removed the method and changed it for the has method on the application context.

CharlesBilbo:
i didnt add a test cause im not quite sure how yall would want this tested i could install octane since its technically a dependency to this PR but i wasnt quite sure on that. LMK and ill write the test if required.

src/Illuminate/Foundation/Console/RouteCacheCommand.php
```diff
@@ -78,10 +79,24 @@ public function handle()
...
+        $application = $this->getFreshApplication();
+
+        $routes = tap($application['router']->getRoutes(), function ($routes) {
             $routes->refreshNameLookups();
             $routes->refreshActionLookups();
         });
+
+        if ($application->has('octane')) {
+            $warning = 'Running route:cache while running swoole decreases memory performance';
+            $this->components->warn($warning);
```
cosmastech, line 91: You could put this all onto one line, I'm not sure there's a good reason to store the warning as a string first.

taylorotwell:
Octane::route is not really a documented stable feature of Octane at the current moment.

PR #46271, opened 2023-02-25
[10.x] Adding Pipeline Facade

byt3sage:
Hey All!
I have a feeling there is a distinct reasoning behind why this hasn't been done before but here it goes. I love pipelines... I also love facades... Why not merge the two and have a pipeline facade?

This is the simplest implementation I could think of. It almost feels too simple so I'm sure there's some form of logic I have missed.

I'd love some feedback to try get this pushed into Laravel as a core facade!

Happy for this to be closed if it's a non-starter!

P.S - I'll be doing some light tweaking of bits, just thought I'd push this up first so I don't waste my time if it's never going to be accepted :)

PR #46449, opened 2023-03-13
[10.x] Catch pipeline exception

talelmishali:
This PR adds a new `catch` method when using a Pipeline object. Following the addition of Pipeline Facade #46271, I think it would be helpful to chain a catch callback directly on the pipeline object.

With this PR, the following would be possible:

```php
            Pipeline::send($this->organization)
                ->via('handle')
                ->through([
                    VerifyOnboardingIsComplete::class,
                    VerifyOrganizationHasValidPaymentMethod::class,
                    ...
                ])
                ->catch(function (VerificationFailed $th) {
                    NewOrganizationHasNotBeenAuthorized::dispatch($this->organization, $th);
                })
                ->then(function (Organization $organization) {
                    NewOrganizationHasBeenAuthorized::dispatch($organization);
                });
```

### DIFF
--- a/src/Illuminate/Contracts/Pipeline/Pipeline.php
+++ b/src/Illuminate/Contracts/Pipeline/Pipeline.php
@@ -41,7 +41,7 @@ interface Pipeline
     /**
      * Set the catch callback for the pipeline.
      *
-     * @param  \Closure $callback
+     * @param  \Closure  $callback
      * @return $this
      */
     public function catch(Closure $callback);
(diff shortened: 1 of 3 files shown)


taylorotwell:
We can't add a method to an interface on a patch release.

talelmishali:
> We can't add a method to an interface on a patch release.

Make sense. Should I target `master`, or are you not interested anyway?

PR #46451, opened 2023-03-13
[11.x] Catch pipeline exception

talelmishali:
This PR adds a new `catch` method when using a Pipeline object. Following the addition of Pipeline Facade #46271, I think it would be helpful to chain a catch callback directly on the pipeline object.

With this PR, the following would be possible:

```php
            Pipeline::send($this->organization)
                ->via('handle')
                ->through([
                    VerifyOnboardingIsComplete::class,
                    VerifyOrganizationHasValidPaymentMethod::class,
                    ...
                ])
                ->catch(function (VerificationFailed $th) {
                    NewOrganizationHasNotBeenAuthorized::dispatch($this->organization, $th);
                })
                ->then(function (Organization $organization) {
                    NewOrganizationHasBeenAuthorized::dispatch($organization);
                });
```

### DIFF
--- a/src/Illuminate/Pipeline/Pipeline.php
+++ b/src/Illuminate/Pipeline/Pipeline.php
@@ -128,13 +128,13 @@ class Pipeline implements PipelineContract
             }
         }
 
-       return $pipeline($this->passable);
+        return $pipeline($this->passable);
     }
 
     /**
      * Set the catch callback for the pipeline.
      * 
-     * @param \Closure $callback
+     * @param  \Closure  $callback
      * @return $this
      */
     public function catch(Closure $callback)
--- a/src/Illuminate/Pipeline/Pipeline.php
+++ b/src/Illuminate/Pipeline/Pipeline.php
@@ -39,6 +39,13 @@ class Pipeline implements PipelineContract
     protected $method = 'handle';
 
     /**
+     * The closures to be called on exceptions.
+     *
+     * @var Closure
+     */
+    protected $catchCallback;
+
+    /**
      * Create a new class instance.
      *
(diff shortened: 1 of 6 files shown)


src/Illuminate/Pipeline/Pipeline.php
```diff
@@ -113,9 +120,30 @@ public function then(Closure $destination)
             array_reverse($this->pipes()), $this->carry(), $this->prepareDestination($destination)
         );
 
+        if ($this->catchCallback) {
+            try {
+                return $pipeline($this->passable);
+            } catch (\Throwable $th) {
+                return call_user_func($this->catchCallback, $th, $this->passable);
```
rodrigopedra, line 127: What happens when a user did not provide a catch callback?

rodrigopedra:
It seems you need to rebase your commit, as it includes unrelated commits from other users.

src/Illuminate/Pipeline/Pipeline.php
talelmishali: If no callback is being provided, and the user is chaining the `catch` method, it would throw an `ArgumentCountError`. If the user does not chain the `catch` method, it would run as normal without the try catch block.

talelmishali:
> It seems you need to rebase your commit, as it includes unrelated commits from other users.

I am not entirely sure how those other commits found themselves here. I think I mistakenly synced the fork.

src/Illuminate/Pipeline/Pipeline.php
rodrigopedra: you are right, I missed the if clause. sorry =)

PR #46453, opened 2023-03-13
adds type-hint to Arr::isList

cosmastech:
#### Problem
I was working with `Arr::isList()` in a project when I noticed a test regression. I was not aware that Arr::isList() requires an array to be passed (and I was working with something that returned `?array`). I figured this would help the next dev who might come up against this.

I figured it was an easy enough fix to just add the type-hints. Arr::isAssoc() already had the type-hint added.

##### Note
I was working on 9.x. Not that it makes much of a difference, but just noting.

#### Other Possible Solutions
Remove the typehints and change to `return is_array($array) && array_is_list($array);`

### DIFF
--- a/src/Illuminate/Collections/Arr.php
+++ b/src/Illuminate/Collections/Arr.php
@@ -408,7 +408,7 @@ class Arr
      * @param  array  $array
      * @return bool
      */
-    public static function isAssoc(array $array)
+    public static function isAssoc(array $array): bool
     {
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hey there - I think the docblocks already correctly specify the types.

taylorotwell:
If we were going to add method and return types we would do it thoroughly all at once.

PR #46456, opened 2023-03-14
[9.x] Typehint tap to input value

bert-w:
This allows `tap($myObject)->someFunctionHere()` to be properly typehinted so that chained functions are inferred as elements from `$myObject`,

I've targeted 9.x because this is code annotation only and useful for everyone.


nunomaduro:
I am sorry, but the currently type system in PHP via annotations does not allow to proper type this function. Thank you for your contribution.

PR #46457, opened 2023-03-14
[10.x] Added tests for the "channel:list" command

ash-jc-allen:
Hey! This PR adds a couple of small tests for the new `channel:list` command. They aren't super in-depth, but I thought they might be useful for providing some coverage for the command output 🙂


taylorotwell:
I appreciate the sentiment, but since this class will likely have very low churn I don't want to take on another test to maintain. 😅

ash-jc-allen:
Yeah, that's totally cool! 😄

PR #46459, opened 2023-03-14
[11.x] Add next due to schedule:work output

chu121su12:
Copy some code from `schedule:list` to extract next due information.

image
image


ankurk91:
This can go to v10?

taylorotwell:
See above.

PR #46468, opened 2023-03-14
Improve readability of resolve method in `getClosure` method

milwad-dev:


taylorotwell:
This will break if we change the variable name. This code is really fine 😅

PR #46469, opened 2023-03-14
[10.x] Add next due to schedule:work output

chu121su12:
Copy some code from `schedule:list` to extract next due information.

image
image

This PR is based on #46459 with added extra commit to move `handle(Schedule $schedule)` to `app()` call to prevent public signature changes.


taylorotwell:

PR #46474, opened 2023-03-15
[10.x] Add whereEnum route constraints

RaazPuspa:
Allow an option to create route constraints based on enum/backed-enum cases, which specifies that the given route parameters must be valid case of provided enum class.

**For basic enum**: route parameter must match enum case name
**For backed enum**: route parameter must match enum case value

_**Note**: parameter value is case sensitive_

#### Usage

```
Route::get('{foo}', [MyController::class, 'index'])->whereEnum('foo', MyEnum::class);

Route::prefix('{foo}')->whereEnum('foo', MyEnum::class)->group(function () {
    Route::get('', [MyController::class, 'index']);
});
```

### DIFF
--- a/src/Illuminate/Routing/CreatesRegularExpressionRouteConstraints.php
+++ b/src/Illuminate/Routing/CreatesRegularExpressionRouteConstraints.php
@@ -2,7 +2,9 @@
 
 namespace Illuminate\Routing;
 
+use BackedEnum;
 use Illuminate\Support\Arr;
+use InvalidArgumentException;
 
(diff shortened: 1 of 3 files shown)


Rizky92:
I thought this was already possible?

```php
enum MyEnum: string
{
    case FOO = 'foo';
    case BAR = 'bar';
    case BAZ = 'baz';
}

// routes/web.php
Route::get('{foo}', fn (MyEnum $foo) => $foo);
```

taylorotwell:
See comment above.

PR #46475, opened 2023-03-15
[10.x] Add custom name for foreign key in foreignId

Dakoni4400:
I stumbled upon an error where I had two very long table names and wanted one to have a foreign key to the other. I used the foreignId Method for that and got an error that the foreign key identifier name was too long. When using the foreign Method you can name your keys if you want so to prevent the error I got, I added the same functionality to foreignId and foreignUuid. 

I tested this in a small test environment with creating and dropping these foreign ids

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -942,7 +942,7 @@ class Blueprint
 
         return $model->getKeyType() === 'int' && $model->getIncrementing()
                     ? $this->foreignId($column ?: $model->getForeignKey(), $name)
-                    : $this->foreignUuid($column ?: $model->getForeignKey());
+                    : $this->foreignUuid($column ?: $model->getForeignKey(), $name);
     }
 
     /**
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -1320,7 +1320,7 @@ class Blueprint
      *
      * @param  string  $column
      * @param  int|null  $length
-     * @param  string|null $name
+     * @param  string|null  $name
      * @return \Illuminate\Database\Schema\ForeignIdColumnDefinition
      */
     public function foreignUlid($column, $length = 26, $name = null)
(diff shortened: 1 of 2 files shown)


taylorotwell:
Can you share the exact code to recreate the problem?

Dakoni4400:
> Can you share the exact code to recreate the problem?

Create the two migrations with the following code:
1. Migration: 
```php
<?php

{
    /**
     * Run the migrations.
     */
    public function up(): void
    {
        Schema::create('character_selector_information', function (Blueprint $table) {
            $table->id();
            $table->timestamps();
        });
    }

    /**
     * Reverse the migrations.
     */
    public function down(): void
    {
        Schema::dropIfExists('character_selector_information');
    }
};
```
2. Migration:
```php
<?php

{
    /**
     * Run the migrations.
     */
    public function up(): void
    {
        Schema::create('character_selector_information_checkboxes', function (Blueprint $table) {
            $table->id();
            $table->foreignId('character_selector_information_id', )->constrained();
            $table->timestamps();
        });
    }

    /**
     * Reverse the migrations.
     */
    public function down(): void
    {
        Schema::dropIfExists('user_foreign');
    }
};
```

Then running the migrations will cause the following error: `SQLSTATE[42000]: Syntax error or access violation: 1059 Identifier name 'character_selector_information_checkboxes_character_selector_information_id_foreign' is too long (Connection: mysql, SQL: alter table `character_selector_information_checkboxes` add constraint `character_selector_information_checkboxes_character_selector_information_id_foreign` foreign key (`character_selector_information_id`) references `character_selector_informations` (`id`))
`

hafezdivandari:
@Dakoni4400 please also consider `foreignUlid`: https://github.com/laravel/framework/blob/0b3fae2e8dd094433a760b09eb4366bb139fcf88/src/Illuminate/Database/Schema/Blueprint.php#L1321

Dakoni4400:
@hafezdivandari added support for ulid.

@taylorotwell I hope the example I gave is enough for you. I can share a repository with the problem if you want.

taylorotwell:
If you need to set a custom name for the foreign key, use the `foreign` method. If we wanted to allow this, it would likely be added on the `constrained` method, not the `foreignId` method.

PR #46477, opened 2023-03-15
Cache remember with locking functionality

hdimitrov1:
The issue this is solving is that if you have a cache::remember call, with a closure that takes significant time, further calls to cache::remember will still run the code which might be heavy on your resources.

Example:

```php
Cache::remember('something',60, function() {
			// heavy DB query that takes 5-10 seconds.
			sleep(5);
			return '5';
		});
````

If another request comes in 2.5 seconds into the sleep, the closure will re-run. This will slow down the response for the end user, and (if the closure is resource intensive) hog twice as many resources.

With the rememberLock function, the second request will hang until the first one is complete without doing much. 

```php
Cache::rememberLock('test',10, function() {
			sleep(5);
			return 'jeff';
		})
```

If the request comes in 2.5 seconds in, your user will only need to wait 5 seconds, only for the first one to complete.

This is not replacing the remember function as it's not supported on all drivers, and it does lead to some additional cache load and it would technically be a breaking change.

I'm happy to add tests to this if there's any interest in this as a framework feature.

### DIFF
--- a/src/Illuminate/Cache/Repository.php
+++ b/src/Illuminate/Cache/Repository.php
@@ -665,6 +665,53 @@ class Repository implements ArrayAccess, CacheContract
     }
 
     /**
+     * Checks if driver supports locks
+     * @return bool
+     */
+    public function supportsLocks(): bool {
+        return method_exists($this->store, 'lock');
+    }
+
+    /**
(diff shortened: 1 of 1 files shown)


taylorotwell:
I would probably just call Cache::lock manually in your own code.

PR #46479, opened 2023-03-15
[11.x] Convert route parameters to camelCase

SamuelNitsche:
This PR is the result of this tweet:
https://twitter.com/pascalbaljet/status/1635970593970331649

Currently, registering routes with dashes results in parameters with an underscore.
```
Route::resource('document-version', DocumentVersionController:: class);
// Route Name: document-version.edit
// URI + Segment name: document-version/{document_version}/edit

Route::singleton('document-version.comments', DocumentVersionCommentsController::class);
// Route Name: document-version.comments.edit
// URI + Segment name: document-version/{document_version}/comments/edit
```

The parameters can be renamed using the `parameter` function, but it would be nice if this was the default behavior.
This PR updates the above examples to the following:

```
Route::resource('document-version', DocumentVersionController:: class);
// Route Name: document-version.edit
// URI + Segment name: document-version/{documentVersion}/edit

Route::singleton('document-version.comments', DocumentVersionCommentsController::class);
// Route Name: document-version.comments.edit
// URI + Segment name: document-version/{documentVersion}/comments/edit
```

### DIFF
--- a/src/Illuminate/Routing/ResourceRegistrar.php
+++ b/src/Illuminate/Routing/ResourceRegistrar.php
@@ -611,7 +611,7 @@ class ResourceRegistrar
             $value = Str::singular($value);
         }
 
-        return str_replace('-', '_', $value);
+        return Str::camel($value);
     }
 
     /**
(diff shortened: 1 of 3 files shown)


GrahamCampbell:
This is quite a breaking change, and could pose a real pain for people with packages that need to support both Laravel 10 and 11, and add unnecessary upgrading frequency. If you want camel case, you can already use camel case directly in your routes.

taylorotwell:

PR #46482, opened 2023-03-15
[11.x] Add missing method to filesystem contract

ProjektGopher:
Adds missing `missing` method to `Illuminate/Contracts/Filesystem/Filesystem`.

### DIFF
--- a/src/Illuminate/Contracts/Filesystem/Filesystem.php
+++ b/src/Illuminate/Contracts/Filesystem/Filesystem.php
@@ -27,6 +27,14 @@ interface Filesystem
     public function exists($path);
 
     /**
+     * Determine if a file is missing.
+     *
+     * @param  string  $path
+     * @return bool
+     */
+    public function missing($path);
+
+    /**
      * Get the contents of a file.
      *
      * @param  string  $path


ProjektGopher:
Original PR #46480 targeted `10.x`. Discovered by phpstan when returning an on-demand disk. Return type of `FilesystemAdapter` worked fine, but stan found the contract being returned instead.

Hoping this change can target next 10.x patch release and not just 11.

GrahamCampbell:
> Hoping this change can target next 10.x patch release and not just 11.

No chance of that happening, as Tim explained. Adding a method to an interface is a major breaking change.

ProjektGopher:
Is that just because we don't know what third parties are implementing this contract, even though we're implementing this method already?

taylorotwell:
@ProjektGopher correct

taylorotwell:
Also I think it was left out of contract originally because you can just do `! $filesystem->exists`

ProjektGopher:
Fair enough - Are you okay enforcing this convenience method in the future @taylorotwell, or would you prefer it be left as-is?

taylorotwell:
I think I'll just leave this as-is since you can `! $file->exists`

PR #46488, opened 2023-03-16
[10.x] Fix deprecations for embedded images in symfony mailer

### LINKED ISSUE #46462: [10.x] Mail-component uses embed-syntax deprecated with Symfony Mailer 6.2
- Laravel Version: 10.0.0

### Description:
According to the Symfony-documentation at https://symfony.com/doc/current/mailer.html#embedding-images the use of embed() and embedFromPath() is deprecated with Symfony 6.2. Laravel 10.0 uses 6.2 and still makes use of the deprecated methods in vendor/laravel/framework/src/Illuminate/Mail/Message.php

Whilel Laravel can keep the existing API it should use the new Symfony-API internally and not the deprecated methods.

Jubeki:
Fixes #46462 

It seems like Symfony has deprecated the old way of embedding images in Mails since 6.2:
https://symfony.com/doc/current/mailer.html#embedding-images

This PR implements the new way of embedding images.

### DIFF
--- a/src/Illuminate/Mail/Message.php
+++ b/src/Illuminate/Mail/Message.php
@@ -7,6 +7,8 @@ use Illuminate\Support\Str;
 use Illuminate\Support\Traits\ForwardsCalls;
 use Symfony\Component\Mime\Address;
 use Symfony\Component\Mime\Email;
+use Symfony\Component\Mime\Part\DataPart;
+use Symfony\Component\Mime\Part\File;
 
 /**
  * @mixin \Symfony\Component\Mime\Email
@@ -344,12 +346,16 @@ class Message
                 function ($path) use ($file) {
                     $cid = $file->as ?? Str::random();
 
-                    $this->message->embedFromPath($path, $cid, $file->mime);
+                    $this->message->addPart(
+                        (new DataPart(new File($path), $cid, $file->mime))->asInline()
+                    );
 
                     return "cid:{$cid}";
                 },
                 function ($data) use ($file) {
-                    $this->message->embed($data(), $file->as, $file->mime);
+                    $this->message->addPart(
+                        (new DataPart($data(), $file->as, $file->mime))->asInline()
+                    );
 
                     return "cid:{$file->as}";
                 }
@@ -358,7 +364,9 @@ class Message
 
         $cid = Str::random(10);
 
-        $this->message->embedFromPath($file, $cid);
+        $this->message->addPart(
+            (new DataPart(new File($file), $cid))->asInline()
+        );
 
         return "cid:$cid";
     }
@@ -373,7 +381,9 @@ class Message
      */
     public function embedData($data, $name, $contentType = null)
     {
-        $this->message->embed($data, $name, $contentType);
+        $this->message->addPart(
+            (new DataPart($data, $name, $contentType))->asInline()
+        );
 
         return "cid:$name";
     }


src/Illuminate/Mail/Message.php
```diff
@@ -344,12 +346,16 @@ public function embed($file)
...
+                    $this->message->addPart(
+                        (new DataPart(new File($path), $cid, $file->mime))->asInline()
+                    );
 
                     return "cid:{$cid}";
                 },
                 function ($data) use ($file) {
-                    $this->message->embed($data(), $file->as, $file->mime);
+                    $this->message->addPart(
+                        (new DataPart($data(), $file->as, $file->mime))->asInline()
```
Jubeki, line 357: This is the only part where I am unsure:
I think `$data()` will return a `string` since it is the `dataStrategy` instead of the `fileStrategy`.

But I could not find any usage of this in the framework.

driesvints (approved):
Lgtm in general 👍 

@Jubeki have you tried out a few scenarios with Helo or something similar?

Jubeki:
> @Jubeki have you tried out a few scenarios with Helo or something similar?

I tried using Mailpint with the following code example (and an image stored in `storage/app/profile.jpg`) and it showed the image inline. I did not try it with any other email provider, but can try it out with `ses` if wanted.

```php
// app\Mail\TestMail.php

<?php

namespace App\Mail;

use Illuminate\Bus\Queueable;
use Illuminate\Mail\Mailable;
use Illuminate\Mail\Mailables\Attachment;
use Illuminate\Mail\Mailables\Content;
use Illuminate\Mail\Mailables\Envelope;

class TestMail extends Mailable
{

    /**
     */
    public function __construct()
    {
        //
    }

    /**
     * Get the message envelope.
     */
    public function envelope(): Envelope
    {
        return new Envelope(
            subject: 'Test Mail',
        );
    }

    /**
     * Get the message content definition.
     */
    public function content(): Content
    {
        return new Content(
            view: 'test',
        );
    }

    /**
     * Get the attachments for the message.
     *
     * @return array<int, \Illuminate\Mail\Mailables\Attachment>
     */
    public function attachments(): array
    {
        return [];
    }
}
```

```blade
{{-- resources/views/test.blade.php --}}
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <meta http-equiv="X-UA-Compatible" content="ie=edge">
    <title>Document</title>
</head>
<body>
    <p>TEST E-Mail</p>
    <img src="{{ $message->embed(storage_path('app/profile.jpg')) }}" alt="" style="height: 100px; width: 100px;" />
</body>
</html>
```

driesvints:
perhaps best indeed that we try it out with ses as well. Also maybe try a data attachment?

Jubeki:
@driesvints 

I updated the above example to also test the embedData method.

Both work with Mailhog and SES.

src/Illuminate/Mail/Message.php
taylorotwell: If you aren't sure, we will need to wait for @timacdonald to review this.

taylorotwell:
Thanks

src/Illuminate/Mail/Message.php
timacdonald: @Jubeki you are correct that `$data()` returns a string.

PR #46495, opened 2023-03-16
[10.x] Added assertDatabaseHasCount

andumy:
# Problem
There is currently no method to assert in a database test how many entries are for a specific set of key/values condition. 

# Use Case
We seed a table with `x` amount of entries. The method we test should set a flag from `false` to `true` for `y` amount of entries.

Currently, we cannot check how many entries had the flag set to `true`.

`assertDatabaseHas` will check only if some entries have the flag set as `true`, while `assertDatabaseCount` will check that there are a total of `x` entries in the table.

`assertDatabaseHasCount('table', $y, ['flag' => true])` will assert that exacly `y` entries have the flag set to `true`

# Solution
`assertDatabaseHasCount` is a combination of `assertDatabaseHas` and `assertDatabaseCount`. 

A new constraint `HasCountInDatabase` was created that checks for both amount and data.
This constrained is then asserted in `InteractsWithDatabase` and wrapped as `assertDatabaseHasCount`

Since the logic follows `assertDatabaseHas` and `assertDatabaseCount` it shouldn't break any existing features.

### DIFF
--- a/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php
+++ b/src/Illuminate/Foundation/Testing/Concerns/InteractsWithDatabase.php
@@ -9,6 +9,7 @@ use Illuminate\Database\Events\QueryExecuted;
 use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Testing\Constraints\CountInDatabase;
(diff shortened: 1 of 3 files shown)


taylorotwell:
You can just use `->assertEquals($expectedCount, DB::someQuery()->count());`

Issue #46355, opened 2023-03-05
"Closure object cannot have properties" at the Illuminate\Queue\Queue::shouldDispatchAfterCommit()


jonasraoni:
- Laravel Version: 9.21.5
- PHP Version: 8.0.27
- Database Driver & Version: MySQL 

### Description:
Fatal error at the method `Illuminate\Queue\Queue::shouldDispatchAfterCommit()`.

The line checks for an object and a property:
https://github.com/laravel/framework/blob/7d436e0af0068b57628073f00d24df620c8c4eb2/src/Illuminate/Queue/Queue.php#L328

Closures are expected (see below), but when a closure is passed, a fatal error is triggered:
https://github.com/laravel/framework/blob/7d436e0af0068b57628073f00d24df620c8c4eb2/src/Illuminate/Queue/Queue.php#L323

### Steps To Reproduce:
```php
isset((fn () => 0)->property);
```

PR #46505, opened 2023-03-17
[9.x] PHP 8.0 fix for Closure jobs

bert-w:
Fixes https://github.com/laravel/framework/issues/46355 (passing a Closure would fail because `isset($closure->someProperty)` throws an error on PHP 8.0).


bert-w:
sometimes I feel like there are little gnomes that cause the tests to fail

PR #46442, opened 2023-03-12
[10.x] Fix parsed input arguments for command events using dispatcher rerouting

bert-w:
This is a draft PR for https://github.com/laravel/framework/pull/46430 where a Symfony dispatcher is created for `\Illuminate\Console\Application` in order to reroute the events to the Laravel counterpart. This has the benefit that the `Input` is properly bound to the command definition, and it is thus available in the event itself.

The complete `\Illuminate\Console\Application::run()` override has been removed.

This opens up the possibility to add some more events as well:

- `\Symfony\Component\Console\ConsoleEvents::COMMAND`
  - (Laravel's version: `\Illuminate\Console\Events\CommandStarting`, included in PR)
- `\Symfony\Component\ConsoleConsoleEvents::TERMINATE`
  - (Laravel's version: `\Illuminate\Console\Events\CommandFinished`, included in PR)
- `\Symfony\Component\Console\ConsoleEvents::SIGNAL`
  - (no Laravel version yet)
- `\Symfony\Component\Console\ConsoleEvents::ERROR`
  - (no Laravel version yet)

PR #46508, opened 2023-03-18
[10.x] Move Symfony events dispatcher registration to `Console\Kernel`

crynobone:
Registering the events under `Illuminate\Console\Application` construct can cause memory leaks especially when running tests caused by #46442.

image

### Test running on `10.3.3`

image

### Test running on `10.4.0`

image


driesvints:
cc @bert-w

PR #46511, opened 2023-03-18
[10.x] Skip isParameterBackedEnumWithStringBackingType for non ReflectionNamedType

vladimirmartsul:
#46483 May cause error `Call to undefined method ReflectionUnionType::getName()` at https://github.com/laravel/framework/blob/10.x/src/Illuminate/Support/Reflector.php#L151 if route action and parameter are defined as a union like this
```
public function edit(TwillModelContract|int $id): mixed {...}
```
(actual code from Twill CMS https://github.com/area17/twill/blob/3.x/src/Http/Controllers/Admin/ModuleController.php#L1131 )

This is because `ReflectionUnionType` and `ReflectionIntersectionType` don't have a `getName()` method, unlike `ReflectionNamedType`. So we need to skip checking if *$parameter* is not an instance of `ReflectionNamedType`.

### DIFF
--- a/src/Illuminate/Support/Reflector.php
+++ b/src/Illuminate/Support/Reflector.php
@@ -148,7 +148,7 @@ class Reflector
      */
     public static function isParameterBackedEnumWithStringBackingType($parameter)
     {
-        if(! $parameter->getType() instanceof ReflectionNamedType) {
+        if (! $parameter->getType() instanceof ReflectionNamedType) {
             return false;
         }
 
--- a/src/Illuminate/Support/Reflector.php
+++ b/src/Illuminate/Support/Reflector.php
@@ -148,6 +148,10 @@ class Reflector
      */
     public static function isParameterBackedEnumWithStringBackingType($parameter)
     {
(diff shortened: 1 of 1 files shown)


driesvints:
Thanks @vladimirmartsul. Can we also add a test for this?

danharrin:
Yeah this is a BC that we're also experiencing in Filament. It seems to also affect any Livewire `mount()` method.

morloderex:
@vladimirmartsul I am so sorry for this issue. There wasn't any tests that broke for me when i did this change.

And i didn't know that we needed to adhere for a special case regarding union types and intersection types. Thank you for the bugfix anyway.

Faks:
> @vladimirmartsul I am so sorry for this issue. There wasn't any tests that broke for me when i did this change.
> 
> And i didn't know that we needed to adhere for a special case regarding union types and intersection types. Thank you for the bugfix anyway.

don't worry, at least we found them (who encountered them), we only could for future create for route registration regression tests to avoid such pitfalls?

PR #46514, opened 2023-03-19
Accept json encoding flag as an argument for casts

xPapay:
This PR adds support to pass [json encoding flags](https://www.php.net/manual/en/function.json-encode.php#:~:text=original%0A%20%20%20%C2%BB%C2%A0RFC%207159.-,[flags](https://www.php.net/manual/en/function.json-encode.php#:~:text=original%0A%20%20%20%C2%BB%C2%A0RFC%207159.-,flags,-Bitmask%20consisting%20of),-Bitmask%20consisting%20of) to casts performing json encoding.

It's not a breaking change. If no flag provided then it behaves the same as before this PR.

### Why is this needed? (Usecases)

Because by default native php `json_encode` function escapes multibyte Unicode characters. 
Some languages uses multibyte characters and current casts implementation escapes them before they're stored to database. This made them hard to query later on.

For example let's assume we have an options property on Delivery model and cast it like this:
```php
// Models/Delivery.php
protected $casts = [
    'options' => AsCollection::class,
];

...

$delivery->options = [
    'deliveryAddress' => 'Čerešňová 32, Banská Štiavnica'
];

$delivery->save();
```
when I check the database it is stored like this
```
{"deliveryAddress":"\u010cere\u0161\u0148ov\u00e1 32, Bansk\u00e1 \u0160tiavnica"}
```
when I try to query the database I can't find my record
```sql
select * from deliveries where options like '%Čerešňová%';
>>> 0 results
```

After this PR we can supply 2nd argument to `json_encode` function and we can do following:
```php
// Models/Delivery.php
protected $casts = [
    'options' => AsCollection::class . ":" .  JSON_UNESCAPED_UNICODE, // <--- utilizing arguments passing as documented here: https://laravel.com/docs/10.x/eloquent-mutators#cast-parameters
];

// NOTE: JSON_UNESCAPED_UNICODE is native predefined PHP constant and can be found here: https://www.php.net/manual/en/json.constants.php

...

$delivery->options = [
    'deliveryAddress' => 'Čerešňová 32, Banská Štiavnica'
];

$delivery->save();
```
when I check the database it is stored like this
```
{"deliveryAddress":"Čerešňová 32, Banská Štiavnica"}
```
when I try to query the database I can find my record
```sql
select * from deliveries where options like '%Čerešňová%';
>>> 1 result
```

Of course there is more usecases.
Main point is to be able to pass 2nd argument to `json_encode` function to customize how we want to encode it.

Other people with same problem:
https://laracasts.com/discuss/channels/laravel/model-cats-to-array-utf-8-json-unescaped-unicode

### DIFF
--- a/src/Illuminate/Database/Eloquent/Casts/AsArrayObject.php
+++ b/src/Illuminate/Database/Eloquent/Casts/AsArrayObject.php
@@ -15,8 +15,15 @@ class AsArrayObject implements Castable
      */
     public static function castUsing(array $arguments)
     {
-        return new class implements CastsAttributes
+        return new class($arguments) implements CastsAttributes
         {
+            protected $arguments;
+
+            public function __construct(array $arguments)
+            {
+                $this->arguments = $arguments;
+            }
+
             public function get($model, $key, $value, $attributes)
             {
                 if (! isset($attributes[$key])) {
@@ -30,7 +37,9 @@ class AsArrayObject implements Castable
 
             public function set($model, $key, $value, $attributes)
             {
-                return [$key => json_encode($value)];
+                $encodingFlag = isset($this->arguments[0]) ? $this->arguments[0] : 0;
+                
+                return [$key => json_encode($value, $encodingFlag)];
             }
 
             public function serialize($model, string $key, $value, array $attributes)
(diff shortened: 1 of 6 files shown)


xPapay:
Of course previous examples assumes database's encoding is utf8 unicode to properly store special characters.

morloderex:
@xPapay What about the raw `array` cast type? It doesn't seem like you are handling that case?

Issue #46492, opened 2023-03-16
MYSQL 8 Migrating text to alternative text type does not change column


liamh101:
- Laravel Version: 10.3.3
- PHP Version: 8.2.3
- Database Driver & Version: MYSQL 8.0.32
- `doctrine/dbal` Version: 3.6.1

### Description:

The column type change will not be made during a text column's migration to medium or long text. Keeping the column as a text type.

Stepping through the process, `Illuminate\Database\Schema\Grammars\ChangeColumn` class line 44 `$tableDiff->isEmpty()` returns `true`. 

Not sure what other columns this affects. This bug occurred during my migration process to Laravel 9 to 10.

### Steps To Reproduce:

1. Set MySQL config in `database.php` as follows (based on MySQL8 docker image): 
```
        'mysql' => [
            'driver' => 'mysql',
            'url' => null,
            'host' => 'mysql',
            'port' => '3306',
            'database' => 'test',
            'username' => 'root',
            'password' => 'root',
            'unix_socket' => '',
            'charset' => 'utf8mb4',
            'collation' => 'utf8mb4_unicode_ci',
            'prefix' => '',
            'prefix_indexes' => true,
            'strict' => true,
            'engine' => null,
            'options' => extension_loaded('pdo_mysql') ? [
                PDO::MYSQL_ATTR_SSL_CA => null,
                PDO::MYSQL_ATTR_SSL_VERIFY_SERVER_CERT => 0,
            ] : [],
        ],
```
2. Create a migration that creates a table with three text columns.
3. Create another migration changing these columns:
 ```
    public function up(): void
    {
        Schema::table('tableName', static function (Blueprint $table) {
            $table->mediumText('col1')->change();
            $table->longText('col2')->change();
            $table->mediumText('col3')->change();
        });
    }
```
4. Run migrations.
5. Columns are still `text` type.

liamh101:
Here's a repo that reproduces the error: https://github.com/liamh101/laravel-column-bug

ankurk91:
See https://github.com/laravel/framework/pull/45487

liamh101:
Thanks for the response. I'll remove dbal from our codebase and check if it doesn't affect anything else. 

However, I think this problem should be fixed. On closer inspection, what I believe is happening is the current type and new type is checked using an enum in the dbal library. However, any text column generates the same enum. Meaning the migration thinks nothing has changed.

If there's no movement by the end of today I'll spend some time and get a fix for this over the weekend.

[...]

PR #46517, opened 2023-03-19
[10.x] Replace Deprecated DBAL Comparator creation with schema aware Comparator

liamh101:
Reference to change in DBAL and why this needs to be done: https://github.com/doctrine/dbal/pull/4746

This is currently causing this bug in Laravel where certain migration changes are ignored: https://github.com/laravel/framework/issues/46492

Not sure if this is affecting other database types.

An example project of the existing bug can be found here: https://github.com/liamh101/laravel-column-bug

### DIFF
--- a/src/Illuminate/Database/Schema/Grammars/ChangeColumn.php
+++ b/src/Illuminate/Database/Schema/Grammars/ChangeColumn.php
@@ -60,7 +60,7 @@ class ChangeColumn
     {
         $current = $schema->introspectTable($grammar->getTablePrefix().$blueprint->getTable());
 
-        return (new Comparator)->compareTables(
+        return $schema->createComparator()->compareTables(
             $current, static::getTableWithColumnChanges($blueprint, $current)
         );
     }
--- a/src/Illuminate/Database/Schema/Grammars/ChangeColumn.php
+++ b/src/Illuminate/Database/Schema/Grammars/ChangeColumn.php
@@ -3,7 +3,6 @@
 namespace Illuminate\Database\Schema\Grammars;
 
 use Doctrine\DBAL\Schema\AbstractSchemaManager as SchemaManager;
-use Doctrine\DBAL\Schema\Comparator;
 use Doctrine\DBAL\Schema\Table;
 use Doctrine\DBAL\Types\Type;
 use Illuminate\Database\Connection;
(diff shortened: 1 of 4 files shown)


hafezdivandari:
@liamh101 What about `TableDiff` on this function, do we have to replace that too?

https://github.com/laravel/framework/blob/7d15f7eef442633cff108f83d9fe43d8c3b8b76f/src/Illuminate/Database/Schema/Grammars/Grammar.php#L324-L331

liamh101:
@hafezdivandari Just need to do the `introspectTable` before creating the `TableDiff` class, and I believe the `tap` function call is now redundant. I'll get that changed and see if there are any side effects.

hafezdivandari:
@liamh101 isn't it better to change this function and use it on `ChangeColumn` class too if possible?

liamh101:
@hafezdivandari can I do some further cleaning as a follow-up PR? I don't want to change too much and introduce additional bugs.

My feeling is that this bug can go under the radar on many projects and have massive consequences, partially multi-tenant projects where fresh database migrations can happen regularly.

hafezdivandari:
@liamh101 sure, it was just a suggestion. But it seems that `RenameColumn` class has the same issue, I think we should not use `TableDiff` directly and use `$schema->createComparator()->compareTables()`. Does renaming a `mediumText` using DBAL column works right now?

liamh101:
Renaming a column will work. However, changing a text type using the default Comparator won't work.

driesvints (approved):
Thanks for digging more into this @liamh101. I feel like we should give this a shot.

liamh101:
Sure!

- Replace deprecated DBAL Comparator creation for a Schema Aware Comparator. ( `ChangeColumn.php`) Reasoning why DBAL deprecated this can be found here:  https://github.com/doctrine/dbal/pull/4746
- Clean up the additional deprecated function call to DBAL in `Grammar.php` however does not affect any behaviour.
- Changes in `DatabaseSchemaBlueprintIntegrationTest.php` to account for SQL changes present between stable and lowest versions of `DBAL 3.x` .
- Add schema builder test to validate text types change in MYSQL. The current version of Laravel is bugged with existing and new migrations. A project demonstrating this issue can be found here: https://github.com/liamh101/laravel-column-bug

No breaking changes were introduced in this PR, just a bug fix.

Once this has been reviewed and merged, I will investigate other places Laravel might use deprecated DBAL functionality and replace them. This PR is to fix the migration issues introduced when upgrading DBAL from 2.x to 3.x.

taylorotwell:
Thanks

PR #46522, opened 2023-03-20
[10.x] Fix default pivot model values

iamgergo:
Recently I faced with a situation, when using the `withPivotValue` method on a  `BelongsToMany` relationship definition does not take effect when saving the pivot record.

-----------

I have the following setup:

```php
class User extends Model
{
    public function categories(): BelongsToMany
    {
        return $this->belongsToMany(Category::class, 'subscriptions')
                    ->withPivotValue('type', 'manager')
                    ->using(Subscription::class);
    }

    public function subscriptions(): BelongsToMany
    {
        return $this->belongsToMany(Category::class, 'subscriptions')
                    ->withPivotValue('type', 'player')
                    ->using(Subscription::class);
    }
}
```

```php
class Subscription extends Pivot
{
    public $table = 'subscriptions';

    protected $attributes = [
        'type' => 'player',
    ];
}
```

Before the fix, when attaching a category to a user, the pivot attributes that are defined in the `withPivotValue` method are overridden by the pivot model's `$attributes` property.

```php
$category = Category::first();
$user = User::first();

$user->categories()->attach($category); // try to attach as a category NOT as a subscription

$user->categories->isEmpty() // true

$user->subscriptions->isEmpty() // false

$user->subscriptions->first()->pivot->role // palyer
```

So, when I try to attach a category to a user using the `categories` relation, since it cannot override the default value using the `withPivotValue`, the attached category model takes place in the wrong relationship. Also, I think this is a bug.

--------

The same Pivot Model can be used in different relationships, so overriding its default attributes from a relationship definition would be great.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
@@ -510,6 +510,8 @@ trait InteractsWithPivotTable
      */
     public function newPivot(array $attributes = [], $exists = false)
     {
+        $attributes = array_merge(array_column($this->pivotValues, 'value', 'column'), $attributes);
+
         $pivot = $this->related->newPivot(
             $this->parent, $attributes, $this->table, $exists, $this->using
         );
--- a/tests/Integration/Database/EloquentPivotTest.php
+++ b/tests/Integration/Database/EloquentPivotTest.php
@@ -26,6 +26,7 @@ class EloquentPivotTest extends DatabaseTestCase
         Schema::create('collaborators', function (Blueprint $table) {
             $table->integer('user_id');
             $table->integer('project_id');
+            $table->string('role')->nullable();
             $table->text('permissions')->nullable();
         });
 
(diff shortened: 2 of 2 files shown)


taylorotwell:
`withPivotValue` has nothing to do with updates. Only retrievals. You need to specify the pivot values you want when saving.

iamgergo:
@taylorotwell @driesvints The `withPivotValue` and the `withPivot` methods are doing two different things.

> withPivotValue has nothing to do with updates

The `withPivotValue` defines a `wherePivot` statement and sets a pivot value when saving the relation. So it is used when saving relations.

The `withPivot` method just retrieves a pivot column.

If I want to retrieve pivot data, I have to use the `withPivot` even if I use `withPivotValue` as well. For example:

```php
    public function categories(): BelongsToMany
    {
        return $this->belongsToMany(Category::class, 'subscriptions')
                    ->withPivotValue('type', 'manager') // sets the type pivot attribute automatically
                    ->withPivot('type') // retrieves the type attribute
                    ->using(Subscription::class);
    }
```
------

Without this fix, the `withPivotValue` does not work in some scenarios. From my point of view, this is a bug. Please, consider to reopen this PR.

Thank you!

PR #46525, opened 2023-03-20
adding not_current_password validation rule

ahmmmmad11:
adding not current password will simplify the process of one of the most repeated tasks which is updating password.

for example:
 
    public function store(Request $request): Response
    {
        $validated = $request->validate([
            'old_password' => ['required', 'current_password'],
            'new_password' => ['required', 'not_current_password']
        ]);
 
        // rest of the logic
    }

of course, someone can say we can do the same thing with:

     'old_password' => ['required', 'current_password'],
     'new_password' => ['required', 'different:old_password']

that's true, but some applications demands' that the password should be changed after some period, and does not require the user to enter the current password, just the new password and its confirmation.

### DIFF
--- a/src/Illuminate/Translation/lang/en/validation.php
+++ b/src/Illuminate/Translation/lang/en/validation.php
@@ -109,6 +109,7 @@ return [
     'missing_with' => 'The :attribute field must be missing when :values is present.',
     'missing_with_all' => 'The :attribute field must be missing when :values are present.',
     'multiple_of' => 'The :attribute field must be a multiple of :value.',
+    'not_current_password' => 'cannot use the current password.',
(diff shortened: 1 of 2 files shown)


driesvints:
Maybe `different_password`?

axlon:
@ahmmmmad11 Couldn't you use [different](https://laravel.com/docs/10.x/validation#rule-different) since you pass both the current and the new password in the same request?

ahmmmmad11:
Hi @axlon 

that's just example, but there is other use case where the user doesn't have to enter the old password to change his current password, or maybe there is a case where a user may have a password and PIN code and there is a condition of the password should not equal the PIN.

there is a lot of use cases for using `not_current_password`

taylorotwell:

PR #46527, opened 2023-03-20
[10.x] Add pipe function to Process layer

WendellAdriel:
This is a follow-up of #46239 which had some issues with the branching.

This PR adds a new function `pipe` to the **Process** layer that receives an array of processes and runs them in sequence passing the output of the previous process as the input for the next one.

The syntax will be like this:

```php
$pipe = Process::pipe(function ($pipe) {
    $pipe->command('cat test.txt'),
    $pipe->command('grep -i "foo"'),
});

$pipe->run()->output(); // "foo"
```

```
Hello, world
foo
```

This can be very handy when we need to chain processes that rely on the result of the previous one to run correctly.


WendellAdriel:
@driesvints @taylorotwell 
Here's the new PR with the git issues fixed and also the new Exception wrapping the errors while running the pipe.

WendellAdriel:
@driesvints @taylorotwell this failed a test that's not related to my changes, do I need to do anything else here?

driesvints:
@WendellAdriel I retriggered those

WendellAdriel:
> @WendellAdriel I retriggered those

Thank you so much!!! 😉

WendellAdriel:
@taylorotwell is there anything else missing that I need to work on for this?

taylorotwell:
Your usage example shows you passing an array to pipe, but the pipe methods expects a `callable` in the code?

WendellAdriel:
> Your usage example shows you passing an array to pipe, but the pipe methods expects a `callable` in the code?

I just updated the example in the PR description!

taylorotwell:
Do we actually need a `ProcessPipeException`? Can we just re-use `ProcessFailedException`?

WendellAdriel:
That's a good point, I think we can just return the last failed process instead now that you mention it.
I'm going to update the PR.

WendellAdriel:
@taylorotwell I updated the code to return the last failed process, removed the `ProcessPipeException` class and added a new test case to check for a failed process pipe run.

taylorotwell:
Another thing - we allow you to add things to the pipe with a key (like pools), but we never pass that key to the output handler like we do with pooled proesses, this means if you pass as output callback to pipe's `run` method, you can't tell which process the output is from.

WendellAdriel:
> Another thing - we allow you to add things to the pipe with a key (like pools), but we never pass that key to the output handler like we do with pooled proesses, this means if you pass as output callback to pipe's `run` method, you can't tell which process the output is from.

@taylorotwell thanks for the heads up, I missed that in the implementation 😅. I just updated the PR with a new commit for that.

taylorotwell:
Thanks!

WendellAdriel:
> Thanks!

I’m happy to help!!! 😉 💪

PR #46530, opened 2023-03-20
[10.x] Configuration value injection

MelchiorKokernoot:
## Discussion
The idea from this pull request has been discussed [here](https://github.com/laravel/framework/discussions/46227) and [here](https://www.reddit.com/r/laravel/comments/11jvzl7/package_for_injection_of_config_values/) previously.

I have wrapped this functionality in a package which already achieves this [here](https://github.com/MelchiorKokernoot/laravel-autowire-config) but received enough positive feedback to believe this way of injecting configurations might be a nice feature for the framework itself.

## Functionality
This PR adds a way of injecting configuration values into class dependencies through the usage of a newly included php attribute, the `InjectedConfig` attribute. An example of this would be:

```php
class Foo {
    public function __construct(
        #[InjectedConfig('app.name')]
        public string $myConfiguredAppName,
    ){}
}
```
in which case the `app.name` config key's value would be injected into the `string` property of `$myConfiguredAppName`.

A clear benefit of this, is that config values can now be supplied as class dependencies, as opposed to using the `config()` helper in code (which I often see being scattered all around the place). This makes it very clear which configuration values a class depends on, and provides (in my opinion) a cleaner way of using config values.

I don't see how this might break backwards compatibility, and don't really think it does, as it hooks in right before the part of the `resolvePrimitive` where a `BindingResolutionException` is thrown. 

I am really interested if this is a feature that is desired, and am open to any and all feedback regarding it.

Kind regards,

Melchior Kokernoot


morloderex:
@MelchiorKokernoot I like the idea but unfortunately the implementation could use some improvement.

Specifically the related part where you do `isset($this['config'])` essentially we are now recursively looking for a class which implements the `\Illuminate\Contracts\Config\Repository` class but not the fqcn of this class what if the config alias to this class gets updated to a class that doesn't adhere to the interface?

Or what if the config component isn't installed i don't know if the Illuminate\Container should be dependent on the \Illuminate\Config component. I am mostly concerned about the persons using the container component outside laravel? I might be wrong but it seems to me that this would (in order to work probably put a soft dependency on the Config component)?

src/Illuminate/Container/Container.php
```diff
@@ -182,8 +184,8 @@ public function when($concrete)
     public function bound($abstract)
     {
         return isset($this->bindings[$abstract]) ||
-               isset($this->instances[$abstract]) ||
-               $this->isAlias($abstract);
+            isset($this->instances[$abstract]) ||
+            $this->isAlias($abstract);
```
driesvints, line 188: Can you please undo all indentation changes in this PR? Keeps the diff clean to review.

src/Illuminate/Container/Container.php
```diff
@@ -1024,6 +1026,26 @@ protected function resolvePrimitive(ReflectionParameter $parameter)
             return [];
         }
 
+        if (isset($this['config'])) {
+            foreach ($parameter->getAttributes() as $attribute) {
```
abenerd, line 1030:
```suggestion
            foreach ($parameter->getAttributes(InjectedConfig::class) as $attribute) {
```

Retrieving all attributes without specifying a name is dangerous and not very performant as we get all attributes specified even ones that can not be instantiated as they are from dev dependencies, causing errors for the users application.

src/Illuminate/Container/Container.php
MelchiorKokernoot: will do thanks, I thought I read somewhere that I should not bother too much because styleCI will fix it, but sure!

MelchiorKokernoot:
@morloderex I changed it to check for the existance of the \Illuminate\Contracts\Config\Repository::class interface, and added an instanceof check, do you think this prevents the situation you described?

src/Illuminate/Config/Attributes/InjectedConfig.php
```diff
@@ -0,0 +1,13 @@
+<?php
+
+namespace Illuminate\Config\Attributes;
+
+use Attribute;
+
+#[Attribute]
+class InjectedConfig
+{
+    public function __construct(string $key = '')
```
mbabker, line 10:
```suggestion
    public function __construct(string $key)
```

The key seems to be required, defaulting it to an empty string doesn't make much sense IMO.

src/Illuminate/Container/Container.php
```diff
@@ -1024,6 +1026,26 @@ protected function resolvePrimitive(ReflectionParameter $parameter)
             return [];
         }
 
+        if ($this->hasConfigRepository()) {
+            foreach ($parameter->getAttributes(InjectedConfig::class) as $attribute) {
+                $attributeName = $attribute->getName();
+
+                if ($attributeName !== InjectedConfig::class) {
+                    continue;
+                }
```
mbabker, line 1035:
```suggestion
```

With the change in the `$parameter->getAttributes()` call, this check is no longer needed.

src/Illuminate/Container/Container.php
```diff
@@ -1024,6 +1026,26 @@ protected function resolvePrimitive(ReflectionParameter $parameter)
...
+
+                if (count($attribute->getArguments()) !== 1) {
+                    throw new \InvalidArgumentException(
+                        'ConfigValueWrapper attribute must have exactly one argument'
+                    );
+                }
```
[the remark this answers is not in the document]
MelchiorKokernoot: Don't really think so, but I will verify asap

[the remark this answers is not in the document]
abenerd: When you define the attribute it is a normal PHP Class, in the constructor you can use property promotion to make the `$key` accessible and instead of saying `$attribute->getArguments()` you can say `$attribute->newInstance()` to get an actual instance of the class and then access `$instance->key`. The user is required to supply the key when using the attribute because they're instantiating a new instance of that class and the constructor requires that argument.

MelchiorKokernoot:
Should be fine now :)

src/Illuminate/Container/Container.php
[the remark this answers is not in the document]
MelchiorKokernoot: Thanks! That's new for me. I am updating it immediately

tests/Config/InjectedConfigTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+    public function testItInjectsValuesForInjectedConfigAttributedProperties(): void
+    {
+        $container = new Container();
+        $mock = m::mock(Repository::class);
+        $mock = $this->mockConfigInjection($mock, 'foo.string', 'Laravel');
+        $mock = $this->mockConfigInjection($mock, 'foo.boolean', true);
+        $mock = $this->mockConfigInjection($mock, 'foo.int', 1);
+        $mock = $this->mockConfigInjection($mock, 'foo.float', 1.1);
+        $mock = $this->mockConfigInjection($mock, 'foo.array', []);
```
X-Coder264, line 29: I don't see the need to mock all this stuff. It makes the test less readable.

You could just do:

```php
use Illuminate\Config\Repository;

$config = new Repository([
     'foo.string' => 'Laravel',
     // all other params
]);
```

and then bind that to the container

MelchiorKokernoot: Thanks for the suggestion!

MelchiorKokernoot: I have reverted the commits for now because I cannot figure out why this is happening

MelchiorKokernoot:
As far as I am concerned, this PR is ready for another review

bert-w:
https://gist.github.com/bert-w/78988fff5e280f353840c25fbcc5ea43 Try this out for the testing class. I also removed the `$object->fooNull` assertion because it throws a warning since it is not a property of the class.

bert-w:
I dont see how that failing test is related to the code in this PR. It has to do something with `mt_srand()` (seeding the RNG so tests are consistent)? Not sure what is happening there.

taylorotwell:
Should default values be supported?

MelchiorKokernoot:
> Should default values be supported?

@taylorotwell haven't really had the need for it myself, but of course I can adjust it so default values are supported. Would you like that?

MelchiorKokernoot:
I have implemented default values through the InjectedConfig attribute's second parameter

johanrosenson:
Does the new method `hasConfigRepository(): bool` need to be public? Does it need to be a method at all? Adding a new method (especially a public method) adds to the long term maintenance burden, it will be hard to rename/move it later because of a potential BC-break.

src/Illuminate/Container/Container.php
```diff
@@ -1280,6 +1290,16 @@ protected function fireCallbackArray($object, array $callbacks)
         }
     }
 
+    /**
+     * @return bool
+     */
+    public function hasConfigRepository(): bool
```
johanrosenson, line 1296: Does this method need to be public? 
Does it need to be a method at all? 

Adding a new method (especially a public method) adds to the long term maintenance burden, it will be hard to rename/move it later because of a potential BC-break.

MelchiorKokernoot: Agreed. I created this because I thought it improved readability a bit. I will merge it back into the main functionality.

src/Illuminate/Container/Container.php
```diff
@@ -1024,6 +1026,17 @@ protected function resolvePrimitive(ReflectionParameter $parameter)
             return [];
         }
 
+        if (isset($this['config']) &&
```
johanrosenson, line 1029: untested code, but i think you could improve readability like this

```
if ($this->has('config') && is_subclass_of($this['config'], \Illuminate\Contracts\Config\Repository::class))
```

`is_subclass_of` will always return `false` for non existing classes/interfaces.

MelchiorKokernoot: Okay I will implement this !

MelchiorKokernoot: Test seems broken now, not sure if it's related though

johanrosenson: > Test seems broken now, not sure if it's related though

Seems to be unrelated, at least "tests / PHP 8.1 - prefer-lowest (pull_request)" is a deprecation warning in an unrelated part of Laravel.

MelchiorKokernoot: Has been fixed now

MelchiorKokernoot:
Hi @taylorotwell, would you please elaborate a little as to why you are leaving it out? It could help me improve for the next time I want to PR something, and it would mean a lot to me :) I'd love to hear from you

MelchiorKokernoot:
@taylorotwell If you have any chance of taking a look at this, I'd still appreciate it a lot!

reincode050:
> A little torn on this one because I do kind of like it, but going to leave it out for now. May revisit this in the future though.

Is 'the future' now @taylorotwell ?

PR #46533, opened 2023-03-21
[10.x] routesAreCached performance

vaawebdev:
## Intro

I develop a large-scale application designed as a feature-based (modular) application.
To decouple features, I use dedicated ServiceProvider for each feature.

Most of the time a feature `ServiceProvider` contains the following:

```php
namespace App\Features\ExampleFeature;

use Illuminate\Support\ServiceProvider as BaseServiceProvider;

final class ServiceProvider extends BaseServiceProvider
{
    public function boot(): void
    {
        $this->loadRoutesFrom(__DIR__ . '/example-feature-routes.php');
    }
}

```

Before registering routes `loadRoutesFrom` method checks if routes are cached

```php
protected function loadRoutesFrom($path)
{
    if (! ($this->app instanceof CachesRoutes && $this->app->routesAreCached())) {
        require $path;
    }
}
```

The `routesAreCached` method does a filesystem call to check if the`cache/routes-v7.php` file exists.

```php
public function routesAreCached()
{
    return $this['files']->exists($this->getCachedRoutesPath());
}
```

## The Problem

Currently my application contains more than 300 features with corresponding route files to be registered, this means that the framework has to do more than 300 filesystem calls to check if routes are cached, and they are always cached in case of our production. The amount of features constantly grows, which leads to a performance bottleneck.

Simple benchmarking gave me insights that currently it takes about 1.5ms solely for `loadRoutesFrom` calls.

## The Solution

I simply memoized the `routesAreCached` call within the application since if routes are cached once, they will be cached for the entire application lifecycle.

After this fix, it takes about `0.031ms` for `loadRoutesFrom` calls, also fixing bottleneck for case when route files amount grows even more.

### DIFF
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -188,6 +188,13 @@ class Application extends Container implements ApplicationContract, CachesConfig
     protected $absoluteCachePathPrefixes = ['/', '\\'];
 
     /**
+     * Indicates if the application cache was resolved.
+     *
+     * @var bool
+     */
+    protected $routesAreCached = false;
+
+    /**
      * Create a new Illuminate application instance.
      *
      * @param  string|null  $basePath
@@ -1120,7 +1127,13 @@ class Application extends Container implements ApplicationContract, CachesConfig
      */
     public function routesAreCached()
     {
-        return $this['files']->exists($this->getCachedRoutesPath());
+        if ($this->routesAreCached) {
+            return true;
+        }
+
+        $this->routesAreCached = $this['files']->exists($this->getCachedRoutesPath());
+
+        return $this->routesAreCached;
     }
 
     /**
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -188,7 +188,7 @@ class Application extends Container implements ApplicationContract, CachesConfig
     protected $absoluteCachePathPrefixes = ['/', '\\'];
 
     /**
-     * Indicates if the application cache was resolved.
+     * Indicates if the application routes cache was resolved.
      *
      * @var bool
      */
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -190,9 +190,9 @@ class Application extends Container implements ApplicationContract, CachesConfig
     /**
      * Indicates if the application routes cache was resolved.
      *
-     * @var bool
+     * @var bool|null
      */
-    protected $routesAreCached = false;
+    protected $routesAreCached = null;
 
     /**
      * Create a new Illuminate application instance.
@@ -1127,11 +1127,9 @@ class Application extends Container implements ApplicationContract, CachesConfig
      */
     public function routesAreCached()
     {
-        if ($this->routesAreCached) {
-            return true;
+        if (is_null($this->routesAreCached)) {
+            $this->routesAreCached = $this['files']->exists($this->getCachedRoutesPath());
         }
-
-        $this->routesAreCached = $this['files']->exists($this->getCachedRoutesPath());
 
         return $this->routesAreCached;
     }
--- a/src/Illuminate/Foundation/Application.php
+++ b/src/Illuminate/Foundation/Application.php
@@ -1127,11 +1127,7 @@ class Application extends Container implements ApplicationContract, CachesConfig
      */
     public function routesAreCached()
     {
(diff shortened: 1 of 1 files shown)


driesvints (approved):

driesvints:
Great find! 👏

src/Illuminate/Foundation/Application.php
```diff
@@ -1120,7 +1127,11 @@ public function getCachedConfigPath()
      */
     public function routesAreCached()
     {
-        return $this['files']->exists($this->getCachedRoutesPath());
+        if (is_null($this->routesAreCached)) {
+            $this->routesAreCached = $this['files']->exists($this->getCachedRoutesPath());
+        }
+
+        return $this->routesAreCached;
```
axlon, line 1134:
```suggestion
        return $this->routesAreCached ??= $this['files']->exists($this->getCachedRoutesPath());
```

Kind of a nit-pick, but this does the same thing and is shorter (also by doing an isset check the explicit null default becomes unnecessary)

driesvints:
@crynobone made the good remark that we also need to make sure this is cleared again for Octane subsequent requests and to check if it doesn't breaks cached routing tests on Testbench.

vaawebdev:
> @crynobone made the good remark that we also need to make sure this is cleared again for Octane subsequent requests and to check if it doesn't breaks cached routing tests on Testbench.

Im not sure if this even must be cleared for Octane. As I know application routing & ServiceProviders are bootstrapped before handling requests.

vaawebdev:
I think that this pipeline test case failure has noting to do with this PR. Can someone take a look?  https://github.com/laravel/framework/actions/runs/4476527864/jobs/7866978257?pr=46533

driesvints:
@vaawebdev yeah I retriggered the build for you.

taylorotwell:
Override this on the application class in your own application. I don't know what kinds of state based breaking changes this would have in people's tests, etc.

vaawebdev:
PHP will not cache file operations. For each call, it will perform a read operation

GrahamCampbell:
The following functions are all cached, if the config is enabled: https://www.w3schools.com/php/func_filesystem_clearstatcache.asp.

GrahamCampbell:
* [stat()](https://www.w3schools.com/php/func_filesystem_stat.asp)
* [lstat()](https://www.w3schools.com/php/func_filesystem_lstat.asp)
* [file_exists()](https://www.w3schools.com/php/func_filesystem_file_exists.asp)
* [is_writable()](https://www.w3schools.com/php/func_filesystem_is_writable.asp)
* [is_readable()](https://www.w3schools.com/php/func_filesystem_is_readable.asp)
* [is_executable()](https://www.w3schools.com/php/func_filesystem_is_executable.asp)
* [is_file()](https://www.w3schools.com/php/func_filesystem_is_file.asp)
* [is_dir()](https://www.w3schools.com/php/func_filesystem_is_dir.asp)
* [is_link()](https://www.w3schools.com/php/func_filesystem_is_link.asp)
* [filectime()](https://www.w3schools.com/php/func_filesystem_filectime.asp)
* [fileatime()](https://www.w3schools.com/php/func_filesystem_fileatime.asp)
* [filemtime()](https://www.w3schools.com/php/func_filesystem_filemtime.asp)
* [fileinode()](https://www.w3schools.com/php/func_filesystem_fileinode.asp)
* [filegroup()](https://www.w3schools.com/php/func_filesystem_filegroup.asp)
* [fileowner()](https://www.w3schools.com/php/func_filesystem_fileowner.asp)
* [filesize()](https://www.w3schools.com/php/func_filesystem_filesize.asp)
* [filetype()](https://www.w3schools.com/php/func_filesystem_filetype.asp)
* [fileperms()](https://www.w3schools.com/php/func_filesystem_fileperms.asp)

vaawebdev:
> The following functions are all cached, if the config is enabled: https://www.w3schools.com/php/func_filesystem_clearstatcache.asp.

Thanks for the info, really helpful! Anyways I already memoized loading routes before making this PR. I don't think that the global filesystem cache is a good solution for my app.

vaawebdev:
> The following functions are all cached, if the config is enabled: https://www.w3schools.com/php/func_filesystem_clearstatcache.asp.

Just made a little infrastructure research, and the benchmark for non-memoized versions that I mentioned in the PR description were already cached by the PHP.

vaawebdev:
@taylorotwell any chance this change will be useful in the next major version?

Issue #46534, opened 2023-03-21
Redis PhpRedisConnactions::hmget should not be using array_values


dave2309:
- Laravel Version: 9.41.0
- PHP Version: 8.0.27
- Database Driver & Version: Redis 5.0.7

### Description:
Illuminate\Redis\Connections\PhpRedisConnection.php
```
    /**
     * Get the value of the given hash fields.
     *
     * @param  string  $key
     * @param  mixed  $dictionary
     * @return array
     */
    public function hmget($key, ...$dictionary)
    {
        if (count($dictionary) === 1) {
            $dictionary = $dictionary[0];
        }

        return array_values($this->command('hmget', [$key, $dictionary]));
    }
```
Using `array_values` remove the actual key pairs array structure which is a real shame as an information is just lost.
I don't see the benefit of doing so.
Keeping the key/value pairs are very handy when dealing with (micro)services that different locations can point to the same redis cache and read the content in a consistent way.
Other framework (leg: cakephp) does keep the return values with the keys passed and their associated values.

### Steps To Reproduce:
Storing using `hmset` keys (+values) like 
```
[
    id => 0,
    q0 => ['test'],
    q1 => [],
]
```
with laravel result of `hmget` (with the same keys) is:
```
[
    0 => 0
    1 => ['test'],
    2 => [],
]
```
With cakephp (for example) same `hmget` would result with
```
[
    'id' => 0,
    'q0' => ['test'],
    'q1' => [],
]
```

Regards,
David

driesvints:
It seems we're not going to be making any changes here, sorry.

PR #46535, opened 2023-03-21
[9.x] Don't use array_values when returning from hmget

dave2309:
As describe by issue #46534, there is no benefit of parsing the
returned redis key/values with `array_values`. In fact, we are
loosing an information (the keys) which could be handy to keep when
it is read from another code/service/framework, as well as, to
confirm that the values correspond to the requested keys.
A last point, is that performance wise not using `array_values`
could be beneficial as well.

ref #46534

### DIFF
--- a/src/Illuminate/Redis/Connections/PhpRedisConnection.php
+++ b/src/Illuminate/Redis/Connections/PhpRedisConnection.php
@@ -114,7 +114,7 @@ class PhpRedisConnection extends Connection implements ConnectionContract
             $dictionary = $dictionary[0];
         }
 
-        return array_values($this->command('hmget', [$key, $dictionary]));
+        return $this->command('hmget', [$key, $dictionary]);
     }
 
     /**
--- a/tests/Redis/RedisConnectionTest.php
+++ b/tests/Redis/RedisConnectionTest.php
@@ -443,6 +443,10 @@ class RedisConnectionTest extends TestCase
                 $redis->hmget('hash', 'name', 'hobby')
             );
 
+            $this->assertEquals(['name' => 'mohamed', 'hobby' => 'diving'],
+                $redis->hmget('hash', 'name', 'hobby')
+            );
+
             $redis->flushall();
         }
     }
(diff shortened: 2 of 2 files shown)


driesvints:
I'm afraid this will be a breaking change as people would expect other keys right now.

dave2309:
@driesvints most probably, I agree.

dave2309:
@driesvints I am wondering if there is a difference between `predis` and `phpredis` regarding the values return when using `hmget` , do we know if `predis` returns the keys as well?

driesvints:
I'm not sure...

driesvints:
@tillkruss do you maybe know?

dave2309:
I guess one can use `hgetall`

dave2309:
@driesvints you can close that PR if you think it is not useful.
May I ask if you have an idea why the test does not pass?

tillkruss:
@dave2309 Do PhpRedis and Predis both return key value pairs?

driesvints:
@tillkruss that's what we're asking you 😅

dave2309:
```
    /**
     * @group connected
     * @requiresRedisVersion >= 2.0.0
     */
    public function testReturnsValuesOfSpecifiedFields(): void
    {
        $redis = $this->getClient();

        $redis->hmset('metavars', 'foo', 'bar', 'hoge', 'piyo', 'lol', 'wut');

        $this->assertSame(['bar', 'piyo', null], $redis->hmget('metavars', 'foo', 'hoge', 'unknown'));
        $this->assertSame(['bar', 'bar'], $redis->hmget('metavars', 'foo', 'foo'));
        $this->assertSame([null, null], $redis->hmget('metavars', 'unknown', 'unknown'));
        $this->assertSame([null, null], $redis->hmget('unknown', 'foo', 'hoge'));
    }
```
So it seems like (at least for tests) that `predis` would returns the values of keys and not the pair key=>value for each.

taylorotwell:

dave2309:
> > So it seems like (at least for tests) that `predis` would returns the values of keys and not the pair key=>value for each.
> 
> Yep. Hence them being removed for PhpRedis. I guess you can call array_merge them yourself afterwards?

 I would use `hgetall` ... It does returns the keys+values :)

tillkruss:
Or `array_combine()` it.

PR #46540, opened 2023-03-21
[10.x] Runs tests in parallel

nunomaduro:
This pull request improves `laravel/framework` core test suite on the CI, by running the tests in parallel.

Before this pull request:
- Locally: 34 seconds
- CI Ubuntu: 70 seconds
- CI Windows: 58 seconds

After this pull request:
- Locally: 7 seconds (a 79.4% improvement)
- CI Ubuntu: 37 seconds (a 47.1% improvement)
- CI Windows: 48 seconds (a 17.2% improvement)


composer.json
```diff
@@ -105,6 +105,7 @@
         "mockery/mockery": "^1.5.1",
         "orchestra/testbench-core": "^8.1",
         "pda/pheanstalk": "^4.0",
+        "pestphp/pest": "^2.0",
```
crynobone, line 108: This will prevent PHPUniit 9 to be installed if we need to debug any of our PHPUnit 9 integration.

Maybe add this on the build script instead? or allow v1 too

ankurk91:
php-unit already supports Parallel tests

https://github.com/paratestphp/paratest

tests/Integration/Console/CommandEventsTest.php
```diff
@@ -171,6 +172,12 @@ protected function writeArtisanScript()
     {
         $path = base_path('artisan');
 
+        while ($this->fs->exists("$path.lock")) {
+            //
```
JurianArie, line 176: Wouldn't it be better to add a sleep in these while loops?
```suggestion
        while ($this->fs->exists("$path.lock")) {
            usleep(250 * 1000);
```

tests/Integration/Console/CommandSchedulingTest.php
```diff
@@ -132,6 +133,12 @@ protected function writeArtisanScript()
     {
         $path = base_path('artisan');
 
+        while ($this->fs->exists("$path.lock")) {
+            //
```
JurianArie, line 137:
```suggestion
            usleep(250 * 1000);
```

composer.json
Jubeki: Three lines below there is already phpunit 10 required so that would already be a problem.

tests/Integration/Console/CommandEventsTest.php
nunomaduro: I will consider it, thanks!

composer.json
nunomaduro: Our test suite only runs over PHPUnit 10 at the moment.

crynobone: So are we dropping PHPUnit 9 support for Laravel 10? because this would affect Testbench dependencies and any packages supporting Laravel 10 with an older version.

Our test suite doesn't run PHPUnit 9 because we want to reduce the number of builds, but we can still use PHPUnit 9. Of course, breaking changes on patch release is unlikely but If in the future there's any new deprecation issue introduced in PHPUnit 9.x we still need to run the test suite with PHPUnit 9 (unless we do intend to remove it).

driesvints: We are not planning on removing PHPUnit v9 support for Laravel v10. Right now, we don't want to run duplicate tests on PHPUnit v9 because we'd have to juggle configurations and CI settings too much.

taylorotwell:
Sometimes the following two tests pass and sometimes they fail:

CleanShot 2023-03-24 at 08 57 59@2x

nunomaduro:
I am putting this on hold because while attempting to run --parallel testing with --order=random, I encountered more shared state than I had anticipated. This indicates that there may be flaky tests in the future.

Issue #46306, opened 2023-03-01
Exceptions in Monolog handlers cause zero exit code in artisan command


fherenius:
- Laravel Version: 9.50.2
- PHP Version: 8.0.27
- Database Driver & Version: PostgreSQL 15

### Description:

When an exception is thrown inside a Monolog handler and `ignore_exceptions => false` is set in `logging.php` then artisan commands will return with a `0` as exit code. This should not be zero.

As an example, see here a seeder that fails for some reason. There is no existing Redis connection, so the `GraylogHandler` fails to log the error and an exception is thrown. This exception is uncaught and the command exits. However, it does so with a 0 exit code.

It can be solved by setting `ignore_exceptions = true`, however I think the command should always exit with a non-zero exit code in case of an uncaught exception.

```
user@c8af5d5c3c28:/usr/src/app$ php artisan db:seed

   INFO  Seeding database.

  Database\Seeders\UsersTableSeeder .......................................................... RUNNING

   RedisException

  php_network_getaddresses: getaddrinfo failed: Temporary failure in name resolution

  at [internal]:0
      1▕

      +16 vendor frames
  17  app/Services/Logging/GraylogHandler.php:45
      Illuminate\Support\Facades\Facade::__callStatic("attempt")

      +7 vendor frames
  25  app/Exceptions/Handler.php:52
      Illuminate\Foundation\Exceptions\Handler::report(Object(RedisException))
user@c8af5d5c3c28:/usr/src/app$ echo $?
0
```

### Steps To Reproduce:
- Write an artisan command that throws an exception that is caught by your logging stack
- Ensure a Monolog handler in the stack also throws an exception
- Set `ignore_exceptions => false` in `logging.php`
- Run artisan command

driesvints:
Can you share the full stacktrace?

fherenius:
Of course! Using the following command:

```php
/**
 * Simple command to throw an exception
 */
class ThrowExceptionCommand extends Command
{
    /** @var string Signature used to call this command */
    protected $signature = "throw_exception";

    /**
     * Handle the command
     *
     * @throws Exception
     * @return int
     */
    public function handle(): int
    {
        throw new Exception("Original error thrown");
    }
}
```
With the following Monolog handler:
```php
/**
 * Example MonologHandler that throws an exception.
 */
class MonologHandlerThrowsException extends AbstractProcessingHandler
{

    /**
     * Writes the record down to the log of the implementing handler
     *
[...]

PR #46541, opened 2023-03-21
[9.x] Return non-zero exit code for uncaught exceptions

bert-w:
This is a fix for https://github.com/laravel/framework/issues/46306.

Any uncaught exceptions inside the exception handler now explicity exit with code 1, indicating an error state. A test has been included to test for the correct exit code in case of an uncaught exception and in the case of successful execution.

EDIT: thanks to Crynobone for cleaning up the testcase to work neatly with the testbench


taylorotwell:
Honestly for a simple change I would just leave all tests unchanged. Just creates more maintenance overhead than it's worth.

src/Illuminate/Foundation/Bootstrap/HandleExceptions.php
```diff
@@ -194,17 +194,23 @@ public function handleException(Throwable $e)
...
         }
 
         if (static::$app->runningInConsole()) {
             $this->renderForConsole($e);
         } else {
             $this->renderHttpResponse($e);
         }
+
+        if ($uncaught) {
+            exit(1);
```
crynobone, line 212: Wouldn't this cause Octane to exit the worker on every uncaught exception, which is not desired?

bert-w: good question. That could be the case yes.

Anyhow, there is no way to say to php "exit with exitCode=1 on your own accord (when its time to exit)". AFAIK you *must* call `exit(1)` yourself which will then still call any remaining `__destruct()` listeners and shutdown handlers.

bert-w: I moved the `exit(1)` call a few lines up now (inside the `$app->runningInConsole()` since this is a command line related issue)

bert-w:
@taylorotwell I prefer to leave the test in since the behaviour does need to be tested. If you have a faster/cleaner way to test this with less code then by all means :) 

One way or another you need to call an external process with a `php artisan ...` command; the `trait MakesArtisanScript` is the most straightforward solution to have the correct boilerplate in the test(s).

bert-w:
@crynobone do you have any idea on how to rewrite the testcase using the testbench such that I can test the exitcode for running an artisan command?

src/Illuminate/Foundation/Bootstrap/HandleExceptions.php
crynobone: Octane workers still will have `$app->runningInConsole() === true` since it is an artisan command.

crynobone:
@bert-w Can you share an artisan command example?

bert-w:
@crynobone in essence doing the following:
```php
/** @var \Illuminate\Foundation\Console\Kernel $kernel */
$exitCode = $kernel->call('throw-exception-command');
```
However since the command is meant to throw and I want to test the exitcode, this call must happen as a separate process or else the testcase itself will throw the exception.

This PR currently includes a wrapper to create an explicit `artisan` file for inside the Testbench but it is a bit convoluted that way.

taylorotwell:
Is there any way an application could enter this section of code without being in an unrecoverable / fatal error state? In other words, is this going to break **any** applications that could have possible kept executing code after this `handleException` method is called?

bert-w:
I don't see how that could be possible since that function is set using `set_exception_handler(...)`. This `handleException` function thus is only executed on an uncaught exception. PHP says "_Execution will stop after the callback is called._" (see https://www.php.net/manual/en/function.set-exception-handler.php) 

Then once `exit(1)` is called, only shutdown handlers `register_shutdown_function(...)` and destructors `__destruct()` are called (see https://www.php.net/manual/en/function.exit.php).

The only destructor being called after this in a default Laravel app is the Monolog handler itself, but whether it is called because of an `exit()` or because of an object dereference shouldnt make a difference.

macropay-solutions:
@bert-w @taylorotwell @crynobone This PR unintentionally introduced the exit(1) call in a method (handleException) used also for register_shutdown_function, making any other shutdown functions to not be executed. 

We discovered this while doing a POC for this issue https://github.com/laravel/framework/issues/58207#issuecomment-3697013293

Question is: Is the exit(1) needed in the php error case? Maybe not.

bert-w:
> @bert-w @taylorotwell @crynobone This PR unintentionally introduced the exit(1) call in a method (handleException) used also for register_shutdown_function, making any other shutdown functions to not be executed.
> 
> We discovered this while doing a POC for this issue [#58207 (comment)](https://github.com/laravel/framework/issues/58207#issuecomment-3697013293)
> 
> Question is: Is the exit(1) needed in the php error case? Maybe not.

I thought about this for a moment: according to the manual https://www.php.net/manual/en/function.register-shutdown-function.php a shutdown handler can register its own shutdown handler as well:

> Shutdown functions may also call register_shutdown_function() themselves to add a shutdown function to the end of the queue.

So perhaps instead of calling `exit(1)` we can register a new shutdown handler that gets pushed to the end of the stack and _that one_ calls `exit(1)`. But it feels somewhat odd...

marius-ciclistu:
@bert-w that works, but is it needed in the fatal error case?

PR #46548, opened 2023-03-22
[10.x] add Storage::json() method to read and decode a json file

lorenzolosa:
Laravel v10.4.0 introduced a `File::json()` method to retrieve json encoded data from a file. That's nice, but often files are accessed using `Storage` and not `File`. Here I'm adding a similar `Storage::json()`, i.e.:

```
$data = Storage::json('sample.json');
```

which is equivalent to:

```
$contents = Storage::get('sample.json');
$data = json_decode($contents, true);
```

This is really the same as #46481, but for `Storage` instead of `File`.

### DIFF
--- a/src/Illuminate/Filesystem/Filesystem.php
+++ b/src/Illuminate/Filesystem/Filesystem.php
@@ -63,14 +63,15 @@ class Filesystem
      * Get the contents of a file as decoded JSON.
      *
      * @param  string  $path
+     * @param  int  $flags
      * @param  bool  $lock
      * @return array
      *
      * @throws \Illuminate\Contracts\Filesystem\FileNotFoundException
      */
-    public function json($path, $lock = false)
+    public function json($path, $flags = 0, $lock = false)
     {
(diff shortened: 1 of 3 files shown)


ankurk91:
Can you add JSON_THROW_ON_ERROR option

taylorotwell:
Flag support added to both methods.

src/Illuminate/Filesystem/Filesystem.php
```diff
@@ -63,14 +63,15 @@ public function get($path, $lock = false)
      * Get the contents of a file as decoded JSON.
      *
      * @param  string  $path
+     * @param  int  $flags
      * @param  bool  $lock
      * @return array
      *
      * @throws \Illuminate\Contracts\Filesystem\FileNotFoundException
      */
-    public function json($path, $lock = false)
```
chrysanthos, line 71: Isn't this a breaking change? The $lock parameter is now third instead of second?

driesvints:
I've sent in a PR to revert this one.

PR #46550, opened 2023-03-22
[10.x] Add support for encrypting date (and any other types) fields

wouterrutgers:
This PR adds support for encrypting any standard cast type. Currently, you have to write custom getters and setters to use standard cast types *and* encryption. With this PR, you can use encryption on columns and also have the standard casts like date, boolean, float, etc.

You can specify the cast type via:

    protected $casts = [
        'date' => 'encrypted:date',
        'bool' => 'encrypted:bool',
        'json' => 'encrypted:json',
        'float' => 'encrypted:float',
        // etc
    ];

Edit:

As it isn't obvious by just the files changed. The logic is semi-already there: https://github.com/laravel/framework/pull/46550/files#diff-59d24f1a8f0dd1e51ee7dffc8778133711634e928ecef4a91b63fc3851cb1579L739-L742.

It just had a hard check on what cast types are allowed:
  
    'encrypted:array',
    'encrypted:collection',
    'encrypted:json',
    'encrypted:object',


ankurk91:
I wish if we can chain multiple cast on the same field.

taylorotwell:

taylorotwell:
I would suggest writing custom casts if you need more custom cast logic than what is already provided by Laravel.

PR #46552, opened 2023-03-22
JSON encoding / decoding customization for JSON based cass

taylorotwell:
This PR allows the customization of `json_encode` and `json_decode` for the `array` and `collection` casts, as well as `AsArrayObject`, `AsCollection`, `AsEnumArrayObject`, etc.

```php
use Illuminate\Database\Eloquent\Casts\Json;

Json::encodeUsing(fn ($value) => json_encode($value));

Json::decodeUsing(fn ($value, $associative) => json_decode($value, $associative));
```

This customization would typically be done in a service provider.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Casts/Json.php
+++ b/src/Illuminate/Database/Eloquent/Casts/Json.php
@@ -1,0 +1,52 @@
+<?php
+
+namespace Illuminate\Database\Eloquent\Casts;
+
+class Json
+{
+    /**
+     * The custom JSON encoder.
+     *
+     * @var callable|null
+     */
+    protected $encoder;
+
+    /**
+     * The custom JSON decode.
+     *
+     * @var callable|null
+     */
+    protected $decoder;
+
+    /**
+     * Encode the given value.
+     */
+    public static function encode(mixed $value): mixed
+    {
+        return isset(static::$encoder) ? (static::$encoder)($value) : json_encode($value);
+    }
+
+    /**
+     * Decode the given value.
+     */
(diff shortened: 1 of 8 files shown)


src/Illuminate/Database/Eloquent/Casts/Json.php
```diff
@@ -0,0 +1,54 @@
...
+     * @var callable|null
+     */
+    protected $encoder;
+
+    /**
+     * The custom JSON decode.
+     *
+     * @var callable|null
+     */
+    protected $decoder;
```
driesvints, line 19: @taylorotwell I guess these need to become static?

src/Illuminate/Database/Eloquent/Casts/Json.php
```diff
@@ -0,0 +1,54 @@
+<?php
+
+namespace Illuminate\Database\Eloquent\Casts;
+
+class Json
```
gocanto, line 5: May we be able to have this in the Support package instead?

taylorotwell: No. That would imply it affects JSON serialization **everywhere**. Having it here demonstrates it is for **casting only**.

morloderex:
What if 2 columns within the same project needs different json encoding / decoding? This customization object seems to be global? Would that require a custom cast?

taylorotwell:
@morloderex I guess that require a custom cast. My thinking was that if you're using something like `JSON_UNESCAPED_UNICODE` you probably want that on all your JSON casts throughout your application.

PR #46553, opened 2023-03-23
Revert "[10.x] add Storage::json() method to read and decode a json file"

driesvints:
Reverts laravel/framework#46548

This PR changes the sequence of parameters of the `json` method and thus causing a breaking change for anyone using the `lock` param.


lorenzolosa:
I see your point, but I would suggest reverting only the addition of the `$flags` parameter instead of the whole PR.

taylorotwell:
This hasn't been documented or released yet.

PR #46560, opened 2023-03-23
[9.x] Fix getDirty method when using `AsEnumArrayObject` / `AsEnumCollection`

JaZo:
Same fix as #38869, but for the `AsEnumArrayObject` / `AsEnumCollection` casts introduced in #45621.


taylorotwell:
Please send this to 10.x branch.

PR #46563, opened 2023-03-23
[10.x] Add a possibility to set a custom on_stats function for the Http Facade

Katalam:
## Idea
I was working on a package that logs the transfer stats of the guzzle requests. To support the Http Facade of the framework I intended to do the same here, but the framework overwrites any custom functions, the idea was to add a possiblity to have both.

## Functionality
In the `sendRequest` Method, we check if we have a user function set and then use a closure to do the default and necessary step to add the stats to the response object and afterward call the user function.

## Disclaimer
The test is not good, but I wasn't able to fake a transferStats Object, and it is my first pull request on a big project so I'm sorry to bother in the first place

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Http\Client;
 
+use Closure;
 use Exception;
 use GuzzleHttp\Client;
 use GuzzleHttp\Cookie\CookieJar;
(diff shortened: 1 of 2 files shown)


driesvints:
It seems the tests here still fail. Please resend this with passing tests and maybe ask for help with it on one of these channels below. Thanks!

- [Laracasts Forums](https://laracasts.com/discuss)
- [Laravel.io Forums](https://laravel.io/forum)
- [StackOverflow](https://stackoverflow.com/questions/tagged/laravel)
- [Discord](https://discordapp.com/invite/KxwQuKb)
- [Larachat](https://larachat.co)
- [IRC](https://web.libera.chat/?nick=laravelnewbie&channels=#laravel)

PR #46564, opened 2023-03-23
[10.x] Pass column to newUniqueId method

hosmelq:
I am creating a trait similar to `HasUlids` that needs to check the uniqueness in the database before creating the record. Therefore, I need to access the `$column` variable, this PR passes the column variable to the `newUniqueId` method. Otherwise, I will need to replicate almost the same code in order to achieve this.


taylorotwell:
It's a breaking change to anyone overwriting this method.

hosmelq:
Are you open to call this with something like `call_user_func`?

PR #46569, opened 2023-03-24
[10.x] Add a possibility to set a custom on_stats function for the Http Facade

Katalam:
## Idea
I was working on a package that logs the transfer stats of the guzzle requests. To support the Http Facade of the framework I intended to do the same here, but the framework overwrites any custom functions, the idea was to add a possiblity to have both.

## Functionality
In the `sendRequest` Method, we check if we have a user function set and then use a closure to do the default and necessary step to add the stats to the response object and afterward call the user function.

## Disclaimer
It is my first pull request on a big project, so I'm sorry to bother in the first place.

Not sure if it is appropriate to tag driesvints, but he declined the first PR because the test were failing, but requested a resend of it. This should pass (hopefully)

### DIFF
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Http\Client;
 
+use Closure;
 use Exception;
 use GuzzleHttp\Client;
 use GuzzleHttp\Cookie\CookieJar;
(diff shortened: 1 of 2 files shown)


src/Illuminate/Http/Client/PendingRequest.php
```diff
@@ -958,11 +959,20 @@ protected function sendRequest(string $method, string $url, array $options = [])
...
+            $this->transferStats = $transferStats;
+        };
+
+        if (data_get($this->options, 'on_stats') instanceof Closure) {
+            $onStats = function ($transferStats) use ($onStats) {
+                $onStats($transferStats);
+                call_user_func($this->options['on_stats'], $transferStats);
+            };
+        }
+
```
bert-w, line 972: Something like this is probably more logical (e.g. your callback can modify the transferStats. It will be assigned to `$this->transferStats` after completion):
```php
$onStats = function ($transferStats) {
    if (($callback = $this->options['on_stats'] ?? false) instanceof Closure) {
        $transferStats = $callback($transferStats);
    }
    $this->transferStats = $transferStats;
};
```

Katalam: Isn't this force the callback to return the transferStats? In my implementation we always have the transferStats assigned to the Response Object, so it wasn't changeable, but in the recommendation if I do not return the transferStats in my custom callback I don't have it right?

I changed it a little bit:
```php
$onStats = function ($transferStats) {
    if (($callback = $this->options['on_stats'] ?? false) instanceof Closure) {
        $transferStats = $callback($transferStats) ?? $transferStats;
    }
    $this->transferStats = $transferStats;
};
```
This should be better if we don't want to force a return on the custom callback

taylorotwell:
I'm confused how this isn't already possible? We merge the options. Can't you already set an `on_stats` option?

Issue #46568, opened 2023-03-24
Job delay doesn't work unless you use the `Dispatchable` trait


SjorsO:
- Laravel Version: v9.36.4
- PHP Version: 8.1
- Database Driver & Version: not relevant

### Description:
Delaying a job like this works:
```php
SomeJob::dispatch()->delay(now()->addMinute());
```

But this doesn't work, it ignores the delay and instantly processes the job:

```php
$job = new SomeJob();

$job->delay(now()->addMinute());

Queue::push($job);
```

As far as I can tell, this happens because [`Queue::push()`](https://github.com/laravel/framework/blob/321c1365c3c051a65745af1d59db1a0a32c94a69/src/Illuminate/Queue/RedisQueue.php#L138) doesn't pass down the `$job->delay` to the [`EnqueueUsing` method](https://github.com/laravel/framework/blob/321c1365c3c051a65745af1d59db1a0a32c94a69/src/Illuminate/Queue/Queue.php#L302).

### Steps To Reproduce:
1. Make sure your queue connection supports delay (so anything except `sync`)
2. Make a job that implements `ShouldQueue`
3. Dispatch the job with a delay like shown above
4. Watch the job get processed instantly without a delay

taylorotwell:
I think this is pretty much expected behavior. You are bypassing the Bus which provides all of that stuff. You would need to use `Queue::later` if you want to hit the queue directly.

PR #46582, opened 2023-03-24
[9.x] Fix Job delay property when dispatched through facade + `data_get()` closure fix

bert-w:
Fix for https://github.com/laravel/framework/issues/46568 and an improvement for https://github.com/laravel/framework/pull/46505.

This also includes a helpful improvement for the `data_get()` global helper, since for PHP 8.0, trying to access properties on a closure (closures are still `is_object()`) will cause it to error. This PR makes it so that when you now call `data_get($closure, 'someProperty')`, it will return null (or some other default).

Sadly I couldn't make any tests for the `enqueueUsing` method to test the correct `$delay` property, since the method calls a closure further on within the function which is different for each and every Queue driver.

### DIFF
--- a/src/Illuminate/Collections/helpers.php
+++ b/src/Illuminate/Collections/helpers.php
@@ -76,7 +76,7 @@ if (! function_exists('data_get')) {
 
             if (Arr::accessible($target) && Arr::exists($target, $segment)) {
                 $target = $target[$segment];
-            } elseif (is_object($target) && isset($target->{$segment})) {
+            } elseif (! $target instanceof Closure && is_object($target) && isset($target->{$segment})) {
                 $target = $target->{$segment};
             } else {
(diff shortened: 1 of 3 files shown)


SjorsO:
Have you tested that this PR? [The `enqueueUsing` method passes down the delay to the callback](https://github.com/laravel/framework/blob/40ad58b58906b178b542caf27178542b947a9de9/src/Illuminate/Queue/Queue.php#L308), but it seems to me like [the callback doesn't actually use that argument](https://github.com/laravel/framework/blob/40ad58b58906b178b542caf27178542b947a9de9/src/Illuminate/Queue/RedisQueue.php#L139-L141), so I have a feeling this fix doesn't work.

taylorotwell:

PR #46586, opened 2023-03-24
[10.x] Set a custom schemaPath for TestCase

cosmastech:
## Problem
We are building a V3 of our application and it doesn't make sense to copy over all of the migrations. We are leveraging the schema dump functionality.

However, by default, migrations attempt to look for a schema that matches the database connection name. This means we would have to maintain two SQL files (one for the `mysql` and one for `mysql_testing`). I originally overrode the `migrateFreshUsing()` method within our application's base TestCase to force it to use the name of our default connection:
```php
    protected function migrateFreshUsing(): array
    {
        $migrateFreshOptions = $this->baseMigrateFreshUsing();

        if ($schemaPath = $this->getSchemaPath()) {
            $migrateFreshOptions['--schema-path'] = $schemaPath;
        }

        return $migrateFreshOptions;
    }

    /**
     * overrides the default schema path checking so that two copies of
     * the schema (with the same data) do not need to be maintained
     *
     * @return string
     */
    protected function getSchemaPath(): string
    {
        return database_path('schema/mysql-schema.dump');
    }
```

I figured if we needed this functionality, maybe others could benefit too?

But I think more useful is the fact that a common test SQL file could be used within tests. Our application has over 1000 test cases, almost all of which touch the database. We could put all of our common seeders into a schema dump which would shorten the dev-test feedback loop.

## Usage
Within a class that extends Tests\TestCase, set a schema path. For example:

```php
use Tests\TestCase;

class UserTest extends TestCase
{
    use RefreshDatabase;

    protected $schemaPath = 'lot_of_users_seeded.dump';

    public function testDoesUserExist()
    {
         $this->assertExists(User::first());
    }
}
```

## Questions
I think that there should be a test case that actually confirms the database schema is loaded. **Where should this test case live and what TestCase it should extend from**? I'm not sure what namespace of directories it would belong to (since it looks like the test harness uses that to determine which runners to use). If someone could steer me in the right direction, I would like to add that.

### DIFF
--- a/src/Illuminate/Foundation/Testing/Traits/CanConfigureMigrationCommands.php
+++ b/src/Illuminate/Foundation/Testing/Traits/CanConfigureMigrationCommands.php
@@ -2,8 +2,6 @@
 
 namespace Illuminate\Foundation\Testing\Traits;
 
-use Illuminate\Support\Facades\File;
-
 trait CanConfigureMigrationCommands
 {
     /**
@@ -85,12 +83,12 @@ trait CanConfigureMigrationCommands
         }
 
         // check if the path is simply the filename of the schema
-        if (file_exists($schemaPathWithinDatabaseMigrationsPath = database_path('schema/' . $schemaPath))) {
+        if (file_exists($schemaPathWithinDatabaseMigrationsPath = database_path('schema/'.$schemaPath))) {
             return $schemaPathWithinDatabaseMigrationsPath;
         }
 
         // check if the path provided is relative to the root database path
-        if (file_exists(($schemaPathWithinDatabasePath = database_path($schemaPath)))) {
+        if (file_exists($schemaPathWithinDatabasePath = database_path($schemaPath))) {
             return $schemaPathWithinDatabasePath;
         }
 
--- a/src/Illuminate/Foundation/Testing/Traits/CanConfigureMigrationCommands.php
+++ b/src/Illuminate/Foundation/Testing/Traits/CanConfigureMigrationCommands.php
@@ -2,6 +2,8 @@
 
 namespace Illuminate\Foundation\Testing\Traits;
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
The docs talk about keeping two schema files - which I think is generally fine. They don't really have to be "maintained", they are just generated. Is there a big issue with this approach?

https://laravel.com/docs/10.x/migrations#squashing-migrations

cosmastech:
> The docs talk about keeping two schema files - which I think is generally fine. They don't really have to be "maintained", they are just generated. Is there a big issue with this approach?
> 
> https://laravel.com/docs/10.x/migrations#squashing-migrations

@taylorotwell It's more error-prone for a team of developers to keep a two copies of a file in their repo.  As we are still in the middle of a V2 to V3 transition, there are still migrations being made within the V2 codebase. That means that those changes need to be squashed and the schema dump put in two places.

It's an inconvenience at worst: like I said, I already wrote a workaround for this for our codebase. However, it took a bit to figure out how to do it and seems like it would be beneficial to be about to override the schema options, even if it was as easy as allowing for a method like `additionalMigrateFreshOptions()`

```php
    protected function migrateFreshUsing()
    {
        $seeder = $this->seeder();
        $additionalMigrateFreshOptions = [];
        if (method_exists($this, 'additionalMigrateFreshOptions')) {
            $additionalMigrateFreshOptions = (array) $this->additionalMigrateFreshOptions();
        }

        return array_merge(
            [
                '--drop-views' => $this->shouldDropViews(),
                '--drop-types' => $this->shouldDropTypes(),
            ],
            $seeder ? ['--seeder' => $seeder] : ['--seed' => $this->shouldSeed()],
            $additionalMigrateFreshOptions,
        );
    }
```

PR #46591, opened 2023-03-25
[9.x] Fix job delay override + `data_get()` improvement to deal with closures

bert-w:
Fix for https://github.com/laravel/framework/issues/46568 and an improvement for https://github.com/laravel/framework/pull/46505. Minor changes have been added for all queue drivers (Database, SQS, Redis, Beanstalkd) so a `$delay` property can be passed.

I would say this still is an issue on the basis of:
```php
$job = new MyTestJob;
$job->delay(now()->addMinute());
Queue::push($job);
```

The code sample above should delay the job obviously (which before it didnt). Setting `public $delay = 60;` in some job class should now always make sure the job is delayed 60 seconds (unless you override it using `Queue::later(30, new MyTestJob)`) which imho is also expected behaviour.

Trace/Explanation: Everything goes through the `Illuminate\Queue::enqueueUsing(...)` which in turn calls the `$callback` which is its fifth argument. This callback differs per queue driver and has a `$queue` param.

### DIFF
--- a/src/Illuminate/Collections/helpers.php
+++ b/src/Illuminate/Collections/helpers.php
@@ -76,7 +76,7 @@ if (! function_exists('data_get')) {
 
             if (Arr::accessible($target) && Arr::exists($target, $segment)) {
                 $target = $target[$segment];
-            } elseif (is_object($target) && isset($target->{$segment})) {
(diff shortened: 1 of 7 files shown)


taylorotwell:
We never document delaying queued jobs this way or say that it will work.

PR #46593, opened 2023-03-25
[10.x] Fix route resource in groups

NeonGC:
@taylorotwell, added tests (#46547)

---

Hi all!

I wanted to use a route resource inside an existing group without an additional name, however this was not possible, so I added the appropriate code.

The reason for this problem was that the ResourceRegistrar relied in creating the route name and the name of the substitution parameters on the received empty name and did not take it from the name of the last group stack, if one exists.

Please, merge it also to [10.x] branch.


taylorotwell:

PR #46595, opened 2023-03-25
[10.x] Allow Blueprint to use an Enum case for column default

cosmastech:
## Problem

I need to create a column with a default value that is sourced from an enum.

```php
use App\Enums\Access\UserRegistrationRequestDispositionEnum;
use App\Models\User;

{
    public function up(): void
    {
        Schema::create('user_registration_requests', function (Blueprint $table) {
            $table->id();
            $table->foreignIdFor(User::class)->constrained()->cascadeOnDelete();
            $table->tinyInteger('disposition')->default(UserRegistrationRequestDispositionEnum::PENDING);
            $table->timestamps();
        });
    }

    public function down(): void
    {
        Schema::dropIfExists('user_registration_requests');
    }
};
```

This throws an exception.

image

Enums do not have a `__toString()` method, so the attempt to cast fails.

The current workaround is use the enum's value:
```php
$table->tinyInteger('disposition')->default(UserRegistrationRequestDispositionEnum::PENDING->value);
```

## Solution
Initially I thought to rely on the `InteractsWithDictionary@getDictionaryKey()` because it already handles this UnitEnum/BackedEnum juggling, however, I ended up splitting the enum juggling to a separate function.

## Other Thoughts
Maybe this should be a global helper function (`getEnumValue()`) instead of embedded within a trait?

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithDictionary.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithDictionary.php
@@ -37,7 +37,7 @@ trait InteractsWithDictionary
      * Get a value from an enum case if it is backed, otherwise use the case name.
      *
      * @param  UnitEnum  $attribute
-     * @return mixed
+     * @return string | int
      */
     protected function getEnumValue(UnitEnum $attribute)
     {
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithDictionary.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithDictionary.php
@@ -24,12 +24,23 @@ trait InteractsWithDictionary
             }
 
             if ($attribute instanceof UnitEnum) {
-                return $attribute instanceof BackedEnum ? $attribute->value : $attribute->name;
+                return $this->getEnumValue($attribute);
             }
 
             throw new InvalidArgumentException('Model attribute value is an object but does not have a __toString method.');
         }
 
         return $attribute;
+    }
+
+    /**
+     * Get a value from an enum case if it is backed, otherwise use the case name.
+     *
+     * @param  UnitEnum  $attribute
+     * @return mixed
+     */
+    protected function getEnumValue(UnitEnum $attribute)
+    {
+        return $attribute instanceof BackedEnum ? $attribute->value : $attribute->name;
     }
 }
--- a/src/Illuminate/Database/Schema/Grammars/Grammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/Grammar.php
@@ -7,6 +7,7 @@ use Doctrine\DBAL\Schema\TableDiff;
 use Illuminate\Contracts\Database\Query\Expression;
 use Illuminate\Database\Concerns\CompilesJsonPaths;
 use Illuminate\Database\Connection;
+use Illuminate\Database\Eloquent\Relations\Concerns\InteractsWithDictionary;
 use Illuminate\Database\Grammar as BaseGrammar;
 use Illuminate\Database\Schema\Blueprint;
 use Illuminate\Support\Fluent;
@@ -16,6 +17,7 @@ use RuntimeException;
 abstract class Grammar extends BaseGrammar
 {
     use CompilesJsonPaths;
+    use InteractsWithDictionary;
 
     /**
      * The possible column modifiers.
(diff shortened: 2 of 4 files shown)


abenerd:
I feel this would be a dangerous thing to do as if you remove the Enum the migration would not work anymore or if you update the Enum you'd get unpredictable results in your migration and table definition.

cosmastech:
@abenerd thanks for your comment!

> I feel this would be a dangerous thing to do as if you remove the Enum the migration would not work anymore or if you update the Enum you'd get unpredictable results in your migration and table definition.

While I can see a case being made for that, in projects I work on:

* If a column has a set of predefined states, then that goes into a PHP Enum.
* The model references that Enum in the `$casts` array
* This maintains consistency between the Model and Migration that both reference a common BackedEnum

Furthermore, if you don't want to use a PHP Enum as the default, you don't have to! But to me, it seems like it's more a pain to have magic strings in migrations _just in case_ an enum changes.

Also, if you want to reference the value as you would have previously with an Enum case's value, you still can.

This is just one of those little annoying "gotyas" I run into when setting up migrations and I figured I would submit this PR to make the developer experience better. If you don't use Enums in your migrations, there's nothing breaking here 😅

abenerd:
> * If a column has a set of predefined states, then that goes into a PHP Enum.
> * The model references that Enum in the $casts array

I agree with those points here and I do the same.
When writing migrations I try not to add any dependencies to them so they are easily reproducible.
If at some point you update the Enum in your codebase you will have to create another migration anyway to represent this in the database. 
But anyway this is not me blocking this but just the opinion of a random stranger on the internet.

src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithDictionary.php
```diff
@@ -24,12 +24,23 @@ protected function getDictionaryKey($attribute)
...
         }
 
         return $attribute;
     }
+
+    /**
+     * Get a value from an enum case if it is backed, otherwise use the case name.
+     *
+     * @param  UnitEnum  $attribute
+     * @return mixed
```
abenerd, line 40:
```suggestion
     * @return string | int
```
Currently Enums can only be backed by a `string` or an `int` so we can be more precise in the return type here I think.

Rizky92:
How do you handle the enum update? In case of the enum class was changed, do you create a new migration or modfiy the existing one?

cosmastech:
> How do you handle the enum update? In case of the enum class was changed, do you create a new migration or modfiy the existing one?

Wouldn't this be the same thing if the default value changed? How would you handle that?

1. Write a new migration to change the column default (if the previous migration already made it to production)
2. Manually change the production database column's default value in order to match the current case's value and leave the original migration as is
3. If the change hasn't made it to production, then I'm not sure what the issue would be?

In my experience/opinion, enums are treated like consts. They do not change, and if they do, there's usually a discussion as to the ramifications. Often times, for better or for worse, they end up not changing because of the downstream headaches it might cause. (named slightly weird, should've started the value at 1 instead of 0, etc)

Outside of early bootstrapping, I don't think I have ever had the need to change an Enum case significantly. And luckily, with an IDE, when I make a change, I can have it changed globally for the project. Just right click > Rename > type the new name > everywhere that references it has been updated.

There's nothing in this MR that is requiring a user to use an Enum for a default value. You can still use magic strings, booleans, integers, consts from classes, consts defined via `define()`, objects, `config()` values, the result of a function, etc.

My thinking is if I can write this code: 
```php
use App\Enums\User\UserStatusEnum;

class User extends Model
{
    protected $casts = [
        'status' => UserStatusEnum::class,
    ];

    protected $attributes = [
        'status' => UserStatusEnum::Pending,
    ];
}
```

Then why not allow me to use an Enum in the migration definition too? For me, it keeps everything very, very clean.

dennisprudlo:
I think the main reason why this PR is so *controversial* is that it adds some maintenance burden for laravel for something that is considered "bad practice" by a lot of developers (saying this without any valuation).

I think "Flyway" (a tool for database migrations) event takes it a step further and generates a migration file hash and stores it with the migration entry. Meaning you can't even change a migration after running because the hashes wouldn't match (This was happening in a Spring Boot project I worked with at least). Simply to prevent users from altering the chain of state changes in a migration.

In theory your migration state can change without touching the migration itself and that is what people here are trying to prevent. It is still possible, you said it yourself, you can use Constants, Model variables, etc. or simply `UserStatusEnum::Pending->value` but I agree with "let's not open that door".

So I don't think the added value of being able to omit `->value` justifies this implementation, but that's just my humble opinion 🙂

taylorotwell:

PR #46604, opened 2023-03-27
Return `View` when `fragmentIf` condition isn't met

leettaylor:
When using `fragmentIf` (or `fragmentsIf`) and ensuring that the condition isn't met, you're unable to use any `Illuminate\Testing\TestView` helpers as a `View` isn't returned.

I believe this is because the `fragmentIf` is currently returning the view content as a string instead of the `View` itself.

**Example of test which breaks:**

```
$this->get(route('example-route'))
    ->assertViewIs('example.index');
```

_Output:_
```
The response is not a view.
```

### DIFF
--- a/src/Illuminate/View/View.php
+++ b/src/Illuminate/View/View.php
@@ -114,7 +114,7 @@ class View implements ArrayAccess, Htmlable, ViewContract
             return $this->fragment($fragment);
         }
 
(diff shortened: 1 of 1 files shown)


taylorotwell:
Can't make a breaking change on a patch release.

PR #46605, opened 2023-03-27
[10.x] allow insert..select (insertUsing()) to have empty $columns

jonnott:
Enables queries which insert *all* columns from the source table into the destination table, e.g. "insert into users select * from users”, when the table schemas exactly match.

(alter testPreservedAreAppliedByInsertUsing() to not conflict with new behaviour)

### DIFF
--- a/src/Illuminate/Database/Query/Grammars/Grammar.php
+++ b/src/Illuminate/Database/Query/Grammars/Grammar.php
@@ -1062,7 +1062,13 @@ class Grammar extends BaseGrammar
      */
     public function compileInsertUsing(Builder $query, array $columns, string $sql)
     {
-        return "insert into {$this->wrapTable($query->from)} ({$this->columnize($columns)}) $sql";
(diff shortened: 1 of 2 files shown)


taylorotwell:
I wonder if we should add a new helper method for this so you don't have to pass an empty array?

taylorotwell:
Updated to also allow `['*']` for columns.

jonnott:
> I wonder if we should add a new helper method for this so you don't have to pass an empty array?

@taylorotwell I think you can't specify * for the INSERT part of an INSERT..SELECT, MySQL (at least) doesn't like it..

e.g. `insert into users (*) select * from users` will result in:

`Query 1 ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '*) select * from users' at line 1`

Also, as it checks empty() on $columns, you can also pass null instead of an empty array, which might feel neater than the empty array.

PR #46608, opened 2023-03-27
Allow array for path

mylesduncanking:
Allow an array to be passed for the path parameter which is then imploded with DIRECTORY_SEPERATOR so paths can be correctly structured without defining "DIRECTORY_SEPERATOR" in the path manually.

For example, instead of 
```
$temp = storage_path('app' . DIRECTORY_SEPARATOR . 'reports' . DIRECTORY_SEPARATOR . 'tmp' . DIRECTORY_SEPARATOR . $this->client->name . '.zip');
```

You could do:
```
$temp = storage_path(['app', 'reports', 'tmp', $this->client->name . '.zip']);
```


taylorotwell:

PR #46611, opened 2023-03-28
[11.x] Add support for first-class callable syntax in `App::call(...)`

Persaeus:
This PR addresses https://github.com/laravel/framework/issues/46587: contextual binding isn't resolved via `App::call(...)` when called using [first class callable syntax](https://www.php.net/manual/en/functions.first_class_callable_syntax.php) (`$object->methodName(...)`).

Aside from style fixes, the main commits are:

1. Addition of passing tests demonstrating that contextual binding _does_ work when using array syntax: `App::call([$object, 'methodName'])`
2. Addition of failing tests demonstrating that contextual binding is broken when using first class syntax: `App::call($object->methodName(...))`
3. Proposed fix to make the tests pass.

This would be my first contribution to this part of the framework. I have done my best to make the style conform with the surrounding code, but it may not be perfect.

I welcome any suggestions or instructions from the maintainers.

### DIFF
--- a/src/Illuminate/Container/Container.php
+++ b/src/Illuminate/Container/Container.php
@@ -648,8 +648,8 @@ class Container implements ArrayAccess, ContainerContract
     {
         $pushedToBuildStack = false;
 
-        if (is_array($callback) && ! in_array(
-            $className = (is_string($callback[0]) ? $callback[0] : get_class($callback[0])),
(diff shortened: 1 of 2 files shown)


taylorotwell:
Should be on 10.x branch. I have merged it into that branch.

taylorotwell:
Thanks!

PR #46612, opened 2023-03-28
[10.x] Add getContainer method to router

iamgergo:
The Router instance's container can be set at any point using the `setContainer` method. It would be nice to have access to the container instance that is set to the router.

```php
Route::group(function (Router $router) {
    // Do something using the container
    $router->getContainer()->...; // it might be more convinient than using the App facade

    $router->get(...);
});
```


GrahamCampbell:
Using services as a service container is always a bad idea. Convenience is not to be confused with spaghetti jank. ;)

PR #46615, opened 2023-03-28
[10.x] Add new acronym method to Str for retrieving acronyms from strings

koenhendriks:
## Description
This Pull Request adds a new method to the `Str` helper class called acronym that retrieves an acronym consisting of the first letter of each word in a given string, concatenated without capitalization (other helper methods can be chained for this).

The acronym method can be used in a variety of contexts, such as generating abbreviations for longer phrases or creating short identifiers from longer names.

## Usage
To use the acronym method, simply call the method on the `Str` class and pass in a string as the argument:

```php
echo Str::acronym('Laravel is freaking everywhere');
// Output: "Life"
```
The acronym method splits the input string into words using spaces as a delimiter, retrieves the first letter of each word, and concatenates the resulting letters into a single string without capitalization.

An optional delimiter can be given which allows for more flexibility in how the resulting acronym is formatted.

```php
echo Str::acronym('Laravel is freaking everywhere', '.');
// Output: "L.i.f.e."
```

## Testing
This Pull Request includes a unit test for the acronym method to ensure that it works correctly and doesn't break any existing functionality.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -69,6 +69,18 @@ class Str
     }
 
     /**
+     * Get an acronym consisting of the first letter of each word in a given string
+     */
+    public static function acronym(string $string): string
+    {
+        $acronym = '';
+        foreach (explode(' ', $string) as $word) {
+            $acronym .= $word[0];
+        }
+        return $acronym;
+    }
+
+    /**
      * Return the remainder of a string after the first occurrence of a given value.
      *
      * @param  string  $subject
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -74,7 +74,7 @@ class Str
     public static function acronym(string $string): string
     {
         $acronym = '';
-        foreach (explode(' ', $string) as $word) {
+        foreach (preg_split('/\s+/', $string) as $word) {
             $acronym .= $word[0];
         }
 
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -69,7 +69,7 @@ class Str
     }
 
     /**
-     * Get an acronym consisting of the first letter of each word in a given string
+     * Get an acronym consisting of the first letter of each word in a given string.
      */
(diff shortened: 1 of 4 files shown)


src/Illuminate/Support/Str.php
```diff
@@ -68,6 +68,19 @@ public static function of($string)
...
     }
 
+    /**
+     * Get an acronym consisting of the first letter of each word in a given string.
+     */
+    public static function acronym(string $string): string
+    {
+        $acronym = '';
+        foreach (explode(' ', $string) as $word) {
+            $acronym .= $word[0];
```
cosmastech, line 78: A warning will be raised here if you pass a string with multiple spaces. `Hello  World` for instance

cosmastech: Should this account for newlines/return carriages?

Should all punctuation be stripped? `Hello World! ! !` would return `HW!!`

koenhendriks: I've updated the method and added additional test cases for this.

taylorotwell:
You can macro this into your own project.

src/Illuminate/Support/Str.php
koenhendriks: I've changed it to `/[^a-zA-Z]+/`  which excludes any non-alphabetic characters.

koenhendriks:
@taylorotwell I know I can, and I've done so in multiple projects. That's one of the reasons why I created this PR.

While it's true that you can create your own macro for generating acronyms in your own Laravel project, adding this feature to the core `Str` helper class in Laravel can provide benefits in terms of consistency, code reuse, and community contributions.

henzeb:
@koenhendriks, I suggest you make this into a package in that case.

koenhendriks:
@henzeb Seemed a bit overkill for such a small feature, but created it anyway. Should be live on packagist and can be used with

`composer require koenhendriks/laravel-str-acronym`

For future people landing here from search engines or github; If you want to use this functionality checkout https://github.com/koenhendriks/laravel-str-acronym

PR #46620, opened 2023-03-28
[10.x] Correct password generation

aktuba:
The current implementation of password generation does not work correctly:

- allows spaces around the edges and a password consisting entirely of spaces;
- when specifying several parameters (letters, numbers, special characters), most people (according to my survey) expect that they will be present in the password, but this is not so;
- there is no way to expand / compress the allowed sets of letters, numbers or special characters.

I propose an implementation that does not break the current interface, but solves the problems described above.

### DIFF
--- a/src/Illuminate/Support/Password.php
+++ b/src/Illuminate/Support/Password.php
@@ -1,0 +1,241 @@
+<?php
+
+namespace Illuminate\Support;
+
+use Exception;
+use Illuminate\Support\Exceptions\PasswordInvalidArgumentsException;
+use Illuminate\Support\Exceptions\PasswordInvalidLengthArgument;
+
+class Password
+{
+    /**
+     * Password length.
+     *
+     * @var int
+     */
+    protected int $length = 32;
+
+    /**
+     * Using letters.
+     *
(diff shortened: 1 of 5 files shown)


taylorotwell:
Only interested in solving the first bullet point of your list.

src/Illuminate/Support/Password.php
```diff
@@ -0,0 +1,241 @@
...
+        while ($chars->count() < $this->length) {
+            $chars = $chars->merge($combination);
+        }
+
+        return $chars
+            ->shuffle()
+            ->take($this->length)
+            ->when($this->withLetters, fn ($c) => $c->splice(1)->merge($letters[array_rand($letters)]))
+            ->when($this->withNumbers, fn ($c) => $c->splice(1)->merge($numbers[array_rand($numbers)]))
+            ->when($this->withSymbols, fn ($c) => $c->splice(1)->merge($symbols[array_rand($symbols)]))
```
valorin, line 224: I know this has been closed already, but it's worth pointing out that `array_rand()` isn't cryptographically secure and shouldn't be used for password generation.

PR #46621, opened 2023-03-28
[10.x] Set container instance on session manager

smortexa:
Related PR: laravel/octane#669


smortexa:
In the case of Octane, The `SessionManager` gives the old instance of `$container` and `$config` to session drivers and session handlers like `CacheBasedSessionHandler`. Even though the configuration and cache manager are being refreshed, the session manager still uses the old instance of `$config` and `$contanier`.

In my case, I needed to switch between drivers on the fly, and the problem was that the session manager didn't receive a new container instance and when I wanted to login with Fortify, the old cache driver was being used.

PR #46624, opened 2023-03-29
[10.x] Add support for `Attribute` union return types

stevelacey:
Attributes currently require `Attribute` be the only return type of the model method, supporting union types allows developers to replicate relationship-ish functionality in models.

Here's my use case, I have a attribute called `scholarships`, and I want to be able to optionally call the method and pass arguments in to get a subset of my dataset, similar to how you might use a `BelongsToMany`:

```php
<?php

namespace App\Concerns;

use App\Support\Dataset;

trait HasScholarships
{
    public function scholarships(int $first_year = null, int $last_year = null): Attribute|Collection
    {
        if (func_get_args()) {
            return Dataset::scholarships(...func_get_args(), school: $this);
        }

        return Attribute::make(fn () => Dataset::scholarships(school: $this));
    }
}
```

This allows me to continue using `$school->scholarships` as I already am currently, and adds support for using the method directly to retrieve a subset of the data via `$school->scholarships(2019, 2021)`.

The idea here is it gives me a fluent api similar to how relationships work where the attribute returns the data and calling the method returns the query builder.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -38,6 +38,7 @@ use LogicException;
 use ReflectionClass;
 use ReflectionMethod;
 use ReflectionNamedType;
+use ReflectionUnionType;
 
 trait HasAttributes
 {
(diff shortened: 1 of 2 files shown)


bert-w:
I like this idea, i've had the usecase as well to provide arguments to a custom attribute as well as keep the standard property call. Imho this does require some extra tests though (retrieving as attribute a property + retrieving attribute as a function call)

taylorotwell:
Agree it feels like we need more tests here.

taylorotwell:
Please mark as ready for review when more tests are added.

stevelacey:
Not sure how to go about testing `HasAttributes` further since its coverage seems quite sparse currently, feel free to push this back to draft and nudge me in the right direction if you want more than this, but I've added some extra assertions.

taylorotwell:
Eh, I don't really think this is a good idea. I see other places in the code that will fail because of this - they expect to be able to depend on an attribute type being returned and now we would have had to add checks in all of those places to make sure it's actually an attribute.

PR #46626, opened 2023-03-29
[10.x] Improves `trait_uses_recursive` docblock

crynobone:
image

It should be possible since the following line would use [`class_uses`](https://www.php.net/manual/en/function.class-uses.php)


Muffinman:
Related question: should this function just use native type declarations?

nunomaduro (approved):

PR #46628, opened 2023-03-29
[10.x] Spaces in password

aktuba:
The current implementation of password generation allows marginal spaces and a password consisting entirely of spaces.
This PR corrects this behavior.


danilopinotti:
> Someone please tell me what is wrong with the pipeline. I can't open the report (((

image

valorin:
This change forces a space into the password when enabled, rather than making it random, which is a different behaviour to the other character toggles. 

I believe it will also lower the entropy of the password - you're essentially removing an entire character and giving `length - 2` possible variants of the reduced length.

(I'm on my phone atm but can give a better explanation later today if that doesn't make sense?)

taylorotwell:
See comment above.

PR #46634, opened 2023-03-30
[10.x] Edit about console output of synchronous running jobs

kemalbal:
In some cases, synchronous jobs can produce confusing console output. 
In the example below, I realized that a second job running synchronously within a job is actually producing output that shouldn't be there. These two jobs need to run in two different queues if they are not to run synchronously.

examplejob
examplejob2

The output I expect when only the "queue1" queue is listened is as follows.
beklenen

But it continues to produce output in synchronous jobs.

jobs_outputs


taylorotwell:

Thanks!

PR #46637, opened 2023-03-30
[10.x] Add global default options feature to HTTP client.

xurshudyan:
This pull request adds a new feature to the Laravel HTTP client that allows global default options to be set for all requests. This makes it possible to define headers, timeouts, and other options that will be used by default for all HTTP requests.

Example Usage:
You can set global default options for all HTTP requests by calling the `withDefaultOptions` method within the `boot` method of your `App\Providers\AppServiceProvider` class. Here's an example:

```php
Http::withDefaultOptions([
    'headers' => [
        'Authorization' => 'Bearer my-secret-token',
    ],
    'timeout' => 10,
]);
```
This sets the Authorization header and a 10 second timeout for all requests made using the Laravel HTTP client.

To remove one or more default options, use the `withoutDefaultOptions` method. For example:

```php
Http::withoutDefaultOptions('headers.Authorization')->get('https://foo.com');
```

This removes the Authorization header from the default options.

If you need to remove multiple default options, you can pass an array of keys to the `withoutDefaultOptions` method:

```php
Http::withoutDefaultOptions(['headers.Authorization', 'timeout'])->get('https://foo.com');
```

This removes the Authorization header and the timeout option from the default options.

Another example.

```php
Http::withDefaultOptions([
    'base_uri' => 'https://api.example.com/v1/',
]);

$response = Http::get('users');

// This sends a GET request to https://api.example.com/v1/users
```

### DIFF
--- a/src/Illuminate/Http/Client/Factory.php
+++ b/src/Illuminate/Http/Client/Factory.php
@@ -7,6 +7,7 @@ use GuzzleHttp\Promise\PromiseInterface;
 use GuzzleHttp\Psr7\Response as Psr7Response;
 use GuzzleHttp\TransferStats;
 use Illuminate\Contracts\Events\Dispatcher;
+use Illuminate\Support\Arr;
 use Illuminate\Support\Str;
 use Illuminate\Support\Traits\Macroable;
 use PHPUnit\Framework\Assert as PHPUnit;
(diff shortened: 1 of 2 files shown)


abenerd:
I like the idea but I wonder what the impact would be for vendor code using the http client. I can imagine this breaking things or leaking stuff.

crynobone:
On one hand yes this would have the potential to break vendor code but in certain environments where any external HTTP request needs to use a proxy there is no way at the moment for any developer to add proxy options for vendor code.

Previous attempt: #40332

taylorotwell:

PR #46648, opened 2023-03-31
[10.x] Query builder `keyBy` functionality

bert-w:
This PR adds functionality for query builder results to be keyed by a given column, instantly using PDO. This effectively speeds up computation about 3 times when comparing the following code samples (see benchmark `Illuminate\Tests\Integration\Database\QueryBuilderKeyByBench`):
```php
// Fast:
$builder->keyBy('title')->get();
// Slow:
$builder->get()->keyBy('title');
```

The top syntax makes use of setting a custom PDO fetch mode namely `PDO::FETCH_UNIQUE|PDO::FETCH_OBJ`. This fetch mode is applied directly onto the PDO statement using `$statement->fetchAll($this->fetchAllMode);`.

Noteworthy is that `FETCH_UNIQUE` strips the first column of the result set and sets it as the array key, thus a column `['a', 'b', 'c']` would effectively appear in the result array as `['a' => ['b', 'c']]`. To counteract this behaviour, the column that you pass to `$builder->keyBy($column)` will be prepended to the requested columns such that it actually selects `['a', 'a', 'b', 'c']`. Now the first column can be freely shifted by PDO and the resulting object will be as expected.

There currently seems to be a bug though where leveraging `$statement->setFetchMode(PDO::FETCH_UNIQUE|PDO::FETCH_OBJ)` ([php.net link](https://www.php.net/manual/en/pdostatement.setfetchmode)) does not work for this specific combination (I could not find any bug report), so the code resorts to manually overriding this by calling to `$statement->fetchAll($this->fetchAllMode)` which does the trick.

Use case: any sort of query where you _instantly_ want to retrieve a model from the array without searching through it. For instance, keeping a local cache of all countries (200+) using `$_cache = Country::keyBy('country_code')->get()` which you can then quickly access countries using `$_cache['us']`. This saves quite some time when you don't need to re-key the collection in PHP.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -2741,8 +2741,8 @@ class Builder implements BuilderContract
      */
     protected function getKeyed($key, $columns = ['*'])
     {
-        return collect($this->onceWithColumns($this->qualifyStarColumns($columns), function () use ($key) {
-            $this->columns = Arr::prepend($this->columns, $key);
+        return collect($this->onceWithColumns(Arr::wrap($columns), function () use ($key) {
+            $this->columns = Arr::prepend($this->qualifyStarColumns($this->columns), $key);
 
             return $this->processor->processSelect($this, $this->runSelect(\PDO::FETCH_UNIQUE));
         }));
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -2741,9 +2741,23 @@ class Builder implements BuilderContract
      */
     protected function getKeyed($key, $columns = ['*'])
     {
-        return collect($this->onceWithColumns(Arr::prepend(Arr::wrap($columns), $key), function () {
+        return collect($this->onceWithColumns(Arr::prepend($this->qualifyStarColumns($columns), $key), function () {
             return $this->processor->processSelect($this, $this->runSelect(\PDO::FETCH_UNIQUE));
         }));
+    }
+
+    /**
+     * Qualify columns with their table name if they have not been assigned.
+     * 
+     * @param  array|string  $columns
+     * @return array
+     */
+    protected function qualifyStarColumns($columns)
+    {
+        return array_map(
+            fn ($column) => $column === '*' ? $this->from.'.'.$column : $column,
+            Arr::wrap($columns)
+        );
     }
 
     /**
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -2773,7 +2773,7 @@ class Builder implements BuilderContract
 
         return tap($this->connection->select(
             $this->toSql(), $this->getBindings(), ! $this->useWritePdo
-        ), fn () => $this->connection->setFetchAllMode(null, false)
+        ), fn () => ! is_null($fetchAllMode) ? $this->connection->setFetchAllMode(null, false) : null
         );
     }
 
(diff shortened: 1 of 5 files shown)


src/Illuminate/Database/Query/Builder.php
```diff
@@ -2703,20 +2723,57 @@ public function soleValue($column)
...
+    /**
+     * Execute the query as a "select" statement where the resulting array is keyed by the given column.
+     *
+     * @param  string  $key
+     * @param  array|string  $columns
+     * @return \Illuminate\Support\Collection
+     */
+    protected function getKeyed($key, $columns = ['*'])
+    {
+        return collect($this->onceWithColumns(Arr::prepend($this->qualifyStarColumns($columns), $key), function () {
```
bert-w, line 2744: some explanation here: the column that you will use as the key will need to be prepended here since PDO will remove it in the process of assigning it as the key (caused by `PDO::FETCH_UNIQUE`). Furthermore, a star `'*'` has to be prefixed with the table name since `SELECT key_column, * FROM my_table` is not valid SQL (apparently putting `*` first IS valid SQL but that doesn't work in this case). The result then becomes `SELECT key_column, my_table.* FROM my_table`.

src/Illuminate/Database/Query/Builder.php
```diff
@@ -2703,11 +2723,63 @@ public function soleValue($column)
...
+     * Execute the query as a "select" statement where the resulting array is keyed by the given column.
+     *
+     * @param  string  $key
+     * @param  array|string  $columns
+     * @return \Illuminate\Support\Collection
+     */
+    protected function getKeyed($key, $columns = ['*'])
+    {
+        return collect($this->onceWithColumns(Arr::wrap($columns), function () use ($key) {
+            $this->columns = Arr::prepend($this->qualifyStarColumns($this->columns), $key);
```
joelharkes, line 2745: does this work well when both the key is in the sql result set as first item and later on in the array? does that not lead to problems with column names or does PDO solve that?

bert-w: I couldnt think of possible issues. You can freely do `SELECT name, name, name FROM users` and you will get the expected result (3x the same column per record). The only thing that happens explicitly here is to add this column *within* the ` onceWithColumns` callback, and also to make sure that any star column is prefixed with a table name since `SELECT name, * FROM users` is not valid SQL.

bert-w: Just to be clear, `PDO::FETCH_UNIQUE` always uses the first column of the result set as the array key. If there are multiple rows/records with the same key, the first record will occupy the array and later values will be discarded.

taylorotwell:

joelharkes:
@taylorotwell  we use keyBy all the time for chunked optimizations, it would be nice if this can be optimized since its already mostly used for performance optimizations.

bert-w:
I've added tests for relationships now as well so they as well can be assigned in `key => value` format. This will shave milliseconds off certain apps which imo is worth it. Even combinations like `$query->groupBy('col')->keyBy('col')` works fast because retrieving particular elements can now be done in `O(1)`.

@taylorotwell what about reopening this as draft and poll it on twitter? There are no breaking changes, except 1 signature change in Collection `mapToDictionary` which has an optional `$keepKeys` param in the latest commit.

decadence:
@bert-w Taylor doesn't look on the closed PR's messages so you can open new PR with your additions.

bert-w:
@decadence i don't know what the etiquette is around here; I want to prevent stuff like "hey i closed this PR already why are you resubmitting this?"

At the same time i imagine that Taylor probably has put the notifications off since there is so much that goes through him nowadays

PR #46651, opened 2023-03-31
[9.x] Throw Exception if a query expression is defined in `order by` for cursor pagination

maximepvrt:
Add check for non-aliased expression in order by clause in cursorPaginate() method

This change adds a check to ensure that expressions used in the order by clause are aliased and also included in the select clause. If a non-aliased expression is used, a RuntimeException will be thrown with an appropriate message.

This change ensures that cursorPaginate() method works correctly with expressions used in the order by clause, and prevents potential issues with non-aliased expressions.

> Query expressions in "order by" clauses are supported only if they are aliased and added to the "select" clause as well.
[https://laravel.com/docs/9.x/pagination#cursor-vs-offset-pagination](https://laravel.com/docs/9.x/pagination#cursor-vs-offset-pagination)

Additionally, I attempted to write a test case that would trigger the RuntimeException when a non-aliased expression is used in the order by clause, but was not able to do so.

Thanks for considering this merge request!


taylorotwell:

maximepvrt:
Hi @taylorotwell, currently without this explicit error, the error returned is "can only flip string and integer values" from this line https://github.com/laravel/framework/blob/9.x/src/Illuminate/Pagination/AbstractCursorPaginator.php#L208. I think it would be good to add explicit errors to encourage adoption of cursor pagination.

maximepvrt:
@taylorotwell, have you seen my last message ?

maximepvrt:
@driesvints 😇

driesvints:
Looks like Taylor doesn't wants to merge this one sorry.

PR #46653, opened 2023-03-31
[10.x] allow `$sleepMilliseconds` parameter receive a Closure in `retry` method from PendingRequest

williamtrevisan:
The `retry` helper accepts either an `int` or a `Closure`. When passing a `Closure` object, the `$attempts` and the `$exception` would be passed, allowed users to implement their own backoff algorithm. The `retry` helper is used in the Pending Request class to perform request retries as well, but in this specific local, the parameters are sent after being informed by `retry` method that belongs to the same class mentioned, and in this method, is not accepted to inform `$sleepMilliseconds` as a `Closure`.

This pull request will allows you to inform `$sleepMilliseconds` parameter as a `Closure` or `int`, opening the possibility to create your own backoff algorithm for retry requests. 

## Example use case

Imagine a scenario where it is necessary to perform a backoff when the exception returned is of type `Request Timeout (408)`, with this modification it would be possible to implement a dynamic strategy as follows:
```PHP
Http::retry(
  times: 2,
  sleepMilliseconds: function (int $attempt, \Throwable $exception): int {
    if ($exception->getCode() === \Illuminate\Http\Response::HTTP_REQUEST_TIMEOUT) {
      return $attempt * 300000; // five minutes in milliseconds
    }
                
    return 0;
  }
);
```

### DIFF
--- a/tests/Http/HttpClientTest.php
+++ b/tests/Http/HttpClientTest.php
@@ -1576,6 +1576,31 @@ class HttpClientTest extends TestCase
         $this->factory->assertSentCount(1);
     }
 
+    public function testRequestsWillBeWaitingSleepMillisecondsReceivedBeforeRetry()
+    {
+        $startTime = microtime(true);
+
+        $this->factory->fake([
+            '*' => $this->factory->sequence()
+                ->push(['error'], 500)
+                ->push(['error'], 500)
+                ->push(['ok'], 200),
+        ]);
+
+        $this->factory
+            ->retry(3, function ($attempt, $exception) {
+                $this->assertInstanceOf(RequestException::class, $exception);
+
+                return $attempt * 100;
+            }, null, true)
+            ->get('http://foo.com/get');
+
+        $this->factory->assertSentCount(3);
+
+        // Make sure was waited 300ms for the first two attempts
+        $this->assertEqualsWithDelta(0.3, microtime(true) - $startTime, 0.03);
+    }
+
     public function testMiddlewareRunsWhenFaked()
     {
         $this->factory->fake(function (Request $request) {
--- a/src/Illuminate/Http/Client/PendingRequest.php
+++ b/src/Illuminate/Http/Client/PendingRequest.php
@@ -131,7 +131,7 @@ class PendingRequest
     /**
      * The number of milliseconds to wait between retries.
      *
-     * @var int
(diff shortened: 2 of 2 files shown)


tests/Http/HttpClientTest.php
```diff
@@ -1576,6 +1576,31 @@ public function testExceptionThrownInRetryCallbackWithoutRetrying()
...
+                $this->assertInstanceOf(RequestException::class, $exception);
+
+                return $attempt * 100;
+            }, null, true)
+            ->get('http://foo.com/get');
+
+        $this->factory->assertSentCount(3);
+
+        // Make sure was waited 300ms for the first two attempts
+        $this->assertEqualsWithDelta(0.3, microtime(true) - $startTime, 0.03);
```
hicham-saddek, line 1601: This could fail in slow machines

CLL-GTA:
Would it be possible to do this on 9.x as well? I don't think there is an issue as the `retry()` helper already accepts a Closure for `$sleepMilliseconds`

williamtrevisan:
> Would it be possible to do this on 9.x as well? I don't think there is an issue as the `retry()` helper already accepts a Closure for `$sleepMilliseconds`

Yes, it would. I'll open a PR to implement that there.

tests/Http/HttpClientTest.php
williamtrevisan: Do you know another approach that I can use here to test this? Or maybe it's not necessary test that functionality knowing that is already tested in retry helper scope?

williamtrevisan:
> > Would it be possible to do this on 9.x as well? I don't think there is an issue as the `retry()` helper already accepts a Closure for `$sleepMilliseconds`
> 
> Yes, it would. I'll open a PR to implement that there.

Well, I have opened an PR to possibility it in 9.x, but folks from laravel core said that 9.x is not open for new features, and people need to upgrade to 10.x. You can see this [here](https://github.com/laravel/framework/pull/47213)

CLL-GTA:
> Well, I have opened a PR to make it possible in 9.x, but folks from laravel core said that 9.x is not open for new features, and people needs to upgrade to 10.x. You can see this https://github.com/laravel/framework/pull/47213

Well, that's disappointing, but thanks for trying.

PR #46658, opened 2023-04-01
[9.x] Test Improvements

crynobone:
Testbench Core 7.24.0 fixed some configuration value leaks between tests due to the way it set environment values including `APP_KEY`, `APP_DEBUG` etc.

https://github.com/orchestral/testbench-core/releases/tag/v7.24.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -101,7 +101,7 @@
         "league/flysystem-read-only": "^3.3",
         "league/flysystem-sftp-v3": "^3.0",
         "mockery/mockery": "^1.5.1",
-        "orchestra/testbench-core": "^7.16",
+        "orchestra/testbench-core": "7.x-dev",
         "pda/pheanstalk": "^4.0",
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -101,7 +101,7 @@
         "league/flysystem-read-only": "^3.3",
         "league/flysystem-sftp-v3": "^3.0",
         "mockery/mockery": "^1.5.1",
-        "orchestra/testbench-core": "^7.16",
+        "orchestra/testbench-core": "7.x-dev",
```
crynobone, line 104:
```suggestion
        "orchestra/testbench-core": "^7.24",
```

crynobone, line 104:
```suggestion
        "orchestra/testbench-core": "^7.24",
```

GrahamCampbell:
Nice. Probably this should go out as v9.53.0 (possibly the last ever minor version bump on the 9.x series). 🎉

PR #46659, opened 2023-04-01
[10.x] Test Improvements

crynobone:
Testbench Core 8.3.0 fixed some configuration value leaks between tests due to the way it set environment values including APP_KEY, APP_DEBUG etc.

https://github.com/orchestral/testbench-core/releases/tag/v8.3.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -103,7 +103,7 @@
         "league/flysystem-read-only": "^3.3",
         "league/flysystem-sftp-v3": "^3.0",
         "mockery/mockery": "^1.5.1",
-        "orchestra/testbench-core": "^8.1",
+        "orchestra/testbench-core": "8.x-dev",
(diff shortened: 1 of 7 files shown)


composer.json
```diff
@@ -103,7 +103,7 @@
         "league/flysystem-read-only": "^3.3",
         "league/flysystem-sftp-v3": "^3.0",
         "mockery/mockery": "^1.5.1",
-        "orchestra/testbench-core": "^8.1",
+        "orchestra/testbench-core": "8.x-dev",
```
crynobone, line 106:
```suggestion
        "orchestra/testbench-core": "^8.3",
```

crynobone, line 106:
```suggestion
        "orchestra/testbench-core": "^8.3",
```

PR #46662, opened 2023-04-02
[10.x] Add `assertPathIs` to TestResponse

milwad-dev:
Adds `assertPathIs()` to allow for something like:

```php
// Route
Route::get('milwad', function () {
    return 'milwad-dev';
})->name('milwad.index');

Route::get('laravel/framework', function () {
    return 'laravel';
})->name('laravel.index');

// Test
$response1 = $this->get(route('milwad.index')); 
$response1->assertPathIs('milwad'); ✅ Pass

$response2 = $this->get(route('laravel.index'));
$response2->assertPathIs('laravel/framework'); ✅ Pass

$response1->assertPathIs('dev'); ❌ Fail
$response2->assertPathIs('laravel'); ❌ Fail
```

If accepted, I will add related tests!

### DIFF
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -980,7 +980,7 @@ class TestResponse implements ArrayAccess
     }
 
     /**
-     * 
+     * Assert that the response url equals the given url.
      *
      * @param  string  $url
      * @return $this
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -980,6 +980,19 @@ class TestResponse implements ArrayAccess
     }
 
     /**
+     * 
+     *
+     * @param  string  $url
+     * @return $this
+     */
+    public function assertUrlIs($url)
+    {
+        PHPUnit::assertEquals($url, request()->path());
+
+        return $this;
+    }
+
+    /**
      * Assert that the response view equals the given value.
      *
      * @param  string  $value
--- a/src/Illuminate/Testing/TestResponse.php
+++ b/src/Illuminate/Testing/TestResponse.php
@@ -980,12 +980,12 @@ class TestResponse implements ArrayAccess
     }
 
     /**
-     * Assert that the response url equals the given url.
+     * Assert that the response url path equals the given url (path).
      *
      * @param  string  $url
      * @return $this
      */
-    public function assertUrlIs($url)
+    public function assertUrlPathIs($url)
     {
         PHPUnit::assertEquals($url, request()->path());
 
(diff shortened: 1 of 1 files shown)


johanrosenson:
This is just testing the path right and not the full URL? a more fitting name maybe could be `assertUrlPathIs`

And is this really testing anything about the response? you pull in the `request()`, I don't think this has anything to do with testing the response, is this not just asserting that `route('milwad.index')` returns the expected path?

milwad-dev:
@johanrosenson 
Fix

crynobone:
Maybe unify the API with Laravel Dusk: https://laravel.com/docs/10.x/dusk#assert-path-is

Also, we can introduce `assertPathIsNot()`

milwad-dev:
@crynobone 
Yes, if merged this PR, we can add `assertPathIsNot`.

crynobone:
@milwad-dev can you change the method to `assertPathIs()` to follow Dusk API?

johanrosenson:
> @johanrosenson Fix

is it really fixed thought? you are still just pulling in the request from the app, not the request associated with the response.

To give you an example, would the following code work?

```
$response1 = $this->get(route('milwad.index'));
$response2 = $this->get(route('another-route'));

$response1->assertUrlPathIs('milwad');
```

milwad-dev:
@crynobone Sure

milwad-dev:
@johanrosenson 
We have a problem, that is to how to access to url !?

johanrosenson:
> @johanrosenson We have a problem, that is to how to access to url !?

If you can't access the URL where we need it, maybe this test doesn't belong on the TestResponse.

milwad-dev:
@johanrosenson 
It needs to be in TestResponse.
Also if write like this, there is not problem !
```php
// Route
Route::get('milwad', function () {
    return 'milwad-dev';
})->name('milwad.index');

Route::get('laravel/framework', function () {
    return 'laravel';
})->name('laravel.index');

// Test
$response1 = $this->get(route('milwad.index'));
$response1->assertPathIs('milwad');

$response2 = $this->get(route('laravel.index'));
$response2->assertPathIs('laravel/framework');

✅ Pass
```

johanrosenson:
```php
$response1 = $this->get(route('milwad.index'));
$response1->assertPathIs('milwad');

$response2 = $this->get(route('laravel.index'));
$response2->assertPathIs('laravel/framework');

// ✅ Pass

// call $response1->assertPathIs() again but with the path of $response2
$response1->assertPathIs('laravel/framework');

// ✅ Also pass
```

The test method is sadly flawed since we can't get the request from the response, so `assertPathIs` as written right now will have a high risk of giving false positives if used in the wrong order. IMHO methods like this should not be dependant on call order or global state.

```php
$this->assertSame('milwad', parse_url(route('laravel.index'), \PHP_URL_PATH));

// this test would give you same information about the route and the generated path.
```

taylorotwell:

PR #46665, opened 2023-04-02
[10.x] Fix api singleton route registration

drradao:
# Issues

### Context

When creating an API Singleton resource we're given the possibility to use the `creatable` method that accepts a `bool $creatable` parameter.

### Problem

This parameter is currently being ignore. It will register a `store` endpoint event when `false` is passed into the creatable method.

### Context

```php
Route::apiSingleton('route', Controller::class)
    ->creatable()
    ->destroyable(false);
```

### Problem

Since the destroyable function's parameter is also ignore we cannot have a resource that is creatable but not destroyable.

### Context

```php
Route::apiSingleton('route', Controller::class)
    ->only(['edit']);
```

### Problem

This registers the edit route.

# Notes

I've included some tests of what I believe to be the expected usage, some of this tests do not pass with the current implementation, my implementation addresses all issues listed above.

Also, this is my first laravel/framework PR, if you feel that something can be improved, please comment and I'll improve it.

### DIFF
--- a/src/Illuminate/Routing/ResourceRegistrar.php
+++ b/src/Illuminate/Routing/ResourceRegistrar.php
@@ -29,6 +29,13 @@ class ResourceRegistrar
     protected $singletonResourceDefaults = ['show', 'edit', 'update'];
 
     /**
+     * The default actions for a api singleton resource controller.
+     *
+     * @var string[]
+     */
+    protected $apiSingletonResourceDefaults = ['show', 'update'];
+
+    /**
      * The parameters set for this resource instance.
      *
      * @var array|string
@@ -147,7 +154,9 @@ class ResourceRegistrar
             return;
         }
 
-        $defaults = $this->singletonResourceDefaults;
+        $defaults = $options['apiSingleton'] ?? false
+            ? $this->apiSingletonResourceDefaults
+            : $this->singletonResourceDefaults;
 
         $collection = new RouteCollection;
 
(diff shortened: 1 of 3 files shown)


2092015:
This is great... I was also having this problem and about to create an issue for it.
Great implementation.

src/Illuminate/Routing/ResourceRegistrar.php
```diff
@@ -28,6 +27,13 @@ class ResourceRegistrar
      */
     protected $singletonResourceDefaults = ['show', 'edit', 'update'];
 
+    /**
+     * The default actions for a api singleton resource controller.
+     *
+     * @var string[]
+     */
+    protected $apiSingletonResourceDefaults = ['show', 'update'];
+
```
jessarcher, line 36: We don't have a corresponding property for `$apiResourceDefaults`, which feels asymmetrical.

In my testing, the `apiResource` method has the same behaviour; it will register the `edit` route if explicitly provided to the `only` method.

I am curious to know how much of an issue this is. Perhaps there are situations where someone may want the `edit` route if they explicitly ask for it.

If we did want to fix this issue, perhaps a property like `$excludedApiActions = ['create', 'edit'];` might be better so that we can just remove those actions from both `apiResource` and `apiSingletonResource`.

jessarcher (commented):
Thanks for fixing this!

In my comment, I've noted that the `apiResource` method will also register `create` and `edit` routes if explicitly provided to the `only` method, but I'm not sure how much of an issue this is.

In order to speed things along, I think it would be good if this PR just focused on the issue of not respecting the value passed to the `createable` and `destroyable` methods and we can consider the other issue in a separate PR.

tests/Routing/RouteRegistrarTest.php
```diff
@@ -1088,6 +1088,68 @@ public function testCanRemoveMiddlewareFromGroupUnregisteredGroup()
...
+        $this->assertTrue($this->router->getRoutes()->hasNamedRoute('users.update'));
+    }
+
+    public function testCanRegisterCreatableApiSingleton()
+    {
+        $this->router->apiSingleton('users', RouteRegistrarControllerStub::class)->creatable();
+
+        $this->assertCount(4, $this->router->getRoutes());
+
+        $this->assertTrue($this->router->getRoutes()->hasNamedRoute('users.store'));
```
axlon, line 1107: I would assert `users.show` and `users.update` here as well (also applies to some of the tests below this one)

axlon:
If we're approaching this as a bugfix, shouldn't this PR go to 9.x?

src/Illuminate/Routing/ResourceRegistrar.php
drradao: Hi Jess, I confess I got a bit carried away and maybe changed a bit too much...

I believe that the `edit` route should never exist in an API resource, am I wrong?

Usually I tend to use a positive approach, instead if using an exclusion approach, I think it's easier to read. I'm going to review this and remove the `$apiSingletonResourceDefaults` I should've used a less intrusive approach.

PR #33612, opened 2020-07-21
[9.x] Flysystem v2

driesvints:
Initial attempt at migrating to (currently unreleased) Flysystem v2. Might be a little too soon for Laravel 8. In draft until Flysystem v2 has been released.

### Breaking Changes

- Any write operations (like `put`, `write`, `writeStream`, etc) now overwrites by default. Checking for existing files needs to be done in user land from now on.
- `writeStream` no longer throws a `FileExistsException`. Since this exception is nowhere used anymore in the framework, it's been removed.
- Similarly, `get` & `readStream` no longer throw a `FileNotFoundException` exception when a file exists. This is because the underlying Flysystem integration no longer distinguishes this specific failure from any other read failure. Both methods will return `null` when they cannot read a file.
- `delete` returns `true` now even if the file wasn't found.
- Using custom filesystem creators in a service provider now requires them to return an instance of `Illuminate\Contracts\Filesystem\Filesystem` rather than `League\Flysystem\FilesystemInterface` like before. See https://github.com/laravel/framework/pull/33612#discussion_r460018589
- The FTP driver now ships as a separate league package and needs to be explicitly required through Composer for it to be used.
- The concept of the cached adapter has been removed from Flysystem v2 and thus also from Laravel.
- `Illuminate\Filesystem\FilesystemAdapter`'s constructor now requires an extra `$adapter` argument and an optional `$options` argument.

PR #46668, opened 2023-04-03
[11.x] Do not change relative path to absolute one

CmdMumm:
Do not change relative path to absolute if SFTP-Driver is created without root path - similiar to createFtpDiver()

### DIFF
--- a/src/Illuminate/Filesystem/FilesystemManager.php
+++ b/src/Illuminate/Filesystem/FilesystemManager.php
@@ -214,7 +214,7 @@ class FilesystemManager implements FactoryContract
     {
         $provider = SftpConnectionProvider::fromArray($config);
 
-        $root = $config['root'] ?? '/';
+        $root = $config['root'] ?? '';
 
         $visibility = PortableVisibilityConverter::fromArray(
             $config['permissions'] ?? []


driesvints:
@CmdMumm will need a much more thorough explanation to let taylor know what this is about.

CmdMumm:
If an SFTP connection is to be established using a relative path and the root path is not set in the connection configuration, the relative path becomes an absolute path by inserting a leading "/", e.g. 'foo/bar' becomes '/foo/bar'

One of the biggest issues here is, that this behaviour is not noticed. In my case through the unnoticed path change from relative to absolute, order files were not longer distributed to the fulfillment center. But the error seen in **phpseclib** (from the sftp-server) never throws an exception. So the error was not recognized.

In my opinion there are two problems:
1.  The change from relative to an absolute path.
2. A possible error from the SFTP-Server (e.g. no rights on the absolute path) isn't noticed.

In Laravel 8.83.27 everything worked as expected, so 'foo/bar' stayed 'foo/bar'.

I think the problem is in laravel/framework/src/Illuminate/Filesystem/FilesystemManger.php, method createSftpDriver():

```
public function createSftpDriver(array $config)
   {
       $provider = SftpConnectionProvider::fromArray($config);

       $root = $config['root'] ?? '/';
    }

```
If $config['root'] is not set, a slash is added.

I would rather expect a behaviour like in the createFtpDriver()-method:

 ```
public function createFtpDriver(array $config)
    {
        if (! isset($config['root'])) {
            $config['root'] = '';
        }
```
If $config['root'] is not set it is set empty and the relative path is preserved.

taylorotwell:
If everything worked as expected in 8.83.27, can we determine the PR or change that introduced this behavior of using `/` as the default root path? I would be curious to know for what reason that behavior was introduced.

driesvints:
@taylorotwell these changes were done by me mostly when we moved to Flysystem v2 in Laravel v9: https://github.com/laravel/framework/pull/33612

Afterwards (I can't remember what triggered it) we needed to fix the FTP root config: https://github.com/laravel/framework/pull/40939. Frank pointed out that this needed to be an empty string (like this PR is trying to do for the SFTP driver). Most likely we should have also changed it for the SFTP driver at the time.

Issue #23042, opened 2018-02-06
Eloquent/Model::withCount doesn't specify database/schema name


foreshadow:
- Laravel Version: 5.5.33
- PHP Version: 7.1.8
- Database Driver & Version: Mysql 5.7

#### Eloquent/Model::withCount doesn't specify database/schema name
I have 2 models User and Problem. Their table locates in different database/schema in same server (`users` locates in connection `mysql` a.k.a. schema `acm`, `problem` locates in connection `mysql2` a.k.a. schema `boj` in my case).
I have defined `protected $connection` for these two models. Anything other works well. 
And when I call `User::with('solved')` it works well, but when calling `User::withCount('solved')`, the generated sql sentence doesn't specify the `problem` table's database/schema, which will cause an error.

(formatted)
```
(3/3) QueryException
SQLSTATE[42S02]: Base table or view not found: 1146 Table 'acm.problem' doesn't exist (SQL: 
```
```sql
select `users`.*, (
    select count(*) from `problem` 
    inner join `boj`.`solution` on `problem`.`problem_id` = `boj`.`solution`.`problem_id` 
    where `users`.`id` = `boj`.`solution`.`acm_user_id` and `boj`.`solution`.`result` = 4
) as `solved_count` 
from `users` order by `active_at` desc limit 150 offset 0)
```

#### Defines
```php
class User
{
    protected $connection = 'mysql';
    // some other defines 

    public function solved()
    {
        $mysql2 = config('database.connections.mysql2.database');
        return $this
            ->belongsToMany(Problem::class, "$mysql2.solution", 'acm_user_id', 'problem_id')
            ->wherePivot('result', 4);
    }
}
```
I believe it doen't matter where the `solution` table locates and how I call it... Although it may looks ugly.
```php
class Problem
{
    protected $connection = 'mysql2';
    protected $table = 'problem';
    // some other defines
}
```
#### Call
```php
User::with('roles')->withCount('solved')->orderBy('active_at', 'desc')->paginate(150);
```

foreshadow:
Besides, `withCount` a `hasMany` relation will cause the same problem.

miki131:
@foreshadow, try this
```php
User::with('roles')
  ->withCount(['solved' => function($query) {
    $mysql2 = config('database.connections.mysql2.database');
    $query->from("$mysql2.problem");
  }])
  ->orderBy('active_at', 'desc')
  ->paginate(150);
```

vesper8:
[...]

PR #46675, opened 2023-04-03
[9.x] Cross-database support for all relations

bert-w:
Related to https://github.com/laravel/framework/issues/23042

### Issue
Various relationship methods do not work with cross database relations including `withCount` and `whereHas`. SQL errors were thrown because the database name was not prefixed in all cases. Some work has been done previously to accomplish this, however it wasn't tested thoroughly and several cases were left unconsidered (like joins and specific relation types).

### Requirements
- Any `MorphToMany`/`BelongsToMany` relations are required to have an intermediary Eloquent model defined like so, because the connection from the intermediary table cant be resolved otherwise:
```php
public function taggables()
{
    return $this->morphToMany(...)->using(Taggable::class);
}
```
- The rest of the relations work out of the box and any joins/froms that happen on a different connection automatically have their database name prefixed.

### Task status
- [x] Test all relationship methods (only correct SQL execution is tested)
- [x] There is still one issue with withCount on a `morphTo()` relation, where `$users->query()->withCount('morphable')->get()` results in a syntax error when no additional arguments are given to the relation (the SQL contains `'laravel_reserved_0'.''` because it is missing a key). The issue is solved though once you hardcode the relation as `return $this->morphTo(null, null, null, 'id')`. I excluded the test case from my tests it is an existing issue.

### Support status
- [x] MySQL 5.7
- [x] MySQL 8.0
- [x] MariaDB 10.0
- [x] SQL Server 2019 (FIXED prefixing `dbo` was required. A custom config override `'schema'` has been added in the database configuration, similar to Postgres)
- [x] NOT SUPPORTED: PostgreSQL 14 (it requires additional pgsql extensions `dblink` and/or `postgres_fdw`)
- [x] NOT SUPPORTED: SQLite

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -750,6 +750,8 @@ class Builder implements BuilderContract
 
         $constraints($relation);
 
+        $this->query->prependDatabaseNameIfCrossDatabaseQuery($relation->getBaseQuery());
+
         // Once we have the results, we just match those back up to their parent models
         // using the relationship instance. Then we just return the finished arrays
         // of models which have been eagerly hydrated and are readied for return.
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -54,6 +54,8 @@ trait QueriesRelationships
             $relation->getRelated()->newQueryWithoutRelationships(), $this
         );
 
+        $this->query->prependDatabaseNameIfCrossDatabaseQuery($hasQuery);
+
         // Next we will call any given callback as an "anonymous" scope so they can get the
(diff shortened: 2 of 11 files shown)


GrahamCampbell:
I don't think cross-database relationships is something that is meant to be supported by the framework.

GrahamCampbell:
In any case, if this is to be a new feature that is wanted, it would need to target 10.x, not 9.x.

bert-w:
Imo its definitely something that needs to be supported. Eloquent already has most of the infrastructure to deal with different connections. Besides that, this fix is relatively small since it only prepends database names when connections differ.

I would be fine with sending it to 10.x but I regard it as a bug since some of the functionality currently works and some doesn't.

bert-w:
decided to send to 10.x instead

bert-w:
decided to send to 10.x instead https://github.com/laravel/framework/pull/46769

PR #46677, opened 2023-04-03
[9.x] Removes unnecessary creatable API singleton parameter

drradao:
# Issue

Currently the `creatable()` and `destroyable()` methods in an API singleton resource route registartion accept boolean parameters that aren't taken into consideration later.

```php
Route::apiSingleton('route', Controller::class)
    ->creatable()
    ->destroyable();
```

# Possible solution

Removal of these parameters as suggested [here](https://github.com/laravel/framework/pull/46665#issuecomment-1494838213)

### DIFF
--- a/src/Illuminate/Routing/PendingSingletonResourceRegistration.php
+++ b/src/Illuminate/Routing/PendingSingletonResourceRegistration.php
@@ -90,12 +90,11 @@ class PendingSingletonResourceRegistration
     /**
      * Indicate that the resource should have creation and storage routes.
      *
-     * @param  bool  $creatable
      * @return $this
      */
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This is strictly speaking a breaking change, so should go to master. In any case this is not a bug fix, so 9.x would not be appropriate.

GrahamCampbell:
Oh, catching up on the other PR. I guess this is a bug fix. ;)

drradao:
I guess it could be breaking if someone is passing a value there...

We could also add a comment in the parameter stating that the feature isn't there so people don't spam the optimize:clear command like maniacs thinking that they're cached 😄

PR #46678, opened 2023-04-04
[10.x] Remove unused param from `assertFound` | `assertMovedPermanently`

milwad-dev:
There is no need to `@param` in phpdoc !!


driesvints:
Thanks!

PR #46681, opened 2023-04-04
[10.x] Issue form for bugs

driesvints:
This PR introduces a new issue form for reporting bugs which looks like the screenshot below. It'll help ensure bug reports are filled out more detailed.

Screenshot 2023-04-04 at 11 36 10


.github/ISSUE_TEMPLATE/Bug_report.yml
```diff
@@ -0,0 +1,46 @@
...
+      required: true
+  - type: input
+    id: database-version
+    attributes:
+      label: Database Driver & Version
+      description: If applicable, please provide the database driver and full version you're using.
+      placeholder: "mysql Ver 8.0.31 for macos13.0 on arm64 (Homebrew)"
+    validations:
+      required: false
+  - type: textarea
```
jbrooksuk, line 32:
```suggestion
  - type: markdown
```

Using this means you can get the full editor toolbar.

.github/ISSUE_TEMPLATE/Bug_report.yml
```diff
@@ -0,0 +1,46 @@
...
+    id: description
+    attributes:
+      label: Description
+      description: Please provide a detailed description of the issue you're facing. Don't write a book but provide enough details for us.
+    validations:
+      required: true
+  - type: textarea
```
jbrooksuk, line 39:
```suggestion
  - type: markdown
```

jbrooksuk (changes requested):
Let's use the Markdown field type so we can retain formatting.

.github/ISSUE_TEMPLATE/Bug_report.yml
driesvints: So this doesn't do what you think it does. I also used this in the beginning because I thought it would render a markdown editor. Turns out the markdown type is just so you can add some text in markdown to your form in between fields. The textarea is actually a markdown editor. Very confusing heh

jbrooksuk (approved):

PR #46697, opened 2023-04-05
[10.x] console.stub: remove void return type from handle

apreiml:
Return type of a command can be the status code as int. The void type suggests that there is no return type expected. This made me search through code and documentation to check if there's now a different way of passing the exit code.

Hence to avoid confusion I suggest to remove the void type and let the user decide later, if it should be void or int.


ankurk91:
I was going to send similar pr, but with `int` return type

PR #46702, opened 2023-04-05
[10.x] Add generic typehints to collections returned from `Stringable` methods

sebdesign:
The `Stringable` class implements a few methods that return collections, like `explode`, `split`, and `ucsplit`. We can specify the `TKey` and `TValue` templates for these collections, since these methods obviously transform a `string` to a `string[]`.

### DIFF
--- a/src/Illuminate/Support/Stringable.php
+++ b/src/Illuminate/Support/Stringable.php
@@ -254,7 +254,7 @@ class Stringable implements JsonSerializable, ArrayAccess
      *
      * @param  string  $delimiter
      * @param  int  $limit
-     * @return \Illuminate\Support\Collection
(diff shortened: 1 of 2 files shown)


src/Illuminate/Support/Stringable.php
```diff
@@ -254,7 +254,7 @@ public function excerpt($phrase = '', $options = [])
      *
      * @param  string  $delimiter
      * @param  int  $limit
-     * @return \Illuminate\Support\Collection
+     * @return \Illuminate\Support\Collection<int, string>
```
nunomaduro, line 257: Can you add tests here? https://github.com/laravel/framework/tree/10.x/types.

nunomaduro (approved):

PR #46703, opened 2023-04-05
[10.x] Add `whereEnum` to the route constraints

osbre:
Here is my use-case:

```php
Route::get('/cards/{type?}', [CardController::class, 'index'])
    ->whereEnum('type', CardFilterType::class)
    ->name('dashboard');

Route::get('/cards/{card}', [CardController::class, 'show'])
    ->whereUuid('card')
    ->name('cards.show');

class CardController
{
    public function index(CardFilterType $type = CardFilterType::Live)
```

If I remove `->whereEnum` call from the routes code, it will serve the `cards.show` route instead of a `dashboard`.


taylorotwell:

Thanks!

PR #46707, opened 2023-04-06
[10.x] Cursor pagination model casting bugfix

gdebrauwer:
In a project, I have a model that uses a custom cast on a datetime column. It converts the UTC datetime from the database to a carbon instance in a timezone defined in another column.

```php
class Booking extends Model
{
	protected $fillable = [
		'starts_at', // In UTC
		'timezone', // e.g. Europe/Brussels
	];

	protected $casts = [
		'starts' => AsDatetimeWithTimezone::class,
	];
}

// A simplified, compact version of the cast
class AsDatetimeWithTimezone implements CastsAttributes, SerializesCastableAttributes
{
    public function get($model, string $key, $value, array $attributes) : ?Carbon
    {
        return Carbon::parse($value)->setTimezone($model->timezone);
    }

    public function set($model, string $key, $value, array $attributes) : array
    {
        return [
            $key => $value->copy()->setTimezone('UTC')->format($model->getDateFormat()),
            'timezone' => $value->getTimezone()->getName(),
        ];
    }
}
```

In a part of the project, I use cursor pagination to get the bookings and sort them by the `starts_at` column. But I now discovered an issue in the cursor pagination because of the cast on the starts-at column.

To create the next/previous cursor, the cursor pagination gets the starts-at from the model, which is a carbon instance in another timezone than the database timezone (UTC), and converts that carbon instance to a string. But now the base64-encoded next/previous cursor contains a non-UTC datetime string. This causes database records to be skipped when using that cursor to get the next page.

```php
// assume a project with these database values is used to create the next cursor
Project::create([
	'starts_at' => '2023-04-01 10:00', // in UTC
	'timezone' => 'Europe/Brussels',
]);

// Then the next cursor will contain the datetime in Europe/Brussels timezone ... 
dump(base64_decode(Project::orderBy('starts_at')->cursorPaginate()->nextCursor()));
// 'starts_at' => '2023-04-01 12:00

// ... Instead of UTC timezone, which it should be because that is the value in the database
// 'starts_at' => '2023-04-01 10:00
```

To fix that, I updated the `getParametersForItem` method in the `AbstractCursorPaginator` to get the raw attribute value from the Eloquent model, so it does not use a casted value.

### DIFF
--- a/src/Illuminate/Pagination/AbstractCursorPaginator.php
+++ b/src/Illuminate/Pagination/AbstractCursorPaginator.php
@@ -214,6 +214,10 @@ abstract class AbstractCursorPaginator implements Htmlable
                 if ($item instanceof Model &&
                     ! is_null($parameter = $this->getPivotParameterForItem($item, $parameterName))) {
                     return $parameter;
+                } elseif ($item instanceof Model) {
+                    return $this->ensureParameterIsPrimitive(
+                        $item->getAttributes()[$parameterName] ?? $item->getAttributes()[Str::afterLast($parameterName, '.')]
+                    );
                 } elseif ($item instanceof ArrayAccess || is_array($item)) {
                     return $this->ensureParameterIsPrimitive(
                         $item[$parameterName] ?? $item[Str::afterLast($parameterName, '.')]
--- a/tests/Database/DatabaseEloquentIntegrationTest.php
+++ b/tests/Database/DatabaseEloquentIntegrationTest.php
@@ -6,6 +6,7 @@ use DateTimeInterface;
 use Exception;
 use Illuminate\Database\Capsule\Manager as DB;
 use Illuminate\Database\Eloquent\Builder;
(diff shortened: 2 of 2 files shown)


taylorotwell:
Couldn't this break other situations where you *do* want the cast value?

gdebrauwer:
A cast on an attribute will always change the attribute's value in some way and that cast (changed) value will not match the value in the database while it should match it, so I don't think it makes sense to ever use the cast value in the next/previous cursor.

taylorotwell:
I dunno, I feel like this is a breaking change and could probably be worked around in your own application. I would probably just define an attribute like `starts_at_utc` when casts to the UTC version of the instance and let Eloquent cursor paginate on that. Or have an attribute like `local_starting_time` that is used to get the shifted instance.

The case could be made that some people may want the behavior you are removing... to me I don't see how their argument would be any less valid. `starts_at` has a cast that puts the instance in another timezone, so that is what Eloquent is using. It seems fairly expected to me?

PR #46710, opened 2023-04-06
Allow use of pivot models on a different connection from the parent model

colincameron:
When using many-to-many relationships across databases, the connection for the pivot model is incorrectly set to the parent model's connection, even if a connection is manually specified on the pivot model.

My example:

```
class User extends Model
{
    protected $connection = 'db-a';

    public function songs()
    {
        return $this->belongsToMany(Song::class, 'UserSongs', 'userID', 'songID')
            ->using(UserSong::class);
    }
}

class Song extends Model
{
    protected $connection = 'db-b';
}

class UserSong extends Pivot
{
    protected $connection = 'db-b';
}
```

```
$song = new Song();
...
$user->songs()->attach($newSong, []);
```

The `attach()` method calls `InteractsWithPivotTable::attachUsingCustomClass()`, which calls `InteractsWithPivotTable::newPivot()`, which calls `Model::newPivot()` on the parent model.
`Model::newPivot()` calls `AsPivot::fromRawAttributes()` on the pivot class, which calls `AsPivot::fromAttributes()`, which sets the connection to the parent's connection.

This fails when the pivot table isn't on the same connection as the parent model.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/AsPivot.php
@@ -46,7 +46,7 @@ trait AsPivot
         // The pivot model is a "dynamic" model since we will set the tables dynamically
         // for the instance. This allows it work for any intermediate tables for the
         // many to many relationship that are defined by this developer's classes.
-        $instance->setConnection($parent->getConnectionName())
+        $instance->setConnection($instance->getConnectionName() ?: $parent->getConnectionName())
             ->setTable($table)
             ->forceFill($attributes)
             ->syncOriginal();


taylorotwell:
Probably a breaking change - also seems very strange to me for a pivot model to not be on the same database connection as its parent?

colincameron:
Thanks for the comments @taylorotwell 

I don't believe this is a breaking change. It is a bug fix to correctly take into account the specified connection on a pivot subclass. I believe this should be the intended behaviour rather than defaulting to the connection of the parent model. 
If a pivot class doesn't have the `protected $connection` property set, this change will have no effect.

---

As for being a strange situation - Laravel supports accessing multiple databases and defining relationships between them. In my case, I am accessing a legacy system where the users are stored in one DB, and the content is stored in a separate DB. In the inverse, the pivot and parent are in the same DB and so the bug doesn't appear.

Many-to-many relationships should be reversible by specifying the inverse relationship ([docs link](https://laravel.com/docs/10.x/eloquent-relationships#many-to-many-defining-the-inverse-of-the-relationship)). In the example given, the following code would work as expected (assuming the users relationship was defined on the song model):

```
$user = new User();
...
$song->users()->attach($user);
```

PR #46711, opened 2023-04-06
Encode array values to JSON for `DB::insert()` and `Model::create()`

newtonjob:
When dealing with Eloquent and the Base Query Builder, it is currently possible to update JSON columns by passing an array as the column value, like this;

```php
DB::table('users')->update([
    'meta' => [
        'key' => 'value'
    ]
]);

User::first()->update([
    'meta' => [
        'key' => 'value'
    ]
]);
```

This correctly encodes the `meta` column value as JSON before executing the query.

However, for inserts, this is not the case.

```php
DB::table('users')->insert([
    'name' => 'Newton Job'
    'meta'  => [
        'key' => 'value'
    ]
]); // Exception: Array to string conversion

User::create([
    'name' => 'Newton Job'
    'meta'  => [
        'key' => 'value'
    ]
]); // Exception: Array to string conversion

DB::insert("insert into users (name, meta) values (?, ?)", [
    'Newton Job'
    ['key' => 'value']
]); // Exception: Array to string conversion
```

This PR fixes that problem by updating the `Illuminate\Database\Connection::prepareBindings` method to also prepare array bindings for execution (similar to booleans and DateTime).

### DIFF
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -716,11 +716,12 @@ class Connection implements ConnectionInterface
             // We need to transform all instances of DateTimeInterface into the actual
             // date string. Each query grammar maintains its own date string format
             // so we'll just ask the grammar for the format to get from the date.
-            if ($value instanceof DateTimeInterface) {
-                $bindings[$key] = $value->format($grammar->getDateFormat());
-            } elseif (is_bool($value)) {
-                $bindings[$key] = (int) $value;
-            }
+            $bindings[$key] = match (true) {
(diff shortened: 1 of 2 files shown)


taylorotwell:
Breaking change.

newtonjob:
@taylorotwell is this really a breaking change? 

I don't think that this will affect users, but only fix a problem that one wouldn't expect out of the box.
If it works for updates, then it's a bit strange not to work for inserts.

What do you think about the situation itself?

Do I resend this to `master` instead?

Curious to know your thoughts.

PR #46712, opened 2023-04-06
[10.x] Fix nested join when not JoinClause instance

joelharkes:
Fixes issue where Nested join would build wrong sql, when joining another Query builder on it.
it would mistakenly cut 'where ' of to 're ' expecting it to say 'on ' (which it doesn't because its not a joinClause but just a regular query builder class inside).

before: 
```
select "users"."id" from "users" inner join "contacts" on "users"."id" = "contacts"."id" and (re "contacts"."id" = ?)
```

now:
```
select "users"."id" from "users" inner join "contacts" on "users"."id" = "contacts"."id" and ("contacts"."id" = ?)
```

### DIFF
--- a/src/Illuminate/Database/Query/Grammars/Grammar.php
+++ b/src/Illuminate/Database/Query/Grammars/Grammar.php
@@ -504,7 +504,7 @@ class Grammar extends BaseGrammar
         // Here we will calculate what portion of the string we need to remove. If this
         // is a join clause query, we need to remove the "on" portion of the SQL and
         // if it is a normal query we need to take the leading "where" of queries.
-        $offset = $query['query'] instanceof JoinClause ? 3 : 6;
+        $offset = $where['query'] instanceof JoinClause ? 3 : 6;
 
         return '('.substr($this->compileWheres($where['query']), $offset).')';
     }
--- a/src/Illuminate/Database/Query/Grammars/Grammar.php
+++ b/src/Illuminate/Database/Query/Grammars/Grammar.php
@@ -504,9 +504,10 @@ class Grammar extends BaseGrammar
         // Here we will calculate what portion of the string we need to remove. If this
         // is a join clause query, we need to remove the "on" portion of the SQL and
         // if it is a normal query we need to take the leading "where" of queries.
-        $offset = $query instanceof JoinClause ? 3 : 6;
+        $sql = $this->compileWheres($where['query']);
+        $offset = $query instanceof JoinClause && !str_starts_with($sql, 'where ') ? 3 : 6;
 
-        return '('.substr($this->compileWheres($where['query']), $offset).')';
+        return '('.substr($sql, $offset).')';
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Illuminate/Database/Query/Grammars/Grammar.php
```diff
@@ -504,9 +504,10 @@ protected function whereNested(Builder $query, $where)
         // Here we will calculate what portion of the string we need to remove. If this
         // is a join clause query, we need to remove the "on" portion of the SQL and
         // if it is a normal query we need to take the leading "where" of queries.
-        $offset = $query instanceof JoinClause ? 3 : 6;
+        $sql = $this->compileWheres($where['query']);
+        $offset = $query instanceof JoinClause && !str_starts_with($sql, 'where ') ? 3 : 6;
```
joelharkes, line 508: alternative solution:
```suggestion
        $offset = $offset = $where['query'] instanceof JoinClause ? 3 : 6; ? 3 : 6;
```

not sure if this would break anything else?

but the code inside again checks the instance but this class and not of the $query variable

joelharkes:
currently have problems with this in my package development: https://github.com/joelharkes/laravel-model-joins

taylorotwell:
Can you explain how you identified the fix and how it fixes the issue?

joelharkes:
> Can you explain how you identified the fix and how it fixes the issue?

I found it when working on anther PR (that you closed without much explaining why, which really bumped me out. i understand not wanting to add the code, but i'd like a proper reason why). This fix was already in that PR, but not as cleanly: https://github.com/laravel/framework/pull/46603

Joining QueryBuilder  on another Query builder

```php
$this->join(
                $aliasToUse, fn (JoinClause $join) => $join->on(
                $modelToJoin->qualifyColumn($joinColumnName),
                '=',
                $baseModel->qualifyColumn($baseColumnName),
            )->addNestedWhereQuery($builderToJoin->getQuery()),
                type: $joinType
            );
            $this->applyScopesWith($builderToJoin->getScopes(), $modelToJoin);
```

The `addNestedWhereQuery` part fails as it generates `where ` code because its not a joinClause, it fails here.
It checks the wrong variable. It checks $query, but $where['query'] is actually send downstream to generate the SQL.

```php
protected function whereNested(Builder $query, $where)
    {
        // Before
        $offset = $sql instanceof JoinClause ? 3 : 6;
        // fixed solution
        $offset = $where['query'] instanceof JoinClause ? 3 : 6;
        return '('.substr($this->compileWheres($where['query']), $offset).')';
    }
```

Downstream `$where['query'] is the one being checked here:
```php
  protected function concatenateWhereClauses($query, $sql)
    {
        $conjunction = $query instanceof JoinClause ? 'on' : 'where';

        return $conjunction.' '.$this->removeLeadingBoolean(implode(' ', $sql));
    }

```

joelharkes:
@taylorotwell 

It fixes the issue because it will no longer try to cut of "on " at the beginning of te string where it actually has "where " because its not a JoinClause

taylorotwell:
Thanks.

Regarding closing the PR - I think the reason is pretty much there, although a bit terse. Every PR I merge I am now responsible for. The person who submitted it is often not around to fix any problems years down the road and we are on our own maintaining that code for the rest of my life maintaining Laravel. Therefore, my default instinct is to reject code, not accept code. I already maintain a lot of code, and there has to be a very compelling reason for me to accept new code to maintain. 

When the PR is simple, it's obviously easier to merge, revert, etc.

PR #46713, opened 2023-04-06
Don't add bug label to every new issue

Jubeki:
I personally liked it more that the bug label was only assigned to confirmed bugs and not to every issue.

Maybe there could be a `needs-triage` label instead.

(Close the PR if you think the current behaviour is wanted)


driesvints:
Agreed, thanks @Jubeki.

PR #46714, opened 2023-04-07
Fixes the logical issue of `isDirty` && `isClean` methods 🐛

mahmoudmohamedramadan:
Hey Folks! 👋

When I was using `isDirty` and `isClean` methods I discovered a logical issue when passing multiple attributes, let's take a look over the next code for more illustration

```PHP
$user = User::first();
$user->name = "The name has changed";

return $user->isDirty(['name']);
```

The result of the previous code will be `true` because the **name** is actually changed, also let's pass the not changed attribute to the `isDirty` method.

```PHP
$user = User::first();
$user->name = "The name has changed";

return $user->isDirty(['email']);
```

The result of the previous code will be `false` because the **false** is not actually changed, but what if we pass a mix of changed and not changed attributes, what would be the result?

```PHP
$user = User::first();
$user->name = "The name has changed";

return $user->isDirty(['name', 'email']);
```

The result of the previous code will be `true` despite the `email` attribute not being changed and it does not make sense, and when I opened the source code I found that Laravel return `true` if the first attribute **only** has changed without checking the rest attributes. So, this PR solves that issue by using the logical operation between all attributes, also I've thought to print an array with boolean values for all attributes and indicate if each attribute has changed.

```PHP
$user = User::first();
$user->name = "The name has changed";

return $user->isDirty(['name', 'email']);
```

I suggest printing the result as shown in the next form instead of printing one boolean value.

```
[true, false]
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -177,6 +177,13 @@ trait HasAttributes
     public static $encrypter;
 
     /**
+     * Indicates whether all attributes are dirty.
+     *
+     * @var bool
+     */
+    protected $isAllAttributesAreDirty = false;
+
+    /**
      * Convert the model's attributes to an array.
      *
      * @return array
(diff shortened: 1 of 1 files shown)


browner12:
This behaves exactly as [documented](https://laravel.com/docs/10.x/eloquent#examining-attribute-changes). This would also be a huge breaking change that would cause unnecessary refactoring, even for a major. If you're looking for this behavior I'd recommend making multiple calls to `isDirty()`.

```php
return [$user->isDirty(['name']), $user->isDirty(['email'])];
```

ankurk91:
This will be a huge breaking change.
You can introduce a new method like
isAllDirty()
isAnyDirty()

driesvints:
We have no plans for this, sorry.

bert-w:
I wouldnt be against a method `areDirty()` to be added. Your code also doesn't work in the current format (`->isDirty(['a','b','c'])` gives the wrong result when a and c are dirty). Tests would be needed obviously.

PR #46716, opened 2023-04-07
[10.x] `setValue()` in Validator

joelharkes:
Followup of: https://github.com/laravel/framework/pull/46162

### Why

For package development, i would like to make it possible for rules to override data to make the validation 'type-safe'.
There is now no possible way to efficiently update a value inline. i can call `setData()` but it would completly reload the rules itself as well cause in big slowdown.

Alternatively I could sub-class Validator and add it myself, but this would greatly risk my package no longer being compatible with other packages AND people their own code repository.

### Usage:

```php
abstract class BaseTypeSafeRule implements ValidationRule, DataAwareRule, ValidatorAwareRule
{
    private array $data;
    private Validator $validator;

    public function setData(array $data)
    {
        $this->data = $data;
    }

    public function setValidator(Validator $validator)
    {
        $this->validator = $validator;
    }

    protected function modifyValue(string $attribute, mixed $value): void
    {
       // with this PR:
       $this->validator->setValue($attribute, $value);

        // previously: very slow!
        Arr::set($this->data, $attribute, $value);
        $this->validator->setData($this->data);
    }

    protected function translate(string $key, array $replacements = []): string|array
    {
        return $this->validator->getTranslator()->get($key, $replacements);
    }
}
```

### DIFF
--- a/src/Illuminate/Validation/Validator.php
+++ b/src/Illuminate/Validation/Validator.php
@@ -1087,6 +1087,17 @@ class Validator implements ValidatorContract
     }
 
     /**
+     * @param string $attribute key to set value for, supports dot notation.
+     * @param mixed $value value to set.
+     * @return void
+     */
+    public function setValue($attribute, $value)
+    {
+        Arr::set($this->data, $attribute, $value);
+    }
+
+
+    /**
      * Get the validation rules.
      *
      * @return array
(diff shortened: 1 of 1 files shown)


taylorotwell:
Please fix the docblock to match the rest of the framework.

joelharkes:
@taylorotwell I fixed to docs according to styleCI. 
I don't see it anywhere but is there a linter for this that i can just run to do it for me?

bert-w:
@joelharkes Remove the parameter descriptions, they are not used anywhere else. I think that should be it. There is no styleci for it atm.

mreduar:
I had been wanting to do this for a while. Thank you @joelharkes!

PR #46719, opened 2023-04-08
[10.x] Add AssertStatusCodes@assertInternalServerError

cosmastech:
Adds a helper to TestResponse for checking if the response code === 500. There is a `assertServerError` method in TestResponse, but it's checking for a range of status codes instead of precisely 500.


taylorotwell:
Hey there - I'm not adding anymore status helpers atm. You can use `assertStatus(500)` ✅

PR #46720, opened 2023-04-08
[10.x] Fix query builder whereBetween method with carbon date period

tegos:
Fixes issue when using query builder method whereBetween with param of custom carbon date period.

```php
$from = '01.03.2023';
$to = '31.03.2023';

$period = CarbonPeriod::create($from, $to);

$orders = Order::query()->whereBetween('created_at', $period)->get();

// the query will be
// select * from `orders` where `created_at` between '2023-03-01 00:00:00' and '2023-03-02 00:00:00'
// instead select * from `orders` where `created_at` between '2023-03-01 00:00:00' and '2023-03-31 00:00:00'
```

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -1253,7 +1253,7 @@ class Builder implements BuilderContract
         $type = 'between';
 
         if ($values instanceof CarbonPeriod) {
-            $values = $values->toArray();
+            $values = [$values->start, $values->end];
         }
 
         $this->wheres[] = compact('type', 'column', 'values', 'boolean', 'not');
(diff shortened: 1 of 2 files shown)


bert-w:
What's happening here? I dont understand the code comment above.

Rizky92:
> What's happening here? I dont understand the code comment above.

`CarbonPeriod` generate list of dates on daily interval when cast into array. So instead of grabbing first and last date, it gets first and next date on the array.

Tegos:
Hello, @bert-w. Don't worry, mate.

In the query builder, method whereBetween accept second parameter iterable $values, which can be a CarbonPeriod instance.

But when use CarbonPeriod parameter the method don't work properly, it take only  first date and next date insted last date of period.

See example

```php
$from = Carbon::now()->startOfYear(); // 2023-01-01
$to = Carbon::now()->endOfYear(); // 2023-12-31

$period = CarbonPeriod::create($from, $to);

$orders = Order::query()->whereBetween('created_at', $period)->get();

```

```sql
#result query
select * from `orders` where `created_at` between '2023-01-01 00:00:00' and '2023-01-02 00:00:00'
```

'2023-01-01 00:00:00' and '**2023-01-02** 00:00:00'

PR #46723, opened 2023-04-08
Convert to English

Noorani-MM:
### Request type :question:

- [ ] :bug: Bug ( fixed or found bug )
- [x] :new: New feature( make new feature or add some things new in project )
- [ ] :star2: Config ( change config files or install )
- [ ] :rocket: performance ( when you do something that can improve performance )
- [ ] :books: Document ( when code has documentation or document change )

### Description :books:

Numbers and some Arabic and Persian characters have different ASCI codes than English numbers. This causes that if these characters are stored incorrectly in the database, it will not be possible to retrieve the correct data in the future when receiving information.
For this reason, it is necessary to apply the same format to the characters so as not to face problems

# For example

Someone signup in our website with his PC. His pc's keyboard type is Arabic. Then after while he want to login with his phone,
But his phone keyboard type is English. if he use Persian or Arabic number for his password he can't login. 

What stored in database with Arabic keyboard by PC=> `'username-۱۲۳۴'`
What send with phone to login => `'uesrname-1234'`
and `'username-1234' != 'username-۱۲۳۴'`

### At store
```php
public function create_user(Request $request) {
    User::query()->firstOrCreate(['username' => Str::toEnglish($request->username)], [
        'password' => Hash::make( Str::toEnglish($request->password) )
    ]);
    return 'success'; 
}
```

### At fetch
```php
public function login (Request $request) {
    $login = false;
    if ($user = User::query()->where('username', '=', Str::toEnglish($request->username'))->first()) {
        $login = Hash::check($user->password, Str::toEnglish($request->password));
    }
    return $login ? 'success' : 'failed' ;
}
```

### Suggest
Use toEnglish method in `perpareForValidation` method. Doing this will make your code cleaner

```php
protected function prepareForValidation() {
     return $this->merge([
         'username' => Str::toEnglish($this->input('username')),
         'password' => Str::toEnglish($this->input('password')),
     ]);
}
```

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -1445,4 +1445,43 @@ class Str
         static::$camelCache = [];
         static::$studlyCache = [];
     }
+
+
+    /**
+     * Persian and Arabic character convert to English letter
+     * @param  string $string
+     * @return array|string[]
+     */
+    public static function toEnglish($string) {
+        if (is_string($string) || is_array($string)) {
+            $newNumbers = range(0, 9);
+
+            $persianDecimal = ['&#1776;', '&#1777;', '&#1778;', '&#1779;', '&#1780;', '&#1781;', '&#1782;', '&#1783;', '&#1784;', '&#1785;'];
+            $arabicDecimal = ['&#1632;', '&#1633;', '&#1634;', '&#1635;', '&#1636;', '&#1637;', '&#1638;', '&#1639;', '&#1640;', '&#1641;'];
+            $arabic = ['٠', '١', '٢', '٣', '٤', '٥', '٦', '٧', '٨', '٩'];
+            $persian = ['۰', '۱', '۲', '۳', '۴', '۵', '۶', '۷', '۸', '۹'];
+
+            $string = str_replace($persianDecimal, $newNumbers, $string);
+            $string = str_replace($arabicDecimal, $newNumbers, $string);
+            $string = str_replace($arabic, $newNumbers, $string);
+            $string = str_replace($persian, $newNumbers, $string);
+            $string = static::toPersianCharacter($string);
+        }
+
+        return $string;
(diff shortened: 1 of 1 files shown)


GrahamCampbell:
This feature doesn't make sense. We already have ascii conversion, which allows specification of the languagae.

GrahamCampbell:
And the other issue you are facing is the incorrect column collation on the database.

Noorani-MM:
> And the other issue you are facing is the incorrect column collation on the database

No that was just an example I will fixed it.

Noorani-MM:
> username-۱۲۳۴

This is the difference between toEnglish result and sample

**code**



**result**

Noorani-MM:
This is a feature that Persian or Arabic speakers usually need to use.

GrahamCampbell:
I don't think you understand either of my points here.

1. Laravel already has a way to normalize input into ascii (`Str::ascii`), and the language paramter (param 2) allows you to tell it the source language, so you can have the conversions that make the most sense for you - I don't know any arrabic, but I do know that many European languages have different conversions, so à in one language may go to `a`, but in another, to two characters.
2. Databases already have a way to specify how you want to identify characters together - it is called collation. For example, `a`, `A, `à` are all treated as the same letter by default.

GrahamCampbell:
You should also be super careful allowing non-ascii characters in user names and mushing them together, as this can result in bad security where two people have different user names that end up being stored as the same username in the database. Best practice is to have strict validation and change the collation, not to transform the input, causing collisions.

Noorani-MM:
OK I got it

Thank you for your guidance 💯

ankurk91:
Graham, you are a wizard.😅

I learnt many thing by just reading your comments on GitHub.

GrahamCampbell:
Glad I could get us pointed in the right direction. 🚀

PR #46725, opened 2023-04-09
[11.x] Efficient and Flexible Execution of Redis Lua Scripts

a1383n:
### Summary

This PR introduces new classes that offer improved handling of Redis Lua scripts, providing greater flexibility and control over script arguments and keys. These enhancements allow for easier execution of Redis Lua scripts with several benefits.

### Benefits of this

Some of the benefits of this PR include:

- Easy specification of keys and arguments to be passed to a script
- Option to enable or disable Redis Lua script caching
- Custom exception class called LuaScriptLoadException for better handling of Lua script errors
- Enhanced efficiency and flexibility of executing Redis Lua scripts, making it ideal for Redis-powered applications

### Reasons for the Changes

The changes in this PR provide an intuitive and efficient way to work with Redis Lua scripts. This is particularly useful for developers working with Redis in complex or high-performance applications. The new class also helps to minimize the network overhead associated with loading and executing Lua scripts in Redis, which can have a significant impact on application performance and latency.

### Strategy

The strategy for this Pull Request is to optimize success by prioritizing speed. To achieve this, we utilized `EVALSHA` by inputting the sha1 hash of the script provided. Initially, an error occurred because the script had not been loaded into Redis. However, we resolved this issue by loading the script into the server and subsequently re-invoking EVALSHA without any further errors.
It's important to note that the Redis script cache is always volatile. The cache may be cleared when the server restarts, during fail-over when a replica assumes the master role, or explicitly by [SCRIPT FLUSH](https://redis.io/commands/script-flush). That means that cached scripts are ephemeral, and the cache's contents can be lost at any time. [Ref](https://redis.io/docs/manual/programmability/eval-intro/#cache-volatility)

### Example usage
```php
use Illuminate\Redis\Lua\LuaScript;
use Illuminate\Redis\Lua\LuaScriptArguments;
use Illuminate\Redis\Lua\LuaScriptExecutor;
use Illuminate\Support\Facades\Redis;

$luaScript = <<<'LUA'
return redis.call("incr", KEYS[1])
LUA;

LuaScriptExecutor::create(Redis::connection())
            ->execute(LuaScript::fromPlainScript($luaScript), LuaScriptArguments::withKeys("key1"));

// Or if we want to integrate this with the Redis facade
Redis::connection()->lua()
            ->execute(LuaScript::fromPlainScript($luaScript), LuaScriptArguments::withKeys("key1"));
```

#### Checklist

- [x] Added new classes with relevant features. 
- [x] Added unit tests
- [x] Integrate with other components that used Redis Lua scripting feature


src/Illuminate/Redis/Lua/LuaScript.php
```diff
@@ -0,0 +1,186 @@
...
+     * The boolean flag that indicates whether the script should be cached or not.
+     *
+     * @var bool
+     */
+    public readonly bool $isCachingEnable;
+
+    /**
+     * Create a new LuaScript instance.
+     *
+     * @param  Connection  $connection  The Redis connection instance.
```
driesvints, line 58: Please use FQN in all DocBlocks.

src/Illuminate/Redis/Lua/LuaScript.php
```diff
@@ -0,0 +1,186 @@
...
+ * The LuaScript class represents a Redis Lua script that can be executed by a Redis connection.
+ */
+class LuaScript
+{
+    /**
+     * The Redis connection instance.
+     *
+     * @var Connection
+     */
+    public readonly Connection $connection;
```
driesvints, line 18: We currently don't use `readonly` or typed properties in the framework for internal code. Please remove these.

src/Illuminate/Redis/Lua/LuaScriptBuilder.php
```diff
@@ -0,0 +1,61 @@
...
+     * @param  Connection  $connection  The Redis connection instance
+     */
+    public function __construct(Connection $connection)
+    {
+        $this->connection = $connection;
+    }
+
+    /**
+     * The static constructor method that's create a new instance of LuaScriptBuilder
+     * @param  Connection  $connection  The Redis connection instance
```
driesvints, line 31: Please make sure to add a whiteline between the first sentence and params.

driesvints (changes requested):

src/Illuminate/Redis/Lua/LuaScript.php
a1383n: Thanks! will do

src/Illuminate/Redis/Lua/LuaScript.php
a1383n: Thanks! will do

src/Illuminate/Redis/Lua/LuaScriptBuilder.php
a1383n: Thanks! will do

a1383n:
@driesvints I have not changed or added any unit test, which is giving an error. Is there a problem with the CI Test in the `master` branch?

src/Illuminate/Redis/Lua/LuaScript.php
```diff
@@ -0,0 +1,237 @@
...
+
+    /**
+     * Check is provided SHA1 key is exists in Redis server.
+     *
+     * @return bool A boolean indicating whether the provided SHA1 key exists in the Redis server.
+     */
+    public function exist(): bool
+    {
+        if ($this->sha1 === null) {
+            return true;
```
Angel5a, line 232: Why is true? If it some kind of workaround for plain script support. it isn't intuitive.

src/Illuminate/Redis/Lua/LuaScript.php
```diff
@@ -0,0 +1,237 @@
...
+    {
+        if ($this->sha1 !== null) {
+            return $this->executeWithHash($this->sha1);
+        } else {
+            if ($this->isCachingEnable) {
+                // Try to execute the script.
+                $hash = sha1($this->script);
+                $result = $this->executeWithHash($hash);
+
+                if ($result !== false) {
```
Angel5a, line 198: Lua has a boolean type. Won't returned false from the Lua script (like "return false") get here as PHP false? Native radis command used to return integers, but I don't knows.

src/Illuminate/Redis/Lua/LuaScript.php
```diff
@@ -0,0 +1,237 @@
...
+
+    /**
+     * Execute the Lua script with the given SHA1 hash using the given Redis connection.
+     *
+     * @param  string  $sha  The SHA1 hash of the Lua script.
+     * @return mixed The result of the script execution.
+     */
+    protected function executeWithHash($sha)
+    {
+        return $this->connection->command('evalsh', [$sha, $this->keys, $this->arguments, count($this->keys)]);
```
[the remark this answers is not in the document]
a1383n: Yes, you are right. That's a mistake.

src/Illuminate/Redis/Lua/LuaScript.php
a1383n: The return value of the Lua script could be a scalar value `int`, `string`, or an `array`. Arrays that are returned can also contain other arrays. if there is an error in executing the Lua script it's it should return `false`.

src/Illuminate/Redis/Lua/LuaScript.php
a1383n: Possibly can't be called a workaround for plain script support, The `exists` method is only useable when we already provide a script sha1 hash, When we have a plain script it should return `true` or `null` or maybe throw some exception.

If you had a better idea to handle this I happy to know that

driesvints:
I have no idea why these tests are failing for you as master is currently passing. Maybe try rebasing with master?

src/Illuminate/Redis/Lua/LuaScript.php
Angel5a: Is "not provided sha1" "exists" in Redis? I think:
- no (false).
- error (null or exception).
null is also "falsy" value, so both negative cases can be checked with a single if.

Next imagine this situation. We get sha1 from some source. Can we call it? Lets check it first?
```
$sha1 = Cache::get('key_for_sha1');
$luaScript = LuaScript::fromSha1(Redis::connection(), $sha1);
if ($luaScript->exists()) {
    $luaScript->execute();
}
```
With a 'true' approach we will fail, because it's never been tested for sha1 correctness. We have to make extra check before instantiating LuaScript from it. Extra check is OK. But we either can throw `InvalidArgumentException` from `fromSha1()` for null to force the error (and to force developer to write extra check), before our code goes to far.

Second thought on it is what about caching? Will we cache our script it would become available via hash. So, asking for Redis "Is this script exists (loaded)" should return true - you're right. But until we cache this script it should return false. Perhaps we need to fill sha1 field after successful call when caching is enabled.

So what the essence of the `exists` method?
- Is the provided sha1 exists in redis? It shouldn't return true for sha1 = null. Will redis itself return true?
- Does the script can be called? It shouldn't return true, if we provide neither sha1 nor script (this check is missed).

src/Illuminate/Redis/Lua/LuaScript.php
Angel5a:
```
sail artisan tinker
Psy Shell v0.11.14 (PHP 8.2.4 — cli) by Justin Hileman
> $redis = \Illuminate\Support\Facades\Redis::connection();
= Illuminate\Redis\Connections\PhpRedisConnection {#3737}

> $redis->eval('return 2 == 2', 0);
= 1

> $redis->eval('return 2 == 3', 0);
= false

> $redis->eval('return not false', 0);
= 1

> $redis->eval('return not true', 0);
= false

> $redis->eval('return not nil', 0);
= 1

> $redis->eval('return nil', 0);
= false

> $redis->eval('redis.call("SET", "key2", "value"); redis.call("SET", "key3", "value"); return redis.call("GET", "key2") == redis.call("GET", "key3")', 0);
= 1

> $redis->eval('redis.call("SET", "key2", "value2"); redis.call("SET", "key3", "value3"); return redis.call("GET", "key2") == redis.call("GET", "key3")', 0);
= false
```
It turns out that Redis can return `false` when the script have been executed successfully. We can move out equality check on redis side (i.e. take a value fron list1, and compare it to value from list 2), we don't want to pull values to our side, we only want the result of the comparison.

Angel5a: Side note. In lua `0`, as well as empty strings, means `true`. Only `false` and `nil` test false. May be, because of it redis doesn't convert false to 0, but do convert true to 1. But idk for sure.

Angel5a: No prob. You are welcome.

Yes, `getLastError()`can help with it. But also `clearLastError()` required before eval call, because there is no any auto-clear with phpredis:
```
> $redis->eval('bad code', 0);
= false

> $redis->client()->getLastError();
= "ERR Error compiling script (new function): user_script:1: '=' expected near 'code'"

> $redis->eval('return 1 == 2', 0);
= false

> $redis->client()->getLastError();
= "ERR Error compiling script (new function): user_script:1: '=' expected near 'code'"

> $redis->eval('return "Hello"', 0);
= "Hello"

> $redis->client()->getLastError();
= "ERR Error compiling script (new function): user_script:1: '=' expected near 'code'"
```
It should be enough for phpredis.

But `predis` doesn't have anything like it. It throw exceptions:
```
> $redis->eval('bad code', 0);

   Predis\Response\ServerException  ERR Error compiling script (new function): user_script:1: '=' expected near 'code'.

> $e = $redis->evalsha(null, 0);

   Predis\Response\ServerException  NOSCRIPT No matching script. Please use EVAL.
```
Or return error object (if exceptions are switched of by config options):
```
> $redis->eval('bad code', 0);
= Predis\Response\Error {#3753}

> $e = $redis->evalsha(null, 0);
= Predis\Response\Error {#3759}

> $e->getMessage()
= "NOSCRIPT No matching script. Please use EVAL."
```
So, catch, check, proceed or throw further (don't know for sure, if it possible to write script that can be successfully cached, but throw an error been executed).

As far as Laravel support both drivers, this LuaScript implementation either should support both of them. We have no way to extend abstract Connection class with any lastError() method - predis doesn't have any, and catching all the errors to implement it manually will introduce a bad BC (user code can relay on exceptions).

It's a bunch of logic here to made this "fast call in hope of success". Strategy template is required.

Angel5a:
@a1383n  As I wrote earlier, phpredis keep last error. With a current implementation if something went wrong before call to `executeWithPlainScript()` (by some external code), then when it get called, it can get false execution error, would script execute successfully be sha1 and return `false`. You missed `clearLastError()` before `executeWithHash()` in try block.

Angel5a:
@a1383n For now you have all the logic duplicated in two classes.

I thought on "composition over inheritance" approach. Something like:
```
DriverSpecificExecutor // perhaps, ScriptExecutionResult would be better to leave only result related logic
{ 
  private $exception;
  private $returnedValue;
  public function isError()
  {
    return $this->exception || $returnedValue instance of PredisError;
  }
  public function isNoScriptError()
  {
    return (bool)"...that stuff to check noscript...";
  }
  public function executeSha1($luaScript)
  {
    try {
      $this->exception = null; // static executor for now
      $this->returnedValue = $luaScript->connection->eval($luaScript->sha1,...);
    } catch($e) {
      $this->exception = $e;
    }
  }
  public function getResult()
  {
    return !$this->exception ? $this->returnedValue : throw $this->exception;
  }
}
```
Then main code would stay near to your first iteration:
```
 // in execute script
 if($this->cached) {
    $executor->executeSha1($this);
    if (!$executor->isNoScriptError()) {
      return $executor->getResult();
    }
    // load, execute....
  }
```
This way we can ass well preserve native error handling, or introduce our own exceptions as well (which also can be determined by strategy). Also, the logic about "update sha1 or leave it empty after loadScript" also could be determined by its own strategy, responsible only on saving sha1.

Just an idea. I'm not forcing you on implementing my thought.

a1383n:
> @a1383n As I wrote earlier, phpredis keep last error. With a current implementation if something went wrong before call to `executeWithPlainScript()` (by some external code), then when it get called, it can get false execution error, would script execute successfully be sha1 and return `false`. You missed `clearLastError()` before `executeWithHash()` in try block.

Oh, you're right, I had completely forgotten

Angel5a:
@a1383n Looks like they want to rise exceptions even for phpredis driver https://github.com/laravel/framework/issues/46838 for now pr was rejected, but issue is still open.

Mastacow:

a1383n:
> @a1383n Looks like they want to rise exceptions even for phpredis driver #46838 for now pr was rejected, but issue is still 
open.

I think throwing an exception is not a good idea at all. I feel it's better to return some class called `ScriptResult` or maybe a `ScriptError`.

I am also considering a better way to make this more readable and easier to use.

driesvints:
@a1383n are you still working on this?

taylorotwell:
I feel like this would maybe be best as an external PHP package that all frameworks and libraries could use? Then, we could explore possible integrating it with Laravel?

PR #46731, opened 2023-04-10
[10.x] Added default value to Database Eloquent and Query Builder `value` and `valueRaw` methods

eusonlito:
This small change allow to add a default value to `value` and `valueRaw` database methods.

This default value will be managed by `value` global helper that allow any value or a callback/Closure.

This change doesn't break backwards compatibility and doesn't include any unwanted side effects.

The result is (for example):

```php
$title = Film::where('id', $id)->value('title', 'No Title Found');
```

```php
$title = Film::where('id', $id)->rawValue('UPPER(title)', 'NO TITLE FOUND');
```

```php
$title = Film::where('id', $id)->rawValue('UPPER(title)', fn () => $this->getDefaultTitle());
```


taylorotwell:

PR #46742, opened 2023-04-11
[10.x] feat: add methods for recursive mapping in collections.

frans-slabbekoorn:
This PR adds methods to the Collection class allowing to map an array recursively.

In my situation, I want to map all array keys to be `CamelCase`:
```php
collect($array)->mapWithKeysRecursive(fn ($item, $key) => [Str::camel($key) => $item]);
```


taylorotwell:
I suggest defining a collection macro in your own project.

PR #46745, opened 2023-04-11
[10.x] Add syntax sugar to the Process::pipe method

WendellAdriel:
## What changes

This adds a syntax sugar for the Process::pipe, allowing it to run simple commands without configuration like this:

```php
$result = Process::pipe([
    'cat test',
    'grep -i "foo"',
]);

$result->output(); // "foo"
```

```
Hello, world
foo
```

## Why

This idea was shared by @nunomaduro on Twitter: https://twitter.com/enunomaduro/status/1645805234151895041 and I think this can be a simple way to run a pipeline of simple commands when we don't need to configure the commands to run.

## Implementation Details

I thought about creating a SimplePipe class to not reuse the Pipe class for this, but IDK if that's a good idea. I'd love to get feedback and update the PR if needed.

### DIFF
--- a/src/Illuminate/Process/Factory.php
+++ b/src/Illuminate/Process/Factory.php
@@ -274,12 +274,14 @@ class Factory
     /**
      * Start defining a series of piped processes.
      *
-     * @param  callable  $callback
+     * @param  callable|array  $callback
      * @return \Illuminate\Process\Pipe
      */
-    public function pipe(callable $callback, ?callable $output = null)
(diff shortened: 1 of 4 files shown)


nunomaduro (changes requested):

WendellAdriel:
Thanks, @nunomaduro and @taylorotwell.
That makes much more sense.

WendellAdriel:
Oh, just saw the changes now
I was going to do it, but thanks for helping me on this, @taylorotwell! You rock!!! 🤘

PR #46748, opened 2023-04-11
Added actionIf method on SimpleMessage class

hasanablak:


ankurk91:
I think we already use when()

taylorotwell:
See `when`.

PR #46753, opened 2023-04-12
[10.x] Trying to get XSRF from cookie if not found anywhere else

m0nclous:
If the XSRF token is passed to the client in the cookie response, why not read it from the request cookie if it hasn't been passed anywhere else?

This will allow you not to send extra headers every time and not to show the decrypted CSRF token in the meta tag.

Reading a token from a cookie supports encryption if the middleware (`\Illuminate\Cookie\Middleware\EncryptCookies::class`) is enabled and will work correctly if disabled.

### DIFF
--- a/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php
+++ b/src/Illuminate/Foundation/Http/Middleware/VerifyCsrfToken.php
@@ -160,6 +160,10 @@ class VerifyCsrfToken
             }
         }
 
+        if (! $token && $this->shouldAddXsrfTokenCookie()) {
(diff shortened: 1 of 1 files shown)


taylorotwell:
I believe this has come up many times before - it introduces a security exploit.

Krisell:
To elaborate slightly on the security aspect – we use CSRF tokens to protect against unintended requests that still carry the session since cookies are (depending on their settings and browser defaults) auto-submitted by browsers, sometimes even when the request is originating from another website. Therefore, in order to add any protection, the CSRF token must be transfered/read via another mechanism than a cookie.

Issue #46737, opened 2023-04-11
Incorrect date format in SQL query from custom date format cast on model


vHeemstra:
### Laravel Version

10.6.2

### PHP Version

8.1.10

### Database Driver & Version

MySQL 10.4.25-MariaDB (using Xampp with Apache) on Windows 11 Pro

### Description

When updating a model using a PATCH request, a custom date attribute is not being correctly transformed to the valid MySQL date value. Resulting in the error:
```
SQLSTATE[22007]: Invalid datetime format: 1292 Incorrect date value: '20-10-2000' for column `users`.`date_of_birth` at row 1 (Connection: mysql, SQL: update `users` set `date_of_birth` = 20-10-2000, `users`.`updated_at` = 2023-04-10 11:38:19 where `id` = 1)
```

### Database (MySQL)
Column `date_of_birth` is type `DATE`.

### Migration
The column was created using `$table->date('date_of_birth')`

### Model
On the default User model, I added the attribute to fillable and set a custom date cast ([as described in the docs](https://laravel.com/docs/10.x/eloquent-mutators#date-casting)).
```php
protected $casts = [
    // other attributes
    'date_of_birth' => 'date:d-m-Y',
];

protected $fillable = [
    // other attributes
    'date_of_birth',
];
```
And according to the [documentation about date serialization](https://laravel.com/docs/10.x/eloquent-serialization#date-serialization), this should not change the format used for storing it in the database, only the format used in PHP array/JSON model instances.

### Steps/flow
1. When loading the page with an edit form, the attribute on the User model is in the correct `d-m-Y` format.
2. The post/patch request sends the date value in `d-m-Y` format as well.
<details>
	<summary>Route definition in `web.php`</summary>

```php
Route::patch('/users/{user}', [UsersController::class, 'update'])->name('users.update');
```

</details>

3. The values are correctly validated by the rule in `UserUpdateRequest`:

```php
public function rules(): array
{
    return [
        // other attributes' rules 
        'date_of_birth' => 'nullable|date_format:d-m-Y',
    ];
}
```

4. The controller fills the values in the model:

<details>
	<summary>`update` method on the controller</summary>

```php
public function update(UserUpdateRequest $request, User $user)
{
    $user->fill($request->validated());

    // dd($user->date_of_birth); <-- Shows it's a Carbon object with the correct value!

    $user->save();

    return response()->json([
        'status' => 'ok',
        'user' => $user,
    ]);
}
```

</details>

[...]

PR #46754, opened 2023-04-12
[9.x] Handle custom formatted `date` and `datetime` attributes casts distinctly

cosmastech:
Closes https://github.com/laravel/framework/issues/46737

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1085,7 +1085,8 @@ trait HasAttributes
     protected function isDateAttribute($key)
     {
         return in_array($key, $this->getDates(), true) ||
-            $this->isDateCastable($key);
+            $this->isDateCastable($key) ||
+            $this->isDateCastableWithCustomFormat($key);
     }
 
     /**
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -299,7 +299,7 @@ trait HasAttributes
             }
 
             if (isset($attributes[$key]) && ($this->isCustomDateTimeCast($value) ||
-                $this->isImmutableCustomDateTimeCast($value))) {
+                $this->isImmutableCustomDateTimeCast($value) || $this->isCustomDateCast($value))) {
                 $attributes[$key] = $attributes[$key]->format(explode(':', $value, 2)[1]);
             }
 
@@ -774,6 +774,7 @@ trait HasAttributes
             case 'collection':
                 return new BaseCollection($this->fromJson($value));
             case 'date':
+            case 'custom_date':
                 return $this->asDate($value);
             case 'datetime':
             case 'custom_datetime':
@@ -866,6 +867,8 @@ trait HasAttributes
             $convertedCastType = 'custom_datetime';
         } elseif ($this->isImmutableCustomDateTimeCast($castType)) {
             $convertedCastType = 'immutable_custom_datetime';
+        } elseif ($this->isCustomDateCast($castType)) {
+            $convertedCastType = 'custom_date';
         } elseif ($this->isDecimalCast($castType)) {
             $convertedCastType = 'decimal';
         } else {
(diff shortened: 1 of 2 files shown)


taylorotwell:
Tests fail - likely indicating this is a breaking change.

cosmastech:
> Tests fail - likely indicating this is a breaking change.

Yes, looking into this more.

cosmastech:
The expected output:
```php
Array &0 (
    0 => '2019-10-01'
    1 => '2019-10-01 10:15:20'
    2 => '2019-10-01'
    3 => '2019-10-01 10:15'
)
```
versus the actual output (from the changes in this MR):
```php
Array &0 (
    0 => '2019-10-01 00:00:00'
    1 => '2019-10-01 10:15:20'
    2 => '2019-10-01 00:00:00'
    3 => '2019-10-01 10:15:00'
)
```

To me, that seems pretty harmless and easy enough to fix (just modify the test case so that it's always including the timestamp, whether or not they will be cut off by the DB engine)

If we want to change the code to fix the issue identifier AND not include extraneous time for a date field, then it's a bigger lift.

There's a lot going on here... if I have a custom format within my model... shouldn't I be able to set the date as a string of that format?

```php
class User extends Model
{
    protected $casts = [
        'custom_date' => 'date:m-d-Y',
    ];
}

$user = new User(['custom_date' => '12-30-2023']);
$user->save(); // throws an exception for the date format being wrong
```

The problem would be that making a change like this would either have to enforce all dates are of the format specified within the cast OR some kind of indication that you want to be able to receive them in that format.

Option one would definitely be a breaking change, and option two seems kind of like bad developer experience.

cosmastech:
@taylorotwell I'm not really sure there's an easy path to fix this.

Right now, the HasAttributes@setAttribute() will never work if you have the date in a format other than `Y-m-d` (or a variant like `Y-m`) unless you explicitly pass the date in the format you want it stored in.

```php
class SomeModel extends Model
{
    protected $casts = [
        'date_field' => 'date:m-d-Y',
    ];
}

SomeModel::create(['date_field' => '12-30-2000']); // blows up because the attribute is stored as-is "12-30-2000"
SomeModel::create(['date_field' => '2000-12-30']); // works fine, though it doesn't match the cast
```

I would say this is clearly not the intended behavior, but even the test cases that this PR broke were inserting the dates in a format that didn't match with the cast. (`TestModel1::create(['date_field' => '2019-10-01']` works but `TestModel1::create(['date_field' => '2019-10'])`, which matches the model's `date_field` cast, fails)

We would could potentially do some juggling like:
```php
public function setAttribute($key, $value)
{
    //...
    elseif (! is_null($value) && $this->isDateAttribute($key)) {
        if (! $this->isStandardDateFormat($value)) {
            $dateFormat = explode(':', $this->getCasts()[$key], 2)[1];
            try {
                $value = Date::createFromFormat($dateFormat, $value);
            } catch (InvalidFormatException $exception) {
                $value = new Date($value);
            }
        }
        $value = $this->fromDateTime($value);
    }
    //...
}
```
But I'm not entirely sure it's worth it, nor do I think it would be a solution that would hit the mark (and adds overhead to this frequently used functionality).

Either that or we force the user to always insert the date in the format of the cast... but I'm sure that's a major breaking change.

It's probably worth adding in the docs something about this behavior unless there's a better plan for how to fix it.

vHeemstra:
Just my two cents here, but from the documentation it seems like:
- the attribute on the model instance is a Carbon instance
- when read as a string it would format using the format in `$casts`
- when set on the model instance it should be either a Carbon/Datetime instance or a string in the format defined in `$casts`
- when reading the attribute to use in the SQL query it should format according to the column type format (or the format set in the model's [`$dateFormat`](https://laravel.com/docs/10.x/eloquent-mutators#date-casting)).

In other words, the `$casts` property seems to only involve how attributes are formatted on the model instance and/or array/JSON representation, and not influence how it is written in the database query. I mean, choosing the format to use in the database table is already done in the migration. Or am I interpreting this wrong?

driesvints:
@cosmastech Taylor doesn't sees draft PRs. Please mark as ready if you want to get feedback.

taylorotwell:
The docs do seem to indicate the format is only to control how the dates are serialized when converting the model to arrays or JSON:

CleanShot 2023-04-13 at 08 54 17@2x

vHeemstra:
> The docs do seem to indicate the format is only to control how the dates are serialized when converting the model to arrays or JSON:
> 
> 
> 
> CleanShot 2023-04-13 at 08 54 17@2x
> 
> 
> 
> For now I don't think we'll take any action here.
> 
> 

That is true, but it should not interfere with how it is stored in the database, according to [this part](https://laravel.com/docs/10.x/eloquent-serialization#date-serialization) of the docs. But that is the case in the bug report scenario.

Issue #46750, opened 2023-04-12
Duplicate SQL constraint name when chaining index() to foreign key creation in schema builder


AyubM:
### Laravel Version

10.1.5

### PHP Version

8.1

### Database Driver & Version

MySQL

### Description

If you chain `->index()` onto a schema builder migration that also has a foreign id constraint, it appears to name the constraint `1`. This name will be used again if doing it elsewhere, causing a `ERROR 1826: Duplicate foreign key constraint name` error.

You can run
```sql
select *
from information_schema.table_constraints
where constraint_schema = 'database_name'
```

To verify that it's creating the foreign id constraint with the name "1"

### Steps To Reproduce

```php
Schema::create('first_table', function (Blueprint $table) {
            $table->id();
            $table->timestamps();
});

Schema::create('second_table', function (Blueprint $table) {
            // ...
            $table->foreignId('foreign_id')->references('id')->on('first_table')->index();
});

Schema::create('third_table', function (Blueprint $table) {
            // ...
            $table->foreignId('foreign_id')->references('id')->on('first_table')->index();
});
```

cosmastech:
Not saying this isn't an issue necessarily, but why are you applying an index to a FK column?

FKs are indexed by default with MySQL.

driesvints:
It seems we don't support what you're trying to do, sorry.

PR #46763, opened 2023-04-13
[9.x] Fixes calling index() on ForeignKeyDefinition

cosmastech:
Closes https://github.com/laravel/framework/issues/46750

The issue that referenced this was using 10.x, but I am assuming this would be considered a bug fix (if accepted) so I targeted 9.x.

Upon a bit more research, it looks like all the databases would already add an index to the FK, so if this PR is not approved, I would suggest adding something in the docs that indicates you should not use `ForeignKeyDefinition@index()` unless providing a different name for the index.


taylorotwell:
This is fixing something that isn't really even documented usage imo. The original issue was doing something that is never shown or discussed in the docs and expecting it to work.

PR #46769, opened 2023-04-13
[10.x] Cross-database support for all relations

bert-w:
Related to https://github.com/laravel/framework/issues/23042

### Issue
Various relationship methods do not work with cross database relations including `withCount` and `whereHas`. SQL errors were thrown because the database name was not prefixed in all cases. Some work has been done previously to accomplish this, however it wasn't tested thoroughly and several cases were left unconsidered (like joins and specific relation types).

### Requirements
- Any `MorphToMany`/`BelongsToMany` relations are required to have an intermediary Eloquent model defined like so, because the connection from the intermediary table cant be resolved otherwise:
```php
public function taggables()
{
    return $this->morphToMany(...)->using(Taggable::class);
}
```
- The rest of the relations work out of the box and any joins/froms that happen on a different connection automatically have their database name prefixed.

### Task status
- [x] Test all relationship methods (only correct SQL execution is tested)

### Support status
- [x] MySQL 5.7
- [x] MySQL 8.0
- [x] MariaDB 10.0
- [x] SQL Server 2019 (FIXED prefixing `dbo` was required. A custom config override `'schema'` has been added in the database configuration, similar to Postgres)
- [x] NOT SUPPORTED: PostgreSQL 14 (it requires additional pgsql extensions `dblink` and/or `postgres_fdw`)
- [x] NOT SUPPORTED: SQLite

### DIFF
--- a/src/Illuminate/Database/Eloquent/Builder.php
+++ b/src/Illuminate/Database/Eloquent/Builder.php
@@ -757,6 +757,8 @@ class Builder implements BuilderContract
 
         $constraints($relation);
 
+        $this->query->prependDatabaseNameIfCrossDatabaseQuery($relation->getBaseQuery());
+
         // Once we have the results, we just match those back up to their parent models
         // using the relationship instance. Then we just return the finished arrays
         // of models which have been eagerly hydrated and are readied for return.
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -54,6 +54,8 @@ trait QueriesRelationships
             $relation->getRelated()->newQueryWithoutRelationships(), $this
         );
 
+        $this->query->prependDatabaseNameIfCrossDatabaseQuery($hasQuery->getQuery());
+
         // Next we will call any given callback as an "anonymous" scope so they can get the
         // proper logical grouping of the where clauses if needed by this Eloquent query
         // builder. Then, we will be ready to finalize and return this query instance.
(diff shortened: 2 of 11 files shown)


Mastacow:

taylorotwell:
I don't make it a point or goal to support relationships across databases. I just don't think it's a good idea.

bert-w:
@taylorotwell then why support it "somewhat" using the existing `Illuminate\Database\Query::prependDatabaseNameIfCrossDatabaseQuery()`?

There are countless usecases and reasons to support this (think of tenancy databases or a separate administration database that links to the main one). But most of all it is yet another reason for people to pick Laravel for their application.

crissi:
@bert-w let me know if you ever make a package for this

Nickno95:
> @taylorotwell then why support it "somewhat" using the existing `Illuminate\Database\Query::prependDatabaseNameIfCrossDatabaseQuery()`?
> 
> There are countless usecases and reasons to support this (think of tenancy databases or a separate administration database that links to the main one). But most of all it is yet another reason for people to pick Laravel for their application.

the separate admin database is the use case I need this functionality for, and I'm requesting this be reconsidered. After using other built in cross database functionality like being able to set relations across models, the lack of being able to query with that relation is disappointing.

Dibbyo456:
I wish this feature was included in the core.

For future readers: there is a package [hoyvoy/laravel-cross-database-subqueries](https://github.com/hoyvoy/laravel-cross-database-subqueries) which does the same. For Laravel 11+ use fork [Dibbyo456/laravel-cross-database-subqueries](https://github.com/Dibbyo456/laravel-cross-database-subqueries) which also support MariaDB driver.

byErikas:
Sorry for necroing, but I'm requesting this to be re-reviewed, and reconsidered. The support for this is already half-way existent, and would greatly assist many multi-db laravel users

PR #46772, opened 2023-04-14
[10.x] Prompts

jessarcher:


nunomaduro (changes requested):
cc @jessarcher

src/Illuminate/Console/OutputStyle.php
```diff
@@ -68,11 +94,24 @@ public function writeln(string|iterable $messages, int $type = self::OUTPUT_NORM
...
+        if ($this->output instanceof static) {
+            return $this->output->newLinesWritten();
+        }
+
+        return $this->newLinesWritten;
+    }
+
     /**
      * {@inheritdoc}
      */
```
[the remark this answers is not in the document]
jessarcher: Yep, thanks, fixed! I thought it would inherit `@deprecated` via `@inheritdoc` but it looks like it doesn't work that way.

src/Illuminate/Console/OutputStyle.php
```diff
@@ -68,11 +94,24 @@ public function writeln(string|iterable $messages, int $type = self::OUTPUT_NORM
...
     public function newLine(int $count = 1)
     {
+        $this->newLinesWritten += $count;
         $this->newLineWritten = $count > 0;
```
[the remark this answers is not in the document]
jessarcher: Yeah, fair point.

.github/workflows/databases.yml
```diff
@@ -29,6 +29,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
jessarcher, line 33: I hate having to do this.

The issue is that `laravel/prompts` depends on `illuminate/collections ^10.0`, which creates a circular dependency. This would ordinarily be fine because the `composer.json` in this project specifies that it provides `illuminate/collections` at `self.version`. However, when doing a shallow clone in CI, none of the tags exist, so `self.version` resolves to the hash of the commit, which doesn't satisfy the dependency.

Alternatives are:

* Make `laravel/prompts` depend on `*` instead of `^10.0` but remove the ability for composer to force a known compatible version.
* Remove the dependency from `laravel/prompts` and rely on built-in PHP methods.
* Add the top-level `version` key to the `composer.json` on this repo and maintain it going forward.
* Add an extra step to CI on this project to automatically add the `version` key to `composer.json` just for the duration of the build, scraping it from `Application.php`. We'd need to be careful with any actions that make commits (e.g. update changelog) to make sure they don't include the change.

I went with this option as it felt the least hacky and problematic. The downside is that the clone is slower and bigger (an additional ~20MB of git history). If this isn't acceptable, I think I'd opt for removing collections from `laravel/prompts`.

crynobone: If `Illuminate\Support\Collection` is only used for type-hint and `$collect instanceof Collection` I believe we can skip it from `laravel/prompts` as it shouldn't cause any issue.

jessarcher: Prompts is using collections functionality in a few places, not just the types :(

PR #46773, opened 2023-04-14
[10.x] Support `insert` method on `HasOneOrMany`

mtawil:
This pull request adds support for inserting related data on `HasOneOrMany` relationships. Before this change, inserting related data using `$parent->children()->insert($data)` would result in the creation of a new child record without the foreign key being set. With this change, inserting related data will automatically set the foreign key to the primary key of the parent record, ensuring that all records are correctly linked together.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
@@ -486,4 +486,34 @@ abstract class HasOneOrMany extends Relation
     {
         return $this->localKey;
     }
+
+    /**
+     * Get the foreign attributes.
+     *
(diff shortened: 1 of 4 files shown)


taylorotwell:
`$parent->children()->create(...)`.

mtawil:
> `$parent->children()->create(...)`.

@taylorotwell I am aware of it; but if you have 1000 records, this will lead to executing 1000 queries, while the insert does only one query

mtawil:
> @taylorotwell I am aware of it; but if you have 1000 records, this will lead to executing 1000 queries, while the insert does only one query

@taylorotwell Have you seen my comment? Please let me know, as I may need to create a new PR if you haven't.

PR #46779, opened 2023-04-14
[10.x] New Validate Attribute , csv for csv mime type

kozhinhikkodan:
for csv mime type validation use `csv` 

after this update u can use
```
 $validator = Validator::make($request->all(), [
              'file' => 'required|csv'
          ]);
```


taylorotwell:
I feel like this could lead down a path of adding of ton of similar methods. "pdf", "sound", "video", etc.

PR #46780, opened 2023-04-14
Add headers

sobhan-m94:


src/Illuminate/Foundation/Application.php
```diff
@@ -1218,7 +1218,7 @@ public function isDownForMaintenance()
     public function abort($code, $message = '', array $headers = [])
     {
         if ($code == 404) {
-            throw new NotFoundHttpException($message);
+            throw new NotFoundHttpException($message, headers: $headers);
```
GrahamCampbell, line 1221: Named parameters should be avoided.

GrahamCampbell (changes requested):

PR #46784, opened 2023-04-14
[10.x] Add when/unless to $fail of closure validation rule

gdebrauwer:
When using a closure validation rule, you always have an if-statement:
```php
    if ($value !== 'foo') {
        return;
    }

    $fail('value should be foo');
}

// or

    if ($value === 'foo') {
        $fail('value should be foo');
    }   
}

// or

    if ($fails) {
        $fail('value should be foo');
    }   
}
```

So I thought: why not add a when/unless-method to the $fail variable? This makes the code a lot cleaner in my opinion.
```php
    // or
    // or
        // split on multiple lines
    });
}
```
It is a pretty small change in the ClosureValidationRule that makes a big DX difference in my opinion.

### DIFF
--- a/src/Illuminate/Validation/ClosureValidationRule.php
+++ b/src/Illuminate/Validation/ClosureValidationRule.php
@@ -59,8 +59,6 @@ class ClosureValidationRule implements RuleContract, ValidatorAwareRule
         $this->failed = false;
 
         $this->callback->__invoke($attribute, $value, function ($attribute, $message = null) {
-            $this->failed = true;
-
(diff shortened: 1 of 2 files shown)


michaelnabil230:
Good job 👍 
But why don't use trait 'Conditionable'

taylorotwell:
I actually really dig the idea but unfortunately we couldn't bring the same behavior to rule objects since the contract forces `Closure`, which feels like a pretty big bummer to me and makes things feel inconsistent. 🥲

https://laravel.com/docs/10.x/validation#using-rule-objects

PR #46788, opened 2023-04-15
[10.x] Add when/unless to $fail of (closure/class) validation rule (improved PR of #46784)

gdebrauwer:
**_Improved version of #46784. The original PR only worked on $fail of a closure validation rule, but this PR adds the same behavior to the $fail of custom validation rule classes_**

When using a closure validation rule (or a [custom validation rule class](https://laravel.com/docs/10.x/validation#using-rule-objects)), you always have an if-statement:
```php
    if ($value !== 'foo') {
        return;
    }

    $fail('value should be foo');
}

// or

    if ($value === 'foo') {
        $fail('value should be foo');
    }   
}

// or

    if ($fails) {
        $fail('value should be foo');
    }   
}
```

So I thought: why not add a when/unless-method to the $fail variable? This makes the code a lot cleaner in my opinion. It also reads natural: "it fails when ..." or "it fails unless ...".
```php
    // or
    // or
        // split on multiple lines
    });
}
```
It is a pretty small change that makes a big DX difference in my opinion.

### DIFF
--- a/src/Illuminate/Validation/ClosureValidationRule.php
+++ b/src/Illuminate/Validation/ClosureValidationRule.php
@@ -59,8 +59,6 @@ class ClosureValidationRule implements RuleContract, ValidatorAwareRule
         $this->failed = false;
 
         $this->callback->__invoke($attribute, $value, function ($attribute, $message = null) {
-            $this->failed = true;
-
             return $this->pendingPotentiallyTranslatedString($attribute, $message);
         });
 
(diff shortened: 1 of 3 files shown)


ankurk91:
For me `if\else` is more readable here.

gdebrauwer:
You can still use if/else. It's an optional syntax.

taylorotwell:
@gdebrauwer invokable validation rules aren't the current, documented way of writing class based validation rules though:

https://laravel.com/docs/10.x/validation#using-rule-objects

gdebrauwer:
@taylorotwell but the "invokable validation rule" class supports both the old way and the new way:

https://github.com/laravel/framework/blob/3e5a6b38f6b180e57b1adb3e89945465892ee216/src/Illuminate/Validation/InvokableValidationRule.php#L95-L103

In the tests that I added, you can also see that I used the new way

PR #46790, opened 2023-04-15
[10.x] Support `insert` method on `HasOneOrMany`

mtawil:
This pull request adds support for inserting related data on `HasOneOrMany` relationships. Before this change, inserting related data using `$parent->children()->insert(...)` would result in the creation of a new child record without the foreign key being set. With this change, inserting related data will automatically set the foreign key to the primary key of the parent record, ensuring that all records are correctly linked together.

Please be aware that using `$parent->children()->create(...)` is not equivalent to using the `insert` method. When dealing with a large number of records (e.g. 1000), the `create` method will create each record one by one, leading to the execution of 1000 individual queries. In contrast, using `insert` will only execute one query.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
@@ -486,4 +486,34 @@ abstract class HasOneOrMany extends Relation
     {
         return $this->localKey;
     }
+
+    /**
+     * Get the foreign attributes.
+     *
+     * @return array
+     */
+    public function getForeignAttributes()
+    {
+        return [$this->getForeignKeyName() => $this->getParentKey()];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function insert(array $values)
+    {
+        if (empty($values)) {
+            return true;
+        }
+
+        if (! is_array(reset($values))) {
+            $values = [$values];
+        }
+
+        foreach ($values as $key => $value) {
+            $values[$key] = array_merge($value, $this->getForeignAttributes());
+        }
+
+        return parent::insert($values);
+    }
 }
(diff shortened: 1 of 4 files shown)


src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
```diff
@@ -486,4 +486,34 @@ public function getLocalKeyName()
...
+        if (empty($values)) {
+            return true;
+        }
+
+        if (! is_array(reset($values))) {
+            $values = [$values];
+        }
+
+        foreach ($values as $key => $value) {
+            $values[$key] = array_merge($value, $this->getForeignAttributes());
```
cosmastech, line 514: should the order be swapped, giving the user the ability to override defaults if they so choose?
```php suggestion
            $values[$key] = array_merge($this->getForeignAttributes(), $value);
```

src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
```diff
@@ -486,4 +486,34 @@ public function getLocalKeyName()
...
+        return [$this->getForeignKeyName() => $this->getParentKey()];
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function insert(array $values)
+    {
+        if (empty($values)) {
+            return true;
```
cosmastech, line 506: Curious: why return true here? In my own code, I would throw an Exception here, but I imagine for adding to the framework, it should return false?

src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
mtawil: Swapping the order is unnecessary since the user cannot manually set foreign key values in the `create` method. Also, the order should be maintained to ensure data integrity in relationships where foreign keys should be set automatically.

src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
mtawil: https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Query/Builder.php#L3261-L3272

cosmastech: Interesting. Never noticed that before. Thanks!

taylorotwell:

PR #46792, opened 2023-04-15
[10.x] Add newCollection method to Fluent class

denniseilander:
This pull request is inspired by the nice solution @stevebauman wrote in an [article](https://stevebauman.ca/create-api-responses-with-eloquent-factories/) about how to solve a problem using of the `Factory::times(2)` method on a Fluent-based Model factory.

This PR resolves an issue where an error will be thrown when using the `MyFactory::times(2)` method with Model Factories using the `Fluent::class` instead of an Eloquent model.

```
TypeError : Illuminate\Database\Eloquent\Factories\Factory::callAfterMaking(): Argument #1 ($instances) must be of type Illuminate\Support\Collection, Illuminate\Support\Fluent given
```

This error occurs because the `Fluent::class` does not have a `newCollection()` method, so passes a single instance of the `Fluent::class` instead of a `Collection` to the `Factory::callAfterMaking()`.

While the Model Factories isn't intended to be used by non-eloquent instances, it really works great for generating objects which doesn't need to be persisted to the database.

Please let me know you what you think.

### DIFF
--- a/src/Illuminate/Support/Fluent.php
+++ b/src/Illuminate/Support/Fluent.php
@@ -96,6 +96,17 @@ class Fluent implements Arrayable, ArrayAccess, Jsonable, JsonSerializable
     }
 
     /**
+     * Create a new Collection instance.
+     *
+     * @param  array  $instances
+     * @return \Illuminate\Support\Collection<TKey, TValue>
+     */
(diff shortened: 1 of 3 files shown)


taylorotwell:
Honestly not sure this method belongs here. Feels really random.

PR #46797, opened 2023-04-16
typo fix

rmrtnv:
small typo found


X-Coder264:
This is **not** a typo. If the logger cannot be resolved from the container we want the original exception to be (re)thrown, we don't wanna throw the container exception that the logger could not be resolved.

driesvints:
Thanks @X-Coder264

PR #46801, opened 2023-04-16
[10.x] Add possibility to set backoff method in broadcast event

dammy001:

### DIFF
--- a/src/Illuminate/Broadcasting/BroadcastEvent.php
+++ b/src/Illuminate/Broadcasting/BroadcastEvent.php
@@ -53,7 +53,6 @@ class BroadcastEvent implements ShouldQueue
         $this->event = $event;
         $this->tries = property_exists($event, 'tries') ? $event->tries : null;
         $this->timeout = property_exists($event, 'timeout') ? $event->timeout : null;
-        $this->backoff = property_exists($event, 'backoff') ? $event->backoff : null;
         $this->afterCommit = property_exists($event, 'afterCommit') ? $event->afterCommit : null;
     }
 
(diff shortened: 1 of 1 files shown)


jmarcher:
This will break a lot of things

dammy001:
@jmarcher @driesvints at which will this cause a breaking change?

driesvints:
You're no longer setting the backoff property at object initialization time.

dammy001:
Yeah but if you check pushOn method inside the queue class, the backoff checks for property or method been provided by the job/event passed.

jmarcher:
What Dries and I meant is that if you create a new BroadcastEvent class the property is no longer there.

dammy001:
@jmarcher @driesvints check the below lines because I'm not sure there's any breaking change

https://github.com/laravel/framework/blob/997218b35f5a7f2d5cb7c640e690a5af03499658/src/Illuminate/Broadcasting/BroadcastManager.php#L171

https://github.com/laravel/framework/blob/997218b35f5a7f2d5cb7c640e690a5af03499658/src/Illuminate/Broadcasting/BroadcastManager.php#L183

https://github.com/laravel/framework/blob/997218b35f5a7f2d5cb7c640e690a5af03499658/src/Illuminate/Queue/Queue.php#L147

https://github.com/laravel/framework/blob/997218b35f5a7f2d5cb7c640e690a5af03499658/src/Illuminate/Queue/Queue.php#L192

jmarcher:
I know what you are thinking, but if I create the class and want to access the backoff property directly, this is no longer initialized as before, your change makes the value of the backoff property depend on another class

dammy001:
@jmarcher i guess I'll add the at the constructor level then

Issue #46537, opened 2023-03-21
Queue job implementing ShouldQueue and ShouldBeUnique dispatched via dispatch()->afterResponse() acquires lock and never releases


spiechu:
- Laravel Version: 9.39.0
- PHP Version: 8.1.8
- Database Driver & Version: MariaDB

### Description:
Queue job implementing `ShouldQueue` and `ShouldBeUnique` and dispatched via `Job::dispatch()->afterResponse()`
acquires lock using `UniqueLock` in `PendingDispatch::shouldDispatch()`.
Then job is run using `app(Dispatcher::class)->dispatchAfterResponse($this->job)`.
After job is done there is nothing that runs something similar to `CallQueuedHandler::ensureUniqueJobLockIsReleased()`.

### Steps To Reproduce:
1. Create queue job implementing `ShouldQueue, ShouldBeUnique`
2. Implement `uniqueId()`
3. Dispatch via `Job::dispatch()->afterResponse()`
4. Dispatch again via `Job::dispatch()->afterResponse()` and it will not work since lock is still acquired.

cosmastech:
Assuming the PR I put up doesn't get merged, I believe you can resolve this by adding:
```php
Bus\Dispatcher::map([YourJob::class => YourUniqueJobHandler::class]);
```

To your `AppServiceProvider@boot()` method. The `UniqueJobHandler` class should have either a `handle()` or `__invoke()` method that receives `YourJob $job` and handles running the job and deleting the lock if successful.

driesvints:
Thanks!

PR #46806, opened 2023-04-17
[9.x] Release lock for job implementing `ShouldBeUnique` that is dispatched `afterResponse()`

cosmastech:
To resolve https://github.com/laravel/framework/issues/46537

* Changed `Dispatchable@dispatchAfterResponse()` so that `PendingDispatch@shouldDispatch()` checks for lock, rather than just pushing immediately to the Application::$terminatingCallbacks array. Could move this to a separate PR, as maybe you would consider it a breaking change. Seems kind of strange that `Job::dispatchAfterResponse()` doesn't check for uniqueness, but `Job::dispatch()->afterResponse()` does.
* Moves the handling of middleware/unique locks out of `CallQueuedHandler@call()` to its own method `CallQueuedHandler@handle()`. I named the method `handle()` so that it could be more easily used with `Bus\Dispatcher@dispatchNow()` (which is expecting the `$handler` argument to implement either `__invoke()` or `handle()`.
  * I think normally I would move this to its own action class or something, but really wanted to make as few changes as possible to get this working.

Jobs dispatched after response should now also be routed through middleware as well. Might make sense to add a test case for that, just let me know.

### DIFF
--- a/src/Illuminate/Bus/Dispatcher.php
+++ b/src/Illuminate/Bus/Dispatcher.php
@@ -263,7 +263,7 @@ class Dispatcher implements QueueingDispatcher
     public function dispatchAfterResponse($command, $handler = null)
     {
         $this->container->terminating(function () use ($command, $handler) {
-            $this->dispatchNow($command, $handler);
+            $this->dispatchSync($command, $handler);
         });
     }
 
--- a/src/Illuminate/Foundation/Bus/Dispatchable.php
+++ b/src/Illuminate/Foundation/Bus/Dispatchable.php
@@ -96,7 +96,7 @@ trait Dispatchable
      */
     public static function dispatchAfterResponse(...$arguments)
     {
-        return app(Dispatcher::class)->dispatchAfterResponse(new static(...$arguments));
+        return self::dispatch(...$arguments)->afterResponse();
     }
 
     /**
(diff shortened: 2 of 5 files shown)


taylorotwell:
Is this not as simple as just changing `dispatchAfterResponse` on the Bus `Dispatcher` to use `dispatchSync` instead of `dispatchNow`.

`dispatchNow` is not recommended and is quasi deprecated. That simple change fixes the issue for me.

```php
public function dispatchAfterResponse($command, $handler = null)
{
    $this->container->terminating(function () use ($command, $handler) {
        $this->dispatchSync($command, $handler);
    });
}
```

cosmastech:
> Is this not as simple as just changing `dispatchAfterResponse` on the Bus `Dispatcher` to use `dispatchSync` instead of `dispatchNow`?
> 
> `dispatchNow` is not recommended and is quasi deprecated because of the fact that it doesn't apply middleware, etc. That simple change fixes the issue for me.

image

You're right.

I've reverted the other changes I made, though I kept the change to `Dispatchable::dispatchAfterResponse()`. Without it, there's never a call to `PendingDispatch@shouldDispatch()`

Tweaked the test cases slightly to use both Job::dispatch()->afterResponse() and Job::dispatchAfterResponse();

PR #46807, opened 2023-04-17
[10.x] Query result indexing using `keyBy` (`PDO::FETCH_UNIQUE`)

bert-w:
**TL; DR**: PHP's PDO currently has a mode called `PDO::FETCH_UNIQUE` which allows you to retrieve query results and assign array keys directly without having to call a collection method like `$query->get()->keyBy()`. This speeds up computation in many cases where you need arrays with a given key. 

-------

`PDO::FETCH_UNIQUE` is a custom fetch mode which is passable to `$pdoStatement->fetchAll(PDO::FETCH_UNIQUE)` ([PHP link](https://www.php.net/manual/en/pdostatement.fetchall.php)). Currently, this mode defaults to `PDO::FETCH_OBJ` which returns the query results as a list containing of objects (note: an array is a "list" when its keys consist of consecutive numbers starting from 0, so there are no custom keys). `PDO::FETCH_UNIQUE` however can be used to set the keys of this array *as the result is coming in through the PDO interface*. 

This speeds up execution considerably when you need a query result that has custom keys. The current way to do this operation is to first retrieve a query using `$query->get()` and *then* key the array using the collection method `$query->get()->keyBy('name')`. This loops the whole array again, which is something that could have been done initially when the array was created.

Retrieving array items by key is essentially an O(1) operation (constant time; it will always be fast even for big arrays) so there is much value in using keyed arrays (see Usecases below), especially when arrays become bigger.

`PDO::FETCH_UNIQUE` eats up the first column of the query resultset as the array key, so the implementation takes care of prepending the custom `keyBy($column)` to the SELECT-list. This extra column can then freely be consumed by PDO to use as the array key, unaffecting the rest of the query result.

#### 1. Key an array by a column:
```php
$users = User::keyBy('email')->get();
$user = $users['foobar@example.com'] ?? null;
```
#### 2. Key an array by a custom expression:
```php
// Retrieve a user for each day (taking the first 10 characters of the timestamp, which is `2023-01-01`):
$users = User::keyBy(DB::raw('SUBSTR(`created_at`, 1, 10)'));
$janFirstUser = $users['2023-01-01'] ?? null;
```
Note that this query can likely be optimized by adding a `->groupBy()` column since there could be multiple users per date. Not including any `groupBy` will loop the complete result array and overwrite any array key that is encountered multiple times. So if multiple users have `created_at = 2023-04-01`, the last user with this value will occupy the array key. This is exactly how `Illuminate\Support\Collection::keyBy()` works as well.
#### 3. Keying relations:
```php
class Post extends BaseModel
{
    public function tags()
    {
        // Return results keyed by the `tag` column (the name of the tag).
        return $this->belongsToMany(Tag::class)->keyBy('tag');
    }
}
$post = Post::with('tags')->first();
$awesomeTag = $post->tags['awesome'] ?? null;
```

#### Example 1: creating a local cache of currency objects, which can be quickly retrieved by their iso-code (EUR,USD,BTC etc) array index:
```php
class SomeClass
{
    private $_cache;

    public function get($isoCode)
    {
        return $this->cache()[$isoCode] ?? null;
    }

    protected function cache()
    {
        if (is_null($this->_cache)) {
            $this->_cache = App\Models\Currency::query()->keyBy('iso_code')->get();
        }

        return $this->_cache;
    }
}

$euroCurrency = (new SomeClass)->get('EUR');
```

#### Example 2: any cases where you need to check key existence in a query result
(see Feature no. 3)

#### Example 3: Optimize any cases where `->get()->keyBy(...)` (or similar) is called:
```php
// Old:
Country::all()->keyBy('country_code');
// New:
Country::keyBy('country_code')->get();
```


Angel5a:
Do you plan to support really any `$query->get()->keyBy()` cases?
```
Post::with('extra.depth')->get()->keyBy('extra.depth.field');
```

crissi:
Could the pluck(key, value) method with two params be optimized in a similar way?

bert-w:
> Could the pluck(key, value) method with two params be optimized in a similar way?

Maybe a combination of `PDO::FETCH_COLUMN|PDO::FETCH_UNIQUE` can improve that function indeed. Good idea for another PR though.

Mastacow:

Hubgov54:
Less try

Hubgov54:
Try

Hubgov54 (commented):
Try

bert-w:
rip branch because of merge issues; i'll remake this

Hubgov54:
I can help you, drop your WhatsApp number so I can handle it sir

Hubgov54:
Okay I can fixed it sir . Send me your WhatsApp or telegram number so I can
text you and fixed it sir

PR #46811, opened 2023-04-17
[10.x] Allow separate directory for locks on filestore

Gaitholabi:
Issue: 

When using the `file` driver, upon executing `php artisan cache:clear` or calling `\Cache::flush();`, all locks are cleared.

Similar to what other drivers offer, this PR adds the ability to define a separate directory for Locks when using the `file` driver.

Example:
```
        'file' => [
            'driver' => 'file',
            'path'   => storage_path().'/framework/cache',
            'lock_path' => storage_path().'/framework/lock',
        ],
 ```

- [X]  Backward compatible, not adding the `lock_path` config would yield the same result as before
- [X]  Makes the FileStore driver more consistent with other drivers namely RedisStore and Database store

### DIFF
--- a/src/Illuminate/Cache/FileStore.php
+++ b/src/Illuminate/Cache/FileStore.php
@@ -36,6 +36,11 @@ class FileStore implements Store, LockProvider
     protected $filePermission;
 
     /**
+     * The path to the lock directory
+     */
+    protected ?string $lockPath = null;
+
+    /**
      * Create a new file cache store instance.
      *
      * @param  \Illuminate\Filesystem\Filesystem  $files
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/Illuminate/Cache/FileStore.php [hunk 1]
```diff
@@ -35,6 +35,11 @@ class FileStore implements Store, LockProvider
      */
     protected $filePermission;
 
+    /**
+     * The path to the lock directory.
+     */
+    protected ?string $lockPath = null;
```
driesvints, line 41: Please remove the property typehint here and add a `@var` annotation

```suggestion
     *
     * @var string|null $lockPath
     */
    protected $lockPath;
```

driesvints (changes requested):

Issue #45800, opened 2023-01-25
Bug with named morphTo relationships and LazyLoadingViolationException


Sergiobop:
- Laravel Version: 9.48
- PHP Version: 8.1
- Database Driver & Version: psql

### Description:
When you have a custom named morphTo relationship, Laravel throws the LazyLoadingViolationException, even though the relationship is loaded.

### Steps To Reproduce:
Given the model "Model", which has a nullable morph, 'item'

Model 

```
public function item(): MorphTo
{
      return $this->morphTo();
}

public function simple_item(): MorphTo
{
      return $this->morphTo('item')->select(['id', 'name']);
}
```

If we do:

```

$models = Model::query()->with(['simple_item'])->get();

foreach($models as $model) {
      $itemName = $model->item->name;
}

```

The exception is thrown:

```
Attempted to lazy load [item] on model [App\Models\Model] but lazy loading is disabled
````

If we dd, the query, the relationship 'item' is filled and the 'simple_item' is null:

```
$models = Model::query()->with(['simple_item'])->get()->dd();
```

```
#relations: array:3 [
      "simple_item" => null
      "item" => App\Models\.. 
]
```

If we debug the model to see what relationships are loaded:
```
dd($this->relationLoaded('item')); // false even though item is filled with the model
dd($this->relationLoaded('simple_item')); // true even though simple_item is null
```

UPDATE:

As a workaround, this works without additional database queries:
```
$models = Model::query()->with(['simple_item'])->get();

foreach($models as $model) {
      $model->loadMissing('item'); // This line actually doesn't call the DB
      $itemName = $model->item->name;
}
```

**Comment by eduance** [comment]
We need to gather some more details to assist you in your journey.

1. Could you specify which model you're trying to actually make morphable?
2.  When you call ``Model::query()``, which model are you actually referring to?
3.  What is the value of ``$modelsShouldPreventLazyLoading`` inside your model?

After creating a fresh Laravel app, reproducing the steps I was unable to recreate this bug. Could you provide a repository with this exact issue?

This was my setup:

```$images = Image::query()->with(['item'])->get();```

```
class Image extends Model
{
    use HasFactory;

    protected $fillable = [
        'url',
        'imageable_id',
        'imageable_type'
    ];

    /**
     * Get the parent model.
     *
     * @return \Illuminate\Database\Eloquent\Relations\MorphTo
     */
    public function imageable()
    {
        return $this->morphTo('item');
    }

    /**
     * Get the parent model.
     *
[...]

PR #46813, opened 2023-04-18
[9.x] Use relation name when loading MorphTo relations

cosmastech:
https://github.com/laravel/framework/issues/45800 Should resolve this issue.

If we call:
```php
$images = Image::query()->with(['simplified_imageable'])->get();

$images[0]->relationLoaded('imageable'); // ❌ false
$images[0]->relationLoaded('simplified_imageable'); // ✅ true
```

Not sure if the desired behavior is to set both relationships or not. To me, if the calling code is eager loading a relationship that relies on another, it shouldn't also set the parent relationship.

## Changes to existing test
I moved the creation of models out of the `defineDatabaseMigrationsAfterDatabaseRefreshed` method and put them directly into the test that utilizes them.


taylorotwell:
We never document morphTo functions with custom names as even being supported.

PR #46814, opened 2023-04-18
[10.x] Add when/unless to $fail of (closure/class) validation rule

gdebrauwer:
Resubmission of #46788. That PR was closed because of a misunderstanding. The `InvokableValidationeRule` class is used for both new validation rule classes (= with a `validate` method) and old validation rule classes (= with an `_invoke` method).
https://github.com/laravel/framework/blob/3e5a6b38f6b180e57b1adb3e89945465892ee216/src/Illuminate/Validation/InvokableValidationRule.php#L95-L103

-----

When using a closure validation rule (or a [custom validation rule class](https://laravel.com/docs/10.x/validation#using-rule-objects)), you always have an if-statement:
```php
    if ($value !== 'foo') {
        return;
    }

    $fail('value should be foo');
}

// or

    if ($value === 'foo') {
        $fail('value should be foo');
    }   
}

// or

    if ($fails) {
        $fail('value should be foo');
    }   
}
```

So I thought: why not add a when/unless-method to the $fail variable? This makes the code a lot cleaner in my opinion. It also reads natural: "it fails when ..." or "it fails unless ...".
```php
    // or
    // or
        // split on multiple lines
    });
}
```
It is a pretty small change that makes a big DX difference in my opinion.

### DIFF
--- a/src/Illuminate/Validation/ClosureValidationRule.php
+++ b/src/Illuminate/Validation/ClosureValidationRule.php
@@ -101,10 +101,17 @@ class ClosureValidationRule implements RuleContract, ValidatorAwareRule
             ? fn ($message) => $this->messages[] = $message
             : fn ($message) => $this->messages[$attribute] = $message;
 
-        $fail = fn () => $this->failed = true;
+        $onFailure = fn () => $this->failed = true;
 
-        return new class($message ?? $attribute, $this->validator->getTranslator(), $destructor, $fail) extends PotentiallyTranslatedString
+        return new class($message ?? $attribute, $this->validator->getTranslator(), $destructor, $onFailure) extends PotentiallyTranslatedString
         {
+            /**
+             * Indicates if the validation callback failed.
+             *
+             * @var bool
+             */
+            public $failed = true;
+
             /**
              * The callback to call when the object destructs.
              *
@@ -117,14 +124,7 @@ class ClosureValidationRule implements RuleContract, ValidatorAwareRule
              *
              * @var \Closure
              */
-            protected $fail;
-
-            /**
-             * Indicates if the validation callback failed.
-             *
-             * @var bool
-             */
-            public $failed = true;
+            protected $onFailure;
 
             /**
              * Create a new pending potentially translated string.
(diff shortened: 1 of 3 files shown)


timacdonald:
@gdebrauwer I dig the idea of this, however I'm wondering if it really generalises as well as we might hope.

For a _single_ failure call it is really nice, but what would something like this look like using this new syntax?

```php
    if (...) {
        $fail('...');
        return;
    }

    if (...) {
        return $fail('...');
        return;
    }

    if (...) {
        return $fail('...');
        return;
    }

    // passes
},
```

You'd have to re-introduce the `if` statements, as `$fail` doesn't throw an exception.

donnysim:
I'm with @timacdonald on this. This looks nice on paper but I don't see it really be used in many cases - most of the rules I've written either returns after a $fail call to prevent useless checks further or the $fail is the last thing in the call with early success returns. This seems like it would only work in one time checks and introduce useless additional checks to avoid double adding failures or even avoiding database calls that would not be present in the first place and you'd just fall back to the initial if/else instead.  Also sometimes you pass a translated message to the $fail where you need some values from the result which yet again, this makes it more complicated.

gdebrauwer:
I guess you could maybe achieve that by assigning variables, something like this?
```php
    $fail('...')->when($foo = ...);
	$fail('...')->when($foo && ...);
},
```
You are right that this syntax might not be the preferred syntax when you check multiple things in a single (closure/class) validation rule. The syntax is at its best when you only have a single check, but multiple checks in one rule is something I have not really encountered a lot in my projects (but that might just be me 😅).

taylorotwell:
@gdebrauwer I don't know if I'm really stupid or just missing something.

Do `make:rule` in a brand new Laravel 10.x application.

The `$fail` parameter is literally type-hinted as a `Closure`, and that type-hint is baked into the interface.

```php
<?php

namespace App\Rules;

use Closure;
use Illuminate\Contracts\Validation\ValidationRule;

class Uppercase implements ValidationRule
{
    /**
     * Run the validation rule.
     *
     * @param  \Closure(string): \Illuminate\Translation\PotentiallyTranslatedString  $fail
     */
    public function validate(string $attribute, mixed $value, Closure $fail): void
    {
        
    }
}

```

PR #46816, opened 2023-04-18
[10.x] Pass boolean validation with "true" and "false" strings

hendrikheil:
This PR aims aims to adjust `boolean` validation in such a way that is also supports the strings `"true"` and `"false"`. 

We noticed that validation is perfect when validating `application/json` as there are discrete types available, but fails when validating requests made with `multipart/form-data`. 
Since `multipart/form-data` can only be a buffer or a string, you'd currently have to send a boolean as `0` or `1` to get proper validation. However, that is quite the hassle when using libraries like Axios, where you just change from `axios.post()` to `axios.postForm()` and don't really control serialization

### DIFF
--- a/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
+++ b/src/Illuminate/Validation/Concerns/ValidatesAttributes.php
@@ -468,7 +468,7 @@ trait ValidatesAttributes
      */
     public function validateBoolean($attribute, $value)
     {
-        $acceptable = [true, false, 0, 1, '0', '1'];
+        $acceptable = [true, false, 0, 1, 'false', 'true', '0', '1'];
 
         return in_array($value, $acceptable, true);
     }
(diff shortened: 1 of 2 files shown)


ankurk91:
I was going to send the same PR

taylorotwell:
It's a breaking change to start accepting these strings. However, you can easily transform those values by writing a middleware that extends Laravel's `TransformsRequest` middleware:

```php
<?php

namespace App\Http\Middleware;

use Illuminate\Foundation\Http\Middleware\TransformsRequest;

class TransformBooleans
{
    /**
     * Transform the given value.
     *
     * @param  string  $key
     * @param  mixed  $value
     * @return mixed
     */
    protected function transform($key, $value)
    {
        return match($value) {
            'true' => true,
            'false' => false,
            default => $value,
        };
    }
}
```

hendrikheil:
@taylorotwell Does it make sense to include this for future versions? I can see this tripping up a lot of people and not everyone debugging this super quickly. The middleware definitely works as a workaround, but it feels like Laravel should validate "true" and "false" just like "1" and "0" too.
Maybe 10.x is not the right target, but 11.x might be?

johanrosenson:
@hendrikheil 

IMO it doesn't make sense, it would make it impossible to cast a validated boolean value without string handling, this works today: 

```php
$boolean = (bool) $validated['boolean_input'];
```

with the proposed change this would not work anymore since the string `"false"` would also be cast to `true`.

PR #46817, opened 2023-04-18
[10.x] whereMorphedTo  null

m-derakhshi:
i use nullableMorphs in table:
```
$table->nullableMorphs('sectionable');
```

and my query is : 

```

select * from visitor_info` where `selectable_type` is null

```

i want use eloquent and phpstorm autocomplate:

```

Model:query()->whereMorphedTo('sectionable', null)->get();

```

laravel error:

```

Call to a member function getMorphClass() on null

```


cosmastech:
I think this is a useful addition, but you should add tests to make it more likely to be approved.

mderakhshi:
> I think this is a useful addition, but you should add tests to make it more likely to be approved.

I don't know where and how to write the test for this section

taylorotwell:

Thanks!

PR #46821, opened 2023-04-18
[10.x] whereMorphedTo null

m-derakhshi:
i use nullableMorphs in table:
```
$table->nullableMorphs('sectionable');
```

and my query is : 

```

select * from visitor_info` where `selectable_type` is null

```

i want use eloquent and phpstorm autocomplate:

```

Model:query()->whereMorphedTo('sectionable', null)->get();

```

laravel error:

```

Call to a member function getMorphClass() on null

```


m-derakhshi (commented):
refresh

Issue #46633, opened 2023-03-30
Filtering Queries Via Intermediate Table Columns on withCount()


ronssij:
- Laravel Version: 9.52.0
- PHP Version: 8.1.14
- Database Driver & Version: MySQL 8.0.31

### Description:
You cannot apply the `$query->wherePivotNotNull('link')` on the `withCount()` function.
And `wherePivotNotNull` is treated as a column `pivot_not_null` like on a `where` clause magic.

### Steps To Reproduce:
SocilaMedia Model
```
public function users()
    {
        return $this->belongsToMany(User::class)
            ->using(SocialMediaUser::class)
            ->as('social_user_link')
            ->withPivot('link')
            ->withTimestamps();
    }
```

```
$links = SocialMedia::query()
            ->with([
                'users' => fn ($query) => $query->wherePivotNotNull('link'),
                'icon'
            ])
            ->withCount([
                'users' => fn ($query) => $query->wherePivotNotNull('link')
            ])
            ->ordered()
            ->get();

return SocialMediaLinkResource::collection($links);
```

The error
```
SQLSTATE[42S22]: Column not found: 1054 Unknown column 'pivot_not_null' in 'where clause' (SQL: select `social_media`.*, (select count(*) from `users` inner join `social_media_user` on `users`.`id` = `social_media_user`.`user_id` where `social_media`.`id` = `social_media_user`.`social_media_id` and `pivot_not_null` = link and `blocked_at` is null) as `users_count` from `social_media` order by `order` asc)",
    "exception": "Illuminate\\Database\\QueryException
```

The generated query
```
SELECT
	`social_media`.*,
	(
		SELECT
			count(*)
		FROM
			`users`
			INNER JOIN `social_media_user` ON `users`.`id` = `social_media_user`.`user_id`
		WHERE
			`social_media`.`id` = `social_media_user`.`social_media_id`
			AND `pivot_not_null` = link
			AND `blocked_at` IS NULL) AS `users_count`
	FROM
		`social_media`
	ORDER BY
		`order` ASC
```

But when you add the where clause to the relation function it works.

```
public function users()
    {
        return $this->belongsToMany(User::class)
            ->using(SocialMediaUser::class)
            ->as('social_user_link')
            ->withPivot('link')
            ->wherePivotNotNull('link')
            ->withTimestamps();
    }
```

ronssij:
@driesvints Yup I already posted it on discord someone commented to post it here.

driesvints:
[...]

PR #46831, opened 2023-04-19
[9.x] Allow Eloquent `Builder@withCount` to apply constraints on Relation

cosmastech:
https://github.com/laravel/framework/issues/46633

Not sure if this is a breaking change, but targeting 9.x as I believe it's a bug.

Say that SomeModel has a BelongsToMany relationship. Previously, if you passed a constrained relationship to withCount(), it would always pass an Eloquent\Builder instance as the first parameter to the closure. With this change, it should pass the Relation (HasMany, HasOne, MorphTo) instead.

```php
SomeModel::query()->withCount([
    'related_model' => fn (BelongsToMany $query) => $query->wherePivotNull('some_pivot_column'),
])->get();
```

Previously, this would've thrown an exception for the closure expecting `BelongsToMany`. Assuming you did not add a type hint, the Builder instance would've handled the constraint like `$query->where('pivot_null', 'some_pivot_column')`.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/QueriesRelationships.php
@@ -633,7 +633,8 @@ trait QueriesRelationships
                 $expression = $column;
             }
 
-            // Apply constraints to the relation.
+            // Apply constraints to the relation so the user can rely on the
+            // scopes/helpers of the relation rather than the Builder.
             $constraints($relation);
 
             // Here, we will grab the relationship sub-query and prepare to add it to the main query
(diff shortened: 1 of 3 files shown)


cosmastech:
It looks like the pipelines are failing because of a docker connection issue.

Is there anything I should do as a contributor when this happens? It doesn't look like I can manually trigger them, so the only other thing would be a force push.

driesvints:
@cosmastech I retriggered the build for you.

Also want to thank you for recent PRs you made 👍

cosmastech:
> @cosmastech I retriggered the build for you.
> 
> Also want to thank you for recent PRs you made 👍

It does bring me pleasure to imagine the eye roll you and Taylor give when you see another PR submitted by me. 😆 

Thank you all for maintaining this framework and associated packages.

cosmastech:
It looks like [the docs indicate that the closure should expect a Builder](https://github.com/laravel/docs/blob/10.x/eloquent-relationships.md?plain=1#L1449) instance as the first parameter.

I personally think there's benefit to giving access to the Relation rather than a Builder instance, so maybe this could target 11.x?

Mastacow:

taylorotwell:

PR #46833, opened 2023-04-19
[10.x] Allow pruning all cancelled and unfinished queue batches

jameshulse:
### Overview

I was previously unable to prune all queue batches on my local system during developemnt. This is due to a bug with how `php artisan queue:prune-batches` handles the flags `--unfinished=0 --cancelled=0`. With this change users will have a slightly nicer experience while developing against the Queue Batches feature.

_Before:_

image

_After:_

image

### Changes in depth

There is a minor bug in the `php artisan queue:prune-batches` command when trying to delete all batches through pruning with the number of hours set to `0`.

The command allows you to prune unfinished and cancelled batches with the `--unfinished` and `--cancelled` flags respectively. These flags take a value of the number of hours to prune from, however, unlike the main `--hours` flag, these flags previously didn't support passing `0` (zero) as a value.

When passing `--unfinished=0` or `--cancelled=0` the command **wouldn't** acknowledge that you had passed the option at all. This was due to the check for these options:

```php
if ($this->option('unfinished')) {
    ...
}
```

When `0` is passed then the above will evaluate to `false`. The fix is to explicitly check whether the option is `null`:

```
if ($this->option('unfinished') !== null) {
    ...
}
```

### Testing

All Laravel tests passed locally after this change. I added a basic test to cover each flag with a value of `0`.

### Future

It could be useful to have a flag like `php artisan queue:prune-batches --all`. Although this flag would be potentially dangerous in production, it would short-hand the command that I have been running during development.

### DIFF
--- a/src/Illuminate/Queue/Console/PruneBatchesCommand.php
+++ b/src/Illuminate/Queue/Console/PruneBatchesCommand.php
@@ -46,7 +46,7 @@ class PruneBatchesCommand extends Command
 
         $this->components->info("{$count} entries deleted.");
 
-        if ($this->option('unfinished')) {
+        if ($this->option('unfinished') !== null) {
             $count = 0;
 
             if ($repository instanceof DatabaseBatchRepository) {
@@ -56,7 +56,7 @@ class PruneBatchesCommand extends Command
             $this->components->info("{$count} unfinished entries deleted.");
         }
 
-        if ($this->option('cancelled')) {
+        if ($this->option('cancelled') !== null) {
             $count = 0;
 
             if ($repository instanceof DatabaseBatchRepository) {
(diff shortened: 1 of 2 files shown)


francoism90:
Does this work with Laravel Horizon as well? It would be really useful on development or when you launched an incorrect batch for example.

dammy001:
@francoism90 No. It doesn't work with horizon.

francoism90:
@dammy001 Ok, maybe it could be added at some point.

Thanks for the PR. :+1:

jameshulse:
> Does this work with Laravel Horizon as well? It would be really useful on development or when you launched an incorrect batch for example.

If I understand your request correctly, it actually _does_ work with Horizon as that was what I was using. Although Horizon uses Redis for tracking jobs, it still uses the database to track batches.

PR #46836, opened 2023-04-20
[10.x] Add toSentence method to Str helper

geisi:
This pull request introduces a new `toSentence` method to the Str helper class, providing a convenient way to convert an array or a Collection of strings into a human-readable `sentence` with commas and a customizable conjunction.

The toSentence method is heavily inspired by the Ruby on Rails [to_sentence helper](https://apidock.com/rails/Array/to_sentence).

I have added the method to the Str Helper, but it would also be convenient to move the method to the Array helper.

### Examples

```PHP
// Using an array of strings
$itemsArray = ['apple', 'banana', 'cherry'];
$sentenceFromArray = Str::toSentence($itemsArray);
// Output: "apple, banana, and cherry"
```
```PHP
// Using a Laravel Collection
$itemsCollection = collect(['apple', 'banana', 'cherry']);
$sentenceFromCollection = Str::toSentence($itemsCollection);
// Output: "apple, banana, and cherry"
```
```PHP
// Using a custom conjunction
$sentenceWithCustomConjunction = Str::toSentence($itemsArray, 'or');
// Output: "apple, banana, or cherry"
```

Thank you very much!

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -1437,9 +1437,8 @@ class Str
     /**
      * Convert a list of strings into a sentence, with commas and a conjunction.
      *
-     * @param iterable<string> $items
-     * @param string           $conjunction
-     *
+     * @param  iterable<string>  $items
+     * @param  string  $conjunction
      * @return string The generated sentence.
      */
     public static function toSentence($items, $conjunction = 'and')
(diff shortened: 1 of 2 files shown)


taylorotwell:
See above. This is possible using `Arr::join` or `Collection::join`. 👍

taylorotwell:
Actually the Rails version works a bit differently. Regardless, I think it would be a better fit for the Arr or Collection class instead of Str.

Issue #46838, opened 2023-04-20
Phpredis may get an error,but not throw RedisException


miaotiao:
### Laravel Version

10.8.0

### PHP Version

8.1.18

### Description

When I use Alibaba Cloud’s Redis services, it display a  "ERR 'EVAL' command keys must in same slot" through function 'getLastError'  ,but does not throw a RedisException .

### Steps To Reproduce

I have no idea ,but look this [phpredis](https://github.com/phpredis/phpredis#eval)

![image](https://user-images.githubusercontent.com/28383349/233341101-a748bdbd-9d5c-4f1f-a626-08878d9e9fc1.png)

miaotiao:
Throwing a Redisexception might be better 😁

Angel5a:
The same doc states, that:
> phpredis throws a [RedisException](https://github.com/phpredis/phpredis#class-redisexception) object if it can't reach the Redis server. That can happen in case of connectivity issues, if the Redis service is down, or if the redis host is overloaded. **In any other problematic case that does not involve an unreachable server (such as a key not existing, an invalid command, etc), phpredis will return FALSE.**

Throwing exceptions would be a breaking BC. If we want that behavior, than we need to provide it via configuration option (which would be "no throw" by default for 10.x branch).

Other solution is to use Laravel-specific exception independent of the underlying driver in upcoming to 11.x feature.

nunomaduro:
Taylor explained that this has the potential to disrupt current applications. It already returns `false` which indicates an error for all other commands. You can get the underlying client via `Redis::client()` to get the last error message.

@miaotiao, feel free to submit a pull request to enhance the documentation on this topic.  Thank you for creating this issue, and for your pull request.

PR #46839, opened 2023-04-20
Phpredis may get an error,but not throw RedisException

miaotiao:
[#46838](https://github.com/laravel/framework/issues/46838#issue-1676460467)


miaotiao:
If the 10.* branch is merged successfully, can it be merged into the 9.* branch?

taylorotwell:
This could break existing applications. It already returns `false` which indicates an error for all other commands. You can get the underlying client via `Redis::client()` to get the last error message.

PR #46844, opened 2023-04-21
[10.x] Add `bindWhen` to Container

milwad-dev:
When you want to add condition for binding you must follow like this:
```php
if (! $this->app->isProduction()) {
    $this->app->bind(Implicit::class, function (Application $app) {
       return new ImplicitMain;
    });
}
```

But is not beautiful for Laravel framework!

We can use `bindWhen`:
```php
$this->app->bindWhen(! $this->app->isProduction(), Implicit::class, function (Application $app) {
   return new ImplicitMain;
});
```

### DIFF
--- a/src/Illuminate/Contracts/Container/Container.php
+++ b/src/Illuminate/Contracts/Container/Container.php
@@ -73,17 +73,6 @@ interface Container extends ContainerInterface
     public function bindIf($abstract, $concrete = null, $shared = false);
 
     /**
-     * Register a binding when condition is true.
-     *
-     * @param  bool  $condition
-     * @param  string  $abstract
-     * @param  \Closure|string|null  $concrete
-     * @param  bool  $shared
-     * @return void
-     */
-    public function bindWhen($condition, $abstract, $concrete = null, $shared = false);
-
-    /**
      * Register a shared binding in the container.
      *
      * @param  string  $abstract
--- a/src/Illuminate/Contracts/Container/Container.php
+++ b/src/Illuminate/Contracts/Container/Container.php
@@ -82,7 +82,7 @@ interface Container extends ContainerInterface
      * @return void
      */
     public function bindWhen($condition, $abstract, $concrete = null, $shared = false);
-    
+
     /**
      * Register a shared binding in the container.
      *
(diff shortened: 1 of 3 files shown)


src/Illuminate/Contracts/Container/Container.php
```diff
@@ -72,6 +72,17 @@ public function bindMethod($method, $callback);
...
+    /**
+     * Register a binding when condition is true.
+     *
+     * @param  bool  $condition
+     * @param  string  $abstract
+     * @param  \Closure|string|null  $concrete
+     * @param  bool  $shared
+     * @return void
+     */
+    public function bindWhen($condition, $abstract, $concrete = null, $shared = false);
```
driesvints, line 84: We can't add methods to contracts on patch releases. Please remove this or send this in for master (11.x). Thanks

milwad-dev:
@driesvints 
If merged, I will add `bindWhen` to Contract in separate PR.

tests/Container/ContainerTest.php
```diff
@@ -66,6 +66,19 @@ public function testBindIfDoesRegisterIfServiceNotRegisteredYet()
...
 
+    public function testBindWhenIfConditionIsTrue()
+    {
+        $container = new Container;
+        $container->bind('surname', function () {
+            return 'Taylor';
+        });
+        $container->bindWhen(true, 'name', function () {
+            return 'Dayle';
+        });
```
binaryk, line 77: What about a test when condition is false?

milwad-dev: The class cannot be `bind'

tests/Container/ContainerTest.php
```diff
@@ -66,6 +66,19 @@ public function testBindIfDoesRegisterIfServiceNotRegisteredYet()
         $this->assertSame('Dayle', $container->make('name'));
     }
 
+    public function testBindWhenIfConditionIsTrue()
+    {
+        $container = new Container;
+        $container->bind('surname', function () {
```
jmarcher, line 72: Shouldn't this be `name` instead of `surname`?

milwad-dev: No, under that is important

jmarcher: Why is it there? I thought your were trying to show how a binding gets overridden.

milwad-dev: Fix it.

ankurk91:
I feel that framework is crowded with such alias/helper method everywhere.
These kind of PRs are distracting the Laravel core team from focusing on real world issues.

taylorotwell:

PR #46845, opened 2023-04-21
[10.x] Removes redundant exception throwing

miranalmehrab:
In these following test methods, we are expecting ``AssertionFailedError`` exceptions to be thrown. But when the actual and expected value does not match, ``assertStatus`` method already throws ``ExpectationFailedException`` exception which ``extends`` the ``AssertionFailedError`` exception class. So according to **Liskov Substitution Principle**, we don't need to manually throw ``AssertionFailedError`` exceptions on our own.


miranalmehrab:
Hi @taylorotwell, I was wondering if you could kindly provide me with some feedback on why the PR was closed. I do value your opinion and would like to understand any issues or concerns that you may have had with the changes that I proposed.

Issue #46849, opened 2023-04-21
[Bug/Feature] Validating files in nested arrays.


pxpm:
### Laravel Version

10.2

### PHP Version

8.2

### Description

Hey guys! 👋 

I am having troubles using the `File` validation rule in nested arrays.

It fails with the following error when using the `File` validation rule in a nested array:
![image](https://user-images.githubusercontent.com/7188159/233690408-bdecfd37-04ba-4de3-bbe4-9c48f5cc64d5.png)

I've set this bug/feature because I am not sure if the behavior is intended or it was forgotten to add a `toString()` method on `File` rule. 

Cheers

### Steps To Reproduce

```php
// this works
array.*.input => ['file', 'max:1024']

//this don't
array.*.input => [File::types(['pdf'])->max(1024)]
```

driesvints:
Thanks. Let's see how your PR goes.

PR #46850, opened 2023-04-21
Add __toString() method to File validation rule

pxpm:
In case #46849 is actually something that can be fixed, this is a possible solution for it.

### DIFF
--- a/src/Illuminate/Validation/Rules/File.php
+++ b/src/Illuminate/Validation/Rules/File.php
@@ -326,4 +326,14 @@ class File implements Rule, DataAwareRule, ValidatorAwareRule
 
         return $this;
     }
+    
+    /**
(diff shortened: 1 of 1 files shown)


driesvints (approved):

taylorotwell:
I'm not sure this is really the right solution? We don't really want to convert it to a string. We want it to be preserved as a rule object so that it's passes method is called, etc. Probably better to fix the rule parsing logic.

pxpm:
Hey @taylorotwell thanks for the review.
From my understanding you will be getting the string only when you require a string representation of the class. 
As far as I understood is how other validation rules work.

Anyway do you plan to work on this or should I attempted the fix on the rule parsing logic? 

Cheers

PR #46853, opened 2023-04-22
[10.x] The model binding should respect the requested parameters by an action

christoph-kluge:
As of now the model binding requires that the paramters within actions will be defined in correct order.
If you only need the last parameter then you need to bloat up the list of parameters.

The following MR will reduce the need and only binds what is requested by the developer.

Before:

```
Route::domain('{account}.example.com')->group(function () {
    Route::post('/post/{post}/tag/{tag}', function (UpdateTagRequest $request, Account $account, Post $post, Tag $tag) {
        // update tag
    });
)};
```

After

```
Route::domain('{account}.example.com')->group(function () {
    Route::post('/post/{post}/tag/{tag}', function (UpdateTagRequest $request, Tag $tag) {
        // update tag
    });
)};
```


Mastacow:

Issue #46852, opened 2023-04-22
[Testing] UploadedFile::fake()->image() produce a 'Test code or tested code did not (only) close its own output buffers'


ivrahhh:
### Laravel Version

10.8.0

### PHP Version

8.2

### Description

Using `UploadedFile::fake()->image('image.jpg')`. Instead of producing a fake image it cuts the test and produce **Test code or tested code did not (only) close its own output buffers**.

This is a snippet code of my test.

```
test('author can create a new book', function () {
    $cover = UploadedFile::fake()->image('image.jpg');

    $response = actingAs($this->user)->post(
        uri: route('books.store') ,
        data:[
            'book_title' => 'Book Title 1',
            'synopsis' => 'Book Test Sample Synopsis',
            'tags' => $this->tags->random(5),
            'genres' => $this->genres->random(3),
            'cover_photo' => $cover,
        ]
    );

    $response->assertValid();
    assertDatabaseHas(Book::class,[
        'book_title' => 'Book Title 1',
        'synopsis' => 'Book Test Sample Synopsis',
    ]);
});

```

Test Result :

```
   WARN  Tests\Feature\Author\AuthorBookTest
  ! author can create a new book → Test code or tested code did not (only) close its own output buffers                                                       1.30s

  Tests:    1 risky (0 assertions)
  Duration: 1.82s
```

Im using PestPHP but I also try it in PHPUnit. It's still the same. But using `UploadedFile::fake()->create()` works.

### Steps To Reproduce

Just how you normally create a new test.

1. `php artisan make:test Author/AuthorBookTest`
2. Type the test.
3. `php artisan test --filter=AuthorBookTest`

cosmastech:
Did you try creating a fresh repo and testing this? I just tried and could not reproduce. I'm not sure if it's something within your controller that is causing the issue or if it is actually UploadedFile::fake() that is causing the problem.

```php
public function testBasic()
{
    $cover = UploadedFile::fake()->image('image.jpg');

    $this->get('/')->assertOk();
}
```

This test is not indicated as risky.

Are you running via Docker? What platform/architecture are you on? What version of PHPUnit are you running? Could you upload a repo with the minimal amount of data to produce the issue so we can triage this problem?

cosmastech:
I submitted [a PR to phpunit](https://github.com/sebastianbergmann/phpunit/pull/5343/files) since the risky message doesn't indicate whether or not output buffers were closed. That would help diagnose the exact problem.

[...]

PR #46859, opened 2023-04-23
[10.x] Throw LogicException when calling `FileFactory@image()` if mimetype is not supported

cosmastech:
The issue here was illuminated by https://github.com/laravel/framework/issues/46852. 

I opted for checking if `function_exists` rather than `extension_loaded` since this would allow for a developer to write their own `imagecreatetruecolor()` and `imagejpeg()` (or whatever extension) to output the contents.

### DIFF
--- a/src/Illuminate/Http/Testing/FileFactory.php
+++ b/src/Illuminate/Http/Testing/FileFactory.php
@@ -49,6 +49,8 @@ class FileFactory
      * @param  int  $width
      * @param  int  $height
      * @return \Illuminate\Http\Testing\File
+     *
+     * @throws \LogicException
      */
     public function image($name, $width = 10, $height = 10)
     {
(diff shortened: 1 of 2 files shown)


Mastacow:

src/Illuminate/Http/Testing/FileFactory.php
```diff
@@ -76,6 +86,12 @@ protected function generateImage($width, $height, $extension)
 
             $image = imagecreatetruecolor($width, $height);
 
+            if (! function_exists($functionName = "image{$extension}")) {
+                ob_get_clean();
+
+                throw new LogicException("{$functionName} function is not defined and image cannot be generated.");
+            }
+
             call_user_func("image{$extension}", $image);
```
cosmastech, line 95: We should change this to:
```php suggestion
            call_user_func($functionName, $image);
```

PR #46866, opened 2023-04-24
[10.x] Update collection docs to use template-covariant for the values

Rocksheep:
This PR will allow developers to use covariants in their collections. Before this was not possible and you would get errors from PHPStan. This is also an issue in older versions of Laravel. The following example would throw PHPStan errors

```php
class Animal {}
class Tiger extends Animal {}
class Lion extends Animal {}
class Zebra extends Animal {}

class Zoo {
    /**
     * @var \Illuminate\Support\Collection<int, Animal>
     */
    private Collection $animals;

        $this->animals = collect([
            new Tiger,
            new Lion,
            new Zebra,
        ]);
    }

    /**
     * @return \Illuminate\Support\Collection<int, Animal>
     */
    public function getWithoutZebras(): Collection {
        return $this->animals->filter(fn (Animal $animal) => ! $animal instanceof Zebra);
    }
}
```

The thrown error would be
```
Property Zoo::$animals (Illuminate\Support\Collection<int, Animal>) does not accept Illuminate\Support\Collection<int, Lion|Tiger|Zebra>.
```

Changing the `@template` to `@template-covariant` fixes this.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -12,7 +12,8 @@ use Traversable;
 
 /**
  * @template TKey of array-key
- * @template TValue
+ *
+ * @template-covariant TValue
  *
  * @implements \ArrayAccess<TKey, TValue>
  * @implements \Illuminate\Support\Enumerable<TKey, TValue>
--- a/src/Illuminate/Collections/Enumerable.php
+++ b/src/Illuminate/Collections/Enumerable.php
@@ -12,7 +12,8 @@ use Traversable;
 
 /**
  * @template TKey of array-key
- * @template TValue
+ *
+ * @template-covariant TValue
  *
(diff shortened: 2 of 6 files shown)


Rocksheep:
@nunomaduro Yeah! Psalm definitely has support for it as can be read on their [documentation](https://psalm.dev/docs/annotating_code/templated_annotations/) and here are screenshots to display that PHPStorm also understands it. 
afbeelding
afbeelding

Rocksheep:
Ah sorry I thought I had pushed it out of the review state because of a push I did earlier. I didn't add it to the database collections because I didn't run into issues with that yet, because I don't usually pull multiple models from the database in a single call.

So I have only changed the files where I ran into the problem and I missed the helper.

And about that arrayable. I thought it would have the same problem, but checking now it doesn't seem so. I will remove it and add it to the collection helpers. 

Do you think the other classes you mentioned need the template change?

Issue #46851, opened 2023-04-21
[Database] Many-to-Many: detach() method builds incomplete pivot object


shehi:
### Laravel Version

10.8.0

### PHP Version

8.1.17

### Database Driver & Version

MariaDb 10.6

### Description

When detaching, e.g.:
```
$ids = [...];
$user->posts()->detach($ids);
```
https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php#L478 is triggered, which builds a Pivot object only with pivot-key attributes (e.g. `user_id` and `post_id`). The remaining pivot attributes don't get included. This becomes a problem especially when this deleted pivot object is passed to some `deleted` event. Deleted object not having all its attributes causes issues where those attributes are needed. And that object being already deleted, unfortunately can't be `refresh()`ed.

At https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php#L513 we can see it potentially could build full pivot object if `$this->pivotValues` were populated. But the latter is empty.

![image](https://user-images.githubusercontent.com/653569/233748212-b2f1fde5-1377-4983-ac30-4ea42f75ddda.png)

### Steps To Reproduce

Description should be enough.

cosmastech:
Looks like the expected behavior is to hydrate the pivot models and delete them yourself. 🤷

cosmastech:
@driesvints can this be closed? The PR to resolve it was rejected, so I'm guessing this is expected behavior.

driesvints:
Thanks. It indeed seems we're not going to make any changes here right now, sorry.

PR #46874, opened 2023-04-24
[10.x] Fetch Pivot fully before detaching

cosmastech:
https://github.com/laravel/framework/issues/46851

This allows for dispatching model events on a Pivot when detaching that includes *all* attributes of the Pivot model. Previously models only contained their foreign and related keys populated.

This was in particular a problem for model event listeners/observers: additional pivot columns would not be available to those functions if they needed to act based on pivot columns within the `deleted` or `deleting` events.

### Additional thoughts
It might be nice to have a distinct set of events for detaching/detached. I think this might be an idea for 11.x. My thinking is that it might provide better granularity to the client code, but I can't think of specific a use case at this time.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
+++ b/src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
@@ -473,32 +473,35 @@ trait InteractsWithPivotTable
      */
     protected function detachUsingCustomClass($ids)
     {
-        $results = 0;
-
-        foreach ($this->parseIds($ids) as $id) {
-            $results += $this->newPivot([
-                $this->foreignPivotKey => $this->parent->{$this->parentKey},
-                $this->relatedPivotKey => $id,
-            ], true)->delete();
-        }
-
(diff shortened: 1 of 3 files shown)


src/Illuminate/Database/Eloquent/Relations/Concerns/InteractsWithPivotTable.php
```diff
@@ -473,32 +473,36 @@ public function detach($ids = null, $touch = true)
...
-    protected function getCurrentlyAttachedPivots()
+    protected function getCurrentlyAttachedPivots($ids = null)
+    {
+        $methodName = is_null($ids) ? 'newPivotQuery' : 'newPivotStatementForId';
+
+        return $this->{$methodName}($ids)->get()->map($this->mapIntoPivot(...));
+    }
+
+    /**
+     * @param  \stdClass  $record
```
driesvints, line 496: Add a 1 line sentence description.

taylorotwell:

michel-silva:
This code resolve the problem. Can you @taylorotwell  considering merge this code?

Issue #46865, opened 2023-04-24
Fallback HTTP error pages (4xx.blade.php and 5xx.blade.php) do not work for certain important error codes.


WEBPerformace:
### Laravel Version

10.4.1

### PHP Version

8.2.4

### Description

I tested it and:

- `4xx.blade.php` does not work for `404` (while working for other `4xx` codes).
- `5xx.blade.php` does not work for `500` and `503` (while working for other `5xx` codes).

This forces us to create duplicate files with the exact error codes in the name, i.e. `404.blade.php`, `500.blade.php`, and `503.blade.php`.

### Steps To Reproduce

1. Add custom `4xx.blade.php` and `5xx.blade.php` files to your `resources/views/errors` directory
2. Test them with different error codes (using `abort($code)`).
3. These templates will not work for `404`, `500`, and `503` errors, but they will work for other `4xx` and `5xx` codes.

driesvints:
This is because Laravel ships with dedicated pages for these. You'll always have to overwrite these yourself.

weshooper:
This seems to contradict the documentation @ https://laravel.com/docs/11.x/errors#fallback-http-error-pages ?

PR #46877, opened 2023-04-25
Make error pages more flexible and informative (#46865)

WEBPerformace:
This change continues the effort to make working with error pages more flexible and easy.

Some time ago, Laravel began supporting "fallback error pages," which was a great improvement. However, the built-in error pages were still provided in an old-fashioned manner, limiting some possibilities and forcing users to create more custom error pages to replace the built-in ones.

This pull request addresses that issue. Now, users can simply place their `4xx.blade.php` and `5xx.blade.php` files in their error template folders, and they're good to go! There is no longer a need to create multiple files for each individual error code to override the built-in pages.

Additionally, this PR adds the error code to the `<title>` tag and the error message to the "message" section if provided.

### DIFF
--- a/src/Illuminate/Foundation/Exceptions/views/4xx.blade.php
+++ b/src/Illuminate/Foundation/Exceptions/views/4xx.blade.php
@@ -1,0 +1,11 @@
+@extends('errors::minimal')
+
+@php
+    $statusCode = $exception->getStatusCode();
+    $statusText = __(\Symfony\Component\HttpFoundation\Response::$statusTexts[$statusCode]);
+    $statusMessage = !empty($exception->getMessage()) ? ': ' . __($exception->getMessage()) : '';
+@endphp
+
+@section('title', "$statusCode: $statusText")
+@section('code', $statusCode)
+@section('message', $statusText . $statusMessage )
--- a/src/Illuminate/Foundation/Exceptions/views/5xx.blade.php
+++ b/src/Illuminate/Foundation/Exceptions/views/5xx.blade.php
@@ -1,0 +1,11 @@
+@extends('errors::minimal')
+
+@php
+    $statusCode = $exception->getStatusCode();
+    $statusText = __(\Symfony\Component\HttpFoundation\Response::$statusTexts[$statusCode]);
+    $statusMessage = !empty($exception->getMessage()) ? ': ' . __($exception->getMessage()) : '';
+@endphp
+
+@section('title', "$statusCode: $statusText")
+@section('code', $statusCode)
+@section('message', $statusText . $statusMessage )
--- a/src/Illuminate/Foundation/Exceptions/views/401.blade.php
+++ b/src/Illuminate/Foundation/Exceptions/views/401.blade.php
@@ -1,5 +1,0 @@
-@extends('errors::minimal')
-
-@section('title', __('Unauthorized'))
-@section('code', '401')
-@section('message', __('Unauthorized'))
--- a/src/Illuminate/Foundation/Exceptions/views/402.blade.php
+++ b/src/Illuminate/Foundation/Exceptions/views/402.blade.php
@@ -1,5 +1,0 @@
-@extends('errors::minimal')
-
-@section('title', __('Payment Required'))
-@section('code', '402')
-@section('message', __('Payment Required'))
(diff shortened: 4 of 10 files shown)


WEBPerformace:
Oops, I see that some tests are failing. Sorry for that, I'll fix my pull request soon!

src/Illuminate/Foundation/Exceptions/views/4xx.blade.php
```diff
@@ -0,0 +1,11 @@
+@extends('errors::minimal')
+
+@php
+    $statusCode = $exception->getStatusCode();
+    $statusText = __(\Symfony\Component\HttpFoundation\Response::$statusTexts[$statusCode]);
+    $statusMessage = !empty($exception->getMessage()) ? ': ' . __($exception->getMessage()) : '';
```
GrahamCampbell, line 6: We should never show raw exception messages to users.

src/Illuminate/Foundation/Exceptions/views/4xx.blade.php
```diff
@@ -0,0 +1,11 @@
+@extends('errors::minimal')
+
+@php
+    $statusCode = $exception->getStatusCode();
```
GrahamCampbell, line 4: How do we know this method always exists on the exception?

src/Illuminate/Foundation/Exceptions/views/5xx.blade.php
```diff
@@ -0,0 +1,11 @@
+@extends('errors::minimal')
+
+@php
+    $statusCode = $exception->getStatusCode();
+    $statusText = __(\Symfony\Component\HttpFoundation\Response::$statusTexts[$statusCode]);
+    $statusMessage = !empty($exception->getMessage()) ? ': ' . __($exception->getMessage()) : '';
```
GrahamCampbell, line 6: For a 5xx, the user has absolutely know business seeing the exception message, that's for sure.

GrahamCampbell (changes requested):

src/Illuminate/Foundation/Exceptions/views/4xx.blade.php
WEBPerformace: We know this because it's a 4xx error page, which only renders when a status code exists in the 400-499 range.

src/Illuminate/Foundation/Exceptions/views/4xx.blade.php
WEBPerformace: But that's exactly what [currently happens](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Foundation/Exceptions/views/403.blade.php#L5) in `403.blade.php` of the Framework.

So, shouldn't we do something about it?

If we agree that we shouldn't show it, I see two options:

1. Show it only if `APP_DEBUG=true`
2. Just remove it from the templates.

I prefer the first option in this case. What are your thoughts?

Thank you for your comments.

taylorotwell:

taylorotwell:
You can already implement this yourself via the `$this->renderable` method in your exception handler.

WEBPerformace:
> Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include.
@taylorotwell I'm not sure about this statement because this PR actually reduces the amount of code (replacing 8 files with just 2) while also providing more flexibility and _eliminating the need for duplicate files for end-users_.

Would you please review my pull request when I fix all the issues mentioned in this discussion?

GrahamCampbell:
I don't feel this change needs to be made to the framework. The duplication is not that bad, and it's already super easy to customize. Lots of people have really custom stuff, and there are loads of packages out there. We don't need to keep changing the framework core.

olivierguerriat:
I respect that you don't want to implement these changes.

However, if the situation is gonna stay as-is, I find the [documentation about fallback errors](https://laravel.com/docs/10.x/errors#fallback-http-error-pages) misleading:

> You may also define a "fallback" error page for a given series of HTTP status codes. This page will be rendered if there is not a corresponding page for the specific HTTP status code that occurred. To accomplish this, define a 4xx.blade.php template and a 5xx.blade.php template in your application's resources/views/errors directory.

I (non native speaker) understood "if there is not a corresponding page for the specific HTTP status code" as in "in my repository", not as in "in Laravel or my repository" and ended up here debugging why my two `4xx` and `5xx` templates weren't used for 404 and 500.

I therefore think the documentation would benefit from an update, explicitly stating the exceptions to that fallback rule.

PR #46878, opened 2023-04-25
[10.x] Str::limit() supports emojis.

Nationalcat:
`Str::limit()` cannot correctly count the number of characters because fullwidth characters are counted as two characters.

```php
// before: 臺灣的珍珠...
// after: 臺灣的珍珠奶茶很好喝🧋...
Str::limit("臺灣的珍珠奶茶很好喝🧋！Taiwan's pearl milk tea is delicious. 🧋", 11);
```


taylorotwell:
Worried this could be a breaking change for some applications. Feel free to define or macro in your own Str::limit function.

PR #46881, opened 2023-04-25
[10.x] Add default method for `withValidator`

tamiroh:
This pull request adds the `withValidator` default method to `FormRequest`. This is mainly for IDE completion and static analysis.

Currently, `withValidator` does not exist by default in `FormRequest`, and developers must refer to the documentation carefully to ensure that it is implemented correctly in their `FormRequest`. This can be a cause of mistakes and also the development experience is not good because completion does not work. In addition, since static analysis does not work well, it is difficult to notice mistakes until it actually runs. This is concerning from a safety viewpoint.

Therefore, this pull request expects to improve the development experience and safety by implementing the default method for `withValidator` in `FormRequest`.

### DIFF
--- a/src/Illuminate/Foundation/Http/FormRequest.php
+++ b/src/Illuminate/Foundation/Http/FormRequest.php
@@ -7,7 +7,7 @@ use Illuminate\Auth\Access\Response;
 use Illuminate\Contracts\Container\Container;
 use Illuminate\Contracts\Validation\Factory as ValidationFactory;
 use Illuminate\Contracts\Validation\ValidatesWhenResolved;
-use Illuminate\Contracts\Validation\Validator;
+use Illuminate\Contracts\Validation\Validator as ValidatorContract;
 use Illuminate\Http\Request;
 use Illuminate\Routing\Redirector;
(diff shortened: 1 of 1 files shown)


rojtjo:
Shouldn't the `withValidator` method also just accept the contract since that's what's returned from the `createDefaultValidator` method?
When overriding the method in your own form request you could always narrow it down to use the implementation and it would still be compatible.

See: https://github.com/laravel/framework/blob/10.x/src/Illuminate/Foundation/Http/FormRequest.php#L111-L117

Angel5a:
@rojtjo Methods should be contra-variant on params. If you declare method that accept Contract in a base class, then you can't override it to accept implementation in a child class.

And, yes, withValidator have to accept Validator contract, not a Validator implementation.

taylorotwell:

tamiroh:
I just noticed that the `after` method is now available as a way to add an after hook. In the latest official documentation, the description of `withValidator` has been removed and a explanation of `after` replaced it.

https://laravel.com/docs/10.x/validation#performing-additional-validation-on-form-requests

It would be better to use `after` instead of `withValidator` from now on!

PR #46882, opened 2023-04-25
[10.x] Add `Str::flat` method

nickbeen:
# Description

This PR adds a `Str::flat()` helper method used to convert values to "flatcase" naming convention. 

```php
// laravelframework
Str::flat('Laravel Framework');

// laravelframework
Str::flat('LARAVEL-FRAMEWORK');
```

# Purpose

Though flatcase naming conventions are more used in other programming languages such as Java, it can be useful for e.g. making hashtags or creating old-style URLs without hyphens.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -364,7 +364,7 @@ class Str
      */
     public static function flat($value)
     {
-        return Str::slug($value, '');
+        return static::slug($value, '');
     }
 
     /**
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -364,9 +364,7 @@ class Str
      */
     public static function flat($value)
     {
-        $string = static::replace(['-', '_', ' '], '', $value);
-
-        return static::lower($string);
+        return Str::slug($value, '');
     }
 
     /**
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -357,6 +357,19 @@ class Str
     }
 
     /**
+     * Convert a value to flat case.
+     *
+     * @param  string  $value
+     * @return string
+     */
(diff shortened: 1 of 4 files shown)


src/Illuminate/Support/Str.php
```diff
@@ -356,6 +356,19 @@ public static function finish($value, $cap)
...
 
+    /**
+     * Convert a value to flat case.
+     *
+     * @param  string  $value
+     * @return string
+     */
+    public static function flat($value)
+    {
+        $string = static::replace(['-', '_', ' '], '', $value);
```
Angel5a, line 367: What would be with a "Hi, there!"? I don't think that "hi,there!" is a valid flatcase. You'd better to use snake/slug/smth. under the hood.

johanrosenson: I think using the `slug` method with a custom `$separator` would implement flatcase correctly.

```
return Str::slug('Hi, there!', ''); // 'hithere'
```

flavio-paqt:
There is a Laravel helper: https://laravel.com/docs/10.x/helpers#method-str-lower

Besides that you can also use the `slug` helper afterwards. It doesn't add much value in my opinion. If possible, you could consider releasing your code as a package so that the community can still take advantage of your contributions.

src/Illuminate/Support/Str.php
```diff
@@ -356,6 +356,17 @@ public static function finish($value, $cap)
...
 
+    /**
+     * Convert a value to flat case.
+     *
+     * @param  string  $value
+     * @return string
+     */
+    public static function flat($value)
+    {
+        return Str::slug($value, '');
```
johanrosenson, line 367: you should probably write it as `static::slug`, that would keep it more in line with how `kebab()` works.

taylorotwell:

PR #46512, opened 2023-03-18
[10.x] Support check constraints

hafezdivandari:
Laravel DB migrations supports almost all column types, indexes and constraints, except `CHECK` constraint. This PR adds support for check constraint on tables and columns without a breaking change.

| Database | Supported version on Laravel | Check constraint support | Drop constraint support |
| --- | --- | --- | --- |
| MariaDB | 10.3+ | ✓ ([Docs](https://mariadb.com/kb/en/alter-table/#add-constraint)) |  ✓ ([Docs](https://mariadb.com/kb/en/alter-table/#drop-constraint)) |
| MySQL | 5.7+ | 8.0.16+ ([Docs](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html#alter-table-foreign-key)) |  8.0.19+ ([Docs](https://dev.mysql.com/doc/refman/8.0/en/alter-table.html#alter-table-foreign-key)) |
| PostgreSQL | 10.0+ | ✓ ([Docs](https://www.postgresql.org/docs/10/sql-altertable.html)) |  ✓ ([Docs](https://www.postgresql.org/docs/10/sql-altertable.html)) |
| SQLite | 3.8.8+ | ✓ ([Docs](https://www.sqlite.org/lang_createtable.html#ckconst)) | ✗ ([Docs](https://www.sqlite.org/omitted.html)) |
SQL Server | 2017+ | ✓ ([Docs](https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-2017#c-adding-an-unverified-check-constraint-to-an-existing-column)) |  ✓ ([Docs](https://learn.microsoft.com/en-us/sql/t-sql/statements/alter-table-transact-sql?view=sql-server-2017#b-dropping-constraints-and-columns)) |

Note: MySQL 5.7 premier support has ended on Oct 2020 and its extended supports ends on Oct 2023.

#### Examples:

- Add check constraints when creating table:
   ```php
   Schema::create('users', function (Blueprint $table) {
       $table->integer('c1')->check('c1 > 10'); // `c1` int not null check (c1 > 10)
       $table->integer('c2');                   // `c2` int not null
       $table->check('c1 + c2 < 100');          // check (c1 + c2 < 100)
       $table->check('c2 <> 0', 'c2_nonzero');  // constraint `c2_nonzero` check (c2 <> 0)
   });
   ```

- Add check constraints when updating table (SQLite doesn't support this):
   ```php
[...]

PR #46883, opened 2023-04-25
[10.x] Allow adding column constraints in migrations (incomplete)

jameshulse:
This is a request for feedback on whether this is an idea that I should keep building out. The PR is a proof of concept currently, but is functional.

## Overview

The basic premise is to have an easy way to add `CHECK` constraints to database tables. These constraints are a useful way to ensure that the data that is persisted to your database is always in a good state. They are supported in [Postgres](https://www.postgresql.org/docs/current/ddl-constraints.html), [MySQL](https://dev.mysql.com/doc/refman/8.0/en/create-table-check-constraints.html) and [MS SQL](https://learn.microsoft.com/en-us/sql/relational-databases/tables/create-check-constraints?view=sql-server-ver16).

Potential use-cases are:
- Ensuring a `month` or `day_of_week` column are valid values in a standard range instead of any integer.
- Validating the format of a string such as a Zip code, Phone number, or tax number with regex
- That a `discount_amount` isn't larger than the `total_amount` for an order or invoice

Although these should be tackled at the application level with validation, for strong database integrity it can be useful to add these checks at a database level too. Encouraging developers to use this feature _should_ improve the quality of some code bases.

## Proposed change

Introduce `->check(...)` functions to the `Blueprint` and `ColumnDefinition` classes to allow for this kind of syntax in migrations:

```
function up() {
    Schema::table('company', function (Blueprint $table) {
        ...
       $table->check('check_day_of_week', 'day_of_week >= 1 and day_of_week <= 7');
    });
}


function down() {
    Schema::table('company', function (Blueprint $table) {
        $table->dropCheck('check_day_of_week');
    });
}
```

There are a few different syntaxes for adding checks, but we wouldn't necessarily have to attempt to support them all. Although adding a check constraint directly at a column level could be useful:

```
$table->smallInteger('day_of_week')->check('day_of_week >= 1 and day_of_week <= 7');
```

## Caveats

We don't have an easy way to automatically name the check constraint as we don't have a list of columns that we are applying the constraint to for example. Because of this I have added a mandatory `$checkName` parameter to the function. This isn't how many of the other migration methods work so may be clunky unless we can come up with another solution.

### DIFF
--- a/src/Illuminate/Database/Schema/Blueprint.php
+++ b/src/Illuminate/Database/Schema/Blueprint.php
@@ -1674,6 +1674,23 @@ class Blueprint
     }
 
     /**
+     * Add a check constraint to the table.
+     *
+     * @param $checkName
+     * @param $constraint
+     * @return Fluent
+     */
+    public function check($checkName, $constraint)
+    {
+        return $this->addCommand('check', compact('checkName', 'constraint'));
+    }
+
+    public function dropCheck($checkName)
+    {
+        return $this->addCommand('dropCheck', compact('checkName'));
+    }
+
+    /**
      * Add a new column to the blueprint.
      *
      * @param  string  $type
--- a/src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/PostgresGrammar.php
@@ -579,6 +579,34 @@ class PostgresGrammar extends Grammar
     }
 
     /**
+     * Compile a table level check command.
+     *
+     * @param Blueprint $blueprint
+     * @param Fluent $command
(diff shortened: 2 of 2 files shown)


morloderex:
@jameshulse this was already tried recently in https://github.com/laravel/framework/pull/46512. 

I am not sure why this other pr was rejected tho just pointing it out.

jameshulse:
@morloderex, good catch! I hadn't seen that PR. I guess I should probably close this then as the previous PR looks like it was more complete than what I have here.

Having a package to handle this with macro's would probably make enough sense that it doesn't need to be added to the core framework.

hafezdivandari:
#56929

PR #46884, opened 2023-04-26
update: add string cast

woenel:
This PR adds `(string)` cast to the value of `mb_strtolower` to fix the following deprecation error from PHP when null is passed:

```
mb_strtolower(): Passing null to parameter #1 ($string) of type string is deprecated in vendor/laravel/framework/src/Illuminate/Support/Str.php on line 508
```

So existing logic that use `Str::lower(config('example.can_be_null')) === 'not_null'` still works and don't throw deprecated error.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -519,7 +519,7 @@ class Str
      */
     public static function lower($value)
     {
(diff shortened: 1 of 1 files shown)


driesvints:
The docblock already indicates you can only pass a string here. You should typecast it before sending it to the method.

PR #46886, opened 2023-04-26
[10.x] Extract common code to `getResourceRegistrar()` method

xurshudyan:
This PR extracts the common code between the singleton() and resource() methods to a new method called getResourceRegistrar(). This avoids code duplication. Both singleton() and resource() now call the getResourceRegistrar() method to retrieve an instance of the ResourceRegistrar class for the current container.


milwad-dev:
I think is an additional method for framework!
If we are going to create method for duplicate code, after we have a lot of methods and it is awful for framework

PR #46887, opened 2023-04-26
[10.x]  Using a named argument instead of a local variable

xurshudyan:
Refactor function to use named argument.


GrahamCampbell:
We can't use named arguments because Laravel does not maintain a BC promise on those.

PR #46888, opened 2023-04-26
[10.x] Store instantiated middleware within Kernel

cosmastech:
## Problem
Right now, all of the Http\Kernel@sendRequestThroughRouter() instantiates all of the middleware to pipe the request through. Then subsequently, Kernel@terminateMiddleware() instantiates all the middleware again just to check if there's a terminate() method on the middleware.

### But is it really a problem?
For a lot of users, this will probably have a pretty minimal impact.

But if you have numerous middleware that require the container to resolve and inject dependencies, every request is made slower by this duplication of instantiation.

## Solution
Store middleware instances within the Kernel.

## Room for improvement
Router@runRouteWithinStack() could push check for terminating middleware and push that to the Kernel's $terminatingMiddlewareInstances array (though there's a dependency conflict).

It might also be nice to have a TerminatingMiddleware interface or a separate middleware stack that's just for terminating middleware.

To me, the co-mingling of the middlewares is very messy.

### DIFF
--- a/src/Illuminate/Foundation/Http/Kernel.php
+++ b/src/Illuminate/Foundation/Http/Kernel.php
@@ -116,7 +116,7 @@ class Kernel implements KernelContract
     /**
      * Stores instances of middleware to be used when terminating the request.
      *
-     * @var array<int, class-string>
+     * @var array<int, callable>
      */
     protected $terminatingMiddlewareInstances = [];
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Potentially a breaking change if dependency bindings are changed during request.

cosmastech:
@taylorotwell - what if I re-opened this and instead of storing instances, stored the class-name string for any middleware that has a `terminate()` method?

Then it can be reinstantiated to refresh bindings without having to instantiate middleware that doesn't have a `terminate()` method and was instantiated previously?

PR #46889, opened 2023-04-26
[10.x] Add missing dependency to `composer.json` in illuminate/support

tamiroh:
`Traits/Localizable` trait in illuminate/support references `Illuminate\Container\Container`, but the dependency on illuminate/container is not listed in `composer.json`. This pull request adds it.


taylorotwell:
We don't add things just for a single reference typically. It could be in suggests, etc.

PR #46890, opened 2023-04-26
[10.x] Add `tags` in flush method in container

milwad-dev:
When flush the binding, the tags are still and not flushed!

Benchmark:
```
0.054ms => Flush without tags

0.041ms => Flush with tags
```


taylorotwell:
This could be a breaking change?

milwad-dev:
@taylorotwell  @driesvints
If break changes, we can add param for `flush` method to flush `tags`

PR #46891, opened 2023-04-26
[9.x] Fix evaluation of factories returned from definition closure

ctalkington:
this was originally fixed in #42344 but in #44107 this was changed to support recycling which removed the `getKey` call so if you had a closure returning a factory in the definition in your factory it would cause an sql error with trying to insert the string representation of the resolved model (in second map of this function) into the database because the factory ends up resolved to model instead of the key. This appears to only happen when using a recycled model via `recycle` in my testing which is when `getRandomRecycledModel` returns model.

note: #44328 changed the method to support multiple models but remained without `getKey` call.

```php
class PostFactory extends Factory
{
    public function definition()
    {
        return [
            'title' => 'post',
            'user_id' => function (array $attributes) {
                return User::factory();
            },
        ];
    }
}
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Factories/Factory.php
+++ b/src/Illuminate/Database/Eloquent/Factories/Factory.php
@@ -471,7 +471,7 @@ abstract class Factory
         return collect($definition)
             ->map($evaluateRelations = function ($attribute) {
                 if ($attribute instanceof self) {
-                    $attribute = $this->getRandomRecycledModel($attribute->modelName())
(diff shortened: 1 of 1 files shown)


nunomaduro:
Could you please add tests to ensure that this new code does not cause any issues with existing applications and functions as intended?

nunomaduro:
Hi @ctalkington, once you have this code/pull request adjusted with tests, please create a new pull request. Thank you!

PR #46892, opened 2023-04-26
[10.x] Store terminating Middleware class-strings to avoid unnecessary middleware instantiation in `Kernel@terminate()`

cosmastech:
https://github.com/laravel/framework/pull/46888 tweaked this slightly to allow the container to make the middleware again where needed in case bindings have changed during the request lifecycle.


taylorotwell:
If you want the container to only resolve some middleware once, just bind them as singletons in the container.

PR #46893, opened 2023-04-26
Fix issue with `e()` function not accepting arrays as arguments

datlechin:
Previously, the `e()` function in Laravel was throwing an error when an array was passed as its first argument. This was due to the fact that the function only accepts strings as its input. To fix this issue, the code was modified to check if the input is an array, and if so, it was encoded as a JSON string using `json_encode()`.

Almost all sites built with Laravel get this error. For example laracasts.com, forge.laravel.com, filamentphp.com

There are the error screenshots when trying to pass array to parameters:

image
image
image

Trying to pass an array to email input
Screenshot from 2023-04-27 01-11-47

oops
Screenshot from 2023-04-27 01-11-56

This ensured that the `e()` function could safely handle arrays and strings.

before implement:
Screenshot from 2023-04-27 00-48-23

after:
Screenshot from 2023-04-27 00-48-33

p/s: if the code of PR is not satisfactory, you (the core team) can fix this error yourself.

### DIFF
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -102,7 +102,7 @@ if (! function_exists('e')) {
     /**
      * Encode HTML special characters in a string.
      *
-     * @param  \Illuminate\Contracts\Support\DeferringDisplayableValue|\Illuminate\Contracts\Support\Htmlable|\BackedEnum|string|null  $value
+     * @param  \Illuminate\Contracts\Support\DeferringDisplayableValue|\Illuminate\Contracts\Support\Htmlable|\BackedEnum|array|string|null  $value
      * @param  bool  $doubleEncode
      * @return string
      */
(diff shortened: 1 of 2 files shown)


rodrigopedra:
I don't think manually changing a parameter a site does not expect, and then expect it to work properly would be a valid bug report.

I agree a site owner could, if they anticipate someone manually trying out unwanted input parameters, handle with a redirect, but that is beyond, such as Google does

https://www.google.com/search?q[]=test

(in Google's case I assume they don't recognize `q[]` a parameter at all, and redirect due to a missing `q` parameter, but still)

IMHO, this is something a developer afraid of this use case can already handle within its app's exception handler.

https://laravel.com/docs/10.x/errors#the-exception-handler

valorin:
There are a lot of reasons completely unrelated to `e()` that would cause a `500` to be thrown when you pass in an array instead of a string parameter. Especially search queries!

It's not the responsibility of the framework to handle unexpected input. That is 100% on the developer.

I can also see this causing security issues - blindly returning an entire array or object rather than failing because the dev forgot to specify which value from a model they wanted.

Consider what `e($user)` could return...

ankurk91:
Json_encode is not safe, use JS::from instead

PR #46895, opened 2023-04-27
Automatically Exclude Soft-Deleted Records in Joins

PouriaSeyfi:
This pull request adds a new joinSoft method to the QueryBuilder in Laravel. The joinSoft method is similar to the existing join method but excludes soft-deleted records. This feature can help to prevent soft-deleted records from being joined.

The joinSoft method accepts the same parameters as the join method, with the addition of a $softDeleteField parameter. The default value for $softDeleteField is deleted_at, but developers can change it to any field that is used for soft-deletes.

I have also added unit tests for the joinSoft method to ensure that it functions correctly and does not introduce any regressions.

### DIFF
--- a/src/Illuminate/Database/Query/Builder.php
+++ b/src/Illuminate/Database/Query/Builder.php
@@ -13,6 +13,7 @@ use Illuminate\Database\Concerns\BuildsQueries;
 use Illuminate\Database\Concerns\ExplainsQueries;
 use Illuminate\Database\ConnectionInterface;
 use Illuminate\Database\Eloquent\Builder as EloquentBuilder;
+use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Query\Grammars\Grammar;
 use Illuminate\Database\Query\Processors\Processor;
@@ -508,7 +509,7 @@ class Builder implements BuilderContract
     /**
      * Add a join clause to the query.
      *
-     * @param  \Illuminate\Contracts\Database\Query\Expression|string  $table
+     * @param  \Illuminate\Contracts\Database\Query\Expression|string|Model  $table
      * @param  \Closure|string  $first
      * @param  string|null  $operator
      * @param  string|null  $second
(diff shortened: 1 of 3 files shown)


ankurk91:
`joinSoft` is not confusing name, it should like `joinWithoutTrashed`

PouriaSeyfi:
I have updated my pull request to use the existing join method instead of creating a new joinSafe method. I have also added a check to see if the provided table parameter is a valid model class. If it is, I retrieve the table name and soft delete field from the model. This way, the method can automatically exclude soft-deleted records if the provided table is a model that uses soft deletes.

This approach reduces code duplication and adheres to the principle of keeping things simple. It also allows developers to use the existing join method with the added benefit of automatically excluding soft-deleted records if the joined table is a model that uses soft deletes.

taylorotwell:

PouriaSeyfi:
Thank you for your review @taylorotwell 
I understand your concern about maintaining the framework, but I believe this change will be valuable for Laravel users who work with soft-deleted models. 
By checking if the table parameter is a class name and automatically excluding soft-deleted records, this change will help developers write more efficient and maintainable code. In addition, this is a relatively small and straightforward change that will not significantly increase the complexity of the framework.
I would be happy to address any further concerns you may have and work together to find a solution that benefits the Laravel community.

dennisprudlo:
This may be undesirable. When moving data because of a database migration I always use the raw query builder instead of the eloquent builder. This allows me to get rid of any pre-filtering or transforming rows such as global scopes and casts.

PouriaSeyfi:
> This may be undesirable. When moving data because of a database migration I always use the raw query builder instead of the eloquent builder. This allows me to get rid of any pre-filtering or transforming rows such as global scopes and casts.

The changes I made only apply to string table names and class names that exist, so they won't affect any raw queries.

bert-w:
@PouriaSeyfi `Query\Builder` != `Eloquent\Builder`. The query builder shouldnt have any notion about an Eloquent model, which is why your PR is not desirable. I hope that clear some stuff up :)

PouriaSeyfi:
> @PouriaSeyfi `Query\Builder` != `Eloquent\Builder`. The query builder shouldnt have any notion about an Eloquent model, which is why your PR is not desirable. I hope that clear some stuff up :)

thank you for explanation @bert-w 

I think a developer can refer to a database table either by its name or by using the corresponding model class name. If the model class name is used in the query builder (wherever), we assume that the developer wants to include model behaviors such as soft delete in the query.

Using the model class name in the query builder is an uncommon practice, but it can be useful when joining tables using a join model, or when developers need to take advantage of model behaviors such as soft delete. However, if these behaviors are not needed, it is generally better to use the table name directly in the queries for simplicity and clarity.

PouriaSeyfi:
When you use Eloquent, soft delete is enabled by default for models, but it is not automatically applied to joined tables. This means that when you join tables in a query, you need to manually include a condition that excludes soft deleted records from the join (and it is likely to be forgotten)

Using the Eloquent model class name in the join statement can help you avoid this mistake because it automatically includes the soft delete behavior for the joined tables. This ensures that only non-deleted records are included in the join, without you needing to manually add a condition to exclude the soft deleted records.

@taylorotwell

bert-w:
@PouriaSeyfi because when you call `$query->join()` you cross the bridge to the QueryBuilder, also when you are in an EloquentBuilder instance. If you want to have SoftDeletes functionality, you should be using relations (which require Models) instead of joins.

PouriaSeyfi:
@bert-w while it is true that using relations with models is a better way to utilize SoftDeletes functionality, there are situations where using joins in QueryBuilder may be necessary or more efficient. For example, when working with a large amount of data, using a join can result in a faster query execution time compared to using a relation. Additionally, there may be cases where we need to join tables that are not related through a model.

PR #46898, opened 2023-04-27
[10.x] Fix error message for `missing_unless` rule

apih:
This PR is intended to fix the error message for `missing_unless` rule. Based on the code snippet below, the error message produced for `missing_rule` was wrong, IMO.

**Code**
```php
$data = [
    'item_1' => '',
    'item_2' => '',
];

$rules = [
    'item_2' => ['missing_unless:item_1,foo,bar'],
];

$v = Validator::make($data, $rules);
$v->passes();

dump($v->errors()->messages());
```

**Output**
```
array:1 [▼ // app/Http/Controllers/TestController.php:27
  "item_2" => array:1 [▼
    0 => "The item 2 field must be missing unless item 1 is ."
  ]
]
```

Based on `required_unless` and `prohibited_unless` rules, the output should be quite similar, which is to list the values needed to satisfy the rule. So the output of error message for `missing_unless` rule should be: `The item 2 field must be missing unless item 1 is in foo, bar.`.

### DIFF
--- a/src/Illuminate/Translation/lang/en/validation.php
+++ b/src/Illuminate/Translation/lang/en/validation.php
@@ -105,7 +105,7 @@ return [
     'min_digits' => 'The :attribute field must have at least :min digits.',
     'missing' => 'The :attribute field must be missing.',
     'missing_if' => 'The :attribute field must be missing when :other is :value.',
-    'missing_unless' => 'The :attribute field must be missing unless :other is :value.',
+    'missing_unless' => 'The :attribute field must be missing unless :other is in :values.',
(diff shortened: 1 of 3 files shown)


taylorotwell:
Fixed - thanks.

apih:
@taylorotwell The current fix didn't take into the account that `missing_unless` rule will check all provided values using `in_array`, not just one value, as shown in the code for [`validateMissingUnless()`](https://github.com/laravel/framework/blob/89ac58aa9b4fb7ef9f3b2290921488da1454ed30/src/Illuminate/Validation/Concerns/ValidatesAttributes.php#L1575).

Issue #46756, opened 2023-04-12
Queued event listener options aren't taking any affect


sts-ryan-holton:
### Laravel Version

10.6.2

### PHP Version

8.1.17

### Database Driver & Version

10.5.16-MariaDB - MariaDB Server

### Description

I'm escalating this to the bugs section in Laravel after exhausting Google searches and debugging myself.

I have a queued event listener called `StoreUptimeHourlySummaryStat` which creates a new entry in my database or gets it and increments some columns. However, despite setting `tries`, `timeout`, `maxExceptions` and a `failed` method, none of these are taking any affect in the application.

For example, here's my listener file:

```php
<?php

namespace App\Listeners;

use App\Events\UptimeHourlySummaryStatCreation;
use Illuminate\Contracts\Queue\ShouldQueue;
use Illuminate\Queue\InteractsWithQueue;
use Illuminate\Support\Facades\Log;
use App\Models\UptimeHourlySummaryStat;
use Carbon\Carbon;
use Throwable;

class StoreUptimeHourlySummaryStat implements ShouldQueue
{
    use InteractsWithQueue;

    /**
     * The name of the connection the job should be sent to.
     *
     * @var string|null
     */
    public $connection = 'database-short-running';

    /**
     * The name of the queue the job should be sent to.
     *
     * @var string|null
     */
    public $queue = 'statistics-hourly';

    /**
     * The number of times the queued listener may be attempted.
     *
     * @var int
     */
    public $tries = 1;

    /**
     * The number of seconds the job can run before timing out.
     *
     * @var int
     */
    public $timeout = 50;

    /**
     * The maximum number of unhandled exceptions to allow before failing.
     *
     * @var int
     */
    public $maxExceptions = 1;

    /**
     * Create the event listener.
     */
    public function __construct()
    {
        //
    }

    /**
     * Handle the event.
     */
    public function handle(UptimeHourlySummaryStatCreation $event): void
    {
        try {
            $statistic = UptimeHourlySummaryStat::firstOrCreate([
                'user_id' => $event->userId,
                'monitor_id' => $event->monitorId,
                'grouped_at' => Carbon::parse($event->lastCheckedAt)->format('Y-m-d H:00:00')
            ]);

            // set count of each check
            $statistic->total_checks_performed = $statistic->total_checks_performed + $event->totalChecksPerformed;
            $statistic->total_successful_checks_performed = $statistic->total_successful_checks_performed + $event->totalSuccessfulChecksPerformed;
[...]

PR #46899, opened 2023-04-27
[10.x] Set CallQueuedHandler queue-related properties from queued EventListener

cosmastech:
CallQueuedHandler does not respect a queued Event Listener's properties, they just default to whatever is set on CallQueuedHandler.

This PR will carry over those properties.

This resolves at least one of the issues identified in https://github.com/laravel/framework/issues/46756

I'll add test cases if there is any reason to.

### DIFF
--- a/src/Illuminate/Events/CallQueuedListener.php
+++ b/src/Illuminate/Events/CallQueuedListener.php
@@ -131,24 +131,10 @@ class CallQueuedListener implements ShouldQueue
      */
     protected function setQueueVariables($handler)
     {
-        if (isset($handler->tries)) {
-            $this->tries = $handler->tries;
-        }
-
-        if (isset($handler->maxExceptions)) {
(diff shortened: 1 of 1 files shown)


src/Illuminate/Events/CallQueuedListener.php
```diff
@@ -123,12 +124,43 @@ protected function setJobInstanceIfNecessary(Job $job, $instance)
...
+        return $handler;
+    }
+
     /**
      * Call the failed method on the job instance.
      *
      * The event instance and the exception will be passed.
      *
-     * @param  \Throwable  $e
+     * @param  Throwable  $e
```
driesvints, line 163: Please revert this change.

taylorotwell:
We already do this here: https://github.com/laravel/framework/blob/10.x/src/Illuminate/Events/Dispatcher.php#L618

???

PR #46904, opened 2023-04-28
[10.x] Siesta

timacdonald:
This PR proposes a new utility class that wraps up the native `sleep` and `usleep` PHP functions into a "fakeable" and testable class.

The motivation behind this PR is to improve testing of sleeping and while also removing the duration of the sleep from the test itself while maintaining confidence in the test.

Often when an application interacts with a 3rd party library it will need to retry interactions after pausing execution.

An example of this may be a HTTP endpoint. The API may be down for a brief moment. Instead of failing, the application may use a "retry":

```php
Http::retry(3, 500)->post(/* ... */);
```

When attempting to test the failure / retry logic, a test would take a _minimum_ of 1.5 seconds.

More generally, when using the `retry` helper, this is also a (minor) annoyance.

```php
return retry(5, function () {
    // Attempt 5 times while resting 100ms between attempts...
}, 100);
```

To address this, we now use the `Siesta` helper in a few places throughout the Framework. The helper may the be "faked" in tests and some basic assertions may also be made.

### Fluent API

```php
// minutes...

Siesta::for(3)->minutes();

Siesta::for(1)->minute();

// seconds...

Siesta::for(3)->seconds();

Siesta::for(1)->second();

// milliseconds...

Siesta::for(30000)->milliseconds();

Siesta::for(1)->millisecond();

// microseconds...

Siesta::for(30000)->microseconds();

Siesta::for(1)->microsecond();
```

### Composed durations

```php
Siesta::for(3)->minutes()
      ->and(5)->seconds()
      ->and(9)->milliseconds();
```

### Siesta until a given time

```php
Siesta::until(now()->addMinute());
```

### 1:1 replacement API

```php
// usleep(30000);

Siesta::usleep(30000);

// sleep(2);

Siesta::sleep(2);

// time_sleep_until(1682641623)

Siesta::until(1682641623);
```

## Testing

As mentioned, the motivation here is to improve testing. If an application simply wants tests to not "sleep", then the following could be added to any test that uses `Siesta`:

```php
{
    parent::setUp();

    Siesta::fake();
}
```

It is also possible to make assertions:

```php
public function testItSendsWebhookSuccessfluly()
{
    Siesta::fake();
    Http::fake(['https://laravel.com' => Http::response('', 200)]);

    Webhook::send('https://laravel.com');

    Siesta::assertInsomniac(); // too much?! 😅
}

public function testItRetriesWithBackoffForFailedWebhooks()
{
    $attempts = 0;
    Siesta::fake();
    Http::fake(['https://laravel.com' => function () use (&$attempts) {
        $attempts++;

        return $attempts++ === 3
            ? Http::response('', 200)
            : Http::response('', 500);
    }]);


    Webhook::send('https://laravel.com');

    $this->assertSame(3, $attempts);
    Siesta::assertSleptTimes(3);
    Siesta::assertSequence([
        Siesta::for(100)->milliseconds(),
        Siesta::for(200)->milliseconds(),
        Siesta::for(300)->milliseconds(),
    ]);
}

// Full API

Siesta::assertSlept(fn (CarbonInterval $duration): bool => /* ... */);

Siesta::assertSlept(fn (CarbonInterval $duration): bool => /* ... */, $times = 5);

// asserts the number of times Siesta was used...
Siesta::assertSleptTimes(5);

// asserts that even if Siesta was used that zero time was spent sleeping...
Siesta::assertInsomniac();
```

Now any call to sleep via Siesta will not actually pause execution throughout the testsuite.

## Naming alternatives

If the framework doesn't jive with "Siesta":

- `Sleep`
- `Pause`
- `Rest`
- Something else?

…but lowkey “Siesta” is the best option 🙉

## Use of Siesta within the Framework

I feel that the places I've used Siesta throughout the framework make sense. If we feel any of them don't make sense, I'm happy to remove any.

## 📌 Notes

The native [`time_sleep_until`](https://www.php.net/manual/en/function.time-sleep-until.php) function will throw an error if the time has already passed; we just do not sleep. When testing, you would need to assert that it slept for zero seconds.

---

The native [`sleep`](https://www.php.net/manual/en/function.sleep.php) function will cast a `float` to an `int`. Our implementation does not do this:

```php
sleep(1.5); // only sleeps for 1 second

Siesta::for(1.5)->seconds(); // sleeps for 1 second and 500 milliseconds

Siesta::sleep(1.5); // sleeps for 1 second and 500 milliseconds
```

---

The native [`usleep`](https://www.php.net/manual/en/function.usleep.php) function may not support values larger than `1000000`. Our implementation fixes this by using a combination of `sleep` and `usleep`.

---

Although PHP does have the ability to sleep for nanoseconds, via the native [time_nanosleep](https://www.php.net/manual/en/function.time-nanosleep.php) function, I didn't feel the need to add an affordance to this.

An implementation would be possible, however I'm not sure it is worth it. If you need something that fine-grained, even our PHP code in-between things is probably going to be problematic.

---

> **Warning**: Time is hard. We do execute some PHP between sleeps, so this will never be _perfect_, although computers also aren't great at time, so it is already unreliable to some degree. YOLO.


thefelipevega:
Thoughts on **_Rest_** for the name?

LasseRafn:
I think there is a typo in the last example regarding float and rounding, it says that Siesta sleeps for 90 seconds when passing 1.5? 👀 or am I misunderstanding something. But I’m loving this idea!

```
sleep(1.5); // only sleeps for 1 second

Siesta::for(1.5)->seconds(); // sleeps for 90 seconds

Siesta::sleep(1.5); // sleeps for 90 seconds
```

timacdonald:
@LasseRafn `1.5` is in base 10, so: `60 * 1.5 = 90`

timacdonald:
@LasseRafn corrected. thank you! Time is hard. Math is hard. Math + Time = 🫠

axlon:
Would personally prefer a name like Sleep, Wait or Rest. Otherwise love this!

antonkomarev:
Rest might add confusion because of the REST.

fourstacks:
Love it!  Would humbly submit `Snooze` as a name.  Does what it says on the tin and sounds _slightly_ weirder every time you say it

timacdonald:
@fourstacks I actually used Snooze when I first conceptualised this - but realised it conflicted with https://github.com/thomasjohnkane/snooze

Sammyjo20:
I absolutely love this 😆 I have a few tests for my own that last 5-10 seconds because of a sleep, would be cool to fake it!

Sammyjo20:
I also agree with @fourstacks I do like `Snooze` too 🤔

Or even being boring and having `Sleep::until()` as to not confuse devs *too* much

dsazup:
Done this a few times by just creating `sleep` method and using `$this->sleep(5)` and mocking that in tests. So really love this idea. But would prefer `Sleep::for(5)` as the name (or `Wait::for(5)`). Siesta is a great name, but a bit too fancy imho. 😄

src/Illuminate/Support/Siesta.php
```diff
@@ -0,0 +1,318 @@
+<?php
+
+namespace Illuminate\Support;
+
+use Carbon\CarbonInterval;
+use DateInterval;
+use PHPUnit\Framework\Assert as PHPUnit;
+use RuntimeException;
+
+class Siesta
```
nunomaduro, line 10: I would prefer this class to be named `Pause`, or something similar.

src/Illuminate/Support/Siesta.php
```diff
@@ -0,0 +1,318 @@
...
+     * Indicates that all sleeping should be faked.
+     *
+     * @var bool
+     */
+    protected static $fake = false;
+
+    /**
+     * The sequence of sleep durations encountered while faking.
+     *
+     * @var array
```
nunomaduro, line 36:
```suggestion
     * @var array<int, \Illuminate\Support\Siesta>
```

src/Illuminate/Support/Siesta.php
```diff
@@ -0,0 +1,318 @@
...
+     *
+     * @return $this
+     */
+    public function minute()
+    {
+        return $this->minutes();
+    }
+
+    /**
+     * Sleep for the duration in seconds.
```
nunomaduro, line 139:
```suggestion
     * Sleep for the duration in seconds.
     *
     * @return $this
```

src/Illuminate/Support/Siesta.php
```diff
@@ -0,0 +1,318 @@
...
+        $remaining = $this->duration->copy();
+
+        if ((int) $remaining->totalSeconds > 0) {
+            sleep((int) $remaining->totalSeconds);
+
+            $remaining = $remaining->subSeconds((int) $remaining->totalSeconds);
+        }
+
+        if ($remaining->totalMicroseconds > 0) {
+            usleep($remaining->totalMicroseconds);
```
nunomaduro, line 253: Is this code correct? Seems that `totalMicroseconds` may be a `float`, while `usleep` only works with `int`.

src/Illuminate/Support/Siesta.php
```diff
@@ -0,0 +1,318 @@
...
+     */
+    public static function assertInsomniac()
+    {
+        PHPUnit::assertSame(0, $count = count(static::$sequence), "Expected [0] pauses but found [{$count}].");
+    }
+
+    /**
+     * Assert sleeping occurred the given times.
+     *
+     * @param  int  $times
```
nunomaduro, line 311:
```suggestion
     * @param  int  $expected
```

nunomaduro (changes requested):

buismaarten:
I actually like `Sleep::for(1)->second()` more than `Siesta::for(1)->second()`.

src/Illuminate/Support/Siesta.php
timacdonald: I’ve mentioned naming in the description 👍

src/Illuminate/Support/Siesta.php
```diff
@@ -0,0 +1,318 @@
...
+use RuntimeException;
+
+class Siesta
+{
+    /**
+     * The total duration to sleep.
+     *
+     * @var \Carbon\CarbonInterval
+     */
+    public $duration;
```
Angel5a, line 17: I know, that it isn't supposed to use Siesta this way, but let's assume this example:
```
function ImAFunOfPrototypeDesignPattern(Siesta $pause)
{
    dump(time());
    (clone $pause)->and(3)->seconds(); // waits 5 seconds as expected.
    dump(time());
    clone $pause; // also waits for 5 seconds instead of 2.
    dump(time());
    // also waits for 5 seconds instead of 2.
    // who doesn't want delay here can write $pause->duration = CarbonInterval::seconds(0); it's ok.
    // or $pause->discard() to also be used in tests.
}

dump(time());
ImAFunOfPrototypeDesignPattern(Siesta::for(2)->seconds());
dump(time());
```
Should the `__clone()` method also be implemented?

src/Illuminate/Support/Siesta.php
```diff
@@ -0,0 +1,318 @@
...
+            throw new RuntimeException('Unknown Siesta duration unit.');
+        }
+
+        if ($this->duration->totalMicroseconds <= 0) {
+            $this->duration = CarbonInterval::seconds(0);
+        }
+
+        if (static::$fake) {
+            if ($this->capture) {
+                static::$sequence[] = $this;
```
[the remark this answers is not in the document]
timacdonald: Good call. I was originally only storing the duration and changed to the whole object at some point - but I agree. Probably not a great idea. Updated.

src/Illuminate/Support/Siesta.php
timacdonald: I don't think we need to consider this. That is not the intended use of this class.

src/Illuminate/Support/Siesta.php
timacdonald: I don't believe this is actually an issue. I'm unable to make this a float, so I can't write a test for it.

I will cast them to an int, just to be safe, but not sure it'll ever be a float tbh.

src/Illuminate/Cache/Lock.php
```diff
@@ -114,7 +115,7 @@ public function block($seconds, $callback = null)
         $starting = $this->currentTime();
 
         while (! $this->acquire()) {
-            usleep($this->sleepMilliseconds * 1000);
```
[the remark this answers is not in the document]
timacdonald: I think at a bare minimum we want it used in the `retry` helper. After that, in order of what I would want to optionally see:

- Timebox
- Cache stuff

timacdonald:
Draft as there are some final tweaks I wanna make.

bert-w:
I prefer just a regular `Sleep` as the class name, that makes the most sense to me. Also a rename for `assertInsomniac` would be fitting.

I don't think I'll ever need this class myself though since I can't think of many practical usecases where I need complex sleep behaviour like in this Siesta class. But for internal/testing usecases it might be handy.

innocenzi:
> I don't think I'll ever need this class myself though since I can't think of many practical usecases where I need complex sleep behaviour like in this Siesta class. But for internal/testing usecases it might be handy.

Definitely has potential use cases for third-party integrations and `retry`. Specifically, the ability to remove the sleep calls during testing would be great.

Sammyjo20:
@timacdonald There are quite a number of people who prefer “sleep” over “siesta”. Have you considered the name change? It’s a great name but I understand people’s feelings against it - but I feel you haven’t addressed it.

timacdonald:
@Sammyjo20 naming considerations and alternatives are part of the description. I lowkey like "Siesta". If Taylor decides this feature is suitable for the framework and wants a name change, I'm happy to change it. No need for polls.

henzeb:
I'd suggest the assertSleptTimes to be assertTimesSlept. Otherwise, love it.

timacdonald:
If we are keen for this feature, but are not keen on the `__destruct` magic, I believe I can maintain the same API without leaning on destruction.

Won't implement it until we are sure we want the feature in the framework.

src/Illuminate/Support/Siesta.php
```diff
@@ -0,0 +1,383 @@
...
+            sleep($seconds);
+
+            $remaining = $remaining->subSeconds($seconds);
+        }
+
+        $microseconds = (int) $remaining->totalMicroseconds;
+
+        if ($microseconds > 0) {
+            usleep($microseconds);
+        }
```
bert-w, line 255: This can be simplified imo. In one of my projects I have added a helper that combines `sleep()` and `usleep()` so i can do fractional sleep easily like so:
```php
function my_sleep($microseconds)
    {
        if ($microseconds <= 0) {
            return;
        }
        if ($usleep = $microseconds % 1000000) {
            usleep($usleep);
        }

        if ($sleep = ($microseconds - $usleep) / 1000000) {
            sleep($sleep);
        }
    }
```

Then you can call that method (maybe even expose it as public/static since it is very handy) and be done with it. The `CarbonInterval $duration` property in your class is probably too complex since you can just use a microseconds integer instead. We are mostly talking about times in that neighborhood anyway (maybe up to a few seconds).

NickSdot:
Copying this over from Twitter. My bad for not keeping it in the PR.🙏

**Context:**
https://github.com/laravel/framework/pull/46904#pullrequestreview-1405176445

https://twitter.com/timacdonald87/status/1653151918863302656?s=46&t=zY_3ch9pBvnnjyXjsgTbnQ

I see where you are coming from, @timacdonald  

I still believe it’s possible to use it wrong.

Actually, imho, there should/must be some enforcement of calling 'minutes()' or whatever else calls 'pullPending()' before the method 'and()' can be chained again.

Let’s say I do something ‚weird‘ like:

'''php
$siesta = Siesta::for(2)->minutes();

if($bigWhateverAmountHereOne) {
 // and(2)
}

if($bigWhateverAmountHereTwo) {
 // and(5)
}

$siesta->minutes()
'''

Theoretically, because both 'if(//)' can be true, I can mess it up. And I believe this shouldn’t be possible.

There should be a check in 'and()' that there is no pending values, and otherwise throw an error.

Disclaimer: on my phone and not able to actually playing around with it. So it’s all theoretical. 🙏

Appreciate your work, Tim! ❤️

Angel5a:
@NickSdot, I think, there is no need on exceptions. Just add every add() to accumulator should be ok. Siesta class (or Sleep if smb wish), shouldn't be responsible on watching for errors in your code.

src/Illuminate/Support/Siesta.php
```diff
@@ -0,0 +1,383 @@
...
+    {
+        static::assertSleptTimes(count($sequence));
+
+        collect($sequence)
+            ->zip(static::$sequence)
+            ->eachSpread(function (?Siesta $expected, CarbonInterval $actual) {
+                if ($expected === null) {
+                    return;
+                }
+
```
Angel5a, line 338: There have been a code to prevent adding passed $sequence into static::$sequence in __destruct.
Is it ok for now to call assertion twice:
```
Seista::fake();
Seista::for(2)->seconds();
Siesta::assertSequence([
    Siesta::for(2)->seconds(), // won't it get into static::$sequence
]);
Siesta::assertSequence([
    Siesta::for(2)->seconds(),  // so this will fail
]);
```

NickSdot:
> @NickSdot, I think, there is no need on exceptions. Just add every add() to accumulator should be ok. 
>>

@Angel5a at first I was thinking that too, as my initial note on it shows. But it allows to accumulate time without defining which unit the time has. 

>
Siesta class (or Sleep if smb wish), shouldn't be responsible on watching for errors in your code.
> 
> 

Whether or not the class should be responsible to allow or disallow me to write wrong code is a good question -- I tend to the latter, but also understand why you see it different. Tim and team will have their take on it, I guess.

taylorotwell:
Renamed to `Sleep`. Added `assertNeverSlept` as alias for `assertInsomniac` (but kept insomniac method since it's a nice Easter egg). 🥚

src/Illuminate/Support/Siesta.php
timacdonald: @Angel5a It is okay to call the assertion twice.

Any time you pass an instance through to an assertion method, we tell the internal class to not capture it.


```php
Siesta::fake();

Siesta::for(2)->seconds();

Siesta::assertSequence([
    Siesta::for(2)->seconds(),
]);
Siesta::assertSequence([
    Siesta::for(2)->seconds(),
]);

Siesta::assertSequence([
    Siesta::for(2)->seconds(),
    Siesta::for(3)->seconds(), << failure
]);
```

This test will fail where marked. The failure will indicate that you expected 2 sleeps, but it only found 1 - which is correct - as the others were for testing only.

timacdonald:
Appreciate all the feedback from everyone ❤️. I hope the feature comes in handy.

Also: On the day of release I'll be all like...

```php
use Illuminate\Support\Sleep as Siesta;
```

😎

Issue #46908, opened 2023-04-28
@json does not work when there is a quoted value with commas in it followed by a comma separated list


craigpipeandpiper:
### Laravel Version

9.50.2

### PHP Version

8.1.11

### Database Driver & Version

MySQL 8.0.32

### Description

Using the `@json` function can cause malformed PHP in the compiled blade template.

My previous issue was closed as too much emphasis seemed to be placed on the best practice of what i had distilled into a proof of concept. Attached is a more realistic example of how the @json function can be broken that is more in line with how views are actually used in the real world. I have managed to reproduce this several ways, if you need more examples please let me know.

In the example below $titles is an array set from the view, $names is also an array set by the view, "last name, first name" is a constant used to define parts of a JSON blob used for AJAX.

Any comma in the quoted first value causes the bug.

The error output from laravel is:

```
Unclosed '[' does not match ')'
```

### Steps To Reproduce

```php
<script>
    window.callProcessNamesList(@json(["last name, first name", $titles, $names]));
</script>
```

driesvints:
The `@json` directive really isn't meant to be used like that. It's best to assign the contents to a variable and then pass it to the directive.

craigpipeandpiper:
uh, this is exactly what i am doing...
Please stop closing and dismissing bugs without looking into them. This is an unhealthy obsession with keeping levels of bugs down to zero.

driesvints:
There is no bug here. If you can come up with a solution for the Blade compiler to allow this then feel free to send in a PR.

craigpipeandpiper:
I shall do this, in my own free time.

The blade compiler should never generate malformed php, full stop. Regardless of if its the "one true way" or not, it should generate correctly formed PHP or generate a proper exception describing why it is not valid.

PR #46910, opened 2023-04-28
[10.x] Fix issue with @json incorrectly compiling when argument has one or more commas

rojtjo:
I ran into the issue as described in #46908 myself recently and fixed it by parsing the arguments using `preg_split` instead of a simple `explode`.

The regex is a slightly modified version of the one from the SO answer below. It now also ensures that commas within parentheses and brackets are ignored.
https://stackoverflow.com/questions/18893390/splitting-on-comma-outside-quotes/18893443#18893443

We should probably add more cases to the `jsonArgumentsProvider` data provider before even considering merging this.

### DIFF
--- a/src/Illuminate/View/Compilers/BladeCompiler.php
+++ b/src/Illuminate/View/Compilers/BladeCompiler.php
@@ -11,6 +11,9 @@ use Illuminate\Support\Str;
 use Illuminate\Support\Traits\ReflectsClosures;
 use Illuminate\View\Component;
 use InvalidArgumentException;
+use PhpParser\Lexer;
+use PhpParser\Node\Arg;
+use PhpParser\ParserFactory;
 
 class BladeCompiler extends Compiler implements CompilerInterface
 {
@@ -158,6 +161,13 @@ class BladeCompiler extends Compiler implements CompilerInterface
      * @var bool
      */
     protected $compilesComponentTags = true;
+
+    /**
+     * The cached PHP parser instance.
+     *
+     * @var \PhpParser\Parser
+     */
+    protected $parser;
 
     /**
      * Compile the view at the given path.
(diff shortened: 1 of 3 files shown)


driesvints:
Appreciate you sending in the PR. I just want to stress being careful with this because we've had numerous occasions in the past where we tried to adjust the compiler only to break other use cases for people.

bert-w:
Your tests do not include the `@json('a', JSON_PRETTY_PRINT)` case (you can provide up to 3 arguments to `@json`) and there are more language structs than the ones that you define (like `[...['a']]` and `a->?b->?c()`).

Although your regex looks good and I cant find any obvious issues with some extra structs that i tested, it seems like the best way forward is to just use simple expressions whenever you are using `@json` (you probably want to delegate most of the logic to a controller/view-component anyway).

rojtjo:
> Your tests do not include the `@json('a', JSON_PRETTY_PRINT)` case (you can provide up to 3 arguments to `@json`) and there are more language structs than the ones that you define (like `[...['a']]` and `a->?b->?c()`).

I didn't include them since they are already tested in the test above, this test only makes sure the arguments are rendered correctly. But I totally agree with your second point, it needs a lot more cases. Also just realised that heredoc and nowdoc probably don't even work.

I guess to get this working for all possible cases we'd have to use `nikic/php-parser` but that definitely seems like overkill.

rojtjo:
It indeed fails on heredoc and nowdoc strings as expected when parsing using regex.

To get it working using the php parser was actually a lot easier than expected. The php parser is currently only available when the dev dependencies are installed. This means we would need to add it as a dependency which is probably not something we would like to do for something like this

bert-w:
Honestly it wouldnt be a bad idea to just bite the bullet and discard/deprecate this blade directive, in favor of something like:

```php
{{ json($params) }}
```

```php
// helpers.php
function json($value, $flags, $depth) {
    return new \Illuminate\Support\HtmlString(json_encode(...func_get_args()));
}
```

(`HtmlString` allows you to call it using `{{ }}` instead of `{!! !!}`)

It makes no sense to apply all this difficult parsing of arguments.

The sole reason we can't do `<?php echo json_encode($expression); ?>` in the blade compiler is because we have defaults for the json arguments; this make parsing the string as php code a pain.

valorin:
You can already do that with the `Js` helper class:

```
<script>
    var app = {{ Js::from($array) }};
</script>
```

I'd suggest adding more methods into that if there are use cases it doesn't support.

https://laravel.com/docs/10.x/blade#rendering-json

bert-w:
Thx I've added a PR that exposes one of the functions in that class for this reason. The `@json` directive has already been removed from the docs, instead the docs only showcase `Js::from()` but that one explicitly echoes javascript code `JSON.parse(...)`.

PR #46911, opened 2023-04-28
add pagination support for flowbite

Mubeen142:
This PR adds support for Flowbite pagination styling.

Flowbite is an open-source library of UI components based on the utility-first Tailwind CSS framework featuring dark mode support, a Figma design system, templates, and more.

Preview darkmode:
image

Preview lightmode:
image

Flowbite: https://flowbite.com/

To use this styling throughout the application, you can edit file `App\Providers\AppServiceProvider` and add `Paginator::useFlowbite();` to the `boot` method
```php
use Illuminate\Pagination\Paginator;
 
/**
 */
public function boot(): void
{
    Paginator::useFlowbite();
}
```

### DIFF
--- a/src/Illuminate/Pagination/AbstractPaginator.php
+++ b/src/Illuminate/Pagination/AbstractPaginator.php
@@ -603,6 +603,16 @@ abstract class AbstractPaginator implements Htmlable
     }
 
     /**
+     * Indicate that Flowbite styling should be used for generated links.
+     *
+     * @return void
+     */
+    public static function useFlowbite()
+    {
+        static::defaultView('pagination::flowbite');
+    }
+
+    /**
      * Indicate that Bootstrap 4 styling should be used for generated links.
      *
      * @return void
--- a/src/Illuminate/Pagination/resources/views/flowbite.blade.php
+++ b/src/Illuminate/Pagination/resources/views/flowbite.blade.php
@@ -1,0 +1,106 @@
+@if ($paginator->hasPages())
+    <nav role="navigation" aria-label="{{ __('Pagination Navigation') }}" class="flex items-center justify-between">
+        <div class="flex justify-between flex-1 sm:hidden">
+            @if ($paginator->onFirstPage())
+                <span class="relative inline-flex items-center px-4 py-2 text-sm font-medium text-gray-500 bg-white border border-gray-300 cursor-default leading-5 rounded-md">
+                    {!! __('pagination.previous') !!}
+                </span>
+            @else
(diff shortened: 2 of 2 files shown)


Angel5a:
What about license? Flowbite is a paid stuff.
"Simple" variant is missing?
What the reason of another tailwind paginator? Bootstrap variations uses totally different classes that's why its two of them.

Mubeen142:
> What about license? Flowbite is a paid stuff. "Simple" variant is missing? What the reason of another tailwind paginator? Bootstrap variations uses totally different classes that's why its two of them.

Flowbite has an optional paid plan but at its core it consists of open source components. Tailwind has paid plans as well but that does not mean it is "Paid stuff". The code used is available on github.
https://github.com/themesberg/flowbite/tree/main/content/components
https://flowbite.com/docs/components/pagination/

The License is MIT

It is not just another tailwind paginator. Using tailwind pagination styling whilst using Flowbite lacks important design details. For instance, the dark mode does not properly adapt to the tailwind paginator and the tailwind pagination styling does not fit that of the styling of most flowbite components

dennisprudlo:
Wouldn't it be more helpful, when the flowbit

Angel5a:
[Mubeen142](https://github.com/Mubeen142), Great. Do you want to say, that you:

1. Took a part of Flowbite library - paginator with great feature of dark mode supported, which is covered by MIT License. It's ok.
2. Made a derivative work, i.e. made blade adaptation of such a cool component. It's ok.
3. Violates the Flowbite's license by providing this PR - I can't see license file of Flowbite attached anywhere, which is a key part of the MIT license.
4. Wants to Laravel Framework itself and everyone who use it to also violate Flowbite's License.

Am I right?

> The above copyright notice and this permission notice shall be included in **all copies or substantial portions** of the Software.
[https://flowbite.com/docs/getting-started/license/](https://flowbite.com/docs/getting-started/license/)

So, as I can see, current PR can't be accepted without violation of Flowbite's License.

GrahamCampbell (changes requested):
I strongly feel that this does not belong in the framework core, but should be a package.

Mubeen142:
> I strongly feel that this does not belong in the framework core, but should be a package.

I found myself struggling when trying to use the default pagination laravel provides and I can imagine that other developers faced the same struggles. The design of the default paginator is not properly adapted and makes the default pagination styling look alien compared to the other design features.

Flowbite has a huge community and it has only been growing. This simple change would be great for other developers. It would not be too practical to make a whole new package for something so small but ultimately I understand that you may not want to include support for it and make it a separate package.

bert-w:
> I found myself struggling when trying to use the default pagination laravel provides and I can imagine that other developers faced the same struggles. The design of the default paginator is not properly adapted and makes the default pagination styling look alien compared to **the other design features**.

Which other design features? The few pagination options are the only "opinionated" html components in the framework besides some standard error pages and mail templates. The pagination components are probably something to be sent to `laravel/ui` as a whole but it will cause a lot of developers to require that package as well since pagination is so omnipresent. For code separation practices it would be a good solution though.

Issue #46909, opened 2023-04-28
whereNotNull check in relations query


mihy:
### Laravel Version

10.x

### PHP Version

8.x

### Database Driver & Version

PostgreSQL 13.8

### Description

This has been a problem for a long time. As we can find, there are multiple requests for that:

https://github.com/laravel/framework/pull/21308

https://github.com/laravel/framework/pull/36905

But always the answer is that it is not a problem. I would not agree to that.

The problem arises when you want to get the count of relationship items in a table with 500K records. 

To get a count of HasMany relationship - Laravel creates SQL:

select count(*) as aggregate from "transactions" where "transactions"."user_id" = 1 and "transactions"."id" is not null

Laravel adds

`and "transactions"."id" is not null` 

because of Illuminate\Database\Eloquent\Relations\HasOneOrMany line 87 in addConstraints() function.

And so, in my case, I have 0,5M records and for comparing, lets look at 2 queries:

A. select count(*) as aggregate from "transactions" where "transactions"."user_id" = 1 and "transactions"."id" is not null
Runs for around 10s

B. select count(*) as aggregate from "transactions" where "transactions"."user_id" = 1
Runs for around 0.5s

So the difference is 20 times.

If there is necessary to keep this **whereNotNull** as default, let it be, but please let's find a way how we can disable this query for specific cases, when it is important to optimize queries. Any ideas?

### Steps To Reproduce

for example -  see issue https://github.com/laravel/framework/pull/21308

cosmastech:
The problem is that the constraints are applied when relation is instantiated.

There could be a static method (similar to Relation::noConstraints()) that would not add this `whereNotNull()` and you could wrap your query in something like this like:

```php
$transactionsCount = HasOneOrMany::doNotApplyNullCheck(fn() => $user->transactions()->count());
```

I would rather see a maintainer say "we would accept a solution like this" rather than push a PR that gets rejected. I'll let @driesvints chime in on that.

cosmastech:
I looked into adding this as a global scope so it could be removed by the client code, but it would break the expectation that `$model->someRelationship()->withoutGlobalScopes()` is removing model related scopes only, as opposed to something that has already been added to the query for many version of Laravel.

Mostly, I don't think the juice is worth the squeeze on this one. My suspicion is that almost anything that would be submitted to rectify this issue would be marked as a breaking change or suggested to build into your own package. 🤷 

You could always write:
```php
$count = null;
[...]

PR #46915, opened 2023-04-28
[10.x] Add HasOneOrMany@withoutForeignKeyNullCheck() to specify whether to apply foreign_key is not null constraint

cosmastech:
This issue seems to be kind of recurring. Most recently brought up here: https://github.com/laravel/framework/issues/46909. I think the biggest complaint I have is that it's a bizarre addition to most queries and I imagine countless developers have spent untold numbers of hours trying to figure out why that was added to their query (and if they did something wrong to make it so).

Here's a non-breaking change solution. Let the calling code specify whether or not the FK check needs to be applied.

I think this makes way to a softer landing were the maintainers to ever wish to drop this unneeded constraint. In a future release, HasOneOrMany::$applyForeignKeyCheck could default to false.

```php
class User extends Model
{
    public function posts()
    {
        return $this->hasMany(Post::class)->withoutForeignKeyNullCheck();
    }
}
```

Might be nice if this was somehow set statically so that within AppServiceProvider's `boot()` it could be called like:

```php
HasOneOrMany::withoutForeignKeyNullCheck();
```

but then that would require there to be both a protected static variable and an instance level variable. I'll happily add that if desired.

### Notes on Tests
You'll notice a lot of the tests were changed. They fell into two categories:
1. The query statement is slightly different (the `foreign_key is not null` is applied at the end after other scopes) but the content is the same
2. Tests that mock the Builder instance expected `whereNotNull()` to be called, but instead we will now check for `beforeQuery()`

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
@@ -26,6 +26,13 @@ abstract class HasOneOrMany extends Relation
     protected $localKey;
 
     /**
+     * Whether to check if the foreign key is null.
+     *
+     * @var bool
+     */
+    protected $applyForeignKeyCheck = true;
+
+    /**
      * Create a new has one or many relationship instance.
      *
      * @param  \Illuminate\Database\Eloquent\Builder  $query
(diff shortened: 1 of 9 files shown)


cosmastech:
Closing to replace with https://github.com/laravel/framework/pull/46916

PR #37451, opened 2021-05-21
[8.x] Improve one-of-many performance

cbl:
## Background

Eager loading one-of-many relationships can become slow for big tables.

## Solution

When eager loading, the related models are loaded using `WHERE foreign_id in (1,2,3,...)`, currently this constraint is applied to the parent query in one of many relationships:

```sql
SELECT *
FROM `logins`
INNER JOIN (
    SELECT MAX(id) AS id
    FROM logins
    GROUP BY logins.user_id
) AS latest_login 
ON latest_login.id = logins.id
WHERE user_id in (1,2,3,4,5) # <---
```

This means that the subselect query gets `MAX(id)` rows for every group in the table, not only the required ones.

This can be improved by adding the constraint to the subquery:

```sql
SELECT *
FROM `logins`
INNER JOIN (
    SELECT MAX(id) AS id
    FROM logins
    WHERE user_id in (1,2,3,4,5) # <---
    GROUP BY logins.user_id
) AS latest_login 
ON latest_login.id = logins.id
```

<details>
<summary>See the `EXPLAIN ANALYZE` results for both queries for more information...</summary>

Results for `4,000` users and `100,000` logins.

Filter rows on parent query:

```sql
-> Nested loop inner join  (cost=50420.32 rows=500500) (actual time=20.806..21.036 rows=5 loops=1)
    -> Index range scan on logins using logins_user_id_index, with index condition: (logins.user_id in (1,2,3,4,5))  (cost=57.51 rows=125) (actual time=0.048..0.190 rows=125 loops=1)
    -> Index lookup on latest_login using <auto_key0> (id=logins.id)  (actual time=0.001..0.001 rows=0 loops=125)
        -> Materialize  (cost=1650.35..1650.35 rows=4004) (actual time=0.167..0.167 rows=0 loops=125)
            -> Index range scan on logins using index_for_group_by(logins_user_id_index)  (cost=1249.95 rows=4004) (actual time=0.012..17.431 rows=4000 loops=1)
```

Filter rows on subquery:

```sql
-> Nested loop inner join  (cost=60.31 rows=125) (actual time=0.121..0.130 rows=5 loops=1)
    -> Filter: (latest_login.id is not null)  (cost=0.13..16.56 rows=125) (actual time=0.113..0.115 rows=5 loops=1)
        -> Table scan on latest_login  (cost=2.50..2.50 rows=0) (actual time=0.000..0.001 rows=5 loops=1)
[...]

PR #46916, opened 2023-04-28
[10.x] Add `HasOneOrMany::shouldAddForeignKeyCheck()`

cosmastech:
https://github.com/laravel/framework/issues/46909

HasOneOrMany adds a `where foreign_key is not null` to all queries, though in most cases it's not needed.

This issue comes up a lot: it makes it hard to debug queries, it's unnecessary in most (all?) cases, and it leads to confusion from developers trying to figure out WHY it was added (if they did something wrong, where it's being added within the framework, etc).

Here is a solution that allows developers to turn it off entirely, or per relationship

```php
{
    public function boot()
    {
        HasOneOrMany::shouldAddForeignKeyCheck(false);
    }
}
```

Voila. Now instead of:

`'select * from "posts" where "posts"."user_id" = ? and "posts"."user_id" is not null'` 

you get

`'select * from "posts" where "posts"."user_id" = ?`

Don't want it turned off globally? Cool, nothing is different.

Want it turned on globally but want it turned off for some relationship? Have at it.

```php
$posts = HasOneOrMany::withoutForeignKeyCheck($user->posts()->get());
```

In a future release, this static property could be set to false by default, and then maybe in a subsequent release, removed entirely.

## final thought
I like how the DB migration traits within test uses a state class with static properties. Would be nice to see this bound to the container so it could be easily swapped within tests. But I don't see this pattern used elsewhere within the framework, so I assumed something so far out of the usual stood less of a chance of being accepted (especially for an issue that has been a long-standing).

### DIFF
--- a/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
+++ b/src/Illuminate/Database/Eloquent/Relations/HasOneOrMany.php
@@ -2,6 +2,7 @@
 
 namespace Illuminate\Database\Eloquent\Relations;
 
+use Closure;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Database\Eloquent\Model;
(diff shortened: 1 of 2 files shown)


bert-w:
Assume we have some `User` parent model. If for some reason the `parentKey` (id) is null, then the `$query->where()` will eventually call it as `$query->whereNull($this->foreignKey)` (see Query Builder `where()`), including all related models with a null foreign-key (which is wrong).

Shouldn't the solution be more something like this?

```php
if (! is_null($this->getParentKey()) {
    // Value available, apply it as the where clause:
    $query->where(...);
} else {
    // Essentially a query-aborter here since we won't get any results (better would be to not query at all):
    $query->where(1, '=', 0);
}
```

I couldn't find a testcase that covers this behaviour and this addition wasn't explained in https://github.com/laravel/framework/pull/37451 either.

PR #46919, opened 2023-04-29
[10.x] Add generics on app() / resolve() / App::make() / App::makeWith()

daison12006013:
My changes are only to the docblocks, nothing fancy or any changes in the core logic. This should prevent us from using too many @var docblocks when using resolvers.

Please refer to my open discussion here: https://github.com/laravel/framework/discussions/46907

**Works perfectly using this plugin:** [PHP Intelephense](https://marketplace.visualstudio.com/items?itemName=bmewburn.vscode-intelephense-client)

* PHPStan shows no issues without forcing the use of `@var` docblock
* Type-hint

* Hovering to the variable

### DIFF
--- a/bin/facades.php
+++ b/bin/facades.php
@@ -3,6 +3,8 @@
 require __DIR__.'/../vendor/autoload.php';
 
 use Illuminate\Cache\Repository;
+use Illuminate\Container\Container;
+use Illuminate\Foundation\Application;
 use Illuminate\Http\Request;
 use Illuminate\Support\Facades\Date;
 use Illuminate\Support\Str;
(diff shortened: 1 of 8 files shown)


nunomaduro:
We appreciate your contribution with Laravel. However, it's not feasible to infer the precise type using static analysis in functions such as `app()`. The computed type can only be determined at runtime, which may work in some cases but also result in wrong resolutions of the type.

To obtain accurate static analysis for such functions, we recommend using Larastan: https://github.com/nunomaduro/larastan.

daison12006013:
Hi @nunomaduro, though this PR is mainly used for generics, anyone can still override the type by using var docblock.

Can you give an example where it can mislead someone from getting the wrong type?

PR #46920, opened 2023-04-30
[10.x] add user() helper function to reduce the need to write guard()->user()

MouhcineMAHFOUD:
During development, I find my self and my team writing every time auth()->user() to access the currently authenticated user, we use this so much that for almost all projects we add this function as a helper function, so I suggest to add it to the framework, in the hope that it will be very helpful for other people as well.

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -169,10 +169,12 @@ if (! function_exists('auth')) {
 if (! function_exists('user')) {
     /**
      * Get the currently authenticated user.
+     *
+     * @param  string|null  $guard
      */
-    function user(): ?Authenticatable
+    function user($guard = null): ?Authenticatable
     {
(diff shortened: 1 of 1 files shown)


ankurk91:
No more global function please.

Your function does not support multiple guards

MouhcineMAHFOUD:
Hello @ankurk91 
You're right,  in most cases I would avoid adding more global functions, however, in this case it's being used a lot, that's why I think it will be useful to the framework.

I have added support for multiple guards.

antonkomarev:
Can't you get it from the request?

`$request->user()`

donnysim:
I, and I can see many others, already create this helper to type it to specific class instead of interface and to shorten the writing, so this most definitely will be a breaking change for some and would kind of still remove the benefit of typing the return type.

milwad-dev:
You can use this helper in you own application!

PR #46925, opened 2023-05-01
[10.x] Mark commands as isolatable

bytestream:
In https://github.com/laravel/framework/pull/44743 there was merit for marking commands as isolated by default. The work was subsequently changed to an `--isolated` flag to reduce the chances of breakages.

This PR allows the developer to configure the default values:
* `protected $isolated = true;` is equivalent to manually passing `--isolated`
* `protected $isolatedCommandExitCode = 12;` is equivalent to manually passing `--isolated=12`

### DIFF
--- a/src/Illuminate/Console/Command.php
+++ b/src/Illuminate/Console/Command.php
@@ -73,7 +73,7 @@ class Command extends SymfonyCommand
      *
      * @var int
      */
-    protected $isolatedCommandExitCode = self::SUCCESS;
+    protected $isolatedExitCode = self::SUCCESS;
 
     /**
      * The console command name aliases.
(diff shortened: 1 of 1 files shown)


johanrosenson:
Just throwing this idea out, but would this be a good use case for an attribute? 

```
#[Isolated]
#[Isolated(exitCode: 12)] // optional exitCode argument
```

emreakay:
It is an excellent feature that I have been searching for.
I am using Kubernetes in production with autoscaling, but certain commands need to be executed as a single thread only.
How does the "--isolated" option prevent the execution of other commands?
Does the "--isolated" feature function across different pods when using Kubernetes?

@taylorotwell @bytestream @bert-w

emreakay:
@dennisprudlo thanks for great expalanation.

PR #46931, opened 2023-05-01
[10.x] Allow configuration set on `PendingRequest` to carry over to `Pool` requests

### LINKED ISSUE #46923: [HTTP Client] Middleware for the HTTP pool is ignored
### Laravel Version

10.8.0

### PHP Version

8.1.17

### Database Driver & Version

_No response_

### Description

The middleware seems to be completely ignored when using this on the HTTP pool. Without using the HTTP pool and invoking requests using the `withMiddleware` method everything seems to work fine.

Interestingly enough when the pool of requests is actually being built it it is filled with the right PendingRequests which do have the proxy handler 'middleware' on it. So at this point in the code all the PendingRequests which need to be fired off are in the right state. https://github.com/illuminate/http/blob/master/Client/PendingRequest.php#L785 

The logic behind the pool buildup makes even less sense when you are trying to set the middleware like this, instead of on each request separately. 

```
Http::withMiddleware(new Proxy)->pool(fn (Pool $pool) => [
    $pool->get('https://this-is-just-an-example.com'),
]);
```

This is perfectly valid code which executes but each time you call `$pool->get()` you get a new `PendingRequest` which does not have any middleware on it at all since there is no reference to the original `PendingRequest` which was instantiated when calling `Http::withMiddleware()`.

**Update**

After some testing it seems this is more related to the options (example from the invokeable class) not sticking than the middleware not applying at all. I haven't been able to pinpoint the exact problem so far though.

### Steps To Reproduce

An example HTTP pool:

```
    $pool->withMiddleware(new Proxy)->get('https://this-is-just-an-example.com'),
]);
```

In this case I just used a simple invokeable class for the middleware

```
public function __invoke(Closure $handler): Closure
{
    return function (RequestInterface $request, array $options) use ($handler) {
        $options['proxy']['https'] = config('clients.proxy');

        return $handler($request, $options);
    };
}
```

cosmastech:
To resolve https://github.com/laravel/framework/issues/46923.

The following code:
```php
Http::withMiddleware(MyRequestMiddleware::class)->pool(fn(Pool $pool) => [
    $pool->get('https://laravel.com'),
    $pool->get('https://laracasts.com'),
]);

Http::withHeaders([
    'This-Header' => 'Does Not Get Passed'
])->pool(fn ($pool) => [$pool->get('https://laravel.com')]);
```

Does not apply theMyRequestMiddleware to the pooled request. Nor does it pass forward headers.

This PR instead swaps Pool to use `PendingRequest` instead of `Http\Client\Factory` as the basis for making requests.

Now, rather than having to apply these additional properties to each array item in the `pool()` closure, you can write something like this:

```php
use Illuminate\Http\Client\Pool;

Http::baseUrl('https://laravel-news.com')
    ->withUserAgent('Laravel News Fetcher')
    ->withHeaders([
        'Accept' => 'text/html',
    ])
    ->pool(fn(Pool $pool) => [
        $pool->get('/blog'), // https://laravel-news.com/blog
        $pool->get('/category/packages'), // https://laravel-news.com/category/packages
    ]);
```

### DIFF
--- a/src/Illuminate/Http/Client/Pool.php
+++ b/src/Illuminate/Http/Client/Pool.php
@@ -5,16 +5,16 @@ namespace Illuminate\Http\Client;
 use GuzzleHttp\Utils;
 
 /**
- * @mixin \Illuminate\Http\Client\Factory
+ * @mixin \Illuminate\Http\Client\PendingRequest
  */
 class Pool
 {
     /**
-     * The factory instance.
+     * The pending request that is cloned for each async request.
      *
-     * @var \Illuminate\Http\Client\Factory
+     * @var \Illuminate\Http\Client\PendingRequest
      */
-    protected $factory;
+    protected $pendingRequest;
 
     /**
      * The handler function for the Guzzle client.
@@ -34,13 +34,17 @@ class Pool
      * Create a new requests pool.
      *
      * @param  \Illuminate\Http\Client\Factory|null  $factory
-     * @param  callable|null  $handler
+     * @param  \Illuminate\Http\Client\PendingRequest|null  $pendingRequest
      * @return void
      */
-    public function __construct(Factory $factory = null, ?callable $handler = null)
+    public function __construct(Factory $factory = null, PendingRequest $pendingRequest = null)
     {
-        $this->factory = $factory ?: new Factory();
-        $this->handler = $handler ?: $this->getDefaultHandler();
+        if (! $pendingRequest) {
+            $factory ??= new Factory();
+            $pendingRequest = $factory->setHandler($this->getDefaultHandler())->newPendingRequest();
+        }
+
+        $this->pendingRequest = $pendingRequest;
     }
 
     protected function getDefaultHandler(): callable
@@ -70,7 +74,7 @@ class Pool
      */
     protected function asyncRequest()
     {
-        return $this->factory->setHandler($this->handler)->async();
+        return (clone $this->pendingRequest)->async();
     }
 
     /**
--- a/src/Illuminate/Http/Client/Pool.php
+++ b/src/Illuminate/Http/Client/Pool.php
@@ -10,18 +10,11 @@ use GuzzleHttp\Utils;
 class Pool
 {
     /**
-     * The pending request that is cloned for each async request.
+     * The PendingRequest that is a basis for each pooled request.
      *
      * @var \Illuminate\Http\Client\PendingRequest
      */
     protected $pendingRequest;
-
-    /**
-     * The handler function for the Guzzle client.
-     *
-     * @var callable
-     */
-    protected $handler;
 
     /**
      * The pool of requests.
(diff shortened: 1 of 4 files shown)


stephan-v:
I think this is good to get the ball rolling ,but I am unsure if this is the way to go because this only fixes the middleware issue.

The problem is that the pool always instantiates new requests for each request, instead of possibly relying on the `PendingRequest` which instantiated the pool in the first place. If the current `PendingRequest` was passed into the pool and that was used instead it would not only unlock `withMiddleware` but also methods like `withOptions`, `withHeaders`, `baseUrl`, etc.

As of now because of how all pool requests are proxied to the factory none of these work either because you always end up creating an entirely new `PendingRequest`.

src/Illuminate/Http/Client/Pool.php
```diff
@@ -34,17 +34,26 @@ class Pool
      * Create a new requests pool.
      *
      * @param  \Illuminate\Http\Client\Factory|null  $factory
+     * @param  \Illuminate\Http\Client\PendingRequest|null  $pendingRequest
      * @return void
      */
-    public function __construct(Factory $factory = null)
+    public function __construct(Factory $factory = null, PendingRequest $pendingRequest = null)
```
stephan-v, line 40: Is it still needed to pass in the factory when you already leverage a `PendingRequest` which was already instantiated with the factory?

It seems highly unlikely this pool will be instantiated at any time without the initial `PendingRequest`.

src/Illuminate/Http/Client/Pool.php
```diff
@@ -65,7 +74,7 @@ public function as(string $key)
      */
     protected function asyncRequest()
     {
-        return $this->factory->setHandler($this->handler)->async();
+        return (clone $this->pendingRequest)->async();
```
stephan-v, line 77: Is there a specific reason to clone the request instead of just re-using the existing `PendingRequest`? In the end it are the HTTP request methods like get, post, patch, etc. that are responsible for creating the promises when the `PendingRequest` has an `async` property which is set to `true`.

stephan-v (commented):
Hi,

I added a small review which might help push this forward. Hope you don't mind.

So far I like where this is heading.

Cheers.

src/Illuminate/Http/Client/Pool.php
cosmastech: I did this so it would be less likely to be considered a breaking change.

src/Illuminate/Http/Client/Pool.php
cosmastech: In PHP, objects are passed by reference, so as soon as you modify a Pool instance (via `as()` for instance), it will also be modifying the underlying `PendingRequest`. This means calling `Http::pool(fn($pool) => [$pool->as('item1')->get('google.com'), $pool->as('item2')->get('laravel.com')])` is relying on the same PendingRequest instance unless we clone it.

cosmastech:
> Hi,
> 
> I added a small review which might help push this forward. Hope you don't mind.
> 
> So far I like where this is heading.
> 
> Cheers.

I appreciate the notes!

src/Illuminate/Http/Client/Pool.php
cosmastech: Happy to see what the maintainer(s) think, as it would be nice to completely remove the Factory variable.

PR #46932, opened 2023-05-01
Adds expressive logging helper methods

miranalmehrab:
This pull request adds new helper functions for logging purposes. Laravel already provides a ``` info($message) ```  helper but I felt that we were missing the same for ```emergency, alert, critical, error, warning, notice,``` and  ```debug``` level messages. Although the ```logger($message)``` helper can be used to log ``` debug ``` level messages, an explicit ```debug($message)``` helper could be more expressive.

**Examples**
```php
emergency($message, []) 
//Write an emergency level message to the log

alert($message, []) 
//Write an alert level message to the log

critical($message, []) 
//Write a critical level message to the log

error($message, []) 
//Write a error level message to the log

warning($message, []) 
//Write a warning level message to the log

notice($message, []) 
//Write a notice level message to the log

debug($message, []) 
//Write a debug level message to the log
```

Thanks.

### DIFF
--- a/src/Illuminate/Foundation/helpers.php
+++ b/src/Illuminate/Foundation/helpers.php
@@ -461,6 +461,90 @@ if (! function_exists('fake') && class_exists(\Faker\Factory::class)) {
     }
 }
 
+if (! function_exists('emergency')) {
+    /**
+     * Write a emergency level message to the log.
+     *
+     * @param  string  $message
+     * @param  array  $context
+     * @return void
+     */
+    function emergency($message, $context = [])
+    {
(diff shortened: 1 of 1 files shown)


taylorotwell:
Would probably need to be namespaced.

miranalmehrab:
Hi @taylorotwell, thanks for your concern. My whole point in this pull request was to add some syntactic sugar in the form of some new helper functions so that we don't need to import the `Log` facade and write `Log::info('')` or chain the `logger()` helper like `logger()->info('')`. Now, if I move the helper functions to another namespaced class, don't you think that we are back to square one?

PR #46934, opened 2023-05-01
[11.x] Split Facades into package

BafS:
This PR extract "Facades" from "Support" to have a better separation of concerns.

Packages like Log, Cache, Encryption, Filesystem, Hashing, Notifications, Pagination, Pipeline, Process, Routing and Translation require "Support" but do not depend on "Facades".

Doing so it could also potentially be possible to use "`replace`" in composer to have more control over facades.

It should not be a BC, I could target 10.x but it seems safer to target the next version. Let me know if you want me to change.

Thanks in advance


taylorotwell:
I don't think there is a huge advantage to the wider user base of doing this.

BafS:
Is there a disadvantage? It would help to decouple the framework more and it would be possible to opt-out from Facades in some scenarios.

GrahamCampbell:
It's still not really possible to truly opt out of facades. Moving them to a package just moves files around, and doesn't reduce any actual coupling.

GrahamCampbell:
You are free to just not use facades in your application code, also. 🤷

BafS:
Thanks for your answer!

> It's still not really possible to truly opt out of facades. Moving them to a package just moves files around, and doesn't reduce any actual coupling.

I talked on #internals and it's possible if you use some packages that don't require "support", but yes it's limited. The way to go to achieve that would be to

- Not depend on Facade in the framework
- Have the package required by default in the project

Like that, no BC and possibility to opt-out. But it's a bit of change internally, I would be fine doing some PRs thought if that's an options.

> You are free to just not use facades in your application code, also. 🤷

Yes but it's always nice to not depend on something that is not used, but indeed you can enforce that with static analysis or architectural tools like deptrac.

dammy001:
@BafS but won't this cause a lot of refactoring on user land?

BafS:
> @BafS but won't this cause a lot of refactoring on user land?

No because the package would be enabled by default, the namespace would be the same, so there should be no BC, the facades would all be there in the vendor folder by default.

Internally we can still use app('xxx'), dependency injection or even string when necessary to register something (`Foo::class` is `'Foo'` after all).

Issue #42822, opened 2022-06-15
@json throw error when array is more than 3 items


ziming:
- Laravel Version: 9.17
- PHP Version: 8.1.6
- Database Driver & Version: MySQL 8

### Steps To Reproduce:

This below is fine. 3 items. Worked perfectly

```blade
@json(['view', 'form', 'tree'])
```

This below throw error. 4 items in the array.

Error message is Unclosed '[' does not match ')'

```blade
@json(['view', 'form', 'tree', 'code'])
```

If more than 3 items, error happen

To give some extra context,I'm using it like this

```blade
// work with 3 items in array
modes: @json($field['modes'] ?? ['form', 'tree', 'code', 'view']),
```

npostman:
Give your array as a variable.. @json can get 3 parameters:
- your data to json encode
- json_encode flags (optional)
- depth (optional)

So:
```php
$array = ['view', 'form', 'tree', 'code'];
@json($array);
```

works

ziming:
what I dont get is why is 1 item, 2 item, 3 items okay. but the 4th item onwards fail?

npostman:
It is because of the way the blade directive gets compiled to PHP. Look at the method that compiles it down:

```php
protected function compileJson($expression)
    {
        $parts = explode(',', $this->stripParentheses($expression));

        $options = isset($parts[1]) ? trim($parts[1]) : $this->encodingOptions;

        $depth = isset($parts[2]) ? trim($parts[2]) : 512;

        return "<?php echo json_encode($parts[0], $options, $depth) ?>";
    }
```

It basically takes your @json parameters as a string, and explodes it at the comma. It then uses the first bit as your JSON data, the second bit as the json encode options and the third bit as the depth parameter.

So, your code `@json($field['modes'] ?? ['form', 'tree', 'code', 'view'])` would compile down to:

```php
<?php echo json_encode($field['modes'] ?? ['form', 'tree', 'code') ?>
```

For your use-case, you'd do something like this:

```php
@php
$modes = $field['modes'] ?? ['form', 'tree', 'code', 'view'];
@endphp
@json($modes);
```

Or you write a method that gets the modes:
```php
class SomeModel extends Model
{
    public function getModes() {
        return $this->modes ?? ['form', 'tree', 'code', 'view'];
    }
}

// And use it in your blade like so
@json($model->getModes());
```

ankurk91:
```blade 
<script>
    var app = {{ Js::from($array) }};
</script>
```

[...]

Issue #21068, opened 2017-09-07
@json awkwardness


KristofMorva:
I might really miss something, but I think Laravel 5.5.3's new `@json` directive does not really seem to work the way it's supposed to do.

For example, with this input: `@json($places, JSON_HEX_APOS)` I get: `<?php echo json_encode(($places, JSON_HEX_APOS), 0, 512) ?>`. Which is actually realistic, because https://github.com/laravel/framework/blob/5.5/src/Illuminate/View/Compilers/Concerns/CompilesJson.php does not parse parse brackets/options like other Blade tags do. Also, as I have seen, it has no corresponding test (which is probably why it is included in the current release this way).

themsaid:
Fixed that in https://github.com/laravel/framework/commit/2a25ee7f2f2d5e2cbc1397cc24abbb2838a9b439

thanks :)

KristofMorva:
Thanks @themsaid, but won't that break if you have comma in the first expression? Like `@json(old('levels', []))`, our code is filled with things like these.

themsaid:
Yeah it will, blade has a few edge cases and this is one of them, personally I'd just use php to encode to json if I'm not passing a variable

PR #46935, opened 2023-05-01
[10.x] Expose `Js::json()` helper

bert-w:
This solves recurring issues where people would use complex expressions inside the `@json(...)` blade directive, like https://github.com/laravel/framework/pull/46910, https://github.com/laravel/framework/issues/42822, https://github.com/laravel/framework/issues/21068, https://github.com/laravel/framework/issues/46908 and others.

If you ever feel the need to use complex expressions, you should use the `{{ Js::json(...) }}` function and benefit from the preset JSON options as well.

For reference: the `Js::from(...)` exists alongside this function and will echo `JSON.parse(...)` (explicit javascript code), so it is different than `Js::json(...)`. The latter just echoes an agnostic JSON string with some predefined encoding options.

### DIFF
--- a/src/Illuminate/Support/Js.php
+++ b/src/Illuminate/Support/Js.php
@@ -74,7 +74,7 @@ class Js implements Htmlable
             $data = $data->value;
         }
 
-        $json = static::json($data, $flags, $depth);
+        $json = static::encode($data, $flags, $depth);
 
         if (is_string($data)) {
             return "'".substr($json, 1, -1)."'";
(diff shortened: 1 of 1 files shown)


valorin:
I just noticed that `Js::encode()` doesn't return the value wrapped inside `Htmlable`, unlike `Js::from()`. What effect will this have if you're loading it into blade with the `{{ ... }}` tags?
I'm guessing it'll add extra encoding, and break the JSON block?

PR #46936, opened 2023-05-01
Add support for setting BackedEnum as column default value

nazmulpcc:
This will allow using `BackedEnum` values as column default values. If we have an enum class like this,
```php
enum Status: string
{
    case ACTIVE = 'active';
    case INACTIVE = 'inactive';
}
```

Currently, we can do this in migration files,
```php
$table->string('status')->default(Status::ACTIVE->value);
```
Which looks a bit ugly. Changes in this PR will allow us to do this instead,
```php
$table->string('status')->default(Status::ACTIVE);
```

### DIFF
--- a/src/Illuminate/Database/Schema/Grammars/Grammar.php
+++ b/src/Illuminate/Database/Schema/Grammars/Grammar.php
@@ -309,6 +309,10 @@ abstract class Grammar extends BaseGrammar
             return $this->getValue($value);
         }
 
+        if ($value instanceof \BackedEnum) {
+            return "'" . (string) $value->value . "'";
+        }
+
         return is_bool($value)
                     ? "'".(int) $value."'"
                     : "'".(string) $value."'";


cosmastech:
https://github.com/laravel/framework/pull/46595

put a similar MR up less than two months ago and it was rejected and received a lot of not-so-positive reactions

nazmulpcc:
@cosmastech Thanks for mentioning that. 
I still think it would be helpful to have this. The arguments for having this feature is stronger in my opinion. You made some good arguments. And it has a very small footprint, not sure why it would be "hard to maintain" or something like that. But I am going to trust the community and close this.
Do you know if there's any way I can have this in my project though? Maybe some macro method?

cosmastech:
> @cosmastech Thanks for mentioning that. 
> 
> I still think it would be helpful to have this. The arguments for having this feature is stronger in my opinion. You made some good arguments. And it has a very small footprint, not sure why it would be "hard to maintain" or something like that. But I am going to trust the community and close this.
> 
> Do you know if there's any way I can have this in my project though? Maybe some macro method?

That is sort of a generic response, I don't think it has much of anything to do with the footprint of the PR.

If I recall correctly, you can bind your own implementation of the Blueprint in the container and override the method.

PR #46070, opened 2023-02-10
[9.x] Don't overwrite existing `Listeners`

xurshudyan:

PR #46938, opened 2023-05-02
Fix artisan `make:listener` command from overriding existing listeners by default

wayz9:
I accidentally ran make:listener with same class name twice and got surprised when same file was created and all of my changes to the original were gone.

One of the solutions is to remove `alreadyExists` method from `ListenerMakeCommand.php`.
CommandGenerator class will call it's original alreadyExists method which works just fine.


xurshudyan:
See https://github.com/laravel/framework/pull/46070

PR #46939, opened 2023-05-02
Fix php error 8.2.4

Nightprince:
Fix error  php 8.2.4:

> Undefined array key 0 in illuminate\collections\Collection.php on line 1736

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -1733,7 +1733,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
      */
     public function offsetGet($key): mixed
     {
-        return $this->items[$key];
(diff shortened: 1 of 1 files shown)


taylorotwell:
Elaborate?

Nightprince:
> Elaborate?

https://www.php.net/manual/en/class.arrayaccess.php

Nightprince:
> > Elaborate?
> 
> https://www.php.net/manual/en/class.arrayaccess.php

@taylorotwell

ziming:
I think it is a bug too

Nightprince:
@taylorotwell check again

rojtjo:
When implementing `ArrayAccess` you make the object act as an array, if you access a non-existing key on an array it also fails. To make it makes sense that it just fails instead of falling back to `null`.

```php
$array = [];
$array[1];
// Warning: Undefined array key 1 in php shell code on line 1
```

PR #46942, opened 2023-05-03
Update doc block to make array notation more consist.

zobay:
## Description
-  Update doc block to make array notation more consist.

### DIFF
--- a/src/Illuminate/Foundation/Support/Providers/EventServiceProvider.php
+++ b/src/Illuminate/Foundation/Support/Providers/EventServiceProvider.php
@@ -25,7 +25,7 @@ class EventServiceProvider extends ServiceProvider
     /**
      * The model observers to register.
      *
-     * @var array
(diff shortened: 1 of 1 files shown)


src/Illuminate/Foundation/Support/Providers/EventServiceProvider.php
```diff
@@ -25,7 +25,7 @@ class EventServiceProvider extends ServiceProvider
     /**
      * The model observers to register.
      *
-     * @var array
+     * @var array<string, array<int, string>>
```
cosmastech, line 28: I can't recall, does Laravel use `class-string` anywhere else?

```php suggestion
     * @var array<class-string, class-string|array<int, class-string>>
```

Either way, observers can be of the format:
```php
$observers = [
    SomeClass::class => SomeClassObserver::class,
    // or
    SomeClass::class => [SomeClassObserver::class],
];

PR #46943, opened 2023-05-03
[9.x] mb_strwidth deprecate warning fix

veneliniliev:
When passing a null value to Str::limit log this error:

`
mb_strwidth(): Passing null to parameter #1 ($string) of type string is deprecated in /app/vendor/laravel/framework/src/Illuminate/Support/Str.php on line 493
`

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -490,7 +490,7 @@ class Str
      */
     public static function limit($value, $limit = 100, $end = '...')
     {
-        if (mb_strwidth($value, 'UTF-8') <= $limit) {
+        if (mb_strwidth($value ?? '', 'UTF-8') <= $limit) {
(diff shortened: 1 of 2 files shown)


bert-w:
this change is debatable, php changed this for 8.1 and the behaviour is now being undone. There are likely many more spots where a similar error occurs.

GrahamCampbell (changes requested):
👎 You are calling the function wrong. You must not pass null, as per the phpdoc on line 486. This PR should not be merged.

driesvints:
As the docblocks indicate, you're required to pass a string.

veneliniliev:
@GrahamCampbell @driesvints but when you use a nullable string, a lot of warnings are logged...

driesvints:
You can't use a nullable string here.

PR #46944, opened 2023-05-03
[10.x] Use container to build `PendingDispatch`

cosmastech:
## Problem
There is no way to swap in an implementation of `PendingDispatch` currently aside from duplicating the code and setting it as an alias. This is a challenge, as it means every time there's a tagged release on this repository, the developer needs to check for changes and if so, modify their `PendingDispatch` implementation for parity.

### Why would I ever want to do that?
Working on a rewrite of an application that will have both a beta version and a legacy application in production running simultaneously.

Jobs dispatched from the beta application to the legacy application need to modify the `PendingDispatch@__destruct()` method so that they are dispatched to a separate Horizon database. (See https://github.com/laravel/horizon/pull/1271)

## Solution
Build `PendingDispatch` from the container when needed in `Dispatchable`.

### Alternatives
* Add a protected static variable to Dispatcher that holds the FQCN for PendingDispatch and a static method to change it. Doesn't feel particularly clean to me and conflicts with how most binding is done in the framework.

* Write a `queue()` method on things that need to dispatch to different Horizon Redis database
This would be easier if `Dispatcher@pushCommandToQueue()` were public.

* Write a custom `Dispatchable` implementation
Again, entirely duplicating something that may change with a framework release.

### DIFF
--- a/src/Illuminate/Foundation/Bus/Dispatchable.php
+++ b/src/Illuminate/Foundation/Bus/Dispatchable.php
@@ -24,6 +24,7 @@ trait Dispatchable
      *
      * @param  static  $job
      * @return \Illuminate\Foundation\Bus\PendingDispatch
+     *
      * @throws \Illuminate\Contracts\Container\BindingResolutionException
      */
     protected static function makeNewPendingDispatch($job)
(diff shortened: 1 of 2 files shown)


taylorotwell:
I'm not sure I totally want to add in all these customization for a single application that is being rewritten.

cosmastech:
> I'm not sure I totally want to add in all these customization for a single application that is being rewritten.

I found a way to work around it. But I do think that there's a discussion to be had about what is built from the container and what isn't. You completely remove the user land's ability to tweak what's under the hood.

Which is nearly the equivalent of declaring a class as final.

@taylorotwell

PR #46948, opened 2023-05-03
[10.x] `hasAttached()` model factory method without pivot data

gdebrauwer:
In a lot of cases, a belongstomany-relationship does not have any extra pivot columns beside the foreign keys. But when using the `hasAttached()` method on a model factory, you still have to provide an empty array as a second parameter. That seems a bit unnecessary.
```php
Company::factory()
    ->hasAttached(User::factory()->count(2), [], 'users')
    ->create();
```
This PR fixes that, so you can provide the relationship name as the second parameter value if you do not have any pivot data. It's a small DX improvement that has been used in other parts of the framework if I'm not mistaken.
```php
Company::factory()
    ->hasAttached(User::factory()->count(2), 'users')
    ->create();
```


taylorotwell:
I would just suggest: `hasAttached(User::factory()->count(2), relationship: 'users')`

PR #46949, opened 2023-05-03
[10.x] ViewComposerMakeCommand

lee-to:
I don’t know why this command hasn’t existed yet, but I often use the view composer class and often look at the documentation for how it should look and where it’s best to store it, the command will immediately create a class in app/View/Composers

`php artisan make:view-composer`


taylorotwell:
I'm personally just not sure using view composers is a great idea in general.

PR #46950, opened 2023-05-03
Fix "Return value must be of type string, null returned"

jkoop:
> Illuminate\Support\HtmlString::__toString(): Return value must be of type string, null returned

The above `TypeError` was thrown in our project. These changes fixes it.


taylorotwell:
Pass a string to the object.

PR #46952, opened 2023-05-04
refactoring Console GeneratorCommand class

miladev95:
1. The code currently modifies the $model variable in place several times. While this is not necessarily a problem, it can make the code harder to follow. Consider using separate variables for the modified versions of $model, rather than overwriting it each time.
2. using the DIRECTORY_SEPARATOR constant. This can make the code more portable between different operating systems.
3. it returns only a string instead of an array or a string, since the is_dir() check is only needed to determine the namespace to


milwad-dev (commented):
It became to unreadable more!!

PR #46960, opened 2023-05-04
[10.x] Fix typo in HasGlobalScopes PHPdoc tag

wouterj:
Fix a minor typo in the PHPdoc introduced in #46922.

This breaks using Doctrine Annotations with the `Model` class, as the `@params` tag is not excluded by default and thus the library tries to autoload a class called `params`.


driesvints:
Thanks @wouterj!

PR #46961, opened 2023-05-04
[9.x] Document some more return types (removes Symfony 6.3 deprecation notices)

wouterj:
In Symfony 6.3, we've added PHPdoc return types to all methods.

I've run the type patch utility on this repository, to make sure all return types are also defined by Laravel to avoid indirect deprecation notices. I'm not sure about your policy regarding adding return types, so some of these might be turned into a real PHP return type.

### DIFF
--- a/src/Illuminate/Console/BufferedConsoleOutput.php
+++ b/src/Illuminate/Console/BufferedConsoleOutput.php
@@ -27,6 +27,8 @@ class BufferedConsoleOutput extends ConsoleOutput
 
     /**
      * {@inheritdoc}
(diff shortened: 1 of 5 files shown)


driesvints (approved):
Cool, thanks @wouterj. I'll let @nunomaduro answer your native return type question.

nunomaduro:
>  I'm not sure about your policy regarding adding return types, so some of these might be turned into a real PHP return type.

Generally, we only use native types into code that may live on user land or to "new" code that gets introduced on the framework.

PR #46968, opened 2023-05-05
[10.x] Throw timeoutException instead of maxAttemptsExceededException when a job times out

PrinsFrank:
Currently, both when a job times out and when the max attempts are exceeded for a job the maxAttemptsExceeded exception is thrown. This makes debugging job timeouts difficult. This PR adds a new exception that is thrown when the SIGALRM signal is fired.

### DIFF
--- a/src/Illuminate/Queue/TimeoutExceededException.php
+++ b/src/Illuminate/Queue/TimeoutExceededException.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Illuminate\Queue;
+
+class TimeoutExceededException extends MaxAttemptsExceededException
+{
+    //
+}
--- a/src/Illuminate/Queue/Worker.php
+++ b/src/Illuminate/Queue/Worker.php
@@ -213,7 +213,7 @@ class Worker
         pcntl_signal(SIGALRM, function () use ($job, $options) {
             if ($job) {
                 $this->markJobAsFailedIfWillExceedMaxAttempts(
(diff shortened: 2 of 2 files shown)


src/Illuminate/Queue/TimeoutExceededException.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+namespace Illuminate\Queue;
+
+use RuntimeException;
+
+class TimeoutExceededException extends RuntimeException
```
GrahamCampbell, line 7: This could be sent to 10.x if you extended MaxAttemptsExceededException instead of RuntimeException.

GrahamCampbell (changes requested):

driesvints:
@SlyDave how did this break your app?

driesvints:
I still don't understand how this breaks your app?

SlyDave:
It doesn't, as I noted, because of the TimeoutExceededException extending the MaxAttemptsExceeededException.

I incorrectly made the assumption that a new Exception was being added, that would be a breaking change, instead one was added that extends the existing one, which does stop the breaking change from occurring, but is really weird as it's utterly unrelated, and just serves to cause confusion in future. 

If one captures only MaxAttemptsExceeededException, it will capture TimeoutExceededException as well, which is _current_ behaviour but I would argue is unintended behaviour that shouldn't be preserved.

PR #46969, opened 2023-05-05
[9.x] Fix schema:dump command - Use old .dump file name if it exists

florianoversmake:
Version 9.50 introduced a new schema dump-file name with the following PR:
#45805. ( [src/Illuminate/Database/Console/DumpCommand.php](https://github.com/laravel/framework/pull/45805/files#diff-68cddf764fafe83042244527f3269a08f693b857ede34814b8b40f4db5ddd866L97) ).

This behavior creates an additional `mysql-schema.sql` file alongside the existing `mysql-schema.dump` file, which may cause failures in existing project toolchains (this happened to us).

The migrate command already checks for the existence of a `mysql-schema.dump` file and will use it instead of the `.sql` file if it is present. [Relevant Code](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Database/Console/Migrations/MigrateCommand.php#L269)

This Pull Request implements the same logic for the `schema:dump` command.

This improvement will assist developers with existing `mysql-schema.dump` files by eliminating the need to modify their current toolchains.

### DIFF
--- a/src/Illuminate/Database/Console/DumpCommand.php
+++ b/src/Illuminate/Database/Console/DumpCommand.php
@@ -102,7 +102,7 @@ class DumpCommand extends Command
     }
 
     /**
-     * Get the path to the to be stored schema for the given connection.
+     * Get the path to the stored schema for the given connection.
      *
      * @param  \Illuminate\Database\Connection  $connection
      * @return string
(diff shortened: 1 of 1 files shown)


taylorotwell:
I don't understand why we would want this in the dump command? We want it to start using the new format.

florianoversmake:
> I don't understand why we would want this in the dump command? We want it to start using the new format.

Objective accomplished. We have adjusted our workflows accordingly.

We prefer using the `.sql` extension as well; however, this change may cause disruptions in existing CI/CD pipelines (such as those involving diffing or linting the database dump).

It might take some time for users to determine the cause of any issues, as this update is only a minor version change and CI/CD logs are hard to debug some times.

taylorotwell:
Going to hold off on this for now.

PR #46975, opened 2023-05-05
[10.x] Leverage PDO fetch modes to improve `pluck()` and add `keyBy()`

bert-w:
[PHP's PDO Fetch Modes](https://www.php.net/manual/en/pdostatement.fetchall.php) allow developers to set different response formats for their query results. Currently in Laravel, this fetch mode defaults to `PDO::FETCH_OBJ` which returns the query results as a list (numerically indexed array, so no custom keys) where the values are objects.

By introducing the helper `onceWithFetchAllMode(PDO::FETCH_..., callable)`, queries can now have their response format modified according to the given fetch mode. The new query builder feature `keyBy()` (similar to Collection's `keyBy()` but for the query builder) and a huge simplification to `pluck()` make use of these new fetch modes.

## Feature: `Query\Builder::keyBy()`
Similar to the Collection method `keyBy()` but it acts directly on the query response. Retrieving array items by key is essentially an O(1) (constant time) operation so performance benefits can be found all around. Several examples:
#### 1. Key an array by a column:
```php
$users = User::keyBy('email')->get();
$user = $users['foobar@example.com'] ?? null;
```
#### 2. Key an array by a custom expression:
```php
// Retrieve a user for each day (taking the first 10 characters of the timestamp, which is `2023-01-01`):
$users = User::keyBy(DB::raw('SUBSTR(`created_at`, 1, 10)'));
$janFirstUser = $users['2023-01-01'] ?? null;
// Note: if multiple users have `created_at = 2023-04-01`, the last user with this value will overwrite the previous. This is exactly how `Illuminate\Support\Collection::keyBy()` works as well. The query can obviously be improved with a GROUP BY clause.
```

#### 3. Keying relations:
```php
class Post extends BaseModel
{
    public function tags()
    {
        // Return results keyed by the `tag` column (the name of the tag).
        return $this->belongsToMany(Tag::class)->keyBy('tag');
    }
}
$post = Post::with('tags')->first();
$awesomeTag = $post->tags['awesome'] ?? null;
```
(test cases included)
#### Example 1: creating a local cache of currency objects, which can be quickly retrieved by their iso-code (EUR,USD,BTC etc) array index:
```php
class SomeClass
{
    private $_cache;

    public function get($isoCode)
    {
        return $this->cache()[$isoCode] ?? null;
    }

    protected function cache()
    {
        if (is_null($this->_cache)) {
            $this->_cache = App\Models\Currency::query()->keyBy('iso_code')->get();
        }

        return $this->_cache;
    }
}

$euroCurrency = (new SomeClass)->get('EUR');
```

#### Example 2: any cases where you need to check key existence in a query result
(see Feature no. 3)

#### Example 3: Optimize any cases where `->get()->keyBy(...)` (or similar) is called:
This saves an extra array loop since the result is indexed immediately when it is retrieved from PDO.
```php
// Old:
Country::all()->keyBy('country_code');
// New:
Country::keyBy('country_code')->get();
```

## 2/2 Improvement: `Query\Builder::pluck()`
This PR also improves the speed for Query Builder `pluck()` and removes 80+ lines of code. The usage and functionality remains the same, but it is greatly simplified by using `PDO::FETCH_COLUMN` (when calling `pluck('col')` and `PDO::FETCH_UNIQUE | PDO::FETCH_COLUMN` (when calling `pluck('col', 'id')`). The query result is now exactly the format you expect from the pluck function, so no further modification is necessary.

### DIFF
--- a/src/Illuminate/Collections/Collection.php
+++ b/src/Illuminate/Collections/Collection.php
@@ -780,14 +780,15 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
      * @template TMapToDictionaryValue
      *
      * @param  callable(TValue, TKey): array<TMapToDictionaryKey, TMapToDictionaryValue>  $callback
+     * @param  bool  $keepKeys
      * @return static<TMapToDictionaryKey, array<int, TMapToDictionaryValue>>
      */
-    public function mapToDictionary(callable $callback)
+    public function mapToDictionary(callable $callback, $keepKeys = false)
     {
         $dictionary = [];
 
-        foreach ($this->items as $key => $item) {
-            $pair = $callback($item, $key);
+        foreach ($this->items as $originalKey => $item) {
+            $pair = $callback($item, $originalKey);
 
             $key = key($pair);
 
@@ -797,7 +798,7 @@ class Collection implements ArrayAccess, CanBeEscapedWhenCastToString, Enumerabl
                 $dictionary[$key] = [];
             }
 
-            $dictionary[$key][] = $value;
+            $keepKeys ? $dictionary[$key][$originalKey] = $value : $dictionary[$key][] = $value;
         }
 
         return new static($dictionary);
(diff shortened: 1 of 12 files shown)


crissi:
would it make sense to make keyBy() return the collection just like pluck() does?
so instead of **->keyBy()->get()** it would just be **keyBy()** to get the collection.

Might simplify the code a bit?!

taylorotwell:

PR #46982, opened 2023-05-06
This MUST be an array as in the base class

alireza1992:


rodrigopedra:
Laravel does not type any class properties like that. And even if it did, as this property didn't have one proper type this would be a breaking change to any class extending this one and overriding this property for any reason (e.g. replace default assignment values).

PR #46984, opened 2023-05-07
[10.x] Prevent throw exception when notifiable id not exist

saeedhosseiinii:
This line throws an exception when notifiable is not an object or object without an id attribute, so this merge fixes it.


taylorotwell:
No example given of when this would happen.

PR #46985, opened 2023-05-08
[10.x] Enable use of non global morph mappings for MorphTo & related

theminecoder:
In certain cases it is useful to have morph mappings that are disconnected from the global mappings - eg working in a db structure where the same custom key means a different class in different tables (thank you external provider).

I have added an option to pass custom mappings to the morph methods I could find which will then override the global mapping for that segment of the query/relationship rather then having to set and unset the global mapping state.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasRelationships.php
@@ -297,9 +297,9 @@ trait HasRelationships
      */
     protected function morphEagerTo($name, $type, $id, $ownerKey, $mappings)
     {
-        return $this->newMorphTo(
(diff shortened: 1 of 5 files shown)


antonkomarev:
This is breaking change, it could be added only to the next major release (master branch).

theminecoder:
It has been intergrated in a way that doesn't require a breaking change in form of [argument shifting](https://github.com/laravel/framework/pull/46985/files#diff-94fc5521c4c3013979aa52c1d6d5dad652e6e10467b7e45481736d3c986e099fR223) in the hopes it can go in sooner then next year. The other option is placing the mappings parameter to the back of the parameter lists but I felt it was batter situated before closures etc but happy to take pointers here.

antonkomarev:
There are changes in public method signature. It's a breaking change, because they may be overridden in some cases

PR #46987, opened 2023-05-08
[10.x] Add arguments to the signed middleware to ignore properties

WendellAdriel:
## Why

Based on this **[tweet](https://twitter.com/ClaudioDekker/status/1655499868326985728)** from Claudio Dekker. This allows passing properties to the `signed` middleware to ignore properties instead of extending it and adding the keys to the `$ignore` property.

Also, this adds the possibility to pass the ignore parameters to the static methods: `relative` and `absolute`.

## Pros of this implementation

- Make the routes file easier to understand
- Remove a lot of 'custom' Middleware files that individually take up a lot of 'visual' space while providing near-zero real value

### Middleware

```php
// Before
Route::get('foo')->middleware('signed');
Route::get('embed')->middleware(EmbeddableSignedRoute::class);
Route::get('callback')->middleware(SignedRouteWithCallback::class);
Route::get('relative')->middleware('signed');

// After
Route::get('foo')->middleware('signed');
Route::get('embed')->middleware('signed:embed');
Route::get('callback')->middleware('signed:callback');
Route::get('callback')->middleware('signed:callback');
Route::get('relative')->middleware('signed:relative');
Route::get('relative-embed')->middleware('signed:relative,embed');
```

### Static Methods

```php
ValidateSignature::absolute('foo', 'bar'); // 'Illuminate\Routing\Middleware\ValidateSignature:foo,bar'
ValidateSignature::relative('foo', 'bar'); // 'Illuminate\Routing\Middleware\ValidateSignature:relative,foo,bar'
```

### DIFF
--- a/src/Illuminate/Routing/Middleware/ValidateSignature.php
+++ b/src/Illuminate/Routing/Middleware/ValidateSignature.php
@@ -71,9 +71,10 @@ class ValidateSignature
             array_shift($args);
         }
 
-        $ignore = empty($args)
-            ? property_exists($this, 'except') ? $this->except : $this->ignore
-            : $args;
+        $ignore = array_merge(
+            property_exists($this, 'except') ? $this->except : $this->ignore,
+            $args
+        );
 
         return [$isRelative, $ignore];
     }
--- a/src/Illuminate/Routing/Middleware/ValidateSignature.php
+++ b/src/Illuminate/Routing/Middleware/ValidateSignature.php
@@ -4,6 +4,7 @@ namespace Illuminate\Routing\Middleware;
 
 use Closure;
 use Illuminate\Routing\Exceptions\InvalidSignatureException;
+use Illuminate\Support\Arr;
 
 class ValidateSignature
 {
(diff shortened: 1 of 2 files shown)


erikgaal:
Hi there,

I guess this is an acceptable solution. Only issue I see is when you would want to ignore the `relative` parameter within an absolute URL. Might be a bit pedantic, but we should not forget about about developer experience and what someone can expect when using these API's. 

Can we maybe update the factory-like static methods to include ignored parameters?

```
ValidateSignature::absolute('ignore', 'me');
```

WendellAdriel:
> Hi there,
> 
> I guess this is an acceptable solution. Only issue I see is when you would want to ignore the `relative` parameter within an absolute URL. Might be a bit pedantic, but we should not forget about about developer experience and what someone can expect when using these API's.
> 
> Can we maybe update the factory-like static methods to include ignored parameters?
> 
> ```
> ValidateSignature::absolute('ignore', 'me');
> ```

That's a nice idea, let me do some testing here!

WendellAdriel:
@erikgaal added the implementation that you suggested, thanks! 😉 💪

taylorotwell:
The middleware already ships in the skeleton though? You don't have to extend or do anything manually.

WendellAdriel:
@taylorotwell this is in cases like in the example where multiple routes will have different ignore parameters, in this case, it's needed to create new middleware classes to extend and add the needed parameters in the ignore/except.

With this PR the current behavior remains the same if people want to use it with the ignore/except property but it will also provide the other way to use it with the list of properties in the middleware definition.

dennisprudlo:
I have not better solution but `signed:embed` kind of suggests that the `embed` parameter is actually the one the will be signed.

But ever Alternative I can think of (`signed:except:embed,state` and so on) makes it even worse 😅

Best alternative I can think of is a fluent `->signed(...)` method but I don't know if this is desired.

taylorotwell:
@WendellAdriel one thought I have with this is should the passed ignore arguments **override** the default "except" ones totally or should they merge with them?

WendellAdriel:
@taylorotwell that's a really good question.
Right now it's set to override and I think it makes sense because you can have your default configuration on the except and if you need something custom, then you can pass manually the arguments to ignore.

taylorotwell:
@WendellAdriel I guess I can see it both ways. You may define the things you **always** want to ignore in the middleware. If you add additional ones on route registration I would potentially expect those (the things you want to **sometimes** ignore) to be **merged** in with the things you always want to ignore.

WendellAdriel:
@taylorotwell tomorrow morning my time I’ll update the PR with that then!

WendellAdriel:
@taylorotwell I updated the PR with some changes to allow it to merge the arguments passed in the middleware with the ignore/except property and added a new test case for that

taylorotwell:
Thanks!

WendellAdriel:
> Thanks!

I’m happy to help!!! :muscle:

PR #46990, opened 2023-05-08
[10.x] Allow no-cache headers to return untouched response with bypass option

FrittenKeeZ:
Follow up from #46986

We're trying to find a way to completely bypass cached responses, including not setting any cache headers, by utilizing the `Cache-Control: no-cache` or `Pragma: no-cache` headers.

After reading [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Caching#force_reload) on caching, I realized it might not be preferred behavior to simply return untouched responses when providing any `no-cache` header, so I opted for a `bypass` option which only responds to `no-cache` for now.
This way there are no breaking changes for people who rely on the existing behavior.

### DIFF
--- a/src/Illuminate/Http/Middleware/SetCacheHeaders.php
+++ b/src/Illuminate/Http/Middleware/SetCacheHeaders.php
@@ -49,6 +49,13 @@ class SetCacheHeaders
             $options = $this->parseOptions($options);
         }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
Hey there - please just extend this middleware in your own application to do what you want it to do.

PR #46992, opened 2023-05-09
[10.x] HasCasts returning false instead of true

jure-knezovic:
In current casting implementation, when class is used as a cast type, hasCast is returning false, as it is falling back to else in case class-string is given, which in turn trims it & makes it lowercase. That means f.e. App\Enums\StatusEnum becomes app\enums\statusenum, which in hasCasts function is case-sensitive compared with actual class and is therefore returning false instead of true. 

This PR offers a simple fix which checks if provided string is a class instance & if it is, it returns it as is from getCastType function. 

Steps to reproduce:
1. Create model with enum cast
2. Try calling $model->hasCast('enum',Enum::class) -> expected true, false received

### DIFF
--- a/tests/Database/DatabaseEloquentModelTest.php
+++ b/tests/Database/DatabaseEloquentModelTest.php
@@ -349,6 +349,15 @@ class DatabaseEloquentModelTest extends TestCase
         $this->assertTrue($model->isDirty('asEnumArrayObjectAttribute'));
     }
 
+    /**
+     * @requires PHP >= 8.1
+     */
+    public function testHasCastsOnEnumAttribute()
+    {
+        $model = new EloquentModelEnumCastingStub();
+        $this->assertTrue($model->hasCast('enumAttribute' , StringStatus::class));
+    }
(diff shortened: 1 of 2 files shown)


tests/Database/DatabaseEloquentModelTest.php
```diff
@@ -349,6 +349,15 @@ public function testDirtyOnEnumArrayObject()
         $this->assertTrue($model->isDirty('asEnumArrayObjectAttribute'));
     }
 
+    /**
+     * @requires PHP >= 8.1
+     */
```
driesvints, line 354: PHP 8.1 is the minimum version so this isn't needed.

driesvints: I removed the rest of these in https://github.com/laravel/framework/pull/46993

jure-knezovic: Merged your changes & removed unecessary checks in latest commit

PR #46995, opened 2023-05-09
[10.x] Add Support\Str methods for inserting substrings

patinthehat:
### Summary

This PR adds several helper methods to the `Support\Str` helper class for inserting substrings into a given string.  All methods include full unit tests as well as comment docblocks.

These methods provide an enhanced developer experience as it makes inserting a substring at a particular point within a string trivial and eliminates the need to do any calculations (i.e. position to insert, location of substring to insert after, etc.).

This does not break any features as it only introduces new features and does not modify any existing code.

### Additional Methods

- `insert()` - inserts a string at the specified position in the given string.
- `insertAfterMatch()` - inserts a string after the specified pattern match.
- `insertAfter()` - inserts a string after the specified substring.

### Examples

```php
$new = Str::insert('HelloWorld', '--', 5); //returns 'Hello--World'
$new = Str::insertAfterMatch('HelloWorld', '/(Hello)/', ' '); // returns 'Hello World'
$new = Str::insertAfter('HelloWorld', 'H', '_'); // returns 'H_elloWorld'
```
As far as making building web applications easier - consider the following use case:

```php
// the identifier in the database is UA-1001, but we want to
// allow the user to provide the identifier without a dash for ease of entry:

$providedIdentifier = 'UA1001';
$actualIdentifier = Str::insertAfter($providedIdentifier, 'UA', '-'); 
$userExists = User::where('identifier', $actualIdentifier)->exists();
```

I believe these additional methods will enhance the overall developer experience and provide value to the overall Laravel framework.

### DIFF
--- a/src/Illuminate/Support/Str.php
+++ b/src/Illuminate/Support/Str.php
@@ -459,9 +459,9 @@ class Str
     /**
      * Insert a substring at a given position in the string.
      *
-     * @param string $subject
-     * @param string $insert
-     * @param int $position
+     * @param  string  $subject
+     * @param  string  $insert
+     * @param  int  $position
      * @return string
      */
     public static function insert($subject, $insert, $position)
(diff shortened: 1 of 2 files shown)


justinexoplanet:
This is functionality I've needed in the past. This would be a great addition.

taylorotwell:

PR #46997, opened 2023-05-09
[10.x] add ability to set broadcast event backoff method

dammy001:
This PR enables setting broadcast event backoff using method

```php
class BroadcastTestEvent implements ShouldBroadcast
{
    use Dispatchable, InteractsWithSockets, SerializesModels;

    public int $tries = 2;
    
    public function backoff()
    {
       return 5;
    }
}
```


driesvints:
Please add descriptions to your PRs.

PR #46999, opened 2023-05-09
[10.x] Prevent missing translations using `Translator::preventMissingTranslations()`

ralphjsmit:
This PR adds a warning system to the `Translator` class, that will allow developers to throw exceptions when a missing translation is detected, similar to the already existing Eloquent strict modes. You can also provide a custom callback for when a missing translation is detected, which is very useful for e.g. production environments, where you don't want to throw an exception, but only be informed immediately. 

Enable missing translations like this:
```php
use Illuminate\Translation\Translator;

Translator::preventMissingTranslations(app()->isLocal());
```

Report violations using a custom callback:

```php
Translator::handleMissingTranslationViolationUsing(function (string $key): void {
    report("Missing translation detected for key [{$key}].");
});
```

Thanks!

### DIFF
--- a/src/Illuminate/Translation/MissingTranslationViolationException.php
+++ b/src/Illuminate/Translation/MissingTranslationViolationException.php
@@ -6,23 +6,23 @@ use RuntimeException;
 
 class MissingTranslationViolationException extends RuntimeException
 {
-	/**
-	 * The key of the missing translation.
-	 *
-	 * @var string
-	 */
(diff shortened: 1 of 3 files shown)


ralphjsmit:
@taylorotwell Allright, would you be open to a different way for reporting missing translations (in production environments)? Like emitting an event (similar to the queue events), which people can attach their own listener to?

PR #47001, opened 2023-05-09
[10.x] Add `Lang::whenMissingTranslation()` callback support

ralphjsmit:
As a follow-up to #46999, this PR implements an alternative, less intrusive, system for detecting missing translations. It is similar to the `DB::whenQueryingForLongerThan()` or the kernel `whenCommandLifecycleIsLongerThan()` handlers.

```php
Lang::whenMissingTranslation(function (string $key): void {
    report("Missing translation for key [{$key}]");
});
```

(I have not added the facade docblock, since I assume it is now generated automatically.)

Thanks!

### DIFF
--- a/src/Illuminate/Contracts/Translation/Translator.php
+++ b/src/Illuminate/Contracts/Translation/Translator.php
@@ -39,4 +39,12 @@ interface Translator
      * @return void
      */
     public function setLocale($locale);
+	
+	/**
+	 * Register a callback to be invoked when the translator detects a missing translation.
+	 *
+	 * @param  callable  $handler
+	 * @return void
+	 */
+	public function whenMissingTranslation($handler);
 }
--- a/src/Illuminate/Contracts/Translation/Translator.php
+++ b/src/Illuminate/Contracts/Translation/Translator.php
@@ -39,12 +39,12 @@ interface Translator
      * @return void
      */
     public function setLocale($locale);
-	
-	/**
-	 * Register a callback to be invoked when the translator detects a missing translation.
-	 *
-	 * @param  callable  $handler
-	 * @return void
-	 */
-	public function whenMissingTranslation($handler);
+
+    /**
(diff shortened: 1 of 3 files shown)


nunomaduro (changes requested):

src/Illuminate/Translation/Translator.php
```diff
@@ -147,7 +162,13 @@ public function get($key, array $replace = [], $locale = null, $fallback = true)
         // If the line doesn't exist, we will return back the key which was requested as
         // that will be quick to spot in the UI if language keys are wrong or missing
         // from the application's language files. Otherwise we can return the line.
-        return $this->makeReplacements($line ?: $key, $replace);
+        $line = $this->makeReplacements($line ?: $key, $replace);
+
+        foreach ($this->missingTranslationHandlers as $missingTranslationHandler) {
+            $missingTranslationHandler($key);
```
ralphjsmit, line 168: @nunomaduro Should we use `call_user_func()` here (I noticed that `call_user_func()` in a few different other places as well) or is this OK?

ralphjsmit:
Not sure why some tests are failing on some DBs - cannot see the reason apart from "This job failed".

src/Illuminate/Translation/Translator.php
nunomaduro: This is fine.

src/Illuminate/Contracts/Translation/Translator.php
```diff
@@ -39,4 +39,12 @@ public function getLocale();
...
      */
     public function setLocale($locale);
+
+    /**
+     * Register a callback to be invoked when the translator detects a missing translation.
+     *
+     * @param  callable  $handler
+     * @return void
+     */
+    public function whenMissingTranslation($handler);
```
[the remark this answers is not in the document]
ralphjsmit: Done!

PR #47004, opened 2023-05-09
[10.x] add expression to DB table doctype

RoccoHoward:
When using DB::raw() against the table method, the parameter type is `Illuminate\Contracts\Database\Query\Expression`

```
DB::table(DB::raw(''))
```

```
Parameter #1 $table of static method Illuminate\Database\Connection::table() expects
         Closure|Illuminate\Database\Query\Builder|Illuminate\Database\Query\Expression|string,
         Illuminate\Contracts\Database\Query\Expression given.
```

### DIFF
--- a/src/Illuminate/Database/Connection.php
+++ b/src/Illuminate/Database/Connection.php
@@ -317,7 +317,7 @@ class Connection implements ConnectionInterface
     /**
      * Begin a fluent query against a database table.
      *
-     * @param  \Closure|\Illuminate\Database\Query\Builder|\Illuminate\Database\Query\Expression|string  $table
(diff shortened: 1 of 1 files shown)


src/Illuminate/Database/Connection.php
```diff
@@ -317,7 +317,7 @@ public function getSchemaBuilder()
     /**
      * Begin a fluent query against a database table.
      *
-     * @param  \Closure|\Illuminate\Database\Query\Builder|\Illuminate\Database\Query\Expression|string  $table
+     * @param  \Closure|\Illuminate\Database\Query\Builder|Illuminate\Contracts\Database\Query\Expression|string  $table
```
bytestream, line 320: Why was the `\` removed? There is no definition of `\Illuminate\Database\Illuminate\Contracts\Database\Query\Expression`

driesvints: Seems this was fixed post-merge 👍

PR #47009, opened 2023-05-10
[8.x] More readable error for mailgun transport

yourchocomate:
Thought, someone didn't notice it or think of it.

Usually it is a pain when one can't find out if the problem is with connecting to the api server or somehow the server rejecting the request. Anything happens with the connection it just throwing the message "Request to Mailgun API failed" which is a headache.

That's why this pull, one can now see what actually happened when connecting to the api.
Hope it helps, Thank you

### DIFF
--- a/src/Illuminate/Mail/Transport/MailgunTransport.php
+++ b/src/Illuminate/Mail/Transport/MailgunTransport.php
@@ -77,7 +77,7 @@ class MailgunTransport extends Transport
                 $this->payload($message, $to)
             );
         } catch (GuzzleException $e) {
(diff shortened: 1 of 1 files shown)


driesvints:
Laravel v8 is not supported anymore.

yourchocomate:
> Laravel v8 is not supported anymore.

Some software still uses laravel v8

dennisprudlo:
> Some software still uses laravel v8

Yes, however it's highly recommended to stay up to date. This is neither a bugfix nor a security fix but even if it was Laravel 8 is end of life: https://laravel.com/docs/10.x/releases#support-policy

PR #47014, opened 2023-05-10
[10.x] Allow passing an array with values to the filled() helper

jochensengier:
This PR will allow you to pass an array of values to the _filled()_ helper method.
The same principle is already applied to the _filled()_ helper method in [Illuminate/Http/Concerns/InteractsWithInput](https://github.com/laravel/framework/blob/10.x/src/Illuminate/Http/Concerns/InteractsWithInput.php#L142).

Instead of this:
```
@if (filled($var1) && filled($var2) && filled($var3))
@endif
```

You would be able to do this:
```
@if (filled([$var1, $var2, $var3]))
@endif
```

I know there are other ways to accomplish the same thing, but this seems like a nice little papercut and an opportunity to streamline both methods.

This might however cause a new confusion with the inverse method of _filled()_, namely _blank()_, where currently one value is allowed.

Thanks for taking the time to review this PR.
I'd appreciate any feedback.

### DIFF
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -148,7 +148,7 @@ if (! function_exists('filled')) {
      */
     function filled($value)
     {
-        $values = is_array($value) ? $value : func_get_args();
+        $values = Arr::wrap($value);
 
         if (empty($values)) {
             return false;
--- a/src/Illuminate/Support/helpers.php
+++ b/src/Illuminate/Support/helpers.php
@@ -150,6 +150,10 @@ if (! function_exists('filled')) {
     {
         $values = is_array($value) ? $value : func_get_args();
 
+        if (empty($values)) {
+            return false;
+        }
+
         foreach ($values as $value) {
             if (blank($value)) {
                 return false;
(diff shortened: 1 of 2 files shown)


nunomaduro (changes requested):
Let's hear @taylorotwell's thoughts on this. Nonetheless, in my view, this pull request brings in breaking changes for all these scenarios:

```
        $this->assertFalse(filled([null])); // true
        $this->assertFalse(filled([''])); // true
        $this->assertFalse(filled(['  '])); // true

        $object = new SupportTestCountable();
        $this->assertFalse(filled([$object])); // true
        $this->assertFalse(filled([[]])); // true
        $this->assertFalse(filled([null, '', '  '])); // true
        $this->assertFalse(filled([10, true, false, 0, 0.0, null, '', '  '])); // true
```

Previously, all of these tests would have failed, indicating to me that the new code has modified the behavior of the filled function when an array is passed as an argument.

Maybe @taylorotwell thinks that we should bring this to Laravel `v11.x`, so let's wait a few minutes.

jochensengier:
Yeah, feeling less confident about this PR as it progresses.
Don't want to be responsible for introducing a BC.

I'll leave it up to you and the rest of the team.

No hard feelings if you decide to close without merging!

taylorotwell:
Breaking change. For example `[' ']` would have previously been considered filled and now it isn't.

Issue #47012, opened 2023-05-10
Enum cast with an accessor does not work on Eloquent model


linaspasv:
### Laravel Version

10.9.0

### PHP Version

8.2

### Description

When I have enum cast and an accessor for the same field, the ``attributesToArray`` method stops to work and produces ``Cannot instantiate enum App\Enums\ProductStatus`` error. 

It looks like the problem originates in ``$caster = $this->resolveCasterClass($key)`` at ``HasAttributes.php`` where enum is incorrectly treated.

### Steps To Reproduce

```php
namespace App\Enums;

enum ProductStatus: string
{
    case ACTIVE = 'active';
    case PAUSED = 'paused'; 
    case OFFLINE = 'offline'; 
}
```

```php
class Product extends Eloquent
{
    protected $table = 'products';
    protected $casts = [
        'status' => ProductStatus::class,
    ];

    protected function status(): Attribute
    {
        return Attribute::get(fn ($value) => $value);
    }
}
```

```php
Product::find(1)->attributesToArray();    // <--- I get Cannot instantiate enum error here
```

lazer-hybiz:
Why would you have both a cast and an accessor?

linaspasv:
> 
```php
    protected function status(): Attribute
    {
        return Attribute::get(
            fn ($value) => $this->attributes['child_status'] ?? $value,
        );
    }
```

because I have an override depending if additional data is loaded.

Anyway, it's not why I have both because it's only happening with enum. I also tried with Castable objects or other ones and it works as intended. :-)

cosmastech:
```php
    protected function status(): Attribute
    {
        return Attribute::get(
            fn ($value) => ProductStatus::from($this->attributes['child_status'] ?? $value)
        );
    }
```

wouldn't this work?

My understanding is that an Attribute is dealing with the raw DB value and the client code is responsible for the appropriate casting.... or does it not work this way with other casts? 🤔 If not, can you give an example?

This could be fixed this way, but I'm not positive this was the intention of `Attribute::get()`

```php
protected function isClassCastable($key)
{
    $casts = $this->getCasts();
    
    if (! array_key_exists($key, $casts)) {
        return false;
    }
    
    $castType = $this->parseCasterClass($casts[$key]);
    
    if (in_array($castType, static::$primitiveCastTypes)) {
        return false;
    }

    // change here
    if (enum_exists($castType)) {
        return false;
    }

    if (class_exists($castType)) {
        return true;
    }
    
[...]

PR #47023, opened 2023-05-10
[10.x] Allows Enum casts to be used together with an accessor

linaspasv:
This PR addresses an issue where using an Enum cast and an accessor for the same field causes the ``attributesToArray`` method to throw a ``Cannot instantiate enum error``. The issue seems to stem from the ``resolveCasterClass`` method in ``HasAttributes.php`` where the Enum case is incorrectly treated. ( #47012 )

Steps to reproduce:
```php
enum ProductStatus: string
{
    case ACTIVE = 'active';
    case PAUSED = 'paused'; 
    case OFFLINE = 'offline'; 
}

class Product extends Eloquent
{
    protected $table = 'products';
    protected $casts = [
        'status' => ProductStatus::class,
    ];

    protected function status(): Attribute
    {
        return Attribute::get(fn ($value) => 'offline' ?? $value);
    }
}

Product::find(1)->attributesToArray(); // <-- Cannot instantiate enum error occurs here
```

After the PR all of these cases will work and output a valid enum ``offline`` string when used with ``attributesToArray``:

```php
   protected function status(): Attribute
   {
        return Attribute::get(fn ($value) => $this->castAttribute('status', 'offline' ?? $value));
    }
```

```php
    protected function status(): Attribute
    {
        return Attribute::get(fn ($value) => 'offline' ?? $value);
    }
```

```php
    public function getStatusAttribute()
    {
        return ProductStatus::OFFLINE;
    }
```

```php
    public function getStatusAttribute()
    {
        return 'offline';
    }
```

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -706,6 +706,12 @@ trait HasAttributes
         } else {
             $value = $this->mutateAttribute($key, $value);
         }
+        
+        if ($this->isEnumCastable($key)) {
+            $value = $this->getEnumCastableAttributeValue($key, $value);
+
+            return $this->getStorableEnumValue($value);
+        }
 
         return $value instanceof Arrayable ? $value->toArray() : $value;
     }
@@ -1585,6 +1591,10 @@ trait HasAttributes
         $castType = $this->parseCasterClass($casts[$key]);
 
         if (in_array($castType, static::$primitiveCastTypes)) {
+            return false;
+        }
+        
+        if (enum_exists($castType)) {
             return false;
         }
 
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -706,7 +706,7 @@ trait HasAttributes
         } else {
             $value = $this->mutateAttribute($key, $value);
         }
-        
+
         if ($this->isEnumCastable($key)) {
             $value = $this->getEnumCastableAttributeValue($key, $value);
 
@@ -1593,7 +1593,7 @@ trait HasAttributes
         if (in_array($castType, static::$primitiveCastTypes)) {
             return false;
         }
-        
+
         if (enum_exists($castType)) {
             return false;
         }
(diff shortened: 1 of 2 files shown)


nunomaduro:
Thank you, @linaspasv, for submitting the pull request. Would you mind adding tests to these changes?

linaspasv:
@nunomaduro PR got updated with tests.

taylorotwell:
I don't think I'm going to fiddle around with Eloquent logic atm. It's complicated and can easily be broken for applications in edge cases. Personally I don't think you should define an accessor and a cast on the same field. I'm not sure if we even have defined behavior for that situation. If you define an accessor, remove the cast and handle the casting manually in your accessor.

linaspasv:
> 

Regarding the handling of primitive types, the current implementation allows to easily mix accessor with a cast. Class castables got their defined rules but there is nothing for Enums when it comes to casting into array right now. The current logic just tries to ``return new $castType(...$arguments);`` Enum and it's completely wrong. :-)

linaspasv:
@taylorotwell so I have tried this to make appear on ``attributesToArray()``.

```php

    protected $appends = [
        'merged_status'
    ];

    protected function mergedStatus(): Attribute
    {
        return Attribute::get(fn () => $this->child_status ?? $this->attributes['status']);
    }
    
```

Unfortunately, I can't use ``ProductStatus::tryFrom(...)`` on accessor because ``attributestoArray()`` just prints this as enum, where it should just be a string (like it would be with a regular cast but without an accessor), and I can't use cast as it results into the error. So in general enum becomes useless in my case and it does not make a difference from a regular ``string`` type. :-/

AMoktar:
hello, I'm getting `Cannot instantiate enum App\Enums\ProductStatus` with the example above 

```json
"laravel/framework": "^10.28.0"
```

```php
<?php

namespace App\Models;

use App\Enums\ProductStatus;

class Product extends Model
{
    use HasFactory;


    /**
     *
     * @var array
     */
    protected $casts = [
        'status' => ProductStatus::class,
    ];


    protected function status(): Attribute
    {
        return Attribute::get(fn ($value) => 'offline' ?? $value);
    }
}
```

```php
<?php

namespace App\Enums;

enum ProductStatus: string
{
    case ACTIVE = 'active';
    case PAUSED = 'paused';
    case OFFLINE = 'offline';
}
```

PR #47025, opened 2023-05-10
[10.x] Adds `Lang::whenMissingTranslation`

nunomaduro:
The original suggestion for this pull request came from https://github.com/laravel/framework/pull/47001. It enables users to report missing translations in their applications in the manner they prefer:

```php
use Illuminate\Support\Facades\Lang;

{
    /**
     */
    public function boot(): void
    {
        Lang::whenMissingTranslation(fn ($key) => logger('Missing translation', ['key' => $key]));
    }
}
```

### DIFF
--- a/src/Illuminate/Contracts/Translation/Translator.php
+++ b/src/Illuminate/Contracts/Translation/Translator.php
@@ -39,4 +39,12 @@ interface Translator
      * @return void
      */
     public function setLocale($locale);
+	
+	/**
+	 * Register a callback to be invoked when the translator detects a missing translation.
+	 *
+	 * @param  callable  $handler
+	 * @return void
+	 */
+	public function whenMissingTranslation($handler);
 }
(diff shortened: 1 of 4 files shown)


taylorotwell:
Can use a community package like: https://github.com/LarsWiegers/laravel-translations-checker

ralphjsmit:
@taylorotwell That does not work for translation keys that are concatenated/generated dynamically.

ralphjsmit:
To clarify, let me explain our situation:

- Many notifications in our app all follow the same structure (header, body, button).
- Therefore, we didn't really want to maintain many very similar markdown/Blade templates.
- We created a base notification class with one Blade template
- Every notification extends that base class, and translation keys are generated dynamically based on the class name
- However, sometimes translation keys get missing (complex setup with Localazy and PRs/merges, multiple team members, not easy fix)
- Users get sent notifications without translations = full of translation keys
- We would like to get alerted immediately in case any notification goes wrong

Issue #46744, opened 2023-04-11
Class cast cache causing trouble when setting a collection object


Propaganistas:
### Laravel Version

10.7.0

### PHP Version

8.2.4

### Description

I have a custom cast class for wrapping value objects in a collection.
The `set()` method serializes the value into JSON, the `get()` method maps all items it receives to a value object and then creates a collection out of it. Quite a simple cast;

```php
public function get($model, $key, $value, $attributes)
{
    if (is_null($value)) {
        return null;
    }

    $data = json_decode($attributes[$key], true);

    return (new Collection($data))->map(function ($item) {
        return new ValueObject($item);
    });
}
```

Things go wrong when providing the same data in different ways:

```php
$model->items = $request->input('items'); // contains an array of arrays
```

When providing the value as an array, it all works fine. The attribute is set properly and returns as the casted value object collection when getting it.

```php
$model->items = $request->collect('items'); // contains a collection of arrays
```

However, when providing the value as a collection, the attribute is set properly BUT returns a collection of arrays instead.
After some debugging I figured that the **class cast cache** is causing trouble. More specifically this part:

https://github.com/laravel/framework/blob/ddbbb2b50388721fe63312bb4469cae13163fd36/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1129-L1133

It seems that, upon setting the value, the original value gets cached if it's an object while it hasn't been casted yet. So in the example above the exact output of `$request->collect('items')` is cached and returned on subsequent retrieval. In the case of `$request->input('items')` an array is provided so the value never makes it into the cache.

The easy fix would consist of calling the `get()` method of the cast and cache that instead, but I'm unsure about any side effects so I'll leave this up to you...

### Steps To Reproduce

examples explained in description

PR #47026, opened 2023-05-10
Ensure cached class cast is actually in expected shape

Propaganistas:
This is a (rudimentary?) take on https://github.com/laravel/framework/issues/46744

---

When defining attribute casts, the framework is able to handle value objects and simple lists (array, collection) fluently.
Imagine now the powerful combination: a cast to handle **a collection of value objects**.
A custom cast class seems like a good fit for this. Here's a slim implementation:

```php
class ValueObjectCollectionCaster implements CastsAttributes
{
    {
        return (new Collection($json_decode($attributes[$key], true)))->map(function ($item) {
            return new ValueObject($item);
        });
    }

    {
        return json_encode($value);
    }
}
```

Because the `set()` method is directly encoding the value into a JSON string, I should be able to pass any encodable value to it:

```php
$model->items = ['foo', 'bar'];
$model->items = [new ValueObject('foo'), new ValueObject('bar')];

$model->items = collect(['foo', 'bar']);
$model->items = collect([new ValueObject('foo'), new ValueObject('bar')]);
```

But these yield very different results. I figured that the **class cast cache** is causing trouble. More specifically this part:

https://github.com/laravel/framework/blob/ddbbb2b50388721fe63312bb4469cae13163fd36/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php#L1129-L1133

It seems that, upon setting the value, the original value gets cached if it's an object. The array example behaves fine. It never makes it into the cache. But the collection gets cached as-is without its items being casted to value objects. Subsequent retrievals of the attribute then results in an uncasted value.

### DIFF
--- a/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
+++ b/src/Illuminate/Database/Eloquent/Concerns/HasAttributes.php
@@ -1137,7 +1137,7 @@ trait HasAttributes
         if ($caster instanceof CastsInboundAttributes || ! is_object($value)) {
             unset($this->classCastCache[$key]);
         } else {
-            $this->classCastCache[$key] = $value;
+            $this->classCastCache[$key] = $this->normalizeCacheableClassCastableAttribute($key, $value, $caster);
         }
     }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
I personally think this is just something that should be fixed by the developer in your own application. You aren't setting the casted attribute to the proper value. You should set it with an object that matches the same type of "shape" returned by your getter.

itsgoingd:
The documentation currently only mentions that the values are cached when "resolved". I'd suggest explicitly documenting the behavior when setting values, as it's not very intuitive and is hard-to-debug.

